PR #16, opened 2022-03-31
 Dependabot security alert: Ant

jukzi:
Improper Handling of Length Parameter Inconsistency in Apache Ant
When reading a specially crafted TAR archive an Apache Ant build can be made to allocate large amounts of memory that finally leads to an out of memory error, even for small inputs. This can be used to disrupt builds using Apache Ant. Apache Ant prior to 1.9.16 and 1.10.11 were affected.


akurtakov:
Thanks! Only that proves to me the trouble moving is worth it - a potential security threat in build tools auto discovered and patch prepared by non-releng person!

PR #26, opened 2022-04-06
Bug 579470 - Add ToolItem background and foreground color properties

nnemkin:
This patch adds new API for recoloring ToolItems. SWT version is bumped to 3.120.0.

Recolored toolbar buttons look decent, but there is some space for discussion. In particular, GTK drop-down buttons are made of two parts (the main button and the arrow button) and the patch recolors them individually. Also, separators ignore color properties, but in theory it's possible for them to respect foreground color.

Snippet153 has been extended for testing this patch. I've tested it on the latest stable versions of macOS, Windows and GNOME (GTK 3.24).

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT PI/win32/library/os.c
+++ b/bundles/org.eclipse.swt/Eclipse SWT PI/win32/library/os.c
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2021 IBM Corporation and others.
+ * Copyright (c) 2000, 2022 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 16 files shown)


vogella:
Thanks. Please merge after M1 is released so that our client can test this.

joel-majano:
In GTK4, testing Snippet 153 when hovering over the ToolItems the background/foreground doesn't change. Looks to be an issue with the MouseMove event not being formed properly, not related to patch. 

updateStyle() uses GTK3 functions, so any time that function gets called the snippet crashes, like when pressing the toggle colours button.

vogella:
@joel-majano can you create a new issue for your discovery?

vogella:
@nnemkin please create a N&N entry.

niraj-modi:
This commit caused the Eclipse IBuild to fail:
https://www.eclipse.org/lists/platform-releng-dev/msg38316.htmlWe also need to update the SWT binaries bundle-version as well

Veroni23:
Downloaded the IBuild today and tried the new API: setBackground/setForeground set colors correctly, but there are no hover, pressed, checked effects. The behavior is different from that of Button.

nnemkin:
@joel-majano thanks for testing. Adding GTK4 support should be pretty easy, but GTK4 it isn't an officially supported platform yet. Is there some king of plan or roadmap for the GTK4 port?

@vogella I'll add N&N.

@niraj-modi sorry, my fault. I was going to bump fragment versions, but the patch got merged.

@Veroni23 thanks for testing. Since this patch is for [bug 579470](https://bugs.eclipse.org/bugs/show_bug.cgi?id=579470), custom state colors are supposed to be provided via CSS. (As far as I understand, the CSS machinery isn't implemented yet.)
I do think it's a good idea to match Button behavior (and it's easy to implement), but automatically derived state colors will conflict with CSS imposed colors. I.e. if we use a darkening effect for the pressed state color, then the color assigned via CSS for that state will get unconditionally darkened too.

Issue #37, opened 2022-04-11
Bug 579626 - RTL text is not rendered correctly on TextLayout


niraj-modi:
The changes made on https://bugs.eclipse.org/bugs/show_bug.cgi?id=565526 caused RTL texts to be rendered incorrectly on TextLayout.

Phillipus:
Thanks for looking at this, Niraj.

I tested commit https://github.com/eclipse-platform/eclipse.platform.swt/commit/521940d173d5dde82c1c7b397a6947b3f08d1e1f but it doesn't seem to have fixed the issue.

`scriptState.fArabicNumContext)` is always returning false in my test cases.

Phillipus:
If I run this snippet the text is wrong direction:

https://bugzillaattachments.eclipsecontent.org/bugs/attachment.cgi?id=288411

Phillipus:
Here's simple test case:

```
import org.eclipse.swt.SWT;
import org.eclipse.swt.custom.StyledText;
import org.eclipse.swt.layout.GridLayout;
import org.eclipse.swt.widgets.Display;
import org.eclipse.swt.widgets.Shell;
import org.eclipse.swt.widgets.Text;

class ArabicTest {

    private static final String ARABIC_TEXT = "جارِ التحميل";

    public static void main(String[] args) {
        final Display display = new Display();
        Shell shell = new Shell(display);
        shell.setLayout(new GridLayout());

        // This is rendered correctly
        Text text = new Text(shell, SWT.BORDER);
        text.setText(ARABIC_TEXT);

        // This is not rendered correctly
        StyledText styled = new StyledText(shell, SWT.BORDER | SWT.MULTI);
        styled.setText(ARABIC_TEXT);

        shell.pack();
        shell.open();

        while(!shell.isDisposed()) {
            if(!display.readAndDispatch()) display.sleep();
        }
    }
}
```

Phillipus:
Hi Niraj,

I tested the PR and it has fixed the problem both in the test snippet and in our RCP app, Archi. Thanks!

niraj-modi:
> Hi Niraj,
> 
> I tested the PR and it has fixed the problem both in the test snippet and in our RCP app, Archi. Thanks!

Thanks Phil for the test/confirmation, closing now.

bbdalai:
Hi @niraj-modi ,

[...]

PR #39, opened 2022-04-11
Bug 579626 part2- RTL text is not rendered correctly on TextLayout #37

niraj-modi:
- Updated check to correctly/consistently identify RIGHT_TO_LEFT
scenario.

Change-Id: Ic1aafc83f0fe68d52cf4c45790cb1fea5c2ecf46


Phillipus:
Hi Niraj,

I tested this PR and it has fixed the problem both in the test snippet and in our RCP app, Archi. Thanks!

Phillipus:
> `resolveTextDirection()` will work only if the TextLayout#setTextDirection(SWT.AUTO_TEXT_DIRECTION) is set.
> That's how I tested my first fix patch, but that's not the case always hence required to call `BidiUtil.resolveTextDirection(text)` instead of `resolveTextDirection()`

Yes, I just tested that and I see what you mean.

Issue #44, opened 2022-04-12
Mixed RTL text(with LTR) is not rendered correctly in StyledText widget


niraj-modi:
Raised this issue to capture bug highlighted in below test Snippet seen StyledText widget:
https://github.com/eclipse-platform/eclipse.platform.swt/issues/37#issuecomment-1096440107

Phillipus:
Thanks for opening this issue, Niraj.

To summarise:

1. Changes made on https://bugs.eclipse.org/bugs/show_bug.cgi?id=565526 caused RTL texts to be rendered incorrectly on TextLayout
2. This has been partially fixed in #37 
3. However there is still a problem with mixed Arabic and non-Arabic text as in `System.out.println("جارِ التحميل");`
4. This is a Windows-only issue. Mac and Linux display RTL text correctly in all cases

Phillipus:
(Excuse my meddling in this, I'm just throwing out some ideas)

I got rid of the check for RTL (`if (BidiUtil.resolveTextDirection(text) != SWT.RIGHT_TO_LEFT)`) and made the for loop like this:
```
for (int i = 0, latestNeutralIndex = -2, latestUnicodeIndex = -2; i < length; i++) {
	char c = chars[i];

	// Check single char for RTL
	if(BidiUtil.resolveTextDirection(Character.toString(c)) == SWT.RIGHT_TO_LEFT) {
		continue;
	}

	if (c >= ' ' && c <= '~' && !Character.isAlphabetic(c)) {
		latestNeutralIndex = i;
	} else if (c > 255) {
		latestUnicodeIndex = i;
	} else {
		continue;
	}

	// If the latest neutral and unicode characters are adjacent
	if (Math.abs(latestNeutralIndex - latestUnicodeIndex) == 1) {
		// Change the neutral into a non-neutral alphabet character
		chars[latestNeutralIndex] = 'A';
	}
}
```

Phillipus:
@niraj-modi I tested PR #45 and that seems to do the trick! But do you think changing `BidiUtil#resolveTextDirection()` might cause problems elsewhere?

niraj-modi:
> @niraj-modi I tested PR #45 and that seems to do the trick! But do you think changing `BidiUtil#resolveTextDirection()` might cause problems elsewhere?

PR #45 fix is inline with JavaDoc of `BidiUtil#resolveTextDirection()`:
- this patch just improves the behavior for complete text String as opposed to starting character.

[...]

PR #45, opened 2022-04-12
Mixed RTL text(with LTR) is not rendered correctly in StyledText widget

niraj-modi:
#44
- Fixed BidiUtil#resolveTextDirection() to work for complete string(not
just first character of the string)

Change-Id: I62ca926cee5cec294f5eacdbaee703f8e50e3c68


bbdalai:
Verified with latest patch. Mix RTL text working as expected.

Issue #49, opened 2022-04-14
[GTK4] Snippet 50 swt_fixed_remove error when closing dialog


joel-majano:
This error is caused by calling swt_fixed_remove with fixedHandle == 0. The fix is to check if the widget to be destroyed is a GtkWindow, and then calling the appropriate gtk_window_destroy() call.

PR #50, opened 2022-04-14
[GTK4] Call gtk_window_destroy if widget to be destroyed is a GtkWindow

joel-majano:
Previously, when the dialog was closed in Snippet 50, an swt_fixed_remove assertion error was logged in the console (issue #49). With this patch that no longer happens. 

- added a call to gtk_window_destroy which happens only if GTK_IS_WINDOW is true.

Tested using Snippet 50.


sravanlakkimsetti (approved):

joel-majano:
Sravan, thanks for the merge!

Issue #54, opened 2022-04-19
Bump min GTK 3 version to 3.22.x


akurtakov:
All our target platforms are at Gtk 3.22 https://www.eclipse.org/projects/project-plan.php?planurl=http://www.eclipse.org/eclipse/development/plans/eclipse_project_plan_4_24.xml#target_environments%20plan so no issue from that side.
Why? In order to start building GTK 4 port officially a builder with Gtk 4.x is needed. CentOS 7.x and 8.x don't have GTK 4 so we have to jump on CentOS Stream 9 which has Gtk 3.24 and newer glib/glibc. Moving to 3.22 now is one step at a time approach to ease migrations by not requiring too big jumps.

PR #56, opened 2022-04-19
Bump min GTK 3 version to 3.22.x

akurtakov:
All target environments, build and test machine are at 3.22 level
already.
This is also a preparation for later move to 3.24 as a prereq for Gtk
4.x binaries being provided.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/54


sravanlakkimsetti (approved):

sravanlakkimsetti:
Probably launcher change will be required in https://github.com/eclipse-equinox/equinox.framework/blob/master/features/org.eclipse.equinox.executable.feature/library/gtk/eclipseGtkInit.c

akurtakov:
> Probably launcher change will be required in https://github.com/eclipse-equinox/equinox.framework/blob/master/features/org.eclipse.equinox.executable.feature/library/gtk/eclipseGtkInit.c

I'll handle it.

Issue #64, opened 2022-04-21
Integration with java.util.concurrent framework


laeubi:
Concurrency could be challenging, especially when it comes to UI and responsiveness. And even more with the single-threading model of SWT (and other UI frameworks as well) as one easily could block the UI then.

Java for several year contains some where useful tool with the [concurrency framework](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/package-summary.html) but it is quite hard for users to interact with that as SWT does not offer some basic primitives (most notable [Executors](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/Executor.html) and [CompletableFuture](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/CompletableFuture.html).

My proposal is to add some basic glues to allow a seamless integration of SWT with then concurrency framework so allow users more save, modern and easy to use calls into the SWT framework that hopefully leads to less blocking calls as it becomes incredible easy to use.

I have prototypesd such components in the past two years in an RCP application and will contribute some patches to integrate this directly in SWT.

akurtakov:
@laeubi as @vogella already approved but we are close to M3, feel free to merge it right after branching 4.25 off so there is whole release cycle to test its usage.

laeubi:
@akurtakov I think you better add this as an approving review to the PR itself, is there a scheduled date already for when " branching 4.25 off " will happen? Then mentioning the date would help too :-)

akurtakov:
Sep 9th (at latest but probably couple days earlier) opening 4.26 stream for development should be complete.

PR #67, opened 2022-04-21
Let Display implement java.util.concurrent.Executor

laeubi:
See https://github.com/eclipse-platform/eclipse.platform.swt/issues/64

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
@@ -621,15 +621,15 @@ public void asyncExec (Runnable runnable) {
  * <li>If the calling thread is the user-interface thread of this display it is
  * executed immediately and the method returns after the command has run, any
  * runtime exception thrown will be rethrown at the caller</li>
- * <li>In all other cases the command is asynchronously execute as with the
+ * <li>In all other cases the the <code>run()</code> method of the runnable is asynchronously executed as with the
  * method {@link Display#asyncExec(Runnable)} at the next reasonable
  * opportunity. The caller of this method continues to run in parallel, and is
  * not notified when the runnable has completed.</li>
  * </ul>
  * <p>
- * This can be used in cases where one want to execute some piece of code should
+ * This can be used in cases where one want to execute some piece of code that should
  * be guaranteed to run in the user-interface thread regardless of the current
- * context thread.
+ * thread.
  * </p>
  *
  * <p>
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
@@ -621,15 +621,16 @@ public void asyncExec (Runnable runnable) {
  * <li>If the calling thread is the user-interface thread of this display it is
  * executed immediately and the method returns after the command has run, any
  * runtime exception thrown will be rethrown at the caller</li>
- * <li>In all other cases the the <code>run()</code> method of the runnable is asynchronously executed as with the
- * method {@link Display#asyncExec(Runnable)} at the next reasonable
- * opportunity. The caller of this method continues to run in parallel, and is
- * not notified when the runnable has completed.</li>
+ * <li>In all other cases the the <code>run()</code> method of the runnable is
+ * asynchronously executed as with the method
+ * {@link Display#asyncExec(Runnable)} at the next reasonable opportunity. The
+ * caller of this method continues to run in parallel, and is not notified when
+ * the runnable has completed.</li>
  * </ul>
  * <p>
- * This can be used in cases where one want to execute some piece of code that should
- * be guaranteed to run in the user-interface thread regardless of the current
- * thread.
+ * This can be used in cases where one want to execute some piece of code that
+ * should be guaranteed to run in the user-interface thread regardless of the
+ * current thread.
  * </p>
  *
  * <p>
@@ -654,14 +655,20 @@ public void asyncExec (Runnable runnable) {
  *
  * @param runnable the runnable to execute in the user-interface thread, never
  *                 <code>null</code>
+ * @throws RejectedExecutionException if this task cannot be accepted for
+ *                                    execution
+ * @throws NullPointerException       if runnable is null
  */
 @Override
-public void execute(Runnable command) {
-	Objects.requireNonNull(command);
+public void execute(Runnable runnable) {
+	Objects.requireNonNull(runnable);
+	if (isDisposed()) {
+		throw new RejectedExecutionException(new SWTException (SWT.ERROR_WIDGET_DISPOSED, null));
+	}
 	if (thread == Thread.currentThread()) {
-		command.run();
+		runnable.run();
 	} else {
-		asyncExec(command);
+		asyncExec(runnable);
 	}
 }
 
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
@@ -666,7 +666,13 @@ public void execute(Runnable runnable) {
 		throw new RejectedExecutionException(new SWTException (SWT.ERROR_WIDGET_DISPOSED, null));
 	}
 	if (thread == Thread.currentThread()) {
-		runnable.run();
+		try {
+			runnable.run ();
+		} catch (RuntimeException exception) {
+			getRuntimeExceptionHandler ().accept (exception);
+		} catch (Error error) {
+			getErrorHandler ().accept (error);
+		}
 	} else {
 		asyncExec(runnable);
 	}
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2019 IBM Corporation and others.
+ * Copyright (c) 2000, 2022 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -10,11 +10,13 @@
  *
  * Contributors:
  *     IBM Corporation - initial API and implementation
+ *     Christoph Läubrich - Issue #64 - Integration with java.util.concurrent framework
  *******************************************************************************/
 package org.eclipse.swt.widgets;
 
 import java.lang.Runtime.*;
 import java.util.*;
+import java.util.concurrent.*;
 import java.util.function.*;
 
 import org.eclipse.swt.*;
@@ -102,7 +104,7 @@ import org.eclipse.swt.internal.cocoa.*;
  * @noextend This class is not intended to be subclassed by clients.
  */
 @SuppressWarnings({"rawtypes", "unchecked"})
-public class Display extends Device {
+public class Display extends Device implements Executor {
 
 	static byte[] types = {'*','\0'};
 	static int size = C.PTR_SIZEOF, align = C.PTR_SIZEOF == 4 ? 2 : 3;
@@ -610,6 +612,56 @@ public void asyncExec (Runnable runnable) {
 	synchronized (Device.class) {
 		if (isDisposed ()) error (SWT.ERROR_DEVICE_DISPOSED);
 		synchronizer.asyncExec (runnable);
+	}
+}
+
+/**
+ * Executes the given runnable in the user-interface thread of this Display.
+ * <ul>
+ * <li>If the calling thread is the user-interface thread of this display it is
+ * executed immediately and the method returns after the command has run, any
+ * runtime exception thrown will be rethrown at the caller</li>
+ * <li>In all other cases the command is asynchronously execute as with the
+ * method {@link Display#asyncExec(Runnable)} at the next reasonable
+ * opportunity. The caller of this method continues to run in parallel, and is
+ * not notified when the runnable has completed.</li>
+ * </ul>
+ * <p>
+ * This can be used in cases where one want to execute some piece of code should
+ * be guaranteed to run in the user-interface thread regardless of the current
+ * context thread.
+ * </p>
+ *
+ * <p>
+ * Note that at the time the runnable is invoked, widgets that have the receiver
+ * as their display may have been disposed. Therefore, it is advised to check
+ * for this case inside the runnable before accessing the widget.
+ * </p>
+ * <p>
+ * Also note that most users would find it more convenient to use
+ * {@link Display#executor()} instead of acquiring a Display first and then use
+ * the Display as an Executor, for example when using
+ * <code>CompletableFuture</code>:
+ *
+ * <pre>
+ * CompletableFuture.supplyAsync(() -&gt; {
+ * 	// do something in the background
+ * }).thenAcceptAsync(computedValue -&gt; {
+ * 	// update the UI with the computed value
+ * }, Display.executor());
+ * </pre>
+ * </p>
+ *
+ * @param runnable the runnable to execute in the user-interface thread, never
+ *                 <code>null</code>
+ */
+@Override
+public void execute(Runnable command) {
+	Objects.requireNonNull(command);
+	if (thread == Thread.currentThread()) {
+		command.run();
+	} else {
+		asyncExec(command);
 	}
 }
 
@@ -1525,6 +1577,19 @@ public static Display getDefault () {
 		if (Default == null) Default = new Display ();
 		return Default;
 	}
+}
+
+/**
+ * Returns an executor that will execute any submitted runnable in the user-interface thread, regardless of the calling thread
+ * @return an {@link Executor} that executes in the user-interface thread.
+ * @since 3.120
+ */
+public static Executor executor() {
+	Display current = Display.getCurrent();
+	if (current != null) {
+		return current;
+	}
+	return Display.getDefault();
 }
 
 /**
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2021 IBM Corporation and others.
+ * Copyright (c) 2000, 2022 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -10,6 +10,7 @@
  *
  * Contributors:
  *     IBM Corporation - initial API and implementation
+ *     Christoph Läubrich - Issue #64 - Integration with java.util.concurrent framework
  *******************************************************************************/
 package org.eclipse.swt.widgets;
 
@@ -19,6 +20,7 @@ import java.lang.reflect.*;
 import java.net.*;
 import java.util.*;
 import java.util.Map.*;
+import java.util.concurrent.*;
 import java.util.function.*;
 import java.util.regex.*;
 import java.util.regex.Pattern;
@@ -111,7 +113,7 @@ import org.eclipse.swt.internal.gtk4.*;
  * @see <a href="http://www.eclipse.org/swt/">Sample code and further information</a>
  * @noextend This class is not intended to be subclassed by clients.
  */
-public class Display extends Device {
+public class Display extends Device implements Executor {
 
 	static boolean strictChecks = System.getProperty("org.eclipse.swt.internal.gtk.enableStrictChecks") != null;
 
@@ -926,6 +928,56 @@ public void asyncExec (Runnable runnable) {
 			}
 		}
 		synchronizer.asyncExec (runnable);
+	}
+}
+
+/**
+ * Executes the given runnable in the user-interface thread of this Display.
+ * <ul>
+ * <li>If the calling thread is the user-interface thread of this display it is
+ * executed immediately and the method returns after the command has run, any
+ * runtime exception thrown will be rethrown at the caller</li>
+ * <li>In all other cases the command is asynchronously execute as with the
+ * method {@link Display#asyncExec(Runnable)} at the next reasonable
+ * opportunity. The caller of this method continues to run in parallel, and is
+ * not notified when the runnable has completed.</li>
+ * </ul>
+ * <p>
+ * This can be used in cases where one want to execute some piece of code should
+ * be guaranteed to run in the user-interface thread regardless of the current
+ * context thread.
+ * </p>
+ *
+ * <p>
+ * Note that at the time the runnable is invoked, widgets that have the receiver
+ * as their display may have been disposed. Therefore, it is advised to check
+ * for this case inside the runnable before accessing the widget.
+ * </p>
+ * <p>
+ * Also note that most users would find it more convenient to use
+ * {@link Display#executor()} instead of acquiring a Display first and then use
+ * the Display as an Executor, for example when using
+ * <code>CompletableFuture</code>:
+ *
+ * <pre>
+ * CompletableFuture.supplyAsync(() -&gt; {
+ * 	// do something in the background
+ * }).thenAcceptAsync(computedValue -&gt; {
+ * 	// update the UI with the computed value
+ * }, Display.executor());
+ * </pre>
+ * </p>
+ *
+ * @param runnable the runnable to execute in the user-interface thread, never
+ *                 <code>null</code>
+ */
+@Override
+public void execute(Runnable runnable) {
+	Objects.requireNonNull(runnable);
+	if (thread == Thread.currentThread()) {
+		runnable.run();
+	} else {
+		asyncExec(runnable);
 	}
 }
 
@@ -2347,6 +2399,19 @@ public static Display getDefault () {
 		if (Default == null) Default = new Display ();
 		return Default;
 	}
+}
+
+/**
+ * Returns an executor that will execute any submitted runnable in the user-interface thread, regardless of the calling thread
+ * @return an {@link Executor} that executes in the user-interface thread.
+ * @since 3.120
+ */
+public static Executor executor() {
+	Display current = Display.getCurrent();
+	if (current != null) {
+		return current;
+	}
+	return Display.getDefault();
 }
 
 static boolean isValidClass (Class<?> clazz) {
@@ -6198,4 +6263,5 @@ static int _getDeviceZoom (long monitor_num) {
 	dpi = dpi * scale;
 	return DPIUtil.mapDPIToZoom (dpi);
 }
+
 }
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
@@ -937,15 +937,16 @@ public void asyncExec (Runnable runnable) {
  * <li>If the calling thread is the user-interface thread of this display it is
  * executed immediately and the method returns after the command has run, any
  * runtime exception thrown will be rethrown at the caller</li>
- * <li>In all other cases the command is asynchronously execute as with the
- * method {@link Display#asyncExec(Runnable)} at the next reasonable
- * opportunity. The caller of this method continues to run in parallel, and is
- * not notified when the runnable has completed.</li>
+ * <li>In all other cases the the <code>run()</code> method of the runnable is
+ * asynchronously executed as with the method
+ * {@link Display#asyncExec(Runnable)} at the next reasonable opportunity. The
+ * caller of this method continues to run in parallel, and is not notified when
+ * the runnable has completed.</li>
  * </ul>
  * <p>
- * This can be used in cases where one want to execute some piece of code should
- * be guaranteed to run in the user-interface thread regardless of the current
- * context thread.
+ * This can be used in cases where one want to execute some piece of code that
+ * should be guaranteed to run in the user-interface thread regardless of the
+ * current thread.
  * </p>
  *
  * <p>
@@ -970,10 +971,16 @@ public void asyncExec (Runnable runnable) {
  *
  * @param runnable the runnable to execute in the user-interface thread, never
  *                 <code>null</code>
+ * @throws RejectedExecutionException if this task cannot be accepted for
+ *                                    execution
+ * @throws NullPointerException       if runnable is null
  */
 @Override
 public void execute(Runnable runnable) {
 	Objects.requireNonNull(runnable);
+	if (isDisposed()) {
+		throw new RejectedExecutionException(new SWTException (SWT.ERROR_WIDGET_DISPOSED, null));
+	}
 	if (thread == Thread.currentThread()) {
 		runnable.run();
 	} else {
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
@@ -982,7 +982,13 @@ public void execute(Runnable runnable) {
 		throw new RejectedExecutionException(new SWTException (SWT.ERROR_WIDGET_DISPOSED, null));
 	}
 	if (thread == Thread.currentThread()) {
-		runnable.run();
+		try {
+			runnable.run ();
+		} catch (RuntimeException exception) {
+			getRuntimeExceptionHandler ().accept (exception);
+		} catch (Error error) {
+			getErrorHandler ().accept (error);
+		}
 	} else {
 		asyncExec(runnable);
 	}
--- a/examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet380.java
+++ b/examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet380.java
@@ -1,0 +1,63 @@
+/*******************************************************************************
+ * Copyright (c) 2022 Christoph Läubrich and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Christoph Läubrich - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.swt.snippets;
+
+import java.text.*;
+import java.util.concurrent.*;
+
+import org.eclipse.swt.*;
+import org.eclipse.swt.events.*;
+import org.eclipse.swt.layout.*;
+import org.eclipse.swt.widgets.*;
+
+public class Snippet380 {
+	public static void main(String[] args) {
+		Display display = new Display();
+		Shell shell = new Shell(display);
+		shell.setSize(400, 100);
+		shell.setLayout(new BorderLayout());
+		shell.setText("Snippet 380 (Executor)");
+		Button button = new Button(shell, SWT.PUSH);
+		button.setLayoutData(new BorderData(SWT.BOTTOM));
+		Label label = new Label(shell, SWT.NONE);
+		label.setLayoutData(new BorderData(SWT.CENTER));
+		button.setText("Calculate");
+		button.addSelectionListener(SelectionListener.widgetSelectedAdapter(e -> {
+			button.setEnabled(false);
+			button.setText("Calculating...");
+			CompletableFuture.supplyAsync(() -> {
+				double value = 0;
+				for (int i = 0; i < 10; i++) {
+					try {
+						TimeUnit.SECONDS.sleep(1);
+					} catch (InterruptedException e1) {
+						break;
+					}
+					value += Math.random();
+				}
+				return value;
+			}).thenApply(result -> NumberFormat.getInstance().format(result))//
+					.thenApply(formatedNumber -> String.format("The result is %s", formatedNumber))//
+					.thenAcceptAsync(label::setText, Display.executor())//
+					.whenCompleteAsync((x, y) -> button.setEnabled(true), Display.executor());
+		}));
+		shell.open();
+		while (!shell.isDisposed()) {
+			if (!display.readAndDispatch()) {
+				display.sleep();
+			}
+		}
+		display.dispose();
+	}
+}
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2020 IBM Corporation and others.
+ * Copyright (c) 2000, 2022 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -10,11 +10,13 @@
  *
  * Contributors:
  *     IBM Corporation - initial API and implementation
+ *     Christoph Läubrich - Issue #64 - Integration with java.util.concurrent framework
  *******************************************************************************/
 package org.eclipse.swt.widgets;
 
 import java.net.*;
 import java.util.*;
+import java.util.concurrent.*;
 import java.util.function.*;
 
 import org.eclipse.swt.*;
@@ -102,7 +104,7 @@ import org.eclipse.swt.internal.win32.*;
  * @see <a href="http://www.eclipse.org/swt/">Sample code and further information</a>
  * @noextend This class is not intended to be subclassed by clients.
  */
-public class Display extends Device {
+public class Display extends Device implements Executor {
 
 	/**
 	 * the handle to the OS message queue
@@ -792,6 +794,56 @@ public void asyncExec (Runnable runnable) {
 	synchronized (Device.class) {
 		if (isDisposed ()) error (SWT.ERROR_DEVICE_DISPOSED);
 		synchronizer.asyncExec (runnable);
+	}
+}
+
+/**
+ * Executes the given runnable in the user-interface thread of this Display.
+ * <ul>
+ * <li>If the calling thread is the user-interface thread of this display it is
+ * executed immediately and the method returns after the command has run, any
+ * runtime exception thrown will be rethrown at the caller</li>
+ * <li>In all other cases the command is asynchronously execute as with the
+ * method {@link Display#asyncExec(Runnable)} at the next reasonable
+ * opportunity. The caller of this method continues to run in parallel, and is
+ * not notified when the runnable has completed.</li>
+ * </ul>
+ * <p>
+ * This can be used in cases where one want to execute some piece of code should
+ * be guaranteed to run in the user-interface thread regardless of the current
+ * context thread.
+ * </p>
+ *
+ * <p>
+ * Note that at the time the runnable is invoked, widgets that have the receiver
+ * as their display may have been disposed. Therefore, it is advised to check
+ * for this case inside the runnable before accessing the widget.
+ * </p>
+ * <p>
+ * Also note that most users would find it more convenient to use
+ * {@link Display#executor()} instead of acquiring a Display first and then use
+ * the Display as an Executor, for example when using
+ * <code>CompletableFuture</code>:
+ *
+ * <pre>
+ * CompletableFuture.supplyAsync(() -&gt; {
+ * 	// do something in the background
+ * }).thenAcceptAsync(computedValue -&gt; {
+ * 	// update the UI with the computed value
+ * }, Display.executor());
+ * </pre>
+ * </p>
+ *
+ * @param runnable the runnable to execute in the user-interface thread, never
+ *                 <code>null</code>
+ */
+@Override
+public void execute(Runnable command) {
+	Objects.requireNonNull(command);
+	if (thread == Thread.currentThread()) {
+		command.run();
+	} else {
+		asyncExec(command);
 	}
 }
 
@@ -1684,6 +1736,19 @@ public static Display getDefault () {
 		if (Default == null) Default = new Display ();
 		return Default;
 	}
+}
+
+/**
+ * Returns an executor that will execute any submitted runnable in the user-interface thread, regardless of the calling thread
+ * @return an {@link Executor} that executes in the user-interface thread.
+ * @since 3.120
+ */
+public static Executor executor() {
+	Display current = Display.getCurrent();
+	if (current != null) {
+		return current;
+	}
+	return Display.getDefault();
 }
 
 /**
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
@@ -848,7 +848,13 @@ public void execute(Runnable runnable) {
 		throw new RejectedExecutionException(new SWTException (SWT.ERROR_WIDGET_DISPOSED, null));
 	}
 	if (thread == Thread.currentThread()) {
-		runnable.run();
+		try {
+			runnable.run ();
+		} catch (RuntimeException exception) {
+			getRuntimeExceptionHandler ().accept (exception);
+		} catch (Error error) {
+			getErrorHandler ().accept (error);
+		}
 	} else {
 		asyncExec(runnable);
 	}
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
@@ -803,15 +803,16 @@ public void asyncExec (Runnable runnable) {
  * <li>If the calling thread is the user-interface thread of this display it is
  * executed immediately and the method returns after the command has run, any
  * runtime exception thrown will be rethrown at the caller</li>
- * <li>In all other cases the command is asynchronously execute as with the
- * method {@link Display#asyncExec(Runnable)} at the next reasonable
- * opportunity. The caller of this method continues to run in parallel, and is
- * not notified when the runnable has completed.</li>
+ * <li>In all other cases the the <code>run()</code> method of the runnable is
+ * asynchronously executed as with the method
+ * {@link Display#asyncExec(Runnable)} at the next reasonable opportunity. The
+ * caller of this method continues to run in parallel, and is not notified when
+ * the runnable has completed.</li>
  * </ul>
  * <p>
- * This can be used in cases where one want to execute some piece of code should
- * be guaranteed to run in the user-interface thread regardless of the current
- * context thread.
+ * This can be used in cases where one want to execute some piece of code that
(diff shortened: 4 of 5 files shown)


bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
```diff
@@ -613,6 +615,16 @@ public void asyncExec (Runnable runnable) {
 	}
 }
 
+@Override
+public void execute(Runnable command) {
```
iloveeclipse, line 619: Please add comprehensive javadoc. The one provided by super class is misleading, because **any** execution in SWT happens on UI thread and nowhere else.
Please also provide some simple example of the possible use of this method *in the javadoc* here, so developers have an idea what it is good for, with a **bold** warning that no guarantees will be given if this returns synchronously or not, because it depends on the thread in which the current code is running.

iloveeclipse (changes requested):
I'm personally OK with that API, but please add proper javadoc.

Please do not submit, until some few other SWT developers give feedback. That kind of API may always trigger discussions.

iloveeclipse:
@jukzi : you've recently added `org.eclipse.swt.widgets.Display.syncCall(SwtCallable<T, E>)`, which qualifies you as a reviewer for this PR :-)

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet380.java
```diff
@@ -0,0 +1,63 @@
...
+					} catch (InterruptedException e1) {
+						break;
+					}
+					value+=Math.random();
+				}
+				return value;
+			}).thenApply(result->NumberFormat.getInstance().format(result))//
+			.thenApply(formatedNumber->String.format("The result is %s", formatedNumber))//
+			.thenAcceptAsync(label::setText, Display.executor())//
+			.whenCompleteAsync((x,y)->button.setEnabled(true), Display.executor());
```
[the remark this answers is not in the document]
laeubi: Sadly I can't came up with a more nice but still simple example here, sorry for that :-)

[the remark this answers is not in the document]
laeubi: Sadly eclipse does not support nice lamda formatting (or has something changed?) and I'm bad with manual formatting code :-D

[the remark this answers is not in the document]
iloveeclipse: I meant `()->{` -> `() -> {`

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
```diff
@@ -2349,6 +2361,19 @@ public static Display getDefault () {
...
+ * Returns an executor that will execute any submitted runnable in the user-interface thread, regardless of the calling thread
+ * @return an {@link Executor} that executes in the user-interface thread.
+ * @since 3.120
+ */
+public static Executor executor() {
+	Display current = Display.getCurrent();
+	if (current != null) {
+		return current;
+	}
+	return Display.getDefault();
```
jukzi, line 2414: I do not like that "Device implements Executor" you could as well return any other class here

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
```diff
@@ -929,6 +931,16 @@ public void asyncExec (Runnable runnable) {
 	}
 }
 
+@Override
+public void execute(Runnable command) {
+	Objects.requireNonNull(command);
+	if (thread == Thread.currentThread()) {
+		command.run();
+	} else {
```
jukzi, line 979: What's the usecase to either run synchronous or asynchronous? You could also offer 3 Method. getSynchronousExecutor, getAsynchronousExecutor, "getRunSomewhenExecutor" ;-) - we could find a better name when we would know the usecase - if any.

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet380.java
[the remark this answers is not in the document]
laeubi: Done.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
laeubi: I have added the javadoc now, I hope it is suitable.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
```diff
@@ -2349,6 +2401,19 @@ public static Display getDefault () {
 	}
 }
 
+/**
+ * Returns an executor that will execute any submitted runnable in the user-interface thread, regardless of the calling thread
+ * @return an {@link Executor} that executes in the user-interface thread.
+ * @since 3.120
+ */
+public static Executor executor() {
```
jukzi, line 2409: Since this is static it could be better placed in a separate class to avoid further bloat of Display - and in a platform independent class.

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet380.java
```diff
@@ -0,0 +1,63 @@
...
+		shell.setText("Snippet 380 (Executor)");
+		Button button = new Button(shell, SWT.PUSH);
+		button.setLayoutData(new BorderData(SWT.BOTTOM));
+		Label label = new Label(shell, SWT.NONE);
+		label.setLayoutData(new BorderData(SWT.CENTER));
+		button.setText("Calculate");
+		button.addSelectionListener(SelectionListener.widgetSelectedAdapter(e -> {
+			button.setEnabled(false);
+			button.setText("Calculating...");
+			CompletableFuture.supplyAsync(() -> {
```
jukzi, line 39: looks like this example is guaranteed to run in Display thread. What is the usecase for a thread where we do not know where we are?

jukzi (changes requested):
I do like the idea, but i do not get the usecase where it is better then asyncExec.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
laeubi: Sure but this would complicate the code/class a lot here so what specifically you don't like in Display implement executor? For sure it could be a inner class as well...

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
laeubi: The idea is that one don't need to know, I always *never* care, e.g. if I want to update the text of a label in what Thread I'm currently running, I just want to update the damn label and as it even do not return a value I even don't care if it happens 'now' and I'm blocked for that while or it happen some microseconds later in another thread.

so this simply prevents me from the usual bloat:

```
if (Display.getCurrent() == null) {
do tomething with the ui
} else {
 Display.getDefault().asyncExec(()->{
  do the same as in the other branch

});
}
```

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
laeubi: People are more used to use the Display class and I don't like introduce more "helper" classes just to hide the SWT bloat.

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet380.java
laeubi: The future computes the value async and then you are not in the ui thread anymore. so when the computation is done you need to re-enter the UI thread (the button handler is no longer executing at that time).

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
jukzi: Theoretically there could be multiple displays. One could write
`Display.getDefault().execute( something)`
but i doubt if it matters that i selected a specific display to run something.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
jukzi: if it does not matter when it runs, you could just always run it async.
The usual problem then is that the code does not know if the widget is already is disposed, one needs to check that first... horrible boiler code. Would be cool to have such a method on a widget, and the code would only run if the widget is not disposed yet ... that would be an improvement!

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet380.java
jukzi: so you could also write something like 
`.thenAcceptAsync(t->display.runAsync(label.setText(t)))`
i do not see the benefit of the new executor.
Note that both could run into disposed which is wrong and a common error

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
laeubi: > Theoretically there could be multiple displays
Not only theoretical.

> Display.getDefault().execute( something)

Sure but why? make it more complicated and less flexible?

> but i doubt if it matters that i selected a specific display to run something.
At least the SWT API suggest that it could make a difference, if each display has its own thread with two displays I can run two Ui actions in parallel without blocking.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
laeubi: > if it does not matter when it runs, you could just always run it async.

This is simply not true, calling `async` is prone to deadloack if the code tries to wait for the "async" operation to complete, my inital implementation always used async and you can easily deadlock with that aproach. Beside that its a perfomance penalty involved scheduling a runnable async.

> The usual problem then is that the code does not know if the widget is already is disposed, one needs to check that first... horrible boiler code. Would be cool to have such a method on a widget, and the code would only run if the widget is not disposed yet ... that would be an improvement!

Sure, but the goal here is not to improve widget disposal but integrate better with the concurrency framework and there it is important that this distinction is made!

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet380.java
laeubi: This creates a new "object" on each call, beside that I need to know the Display beforehand, even though I have it in this case I sometimes do not. beside that is much more readable instead of repeating the same code sequence over and over again. Also as noted before display.runAsync is not always suitable as you can easily deadlock then.

> Note that both could run into disposed which is wrong and a common error

Well this is a simplified example. as described before, disposal handling is not my intention here.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
jukzi: Why would code wait for async? then it should take syncexec. the perfomance penalty is neglegible when you want to update a text - compared to drawing the text.

You could add that concurrency framework thing to widget - and would solve two problems with a single call.
Please, convince me that your change is worth by showing a benefit in a nontrivial example  with correct error handling against using the current API in the best manner. You could also use such example for the noteworthy new.
It should be like ... "till now you used this and now you can use that and it's x commands less" - win.

laeubi: > You could add that concurrency framework thing to widget - and would solve two problems with a single call.

This is not about calling a single method, I want to execute code in the UI thread and that's what executors are for, that the caller does not has to fiddle around what thread it is in, what display to acquire and so on.

> You could add that concurrency framework thing to widget

Having such API does not make sense at the widget level as it is not about calling a single widget method.

> Please, convince me that your change is worth by showing a benefit in a nontrivial example with correct error handling

Just because it is not useful for /you/ does not mean it is not useful for others. I have already explained why manual coding does not satisfies my needs and added an explanation to https://github.com/eclipse-platform/eclipse.platform.swt/issues/64 why this is useful, my goal is not to convince specifically **you** to use that APi feel free to ignore it if you think the good old way is the best of all choices ;-)

> "till now you used this and now you can use that and it's x commands less" - win.

Till now I need to use (+ any not show error/dispose handling not show):

```
if (Display.getCurrent() == null) {
do something with the ui
} else {
 Display.getDefault().asyncExec(()->{
  do the same as in the other branch

});
}
```
and now I can use (+ any not show error/dispose handling not show):
```
Display.executor(()->{
do something with the ui
});
```

I even can take a CompletableFuture and in input parameter in

```
void handleError(CompletableFuture<?> future) {
      future.handleAsync((result, error)->{
		if (error != null) {
			//show my favorite error dialog
			
		}
		return result;
	}, Display.executor());
}
```

without having to know what thread the caller uses, when in an unforeseeable future an error might occurs.

All these are "wins" for me and for sure I can just write another `SwtUtil` class (why org.eclipse.swt.widgets.Display.syncCall(SwtCallable<T, E>) was not put into such a static? It does not uses any internal state by the way?) and let other people do the same or just get mad about SWT being stuck at java 1.4 ...

merks:
Somehow the whole "implements Executor" things feels gratuitous and the same for a static method to get an/the Executor.   There's just no compelling need for either.   There's already a static method to get the Display and a execWhenever method on Display would suffice. But I also I don't see the use case for that and have never felt a need for that.   One needs to know if one has to check for disposed.  And, if I don't care when it gets done, I can just call asyncExec which will also do it whenever, and given I don't care, why isn't that good enough then?

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
jukzi: These unrealistic trivial example does not convince me. Neither that the executor is helpful nor that it needs to be in Display. 
As an alternative a label.executor(()->label.setText("new")) would be neat to avoid dispose checks.

laeubi: > As an alternative a label.executor(()->label.setText("new")) would be neat to avoid dispose checks.

I don't want to cover a single call, please open another ticket/enahncement request if that's important for you.
I want to execute an arbitrary code block in the ui thread...

From my POV your example is unrealistic and trivial as I rather would like call:
`label.setText("new"):`
and don't care:

1. if I'm on swt thread or not (if not spin an async call for me)
2. label is disposed (as it is not visible I dont care if my call would simply be a noop)

but thats completely out of scope of this cahnge and it is **not** what I'd like to archive. I'D like to archie modern programming primitives instead of all the UI frezze stuff that regular happen because people don't do all the checks.

Here is a simple example that shows that it **is** important to know the thread and why you cant simply always use async exec, and yes I know that in that example I could have know that it might block, but as noted above: I don't want to check all the time if I'm in or out the Ui thread ...

```
public class SnippetBlocking {

	public static void main(String[] args) throws InterruptedException, ExecutionException {
		shell.setText("Snippet 1");
		button.setText("Press me");
		Button check = new Button(shell, SWT.CHECK);
		check.setText("Do you like to block?");
		button.addSelectionListener(new SelectionListener() {

			@Override
			public void widgetSelected(SelectionEvent e) {
				Executor willBlock = display::asyncExec;
				Executor willNotBlock = r->{
					if (Display.getCurrent()==null) {
						Display.getDefault().asyncExec(r);
					} else {
						r.run();
					}

				};
				Executor ex;
				if (check.getSelection()) {
					ex = willBlock;
				} else {
					ex = willNotBlock;
				}
				CompletableFuture.completedFuture("Hello").thenApplyAsync(str -> str + " Stage 1", ex)
						.thenAccept(System.out::println).join();
			}

			@Override
			public void widgetDefaultSelected(SelectionEvent e) {

			}
		});
		shell.open();
		while (!shell.isDisposed()) {
			System.out.println("Dispatch...");
				display.sleep();
		}
		display.dispose();
	}
}
```

laeubi:
> here's already a static method to get the Display and a execWhenever method on Display would suffice.

I already have explained that this is wrong and repeating it multiple times does not make this true but shows that we need such a method as even people very familiar with SWT running into the same deadlock issues.

> I can just call asyncExec which will also do it whenever, and given I

See above... I can't call asyncExec *always* I can only call it if I'm not in an UI thread and I probably not always know that. Beside that, I always need to create a lamda for each and every call and as noticed by equinix recently lamdas are not 'free' as well as sceduling async operations are not free...

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
iloveeclipse: > calling async is prone to deadloack if the code tries to wait for the "async" operation to complete

Couldn't resist: sorry, it's join() that is prone to deadlock, not asyncExec(). 

I reviewed lot of code/bugs in UI area in last few years and **never** saw a deadlock caused by asyncExec(). join() and syncExec() are best friends for deadlocks, not asyncExec().

And after deadlocks via join() or syncExec() most prominent error is to **not** check if the widget is already disposed inside the asyncExec() block.

So from this point of view, providing an **async** API that executes a task on a widget **only** it it is not disposed yet would prevent lot of bugs and boiler plated code.

laeubi: > Couldn't resist: sorry, it's join() that is prone to deadlock, not asyncExec().

This is the problem here, but it was claimed that I can always use `asyncExec()` instead of my executor and that is simply not true, I have to know if I'm on the Ui thread or not, and that's what the executor does I just don't need to know. and as described in the issue description this is specifically to integrate with the concurrent API, and the executor is the most basic building block to archive that.

> I reviewed lot of code/bugs in UI area in last few years and **never** saw a deadlock caused by asyncExec().

But you also review a lot of code that deadlocks in certain circumstance and flapping UI test, and thats because we always try to be smart and "know" what happens when, but then there are cases where it is not obvious, we need to wait and alike. That's what one could avoid using the concurrent API but currently SWT requires me to do repetive check if I really like to use it or know the full context of a call chain.

> join() and syncExec() are best friends for deadlocks

True and that's why I'd like to add some easy to use integration with the concurrent API and Completable future, because that allows people to fulfill their goals without the need of ever calling those `syncExec()` and repeatably check for what thread they are in. To make things wore we now not only have `syncExec()` but also `syncCall()` ... still I can understand that people sometimes need to get the result of the operation and there we come to the end beginning again.

> So from this point of view, providing an **async** API that executes a task on a widget **only** it it is not disposed yet would prevent lot of bugs and boiler plated code.

The good think is, that once you have such basic building block you can easily write convenient API on top of it, given that you have a Widget class (I used the Text Widget here) you can now add a method safeCall:

```
public void safeRun(Consumer<? super Text> consumer) {
	getDisplay().execute(() ->{
		if (isDisposed()) {
			return;
		}
		consumer.accept(this);
	});
}
```

and call it like

```
Text text = ...
text.safeRun(t->t.setText("Will never be disposed"));
```

This has now two properties:

1. I can call it fro **any** thread, so no need to check or "know for sure" that you are on an UI thread, this is extremely useful if you implement API / Services where you simply don't know who calls then when as the APi contract simply don't know anything about SWT!
2. If the widget is disposed in the meanwhile, the code will be just a no-op

If you better like a static method use

merks:
> See above... I can't call asyncExec _always_ I can only call it if I'm not in an UI thread and I probably not always know that. Beside that, I always need to create a lamda for each and every call and as noticed by equinix recently lamdas are not 'free' as well as sceduling async operations are not free...

The assertion "I can only call it if I'm not on a UI thread" is simply not true, or perhaps its somehow stated out of context.  What is the basis for this statement?  The Javadoc states this:

> Causes the run() method of the runnable to be invoked by the user-interface thread at the next reasonable opportunity. The caller of this method continues to run in parallel, and is not notified when the runnable has completed. 

Clearly you can call this anywhere, any time, on any thread, including the UI thread, and the runnable will run at some time in the future.  If you say you don't care when it runs, then that certainly satisfies the "don't care" requirement.   But I assume you know this so there must be something unstated or I'm reading your statement out of context.

We have utilities like this in Oomph to deal with the common pattern of disposal checking:
```
  public static void asyncExec(final Control control, final Runnable runnable)
  {
    try
    {
      if (control.isDisposed())
      {
        return;
      }

      control.getDisplay().asyncExec(new Runnable()
      {
        @Override
        public void run()
        {
          if (!control.isDisposed())
          {
            runnable.run();
          }
        }
      });
    }
    catch (SWTException ex)
    {
      if (ex.code != SWT.ERROR_WIDGET_DISPOSED)
      {
        throw ex;
      }

      //$FALL-THROUGH$
    }
  }
```
But never have I had the need for a method that will just run if I'm on the UI thread and async if I'm not; I would still have to be careful about disposal checked.  I'm not saying no one needs that.  I see this exists for example:

org.eclipse.jdt.internal.ui.util.SWTUtil.execDirectOrAsyncIfNecessary(Supplier<Control>, Runnable)

There are a great many SWTUtil(s) and not how this one does all the dispose checking.

I  do question if enough people need that that it should be pushed into the API of Display.  It seems to me that's just one way of providing such a utility and perhaps a little in-your-face...

laeubi:
> The Javadoc states this:

Yes and the javadoc is obviously wrong, as there is only one UI thread, it is impossible that 

> the caller of this method continues to run in parallel

if the caller is in the UI thread, how should this work?

iloveeclipse:
> if the caller is in the UI thread, how should this work?

Come on, it's just a slight simplification, it should mean "The caller of this method **is not blocked** and continues to run ~in parallel~". Javadoc says clearly: "Causes the run() method of the runnable to be invoked by the user-interface thread at the next reasonable opportunity". 

if something is not clear, please propose an improved version, but `asyncExec()` works as specified.

laeubi:
> Come on, it's just a slight simplification,

It is not a simplification, it is wrong. caller not being blocked until the runnable *finished running* is something totally different to *running in parallel* with the caller. Beside this I didn't start this being pedantic and refering to the javadoc to state,

> Clearly you can call this anywhere, any time, on any thread, including the UI thread, and the runnable will run at some time in the future.

Just say I'm in the UI thread, I call `asyncExec` and entering a loop that do not drives the event queue, the runnable **will never run**.

So this really does not help and is irrelevant, as I never stated  `asyncExec` should never called, just calling it from the UI thread could be problematic if any kind of synchronization is about to happen between the calling code and the code scheduled async and for that reason the executor check this and prevent this case from happen.

And this is something that must be taken into account for the usage of an Executor in the context of Completable future (by the way that's what this change is about if someone remember) [because in its javadoc it says](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/CompletableFuture.html):

> All async methods without an explicit Executor argument are performed using the [ForkJoinPool.commonPool()](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/ForkJoinPool.html#commonPool()) (unless it does not support a parallelism level of at least two, in which case, a new Thread is created to run each task).

This implies that the async call can *not* work without special consideration if there is only *one* Thread, it was just neglected that this is required or useful and stated that I can always call `asyncExec` without **any** checks and this is obviously not true regardless of what the javadoc states.

So this is a technical act and I have shown it with an example and instead it is argued that some *feel* it *might* be not required.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
akurtakov: @iloveeclipse are you satisfied with the changed javadoc?

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
akurtakov: Did you come to an agreement here guys? Even in the case of multiple displays the UI thread would be single one so not much difference.

jukzi:
As i understand 3 committers voted we do not need this change in Display.

laeubi:
I don't see any official "vote" here (and I don't think its necessary either) just some concerns from people that think they don't need it .. fair enough. But there is a difference between "I don't need it" and "we don't need it", it was also stated that we don't need a change in the job API because it "should work", I already have given the references that shows one actually needs a special executor and can't always use `asyncExec`.

So I don't see a compelling reason why "we" don't need it e.g. because there is already such a function that could be used in a similar fashin. If it helps to convincing people I can extract it in just another class but the requirements remain to have a save, user friendly way to interact with the `Executor` / `CompletableFuture` as this is the basic building block for other features and SWT has a wide enough audience to assume that there are more use-case we currently can think of.

akurtakov:
I haven't cast any vote but rather trying to understand the status, concerns, implementation and etc. here. As I already see 2 blocking requests - I try to understand them first.

laeubi:
I think @jukzi refers to:

- himself (concerned about adding it to Display?)
- @iloveeclipse (who asked for better javadoc)
- @merks that had general concerns bit haven't left a review

vogella:
I personally really like the idea of making the execution in the main thread easier for CompletableFutures. I have implemented something similar for clients with a customer scheduler in project reactor and it would to have such a functionality in SWT.

sravanlakkimsetti:
requesting @lshanmug and @niraj-modi for review

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
mickaelistria: I'm looking at this interesting change for the 1st time and really like that
1. it provides a standard way to run in UI Thread without asking clients to verify whether they're in UI Thread or not first, so it replaces a lot of annoying code
2. It uses futures for which we can chain things very efficiently, ultimately removing need for complex chains/listeners of system vs UI jobs for patterns like "User clicks a button, a long computation happens out of the UI Thread, but we don't need a Job for it because it's in a modal dialog and we report progress differently, then the UI reacts upon completion". We happen to have it for example in Corrosion whether the wizard allows to provision the language-server.

However, I'm not sure this `executor()` static method is really useful for this story. Couldn't we just remove it if there is no clear agreement what it should do, and it there is no story to make it shiny?
I see it as yet-another-static-entry-point while we all know that it's better to infer the target from the context. Getting the display in a context (from a dialog, a widget, from an Event) is really easy, so this method wouldn't add much value, and would on the contrary lead people to write not-so-good code.
Also, I'm not knowledgeable with the multiple displays story, but I assume that in such case, it's really important that people do manage their display properly and that such method will make that code that relies on it (just as much as `Display.getDefault()` will hardly work for them). So if we can get rid of it, it would be better. `Display.getDefault()` is already "bad enough", no need to start wrapping it even more.

laeubi: I think a static method is not *strictly* required of course but:

- I frequently see `Display.getDefault()` (more than 600 references in platform) so it seems it is not *that* obvious where to get the current context display and may of those call `Display.getDefault().asyncExec` what seems to imply that one is not in an UI context yet or could acquire one other than calling a static method.
- The stream / future API makes heavy use of such static methods even if you can get it otherwise, to improve readability
- This might provides us with some opportunity here to interfere calls later on and hide some details.

mickaelistria: You're more motivating me to audit why some many Display.getDefault() in Platform than convincing me we need some static stuff on Display object.
I see a difference that most APIs in futures are written as utility or services and are hiding the implicit executor. But here the Display is more an object, and there can easily be multiple ones. If display implements `Executor` already, one can write
`future.thenAcceptAsync(widget::setText, widget.getDisplay())`, or `future.thenAcceptAsync(res -> findWidget().setText(res), Display.getDefault())`. I don't think introducing an API method makes it simpler in practice, and the "might provides us with some opportunity" is a bit too theoric to be a compelling reason to add one API. And even if we add an API, why not a `Display.currentOrDefault()` or why not calling it `Display.currentOrDefaultExecutor()`... This brings some more questions that we could easily just skip if we don't introduce a new API.

1 more question: what does happen if I chain `apply` on the same executors, ie if I run `CompletableFuture.completedFuture("hello").thenAcceptAsync(text -> shell::setText, shell.getDisplay()).thenRun(() -> shell.setVisible(true))`? Would that work in both UI and non-UI Thread?

laeubi: > You're more motivating me to audit why some many Display.getDefault() in Platform than convincing me we need some static stuff on Display object.

What ever fits your plans :-D

I think most uses are because code is called and caller do not know for sure if UI thread or not, something that this change should address. Maybe its just more convenient... 

> And even if we add an API, why not a `Display.currentOrDefault()` or why not calling it `Display.currentOrDefaultExecutor()`

At least I don't mind much how it is called, the rationale was just:

- If I normally would do something like `Display.getDefault()` then I could use `Display.executor()` now
- If I would call `display.anyThing` I could just pass the display

I more choose the name `executor()` as whether I get the default or the current is not of much interest when calling that method.

> what does happen if I chain `apply` on the same executors, ie if I run `CompletableFuture.completedFuture("hello").thenAcceptAsync(text -> shell::setText, shell.getDisplay()).thenRun(() -> shell.setVisible(true))`? Would that work in both UI and non-UI Thread?

I think this might describe it better than I can: https://stackoverflow.com/questions/65685715/will-a-chain-of-method-calls-completablefuture-api-execute-asynchronously-if-t

mickaelistria:
Please provide as part of this PR, or in some other ones, some example of code existing in Platform (production code, not examples nor tests) that would be made simpler by this proposal and that would leverage the proposed methods. This will make it easier to sort out what's needed and what's not, and how to name things.

laeubi:
> Please provide as part of this PR, or in some other ones, some example of code existing in Platform (production code, not examples nor tests) that would be made simpler by this proposal and that would leverage the proposed methods. This will make it easier to sort out what's needed and what's not, and how to name things.

Sorry but this is out of scope for me at the moment I won't invest more time here if there is no consent, I already added a very detailed javadoc without any reaction and it seems the effort is more to be vanished once again because of some vague concerns so this is all I can offer for now.

mickaelistria:
Sorry, but if this code is not proven to be helpful for Platform itself, then it's better to not add it.

vogella:
> Sorry, but if this code is not proven to be helpful for Platform itself, then it's better to not add it.

Didn't you use CompletableFutures in code minings? I did not look at the implementation but assume that you have to return to the main thread after calculation.

@laeubi if we add API we always try to use in our code to catch issues with the API in our day to day usage. We already skipped that with the new GroupLayout API but adding new API without using it in platform should be avoided.

laeubi:
> Sorry, but if this code is not proven to be helpful for Platform itself, then it's better to not add it.

That's why I won't waste any time on this, its always the same, code is not helpful, examples are bad 'we' don't need it and alike. Funny enough, if the 'right' people suggest something it is proven by design to be helpful, I'm really sick about this and this clearly repels anyone trying to improve "the platform" that claims to be useful for others but only evolves for its internal needs.

Instead we complain about to less contributions, people are migrating to IDEA/whatever/... and wonder why no one sees the glory of the Eclipse Platform 'helping' them and is slowly dying.

mickaelistria:
> Didn't you use CompletableFutures in code minings? I did not look at the implementation but assume that you have to return to the main thread after calculation.

It's done calling asyncExec, and as I already mentioned it would be great to do it with `thenAccept(res -> widget::setText, display)`. Other methods are questionable, and I made comments about those. I've suggested that the proposal needs to be enriched with more concrete code improvements in platform to show the substantial value, but my request is not welcome.

> That's why I won't waste any time on this

Let's close then.

laeubi:
> Let's close then.

It is still useful, I just wont try to search for "concrete code examples" jsut for the sake being told you can sue async exec instead...

Its really disturbing given that @mickaelistria and @vogella have claimed to have concrete use-cases for this to now state there is no use-case for this.

vogella:
>  @vogella have claimed to have concrete use-cases for this to now state there is no use-case for this.

Where did you get this from. I still think it useful and listed a potential use case in platform for code minings.

laeubi:
> Where did you get this from?

Sorry Lars, I might have misread/interpreted that. I'll see if I can prepare one more PR for that I just don't see that this will convince anyone that is not convinced right now.

To be clear I'm not strictly needing this functionality as I have working solution for this (and even more helpful things in that area I actually have had planed to contribute), I'm just

1. being sick to copy that helper class for each new bigger project
2. found it incredibly useful once it their and thought it would be useful for others
3. falsely assumed SWT might be open to adopting to more modern programming primives

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
laeubi: @iloveeclipse could you check if the javadoc fit now or like to recommend more changes? I would otherwise assume this as resolved?

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
laeubi: > I do like the idea, but i do not get the usecase where it is better then asyncExec.

@jukzi  Are there any concerns given the examples, possibles use cases and api references quite explain for what cases it might be required to have such a specialized executor?

merks:
Would these places be improved using the new facilities?

org.eclipse.jface.action.StatusLine.inUIThread(Runnable)
org.eclipse.jface.util.Policy.getDummyStatusHandler()

I found those by searching for `Display.getCurrent() != null` in the SDK workspace.  You might consider doing the same.

I've previously mentioned the following as similar but it is more complex though it also does useful disposal checking:

org.eclipse.jdt.internal.ui.util.SWTUtil.execDirectOrAsyncIfNecessary(Supplier<Control>, Runnable)

While we are busy criticizing others, we should be prepared to accept criticism too.  I sympathize because I must confess I too was rather annoyed during the previous week by changes apparently being held to a higher standard than the crappy code that was there currently---but it really was more work to argue about it than it was to do what was suggested.  Not only that, in the end the code was much nicer than it would have been if I continued to argue that an improvement to crappy code that was still somewhat crappy was still an improvement, which is a fact, but not one that feels in need to so much justification.   Sometimes it pays off to listen well, even when it starts to feel annoying.  High standards are both a curse and a blessing (and sometimes it does feel like there is a double standard).

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
iloveeclipse: Looks better now, just few smaller issues.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
```diff
@@ -1527,6 +1579,19 @@ public static Display getDefault () {
 	}
 }
 
+/**
+ * Returns an executor that will execute any submitted runnable in the user-interface thread, regardless of the calling thread
+ * @return an {@link Executor} that executes in the user-interface thread.
+ * @since 3.120
+ */
+public static Executor executor() {
```
iloveeclipse, line 1587:
```suggestion
public static Display executor() {
```

iloveeclipse: Looking on https://github.com/eclipse-platform/eclipse.platform.ui/pull/51 I wonder if the javadoc should mention, that the returned executor is not guaranteed to work if disposed by other code.

Also this would mean, the returned executor shouldn't be cached.

iloveeclipse (commented):
One thing upfront: in general I'm neutral to this API addition, I think if Christoph finds it useful, it is fine to be added except there are valid technical issues regarding proposed API (e.g. it doesn't work in some cases or violates some SWT rules etc).

laeubi:
Here is an example, saving an additional helper method: https://github.com/eclipse-platform/eclipse.platform.text/pull/16

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
jukzi: I still think we still need a *good* example. Preferably from within eclipse code where it can be useful. The only situation i know,  where the "compute now or later - i do dont care "- pattern is used is org.eclipse.jface.util.Throttler.throttledExec(). But even there i don't see that a Executor would simplify anything. I still like the idea of supporting the Executor style - but would rather see it in a platform independent class, as there is not need to extend Display. So if you would give an example from within eclipse where it can be used and change the Implementation i would support this kind of API. Another technical detail might be that it should be tested for multi-display usage which we got aware of could be used in win. If on the other hand multi display usage requires a (non static!) method in Display then i would support even a extension of Display.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
laeubi: Disposal is a good topic to mention here, I can think about that we check if the display is disposed and then simply *not* execute, this might be what most users actually want, but is a bit contradicted to the API of executor.

Beside that if we want to follow the Executor contract, I would simply throw `RejectedExecutionException` in that case, I think most of the time where the Display is disposed there is either something really wrong, or the system is shutting down, in both cases I won't mind to get a proper exception.

iloveeclipse: "Not execute" would hide programming errors. 
Display will throw "Device disposed" anyway, do we really need extra handling here? 
And yes, I assume in most cases that shouldn't be a problem, but it would make sense to add a note that there **is** a possibility that the executor will not always work.

laeubi:
> Would these places be improved using the new facilities?

possibly yes, I'll see if I can find some time to look in detail there, in the meantime I created an example as suggested by @vogella  here https://github.com/eclipse-platform/eclipse.platform.text/pull/16

> Sometimes it pays off to listen well, even when it starts to feel annoying.

Sure, but that's for both sides and I'm always open to suggestions and improvements and its always good to have different sights, but it becomes really frustrating (and not just annoing) if one get the feeling (might be wrong) that these high standards are just applied to argue against a change. We are reading here about just a few simple lines of code, and the wast majority of this are documentation and test, then answering the suggestions and such one could easily invest a lot of time and effort and one should assume that this is not done because the person is just bored, we all busy I think and it is really safe to assume that there *is* a use-case even though its not always easy to find a small, convincing and easy to demonstrate use-case in platform itself. If we just want that, it would probably be more easy to mark everything as `x-internal` and don't mind at the outer world...

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
laeubi: > Display will throw "Device disposed" anyway

Yes, but that's a bit SWT specific here so I think throwing `RejectedExecutionException` with the SWT error as a cause would be suitable as it is a declared exception of the api contract:

> @throws RejectedExecutionException if this task cannot be accepted for execution

e.g. there are executors that only allow to have a fixed bounded list of pending executions ... I think this is a similar case here.

merks:
Don't take this the wrong way, but I love the expression 'Our own personal faults are so much more glaring when we see them in others.'    I know I have faults, many, many fault, but I prefer to see them elsewhere because of course no one is more open-minded than me!

It probably took you longer to write the paragraph above than it would have taken to look at one example:

https://github.com/eclipse-platform/eclipse.platform.ui/blob/8c215cb8303120ce9c61315a5a13376add2348ca/bundles/org.eclipse.jface/src/org/eclipse/jface/action/StatusLine.java#L375-L385

And if you looked at that, you would be able to say, yes this example, and many like it, could and would be simplified by the proposal...

mickaelistria:
For Platform APIs, dogfooding is important. That's why I'm asking for some change in production code to leverage this proposal.

laeubi:
> Don't take this the wrong way

All fine, no need to worry.

> And if you looked at that, you would be able to say, yes this example, and many like it, could and would be simplified by the proposal...

Thanks for the reference, that really looks like a good example I just have to finish some other task right now but will come back to this as well! 

> For Platform APIs, dogfooding is important. That's why I'm asking for some change in production code to leverage this proposal.

Sure I already prepared one and I think the reference from Ed is also good as well. When changing actual production code it is said to be to complex to broad .... so finding good examples are (at laest for me) often harder than "please just provide an example" I'm just not *that* familiar with the platform code to always find a good example right now so it help if someone suggest so thanks again to @merks and @vogella suggesting good places to investigate.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
laeubi: About the return type I think `Executor` would be really better here to be less specific than possible and people not tend to accidently call sync/async exec for example?

jukzi:
> Here is an example, saving an additional helper method: [eclipse-platform/eclipse.platform.text#16](https://github.com/eclipse-platform/eclipse.platform.text/pull/16)

Thanks. That is also a good example of showing that it would be even better to put that execute() in 
org.eclipse.swt.widgets.Widget next to org.eclipse.swt.widgets.Widget.isDisposed(), as it would allow to reduce the text.isDisposed() check.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
iloveeclipse: Why shouldn't they be able to do that? At the end, it's a Display that implements Executor now. Or should it not implement but delegate? But then the code would not be so easy to use as proposed.

laeubi:
The aim is not to save some lines of code, a Widget has always as display and we might want to add additional methods to the widget that ultimately delegates to the display but not the other way round.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
laeubi: I think it should implement it to prevent additional roundtripps if you have a display (e.g. from a widget), but here the method is to get an `Executor` (and not a `Display`) that is we should not expose the implementation it self here but if we think its useful of course both would work. I just try to keep return values as open as necessary, e.g. returning Collection instead of List even though internally maybe an ArrayList is used.

jukzi:
> The aim is not to save some lines of code, a Widget has always as display and we might want to add additional methods to the widget that ultimately delegates to the display but not the other way round.

What would be the usecase to execute something in the Display thread - beside doing something with a Widget?

iloveeclipse:
> The aim is not to save some lines of code

Huh, I assumed that **is** the ultimate goal here? Providing API that allows to replace some boiler-plated code?

But I agree with Jörg, could you create a proposal that adds `Widget.execute()` too, and how that would look like?
Most cases I saw is that someone has some UI in the hand and wants to do something in the job, and afterwards do something with exact that UI but only if the UI is still there. Today we write lot of code that checks UI state before scheduling, after etc. Would be easier to say: "widget, if you are still there, please execute that piece of code"

laeubi:
Just take a look at `org.eclipse.jface.databinding.swt.DisplayRealm` for example there is no widget at all, or suppose you like to open a new Shell you can't create a shell without an (implicit) Display ...

laeubi:
> Huh, I assumed that **is** the ultimate goal here? Providing API that allows to replace some boiler-plated code?

Sure, but I defiantly won't like to save *every* possible boiler-plated wit *that* change, the goal is to maybe add on top of this...

> But I agree with Jörg, could you create a proposal that adds `Widget.execute()` too, and how that would look like?

As part of the current change? I'm not feeling its good to broaden the context too much, its hard enough to getting it done for that *very limited* scope already :-)

> Most cases I saw is that someone has some UI in the hand and wants to do something in the job, and afterwards do something with exact that UI but only if the UI is still there.

Yes for sure, that's why I try to setup some *basic building blocks* so we later can base on those, but not *all at once* to limit the scope and discussions. e.g changing/adding to widget might be even more controversial, as this could be considered API break as Widgets are extensible (Display is not)!

laeubi:
> org.eclipse.jface.action.StatusLine.inUIThread(Runnable)

https://github.com/eclipse-platform/eclipse.platform.ui/pull/52 replaces the method `inUIThread`

laeubi:
> org.eclipse.jface.util.Policy.getDummyStatusHandler()

Added to example https://github.com/eclipse-platform/eclipse.platform.ui/pull/52

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
laeubi: > would be a much more explicit and more powerful method

I could even add both ;-)

I just find this to verbose and trying to tell to much about internals, so following that we must rename `getCurrent()` -> `getCurrentOrNullIfNotUIThread()` and `getDefault()` -> `getDefaultOrCreateAndMakeCurrentThreadUIThread()` ...

But as mentioned before, if its just the name I could name it as you like ...

mickaelistria: > I could even add both ;-)

As usual with API, it's not the more the better, it's more the opposite. So let's try to only add what seems best.

> I just find this to verbose

I like verbose stuff, they help me in not having to read the documentation. However, any proposal to a shorter and as explicit name would be welcome. The name I suggest here is the one I thought of while reading the examples because I believe it makes everything clearer on consumer side.

> trying to tell to much about internals,

For the display, such details must matter to consumer. Someone who consumes a Display needs to get a sense of which one(s) they're using and get a sense that maybe `Display.executor()` is not the API they want because it's not necessarily the best display accessor to use in their context. We don't want things that work "magically", we want API that make user write good code. For this class, I believe exposing a bit more about the behavior in the method name allows that better.

laeubi:
> But I agree with Jörg, could you create a proposal that adds `Widget.execute()` too, and how that would look like?

I have added a possible implementation here: 
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/94

Of course one could build others on top of the executor primitive as well including one returning a `CompleableFuture` in case one is interested in accessing/processing return values...

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
laeubi: Well there is not much magic and thats why I have added the javadoc do clarify it do you think the javadoc is missing someting? At least for me the javadoc is more normative than any method name.

mickaelistria: > At least for me the javadoc is more normative than any method name.

People don't read doc. Naming is one of the most important part of software development (cf "Clean Code" by Bob Martin). Javadoc mustn't be a substitution to good naming.

iloveeclipse:
> > But I agree with Jörg, could you create a proposal that adds `Widget.execute()` too, and how that would look like?
> 
> I have added a possible implementation here:

OK, my original request was unfortunately not precise, and so the result is not what I've meant.
I meant actually "Widget.asyncExecute()", to do the work **later**, but your proposal would run the code immediately if we are on UI thread. Not your fault, I should have been more precise.

So let's rephrase again what I've meant: what about adding this to `Widget`:

```
public void asyncExec(Runnable r)
```
which would run the given task later **only** if the current widget is not disposed yet.
I think this is not related to the **current** PR, as it doesn't require Executor.

mickaelistria:
I opened https://github.com/eclipse-platform/eclipse.platform.debug/pull/15 as 1 example of the kind of pattern this proposal enables.

mickaelistria:
@laeubi it seems like most of the discussion happens on the static method. Maybe it would be worth building 2 distinct patches: 1 for the `implements Executor` and the other for the new method? At this stage, I believe splitting it would bring faster outcome.

jukzi:
Meanwhile, given the examples, i agree that it's suitable to have Display implements Executor.
However it's not good to have or use a static currentOrDefault() for those usecases as getDefault() would create a Display (possibly in any background worker thread) which could have just already have been disposed intentionally in the former ui thread. There are many example that getDefault() is used, but they should not have used that but they should have used widget.getDisplay() and we are going to improve that examples we should not follow wrong usage. Also most examples would better fit a widget.execute() and there is little use of Display.execute() then. It's ok to have Display.execute() but in all those example where a widget.execute() could be used it should be used to give best practice examples. please do not let in a N&N where a getDefault() is advocated and the necessary dispose check is missing.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
```diff
@@ -613,6 +615,63 @@ public void asyncExec (Runnable runnable) {
...
+public void execute(Runnable runnable) {
+	Objects.requireNonNull(runnable);
+	if (isDisposed()) {
+		throw new RejectedExecutionException(new SWTException (SWT.ERROR_WIDGET_DISPOSED, null));
+	}
+	if (thread == Thread.currentThread()) {
+		runnable.run();
+	} else {
+		asyncExec(runnable);
+	}
```
jukzi, line 672: What i do not like at this implementation is the handling of unchecked Exceptions. Let there be a NPE in runnable.run() then it depends on the current thread if that NPE is thrown in the calling thread or not. That contradicts the usecase of "i dont care when it is run".

laeubi: That's consistent to how executors works and documented in the javadoc as well.

laeubi: I have now adopted what the AsyncHandler is doing by calling the error/runtime exception handlers of the display.

jukzi: the default handler will still just throw the exception, while asyncExec will catch it in org.eclipse.swt.widgets.Synchronizer.runAsyncMessages(boolean)
```
       } catch (Throwable t) {
				SWT.error (SWT.ERROR_FAILED_EXEC, t);
	}
```
and then the SWT.error will wrap in SWTException .. and then the eventloop may care about the exception....

would it not better to simply always call asyncExec to get a consistent behavior?

laeubi:
As suggested by @mickaelistria I created a derivative of this that do not adds new static API:

https://github.com/eclipse-platform/eclipse.platform.swt/pull/289

I still think that a `Display.executor()` would be more semantic, but as it is not a technical requirement I won't want this to be further delayed because of this.

PR #71, opened 2022-04-23
Include in setup support for eclipse-platform/www.eclipse.org-swt

merks:


merks:
This doesn't affect the build so I aborted it.

Issue #74, opened 2022-04-28
Very slow handling of SWT messages when millions of messages queued


jukzi:
During https://github.com/eclipse-pde/eclipse.pde/issues/53 is see terrible hotspots in

- org.eclipse.swt.widgets.Synchronizer.addLast ()
- org.eclipse.swt.graphics.Device.isDisposed ()
- org.eclipse.swt.widgets.Synchronizer.asyncExec ()
- Display.findDisplay ()

while the amount of messages in org.eclipse.swt.widgets.Synchronizer.addLast(RunnableLock) was > 1million.

All those functions show "synchronized" blocks, which could be probably avoided with the usage of appropriate classes of java.util.concurrent.

Sampling of SWT Thread:
![image](https://user-images.githubusercontent.com/51790620/165740780-cfe93040-24a1-4bba-a803-770c6d855459.png)

And asyncExec() seen in a sampling of a background thread:
![image](https://user-images.githubusercontent.com/51790620/165741005-e726c671-c742-4011-b083-c79abdfa4087.png)

merks:
Good finding!  My sense is that org.eclipse.pde.internal.ui.editor.context.InputContextManager.resourceChanged(IResourceChangeEvent) should just build sets of added and removed resources and then do a single asyncExec in which it calls org.eclipse.pde.internal.ui.editor.context.InputContextManager.structureChanged(IFile, boolean) for each of those added/removed things.  Then likely all the other methods will not be a significant performance concern anymore.

vogella:
PR request are welcome to replace the sync blocks, I assume that would benefit all parts of SWT.

iloveeclipse:
Most parts of SWT don't use any sync blocks, because SWT is single threaded. There are only few that contact with the "outer" world :-)

vogella:
> @vogella
> 
> Of course improvements are always welcome. :-) But no amount of improvement will make doing asyncExec per changed resource when processing a resource delta as reasonable design! :-P That's definitely just bad design and should be fixed.

Of course but this discovery is still good as it may speed up SWT in general.

jukzi:
> @jukzi Are you planning on fixing the "too many asyncExecs" problem?"

Not at the moment. I submitted an Issue on PDE for that so that hopefully PDE team could take a look. I will take a short look if i find a quick improvement for SWT meanwhile  (my build is still running dequeuing the events after hours 😒 ).

merks:
@vogella 

[...]

PR #75, opened 2022-04-28
Avoid synchronized in Device.isDisposed #74

jukzi:
Device.isDisposed() is called much more frequently then dispose() and
should not be costly synchronized. The isDisposed() check in dispose()
is still synchronized so that still there will be no double dispose().

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Device.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Device.java
@@ -54,7 +54,7 @@ public abstract class Device implements Drawable {
 	long fontCollection;
 	String[] loadedFonts;
 
-	boolean disposed;
+	volatile boolean disposed;
 
 	/* Auto-Scaling*/
 	boolean enableAutoScaling = true;
@@ -766,9 +766,7 @@ public abstract void /*long*/ internal_dispose_GC (long hDC, GCData data);
  * @return <code>true</code> when the device is disposed and <code>false</code> otherwise
  */
 public boolean isDisposed () {
-	synchronized (Device.class) {
-		return disposed;
-	}
+	return disposed;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Device.java
```diff
@@ -336,6 +335,8 @@ public void dispose () {
 					trackingLock = null;
 				}
 			}
+		} finally {
```
vogella, line 338: Is this change related to the removal of the synchronization or just better style?

vogella:
Same change needs to be done for SWT Mac and SWT GTK.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Device.java
jukzi: While it probably not any usual usecase to dispose() Device i wanted to make sure that disposed flag is updated as the very last statement during dispose(). One could also change it to be the very first statement- in any case if isDisposed() is checked in any other thread it has almost no meaning as it could be disposed in the next blink of an eye. So it does not really matter.

SyntevoAlex:
Note that `Device.dispose()` is also synchronized. To my understanding, if one thread is disposing and the other is querying `isDisposed()`, the latter would wait and finally return `disposed = true`.

With new code, `isDisposed()` is no longer blocked and will return `false` even though the `Device` is already in the process of being disposed.

However, this new problem will probably not make things any worse. Because it's still wrong to call `isDisposed()` outside of lock and then expecting anything on the next line.

jukzi:
> Same change needs to be done for SWT Mac and SWT GTK.

done

vogella:
> > However, this new problem will probably not make things any worse. Because it's still wrong to call `isDisposed()` outside of lock and then expecting anything on the next line.
> 
> Thats what i think too!

Calling isDisposed outside the main thread and expecting no change would be wrong IMHO with or without a lock on this method. So I agree the change is fine.

lshanmug:
The javadoc of Device.dispose() states that "After this method has been invoked, the receiver will answer <code>true</code> when sent the message <code>isDisposed()</code>". Will this remain true after this change to Device.isDisposed() or can it return false in some cases?

jukzi:
> Maybe we should add to the Javadoc that we expect isDisposed to be called from the main thread?

no. there are a lot of usages where isDisposed() is called from other threads to avoid queuing up messages if widget is already disposed. Still needs a double isDisposed() check within SWT thread though.

PR #76, opened 2022-04-28
make Display.Displays size 1 by default #74

jukzi:
There is only 1 display - no need to search in an array of 4 displays
during Display.findDisplay()

It's a minor optimization for Display.findDisplay (). The greatest
bottle neck is still the "synchronized" statement.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
@@ -333,7 +333,7 @@ public class Display extends Device {
 
 	/* Multiple Displays. */
 	static Display Default;
-	static Display [] Displays = new Display [4];
+	static Display [] Displays = new Display [1];
 
 	/* Skinning support */
 	static final int GROW_SIZE = 1024;
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
@@ -496,7 +496,7 @@ public class Display extends Device {
 
 	/* Multiple Displays. */
 	static Display Default;
-	static Display [] Displays = new Display [4];
+	static Display [] Displays = new Display [1];
 
 	/* Skinning support */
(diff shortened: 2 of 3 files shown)


vogella:
iF there is only one Display why use an array at all?

jukzi:
@niraj-modi any objections?

lshanmug:
> There is only 1 display - no need to search in an array of 4 displays
> during Display.findDisplay()

There can be more than 1 display on Windows. Creating multiple displays is supported on Windows, but not on Linux and macOS. For consistency, the code is similar across the platforms.
`Display.checkDisplay(Thread thread, boolean multiple)` is called with true when multiple displays are supported and false otherwise.

merks:
This was a question about the default size here:
```
	static Display [] Displays = new Display [4];
```
That could be 1 instead of 4 and it will grow if an additional display is registered/needed, which is atypical, right?

mickaelistria:
> There can be more than 1 display on Windows.

Is it worth making 4 the default value though? I see that `register` would increase the array size if necessary. Since I imagine the vast majority of SWT-based apps use single display, we should maybe make 1 the default to save a few useless bytes?

> For consistency, the code is similar across the platforms.

Is this something you think should be kept or is it fine to make the code optimized according to the Platform even if it causes a few divergences?

merks:
> > There can be more than 1 display on Windows.
> 
> Is it worth making 4 the default value though? I see that `register` would increase the array size if necessary. Since I imagine the vast majority of SWT-based apps use single display, we should maybe make 1 the default to save a few useless bytes?
> 

The point isn't so much the useless bytes but rather the useless iterations in findDisplay; though I suspect the synchronized is the expensive part, not a few simple == comparisons.  But still it seems sensible to optimize for the common case...

lshanmug:
> > > There can be more than 1 display on Windows.
> > 
> > 
> > Is it worth making 4 the default value though? I see that `register` would increase the array size if necessary. Since I imagine the vast majority of SWT-based apps use single display, we should maybe make 1 the default to save a few useless bytes?
> 
> The point isn't so much the useless bytes but rather the useless iterations in findDisplay; though I suspect the synchronized is the expensive part, not a few simple == comparisons. But still it seems sensible to optimize for the common case...

Agreed. Wasn't opposing the change, wanted to clarify that multiple displays are supported.

niraj-modi (commented):
Fine with the change.

Issue #79, opened 2022-04-29
let readAndDispatch() handle multiple messages


jukzi:
During https://github.com/eclipse-platform/eclipse.platform.swt/pull/78 is see a hotspots in readAndDispatch for the costly checks for system messages. We could avoid those if we handle multiple internal messages before polling OS again. 
To avoid unresponsiveness to OS we could limit that internal message handling to a duration of for example 1ms.

This improves the BenchmarkSwtMultithreading handling from ~ 190ms to ~ 88ms on my Windows computer.
WDYT?
I know it is against the javadoc which clearly states readAndDispatch() will "Reads an event " which sounds like _single_ event.

laeubi:
I wonder if there is actually a use-case where it is important that only *one* event is read? If not I think it is fine handle multiple events. `EventLoopProgressMonitor` even uses 50ms so even 10ms would be fine I think as in most cases there is even not much to do than reading another event from the queue or sleep ...

mickaelistria:
> I know it is against the javadoc which clearly states readAndDispatch() will "Reads an event " which sounds like single event.

That says it all. Documented shouldn't be changed, and if there is a good reason to challenge them, they should be brought for consideration to PMC with an argumentation.
I think it was already discussed in another issue that some OS *might* manipulate their event queue while applications are processing. For example we can imagine that if one event causes an operation takes a long time in the SWT UI thread, then further OS events are generated in the meantime, but some of them may be removed/replaced because the OS know some intermediary events are not relevant. I'm not sure it's a realistic case, but theorically, it could be 1 good reason to request OS more often instead of getting a local copy of events.
However one may want to create and use an extra method to allow processing a batch of events.

laeubi:
> That says it all. Documented shouldn't be changed

I think there is still room for interpretation. e.g what is "an event", e.g. if I double click, the OS might or might not generate numerous events, but from an outer perspective I might see this as "one double click event".

> I think it was already discussed in another issue that some OS _might_ manipulate their event queue while applications are processing.

As SWT is OS dependent one might still optimize that on OS that do not do this, or check some "no longer relevant" flags.

mickaelistria:
[...]

PR #80, opened 2022-04-29
let readAndDispatch() handle multiple messages #79

jukzi:
avoid polling OS messages while internal messages are in the queue for
1ms

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
@@ -3655,7 +3655,7 @@ public boolean readAndDispatch () {
 		runDeferredEvents ();
 		return true;
 	}
-	return isDisposed () || runAsyncMessages (false);
+	return isDisposed () || runAsyncMessages (null); // handle multiple messages to avoid costly checks
 }
 
 static void register (Display display) {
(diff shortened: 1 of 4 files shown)


jukzi:
NPE in org.eclipse.swt.tests.junit.Test_org_eclipse_swt_widgets_Display.test_asyncExecLjava_lang_Runnable_dispose

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
```diff
@@ -3655,7 +3655,7 @@ public boolean readAndDispatch () {
 		runDeferredEvents ();
 		return true;
 	}
-	return isDisposed () || runAsyncMessages (false);
+	return isDisposed () || runAsyncMessages (null); // handle multiple messages to avoid costly checks
```
iloveeclipse, line 3658: This is very sensitive area & should be tested on all 3 platforms.
I'm not sure a potential improvement for the synthetic use case of billions async messages is worth introducing a subtle regression.

Running "all" "asyncExec" pending tasks (even time bound) might "ignore" OS events triggered by one of them, so we will see OS native events handled "too late", and see strange effects no one will ever be able to trace to this change.

jukzi:
Let's keep as-is until we find a need. I'll keep in mind that there is optimization potential for unusual cases.

Issue #32, opened 2022-04-08
NullPointerException in BorderLayout when Layout Data is null


laeubi:
It is assumed that a null layout data implies the center region, but BorderLayout.getTotal is not aware of this and throws:

```
Caused by: java.lang.NullPointerException
	at org.eclipse.swt.layout.BorderLayout.lambda$4(BorderLayout.java:233)
	at java.base/java.util.stream.ReferencePipeline$4$1.accept(ReferencePipeline.java:212)
	at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1655)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
	at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:913)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.base/java.util.stream.IntPipeline.reduce(IntPipeline.java:491)
	at java.base/java.util.stream.IntPipeline.sum(IntPipeline.java:449)
	at org.eclipse.swt.layout.BorderLayout.getTotal(BorderLayout.java:233)
	at org.eclipse.swt.layout.BorderLayout.computeSize(BorderLayout.java:147)
	at org.eclipse.swt.widgets.Composite.computeSizeInPixels(Composite.java:269)
	at org.eclipse.swt.widgets.Control.computeSize(Control.java:837)
	at org.eclipse.swt.widgets.Control.pack(Control.java:1609)
	at org.eclipse.swt.widgets.Control.pack(Control.java:1584)
```

niraj-modi:
Similarly with null Layout Data and layout.type = SWT.VERTICAL; BorderLayout.getMax() also throws NPE as below:
Exception in thread "main" java.lang.NullPointerException
	at org.eclipse.swt.layout.BorderLayout.lambda$5(BorderLayout.java:244)
	at java.base/java.util.stream.ReferencePipeline$4$1.accept(ReferencePipeline.java:212)
	at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1654)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
[...]

PR #83, opened 2022-05-01
NullPointerException in BorderLayout when Layout Data is null

laeubi:
Fix https://github.com/eclipse-platform/eclipse.platform.swt/issues/32
FYI @joel-majano


niraj-modi:
Suggest we also tweak Snippet379.java to cover null LayoutData scenario for testing purpose.

laeubi:
> Suggest we also tweak Snippet379.java to cover null LayoutData scenario for testing purpose.

Thanks for the review, I have added a `pack()` call to the example to cover that case.

niraj-modi (approved):
Looks good now, tested on Win10 for both:
- layout.type = SWT.HORIZONTAL
- layout.type = SWT.VERTICAL

PR #89, opened 2022-05-03
RTFTransfer fix, small fixes for empty clipboard and TextTransfer IDs

joel-majano:
- added handling for Rich text (RTF)
- Fixed errors logged to console if trying to paste with empty clipboard
- made sure registerType is not called in TextTransfer

I discussed other ways of doing RTF with the GTK devs through IRC. There were a few suggestions that did not work, and their suggestion was to use the gdk_clipboard_read function. Since this function is async, it adds another level of complexity that did not exist with the wait_for_contents function in GTK3. The solution implemented was the most reasonable one, and maintains similar functionality.


akurtakov:
Looks good.

PR #94, opened 2022-05-05
Showing possible widget usage of Display executor

laeubi:
Just a possible implementation for the widget class.


akurtakov:
Is this one still needed?

laeubi:
Not from my side ...

PR #99, opened 2022-05-06
Windows Compile fails due to missing Display#isXMouseActive

niraj-modi:
eclipse-platform#98

Change-Id: Id36bd8491165e645334cddba3fb6b30fc745ea8e

### DIFF
--- a/tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/tests/win32/AllWin32Tests.java
+++ b/tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/tests/win32/AllWin32Tests.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2018, 2022 Red Hat and others. All rights reserved.
+ * Copyright (c) 2018 Red Hat and others. All rights reserved.
  * The contents of this file are made available under the terms
  * of the GNU Lesser General Public License (LGPL) Version 2.1 that
  * accompanies this distribution (lgpl-v21.txt).  The LGPL is also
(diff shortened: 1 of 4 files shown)


tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/tests/win32/AllWin32Tests.java
```diff
@@ -23,7 +22,7 @@
 @RunWith(Suite.class)
 @Suite.SuiteClasses({
 	Test_org_eclipse_swt_dnd_DND.class,
-	Test_org_eclipse_swt_widgets_Display.class,
+//	Test_org_eclipse_swt_widgets_Display.class,
```
laeubi, line 25: Just wondering as it is named `AllWin32Tests` should it not actually *compile* when compiling for/with windows?

laeubi:
Just to not confuse here I just want to compile all sources/test not running them (yet), so it seems the compiler here gets a "wrong" reference to compile the test. As also the running later fails, I assume there is something missing in the maven workflow currently so this seems a good first start to verify windows-compile will use the windows compiled sources.

laeubi:
@niraj-modi currently you disable the **execution** of the test, but here the **compilation** fails! @akurtakov do you have any idea what could be wrong here? I have now removed all but the compilation phase but it seems the test compile is pulling in the linux part instead of the windows part ...

niraj-modi:
> @niraj-modi currently you disable the **execution** of the test, but here the **compilation** fails! @akurtakov do you have any idea what could be wrong here? I have now removed all but the compilation phase but it seems the test compile is pulling in the linux part instead of the windows part ...

For now I could work-around this compilation failure using reflection.

PR #100, opened 2022-05-07
Test GH action compile error on different platforms

laeubi:
**DO NOT MERGE**

These are just some tests of the new GH action verification

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -19,9 +19,9 @@ jobs:
       fail-fast: false
       matrix:
         config: 
-          - { name: Linux, os: ubuntu-latest, native: gtk.linux.x86_64 }
-          - { name: Windows, os: windows-latest, native: win32.win32.x86_64 }
-          - { name: MacOS, os: macos-latest, native: cocoa.macosx.x86_64 }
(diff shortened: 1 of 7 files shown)


laeubi:
I'm playing around ab bit with this now and have added a [compile error](https://github.com/eclipse-platform/eclipse.platform.swt/pull/100/commits/7dc5216756bab00746190d2df7f2983103a90fa1) to the Linux Display part.

The problem is, that the [all checks fails](https://github.com/eclipse-platform/eclipse.platform.swt/actions/runs/2285202257), if I have [checkAll = false](https://github.com/eclipse-platform/eclipse.platform.swt/pull/100/commits/0fa2b17fbf746d2bf08241849eedf8bf96b01f8d) then it seems it is not compiled and [all checks pass](https://github.com/eclipse-platform/eclipse.platform.swt/actions/runs/2285238615)

### Question:
How do I enable that each check only compiles the code for one platform?

laeubi:
I have now adjusted the ant script to allow compilation for a single platform and also adjusted the test build to get only the relevant target environment configuration what fixes the previous failing windows compile and thus can revert the workaround by @niraj-modi I will now squash everything and wait if the CI to see that it does not creates issues in the jenkins verification build.

laeubi:
There is one focus test failing on Jenkins, I just don't see how this could be related here, triggered a rebuild...

PR #101, opened 2022-05-07
Enable Tests for GH action

laeubi:
Lets see if we can get the tests running...

### DIFF
--- a/.github/workflows/junit.yml
+++ b/.github/workflows/junit.yml
@@ -1,0 +1,36 @@
+name: Publish Unit Test Results
+
+on:
+  workflow_run:
+    workflows: ["SWT Matrix Build"]
+    types:
+      - completed
+
+jobs:
+  unit-test-results:
+    name: Unit Test Results
+    runs-on: ubuntu-latest
+    if: github.event.workflow_run.conclusion == 'success'
+
+    steps:
+      - name: Download and Extract Artifacts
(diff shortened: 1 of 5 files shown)


laeubi:
Test can now be run on linux and windows but mac complains:

> ***WARNING: Display must be created on main thread due to Cocoa restrictions. Use vmarg -XstartOnFirstThread

even though the pom seem to already pass that option it seems not be picked up

laeubi:
MacOS Tests are running now but seem been blocked in `Running Test_org_eclipse_swt_browser_Browser#test_stop` by the following message:

> 2022-05-07 16:25:26.446 java[2764:12015] App Transport Security has blocked a cleartext HTTP (http://) resource load since it is insecure. Use HTTPS instead or add Exception Domains to your app's Info.plist.

can anyone with mac knowledge tell whats wrong here? Do we need additional configuration for macos or should we adjust the tests?

laeubi:
I tried to replace all `http` with `https` in tests now...

laeubi:
Now all verification steps succeed, lets see if it has any impact on the Jenkins CI ...

lshanmug:
> I tried to replace all `http` with `https` in tests now...

On Mac, the Browser does not allowed http now in due to App Transport Security, so we can fix the tests to use https.

PR #102, opened 2022-05-09
Check OS specific tests

laeubi:
This is for testing the OS specific tests:

1. Current failing tests
2. Make sure OS specific fragments are not only compiled but executed
3. ... maybe other things to check/adjust ...


tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_printing_Printer.java
```diff
@@ -157,19 +161,34 @@ public void test_getPrinterList() {
...
+				 * Windows also has some default (non hardware) printer by default
 				 */
-				assertEquals("GtkPrintBackendFile", list[0].driver);
-			}
-		} else {
-		assertEquals("printer list contains items even though there are no printers",
-				0 , list.length);
+				if ("Microsoft Print to PDF".equalsIgnoreCase(pd.name) || "Microsoft XPS Document Writer".equalsIgnoreCase(pd.name) || "Fax".equalsIgnoreCase(pd.name)) {
+					return false;
+				}
```
niraj-modi, line 184: IMO 'switch-case' would be bit more readable here, something like below:
switch (pd.name.toLowerCase()) {
case ("fax"):
case ("microsoft print to pdf"):
case ("microsoft xps document writer"):
	return false;
}

laeubi: done.

niraj-modi (approved):
Looks good to me.

PR #113, opened 2022-05-10
Test that OS specific test are failing and are reported

laeubi:


jukzi:
@laeubi
somehow such actions look potential useful. thanks for workin on them. can you also add a https://github.com/marketplace/actions/sonarcloud-scan action?

laeubi:
@jukzi my personal opinion is that the SonarCloud scans are not very helpful in some cases and some alerts are questionable. So I have not used it in the past. Given that are some complains about not enough dev resources it might also even disturb people to see more warnings as we already have with default eclipse code warnings.

So I think the best would be to open a [new discussion of type "poll"](https://github.com/eclipse-platform/eclipse.platform.swt/discussions/new) to gather some feedback, then one needs to check how we get an appropriate (eclipse?) account and store the secrets / configs as described here: https://github.com/SonarSource/sonarcloud-github-action and then it should be rather trivial to add the check itself.

akurtakov:
@jukzi I do have (sonarlint plugin) [https://www.sonarlint.org/eclipse/] installed in my IDE and I fix things as I see them (quite helpful tbh). Considering how few pay attention to https://download.eclipse.org/eclipse/downloads/drops4/I20220510-1800/testResults.php#PluginsErrors (and this has been around ever since 2008 at least!) and how many people pay attention to warnings in the IDE itself - the only working solution I found is to fix warnings wherever I can, raise some to errors in project properties so things no longer compile for people if they introduce such and etc. 
Unfortunately, generally improving the codebase is not a task embraced by many and just showing extra warnings while we have hundreds visible already will not make people magically look at them.

jukzi:
i would like to get a overview about the most severe (security, multithreading) issues.
@akurtakov i guess few people care about the nightly build errors is that they are not obvious related to commits.  it would help if all new errors on different platforms are reported before pushing a commit. And therefore those github actions to test all platforms could improve the situation. I do not know if it is possible to configure, but if we could get sonar warnings on new commits only - i.e. only new warning reported - it might also help to at least not introduce new errors.

laeubi:
@jukzi You can try that out with your fork and a private sonatype account, just setup everything in a branch, then create a PR with that branch as a target.

jukzi:
Was not straight forward to configure. Anyway it finally runs. take +10min for the analysis and found 1 security issue - which for me proves we should run analysis like that:
https://sonarcloud.io/project/issues?resolved=false&types=VULNERABILITY&id=jukzi_eclipse.platform.swt

vogella:
@jukzi just in case you don't know this, but SonarLint can be installed into your IDE and allows a batch run for selected projects.

vogella:
Btw. +1 for adding this to the GH action. Any additional quality check done in the background is good. Would be good if we fails the build in case more compiler errors or Sonar warnings are introduced with a change but that is a different topic.

jukzi:
it would be better to have a regular analysis. does not matter if bugs get fixed and later reappear because its not analysed anymore

PR #117, opened 2022-05-11
Try out ubuntu-22.04

laeubi:


akurtakov:
Should this one be kept ?

laeubi:
If you think it is useful I can add this as another matrix option to the github verification, currently we only use ubuntu-latest what is ubuntu-20.04 at the moment.

laeubi:
I think the length of checks should really not be a measure but if it helps us to find issues we can even have more java and more os variants I think given that these checks often complete before the regular Jenkins CI...

mickaelistria:
> I think the length of checks should really not be a measure

There is plenty of litterature about how a shorter "feedback loop" has a positive impact on project quality and productivity.
I'm not saying it's the highest priority here though, but it's is something important to keep in mind.

laeubi:
> > I think the length of checks should really not be a measure
> 
> There is plenty of litterature about how a shorter "feedback loop" has a positive impact on project quality and productivity. I'm not saying it's the highest priority here though, but it's is something important to keep in mind.

Sure I just wanted to note that "I need to scroll" is a bad measure weather or not we have a certain check, so we might tell people using platform X that we only test/support/care/... about A,BC for this or that reason and I think that's fine, but we should not tel them we could have checked that but find the scrollbar on the checkmarks seem to be too long :-)

PR #120, opened 2022-05-11
Add minimal sonarcloud parameters to pom.xml

jukzi:
This itself does not enable sonarcloud. But just in case someone would like to run sonarcloud on a fork: At least these parameters need to be added till a result can be obtained.
Example result see:
https://sonarcloud.io/summary/overall?id=jukzi_eclipse.platform.swt


akurtakov:
Doesn't hurt anything so let's push it.

PR #123, opened 2022-05-11
Update SWT_AWT.java

akurtakov:
Do not submit! Test for codeowners functioning.


akurtakov:
And it seems to not function

PR #126, opened 2022-05-11
Consistently fire 'SWT.ZoomChange' event on OS zoom changes with single monitor setup.

niraj-modi:
eclipse-platform#125

Change-Id: I2d1724ae2f6bc9de742ef931ebaa7cd0ab514f1e


akurtakov:
@niraj-modi Is this one planned for M3?

merks:
I'm not sure this is actually a good idea.  When SWT does scaling it generally does a poor job except perhaps for integral scaling.  Until the various scaling issues are addressed, e.g., each monitor can have its own scaling factor but SWT doesn't actually support that.  See https://github.com/eclipse-platform/eclipse.platform.swt/issues/286 for more details.

In any case, the following code will need to be revisited:

https://github.com/eclipse-platform/eclipse.platform.ui/blob/5ce73d59d2e9f26b3f80ad7faf00538fbb3970f9/bundles/org.eclipse.ui.workbench/Eclipse%20UI/org/eclipse/ui/internal/WorkbenchWindow.java#L940-L954

akurtakov:

PR #134, opened 2022-05-12
Simplyfies if statements in printing package

vogella:
No need for if (value!=false), can simply be if (value).


niraj-modi (commented):
Cocoa's Printer.java also needed a similar fix, just created #135 for the same.

Issue #95, opened 2022-05-05
FileDialog does not remember previous directory


Phillipus:
Windows 10
Eclipse 4.23 and 4.24

This seems to be a regression as this is working in Eclipse 4.22 but not in 4.23 and 4.24.

In the SWT as used in Eclipse 4.22 and earlier, a FileDialog would remember the last open directory. In the SWT in Eclipse 4.23 and later it reverts to the root "Computer" directory on Windows.

Run the snippet below against a target of Eclipse 4.22 and compare with a target of 4.23. Click the "Open" button, navigate to a file and open this. Repeat. On Eclipse 4.22 target the previous directory is remembered. Not on 4.23 target.

```java
import org.eclipse.swt.SWT;
import org.eclipse.swt.events.SelectionAdapter;
import org.eclipse.swt.events.SelectionEvent;
import org.eclipse.swt.layout.GridLayout;
import org.eclipse.swt.widgets.Button;
import org.eclipse.swt.widgets.Display;
import org.eclipse.swt.widgets.FileDialog;
import org.eclipse.swt.widgets.Shell;

public class FileOpen {

    public static void main(String[] args) {
        Display display = new Display();
        Shell shell = new Shell(display);
        shell.setLayout(new GridLayout());
        shell.setSize(300, 200);
        
        Button button = new Button(shell, SWT.PUSH);
        button.setText("Open");
        button.addSelectionListener(new SelectionAdapter() {
            @Override
            public void widgetSelected(SelectionEvent e) {
                new FileDialog(shell, SWT.OPEN).open();
            }
        });
        
        shell.open();

        while(!shell.isDisposed()) {
            if(!display.readAndDispatch()) display.sleep();
        }
        display.dispose();
    }
}
```

Phillipus:
Just discovered a workaround:

`FileDialog.setFilterPath(null);`

niraj-modi:
@nnemkin is working on this issue via bugzilla: https://bugs.eclipse.org/bugs/show_bug.cgi?id=579359

niraj-modi:
With is issue fixed below bug is now reopened:
Bug https://bugs.eclipse.org/bugs/show_bug.cgi?id=577190 is reopened now.

PR #142, opened 2022-05-13
[Win32] Restore historical FileDialog behavior

nnemkin:
Miration to COM API in Bug 571571 resulted in unexpected changes in behavior. This patch restores original behavior.

 * Allow non-existent file names in SWT.OPEN dialog (Bug 579359).

 * Let the system remember last used directory and override FileDialog.setFilterPath (Bug 577773 and #95).
   This change reverts the fix for Bug 577190 and Bug 578415.

@niraj-modi please see my comment https://bugs.eclipse.org/bugs/show_bug.cgi?id=577190#c17


niraj-modi (approved):
This patch works and it effectively reopens bug:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=577190

Issue #144, opened 2022-05-06
Word wrap after scrolling is broken since 4.19


iloveeclipse:
Open "New untitled editor" and paste this exception stack below.
After that put cursor on first line and try to toggle "Word Wrap" in the main toolbar.
Error with the same exception is logged in the log, line numbers disappear from line number ruler.

Same happens with almost every Java editor opened on a non trivial Java file that has longer lines. I haven't found smallest possible text that causes the error yet.

Either some StyledText change broke the editor, or some changes in text framework, I haven't bisected that yet.

**Update**: to reproduce, editor should have scrollbars and should scroll at least once down and back. 
I see it back to **4.19**, so regression is somewhere there, 4.18 is OK. OMG.

```
org.eclipse.core.runtime.AssertionFailedException: assertion failed: 
	at org.eclipse.core.runtime.Assert.isTrue(Assert.java:113)
	at org.eclipse.core.runtime.Assert.isTrue(Assert.java:99)
	at org.eclipse.jface.text.Position.<init>(Position.java:65)
	at org.eclipse.jface.text.AbstractDocument.getPositions(AbstractDocument.java:1608)
	at org.eclipse.core.internal.filebuffers.SynchronizableDocument.getPositions(SynchronizableDocument.java:246)
	at org.eclipse.jface.text.source.AnnotationModel.getRegionAnnotationIterator(AnnotationModel.java:716)
	at org.eclipse.jface.text.source.AnnotationModel.getAnnotationIterator(AnnotationModel.java:678)
	at org.eclipse.jface.text.source.AnnotationRulerColumn.doPaint(AnnotationRulerColumn.java:642)
	at org.eclipse.jface.text.source.AnnotationRulerColumn.doubleBufferPaint(AnnotationRulerColumn.java:548)
	at org.eclipse.jface.text.source.AnnotationRulerColumn.lambda$1(AnnotationRulerColumn.java:274)
	at org.eclipse.swt.widgets.TypedListener.handleEvent(TypedListener.java:234)
	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:89)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:5794)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1529)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1555)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1538)
[...]

PR #145, opened 2022-05-13
[StyledText] Test scroll then wordwrap

mickaelistria:
Reproduces issue
https://github.com/eclipse-platform/eclipse.platform.swt/issues/144


mickaelistria:
A 1-char fix is now included. This does also fix the downstream test case from https://github.com/eclipse-platform/eclipse.platform.text/pull/28 .
I think it's ready to be reviewed/merged.

iloveeclipse:
Thanks.

Issue #146, opened 2022-05-14
Clipboard ImageTransfer does not work on Linux x86_64 - possible fix


Phillipus:
For some time now (13 years) this has not worked on Linux 64-bit:

https://bugs.eclipse.org/bugs/show_bug.cgi?id=283960

While investigating #143 (and using the test snippet there) I took a look at `org.eclipse.swt.dnd.ImageTransfer` and line 117:

`transferData.format = 32;`

I simply changed the value to 16 and, voila, it works.

Could this be a simple fix?

iloveeclipse:
Just push a PR. It is broken anyway, so what could go wrong?

Phillipus:
> Just push a PR. It is broken anyway, so what could go wrong?

OK, but I just wanted to understand what that change actually does, or if there is something else in that chunk of code that could be changed.

```
public void javaToNative(Object object, TransferData transferData) {
	if (!checkImage(object) || !isSupportedType(transferData)) {
		DND.error(DND.ERROR_INVALID_DATA);
	}
	ImageData imgData = (ImageData)object;
	if (imgData == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
	Image image = new Image(Display.getCurrent(), imgData);
	long pixbuf = ImageList.createPixbuf(image);
	if (pixbuf != 0) {
		String typeStr = "";
		if (transferData.type ==  JPEG_ID) typeStr = "jpeg";
		else if (transferData.type ==  PNG_ID) typeStr = "png";
		else if (transferData.type ==  BMP_ID) typeStr = "bmp";
		else if (transferData.type ==  EPS_ID) typeStr = "eps";
		else if (transferData.type ==  PCX_ID) typeStr = "pcx";
		else if (transferData.type ==  PPM_ID) typeStr = "ppm";
		else if (transferData.type ==  RGB_ID) typeStr = "rgb";
		else if (transferData.type ==  TGA_ID) typeStr = "tga";
		else if (transferData.type ==  XBM_ID) typeStr = "xbm";
		else if (transferData.type ==  XPM_ID) typeStr = "xpm";
		else if (transferData.type ==  XV_ID) typeStr = "xv";
		byte[] type = Converter.wcsToMbcs(typeStr, true);
		long [] buffer = new long [1];
		long [] len = new long [1];
		if (type == null) return;
		GDK.gdk_pixbuf_save_to_bufferv(pixbuf, buffer, len, type, null, null, null);
[...]

PR #148, opened 2022-05-14
Fix Clipboard ImageTransfer not working on Linux x86_64

Phillipus:
- Change value of transferData.format from 32 to 16
- See #146

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Drag and Drop/gtk/org/eclipse/swt/dnd/ImageTransfer.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Drag and Drop/gtk/org/eclipse/swt/dnd/ImageTransfer.java
@@ -114,7 +114,11 @@ public void javaToNative(Object object, TransferData transferData) {
 		transferData.pValue = buffer[0];
 		transferData.length = (int)(len[0] + 3) / 4 * 4;
 		transferData.result = 1;
-		transferData.format = 32;
+		// The following value has been changed from 32 to 8 as a simple fix for #146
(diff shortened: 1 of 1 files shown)


iloveeclipse:
@akurtakov , @sravanlakkimsetti, @SyntevoAlex : I'm inclined to merge this PR today so we can get it into **4.24 M3**.

Any objections?

iloveeclipse:
@akurtakov , @sravanlakkimsetti, @SyntevoAlex : I'm inclined to merge this PR today so we can get it into **4.24 M3**.

Any objections?

Phillipus:
@iloveeclipse Do you think we set the value to 8 because of https://github.com/eclipse-platform/eclipse.platform.swt/issues/146#issuecomment-1126800896

**16.3 Supplying the selection** `The format field is actually important here - the X server uses it to figure out whether the data needs to be byte-swapped or not. Usually it will be 8 - i.e. a character - or 32 - i.e. a. integer.`

akurtakov:
No objection from my side. Local test (Fedora 36, Wayland) showed that format=32 makes image (copied from ClipboardExample) to be pasted(in Gimp) partially. format=8 and format=16 work.

Phillipus:
OK, I'll change the value to 8.

What's best practice of PRs? Do I just add another commit on my fork, or something else?

> and document that we simply guess it makes more sense...

Where to do that? In the git commit? In code?

Phillipus:
Forced push the changes now.

Happy to make changes to the comment if needed.

akurtakov (approved):

iloveeclipse (approved):
Thank you!

Phillipus:
Thanks for the fast review, feedback and advice!

PR #166, opened 2022-05-24
Try to generate a badge with the test results

laeubi:


akurtakov:

laeubi:
I wanted to wait util release to not create too much noise, sadly this could not be verified with a PR so I need to first check this with my fork or we merge this and improve it later fi there are any issues.

akurtakov:
As it's GH specific just merge it and improve.

akurtakov:
Let's see what will come out of this one.

.github/workflows/junit.yml
```diff
@@ -30,7 +30,39 @@ jobs:
...
+          esac
+
+      - name: Create badge
+        if: github.ref == 'refs/heads/master'
+        uses: emibcn/badge-action@d6f51ff11b5c3382b3b88689ae2d6db22d9737d1
+        with:
+          label: Tests
+          status: '${{ fromJSON( steps.test-results.outputs.json ).stats.tests }} tests, ${{ fromJSON( steps.test-results.outputs.json ).stats.runs }} runs: ${{ fromJSON( steps.test-results.outputs.json ).conclusion }}'
+          color: ${{ env.BADGE_COLOR }}
+          path: badge.svg
```
SyntevoAlex, line 60: Note this filename

.github/workflows/junit.yml
```diff
@@ -30,7 +30,39 @@ jobs:
...
 
+      - name: Upload badge to Gist
+        if: github.ref == 'refs/heads/master'
+        uses: andymckay/append-gist-action@1fbfbbce708a39bd45846f0955ed5521f2099c6d
+        with:
+          token: ${{ secrets.GIST_TOKEN }}
+          gistURL: https://gist.githubusercontent.com/eclipse-platform/eclipse.platform.swt/
+          file: test_results_badge.svg
```
SyntevoAlex, line 68: Note this filename

SyntevoAlex (commented):
Unfortunately GitHub side unit tests fail after this was merged, with
```
 Error: ENOENT: no such file or directory, open 'test_results_badge.svg'
    at Object.openSync (fs.js:462:3)
    at Object.readFileSync (fs.js:364:35)
    at Object.<anonymous> (/home/runner/work/_actions/andymckay/append-gist-action/1fbfbbce708a39bd45846f0955ed5521f2099c6d/gist.js:12:19)
    at Module._compile (internal/modules/cjs/loader.js:999:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1027:10)
    at Module.load (internal/modules/cjs/loader.js:863:32)
    at Function.Module._load (internal/modules/cjs/loader.js:708:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:60:12)
    at internal/main/run_main_module.js:17:47 {
```

laeubi:
@SyntevoAlex the problem is that we need a gist first created by the bot see:

https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/1322#note_809508

but sadly EF infra seem to be too busy to work on this :-\

SyntevoAlex:
That's sad.
I understand that it makes sense to revert the patch for now, because current;y all GitHub JUnit tests report as failed?

SyntevoAlex:
I investigated a different approach with uploading badge as artifact via `actions/upload-artifact@v2`. That would require no changes from infra, as I understand it. Yet alas, there seems to be no easy way to obtain the URL of the uploaded artifact.

laeubi:
@SyntevoAlex yes I think the best would be to disable/revert this for now, I though it would just be a matter of a day but seem either get out of scope or EF infra is to much overloaded with other stuff.

The problem is that one would need a constant URL to use it somewhere in the Readem... and I honestly don't want to use to much different services/techniques so the gist approach seems a good way.

On the other hand it seems not much people have missed that anyway to see the status of the test ...

SyntevoAlex:
I did some more experimenting.

Specifically, I tried to just create the gist from GitHub Action. I used code (see #393)
```
env:
  GH_TOKEN: ${{secrets.GITHUB_TOKEN}}

<...>

echo "GITHUB_TOKEN = ${{secrets.GITHUB_TOKEN}}"
echo "GIST_TOKEN = ${{secrets.GIST_TOKEN}}"
gh gist create --desc "eclipse.platform.swt.tests" --public
```

The output was
```
GITHUB_TOKEN = ***
GIST_TOKEN =
- Creating gist...
X Failed to create gist: HTTP 403: Resource not accessible by integration (https://api.github.com/gists)
```

My conclusions is that `GIST_TOKEN` is not in fact available in GitHub Action.
https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/1322#note_809508 says "GIST_TOKEN for user eclipse-releng-bot has been created", but I don't think that it's the `eclipse-releng-bot` who runs GitHub action.

Furthermore, making token available to GitHub Action would probably create a security issue. See, anyone can create a PR, and in it, it can modify the Action script to output GitHub token and steal it and/or use it for evil purposes. Just like I did in #393.

laeubi:
You need to use a personal access token, at least with my testing this has worked and thus it should work with the bot account as well but as far as I understand you can not create a gist!

About security, not *everyone* can do that, first time contributors needs approval from a committer and one can do much more "evil" already bejond modifying a gist anyways already :-)

PR #174, opened 2022-05-26
Request symbolic images on GTK but prefer colorful

### LINKED ISSUE #173: Too small icons for system images with Adwaita theme
As of version 42.0 Adwaita theme (default Gnome theme) no longer ships non symbolic images for "dialog-*" images. Due to this a 16x16 default one is loaded.
Screenshot showing the issue
The fix would be to request symbolic images with the GTK_ICON_LOOKUP_FORCE_REGULAR flag so the colorful one is shown if available in the team but to show symbolic image otherwise.

akurtakov:
Only symbolic images are "guaranteed" to be available and actually
available in adwaita-icon-theme 42.0. Thus request symbolic images but
pass the flag to prefer the colorfull images if available.
Fixes #173 .


akurtakov:
@sravanlakkimsetti Would you please review this one for RC2? It might even fix the NPE you had on Ubuntu with some theme.

sravanlakkimsetti (approved):
I checked with ubuntu with gnome 
if adwaita-icon-theme-full package installed I do get colorful icons. if this package is not there earlier I was getting NPE. now I am getting symbolic images.

By the way gnome desktop environment also behaves the same way colorful images are show only if adwaita-icon-theme-full package installed

With Ubuntu DE I am getting symbolic image though the flag is set.
without patch
wthout patch

with patch
with patch

But I don't think this is a problem for now. Most likely the DE needs to be adopt the new theming.

Issue #55, opened 2022-04-19
Allow to easily spot which CTabFolder/CTabItem has focus (eg in Workbench)


mickaelistria:
1 issue that theme have to resolve is that when there are multiple CTabFolders on the same window (eg the workbench), it's not clear which current folder has the focus and which actual tabs has focus. The default CTabFolderRenderer should try to resolve that issue, maybe with some option; by applying some color or border to better identify the current focus in case multiple CTabFolders are shown on the same window.

mickaelistria:
Here is a proposal of how this would look like (the color would be the one defined in SWT.SELECTION_COLOR_BACKGROUND constant or similar)
![ctab-before-after](https://user-images.githubusercontent.com/469058/170983553-baba48ae-02ff-4077-8923-c4f88c9473d1.png)

PR #183, opened 2022-05-31
More visible highlight of active CTab(Folder) #55

mickaelistria:
Adds a thick line under the active tab of each folder; and use the
selection color for the "active with focus".

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
@@ -3474,7 +3474,7 @@ public void setSelectionForeground (Color color) {
 /**
  * Sets the thickness of the highlight bar on the selected tab. The highlight bar is drawn in the top margin of the selected tab.
  *
- * @param thickness the desired thickness. Most be positive and lower than {@link CTabFolderRenderer#ITEM_TOP_MARGIN} (that is either {@code 0} {@code 1} or {@code 2} at the moment),for correct results.
+ * @param thickness the desired thickness. Must be positive and lower than {@link CTabFolderRenderer#ITEM_TOP_MARGIN} (that is either {@code 0} {@code 1} or {@code 2} at the moment),for correct results.
  *                  Set to {@code 0} to not draw a highlight bar.
  * @exception SWTException <ul>
  *    <li>ERROR_WIDGET_DISPOSED - if the receiver has been disposed</li>
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
@@ -189,6 +189,7 @@ public class CTabFolder extends Composite {
 	boolean selectionGradientVertical;
 	Color selectionForeground;
 	Color selectionBackground;
+	int selectionHighlightBarThickness = 2;
 
 	/* Unselected item appearance */
 	Color[] gradientColors;
@@ -3471,6 +3472,27 @@ public void setSelectionForeground (Color color) {
 }
 
 /**
+ * Sets the thickness of the highlight bar on the selected tab. The highlight bar is drawn in the top margin of the selected tab.
+ *
+ * @param thickness the desired thickness. Most be positive and lower than {@link CTabFolderRenderer#ITEM_TOP_MARGIN} (that is either {@code 0} {@code 1} or {@code 2} at the moment),for correct results.
+ *                  Set to {@code 0} to not draw a highlight bar.
+ * @exception SWTException <ul>
+ *    <li>ERROR_WIDGET_DISPOSED - if the receiver has been disposed</li>
+ *    <li>ERROR_THREAD_INVALID_ACCESS - if not called from the thread that created the receiver</li>
+ *    <li>ERROR_INVALID_ARGUMENT - if the parameter value is invalid</li>
+ * </ul>
+ * @implNote Default {@link CTabFolderRenderer} currently ignores this setting if {@link #getSimple()} is {@code true}.
+ * @since 3.121
+ */
+public void setSelectionBarThickness(int thickness) {
+	checkWidget();
+	if (thickness < 0) {
+		SWT.error(SWT.ERROR_INVALID_ARGUMENT);
+	}
+	this.selectionHighlightBarThickness = thickness;
+}
+
+/**
  * Sets the shape that the CTabFolder will use to render itself.
  *
  * @param simple <code>true</code> if the CTabFolder should render itself in a simple, traditional style
@@ -4192,7 +4214,7 @@ public void setHighlightEnabled(boolean enabled) {
  * highlighted.
  *
  * @return <code>true</code> if the selected tab is rendered as
- *         highlighted
+ *         highlighted.
  *
  * @exception SWTException
  *                <ul>
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
@@ -1354,9 +1354,17 @@ public class CTabFolderRenderer {
 					}
 				}
 
-				//Highlight MUST be drawn before the outline so that outline can cover it in the right spots (start of swoop)
+				//Highlight colors MUST be drawn before the outline so that outline can cover it in the right spots (start of swoop)
 				//otherwise the curve looks jagged
 				drawHighlight(gc, bounds, state, rightEdge);
+
+				// draw highlight marker of selected tab
+				if (parent.selectionHighlightBarThickness > 0 && parent.simple) {
+					Color previousColor = gc.getBackground();
+					gc.setBackground(item.getDisplay().getSystemColor(parent.shouldHighlight() ? SWT.COLOR_LIST_SELECTION : SWT.COLOR_WIDGET_DISABLED_FOREGROUND));
+					gc.fillRectangle(x + 1 /* outline */, parent.onBottom ? y + height - 1 - parent.selectionHighlightBarThickness : y + 1, width - 2 /*outline*/, parent.selectionHighlightBarThickness);
+					gc.setBackground(previousColor);
+				}
 
 				// draw outline
 				shape[0] = Math.max(0, borderLeft - 1);
@@ -1797,4 +1805,5 @@ public class CTabFolderRenderer {
 	boolean useEllipses() {
 		return parent.simple;
 	}
+
 }
--- a/bundles/org.eclipse.swt/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.swt/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-Name: %pluginName
 Bundle-Vendor: %providerName
 Bundle-SymbolicName: org.eclipse.swt; singleton:=true
-Bundle-Version: 3.120.100.qualifier
+Bundle-Version: 3.121.0.qualifier
 Bundle-ManifestVersion: 2
 Bundle-Localization: plugin
 DynamicImport-Package: org.eclipse.swt.accessibility2
--- a/bundles/org.eclipse.swt/pom.xml
+++ b/bundles/org.eclipse.swt/pom.xml
@@ -21,7 +21,7 @@
     </parent>
     <groupId>org.eclipse.swt</groupId>
     <artifactId>org.eclipse.swt</artifactId>
-    <version>3.120.100-SNAPSHOT</version>
+    <version>3.121.0-SNAPSHOT</version>
     <packaging>eclipse-plugin</packaging>
 
     <properties>


mickaelistria:
@vogella The goal is to allow this for SWT without any help from CSS. Ultimately, I want to get rid of the System theme as you suggested earlier, and to improve the no-theme so I can disable theming with less loss. Good highlighting of the tab is something plain SWT should provide first.

iloveeclipse:
@mickaelistria : could you please provide same screenshot with the highlight *at top* of the tab? Highlighting at the bottom looks unnatural to me, it "breaks" the "tab" idea. Look how gedit does it:
image

But! I would avoid adding things to the SWT which do not match "native" L&F, which is here the case. Native tabs on Windows don't have such "colored bars", and I guess Mac also doesn't follow "Adwaita" style. 

So ideally we should use themes for playing with "better" L&F, not SWT.

vogella:
> @vogella The goal is to allow this for SWT without any help from CSS. Ultimately, I want to get rid of the System theme as you suggested earlier, and to improve the no-theme so I can disable theming with less loss. Good highlighting of the tab is something plain SWT should provide first.


Sounds fine to me, if SWT changes, we may have to update the dark theme or can remove the additional support in CSS for this. Please test this change with the dark theme if you can.

laeubi:
> 2\. This should be done for each platform differently, current proposal is based on Gnome L&F, which is not native for Windows / Mac.

Isn't the `C<Component>` already "non platform specific", thus the "Custom" prefix?

merks:
Note that Excel looks like this:

image

And Thunderbird looks like this:

image

So such bars are on Windows applications.  But I personally like the bar at the top for tabs above  (Thunderbird) and the bar below for tabs at the bottom (Excel)....

iloveeclipse:
> Isn't the C<Component> already "non platform specific", thus the "Custom" prefix?

Yes, it is custom SWT widget, but still it should match the rest of the SWT / current platform.

mickaelistria:
> could you please provide same screenshot with the highlight at top of the tab?

I can, but on newer GTK, the highlight of tab is on the bottom (see attached screenshot); so if the goal is consistency with the OS, then it becomes a challenge.
tabs-newgtk

> it "breaks" the "tab" idea. 

Note that the current proposal doesn't touch the tab borders, it's more a form of underlining the current item. Seeing that, I don't really get what you feel is broken here. Can you please elaborate.

> But! I would avoid adding things to the SWT which do not match "native" L&F, which is here the case. 

The CTabFolder already doesn't match any existing L&F of any OS ;)

> . Native tabs on Windows don't have such "colored bars",

Yet see the screenshots posted by @merks . Some Windows apps do have such bars. On Windows, there is no more concept of native L&F by default, everything is inconsistent and application-specific.

mickaelistria:
> Just look how GTK version switches are used elsewhere in "native" SWT widgets code.

CTabFolder is not native, we cannot invoke reference `org.eclipse.swt.internal.gtk.OS` nor `org.eclipse.swt.internal.gtk.GTK4` in it.

mickaelistria:
I just pushed a new patch that adds the line only to Linux, so Windows and Cocoa are not impacted. People interested in Windows or Cocoa support are recommended to open a separate issue.

vogella:
> > could you please provide same screenshot with the highlight at top of the tab?
> 
> I can, but on newer GTK, the highlight of tab is on the bottom (see attached screenshot); so if the goal is consistency with the OS, then it becomes a challenge. tabs-newgtk

CSS files are OS specific, so maybe reconsider using the existing CSS support for tabs?

> I just pushed a new patch that adds the line only to Linux, so Windows and Cocoa are not impacted.

-1 for having a SWT Linux only solution.

mickaelistria:
> CSS files are OS specific, so maybe reconsider using the existing CSS support for tabs?

I really don't want to rely on theme for CTabFolder to be comfortable. I did it once with the System theme, and the result is that you have yourself argued it should be removed and I'm investing in this topic to follow your advice and allowing to remove the System theme (and to allow me to disable theming without major loss of usability).\
Anyway SWT should work nicely (and I agree closely to native) out of the box without CSS theming. On Linux, the CTabFolder needs to be improved for both goals, and the current patch is an improvement both on "more nicely" and "more native" (just the position of the line can be argued, but at least we go from no-line to a line, so that's progress).

> -1 for having a SWT Linux only solution.

So please have a fight with @iloveeclipse and come back as soon as possible to tell me whom I should obey to.
I would personally be happy with a Linux-only solution here; but would be equally happy if this change can serve other users who've seen this patterns in many other apps.

vogella:
> So please have a fight with @iloveeclipse and come back as soon as possible to tell me whom I should obey to.

So you assume @iloveeclipse will not care of Linux only changes? ;-)

I'm still -1 here for Linux only changes, but I'm fine if you override me or feel like this is a helpful change for SWT on Linux only.  

I guess you will have to make this optional so that existing clients need to opt into this new style. I assume not all SWT clients will like to have their UI changed.

mickaelistria:
> So you assume @iloveeclipse will not care of Linux only changes? ;-)

At least, it covers this half of the change request "This should be done for each platform differently, current proposal is based on Gnome L&F, which is not native for Windows / Mac."
I think there is room for building a consensual strategy on Linux here. I'm not equipped technically and morally to deal with other OS.

> I guess you will have to make this optional so that existing clients need to opt into this new style. I assume not all SWT clients will like to have their UI changed.

I'd rather not make it optional until someone complains. Otherwise, we'll end up putting switch to enable/disable every patch ;)
As mentioned above, I think we can end up with a consensual improvement -maybe not the very perfection of plain native experience- but something that is better than the current state to everyone.

akurtakov:
I have few comments here: 
* CWidgets (CTabFolder, CLabel, CCombo) are worst part of SWT and they don't match anything on any OS to be honest. It's SWTs implementation stuck in the 90s in terms of look. Personally, I never use them and recommend others to not use them when someone asks me - if you want real native LnF these just don't serve well. I'm not going to even start on a11y, performance and etc. 
* Mixture of SWT and IDE - the fact that smth is doable with CSS means nothing for pure SWT apps (e.g. fix in SWT will help BiglyBT and maybe SmartGit (@SyntevoAlex ?) where CSS solution is not possible. So whenever we speak about SWT patch try to think about these use cases too.
* OS specific changes - I really think we are casting out SWT's strength by trying to make things look exactly the same on every OS. The goal of SWT is (and its initial goal was!) to have "native" LnF thus every single bit that could be made to look more native on every platform would better be welcomed or there is no point in developing SWT at all .

laeubi:
> * The goal of SWT is (and its initial goal was!) to have "native" LnF thus every single bit that could be made to look more native on every platform would better be welcomed

I think the problem is that the native widgets are sometimes to restricted (or the implementation/mapping to OS is...?)

> or there is no point in developing SWT at all

Yes maybe SWT is just outdated, don't know, so probably having an os-independet mapping would be a better soloution where all components are owner-drawn :-)

laeubi:
> * where CSS solution is not possible

By the way if I rember right last time when looking into the CSS stuff, GTK has native support for CSS, so probably it would make sense to simply port the CSS support to swt instead of eclipse (e4?)

laeubi:
> just the position of the line can be argued

maybe use a SWT.STYLE bit to control if it is top or bottom? At least that would be "swt style" of doing things, then we can even have:

1. no line
2. top line
3. bottom line
4. top and bottom line

akurtakov:
> > * where CSS solution is not possible
> 
> By the way if I rember right last time when looking into the CSS stuff, GTK has native support for CSS, so probably it would make sense to simply port the CSS support to swt instead of eclipse (e4?)

Unfortunately, the e4 one will have to stay for win/mac. As they don't provide any support for CSS styling (sometimes no styling at all for some widgets) SWT can not provide such API. We have been requesting minimum 2 different WS implementation before exposing as API. 
With all that said if you're Linux only there is a nice hack for you - Widget.setData("org.eclipse.swt.internal.gtk.css", "your_gtk_css_string") - supports everything your GTK version supports incl. animations and etc.  If you want to style the whole app with GTK css file just pass -Dorg.eclipse.swt.internal.gtk.cssFile=your_css_file to the app. One can make wonders that way and totally outshine many UIs without writing single code of line. See https://akurtakov.fedorapeople.org/buttoncss.webm for an old (2016!) presentation . Unfortunately this never made it to be used in IDE due to Win32/Cocoa lacking such capabilities.

laeubi:
> I don't think it's the right place to host theme when native already provides it.

I mean: If gtk support it, it might be good to *emulate* it for win/mac and then have 'native' css support. e.g. if one stores the css in a data `org.eclipse.swt.css` then gtk can use it as-is and win/mac read it for emulating it.

> SWT can not provide such API. We have been requesting minimum 2 different WS implementation before exposing as API

There is always an exception for a rule :-)

laeubi:
Isn't there one in e4 already?

akurtakov:
> Isn't there one in e4 already?

Let's not pollute this PR. I would be more than happy to discuss in separate issue though.

iloveeclipse:
> the current proposal doesn't touch the tab borders, it's more a form of underlining the current item. Seeing that, I don't really get what you feel is broken here. Can you please elaborate.

The UX visual heritage of tabs is to represent some "tab" **connected** to some folder that allows easy way to find right file, see https://www.interaction-design.org/literature/topics/tabs, so drawing a line that **separates** the tab handler from the tab content is against this concept.

Note: I personally would be **always** against such line **on bottom** position and would for sure vote for an option to disable that.

You are coming [from here](https://github.com/eclipse-platform/eclipse.platform.ui/issues/3#issuecomment-1102472718)
> I've thought a bit more about it, as I agree 1 goal is to remove the System theme all together and replace it with native

And since the main driver for you is to get rid of "Native" theme and switch to no theme by default, which I can only support,  let's try something that is not **that** Adwaita like, so it could be accepted by other OS users and doesn't immediately looks alienated.

So I would vote for more subtle line on top of the tab (1/2 of the tab width), drawn as oval to use "fuzziness" effect.

It looks like this on RHEL 7.9:

image


See https://github.com/eclipse-platform/eclipse.platform.swt/commit/2a1582a629835db17f71937fbcf4b37ad59010d1

Note: the patch also fixes tab drawing for tabs on bottom.

mickaelistria:
> How does it look like when 1 of the tabfolder is "active"? That's the main issue with current CTabFolders: when you have multiple CTabFolder, it's not easy to find which CTabFolder is the parent of the current focus/selection.

I tried it and think it's OK. Please submit a PR for it and we can merge it ASAP.

mickaelistria:
> so it could be accepted by other OS users and doesn't immediately looks alienated.

FWIW, compared to screenshots shared by @merks for popular Windows applications, and compared to your screenshots on Linux, and compared to existing Eclipse themes, and compared to Intellij... your proposal looks much more alienated to most users than the current/previous one.
Not that I care much and I'd be happy to see your proposal merged, just highlighting here I don't buy the argument that this looks less alienated, on the contrary. But I've got nothing against aliens, I just want the current CTabFolder to be more visible.

iloveeclipse:
> > so it could be accepted by other OS users and doesn't immediately looks alienated.
> 
> FWIW, compared to screenshots shared by @merks for popular Windows applications, and compared to your screenshots on Linux, and compared to existing Eclipse themes, and compared to Intellij... your proposal looks much more alienated to most users than the current/previous one. 

Let say it this way: **any** change initiated here would look alienated compared to original state with a simple tab and no extras.

I've tried my patch on my Windows notebook and realized that on Windows `drawOval()` seem to produce a **perfect rectangle** (why not), I guess because it is just 2 pixel height and no aliasing enabled. 

The result look like this, and it looks differently and while it improves selected tab discovering, it will for sure make some users unhappy, and some products look strange.

image
 
So with that, I would prefer we would provide a system flag that would allow product owners disable this feature.

Could you please take over my changes related to the placing / size of the highlight, change back `drawOval` to `drawRectangle` and add a kill switch to disable this if a system flag is set?

iloveeclipse:
Couldn't resist: with gerrit I could just push your patchset with my cnanges to gerrit without configuring anything else on my side (sure I can add you repo as remote and try push to it, but I honestly don't see why should I do that with each and every contributor repo where I would like to suggest another patch set)...

laeubi:
> Couldn't resist: with gerrit I could just push your patchset with my cnanges to gerrit without configuring anything else on my side (sure I can add you repo as remote and try push to it, but I honestly don't see why should I do that with each and every contributor repo where I would like to suggest another patch set)...

You do not need to fetch @mickaelistria repo, you can just fetch this PR:

1. Go to Repository perspective
2. Right click on the swt repo
3. Choose "Fetch Github Pull request"
4. paste the URL https://github.com/eclipse-platform/eclipse.platform.swt/pull/183
5. You can now edit the branchname if you like
6. Push that to your fork and open a PR

If you like to add to @mickaelistria PR he needs to enable "allow access for contributors" first, then you could even use the github UI to edit files on his branch.

iloveeclipse:
> > Couldn't resist: with gerrit I could just push your patchset with my cnanges to gerrit without configuring anything else on my side (sure I can add you repo as remote and try push to it, but I honestly don't see why should I do that with each and every contributor repo where I would like to suggest another patch set)...
> 
> You do not need to fetch @mickaelistria repo, you can just fetch this PR:

Guess how I've pushed my commit? But I do **not** want to create **another** PR, that's the point.

laeubi:
> But I do **not** want to create **another** PR, that's the point.

Probably @mickaelistria might not want you to push to his repository, anyways if you are on windows (or mac) you might want to use https://desktop.github.com/ it claims to offer making such things "automatic":

> Checkout branches with pull requests and view CI statuses
> 
> See all open pull requests for your repositories and check them out as if they were a local branch, even if they're from upstream branches or forks. See which pull requests pass commit status checks, too!

And as mentioned else-where, it would be better to request similar features from Egit/Jgit instead of complaining what gerrit has done different :-)

In fact gerrit only allowed such workflows for committer and the user has no way to "opt-out", so everything has pro&con.

mickaelistria:
@iloveeclipse new patch implements your proposal with some fix and extra things I'll leave tests to discover.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
```diff
@@ -1348,16 +1349,28 @@ void drawSelected(int itemIndex, GC gc, Rectangle bounds, int state ) {
...
 
-				//Highlight MUST be drawn before the outline so that outline can cover it in the right spots (start of swoop)
+				//Highlight colors MUST be drawn before the outline so that outline can cover it in the right spots (start of swoop)
 				//otherwise the curve looks jagged
 				drawHighlight(gc, bounds, state, rightEdge);
 
+				// draw highlight marker of selected tab
+				Color previousColor = gc.getBackground();
+				gc.setBackground(item.getDisplay().getSystemColor(parent.shouldHighlight() ? SWT.COLOR_LIST_SELECTION : SWT.COLOR_WIDGET_DISABLED_FOREGROUND));
+				double widthRatio = parent.shouldHighlight() ? 0.8 : 0.25;
```
iloveeclipse, line 1365: Interesting. I like the change in the width, gives nice "animation" effect.
I'm still undecided what is the best width here. With `1 : 0.2` I see "OK" results on Windows and Linux, interestingly "1" doesn't mean, the line is full tab width.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
```diff
@@ -98,6 +98,7 @@ public class CTabFolderRenderer {
 	static final int ITEM_LEFT_MARGIN = 4;
 	static final int ITEM_RIGHT_MARGIN = 4;
 	static final int INTERNAL_SPACING = 4;
+	static final int HIGHLIGHT_BAR_THICKNESS = Boolean.getBoolean("SWT_CTABFOLDER_DISABLE_HIGHLIGHT") ? 0 : 2;
```
iloveeclipse, line 101: The convention for system flags differs from environment variables, so `org.eclipse.swt.CTabFolder.disableHighlight` would be the more consistent key.

iloveeclipse (changes requested):
I've tested on Windows with all themes, and it looks like there are changes needed in most of them.

 - "Light" theme seem to fully override changes for tabs on top, but it does't style tabs on bottom, so only "bottom" tabs are highlighted.
 - "Dark" theme overrides everything and, surprisingly, already shows highlights but as separation between tab and content, which looks wrong.
 - "System" theme looks worst with/without the patch (and somewhat similar to the patch here), but with "1:0.2" width ration in the patch  the result **without** the theme used looks far better than "System" theme.
 - "Classic" theme is the only one where everything works as expected, but there this change is not needed at all, because selected tabs are clearly highlighted there by background color anyway. 

So assuming we will get some agreement on final L&F in SWT here, we should update themes to be consistent and use SWT tab highlighting for all tabs or fully override this patch (for example in the "Classic" theme) or adopt this patch so we give themes use some API to switch this patch off/on.

trancexpress:
I don't like how this looks. I can tell which tab is active just fine without such lines (especially considering that most of the time views are very different). Please make this configurable, so I can turn it off (it will be the first thing I do when I download the newer Eclipse version).

If the lines should exist, IMO they should be consistent with what ```gedit``` shows (I assume that is the "inspiration" for the change). So on my system (RHEL 7.9, GTK 3.22.30-8),  the blue bar should be above the tab. On whichever version the bars are below, should also be below in Eclipse. It should cover the whole tab either way.

Also why is it shown for the non-active part stack? Does ```gedit``` do this?

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
```diff
@@ -1348,16 +1349,28 @@ void drawSelected(int itemIndex, GC gc, Rectangle bounds, int state ) {
...
+				//Highlight colors MUST be drawn before the outline so that outline can cover it in the right spots (start of swoop)
 				//otherwise the curve looks jagged
 				drawHighlight(gc, bounds, state, rightEdge);
 
+				// draw highlight marker of selected tab
+				Color previousColor = gc.getBackground();
+				gc.setBackground(item.getDisplay().getSystemColor(parent.shouldHighlight() ? SWT.COLOR_LIST_SELECTION : SWT.COLOR_WIDGET_DISABLED_FOREGROUND));
+				double widthRatio = parent.shouldHighlight() ? 0.8 : 0.25;
+				if(parent.onBottom) {
+					gc.fillOval((int)(x + 1 /* outline */ + (1 - widthRatio) * width / 2), y + height - 1 - HIGHLIGHT_BAR_THICKNESS, (int)(width * widthRatio), HIGHLIGHT_BAR_THICKNESS);
```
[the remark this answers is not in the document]
mickaelistria: I'm personally fine with rectangles as it was used in the initial patch and would happily revert to it.

akurtakov:
As long as nothing else but "THE PERFECT" solution is accepted we would never move ahead!

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
mickaelistria: OK, let's go for 1 : 0.2 then.

akurtakov:
> > I can't say anything else but - This issue is great example why our UI stays in 90s !!!
> 
> Could you elaborate it a bit more? What exactly is wrong with this issue and/or our UI's?

Cluttered, non-intuitive, ancient looking, slow - just let students use it and talk to them.

akurtakov:
> > Cluttered, non-intuitive, ancient looking, slow - just let students use it and talk to them.
> 
> Well, it is like it is - with given resources it's best we can do. Feel free to give students a link to github repo so they could fix whatever they dislike.

Very few people (if there are such thick skin still!) not only student will lose their time in such kind of discussions. How would anyone join a team which has 4 members with 5 different visions?

mickaelistria:
> "Dark" theme overrides everything and, surprisingly, already shows highlights but as separation between tab and content, which looks wrong.

Not so surprisingly to millions of people who've been using Dark theme for years without complaining. Pragmatically, people are just fine with it.

> we should update themes to be consistent and use SWT tab highlighting for all tabs or fully override this patch (for example in the "Classic" theme) or adopt this patch so we give themes use some API to switch this patch off/on.

The themes do use e4's CTabRendering class. We can most likely find a way to override the tab drawing here; as it's actually already done.

> Please make this configurable, so I can turn it off (it will be the first thing I do when I download the newer Eclipse version).

Already done.

> If the lines should exist, IMO they should be consistent with what gedit shows

This is already discussed above. Please look at history of why we cannot make a custom widget reliably mimic the native behavior and just can do "our best" or take an opinionated decision.

> Also why is it shown for the non-active part stack? Does gedit do this?

Please read the description in linked issue. This issue is more Eclipse workbench-specific as gedit doesn't show multiple Tab groups/folders as siblings on the same window.

> setSelectionBackground() to set a single color or gradient colors
> setSelectionForeground()
> setSelectionBackground(Image image)
> @mickaelistria can any of these be useful to highlight the active tab for you?

As my investment on the System theme earlier and this issue now can hightlight, no, those are not enough. And I don't want more configuration opportunities, more theming an so on. I just want CTabFolder and workbench to be slightly more usable by default without need for configuration.

humphreygao:
Is SWT able to simulate Windows Fluent Design like https://github.com/JFXtras/jfxtras-styles

trancexpress:
> and I can continue ....

And so for my themes/work environments where there is nothing like that, suddenly Eclipse should have it?

But whatever, if you think this will make Eclipse more up-to-date do go on. I can look back at many Eclipse mailing list discussions that were ended with the "we will never move forward if we don't do XYZ" "argument". I'm not seeing Eclipse move forward either way.

akurtakov:
> s nothing like that, suddenly Eclipse should have it?
> 
> But whatever, if you think this will make Eclipse more up-to-date do go on. I can look back at many Eclipse mailing list discussions that were ended with the "we will never move forward if we don't do XYZ" "argument". I'm not seeing Eclipse move forward either way.

Exactly, my point! The issue is not XYZ, the issue is how we get to it - it's so exhausting (not from real work!!!) that no one cares to finish it or drive it further at the end!

iloveeclipse:
> Very few people (if there are such thick skin still!) not only student will lose their time in such kind of discussions. 

Here is the quote from [Code of Conduct](https://www.eclipse.org/org/documents/Community_Code_of_Conduct.php)
> Being respectful of differing viewpoints and experiences

So what kind of discussion would be acceptable for you? The one where everyone immediately agrees with our opinion? That is not a discussion but a master/slave kind of communication.

> How would anyone join a team which has 4 members with 5 different visions?

I though we are all about diversity these days, so one should welcome different visions? 
Or they are only welcome if they are all following the **right** one, which is defined by some higher authority? Well, that's again not the way how open source projects are supposed to work. In a company you might be able to get that kind of work and that will be for sure faster, but here we are all equal and different visions must be considered while taking decisions.

akurtakov:
> > Very few people (if there are such thick skin still!) not only student will lose their time in such kind of discussions.
> 
> Here is the quote from [Code of Conduct](https://www.eclipse.org/org/documents/Community_Code_of_Conduct.php)
> 
> > Being respectful of differing viewpoints and experiences
> 
> So what kind of discussion would be acceptable for you? The one where everyone immediately agrees with our opinion? That is not a discussion but a master/slave kind of communication.

To me discussion means smth quite different than giving mutually exclusive statements.

lshanmug:
Drawing the highlight or not should be configurable via API similar to the other properties in CTabFolder, so that the client code can make the choice on how the CTabFolder will look. For example: What if the user would like to have both underline highlight  and selectionBackground? Or what if user wants to disable the highlight line for a particular CTabFolder. 

We can choose how the CTabFolder looks in Eclipse and have the highlight enabled by default in it. But, CTabFolder is used by other SWT clients as well and should be configurable for them.

mickaelistria:
Patch was updated to add one API and is fully opt-in now.
Companion patch to enable this feature on the workpace is https://github.com/eclipse-platform/eclipse.platform.ui/pull/90

mickaelistria:
> Why is the new API called on CTabFolderRenderer instead of CTabFolder?

Because we want to customize 1 particular renderer instance, we don't want to tell all renderers (eg the e4 CTabRendering) that it should draw that line. But if you prefer we can move it to CTabFolder directly. As long as I'm not mandated to editor e4's CTabRendering by the way, I don't really mind.

> Can we name the API in line with the existing setSelection* APIs which set the appearance of the selected tab too?

OK;

mickaelistria:
@lshanmug next steps a pending on your answer to `But if you prefer we can move it to CTabFolder directly. As long as I'm not mandated to editor e4's CTabRendering by the way, I don't really mind.`

lshanmug:
> @lshanmug next steps a pending on your answer to `But if you prefer we can move it to CTabFolder directly. As long as I'm not mandated to editor e4's CTabRendering by the way, I don't really mind.`

I think we should move it CTabFolder to keep it consistent with rest of the API. But, I've not checked the e4's CTabRendering code to see how/if any changes will be required there.

mickaelistria:
@lshanmug newer patch implements the requested changes.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
```diff
@@ -3470,6 +3471,17 @@ public void setSelectionForeground (Color color) {
 	if (selectedIndex > -1) redraw();
 }
 
+/**
+ * Sets the thickness of the highlight bar on the selected tab.
+ *
+ * @param thickness the desired thickness. Set to {@code 0} to not draw a highlight bar
+ * @since 3.121
+ */
+public void setSelectionBarThickness(int thickness) {
```
lshanmug, line 3480: What are the allowed values for thickness? May be a value between 0 to 3? We should check if thickness is within the allowed range. Currently user code can specify any value including negative values or large values.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
```diff
@@ -449,8 +449,8 @@ protected Rectangle computeTrim (int part, int state, int x, int y, int width, i
 					if (!parent.simple && !parent.single && (state & SWT.SELECTED) != 0) {
 						width += curveWidth - curveIndent;
 					}
-					y = y - ITEM_TOP_MARGIN;
-					height = height + ITEM_TOP_MARGIN + ITEM_BOTTOM_MARGIN;
+					y = y - ITEM_TOP_MARGIN - parent.selectionHighlightBarThickness;
+					height = height + ITEM_TOP_MARGIN + ITEM_BOTTOM_MARGIN + parent.selectionHighlightBarThickness;
```
lshanmug, line 453: Can we draw the bar in the Item Margin instead of increasing the item height? Currently this causes the height of the TabItem header to increase/decrease based on the bar thickness.
AFAICS the current implementation in Eclipse Dark theme also draws in the available space - https://github.com/eclipse-platform/eclipse.platform.ui/blob/master/bundles/org.eclipse.e4.ui.workbench.renderers.swt/src/org/eclipse/e4/ui/workbench/renderers/swt/CTabRendering.java#L584

lshanmug (commented):
Thanks! API looks good (and is different from previous versions). Please see comments below. Also, noticed that the bar on the selected tab in inactive CTabFolder still shows a partial line instead of full line. Is it as intended?

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
mickaelistria: Any positive value should be fine.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
mickaelistria: ok, I'll try that.

mickaelistria:
> Also, noticed that the bar on the selected tab in inactive CTabFolder still shows a partial line instead of full line. Is it as intended?

I'll revert to a whole line.

mickaelistria:
@lshanmug i implemented the suggested changes.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
```diff
@@ -3470,6 +3471,18 @@ public void setSelectionForeground (Color color) {
 	if (selectedIndex > -1) redraw();
 }
 
+/**
+ * Sets the thickness of the highlight bar on the selected tab.
+ *
+ * @param thickness the desired thickness. Most be positive and lower that {@link CTabFolderRenderer#ITEM_TOP_MARGIN} for correct results.
```
niraj-modi, line 3477: lower that >> lower than

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
```diff
@@ -189,6 +189,7 @@ public class CTabFolder extends Composite {
 	boolean selectionGradientVertical;
 	Color selectionForeground;
 	Color selectionBackground;
+	int selectionHighlightBarThickness = 0;
```
niraj-modi, line 192: - Suggest we keep default value as 1(or something which enables this feature by default)
- Default value can be revised based on feedback

niraj-modi (changes requested):
- "org.eclipse.swt/pom.xml" also needs to be updated for bundle version similar to MANIFEST.MF

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
mickaelistria: Made it 2 as 1 pixel is to thin to easily distinguish the different colors.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
lshanmug: The allowed values are still not clear. CTabFolderRenderer#ITEM_TOP_MARGIN is 2 so going by javadoc they values can be only 0 and 1, but I think that's not intended. It's better to sanity check the thickness (especially for negative and large values) to prevent unexpected behavior.

lshanmug (commented):
Tested the changes on Mac and looks good. Please see comments inline. 
P.S: Sorry for the delay (I'm on leave most of this week)

iloveeclipse (dismissed):
The top highlight line is decent now, but there is now a 1 pixel gray line between tab handle and tab content (screenshot taken on RHEL 7 with no CSS themes enabled). This is not nice and should be fixed.

image

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
lshanmug: I'm still not sure about pushing the changes to the widget by default for the same reasons (as discussed before). But, may be we can keep it on for now to get some testing and feedback.

mickaelistria:
> it can be fixed with this patch?

I don't think it's possible to dynamically decide whether it's best to show it or not from according to the theme in the SWT layer. So I don't think we can fix it as part of this patch.

> Do you plan to fix Classic theme after merge

I'll try to provide a patch for this soon. We can wait for it before merging.

mickaelistria:
> I don't think it's possible to dynamically decide whether it's best to show it or not from according to the theme in the SWT layer. So I don't think we can fix it as part of this patch.

I was wrong, the state for "classic" tabs is in CTabFolder (`setSimple(...)`). So we can skip the highlight for that case.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
```diff
@@ -3470,6 +3471,22 @@ public void setSelectionForeground (Color color) {
 	if (selectedIndex > -1) redraw();
 }
 
+/**
+ * Sets the thickness of the highlight bar on the selected tab. The highlight bar is drawn in the top margin of the selected tab.
+ *
+ * @param thickness the desired thickness. Most be positive and lower than {@link CTabFolderRenderer#ITEM_TOP_MARGIN} (that is either {@code 0} {@code 1} or {@code 2} at the moment),for correct results.
+ *                  Set to {@code 0} to not draw a highlight bar.
+ * @implNote Default {@link CTabFolderRenderer} currently ignores this setting if {@link #getSimple()} is {@code true}.
+ * @since 3.121
```
niraj-modi, line 3480: New API Javadoc need to call-out the exception:
<details>
<summary>Snippet</summary>

```java
 * @exception SWTException <ul>
 *    <li>ERROR_INVALID_ARGUMENT - if called with a thickness of less than 0</li>
 * </ul>
 *
 * @since 3.121
 */
```
</details>

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
```diff
@@ -1354,10 +1354,18 @@ void drawSelected(int itemIndex, GC gc, Rectangle bounds, int state ) {
 					}
 				}
 
-				//Highlight MUST be drawn before the outline so that outline can cover it in the right spots (start of swoop)
+				//Highlight colors MUST be drawn before the outline so that outline can cover it in the right spots (start of swoop)
 				//otherwise the curve looks jagged
 				drawHighlight(gc, bounds, state, rightEdge);
 
+				// draw highlight marker of selected tab
+				if (parent.simple) {
```
niraj-modi, line 1362: Draw highlight marker only when thinkness is greater than 0:
if (parent.selectionHighlightBarThickness > 0 && parent.simple) {

niraj-modi (changes requested):
Overall behavior-wise looks good now:
- Suggested some possible improvements for code/javadoc inline.

niraj-modi (approved):
Looks good now.

lshanmug:
@mickaelistria  Please add an N&N entry for the new feature/API.

mickaelistria:
See https://github.com/eclipse-platform/www.eclipse.org-eclipse-news/pull/44

jukzi:
I did not follow the discussion but it feels inconsistently to see 
1. those bars in eclipse only in very few Dialogs (Search+Launch configuration+??) but not in View Tabs
2. the dashed selection box is drawn at the same location as the blue bar .. ugly (and XORed to red??)
Example (windows):
image
The Dashes on a selected Button look much better aligned to the blue border:
image

Overall it is a feature that i never missed, since using the keyboard navigations the dashes already showed the active selection, and the selected tab already had a higlighted background which was always enough.

PR #188, opened 2022-06-02
[GTK4] Various Shell fixes to allow more snippets to work

joel-majano:
Continues fixes for https://bugs.eclipse.org/bugs/show_bug.cgi?id=567333

- guarded call to gtk_window_set_keep_above as API not available in GTK4
- removed modification to allocation.height in forceResize as it was causing wrong behaviour when displaying contents
- Solved gtk size allocation errors by taking into account the "y" position of the widget (gtk appears to be subtracting it from the height requested)
- guarded calls to gtk3 only functions as gtk4 snippets appear to work fine without it (will be monitoring this fix to make sure it didnt break something else)

Tested with Snippet 50, 63, 295, 233, 104, 99, 248, 4


akurtakov:
Thanks Joel!

PR #192, opened 2022-06-04
Version bumps for 4.25 stream.

akurtakov:
Tracked in
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/313


akurtakov:
Verification can't succeed until versions in the 2 swt repos are aligned.

PR #195, opened 2022-06-06
Don't check LocationListener changing js eval on WebkitGTK

akurtakov:
It's same backend (Webkit) as MacOS and behaves in the same way.


akurtakov:
Unrelated test failures.

PR #196, opened 2022-06-06
Improve tests that fail on jenkins

akurtakov:
Preliminary changes to get a bit saner tests before looking why they fail on Jenkins.


akurtakov:
Test failures are what caused this cleanup initially.

PR #216, opened 2022-06-20
Don't execute listeners on disposed widgets and items.

nnemkin:
The problem occurs when there are multiple listeners of the same type and one of them triggers widget dispose (either directly or via its parent).

Note that SWT.Dispose events are not affected, because they are sent before the widget is marked as disposed.

Inspired by eclipse-platform/eclipse.platform.ui#164.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/widgets/EventTable.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/widgets/EventTable.java
@@ -90,6 +90,10 @@ public void sendEvent (Event event) {
 					} catch (Error | RuntimeException ex) {
 						exceptions.stash (ex);
 					}
+					if (event.widget != null && event.widget.isDisposed()
+							|| event.item != null && event.item.isDisposed()) {
+						break;
+					}
 				}
 			}
 		}


vogella (approved):

iloveeclipse:
Haven't tried in debugger, but wouldn't this mean, from two dispose listeners only the first one will get disposed event?

nnemkin:
> Haven't tried in debugger, but wouldn't this mean, from two dispose listeners only the first one will get disposed event?

All `Dispose` listeners are executed as expected, because `widget.isDisposed()` becomes `true` later in the dispose cycle, after the events are sent. (I've tested it, just to be sure.)

The only exception to this is a `Dispose` listener itself calling `widget.dispose()`. Then the following listeners will be skipped.

iloveeclipse:
> The only exception to this is a `Dispose` listener itself calling `widget.dispose()`. Then the following listeners will be skipped.

OK, this would be a bug anyway.

niraj-modi (changes requested):
Suggest we add a new JUnit test-case covering this particular scenario.

lshanmug (changes requested):
I'm afraid this will be a breaking change in listener APIs. 
`Widget.addListener()` - adds the listener to the collection of listeners who will be notified when an event of the given type occurs. 
`Widget.notifyListeners()` - Notifies all of the receiver's listeners for events of the given type that one such event has occurred by invoking their handleEvent() method.

When an event is sent, **\*all\*** the added listeners are expected to be notified, even if the widget is disposed in one of the listeners. With this change, if the widget is disposed in one of the listeners, the following listeners of the event will not be notified causing unexpected behavior. 
For example, a listener client code has some logic to execute if the widget is disposed, but with this change, the listener client code may never be called.

nnemkin:
> When an event is sent, *all* the added listeners are expected to be notified, even if the widget is disposed in one of the listeners.

I don't see how this follows. When a widget is disposed, it's gone. Delivering events to something that doesn't exist anymore makes no sense.

> For example, a listener client code has some logic to execute if the widget is disposed, but with this change, the listener client code may never be called.

I'm pretty sure that there are no such listeners anywhere in the world. Also, there is a dedicated event for tracking disposal (`SWT.Dispose`).

lshanmug:
> > When an event is sent, _all_ the added listeners are expected to be notified, even if the widget is disposed in one of the listeners.
> 
> I don't see how this follows. When a widget is disposed, it's gone. Delivering events to something that doesn't exist anymore makes no sense.

It follows from the javadoc as mentioned before: 
> `Widget.addListener()` - adds the listener to the collection of listeners who will be notified when an event of the given type occurs. 
`Widget.notifyListeners()` - Notifies all of the receiver's listeners for events of the given type that one such event has occurred by invoking their handleEvent() method.

It's not delivering events to the disposed widget. It's only notifying **all** the listeners about a event that has occurred on the widget. 
The widget may be disposed in one of the listeners, but that shouldn't matter if the other listener of the event is going to perform further action based on event notification and not necessarily on the disposed widget.

merks:
A change such as this one is more likely to cause a new problem than it is to fix any existing problem.   In fact I don't see that it fixes any reported problem and there appears to be no test case to test this new-but-questionable-behavior...

PR #226, opened 2022-06-23
[GTK4] TabFolder wrong background color fix (Snippet 41 fixes)

joel-majano:
- Removed setInitialBackgroundGTK4 in TabFolder.java

In Gerrit change https://git.eclipse.org/r/c/platform/eclipse.platform.swt/+/188699, the initial background of the tabfolder "stack" was changed to the default SWT Background color. This would be correct in ControlExample since in GTK3 the folder background is the SWT default, but in a snippet with only a tabfolder, the background is white.  That issue appears to be more theme related as backgrounds are not explicitly set for TabFolders in SWT-GTK3. This patch reverts that change, and Snippet 41 now displays correctly.


akurtakov:
What happens in ControlExample?

joel-majano:
> What happens in ControlExample?

ControlExample will now show a white background in the TabFolders again. Snippet 41 was showing the wrong background color (should be white) in GTK4 due to this previous change. I'll need to reopen the previous bug about ControlExample's background.

akurtakov (approved):

Issue #48, opened 2022-04-14
[GTK4] FileDialog created with SWT.OPEN displays error if no FilterPath set


joel-majano:
If a FileDialog is created with SWT.OPEN but setFilterPath is not used to set a path other than the default empty string, the error: "The folder contents could not be displayed" "Operation not supported" is shown.

This does not happen with FileDialog created with SWT.SAVE or if setFilterPath sets the path to something valid. 

Tested using Snippet72 and commenting out all relevant lines that set filter related fields. 

Tracked in #228

joel-majano:
This further only happens if a file name is not set using setFileName.

akurtakov:
This one has been fixed.

PR #229, opened 2022-06-23
Issue #48 FileDialog created displays error if no fileFilter Path set

jasonhy-wang:
+if no filename set defaults to location of file

The issue was caused by if no filename was set then it would try to open file "" as a default which creates an error. This is simply fixed by if no filename is set then set the filename to location of the file

Signed-off-by: Jason Wang <jaswan@redhat.com>

Original Issue #48

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/FileDialog.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/FileDialog.java
@@ -479,8 +479,6 @@ void presetChooserDialog () {
 
 				if (fileName.length() > 0) {
 					GTK4.gtk_file_chooser_set_file (handle, file, 0);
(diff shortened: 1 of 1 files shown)


joel-majano (approved):
Verified using Snippet 48, ClipboardExample "Choose file" dialog and TextEditor.java "Open" feature. Issue is fixed.

akurtakov:
This feels more like workaround rather than bugfix. IMHO a better fix would be needed that properlty sets the chooser up to not require this hack. Can you reproduce the issue in pure C code?

iloveeclipse (dismissed):
Sorry, it looks like you changed way more files as needed, most likely due some bad merge operation.

Please only push **one** commit with only changes you really wanted to make.

akurtakov:
@iloveeclipse as @jasonhy-wang rebased properly I'll dismiss your change request to not block it further.

akurtakov (approved):

Issue #233, opened 2022-06-29
Add API to set labels for MessageBox buttons


akurtakov:
In https://github.com/eclipse-equinox/p2/pull/97 pointed a missing API in MessageBox which would allow to not use JFace's MessageDialog/PlainMessageDialog which are no used in case when one just wants to show e.g. "Revert" instead of "OK". By providing such an API developers can go back to use SWT's MessageBox directly and thus let the OS decide whether to show icon, button order and so on.
I've been thinking of API like:
```
public void setStandardButtonLabels(Map<Integer buttonId, String buttonLabel) {}
```
that would allow using it like "messageBox.setStandardButtonLabels(Map.of(SWT.OK,ProvUIMessages.RevertProfilePage_RevertLabel, SWT.CANCEL, ProvUIMessages.RevertDialog_CancelButtonLabel));"
I'll provide a proof of concept for Linux.

akurtakov:
Screenshot on Linux/Gnome with the change: 
![image](https://user-images.githubusercontent.com/574788/176614385-a52a62f7-5974-4b48-a713-502e76d5d93c.png)

niraj-modi:
> Thanks for testing @sravanlakkimsetti . So we need win32 implementation.

@akurtakov 
Windows API changes are also ready: #242 
Screenshot on Windows:
![image](https://user-images.githubusercontent.com/92304759/178258088-21462cb9-d301-4ee1-9997-9dee085e8773.png)

akurtakov:
Thanks, Niraj. I'll apply your patch in my PR and amend it so we can have all implementations in one go if that's fine with you.

PR #234, opened 2022-06-29
[WIP] Add API to set labels for MessageBox buttons

akurtakov:
Proof of concept for
https://github.com/eclipse-platform/eclipse.platform.swt/issues/233 .
Implemented only on GTK for now but having similar implementation for
Cocoa would be easy.


akurtakov:
Windows implementation might be trickier as I don't see SWT.getMessage("SWT_OK") used in that part. @niraj-modi  can you give some hints?

niraj-modi:
> Windows implementation might be trickier as I don't see SWT.getMessage("SWT_OK") used in that part. @niraj-modi can you give some hints?

Unlike GTK/Cocoa: Windows controls the [MessageBox](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-messageboxexw) button-text and it's language translation as well.
Only predefined icons/buttons are supported which are passed as MessageBox style bits currently.

IMO, since we have implementation for two platforms:
- New API MessageBox#setButtonLabels() can still go in with JavaDoc hint for platform support/limitation and Windows implementation remains as no-op.

@SyntevoAlex : WDYT ?

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/MessageBox.java
```diff
@@ -263,4 +273,12 @@ private static int checkStyle (int style) {
 	style = (style & ~mask) | SWT.OK;
 	return style;
 }
+
+/**
+ * @since 3.121
+ */
```
niraj-modi, line 279: - JavaDoc for new API needed.

niraj-modi (changes requested):

SyntevoAlex:
For Windows, I see three paths:
1) Replace older `MessageBox` WINAPI with newer [`TaskDialogIndirect`](https://docs.microsoft.com/en-us/windows/win32/api/commctrl/nf-commctrl-taskdialogindirect). This sounds like the best approach.
2) Use `SetWindowsHookEx(WH_CBT)` temporarily to catch `HCBT_ACTIVATE` and use `SetWindowText()` to replace button texts.
3) Stop using WINAPI and fully custom-implement `MessageBox`. Not recommended: there will be various compatibility issues (such as where text was wrapped before and after), and WINAPI has non-obvious features (such as Ctrl+C copying entire content of the `MessageBox`, which is super useful)

Unfortunately I won't be able to spend time to provide an implementation.

niraj-modi:
> For Windows, I see three paths:
> 
>     1. Replace older `MessageBox` WINAPI with newer [`TaskDialogIndirect`](https://docs.microsoft.com/en-us/windows/win32/api/commctrl/nf-commctrl-taskdialogindirect). This sounds like the best approach.
> 
>     2. Use `SetWindowsHookEx(WH_CBT)` temporarily to catch `HCBT_ACTIVATE` and use `SetWindowText()` to replace button texts.
Thanks @SyntevoAlex  for your inputs, I could add the new API to Windows implementation based on second approach:
https://github.com/eclipse-platform/eclipse.platform.swt/pull/242

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/MessageBox.java
akurtakov: Done.

akurtakov:
@sravanlakkimsetti @lshanmug  @niraj-modi Please review this PR . It's finally ready for full review and eventually merge.

niraj-modi (approved):
Looks good.

sravanlakkimsetti (approved):

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/MessageBox.java
```diff
@@ -269,6 +272,13 @@ public int open () {
 	return userResponse;
 }
 
+private NSString getLabelForButton(int buttonId, String messageId) {
+	if (labels !=null && labels.containsKey(buttonId)) {
+		return NSString.stringWith(labels.get(buttonId));
```
lshanmug, line 277: API looks good. It'll be good to have a null check here in case returned label is null.

akurtakov: There is already null check and one can set label to anything incl. null, empty string, any whitespace character and etc. and all would not make sense . We are also not making any such check in e.g. Button.setText

akurtakov:
As overall there is agreement and comments are handled I plan on merging this on and handle further requests separately so there is progress shortly.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/MessageBox.java
lshanmug: If labels.get(buttonId) returns null, NSString.stringWith() results in NPE. In case of Button.setText, it doesn't accept null string.

PR #236, opened 2022-06-29
[GTK4] Snippet 282 needs clarity improvements

joel-majano:
This snippet currently does not provide any visual cue to let the user know that the image was pasted correctly.
For the user, it looks like the image did not change at all if the snippet works correctly.

With this patch, a simple console message is printed when an image is succesfully copied, or when it fails.

### DIFF
--- a/examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet282.java
+++ b/examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet282.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016 IBM Corporation and others.
+ * Copyright (c) 2016, 2022 IBM Corporation and others.
  *
  * This program and the accompanying materials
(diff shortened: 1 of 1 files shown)


examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet282.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016 IBM Corporation and others.
+ * Copyright (c) 2022 IBM Corporation and others.
```
akurtakov, line 2: This is wrong change of Copyright - It should  become 2016, 2022

akurtakov (changes requested):

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet282.java
joel-majano: Made this change.

akurtakov (approved):

PR #238, opened 2022-06-30
Part4: Table remains visible though setVisible(false) was called

niraj-modi:
- Fix Eclipse code completion popup which is currently empty unless
re-triggered.

Change-Id: I2c985d221d2d8048e5ad63a184145bff63906b7c


niraj-modi (commented):
Tested code completion in Eclipse self-hosted mode.

PR #242, opened 2022-07-04
Add API to set labels for MessageBox buttons

niraj-modi:
- Windows Implementation for new API MessageBox#setButtonLabels()
#233

Change-Id: I3a3c70c31475e378c1b54d09ac540104314988a5

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MessageBox.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MessageBox.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2015 IBM Corporation and others.
+ * Copyright (c) 2000, 2022 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -14,7 +14,10 @@
 package org.eclipse.swt.widgets;
 
 
+import java.util.*;
+
 import org.eclipse.swt.*;
+import org.eclipse.swt.internal.*;
 import org.eclipse.swt.internal.win32.*;
 
 /**
@@ -42,6 +45,10 @@ import org.eclipse.swt.internal.win32.*;
  */
 public  class MessageBox extends Dialog {
 	String message = "";
+	private int buttonBits = 0;
+	private long cbtHook;
+	private LinkedList<Long> hwdButtons = new LinkedList<>();
+	private Map<Integer, String> labels;
 
 /**
  * Constructs a new instance of this class given only its parent.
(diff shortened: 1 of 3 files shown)


niraj-modi (commented):
Windows API implementation is based on below discussion:
https://github.com/eclipse-platform/eclipse.platform.swt/pull/234#issuecomment-1173912554

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MessageBox.java
```diff
@@ -99,6 +106,88 @@ public MessageBox (Shell parent) {
 public MessageBox (Shell parent, int style) {
 	super (parent, checkStyle (parent, checkStyle (style)));
 	checkSubclass ();
+	/* Create the callback hook */
+	int threadId = OS.GetCurrentThreadId ();
+	Callback cbtCallback = new Callback (this, "CBTProc", 3); //$NON-NLS-1$
```
SyntevoAlex, line 111: This is a leak, and a pretty bad one, because the limit for Callbacks is quite low (see `MAX_CALLBACKS`)

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MessageBox.java
```diff
@@ -99,6 +106,88 @@ public MessageBox (Shell parent) {
...
+	cbtHook = OS.SetWindowsHookEx (OS.WH_CBT, cbtCallback.getAddress (), 0, threadId);
```
SyntevoAlex, line 112: I would rather set hook immediately before `OS.MessageBox()`

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MessageBox.java
```diff
@@ -99,6 +106,88 @@ public MessageBox (Shell parent) {
...
+}
+
+long CBTProc(long nCode, long wParam, long lParam) {
+	char[] buffer = new char[128];
+	int length = OS.GetClassName(wParam, buffer, buffer.length);
+	String className = new String(buffer, 0, length);
+	if (labels != null) {
+		switch ((int) nCode) {
+		/* Cache the button handles in the order of creation */
```
SyntevoAlex, line 121: Use `GetDlgItem()` instead. See this for an example:
https://www.codeguru.com/windows/messagebox-with-custom-button-captions/

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MessageBox.java
```diff
@@ -99,6 +106,88 @@ public MessageBox (Shell parent) {
...
+		case OS.HCBT_CREATEWND: {
+			if (className.equals("Button")) { //$NON-NLS-1$
+				hwdButtons.add(wParam);
+			}
+			break;
+		}
+		/* The system is about to activate a window */
+		case OS.HCBT_ACTIVATE: {
+			if (className.equals("#32770")) { //$NON-NLS-1$
```
SyntevoAlex, line 130: Not sure if this is needed.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MessageBox.java
SyntevoAlex: Also, this is another leak. I suggest to call `UnhookWindowsHookEx()` immediately after the button labels are set.

SyntevoAlex (changes requested):
I had a quick glance only, could do a proper review tomorrow.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MessageBox.java
niraj-modi: After `UnhookWindowsHookEx()` called `cbtCallback.dispose()` to avoid this leak.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MessageBox.java
niraj-modi: Done.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MessageBox.java
niraj-modi: Done.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MessageBox.java
niraj-modi: Removed.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MessageBox.java
```diff
@@ -199,7 +266,16 @@ public int open () {
...
+
+	/* Create the callback hook */
+	Callback cbtCallback = null;
+	if (labels != null) {
+		cbtCallback = new Callback (this, "CBTProc", 3); //$NON-NLS-1$
+		cbtHook = OS.SetWindowsHookEx (OS.WH_CBT, cbtCallback.getAddress (), 0, OS.GetCurrentThreadId ());
+	}
+
 	int code = OS.MessageBox (hwndOwner, buffer1, buffer2, bits);
+	if (cbtCallback != null) cbtCallback.dispose();
```
SyntevoAlex, line 278: Just for the extra safety, I would `OS.UnhookWindowsHookEx()` here (just before disposing callback) as well if it wasn't uninstalled before. Otherwise, the hook might remain installed if something goes contrary to expected (for example for whatever reason `HCBT_ACTIVATE` never arrives). This would in turn cause a JVM crash when it tries to call a disposed callback.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MessageBox.java
```diff
@@ -101,6 +107,68 @@ public MessageBox (Shell parent, int style) {
...
+		case OS.MB_ABORTRETRYIGNORE:
+			setButtonText(wParam, SWT.ABORT, OS.IDABORT);
+			setButtonText(wParam, SWT.RETRY, OS.IDRETRY);
+			setButtonText(wParam, SWT.IGNORE, OS.IDIGNORE);
+			break;
+		default:
+			break;
+		}
+		/* Exit the CBT hook */
+		OS.UnhookWindowsHookEx(cbtHook);
```
SyntevoAlex, line 138: I understand that you copied the comment from the example, but it's not very useful. I'd just delete it.

The bigger problem here is that you call `OS.CallNextHookEx()` on an already destroyed hook instead of just returning. Note that the example had a different problem - it didn't call the next hook, which may be present. The proper approach would be to call next hook and THEN uninstall and return.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MessageBox.java
```diff
@@ -101,6 +107,68 @@ public MessageBox (Shell parent, int style) {
 	checkSubclass ();
 }
 
+long CBTProc(long nCode, long wParam, long lParam) {
+	if (labels != null && (int) nCode == OS.HCBT_ACTIVATE) {
+		/* The system is about to activate MessageBox window */
+		switch (buttonBits) {
+		case OS.MB_OKCANCEL:
+			setButtonText(wParam, SWT.CANCEL, OS.IDCANCEL);
```
SyntevoAlex, line 115: Just a small rant, feel free to ignore: I would rather not use case fallthrough because it makes code less readable.

Also note that instead of `buttonBits` it was wiser to just use `this.style` and avoid `switch` completely.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MessageBox.java
```diff
@@ -101,6 +107,68 @@ public MessageBox (Shell parent, int style) {
...
+		}
+		/* Exit the CBT hook */
+		OS.UnhookWindowsHookEx(cbtHook);
+	}
+	return OS.CallNextHookEx(cbtHook, (int) nCode, wParam, lParam);
+}
+
+void setButtonText(long wParam, int style, int id) {
+	if (labels.get(style) != null) {
+		long hwdButton = OS.GetDlgItem(wParam, id);
```
SyntevoAlex, line 145: did you intend to name it hw**n**d ?

SyntevoAlex (changes requested):
Yes, hooks are hard. One of the reasons I preferred the other solution.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MessageBox.java
niraj-modi: Done.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MessageBox.java
niraj-modi: Done.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MessageBox.java
niraj-modi: Done.

niraj-modi:
> Yes, hooks are hard. One of the reasons I preferred the other solution.

Thanks @SyntevoAlex, appreciate the detailed review here.

SyntevoAlex (approved):
Looks good to me.

Issue #225, opened 2022-06-23
[GTK4] Snippet 41 has various bugs (TabFolder color, Shell tooltips not working)


joel-majano:
**Describe the bug**
Under GTK4, Snippet 41 has the wrong background color on its TabFolder and no shell tooltips when hovered over the window or the tabfolder shell.

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [x] All OS
- - [ ] Windows
- - [x] Linux
- - [ ] macOS 

2. Additional OS info (e.g. OS version, Linux Desktop, etc)
Fedora 36, Gnome 42.2, GTK 4.6.5

3. JRE/JDK version
OpenJDK 17

akurtakov:
Just tested and this one is no longer reproducible with master.

PR #243, opened 2022-07-05
[GTK4] Fix for shell tooltips not working

joel-majano:
In GTK4 there was a missing function in gtk4_enter_event that was called in the equivalent GTK3 function.
This is the initial shell tooltip fix, but right now TabFolder tooltips aren't behaving correctly.

Partly fixes for #225


akurtakov (approved):
Good catch!

PR #245, opened 2022-07-07
Support 16bpp ImageData (for images such as 16bpp grayscale TIFF)

### LINKED ISSUE #232: Support 16bpp palette images (such as 16bpp grayscale TIFF)
Currently, 16bpp palette images fail on Windows and on macOS.

On Windows, the image loads, but the colors are all wrong.
For example, here's the actual image:
image
But SWT shows it this way:
image

On macOS, image fails to load with exception `SWT.ERROR_UNSUPPORTED_DEPTH`:
```
Caused by: org.eclipse.swt.SWTException: code=38
	at org.eclipse.swt.SWT.error(SourceFile:4918)
	at org.eclipse.swt.SWT.error(SourceFile:4833)
	at org.eclipse.swt.SWT.error(SourceFile:4804)
	at org.eclipse.swt.graphics.Image.createRepresentation(SourceFile:1002)
	at org.eclipse.swt.graphics.Image.init(SourceFile:1403)
	at org.eclipse.swt.graphics.Image.<init>(SourceFile:573)
```

SyntevoAlex:
Fixes #232


SyntevoAlex:
New unit tests fail on GItHub because I added `org.eclipse.swt.graphics.ImageDataTestHelper` into SWT package in JUnit project. As a result, classes from the same package are in two different jars, and Java throws `java.lang.SecurityException: class "org.eclipse.swt.graphics.ImageDataTestHelper"'s signer information does not match signer information of other classes in the same package`.

I tried to disable security manager in tests, but didn't succeed so far.

The patch itself is good.

niraj-modi (commented):
My initial impressions:
- The patch works as intended on Win10 as tested with the associated test snippet `Issue0232_16bppIndexedImageData.java`
- The new tests `test_blit` and `test_blit_MsbLsb` too passing locally.

SyntevoAlex:
I removed the commit that tried (but did not succeed) to get JUnit to work on GitHub server. I'll try to fix it in the next days, but if I don't succeed, I suggest to merge as is for now. Because actual patch is good, and JUnit is good when run locally. The problem is only in JUnit on GitHub, and it's not exactly the problem of my patch.

SyntevoAlex:
The problem with unit tests is that GitHub tests published SWT instead of SWT submitted in the PR (see #251). Therefore, it's a completely different problem and this PR should be merged independently.

SyntevoAlex:
Yes, not everything is supported. My goal was limited to making the image load and show. Our product doesn't need saving images, so I won't be able to spend time implementing the rest, sorry.

niraj-modi:
> Tested on Mac with the ImageAnalyzer example. Opening the image file Issue0232_16bppTiff works file. But, saving the file with ImageLoader.save() is not supported yet. Tying to save with Save or Save As menu fails with Unsupported or unrecognized format error.
On saving 16bpp Tiff I notice the same error on Windows too.

niraj-modi (approved):
Approving for Windows:
- this patch enables Opening/Loading operation of 16bpp Tiff format.

SyntevoAlex:
> It'll be good to mention what is supported in the commit message for clarity.

Done.

SyntevoAlex:
I tested on Ubuntu 21.04:
* Tested with my snippet
* Tested with our product
* Tested with Eclipse 2022-06 (created a template plugin project and debugged it)

I didn't encounter any problems.

Issue #251, opened 2022-07-11
GitHub unit tests are testing irrelevant swt.jar


SyntevoAlex:
** Summary **
When GitHub is running unit tests for a pull request, it tests latest published SWT instead of SWT in the pull request!

** Analysis **
In #245 I ran into a problem where GitHub side unit tests fail with error `java.lang.SecurityException: class "org.eclipse.swt.graphics.ImageDataTestHelper"'s signer information does not match signer information of other classes in the same package`.

Eventually I added this code in my unit test
```
Class swtClass = SWT.class;
Class tstClass = Test_org_eclipse_swt_SWT.class;
URL swtURL = swtClass.getProtectionDomain().getCodeSource().getLocation();
URL tstURL = tstClass.getProtectionDomain().getCodeSource().getLocation();
```

and ran maven locally (where the problem is also reproducible). The output was
```
SWT: (file:/C:/Users/User/.m2/repository/p2/osgi/bundle/org.eclipse.swt.win32.win32.x86_64/3.121.0.v20220701-1002/org.eclipse.swt.win32.win32.x86_64-3.121.0.v20220701-1002.jar [
  Version: V3
  Subject: EMAILADDRESS=webmaster@eclipse.org, CN="Eclipse.org Foundation, Inc.", OU=IT, O="Eclipse.org Foundation, Inc.", L=Ottawa, ST=Ontario, C=CA
  Signature Algorithm: SHA256withRSA, OID = 1.2.840.113549.1.1.11
  <...>])
TST: (file:/C:/Path/eclipse.platform.swt/tests/org.eclipse.swt.tests/target/classes/ <no signer certificates>)
```

That is, when running unit tests, maven uses a published SWT instead of the one built from the repo!

iloveeclipse:
@laeubi : FYI. 
If it can't be fixed, we should remove the githib test matrix from PR's, otherwise it suggests something that is completely irrelevant.

SyntevoAlex:
I submitted #252 with a unit test to demonstrate the problem. In its GitHub tests, I can see the problem being reported in the new test:

```
java.lang.AssertionError: Tested SWT path is on maven cache: (

[...]

PR #252, opened 2022-07-11
Issue #251 - Unit test to ensure that GitHub tests currently built SWT

SyntevoAlex:
A unit test to show (and prevent in future) #251

### DIFF
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_SWT.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_SWT.java
@@ -24,6 +24,13 @@ import org.eclipse.swt.SWTError;
 import org.eclipse.swt.SWTException;
 import org.junit.Test;
 
+import java.nio.file.Paths;
+import java.security.CodeSigner;
+import java.security.cert.Certificate;
+import java.security.cert.X509Certificate;
+import java.util.ArrayList;
+import java.util.List;
+
 /**
  * Automated Test Suite for class org.eclipse.swt.SWT
  *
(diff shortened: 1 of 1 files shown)


tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_SWT.java
```diff
@@ -117,6 +119,51 @@ public void test_getMessageLjava_lang_String() {
...
+
+	String swtPath = pathFromClass(SWT.class);
+	String tstPath = pathFromClass(Test_org_eclipse_swt_SWT.class);
+
+	// It's hard to devise a condition that would hold for all possible
+	// locations of compiled classes. For example, my own project uses
+	// output directories different from what Eclipse does by default.
+
+	// One thing for sure, SWT shouldn't be on maven cache!
+	if (swtPath.contains("/.m2/")) {
```
laeubi, line 155: I think this needs be more specific like 
```suggestion
	if (swtPath.contains("/.m2/repository/p2/osgi/bundle/")) {
```

SyntevoAlex: `/.m2/repository` part is probably good, and I'm not sure about the rest. I think it's best to just leave it as is for now, and if it causes false positives, it can be improved later when it's clear when exactly this could happen.

laeubi: No, it is fine that the jar comes from the m2 repository, but it should not come from P2 (p2 artifacts are stored under` p2/osgi/bundle/`).

> if it causes false positives

It already does, see the Jenkins-job where it actually is at the expected location:

/home/vnc/.m2/repository/org/eclipse/swt/org.eclipse.swt.gtk.linux.x86_64/3.121.0-SNAPSHOT/org.eclipse.swt.gtk.linux.x86_64-3.121.0-SNAPSHOT.jar

SyntevoAlex:
> the problem also show up in the jenkins:
> https://ci.eclipse.org/platform/job/eclipse.platform.swt/job/PR-252/4/testReport/junit/org.eclipse.swt.tests.junit/Test_org_eclipse_swt_SWT/test_isLocal/

Hmm, the linked test results say

```
Tested SWT path is on maven cache: (
SWT='/home/jenkins/agent/workspace/eclipse.platform.swt_PR-252/eclipse.platform.swt/home/vnc/.m2/repository/org/eclipse/swt/org.eclipse.swt.gtk.linux.x86_64/3.121.0-SNAPSHOT/org.eclipse.swt.gtk.linux.x86_64-3.121.0-SNAPSHOT.jar'
Test='/home/jenkins/agent/workspace/eclipse.platform.swt_PR-252/eclipse.platform.swt/home/jenkins/agent/workspace/eclipse.platform.swt_PR-252/eclipse.platform.swt/tests/org.eclipse.swt.tests/target/classes
```

Note that SWT is on maven cache, but the jar has name `3.121.0-SNAPSHOT.jar`. Which makes me wonder: is it a published SWT or a locally built SWT which for some reason was cached immediately? If it's the second, then test should be reworked. Probably to ensure that swt.jar is not signed.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_SWT.java
SyntevoAlex: Thanks for clarification. I have almost zero maven experience, and two days ago, it was zero exactly :)

SyntevoAlex:
I have updated the patch. I'll wait for test results and remove debugging code and put signature test first.

SyntevoAlex:
OK, I did everything I wanted, please continue reviewing.
Note that signature test is first now, because it should be more reliable.

laeubi:
Thanks for the test case (now we need to build the fragments also in GH action ...)

PR #253, opened 2022-07-12
Checkout an build the binary fragments

laeubi:
Currently always the fragments from the latest ibuild are used.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/251

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -30,6 +30,12 @@ jobs:
       with:
        path: 'eclipse.platform.swt'
        fetch-depth: 0 # required for jgit timestamp provider to work
+    - name: checkout binaries
+      uses: actions/checkout@v3
+      with:
+       path: 'eclipse.platform.swt.binaries'
+       repository: 'eclipse-platform/eclipse.platform.swt.binaries'
+       fetch-depth: 0 # required for jgit timestamp provider to work
     - name: Install Linux requirements
       run: sudo apt-get update -qq && sudo apt-get install -qq -y webkit2gtk-driver
       if: ${{ matrix.config.native == 'gtk.linux.x86_64'}}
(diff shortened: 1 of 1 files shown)


laeubi:
@akurtakov can you help here? I'm not sure if actually we want to build the **binaries**, but build the **jar fragment** I just don't see how this actually works...

If I look at https://github.com/eclipse-platform/eclipse.platform.swt/runs/7304931962?check_suite_focus=true then I see that the org.eclipse.swt.tests.gtk 3.108.100-SNAPSHOT is build but *after* the tests are executed....

akurtakov:
We want to build and install the binaries fragments as -Pbuild-individual-bundles resolves dependencies outside of the current reactor build from local maven repo or I-builds repo (whichever is newer). org.eclipse.swt.tests.gtk is a bundle with manual tests and etc. used during development. Real JUnit tests are in org.eclipse.swt.tests.

laeubi:
But I don't get how it is supposed to work if _first_ binary is build and swt _after_ that? I remember that there was a parameter to specify the location of the binary somewhere.... but I don't see it set in the Jenkinsfile ...

akurtakov:
Building binaries uses sources in eclipse.platform.swt repo to create the fragrments https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/bundles/org.eclipse.swt/buildFragment.xml . Bundle org.eclipse.swt is just a shell so building it really creates nothing in terms of actual class files.

akurtakov:
What is the status of this one? It doesn't look like it works fine to me.

laeubi:
No I havent had time to further look into it and it seems my inital idea to build the binaries was not useful, looking at the build I don't see why the bundles should **not** be build, maybe they are just not installed in the local repository?

PR #255, opened 2022-07-13
Fix failing test case on Mac - TextLayout.test_getLineOffsets()

lshanmug:
Test has been adapted based on Mac behavior


niraj-modi (approved):
Change looks good:
- no side effect on Windows.

PR #260, opened 2022-07-14
Replace lambdas with method handles.

akurtakov:
Examples and snippets are covered.


akurtakov:

PR #263, opened 2022-07-20
Improve assertions in ImageData tests

akurtakov:
Use assertThrows and other smaller assertion improvements.
Prep work before investigating test failure as reported in
https://github.com/eclipse-platform/eclipse.platform.swt/issues/232


akurtakov:

PR #264, opened 2022-07-20
Make SWT tests fragment.

akurtakov:
Effort to reduce test failures as seen in
https://github.com/eclipse-platform/eclipse.platform.swt/issues/232 .


akurtakov:

SyntevoAlex:
Is it expected that you didn't change `Bundle-Localization: plugin` and didn't change platform-specific test projects' manifests?

akurtakov:
Platform specific test projects contain only manually executed tests AFAICT so no need to do it to fix the test failure. 
Bundle-Localization just points to the properties file to use for localizations so I think waiting to see that tests run before renaming plugin.properties might be best.

SyntevoAlex:
OK, thanks for working on it!

PR #274, opened 2022-07-26
Fix #256 Menus does not show on macOS Ventura

### LINKED ISSUE #256: Menus does not show on macOS Ventura
Having tested Eclipse on macOS Ventura beta (22A5295i) using an M1 based computer. I found that everything appears to be in order, except that menus do not pop up. This includes right clicking on any element in an editor or in a view along with clicking on main items in the menu bar. Pressing shortcuts to open menus, i.e. ⌥⌘N in order to produce the new file menu – does not yield a result either.

**To Reproduce**

Install Eclipse 2022-06 on macOS Ventura. Try to open any menu.

**Expected behavior**

Menus should appear

**Environment:**
- - [ ] All OS
- - [ ] Windows
- - [ ] Linux
- - [X] macOS 

* macOS Ventura beta (22A5295i)
* Eclipse 2022-06
* AArch64
* JustJ JRE 17.0.3 (built-in)

While this is tested on a beta-version of macOS I don't think this problem will go away by itself.

turesheim:
This fix basically disables the workaround for [Eclipse issue 578171](https://bugs.eclipse.org/bugs/show_bug.cgi?id=578171)  if running on macOS 13 (Ventura) or later. I have not found related information in the [macOS release notes](https://developer.apple.com/documentation/macos-release-notes/macos-13-release-notes), but I think it is safe to assume that the issue referenced in 578171 has been fixed.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT PI/cocoa/org/eclipse/swt/internal/cocoa/OS.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT PI/cocoa/org/eclipse/swt/internal/cocoa/OS.java
@@ -178,7 +178,7 @@ public class OS extends C {
 	 * Custom message that will be sent when setTheme is called for example from Platform UI code.
 	 */
 	public static final long sel_appAppearanceChanged = OS.sel_registerName("appAppearanceChanged");
-	
+
 	/**
 	 * Experimental API for dark theme.
 	 * <p>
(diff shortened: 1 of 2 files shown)


Phillipus:
> but I think it is safe to assume that the issue referenced in 578171 has been fixed.

Not in my testing with Ventura Beta 3.

Phillipus:
@turesheim It seems that this is not needed now, as @SyntevoAlex has a fix in #277

PR #277, opened 2022-07-27
Issue #256 - [cocoa] Menus does not show on macOS 13 Ventura

### LINKED ISSUE #256: Menus does not show on macOS Ventura

SyntevoAlex:
It was caused by undocumented setting used to prevent Bug 578171.

This patch changes the workaround to a different one, where instead
of trying to make menu behave, it's now canceled just before showing
Shell.

Fixes #256

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
@@ -878,33 +878,6 @@ static private void configureSystemOptions () {
 	if (OS.isBigSurOrLater ()) {
 		configureSystemOption ("NSViewUsesAutomaticLayerBackingStores", false);
 	}
-
-	/*
-	 * Bug 578171: There is new code in macOS 12 that remembers which
-	 * Shell was active before menu popup was shown and tries to
-	 * re-activate after menu popup is closed. Unfortunately there is a
-	 * bug in this code: if window list changes, it activates a wrong
-	 * Shell.
-	 *
-	 * This is a bug on its own, but worse yet, this causes a JVM crash
-	 * because activating a new Shell causes menu bar to reset its
-	 * internal data, which is unexpected to the macOS's menu tracking
-	 * loop.
-	 *
-	 * Both bugs are bugs of macOS itself. The workaround is to disable
-	 * the new macOS 12 behavior.
-	 *
-	 * The condition should be for (macOS >= 12), but it's not possible
-	 * to reliably distinguish 11 from 12, see comment for OS.VERSION.
-	 * That's fine: older macOS don't know this setting and will not
-	 * check for it anyway.
-	 */
-	if (OS.isBigSurOrLater ()) {
-		// The name of the option is misleading. What it really means
-		// is whether '-[NSMenuWindowManagerWindow _setVisible:]' shall
-		// save/restore current key window or not.
-		configureSystemOption ("NSMenuWindowManagerWindowShouldSetVisible", true);
-	}
 }
 
 /**
@@ -4950,6 +4923,13 @@ void setDeviceZoom() {
 	DPIUtil.setDeviceZoom (getDeviceZoom());
 }
 
+static void cancelRootMenuTracking () {
+	long rootMenu = OS.AcquireRootMenu ();
+	if (rootMenu == 0) return; // Extra safety, not sure if it can happen
+	OS.CancelMenuTracking (rootMenu, true, 0);
+	OS.CFRelease (rootMenu);
+}
+
 void setMenuBar (Menu menu) {
 	// If passed a null menu bar don't clear out the menu bar, but switch back to the
 	// application menu bar instead, if it exists.  If the app menu bar is already active
@@ -4965,7 +4945,7 @@ void setMenuBar (Menu menu) {
 	* event loop. The fix is to use CancelMenuTracking() instead.
 	*/
 //	menubar.cancelTracking();
-	OS.CancelMenuTracking (OS.AcquireRootMenu (), true, 0);
+	cancelRootMenuTracking ();
 	long count = menubar.numberOfItems();
 	while (count > 1) {
 		menubar.removeItemAtIndex(count - 1);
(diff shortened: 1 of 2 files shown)


Phillipus:
Tested this and all looks good!

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
```diff
@@ -4950,6 +4923,16 @@ void setDeviceZoom() {
 	DPIUtil.setDeviceZoom (getDeviceZoom());
 }
 
+static void cancelRootMenuTracking () {
+	long rootMenu = OS.AcquireRootMenu ();
+	if (rootMenu == 0) return; // Extra safety, not sure if it can happen
+	OS.CancelMenuTracking (rootMenu, true, 0);
+
+	// Notice from a reader:
+	// OS.ReleaseMenu() is needed here according to docs in AcquireRootMenu().
```
lshanmug, line 4932: Can you clarify this comment? What needs to be fixed?

lshanmug (approved):
Patch fixes the crash on macOS 12 and is safer as it doesn't depend on undocumented option. Thanks @Phillipus for verifying the fix on macOS 13 beta. I don't have it installed yet...

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
SyntevoAlex: Quite literally, `OS.ReleaseMenu()` call is missing. It's needed to release the object in order to prevent the leak. I didn't want o fix it because it will need new native libs, which will prevent the patch from being cherry picked easily.

Phillipus: It has always been like this, no? But does `OS.CancelMenuTracking` actually acquire the menu so it needs to be released?

SyntevoAlex: Yes, it has always been like this.
docs in `AcquireRootMenu()` say:
```
 *    The refcount of the root menu is incremented by this API. The
 *    caller should release a refcount with ReleaseMenu when it's done
 *    with the menu.
```

Phillipus: But where is `AcquireRootMenu()` called?

SyntevoAlex: One line above. Check full patch diffs instead of preview.

Phillipus: Aarggh!! Sorry for the noise!

SyntevoAlex: It happens :) Thanks for the testing effort you've put into this, it's been helpful!

turesheim:
Thanks @SyntevoAlex. I can also confirm that this fixes the issue.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
lshanmug: Thanks Alexandr! The code has been like this, so let's merge this change. But, since we may have additional calls to AcquireRootMenu() now, it would be good to fix the leak as well. Can you please create a PR to address the leak?

Phillipus: Apple docs for `ReleaseMenu()` say "Deprecated: In Mac OS X 10.2 and later, use CFRelease instead of ReleaseMenu."

So could it be as simple as calling `OS.CFRelease(rootMenu);` ? I tried that and it didn't crash. ;-)

Phillipus: I called `OS.objc_msgSend(rootMenu, OS.sel_retainCount);` before and after `OS.CFRelease(rootMenu);` and the retain count is decremented by 1.

Phillipus:
I did some more testing with the addition of `OS.CFRelease(rootMenu);` as per my earlier comment, and it all seems OK.

@SyntevoAlex @lshanmug WDYT?

lshanmug:
> I did some more testing with the addition of `OS.CFRelease(rootMenu);` as per my earlier comment, and it all seems OK.
> 
> @SyntevoAlex @lshanmug WDYT?

Thanks @Phillipus for trying this out. Since CFRelease() is already available in OS.java and no native changes are required, I think it it would be good to add it to the PR. @SyntevoAlex WDYT?

SyntevoAlex:
Sorry for delays, been quite busy.
The suggestion about `CFRelease()` sounds reasonable to me. I will adjust the patch soon, hopefully tomorrow. @Phillipus  Thanks for your testing!

SyntevoAlex:
I have adjusted the patch, please continue reviewing.

lshanmug (approved):
Looks good to me.

Phillipus:
Did a quick test and can confirm menus are working. Thanks @SyntevoAlex

lshanmug:
Thanks @SyntevoAlex and @Phillipus!

SyntevoAlex:
Thanks @lshanmug @Phillipus  for your assistance!

SyntevoAlex:
See also #337

PR #279, opened 2022-07-29
Add list of Snippets as a Markdown file

Phillipus:
- See https://github.com/eclipse-platform/eclipse.platform.swt/issues/258


mickaelistria:
Thanks.
As this is pure decoration that is not changing the build outcome, we can maybe merge it right now. @akurtakov What do you think?

akurtakov:
Yes, let's merge it.

PR #289, opened 2022-08-06
Integrate the Display with the Executor Framework

laeubi:
Concurrency could be challenging, especially when it comes to UI and
responsiveness. And even more with the single-threading model of SWT
(and other UI frameworks as well) as one easily could block the UI then.

Java for several year contains some where useful tool with the
concurrency framework but it is quite hard for users to interact with
that as SWT does not offer some basic primitives (most notable Executors
and CompletableFuture).

This add some Executor support to the Display to allow integration with
the Executor Framework handling some special requirements of SWT, e.g
the single thread model, disposal checking and exception handling.

See also https://github.com/eclipse-platform/eclipse.platform.swt/issues/64

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2019 IBM Corporation and others.
+ * Copyright (c) 2000, 2022 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -10,11 +10,13 @@
  *
  * Contributors:
  *     IBM Corporation - initial API and implementation
+ *     Christoph Läubrich - Issue #64 - Integration with java.util.concurrent framework
  *******************************************************************************/
 package org.eclipse.swt.widgets;
 
 import java.lang.Runtime.*;
 import java.util.*;
+import java.util.concurrent.*;
 import java.util.function.*;
 
 import org.eclipse.swt.*;
@@ -102,7 +104,7 @@ import org.eclipse.swt.internal.cocoa.*;
  * @noextend This class is not intended to be subclassed by clients.
  */
 @SuppressWarnings({"rawtypes", "unchecked"})
-public class Display extends Device {
+public class Display extends Device implements Executor {
 
 	static byte[] types = {'*','\0'};
 	static int size = C.PTR_SIZEOF, align = C.PTR_SIZEOF == 4 ? 2 : 3;
@@ -610,6 +612,49 @@ public void asyncExec (Runnable runnable) {
 	synchronized (Device.class) {
 		if (isDisposed ()) error (SWT.ERROR_DEVICE_DISPOSED);
 		synchronizer.asyncExec (runnable);
+	}
+}
+
+/**
+ * Executes the given runnable in the user-interface thread of this Display.
+ * <ul>
+ * <li>If the calling thread is the user-interface thread of this display it is
+ * executed immediately and the method returns after the command has run, as with
+ * the method {@link Display#syncExec(Runnable)}.</li>
+ * <li>In all other cases the <code>run()</code> method of the runnable is
+ * asynchronously executed as with the method
+ * {@link Display#asyncExec(Runnable)} at the next reasonable opportunity. The
+ * caller of this method continues to run in parallel, and is not notified when
+ * the runnable has completed.</li>
+ * </ul>
+ * <p>
+ * This can be used in cases where one want to execute some piece of code that
+ * should be guaranteed to run in the user-interface thread regardless of the
+ * current thread.
+ * </p>
+ *
+ * <p>
+ * Note that at the time the runnable is invoked, widgets that have the receiver
+ * as their display may have been disposed. Therefore, it is advised to check
+ * for this case inside the runnable before accessing the widget.
+ * </p>
+ *
+ * @param runnable the runnable to execute in the user-interface thread, never
+ *                 <code>null</code>
+ * @throws RejectedExecutionException if this task cannot be accepted for
+ *                                    execution
+ * @throws NullPointerException       if runnable is null
+ */
+@Override
+public void execute(Runnable runnable) {
+	Objects.requireNonNull(runnable);
+	if (isDisposed()) {
+		throw new RejectedExecutionException(new SWTException (SWT.ERROR_WIDGET_DISPOSED, null));
+	}
+	if (thread == Thread.currentThread()) {
+		syncExec(runnable);
+	} else {
+		asyncExec(runnable);
 	}
 }
 
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2021 IBM Corporation and others.
+ * Copyright (c) 2000, 2022 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -10,6 +10,7 @@
  *
  * Contributors:
  *     IBM Corporation - initial API and implementation
+ *     Christoph Läubrich - Issue #64 - Integration with java.util.concurrent framework
  *******************************************************************************/
 package org.eclipse.swt.widgets;
 
@@ -19,6 +20,7 @@ import java.lang.reflect.*;
 import java.net.*;
 import java.util.*;
 import java.util.Map.*;
+import java.util.concurrent.*;
 import java.util.function.*;
 import java.util.regex.*;
 import java.util.regex.Pattern;
@@ -111,7 +113,7 @@ import org.eclipse.swt.internal.gtk4.*;
  * @see <a href="http://www.eclipse.org/swt/">Sample code and further information</a>
  * @noextend This class is not intended to be subclassed by clients.
  */
-public class Display extends Device {
+public class Display extends Device implements Executor {
 
 	static boolean strictChecks = System.getProperty("org.eclipse.swt.internal.gtk.enableStrictChecks") != null;
 
@@ -926,6 +928,49 @@ public void asyncExec (Runnable runnable) {
 			}
 		}
 		synchronizer.asyncExec (runnable);
+	}
+}
+
+/**
+ * Executes the given runnable in the user-interface thread of this Display.
+ * <ul>
+ * <li>If the calling thread is the user-interface thread of this display it is
+ * executed immediately and the method returns after the command has run, as with
+ * the method {@link Display#syncExec(Runnable)}.</li>
+ * <li>In all other cases the <code>run()</code> method of the runnable is
+ * asynchronously executed as with the method
+ * {@link Display#asyncExec(Runnable)} at the next reasonable opportunity. The
+ * caller of this method continues to run in parallel, and is not notified when
+ * the runnable has completed.</li>
+ * </ul>
+ * <p>
+ * This can be used in cases where one want to execute some piece of code that
+ * should be guaranteed to run in the user-interface thread regardless of the
+ * current thread.
+ * </p>
+ *
+ * <p>
+ * Note that at the time the runnable is invoked, widgets that have the receiver
+ * as their display may have been disposed. Therefore, it is advised to check
+ * for this case inside the runnable before accessing the widget.
+ * </p>
+ *
+ * @param runnable the runnable to execute in the user-interface thread, never
+ *                 <code>null</code>
+ * @throws RejectedExecutionException if this task cannot be accepted for
+ *                                    execution
+ * @throws NullPointerException       if runnable is null
+ */
+@Override
+public void execute(Runnable runnable) {
+	Objects.requireNonNull(runnable);
+	if (isDisposed()) {
+		throw new RejectedExecutionException(new SWTException (SWT.ERROR_WIDGET_DISPOSED, null));
+	}
+	if (thread == Thread.currentThread()) {
+		syncExec(runnable);
+	} else {
+		asyncExec(runnable);
 	}
 }
 
@@ -6198,4 +6243,5 @@ static int _getDeviceZoom (long monitor_num) {
 	dpi = dpi * scale;
 	return DPIUtil.mapDPIToZoom (dpi);
 }
+
 }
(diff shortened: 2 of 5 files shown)


laeubi:
Only the failures we see on master too:
- #294 
- #295

vogella (approved):

szarnekow:
Please add a test and documentation that warns about the deadlock potential when the UI thread does a blocking.get on a future that's being scheduled on the display's executor.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
```diff
@@ -929,6 +931,55 @@ public void asyncExec (Runnable runnable) {
...
+public void execute(Runnable runnable) {
+	Objects.requireNonNull(runnable);
+	if (isDisposed()) {
+		throw new RejectedExecutionException(new SWTException (SWT.ERROR_WIDGET_DISPOSED, null));
+	}
+	if (thread == Thread.currentThread()) {
+		try {
+			runnable.run ();
+		} catch (RuntimeException exception) {
+			getRuntimeExceptionHandler ().accept (exception);
```
szarnekow, line 974: Assuming this was runnable was executed in a future. If it throws an exception, I'd expect the future to fail. Can you please add a test for that usecase, too?

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
```diff
@@ -613,6 +615,55 @@ public void asyncExec (Runnable runnable) {
...
+ * </p>
+ *
+ * @param runnable the runnable to execute in the user-interface thread, never
+ *                 <code>null</code>
+ * @throws RejectedExecutionException if this task cannot be accepted for
+ *                                    execution
+ * @throws NullPointerException       if runnable is null
+ */
+@Override
+public void execute(Runnable runnable) {
```
szarnekow, line 649: Another question about the contract here:
Would we expect that exceptionHandler or errorHandler to be called if the runnable is a FutureTask where the wrappedCallable threw an exception?

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
laeubi: You cannot execute a `Future` directly with an `Executor`, beside that internal implementation details of the runnable are not to be handled here... This simply replicates how `org.eclipse.swt.widgets.Synchronizer.syncExec(Runnable)` already works.

Of course a `Runnable` could implement whatever exception handling seems suitable.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
laeubi: If the runnable handles exceptions itself then of course no SWT code will ever be used to handle this.

laeubi:
> Please add a test and documentation that warns about the deadlock potential when the UI thread does a blocking.get on a future that's being scheduled on the display's executor.

This has nothing to do with this implementation and/or change. An execute will never deadlock itself but only when inappropriate waiting code is used, so the usual SWT rules apply here.

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet380.java
```diff
@@ -0,0 +1,63 @@
...
+						TimeUnit.SECONDS.sleep(1);
+					} catch (InterruptedException e1) {
+						break;
+					}
+					value += Math.random();
+				}
+				return value;
+			}).thenApply(result -> NumberFormat.getInstance().format(result))//
+					.thenApply(formatedNumber -> String.format("The result is %s", formatedNumber))//
+					.thenAcceptAsync(label::setText, Display.getDefault())//
```
mickaelistria, line 52: I think we should encourage using the "local" display instead of the singleton, so maybe
```suggestion
					.thenAcceptAsync(label::setText, label.getDisplay())//
```

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet380.java
```diff
@@ -0,0 +1,63 @@
...
+					.whenCompleteAsync((x, y) -> button.setEnabled(true), Display.getDefault());
```
mickaelistria, line 53: and
```suggestion
					.whenCompleteAsync((x, y) -> button.setEnabled(true), button.getDisplay());
```

vogella:
Can this be finally merged?

laeubi:
> Can this be finally merged?

Ready if you are ;-)

vogella:
+1 for the merge unless someone still has issues. Shall we say, please additional/ remaining raise concerns until 29.09?

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
```diff
@@ -613,6 +615,55 @@ public void asyncExec (Runnable runnable) {
...
+	Objects.requireNonNull(runnable);
+	if (isDisposed()) {
+		throw new RejectedExecutionException(new SWTException (SWT.ERROR_WIDGET_DISPOSED, null));
+	}
+	if (thread == Thread.currentThread()) {
+		try {
+			runnable.run ();
```
szarnekow, line 656: Judging from the logic in `syncExec` this appears to bypass a few callbacks, e.g. `Synchronizer.syncExec` will call `display.sendPreEvent(..)` and `sendPostEvent` respectively. 

Also this might cause reordering of events since the runnable will execute immediately on the DisplayThread ignoring other UI events that have been queued in the meantime. This doesn't feel right.

szarnekow: > Also this might cause reordering of events since the runnable will execute immediately on the DisplayThread ignoring other UI events that have been queued in the meantime.

Never mind, this will happen with a syncExec issued from the DisplayThread anyhow. 

Instead of calling runnable.run I'd rather call syncExec if the current thread is the display thread.

laeubi: This is more to call the code of the runnable immediately without any extra work.

szarnekow: You're saying you don't want to keep the contract of Display.syncExec if one uses the display as an Executor, but you do want to keep the contract for asyncExec?

laeubi: > You're saying you don't want to keep the contract of Display.syncExec if one uses the display as an Executor, but you do want to keep the contract for asyncExec?

The only contract is that the Executor guarantees is the given code is always executed in the UI thread. If the current Thread the UI Thread this contract is fulfilled by invoking the code right now. I don't see how `syncExec` would improve anything here?

szarnekow: It would send the proper sendPre and sendPost event notification keeping all observers of these events up-to-date and not cloaking a random runnable as part of another runnable that's currently mid-flight. I don't understand why you would want to bypass these in favor of duplicating the exception handling here that is already done in the Synchronizer.

laeubi: > It would send the proper sendPre and sendPost event notification keeping all observers of these events up-to-date and not cloaking a random runnable as part of another runnable that's currently mid-flight. I

Why should this be desired/required? You seem to searching for problems that simply don't extis, if I have a method and call for example

```
label.setText("whatever")
label.setEnabled(true)
label.setWhatever()

```
no one would complain I should instead call `syncExec` for any single call...

szarnekow: > no one would complain 

Neither would I.

I'd also not expect someone to write these three method calls as a sequence of futures based on the executor framework.

But if I see a sequence of futures that exercise the Display as an Executor, I'd expect the usual things to still work for the runnables that are submitted via that API, e.g. UI freeze monitoring. Can you guarantee that if you bypass the Synchronizer?

szarnekow: As starting point, please look into org.eclipse.ui.internal.monitoring.EventLoopMonitorThreadManualTests.testFixedWork() and change `display.syncExec(doFixedAmountOfWork);` to something of the shape
```java
CompletableFuture<Void> future = CompletableFuture.runAsync(()->{}).thenApplyAsync(any->{
	return null;
}, display).thenRunAsync(doFixedAmountOfWork, display);
while(!future.isDone()) {
	while(display.readAndDispatch ()) {
		//dispatch
	}
}
```

You'll no longer see any stacks being reported from the long-running runnable.

laeubi: I can't think about a case where it makes a difference, maybe you have an example? For sure this dont prevent all possible deadlock and if one want can for sure deadlock as well as one can with the traditional API but that's also not the goal.

laeubi: > You'll no longer see any stacks being reported from the long-running runnable.

Long running runnable should never be executed in the UI thread so why should I want to do that? :-)

The desired use-case would be:

`CompletableFuture.supplyAsync(()->return doLenthyComputation()).thenAcceptAsync(result->updateTheUI(result), display);`

so no need for blocking programming style...

szarnekow: I back off from this discussion. You're saying the pre and post events are useless, I'm saying the opposite. I don't see any progress. Since I'm not a committer on SWT, I let someone make the call.

merks: I'm not sure I understand the adamant resistance to what looks to me like a simple and constructive suggestion.   Is there some significant harm, or some violated principle, that comes from doing what's suggested?   These are after all methods being added to Display, so it would seem to me that the more these new methods behave (as expected) like the existing methods that they emulate or are intended to replace, the better.  

What exactly is the strong counter argument to not do what @szarnekow suggests?  I am a committer and I would do as he suggests.

laeubi: The general contract of the `Executor` interface:

>Executes the given command at some time in the future.  The command
> may execute in a new thread, in a pooled thread, or in the calling
> thread, at the discretion of the {@code Executor} implementation.

So we are actually free tom implement **what ever suffice**.

> I'm not sure I understand the adamant resistance to what looks to me like a simple and constructive suggestion.

Why is the current approach not "simple and constructive"? Why is it "adamant" if one likes to understand why we should choose a much more expensive construct that has a different semantics here?

> Is there some significant harm, or some violated principle, that comes from doing what's suggested?

Due to how SWT works, executing the code immediately or using syncExec has different semantics and different impact:

1. Using `syncExec` you might get code executed "in between" as per the API:  "At the time the runnable is invoked, widgets that have the receiver as their display may have been disposed", that might not be an issue, but could be an issue if such code is not just disposing a widget but changing its state (I think that what @szarnekow referes to with event reordering).
2. If you look at the implementation (e.g. GTK), there is a lot of overhead involved, that is calling the Synchronizer (what might or might not trigger more code, e.g Platform has `UISyncronizer` doing additional checks)), synchronizing on locks and performing system calls.
3. Whenever I see code in our platform code or other libs, the usual idiom is to check if one is on the UI thread if yes, execute the code, if no either `asyncExec` or `syncExec` code (depending on weather would want to be blocked or not), I never found an example where `syncExec` is used just randomly if one is already in the UI thread.

That's why I have given the example with just calling on (or more) methods that strictly requires a UI-Thread, something that is common, no one ever would argue you should use `syncExec` all the time to trigger events, nor one would suggest that it is Ok to run lengthy operation when using `syncExec` ...

So for me it sounds not that bad simply replicate this common pattern. If the other approach has strong benefits I'm open to change that, but it seems a very costly operation and I can't tell if the event reordering can cause any harm.

Whatever is decided here the implementation will still never:

1. Prevent the user from any deadlocks, as due the single-threaded nature you can always created a situation where deadlocks occur when you use blocking calls, **but** will give the user the opportunity to use non blocking programming paradigms.
2. Make it "save" to do heavy weight computations or long running task in the UI thread, **but**  will give the user the opportunity to simply fire this work "in the background" and using the result to do the necessary UI work as short as possible.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
```diff
@@ -613,6 +615,55 @@ public void asyncExec (Runnable runnable) {
 	}
 }
 
+/**
+ * Executes the given runnable in the user-interface thread of this Display.
+ * <ul>
+ * <li>If the calling thread is the user-interface thread of this display it is
+ * executed immediately and the method returns after the command has run, any
+ * runtime exception thrown will be rethrown at the caller</li>
+ * <li>In all other cases the the <code>run()</code> method of the runnable is
```
merks, line 624: "the the"

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
szarnekow: Please ensure that UI freeze monitoring works for tasks submitted via the newly introduced method. I’m not asking for more than that. You’ll get it for free when using syncExec after checking that the current thread is the display thread.

laeubi:
@szarnekow 
@mickaelistria
@merks 

All suggestions should now be applied so please check if anything is left so we can have this before M3 and do not miss the release!

laeubi:
Only usual build failures....

szarnekow:
Thank you for following up on this, @laeubi. As stated previously I cannot merge this, but it looks good to me and I'd approve.

mickaelistria (approved):
This API addition is pretty clear and very useful to write better modern code. I'm fine with it being merged for M3.

merks:
It looks fine.

mickaelistria:
@laeubi Please add a note about this new API in the noteworthy change doc.

lshanmug:
@laeubi Please add an N&N entry for the new API in [platform_isv.html](https://github.com/eclipse-platform/www.eclipse.org-eclipse-news/blob/master/4.26/platform_isv.html)

laeubi:
See 
- https://github.com/eclipse-platform/www.eclipse.org-eclipse-news/pull/68

PR #293, opened 2022-08-08
GH: Build the swt fragments together with the rest of the build

laeubi:
Currently the fragments are not build but taken from the I-Build in the
Gh actions, this fixes the problem by checkout and build the binary
repository as well.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/251

### DIFF
--- a/.github/aggregator.pom
+++ b/.github/aggregator.pom
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>eclipse.platform.swt</groupId>
+    <artifactId>aggregator</artifactId>
+    <packaging>pom</packaging>
+    <version>1</version>
+	<modules>
+		<module>..</module>
+		<module>../../eclipse.platform.swt.binaries</module>
+	</modules>
+</project>
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -30,6 +30,12 @@ jobs:
       with:
        path: 'eclipse.platform.swt'
        fetch-depth: 0 # required for jgit timestamp provider to work
+    - name: checkout swt.binaries
+      uses: actions/checkout@v3
+      with:
+       path: eclipse.platform.swt.binaries
+       fetch-depth: 0 # required for jgit timestamp provider to work
+       repository: 'eclipse-platform/eclipse.platform.swt.binaries'
     - name: Install Linux requirements
       run: sudo apt-get update -qq && sudo apt-get install -qq -y webkit2gtk-driver
       if: ${{ matrix.config.native == 'gtk.linux.x86_64'}}
@@ -43,7 +49,7 @@ jobs:
       uses: GabrielBB/xvfb-action@v1
       with:
        run: >- 
-        mvn --batch-mode
+        mvn --batch-mode -V -U
         -Pbuild-individual-bundles
         -Pbuild-individual-platform
         -Dswt.build.platform_cp=${{ matrix.config.cp }}
@@ -53,7 +59,8 @@ jobs:
         -Dmaven.test.failure.ignore=true
         -DskipNativeTests=false
         -DfailIfNoTests=false
-        clean verify
+        -f .github/aggregator.pom
+        clean install
        working-directory: eclipse.platform.swt
     - name: Upload Test Results for ${{ matrix.config.name }} / Java-${{ matrix.java }}
       uses: actions/upload-artifact@v3


laeubi:
I now see that the fragments get installed:
```
 --- maven-install-plugin:2.3.1:install (default-install) @ org.eclipse.swt.win32.win32.x86_64 ---
 [INFO] Installing /home/runner/work/eclipse.platform.swt/eclipse.platform.swt/eclipse.platform.swt.binaries/bundles/org.eclipse.swt.win32.win32.x86_64/target/org.eclipse.swt.win32.win32.x86_64-3.121.0-SNAPSHOT.jar 
to /home/runner/.m2/repository/org/eclipse/swt/org.eclipse.swt.win32.win32.x86_64/3.121.0-SNAPSHOT/org.eclipse.swt.win32.win32.x86_64-3.121.0-SNAPSHOT.jar
```
And in the Reactor Summary also contains them:
```
[INFO] Reactor Summary:
[INFO] 
[INFO] eclipse.platform.swt 4.25.0-SNAPSHOT ............... SUCCESS [ 18.933 s]
[INFO] org.eclipse.swt 3.121.0-SNAPSHOT ................... SUCCESS [ 34.017 s]
[INFO] org.eclipse.swt.gtk.linux.x86_64 3.121.0-SNAPSHOT .. SUCCESS [ 19.957 s]
[INFO] org.eclipse.swt.gtk.linux.ppc64le 3.121.0-SNAPSHOT . SUCCESS [ 14.546 s]
[INFO] org.eclipse.swt.fragments.localbuild 3.105.0-SNAPSHOT SUCCESS [  0.482 s]
[INFO] org.eclipse.swt.tools.base 3.107.400-SNAPSHOT ...... SUCCESS [ 11.066 s]
[INFO] org.eclipse.swt.tools.spies 3.108.400-SNAPSHOT ..... SUCCESS [  3.110 s]
[INFO] org.eclipse.swt.tools 3.109.500-SNAPSHOT ........... SUCCESS [  2.465 s]
[INFO] org.eclipse.swt.examples 3.107.400-SNAPSHOT ........ SUCCESS [  2.124 s]
[INFO] org.eclipse.swt.examples.browser.demos 3.107.100-SNAPSHOT SUCCESS [  1.463 s]
[INFO] org.eclipse.swt.examples.launcher 3.107.100-SNAPSHOT SUCCESS [  1.166 s]
[INFO] org.eclipse.swt.win32.win32.x86_64 3.121.0-SNAPSHOT  SUCCESS [ 16.609 s]
[INFO] org.eclipse.swt.examples.ole.win32 3.108.0-SNAPSHOT  SUCCESS [  3.762 s]
[INFO] org.eclipse.swt.examples.views 3.107.100-SNAPSHOT .. SUCCESS [  1.631 s]
[INFO] org.eclipse.swt.tests 3.106.1800-SNAPSHOT .......... SUCCESS [02:53 min]
[INFO] org.eclipse.swt.tools.feature 3.108.600-SNAPSHOT ... SUCCESS [  0.779 s]
[INFO] org.eclipse.swt.tests.gtk 3.108.100-SNAPSHOT ....... SUCCESS [  5.166 s]
[INFO] org.eclipse.swt.cocoa.macosx.x86_64 3.121.0-SNAPSHOT SUCCESS [ 15.807 s]
[INFO] org.eclipse.swt.cocoa.macosx.aarch64 3.121.0-SNAPSHOT SUCCESS [ 15.353 s]
[INFO] org.eclipse.swt.gtk.linux.aarch64 3.121.0-SNAPSHOT . SUCCESS [ 14.157 s]
[INFO] eclipse.platform.swt.binaries 4.25.0-SNAPSHOT ...... SUCCESS [  0.005 s]
[INFO] aggregator 1 ....................................... SUCCESS [  0.115 s]
```

But the test `org.eclipse.swt.tests.junit.Test_org_eclipse_swt_SWT#isLocal` still complains:
java.lang.AssertionError: Tested SWT is signed by Eclipse: (SWT='D:/a/eclipse.platform.swt/eclipse.platform.swt/eclipse.platform.swt.binaries/bundles/org.eclipse.swt.win32.win32.x86_64/target/org.eclipse.swt.win32.win32.x86_64-3.121.0-SNAPSHOT.jar' signer='1.2.840.113549.1.9.1=#16157765626d61737465724065636c697073652e6f7267,CN=Eclipse.org Foundation\, Inc.,OU=IT,O=Eclipse.org Foundation\, Inc.,L=Ottawa,ST=Ontario,C=CA')

I think this is because of:
```
[INFO] MavenProject: org.eclipse.swt:org.eclipse.swt.win32.win32.x86_64:3.121.0-SNAPSHOT @ /home/runner/work/eclipse.platform.swt/eclipse.platform.swt/eclipse.platform.swt.binaries/bundles/org.eclipse.swt.win32.win32.x86_64/pom.xml
2022-08-08T09:52:53.9267822Z     The main artifact has been replaced with the baseline version.
```

so I think the `org.eclipse.swt.tests.junit.Test_org_eclipse_swt_SWT#isLocal` can not work reliable here because it requires a change in those fragments to work.

I'll still merge this as it at least makes sure these parts of the code are build even though the test itself can not detect this at the moment.

SyntevoAlex:
Previously I noticed that the problem is reproducible in a local maven build. If I understand it correctly, you only changed github-related scripts. To me, it sounds like the problem should still be there. What do you think?

laeubi:
for a local build, you need to checkout both repos side-by-side and then use the aggreagator pom like in the github action.

SyntevoAlex:
I'm not sure if we understand each other.

I'm saying that I can download SWT repo and then use `verify` maven target in it.
My expectation is that it will build SWT and then run tests in it.
What actually happens is that it downloads latest published SWT and runs tests against it, which makes no sense to me.

laeubi:
> My expectation is that it will build SWT and then run tests in it.

The expectation is wrong, see https://www.eclipse.org/swt/fixbugs.php SWT uses a very special build setup that can't be replicated by standard maven.

> What actually happens is that it downloads latest published SWT and runs tests against it, which makes no sense to me.

If you want to replicate a full build you need:

1. checkout [eclipse.platform.swt](https://github.com/eclipse-platform/eclipse.platform.swt) 
2. checkout [eclipse.platform.swt.binaries](https://github.com/eclipse-platform/eclipse.platform.swt.binaries)
3. run inside `eclipse.platform.swt` folder:
```
mvn --batch-mode -V -U
        -Pbuild-individual-bundles
        -Pbuild-individual-platform
        -Dswt.build.platform_cp=[any of .classpath_gtk or .classpath_win32 or .classpath_cocoa]
        -DforkCount=1
        -Dcompare-version-with-baselines.skip=false
        -Dmaven.compiler.failOnWarning=true
        -Dmaven.test.failure.ignore=true
        -DskipNativeTests=false
        -DfailIfNoTests=false
        -f .github/aggregator.pom
        clean install
```

SyntevoAlex:
I'm sorry if I'm being stupid, but I still don't understand.

Let's put native binaries aside for now, even though they are a problem as well. For now, let's discuss Java side of SWT.

Now, I'm building SWT with maven, locally, without any github files, using this commandline I took from older (before this issue was created) build bot logs:
`mvn --batch-mode -Pbuild-individual-bundles -Pbuild-individual-platform "-Dswt.build.platform_cp=.classpath_win32" "-DforkCount=1" "-Dcompare-version-with-baselines.skip=false" "-Dmaven.compiler.failOnWarning=true" "-Dmaven.test.failure.ignore=true" "-DskipNativeTests=false" "-DfailIfNoTests=false" clean verify`

I'm confident that maven actually builds SWT, for a lot of reasons:
* The presence of `-Pbuild-individual-platform` implies that it's built
* maven output mentions `[javac] Compiling 615 source files`
* If I make an intentional syntax error in for example `Display.java`, maven build fails with a corresponding error.

However, when build completes, the tests are run against **published** SWT instead of the built one.
This is reinforced by the following test:
1. Add `static { System.out.println("Hello from custom SWT!"); }` in `Display.java`
2. This will not be seen in maven test output.

I assert that this makes no sense and tests shall be run against SWT that was just built.

Furthermore, you mention that you see this
`The main artifact has been replaced with the baseline version.`

And you observe that SWT.jar is signed, which is detected by my test.

By I understand that GitHub doesn't have Eclipse's signing key and anything built on GitHub can NOT turn out to be signed.

laeubi:
> Let's put native binaries aside for now, even though they are a problem as well.

They are the "key" to your understandind as "the swt jar" is simply just an almost empty bundle that only is the host for the fragments that

1. contain the compiled javacode
2. contain the compile native codes

> However, when build completes, the tests are run against **published** SWT instead of the built one.

because you cannot "compile" all SWT, you always compile just one part and need to publish that (that's is the swt.win32, swt.gkt, ..) so you need to build *and publish* them and thats what the swt.binaries part do, so for your test, the compiled code simpl do not exits as `clean verify` only compiles code but never publish it to your local repository and thus the tests can not find them!

> Furthermore, you mention that you see this `The main artifact has been replaced with the baseline version.`
> And you observe that SWT.jar is signed, which is detected by my test.

Yes that means that the code is compiled, compared with the baseline and then it was detected that they are equal and thus replaced with the signed(!) baseline artifacts. So if you ever want to "see" you local build items you need to increase the version of the bundle(s).

> By I understand that GitHub doesn't have Eclipse's signing key and anything built on GitHub can NOT turn out to be signed.

Baselineartifacts **are** signed and they replace the original compiled jar, in github in jenkins and also on your local machine.

SyntevoAlex:
I now understood that native libraries pose some problem.
Still, maven build sounds weird to me, but I guess I'll leave it alone, because I don't use maven to build.

I have now tested things in #0305:
* Changing java files is now correctly tested by GitHub when PR is pushed
* Changing native libraries is ignored. This is not very good for patches that involve native libraries and make a false impression that things work, because previously Jenkins builds tested both.

laeubi:
> * Changing native libraries is ignored. This is not very good for patches that involve native libraries and make a false impression that things work, because previously Jenkins builds tested both.

Then simply look at the Jenkins build ;-)

The github actions is to **run** the tests on the **real** platform, while the Jenkins build only compile (for all platforms) and run test on **linux**.

SyntevoAlex:
Maybe I was under a false impression from jenkins build as well. I think it can't be tested anymore.

> If you like to contribute the real native compile, this would be welcome, but it is hard to setup and would require a lot of more setup in the GH action parts.

Hmm, to my understanding, to build native libs, we just need to run a single ant target? That's how I build them myself anyway (using ant `swtbuild.build_libraries` from `eclipse.platform.swt.binaries/bundles/org.eclipse.swt.<platform>.x86_64/build.xml`)

On Windows and macOS, it would need a `runs-on` image with a compiler (Visual Studio and XCode respectively), which sounds easy with GitHub actions. I wouldn't be surprised if current images already have the compiler. On Linux, a couple more `apt- install` may be needed.

laeubi:
As said, if you think its easy feel free to prose a patch, that would be more than welcome, I never compiled SWT and the documentation is outdated, no one really knows the details and questions are mostly gone, I completely reverse-engineered the current state from the built-scripts and as you know no one noticed for a long time that it was flawed ...

> I think it can't be tested anymore, so can't say for sure.

At least the last time I asked it was mentioned that there are special builds that publish this to the swt.binaries repository so we can consume it from there, also the I-Builds are running on "real" windows/mac/linux, but I'm not aware of any just-in-time verification at the time you create a PR, I think with gerrit it was somehow possible to request a windows/mac node for build but nothing automatic.

SyntevoAlex:
I *think* that gerrits were unit tested properly. But I'm not entirely sure.

The problem with me is, as usual, complete lack of spare time :(  On top of that, I'm not even a Java programmer, so all these maven and ant scripts are completely new to me. I merely found way to build things on all 3 platforms that works for me and that's it.

Alas, it seems that someone else needs to pick this issue up.

PR #224, opened 2022-06-23
Table remains visible though setVisible(false) was called

niraj-modi:
fixes #223 

Change-Id: Ie6ea38db33f395efbba13d2451584fed84bfe395
Signed-off-by: Niraj Modi <niraj.modi@in.ibm.com>

PR #303, opened 2022-08-10
Issue #301 - wrong transparency when drawing image (regression from PR #224)

### LINKED ISSUE #301: Regression drawing image arising from fix to #185 and other regressions in Image class
Windows only.

There have been a few  changes to the way transparency is handled in the `Image` class on Windows from @jonathan-meier recently. However, we were left with a bug that caused problems that was apparently fixed here with https://github.com/eclipse-platform/eclipse.platform.swt/issues/185

Unfortunately this fix has caused yet another bug which now makes our RCP application broken when targetted with the latest SWT version.

The regression is caused by the fix in 8def011fdd013543a5ea5513dacef5914c3c8662 - however the underlying cause is somewhere deeper than that arising from previous changes to `Image` class on Windows

I have put together a snippet that shows the problem (see later post)

If you run the snippet with a version of SWT prior to 8def011fdd013543a5ea5513dacef5914c3c8662 a red box should be drawn:

correct

But running the snippet with a version of SWT after 8def011fdd013543a5ea5513dacef5914c3c8662 gives a wrong colored image:

incorrect

jonathan-meier:
Fixes #301.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
@@ -1862,7 +1862,11 @@ static long [] init(Device device, Image image, ImageData i) {
 					}
 					break;
 				case 32:
-					if (!(redMask == 0xFF00 && greenMask == 0xFF0000 && blueMask == 0xFF000000)) {
+					if (i.getTransparencyType() != SWT.TRANSPARENCY_MASK) {
+						newDepth = 24;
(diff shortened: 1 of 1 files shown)


vogella:
@SyntevoAlex can you reiview and merge this one?

SyntevoAlex:
I could have a look, hopefully today or tomorrow.

niraj-modi (approved):
This patch works as expected, tested with below snippet:
https://github.com/eclipse-platform/eclipse.platform.swt/issues/301#issuecomment-1209832523

Phillipus:
Thanks, everyone! 👍

Issue #295, opened 2022-08-08
org.eclipse.swt.tests.junit.Test_org_eclipse_swt_graphics_TextLayout.test_getLineOffsets fail on master


laeubi:
See https://ci.eclipse.org/platform/job/eclipse.platform.swt/job/master/571/testReport/junit/org.eclipse.swt.tests.junit/Test_org_eclipse_swt_graphics_TextLayout/test_getLineOffsets/

joel-majano:
Getting a 404 Error on that link. I would link to https://ci.eclipse.org/platform/job/eclipse.platform.swt/job/master/ then look at the latest build (currently #576).

laeubi:
571 is gone but the just finished master shows the same problem:
https://ci.eclipse.org/platform/job/eclipse.platform.swt/job/master/576/testReport/junit/org.eclipse.swt.tests.junit/Test_org_eclipse_swt_graphics_TextLayout/test_getLineOffsets/

lshanmug:
I assume this is on Linux as the Jenkins job runs the tests the Linux. Also, this failure is not seen in the I-build test results - https://download.eclipse.org/eclipse/downloads/drops4/I20220809-1800/testresults/html/org.eclipse.swt.tests_ep425I-unit-cen64-gtk3-java17_linux.gtk.x86_64_17.html

laeubi:
See https://ci.eclipse.org/platform/job/eclipse.platform.swt/job/master/582/#showFailuresLink

akurtakov:
This is caused by Pango 1.44 putting soft hyphenation (SHY) when wrapping now . It's explained in https://blogs.gnome.org/mclasen/2019/07/27/more-text-rendering-updates/ . From the documentation pango_attr_insert_hyphens_new should be cure but I haven't managed to make it work so far.

akurtakov:
One can run `pango-view --text WMWM --wrap word-char --width 28` and see the SHY displayed. There seems to be extra issue on SWT side in g_utf16_pointer_to_offset to consider it thus calculating wrong offsets.

akurtakov:
And `pango-view --wrap word-char --width 28 --markup --text "<span insert_hyphens='false'>WMWM</span>"` makes the hyphens disappear so this is the ultimate goal for SWT too.

PR #306, opened 2022-08-11
Fix TextLayout.getOffsets on Linux

akurtakov:
Pango started to use soft hyphens when char wrapping starting with
version 1.44 and SWT is not prepared for that thus returning wrong
offsets.
Disabling the hyphens is best as SWT never relied nor used hyphens at
any point in time.
Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/295


akurtakov:
Fixes the test failure.

PR #310, opened 2022-08-12
Fix oval window close button on Gtk under Wayland

akurtakov:
This was caused by a fix for
https://bugs.eclipse.org/bugs/show_bug.cgi?id=549117 which set the css
for buttons on latest Gtk versions to have different paddings for
top/bottom and left/right.
Changing it to be equal returns the proper round size of the close
button while preserving the fix in the mentioned.


akurtakov:
@iloveeclipse  I think you've complained about this one .

sravanlakkimsetti (approved):

iloveeclipse:
> I think you've complained about this one .

No, not really, I don't have Wayland yet on RHEL 7.9 :-(

PR #319, opened 2022-08-17
Add problem matcher to github for api and baseline problems

laeubi:
To help people finding problems without looking at the logs.

### DIFF
--- a/.github/matcher.json
+++ b/.github/matcher.json
@@ -1,0 +1,27 @@
+{
+  "problemMatcher": [
+    {
+      "owner": "api-problems",
+      "pattern": [
+        {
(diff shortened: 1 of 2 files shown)


laeubi:
@akurtakov I think freeze period do not really matter here?

akurtakov:
Yes, this one shouldn't be a problem. It's good practice to hold even such commits just on Thu for the sake of simplicity if we have to run a rebuild to have only single change in the gitlog as otherwise we will have to look what the other one is.

laeubi:
> It's good practice to hold even such commits just on Thu for the sake of simplicity if we have to run a rebuild to have only single change in the gitlog as otherwise we will have to look what the other one is.

Alright I think it would be good if the freeze check could add/update a comment that summarize such things (automatically adding when freeze period ends would be useful as well).

akurtakov:
This one needs manual rebase.

PR #325, opened 2022-08-18
Replicate Jenkins build in GH action

akurtakov:
This should make
https://github.com/eclipse-platform/eclipse.platform.swt/issues/251
work.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -45,6 +45,13 @@ jobs:
         java-version: ${{ matrix.java }}
         distribution: 'temurin'
         cache: maven
+    - name: Build swt.binaries fragments with Maven
+      uses: GabrielBB/xvfb-action@v1
+      with:
+       run: >- 
+        mvn --batch-mode -Pbuild-individual-bundles -DforceContextQualifier=zzz
+        -DskipJni -Dcompare-version-with-baselines.skip=true -Dmaven.compiler.failOnWarning=true install
+       working-directory: eclipse.platform.swt.binaries
     - name: Build with Maven
       uses: GabrielBB/xvfb-action@v1
       with:
@@ -59,7 +66,6 @@ jobs:
         -Dmaven.test.failure.ignore=true
         -DskipNativeTests=false
         -DfailIfNoTests=false
-        -f .github/aggregator.pom
         clean install
        working-directory: eclipse.platform.swt
     - name: Upload Test Results for ${{ matrix.config.name }} / Java-${{ matrix.java }}


akurtakov:
@laeubi @SyntevoAlex Would you please check that this one works for you? The important change is passing -DforceContextQualifier=zzz to swt fragments builds which ensures that they always override the repo one, just like if a version bump has happened. This is also the reason why both repos can't be build with single aggregator build.

laeubi (approved):
Looks good to me, still wondering why baseline checks are enabled if we effectively disable them afterwards...

akurtakov:
We want them enabled in the aggregator build but not in the verification build.

SyntevoAlex:
I suggest to just merge it and see what happens next. If my unit test #252 will stop finding problems, then I guess we could declare it a victory.

laeubi:
> We want them enabled in the aggregator build but not in the verification build.

Then it should also work to simply disable it for the complete build with `-Dcompare-version-with-baselines.skip=true` then also aggregator build should work as desired.

laeubi:
> I suggest to just merge it and see what happens next.

You can find the results if this right here: https://github.com/eclipse-platform/eclipse.platform.swt/runs/7901664228 I don't see a problem with that test there (what is not surprising because baseline checks are disabled for the first build).

SyntevoAlex:
Going through that link, I see this in logs for Windows Java 11:
```
Warning:  MavenProject: org.eclipse.swt:org.eclipse.swt:3.121.0-SNAPSHOT @ D:\a\eclipse.platform.swt\eclipse.platform.swt\eclipse.platform.swt\bundles\org.eclipse.swt\pom.xml: baseline and build artifacts have same version but different contents
   no-classifier: different
      .api_description: different
      META-INF/p2.inf: different
      about.html: different
      about_files/IJG_README: different

[INFO] MavenProject: org.eclipse.swt:org.eclipse.swt:3.121.0-SNAPSHOT @ D:\a\eclipse.platform.swt\eclipse.platform.swt\eclipse.platform.swt\bundles\org.eclipse.swt\pom.xml
    The main artifact has been replaced with the baseline version.
```

SyntevoAlex:
I don't know what this means, but based on previous discussions, it looks supsicious

akurtakov:
org.eclipse.swt artifact is plain shell, aka it contains nothing so it will be replaced almost always. I'm merging this one if there are issues please reopen.

SyntevoAlex:
OK, thanks for clarification!
@akurtakov thanks for fixing it!

PR #327, opened 2022-08-22
[GTK] Disabled toolbar icons and button images are double size on hi-res screen

### LINKED ISSUE #315: [GTK] Disabled toolbar icons and button images are double size on hi-res screen
**Describe the bug**
I'm bringing this over from https://bugs.eclipse.org/bugs/show_bug.cgi?id=575492 as the problem still exists.

Disabled images in buttons and toolbar items are the wrong size when DPI scaling is at 200%

**To Reproduce**
Run Snippet at 200% scaling to see that the disabled button image is scaled incorrectly and is too big and off-centred.

**Expected behavior**
The disabled button image should be scaled correctly when DPI is at 200% scale.

**Screenshots**
Disabled button at 100% scale:

100-scale

Disabled button at 200% scale:

200-scale

**Environment:**
- - [ ] All OS
- - [ ] Windows
- - [x] Linux
- - [ ] macOS 

Linux Mint Cinnamon X11. Didn't test on Wayland.

3. JRE/JDK version
AdoptOpeJDK 11.0.15
 
**Version since**
Worked OK in Eclipse 4.15, but not since

joel-majano:
This issue was caused by logic running Cairo scaling multiple times when it was not necessary.
There are two locations that were problematic, one, Image.java was applying a second scale when copying an image.
Two, ImageList.java was interpreting width/height returned from a Cairo function incorrectly.

+ Guarded cairo_surface_set_device_scale call if copy and original have the same device zoom in Image.java
+ Corrected width and height variables to interpret size correctly in ImageList.java

Fixes #315

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Image.java
@@ -274,7 +274,7 @@ public Image(Device device, Image srcImage, int flag) {
 	boolean hasAlpha = format == Cairo.CAIRO_FORMAT_ARGB32;
 	surface = Cairo.cairo_image_surface_create(format, width, height);
 	if (surface == 0) SWT.error(SWT.ERROR_NO_HANDLES);
-	if (DPIUtil.useCairoAutoScale()) {
+	if (DPIUtil.getDeviceZoom() != currentDeviceZoom && DPIUtil.useCairoAutoScale()) {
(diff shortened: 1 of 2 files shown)


Phillipus:
Nice. Is it too late for 4.25?

akurtakov:
This kind of changes are too risky as quite often they introduce regressions so I would prefer to defer it for next release.

Phillipus:
> This kind of changes are too risky as quite often they introduce regressions so I would prefer to defer it for next release.

Understood.

sravanlakkimsetti (approved):

AlexisDrogoul:
Thanks for the fix !

Issue #228, opened 2022-06-23
GTK4 Issue Tracker


joel-majano:
This issue serves as an umbrella issue for all GTK4 port related work.

The previous GTK4 issue tracker is here: https://bugs.eclipse.org/bugs/show_bug.cgi?id=540385.

To add to this list simply reference this issue's number.

medmedin2014:
One advice is to never use that cursed LibAdwaita adopted by GNOME apps, it locks the theming of the app to that ugly Adwaita, please us only pure GTK4 widgets. That library is hated on every Linux desktop.

probonopd:
Second this. [libadwaita doesn't work well](https://www.youtube.com/watch?v=WFqJA1q-IKY) outside of the Gnome desktop.  Switching to libadwaita would significantly reduce the potential user base for this application, essentially limiting it to the fraction of users who happen to use Gnome.

akurtakov:
Closing this one as it's a tracker and we have https://github.com/orgs/eclipse-platform/projects/7 for tracking purposes

PR #328, opened 2022-08-22
[GTK4] Text entered in StyledText does not show up until window resized

joel-majano:
When entering text using a keyboard in StyledText snippets or TextEditor example, text would not show up until the window was resized.

+ Added gtk_widget_queue_draw in update()

Tested with TextEditor.java and a simple StyledText snippet.

Tracked in #228


akurtakov:
Unrelated test failure.

PR #329, opened 2022-08-22
[GTK4] StyledText Accessible function calls flood the console with errors

joel-majano:
Temporarily disable the two main calls (TextChanged & CaretMoved) until Accessible is working in GTK4.

Tracked in #228


akurtakov:
StyledText is shared between all WS and on Mac/Win there is non GTK* classes so this is patch is total no-go.

Issue #333, opened 2022-08-23
Multiple org.eclipse.debug.internal.ui.viewers.model.TreeModelContentProvider$DelayedDoModelChangedJob errors if I start an applciation in debug mode


vogella:
Steps to reproduce:

1. Ensure Debug View is open
2. Debug child Eclipse or RCP app
3. Debug Tree is now corrupt and error dialog opens

With Eclipse SDK
Version: 2022-09 (4.25)
Build id: I20220821-1800
OS: Windows 10, v.10.0, x86_64 / win32
Java vendor: IBM Corporation
Java runtime version: 17.0.4+8
Java version: 17.0.4

I see multiple errors with the following stacktrace.
I only see this (currently) in a client installation, so I do not know how to reproduce.

```
java.lang.IllegalArgumentException: Index out of bounds
	at org.eclipse.swt.SWT.error(SWT.java:4899)
	at org.eclipse.swt.SWT.error(SWT.java:4833)
	at org.eclipse.swt.SWT.error(SWT.java:4804)
	at org.eclipse.swt.widgets.Widget.error(Widget.java:450)
	at org.eclipse.swt.widgets.Tree.getItem(Tree.java:3182)
	at org.eclipse.jface.viewers.TreeViewer.replace(TreeViewer.java:439)
	at org.eclipse.debug.internal.ui.viewers.model.TreeModelContentProvider.expand(TreeModelContentProvider.java:1429)
	at org.eclipse.debug.internal.ui.viewers.model.TreeModelContentProvider.handleExpand(TreeModelContentProvider.java:1396)
	at org.eclipse.debug.internal.ui.viewers.model.TreeModelContentProvider.handleExpand(TreeModelContentProvider.java:1394)
	at org.eclipse.debug.internal.ui.viewers.model.TreeModelContentProvider.updateNodes(TreeModelContentProvider.java:1300)
	at org.eclipse.debug.internal.ui.viewers.model.TreeModelContentProvider.updateNodes(TreeModelContentProvider.java:1312)
	at org.eclipse.debug.internal.ui.viewers.model.TreeModelContentProvider.updateNodes(TreeModelContentProvider.java:1312)
	at org.eclipse.debug.internal.ui.viewers.model.TreeModelContentProvider.updateModel(TreeModelContentProvider.java:578)
	at org.eclipse.debug.internal.ui.viewers.model.TreeModelContentProvider.doModelChanged(TreeModelContentProvider.java:543)
[...]

Issue #287, opened 2022-08-04
Regression in Tree from Bug 575787


jonahgraham:
[Bug 575787](https://bugs.eclipse.org/bugs/show_bug.cgi?id=575787) which optimized `Tree.setItemCount` causes `Tree.indexOf(TreeItem)` to return negative numbers.

This issue was originally reported as a bug with CDT in [Bug 580518](https://bugs.eclipse.org/bugs/show_bug.cgi?id=580518)

**To Reproduce**

At the moment the steps to reproduce involve using CDT on Windows:

1. Install CDT 10.7 in Platform 4.24, gcc+gdb will be needed too
2. Create two Hello World projects
3. In project A insert 5 breakpoints
4. In project B insert 2 breakpoints
5. Debug project B
6. In the Breakpoints view toggle "Show Breakpoints Supported by Selected Target"

I'll try to find time to make a standalone test case. The above test case uses virtual trees + platform debug's async view model which makes it fairly complicated.

**Expected behavior**

When toggle is on, only 2 breakpoints of project B displayed. When toggle is off, all 7 breakpoints displayed.

**Actual behaviour**

When toggled back off, the breakpoints view is "corrupted" - see screenshot. There is also an exception in the CDT code due to the index not being in the expected range of >= -1. See [bug 580518](https://bugs.eclipse.org/bugs/show_bug.cgi?id=580518#c2) for full stack trace.

**Screenshots**

![image](https://user-images.githubusercontent.com/679236/182942391-022cf90b-9ba9-4352-9a37-8e5417fae369.png)

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 

Definitely Windows, can't seem to reproduce on Linux.

7. Additional OS info (e.g. OS version, Linux Desktop, etc)

8. JRE/JDK version

JustJ Java 17 as shipped with EPP.
 
**Version since**

Since commit 17b4024759c8084afa680a06b7b2803463e84114 - reverting 17b4024759c8084afa680a06b7b2803463e84114 resolves the issue.

jonahgraham:
Instead of the steps in my original I can reproduce with Java + some other debug environment. By creating 5ish Java breakpoints and 5ish non-Java breakpoints and then toggling "Show Breakpoints Supported by Selected Target" while debugging the Java project will lead to the same problem, with the following exception:

```
[...]

PR #347, opened 2022-08-25
Bug 575787 - [win32] Optimize Tree.setItemCount

SyntevoAlex:
On Windows, Tree maintains its items with a single linked list. For this
reason, "find previous" operation is O(N) because it requires walking
from start of the list.

When inserting an item, it needs to be linked into the list, which
involves finding the previous item and updating its 'next' field, which
is O(N) for one item and O(N*N) for inserting N items.

The workaround is to bulk insert all items at the same position. This
means that if TreeItem had no children, inserting N items into it is
O(N) instead of O(N*N).

On top of that, this patch preallocates 'Tree.items[]' to hold exactly
the required number of items.

```
Resulting performance improvement on my machine:
                |      before patch     |      after patch      |
                +-----------+-----------+-----------+-----------+
                |   REGULAR |   VIRTUAL |   REGULAR |   VIRTUAL |
----------------+-----------+-----------+-----------+-----------+
   10_000 items |   0,44sec |   0,43sec |   0,07sec |   0,06sec |
  100_000 items | 113,83sec | 111,95sec |   0,70sec |   0,63sec |
1_000_000 items |  too long |  too long |   6,82sec |   6,28sec |
```

This patch combines:
* Original patch for Bug 575787 (the parts that were not reverted later)
* Fix for Issue #287
* Fix for Issue #333
* Converting some test snippets into JUnit tests

Signed-off-by: Alexandr Miloslavskiy <alexandr.miloslavskiy@syntevo.com>


SyntevoAlex:
Hopefully this patch will not trigger any more regressions. Sorry about how it went before!

niraj-modi:
@Phillipus @vogella 
Let's test this patch: specifically self-hosted Eclipse use-cases.
Anyways we have some time, targeted it for 4.26 M1.

SyntevoAlex:
Let's merge maybe?

akurtakov:
Please do . I don't have access to Windows to test so can't verify it.

SyntevoAlex:
OK, merged it. Hopefully it will be a success this time

iloveeclipse:
> OK, merged it. Hopefully it will be a success this time

Sorry for that, I fear we have regression, see #396

Issue #360, opened 2022-09-01
Slow UI Freeze Monitoring


jukzi:
I never found the "UI Freeze Monitoring" very helpful (far less information then sampling with visualVM) but i just learned, that itself can cause large part of freezes.

**To Reproduce**
In a eclipse Platform workspace click on any item in the BrandingInfo of the OOmph configuration:
![image](https://user-images.githubusercontent.com/51790620/188027303-8fea0306-eb75-408b-98fc-97443cff6a10.png)
=> ~ 3 sec lag - but no freeze report (LOL)

**Expected behavior**
no freeze

**Screenshots**
sampled with VisualVM:
![image](https://user-images.githubusercontent.com/51790620/188028117-d6458390-9276-4c55-83b0-ca8ab2d39510.png)
=> most time (2s) is spend in org.eclipse.swt.widgets.Display.getLastEventTime() due to 
sendPostExternalEventDispatchEvent(), sendPreExternalEventDispatchEvent (totally 2.3/3 seconds)
![image](https://user-images.githubusercontent.com/51790620/188029044-b7c92ec7-cf0e-4b6c-bda1-bc753c5d17d8.png)

**Environment:**
- - [x] Windows
- - [?] Linux
- - [?] macOS 

2. Additional OS info (e.g. OS version, Linux Desktop, etc)
Win 10
3. JRE/JDK version
JDK 11
 
**Version since**
probably with initial commit of UI Freeze Monitoring 2014

**Workaround (or) Additional context**
PreExternalEventDispatch and PostExternalEventDispatch are rapidly send via Display event without setting a time in the event, which is then slowly requested from OS.
Setting the event time to any value != 0 should solve the problem (time is not read from those events anyway - see org.eclipse.ui.internal.monitoring.EventLoopMonitorThread.EventLoopState.handleEvent(Event). 

It would also be a useful to not send those events if UI Freeze Monitoring is disabled.

It could also be analysed if the slow getLastEventTime() is useful for any call of org.eclipse.swt.widgets.Display.sendEvent(int, Event). If the event time is only used for debugging purpose and measure mouse DoubleClickTime it would be enough to use System.nanotime/100_000.

mickaelistria:
> I never found the "UI Freeze Monitoring" very helpful 

[...]

PR #362, opened 2022-09-02
WIN: do not query OS lastEventTime for UI Freeze Monitoring #360

jukzi:
getLastEventTime() can be slow on windows and is not needed for UI
Freeze Monitoring


jukzi:
only unrelated errors from tests that fail anyway

vogella:
Should the same change be done on Linux and Mac?

jukzi:
> Should the same change be done on Linux and Mac?

Probably its not an issue on Linux<:
GTK: getLastEventTime() returns Display.lastEventTime -> fast

cocoa: i don't know
.
However it might better to keep implementations similar to even if it's not a problem on Linux/Mac. 
WDYT?

vogella:
> However it might better to keep implementations similar to even if it's not a problem on Linux/Mac.
WDYT?

Yes, keeping the implementations the same would be good

Issue #114, opened 2022-05-10
[win32] Test_org_eclipse_swt_widgets_Display has classpath configuration problems


laeubi:
As one can see here https://github.com/eclipse-platform/eclipse.platform.swt/runs/6372153691 the test fails to execute in maven with the following error:

> java.lang.IllegalAccessError: class org.eclipse.swt.widgets.Test_org_eclipse_swt_widgets_Display tried to access method 'boolean org.eclipse.swt.widgets.Display.isXMouseActive()' (org.eclipse.swt.widgets.Test_org_eclipse_swt_widgets_Display is in unnamed module of loader org.eclipse.osgi.internal.loader.EquinoxClassLoader @a202ccb; org.eclipse.swt.widgets.Display is in unnamed module of loader org.eclipse.osgi.internal.loader.EquinoxClassLoader @4cc6fa2a)
> 	at org.eclipse.swt.widgets.Test_org_eclipse_swt_widgets_Display.test_isXMouseActive(Test_org_eclipse_swt_widgets_Display.java:24)

jukzi:
The error can be reproduced in IDE if and only if instead of running a simple Junit test explicitly a Junit "Plug-in Test" for org.eclipse.swt.tests.win32 is started.

HeikoKlare:
Those tests are not supposed to be executed as JUnit plug-in tests anyway. They create a display for the current thread, which fails as a display is already present in case of execution as JUnit plug-in test. The execution as plain JUnit tests does not seem to produce any such problems.

PR #363, opened 2022-09-02
win32: fix test_isXMouseActive #114

jukzi:
bundles can not share package.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
@@ -2859,7 +2859,11 @@ public void internal_dispose_GC (long hDC, GCData data) {
 	OS.ReleaseDC (0, hDC);
 }
 
-boolean isXMouseActive () {
+/**
+ * Invokes platform specific functionality
+ * @noreference This method is not intended to be referenced by clients.
+ */
+protected boolean isXMouseActive () {
 	/*
 	* NOTE: X-Mouse is active when bit 1 of the UserPreferencesMask is set.
 	*/
--- a/tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/tests/win32/AllWin32Tests.java
+++ b/tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/tests/win32/AllWin32Tests.java
@@ -14,7 +14,6 @@
  */
 package org.eclipse.swt.tests.win32;
 
-import org.eclipse.swt.widgets.Test_org_eclipse_swt_widgets_Display;
 import org.junit.runner.JUnitCore;
 import org.junit.runner.RunWith;
 import org.junit.runners.Suite;
--- a/tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/tests/win32/Test_org_eclipse_swt_widgets_Display.java
+++ b/tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/tests/win32/Test_org_eclipse_swt_widgets_Display.java
@@ -11,15 +11,27 @@
  * Contributors:
  *     Joerg Kubitz - initial API and implementation
  *******************************************************************************/
-package org.eclipse.swt.widgets;
+package org.eclipse.swt.tests.win32;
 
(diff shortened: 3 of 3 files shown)


laeubi:
> bundles can not share package.

Can you explain? Fragments actually can *extend* a package of the host ...

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
```diff
@@ -2859,7 +2859,11 @@ public void internal_dispose_GC (long hDC, GCData data) {
 	OS.ReleaseDC (0, hDC);
 }
 
-boolean isXMouseActive () {
+/**
+ * Invokes platform specific functionality
+ * @noreference This method is not intended to be referenced by clients.
+ */
+protected boolean isXMouseActive () {
```
akurtakov, line 2866: You can not add such API as it is not existing on Mac/Linux, lacks documentation what is it about at all and probably doesn't make any sense there at all.

jukzi: doesn't @noreference prevent it from being API?

jukzi:
> Fragments actually can _extend_ a package of the host ...

how - so that the problem is solved?

laeubi:
https://github.com/eclipse-platform/eclipse.platform.swt/blob/dd9fe8ccf43b533a4de9d347abf282cf95e87a21/tests/org.eclipse.swt.tests.win32/META-INF/MANIFEST.MF should be a fragment of `org.eclipse.swt`

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
akurtakov: Java API has to be the same for all ports, @noreference is just a documentation from Java POV which happen to have special treatment by PDE.

akurtakov:
> https://github.com/eclipse-platform/eclipse.platform.swt/blob/dd9fe8ccf43b533a4de9d347abf282cf95e87a21/tests/org.eclipse.swt.tests.win32/META-INF/MANIFEST.MF should be a fragment of `org.eclipse.swt`

I don't think this will work as one fragment doesn't see classes from another fragment and real swt implementation comes from platform specific fragments and org.eclipse.swt is just a shell.

laeubi:
> I don't think this will work as one fragment doesn't see classes from another fragment and real swt implementation comes from platform specific fragments and org.eclipse.swt is just a shell.

All fragments of a host are merged within one classpace, see:
http://docs.osgi.org/specification/osgi.core/7.0.0/framework.module.html#i3174728

`Bundle Space - The bundle space consists of the JAR file that is associated with the bundle, plus any additional JAR that are closely tied to the bundle, like fragments, see [Fragment Bundles](http://docs.osgi.org/specification/osgi.core/7.0.0/framework.module.html#framework.module.fragmentbundles)`

and http://docs.osgi.org/specification/osgi.core/7.0.0/framework.module.html#framework.module.fragmentbundles

> A host bundle's class path is searched before a fragment's class path. This implies that packages can be split over the host and any of its fragments.

akurtakov:
> All fragments of a host are merged within one classpace, see: http://docs.osgi.org/specification/osgi.core/7.0.0/framework.module.html#i3174728
> 
> > Bundle Space - The bundle space consists of the JAR file that is associated with the bundle, plus any additional JAR that are closely tied to the bundle, like fragments, see [Fragment Bundles](http://docs.osgi.org/specification/osgi.core/7.0.0/framework.module.html#framework.module.fragmentbundles)
> 
> and http://docs.osgi.org/specification/osgi.core/7.0.0/framework.module.html#framework.module.fragmentbundles
> 
> > A host bundle's class path is searched before a fragment's class path. This implies that packages can be split over the host and any of its fragments.

Then it should be an issue with tycho/maven-surefire/etc. as I had to revert the effort trying to make swt.tests a fragment https://github.com/eclipse-platform/eclipse.platform.swt/issues/232#issuecomment-1192172080

laeubi:
Maybe but hard to tell what actually was/is the issue fragments are somehow special and it is possible that such a setup requires additional setup in Tycho.

jukzi:
according to your comments this is the wrong direction

PR #370, opened 2022-09-03
[win32] Fix various warnings in native libs, treat any new warnings as errors

### LINKED ISSUE #369: [win32] Native libs: treat warnings as errors
I noticed that some warnings occur when building native libs on Windows. These easily get overlooked, because nobody is watching. Previously, in Bug 570178, I already changed GTK native libs to treat build warnings as errors. Let's do the same for Windows.

SyntevoAlex:
Previously, I already changed GTK build to treat warnings as errors, see Bug 570178

The commits come in reverse order: first, (not yet visible) warnings are fixed, and then strict `/W4` is enabled. This is so that each commit can be built successfully and there are no broken commits in the chain.

If you want to see the warnings being fixed (for review purposes), move the last commit to be the first. On the other hand, the nature of the warnings is rather evident from commits that fix them.

The patch set only targets Windows. Other platforms are only trivially affected by injecting new Windows-specific settings into all JNI native sources.

Noteworthy commits (all of them at the end):
* `Get rid of deprecated GetVersion()`
  Clears most of Windows version checks from SWT.
* `Deal with warnings about unreferenced parameters`
  Changes `JniGenerator` to inject warning settings into all generated sources.
* `Fix potentially uninitialized variable`
  That seems to be a good catch, where warning pointed to a potential JVM crash.

Fixes #369

### DIFF
--- a/bundles/org.eclipse.swt.tools/JNI Generation/org/eclipse/swt/tools/internal/NativesGenerator.java
+++ b/bundles/org.eclipse.swt.tools/JNI Generation/org/eclipse/swt/tools/internal/NativesGenerator.java
@@ -76,12 +76,13 @@ public void generate(JNIClass clazz) {
 	if (i == methods.length) return;
 	sort(methods);
 	generateNativeMacro(clazz);
+	generateWarningSettings();
 	generateExcludes(methods);
 	generate(methods);
 }
 
 public void generate(JNIMethod[] methods) {
-	sort(methods);	
+	sort(methods);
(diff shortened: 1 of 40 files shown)


akurtakov:
@niraj-modi The core of this patch is win32 specific thus please review.

akurtakov:
@SyntevoAlex If you're confident enough please merge. It's best to have these changes in M1.

SyntevoAlex:
I don't expect trouble. Let's merge.

Issue #340, opened 2022-08-24
[a11y] - Content Types Preference description can't be read by JAWS


niraj-modi:
Steps to reproduce on Windows:
1. Open JAWS
2. Launch Eclipse
3. Click Window > Preferences > General > Content Types
![Eclipse_Preference_ContentType_JAWS](https://user-images.githubusercontent.com/92304759/186392819-7a02a9f1-3bd9-4d83-9ecb-17a09100a677.png)

===== EXPECTED RESULT =====
The content should be read out for user.

niraj-modi:
Based on discussion in https://github.com/eclipse-platform/eclipse.platform.ui/pull/288#issuecomment-1225653035
Will share a PR on SWT side shortly.

niraj-modi:
Verified using Build id: I20220824-1800 on Win10 using JAWS 2021

niraj-modi:
Fix is back-ported to R4_23_maintenance:
- Also done with the required SWT build input on R4_23_maintenance branch.

Thanks @sravanlakkimsetti for your quick help in fixing the SWT maintenance jobs.

PR #372, opened 2022-09-05
[a11y] - Content Types Preference description can't be read by JAWS

niraj-modi:
- Back-port fix to 4.23 #340


niraj-modi:
Fix back-port to R4_23_maintenance

Issue #356, opened 2022-08-29
Loading an (out of spec?) .bmp produces blank ImageData


rcblanke:
When loading [a specific mobygames.com image](https://www.mobygames.com/images/shots/l/26512-transylvania-dos-screenshot-you-are-facing-an-ancient-stump.gif) with new ImageData(InputStream), the resulting ImageData data is empty, that is, the entire byte[] contains zeros resulting in a black 320x200 image when displayed on screen. Please note that although the url suggests a .gif, the image is in fact a .bmp file.

The reason appears to be the fact that the .bmp's 'compressedSize', calculated using its infoHeader (see line 469 in org.eclipse.swt.internal.image.WinBMPFileFormat, SWT version 4.24) comes out as being 0. This in turn results in 0 bytes being read as 'compressed data'.

According to the [BMP file format specs](https://en.wikipedia.org/wiki/BMP_file_format), 'the image size' in the [Windows BITMAPINFOHEADER]  is "the size of the raw bitmap data; a dummy 0 can be given for BI_RGB bitmaps". The Mobygames .bmp in question is compressed using BI_RLE4 (cmp on line 460 equals 2), though.

Thus, one might argue that the current behaviour is correct. On the other hand, browsers and image viewers that I've tried all seem to display the .bmp just fine. Also, the 'spec' at wikipedia might be incomplete, although I've not been able to find any further information on this unusual situation (with cmp!=0 and compressedSize==0).

That's why I would like to suggest to handle this specific case such that either:

1. loading the file in question throws SWT.ERROR_INVALID_IMAGE so that the programmer can at least expect that rendering it may be problematic
2. or find a workaround to calculate the correct 'compressedSize' (maybe using the [Bitmap file header] [size of the BMP file in bytes] - [The offset, i.e. starting address, of the byte where the bitmap image data (pixel array) can be found])

FWIW I've tried the latter, and this seems to work properly _for this specific .bmp_ (that is, it decompresses and displays correctly).

Kindest regards,
Ronald

mickaelistria:
Would you qualify the image as "correct"? SWT and Eclipse in general is a bit too short on code cintribution resources to focus in incorrect cases.
But if you're willing to implement an improvement that helps supporting a common sensible practice,  code contributions are welcome in Pull Requests.

rcblanke:
This really seem to be an edge case, though I believe SWT can at least do a little bit better. Please take a look at my PR, questions and comments are very much welcome!

[...]

PR #374, opened 2022-09-06
fix issue 356  .bmp produces blank ImageData

rcblanke:
[fix issue 356 .bmp produces blank ImageData](https://github.com/eclipse-platform/eclipse.platform.swt/issues/356)


mickaelistria:
Would it be possible for you to add a test case for this, adding a minimal bmp file with this particular format and verifying the WinBMPFileFormat (or other consumer) can process it decently?

mickaelistria:
As mentioned on the ticket, a better implementation would be to replace loading of images using this custom class by the Windows image loading API directly, so Eclipse IDE would be as capable as the OS.
Do you think you can work on such alternative approach? That would be much more sustainable and reliable.

rcblanke:
I've implemented a JUnit test for this issue.

Sorry, redoing the Windows ImageLoader implementation using OS system calls is out of my league.

Regards,
Ronald

SyntevoAlex:
I spent some time to understand the problem and it seems that the value of `BITMAPINFOHEADER.biSizeImage` is not interesting. When bitmap is decompressed, it does so until it finds "end of bitmap" marker, see
* https://learn.microsoft.com/en-us/windows/win32/gdi/bitmap-compression
* https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/bundles/org.eclipse.swt/Eclipse%20SWT/common/org/eclipse/swt/internal/image/WinBMPFileFormat.java#L312

To double check, I had a look at how GTK (the main Linux implementation for image loading) does it:
https://gitlab.gnome.org/GNOME/gdk-pixbuf/-/blob/master/gdk-pixbuf/io-bmp.c#L256

It ignores `BITMAPINFOHEADER.biSizeImage` completely (No uses of `BIH[20]`).

To my understanding, SWT only uses `compressedSize` to pre-load stream into `byte[]` which seems rather useless.
I understand that SWT code shall be changed as follows:
1) Remove `int compressedSize` and `byte[] compressed` completely.
2) In `decompressRLE4Data()` and `decompressRLE8Data()` use `inputStream` directly to read bytes from it.

This would fix the problem by getting rid of the code that causes trouble.

lshanmug:
> I spent some time to understand the problem and it seems that the value of `BITMAPINFOHEADER.biSizeImage` is not interesting. When bitmap is decompressed, it does so until it finds "end of bitmap" marker, see
> 
>     * https://learn.microsoft.com/en-us/windows/win32/gdi/bitmap-compression
> 
>     * https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/bundles/org.eclipse.swt/Eclipse%20SWT/common/org/eclipse/swt/internal/image/WinBMPFileFormat.java#L312
> 
> 
> To double check, I had a look at how GTK (the main Linux implementation for image loading) does it: https://gitlab.gnome.org/GNOME/gdk-pixbuf/-/blob/master/gdk-pixbuf/io-bmp.c#L256
> 
> It ignores `BITMAPINFOHEADER.biSizeImage` completely (No uses of `BIH[20]`).
> 
> To my understanding, SWT only uses `compressedSize` to pre-load stream into `byte[]` which seems rather useless. I understand that SWT code shall be changed as follows:
> 
>     1. Remove `int compressedSize` and `byte[] compressed` completely.
> 
>     2. In `decompressRLE4Data()` and `decompressRLE8Data()` use `inputStream` directly to read bytes from it.
> 
> 
> This would fix the problem by getting rid of the code that causes trouble.

@SyntevoAlex Thanks for investigating.
@rcblanke  WDYT? Can you make the suggested changes?

akurtakov:
Is there some interest in this one still?

akurtakov:
Closing due to no response to my question about interest in this one from 2022 although there have been some rebases.

Issue #110, opened 2022-05-10
Test_org_eclipse_swt_browser_Browser.test_LocationListener_ProgressListener_noExtraEvents fails since I20220508-1800


iloveeclipse:
`Test_org_eclipse_swt_browser_Browser.test_LocationListener_ProgressListener_noExtraEvents` fails since I20220508-1800.

See https://download.eclipse.org/eclipse/downloads/drops4/I20220508-1800/testresults/html/org.eclipse.swt.tests_ep424I-unit-cen64-gtk3-java11_linux.gtk.x86_64_11.html

```
Incorrect event sequences. Events missing or too many fired: Expected one of each, but received: Changed count: 1 Completed count: 0

java.lang.AssertionError:
Incorrect event sequences. Events missing or too many fired:
Expected one of each, but received:
Changed count: 1
Completed count: 0
at org.junit.Assert.fail(Assert.java:89)
at org.junit.Assert.assertTrue(Assert.java:42)
at org.eclipse.swt.tests.junit.Test_org_eclipse_swt_browser_Browser.test_LocationListener_ProgressListener_noExtraEvents(Test_org_eclipse_swt_browser_Browser.java:479)
```

iloveeclipse:
See https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/1274

laeubi:
I found that the Browser Tests are very fragile in regard to OS and configuration.

PR #376, opened 2022-09-08
Fix Browser tests failing in I-build

akurtakov:
Wait for the events that have to appear first before checking that no extra events are received.
Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/110


akurtakov:
Test failure is seen  on Jenkins for quite some time now but not in I-builds and has to be invested separately

PR #382, opened 2022-09-13
[GTK4] Add WebkitGtk support

akurtakov:
Now that Fedora 37 ships webkitgtk compiled against Gtk 4.x the new *.so name is finally known so we know which lib to dlopen. Guards preventing Browser instantiation when running Gtk 4.x are removed.
Extra change was to no longer manually connect signals like button_press/release, focus_in/out and etc. as they are by default now as per https://trac.webkit.org/changeset/262184/webkit (the web view is always the target of the events, so we don't really  need to send the events to GTK to process them.)


akurtakov:
Screenshot from 2022-09-13 09-20-20 showing it works!

Issue #292, opened 2022-08-07
Support HTML as a Clipboard format for StyledText


mihnita:
The current `StyledText` can copy / paste styled text to clipboard, but only in RTF format.

If you copy a piece of code (with syntax) and paste it in MS Word, or PowerPoint, the colors and font are preserved
(although the background color is lost).
But only plain text is pasted into Google GMail, Docs, Slides, or Yahoo Mail.

**Describe the solution you'd like**

With web-based applications as common as they are, it would be really handy to also support HTML as a clipboard format.
So that one can copy code and paste it in something like Google Slides, for example, while preserving the colors (and font).
The way it works today in PowerPoint.

**Describe alternatives you've considered**

Copy / paste in MS Word works, then copy / paste from MS Word to web-based apps works.
Opening the code in Visual Studio Code and copying from there also works.

But it would be nice to have that working from Eclipse.

**Additional context**

Not much, I think it is pretty obvious.

BUT: I offer to implement this, if you are interested.

Thank you very much,
Mihai

mickaelistria:
Such a contribution would be very welcome!

mihnita:
Hi, 

Sorry, nothing new, but I just to make sure I manage expectations :-)

I did not drop out, but I think this is not something for 2022-09.
My guess is that it is a bit too late for that.
(I also don't have much time, so I would rather focus on the color support in the Console, for now)

But I can share a bit the big picture plan:
* Optional: extract `TextWriter` and `RTFWriter` out of `org.eclipse.swt.custom.StyledText`
* Add an `HTMLWriter`, very similar to the `RTFWriter`
* Change the `StyledText` copy operation to also put the HTML in the clipboard

The HTML I would produce is very "primitive" for best compatibility.
Using `<b>` and `<i>` and `<span style=...>`, no stylesheets or anything fancy.
I've compared the kind of HTML put in the clipboard by all kinds of apps, from (the "ancient") Office 2010 to Google Slides/Docs/GMail.
And that kind of HTML seems to be the norm.

Regards,
Mihai

tmssngr:
Maybe such code can be placed outside the (already large) `StyledText` code?

mickaelistria:
@tmssngr I believe this is what was meant with "Optional: extract TextWriter and RTFWriter out of org.eclipse.swt.custom.StyledText"

mihnita:
Agree with `HTMLWriter` outside `StyledText`.

[...]

PR #388, opened 2022-09-16
Extract TextWriter & RTFWriter from StyledText (issue #292)

mihnita:


mickaelistria:
Failure is not related. Let's merge. Thanks!

PR #391, opened 2022-09-17
[win32] Native libs: treat warnings as errors

### LINKED ISSUE #369: [win32] Native libs: treat warnings as errors

SyntevoAlex:
Fixes #369


SyntevoAlex:
@akurtakov I understand that Niraj Modi is currently busy, so there's nobody to review and I will just merge it. This time, it fixes even more warnings, so outdated Eclipse build machines with VS2019 can also build it fine.

PR #394, opened 2022-09-18
Fix keyboard handling issues

### LINKED ISSUE #351: Fix keyboard handling issues
There is a number of issues when handling key presses in Windows, such as:
* [win32] JVM crash when pressing `1` in Nepali keyboard layout
* [win32] Bug 536041: Digit keys in Lithuanian type additional "trash" characters
* [win32] Digit keys in French and many other languages don't invoke corresponding keyboard shortcuts
* [win32] Various issues with dead keys reporting
* [cocoa] Bug 573487: Keyboard shortcuts don't work in non-latin layouts
* etc

I'm already working on a patch to fix these.

SyntevoAlex:
This is something I've been working on for a while. It fixes a number of keyboard-related issues in SWT.

Due to the nature of the changes, it may be possible that some scenarios have been changed for the worst. I did quite a bit of testing, and don't expect immediate trouble, but it's still possible. I guess, the only way to find the corner cases is to merge and see.

This PR introduces massive documentation that follows my (also massive) research. It also provides debugging snippets for all platforms - Java, win32, gtk, cocoa.

On Windows, I fixed many issues, and introduced A LOT of unit tests. Sadly, I didn't have the energy to extend the tests to other platforms.

On macOS, there's just a single change, hopefully it will be enough to address most problems.

Probably "Noteworthy", especially on macOS, where keyboard shortcuts shall now work in non-latin layouts (Bulgarian, Arabic, Japanese, Chinese, etc etc).

Fixes #0351


akurtakov:
@SyntevoAlex There were extra test failures with this patch. I've rebased it now but please pay attention and push yourself when there are no regressions.

SyntevoAlex:
The test failures are because GitHub tester does not build native libraries. Yes, yes, again the same problem.
The patch set itself should be fine and tests well when tested locally.
So, shall I merge it?

iloveeclipse:
> The test failures are because GitHub tester does not build native libraries.

Did you mean this fail: org.eclipse.swt.tests.junit.Test_org_eclipse_swt_custom_StyledText.test_caretSizeAndPositionVariableGlyphMetrics

If yes, it seem to happen also without your patch.

SyntevoAlex:
I think that @akurtakov meant numerous `org.eclipse.swt.tests.win32.Test_org_eclipse_swt_events_KeyEvent` failures here:
https://github.com/eclipse-platform/eclipse.platform.swt/runs/8652818452

akurtakov:
In that case feel free to merge yourself. As changes are windows (mostly) and you're the most active win32 developer it doesn't look like you'll get reviews.

SyntevoAlex:
Alrighty, let's have it. I did quite a bit of testing, but still, the patches change a lot, so regressions are possible. If you come across anything keyboard input related for Windows and macOS that happens after this merge, please let me know.

PR #397, opened 2022-09-20
Try older image for swtbuild

mickaelistria:


mickaelistria:
@akurtakov this extra test shows the getTextBounds seems to behave differently than expected, basically considering the whole line height instead of the character "box".

PR #402, opened 2022-09-23
Extracting common code between RTFWriter and HTMLWriter in a base class

mihnita:
I don't know if it is generic enough to support other future formats.
But at least it puts in a common place the relatively complicated code that iterates and  outputs the line styled ranges.

I've tried to split it into separate commits that might be easier to follow, but I don't know if you prefer that style, or one single commit.
Let me know and I can squash.
Or only squash after you reviewed.

Also, if you think it is overengineered and not worth doing, that's also fine.

Thanks,
Mihai

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/HTMLWriter.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/HTMLWriter.java
@@ -225,12 +225,28 @@ class HTMLWriter extends StyledTextWriterBase {
 	// ==== Helper methods ====
 
 	@Override
-	String escapeText(String text) {
-		return text
-				.replaceAll("&", "&amp;")
-				.replaceAll("\"", "&quot;")
-				.replaceAll("<", "&lt;")
-				.replaceAll(">", "&gt;");
+	String escapeText(String string) {
+		// The resulting string is at least as long as the original.
+		StringBuilder result = new StringBuilder(string.length());
+		string.chars().forEach(ch -> {
+			switch (ch) {
+				case '&':
+					result.append("&amp;");
+					break;
+				case '"':
+					result.append("&quot;");
+					break;
+				case '<':
+					result.append("&lt;");
+					break;
+				case '>':
+					result.append("&gt;");
+					break;
+				default:
+					result.append((char) ch);
+			}
+		});
+		return result.toString();
 	}
 
 	// TODO: do we also want support for alpha?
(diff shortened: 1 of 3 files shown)


mihnita:
Note: I was unable to test properly.

Fails in command line:
```
[ERROR] Failures:
[ERROR]   Test_org_eclipse_swt_SWT.test_isLocal:180 Tested SWT is signed by Eclipse: (SWT='C:/Users/mihai/.m2/repository/p2/osgi/bundle/org.eclipse.swt.win32.win32.x86_64/3.121.100.v20220919-1402/org.eclipse.swt.win32.win32.x86_64-3.121.100.v20220919-1402.jar' signer='1.2.840.113549.1.9.1=#16157765626d61737465724065636c697073652e6f7267,CN=Eclipse.org Foundation\, Inc.,OU=IT,O=Eclipse.org Foundation\, Inc.,L=Ottawa,ST=Ontario,C=CA')
[INFO]
[ERROR] Tests run: 3170, Failures: 1, Errors: 0, Skipped: 38
```
Worse part is that the failure seems to hint that I am not even testing the changes I've made, I am testing the wrong artifact.

And fails (differently) in Eclipse.

Bot nothing seems related to what I did.
Will see what Jenkins things about it :-)

---

How I tested:
* I've used the SWT app I shared before (that creates all kind of styled text, copy, then dump all the formats in the clipboard).
* Added a line with characters that need escaping the rtf/html (`&<>"'` and `\{}`, + some Unicode)
* Dumper all clipboard formats before / after refactoring and binary compared. They are identical.

Cheers,
M.

SyntevoAlex:
> Test_org_eclipse_swt_SWT.test_isLocal:180 Tested SWT is signed by Eclipse
> failure seems to hint that I am not even testing the changes I've made

Yes, it means exactly that :)  I've be trying to figure it out in #251 #252, but I remain rather confused. To my understanding, local maven builds are broken in a sense that they always test something irrelevant, and nobody has the energy to fix that. GitHub side tests seem to run properly... kind of... maybe

mihnita:
Anyway...

I did test what I did though, the best I could :-)

---

Command line I get the same failures with jdk 11.
It looks like some artifact somewhere was built with 17 and ended up in maven.
```
Caused by: java.lang.UnsupportedClassVersionError: org/eclipse/tycho/p2maven/repository/P2ArtifactRepositoryLayout has been compiled by a more recent version of the Java Runtime (class file version 61.0), this version of the Java Runtime only recognizes class file versions up to 55.0
```

Building with jdk 17 I get that signature / local build error described above.

---

Running the unit tests in Eclipse worked, going through the code I just wrote.
I can step through it, and introducing an intentional mistake makes the test fail.
There are other tests failing there, but they are not related to my code (they also fail without my changes)

So that is testing.

---

And I've also tested with my own small app as described (diff the results before / after the refactoring).
See https://mihai-nita.net/eclipse_clip_html/ for the app.
It creates a StyledText widget, populates it with test and applies all possible styles on various regions,
copy to clipboard, then iterates all clipboard, all the formats, and saves the content in files.
That also testes the new code (changing the escape method resulted in the expected file differences)

---

**TLDR**
I am confident that my changes are fine, and **I did** test them.
But it is somewhat inconvenient that I can't test the way the CI does.

Mihai

mickaelistria:
@SyntevoAlex would you like to finalize review here?

mickaelistria:
> I have somewhat exceeded my limit for Eclipse issues that are of no interest to my boss. So I would rather prefer that someone else reviews it, sorry

OK, no problem.

PR #405, opened 2022-09-23
Bug 399 - SWT.MouseDown click not regognized in tree component

jasonhy-wang:
+ Call postOrSend function for SWT.MouseDown signal

This issue was caused by a previous fix for another issue where SWT.MouseDown always returns before calling postOrSend, however this should only be the case if event is null other wise postOrSend function should be called

Signed-off-by: Jason Wang (jaswan@redhat.com)


akurtakov:
So I tested this on Gtk 3 and 4 on Fedora 37 and didn't manage to reproduce the misisng mousedown with Snippet90.

PR #408, opened 2022-09-25
[Issue 367] Button Image is wrong size at hi-res scale

### LINKED ISSUE #367: [GTK] Button Image is wrong size at hi-res scale
**Describe the bug**
This follows on from https://github.com/eclipse-platform/eclipse.platform.ui/issues/261 where buttons with color images are not displayed correctly on Linux on hi-res screens (second screenshot in that bug report related to [ColorSelector](https://github.com/eclipse-platform/eclipse.platform.ui/blob/master/bundles/org.eclipse.jface/src/org/eclipse/jface/preference/ColorSelector.java)). To test, set environment variable  `GDK_SCALE=2` on low-res screens.

This is only happening on Linux. Mac and Windows work correctly.

**To Reproduce**
Snippet to follow

**Expected behavior**
The button image should be the correct size.

**Environment:**
- - [ ] All OS
- - [ ] Windows
- - [x] Linux
- - [ ] macOS 

- Linux Mint Cinnamon
- Hi-res scaling (set `GDK_SCALE=2` on normal monitor)

4. JRE/JDK version
- Temurin 11 JDK
 
**Version since**
Present in Eclipse 4.23 and later, probably earlier versions, too

majanojoel:
Button image was incorrectly scaled due to set_from_surface not behaving as expected.

+ Refactored image setting to use GdkPixbuf instead, which fixes the issue.
+ Added gtk_image_set_from_pixbuf to GTK3.java

Signed-off-by: Joel Majano <majanojoel@gmail.com>

Fixes #367

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT PI/gtk/library/gtk3.c
+++ b/bundles/org.eclipse.swt/Eclipse SWT PI/gtk/library/gtk3.c
@@ -1118,6 +1118,16 @@ fail:
 }
 #endif
 
+#ifndef NO_gtk_1image_1set_1from_1pixbuf
+JNIEXPORT void JNICALL GTK3_NATIVE(gtk_1image_1set_1from_1pixbuf)
+	(JNIEnv *env, jclass that, jlong arg0, jlong arg1)
+{
+	GTK3_NATIVE_ENTER(env, that, gtk_1image_1set_1from_1pixbuf_FUNC);
+	gtk_image_set_from_pixbuf((GtkImage *)arg0, (GdkPixbuf *)arg1);
+	GTK3_NATIVE_EXIT(env, that, gtk_1image_1set_1from_1pixbuf_FUNC);
+}
+#endif
+
 #ifndef NO_gtk_1image_1set_1from_1surface
 JNIEXPORT void JNICALL GTK3_NATIVE(gtk_1image_1set_1from_1surface)
 	(JNIEnv *env, jclass that, jlong arg0, jlong arg1)
--- a/bundles/org.eclipse.swt/Eclipse SWT PI/gtk/library/gtk3_stats.c
+++ b/bundles/org.eclipse.swt/Eclipse SWT PI/gtk/library/gtk3_stats.c
@@ -113,6 +113,7 @@ char * GTK3_nativeFunctionNames[] = {
 	"gtk_1image_1new_1from_1icon_1name",
 	"gtk_1image_1new_1from_1surface",
 	"gtk_1image_1set_1from_1icon_1name",
+	"gtk_1image_1set_1from_1pixbuf",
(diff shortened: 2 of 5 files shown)


bundles/org.eclipse.swt/Eclipse SWT PI/gtk/library/atk.c
```diff
@@ -24,11 +24,6 @@
 #define ATK_NATIVE(func) Java_org_eclipse_swt_internal_accessibility_gtk_ATK_##func
 #endif
 
-#ifdef _WIN32
```
akurtakov, line 27: Why do you remove these ifdef _WIN32? While I don't say they should be preserved these are clearly noise in this patch.

akurtakov (changes requested):
Please remove the removals of ifdef _WIN32 from this patch.

bundles/org.eclipse.swt/Eclipse SWT PI/gtk/library/atk.c
majanojoel: Agreed, not removed on purpose. They were removed when I built the bindings using the build.sh. I used the latest release but not the latest I build. Is my SWTTools out of date? How can I prevent them from being removed?

akurtakov: Use the latest I-build SWTTools always, that should fix the issue.

majanojoel: Roger, will fix within a day.

majanojoel: Fixed, check latest commit

majanojoel:
Updated to latest I-build SWT Tools and now things look good. Let me know if there's any further changes needed.

akurtakov (approved):

Phillipus:
I'm trying this now with latest commit but get an exception:

```
java.lang.UnsatisfiedLinkError: 'void org.eclipse.swt.internal.gtk3.GTK3.gtk_image_set_from_pixbuf(long, long)'
	at org.eclipse.swt.internal.gtk3.GTK3.gtk_image_set_from_pixbuf(Native Method)
	at org.eclipse.swt.widgets.Button._setImage(Button.java:1167)
	at org.eclipse.swt.widgets.Button.setImage(Button.java:1150)
```

akurtakov:
That happens automatically every night. You have to either wait for that or rebuild natives locally for testing purposes.

akurtakov:
https://github.com/eclipse-platform/eclipse.platform.swt.binaries/commit/dc091c353b59a3d1eefcdc841a18477576298541 can be used for testing

PR #409, opened 2022-09-26
Add zoom level support for program image data

ptief:
Add zoom level support for Program image data.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Program/cocoa/org/eclipse/swt/program/Program.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Program/cocoa/org/eclipse/swt/program/Program.java
@@ -333,14 +333,29 @@ public boolean execute (String fileName) {
 	}
 }
 
+
 /**
- * Returns the receiver's image data.  This is the icon
- * that is associated with the receiver in the operating
- * system.
+ * Returns the receiver's image data at 100% zoom level.  
+ * This is the icon that is associated with the receiver 
+ * in the operating system.
  *
  * @return the image data for the program, may be null
  */
 public ImageData getImageData () {
+  return getImageData (100);
+}
+
+/**
+ * Returns the receiver's image data based on the given zoom level.  
+ * This is the icon that is associated with the receiver in the 
+ * operating system.
+ * 
+ * @param zoom
+ *            The zoom level in % of the standard resolution
+ * 
+ * @return the image data for the program, may be null
+ */
+public ImageData getImageData (int zoom) {
 	NSAutoreleasePool pool = (NSAutoreleasePool) new NSAutoreleasePool().alloc().init();
 	try {
 		NSWorkspace workspace = NSWorkspace.sharedWorkspace();
@@ -358,7 +373,7 @@ public ImageData getImageData () {
 				nsImage.setSize(size);
 				nsImage.retain();
 				Image image = Image.cocoa_new(Display.getCurrent(), SWT.BITMAP, nsImage);
-				ImageData imageData = image.getImageData();
+				ImageData imageData = image.getImageData(zoom);
 				image.dispose();
 				return imageData;
 			}
--- a/bundles/org.eclipse.swt/Eclipse SWT Program/cocoa/org/eclipse/swt/program/Program.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Program/cocoa/org/eclipse/swt/program/Program.java
@@ -354,6 +354,7 @@ public ImageData getImageData () {
  *            The zoom level in % of the standard resolution
  * 
  * @return the image data for the program, may be null
+ * @since 3.125
  */
 public ImageData getImageData (int zoom) {
 	NSAutoreleasePool pool = (NSAutoreleasePool) new NSAutoreleasePool().alloc().init();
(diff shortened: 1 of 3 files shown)


lshanmug:
Changes look good for Windows and Mac.

On Linux, the zoom passed in the new API is not used, so behavior is same as Program.getImageData(). I'm unable to test on Linux, so not sure if the Program.getImageData() already returns ImageData based on the system scaling. If that's the case, I think we can merge the change as Program.getImageData(zoom) is working as expected. Behavior of Program.getImageData() would be different compared to other platforms, but that would be a different issue. @akurtakov Can you please verify the behavior of Program.getImageData on Linux?

lshanmug:
> Changes look good for Windows and Mac.
> 
> On Linux, the zoom passed in the new API is not used, so behavior is same as Program.getImageData(). I'm unable to test on Linux, so not sure if the Program.getImageData() already returns ImageData based on the system scaling. If that's the case, I think we can merge the change as Program.getImageData(zoom) is working as expected. Behavior of Program.getImageData() would be different compared to other platforms, but that would be a different issue. @akurtakov Can you please verify the behavior of Program.getImageData on Linux?

@ptief targeting this for 4.27M1. From the code looks like no changes are required for Linux, but I couldn't confirm it. If required let's track it in a separate issue.

deepika-u:
@ptief @lshanmug 
I have tried pulling both SWT and Platform UI patches on windows and seeing the behavior as below with my observations summarized::

409_issue

Observation 1 ::
On windows - 100%, 125%(recommended), 150% and 175% captured section of images.
1. 125% and 150% icons are having some problem(not center/not complete if we observe keenly).
2. browser icon say mozilla icon and zip file kind of icons are getting some black background irrespective of zoom size.

Observation 2 ::
When i checked on the file changes and went through below link => https://learn.microsoft.com/en-us/windows/win32/api/shellapi/nf-shellapi-extracticonexa
They are providing both phiconLarge and phiconSmall but in the fix proposed now with zoom being passed, you are only using phiconSmall. Any specific reason why you are always using phiconSmall?

iloveeclipse:
Please retarget this for the appropriate milestone in 4.28.

HeikoKlare:
@deepika-u could you please check whether your concerns in https://github.com/eclipse-platform/eclipse.platform.swt/pull/409#issuecomment-1427958143 have been addressed with the recent changes? The icons should now be sized correctly and visible completely, no matter which scale factor.

Regarding your question concerning usage of `phIconLarge` instead of `phIconSmall`: I did some experiments with that but found the "large" icons to even be less sharp than the small ones on 200%, as the "small" ones were already sharp and the larger ones had to be scaled down. My impression was that the Windows API already delivers the "large" image in case the primary monitor has a scale factor of more than 100%. But it may also be that there is still something broken in how the icon is loaded and then rescaled. I would propose to first process this PR (together with the related Platform UI PR) and then investigate the usage of `phIconLarge` again afterwards.

Since recent changes did only affect the Windows implementation, there is no need to re-verify the effects on Mac.

deepika-u:
@HeikoKlare 

Thanks alot for your efforts in pushing this pr in. This is my first pr i started off with SWT ```:)``` so was one of the reasons and another reason being it should not get outdated(effort from @ptief should not go for toss) since we need this fix anyway.

```First of all i am happy with this pr now and my vote goes for this fix 👍 ```, yeah but sorry for my late reply since i tried multiple times across resolutions to be sure of my observations.

Please find my observations at 100%, 125%, 150%, 175%, 200% as below 

Without any pr -> 

without_pr_409


With 409 pr alone, the images are not clear as you said -> 

pr_409_new


With both prs 409 and https://github.com/eclipse-platform/eclipse.platform.ui/pull/295 applied, images are PRETTY clear and complete -> 

pr_409_295_new

So along with this pr expecting https://github.com/eclipse-platform/eclipse.platform.ui/pull/295 to be merged ASAP. Thanks once again.


Least but not last i have a small observation here in the last image when both pr 409 and pr 295 applied case => 
The images from 100%, 125%, 150% of the various icons almost look similar size if you keenly observe.  Where as in 175%, 200% icons sizes increase proportionately. Yeah but may be of less importance as of now i feel.

HeikoKlare:
Thanks a lot for that detailed verification, @deepika-u!

> Least but not last i have a small observation here in the last image when both pr 409 and pr 295 applied case =>
> The images from 100%, 125%, 150% of the various icons almost look similar size if you keenly observe(150% icons look slightly shrunk than 125% but definitely look better after the fix). Where as in 175%, 200% icons sizes increase proportionately. Yeah but may be of less importance as of now i feel.

That's a nice observation. Unfortunately, that is something we cannot change. The reason is the following: When retrieving an icon via the Windows API, it returns that icon scaled according to the primary monitor scaling, so in your example in 100% (16x16 px), 125% (20x20 px), 150% (24x24 px) and so on. SWT is currently configured to only use 100% and 200% images (due to 
`swt.autoScale=integer200`), so it scales the icons down to 16x16 px or up to 32x32 px. The performed rescaling is what produces the results that differ in quality for different scale values.
In my opinion, the "correct" scaling mode is `swt.autoScale=quarter`, because then images have the "correct" size (i.e., when scaling down a screenshot of a 150% UI to 100% and compare it to a native 100% UI, the images have the same size). In that case, icons do not need to be resized as they can be drawn with the scaling in which the Windows API provides them. You can see how that looks like in my screenshots in https://github.com/eclipse-platform/eclipse.platform.swt/pull/409#issuecomment-1908275138.
I am not sure why `swt.autoScale=quarter` isn't the default on Windows (maybe it's because there have been several problems with image scaling, but they should have been addressed recently and with this PR). I appreciate input on that and will also start a discussion about that soon.

merks:
I'm really impressed not only by the work here but also by the collaborative teamwork!

HeikoKlare (approved):
https://github.com/eclipse-platform/eclipse.platform.ui/pull/295 should be merged right afterwards.

PR #428, opened 2022-10-08
Win Display: use java map instead of OS for control map

jukzi:
Calling OS is heavy overhead compared to jvm internal map.get().

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
@@ -137,13 +137,8 @@ public class Display extends Device {
 	boolean externalEventLoop; // events are dispatched outside SWT, e.g. TrackPopupMenu or DoDragDrop
 
 	/* Widget Table */
-	int freeSlot;
-	int [] indexTable;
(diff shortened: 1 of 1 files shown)


jukzi:
Sampled eclipse start main thread:
image

Does not make sense to inform/query the OS about internal index for each control.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
```diff
@@ -138,12 +139,8 @@ public class Display extends Device {
 
 	/* Widget Table */
 	int freeSlot;
```
iloveeclipse, line 141: - freeSlot should be unused now
- SWT is single threaded, why do we need ConcurrentHashMap here?
- I wonder if the handle / control is really 1:1 relationship. At least in the table it was possible to map 1:N. Have you tried to run all SWT tests **locally** on Windows with the new code?

jukzi: I ran test locally and eclipse ide still starts normally.
It would still be possible to add multiple handles per control.  The getControl() signature however clearly renders multiple controls per handle as not supported.

Issue #420, opened 2022-10-02
[High DPI] "swt.autoScale" should add new "half" option


Dani-Hub:
**Is your feature request related to a problem? Please describe.**
High-DPI support of applications usually has to take a middle-ground between user-flexibility ("Support of multiple zoom options") and application economy ("Try to minimize the number of supported zoom options"). Eclipse jface provides via the "@NNx" file name heuristics a nice auto detection for high resolution image files targeting 150% and 200% zoom option. This is more flexible than the "integer200" default, but among the currently supported "swt.autoScale" options there is currently none that targets this sweet spot. Hard-coding 150% is rarely an acceptable option for real-world applications and the "quarter" option may be too fine-grained for applications that just wish to support 150% and 200% zoom.

**Describe the solution you'd like**
I suggest to add a new "swt.autoScale" option (tentatively suggested name would be "half") that is similar to the existing "quarter" option, but folds the quarter cases to the next 50% ("halves"). The effective formula (as of [DPIUtil.java#L454](https://github.com/eclipse-platform/eclipse.platform.swt/blob/ffbfd4e2461e63ebc3cbc2351a19b1ca7ac4d317/bundles/org.eclipse.swt/Eclipse%20SWT/common/org/eclipse/swt/internal/DPIUtil.java#L454)) would be
```
zoom = Math.round(nativeDeviceZoom / 50f) * 50;
```

**Describe alternatives you've considered**
In our team we discussed the existing options and agreed that none of them can be considered as a suitable alternative.

**Additional context**
None.

Dani-Hub:
@sravanlakkimsetti, I would like to suggest that we can close this issue now with the acceptance of #429.

Issue #294, opened 2022-08-08
org.eclipse.swt.tests.junit.Test_org_eclipse_swt_custom_StyledText.test_caretSizeAndPositionVariableGlyphMetrics fails on master


laeubi:
see https://ci.eclipse.org/platform/job/eclipse.platform.swt/job/master/571/testReport/junit/org.eclipse.swt.tests.junit/Test_org_eclipse_swt_custom_StyledText/test_caretSizeAndPositionVariableGlyphMetrics/

jukzi:
Please revert whatever caused that or fix the test.

akurtakov:
This failure is due to change in Pango library which changed in the OS. Reverting to older docker image is not an option for security reasons. I have spent some time trying to fix this one but still can't figure it. If there is anyone knowledgable of pango details help is more than welcome.

iloveeclipse:
As discussed on https://github.com/eclipse-platform/eclipse.platform.swt/pull/504, if we don't have solution for 4.27 M3 we will "mute" the test via https://github.com/eclipse-platform/eclipse.platform.swt/pull/504.

akurtakov:
If only there was XFAIL concept in JUnit so these can stay easily findable while not breaking the build.

HannesWell:
> This failure is due to change in Pango library which changed in the OS. Reverting to older docker image is not an option for security reasons. I have spent some time trying to fix this one but still can't figure it. If there is anyone knowledgable of pango details help is more than welcome.

In #514 I noticed that all tests pass, as mentioned in https://github.com/eclipse-platform/eclipse.platform.swt/pull/504#issuecomment-1399351991, this is maybe because with that change the tests run on the standard centos-latest/8 image?

akurtakov:
Running the build on older pango version will make the tests work but they would still fail on newer one.

HannesWell:
> Running the build on older pango version will make the tests work but they would still fail on newer one.

So the centos-latest images use the old pango version and the tests will fail again, if those images use newer versions?
I have to admit, that until now I never heard of pango.^^

akurtakov:
> > Running the build on older pango version will make the tests work but they would still fail on newer one.
> 
> So the centos-latest images use the old pango version and the tests will fail again, if those images use newer versions? I have to admit, that until now I never heard of pango.^^

[...]

PR #429, opened 2022-10-09
Issue #420: "swt.autoScale" should add new "half" option

Dani-Hub:
- Add new "half" option for the "swt.autoScale" system property and evaluate in DPIUtil#getZoomForAutoscaleProperty
- Remove unneeded casts to int (Math.round with float argument already returns int)

Signed-off-by: Daniel Krügler <daniel.kruegler@gmail.com>

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
@@ -55,9 +55,9 @@ public class DPIUtil {
 	 *     generally rounded down (e.g. at 150%, will use 100%), unless close to
 	 *     the next integer multiple (currently at 175%, will use 200%).</li>
 	 * <li><b>integer200</b>: like <b>integer</b>, but the maximal zoom level is 200%.</li>
-	 * <li><b>half</b>: deviceZoom depends on the current display resolution,
+	 * <li><b>half-even</b>: deviceZoom depends on the current display resolution,
 	 *     but only uses integer multiples of 50%. The detected native zoom is
-	 *     rounded to the closest permissible value.</li>
+	 *     rounded to the closest permissible value, with tie-breaker towards even.</li>
 	 * <li><b>quarter</b>: deviceZoom depends on the current display resolution,
 	 *     but only uses integer multiples of 25%. The detected native zoom is
 	 *     rounded to the closest permissible value.</li>
@@ -460,7 +460,7 @@ public static int getZoomForAutoscaleProperty (int nativeDeviceZoom) {
 	if (autoScaleValue != null) {
 		if ("false".equalsIgnoreCase (autoScaleValue)) {
 			zoom = 100;
-		} else if ("half".equalsIgnoreCase (autoScaleValue)) {
+		} else if ("half-even".equalsIgnoreCase (autoScaleValue)) {
 			// Math.round rounds 125->150 and 175->200, 
 			// Math.rint rounds 125->100 and 175->200 matching
 			// "integer200"
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
@@ -55,7 +55,7 @@ public class DPIUtil {
 	 *     generally rounded down (e.g. at 150%, will use 100%), unless close to
 	 *     the next integer multiple (currently at 175%, will use 200%).</li>
 	 * <li><b>integer200</b>: like <b>integer</b>, but the maximal zoom level is 200%.</li>
-	 * <li><b>half-even</b>: deviceZoom depends on the current display resolution,
+	 * <li><b>half</b>: deviceZoom depends on the current display resolution,
 	 *     but only uses integer multiples of 50%. The detected native zoom is
 	 *     rounded to the closest permissible value, with tie-breaker towards even.</li>
 	 * <li><b>quarter</b>: deviceZoom depends on the current display resolution,
@@ -460,7 +460,7 @@ public static int getZoomForAutoscaleProperty (int nativeDeviceZoom) {
 	if (autoScaleValue != null) {
 		if ("false".equalsIgnoreCase (autoScaleValue)) {
 			zoom = 100;
-		} else if ("half-even".equalsIgnoreCase (autoScaleValue)) {
+		} else if ("half".equalsIgnoreCase (autoScaleValue)) {
 			// Math.round rounds 125->150 and 175->200, 
 			// Math.rint rounds 125->100 and 175->200 matching
 			// "integer200"
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
@@ -10,6 +10,7 @@
  *
  * Contributors:
  *     IBM Corporation - initial API and implementation
+ *     Daniel Krügler - #420 - [High DPI] "swt.autoScale" should add new "half" option
  *******************************************************************************/
 package org.eclipse.swt.internal;
 
@@ -54,6 +55,9 @@ public class DPIUtil {
 	 *     generally rounded down (e.g. at 150%, will use 100%), unless close to
 	 *     the next integer multiple (currently at 175%, will use 200%).</li>
 	 * <li><b>integer200</b>: like <b>integer</b>, but the maximal zoom level is 200%.</li>
+	 * <li><b>half</b>: deviceZoom depends on the current display resolution,
+	 *     but only uses integer multiples of 50%. The detected native zoom is
+	 *     rounded to the closest permissible value.</li>
 	 * <li><b>quarter</b>: deviceZoom depends on the current display resolution,
 	 *     but only uses integer multiples of 25%. The detected native zoom is
 	 *     rounded to the closest permissible value.</li>
@@ -456,8 +460,10 @@ public static int getZoomForAutoscaleProperty (int nativeDeviceZoom) {
 	if (autoScaleValue != null) {
 		if ("false".equalsIgnoreCase (autoScaleValue)) {
 			zoom = 100;
+		} else if ("half".equalsIgnoreCase (autoScaleValue)) {
+			zoom = Math.round (nativeDeviceZoom / 50f) * 50;
 		} else if ("quarter".equalsIgnoreCase (autoScaleValue)) {
-			zoom = (int) (Math.round (nativeDeviceZoom / 25f) * 25);
+			zoom = Math.round (nativeDeviceZoom / 25f) * 25;
 		} else if ("exact".equalsIgnoreCase (autoScaleValue)) {
 			zoom = nativeDeviceZoom;
 		} else {
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
@@ -461,9 +461,10 @@ public static int getZoomForAutoscaleProperty (int nativeDeviceZoom) {
 		if ("false".equalsIgnoreCase (autoScaleValue)) {
 			zoom = 100;
 		} else if ("half".equalsIgnoreCase (autoScaleValue)) {
-			// Math.round would round 125->150 and 175->200, 
-			// using truncation round 125->100 and 175->150
-			zoom = (nativeDeviceZoom / 50) * 50;
+			// Math.round rounds 125->150 and 175->200, 
+			// Math.rint rounds 125->100 and 175->200 matching
+			// "integer200"
+			zoom = (int) Math.rint(nativeDeviceZoom / 50d) * 50;
 		} else if ("quarter".equalsIgnoreCase (autoScaleValue)) {
 			zoom = Math.round(nativeDeviceZoom / 25f) * 25;
 		} else if ("exact".equalsIgnoreCase (autoScaleValue)) {
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
@@ -461,9 +461,11 @@ public static int getZoomForAutoscaleProperty (int nativeDeviceZoom) {
 		if ("false".equalsIgnoreCase (autoScaleValue)) {
 			zoom = 100;
 		} else if ("half".equalsIgnoreCase (autoScaleValue)) {
-			zoom = Math.round (nativeDeviceZoom / 50f) * 50;
+			// Math.round would round 125->150 and 175->200, 
+			// using truncation round 125->100 and 175->150
+			zoom = (nativeDeviceZoom / 50) * 50;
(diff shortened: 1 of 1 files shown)


Dani-Hub:
Could please someone provide some guidance? I looked into the reported test failure, but to me this seems completely unrelated to my PULL request. Thank you.

Dani-Hub:
> Could please someone provide some guidance? I looked into the reported test failure, but to me this seems completely unrelated to my PULL request. Thank you.

Ping, this is a friendly reminder asking for support.

SyntevoAlex:
Which test failure are you talking about? As far as I can see, tests wasn't run before you're a first time contributor?

Dani-Hub:
> Which test failure are you talking about? As far as I can see, tests weren't run before you're a first time contributor?

Sorry for being less informative than I could have been. The test failure was (and still is) [here](https://ci.eclipse.org/platform/job/eclipse.platform.swt/job/PR-429/2/testReport/junit/org.eclipse.swt.tests.junit/Test_org_eclipse_swt_custom_StyledText/test_caretSizeAndPositionVariableGlyphMetrics/). I'm very confident that it is unrelated to my PULL request, but I'm happy to be taught that I was wrong.

SyntevoAlex:
Yes, `test_caretSizeAndPositionVariableGlyphMetrics` test is broken, see #294. Indeed it's not related to your PR.

SyntevoAlex:
Can you give a more detailed explanation of how does it help to limit to multiples of 50% ?
If user actually runs 125%, and SWT would use 100%, doesn't it result in blurry painting?

Dani-Hub:
> Can you give a more detailed explanation of how does it help to limit to multiples of 50% ? If user actually runs 125%, and SWT would use 100%, doesn't it result in blurry painting?

I tried to explain this in issue #420: The suggestion to add a new half option is considered to provide a middle ground between user-flexibility and implementation economy. The current default "integer200" is a very rough candidate and would not help in either 150% nor 125%. The advantage of supporting 50% steps is that jface directly supports both 150% and 200% via a "@NnX" file name heuristics, so when the user accepts the (s)he cannot get direct support for 125% but for 150%, that seems to be an improvement. The disadvantage for being currently forced to use the "quarter" option is that 125% nor 175% are currently not well supported directly through the "@NnX" file name heuristics. While I could open a request for jface to support these as well, this is not what I wanted to to, because of the economic argument, because in this case an application would need to support about four variants of an icon instead of two.

SyntevoAlex:
Sorry but I have no idea what the `@NnX` file name heuristic is.

Trying to read between the lines, it seems to me that the problem is about finding a most matching image to paint, where jface can automatically find a variant of image based on its filename or something.

Also, I don't understand that nature of the problem with this heuristic and jface. What exactly does it mean that "not well supported directly through" ?

It is not clear to me what happens if jface fails to find an image with desired DPI? I would assume that it should grab some other image and stretch it. That would result in some blurry images sometimes, but how does it help if you suggest that it loads 150% image?

Why don't you use `ImageDataProvider` ? Its goal is exactly to let you choose the most matching image.

SyntevoAlex:
When you force `DPIUtil` to 150% where user runs at 175%, doesn't it cause problems in various places not related to Images? Such as font sizes, Control sizes, etc etc?

Dani-Hub:
> When you force `DPIUtil` to 150% where user runs at 175%, doesn't it cause problems in various places not related to Images? Such as font sizes, Control sizes, etc etc?

I would like to remind that the current default is "integer200", where this argument is even worse. But even with that option the situation is less worse as you may assume, because the "effective" SWT zoom level is correctly used throughout the Eclipse code and a consistent behaviour is observed by the user. What we would like to realize is that we get better than  "integer200" by supporting the 50% level support, which means that we are willing to add two additional images per image to our application to provide optimal user experience for this granularity. Currently there is no option in "swt.autoScale" that allows me to get these effects. The next possible choice would be 25% granularity, which means I'm required to provide four additional images per image. We consider this choice as too expensive.

SyntevoAlex:
You're still not quite answering my question.

What exactly does user see when you use `"quarter"` option at 125%?
What exactly does user see when you use `"half"` option at 125%?
Can we see some screenshots?

Dani-Hub:
> Sorry but I have no idea what the `@NnX` file name heuristic is.
> 

Jface provides an auto-detection mechanism to select the optimal icon for a given zoom level based on a naming convention. The most prominent one is that if a file name ends with "@2x" or with "@1.5x" it will select it when the SWT device zoom level is 200% or 150%, respectively. This is very helpful for larger RCP clients which are build on-top of jface and not being restricted to SWT.

> Trying to read between the lines, it seems to me that the problem is about finding a most matching image to paint, where jface can automatically find a variant of image based on its filename or something.
> 

That is the correct idea, but not the exact point. Of-course there exists the `ImageDataProvider` option and we could use it at places where we have full control over the image loading step (We use that interface in other programs that are solely SWT-based). But if you are developing an RCP client based on richer higher-level API you usually have to provide a jface `ImageDescriptor` or an image URL (that somewhat latter is converted to an ImageDescriptor) and the actual image loading is out of my hands. I do depend on the ImageDescriptor logic uses the (a) the swt device zoom level  and (b)to determine

> Also, I don't understand that nature of the problem with this heuristic and jface. What exactly does it mean that "not well supported directly through" ?
> 

I'm talking about the direct support for 150% and 200% image detection see above. I _could_ request to similarly support 125% and 175% by the same mechanism but that is intentionally not done due to the economic argument, see above.

> It is not clear to me what happens if jface fails to find an image with desired DPI? I would assume that it should grab some other image and stretch it. That would result in some blurry images sometimes, but how does it help if you suggest that it loads 150% image?
> 

Your interpretation is correct. It would stretch it, but if I have a matching filename that fits the current zoom level, it loads it *instead* of stretching it.

> Why don't you use `ImageDataProvider` ? Its goal is exactly to let you choose the most matching image.

See above.

SyntevoAlex:
> Your interpretation is correct. It would stretch it, but if I have a matching filename that fits the current zoom level, it loads it instead of stretching it.

If user runs at 125%, and you only have 1x and 2x images, then there is no matching file to load.
Then, you could use `"quarter"` option, which would resulting in stretching 1x image to 1.25x.
Or you could use `"half"` option which... And here's where I don't understand, how does it help?

merks:
> > Your interpretation is correct. It would stretch it, but if I have a matching filename that fits the current zoom level, it loads it instead of stretching it.
> 
> If user runs at 125%, and you only have 1x and 2x images, then there is no matching file to load. Then, you could use `"quarter"` option, which would result in stretching something, probably 1x image to 1.25x. Or you could use `"half"` option which... And here's where I don't understand, how does it help?

As I understand it, the real native zoom might be 100, 125, 150, 175, 200, 225, 250, and so on, but with this additional option, the computed value will be 100, 150, 200, 250, and so on, and then someone building an RCP product could use this option for their product and provide "perfect"  images for those resolutions just as Eclipse itself does for 100% and 200%, relying on rounding to the nearest 100% which is the default:

image

Note that you can try out how each scale factor looks by passing -Dswt.autoScale=125 (or whatever value) either in the eclipse.ini or in the VM args of a launch config.

So the only point, again as I understand it, is that there as a option to round to the nearest 50% just as there is already an option to round to the nearest 25% such that one can reduce the number of provided images and force the rounding value to match the sizes of the provided "perfect" images.  No one is forced to use this so it's just a nice-to-have if someone wants that for their product.  it sounds 100% reasonable to me...

Dani-Hub:
> As I understand it, the real native zoom might be 100, 125, 150, 175, 200, 225, 250, and so on, but with this additional option, the computed value will be 100, 150, 200, 250, and so on, and then someone building an RCP product could use this option for their product and provide "perfect" images for those resolutions just as Eclipse itself does for 100% and 200%, relying on rounding to the nearest 100% which is the default:
> 
> image
> 

Yes, this is exactly the point I'm trying to make. I'm sorry that I failed to do so.

> So the only point, again as I understand it, is that there as a option to round to the nearest 50% just as there is already an option to round to the nearest 25% such that one can reduce the number of provided images and force the rounding value to match the sizes of the provided "perfect" images. No one is forced to use this so it's just a nice-to-have if someone wants that for their product. it sounds 100% reasonable to me...

Thank you for the support. I will provide a separate comment that shows a table of the effects, because due to this very helpful conversation I actually found a potential issue with my proposed resolution (that can be easily resolved, though).

Dani-Hub:
> You're still not quite answering my question.

I'm sorry that you got this impression. I can ensure you that I'm trying my best to make my suggestion clear.

> What exactly does user see when you use `"quarter"` option at 125%? 

I'm replying to this and the follow-up question under the assumption that I have prepared my RCP client images to add image variants @1.5x and @2x. Under 125% none of these images will be considered (which makes sense, because they don't match this zoom value), so jface will stretch the provided 100% image using the [DPIUtil#autoScaleImageData](https://github.com/eclipse-platform/eclipse.platform.swt/blob/ffbfd4e2461e63ebc3cbc2351a19b1ca7ac4d317/bundles/org.eclipse.swt/Eclipse%20SWT/common/org/eclipse/swt/internal/DPIUtil.java#L221), which makes the image look like a low-quality image.

>What exactly does user see when you use `"half"` option at 125%? 

Under my originally suggested formula (But see my follow-up comment and refined PULL request), this will compute an SWT effective device zoom of 150 and this has the effect that instead of scaling my 100% image the @1.5x variant will be selected, leading to a high-quality image impression for the user.

> Can we see some screenshots?

I'm not able to produce them for the place I'm sitting at. @merks has tried to describe the effects in better words than I did. Does this help?

Dani-Hub:
This comment intends to show the computed effective SWT device zoom value depending on the current "swt.autoScale" value:

|swt.autoScale| 100 | 125 | 150 | 175 | 200
|---|---|---|---|---|---|
|"integer200"|100|100|100|200|200|
|"quarter"|100|125|150|175|200|
|"half"|100|150|150|200|200|

So for the "half" option we need only to provide two extra high-DPI images per icon to get optimum quality for the user assuming that we provide @1.5x and @2x variants for each.

The interesting question is (And I would like to thank all participants of this discussion to their helpful contributions that helped me looking with an critical eye on the proposed solution): Do we agree that the effective rounding-up results here as part of my suggested PULL request? (Note that I'm using the exactly same approach as we already have for the "quarter" solution, but the effect of `Math#round` is to "tie rounding to positive infinity", which has the effect that the 125% case is rounded to 150 (Not 100) and the 175% case is rounded to 200 (Not 150). Are we happy with this or would we prefer the following effects instead (emphasis mine):

|swt.autoScale| 100 | 125 | 150 | 175 | 200
|---|---|---|---|---|---|
|"integer200"|100|100|100|200|200|
|"quarter"|100|125|150|175|200|
|"half"|100|*100*|150|*150*|200|

Thanks for your opinion.

merks:
My sense is that just as 150 is rounded down to 100 for `integer200` probably 125 and 175 should be rounded down for `half`.

Dani-Hub:
> My sense is that just as 150 is rounded down to 100 for `integer200` probably 125 and 175 should be rounded down for `half`.

Thanks for the feedback - PULL request adjusted.

Dani-Hub:
My hopefully last rounding refinement attempt. By means of using `Math.rint` we have the nice effect that for 125% and 175% we get the same rounding behaviour as the default "integer200" (one of them is rounded down and the other up due to half-even rule). The effects are as follows:

|swt.autoScale| 100 | 125 | 150 | 175 | 200
|---|---|---|---|---|---|
|"integer200"|100|100|100|200|200|
|"quarter"|100|125|150|175|200|
|"half"|100|*100*|150|*200*|200|

Does this suggestion gets more consensus as the two preceding ones?

sravanlakkimsetti:
With `swt.autoScale=quarter`, apart from blurry images, we have quite a bit widget positioning problems and crashes. Using `half` could provide a good middle path. 

As far as remember for `integer200` we use 100% image till 174. At 175 it gets rounded up to 200. I feel latest implementation is going is the right direction. 

Also as far I remember `swt.autoScale` will work only with `windows` implementation. `Mac` and `linux` will not have any impact with this one. First hi-dpi implementation of linux used swt.autoScale param but the second implementation where we were using cairo scaling does not use swt.autoScale param. 

I would suggest focusing on fixing the positioning problems and crashes arising from rounding errors when we are using fractional scaling.

Dani-Hub:
Since my different rounding forms may have confused reviewers, I have now given the most recently suggested "half" form the name "half-even", since it technically corresponds to the `RoundingMode#HALF_EVEN` characteristics. Naming the newly suggested `swt.autoScale` form this way may help people understanding the effects and would potentially allow in the future to add other "half" forms as well. Please get me right: I'm _not_ proposing different "half" forms as part of this pull request. For the sake of the discussion here, my original rounding form could be termed "half-up" (Since we rounded up) and my second suggestion could be termed "half-down" (Since we used truncation). I'm open to any of these half-rounding forms, but would tentatively prefer "half-even", because it has statistically the least rounding bias.

Dani-Hub:
> I prefer calling this one as `half` rather than `half-even`. 

Sure, no problem. I will revert this part of my PULL request this evening.

I would prefer enabling `half` and keep working on the positioning of widgets and crashes that occur due to rounding errors. Once we are fine with the problems fixed, change
> 
> https://github.com/eclipse-platform/eclipse.platform.swt/blob/6c09b679b5d0e72461287b56c5dbb16dd5101562/bundles/org.eclipse.swt/Eclipse%20SWT/common/org/eclipse/swt/internal/DPIUtil.java#L473
> to support `half`.

I'm sorry, I don't understand this part of the suggestion. In which sense does this default branch has any relation to the suggested "half" proposal? Could you please elaborate?

sravanlakkimsetti:
> I'm sorry, I don't understand this part of the suggestion. In which sense does this default branch has any relation to the suggested "half" proposal? Could you please elaborate?

Let me explain the origin of `swt.autoScale` parameter. 
We were actually tried to support fractional scaling when we started the Hi-Dpi implementation. In hidpi implementation we actually manipulate the widget sizes based on the dpi used(now this implementation is used in Windows only). 

Because of this widget size manipulation we ended up having wrong sizes. Sometimes a widget position can be off by a pixel or accessing uninitialized area. In case of images the dimensions and array sizes used to hold the image do not reconcile. This caused a lot of crashes. 

To avoid this by default(for most customers) we added support of integer scale factors. This implementation is there in https://github.com/eclipse-platform/eclipse.platform.swt/blob/6c09b679b5d0e72461287b56c5dbb16dd5101562/bundles/org.eclipse.swt/Eclipse%20SWT/common/org/eclipse/swt/internal/DPIUtil.java#L473

We added `swt.autoScale` as a way to test fractional scalefactors for the community. Since Windows uses the scalefactors in multiples of 25% we selected `quarter`. 

My suggestion here is add `half` as one of the option but look out for the widget positioning problems and crashes. `half` option will have less number of problems than `quarter` if we are satisfied with testing we should make eclipse support half by default without any extra parameter by modifying https://github.com/eclipse-platform/eclipse.platform.swt/blob/6c09b679b5d0e72461287b56c5dbb16dd5101562/bundles/org.eclipse.swt/Eclipse%20SWT/common/org/eclipse/swt/internal/DPIUtil.java#L473

PS: Linux and Mac implementations delegate calculation of widget sizes and positions to the OS. So those problems are visible in those Operating Systems.

Dani-Hub:
> Let me explain the origin of `swt.autoScale` parameter. We were actually tried to support fractional scaling when we started the Hi-Dpi implementation. In hidpi implementation we actually manipulate the widget sizes based on the dpi used(now this implementation is used in Windows only).
> [...]

Thanks for the complete picture, that was indeed helpful.

Dani-Hub:
Based on feedback from @sravanlakkimsetti I reverted to the "half" name suggestion.

SyntevoAlex:
In this specific PR, it would have been wise to squash commits before merging them. In other PRs however, commits should be merged as is. This is a decision that needs to be made per PR based on the quality of splitting work into fine-grained commits.

PR #430, opened 2022-10-09
Win Tree: cache column order

jukzi:
Since order is frequently needed but rarely changed. Especially the index of the first column is frequently used for each TreeItem.
On Win even such simple systemcalls are more then 50 times slower then jvm internal cache.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
@@ -104,6 +104,7 @@ public class Tree extends Composite {
 	int headerForeground = -1;
 
 	// Cached variables for fast item lookup
+	int[] cachedItemOrder;
 	long cachedFirstItem;   // Used to figure when other cache variables need updating
 	long cachedIndexItem;   // Item for which #cachedIndex is saved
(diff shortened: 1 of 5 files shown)


jukzi:
@SyntevoAlex there is a lot of caching in Tree. Is there any reason why we ask the OS about the items at all? If only SWT creates, removes and reorders items can SWT just maintain a list of all items instead of (errorprone) cache and sometimes ask the OS to find an item? Can items be created/deleted from outside SWT?

SyntevoAlex:
Do you handle the case of drag&drop reordering of columns?

SyntevoAlex:
I didn't implement that caching, just struggled with it.

I also thought about maybe having everything inside Tree, but then I thought about more complicated cases (such as moving an item with children around) and gave up. I needed to just get it done and switch back to another task.

jukzi:
> Do you handle the case of drag&drop reordering of columns?

not explicitly. this impl relies on setColumnOrder(). i am not used to DND. is that a OS functionality or implemented in SWT? is there an programatic example?

SyntevoAlex:
See `TreeColumn.setMoveable()`. When enabled, you just click on a column and drag it to a new position.

jukzi:
> See `TreeColumn.setMoveable()`. When enabled, you just click on a column's header and drag it to a new position.

Snippet055HideShowColumn creates such a table and seems to work fine: when start dragging cache is fixed and later new values queried from OS

PR #437, opened 2022-10-18
Add canPaste method to StyledText

mihnita:
Handy to have, to make some code simpler.
See for example https://github.com/eclipse-pde/eclipse.pde/pull/363, where 14 classes have to do the same thing.
And they have to know (and keep up with) what formats `StyledText` supports.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/StyledText.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/StyledText.java
@@ -7165,6 +7165,34 @@ void paintObject(GC gc, int x, int y, int ascent, int descent, StyleRange style,
 	}
 }
 /**
+ * Detects if any of the content types in the clipboard are in a format
+ * that can be pasted into a {@code StyledText}.
+ *
+ * <p>The formats currently supported are plain text, HTML, and RTF.</p>
+ *
+ * @param clipboard the clipboard to check for supported types
+ * @return true if any of he types in clipboard can be pasted into a {@code StyledText}.
+ *
+ * @since 3.121
+ */
+public static boolean canPaste(Clipboard clipboard) {
+	TransferData[] types = clipboard.getAvailableTypes();
+	Transfer[] transfers = new Transfer[] {
+			TextTransfer.getInstance(),
+			HTMLTransfer.getInstance(),
+			RTFTransfer.getInstance()
+	};
+	for (TransferData type : types) {
+		for (Transfer transfer : transfers) {
+			if (transfer.isSupportedType(type)) {
+				System.out.printf("%s.isSupportedType(%s)", transfer, type);
+				return true;
+			}
+		}
+	}
+	return false;
+}
+/**
  * Replaces the selection with the text on the <code>DND.CLIPBOARD</code>
  * clipboard  or, if there is no selection,  inserts the text at the current
  * caret offset.   If the widget has the SWT.SINGLE style and the
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/StyledText.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/StyledText.java
@@ -7165,29 +7165,18 @@ void paintObject(GC gc, int x, int y, int ascent, int descent, StyleRange style,
 	}
 }
 /**
- * Detects if any of the content types in the clipboard are in a format
- * that can be pasted into a {@code StyledText}.
+ * Detects if the provided Clipboard content can be pasted into this widget.
  *
- * <p>The formats currently supported are plain text, HTML, and RTF.</p>
- *
- * @param clipboard the clipboard to check for paste
- * @return true if any of he types in clipboard can be pasted into a {@code StyledText}.
+ * @param clipboard the clipboard to check for paste support.
+ * @return true if any of the content in clipboard can be pasted.
  *
  * @since 3.121
  */
 public static boolean canPaste(Clipboard clipboard) {
-	TransferData[] types = clipboard.getAvailableTypes();
-	Transfer[] transfers = new Transfer[] {
-			TextTransfer.getInstance(),
-			HTMLTransfer.getInstance(),
-			RTFTransfer.getInstance()
-	};
-	for (TransferData type : types) {
-		for (Transfer transfer : transfers) {
-			if (transfer.isSupportedType(type)) {
-				System.out.printf("%s.isSupportedType(%s)", transfer, type);
-				return true;
-			}
+	Transfer textTransfer = TextTransfer.getInstance();
+	for (TransferData type : clipboard.getAvailableTypes()) {
+		if (textTransfer.isSupportedType(type)) {
+			return true;
 		}
 	}
 	return false;
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/StyledText.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/StyledText.java
@@ -7170,7 +7170,7 @@ void paintObject(GC gc, int x, int y, int ascent, int descent, StyleRange style,
  *
  * <p>The formats currently supported are plain text, HTML, and RTF.</p>
  *
- * @param clipboard the clipboard to check for supported types
+ * @param clipboard the clipboard to check for paste
  * @return true if any of he types in clipboard can be pasted into a {@code StyledText}.
  *
  * @since 3.121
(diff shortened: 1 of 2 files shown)


bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/StyledText.java
```diff
@@ -7164,6 +7164,34 @@ void paintObject(GC gc, int x, int y, int ascent, int descent, StyleRange style,
 		notifyListeners(ST.PaintObject, event);
 	}
 }
+/**
+ * Detects if any of the content types in the clipboard are in a format
+ * that can be pasted into a {@code StyledText}.
+ *
+ * <p>The formats currently supported are plain text, HTML, and RTF.</p>
+ *
+ * @param clipboard the clipboard to check for supported types
```
mickaelistria, line 7173:
```suggestion
 * @param clipboard the clipboard to check for paste
```

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/StyledText.java
```diff
@@ -7164,6 +7164,34 @@ void paintObject(GC gc, int x, int y, int ascent, int descent, StyleRange style,
...
+	TransferData[] types = clipboard.getAvailableTypes();
+	Transfer[] transfers = new Transfer[] {
+			TextTransfer.getInstance(),
+			HTMLTransfer.getInstance(),
+			RTFTransfer.getInstance()
+	};
+	for (TransferData type : types) {
+		for (Transfer transfer : transfers) {
+			if (transfer.isSupportedType(type)) {
+				System.out.printf("%s.isSupportedType(%s)", transfer, type);
```
mickaelistria, line 7188: Please remove

mickaelistria:
I don't think this patch is correct. The getClipboardContent seems only capable of pasting plain TextTransfer, so this canPaste method would be incorrect.
I believe the issue is more that the clipboard is missing a TextTransfer when copying HTML.

mickaelistria (changes requested):

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/StyledText.java
mihnita: Done.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/StyledText.java
```diff
@@ -7164,6 +7164,34 @@ void paintObject(GC gc, int x, int y, int ascent, int descent, StyleRange style,
 		notifyListeners(ST.PaintObject, event);
 	}
 }
+/**
+ * Detects if any of the content types in the clipboard are in a format
```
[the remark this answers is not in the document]
mihnita: Done.

mickaelistria:
But then, I don't think it's necessary to add this method at the moment as it wouldn't fix the issue in editors: canPaste() in editors would still return true if there is a TextTransfer in clipboard, but the issue seems them that there is no TextTransfer in clipboard, and this is not really an issue from StyledText, is it?

mickaelistria:
But why do other editors claim they can support pasting RTF (and HTML with your change) while StyledText actually cannot? Isn't it just a bug that those editor do claim the can paste more than a TextTransfer? What happens when pasting a RTFTransfer?
What are the application that can feed the clipboard with a RTFTransfer but not with a HTMLTransfer? How to reproduce this issue in practice?

mihnita:
> But why do other editors claim they can support pasting RTF (and HTML with your change) while StyledText actually cannot? 
> Isn't it just a bug that those editor do claim the can paste more than a TextTransfer? What happens when pasting a RTFTransfer?

Because they really can.
To paste a format means you that the application (or widget) would read the content from the clipboard, parse it, and "interpret it" to it's own native format.

So for `StyledText` to paste HTML (or RTF) would need to do the opposite of what it does in the `HTMLWriter` / `RTFWriter`
In the `*Writer` we iterate on lines, then iterate on `StyleRange`s and produce html/rtf "by hand"

For paste we would need to some classes (Reader?) doing the opposite operations: parse the HTML / RTF, add the text to the document and `StyleRanges`.

But I am not sure how that would really work, since `StyledText` seems to use two models that can't coexist.
One is to contain the formatting info in the document itself, and another generates the styles "on demand" (in a listener).

So if the current widget goes the listener route, I can't think of a clean way to set styles from clipboard.
And if I paste in a Java editor I don't want the bold and colors from MS Word anyway.
So maybe paste would only make sense for the non-listener kind of widget.

---

When you say "other editors" you think something like MS Word?
MS Word does indeed understand both RTF and HTML. Parses them and converts to the format to the internal representation.

But Chrome (and GMail, and many web based apps) don't understand RTF, only plain text and HTML.
There was a time when MS Word did not support HTML (and I'm old enough to remember that time :-)
And the Windows native [RichEdit control](https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.richtextbox) still does not support HTML.

So the idea is to put in the clipboard as many formats as you can, hoping that the other applications can parse (and paste) at lease one of your formats. And they are smart enough to choose the richest format.

That is also why all apps I know that deal with text put in the clipboard some plain text format, and one or more forms of styled format. Because plain text is easy to produce / consume, and it is widely understood.

---

> What are the application that can feed the clipboard with a RTFTransfer but not with a HTMLTransfer?
> How to reproduce this issue in practice?

The `StyledText` widget before my changes.
The Windows RichEdit control (MS calls the Windows widgets "controls").

And WordPad (the Windows app) is a very thin wrapper around RichEdit control.
It can easily copy / paste plain text and RTF, but not HTML.

To reproduce you can start WordPad, write something, format it, and copy.
The clipboard will contain plain text, RTF, and a few proprietary formats (some kind of OLE stuff ("Wordpad Document Object"), and an "Enhanced Metafile").

Or take Eclipse 2022-09 (or below) and copy some Java code.

Then got to MS Word and try "Paste Special" (Alt+E for "Edit", Alt+S for "Special", or the "Home tab", "Paste" button has an arrow down under it, and when you click it the options are "Paste" and "Paste Special"

MS Word Paste Special from Eclipse 2022-09:
image

MS Word Paste Special from Eclipse 2022-12 nightly (I20221013-1800):
image

mickaelistria:
Ok, so if the editors can understand HTML/RTF but StyledText cannot, and they really override the paste behavior then my initial idea was wrong and it doesn't make sense to try factorizing it in StyledText at the moment; and editors need to be updated as you initially did.

mihnita:
Nitpick:
> ... can paste more than a TextTransfer? What happens when pasting a RTFTransfer?

Applications don't paste TextTransfer / RTFTransfer, those are SWT concepts.
Each platform has its own thing.

I am more familiar with the Windows one.
https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setclipboarddata
So there are [Standard Clipboard Formats](https://learn.microsoft.com/en-us/windows/desktop/dataxchg/standard-clipboard-formats) and [Registered Clipboard Formats](https://learn.microsoft.com/en-us/windows/desktop/dataxchg/clipboard-formats).

RTF and HTML are registered formats.
The *Transfer.java in `eclipse.platform.swt\bundles\org.eclipse.swt\Eclipse SWT Drag and Drop\win32\org\eclipse\swt\dnd` map between Java to native.

mihnita:
Sorry for the long blurb.

To summarize:
* `StyledText` can copy plain text, RTF, and now HTML
* `StyledText` can paste plain text only
* the 14 classes I'm changing in https://github.com/eclipse-pde/eclipse.pde/pull/363 are incorrect when they check for `RTFTransfer` and claim that if present it `canPaste`
* but that incorrect behavior is not a observable. Even if the code detects an RTFTransfer in the clipboard, it will be very-very-very likely that there is also a TextTransfer in the clipboard. So it is possible to paste, and unlikely to have visible buggy behavior. `canPaste` returns true because it found a `RTFTransfer`, the "Paste" menu is enabled, and when I select it the `paste()` operation in `StyleText` is invoked, which will paste the `TextTransfer`
* it would also be wrong to add HTMLTransfer to those tests in the 14 classes, the way my PR intended to do.
* removing the RTFTransfer in the 14 classes would make things correct, but probably not fix any observable bug. For a bug to be observable you need an application that puts RTF in the clipboard, but not text. Which is something I've never seen. Putting text in the clipboard (a native format) is a lot easier than putting RTF (a registered format). So if you do RTF it is hard to get the text part wrong or not do it at all. That's why "leave things as is" is not a bad option.

mihnita:
Proposal:
 * Take a look and maybe review / approve https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/296
   That is a direct fix for the observable bug, where you copy code from the Java editor, paste into Google Slides / Doc / Gmail, and it is not colored (https://github.com/eclipse-platform/eclipse.platform.text/issues/101).
   Pretty straightforward fix: we used to put `TextTransfer` and `RTFTransfer` in the clipboard, now we also put `HTMLTransfer`
 * I drop this and https://github.com/eclipse-pde/eclipse.pde/pull/363, which are more unclear.
    Things would not be worse than they are, a bug would probably be observable very-very rarely.

That way we have https://github.com/eclipse-platform/eclipse.platform.text/issues/101 fixed for 2022-12.

WDYT?

Thank you,
Mihai

Issue #440, opened 2022-10-19
ArrayIndexOutOfBoundsException


jukzi:
When i hover a project in the package Explorer:
```
java.lang.ArrayIndexOutOfBoundsException: Index 0 out of bounds for length 0
	at org.eclipse.swt.widgets.Tree.findCell(Tree.java:2799)
	at org.eclipse.swt.widgets.Tree.WM_MOUSEMOVE(Tree.java:6970)
	at org.eclipse.swt.widgets.Control.windowProc(Control.java:4790)
	at org.eclipse.swt.widgets.Tree.windowProc(Tree.java:6142)
	at org.eclipse.swt.widgets.Display.windowProc(Display.java:4977)
	at org.eclipse.swt.internal.win32.OS.DispatchMessage(Native Method)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:3595)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine$5.run(PartRenderingEngine.java:1155)
	at org.eclipse.core.databinding.observable.Realm.runWithDefault(Realm.java:338)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine.run(PartRenderingEngine.java:1046)
	at org.eclipse.e4.ui.internal.workbench.E4Workbench.createAndRunUI(E4Workbench.java:155)
	at org.eclipse.ui.internal.Workbench.lambda$3(Workbench.java:643)
	at org.eclipse.core.databinding.observable.Realm.runWithDefault(Realm.java:338)
	at org.eclipse.ui.internal.Workbench.createAndRunWorkbench(Workbench.java:550)
	at org.eclipse.ui.PlatformUI.createAndRunWorkbench(PlatformUI.java:171)
	at org.eclipse.ui.internal.ide.application.IDEApplication.start(IDEApplication.java:152)
	at org.eclipse.equinox.internal.app.EclipseAppHandle.run(EclipseAppHandle.java:203)
	at org.eclipse.core.runtime.internal.adaptor.EclipseAppLauncher.runApplication(EclipseAppLauncher.java:136)
	at org.eclipse.core.runtime.internal.adaptor.EclipseAppLauncher.start(EclipseAppLauncher.java:104)
	at org.eclipse.core.runtime.adaptor.EclipseStarter.run(EclipseStarter.java:402)
[...]

PR #442, opened 2022-10-19
fix cachedItemOrder #440

jukzi:
Prevent ArrayIndexOutOfBoundsException and cache old order till new order is available after drag and drop.
Drag and drop can be tested with Problems View: A wrong order would show vertical separator lines in the table that do not match the header column.


jukzi:
ignoring https://github.com/eclipse-platform/eclipse.platform.swt/issues/294

PR #446, opened 2022-10-21
Issue #445: [GTK] Fix IAE on Linux with smooth antialiasing enabled

### LINKED ISSUE #445: IllegalArgException on Linux with smooth antialiasing enabled
**Describe the bug**

Running on Linux, with a HiDPI screen, with a scaling factor of 2, with smooth anti-aliasing enabled, my application exits with the following error:
```
java.lang.IllegalArgumentException: Argument not valid
	at org.eclipse.swt.SWT.error(SWT.java:4899)
	at org.eclipse.swt.SWT.error(SWT.java:4833)
	at org.eclipse.swt.SWT.error(SWT.java:4804)
	at org.eclipse.swt.graphics.GC.drawImage(GC.java:873)
	at org.eclipse.swt.graphics.GC.drawImage(GC.java:854)
	at org.eclipse.swt.internal.DPIUtil.autoScaleImageData(DPIUtil.java:243)
	at org.eclipse.swt.internal.DPIUtil.autoScaleUp(DPIUtil.java:279)
	at org.eclipse.swt.graphics.Image.<init>(Image.java:417)
	at org.eclipse.jface.resource.URLImageDescriptor.createImage(URLImageDescriptor.java:260)
	at org.eclipse.jface.resource.ImageDescriptor.createImage(ImageDescriptor.java:290)
	at org.eclipse.jface.resource.ImageDescriptor.createImage(ImageDescriptor.java:268)
```

**To Reproduce**

Either use a HiDPI environment with the scaling factor set to 2 (200%) or set `GDK_SCALE=2` in the environment.

When running the following snippet, pass the system property `-Dswt.autoScale.method=smooth`

Run the following snippet, where "gear_icon1.png" is some 16x16 icon, e.g.: gear_icon1

```
public class ScalingTest2 {

	public static void main(String[] args) throws MalformedURLException {
		shell.setSize(100, 100);

		System.out.println("Device Zoom: " + DPIUtil.getDeviceZoom());

		ImageDescriptor desc1x = ImageDescriptor.createFromURL(new URL("file:gear_icon1.png"));
		Image image1x = desc1x.createImage();

		ToolBar bar = new ToolBar(shell, SWT.BORDER | SWT.FLAT);
		Rectangle clientArea = shell.getClientArea();
		bar.setBounds(clientArea.x, clientArea.y, 100, 32);

		ToolItem item1 = new ToolItem(bar, SWT.NONE);
		item1.setImage(image1x);
		item1.setToolTipText("1x");

		shell.open();
		while (!shell.isDisposed()) {
				display.sleep();
			}
		}
		image1x.dispose();
		display.dispose();
	}
}
```

**Screenshots**

| | Default antialiasing  | With `-Dswt.autoScale.method=smooth` |
| ------------- | ------------- | ------------- |
| Windows 10 | windows_default | windows_smooth |
| Linux | linux_default | java.lang.IllegalArgumentException: Argument not valid<br />(see stack trace above) |

**Expected behavior**

No crash, the icon should be antialiased when scaled up 2x, similar to the Windows 10 screenshot above.

**Environment:**

Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [ ] Windows
- - [x] Linux
- - [ ] macOS 

Fedora 36
GTK version: gtk3-3.24.34-1.fc36.x86_64
Java version: java-17-openjdk-17.0.4.1.1-1.fc36.x86_64
SWT version: git commit 6c09b67 (15th Oct 2022)
 
**Version since**

Unknown

Workaround is don't enable smooth antialiasing :-)

mbooth101:
Incorrect dimensions were being used for the source image during an autoScaleImageData operation. Instead take the source dimensions from the given image's ImageData which contains the actual source image information whose dimensions are accurate.

Fixes #445


mbooth101:
With this change, the combination of `GDK_SCALE=2` and `-Dswt.autoScale.method=smooth` now has the correct results for antialiasing images that do not have Hi-DPI @2x versions:

 
Screenshot from 2022-10-21 11-11-43

iloveeclipse:
Considering the original code throws exception, do you think you could craft a regression test?

mbooth101:
> Considering the original code throws exception, do you think you could craft a regression test?

Hmm, because the autoScale settings are statically initialised in the DPIUtil I don't think I can set the system properties early enough in a automated unit -- at least not without affect all other tests.

I can however at least add a self-contained manual test snippet -- see latest patch revision.

mbooth101:
Hi @iloveeclipse 

I've spent the last few days thinking about your comment, and I found a way to reproduce the exception in an automated tests, so I've updated the patchset to include test suite improvements.

The process of adding new automated tests actually revealed another bug in GTK image scaling (scaling factor incorrectly truncated instead) so this patchset now contains a bonus extra bugfix!

I would welcome a re-review from you (or anyone else) if you have time :-)

mbooth101:
I take a lack of dissent as tacit approval :-)

akurtakov:
Absolutely, we are very short of reviewers lately so I strongly encourage committers to not only push their own patches but review others too!

iloveeclipse:
@mbooth101, @akurtakov : this seem to introduce a regression, see https://github.com/eclipse-platform/eclipse.platform.swt/issues/2451#issuecomment-3228102378

Could you please check?

akurtakov:
I will try to find time for it. As the regression has been around for ~3 years, it is not a blocker for 4.37  IMO.

Issue #450, opened 2022-10-24
Control.setFocus must not bring the application to front if it is in the background


tmssngr:
Please run the attached snippet. It opens 3 shells, simulates some loading and then wants to set the focused control for each of the shells. Unfortunately, (sometimes) it brings the shell to front, even if it was in the background. If I would want to bring a shell to front, I would invoke `shell.setActive()`, but I just want to change the focuses control inside a shell.
```java
import org.eclipse.swt.*;
import org.eclipse.swt.layout.*;
import org.eclipse.swt.widgets.*;

public class FocusToFrontTest {

	public static void main(String[] args) {
		final FocusToFrontTest test = new FocusToFrontTest();
		test.openShell();
		test.openShell();
		test.openShell();
		test.show();
	}

	private final Display display = new Display();

	private int count;

	private FocusToFrontTest() {
	}

	private void openShell() {
		count++;
		final Shell shell = new Shell(display);
		final int number = count;
		shell.setText("Window " + number);
		shell.setLayout(new FillLayout());
		shell.addListener(SWT.Dispose, event -> count--);
		shell.addListener(SWT.Activate, event -> System.out.println("Activated " + number));
		shell.setSize(400, 300);

		final Text text1 = new Text(shell, SWT.BORDER | SWT.MULTI);
		text1.setText("Loading...");

		final Text text2 = new Text(shell, SWT.BORDER | SWT.MULTI | SWT.WRAP);

		shell.open();

		display.timerExec(1_000 * number, () -> {
			if (shell.isDisposed()) {
				return;
			}

			text1.setText("");
			text2.setText("Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.");
			text2.setFocus();
		});
	}

	private void show() {
		while (count > 0) {
			if (!display.readAndDispatch()) {
				display.sleep();
			}
		}

		display.dispose();
	}
}
```
This ticket is related to https://bugs.eclipse.org/bugs/show_bug.cgi?id=578929, but instead of adding new API it should fix the existing one.

[...]

PR #451, opened 2022-10-24
Issue #450 - Control.setFocus must not bring the application to front…

tmssngr:
… if it is in the background

This should prevent bringing the application to front, if the active shell is different from the control's shell. The only way of changing the control that has the focus, is by invoking control.setFocus(). If the shell also should be brought to front,
control.getShell().setActive() should be invoked before invoking control.setFocus().

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Control.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Control.java
@@ -1444,7 +1444,7 @@ public boolean forceFocus () {
 	if (display.focusEvent == SWT.FocusOut) return false;
 	Decorations shell = menuShell ();
 	shell.setSavedFocus (this);
-	if (!isEnabled () || !isVisible () || !isActive ()) return false;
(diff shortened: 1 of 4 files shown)


akurtakov:
Patch looks good. I can't test non Linux behavior but as the change is the same I think it's fine. The issue didn't happen on Wayland but only on x11 but still important improvement.

iloveeclipse:
@tmssngr : I see a bigger regression now, see https://github.com/eclipse-platform/eclipse.platform.swt/issues/450#issuecomment-1352853224 and plan to revert the commit this evening.

Please check if you can provide a fix.
Steps to reproduce: in SDK, use Ctrl+Shift+T to open "Open Type" dialog.
Right after dialog is opened, the filter text should have focus, so user can type into directly after opening the dialog.
With latest change from SWT this doesn't work anymore, because widget sets the focus before shell is opened.

The main code is in org.eclipse.ui.dialogs.FilteredItemsSelectionDialog that is used by lot of IDE code.

I assume that can be easily reproduced without JFace / Platform UI etc, just by opening a dialog with a text from a shell, where the text is created & focused after shell creation but before it is opened/made visible.

PR #469, opened 2022-11-22
Fix #467: Paste as HTML collapses white spaces

### LINKED ISSUE #467: Paste as HTML (copied from a StyledText) collapses white spaces
When StyledText saves HTML to clipboard and the result is pasted into some application, in some cases the multiple spaces are collapsed into one.

This happens in GMail, MS Word, and Powerpoint.
The spaces are preserved in Google Docs and Google Slides.

This is most obvious for code, which needs indent.

**To Reproduce**

Run this code in Eclipse:
```
public class Main {
		System.out.println("\033[31mpublic class\033[m Main {");
		System.out.println("    \033[31mpublic static void\033[m main(String[] args) {");
		System.out.println("        \033[32m// TODO Auto-generated method stub\033[m");
		System.out.println("        System.\033[34mout\033[m.println(\033[34m\"Hello world!\"\033[m);");
		System.out.println("    }");
		System.out.println("}");
	}
}
```
The console should look like syntax colored "Hello world" app.
Select the text in console and copy to clipboard. You can use Ctrl+Insert in Win / Linux (Option+Insert on Mac) to copy without escape codes.

Paste the clipboard content in Gmail (create new email), or in MS Word / MS PowerPoint.

**Expected behavior**

The spaces should be preserved, the result when pasted into an application should look as close as possible to what we see in the StyledText (for example in Eclipse).

**Screenshots**

image

**Environment:**
- - [x] All OS
- - [ ] Windows
- - [ ] Linux
- - [ ] macOS 

Tried on Win 11, MacOS, Debian

3. JRE/JDK version
 
11, but very likely not relevant.

**Version since**

Eclipse 2022-12 M3 (4.26.0 M3)
HTML was not available as clipboard format until recently (Sep 19, 2022)
This was discovered while testing https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/320.
Before it was harder to notice because I did not copy-paste text with multiple spaces, which is more common in code.

No workarounds.

But the cause is easy to see if we dump the clipboard and inspect the HTML (or look at the code).
The `div` wrapping the content specifies `style=' ... white-space:nowrap;'`

But `white-space:nowrap` _"collapses white space as for normal, but suppresses line breaks (text wrapping) within the source."_
https://developer.mozilla.org/en-US/docs/Web/CSS/white-space

Need to use `white-space:pre`
_"Sequences of white space are preserved. Lines are only broken at newline characters in the source and at `<br>` elements."_

The code is here:
https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/bundles/org.eclipse.swt/Eclipse%20SWT%20Custom%20Widgets/common/org/eclipse/swt/custom/HTMLWriter.java#L77

mihnita:
Tested on Win 11, MacOS  Ventura 13.0.1, Debian 10.

Tested Gmail, Google Docs, Google Slides:
- Win: Chrome & Firefox
- Debian: Chrome & Firefox
- MacOS: Chrome, Firefox, and Safari

Tested platform native apps:
 - MS Word 2000 and PowerPoint 2000 on Win
 - LibreOffice Writer (~Word / GDocs) and LibreOffice Impress (~PowerPoint / GSlides )
 - The Apple Mail, Notes, and TextEdit apps on MacOS (they come with the system)

Screenshots for everything here: https://mihai-nita.net/eclipse_swt_467/

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/HTMLWriter.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/HTMLWriter.java
@@ -37,8 +37,8 @@ class HTMLWriter extends StyledTextWriterBase {
 	@Override
 	public void close() {
 		if (!isClosed()) {
-			write("</div>\n");
-			write("</div>\n");
+			write("</div>");
+			write("</div>");
 			super.close();
 		}
 	}
@@ -52,29 +52,27 @@ class HTMLWriter extends StyledTextWriterBase {
 		appendStyle(innerDivStyle, "color:", styledText.getForeground(), ";");
 		appendStyle(innerDivStyle, "background-color:", styledText.getBackground(), ";");
 
-		appendStyle(outerDivStyle, "padding-left:", styledText.getLeftMargin(), "px;");
-		appendStyle(outerDivStyle, "padding-top:", styledText.getTopMargin(), "px;");
-		appendStyle(outerDivStyle, "padding-right:", styledText.getRightMargin(), "px;");
-		appendStyle(outerDivStyle, "padding-bottom:", styledText.getBottomMargin(), "px;");
+		appendStyle(outerDivStyle, "padding:"
+						+ styledText.getTopMargin() + "px "
+						+ styledText.getRightMargin() + "px "
+						+ styledText.getBottomMargin() + "px "
+						+ styledText.getLeftMargin() + "px;");
 
 		String language = appendFont(innerDivStyle, styledText.getFont(), 0);
 
-		// Using wrapIndent like this messes up the line background (if set).
-//		int wrapIndent = styledText.getWrapIndent();
-//		if (wrapIndent != 0) {
-//			appendStyle(globalStyle, "padding-left:", wrapIndent, "px;");
-//		}
-//		int indent = styledText.getIndent() - wrapIndent;
 		int indent = styledText.getIndent();
 		if (indent != 0) {
 			appendStyle(innerDivStyle, "text-indent:", indent, "px;");
 		}
 
-		// TODO?
-//		int lineSpacing = styledText.getLineSpacing();
-
-		if (!styledText.getWordWrap()) {
-			appendStyle(innerDivStyle, "white-space:nowrap;");
+		if (styledText.getWordWrap()) {
+			// Sequences of white space are preserved.
+			// Lines are broken at newline characters, at <br>, and as necessary to fill line boxes.
+			appendStyle(innerDivStyle, "white-space:pre-wrap;");
+		} else {
+			// Sequences of white space are preserved.
+			// Lines are only broken at newline characters in the source and at <br> elements.
+			appendStyle(innerDivStyle, "white-space:pre;");
 		}
 
 		appendAlignAndJustify(innerDivStyle, styledText.getAlignment(), styledText.getJustify());
@@ -83,17 +81,11 @@ class HTMLWriter extends StyledTextWriterBase {
 			appendStyle(innerDivStyle, "direction:rtl;");
 		}
 
-		// Do we have any use for these?
-//		Color selectionBackground = styledText.getSelectionBackground();
-//		Color selectionForeground = styledText.getSelectionForeground();
-//		int tabs = styledText.getTabs();
-//		boolean blockSelection = styledText.getBlockSelection();
-
-		write("<div style='" + outerDivStyle + "'>\n");
+		write("<div style='" + outerDivStyle + "'>");
 		if (language == null || language.isEmpty()) {
-			write("<div style='" + innerDivStyle + "'>\n");
+			write("<div style='" + innerDivStyle + "'>");
 		} else {
-			write("<div lang='" + language + "' style='" + innerDivStyle + "'>\n");
+			write("<div lang='" + language + "' style='" + innerDivStyle + "'>");
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


mihnita:
**Note:** this depends on https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/320, otherwise copying code from the editor will not store HTML in clipboard.

**Note:** I know this is late. But copy HTML to clipboard did not exist before 2022-12, it is a new feature.
So this can't break some existing functionality
(OK, it can break copy as plain text and as RTF, but those changes are already in, not affected by this fix)

You can see screenshots for all the tests.

No case got worse, in most cases things improved, and in some cases they stayed the same.
It might be that the paste used the RTF form, but the application had no option to say "Paste as <clipboard_format>", so no way to tell (Apple-Notes-macos, Apple-TextEdit-macos)

I am preparing a summary of what cases where fixed and what cases were not.

mihnita:
| Case                      | Indent      | Line height | Empty line  | Comment |
| ------------------------- | ----------- | ----------- | ----------- | ------- |
| Apple-Mail-macos          | better      | better      | better      |         |
| Apple-Notes-macos         | same (good) | same (good) | same (good) | No option to paste special. So it might be RTF, or not. |
| Apple-TextEdit-macos      | same (good) | same (good) | same (good) | No option to paste special. So it might be RTF, or not. |
| GDocs-Chrome-debian       | same (good) | better      | better      |         |
| GDocs-Chrome-macos        | same (good) | better      | better      |         |
| GDocs-Chrome-win          | better      | better      | better      |         |
| GDocs-Firefox-debian      | same (good) | better      | better      |         |
| GDocs-Firefox-macos       | same (good) | better      | better      |         |
| GDocs-Firefox-win         | better      | better      | better      |         |
| GDocs-Safari-macos        | better      | better      | better      |         |
| Gmail-Chrome-debian       | better      | better      | better      |         |
| Gmail-Chrome-win          | better      | better      | better      |         |
| Gmail-Firefox-debian      | better      | better      | better      |         |
| Gmail-Firefox-macos       | better      | better      | better      |         |
| Gmail-Firefox-win         | better      | better      | better      |         |
| Gmail-Safari-macos        | better      | better      | better      |         |
| GSlides-Chrome-debian     | same (good) | better      | better      |         |
| GSlides-Chrome-macos      | same (good) | better      | better      |         |
| GSlides-Chrome-win        | better      | better      | better      |         |
| GSlides-Firefox-debian    | same (good) | better      | better      |         |
| GSlides-Firefox-macos     | same (good) | better      | better      |         |
| GSlides-Firefox-win       | better      | better      | better      |         |
| GSlides-Safari-macos      | better      | better      | better      |         |
| LibreOfficeImpress-debian | same (bad)  | same (bad)  | same (good) |         |
| LibreOfficeWriter-debian  | same (bad)  | same (bad)  | better      |         |
| MSPowerPoint2000-win      | same (bad)  | better      | better      |         |
| MSWord2000-win            | better      | better      | better      |         |
| WordPad-win               | same (good) | same (bad)  | better      |         |

Legend:
* better : was incorrect before the fix, and correct after (fixed)
* same (good) : was correct before the fix, and correct after
* same (bad)  : was incorrect before the fix, and incorrect after

iloveeclipse:
Note: LibreOfficeWriter is same for me as without the patch (no indentation), also looking on your screenshots I don't see any improvement.

mihnita:
> Note: LibreOfficeWriter is same for me as without the patch (no indentation), also looking on your screenshots I don't see any improvement.

Yes, this is also what I see. This is why the summary says "same (bad)" in the first 2 columns.
The only thing that seems to improve was the treatment of empty line (between `package` and `class`) in the Writer.
So I didn't claim any improvement there.
But there are a lot of cells saying "better"

My theory (right now) is that the LibreOffice support for HTML clipboard is poor.
I will try to copy paste from other tools that put HTML in clipboard (browser, VS Code), to see if it is any better.

But even without LibreOffice, I home that the improvements are worth landing?

mickaelistria:
This change seems quite good. Please add a commit that increases version by +0.0.1 and we'll try to have it merged early after 4.26 is released.

mihnita:
Thank you.

Is there a place where I can read about what goes in each Eclipse stage (Mx, RCx)?
Because all I'm familiar with is what the Wiki says: https://wiki.eclipse.org/Category:SimRel-2022-12
But from that it is unclear to me if bug fixes that don't change public APIs, go, and up to what stage.
If new features don't go in after RC1, and bug fixes don't go in after RC1, then nothing goes in :-)
I know that in reality "important bug fixes" would go in.
But different project have different rules for what's important enough to make it.

Anyway, would be good if I can get familiar with the rules, so that I don't waste people time asking for reviews on issues that don't have a chance to make it. Especially at the time when everybody is busy with the last weeks before the release.
(like I did with https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/320).
I want to help, not to be a nuisance :-)

Thank you very much,
Mihai

akurtakov:
I think you're looking for https://www.eclipse.org/eclipse/development/plans/freeze_plan_4_26.php

mihnita:
Thanks,
Kind of sore of :-)

It still does not clarify much, as "project lead or PMC vote on the bug report" explains the process, but not the rules.
As in "what makes a bug important enough to escalate" to a PMC?
For example, is this a important enough to bother a PMC?
(I know that probably not, it's just an example. But that's the kind of guideline I was looking for)

But I know a bit more than before :-)

Thank you!
M.

mickaelistria:
Usually in RC phases, only the regressions compared to latest releases are considered as important enough.

mihnita:
Got you! Thanks!
M.

PR #471, opened 2022-11-23
Add methods to execute tasks as CompletableFuture with a "busy" user feedback to BusyIndicator

laeubi:
Currently it is very hard in SWT to have code that takes a bit longer than executed directly but do not take long enought to show "heavy" feedback to the user like a progress monitor. BusyIndicator claims to help with this task but is hard to be used correctly and cumbersume to use when it comes to further work with the result data, and the Snippet130 demonstrate this complexity quite well: https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet130.java

This adds new methods to BusyIndicator that allows easy usage by integration using CompletableFutures. The Snippet 130 is adjusted to show this much more modern aproach that shows how easy is is to acomplish non blocking mid-lenth running operations.

This is another one towards
- https://github.com/eclipse-platform/eclipse.platform.swt/issues/64

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/BusyIndicator.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/BusyIndicator.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2011 IBM Corporation and others.
+ * Copyright (c) 2000, 2022 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -10,11 +10,13 @@
  *
  * Contributors:
  *     IBM Corporation - initial API and implementation
+ *     Chrsitoph Läubrich - add methods to work with {@link CompletableFuture}s
  *******************************************************************************/
 package org.eclipse.swt.custom;
 
-
 import java.util.*;
+import java.util.concurrent.*;
+import java.util.concurrent.atomic.*;
 
 import org.eclipse.swt.*;
 import org.eclipse.swt.graphics.*;
@@ -28,7 +30,7 @@ import org.eclipse.swt.widgets.*;
  */
 public class BusyIndicator {
 
-	static int nextBusyId = 1;
+	private static final AtomicInteger nextBusyId = new AtomicInteger();
 	static final String BUSYID_NAME = "SWT BusyIndicator"; //$NON-NLS-1$
 	static final String BUSY_CURSOR = "SWT BusyIndicator Cursor"; //$NON-NLS-1$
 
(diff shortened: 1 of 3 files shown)


laeubi:
If there are no further concerns / feedback I plan to merge this end of next week.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/BusyIndicator.java
```diff
@@ -59,32 +61,154 @@ public static void showWhile(Display display, Runnable runnable) {
...
+	 *                 the {@link CompletableFuture}
+	 * @param executor the Executor to perform the computation in case this is
+	 *                 called from a {@link Display} thread, passing a
+	 *                 {@link Display} will throw an
+	 *                 {@link IllegalArgumentException} as this will lead to a
+	 *                 blocking UI and violates the contract of this method
+	 * @since 3.122
+	 */
+	public static <E extends Exception> CompletableFuture<?> execute(SwtRunnable<E> action, Executor executor) {
+		return compute(() -> {
```
szarnekow, line 110: I don't the see passed Executor being used here. I'm certainly missing something.

szarnekow: Ah, ok it's just missing as an argument to the delegate compute()

laeubi: Good catch, this should be fixed now!

mickaelistria:
Please add a note to the N&N doc about this new API.

iloveeclipse:
This change is incomplete, it produces API error (obviously)

iloveeclipse:
@laeubi : would you be able to update all above today / tomorrow? The M1 build is scheduled for this week.

iloveeclipse:
I've created #520 to follow up.

laeubi:
@iloveeclipse I also have seen this in the IDE (strange enough the build her succeeds) but still not sure if it is an API tools glitch or something to do, I created:

- https://github.com/eclipse-platform/eclipse.platform.swt/pull/511

and
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/510

but both builds indicate a success regarding API checks so I'm a bit lost here what is the right approach...

iloveeclipse:
> but still not sure if it is an API tools glitch or something to do

Reading through https://wiki.eclipse.org/Version_Numbering#When_to_change_the_minor_segment should give a hint that something is wrong. Adding new API without incrementing bundle minor segment is always an error.

I assume command line API check **in jenkins**  is broken for SWT (or all bundles?).

I see `[INFO] --- tycho-apitools-plugin:4.0.0-SNAPSHOT:generate (generate) @ org.eclipse.swt ---
[INFO] ` entries but I have no idea what tycho/maven does there and if that is doing anything at all.

Please check and create a bug for SWT or tycho or whoever could be involved here.

mickaelistria:
Indeed, API Tools validation at build time is not configured for SWT at the moment. The reason is that it appeared tricky back then. For API Tools to work, project must be well configured, ie have a valid .classpath. We should maybe add to the org.eclipse.swt pom file a step that sets up the right .classpath so hopefully it will make API Tools proceed.

PR #477, opened 2022-11-28
Bug 476 - SWT clipboard HTML transfer doesn't work

jasonhy-wang:
+ implemented HTML transfer in gtk4 clipboard snippet

Signed-off-by: Jason Wang

Tracked in #476


akurtakov:
Thanks Jason!

PR #484, opened 2022-12-01
[GTK] Added input checking for DateTime.java SWT.TIME Widget

### LINKED ISSUE #180: DateTime's behavior in Linux is undesirable
**Describe the bug**
On Ubuntu, it's possible to input more than 2 digits when trying to input a time. The behavior is different than the one in Windows, where the time gets automatically edited to the next meaningfull integer based on last input.

**To Reproduce**
1. Create a time picker by writing 
`DateTime time = new DateTime(timeComposite, SWT.TIME | SWT.SHORT | SWT.BORDER | SWT.CENTER);`
2. Interact with it by typing numbers that would result in non-valid hours or minutes, which are shown nevertheless.

**Expected behavior**
Scenario: 3:00 AM is set.
Try and edit the hours by typing 31. The hours should change to a 1, since 31 would not be valid.
If 1 is pressed again, the hours should change to 11.
If 1 is pressed again, the hours should change to 1, again.

**Screenshots**
image

**Environment:**
- Select the platforms on which the behavior is seen.
[] All OS
[ ] Windows
[x] Linux
[ ] macOS (can't verify)
 
 - Additional OS info (e.g. OS version, Linux Desktop, etc)
 Ubuntu 20.04 running in a VM.
 
 - JRE/JDK version
 Temurin jdk-17.0.2+8.
 
**Version since**
 Always been a problem AFAIK.

There are no possible workarounds, since the text that's displayed while in editor mode is inside an incapsulated variable of DateTime.java.

A possible solution may be to make the method "onNumberKeyInput" aware of what's been edited and hence set constraints.

majanojoel:
+ Added validateNewTimeValue() and findNumTimeSeparators() to check for valid key input and gets corrected to a more correct state.

+ Added a check for valid key input for the AM_PM field due to all characters being accepted by this field. Now only alphabetic characters change this field.

Tested using Snippet 250 and ran the DateTime Junit test suite.

Fixes #180 

Signed-off-by: Joel Majano <majanojoel@gmail.com>

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/DateTime.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/DateTime.java
@@ -1828,6 +1828,10 @@ boolean onNumberKeyInput(int key) {
 		if ((style & SWT.TIME) != 0 && fieldName != Calendar.AM_PM && !Character.isDigit(newText)) {
 			return false;
 		}
+		// Only allow non-digit characters to modify AM/PM field
+		if(fieldName == Calendar.AM_PM && !Character.isAlphabetic(newText)) {
+			return false;
+		}
 		if (!Character.isAlphabetic(newText) && !Character.isDigit(newText)) {
 			return false;
 		}
@@ -1851,6 +1855,9 @@ boolean onNumberKeyInput(int key) {
 		typeBufferPos++;
 	}
 	StringBuilder newText = new StringBuilder(prefix);
+	if((style & SWT.TIME)!=0 && fieldName != Calendar.AM_PM){
+		typeBuffer = validateNewTimeValue(prefix.toString(), String.valueOf(keyToString(key)));
+	}
 	newText.append(typeBuffer);
 	newText.append(suffix);
 	setText(newText.toString());
@@ -1866,6 +1873,58 @@ boolean onNumberKeyInput(int key) {
 		}
 	}
 	return false;
+}
+
+private StringBuilder validateNewTimeValue(String prefix, String keyValue) {
+	if(typeBuffer.toString().isEmpty()) return typeBuffer;
+	int numSeparators = findNumTimeSeparators(prefix);
+	String fixedBuffer = "";
+	int val;
+	try{
+		val = Integer.parseInt(typeBuffer.toString());
+	}
+	catch(Exception e){
+		return typeBuffer;
+	}
+	if(numSeparators == 0) {
+		if (val > 23 || val <0) {
+			fixedBuffer = keyValue;
+		}
+	}
+	else if(numSeparators == 1 || numSeparators == 2) {
+		if (val > 59 || val <0) {
+			fixedBuffer = keyValue;
+		}
+	}
+	else {
+		return typeBuffer;
+	}
+	if (fixedBuffer.length() == 1) {
+		typeBufferPos = 1;
+	}
+	/* 	For typeBuffer containing 0, we need to trim
+	 *  the buffer to contain at max 2 0s.
+	 *  Reset the length if greater than 2.
+	 */
+	if(typeBuffer.length() > 2) {
+		fixedBuffer = typeBuffer.substring(2);
+		typeBufferPos = fixedBuffer.length();
+	}
+	return fixedBuffer.isEmpty() ? typeBuffer : new StringBuilder(fixedBuffer);
+}
+
+private int findNumTimeSeparators(String str) {
+	int numSeparators = 0;
+	if(str.isEmpty()) return 0;
+	char prev = str.charAt(0);
+	for(int i = 1; i < str.length(); i++) {
+		char c = str.charAt(i);
+		if (!Character.isDigit(c) && Character.isDigit(prev)) {
+			numSeparators++;
+		}
+		prev = c;
+	}
+	return numSeparators;
 }
 
 private char keyToString(int key) {


majanojoel:
Forgot to mention that this bug was creating Gtk-CRITICAL warning messages from GtkSpinButton when the text became longer than 4 characters. Those are not present anymore.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/DateTime.java
```diff
@@ -1868,6 +1872,28 @@ boolean onNumberKeyInput(int key) {
 	return false;
 }
 
+private StringBuilder validateNewTimeValue(String prefix, String keyValue) {
+	if(typeBuffer.toString() == "") return typeBuffer;
+	int numColons = (prefix.length() - prefix.replace(":", "").length()) / ":".length();
+	String finalVal;
+	//numColons mapping: 0 -> hours, 1 -> minutes, 2-> seconds
+	int val = Integer.valueOf(typeBuffer.toString());
```
akurtakov, line 1880: Please use Integer.parseInt to avoid useless unboxing.

akurtakov: Furthermore this tries to parse AM/FM which  leads into yet another NumberFormatException

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/DateTime.java
```diff
@@ -1831,6 +1831,9 @@ boolean onNumberKeyInput(int key) {
 		if (!Character.isAlphabetic(newText) && !Character.isDigit(newText)) {
 			return false;
 		}
+		if (fieldName == Calendar.AM_PM && (newText != 'p' && newText !='a')){
```
akurtakov, line 1834: Upper case A and P do nothing and even thought what if other locales have different indicators for AM/PM ? As per https://en.wikipedia.org/wiki/12-hour_clock#Abbreviations : The equivalents in [Greek](https://en.wikipedia.org/wiki/Greek_language) are π.μ. and μ.μ., respectively, and in [Sinhala](https://en.wikipedia.org/wiki/Sinhala_language) පෙ.ව. (pe.va.) for පෙරවරු (peravaru, පෙර pera – fore, pre) and ප.ව. (pa.va.) for පස්වරු (pasvaru, පස්සේ passē – after, post). So any such filtering should really not happen

akurtakov:
Please redo to only check the numbers to be of the proper size.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/DateTime.java
majanojoel: Localization support right, okay I'll remove this.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/DateTime.java
```diff
@@ -1868,6 +1872,28 @@ boolean onNumberKeyInput(int key) {
 	return false;
 }
 
+private StringBuilder validateNewTimeValue(String prefix, String keyValue) {
+	if(typeBuffer.toString() == "") return typeBuffer;
```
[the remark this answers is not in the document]
majanojoel: For this case, you mention that you were able to delete the hour field, does this mean you were able to delete the colon (":") in the field? If yes, shouldn't this not be allowed. I.e if the key is "Delete" and the charAt the position to be deleted is a colon, this input should be discarded?

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/DateTime.java
majanojoel: Okay, this seems simple enough to fix, run the validtime check only if SWT.TIME && field != AM_PM. I overlooked this point but it is what it should have been in the first place.

majanojoel:
> Please redo to only check the numbers to be of the proper size.

I can keep the check for valid time (0-23, 0-59), and the checks for size. For the AM_PM field, due to localization support, I'm not sure of a completely safe way to limit input size, unless it is guaranteed that all locales have the same number of characters in their "A.M"/"P.M" equivalents.

majanojoel:
Updated the PR with the required fixes and added a check to only allow non-digit values for the AM_PM field. Let me know if this is still not correct. This still doesn't address the AM_PM field typeBuffer getting too large, but that should be made in a different PR.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/DateTime.java
```diff
@@ -1868,6 +1875,38 @@ boolean onNumberKeyInput(int key) {
 	return false;
 }
 
+private StringBuilder validateNewTimeValue(String prefix, String keyValue) {
+	if(typeBuffer.toString().isEmpty()) return typeBuffer;
+	int numColons = (prefix.length() - prefix.replace(":", "").length()) / ":".length();
```
mihnita, line 1880: Not all locales use ":" as time separator.
Some use "." (be (Belarusian), it (Italian), lt (Lithuanian), sq (Albanian), sr (Serbian))

Worse, this is not 100% stable, might change between Java versions.
The list above is from Java 8.
Java 17 moved to use CLDR by default, so the list of locales using "." grows (with Assamaese, Danish, more)
Worse, separators in 17 go beyond ":" and ".", French-Canada uses " h " ("19 h 43").
And 19:43 in Dzonga (dz) is "ཆུ་ཚོད་ ༧ སྐར་མ་ ༤༣ ཕྱི་ཆ་" (the separator is "སྐར་མ་")

majanojoel: Updated PR to better handle different locales.

majanojoel:
I generalized finding the number of separators which seems to work in the en and fr locales (":" and " h" / "min"). I also modified the typing behavior to make it more intuitive. Please let me know if further changes are needed.

akurtakov:
Closing as no one seems to be interested in finishing and pushing this one.

PR #485, opened 2022-12-02
Simplify *Transfer check* methods

akurtakov:
* instanceof already checks nulls
* String.isEmpty used


akurtakov:
Long failing test

Issue #486, opened 2022-12-02
SWT Pattern corruption.


mikevwdriver:
**Describe the bug**
I've been able to identify a commit which causes Patterns to become corrupted.  
22797d3f9f09ca67bfd29c864f2aaf8eb452c75c

Sometimes the image becomes corrupted... but other times, the pattern image becomes other images, such as the icons from other ToolItems. 

This issue continues to be a problem in the latest revisions (3d9d8fb030873ee3589db14a11f0b366b4b0478a at the time of writing this)

**To Reproduce**
I've put together a standalone snippet to reproduce the issue.
[snippets.zip](https://github.com/eclipse-platform/eclipse.platform.swt/files/10143667/snippets.zip)

**Expected behavior**
I would expect the pattern to never become corrupted.

**Screenshots**
Before, from an older revision (1bb1a46ea5f4e6b22fece451ca6247f8999d3a10):
![image](https://user-images.githubusercontent.com/16846123/205368451-2febcc3e-e395-4d85-8cd1-6f2686e911b7.png)

After, with this revision:
![image](https://user-images.githubusercontent.com/16846123/205368775-9bb609f7-e3e6-47b7-97d6-c20f7509b9f7.png)

The effect does change when the shell is resized.

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 

2. Additional OS info (e.g. OS version, Linux Desktop, etc)
Windows 10, but a colleague reproduced the same issue on Windows 11.

3. JRE/JDK version
 Java 17

**Version since**
SWT commit 22797d3f9f09ca67bfd29c864f2aaf8eb452c75c

This is the image referred to in the code snippet.
![transparentBackground](https://user-images.githubusercontent.com/16846123/205367693-fe2f0b6e-ac65-4132-b96c-b1c25e6755ec.png)

mikevwdriver:
@niraj-modi

mikevwdriver:
From our application:
![image](https://user-images.githubusercontent.com/16846123/205370948-5a239eb4-8a0c-466a-b5dd-a4dc5d501f4b.png)
As you can see, the Explorer view icon has somehow become the pattern.

mikevwdriver:
And one more final example:
![image](https://user-images.githubusercontent.com/16846123/205371325-7bd87ac3-4c87-4744-9cf4-8011f63d5058.png)

[...]

PR #487, opened 2022-12-04
Fix issue #486 - [win32]: use straight alpha format to create GDI+ bitmaps from images with alpha data

jonathan-meier:
GDI+ texture brushes used in the `Pattern` class create corrupted images if the brush was initialized with a GDI+ bitmap having premultiplied alpha format. This seems to be a bug in GDI/GDI+ since manually drawing the pattern by repeatedly drawing the GDI+ image having premultiplied alpha format works well.

The workaround is to always return GDI+ bitmaps having straight alpha format instead of premultiplied from `Image.createGdipImage`. Unfortunately, this means we have to manually convert from premultiplied to straight alpha format.

Fixes issue #486.


jonathan-meier:
Closing as there is a revised fix for issue #486 in PR #489. See also the comments in issue #486.

PR #488, opened 2022-12-05
Improve Jenkinsfile

mickaelistria:
* Use tools
* do not fail on test failure/error (enable "unstable" build result)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -83,7 +83,6 @@ spec:
 		stage('Prepare-environment') {
 			steps {
 				container('container') {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -83,7 +83,6 @@ spec:
 		stage('Prepare-environment') {
 			steps {
 				container('container') {
-					sh 'mutter --replace --sm-disable &'
```
szarnekow, line 86: Why did you replace `mutter` here? What do you refer to by `use tools` in the PR message?

PR #489, opened 2022-12-05
Fix issue #486: [win32] Fix use-after-free in Pattern

jonathan-meier:
The bitmap used to create a GDI+ texture brush must not be deleted until after the brush is no longer used.

Fixes issue #486.


SyntevoAlex:
The code changes make sense and look good to me.
I tested it briefly and my debugging tools no longer detect any errors.
Also, the reported regression of corrupted painting is resolved.
Therefore, LGTM.

PR #500, opened 2022-12-13
Fix #475 : CTabFolderRenderer should provide more protected methods

### LINKED ISSUE #475: CTabFolderRenderer should provide more protected methods
To facilitate the tab folder customization, the following methods should be protected :

- drawMinimize
- drawMaximize
- drawBorder
- drawBody
- drawXXX in general...

opcoach:
Changes are only about protected access in the class. 

No code has changed. 

Once this PR will be merged, it will be possible to fix the ugly bug in platform (e4 class name referred in swt code ... See :  https://github.com/eclipse-platform/eclipse.platform.ui/issues/462)

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
@@ -30,34 +30,34 @@ public class CTabFolderRenderer {
 
 	protected CTabFolder parent;
 
-	int[] curve;
-	int[] topCurveHighlightStart;
-	int[] topCurveHighlightEnd;
-	int curveWidth = 0;
-	int curveIndent = 0;
-	int lastTabHeight = -1;
+	protected int[] curve;
+	protected int[] topCurveHighlightStart;
+	protected int[] topCurveHighlightEnd;
+	protected int curveWidth = 0;
+	protected int curveIndent = 0;
+	protected int lastTabHeight = -1;
 
-	Color fillColor;
+	protected Color fillColor;
 	/* Selected item appearance */
-	Color selectionHighlightGradientBegin = null;  //null == no highlight
+	protected Color selectionHighlightGradientBegin = null;  //null == no highlight
 	//Although we are given new colours all the time to show different states (active, etc),
 	//some of which may have a highlight and some not, we'd like to retain the highlight colours
 	//as a cache so that we can reuse them if we're again told to show the highlight.
 	//We are relying on the fact that only one tab state usually gets a highlight, so only
 	//a single cache is required. If that happens to not be true, cache simply becomes less effective,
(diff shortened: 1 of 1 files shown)


akurtakov:
For me this is a total no-go. This practically locks everything as API and makes CTabFolder unmanageable. Not to mention that making members protected kills all kind of validation ensurance.

sravanlakkimsetti:
I think we are creating more API here by making them protected. I don't think this is a good idea. Is there way we can avoig making all the variables as api?

opcoach:
Hi Alexander ! 

I will explain why I propose this patch... 

I am working on a project in which we have to customize for instance the way to draw the minimize and maximize buttons displayed on the right of the CTabFolder, or the font, colors of tabs, and why not margin spaces between tabs...

Actually we could expect to extend easily the **CTabFolderRenderer** or its subclass **CTabRendering** and then to use these Renderer as css properties. Unfortunately everything is locked inside... because nothing is protected. 

Access to the **CTabFolder** is also complicated  because all the fields are defined as package, and, for the **CTabRendering**  they are all read using reflection to get their values (see the **CTabFolderWrapper** inside the **CTabRendering**). 

I totally agree with you when you say that all this code is complicated and dangerous to change because all the tabs are drawn manually (and even the min max buttons)...  all the rectangles and coordinates are computed depending on the size of the parent, of the tab texts, etc... 

What I propose is to make this code extendable and allow to redefine some drawing area in an easier way. There are no functional changes...

Could you explain why it locks everything as API ?  And why it makes CTabFolder unmanageable as I didn't change it ? 

May be you have another point of view and you know another way to customize this  **CTabFolderRender** (used by the **PartStackRenderer**) without having to copy everything in our project just to change any graphical customization?

Thanks

akurtakov:
> Could you explain why it locks everything as API ? And why it makes CTabFolder unmanageable as I didn't change it ?

If all these draw* methods are made protected, everyone will be able to override them in their own way thus locking current behavior and even pixel by pixel look  as they rely strongly on knowlegde about each other. This would practically mean that no change in CTabFolder could be done without breaking adopters and I consider this unmanageable state.

opcoach:
Hi Alexander, 

I agree with you, my solution is not the best and I have found another solution by only overriding the draw method from the CTabFolderRenderer.  This PR learned me how to use these renderer which is clearly complicated... 

I wanted also, with this PR,   to fix the bug about the updateCurves in CTabFolderRenderer (see description in https://github.com/eclipse-platform/eclipse.platform.ui/issues/462) but I found also a patch by rewriting only the computeSize in CTabRendering... 

So I will close this PR which is not relevant, and I will close the related bug with 'Won't fix'. 

Thank you for your remarks.
Olivier

akurtakov:
Thanks @opcoach ! This is one of the most complicated widgets in SWT (together with StyledText) and the worst one in terms of maintenance . So many bad memories came from this discussion and reminded me why I strongly believe people should be using plain TabFolder instead :).

opcoach:
Yes I understand, I have seen the mess inside :)  !

PR #510, opened 2022-12-25
Missing version bump indicated by API tools after merge

laeubi:
Inside the IDE the API tools indicated there should be a minor version bump, let see what CI tells about.


iloveeclipse (changes requested):
This is going int the right direction, but not enough.

Please update the pom's and manifests of SWT **fragments** to match host version, see the example list in [2e9528ed10a7a14f9c8afcd9c2fdcf0374773e96](https://github.com/eclipse-platform/eclipse.platform.swt.binaries/commit/2e9528ed10a7a14f9c8afcd9c2fdcf0374773e96)

laeubi:
But this needs to be done in swt.binry repository? Just not sure if it must be done *before* or *After* this change (probably it doesn't matter?) @akurtakov ?

iloveeclipse:
> But this needs to be done in swt.binry repository? Just not sure if it must be done _before_ or _After_ this change

This must be done **in parallel**, so next nightly SDK build can see **both** PR's.

iloveeclipse (commented):
Could you please refer in the commit message to https://github.com/eclipse-platform/eclipse.platform.swt/issues/520 or original issue?

iloveeclipse (approved):

PR #511, opened 2022-12-25
Just a test without VB

laeubi:


akurtakov:
I guess this test served its purpose and can be closed now.

PR #512, opened 2022-12-27
[Build] Skip default checkout, add timestamps and only use tabs

HannesWell:
Skip the unncessary default check-out, because the build operates only on the later site-by-site checked-out sources of o.e.swt and o.e.swt.binaries. Skipping the default checkout can save about a minute for each verification build.

Only use tabs for indentation except for the yml section, where tabs are not supported.
And format lines to make them shorter.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,14 +1,16 @@
 pipeline {
 	options {
+		skipDefaultCheckout() // Specialiced checkout is performed below
+		timestamps()
 		timeout(time: 90, unit: 'MINUTES')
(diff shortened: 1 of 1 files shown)


HannesWell:
The build results seems to be the same like on the master branch. Two Maven builds and the same single test-case failing.
And we have the anticipated save of 2min in each build due to the now skipped default checkout. Although the `Prepare environment` now seems to run slightly longer (~10s), if this is not by chance it might be that the bytes where cached before.
Nevertheless this looks good to go.

Issue #513, opened 2022-12-27
Simplify the build of SWT-binaries and make it more Maven-oriented


HannesWell:
### Discussed in https://github.com/eclipse-platform/eclipse.platform.swt/discussions/497

<div type='discussions-op-text'>

<sup>Originally posted by **HannesWell** December 11, 2022</sup>
The build of the SWT-binaries is a quite complex process and I have the impression that it is even more complex than it would have to be, from the fact that o.e.swt contains the sources for all platforms but the build produces a fragment for each platform that contains all, the common and platform specific compiled code and native libraries.
I'm just about the learn the entire build-process of SWT (because of https://github.com/eclipse-platform/eclipse.platform.swt/issues/490), so please correct me if I'm wrong or missing something.

One thing that makes it IMHO complex is the usage of multiple different ant-scripts that copy sources multiple times, back and fourth and that it looks like that the java-code is compiled multiple times and that besides the standard main- and source-jar there is also a zip produced for each platform-specific fragment.
The zip seems to be the one that is presented in the SWT section of the I-build drops:
https://download.eclipse.org/eclipse/downloads/drops4/I20221209-1800/
Is it correct that the `o.e.swt` and platform specific swt fragments that end up in the eclipse-updates repos (like https://download.eclipse.org/eclipse/updates/4.27-I-builds/) are the regular main- and sources-jar of the corresponding maven project?

I wonder why the code is compiled twice? And from my understanding of the build-process, as described above, this means that the class-files in the jars that end up in the p2-repo are different ones than those presented as swt-zips in the drops.

In general I think that it would be good if the swt.binaries build would be more maven oriented and I think it could be just like:
1. For each platform-specific fragment copy the corresponding java source-files from o.e.swt to the src folder
2. Build each fragment the usual maven way
3. Assemble the swt-*.zip for each fragment, but without recompiling the java-files.

This would also include that all ant based launch configs or procedures are replaced by ones that include executing a Maven build.

A second bigger point that could be improved is where/when the native-libs in each platform-specific fragment are build.
At the moment this happens in https://ci.eclipse.org/releng/job/SWT-Increment_if_needed, which is triggered in the beginning of the daily releng/I-build job.
[...]

PR #514, opened 2022-12-27
Perform SWT-natives build as part of the master/verification build

HannesWell:
First part of https://github.com/eclipse-platform/eclipse.platform.swt/issues/513

The build procedure is based on the configurations in https://ci.eclipse.org/releng/view/SWT%20Natives/
Namely these are:
- https://ci.eclipse.org/releng/job/SWT-Increment_if_needed/configure
- https://ci.eclipse.org/releng/job/cocoa_aarch64/configure
- https://ci.eclipse.org/releng/job/cocoa_x86_64/configure
- https://ci.eclipse.org/releng/job/gtk_linux_aarch64/configure
- https://ci.eclipse.org/releng/job/gtk_linux_ppc64le/configure
- https://ci.eclipse.org/releng/job/gtk_linux_x86_64/configure
- https://ci.eclipse.org/releng/job/win32_x86_64/configure

The steps of all those jobs are now perform in the Jenkins-build of this o.e.swt repo and thus makes all the mentioned jobs obsolete. This also makes building the swt-binaries as part of the I-builds obsolete. It has also the advantage that one can get immediate feedback in verification-builds for changes that affect the SWT-binaries.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,124 +1,303 @@
-pipeline {
-	options {
-		skipDefaultCheckout() // Specialiced checkout is performed below
-		timestamps()
-		timeout(time: 90, unit: 'MINUTES')
-		buildDiscarder(logRotator(numToKeepStr:'5'))
-		disableConcurrentBuilds(abortPrevious: true)
-	}
-	agent {
-		kubernetes {
-			label 'swtbuild-pod'
-			defaultContainer 'container'
-			yaml """
+def nativeBuildAgent(String platform, Closure body) {
+	def final nativeBuildStageName = 'Build SWT-native binaries'
+	if(platform == 'gtk.linux.x86_64') {
+		return podTemplate(yaml: '''
 apiVersion: v1
 kind: Pod
 spec:
   containers:
-  - name: jnlp
+  - name: "swtbuild"
+    image: "eclipse/platformreleng-centos-swt-build:8"
+    imagePullPolicy: "Always"
     resources:
+      limits:
+        memory: "4096Mi"
+        cpu: "2000m"
       requests:
         memory: "512Mi"
-        cpu: "100m"
-      limits:
-        memory: "512Mi"
-        cpu: "500m"
-  - name: container
-    image: akurtakov/swtbuild@sha256:43085feb91b1703e019a282188d996607385dcd746a7441cb7d2ca453a0adcc9
+        cpu: "1000m"
+    command:
+    - cat
     tty: true
-    command: [ "uid_entrypoint", "cat" ]
-    resources:
-      requests:
-        memory: "4Gi"
-        cpu: "1"
-      limits:
-        memory: "4Gi"
-        cpu: "1"
     volumeMounts:
-    - name: "settings-xml"
-      mountPath: "/home/jenkins/.m2/settings.xml"
-      subPath: "settings.xml"
-      readOnly: true
-    - name: toolchains-xml
-      mountPath: /home/jenkins/.m2/toolchains.xml
-      subPath: toolchains.xml
-      readOnly: true
-    - name: "settings-security-xml"
-      mountPath: "/home/jenkins/.m2/settings-security.xml"
-      subPath: "settings-security.xml"
-      readOnly: true
-    - name: m2-repo
-      mountPath: /home/jenkins/.m2/repository
-    - name: "tools"
-      mountPath: "/opt/tools"
+    - name: tools
+      mountPath: /opt/tools
   volumes:
-  - name: settings-xml
-    secret:
-      secretName: m2-secret-dir
-      items:
-      - key: settings.xml
-        path: settings.xml
-  - name: toolchains-xml
-    configMap:
-      name: m2-dir
-      items:
-      - key: toolchains.xml
-        path: toolchains.xml
-  - name: settings-security-xml
-    secret:
-      secretName: m2-secret-dir
-      items:
-      - key: settings-security.xml
-        path: settings-security.xml
-  - name: m2-repo
-    emptyDir: {}
   - name: tools
     persistentVolumeClaim:
       claimName: tools-claim-jiro-releng
-"""
-		}
+''') { node(POD_LABEL) { stage(nativeBuildStageName) { container('swtbuild') { body() } } } }
+	} else {
+		return node('swt.natives-' + platform) { stage(nativeBuildStageName) { body() } }
+	}
+}
+
+pipeline {
+	options {
+		skipDefaultCheckout() // Specialiced checkout is performed below
+		timestamps()
+		timeout(time: 180, unit: 'MINUTES')
+		buildDiscarder(logRotator(numToKeepStr:'5'))
+		disableConcurrentBuilds(abortPrevious: true)
+	}
+	agent {
+		label 'centos-latest'
 	}
 	environment {
 		MAVEN_OPTS = "-Xmx4G"
 		PR_VALIDATION_BUILD = "true"
 	}
+	parameters {
+		booleanParam(name: 'forceNativeBuilds', defaultValue: false, description: 'Forces to run the native builds of swt\'s binaries. Useful in debugging.')
+		booleanParam(name: 'skipCommit', defaultValue: false, description: 'Stops committing to swt and swt binaries repo at the end. Useful in debugging.')
+	}
 	stages {
-		stage('Prepare environment') {
+		stage('Checkout swt git repos') {
 			steps {
-				container('container') {
-					dir ('eclipse.platform.swt') {
-						checkout scm
+				dir ('eclipse.platform.swt') {
+					checkout scm
+					script {
+						def authorMail = sh(script: 'git log -1 --pretty=format:"%ce" HEAD', returnStdout: true)
+						echo 'HEAD commit author: ' + authorMail
+						if ('eclipse-releng-bot@eclipse.org'.equals(authorMail)) {
+							// Prevent endless build-loops due to self triggering because of a previous automated build of SWT-natives and the associated updates.
+							currentBuild.result = 'ABORTED'
+							error('Abort build only triggered by automated SWT-natives update.')
+						}
 					}
-					dir ('eclipse.platform.swt.binaries') {
-						checkout([$class: 'GitSCM', branches: [[name: '*/master']],
-							extensions: [[$class: 'CloneOption', timeout: 120, depth: 1, shallow: true]],
-							userRemoteConfigs: [[url: 'https://github.com/eclipse-platform/eclipse.platform.swt.binaries.git']]
-						])
+					sh '''
+						git fetch --all --tags --quiet
+						git remote set-url --push origin git@github.com:eclipse-platform/eclipse.platform.swt.git
+					'''
+				}
+				dir ('eclipse.platform.swt.binaries') {
+					checkout([$class: 'GitSCM', branches: [[name: 'refs/heads/master']],
+						extensions: [[$class: 'CloneOption', timeout: 120, noTags: false ]],
+						userRemoteConfigs: [[url: 'https://github.com/eclipse-platform/eclipse.platform.swt.binaries.git']]
+					])
+					sh 'git remote set-url --push origin git@github.com:eclipse-platform/eclipse.platform.swt.binaries.git'
+				}
+			}
+		}
+		stage('Check if SWT-binaries build is needed') {
+			steps {
+				withAnt(installation: 'apache-ant-latest', jdk: 'openjdk-jdk11-latest') { // nashorn javascript-engine required in ant-scripts
+					sh'''
+						java -version
+						git config --global user.email 'eclipse-releng-bot@eclipse.org'
+						git config --global user.name 'Eclipse Releng Bot'
+						
+						ant -f eclipse.platform.swt/bundles/org.eclipse.swt/buildSWT.xml check_compilation -DTAG=HEAD
+						ant -f eclipse.platform.swt/bundles/org.eclipse.swt/buildSWT.xml new_build_with_create_file -DTAG=HEAD
+					'''
+				}
+			}
+		}
+		stage('Build SWT-binaries, if needed') {
+			when {
+				anyOf {
+					expression { return params.forceNativeBuilds }
+					expression { return fileExists ('tmp/build_changed.txt') && fileExists ('tmp/natives_changed.txt') }
+				}
+			}
+			matrix {
+				axes {
+					axis {
+						name 'PLATFORM'
+						values 'cocoa.macosx.aarch64' , 'cocoa.macosx.x86_64', 'gtk.linux.aarch64', 'gtk.linux.ppc64le', 'gtk.linux.x86_64', 'win32.win32.x86_64'
+					}
+				}
+				stages {
+					stage("Collect SWT-native's sources") {
+						steps {
+							dir('eclipse.platform.swt.binaries/bundles'){
+								withAnt(installation: 'apache-ant-latest', jdk: 'openjdk-jdk11-latest') { // nashorn javascript-engine required in ant-scripts
+									sh '''
+										pfSpec=(${PLATFORM//"."/ })
+										ant -f binaries-parent/build.xml copy_library_src_and_create_zip -Dws=${pfSpec[0]} -Dos=${pfSpec[1]} -Darch=${pfSpec[2]}
+									'''
+								}
+								dir("org.eclipse.swt.${PLATFORM}/tmpdir") {
+									stash name:"swt.binaries.sources.${PLATFORM}",	includes: "org.eclipse.swt.${PLATFORM}.master.zip"
+								}
+							}
+						}
+					}
+					stage('Build SWT-natives') {
+						options {
+							timeout(time: 120, unit: 'MINUTES') // Some build agents are rare and it might take awhile until they are available.
+						}
+						steps {
+							script {
+								nativeBuildAgent("${PLATFORM}") {
+									cleanWs() // Workspace is not cleaned up by default, so we do it explicitly
+									unstash "swt.binaries.sources.${PLATFORM}"
+									
+									def (ws, os, arch) = env.PLATFORM.split('\\.')
+									echo "OS: ${os}"
+									echo "ARCH: ${arch}"
+									def javaHome = env.JAVA_HOME
+									// Some of the native-build agents don't have their JAVA_HOME properly set. Actually that should be done in the agents
+									//TODO: ask the infra-team to fix the setup. This is a infra-setup detail and should not be handled in the pipeline.
+									if(os =='linux' && arch == 'aarch64'){
+										def armJDK = '/usr/lib/jvm/java-11-openjdk-arm64'
+										javaHome = fileExists(armJDK) ? armJDK : '/usr/lib/jvm/java-11-openjdk'
+									} else if (os =='linux' && arch == 'ppc64le') {
+										javaHome = '/usr/lib/jvm/java-11-openjdk-11.0.15.0.10-3.el8.ppc64le'
+									} else if (os =='linux' && arch == 'x86_64') {
+										javaHome = tool(type:'jdk', name:'temurin-jdk11-latest')
+									} else if(os == 'macosx') {
+										javaHome = tool(type:'jdk', name:'temurin-jdk11-latest')
+									}
+									echo 'JAVA_HOME: ' + javaHome
+									// TODO: don't zip the sources and just (un)stash them unzipped! That safes the unzipping and removal of the the zip
+									withEnv(['MODEL=' + arch, "OUTPUT_DIR=${WORKSPACE}/libs", 'JAVA_HOME=' + javaHome]) {
+										if(isUnix()){
+											sh '''
+												unzip -aa org.eclipse.swt.${PLATFORM}.master.zip
+												rm org.eclipse.swt.${PLATFORM}.master.zip
+												mkdir libs
+												
+												#TODO: unify build script arguments?!
+												if [[ ${PLATFORM} == cocoa.macosx.* ]]; then
+													sw_vers -productVersion
+													sh build.sh install
+												elif [[ ${PLATFORM} == gtk.linux.* ]]; then
+													sh build.sh clean install
+												fi
+												ls -1R libs
+											'''
+										} else {
+											withEnv(['PATH=C:\\tools\\cygwin\\bin;' + env.PATH]) {
+												bat '''
+													unzip org.eclipse.swt.%PLATFORM%.master.zip
+													rm org.eclipse.swt.%PLATFORM%.master.zip
+													mkdir libs
+													
+													@rem rustup show
+													cmd /c build.bat x86_64 all install
+													ls -1R libs
+												'''
+											}
+										} 
+									}
+									dir('libs') {
+										stash "swt.binaries.${PLATFORM}"
+									}
+								}
+							}
+						}
+					}
+					stage('Collect and sign binaries') {
+						steps {
+							dir("libs/${PLATFORM}") {
+								unstash "swt.binaries.${PLATFORM}"
+								
+								sh '''#!/bin/bash -x
+									fn-sign-files()
+									{
+										extension=$1
+										signingServiceAddress=$2
+										if [[ "${BRANCH_NAME}" == master ]] || [[ "${BRANCH_NAME}" =~ R[0-9]+_[0-9]+(_[0-9]+)?_maintenance ]]; then
+											for filename in $(ls *.${extension})
+											do
+												mv ${filename} unsigned-${filename}
+												curl -f -o ${filename} -F file=@unsigned-${filename} $signingServiceAddress
+												if [[ $? != 0 ]]; then
+													echo "Signing of ${filename} failed"
+													exit 1
+												else
+													rm unsigned-${filename}
+												fi
+											done
+										fi
+									}
+									
+									binaryFragmentsRoot=${WORKSPACE}/eclipse.platform.swt.binaries/bundles
+									
+									if [[ ${PLATFORM} == cocoa.macosx.* ]]; then
+										#TODO: Instead use (with adjusted URL): https://github.com/eclipse-cbi/org.eclipse.cbi/tree/main/maven-plugins/eclipse-winsigner-plugin
+										# The mac-signer mojo only works for Mac-applications and not for jnilib files, but the winsigner-plugin with adjusted URL seems to do what we want.
+										echo "Sign ${PLATFORM} libraries"
+										fn-sign-files jnilib https://cbi.eclipse.org/macos/codesign/sign
+		
+										cp *.jnilib ${binaryFragmentsRoot}/org.eclipse.swt.${PLATFORM}/
+									
+									elif [[ ${PLATFORM} == gtk.linux.* ]]; then
+										#TODO: can the webkit handling be removed?!
+										#echo "Removing existing webkitextensions"
+										#rm -r ${binaryFragmentsRoot}/org.eclipse.swt.${PLATFORM}/webkit*/
+										#cp -r webkitextensions* ${binaryFragmentsRoot}/org.eclipse.swt.${PLATFORM}/
+										
+										cp *.so ${binaryFragmentsRoot}/org.eclipse.swt.${PLATFORM}/
+									
+									elif [[ ${PLATFORM} == win32.win32.* ]]; then
+										#TODO: Instead use: https://github.com/eclipse-cbi/org.eclipse.cbi/tree/main/maven-plugins/eclipse-winsigner-plugin
+										echo "Sign ${PLATFORM} libraries"
+										fn-sign-files dll https://cbi.eclipse.org/authenticode/sign
+										
+										cp *.dll ${binaryFragmentsRoot}/org.eclipse.swt.${PLATFORM}/
+									fi
+								'''
+							}
+						}
 					}
 				}
 			}
 		}
-		stage('Build') {
+		stage('Commit SWT-native binaries, if build') {
+			when {
+				expression { return params.forceNativeBuilds || fileExists ('tmp/build_changed.txt') }
+			}
 			steps {
-				container('container') {
-					wrap([$class: 'Xvnc', useXauthority: true]) {
-						withEnv(["JAVA_HOME=${ tool 'openjdk-jdk17-latest' }"]) {
-							dir ('eclipse.platform.swt.binaries') {
-								sh '''
-									/opt/tools/apache-maven/latest/bin/mvn install \
-										--batch-mode -Pbuild-individual-bundles -DforceContextQualifier=zzz -Dnative=gtk.linux.x86_64 \
-										-Dcompare-version-with-baselines.skip=true -Dmaven.compiler.failOnWarning=true
-								'''
-							}
-							dir ('eclipse.platform.swt') {
-								sh '''
-									/opt/tools/apache-maven/latest/bin/mvn clean verify \
-										--batch-mode -Pbuild-individual-bundles -DcheckAllWS=true -DforkCount=0 \
-										-Dcompare-version-with-baselines.skip=false -Dmaven.compiler.failOnWarning=true \
-										-Dmaven.test.failure.ignore=true -Dmaven.test.error.ignore=true
-								'''
-							}
-						}
+				withAnt(installation: 'apache-ant-latest', jdk: 'openjdk-jdk11-latest') { // nashorn javascript-engine required in ant-scripts
+					//The maven build reads the git-history so we should have to commit the native-binaries before building
+					sh '''
+						pushd eclipse.platform.swt.binaries
+						git add --all *
+						echo "git status after add"
+						git status
+						popd
+
+						ant -f eclipse.platform.swt/bundles/org.eclipse.swt/buildInternal.xml write_qualifier -Dlib.dir=${WORKSPACE} -Dbuild_changed=true
+						ant -f eclipse.platform.swt/bundles/org.eclipse.swt/buildSWT.xml commit_poms_and_binaries
+						ant -f eclipse.platform.swt/bundles/org.eclipse.swt/buildSWT.xml tag_projects
+	
+						pushd eclipse.platform.swt
+						git status
+						git log -p -2
+						popd
+						
+						pushd eclipse.platform.swt.binaries
+						git status
+						git log -p -1
+						popd
+					'''
+				}
+			}	
+		}
+		stage('Build') {
+			tools {
+				// Define tools only in this stage to not interfere with default environemts of SWT-natives build-agents
+				jdk 'openjdk-jdk17-latest'
+				maven 'apache-maven-latest'
+			}
+			steps {
+				xvnc(useXauthority: true) {
+					dir ('eclipse.platform.swt.binaries') {
+						sh '''
+							mvn install \
+								--batch-mode -Pbuild-individual-bundles -DforceContextQualifier=zzz \
+								-Dcompare-version-with-baselines.skip=true -Dmaven.compiler.failOnWarning=true
+						'''
+					}
+					dir ('eclipse.platform.swt') {
+						sh '''
+							mvn clean verify \
+								--batch-mode -Pbuild-individual-bundles -DcheckAllWS=true -DforkCount=0 \
+								-Dcompare-version-with-baselines.skip=false -Dmaven.compiler.failOnWarning=true \
+								-Dmaven.test.failure.ignore=true -Dmaven.test.error.ignore=true
+						'''
 					}
 				}
 			}
@@ -130,6 +309,46 @@ spec:
 					// discoverGitReferenceBuild referenceJob: 'eclipse.platform.swt/master'
 					// recordIssues publishAllIssues: true, tools: [java(), mavenConsole(), javaDoc()]
 					publishIssues issues:[scanForIssues(tool: java()), scanForIssues(tool: mavenConsole())]
+				}
+			}
+		}
+		stage('Push SWT-native binaries, if build') {
+			when {
+				expression { return params.forceNativeBuilds || fileExists ('tmp/build_changed.txt') }
+			}
+			steps {
+				sshagent(['github-bot-ssh']) {
+					withAnt(installation: 'apache-ant-latest', jdk: 'openjdk-jdk11-latest') { // nashorn javascript-engine required in ant-scripts
+						sh '''
+							# Check for the master-branch as late as possible to have as much of the same behaviour as possible
+							if [[ "${BRANCH_NAME}" == master ]] || [[ "${BRANCH_NAME}" =~ R[0-9]+_[0-9]+(_[0-9]+)?_maintenance ]]; then
+								if [[ ${skipCommit} != true ]]; then
+									
+									pushd eclipse.platform.swt
+									git fetch origin refs/heads/${BRANCH_NAME}:refs/remotes/origin/${BRANCH_NAME}
+									popd
+									
+									pushd eclipse.platform.swt.binaries
+									git fetch origin refs/heads/${BRANCH_NAME}:refs/remotes/origin/${BRANCH_NAME}
+									popd
+									
+									ant -f eclipse.platform.swt/bundles/org.eclipse.swt/buildSWT.xml push_remote_from_workspace
+									exit 0
+								else
+									echo Committing is skipped
+								fi
+							else
+								echo Skip pushing changes of native-binaries for branch "${BRANCH_NAME}"
+							fi
+							# The commits are not pushed. At least list them, so one can check if the result is as expected.
+							pushd eclipse.platform.swt
+							git log -n 2
+							popd
+							pushd eclipse.platform.swt.binaries
+							git log -n 2
+							popd
+						'''
+					}
 				}
 			}
 		}
--- a/bundles/org.eclipse.swt/buildSWT.xml
+++ b/bundles/org.eclipse.swt/buildSWT.xml
@@ -974,7 +974,7 @@
 			<arg line="rebase origin/${TAG}"/>
 		</exec>
 		<exec dir="${repo.src}" executable="git" failonerror="true">
-			<arg line="push origin ${TAG}"/>
+			<arg line="push origin HEAD:refs/heads/${TAG}"/>
 		</exec>
 		<exec dir="${repo.bin}" executable="git" failonerror="true">
 			<arg line="fetch"/>
@@ -986,7 +986,7 @@
 			<arg line="rebase origin/${TAG}"/>
 		</exec>
 		<exec dir="${repo.bin}" executable="git" failonerror="true">
-			<arg line="push origin ${TAG}"/>
+			<arg line="push origin HEAD:refs/heads/${TAG}"/>
 		</exec>
 		<exec dir="${repo.src}" executable="git" failonerror="true">
 			<arg line="push origin refs/tags/${swt_tag}"/>


HannesWell:
On thing that needs to be done to make this work is making the specialized nodes in the releng-Jenkins (https://ci.eclipse.org/releng/computer/) available to the `eclipse.platform.swt` job of the platform-Jenkins (https://ci.eclipse.org/platform/job/eclipse.platform.swt/).
@akurtakov I assume I can request that from the EFN infra-team via helps-desk?

HannesWell:
> On thing that needs to be done to make this work is making the specialized nodes in the releng-Jenkins (https://ci.eclipse.org/releng/computer/) available to the `eclipse.platform.swt` job of the platform-Jenkins (https://ci.eclipse.org/platform/job/eclipse.platform.swt/). @akurtakov I assume I can request that from the EFN infra-team via helps-desk?

Just created https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/2448

HannesWell:
Another adjustment of the Jenkins infra that probably will allow further simplifications is to assign better labels to Jenkins nodes prepared to build the SWT native binaries. For example
- `swt.natives-cocoa.macosx.aarch64` for node capable SWT natives for cocoa.macosx.aarch64
- `swt.natives-cocoa.macosx.x86_64` for node capable SWT natives for cocoa.macosx.x86_64
- `swt.natives-gtk.linux.aarch64` for node capable SWT natives for gtk.linux.aarch64
- `swt.natives-gtk.linux.ppc64le` for node capable SWT natives for gtk.linux.ppc64le
- `swt.natives-gtk.linux.x86_64` for node capable SWT natives for gtk.linux.x86_64
- `swt.natives-win32.win32.x86_64` for node capable SWT natives for win32.win32.x86_64

This would (probably) allow to convert the currently parallel stage to build the SWT natives into a matrix step.
I tried to use a groovy map that maps the platform spec to a node label, but it looks like that does not work. But what should work is to use `agent { label 'swt.natives-${PLATFORM}' }`, where `PLATFORM` is the matrix axis that has all supported platforms as argument (as listed above).

I plan to request that from the infra team, once https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/2448 is resolved and we have a working (parallel) pipeline.

akurtakov:
I wonder can't we make it the other way - build the natives on master change on releng JIPP. Reason for that is that some of the entries like ppc/arm/win/mac machines are very low resource and are ment to be used only for native rebuilds and thus it would be best if they are not generally available in Platform JIPP. We should be able to do the very same build on Releng JIPP.

HannesWell:
> I wonder can't we make it the other way - build the natives on master change on releng JIPP.

I don't think it is possible to refer to agents on a different JIPP, but please correct me if I'm wrong. Or do you mean something else?
For the sake of simplicity, comprehensibility and faster feedback it would be great to define the natives build in the same pipeline as it is drafted here. The current approach has the advantage that one gets feedback about changes in natives even in verification builds and that all uses of the ANT task are specified in a file in this or the swt.binaries repo. I already identified some unused task, but that is harder if you have to check multiple pipelines/jobs.

But what should work technically is to move the entire `o.e.swt` project from the platform JIPP to the releng JIPP.

> Reason for that is that some of the entries like ppc/arm/win/mac machines are very low resource and are ment to be used only for native rebuilds and thus it would be best if they are not generally available in Platform JIPP. We should be able to do the very same build on Releng JIPP.

Fully understand and agree with that. Maybe it is possible to make nodes/computers only available to certain projects in a JIPP?
Otherwise we have to make sure that no pipeline uses `agent any`. I checked a few and it looks like `centos-latest` is usually used as agent label. But if we go with the currently suggested approach, I can check that all projects in the platform JIPP use that.

mickaelistria:
I believe it would be more sustainable to reduce the amount of commands in Jenkinsfile and move more of those into pom.xml to more easily enable local builds, or builds on other CIs (eg GitHub).
Could the commands that build the binary be entirely defined in the SWT fragment that's supposed to embed the library?

HannesWell:
> I believe it would be more sustainable to reduce the amount of commands in Jenkinsfile and move more of those into pom.xml to more easily enable local builds, or builds on other CIs (eg GitHub). Could the commands that build the binary be entirely defined in the SWT fragment that's supposed to embed the library?

This is just the very first step of #513 to have all the build steps that are currently defined in those seven releng Jenkins jobs https://ci.eclipse.org/releng/view/SWT%20Natives/ defined into this repo.
This would avoids to trigger them as (and consequently build the SWT-binaries) only in the beginning of the I-Build:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/c2c0b2e1f87c239d1d784d6c6094f85dd7401453/JenkinsJobs/Builds/I_build.groovy#L232-L236

Once all the steps to build the SWT-binaries are carried out in this repo it will be much simpler to implement further improvements and more 'mavenization' in subsequent changes.

Furthermore this could also simplify back-porting. At the moment I have no clue how changes to the natives are back-ported since the jobs in https://ci.eclipse.org/releng/view/SWT%20Natives/ always take the latest artifacts.

akurtakov:
Both Fred and me think it's best to run natives build on releng jipp for already discussed reasons. Are you interested in looking that path?

HannesWell:
> Both Fred and me think it's best to run natives build on releng jipp for already discussed reasons. Are you interested in looking that path?

Isn't my response (don't use `agent any`) to your concern (usage of limited resources by other jobs that don't need it) a valid solution?
For [Fred's concern](https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/2448#note_1059376) I have no immediate solution.
I believe it is mainly about duplicated configurations, but @fredg02 could you please explain the reason why you prefer to not duplicate the agents? Maybe there are some smart ways to handle this in https://github.com/eclipse-cbi/jiro/tree/master/instances

Can tell me how to trigger builds on another Jenkins instance or even better how to use agents from another instance? Then I can of course try that. And as said in https://github.com/eclipse-platform/eclipse.platform.swt/issues/513#issuecomment-1370533863, another option would be to move the [eclipse.platform.swt Jenkins project](https://ci.eclipse.org/platform/job/eclipse.platform.swt/) to the [releng Jenkins](https://ci.eclipse.org/releng/).

Because maintaining the binary build steps somewhere else makes the build IMHO very hard to understand. I spend a significant amount of time to figuring out and understand how things are working together and the presented approach is, in my opinion, a significant reduction in complexity (~200 lines in one pipeline vs 7 additional Jobs defined somewhere else in another Jenkins instance).

fredg02:
> For [Fred's concern](https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/2448#note_1059376) I have no immediate solution.
I believe it is mainly about duplicated configurations, but @fredg02 could you please explain the reason why you prefer to not duplicate the agents?

Mostly duplication. There is also a potential issue because two agents from different JIPPs do not share a build queue. So build times can increase significantly if two builds (one from JIPP A and one from JIPP B) run at the same time.

> Maybe there are some smart ways to handle this in https://github.com/eclipse-cbi/jiro/tree/master/instances

Nope. Jiro does not support making the duplication of build agents easier. :)

> Can you tell me how to trigger builds on another Jenkins instance or even better how to use agents from another instance?

You can trigger builds from another Jenkins instance by using the "remote build trigger" functionality (here is a tutorial explaining it for freestyle builds: https://www.middlewareinventory.com/blog/jenkins-remote-build-trigger-url/). For Jenkins pipelines the keyword is `remoteBuildTrigger` (see https://www.jenkins.io/doc/pipeline/steps/params/pipelinetriggers/). It requires user authentication, so a few extra steps are involved.

You cannot (easily) use agents from another instance unless you duplicate the agents.

> Then I can of course try that. And as said in https://github.com/eclipse-platform/eclipse.platform.swt/issues/513#issuecomment-1370533863, another option would be to move the [eclipse.platform.swt Jenkins project](https://ci.eclipse.org/platform/job/eclipse.platform.swt/) to the [releng Jenkins](https://ci.eclipse.org/releng/).

Moving the builds to the releng JIPP is recommended.

> Because maintaining the binary build steps somewhere else makes the build IMHO very hard to understand. I spend a significant amount of time to figuring out and understand how things are working together and the presented approach is, in my opinion, a significant reduction in complexity (~200 lines in one pipeline vs 7 additional Jobs defined somewhere else in another Jenkins instance).

I agree that the build is complicated and the releng JIPP handling builds for/from other projects is not straight forward.
I'd recommend to improve the documentation to make it easier to find the right build jobs and describe the reasoning behind it.

HannesWell:
Thank you for elaborating.

> > Then I can of course try that. And as said in [#513 (comment)](https://github.com/eclipse-platform/eclipse.platform.swt/issues/513#issuecomment-1370533863), another option would be to move the [eclipse.platform.swt Jenkins project](https://ci.eclipse.org/platform/job/eclipse.platform.swt/) to the [releng Jenkins](https://ci.eclipse.org/releng/).
> 
> Moving the builds to the releng JIPP is recommended.

Yes, after all that looks like the way with full benefits but the fewest extra work.

> 
> > Because maintaining the binary build steps somewhere else makes the build IMHO very hard to understand. I spend a significant amount of time to figuring out and understand how things are working together and the presented approach is, in my opinion, a significant reduction in complexity (~200 lines in one pipeline vs 7 additional Jobs defined somewhere else in another Jenkins instance).
> 
> I agree that the build is complicated and the releng JIPP handling builds for/from other projects is not straight forward. I'd recommend to improve the documentation to make it easier to find the right build jobs and describe the reasoning behind it.

With this approach everything about building the SWT-binaries is specified in this repo in the Jenkinsfile and Maven/ANT files and I have the intention the further simplify that subsequently. This 'Code' is IMHO the best documentation that never goes out of sync, given it is not too complicated to understand.
But in the swt.binaries repo there could be README that shortly summaries where the binaries are coming from (at least as long that repo is in use).

Jenkinsfile
```diff
@@ -7,128 +7,285 @@ pipeline {
...
+						}
+					}
+				}
+			}
 			post {
 				always {
 					junit 'eclipse.platform.swt/tests/*.test*/target/surefire-reports/*.xml'
 					archiveArtifacts artifacts: '**/*.log,**/*.html,**/target/*.jar,**/target/*.zip'
-					discoverGitReferenceBuild referenceJob: 'eclipse.platform.swt/master'
-					recordIssues publishAllIssues: true, tools: [java(), mavenConsole(), javaDoc()]
```
HannesWell, line 129: This was added in #521, but the releng JIPPs seems to not have the corresponding plug-ins.
Eventually they should be installed so that can be used again.

HannesWell:
Finally a first success, for those platforms for which native-build agents are available (windows and linux aarch64 and ppc), the full pipeline seems to work.
Once the mac agent (https://ci.eclipse.org/releng/computer/b9h15%2Dmacos10.15/) is back, I'll try that as well. Then only the agent for linux-x86_64 is missing to complete this: https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/2448#note_1059781

mickaelistria:
I don't plan to review it in details. Is there anything particular you'd like me to look at?

HannesWell:
> I don't plan to review it in details. Is there anything particular you'd like me to look at?

Nothing particular, but of course a general review is welcome. 

My main problem at the moment is that I'm waiting for the mac-agent to be revived and a linux-x86_64 to be provisioned (See my previous comment). It would we great, if you could help with that.

HannesWell:
> Once the mac agent (https://ci.eclipse.org/releng/computer/b9h15%2Dmacos10.15/) is back, I'll try that as well. Then only the agent for linux-x86_64 is missing to complete this: https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/2448#note_1059781

@fredg02, @akurtakov, @mickaelistria could you or anybody please review and/or help with https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/2448#note_1059781 respectively https://github.com/eclipse-cbi/jiro/pull/261?
That and the offline mac-agent are blocks for this PR and I'm waiting now for two weeks to get them resolved. I tried to do as much as I can do to resolve them by myself, but I need help in this regard.

I totally understand that everybody is busy with other more important things, but it is unpleasant to be blocked for that long. For me personally this change will probably have almost zero benefit and I only did out of curiosity and because I saw this opportunity to contribute an improvement that hopefully makes the life of swt-natives developers (because of fast feedback in verification builds) and of the releng/infra-team (because of seven non-trivial Jenkins builds less) easier.
I don't want to complain, but I want to ask kindly to increase your priority for this change or tell me what to do in order to help you helping me or at least give me a rough estimation when you will have the time to resolve the mentioned issues so that I can put this on hold and do other things in the meantime.

akurtakov:
Only @fredg02 can help with the infra changes. Everyone else involved in this ticket can do as much as you on the topic.

HannesWell:
> Only @fredg02 can help with the infra changes. Everyone else involved in this ticket can do as much as you on the topic.

Yes. Could you please have a look at https://github.com/eclipse-cbi/jiro/pull/261 and tell if you think the configuration of the additional kubernetes-agent configuration is fine? From the other discussions I got the impression that you know quite well how the machine that builds the swt-natives has to be set-up, which I don't.

Jenkinsfile
```diff
@@ -7,120 +7,272 @@ pipeline {
...
+						}
+					}
+					stage('Build SWT-native binaries') {
+						options {
+							skipDefaultCheckout()
+							timestamps()
+							timeout(time: 15, unit: 'MINUTES')
+						}
+						agent {
+							label "swt.natives-${PLATFORM}"
```
mickaelistria, line 87: Couldn't we have at the beginning of the file a map PLATFORM->agent with defined agents such as
```groovy
def agents = [ "gtk.linux.aarch64": agent { label "swt.natives.gtk.linux.aarch64" },
  ...
  "gtk.linux.x86_64": agent { label "centos-8" }]
```
and then let `agent` here just resolve the map with something like `agent[${PLATFORM}]`?

That would allow to have heterogeneous way of defining agents and to rely less on Eclipse infra convention. Later on, we could even consider replacing some agents defined with labels with some defined with a kubernetes config, typically if we want to use another linux before Eclipse infra sets it up.

fredg02:
> That and the offline mac-agent are blocks for this PR and I'm waiting now for two weeks to get them resolved. I tried to do as much as I can do to resolve them by myself, but I need help in this regard.

I was on vacation last week. I will try to revive the Mac agent and review your PR this week.

Jenkinsfile
HannesWell: I tried that but failed to do it in a declarative pipeline, with errors like:
`WorkflowScript: 20: Method calls on objects not allowed outside "script" blocks.`
Therefore I think it is not possible (by design) to have such dynamic agents in a declarative pipeline.

But after thinking about this again and as the error message actually suggests, I had the idea to create the native-builds `node`s in scripted-pipeline style in the `script`-step I'm using for the build native stage anyway. In scripted style this should work.
This probably also enables to execute the third-stage of the matrix (to collect the build native-binaries) on the initial/pipeline agent again the (as I asked here https://stackoverflow.com/q/75201394/14542697).


> Later on, we could even consider replacing some agents defined with labels with some defined with a kubernetes config, typically if we want to use another linux before Eclipse infra sets it up.

In general I'm wondering if it is better to specify the infrastructure in the pipeline in this repo or if it's better to define it externally (e.g. in the jiro repo)? Initially I tended to the latter answer because external changes are often not that relevant (e.g. if a new version of cent-os is used or if the JIBB is restructured), but on the other hand it also allows specify more of the exact build environment used to build the natives, which is also good and allows to try out new environments in a verification build (as you said).

Nevertheless I'll try the scripted heterogeneous build-node definition now and if that works we can still decide to keep it or to proceed with https://github.com/eclipse-cbi/jiro/pull/261 and just use 'labeled' agents for all platforms.

HannesWell:
> > That and the offline mac-agent are blocks for this PR and I'm waiting now for two weeks to get them resolved. I tried to do as much as I can do to resolve them by myself, but I need help in this regard.
> 
> I was on vacation last week. I will try to revive the Mac agent and review your PR this week.

No problem and I'm sorry if my previous message was a bit harsh. Thank you for your help!

Jenkinsfile
HannesWell: > Couldn't we have at the beginning of the file a map PLATFORM->agent with defined agents such as
> 
> ```groovy
> def agents = [ "gtk.linux.aarch64": agent { label "swt.natives.gtk.linux.aarch64" },
>   ...
>   "gtk.linux.x86_64": agent { label "centos-8" }]
> ```
> 
> and then let `agent` here just resolve the map with something like `agent[${PLATFORM}]`?
> 
> That would allow to have heterogeneous way of defining agents and to rely less on Eclipse infra convention. Later on, we could even consider replacing some agents defined with labels with some defined with a kubernetes config, typically if we want to use another linux before Eclipse infra sets it up.

All of that worked as hoped. 🎉 So if you prefer the current approach over https://github.com/eclipse-cbi/jiro/pull/261 it is possible. Instead of maintaining a map with closure-values I use a method that only handles the `linux.x86_64` case specially. This avoids the need to list each platform twice (once in the map and once in the matrix axis).

mickaelistria: Awesome and congrats for this! It looks like an advanced and useful solution of a usual pattern. I suspect it would be worth documenting somewhere (eg https://wiki.eclipse.org/Jenkins and/or some blog post if you use a blog).

IMO, it's better to get as much as possible of the infrastructure requirement specified in the project. It will make it easier for project to evolve its target infrastructure without having to involve EF infra foundation staff (who may not be available or busy with other things). I believe having the more infra defined in Jenkinsfile increases agility for the project.

HannesWell: > Awesome and congrats for this! It looks like an advanced and useful solution of a usual pattern. I suspect it would be worth documenting somewhere (eg https://wiki.eclipse.org/Jenkins and/or some blog post if you use a blog).

Thanks. I don't have a blog, but documenting it in https://wiki.eclipse.org/Jenkins#FAQ sounds good. Looks like I don't have write permissions there. But if the infra-team is interested (@fredg02 ?) I can provide a little text with an example and one of the infra-team can adjust it as desired and add it to the side. If one of them write an FAQ entry all by themself I'm also fine with that. :)


> 
> IMO, it's better to get as much as possible of the infrastructure requirement specified in the project. It will make it easier for project to evolve its target infrastructure without having to involve EF infra foundation staff (who may not be available or busy with other things). I believe having the more infra defined in Jenkinsfile increases agility for the project.

That's right. But then the project also has to take care, for example of security updates etc. Not saying that is per-se a blocker, I understand and agree with your point, but I think the latter is something to consider as well.

fredg02:
Mac agent is back online.... BUT I did not manage to install Ant so far (which when installed via homebrew, requires the openjdk homebrew-package, which in turn requires the xcode app for compilation (?), which can't be installed due to the old macOS version (10.15)).

HannesWell:
> Mac agent is back online.... BUT I did not manage to install Ant so far (which when installed via homebrew, requires the openjdk homebrew-package, which in turn requires the xcode app for compilation (?), which can't be installed due to the old macOS version (10.15)).

Thanks a lot!
Ant is actually not necessary at the build agents, at least for the purpose of building SWT-natives. But I would not object to update to a more recent macos version, but I'm not familiar with the requirements and can therefore not tell if it is ok to upgrade?

In the current state the build of the natives succeeds for macosx.x86_64 but fails for macos.aarch64.
On that agent, after inflating the source zip I get the following printout:

```
  + sh build.sh clean install
  10.15.5
  rm -f *.jnilib *.o
  cc -c -xobjective-c -Wall -arch arm64 -DSWT_VERSION=4958r3   -DUSE_ASSEMBLER -DCOCOA -DATOMIC -I /Library/Java/JavaVirtualMachines/temurin-19.jdk/Contents/Home/include -I /Library/Java/JavaVirtualMachines/temurin-19.jdk/Contents/Home/include/darwin -I /System/Library/Frameworks/Cocoa.framework/Headers -I /System/Library/Frameworks/JavaScriptCore.framework/Headers swt.c
  In file included from swt.c:15:
  In file included from ./swt.h:26:
  In file included from /Library/Java/JavaVirtualMachines/temurin-19.jdk/Contents/Home/include/jni.h:39:
  In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/include/stdio.h:64:
  In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/include/_stdio.h:68:
  /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/include/sys/cdefs.h:807:2: error: Unsupported architecture
  #error Unsupported architecture
   ^
  In file included from swt.c:15:
  In file included from ./swt.h:26:
  In file included from /Library/Java/JavaVirtualMachines/temurin-19.jdk/Contents/Home/include/jni.h:39:
  In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/include/stdio.h:64:
  In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/include/_stdio.h:71:
  In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/include/_types.h:27:
  In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/include/sys/_types.h:33:
  /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/include/machine/_types.h:34:2: error: architecture not supported
  #error architecture not supported
   ^
  In file included from swt.c:15:
  In file included from ./swt.h:26:
  In file included from /Library/Java/JavaVirtualMachines/temurin-19.jdk/Contents/Home/include/jni.h:39:
  In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/include/stdio.h:64:
  In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/include/_stdio.h:71:
  In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/include/_types.h:27:
  /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/include/sys/_types.h:55:9: error: unknown type name '__int64_t'
  typedef __int64_t       __darwin_blkcnt_t;      /* total blocks */
          ^
  /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/include/sys/_types.h:56:9: error: unknown type name '__int32_t'; did you mean '__int128_t'?
  typedef __int32_t       __darwin_blksize_t;     /* preferred block size */
          ^
  note: '__int128_t' declared here
  /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/include/sys/_types.h:57:9: error: unknown type name '__int32_t'; did you mean '__int128_t'?
  typedef __int32_t       __darwin_dev_t;         /* dev_t */
          ^
  note: '__int128_t' declared here
  /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/include/sys/_types.h:60:9: error: unknown type name '__uint32_t'; did you mean '__uint128_t'?
  typedef __uint32_t      __darwin_gid_t;         /* [???] process and group IDs */
          ^
  note: '__uint128_t' declared here
  /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/include/sys/_types.h:61:9: error: unknown type name '__uint32_t'; did you mean '__uint128_t'?
  typedef __uint32_t      __darwin_id_t;          /* [XSI] pid_t, uid_t, or gid_t*/
          ^
  note: '__uint128_t' declared here
```

Comparing this with the [last successful build of the existing SWT-natives job for cocoa_aarch64](https://ci.eclipse.org/releng/view/SWT%20Natives/job/cocoa_aarch64/174/console), I noticed that the return value of `sw_vers -productVersion` decreased from `10.15.7` to `10.15.5` on the current agent
and that the first c-compiler invocation now includes headers from `temurin-19.jdk` and before from `temurin-11.jdk`:
```
cc -c -xobjective-c -Wall -arch arm64 -DSWT_VERSION=4958r2   -DUSE_ASSEMBLER -DCOCOA -DATOMIC -I /Library/Java/JavaVirtualMachines/temurin-11.jdk/Contents/Home/include -I /Library/Java/JavaVirtualMachines/temurin-11.jdk/Contents/Home/include/darwin -I /System/Library/Frameworks/Cocoa.framework/Headers -I /System/Library/Frameworks/JavaScriptCore.framework/Headers swt.c
```
```
cc -c -xobjective-c -Wall -arch arm64 -DSWT_VERSION=4958r3   -DUSE_ASSEMBLER -DCOCOA -DATOMIC -I /Library/Java/JavaVirtualMachines/temurin-19.jdk/Contents/Home/include -I /Library/Java/JavaVirtualMachines/temurin-19.jdk/Contents/Home/include/darwin -I /System/Library/Frameworks/Cocoa.framework/Headers -I /System/Library/Frameworks/JavaScriptCore.framework/Headers swt.c
```
AFAICT the existing SWT-natives build job for macos.aarch64 just uses the system java-installation.

Could the older product version or the more recent jdk-headers be the cause for that error?

SyntevoAlex:
In my own project, I also noticed that `CommandLineTools` on macOS 10.15 can't build `arm64` architecture.
Most likely this problem is not related to JDK in any way.

SyntevoAlex:
When building on macOS 11, this problem doesn't occur.

SyntevoAlex:
From a random website:
> You need the MacOS SDK v11. This comes with XCode 12.2. You cannot use previous versions (e.g. v10.15) of the MacOS SDK because they do not contain the new architecture ARM64. They only contain the x86_64 architecture.

SyntevoAlex:
It is probably possible to install newer Command line tools manually, via a download like
https://download.developer.apple.com/Developer_Tools/Command_Line_Tools_for_Xcode_14/Command_Line_Tools_for_Xcode_14.dmg

SyntevoAlex:
The latest version that agrees to install on macOS 10.15 is [`Command Line Tools for Xcode 12.4`](https://download.developer.apple.com/Developer_Tools/Command_Line_Tools_for_Xcode_12.4/Command_Line_Tools_for_Xcode_12.4.dmg)

SyntevoAlex:
My experiments:

Default macOS 10.15:
```
$ gcc --version
	Apple clang version 11.0.3 (clang-1103.0.32.59)
	Target: x86_64-apple-darwin19.6.0
$ echo "#include <stdio.h>" > 1.c
$ gcc -c -arch arm64 1.c
	In file included from 1.c:1:
	In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/stdio.h:64:
	In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/_stdio.h:68:
	/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/sys/cdefs.h:807:2: error:
				Unsupported architecture
	#error Unsupported architecture
```

After installing [`Command Line Tools for Xcode 12.4`](https://download.developer.apple.com/Developer_Tools/Command_Line_Tools_for_Xcode_12.4/Command_Line_Tools_for_Xcode_12.4.dmg) 
```
$ gcc --version
	Apple clang version 12.0.0 (clang-1200.0.32.29)
	Target: x86_64-apple-darwin19.6.0
$ echo "#include <stdio.h>" > 1.c
$ gcc -c -arch arm64 1.c
	Still gives the same error.
	For some reason it still uses SDK from 10.15:
	/Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/include/stdio.h
```

Fine, let's also pass `-isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk` to compiler
```
$ gcc -c -arch arm64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk 1.c
	Succeeds
```

Conclusions: macOS 10.15 can compile arm64 with some additional setup.

HannesWell:
> After installing [`Command Line Tools for Xcode 12.4`](https://download.developer.apple.com/Developer_Tools/Command_Line_Tools_for_Xcode_12.4/Command_Line_Tools_for_Xcode_12.4.dmg)

Thank you very much for this analysis and help. Since I'm not using Mac at all it would have take me a while to find that out (and I cannot access the websites you linked).

@fredg02 could you install the mentioned `Command Line Tools for Xcode 12.4` on the mac agent?
Or @akurtakov would it be sufficient to upgrade to a more recent mac-os version that ships with the required SDK already?

akurtakov:
According to https://endoflife.date/macos 10.5 is out of support by Apple itself so moving to macos 11 should be fine.

akurtakov:
On https://ci.eclipse.org/releng/ I see two MacOS executors - https://ci.eclipse.org/releng/computer/b9h15%2Dmacos10.15/ and https://ci.eclipse.org/releng/computer/nc1ht%2Dmacos11%2Darm64/ from jobs assigned to each it looks like running tests on MacOS 10.15 is no longer a thing thus compiling on 11 sounds fine to me too.

HannesWell:
Sounds good. :)

So should [b9h15-macos10.15](https://ci.eclipse.org/releng/computer/b9h15%2Dmacos10.15/) be upgraded to macos11 and both mac-versions should still be build there?
Or should `macos.aarch64` be build on [nc1ht-macos11-arm64 ](https://ci.eclipse.org/releng/computer/nc1ht%2Dmacos11%2Darm64/) and `macos.x86_64` remains on [b9h15-macos10.15](https://ci.eclipse.org/releng/computer/b9h15%2Dmacos10.15/).
Or should both mac-versions just be build on [nc1ht-macos11-arm64](https://ci.eclipse.org/releng/computer/nc1ht%2Dmacos11%2Darm64/)? But from the label I'm not sure if this is a 'pure' arm64 machine and if that can build x86_64?

In case option 2 or 3 is the one to go, should [b9h15-macos10.15](https://ci.eclipse.org/releng/computer/b9h15%2Dmacos10.15/) still be upgraded?

SyntevoAlex:
> (and I cannot access the websites you linked).

Requires an Apple ID that can be registered for free.

> if this is a 'pure' arm64 machine and if that can build x86_64?

I'm rather confident that both x86_64 and arm64 can build both architectures.

> According to https://endoflife.date/macos 10.15 is out of support by Apple itself so moving to macos 11 should be fine

Yes if machine can be upgraded to macOS 11, then it is probably the simplest solution.

laeubi:
> Requires an Apple ID that can be registered for free.

# [The "Free" Model](https://geek-and-poke.com/geekandpoke/2010/12/21/the-free-model.html)

SyntevoAlex:
> Since I'm not using Mac at all

Happy man

akurtakov:
> > Since I'm not using Mac at all
> 
> Happy man

TBH, last time I used a Mac for more than 5 mins was ~2005 and only time I have to deal with anything Mac is for Eclipse releng. Without even knowing what it looks like it's in the group of aarch64, ppc64le weird configs Eclipse provides builds for :)

sravanlakkimsetti:
> Sounds good. :)
> 
> So should [b9h15-macos10.15](https://ci.eclipse.org/releng/computer/b9h15%2Dmacos10.15/) be upgraded to macos11 and both mac-versions should still be build there? Or should `macos.aarch64` be build on [nc1ht-macos11-arm64 ](https://ci.eclipse.org/releng/computer/nc1ht%2Dmacos11%2Darm64/) and `macos.x86_64` remains on [b9h15-macos10.15](https://ci.eclipse.org/releng/computer/b9h15%2Dmacos10.15/). Or should both mac-versions just be build on [nc1ht-macos11-arm64](https://ci.eclipse.org/releng/computer/nc1ht%2Dmacos11%2Darm64/)? But from the label I'm not sure if this is a 'pure' arm64 machine and if that can build x86_64?
> 
> In case option 2 or 3 is the one to go, should [b9h15-macos10.15](https://ci.eclipse.org/releng/computer/b9h15%2Dmacos10.15/) still be upgraded?

Actually we can build both `x86_64` and `arm64` architectures using either of the machines.

SyntevoAlex:
arm64 can't be built on default 10.15, see https://github.com/eclipse-platform/eclipse.platform.swt/pull/514#issuecomment-1407654909

sravanlakkimsetti:
> `arm64` can't be built on default 10.15, see [#514 (comment)](https://github.com/eclipse-platform/eclipse.platform.swt/pull/514#issuecomment-1407654909)

to build `arm64` you'll need Xtools 12.2. We have installed xtools 12.2 on that machine. That way we are able to build both architectures

HannesWell:
When I try to push the results of the build to a test branch using
`git push origin HEAD:refs/heads/testNativeBuildInPipeline`
the command is stuck for a while.

Is it possible that the [`releng/eclipse.platform.swt`](https://ci.eclipse.org/releng/job/eclipse.platform.swt/) needs again permissions to push to the eclipse.platform.swt repo? Or do all jobs in the releng Jenkins have that permission already?

HannesWell:
> When I try to push the results of the build to a test branch using `git push origin HEAD:refs/heads/testNativeBuildInPipeline` the command seems to be stuck forever and does nothing.
> 

That is probably because the repo is cloned using https. Setting the SSH address as pushurl will probably fix that. Will try that.

fredg02:
@sravanlakkimsetti can you confirm that the `arm64` build worked on `b9h15-macos10.15` before it went AWOL?

iloveeclipse:
Is the fail here (on innocent change in SWT Linux code) related?
https://ci.eclipse.org/releng/job/eclipse.platform.swt/job/PR-537/3/console

```
11:32:47  [ERROR] Failed to execute goal org.apache.maven.plugins:maven-antrun-plugin:3.1.0:run (normal) on project org.eclipse.swt.cocoa.macosx.x86_64: An Ant BuildException has occured: The following error occurred while executing this line:
11:32:47  [ERROR] /home/jenkins/agent/workspace/eclipse.platform.swt_PR-537/eclipse.platform.swt.binaries/bundles/binaries-parent/build.xml:14: Cannot find /home/jenkins/agent/workspace/eclipse.platform.swt_PR-537/eclipse.platform.swt.binaries/bundles/binaries-parent/${buildDirectory}/plugins/org.eclipse.swt/buildFragment.xml imported from /home/jenkins/agent/workspace/eclipse.platform.swt_PR-537/eclipse.platform.swt.binaries/bundles/binaries-parent/build.xml
11:32:47  [ERROR] around Ant part ...<ant antfile="../binaries-parent/build.xml" target="copy.cocoa.src" />... @ 5:75 in /home/jenkins/agent/workspace/eclipse.platform.swt_PR-537/eclipse.platform.swt.binaries/bundles/org.eclipse.swt.cocoa.macosx.x86_64/target/antrun/build-main.xml
```

fredg02:
@akurtakov @sravanlakkimsetti @HannesWell @iloveeclipse PLEASE NOTE:

Build agent `b9h15-macos10.15` has been upgrade to macOS11 (Big Sur) and therefore renamed to `b9h15-macos11-x86_64`. Also the associated tag `macos-10.15` has been replaced with the tag `macos-11`. The build jobs that were tied to the machine have been adapted accordingly.

Any other build jobs that reference the old name will need to be updated.

HannesWell:
> Build agent `b9h15-macos10.15` has been upgrade to macOS11 (Big Sur) and therefore renamed to `b9h15-macos11-x86_64`. Also the associated tag `macos-10.15` has been replaced with the tag `macos-11`. The build jobs that were tied to the machine have been adapted accordingly.
> 

Thank you @fredg02 for that.
It isn't affecting this PR, but looking at [`b9h15-macos11-x86_64`](https://ci.eclipse.org/releng/computer/b9h15%2Dmacos11%2Dx86%5F64/) and [`nc1ht-macos11-arm64`](https://ci.eclipse.org/releng/computer/nc1ht%2Dmacos11%2Darm64/) I just noticed that the former is labeled with `macos-11` while the latter is labeled with `macos11`.

fredg02:
> I just noticed that the former is labeled with macos-11 while the latter is labeled with macos11.

Good catch. I copied from the existing format. If @sravanlakkimsetti agrees, I can change the tag for `nc1ht-macos11-arm64` to `macos-11` as well.

sravanlakkimsetti:
> > I just noticed that the former is labeled with macos-11 while the latter is labeled with macos11.
> 
> Good catch. I copied from the existing format. If @sravanlakkimsetti agrees, I can change the tag for `nc1ht-macos11-arm64` to `macos-11` as well.

Yes. Please relabel them as `macos-11`

HannesWell:
@fredg02 on the [b9h15-macos11-x86_64](https://ci.eclipse.org/releng/computer/b9h15%2Dmacos11%2Dx86%5F64/) I see the following error for the macosx.x86_64 build of SWT-natives.
Could it be that this is missing?
`xcrun: error: invalid active developer path (/Library/Developer/CommandLineTools), missing xcrun at: /Library/Developer/CommandLineTools/usr/bin/xcrun`

HannesWell:
Besides that it looks like the project is not prepared to push to the swt(-natives)-repo:
```
  + git push origin HEAD:refs/heads/testSWTNativesInPipeline
  git@github.com: Permission denied (publickey).
  fatal: Could not read from remote repository.
  
  Please make sure you have the correct access rights
  and the repository exists.
```

fredg02:
> @fredg02 on the [b9h15-macos11-x86_64](https://ci.eclipse.org/releng/computer/b9h15%2Dmacos11%2Dx86%5F64/) I see the following error for the macosx.x86_64 build of SWT-natives.
> Could it be that this is missing?
> xcrun: error: invalid active developer path (/Library/Developer/CommandLineTools), missing xcrun at: /Library/Developer/CommandLineTools/usr/bin/xcrun

Please try again, and in case it's still failing, please point me to the build.

HannesWell:
> > @fredg02 on the [b9h15-macos11-x86_64](https://ci.eclipse.org/releng/computer/b9h15%2Dmacos11%2Dx86%5F64/) I see the following error for the macosx.x86_64 build of SWT-natives.
> > Could it be that this is missing?
> > xcrun: error: invalid active developer path (/Library/Developer/CommandLineTools), missing xcrun at: /Library/Developer/CommandLineTools/usr/bin/xcrun
> 
> @HannesWell Please try again, and in case it's still failing, please point me to the build.

Looks like the initial error is fixed, but there seems to be another one on the new macos-x86_64 node:
```
00:04:12.118  cc -c -xobjective-c -Wall -arch x86_64 -DSWT_VERSION=4958r3   -DUSE_ASSEMBLER -DCOCOA -DATOMIC -I /Library/Java/JavaVirtualMachines/temurin-19.jdk/Contents/Home/include -I /Library/Java/JavaVirtualMachines/temurin-19.jdk/Contents/Home/include/darwin -I /System/Library/Frameworks/Cocoa.framework/Headers -I /System/Library/Frameworks/JavaScriptCore.framework/Headers os.c
00:04:12.543  In file included from os.c:19:
00:04:12.543  In file included from ./os_structs.h:18:
00:04:12.543  In file included from ./os.h:22:
00:04:12.543  In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk/System/Library/Frameworks/Cocoa.framework/Headers/Cocoa.h:12:
00:04:12.543  In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk/System/Library/Frameworks/Foundation.framework/Headers/Foundation.h:12:
00:04:12.543  /Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk/System/Library/Frameworks/Foundation.framework/Headers/NSBundle.h:91:143: error: function does not return NSString
00:04:12.543  - (NSAttributedString *)localizedAttributedStringForKey:(NSString *)key value:(nullable NSString *)value table:(nullable NSString *)tableName NS_FORMAT_ARGUMENT(1) NS_REFINED_FOR_SWIFT API_AVAILABLE(macos(12.0), ios(15.0), watchos(8.0), tvos(15.0));
```
See https://ci.eclipse.org/releng/job/eclipse.platform.swt/view/change-requests/job/PR-514/136/execution/node/275/log/

HannesWell:
> Besides that it looks like the project is not prepared to push to the swt(-natives)-repo:
> 
> ```
>   + git push origin HEAD:refs/heads/testSWTNativesInPipeline
>   git@github.com: Permission denied (publickey).
>   fatal: Could not read from remote repository.
>   
>   Please make sure you have the correct access rights
>   and the repository exists.
> ```

This was because I oversaw the SSH-agent setup in the existing/old `SWT-Increment_if_needed` job. After adding a `sshagent(['github-bot-ssh'])` step, the pipeline now successfully pushes to the swt and swt.binaries repo.

You can see the results here:
- https://github.com/eclipse-platform/eclipse.platform.swt/tree/testSWTNativesInPipeline
- https://github.com/eclipse-platform/eclipse.platform.swt.binaries/tree/testSWTNativesInPipeline

But while testing pushing I noticed that the automated natives-build respectively the automated enforced-qualifier commits will trigger another round of automated updates, which likely leads to an infinite update-cycle.
So I will need to think about a mechanism to detect if the latest commit was created by the automated SWT-update.
Probably based on the commit message and committer/author.
On the long run, the enforced qualifier updates should not be necessary at all, which then hopefully makes this mechanism obsolete. But I don't want to change that in this PR.

fredg02:
> Looks like the initial error is fixed, but there seems to be another one on the new macos-x86_64 node

That error should be fixed as well now. It required to update the MacOS Developer Command Line Tools from version 12.5 to 13.2 (with `softwareupdate -i "Command Line Tools for Xcode-13.2"`) .

HannesWell:
> > Looks like the initial error is fixed, but there seems to be another one on the new macos-x86_64 node
> 
> That error should be fixed as well now. It required to update the MacOS Developer Command Line Tools from version 12.5 to 13.2 (with `softwareupdate -i "Command Line Tools for Xcode-13.2"`) .

Great. Thank you! I can confirm that the mac build is now working for both architectures.

> But while testing pushing I noticed that the automated natives-build respectively the automated enforced-qualifier commits will trigger another round of automated updates, which likely leads to an infinite update-cycle.
> So I will need to think about a mechanism to detect if the latest commit was created by the automated SWT-update.
> Probably based on the commit message and committer/author.

I prevent such self-loop now by performing a check if the HEAD commit was committed by the Eclipse Releng Bot. In that case the build simply aborts, as you can see for example in https://ci.eclipse.org/releng/job/eclipse.platform.swt/job/testSWTNativesInPipeline/8/console.
This way and endless loop of builds where the automated update triggers a new build immediately which again triggers a new build is prevented.

So I think this is ready now. You can see the results here:
- https://github.com/eclipse-platform/eclipse.platform.swt/tree/testSWTNativesInPipeline
- https://github.com/eclipse-platform/eclipse.platform.swt.binaries/tree/testSWTNativesInPipeline

The only thing to do, before this can be submitted is to remove the last commit (which was only used to test pushing) and to squash the others.

@mickaelistria, @akurtakov or anybody else: Do you want to review this?
I assume you prefer to wait until the master branch opens for the 2023-06 before submitting this?

mickaelistria:
> @mickaelistria, @akurtakov or anybody else: Do you want to review this?

I've personally lost track (again) on this topic and don't intend to refocus on it soon enough to give a helpful review, sorry.

> I assume you prefer to wait until the master branch opens for the 2023-06 before submitting this?

Yes please.

HannesWell:
> > @mickaelistria, @akurtakov or anybody else: Do you want to review this?
> 
> I've personally lost track (again) on this topic and don't intend to refocus on it soon enough to give a helpful review, sorry.
> 
OK, no problem. I did some more testing (changes and a few more adjust and I'm now relatively confident that this change is fine as it is.
Therefore I suggest to submit this as soon as the master opens again and if anyone encounters problems I'll try to fix them as soon as possible.
Of course everybody interested is invited to review this PR until then.

HannesWell:
Since the master branch is now open I plan to submit this PR tomorrow evening, unless anybody objects.

HannesWell:
All tests on Jenkins are green and the GH-workflow failures are the same like on previous master builds. Therefore I believe they are unrelated.
So this is ready for submission now; Merging.
I will monitor the next master branch build to verify that it behaves as expected.

Thanks to everyone that helped to make this happen, especially @fredg02!

HannesWell:
So the build succeeded as expected in https://ci.eclipse.org/releng/job/eclipse.platform.swt/job/master/24/ 🎉 
And as intended the subsequent master-branch triggered from the push of the updated timestamps was aborted to prevent an endless build loop: https://ci.eclipse.org/releng/job/eclipse.platform.swt/job/master/25/

Unfortunately we will now see aborted builds in the master queue, but I didn't found an elegant way to suppress. Maybe it would be possible with adjusting the Jenkins/GH-hook configuration. But I don't have access to that.




> The build procedure is based on the configurations in https://ci.eclipse.org/releng/view/SWT%20Natives/ Namely these are:
> 
>     * https://ci.eclipse.org/releng/job/SWT-Increment_if_needed/configure
> 
>     * https://ci.eclipse.org/releng/job/cocoa_aarch64/configure
> 
>     * https://ci.eclipse.org/releng/job/cocoa_x86_64/configure
> 
>     * https://ci.eclipse.org/releng/job/gtk_linux_aarch64/configure
> 
>     * https://ci.eclipse.org/releng/job/gtk_linux_ppc64le/configure
> 
>     * https://ci.eclipse.org/releng/job/gtk_linux_x86_64/configure
> 
>     * https://ci.eclipse.org/releng/job/win32_x86_64/configure
> 
> 
> The steps of all those jobs are now perform in the Jenkins-build of this o.e.swt repo and thus makes all the mentioned jobs obsolete. This also makes building the swt-binaries as part of the I-builds obsolete. It has also the advantage that one can get immediate feedback in verification-builds for changes that affect the SWT-binaries.

Since the mentioned seven jobs are now not necessary anymore I suggest we delete them on the long run. For some time I think it is reasonable to keep them as a reference but once we are confident everything works I think we can just delete them.
From how the jobs are configured I assume they are not used for any maintenance work, is that correct.

PR #516, opened 2022-12-28
[Build] Perform swt.binaries clone non-shallow to speed it up

HannesWell:
In other PRs I had the impression that a non-shallow clone significantly speed up the cloning (contrary to the intuition).

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -89,8 +89,8 @@ spec:
 						checkout scm
 					}
 					dir ('eclipse.platform.swt.binaries') {
-						checkout([$class: 'GitSCM', branches: [[name: '*/master']],
-							extensions: [[$class: 'CloneOption', timeout: 120, depth: 1, shallow: true]],
+						checkout([$class: 'GitSCM', branches: [[name: 'refs/heads/master']],
+							extensions: [[$class: 'CloneOption', timeout: 120]],
 							userRemoteConfigs: [[url: 'https://github.com/eclipse-platform/eclipse.platform.swt.binaries.git']]
 						])
 					}


HannesWell:
OK, the Jenkins Stage overview did not show a improvement, it was even a bit slower).
Maybe the improvment I noticed in https://github.com/eclipse-platform/eclipse.platform.swt/pull/514 was because there the build is not running in a container anymore.
Anyway then this is not a relevant standalone change.

SyntevoAlex:
@HannesWell  Please see discussion in https://bugs.eclipse.org/bugs/show_bug.cgi?id=562937#c6
with proper commandline, shallow clone only takes 1 second instead of ~7 minutes on my machine.

HannesWell:
> @HannesWell Please see discussion in https://bugs.eclipse.org/bugs/show_bug.cgi?id=562937#c6 with proper commandline, shallow clone only takes 1 second instead of ~7 minutes on my machine.

Thank you for that hint. Can you say what was wrong here that I didn't get such an improvement here?
Nevertheless I'm currently working on improving the SWT build. And eventually we maybe don't need the tags anymore or another idea (from others) was to store the binaries in a artifact repository, which should also speed up the retrial of existing artifacts.

HannesWell:
The first step of the build improvements are in https://github.com/eclipse-platform/eclipse.platform.swt/pull/514.
There the checkout also takes only 1~2 min instead of 5min. I'm currently not sure why this is the case. The main difference in this regard is that with that PR the build is running on the centos-latest agents and the master currently runs on a specific kubernetes agent.

SyntevoAlex:
binaries repo contains a lot of binaries, which as of [2022-03-10](https://bugs.eclipse.org/bugs/show_bug.cgi?id=562937#c6) downloaded 1.21gb. Clearly this is not fast.

Shallow clone downloads only one latest commit, this downloaded only 9.5mb and resulted in speed improvement from 7 minutes to 1 second (420x speedup huh).

However, in jenkins, also all tags are requested. And in binaries repo, pretty much every single commit is tagged.

So it's asking for "I only want one last commit, but please also include all other commits" which results in no improvement.

I have no idea why all tags are needed in checkout. Probably they are not needed at all and just happen to be there by default, or by oversight, or whatever.

SyntevoAlex:
Try adding `,noTags: true` just after `shallow: true`

PR #521, opened 2023-01-04
Enable publish of code metrics to github

laeubi:
FYI @akurtakov


laeubi:
As it is just build-configuration change freeze period should not apply here.

PR #535, opened 2023-01-23
#450/Bug 578929 - Control.setFocus brings windows to front

tmssngr:
If the application is in the background and the GUI is updated, especially a different control is focused, it might happen that the shell will be brought to front.

This patch should fix that if in your SWT-based application you add

display.setData(Display.ACTIVATE_SHELL_ON_FORCE_FOCUS, false);

after having created the display instance. If you want an application window be brought to front, use shell.forceActive().

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Control.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Control.java
@@ -1445,6 +1445,7 @@ public boolean forceFocus () {
 	Decorations shell = menuShell ();
 	shell.setSavedFocus (this);
 	if (!isEnabled () || !isVisible () || !isActive ()) return false;
+	if (display.getActiveShell() != getShell() && !Display.isActivateShellOnForceFocus()) return false;
 	if (isFocusControl ()) return true;
 	shell.setSavedFocus (null);
 	NSView focusView = focusView ();
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
@@ -6946,4 +6946,7 @@ static long windowProc(long id, long sel, long arg0, long arg1, long arg2, long 
 	}
 }
 
+static boolean isActivateShellOnForceFocus() {
+	return "true".equals(System.getProperty("org.eclipse.swt.internal.activateShellOnForceFocus", "false")); //$NON-NLS-1$
+}
 }
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Control.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Control.java
@@ -2914,6 +2914,7 @@ public boolean forceFocus () {
 	Shell shell = getShell ();
 	shell.setSavedFocus (this);
 	if (!isEnabled () || !isVisible ()) return false;
+	if (display.getActiveShell() != getShell() && !Display.isActivateShellOnForceFocus()) return false;
 	shell.bringToTop (false);
 	return forceFocus (focusHandle ());
 }
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
@@ -6244,4 +6244,7 @@ static int _getDeviceZoom (long monitor_num) {
 	return DPIUtil.mapDPIToZoom (dpi);
 }
 
+static boolean isActivateShellOnForceFocus() {
+	return "true".equals(System.getProperty("org.eclipse.swt.internal.activateShellOnForceFocus", "false")); //$NON-NLS-1$
+}
 }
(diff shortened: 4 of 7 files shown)


bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Control.java
```diff
@@ -1445,6 +1445,7 @@ public boolean forceFocus () {
 	Decorations shell = menuShell ();
 	shell.setSavedFocus (this);
 	if (!isEnabled () || !isVisible () || !isActive ()) return false;
+	if (display.getActiveShell() != getShell() && !getDisplay().activateShellOnForceFocus) return false;
```
iloveeclipse, line 1448: Is `getDisplay().` intentionally used instead of `display.` here?

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
```diff
@@ -117,6 +117,18 @@ enum APPEARANCE {
...
+	/**
+	 * This constant is used for configuring whether control.forceFocus() activates the parent shell.
+	 * Expects a <code>boolean</code> value.
+	 * To prevent control.forceFocus() to activate the shell, invoke
+	 * <pre>
+	 *   display.setData(Display.ACTIVATE_SHELL_ON_FORCE_FOCUS, false);
+	 * </pre>
+	 * after having created the display.
+	 */
+	public static final String ACTIVATE_SHELL_ON_FORCE_FOCUS = "org.eclipse.swt.activateShellOnForceFocus"; //$NON-NLS-1$
```
iloveeclipse, line 129: This adds **public** API and API tools reports an error (missing since 3.123).
**If** the field need to be public, it needs a proper \@since comment.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
```diff
@@ -117,6 +117,18 @@ enum APPEARANCE {
...
 	static final String USE_SYSTEM_THEME = "org.eclipse.swt.display.useSystemTheme";
 
+	/**
+	 * This constant is used for configuring whether control.forceFocus() activates the parent shell.
+	 * Expects a <code>boolean</code> value.
+	 * To prevent control.forceFocus() to activate the shell, invoke
+	 * <pre>
+	 *   display.setData(Display.ACTIVATE_SHELL_ON_FORCE_FOCUS, false);
+	 * </pre>
+	 * after having created the display.
```
iloveeclipse, line 127: Please provide information about expected **default** value / behavior.

iloveeclipse (changes requested):
In general, wouldn't be a `Shell()` constructor argument or `set/get` methods would be more suitable for the API that changes Shell behavior?
I see that the Display "key's" you are trying to mimic were all **internal**, not API.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Control.java
tmssngr: No, the `getDisplay()` was not by intention. I've replaced it with `display` in the new pull request.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
tmssngr: IMHO it makes sense to have this constant being public, so it could be invoked in the user's code which usually uses different packages. In the new pull request I've added the @since tag, but I'm unsure about the version - I've used `@since 4.959` because this is the version that occurs in the `version.txt` inside the swt.jar.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
tmssngr: Done in the new pull request.

tmssngr:
> In general, wouldn't be a `Shell()` constructor argument or `set/get` methods would be more suitable for the API that changes Shell behavior?

I don't think so, because I rather consider it an application wide option, not one which is enabled for this shell, but disabled for another.

> I see that the Display "key's" you are trying to mimic were all **internal**, not API.

Yes, on GTK and Win32 the constants have the `internal` part, but for macOS there is
```
static final String USE_SYSTEM_THEME = "org.eclipse.swt.display.useSystemTheme";
```
I'm indifferent - you decide. If you want me to change the constant value to `org.eclipse.swt.internal.activateShellOnForceFocus` or something else, let me know.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
iloveeclipse: Thomas, you should use the version defined in the manifest.mf. This is what is checked by API tooling to make sure the API's aren't broken / changed on next release without proper version update.

iloveeclipse:
> I'm indifferent - you decide. If you want me to change the constant value to `org.eclipse.swt.internal.activateShellOnForceFocus` or something else, let me know.

I'm indifferent either.
Let's ask for @SyntevoAlex @niraj-modi (and any anyone other's listening) opinion.

It is easy to add new (inconsistent) API but it will be impossible to get rid of it (move/rename) after release (considering the long grace period for API removal).

SyntevoAlex:
> Let's ask for @SyntevoAlex (and any anyone other's listening) opinion.

I replied in https://github.com/eclipse-platform/eclipse.platform.swt/issues/450
I think that it's better to fix the old API.

tmssngr:
Maybe it would make more sense to
- make the option non-public API
- change the default to `false` (don't activate on `setFocus`) and
- use a VM option instead the `display.setData`, so there is a quick workaround if an Eclipse user might find some (further) problems. What do you think?

Issue #536, opened 2023-01-26
[GTK] Shell.open: incorrect order of bringToTop and setVisible


tmssngr:
**Describe the bug**
The beginning of `Shell.open()` looks as this:
```java
	...
	bringToTop (false);
	setVisible (true);
	...
```
The problem is that `bringToTop` is ignored if the shell is invisible. Hence, `bringToTop` is only invoked as a undesired side-effect of `control.setFocus` (see #450).

**To Reproduce**
Debug a `shell.open` invocation.

**Expected behavior**
`bringToTop` should be invoked after `setVisible`, so it does something useful.

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [ ] Windows
- - [x] Linux
- - [ ] macOS

tmssngr:
Are there objections against this patch or is currently a release period where no serious pull requests are accepted?

tmssngr:
OK, this sounds reasonable. Removing the `bringToTop (false)` in #536 and reintroduce it in #450 as a fix.

iloveeclipse:
Thanks Thomas, fixed via #537.

PR #537, opened 2023-01-26
#536: [GTK] Shell.open: incorrect order of bringToTop and setVisible

tmssngr:
bringToTop is ignored if the shell is not visible. Without the shell being activated, the behavior of restoreFocus requires the undesired shell activation of setFocus.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Shell.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Shell.java
@@ -2029,7 +2029,6 @@ long notifyState (long object, long arg0) {
  */
 public void open () {
 	checkWidget ();
-	bringToTop (false);
 	setVisible (true);
 	if (isDisposed ()) return;
 	/*


iloveeclipse:
Is this only the fix for https://github.com/eclipse-platform/eclipse.platform.swt/issues/536 or does it at same time adresses  https://github.com/eclipse-platform/eclipse.platform.swt/issues/450?

tmssngr:
I've noticed it when debugging #450 and it looks like a good preparation fix.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Shell.java
```diff
@@ -2029,8 +2029,8 @@ long notifyState (long object, long arg0) {
  */
 public void open () {
 	checkWidget ();
-	bringToTop (false);
 	setVisible (true);
+	bringToTop (false);
```
iloveeclipse, line 2033: Since many years long this call did **literally nothing**, would be safer to simply remove that?

iloveeclipse: > Since many years long this call did **literally nothing**, would be safer to simply remove that?

@SyntevoAlex : WDYT?

iloveeclipse:
Jenkins build aborted due https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/2478 :-(

iloveeclipse (approved):
LGTM

Issue #106, opened 2022-05-09
test_cut (org.eclipse.swt.tests.junit.Test_org_eclipse_swt_widgets_Text)  is flapping


laeubi:
The test fails from one build to the other without any SWT code changes.

SyntevoAlex:
I have investigated that.

All clipboard-related tests can fail, I myself seen these:
```
Test_org_eclipse_swt_widgets_Text.test_consistency_Segments
Test_org_eclipse_swt_widgets_Text.test_copy
Test_org_eclipse_swt_widgets_Text.test_paste
Test_org_eclipse_swt_widgets_Combo.test_consistency_Segments
Test_org_eclipse_swt_widgets_Combo.test_copy
Test_org_eclipse_swt_custom_CCombo.test_cut
```

The failure is chance based, with around 1% chance.

It's easy to reproduce by running this code:
```java
final Display display = new Display();
final Shell shell = new Shell(display);
shell.setLayout (new GridLayout (1, true));

Text text = new Text(shell, 0);
System.out.format("%X = handle%n", text.handle);

shell.pack();
shell.open();
while (display.readAndDispatch()) {};

for (int i = 0; i < 10000; i++) {
	text.setText("01234567890");
	text.setSelection(2, 5);
	text.cut();
	if (!text.getText().equals("01567890")) {
		System.out.println("Error!");
	}
}

while (!shell.isDisposed()) {
	if (!display.readAndDispatch()) {
		display.sleep();
	}
}

display.dispose();
```

I have debugged, the problem is that `OpenClipboard()` fails inside WINAPI handler for messages like `WM_COPY` or `WM_CUT`.
Can be seen with WinDBG breakpoint: `bp win32u!NtUserOpenClipboard+0x14 ".if (@rax == 1) {g}"`.

I'm not exactly sure what causes the failure.
It doesn't make sense that some other app repeatedly claims ownership of clipboard, but maybe that is indeed the case.
Maybe when SWT copies something to clipboard, some other app checks that, and if SWT tries to copy again while app is checking, it fails.
Getting a definitive answer would require kernel debugging, and I don't think the Issue is important enough to spend time on that.

basilevs:
https://github.com/eclipse-platform/eclipse.platform.swt/actions/runs/7674714373/job/20919789301
```
org.junit.ComparisonFailure: expected:<[00000]> but was:<[]>
	at org.junit.Assert.assertEquals(Assert.java:117)
[...]

PR #553, opened 2023-02-16
#450/Bug 578929 - Control.setFocus brings windows to front

tmssngr:
If the application is in the background and the GUI is updated, especially a different control is getting focused, it might happen that the shell will be brought to front. This patch should fix that.

To get the previous behavior (activate the shell on control.forceFocus(), set the system property
"org.eclipse.swt.internal.activateShellOnForceFocus" to "true").

If you want an application window be brought to front, instead of relying on control.forceFocus() to bring the shell to front, use shell.forceActive().

Shell.open: an explicit bringToTop() is needed now, because the implicit one from setFocus/forceFocus() is not happening any more. For the first shell it needs to be forced, because otherwise it won't do anything (display.activeShell is null).

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
@@ -6946,4 +6946,7 @@ static long windowProc(long id, long sel, long arg0, long arg1, long arg2, long 
 	}
 }
 
+static boolean isActivateShellOnForceFocus() {
+	return "true".equals(System.getProperty("org.eclipse.swt.internal.activateShellOnForceFocus", "false")); //$NON-NLS-1$
+}
 }
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Shell.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Shell.java
@@ -1604,6 +1604,7 @@ long gtk_focus_in_event (long widget, long event) {
 	} else {
 		ignoreFocusIn = false;
 	}
+	restoreFocus();
 	return 0;
 }
 
@@ -2030,6 +2031,8 @@ long notifyState (long object, long arg0) {
 public void open () {
 	checkWidget ();
 	setVisible (true);
+	// force is necessary, because otherwise it won't do anything
+	bringToTop (true);
 	if (isDisposed ()) return;
 	/*
 	 * When no widget has been given focus, or another push button has focus,
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
@@ -1071,6 +1071,7 @@ public boolean forceFocus () {
 	Decorations shell = menuShell ();
 	shell.setSavedFocus (this);
 	if (!isEnabled () || !isVisible () || !isActive ()) return false;
+	if (display.getActiveShell() != shell && !Display.isActivateShellOnForceFocus()) return false;
 	if (isFocusControl ()) return true;
 	shell.setSavedFocus (null);
 	/*
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Control.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Control.java
@@ -1461,6 +1461,7 @@ public boolean forceFocus () {
 	Decorations shell = menuShell ();
 	shell.setSavedFocus (this);
 	if (!isEnabled () || !isVisible () || !isActive ()) return false;
+	if (display.getActiveShell() != shell && !Display.isActivateShellOnForceFocus()) return false;
 	if (isFocusControl ()) return true;
 	shell.setSavedFocus (null);
 	NSView focusView = focusView ();
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Control.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Control.java
@@ -2944,6 +2944,7 @@ public boolean forceFocus () {
 	Shell shell = getShell ();
 	shell.setSavedFocus (this);
 	if (!isEnabled () || !isVisible ()) return false;
+	if (display.getActiveShell() != shell && !Display.isActivateShellOnForceFocus()) return false;
 	shell.bringToTop (false);
 	return forceFocus (focusHandle ());
 }
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
@@ -6244,4 +6244,7 @@ static int _getDeviceZoom (long monitor_num) {
 	return DPIUtil.mapDPIToZoom (dpi);
 }
 
+static boolean isActivateShellOnForceFocus() {
+	return "true".equals(System.getProperty("org.eclipse.swt.internal.activateShellOnForceFocus", "false")); //$NON-NLS-1$
+}
 }
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
@@ -5194,4 +5194,7 @@ static char [] withCrLf (char [] string) {
 	return result;
 }
 
+static boolean isActivateShellOnForceFocus() {
+	return "true".equals(System.getProperty("org.eclipse.swt.internal.activateShellOnForceFocus", "false")); //$NON-NLS-1$
+}
 }
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_widgets_Shell.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_widgets_Shell.java
@@ -19,6 +19,7 @@ import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertSame;
 import static org.junit.Assert.assertTrue;
 
 import java.io.IOException;
@@ -38,6 +39,7 @@ import org.eclipse.swt.widgets.Button;
 import org.eclipse.swt.widgets.Display;
 import org.eclipse.swt.widgets.Event;
 import org.eclipse.swt.widgets.Shell;
+import org.eclipse.swt.widgets.Text;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
@@ -921,4 +923,52 @@ public void test_bug558652_scrollBarNPE() {
 	}
 }
 
+@Test
+public void test_Issue450_NoShellActivateOnSetFocus() {
+	Display display = shell.getDisplay();
+	final int timeout = 1000;
+
+	// First shell
+	final Shell shell1 = new Shell(shell);
+	final Text text11 = new Text(shell1, SWT.BORDER | SWT.SINGLE);
+	final Text text12 = new Text(shell1, SWT.BORDER | SWT.SINGLE);
+
+	// Second shell
+	final Shell shell2 = new Shell(shell);
+	final Text text21 = new Text(shell2, SWT.BORDER | SWT.SINGLE);
+	final Text text22 = new Text(shell2, SWT.BORDER | SWT.SINGLE);
+
+	// System.out.println("open 1st shell");
+	assertTrue(SwtTestUtil.waitEvent(shell1::open, shell1, SWT.Activate, timeout));
+	assertSame("expecting the 1st shell to be activated", display.getActiveShell(), shell1);
+	assertTrue("expecting the 1st text field to be focused in the 1st shell", text11.isFocusControl());
+
+	// System.out.println("open 2nd shell");
+	assertTrue(SwtTestUtil.waitEvent(shell2::open, shell2, SWT.Activate, timeout));
+	assertSame("expecting the 2nd shell to be activated", display.getActiveShell(), shell2);
+	assertTrue("expecting the 1st text field in 2nd shell to be focused", text21.isFocusControl());
+
+	// System.out.println("setting focus to 2nd text field in 1st (unactivated) shell");
+	text12.setFocus();
+	assertSame("expecting the 2nd shell to remain activated", display.getActiveShell(), shell2);
+
+	// System.out.println("activating 1st shell");
+	// Fails here? Check Bug 575712 that only occurs on Ubuntu, and only Ubuntu 21.04+
+	assertTrue(SwtTestUtil.waitEvent(shell1::setActive, shell1, SWT.Activate, timeout));
+	assertSame("expecting the 1st shell to be activated", display.getActiveShell(), shell1);
+	assertTrue("expecting the the 1st shell to have remembered the previous setFocus and with the shell activation setting it to the 2nd text field", text12.isFocusControl());
+
+	// System.out.println("disposing 1st shell");
+	assertTrue(SwtTestUtil.waitEvent(shell1::dispose, shell2, SWT.Activate, timeout));
+	assertSame("expecting the 2nd shell to be activated after the 1st, active has been disposed", display.getActiveShell(), shell2);
+	assertTrue("expecting the 1st text field in the 2nd shell to still have the focus because it hasn't been changed", text21.isFocusControl());
+
+	// System.out.println("setting the focus to the 2nd text field");
+	text22.setFocus();
+	assertSame("expecting the 2nd shell to remain activated", display.getActiveShell(), shell2);
+	assertTrue("expecting the 2nd text field to have received the focus", text22.isFocusControl());
+
+	// System.out.println("disposing the 2nd shell, too");
+	shell2.dispose();
+}
 }


lshanmug:
@elsazac @deepika-u Can you please verify the PR?

deepika-u:
@tmssngr 
Attempted with FocusToFrontTest.java snippet.

As per my understanding, expected behavior is : 
3 windows are being opened and asked to load something. 3rd window being the last once finished loading needs to continue to be focused. (1st and 2nd windows since started loading in the background should never get focused (irrespective of finishing) till they are exclusively focused by user/invoked programatically.)

Before the fix : 
3 windows are opened and asked to load something in the sequence(window 1 followed by window 2 followed by window 3). 
Window 3 is on focus right now. 
Then from window 3 focus goes to window 1 as it finished loading. 
Then window 2 gets the focus as it finished loading. 
Then window 3 gets the focus as it finished lastly and remains focused.

After the fix :
It is working as expected and focus continues to remain on window 3(though it finishes loading after window 1 and window 2).

This fix looks fine for me on windows 11. Please correct/add if my understanding is fine.

elsazac:
@tmssngr 
Attempted with FocusToFrontTest.java snippet on MacOS aarch 64.

As per my understanding, Expected Behavior is:
Window 1 shows up and starts loading.
Window 2 shows up and starts loading.
Window 3 shows up and starts loading.
Now focus remains on window 3, window 1 finished loading in the background, then window 2 finished loading in the background, window 3 finished loading on the foreground and focus continues to remain on window 3.

Without PR :
Window 1 and Window 2 doesn’t show up and hence looks like loading/not loading.
Window 3 shows up and starts loading.
Then window 1 finished loading and shows up focused.
Then window 2 finished loading and shows up focused.
Then window 3 finished loading and shows up focused.
Now focus remains on window 3.

With PR:
only the window 3 shows up loading and finished loading. After that , only if we perform an external action (for eg minimise the window 3 or close the the window 3)we will see window 2. When window 2 is minimised or closed window 1 shows up.

The behaviour in Mac is totally different from that in windows. 
Irrespective of the fix, window 1 and window 2 are not showing up. Looks like this is another issue.
Focus is not shifting from window 3( irrespective of window 1, window 2 loading) and continues to remain on window 3.So this fix looks good to me.

tmssngr:
To what `FocusToFrontTest.java` you are referring?

elsazac:
The first snippet that was provided along with the issue.

tmssngr:
Sorry for the delay. On my Mac mini M1 (macOS 13.2.1) the `FocusToFrontTest` test behaves the same as on all other platforms. Without patch, 3 windows open that get activated for each "loading finished". With patch the windows open but the top most shell remains the active one independent of the others.

elsazac:
This is the behaviour without patch.
Screenshot 2023-03-30 at 6 07 21 PM

When I run this snippet window 1 and 2 doesn't come up.Window 3 comes up and starts loading.Then window 1 finishes loading and comes to the front followed by window 2 and window 3.Focus remains on window 3.

This is the behaviour with patch
Screenshot 2023-03-30 at 6 09 54 PM

Only window 3 opens up and finishes loading.If I close or minimise the window 3 only then I could see window 2.If I close or minimise window 2 , only then window 1 shows up. Else I cannot see window 1 or window 2 are open or not.

In both the cases opening of window 1 and window 2 doesn't show up at all.Only after loading finished window 1 and window 2 shows up without patch .These 2 windows (window 1 and window 2 before loading)show up on windows but not on Mac.

tmssngr:
For me it looks before patch:
before-patch
and with patch:
with-patch

tmssngr:
Are these failing builds caused by my pull request?

tmssngr:
What else is needed from my side to merge this PR?

iloveeclipse:
> What else is needed from my side to merge this PR?

A positive review from @SyntevoAlex?

deepika-u:
> For me it looks before patch: before-patch [ before-patch ](https://user-images.githubusercontent.com/6029561/228876396-0db385b1-b1a4-4f4f-b9ba-e43e875b9731.gif) [ ](https://user-images.githubusercontent.com/6029561/228876396-0db385b1-b1a4-4f4f-b9ba-e43e875b9731.gif) and with patch: with-patch [ with-patch ](https://user-images.githubusercontent.com/6029561/228876488-4daaa6e2-3c19-444a-9769-897d7cdbaf66.gif) [ ](https://user-images.githubusercontent.com/6029561/228876488-4daaa6e2-3c19-444a-9769-897d7cdbaf66.gif)

My vote goes for the fix.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
```diff
@@ -6946,4 +6946,7 @@ static long windowProc(long id, long sel, long arg0, long arg1, long arg2, long
 	}
 }
 
+static boolean isActivateShellOnForceFocus() {
+	return "true".equals(System.getProperty("org.eclipse.swt.internal.activateShellOnForceFocus", "false")); //$NON-NLS-1$
```
SyntevoAlex, line 6950: `Boolean.getBoolean()` ?

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Shell.java
```diff
@@ -2030,6 +2031,7 @@ long notifyState (long object, long arg0) {
 public void open () {
 	checkWidget ();
 	setVisible (true);
+	bringToTop (true);
```
SyntevoAlex, line 2034: Please move the explanation for `true` from commit message to code comment here.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
```diff
@@ -1071,6 +1071,7 @@ public boolean forceFocus () {
 	Decorations shell = menuShell ();
 	shell.setSavedFocus (this);
 	if (!isEnabled () || !isVisible () || !isActive ()) return false;
+	if (display.getActiveShell() != shell && !Display.isActivateShellOnForceFocus()) return false;
```
SyntevoAlex, line 1074: Regarding `display.getActiveShell() != shell` :
The condition looks weird in this code. If the Shell is already active, does it make sense to activate it again? I would say that no. I agree that previously it already worked like that, but adding another confusing piece to this code makes it too many already.

The other problem is the `return false` part you're adding.
This function's logic is somewhat hard to grasp (as in, one has to stop and think to fully understand it), and that early return only adds to confusion, because it's not exactly an error and is more like an expected behavior. If you don't agree yet, try to answer this question quickly: should this line be moved up or down or stay exactly where it is?

After meditating on this code for a bit, I think I would prefer it to be laid out this way:
```java
boolean activateShellOnFocus(Shell shell) {
	// Already active?
	if (display.getActiveShell() == shell) return true;

	// Issue 450: When forcing focus to a control owned by a background Shell, the Shell should stay in background
	if (!Display.isActivateShellOnForceFocus()) return false;

	// <Here, place code to activate Shell from `Control.forceFocus()`)
}

public boolean forceFocus () {
	// <function prologue is kept as is>

	// Only focus controls in active shells. Otherwise, `shell.setSavedFocus()` will cause Control to be activated the next time Shell becomes active.
	if (!activateShellOnFocus(shell)) return false;

	// <Here, keep code that was focusing Control after activating Shell>
}
```

Note that `return true` on first line of `activateShellOnFocus()` modifies your patch and needs to be tested. But it makes things easier to understand and hopefully removes unneeded API calls. Otherwise, you could leave it as it was, but explain in the code comment that you didn't risk to change old behavior even if it sounded reasonable.

SyntevoAlex (commented):
I support the idea of this change, but I feel uneasy about potential regressions. I guess the only way to move forward here is trial&error - merge, see what happens, revert/fix/reapply as needed.

I would really like to see a set of UnitTests here (one per each fixed problem and discovered regression). Because
* there was a number of regressions
* multiple platforms are affected
* the change has potentially far-reaching consequences
* I find it likely that there will be some future tweaks to this logic

In your testing snippet, there are 1000ms pauses. Try to fix that when converting to unit test.
Maybe you will need to add `SwtTestUtil.processEvents()` somewhere instead of those sleeps.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
tmssngr: No, because with `Boolean.getBoolean()` you don't have control over the default if no option is set.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
tmssngr: I think, changing the code - like you suggest - can be done later if there is time or a good reason to do it, e.g. fixing another problem. At the moment, I think, your suggested much larger changes would just delay this fix for at least another couple of months and comes with a higher risk of breaking something else.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Shell.java
tmssngr: Done.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
SyntevoAlex: I guess this depends on the expectation of how likely is it that default will be changed. If not too likely, then simpler code is better. Afterall the one who'll change the default could rewrite the code back to how it is now. And if complexity can be done later at the same cost, then there's no need to add it now. Anyways, if you believe that it's likely to be changed, it's fine to keep is as is.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
SyntevoAlex: In such case, please still extract the function, but keep code logically the same and add an explaining code comment. Adding more complexity to already complex code is not a good thing to do.

tmssngr: I'd disagree. My intention for this whole RFE is to separate setting focus from activating the shell. The `Display.isActivateShellOnForceFocus()` is only a safety net for users who report a regression by this RFE to test-drive the current behavior and have a quick fix. In a second step (1 year later?) I would go ahead and remove `Display.isActivateShellOnForceFocus()` if no regression would have been reported. I think, an extra method `activateShellOnFocus()` would carve the existing behavior in stone, contradicting the separation of tasks.

SyntevoAlex: I anticipate that code won't be removed (simply because everyone would forget to). Also, deleting a dedicated function is easier than figuring which lines of code to delete. Either way, even though I don't like your approach, I won't block the patch over it. Consider this a suggestion.

SyntevoAlex:
I would still very much like to see unit tests here.
Otherwise, it would be too hard to continue iterating on it if there's yet another regression. Also, due to how non-obvious things are (especially how other code relies on old behavior) I fear that things will get broken later.

tmssngr:
Please look at this unit test. It reliably produces an NPE which might be caused by the disposing of the display inside the shell's dispose listener. I'm quite sure this NPE is independent of the patch.

SyntevoAlex:
Regarding new unit test:
1. Please add code comments to explain the idea behind specific tests.
   Like, "when setting focus to a control, Shell shall not activate", "but when activating it later, control shall gain focus"
2. If possible, please rework delay into waiting for specific event
   For example, could you wait for `SWT.Paint` instead of waiting fixed time?
   Or maybe there's some even better event to wait for.
   Fixed small pauses are unreliable and will give false positives.
   Fixed larger pauses will make the test suite slower.

> It reliably produces an NPE which might be caused by the disposing of the display inside the shell's dispose listener

I'm not sure if it's valid to dispose display from any listener at all.
I don't know what is the design here.
I suggest that you simply don't do that in your test.
In your listener, when test is done, dispose main shell, and change display loop condition to wait for shell.

tmssngr:
Alex, I think I've changed the code according to your wishes in the last pull request.

SyntevoAlex:
You forgot to add the new JUnit class into the list of all tests. Please add in `AllWidgetTests`.

Also, I see that pretty much all other tests refrain from polluting output with `System.out.println`. I think that it would make sense to comment out your `System.out.println`.

With these changes, I think I'll merge the patch and then we see if it causes any new problems.

tmssngr:
> You forgot to add the new JUnit class into the list of all tests. Please add in `AllWidgetTests`.

I didn't know that it needs to be hard-linked there. Usually, JUnit collects all test cases it finds.

> Also, I see that pretty much all other tests refrain from polluting output with `System.out.println`. I think that it would make sense to comment out your `System.out.println`.

Done.

tmssngr:
Does someone has an idea how to fix this error?
```
Error:  Errors: 
Error:    Test_org_eclipse_swt_widgets_Control_focus.testSetFocusDoesNotActivateShell:37 » SWT Not implemented [multiple displays]
[INFO] 
Error:  Tests run: 4037, Failures: 0, Errors: 1, Skipped: 6

```
It looks like SWT does not like a new instance of `Display`.

merks:
Yes, not all platforms support creating multiple displays so you should not do that.   I see other tests use org.eclipse.swt.widgets.Display.getDefault().

SyntevoAlex:
Let's give it a try. In master now.

@iloveeclipse @lshanmug this PR is kind of breaking change. When app sets focus to a control and relies (incorrectly) that Shell will also become active, this will no longer be the case. Explicit `Shell.setActive()` will need to be added. I'm guess I'll have to ... uh, "volunteer"... to fix these issues.

elsazac:
@SyntevoAlex 

There is a test failure in the build pertaining to this PR.
`org.eclipse.swt.tests.junit.Test_org_eclipse_swt_widgets_Shell`

iloveeclipse:
> @elsazac Where can the stacktrace be found?

See latest nightly build root page:
https://download.eclipse.org/eclipse/downloads/drops4/I20230703-1800/testResults.php

Go from there to Mac swt results

image

- https://download.eclipse.org/eclipse/downloads/drops4/I20230703-1800/testresults/html/org.eclipse.swt.tests_ep429I-unit-mac64-java17_macosx.cocoa.x86_64_17.html
- https://download.eclipse.org/eclipse/downloads/drops4/I20230703-1800/testresults/html/org.eclipse.swt.tests_ep429I-unit-macM1-java17_macosx.cocoa.aarch64_17.html

```
java.lang.AssertionError
at org.junit.Assert.fail(Assert.java:87)
at org.junit.Assert.assertTrue(Assert.java:53)
at org.eclipse.swt.tests.junit.Test_org_eclipse_swt_widgets_Shell.test_Issue450_NoShellActivateOnSetFocus(Test_org_eclipse_swt_widgets_Shell.java:962)
```

Please also check the windows fail in a different test (which I haven't seen failing before):
- https://download.eclipse.org/eclipse/downloads/drops4/I20230703-1800/testresults/html/org.eclipse.swt.tests_ep429I-unit-win32-java17_win32.win32.x86_64_17.html

```
expected:<01[]567890> but was:<01[234]567890>

org.junit.ComparisonFailure: expected:<01[]567890> but was:<01[234]567890>
at org.junit.Assert.assertEquals(Assert.java:117)
at org.junit.Assert.assertEquals(Assert.java:146)
at org.eclipse.swt.tests.junit.Test_org_eclipse_swt_widgets_Text.test_cut(Test_org_eclipse_swt_widgets_Text.java:374)
```

iloveeclipse:
Note, there is also now a fail in org.eclipse.jface.tests.databinding

https://download.eclipse.org/eclipse/downloads/drops4/I20230703-1800/testresults/html/org.eclipse.jface.tests.databinding_ep429I-unit-win32-java17_win32.win32.x86_64_17.html

```
java.lang.AssertionError
at org.junit.Assert.fail(Assert.java:87)
at org.junit.Assert.assertTrue(Assert.java:53)
at org.eclipse.jface.tests.internal.databinding.swt.ControlObservableValueTest.testObserveFocus(ControlObservableValueTest.java:223)
```

that seem be directly related to this PR.

deepika-u:
Is anyone aware of the cause of the windows failure? seeing for the first time. Else please let me know if i might have to open new issue for it if it is unrelated.

https://download.eclipse.org/eclipse/downloads/drops4/I20230703-1800/testresults/html/org.eclipse.swt.tests_ep429I-unit-win32-java17_win32.win32.x86_64_17.html

tmssngr:
> For x86 https://download.eclipse.org/eclipse/downloads/drops4/I20230703-1800/testresults/html/org.eclipse.swt.tests_ep429I-unit-mac64-java17_macosx.cocoa.x86_64_17.html
> 
> for arch 64 https://download.eclipse.org/eclipse/downloads/drops4/I20230703-1800/testresults/html/org.eclipse.swt.tests_ep429I-unit-macM1-java17_macosx.cocoa.aarch64_17.html

I wonder what is different for this test machine than for the one which ran yesterday successfully for this (unmerged) pull request.

deepika-u:
> For me it looks before patch: before-patch [ before-patch ](https://user-images.githubusercontent.com/6029561/228876396-0db385b1-b1a4-4f4f-b9ba-e43e875b9731.gif) [ ](https://user-images.githubusercontent.com/6029561/228876396-0db385b1-b1a4-4f4f-b9ba-e43e875b9731.gif) and with patch: with-patch [ with-patch ](https://user-images.githubusercontent.com/6029561/228876488-4daaa6e2-3c19-444a-9769-897d7cdbaf66.gif) [ ](https://user-images.githubusercontent.com/6029561/228876488-4daaa6e2-3c19-444a-9769-897d7cdbaf66.gif)

This it is exactly working for me even.

I have verified on below environment
Eclipse SDK
Version: 2023-09 (4.29)
Build id: I20230703-1800
Java vendor: Eclipse Adoptium
Java runtime version: 20+36
Java version: 20

SyntevoAlex:
Going to investigate test failures now.

SyntevoAlex:
> Attempted with FocusToFrontTest.java snippet.
> 
> Expected behavior is : 3 windows are being opened and asked to load something. 3rd window being the last once finished loading needs to continue to be focused. (1st and 2nd windows since started loading in the background should never get focused (irrespective of finishing) till they are exclusively focused by user/invoked programatically.)
> 
> This is how it is exactly working for me even.

@deepika-u sorry failed to parse. Do you mean that it's good after the patch?

SyntevoAlex:
`test_Issue450_NoShellActivateOnSetFocus` failures on macOS will hopefully be fixed by #725

SyntevoAlex:
> Test_org_eclipse_swt_widgets_Text.test_cut

Fails for a long time, unrelated to this Issue, see #106

SyntevoAlex:
> org.eclipse.jface.tests.databinding

The problem is in test only.
It can be seen how it struggles to activate Shell:

https://github.com/eclipse-platform/eclipse.platform.ui/blob/f0ce6e75ebab0142f7ff9f13afda67554e0f3818/tests/org.eclipse.jface.tests.databinding/src/org/eclipse/jface/tests/internal/databinding/swt/ControlObservableValueTest.java#L217-L221

And here (old logs before patch) it can be seen that this never worked on Windows:
https://download.eclipse.org/eclipse/downloads/drops4/I20230612-1800/testresults/ep429I-unit-win32-java17_win32.win32.x86_64_17/org.eclipse.jface.tests.databinding.BindingTestSuite.txt
```
ControlObservableValueTest.testObserveFocus() start active shell: null
active shell (2): null
active shell (3): null
```

So the test was always kind of broken.

I think that test needs to wait for `SWT.Activate` properly, like we did here:
https://github.com/syntevo/eclipse.platform.swt/blob/9e43c2659d35e66a350260999e25b432c1344f50/tests/org.eclipse.swt.tests/JUnit%20Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_widgets_Shell.java#L942

deepika-u:
> > Attempted with FocusToFrontTest.java snippet.
> > Expected behavior is : 3 windows are being opened and asked to load something. 3rd window being the last once finished loading needs to continue to be focused. (1st and 2nd windows since started loading in the background should never get focused (irrespective of finishing) till they are exclusively focused by user/invoked programatically.)
> > This is how it is exactly working for me even.
> 
> @deepika-u sorry failed to parse. Do you mean that it's good after the patch?

From the past iteration of testing, i have understood that expected behavior is like that. So it is working as per that after the fix is delivered into master(leaving the junit failures aside).

iloveeclipse:
> `test_Issue450_NoShellActivateOnSetFocus` failures on macOS will hopefully be fixed by #725



> `test_Issue450_NoShellActivateOnSetFocus` failures on macOS will hopefully be fixed by #725

Unfortunately not:
- https://download.eclipse.org/eclipse/downloads/drops4/I20230704-1800/testResults.php
- https://download.eclipse.org/eclipse/downloads/drops4/I20230704-1800/testresults/html/org.eclipse.swt.tests_ep429I-unit-mac64-java17_macosx.cocoa.x86_64_17.html
- https://download.eclipse.org/eclipse/downloads/drops4/I20230704-1800/testresults/html/org.eclipse.swt.tests_ep429I-unit-macM1-java17_macosx.cocoa.aarch64_17.html

```
Shell did not become active

java.lang.AssertionError: Shell did not become active
at org.eclipse.swt.tests.junit.SwtTestUtil.waitShellActivate(SwtTestUtil.java:418)
at org.eclipse.swt.tests.junit.Test_org_eclipse_swt_widgets_Shell.test_Issue450_NoShellActivateOnSetFocus(Test_org_eclipse_swt_widgets_Shell.java:961)
at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
```

iloveeclipse:
I've created dedicated issue for the fail, we don't need to abuse this PR anymore: https://github.com/eclipse-platform/eclipse.platform.swt/issues/727

iloveeclipse:
> > org.eclipse.jface.tests.databinding
> The problem is in test only. It can be seen how it struggles to activate Shell:

Thanks for analysis

> So the test was always kind of broken.

But now it consistently fail, which is not nice... Anyway, moving to dedicated issue: https://github.com/eclipse-platform/eclipse.platform.ui/issues/910

PR #555, opened 2023-02-17
[GTK4] Linkage error in Buttons with SWT.ARROW style

mbooth101:
When setting the alignment or orientation of arrow buttons we get a UnsatisfiedLinkError, which can be seen in a GTK4 unit test failure and in the non-functioning of SWT Snippet366:

```
Exception in thread "main" java.lang.UnsatisfiedLinkError: 'void org.eclipse.swt.internal.gtk3.GTK3.gtk_image_set_from_icon_name(long, byte[], int)'
	at org.eclipse.swt.internal.gtk3.GTK3.gtk_image_set_from_icon_name(Native Method)
	at org.eclipse.swt.widgets.Button.setOrientation(Button.java:1190)
	at org.eclipse.swt.widgets.Control.setOrientation(Control.java:5808)
	at org.eclipse.swt.snippets.Snippet366.lambda$0(Snippet366.java:62)
	at org.eclipse.swt.events.SelectionListener$1.widgetSelected(SelectionListener.java:84)
	at org.eclipse.swt.widgets.TypedListener.handleEvent(TypedListener.java:252)
	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:89)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:5855)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1529)
	at org.eclipse.swt.widgets.Display.runDeferredEvents(Display.java:5065)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:4517)
	at org.eclipse.swt.snippets.Snippet366.main(Snippet366.java:46)
```

Add a new GTK4 function `gtk_image_set_from_icon_name()` which is the same as the GTK3 version but with fewer arguments, and call it instead when on GTK4.

This change fixes the `Button#test_setAlignmentI` unit test when run on GTK4 and adds a new test `Button#test_setOrientation` to cover the orientation case.

Also, update Snippet366 to make it more correct (i.e. use selection listener instead of mousedown listener since some platforms might swallow the mousedown event for buttons.)

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT PI/gtk/library/gtk4.c
+++ b/bundles/org.eclipse.swt/Eclipse SWT PI/gtk/library/gtk4.c
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2022 IBM Corporation and others.
+ * Copyright (c) 2000, 2023 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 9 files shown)


akurtakov:
I like the patch and will push it once next stream opens.

mickaelistria:
Can you please add a commit that bumps versions for org.eclipse.swt.tests by +0.0.100?

mbooth101:
> Can you please add a commit that bumps versions for org.eclipse.swt.tests by +0.0.100?

Done! Thanks @mickaelistria

Issue #570, opened 2023-03-08
[OpenGL] Include a default binding based on Project Panama


laeubi:
I have experimented lately with [Project Panama](https://openjdk.org/projects/panama/) (available with JDK17 as an incubator feature) and created a prototype for SWT+OpenGL.

I was able to get it working to generate the bindings and adapted the [Snippet195](https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet195.java) to use this what results in a pure Java-Code only binding of OpenGL at my Linux system:

![grafik](https://user-images.githubusercontent.com/1331477/223622071-c35df938-624d-491d-ab01-e1082209516b.png)

I therefore like to suggest we start with incubating this feature for SWT and ship the bindings by default so people can use SWT+OpenGL without any additional effort, using custom bindings (like JOGL / JWGL / ..) will still be possible but we have an out of the box solution then and i think this will really boost the usefulness of SWT+OpenGL!

What needs to be done:

- [x] SWT would require Java 17 for this feature (as far as I know this would be fine for this release @akurtakov ?)
- [x] Cleanup the POC and provide it as a PR, add a new snippet
- [ ] Check this on the different linux variants (I have tested with Debian x64)
- [ ] Provide a POC for Windows @HannesWell wanted to look into this
- [ ] Provide a POC for Mac (any one interested here?)
- [ ] Somehow  update https://www.eclipse.org/swt/opengl/ to link the new snippet and tell about this opportunity
- [ ] Possibly providing feedback to the [Project Panama](https://openjdk.org/projects/panama/)  about any issues we find

My goal would be to have at least Windows+Linux support for the 2023-06 release we can ship with an explicit notice that it is _experimental_ to gather feedback for both SWT + [Project Panama](https://openjdk.org/projects/panama/) and people can start using it, once we got enough confidence and the APi is finalized and SWT has reached that jdk level (most probably Java 21) we can make this an official API and support backward compatibility for older java versions by using Multi-Release-jars, support for this was [recently added to Tycho](https://github.com/eclipse-tycho/tycho/blob/master/RELEASE_NOTES.md#building-multi-release-jars).

laeubi:
For the first two items I now created a PR:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/571

PR #571, opened 2023-03-08
Add panama binding for OpenGL to gtk

laeubi:
This do the follwoing:
- Upgrade to Java 17
- Add information about Project Panama to the readme
- Add the generated bindings
- Add a GL class for public access of all GL operations
- Add a snippet to demonstrate its usage

For more details see
- https://github.com/eclipse-platform/eclipse.platform.swt/issues/570


akurtakov:
I am in favor of the general idea but won't be able to review this week as I've taken the rest of it off. It would be nice if someone else reviews meanwhile.

mbooth101:
What is the benefit of this over using a well-maintained GL binding like LWJGL?

mickaelistria:
> What is the benefit of this over using a well-maintained GL binding like LWJGL?

I think the benefit is that Panama allows to access the C-level API directly, while LWJGL encapsulates the binding and provides some higher-level API. Some people may prefer one or the other depending on their experience and expertise. OpenGL expects could prefer the Panama style; while Java experts may enjoy LWJGL better.
However, I'm not sure the SWT project is the best place to host such a 1-1 binding with OpenGL APIs. Is this really in the scope of SWT project or would it be also profitable as an alternative project (and Snippets would demonstrate how to make use of it on some SWT Canvas)?

laeubi:
> demonstrate how to make use of it on some SWT Canvas)

SWT already provides the `GLCanvas` so I think OpenGL is in the scope of SWT already, but this does not provide a way to actually use any OpenGL function, so one always needs an external lib, what has several drawbacks:

1. I was never able to run any of the SWT OpenGL snippets, they only vaguely refer to some lib but never show how to install use that lib and connect it with an SWT app
2. I still tried this from time to time and gave up on this because the lib was not up-to-date, crashed with the current SWT version or was to complex to setup, missing OSGi headers and so on
3. The only thing i managed was a software-gl renderer but that don't work with GLCanvas at all so fall back to stupid copy-the-image and was quite slow
4. If something do not work, it usually is blamed to SWT, to OSGi or whatever by the lib-author, if complained at SWT it is blamed at the library (see also (1) as we have no runnable examples)

So I think as it is now possible with pure java SWT should offer such solution, it must not be strictly part of the swt.jar (could be an extra fragment), but at least I should be able to download the swt distro, and start the snippet and see a result.

mickaelistria:
> SWT already provides the GLCanvas so I think OpenGL is in the scope of SWT already

Showing OpenGL content in an SWT app is in the scope of SWT; but providing OpenGL bindings is another story which is not really/only about SWT or even Eclipse Platform. Basically, it would make more sense to have those bindings directly produced and maintained as a dedicated artifact in LWJGL or other Java-OpenGL binding/API and showcase in SWT how to use those.
But adding the burden of maintaining OpenGL bindings inside SWT directly, whle it looks like there are not many people even using it, thus even less interested in maintaining them, seems like a wrong organization to me.

laeubi:
> But adding the burden of maintaining OpenGL bindings inside SWT directly, whle it looks like there are not many people even using it, thus even less interested in maintaining them, seems like a wrong organization to me.

Well the problem is that as SWT does not offering such, not many people are using it ... because it is hard to setup and maintain. But if there is a common ground that came just with the UI library the people can build a lib based on that already instead of writing it for one that then is incompatible with another and forcing people to install even different libs.

An as these are just some java classes (no JNI, no dll/s/...) anyone can simply ignore it, as written this does not prevent using any other binding as well, and as SWT already claims to integrate with the native sub-system and handle the integration with Widgets this can greatly improve the usefulness of SWT (and also platform/RCP) because with that at hand one instantly can start building applications with rich and high performance graphic capabilities.

Beside that SWT uses already a lot of native (handwritten) bindings in other areas, so this could even be a starting point for other improvement in the SWT project, as that this area seems suitable as it does not directly influence any of the existing SWT sub systems.

mbooth101:
> > demonstrate how to make use of it on some SWT Canvas)
> 
> SWT already provides the `GLCanvas` so I think OpenGL is in the scope of SWT already,

The canvas wraps the rendering context, which is provided by the platform (e.g. egl/wgl/glx). It is not provided by OpenGL.

mbooth101:
> * I was never able to run any of the SWT OpenGL snippets, they only vaguely refer to some lib but never show how to install use that lib and connect it with an SWT app
> 
> * I still tried this from time to time and gave up on this because the lib was not up-to-date, crashed with the current SWT version or was to complex to setup, missing OSGi headers and so on
> 
> * The only thing i managed was a software-gl renderer but that don't work with GLCanvas at all so fall back to stupid copy-the-image and was quite slow
> 
> * If something do not work, it usually is blamed to SWT, to OSGi or whatever by the lib-author, if complained at SWT it is blamed at the library (see also (1) as we have no runnable examples)

I am happily able to use SWT to render 3D scenes using pure SWT app or as part of an OSGI app using LWJGL as the GL binding. Maybe just better documentation and examples are needed -- when I find the time, I will try to contribute in this area, because I have a few SWT-based OpenGL projects here.

laeubi:
> The canvas wraps the rendering context, which is provided by the platform (e.g. egl/wgl/glx). It is not provided by OpenGL.

Well the package-name tells differently:

https://github.com/eclipse-platform/eclipse.platform.swt/blob/ac8f665156d0275c9cfdfd6f1a63987389649321/bundles/org.eclipse.swt/Eclipse%20SWT%20OpenGL/gtk/org/eclipse/swt/opengl/GLCanvas.java#LL14

also the docs here say its support for OpenGL:
https://www.eclipse.org/swt/opengl/

> OpenGL applications use two separate APIs: the window-system independent drawing API (OpenGL)

and 

> a window-system specific integration layer (WGL under Windows, GLX under X, etc).

so if there is a " window-system independent drawing API " why should SWT not provide it with the tools of a platform independent language (java)?

mbooth101:
> lats LWJGL release is over one year old

Note that the GL spec itself hasn't changed for a while (the push for modern graphics is Vulkan) and LWJGL 3 is quite mature. Infrequent LWJGL releases is irrelevant IMO.

mbooth101:
> so if there is a " window-system independent drawing API " why should SWT not provide it with the tools of a platform independent language (java)?

SWT already provides this in GLCanvas. It wraps WGL under Windows, and GLX under X. You already quoted me saying this:

> > The canvas wraps the rendering context, which is provided by the platform (e.g. egl/wgl/glx). It is not provided by OpenGL.

mickaelistria:
> Beside that SWT uses already a lot of native (handwritten) bindings in other areas,

There is a difference between using internally some bindings to fulfill the API defined by SWT, and providing a 3200 new classes of new public APIs derived from 3rd party API and which SWT as a project is supposed to maintain according to typical API rules and quality gates of the project.
Are you actively willing to take care of the maintenance whenever OpenGL header change? Are other committers on Eclipse Platform and SWT sharing the enthusiasm?
Also, there is a legal issue here: what is the license of the derived content here? Is it legal to just port the API to another language and ship it to EPLv2?

As the discussion goes on, I'm growing the believe that offering this binding as API doesn't fit in SWT.

laeubi:
> There is a difference between using internally some bindings to fulfill the API defined by SWT, and providing a 3200 new classes of new public APIs derived from 3rd party API and which SWT as a project is supposed to maintain according to typical API rules and quality gates of the project.

Please look at the proposed code, it does only introduce **one** public API ... and there is nothing to "maintain" as this is fully generated in an automatic way (beside a one line code change to not make it "public API).

> Are you actively willing to take care of the maintenance whenever OpenGL header change?

@mbooth101 already claimed that they actually do not change and even if that only means one can not use **new features** but simply use the old API unless someone is "enthusiastic enough" to run the generator again...

> Also, there is a legal issue here: what is the license of the derived content here? 

Auto generated code do not fall under IP, otherwise every code-template used in JDT would cause "legal issues".

> As the discussion goes on, I'm growing the believe that offering this binding as API doesn't fit in SWT.

So why does then (for example) a Brwoser API "fit" SWT ... the same arguments would hold here, still SWT offers "Browser Bindings", it even offers Windos OLE Bindings and so on, in the end everything could pe provided by an external project, still it reduces the usefullness of this project if users need to plug in common things by external projects.

mickaelistria:
> Please look at the proposed code, it does only introduce one public API ... 

I'm seeing a lot of `public interface`s.

> and there is nothing to "maintain" as this is fully generated in an automatic way

Assuming that generated code doesn't require maintenance and that we can always regenerate things easily is a bit too optimistic. Look at code generated by EMF, code generated by compilers, code generated by X or Y; you often reach a point where you cannot just easily regenerate, and need to either take the generated code in your bucket of direct maintenance, or invest in making the generator remain compatible with your desires.

> Auto generated code do not fall under IP, otherwise every code-template used in JDT would cause "legal issues".

This generated code seems to be directly derived from an API design; which qualifies as important IP work; so one needs to know whether deriving the API has constraints on the license. Assuming there are no legal issues is also too optimistic here; I hope it is the case, but it can definitely not be something safe to assume by default.
JDT code templates, and code templates in general, are intentionally qualified as non-IP substantial work to allow derivative. I don't think OpenGL qualifies as such.
Anyway, if we want to move forwards with including Panama bindings, we will need to consult Eclipse IP team to ensure it's not wrong to do so.

> So why does then (for example) a Brwoser API "fit" SWT ... 

The Browser widgets provides some APIs to interact with the browser, internally using bindings, but doesn't expose the particular browser APIs in the SWT APIs and doesn't maintain the complete binding, just maintain a working SWT Browser API that is defined by SWT.

laeubi:
> I'm seeing a lot of public interfaces.

What don't makes them public API ... the package export is marked as internal and you will find numerous "public" methods in SWT code that have a hint that they are not public API only in the javadoc.

> This generated code seems to be directly derived from an API design; which qualifies as important IP work; so one needs to know whether deriving the API has constraints on the license. Assuming there are no legal issues is also too optimistic here; 

The generated code does not derive any work on behalf of the header files, the result only includes these portions that make up the ABI of the code that is method names and parameter types that are essential for interacting with the library, it is even not based on the source at all but using CLANG compiler that builds a model from it.

But if in doubt it would be very welcome to open an ticket with the EMO to clarify the situation, as it is not limited to this situation.

mbooth101:
> @mbooth101 already claimed that they actually do not change

Please reread. I actually said the *spec* hasn't changed _in a while._  I cannot say it won't change again in the future.

Another thing that makes me hesitant is that LWGL takes great effort to maximise performance, and intentionally avoids a lot of Java heap allocation costs when sending stuff across the native barrier. With LWJGL no data is copied when sending geometry to the GPU, due to use of direct buffers.

I would be very interested to see a perf comparison with a 1:1 panama mapping of the API.

mbooth101:
> As the discussion goes on, I'm growing the believe that offering this binding as API doesn't fit in SWT.

Same @mickaelistria -- I am convinced that a Panama-based version of GL/Vulkan bindings should be driven by the experts at e.g. the LWJGL project.

We at SWT should just provide the facility to render a GL widget or a Vulkan widget -- in the same way that GTK does with it's GdkGLContext and GdkVulkanContext objects.

I would love to see a SWT VulkanCanvas widget BTW -- no Vulkan support is a real limitation that SWT has.

laeubi:
> I would be very interested to see a perf comparison with a 1:1 panama mapping of the API.

Let me know if I can help with that, my OpenGL skills are just to limited to provide a meaningful benchmark here but it is quite easy to convert an JWGL example to a Panama one as the namings match 1:1, just constants are mapped as methods. I think project panama would also be interested in such measures.

> I would love to see a SWT VulkanCanvas widget BTW -- no Vulkan support is a real limitation that SWT has.

I planned to try that next that but it seems not very valuable if even the "simple" OpenGL is not deemed useful :-)

laeubi:
> I am convinced that a Panama-based version of GL/Vulkan bindings should be driven by the experts at e.g. the LWJGL project.

If they are willing to do so fine, as long I I do need just a singel jar with just the bindings as a maven dependency from central I would be fine ;-)

As said I don't claim my approach here as any IP relevant so the essential steps are described here:
https://github.com/laeubi/eclipse.platform.swt/blob/opengl_panama_gtk/bundles/org.eclipse.swt/Readme.Linux.md#generating-panama-bindings

and the working snippet is here:
https://github.com/laeubi/eclipse.platform.swt/blob/opengl_panama_gtk/examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet255.java

if there are any questions regarding building the stuff just let me know....

mbooth101:
Yikes I would definitely veto that snippet .... No-one should be using the fixed function pipeline in new code, it has be deprecated since 2009!! :-D

This is something else LWJGL does very well -- Exposing only the functions for the specific OpenGL profile available to the context. If the panama binding just exposes *everything* that is quite bad form.

iloveeclipse:
> What don't makes them public API

Well, `org.eclipse.swt.opengl.GL` **is** public API (and APi tooling also complains that the bundle version is not bumping adding this class). And `GL` itself, even if seem to be empty, extends `org.eclipse.swt.opengl.panama.glut_h` and so exposes all the public static methods & types used in method signatures.

I don't think we can guarantee here anything to clients so strictly speaking this `GL` can't be exported as public API too.

laeubi:
> Yikes I would definitely veto that snippet .... No-one should be using the fixed function pipeline in new code, 

This code is not new, I only adapted the imports, please open a PR to update them at best aslo enhancing instructions on how to run the snippets:

- https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet195.java
- https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet209.java

mbooth101:
Sure, I already suggested I could provide such contributions.

laeubi:
I'll close the PR for now but feel free to continue discussion, at least it seems to trigger some actions to bring SWT/OpenGL forward...

laeubi:
I poked around a bit and it seems the ultimate source might be https://github.com/KhronosGroup/OpenGL-Registry/tree/main/api/GL what is MIT licenced.

For this experiment I used my system headers and they seem to contain a lot of stuff not strictly required (therefore the large binding as I also included GLUT). ... so strictly speaking there is an offical API and even if there are license concerns MIT seems quite open, I'll probabbly give it a try to see what the outcome is with those.

Issue #441, opened 2022-10-19
win: forgotten redraw on resized columns


jukzi:
Testing the reording cache of columns i saw this effect, which i can also reproduce with old eclipse versions:
reorder the columns in the Problems View and then resize the columns with drag&drop. Eventually swt forgets to redraw/invalidate the resized area:
![image](https://user-images.githubusercontent.com/51790620/196746255-26b2f980-24cc-4063-8f3c-c7398e597fc7.png)

PR #576, opened 2023-03-08
Fix for #400 [GTK][Wayland] Non-standard mouse events during Drag&drop

### LINKED ISSUE #400: Drag and drop with different mouse event sequence on Ubuntu 22.04
**Describe the bug**
On the latest Manjaro (Gnome) version as well as on macOS the sequence of mouse/drag events is for a drag operation (on the same control):
- mouse down
- drag start
- drag enter
- drop over (repeats quickly)
- drag leave
- drop accept
- drag set data
- drop
- drag finished
For Ubuntu 22.04, however it is:
- drag start
- drag enter
- drop over (repeats quickly)
- drag leave
- drop accept
- drag set data
- drop
- drag finished
- mouse up
Notice the missing **mouse down** at the start and the new **mouse up** at the end.

**To Reproduce**
Please see the attached snippet that simply logs the events.
[MouseEventSequenceTest.java.zip](https://github.com/eclipse-platform/eclipse.platform.swt/files/9626311/MouseEventSequenceTest.java.zip)

**Expected behavior**
All Linux versions should produce the same sequence of events.

**Environment:**
- - [ ] All OS
- - [ ] Windows
- - [x] Linux
- - [ ] macOS 

Stock Ubuntu 22.04 (Gnome).

4. JRE/JDK version
Tried Java 17.
 
**Version since**
I've tried version 4.954.

No known (and expected) workaround.

SyntevoAlex:
Fixes #400
Please refer to individual commit messages.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Drag and Drop/gtk/org/eclipse/swt/dnd/DragSource.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Drag and Drop/gtk/org/eclipse/swt/dnd/DragSource.java
@@ -202,12 +202,12 @@ public DragSource(Control control, int style) {
 		}
 		control.setData(DND.DRAG_SOURCE_KEY, this);
 
+		// There's a native GTK snippet available, find 'Issue0400_WaylandDndEvents.cpp' in this repo.
+		// It may be helpful in understanding / debugging bugs.
 		OS.g_signal_connect(control.handle, OS.drag_begin, DragBegin.getAddress(), 0);
 		OS.g_signal_connect(control.handle, OS.drag_data_get, DragGetData.getAddress(), 0);
 		OS.g_signal_connect(control.handle, OS.drag_end, DragEnd.getAddress(), 0);
 		OS.g_signal_connect(control.handle, OS.drag_data_delete, DragDataDelete.getAddress(), 0);
-
-
 
 		controlListener = event -> {
 			if (event.type == SWT.Dispose) {
--- a/bundles/org.eclipse.swt/Eclipse SWT Drag and Drop/gtk/org/eclipse/swt/dnd/DragSource.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Drag and Drop/gtk/org/eclipse/swt/dnd/DragSource.java
@@ -502,13 +502,6 @@ void dragEnd(long widget, long context){
 			long selection = GTK.gtk_tree_view_get_selection (widget);
 			GTK.gtk_tree_selection_set_select_function(selection,0,0,0);
 		}
-
-		/*
-		 * send a mouse Up signal for >GTK3.14 as Wayland (support as of 3.14)
-		 * does not trigger a MouseUp/Mouse_release_event on DragEnd.
-		 * See Bug 510446.
-		 */
-		control.notifyListeners(SWT.MouseUp, event);
 	}
 	moveData = false;
 }
--- a/bundles/org.eclipse.swt/Eclipse SWT Drag and Drop/gtk/org/eclipse/swt/dnd/DropTarget.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Drag and Drop/gtk/org/eclipse/swt/dnd/DropTarget.java
@@ -170,6 +170,8 @@ public DropTarget(Control control, int style) {
 		}
 		control.setData(DND.DROP_TARGET_KEY, this);
 
+		// There's a native GTK snippet available, find 'Issue0400_WaylandDndEvents.cpp' in this repo.
+		// It may be helpful in understanding / debugging bugs.
 		drag_motion_handler = OS.g_signal_connect(control.handle, OS.drag_motion, Drag_Motion.getAddress(), 0);
 		drag_leave_handler = OS.g_signal_connect(control.handle, OS.drag_leave, Drag_Leave.getAddress(), 0);
 		drag_data_received_handler = OS.g_signal_connect(control.handle, OS.drag_data_received, Drag_Data_Received.getAddress(), 0);
--- a/bundles/org.eclipse.swt/Eclipse SWT PI/gtk/org/eclipse/swt/internal/gtk/OS.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT PI/gtk/org/eclipse/swt/internal/gtk/OS.java
@@ -269,7 +269,7 @@ public class OS extends C {
 	/** Signals */
 	public static final byte[] accel_closures_changed = ascii("accel-closures-changed");		// Gtk3,4
 	public static final byte[] activate = ascii("activate");	// ?
-	public static final byte[] angle_changed = ascii("angle_changed");	// Gtk3/4, Guesture related.
+	public static final byte[] angle_changed = ascii("angle-changed");	// Gtk3/4, Guesture related.
 	public static final byte[] backspace = ascii("backspace");
 	public static final byte[] begin = ascii("begin");
 	public static final byte[] button_press_event = ascii("button-press-event");
@@ -291,14 +291,14 @@ public class OS extends C {
 	public static final byte[] delete_text = ascii("delete-text");
 	public static final byte[] direction_changed = ascii("direction-changed");
 	public static final byte[] dpi_changed = ascii("notify::scale-factor");
-	public static final byte[] drag_begin = ascii("drag_begin");
-	public static final byte[] drag_data_delete = ascii("drag_data_delete");
-	public static final byte[] drag_data_get = ascii("drag_data_get");
-	public static final byte[] drag_data_received = ascii("drag_data_received");
-	public static final byte[] drag_drop = ascii("drag_drop");
-	public static final byte[] drag_end = ascii("drag_end");
-	public static final byte[] drag_leave = ascii("drag_leave");
-	public static final byte[] drag_motion = ascii("drag_motion");
+	public static final byte[] drag_begin = ascii("drag-begin");
+	public static final byte[] drag_data_delete = ascii("drag-data-delete");
+	public static final byte[] drag_data_get = ascii("drag-data-get");
+	public static final byte[] drag_data_received = ascii("drag-data-received");
+	public static final byte[] drag_drop = ascii("drag-drop");
+	public static final byte[] drag_end = ascii("drag-end");
+	public static final byte[] drag_leave = ascii("drag-leave");
+	public static final byte[] drag_motion = ascii("drag-motion");
 	public static final byte[] prepare = ascii("prepare");
 	public static final byte[] draw = ascii("draw");
 	public static final byte[] end = ascii("end");
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Control.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Control.java
@@ -3499,6 +3499,16 @@ long gtk_button_press_event (long widget, long event) {
 	return gtk_button_press_event (widget, event, true);
 }
 
+boolean wantDragDropDetection () {
+	// Drag&drop detection has its costs: when mouse button is pressed,
+	// mouse events are not reported until SWT can decide if drag&drop
+	// was triggered or not. For some applications, this could be a problem.
+	// Consider for example a graphical drawing application: it would expect
+	// to begin drawing as soon as mouse button is pressed.
+	// If app is interested in drag&drop, it will likely listen to `DragDetect`.
+	return hooks (SWT.DragDetect);
+}
+
 long gtk_button_press_event (long widget, long event, boolean sendMouseDown) {
 	mouseDown = true;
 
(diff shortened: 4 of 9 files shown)


mbooth101:
The commit _"Missing SWT.MouseDown before drag&drop"_ definitely fixes some click+drag weirdness I had on Wayland thanks :+1:

SyntevoAlex:
What kind of weirdness? I'd better know the details in case I'll need to change something later

mbooth101:
For example I have a simple event listener class to log events:

    private class MouseDragTracker implements DragDetectListener, MouseListener, MouseMoveListener {
    }

Event sequence before and after your patch:

| Before Your Patch  | After Your Patch |
| ------------- | ------------- |
| dragDetected | >**mouseDown**< |
| dragDetected | dragDetected |
| dragDetected | dragDetected |
| dragDetected | dragDetected |
| dragDetected | dragDetected |
| dragDetected | dragDetected |
| dragDetected | dragDetected |
| >**mouseDown**< | dragDetected |
| mouseUp | mouseUp |

I can remove a nasty hack in my application to catch the first dragDetect event...

SyntevoAlex:
Hmm that's pretty weird!
Before my patches, on Wayland, `mouseDown` did never arrive.
And `mouseUp` shall no longer arrive after all patches.
Can you track where did you get `mouseDown` from?

mbooth101:
I'm not sure what emits the event. Here's the reproducer I am using:

```

public class MyApp {
	private static class MouseDragTracker implements DragDetectListener, MouseListener {

		@Override
		public void dragDetected(DragDetectEvent e) {
			System.out.println("dragDetected");
		}
		@Override
		public void mouseDown(MouseEvent e) {
			System.out.println("mouseDown");
		}
		@Override
		public void mouseUp(MouseEvent e) {
			System.out.println("mouseUp");
		}
		@Override
		public void mouseDoubleClick(MouseEvent e) {
			System.out.println("mouseDouble");
		}
	}

		Canvas canvas = new Canvas(shell, SWT.NONE);
		MouseDragTracker l = new MouseDragTracker();
		canvas.addDragDetectListener(l);
		canvas.addMouseListener(l);

		shell.setSize(640, 480);
		shell.open();
		while (!shell.isDisposed()) {
				display.sleep();
			}
		}
		display.dispose();
	}
}
```

gtk3-3.24.35-1.fc36.x86_64
libwayland-server-1.21.0-1.fc36.x86_64

> Hmm that's pretty weird! Before my patches, on Wayland, `mouseDown` did never arrive. And `mouseUp` shall no longer arrive after all patches. Can you track where did you get `mouseDown` from?

If I get a mouseDown, I need to get a corresponding mouseUp, no?

SyntevoAlex:
> Here's the reproducer I am using:

Thanks, I'll give it a try.

> If I get a mouseDown, I need to get a corresponding mouseUp, no?

Not if drag&drop was started. I tested on Windows, macOS, Linux-GTK-X11 and Linux-GTK-Wayland (the last one with native snippet). All of them behave this way.

mbooth101:
> 

But mouseDown happens before you move the mouse....!? What if I press the mouse button and don't move it?  It's not a drag until you start moving the mouse, so I expect a mouseDown whatever happens. Surely only after I move the mouse does it become a drag....

SyntevoAlex:
Both before my patch and also after it, SWT temporarily puts `MouseDown` away until it decides about whether drag&drop has started. Before my patch, it didn't put it back. My patch fixes it.

mbooth101:
FWIW I see the same on Windows:

image

SyntevoAlex:
Yes, thanks for sharing your snippet, it was really worth looking at.

The thing about it, and I completely didn't expect it, is that you're listening to drag&drop events without actually starting drag&drop (because you didn't assign `DragSource` to your `Composite`).

This turns entire drag&drop logic upside down.
* You're seeing `SWT.MouseUp` event because, huh, you didn't start drag&drop, so you're seeing completely regular `SWT.MouseUp` as if there was no drag&drop stuff involved at all.
* You're seeing `SWT.MouseDown` event because SWT thinks that drag&drop didn't start, and it really didn't, so previously it had to give up its cached `SWT.MouseDown` event just before letting you see `SWT.MouseUp`. Now I moved the code and by a happy accident, it solves both problems - the one I saw and the one I didn't.

SyntevoAlex:
In shorter words, your snippet doesn't really have drag&drop.
So you're seeing `SWT.MouseUp` as normal, and due to my changes, `SWT.MouseDown` is now flushed earlier.

mbooth101:
> In shorter words, your snippet doesn't really have drag&drop.

Yes, I tried to be clear by saying "click+drag" and not "drag+drop" -- sorry for the confusion!

I just want to move things around on the canvas with my mouse :-)

SyntevoAlex:
With actual drag&drop however
* `SWT.MouseUp` shall not arrive (GTK doesn't send it, but it was emulated before, I removed that)
* `SWT,MouseDown` did not arrive (because SWT forgot to flush it) and I fixed that by flushing it just before `SWT.DragDetect`.

SyntevoAlex:
Abusing `SWT.DragDetect` to "just move things around with mouse" is something I wouldn't have imagined :)

mbooth101:
> Abusing `SWT.DragDetect` to "just move things around with mouse" is something I wouldn't have imagined :)

Well now I properly get a mouseDown event on Wayland, I don't have to abuse it so much :+1:

SyntevoAlex:
Right, thanks for the last clarification, I didn't quite get it before.

I now added a passage to commit message
```
One other scenario fixed in this commit is when there is no `DragSource`
involved at all. In this case, previously, `SWT.MouseDown` was cached
to detect drag&drop and then forgot later (just as in the first problem).
This caused problems to programs that merely wanted to use "press and
drag" action instead of drag&drop (imagine something as simple as
graphical editor with wants to drawing a line with mouse cursor when
 mouse is pressed). Now 'SWT.MouseDown' is correctly flushed quickly and
is not lost forever.
```

mbooth101 (approved):

SyntevoAlex:
@mbooth101 I noticed that without `canvas.addDragDetectListener(l)`, your snippet behaves well - SWT doesn't even try to detect drag&drop, avoiding the bug. You said that you only used it to avoid the bug, which is a contradiction. Could you please clarify?

SyntevoAlex:
Let's have this.
Hopefully the good deed goes unpunished this time and there will be no new regressions.

mbooth101:
Hi @SyntevoAlex -- Consider the case when a control *does* already have a drag detect listener, unrelated to my use-case. And then I have a new requirement need to track the mouse position while the mouse button is depressed:

Run this snippet using SWT before your patches, compare X11 versus Wayland behaviour:

```
public class MyApp {
	private static class MouseDragTracker implements MouseListener, MouseMoveListener {
		private boolean clickDrag = false;

		@Override
		public void mouseDown(MouseEvent e) {
			clickDrag = true;
			System.out.println("mouseDown");
		}
		@Override
		public void mouseUp(MouseEvent e) {
			clickDrag = false;
			System.out.println("mouseUp");
		}
		@Override
		public void mouseDoubleClick(MouseEvent e) {
			System.out.println("mouseDouble");
		}
		@Override
		public void mouseMove(MouseEvent e) {
			if (clickDrag)
				System.out.println("mouseMove " + e.x + "," + e.y);
		}
	}

		Canvas canvas = new Canvas(shell, SWT.NONE);
		MouseDragTracker l = new MouseDragTracker();
		canvas.addDragDetectListener(e -> System.out.println("An unrelated drag detect listener"));
		canvas.addMouseListener(l);
		canvas.addMouseMoveListener(l);

		shell.setSize(640, 480);
		shell.open();
		while (!shell.isDisposed()) {
				display.sleep();
			}
		}
		display.dispose();
	}
}
```

On Wayland you see `DragDetect` events instead of `MouseMove` events -- this is the original reason I had to abuse `DragDetect`

SyntevoAlex:
Thanks for the explanation.

In my patches, I fixed the problem of missing `MouseMove` events as well.
Your new snippet looks very similar to my snippet on tab Problem#2.

HannesWell:
Just in case you didn't knew this, if you prefix the issue number with a hash (like #400) in the commit message, GH can link the issue if you view the message in the browser. The same is if you add the full URL in the commit message.
The advantage of the latter approach is that you can open full URLs from EGit's commit view without any extra plugin or login.

merks:
Also, with the full URI, you can reference any issue in any organization to create back in the issue to the commit/PR...

SyntevoAlex:
Before GitHub, commits were uniformly named like "Bug 123456 - Title".
After the move, it's more like a chaos and everyone names commits differently.

I guess we need to start using something like
```
Issue #400 - [GTK][Wayland] Non-standard mouse events during Drag&drop
```

or 
```
#400 - [GTK][Wayland] Non-standard mouse events during Drag&drop
```

uniformly.

merks:
This was discussed at length during the migration.  Some people objected to needing issue for every commit because a PR is just like an issue and hence sufficient.  Also a comment ´fixes  issue-uriˋ ( I think) closes the issue when committed.  Mostly folks add the issue URI in the body of the commit, if there is one 

If you want to change that now, you should restart that discussion and be prepared for the fun

SyntevoAlex:
I'm not objecting to "PR is sufficient" part.

I'm just saying that if there is an issue, we better put it in the beginning of commit message, like it was before with bugs.

So instead of
`win: fix forgotten redraw on resized columns #441`
We should
`Issue #441 - win: fix forgotten redraw on resized columns`

Commit 8457de26 has no mention of Issue or PR anywhere, and instead of
`Win Display: use java map instead of OS for control map`
we should probably
`PR #428 - Win Display: use java map instead of OS for control map`

Instead of
`[Issue 367] Button Image is wrong size at hi-res scale`
we should
`Issue #367 - Button Image is wrong size at hi-res scale`

Instead of
```
Try to use Java 11 for native SWT build on Windows
See https://github.com/eclipse-platform/eclipse.platform.swt/issues/417
```
we should
```
Issue #417 - Try to use Java 11 for native SWT build on Windows
See https://github.com/eclipse-platform/eclipse.platform.swt/issues/417
```

etc etc.

SyntevoAlex:
I really, really don't like it when no PR or Issue is mentioned anywhere at all.

When it's mentioned somewhere, but in chaotic format, I strongly prefer that it's mentioned uniformly and in the beginning instead.

SyntevoAlex:
And there's a whole lot of recent commits that don't mention anything. This has to end.

laeubi:
> Commit [8457de2](https://github.com/eclipse-platform/eclipse.platform.swt/commit/8457de260355f8e0f0dfec32bd6a3fed8858348d) has no mention of Issue or PR anywhere, and instead of
> `Win Display: use java map instead of OS for control map`

Its all there...

its-all-there

> we should probably `PR #428 - Win Display: use java map instead of OS for control map`

Before you create the PR you won't know the number, so this would require to (force) push the change immediately after creating it, that do not sounds very useful, as you can see github automatically records the PR with the merge so I really don't see any advantage of that, writing more complete commit messages that include a description that could be understood without any PR/Issue number is much more useful.

> And there's a whole lot of recent commits that don't mention anything. 

Everyone can step up and review PRs, request changes for commit messages if desired, so the current state reflects the time people can/will invest here.

> This has to end.

or ... [we are all fired](https://www.youtube.com/watch?v=ejjtlnZOdgo)  :-)

SyntevoAlex:
I didn't know that GitHub shows that. I was looking at commit message in git directly.

I however don't think it's a good idea to rely on that.
Imagine that Jenkins did that before. Let's even assume hat everyone knew about that, even the new contributors. Now SWT has migrated off Bugzilla/Jenkins, and they are supposed to be turned off eventually. Old discussions will probably be dumped as text somewhere. So, how would someone match a commit to PR/Issue in this case?

I basically mean that a day will come and SWT will migrate from GitHub elsewhere, and mapping between commits will be lost.

And I very often have to read not only commit messages and discussions in Bug, but even study Gerrit revisions to try to figure when something was done this way and not another.

SyntevoAlex:
> Before you create the PR you won't know the number, so this would require to (force) push the change immediately after creating it

That is correct. It may be slightly inconvenient, but things need to be documented. The next step in this rant could be "it's inconvenient that after I changed code I also need to write a commit message"

SyntevoAlex:
> Everyone can step up and review PRs, request changes for commit messages if desired, so the current state reflects the time people can/will invest here.

I would say that this is because no decision has been made after the move and it plunged in chaos. It only takes a couple interactions to educate each committer. I could do that myself, but a decision is needed first.

HannesWell:
I would be glad if the PR's full URL would be appended to each commit added with a PR, for the reasons you mentioned. Similar to how it was done in Gerrit in it's late days:
https://github.com/eclipse-platform/eclipse.platform.swt/commit/0794f8be6e871ceb8248f32b5f71c2f80c1dcc5e

But I think that should be done automatically by some tooling when the PR is integrated. Then additionally reviewers could be adde, for example. This would probably mean that we cannot use the `Rebase/Squash-and-Merge` Buttons in the GH web interface anymore, but would for use some commands that trigger a GH workflow or app.
You can see that for example in the OpenJDK: https://github.com/openjdk/jdk/pulls

I suggested that during the mentioned discussions about the migration, but until now nobody implemented such tooling. Please correct me if I'm wrong, but I wouldn't expect much resistance if you have a corresponding solution. :)

However I don't think it will work well if it is mandated to do that manually. It is inconvenient, we would also have to ask each contributor, people will use different formats or will just forget it.

Btw, if you select Squash-And-Merge GH appends the PR Number to the suggested headline. But of course then you end up with one commit.

SyntevoAlex:
Thanks for your insight.
Indeed, appending things automatically makes a lot of sense.
Let me create an issue for that.

SyntevoAlex:
https://github.com/eclipse-platform/eclipse.platform.swt/issues/582

PR #588, opened 2023-03-14
Isseu 586: [macOS aarch64] NPE in org.eclipse.swt.widgets.FileDialog.open

tmssngr:
`NSSavePanel.savePanel()` and `NSOpenPanel.openPanel()` can return null, so we need to handle this case gracefully.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/FileDialog.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/FileDialog.java
@@ -352,6 +352,7 @@ public String open () {
 	fullPath = null;
 	if ((style & SWT.SAVE) != 0) {
 		NSSavePanel savePanel = NSSavePanel.savePanel();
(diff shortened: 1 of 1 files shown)


lshanmug (approved):
Null check was added in the past to DirectoryDialog, but FileDialog was missed.

lshanmug:
Thanks @tmssngr for the patch!

kysmith-csg:
I'm encountering an issue with this change. Returning `null` from this method is unsafe because `FileDialog.open` expects it to never be `null`.

https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/bundles/org.eclipse.swt/Eclipse%20SWT/cocoa/org/eclipse/swt/widgets/FileDialog.java#L358

`return openDialog().orElse(null)`

and the resulting `NullPointerException` when we return `null` inside `openDialog()`.

This should be changed to 

```java
NSSavePanel savePanel = NSSavePanel.savePanel();
if (savePanel == null) return Optional.ofNullable(null);
```

and same for OPEN.

tmssngr:
@kysmith-csg Please feel free to provide a patch!

PR #590, opened 2023-03-15
[Linux] Fix natives build when NATIVE_STATS enabled

mbooth101:
When native stats collection is enabled, i.e. by passing `-DNATIVE_STATS` to the compiler then the natives fail to build:

```
os_stats.c:820: error: "NATIVE_FUNCTION_COUNT" redefined [-Werror]
  820 | #define NATIVE_FUNCTION_COUNT sizeof(GTK_nativeFunctionNames) /
      sizeof(char*)
os_stats.c:240: note: this is the location of the previous definition
  240 | #define NATIVE_FUNCTION_COUNT sizeof(GDK_nativeFunctionNames) /
      sizeof(char*)
```

This change causes the generator to prefix the NATIVE_FUNCTION_COUNT macro name with the class whose native functions are being counted and includes regenerated *_stats.c files.

The change also fixes up the NativeStats tool to know about the discrete GTK libraries and allows native stats tool to be activated by environment variable instead of requiring user to manually change the makefile.

### DIFF
--- a/bundles/org.eclipse.swt.tools/JNI Generation/org/eclipse/swt/tools/internal/StatsGenerator.java
+++ b/bundles/org.eclipse.swt.tools/JNI Generation/org/eclipse/swt/tools/internal/StatsGenerator.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2004, 2013 IBM Corporation and others.
+ * Copyright (c) 2004, 2023 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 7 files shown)


akurtakov:
Why is NATIVE_STATS still needed? What is your use case for it? If there is not a strong one maybe we would be better dropping them entirely?

mbooth101:
> Why is NATIVE_STATS still needed? What is your use case for it? If there is not a strong one maybe we would be better dropping them entirely?

I knew you would ask this @akurtakov :-) TBH I don't have a use real use-case I just noticed it was broken as I was poking around and it was easily fixed.

I am not opposed to deleting it all.

akurtakov:
Let's give a week or so for others to jump in and show a use case and if not - we can get rid of it.

HannesWell:
> I guess no-one has tried building SWT with native stats enabled on Linux for a while, it looks like it has not worked for years.

You could now do that as part of this PR's verification build. You just have to adjust either the pipeline or the relevant build script for linxu (I don't know where to place that exactly).

akurtakov:
@mbooth101 Sorry, I totally forgot about this one. I'll work on getting rid of NATIVE_STATS now.

akurtakov:
Removal on NATIVE_STATS handled via https://github.com/eclipse-platform/eclipse.platform.swt/issues/640

Issue #591, opened 2023-03-17
Test_org_eclipse_swt_widgets_Button.test_setOrientation fails on Mac


iloveeclipse:
See https://download.eclipse.org/eclipse/downloads/drops4/I20230316-1800/testresults/html/org.eclipse.swt.tests_ep428I-unit-mac64-java17_macosx.cocoa.x86_64_17.html

I can't remember it failing before, seem some recent change caused it.

```
expected:<67108864> but was:<33554432>

java.lang.AssertionError: expected:<67108864> but was:<33554432>
at org.junit.Assert.fail(Assert.java:89)
at org.junit.Assert.failNotEquals(Assert.java:835)
at org.junit.Assert.assertEquals(Assert.java:647)
at org.junit.Assert.assertEquals(Assert.java:633)
at org.eclipse.swt.tests.junit.Test_org_eclipse_swt_widgets_Button.test_setOrientation(Test_org_eclipse_swt_widgets_Button.java:192)
```

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [ ] Windows
- - [ ] Linux
- - [x] macOS

Phillipus:
`Button#setOrientation` (actually `Control#setOrientation`) does nothing. Orientation should be set in the constructor:

`Button button = new Button(shell, SWT.PUSH | SWT.RIGHT_TO_LEFT);`

iloveeclipse:
@Phillipus : thanks for looking into.

@mbooth101 : would you please fix the new test to not run on Mac, see comment above?

mbooth101:
I will take a look -- I did not notice any MacOS test failures on the original PR: https://github.com/eclipse-platform/eclipse.platform.swt/actions/runs/4366256928/jobs/7636013659?pr=555

Are tests not actually run there?

iloveeclipse:
> It seems like the GitHub job is always green even when there is test failure!!!! Can that be fixed?

Before we can fix an issue, it should be 1) found & 2) reported. We are at 1) now :-)
Please create an issue.

mbooth101:
I will do so, but I meant more like "is it within the realm of possibility?" and some limitation of GitHub :-)

mbooth101:
> 
> **Obviously** it is :-) 

[...]

PR #595, opened 2023-03-17
Issue #591: Fix test failure on MacOS

mbooth101:
The test was introduced as part of dfaed5e but unfortunately the method under test has no effect on MacOS, so don't run the test on that platform.

For history see also:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=261270


iloveeclipse:
Thanks Matt.

PR #596, opened 2023-03-18
Explicitly disable API tools execution

laeubi:
Currently SWT do not use the API tools in the build verification, but enabling the profile (-Papi-check) leads to multiple errors and then fails the build.

To allow using this build to unify the verification build infrastructure, SWT can simply declare that even if the profiles are active they are skipped.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -39,6 +39,8 @@
     <sonar.c.file.suffixes>-</sonar.c.file.suffixes>
     <sonar.cpp.file.suffixes>-</sonar.cpp.file.suffixes>
     <sonar.objc.file.suffixes>-</sonar.objc.file.suffixes>
+    <skipAPIDescription>true</skipAPIDescription>
+    <skipAPIAnalysis>true</skipAPIAnalysis>
   </properties>
 
   <!-- 


HannesWell:
I'm not sure if disabling it for all modules isn't too much. Maybe it should only be disabled for the o.e.swt Plugin?

laeubi:
It is currently disabled for all modules already as we do not use the `-Papi-check` profile anywhere.

HannesWell:
> It is currently disabled for all modules already as we do not use the `-Papi-check` profile anywhere.

Indeed. I wonder why I did notice that when reworking the SWT build pipeline. 😅

Maybe we should investigate for which artifacts the api-check works, maybe only for the fragments? Because in general I think that would also be valuable for SWT.

laeubi:
At least currently that fails due to so build structure, but maybe later :-)

So I think we can merge this for now and go on.

pom.xml
```diff
@@ -39,6 +39,8 @@
     <sonar.c.file.suffixes>-</sonar.c.file.suffixes>
     <sonar.cpp.file.suffixes>-</sonar.cpp.file.suffixes>
     <sonar.objc.file.suffixes>-</sonar.objc.file.suffixes>
+    <skipAPIDescription>true</skipAPIDescription>
```
HannesWell, line 42: After thinking about this again and checking the current artifacts in the I-build repo, I think we must not disable the api-description generation. Skipping the analysis is fine, but the description is currently generated and the o.e.swt Plugin as well as its platform specific fragments (I only checked windows) contain an `.api_description` file.

I think they have to be retained, so that part should be reverted and the api-description generation has to be fixed in another way.

laeubi: The api-description works fine, so we can simply set this to false here, or delete the line ...

HannesWell: Then better remove it or one day someone will wonder what this is about and will spend some time to find out that is actually unused.

HannesWell: Done in #598.

Issue #594, opened 2023-03-17
GitHub PR build results always report green


mbooth101:
In #591 it was reported that a new failing test was introduced.

However it was not spotted because the Github PR continuous integration always report green _even when a change introduces a new test failure!_

E.g. See this build result: https://github.com/eclipse-platform/eclipse.platform.swt/actions/runs/4366256928/jobs/7636013659?pr=555

You have to read the build logs very closely to spot there was a new MacOS test failure.

Is it possible to have such cases report a "failure" or "unstable" result as Jenkins might do? It would be great to catch these before merge.

iloveeclipse:
Matt, can you push some dummy PR with a trivial change to see if SWT test in question is actually **executed** at all on Mac at github action?

laeubi:
Usually there is a github action that publish the unit test results see:

https://github.com/eclipse-platform/eclipse.platform.swt/actions/workflows/junit.yml

I don't know why this currently do not kick in I can take a look the next days where I already planned to update the jobs for platform repositories.

mbooth101:
You can see it in any recent PR, either the one I linked to above, or this from a couple days ago, for example:

https://github.com/eclipse-platform/eclipse.platform.swt/actions/runs/4414773547/jobs/7736908402?pr=588#step:7:1607

iloveeclipse:
> I can take a look the next days where I already planned to update the jobs for platform repositories.

Thanks.

mbooth101:
For a while the Win32 build was failing frequently, and there are no test results for any of these runs,:

https://github.com/eclipse-platform/eclipse.platform.swt/actions/runs/4368161788
https://github.com/eclipse-platform/eclipse.platform.swt/actions/runs/4369053751
https://github.com/eclipse-platform/eclipse.platform.swt/actions/runs/4368159636

@laeubi so it's maybe it's possible that if one platform fails to build, then it never bothers to collate the test results for the other successful platforms. It looks like this is why I missed the new test failure.

laeubi:
The primary goal should be to fix the build (make test more robust, disable or delete them) otherwise the confidence in these tests will degrade and we will end up in such situation like this where it is only accidentally noticed that the test are not run/reported...

laeubi:
@iloveeclipse @mbooth101 
At least here I see test results published:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/596#issuecomment-1474809070

iloveeclipse:
[...]

PR #597, opened 2023-03-18
Fail matrix job if there are test-failures as well as show delta

laeubi:
Currently the matrix job succeeds if there are test-failures, this can be confusing for people. Also we can modernize the usage of the Test results check to show a +/- delta compared to current master build.

Fix https://github.com/eclipse-platform/eclipse.platform.swt/issues/594
FYI @iloveeclipse @mbooth101


laeubi:
Build failures are as requested...

Issue #346, opened 2022-08-25
[GTK4] gtk_shortcut_controller_remove_shortcut crash when using Eclipse


joel-majano:
After opening a file in GTK4 and moving around in the IDE a crash happens in gtk_shortcut_controller_remove_shortcut.

Here is the log: [hs_err_pid174210.log](https://github.com/eclipse-platform/eclipse.platform.swt/files/9427344/hs_err_pid174210.log)

Environment: GTK 4.6.5, Fedora 36, Wayland

Tracked in #228

mbooth101:
Hmm, I believe I reproduced this, with the following snippet:

```
	public static void main(String[] args) {
		Display display = new Display();
		Shell shell = new Shell(display);
		shell.setText("Snippet 29");
		Menu bar = new Menu(shell, SWT.BAR);
		shell.setMenuBar(bar);
		MenuItem fileItem = new MenuItem(bar, SWT.CASCADE);
		fileItem.setText("&File");
		Menu submenu = new Menu(shell, SWT.DROP_DOWN);
		fileItem.setMenu(submenu);
		MenuItem addItem = new MenuItem(submenu, SWT.PUSH);
		MenuItem removeItem = new MenuItem(submenu, SWT.PUSH);
		addItem.addListener(SWT.Selection, e -> {
			System.out.println("Adding Accelerators");
			addItem.setText("&Add Accelerators\tCtrl+A");
			addItem.setAccelerator(SWT.MOD1 | 'A');
			removeItem.setText("&Remove Accelerators\tCtrl+R");
			removeItem.setAccelerator(SWT.MOD1 | 'R');
		});
		removeItem.addListener(SWT.Selection, e -> {
			System.out.println("Removing Accelerators");
			addItem.setText("&Add Accelerators");
			addItem.setAccelerator(0);
			removeItem.setText("&Remove Accelerators");
			removeItem.setAccelerator(0);
		});
		addItem.setText("&Add Accelerators\tCtrl+A");
		addItem.setAccelerator(SWT.MOD1 | 'A');
		removeItem.setText("&Remove Accelerators\tCtrl+R");
		removeItem.setAccelerator(SWT.MOD1 | 'R');
		shell.setSize(200, 200);
		shell.open();
		while (!shell.isDisposed()) {
			if (!display.readAndDispatch())
				display.sleep();
		}
		display.dispose();
	}

```

I always get GTK-CRITICALs like this:

```
[...]

PR #599, opened 2023-03-20
Issue #346: [GTK4] gtk_shortcut_controller_remove_shortcut crash

mbooth101:
Null the handle reference after removing the shortcut from the menu item, to prevent it from trying to remove the shortcut more than once.

Also includes the manual test snippet.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/MenuItem.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/MenuItem.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2019 IBM Corporation and others.
+ * Copyright (c) 2000, 2023 IBM Corporation and others.
  *
  * This program and the accompanying materials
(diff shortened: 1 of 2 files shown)


akurtakov:
@mbooth101 What's the story with many failing tests ?

mbooth101:
@akurtakov There are no new test failures compared with current master HEAD: https://github.com/eclipse-platform/eclipse.platform.swt/actions/runs/4456702886/jobs/7827421993

Since 153a6ce test failures are no longer ignored

akurtakov:
So is anyone working to fix these or we should still look at JIPP results only? The patch looks good to me so feel free to push it depending on how tests work is planned.

laeubi:
If test only work on JIPP, we should ship each SWT user an own JIPP instance ... or there is a problem that should be fixed.

akurtakov:
> If test only work on JIPP, we should ship each SWT user an own JIPP instance ... or there is a problem that should be fixed.

I'm all for someone fixing the issues :+1:

mbooth101:
I took a look at the most severe one (`Invalid thread access` causing test error) in #600 

If I find the time, I will try to look at the others too.

Issue #593, opened 2023-03-17
Copy formatted text adds extra line(s) and format


jukzi:
**Describe the bug**
STRG+C a text without linebreaks from an editor
STRG+V into microsoft word or Teams => pastes copied text + one or even two newlines (NOK!)
STRG+V  into notepad++ => pastes only the copied text (OK)
STRG+SHIFT+V into microsoft word or Teams => pastes only the copied text (OK)

It's also not helpfull that the whole format is copied at all since the format depends on the current selection -> the copied text is shown ontop of light blue background, only because it was the selected line. 

**To Reproduce**
copy "main" from "public static void main(String[] args)" and paste it with format:
![image](https://user-images.githubusercontent.com/51790620/225929008-5146d2a0-f2e1-401a-8a8b-7d1c41256a42.png)

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [x] Windows
- - [x] Linux
- - [?] macOS 

**Version since**
first recognized with 2022-12

**Workaround**
paste without format (STRG+SHIFT+V)

jukzi:
happens also when all CTRL+C in the key bindings are disabled.

iloveeclipse:
I assume this is the effect of #292 ?

jukzi:
Extra newline for multi-line copy was solved.  Copy a "line fragment" still has issue.

drapichrust:
@jukzi  is it in anyway related to an issue that I see:

using simple text file on plain eclipse for java

I put some text to the clipboard via ctrl + c

I'm creating multi caret selection by using "multi selection down relative to anchor selection" (alt + j in my case) multiple times. Now depending on how many carets I have if I click ctrl + v the pasted text is multiplied by the amount of carets. 2 carets will give me double insertion per caret, 4 will give me 4 per caret (16).

What is weird is that if I make selection first via ctrl + j and then copy text and paste I only get one insertion per caret.

PR #609, opened 2023-03-29
No extra paragraph in the HTML clipboard

mihnita:
Fixing issue https://github.com/eclipse-platform/eclipse.platform.swt/issues/593

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/StyledTextWriterBase.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/StyledTextWriterBase.java
@@ -147,6 +147,12 @@ abstract class StyledTextWriterBase extends TextWriter {
 		String atLineEnd = writeLineStart(lineBackground, indent, verticalIndent, alignment, justify);
 
 		int endOffset = startOffset + super.getCharCount();
+
+		// We are already at the end, should not write an empty paragraph
+		if (lineOffset >= endOffset) {
+			return;
+		}
+
 		int lineEndOffset = Math.min(lineLength, endOffset - lineOffset);
 
 		int outTextLen = 0; // collect the length of the text we output (unescaped)


mihnita:
image

mihnita:
Note: I am aware that if you select the text until the end of the last line but without the final newline (stop right after the `}`) there is still a difference between RTF and HTML.

The RTF would look like in the screenshot but no newline after 🡸 (`...}🡸`), while the HTML will have the newline (the way the screenshot shows for the "fixed" version)

There is not much I can do for that case.

No end of paragraph means that last line would not behave like a normal paragraph, so no alignment, justification, etc.
(all the features possible for a block element).
And I think that is worse than the extra line (breaks more than it fixes)
One can always delete the extra newline, if really bothered by it.

jukzi:
Thanks for working on it.

When i copy a text inside a line (like "my" from "hello my world") and html-insert into Word it still produces a new line which it does not for RTF.  I guess when someone copies text without a single line break it should not result in block elements.
For multiline copy the paragraph height is still wrong after pasting into Word.

I would not bother at all if the paste would not automatically prefer html over RTF - the default paste still looks wrong.

mihnita:
> When i copy a text inside a line (like "my" from "hello my world") and html-insert into Word it still produces a new line which it does not for RTF. I guess when someone copies text without a single line break it should not result in block elements.

I understand, and I am very split about that.
To support that kind of functionality I would need to drop the `div`s, and that would break quite a few things that now actually work better than the RTF.
And yes, I understand that Word can do it when copy-ing from Word and then paste as HTML.
But as explained, that "Word HTML" is not quite standard, there are quote a few hacks there.

When deciding to do what I did here I compared the html produced by several apps.

> For multiline copy the paragraph height is still wrong after pasting into Word.

I cannot reproduce that. As you can see in the screenshot the paragraph height is the same between rtf and html.

> I would not bother at all if the paste would not automatically prefer html over RTF - the default paste still looks wrong.

Yes, but Eclipse can't control that, it is a MS Word decision.
I can maybe add an option to disable putting HTML in the clipboard?

But I wonder how many people would really be bothered by this...

M.

jukzi:
You would not need to drop divs. just replace them by span. 

For the line hight i found that when i start an empty document on my computer the paragraph is already set to have a gap. When i then paste the RTF that setting is overwritten by 0 and looks fine (as formated as in eclipse) - but when pasting html the setting is sometimes(depends on what i copy?) overwritten with some value (or "automatic"). So it should be possible to  provide a style with 0 margin - but i don't know which html style attribute word would take. i tried margin, mso-margin-top-alt, margin-top without success :-(

What does work with word is to use divs for format and an empty paragraph "<p></p>" like a "<br>" => 0 height and a new paragraph.

mihnita:
> You would not need to drop divs. just replace them by span.

Sorry, I don't understand this.

A span can't replace a div, because a span is an inline HTML element, divs are block elements.
So it can't do the things that a block can do, like horizontal align justify, etc.
In MS Word concepts this spans are character styles and divs are paragraph styles.

> i tried margin, mso-margin-top-alt, margin-top without success :-(

Those are not standard html attributes, and not portable.

---

Can we though keep that disscution in the bug itself, and try to deal with the PR at hand?

Do you agree it is an improvement and accept it? (see screenshot with before / after)
And then we can debate in the bug if we consider it fixed / acceptable as is / still a major problem?

mihnita:
> a) copy some text that is only part of the line
> b) copy multiple lines
> c) There is a corner case as a mixture of both where the user selects selects only part of multiple lines, starting in the middle of a line and ending in a line. Or even a block selection. But that is so rare that i don't care about it.

OK, I  tried a), and it is actually (much) worse in RTF than in HTML. The RTF drops all color info, except background.
See case 5 in the screenshot below.

Case b) is case 1 in the screenshot, and works the same.

Case c) (mixed) corresponds to cases 2, 3, 4 in the screenshot. And case 2 (start in middle, end with full line) is the same.
Cases 3 and 4 (last line partially selected) results in an extra line in HTML
(already in my initial note, saying that _"I am aware that if you select the text until the end of the last line but without the final newline ..."_)

To summarize:
  case a) is (a lot) better with HTML
  case b) is the same
  case c) (that you don't care about) is a bit worse in 2/3 of the cases. And extra line that is easy to delete.

Animation capture showing I didn't cheat here:
https://mihai-nita.net/eclipse_test/MSWord_SelectionTypes.gif 

The final status in MS Word:
image

mihnita:
> I would prefer a single commit for all the HTML related issues.

I'm afraid I don't have time for that, especially (as explained) doing that will break other things.

Because (for example) making the last line `<span>` instead of `<p>`  changes it from block element to inline element.
That is how HTML works, and I don't plan to waste much more time trying to work around that.

Fighting against standards is not a good way to spend the time.

And in fact that is also how MS Word works. There is only so much one can fake in a char style to "simulate" what a paragraph style can do.

---

I think that with this PR it is close enough to RTF (and in some cases better).
And extra newline is easy to delete, if someone is bothered by it.
Fixing alignment or justification is more work than that.

I've spent a lot of time in the first two iterations (before this bug) to make it work everywhere and in all kind of uses (beyond just the code editor, but general swt use).
Inspecting clipboard html produced by all kind of apps, and pasting html in all kind of apps.

So by now I kind of know what it works and what not.
If you don't believe me, please give it a try.
And I will test it outside the MS Word "monoculture" and I will show that it breaks other things that now work in HTML and don't work in RTF.
It will make the HTML as bad as the RTF.

jukzi:
To say the HTML looks better is subjective. I would clearly prefer to not see the dark theme background when pasting into a bright theme document. Especially it should not change the format of the surrounding text. But that is what is does and also visible in your screenshots: Always the preexisting "HTML:" is changed by the pasting.
The typical example is to copy a class name only and paste it after a text like "the error is in:". With current HTML that messes up the format of "the error is in:". i don't see any reason why it should be good to change the format of the surrounding block if you only copy a word.
image

mihnita:
> To say the HTML looks better is subjective.

My comment was _"case a) is (a lot) better with HTML"_

Look at the screenshot, "case a)" is "mid-start, mid-end (one line)", case 5 in the screenshot.

The RTF has background color, same as the HTML.
But lost all the syntax coloring.
If the HTML is not objectively better in that case, I don't know what is.

---

> Especially it should not change the format of the surrounding text

That is a Word bug.
When one pastes a block kind of content there is no reason to merge the existing text into it.

jukzi:
> But lost all the syntax coloring.

We where used to it over decades and it was fine. Now it pastes a backgroundcolor that is only there to highlight the current selection. That is clearly unintended and the wrong color.

> That is a Word bug.

It's an incompatibility between your copy and Word. Another example that we use: Atlassian's jira is a webtool and we can't paste a single word without linebreaks there anymore too. If i do and then continue to write the rest of the text gets that unintented backgroundcolor too. Example:
Eclipse IDE:
image
Jira:
image

jukzi:
Or when pasting between existing text into Jira still 2 unintended linebreaks:
image

mihnita:
That example does not work in Firefox.
Select from "on" in the first line to "in the div" in the second line.

And it does not really matter, the html is incorrect, and each browser will do "a best effort"
Also, the two lines (_"red on yellow"_ and _"We are still in the div element."_) should not be directly in a `<div>`, should be paragraphs.

Each line of text, all the way to the explicit newline (even if it wraps) is a paragraph.
A paragraph can be aligned. Try making the first line right aligned, and the last one center aligned.
You can't, because they are not block elements.
The way to do that is to wrap them a block element. The correct one is `<p>`
And if you do that, and select from the middle of line 1 to the middle of line 3, MS Word will ignore the alignment when pasted.

See:
https://www.w3.org/TR/2011/WD-html5-author-20110809/the-div-element.html
_"mark up semantics common to a group of consecutive elements"_
_"On the other hand, [div](https://www.w3.org/TR/2011/WD-html5-author-20110809/the-div-element.html#the-div-element) elements can be useful for stylistic purposes or to wrap multiple paragraphs within a section that are all to be annotated in a similar way."_

> We where used to it over decades and it was fine. Now it pastes a backgroundcolor that is only there to highlight the current selection. That is clearly unintended and the wrong color.

There is no difference in how RTF works. It pastes background color now, same as before. That didn't change.
The fact that you didn't notice and nobody complained until now shows that in fact nobody needed that.

---

Try copy paste-ing from Visual Studio Code, for curiosity.

---

Sorry, I am getting tired of this, and I suppose you do too.

So let's try to see if we can approach this in a more constructive way.
Instead of you trying to force me to to fix things a certain way (drop the divs), let's try to focus on the observed behavior.

You shared the use cases that you consider critical, and I agreed: full block of code (with first and last lines of code selected), and part of one line.

The first one (the block) was the origin of this bug (and extra line), and it is fixed by this PR.

The second one (fragment of one line) is very similar between RTF and HTML, in that they both paste with background.
(In my opinion) the RTF is worse, because the syntax highlighting is not preserved.
And syntax highlighting is probably why someone would want to copy paste from a code editor (like Eclipse) to begin with.

> I would clearly prefer to not see the dark theme background when pasting into a bright theme document

I hear you, but I do not.
I was annoyed many-many times by the fact that I keep my editor in dark mode, and when I paste the background is not preserved.
Yes, one can make the argument that a bright theme document should not "inherit" that dark background.
But the trouble is that for Eclipse in dark mode the text is light.
So I paste white text (and light colors in general) in a light document. Which makes the text invisible.
You can see in the screenshot above that the punctuation is invisible

I can work to fix that special case (fragment of one line) to not add a newline when pasting from HTML.
But that is a different fix.

---

Might also help to share "the values" used to judge if something is better or worse.

What I valued is:
- The results of the paste should be as close as possible to the way things look in the `StyledText` widget. And that should include as many properties as possible (indent, hanging, alignment, justification, RTL / LTR control, bullets, margins, and so on). 
- That should work across platforms and browsers
- This is not only about Eclipse, but about SWT in general. So it should support properties that are not necessarily exercised by the current code editors

---

If someone really likes the old RTF clipboard style (which didn't change) we can consider adding an option for the formats to put in the clipboard, so that a user can say "I don't want HTML in the clipboard, ever"

That is not common, in most cases the app serving the content puts in the clipboard as many formats as possible, and the client chooses what is considered the best match, with the options of a paste special. The way MS Word behaves.

mihnita:
TLDR: this PR fixes the bug as filed (the extra newline) in block mode (the most common use case)
Let's land it.

We can then debate what to do about a fragment of a line that is intended to be pasted in a middle of a paragraph.
That has an extra paragraph in HTML, and no colors in RTF.
What do we want the behavior to be?
And do we want syntax colors for RTF?

We can also check with the people in Eclipse that decide on UX.
Should we add options for the users to configure what to put in the clipboard?
Maybe an option to preserve the background color or not.

laeubi (approved):
If that improves anything (and even the smallest part) please merge this as of today so it can go into M1 and maybe improve/discuss further things in a different PR.

The change it self looks quite focused and clear enough.

mihnita:
Thank you.
M.

jukzi:
I agree it would be inconsistent to paste a dark-mode color into a light document. It's better to always paste color+backgroundcolor together. Just it's wrong which color is copied as background color. The copied backgroundcolor should not be influenced by the current selection - but is. 
You should not feel forced to use <span>. i only investigates the possibilities which would work.  If you find any other working pattern feel totally free to use it, just please fix the inline-copy additional new-line think somehow.

mihnita:
@jukzi

I will take a look at "line fragment" from HTML, to eliminate the extra newline, and maybe also (hopefully) how to not "contaminate" the text on the left of the paste. That is really not a block element behavior, so there is no need for `<div>` or `<p>`, simple `<span>`(s) will do.

Since I tinker that, should I also take a look at the RTF single-line fragment?
Right now the newlines are fine (none are added).
But the background is dark, and there are no syntax colors
(see "mid-start, mid-end, one line" (case 5) in screenshot here https://github.com/eclipse-platform/eclipse.platform.swt/pull/609#issuecomment-1491169740)

What would you prefer:
1. Don't touch it. It's been like this for a long time
2. Bring in the syntax colors and keep the background dark
    Pros: it is what it does now, we get light text on light background, is consistent with the HTML
    Cons: not consistent with block RTF (which uses light background (and light text))
4. Bring in the syntax colors and make the background light
   Pros: consistent with the "block RTF"
   Cons: light text on light background.

Note: I am not familiar with the RTF part. But willing to investigate.

jukzi:
I have no preferences about RTF, since i am using light mode in Eclipse and target document as well and have no problems with the RTF. Also if HTML works and is the preferred source it is unlikely i will ever (manually) insert as RTF ever again. I am not aware of any target that prefers RTF over HTML, so just keeping RTF as is would be probably make least problems. Otherwise bring up a use case for someone who would really use RTF in future.

Issue #616, opened 2023-04-06
Inconsistent behavior of Composite.setFocus across operating systems


HeikoKlare:
**Describe the bug**
The behavior of `setFocus` of a `Composite` is different between Linux and the other operating systems. While the [Linux implementation](https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/bundles/org.eclipse.swt/Eclipse%20SWT/gtk/org/eclipse/swt/widgets/Composite.java#L1672) only tries to set focus on children when they are visible, the [implementation for Windows](https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/bundles/org.eclipse.swt/Eclipse%20SWT/win32/org/eclipse/swt/widgets/Composite.java#L1078-L1080) and [the one for MacOS](https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/bundles/org.eclipse.swt/Eclipse%20SWT/cocoa/org/eclipse/swt/widgets/Composite.java#L1127) call `setFocus` on their children even if they are invisible. This can lead to unexpected behavior such as views processing `setFocus` (with potentially complex functionality) while they are not visible. An example can be seen in the  `TextMergeViewer`, which is recalculated every time one of the compared documents is saved, even if it is not visible. This is due to all views being temporarily disabled while saving, which in turn triggers a call of `setFocus`: eclipse-platform/eclipse.platform#121

Of course, a view should not process complex logic in `setFocus` when it is not visible. However, the framework could at least avoid that views that are not even visible perform actions for which the user does not understand where they belong, as they do not see the view.

**To Reproduce**
Taken from eclipse-platform/eclipse.platform#121:
1.) Do a change to a larger text and displays its delta via the compare editor
2.) Ensure the compare editor is not visible
3.) Perform changes in other files and save -> Frequently you see "Calculating deltas" in the job monitor
4.) Change the file itself in a regular editor -> -> Frequently you see "Calculating deltas" in the job monitor

**Expected behavior**
I would expect that propagating `setFocus` to children is not necessary when they are invisible. Thus I propose to add a check for whether children are visible before trying to set focus to them for Windows and MacOS, as already implemented for Linux. At least, I would expect the behavior to be consistent across systems.

**Screenshots**
![image](https://user-images.githubusercontent.com/755472/230432010-48dda786-cd4a-434b-aa19-1200f94a8b99.png)

**Environment:**
1. Select the platform(s) on which the behavior is seen:
[...]

PR #617, opened 2023-04-06
Avoid setting focus on invisible children of Composite #616

HeikoKlare:
In the Windows and MacOS implementations of a `Composite`, the `setFocus` method tries to set focus on children even if they are not visible and will thus never receive focus. Depending on the realization of `setFocus` on the children, functionality may be executed that confuses the user as it belongs to a view they do not see (see #616).
The Linux implementation already validates whether a child is visible before setting focus on it. This change makes the implementations consistent across operating systems and ensures the consistent behavior with according regression tests.

https://github.com/eclipse-platform/eclipse.platform.swt/blob/d65eb73da26084e65c3090e7ede0892b3016a112/bundles/org.eclipse.swt/Eclipse%20SWT/gtk/org/eclipse/swt/widgets/Composite.java#L1667-L1675

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Composite.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Composite.java
@@ -1124,7 +1124,7 @@ public boolean setFocus () {
 	checkWidget ();
 	Control [] children = _getChildren ();
 	for (int i= 0; i < children.length; i++) {
-		if (children [i].setFocus ()) return true;
+		if (children [i].getVisible() && children [i].setFocus ()) return true;
 	}
 	return super.setFocus ();
 }
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Composite.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Composite.java
@@ -1073,10 +1073,10 @@ public boolean setFocus () {
 	checkWidget ();
 	Control [] children = _getChildren ();
 	for (Control child : children) {
-		if (child.setRadioFocus (false)) return true;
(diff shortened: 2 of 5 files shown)


mickaelistria:
I'm not familiar with the GitHub actions verification. Is it acceptable that they are failing?
I think the overall idea is very correct, patch looks good enough, and would be easy to revert if necessary, so this is safe to merge. So if there is no red flag, let's merge ASAP to get feedback in next builds.

mickaelistria (approved):

HeikoKlare:
> Sadly there are very few maintainers available to review changes on this project currently.

Thanks for having a look at this anyway! I will try to support the project as best as I can.

> I'm not familiar with the GitHub actions verification. Is it acceptable that they are failing?

Three of the tests in the Matrix build seem to fail on every build. The fourth failing test is added with this PR and only fails on Windows systems. I will have a look at that tomorrow and update the PR with a fixed tests. Please wait for that before merging.

```
1 out of 3 runs failed: test_setFocus_withVisibleAndInvisibleChild (org.eclipse.swt.tests.junit.Test_org_eclipse_swt_browser_Browser)
artifacts/test-results-win32.win32.x86_64-java17/eclipse.platform.swt/tests/org.eclipse.swt.tests/target/surefire-reports/TEST-org.eclipse.swt.tests.junit.Test_org_eclipse_swt_browser_Browser.xml
```

HeikoKlare:
I have now disabled the new tests regarding setting focus for the `Browser` as a specific `Composite`, like it was the case for the other `setFocus` tests on browser due to the different OS having inconsistent behavior w.r.t. setting focus.
The performed Jenkins and GitHub Actions test runs do not show any regressions.

According to your proposal, @mickaelistria, it would be okay for me to merge this.

mickaelistria:
thanks

vogella:
Thanks @HeikoKlare I assume that this will also be a little bit faster, if children are not visible.

HeikoKlare:
I hope so too. Since the code is executed quite often, it may have some positive performance impact.

Issue #620, opened 2023-04-08
CTabFolder should allow to listen on tab count changes


iloveeclipse:
Coming from https://github.com/eclipse-platform/eclipse.platform.ui/issues/708.

Currently there is no way for `CTabFolder2Listener` to detect a change in `CTabFolder` if a tab is added or removed.  For the issue above we want know if the onboarding area should be rendered or not if the first editor is opened or the last editor is closed.

Note, there is a `close` event that can be used to detect tab closing, however it also allows listeners to **prevent** closing of tabs - so in fact, it may be not correct for our use case.

So proposal is to add a simple `itemsCount` event that would be triggered if the tabs **really** added or removed from `CTabFolder`.

I will push a PR.

PR #621, opened 2023-04-08
CTabFolder should allow to listen on tab count changes

iloveeclipse:
- added CTabFolder2Listener.itemsCount(CTabFolderEvent) API
- added CTabFolder2Listener.itemsCountAdapter(Consumer<CTabFolderEvent>) API
- updated CTabFolder to notify CTabFolder2Listener on creating/removing items
- added unit test
- bumped minor version segment for new API added

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/620

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
@@ -744,6 +744,14 @@ Image createButtonImage(Display display, int button) {
 	image = new Image(display, new AutoScaleImageDataProvider(display, imageData, DPIUtil.getDeviceZoom()));
 	return image;
 }
+
+private void notifyItemCountChange() {
+	CTabFolderEvent e = new CTabFolderEvent(this);
+	for (CTabFolder2Listener listener : folderListeners) {
+		listener.itemsCount(e);
+	}
+}
+
 void createItem (CTabItem item, int index) {
 	if (0 > index || index > getItemCount ())SWT.error (SWT.ERROR_INVALID_RANGE);
 	item.parent = this;
@@ -769,6 +777,7 @@ void createItem (CTabItem item, int index) {
 	} else {
 		updateFolder(REDRAW_TABS);
 	}
+	notifyItemCountChange();
 }
 void destroyItem (CTabItem item) {
 	if (inDispose) return;
@@ -789,6 +798,7 @@ void destroyItem (CTabItem item) {
 		updateButtons();
 		setButtonBounds();
 		redraw();
+		notifyItemCountChange();
 		return;
 	}
 
@@ -820,6 +830,7 @@ void destroyItem (CTabItem item) {
 
 	requestLayout();
 	updateFolder(UPDATE_TAB_HEIGHT | REDRAW_TABS);
+	notifyItemCountChange();
 }
 
 /**
@@ -1631,7 +1642,6 @@ void onKeyDown (Event event) {
 						Rectangle chevronRect = chevronItem.getBounds();
 						chevronRect = event.display.map(chevronTb, this, chevronRect);
 						CTabFolderEvent e = new CTabFolderEvent(this);
-						e.widget = this;
 						e.time = event.time;
 						e.x = chevronRect.x;
 						e.y = chevronRect.y;
@@ -1953,7 +1963,6 @@ void onMouse(Event event) {
 					redraw(item.closeRect.x, item.closeRect.y, item.closeRect.width, item.closeRect.height, false);
 					if (!selected) return;
 					CTabFolderEvent e = new CTabFolderEvent(this);
-					e.widget = this;
 					e.time = event.time;
 					e.item = item;
 					e.doit = true;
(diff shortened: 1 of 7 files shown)


iloveeclipse:
@SyntevoAlex : any objections regarding the new API? I feel not OK to add API without anyone else looked at it.
The example IDE code that would use it is in https://github.com/eclipse-platform/eclipse.platform.ui/pull/710/files#diff-c065602fc1c20bba7ec8effcd0f47262b6cc7e9a32a9c9dc4b39bdfc993c5909R713

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
```diff
@@ -769,6 +778,7 @@ void createItem (CTabItem item, int index) {
 	} else {
 		updateFolder(REDRAW_TABS);
 	}
+	notifyItemCountChange();
 }
 void destroyItem (CTabItem item) {
 	if (inDispose) return;
```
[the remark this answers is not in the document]
SyntevoAlex: Also, I'm not sure what `if (index == -1) return` is for below. How is it even possible for TabItem to not be present in parent?

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
```diff
@@ -744,6 +744,15 @@ Image createButtonImage(Display display, int button) {
 	image = new Image(display, new AutoScaleImageDataProvider(display, imageData, DPIUtil.getDeviceZoom()));
 	return image;
 }
+
+private void notifyItemCountChange() {
+	CTabFolderEvent e = new CTabFolderEvent(this);
+	e.widget = CTabFolder.this;
```
SyntevoAlex, line 750: You're not populating `display` (here, you're not alone) and `time`. I have no opinion if you should.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder2Listener.java
```diff
@@ -115,6 +115,34 @@ public interface CTabFolder2Listener extends SWTEventListener {
  */
 public void showList(CTabFolderEvent event);
 
+/**
+ * Sent when the tab items count changes
+ * 
+ * @param event from observed tab folder
+ * @since 3.124
+ */
+public default void itemsCount(CTabFolderEvent event) {
```
SyntevoAlex, line 124: I would rather call it `itemsCountChanged` so that it doesn't sound like a verb.

SyntevoAlex (commented):
My comments are mostly nitpicks, so feel free to ignore.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_custom_CTabFolder.java
```diff
@@ -143,6 +147,48 @@ private void createTabFolder(List<String> events, int numItems) {
 	ctabFolder.setSelection(ctabFolder.getItem(0));
 }
 
+@Test
+public void test_ItemCountListsners() {
```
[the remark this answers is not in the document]
iloveeclipse: Done

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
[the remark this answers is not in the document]
iloveeclipse: > Do you think events are not needed when TabFolder is being disposed? I have no opinion.

I think no. As a client I'm interesting if tabs are added / removed in a **living** folder instance. How many items will be / will be not in the dead instance, no one should be interested.

> Also, I'm not sure what if (index == -1) return is for below. How is it even possible for TabItem to not be present in parent?

I guess because `org.eclipse.swt.custom.CTabFolder.items` is accessible from outside of CTabFolder (horrible!) and anyone in the package can change items.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder2Listener.java
iloveeclipse: Funny enough, that was my first name. But looking at other events, I see they all are verbs: `maximize`, `close`, etc (instead of `maximized`, `closed` etc). To be consistent, I would keep the "verb" form here, because it's already there.

SyntevoAlex: OK

iloveeclipse:
Thanks Alex for review.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
```diff
@@ -744,6 +744,14 @@ Image createButtonImage(Display display, int button) {
 	image = new Image(display, new AutoScaleImageDataProvider(display, imageData, DPIUtil.getDeviceZoom()));
 	return image;
 }
+
+private void notifyItemCountChange() {
+	CTabFolderEvent e = new CTabFolderEvent(this);
```
laeubi, line 749: You could init the event lazy in the loop, so if there are no listeners you don't create a uselsess object that is instantly garbage collected.
Alternatively, if folderListeners.isEmpty() one might early exit this method.

PR #629, opened 2023-04-12
[Jenkins-Build] Use temurin-jdk to provide JVM headers for natives build

HannesWell:
The JDK referenced by the JAVA_HOME variable during the SWT natives build stage only provides the headers used for the JNI code.
But since the Java JNI code headers should be equivalent for all platforms it should be sufficient to use the same JVM for all platforms, which simplifies the Jenkins pipeline.


HannesWell:
This does not work since tool(type:'jdk', name:'temurin-jdk11-latest') returns the path of the jdk tool path at the controller (which is a linux machine), that are not valid/existing at the machines used to build the SWT natives.

Issue #254, opened 2022-07-13
[macOS] Possible NPE in Combo.mouseDown


tmssngr:
**Describe the bug**
We've received a couple of thread dumps from users similar to this one from SmartGit 21.2.3 (which uses SWT based on commit a274d744):
```
java.lang.NullPointerException
	at org.eclipse.swt.widgets.Combo.mouseDown(SourceFile:1158)
	at org.eclipse.swt.widgets.Display.windowProc(SourceFile:6149)
	at org.eclipse.swt.internal.cocoa.OS.objc_msgSendSuper
	at org.eclipse.swt.widgets.Widget.callSuper(SourceFile:235)
	at org.eclipse.swt.widgets.Widget.windowSendEvent(SourceFile:2150)
	at org.eclipse.swt.widgets.Shell.windowSendEvent(SourceFile:2487)
	at org.eclipse.swt.widgets.Display.windowProc(SourceFile:6273)
	at org.eclipse.swt.internal.cocoa.OS.objc_msgSendSuper
	at org.eclipse.swt.widgets.Widget.callSuper(SourceFile:235)
	at org.eclipse.swt.widgets.Widget.mouseDownSuper(SourceFile:1146)
	at org.eclipse.swt.widgets.Widget.mouseDown(SourceFile:1138)
	at org.eclipse.swt.widgets.Control.mouseDown(SourceFile:2631)
	at org.eclipse.swt.widgets.Combo.mouseDown(SourceFile:1156)
	at org.eclipse.swt.widgets.Display.windowProc(SourceFile:6149)
	at org.eclipse.swt.internal.cocoa.OS.objc_msgSendSuper
	at org.eclipse.swt.widgets.Widget.callSuper(SourceFile:235)
	at org.eclipse.swt.widgets.Widget.windowSendEvent(SourceFile:2150)
	at org.eclipse.swt.widgets.Shell.windowSendEvent(SourceFile:2487)
	at org.eclipse.swt.widgets.Display.windowProc(SourceFile:6273)
	at org.eclipse.swt.internal.cocoa.OS.objc_msgSendSuper
	at org.eclipse.swt.widgets.Display.applicationSendEvent(SourceFile:5517)
	at org.eclipse.swt.widgets.Display.applicationProc(SourceFile:5660)
	at org.eclipse.swt.internal.cocoa.OS.objc_msgSend
	at org.eclipse.swt.internal.cocoa.NSApplication.sendEvent(SourceFile:117)
[...]

PR #630, opened 2023-04-13
#254 - [macOS] Possible NPE in Combo.mouseDown

tmssngr:
better cache the used display before it gets unset

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Combo.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Combo.java
@@ -1144,17 +1144,19 @@ void menuDidClose(long id, long sel, long menu) {
 
 @Override
 void mouseDown(long id, long sel, long theEvent) {
+	// If the control is diposed in below 'super.mouseDown()', both fields
+	// 'view' and 'display' will be nulled. Hence keep their references:
+	final Display display = this.display;
+	final NSView view = this.view;
+
+	display.sendPreExternalEventDispatchEvent();
 	// If this is a combo box with an editor field and the control is disposed
 	// while the view's cell editor is open we crash while tearing down the
 	// popup window. Fix is to retain the view before letting Cocoa track
 	// the mouse events.
-
-	display.sendPreExternalEventDispatchEvent();
-	// 'view' will be cleared if disposed during the mouseDown so cache it.
-	NSView viewCopy = view;
-	viewCopy.retain();
+	view.retain();
 	super.mouseDown(id, sel, theEvent);
-	viewCopy.release();
+	view.release();
 	display.sendPostExternalEventDispatchEvent();
 }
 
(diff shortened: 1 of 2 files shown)


bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Combo.java
```diff
@@ -1146,15 +1146,16 @@ void menuDidClose(long id, long sel, long menu) {
...
-	// popup window. Fix is to retain the view before letting Cocoa track
-	// the mouse events.
+	// popup window. Fix is to retain the display and view before letting Cocoa
+	// track the mouse events.
+
+	final Display display = this.display;
+	final NSView view = this.view;
 
 	display.sendPreExternalEventDispatchEvent();
-	// 'view' will be cleared if disposed during the mouseDown so cache it.
```
lshanmug, line 1153: Let's retain the comment and variable name as they don't need any changes.

lshanmug (changes requested):

SyntevoAlex:
I'd like to see a unit test here.
Most likely, if you dispose `Combo` in its `SWT.MouseDown` listener, it will reproduce reliably.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Combo.java
SyntevoAlex: I would say that this is a matter of taste. And these a better left to author, if there's no obvious harm.

tmssngr:
I've tried to add a unit test, but failed, because I was not able to send mouse events correctly, so this part of the code would be executed. See https://github.com/eclipse-platform/eclipse.platform.swt/compare/master...syntevo:eclipse.platform.swt:feature/254-macos-npe-mousedown-failed-test. I've also tried with `display.post`, but then other unit tests broke.

SyntevoAlex:
If creating an automated test is difficult, then a manual snippet is just fine.

tmssngr:
Is it OK now?

SyntevoAlex:
The code comments were misplaced in your patch.
I fixed that and added commits to your branch. Please review and squash.

SyntevoAlex:
The code comments were misplaced in your patch.

The old top-level comment was in fact explaining why `view.retain()` is needed.
You edited it incorrectly, `display` has nothing to do with it.

The `so cache it` code comment got misplaced after you moved variable declaration.

I fixed that and added commits to your branch. Please review and squash.

SyntevoAlex:
Regarding the failing `test_setFocus_withVisibleAndInvisibleChild` test, see https://github.com/eclipse-platform/eclipse.platform.swt/issues/616#issuecomment-1616669535

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Combo.java
SyntevoAlex: Not sure if reply is expected here. I'll resolve the conversation to merge PR.

SyntevoAlex:
In master now.

elsazac:
Verified on :

Eclipse SDK
Version: 2023-09 (4.29)
Build id: I20230703-1800
OS: Mac OS X, v.13.4.1, aarch64 / cocoa
Java vendor: Eclipse Adoptium
Java runtime version: 20+36
Java version: 20

Issue #631, opened 2023-04-13
[Mac] GC text foreground color not updated if string length is 1 and transparent background set


Phillipus:
**Describe the bug**
This bug originates from https://github.com/eclipse/gef-classic/issues/173

Due to a change in `GC#drawText` on Mac introduced in https://github.com/eclipse-platform/eclipse.platform.swt/commit/2a2325b0699615424fc557f8d6db2190dd304048 drawing a single character onto the same `GC` instance with different foreground colors is not drawn correctly.

This affects `Draw2d` figures that draw the same single character with different foreground colors because Draw2d uses the same `GC` instance for all figures in a canvas.

**To Reproduce**
Run  this snippet:

```java
import org.eclipse.swt.SWT;
import org.eclipse.swt.graphics.Color;
import org.eclipse.swt.layout.FillLayout;
import org.eclipse.swt.widgets.Composite;
import org.eclipse.swt.widgets.Display;
import org.eclipse.swt.widgets.Shell;

public class SWTSnippet {

    public static void main(String[] args) {
        Display display = new Display();
        Shell shell = new Shell(display);
        shell.setLayout(new FillLayout());
        shell.setSize(200, 180);

        String text = "A";
        Color red = new Color(255, 0, 0);
        Color blue = new Color(0, 0, 255); 

        Composite comp = new Composite(shell, SWT.None);
        
        comp.addPaintListener(e -> {
            e.gc.setForeground(red);
            e.gc.drawText(text, 0, 0, true);
            
            e.gc.setForeground(blue);
            e.gc.drawText(text, 20, 0, true);
        });
        
        shell.open();

        while(!shell.isDisposed()) {
            if(!display.readAndDispatch()) {
                display.sleep();
            }
        }
    }
}
```

**Expected behavior**
The first character should be red and the second one blue. Because of the caching in `GC` the foreground color is not updated.

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [ ] Windows
- - [ ] Linux
- - [x] macOS 

2. Additional OS info (e.g. OS version, Linux Desktop, etc)
Mac

3. JRE/JDK version
Java 17
 
**Version since**
Since https://github.com/eclipse-platform/eclipse.platform.swt/commit/2a2325b0699615424fc557f8d6db2190dd304048

Phillipus:
Alternatively, the current foreground color could be used as part of the caching key string:

```java
[...]

PR #632, opened 2023-04-13
Revert single character text caching on Mac GC

Phillipus:
- See https://github.com/eclipse-platform/eclipse.platform.swt/issues/631
- This caching was only being used in one case when drawing text of one character on a transparent background
- This means that if the GC is used to draw more than one text string with different foreground colors the cached color is used instead of the new one
- This reverts commit 2a2325b0699615424fc557f8d6db2190dd304048.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/GC.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/GC.java
@@ -13,8 +13,6 @@
  *******************************************************************************/
 package org.eclipse.swt.graphics;
 
-import java.util.*;
-
 import org.eclipse.swt.*;
 import org.eclipse.swt.internal.*;
 import org.eclipse.swt.internal.cocoa.*;
@@ -76,7 +74,6 @@ public final class GC extends Resource {
 
 	Drawable drawable;
 	GCData data;
-	GCTextData.Cache textDataCache = new GCTextData.Cache(20);
 
 	CGPathElement element;
 	int count, typeCount;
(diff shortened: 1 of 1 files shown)


lshanmug:
@Phillipus Will review this early next week. AFAIR the change was made to improve the slow scrolling on Mac text editors. We don't want to bring back the issue.

lshanmug:
@elsazac Can you pls try the PR?

Phillipus:
@lshanmug Thanks. I also made another suggestion in https://github.com/eclipse-platform/eclipse.platform.swt/issues/631#issuecomment-1508377944

vogella:
@sratz could you review / test? This affects GEF-classic and therefore would be nice to get fixed.

sratz:
> I assume @sratz you have a Mac. Let me know if that is not true...

No, I don't have one.

However, according to
https://bugs.eclipse.org/bugs/show_bug.cgi?id=366471#c272
and
https://git.eclipse.org/r/c/platform/eclipse.platform.swt/+/92329/
the change 2a2325b0699615424fc557f8d6db2190dd304048 helps significantly when
'Show whitespace characters' is checked, due to all the `·` characters being drawn.

Phillipus:
I'm testing on a Mac Mini M1 and a 2020 MacBook Air Intel with Monterey. I have "show whitespace characters" on and I'm on a 4k monitor. I can't detect any lag when scrolling a large text file (using GC.Java to test).

The original bug https://bugs.eclipse.org/bugs/show_bug.cgi?id=366471 was reported in 2011, and things haven gotten a lot faster on Mac since.

sratz:
No objections then to revert the change.

We can do some tests once the I-Builds are available to see if we notice any performance regressions.

vogella:
AFAICS the test failures are not related. Merging.

Issue #626, opened 2023-04-12
SWT build still requires Java 11


iloveeclipse:
Nashorn JavaScript Engine was removed in Java 15, see https://openjdk.org/jeps/372

However, SWT build still uses it in ant file:
See 
- https://github.com/eclipse-platform/eclipse.platform.swt/blob/c26d1ff0248c9191a81e6cd5a94607bf77c4f740/bundles/org.eclipse.swt/buildSWT.xml#L371
- https://github.com/eclipse-platform/eclipse.platform.swt/blob/c26d1ff0248c9191a81e6cd5a94607bf77c4f740/bundles/org.eclipse.swt/buildSWT.xml#L385
- https://github.com/eclipse-platform/eclipse.platform.swt/blob/c26d1ff0248c9191a81e6cd5a94607bf77c4f740/bundles/org.eclipse.swt/buildSWT.xml#L415

This has to be solved if we want to get rid of Java 11 in SWT.

Once that is done, we also can change `Jenkinsfile`

Beside supporting ant execution on 11, Jenkinsfile uses different JDK 11 versions also in other places, see [this comment](https://github.com/eclipse-platform/eclipse.platform.swt/pull/627#issuecomment-1505780265).
and following places:
- https://github.com/eclipse-platform/eclipse.platform.swt/blob/04678a28607971268382a53f4b85c0b176b02add/Jenkinsfile#L143
- https://github.com/eclipse-platform/eclipse.platform.swt/blob/04678a28607971268382a53f4b85c0b176b02add/Jenkinsfile#L146
- https://github.com/eclipse-platform/eclipse.platform.swt/blob/04678a28607971268382a53f4b85c0b176b02add/Jenkinsfile#L148

mickaelistria:
It's possible to use the `java` task instead; but we need to rework things a bit to avoid dereferencing the project in the code, but instead pass the properties we're willing to process. eg
```xml
<project default="blah">
	<target name="blah">
		<property name="tags"><![CDATA[
			a
			b
			c]]></property>
		<echo file="dummyTags.java">
			<![CDATA[
				public class lastTag { public static void main(String... args) {
					var tags = args[args.length - 1].split("\n");
					System.out.println(tags[tags.length - 1].strip());
				}}
			]]>
		</echo>
		<java fork="true" outputproperty="swt_tag" sourcefile="dummyTags.java">
			<arg value="${tags}"/>
		</java>
		<echo>${swt_tag}</echo>
[...]

Issue #625, opened 2023-04-12
SWT should require Java 17 runtime in 4.28


iloveeclipse:
### Discussed in https://github.com/eclipse-platform/eclipse.platform.swt/discussions/623

I've just noticed that SWT is still requires at least Java 11, even if the almost everything in platform (except equinox) is moved to Java 17.

Note that the [Eclipse project Plan for 4.28](https://www.eclipse.org/projects/project-plan.php?planurl=https://www.eclipse.org/eclipse/development/plans/eclipse_project_plan_4_28.xml) only sees Java 17 as supported platform.

We should switch SWT to require Java 17 runtime for 4.28 M2.

I will provide a PR.

iloveeclipse:
@HannesWell : For ppc64le build, which JVM is to use here:
https://github.com/eclipse-platform/eclipse.platform.swt/blob/c26d1ff0248c9191a81e6cd5a94607bf77c4f740/Jenkinsfile#L146

iloveeclipse:
Currently I'm stuck on some strange Jenkins build results in https://github.com/eclipse-platform/eclipse.platform.swt/pull/627. If anyone has ideas, please comment there.

HannesWell:
> @HannesWell : For ppc64le build, which JVM is to use here:
> 
> https://github.com/eclipse-platform/eclipse.platform.swt/blob/c26d1ff0248c9191a81e6cd5a94607bf77c4f740/Jenkinsfile#L146

That's a JVM/JDK installed by the EFDN infra team and you want to use another one you have to open a ticket and ask them to add another one tot the build machines dedicated for ppc64le (see https://ci.eclipse.org/releng/label/swt.natives-gtk.linux.ppc64le/), which is at the moment only https://ci.eclipse.org/releng/label/swt.natives-gtk.linux.ppc64le/.

But that and the other JDKs in that part assigned to the JAVA_HOME env variable only provide the headers used for the JNI code.
Since (I assume that) the Java JNI code headers should be equivalent over the Java versions it should be sufficient the leave them untouched. The SWT natives build against the Java-11 headers work with java-17 JRE already and I assume they will do it even when the bundles require java-17.
Nevertheless on the long run those JVMs should be updated. Looking at the win32/library/build.bat it seems that for windows only the SWT_JAVA_HOME is honored, while for other Platforms JAVA_HOME is used:
[...]

PR #633, opened 2023-04-13
Build SWT-natives against Java-17 JDK header files

HannesWell:
As described in https://github.com/eclipse-platform/eclipse.platform.swt/issues/626#issuecomment-1507719226 the SWT-natives are currently build against the JDK header files for different Java versions raging from 11 to 19. Because SWT now requires Java-17 (https://github.com/eclipse-platform/eclipse.platform.swt/issues/625), they all should be build against Java-17 headers.

This PR aims to achieve that and at the same time fixes the inconsistencies regarding the headers/libs in the natives build described in https://github.com/eclipse-platform/eclipse.platform.swt/issues/626#issuecomment-1507719226. 

- Use the C header files and shared native libraries from minimal JustJ JDKs when building the SWT native binaries on specialized build machines.
- Always specify the location of the JDK providing the headers/libs using the 'SWT_JAVA_HOME' environment variable. This allows to use another JDK as the one specified in the global JAVA_HOME variable.
- When building the natives for individual platforms via Maven (with the `build-individual-platform` profile active and `-Dnative=<platform>` specified) set the value of `SWT_JAVA_HOME` to the location of the running JDK (which must of course be present). If one wants to use a custom JDK to build the SWT natives against, one can set set for example the Maven CLI argument
`-DSWT_JAVA_HOME=<path-to-desired-JDK>`

Because in the CI and for Maven builds that build natives now always have a JDK specified the auto-search algorithms to find one is now obsolete and removed from the scripts.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/626 (at least the part regarding the Java-11 headers).

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -32,6 +32,11 @@ spec:
 	}
 }
 
+/** Returns the download URL of the JDK against whoose C headers (in the 'include/' folder) and native libaries the SWT natives are compiled.*/
+def getNativeJdkUrl(String os, String arch){ // To update the used JDK version update the URL template below
+	return "https://download.eclipse.org/justj/jres/17/downloads/20230428_1804/org.eclipse.justj.openjdk.hotspot.jre.minimal.stripped-17.0.7-${os}-${arch}.tar.gz"
+}
+
 pipeline {
 	options {
 		skipDefaultCheckout() // Specialiced checkout is performed below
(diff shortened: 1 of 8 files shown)


HannesWell:
Created https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/2957 to ask the Infra-team to add corresponding JDKs to the SWT-natives build machines.
Maybe we are lucky and they are already there. As far as I know SWT is not the only user of those machines, so I believe the chances are not too low, but I don't know the list of available JDKs in these specific agents.

Jenkinsfile
```diff
@@ -91,6 +96,21 @@ pipeline {
 						ant -f eclipse.platform.swt/bundles/org.eclipse.swt/buildSWT.xml new_build_with_create_file -DTAG=HEAD
 					'''
 				}
+				script { // Stash jdk-headers for transfer to the natives build machines
+					if ( params.forceNativeBuilds || (fileExists ('tmp/build_changed.txt') && fileExists ('tmp/natives_changed.txt')) ){
+						dir('jdkHeaderDownloads') {
+							for(os in [ 'macosx', 'linux', 'win32' ]) {
+								sh """
+									curl ${getJDKHeaderProviderURL(os)} -o jdk.tar.gz
+									tar -xzf jdk.tar.gz ${ os == 'win32' ? './include/': 'include/'}
```
HannesWell, line 105: @merks can you tell why in the Windows JDKs available at `https://download.eclipse.org/justj/jres/17/downloads/20230204_0657/` there is always an extra folder named `.` in the tar.gz root which contains the jdk content?
For Linux and Macos the jdk is always contained directly in the zip's root.

merks: I think it's produced the same way on all platforms by this:

https://github.com/eclipse-justj/justj/blob/400371be17b1e5b4819ef1d977cb4133d7bbf277/releng/org.eclipse.justj.releng/build-jre.sh#L373-L376

It should be logged here:

https://ci.eclipse.org/justj/job/build-jres/248/

But there I see the following as if the * is expanded to "."  which seems rather strange behavior.
```
+ tar -czf ../org.eclipse.justj.openjdk.hotspot.jre.full-17.0.6-win32-x86_64.tar.gz .
```
It should be possible to run this script locally and reproduce such a problem.   But I won't have time today or tomorrow to investigate that.

merks: Found time for a quick look...  

It's this that is repackaging after signing all the executables and dlls:

https://github.com/eclipse-justj/justj/blob/400371be17b1e5b4819ef1d977cb4133d7bbf277/releng/org.eclipse.justj.releng/Jenkinsfile#L513

I thing that should be '*' and not '.' to avoid this problem.

merks: I've kicked off a build to build new JREs that should fix this problem...

BTW, this file contains relative URIs of all the latest Java 17 JREs:

https://download.eclipse.org/justj/jres/17/downloads/latest/justj.manifest

So in principle one could use that to avoid hard coding the specific URL of a current latest...

HannesWell: > I've kicked off a build to build new JREs that should fix this problem...

Thanks a lot! The windows tar.gz at https://download.eclipse.org/justj/www/download.eclipse.org.php?file=jres/17/downloads/20230423_0633 look perfectly fine now!



> I've kicked off a build to build new JREs that should fix this problem...
> 
> BTW, this file contains relative URIs of all the latest Java 17 JREs:
> 
> https://download.eclipse.org/justj/jres/17/downloads/latest/justj.manifest
> 
> So in principle one could use that to avoid hard coding the specific URL of a current latest...

Thanks for that hint. I was already wondering if there is a latest URL from which one can just download them directly.
But since actually only the C header files provided by a JDK (in the `include`) are of interest in this case and I believe that they don't change often I was fine with how it is.
But I'll have a look so that always the latest tar.gz are used.

HannesWell:
It turns out not only the header files but also some shared libraries (at least jawt) are used in the SWT natives build. As far as I have understood it, some objects are extracted from it. 

But that's not too bad, because I have changed the workflow already to simply download a minimal JDK from JustJ for each supported platform each time the natives are really build. This has the advantage that it is ensured that the natives use the same JDK version and provider for all platforms and we don't have to ask the infra team when we want to update the used JDK version.
The the minimal jdks are between 20~30MB in size and take about a second to download on the build machines (they seem to have quite fast internet there or they run in the same network as the file-server hosting the justJ download site).
At the moment only the jdk for Linux ppc64le is not available in JustJ and therefore the libs are downloaded from adoptium (and the linux x86_64 headers are used which are identical). But I have already asked to provide Linux ppc64le in JustJ so later that will be hopefully available too:
- https://github.com/eclipse-justj/justj/issues/12

@iloveeclipse do you have any objection about this approach? Can you tell if somebody else is interested in this?

HannesWell:
For me this PR is ready.
I just updated the PR description to reflect the final state.
@iloveeclipse or someone else, do you want to review this?

akurtakov:
Even if this one causes some issues it's a move in the right direction so we would better merge it now and deal with possible consequences.

HannesWell:
> Even if this one causes some issues it's a move in the right direction so we would better merge it now and deal with possible consequences.

As far as I can tell and as far as I have tested it, everything should be fine. But eventually we will only know for sure if we try it, therefore I'm merging this one.
Thanks Aleks for the review.

PR #638, opened 2023-04-20
Enable build-individual-bundles profile in maven.config

akurtakov:
Makes mvn calls work out of the box


vogella:
Nice!

HannesWell:
Created a follow-up to remove the now unnecessary references to that profile:
https://github.com/eclipse-platform/eclipse.platform.swt/pull/642

PR #643, opened 2023-04-25
[GH-Workflow] Remove now unnecessary toolchains.xml

HannesWell:
Since SWT has moved to Java-17 using Maven toolchains is not necessary anymore and should be removed to not let the build appear more complex than it is.


laeubi (approved):
The toolchains are actually automatically generated by the `setup-java` action.

akurtakov:
Thank you guys for improving and simplifying the releng so much !

HannesWell:
> The toolchains are actually automatically generated by the `setup-java` action.

Yes I know. I tried to use that recently for Xtext, but somehow the generated toolchain only contained the last JDK. But now I cannot reproduce this.

PR #646, opened 2023-04-26
Remove unused import in tests

akurtakov:


SyntevoAlex:
Right, sorry. IDEA only considers this as low-priority issue.

akurtakov:
> Right, sorry. IDEA only considers this as low-priority issue.

Isn't there smth like "save action" to auto do it ?

SyntevoAlex:
It would be best if GitHub build action also checked that to reject such patches

mickaelistria:
The .settings can probably be set to report such unused import as errors, and tycho would read the settings and fail the compilation in such case

PR #648, opened 2023-04-27
Fix regression in 631

Phillipus:
This fixes the regression in [#631.](https://github.com/eclipse-platform/eclipse.platform.swt/issues/631#issuecomment-1524127112)

There are two commits to make things clearer:

1. Revert the revert of 9e996e825dbcaf2bc95a954cadadd24729b2ab8e
2. Use a unique cache key for each text character with different alpha, font and foreground color attributes

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/GC.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/GC.java
@@ -123,15 +123,20 @@ public final class GC extends Resource {
 		private NSRange range;
 
 		/**
+		 * Key object to use for Cache consisting of text string, alpho, font and foreground color
+		 */
+		private static record Key(String string, int alpha, long font, double[] fgColor) {}
+
+		/**
 		 * Simple {@link GCTextData} cache limited to a set number of items.
 		 */
 		private static class Cache {
 			private final int cacheSize;
-			private final Map<String, GCTextData> cache = new LinkedHashMap<String, GCTextData>() {
+			private final Map<Key, GCTextData> cache = new LinkedHashMap<>() {
 				private static final long serialVersionUID = 1L;
 
 				@Override
-				protected boolean removeEldestEntry(Map.Entry<String, GCTextData> eldest) {
+				protected boolean removeEldestEntry(Map.Entry<Key, GCTextData> eldest) {
 					if (size() >= cacheSize) {
 						((GCTextData) eldest.getValue()).release();
 						return true;
@@ -151,12 +156,12 @@ public final class GC extends Resource {
 				cache.clear();
 			}
 
-			public GCTextData get(String string) {
-				return cache.get(string);
+			public GCTextData get(Key key) {
+				return cache.get(key);
 			}
 
-			public void put(String string, GCTextData data) {
-				cache.put(string, data);
+			public void put(Key key, GCTextData data) {
+				cache.put(key, data);
 			}
 		}
 
@@ -1807,14 +1812,20 @@ private void doFastDrawText(String string, int x, int y) {
 }
 
 private GCTextData getTextData(String string) {
-	GCTextData data = (GCTextData) textDataCache.get(string);
-	if (data == null) {
+	// Create a cache key from the string, alpha, font and foreground color
+	// This ensures that the same string with different attributes is drawn correctly
+	GCTextData.Key key = new GCTextData.Key(string, data.alpha,
+			(data.font == null || data.font.handle == null) ? 0 : data.font.handle.id,
+			data.foreground);
+
+	GCTextData gcData = textDataCache.get(key);
+	if (gcData == null) {
 		NSAttributedString attribStr = createString(string, 0, true);
-		data = new GCTextData(attribStr);
+		gcData = new GCTextData(attribStr);
 		attribStr.release();
-		textDataCache.put(string, data);
+		textDataCache.put(key, gcData);
 	}
-	return data;
+	return gcData;
 }
 
 private void doDrawText(String string, int x, int y, int flags) {
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/GC.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/GC.java
@@ -13,6 +13,8 @@
  *******************************************************************************/
 package org.eclipse.swt.graphics;
 
+import java.util.*;
+
 import org.eclipse.swt.*;
 import org.eclipse.swt.internal.*;
 import org.eclipse.swt.internal.cocoa.*;
@@ -74,6 +76,7 @@ public final class GC extends Resource {
 
 	Drawable drawable;
 	GCData data;
+	GCTextData.Cache textDataCache = new GCTextData.Cache(20);
 
 	CGPathElement element;
 	int count, typeCount;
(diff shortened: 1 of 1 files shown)


Phillipus:
@SyntevoAlex Please review.

SyntevoAlex:
> @SyntevoAlex Please review.

Hopefully on Saturday.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/GC.java
```diff
@@ -1807,14 +1807,25 @@ private void doFastDrawText(String string, int x, int y) {
 }
 
 private GCTextData getTextData(String string) {
-	GCTextData data = (GCTextData) textDataCache.get(string);
-	if (data == null) {
+	// Create a cache key from the string, alpha, font and foreground color
+	// This ensures that the same string with different attributes is drawn correctly
+	StringBuilder str = new StringBuilder(string).append(data.alpha);
+	if(data.font != null) {
+		str.append(data.font.hashCode());
```
sratz, line 1814: `hashCode()` is not collission-free and therefore should not be used as a unique key.

Instead, I propose to either use a nested `Map<String, Map<Font>, Map<Color>>>` or create a small wrapper `Record` combining the 3 attributes and use that as key.

sratz (changes requested):

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/GC.java
Phillipus: Not sure how to do the nested Map. For the wrapper `Record` do you mean something like this?

```java
record Key(String string, int alpha, Font font, double[] fgColor) {}

String key = new Key(string, data.alpha, data.font, data.foreground).toString();
```

I tried it and it's a bit slower than using the `hashCode()` method. I used `hashCode()` to make the key String a bit shorter.

Phillipus: A I only used `hashCode()` to make the key String a bit shorter could we just not use the objects? 

```java
StringBuilder str = new StringBuilder(string).append(data.alpha);
if(data.font != null) {
	str.append(data.font);
}
if(data.foreground != null) {
	str.append(data.foreground);
}
String key = str.toString();
```

sratz: > Not sure how to do the nested Map. 

```java
Map<String, Map<Font, Map<Color, GCData>>> cache = new HashMap<>();
GCData gcdata = cache.computeIfAbsent(string, k -> new HashMap<>()).computeIfAbsent(font, k -> new HashMap<>()).get(color);
```

> For the wrapper Record do you mean something like this?
> ```java
> record Key(String string, int alpha, Font font, double[] fgColor) {}
> 
> String key = new Key(string, data.alpha, data.font, data.foreground).toString();
> ```

Yes, but without `toString()`.

```java
record Key(String string, int alpha, Font font, double[] fgColor) {}
Map<Key, GCData> cache;
Key key = new Key(string, data.alpha, data.font, data.foreground);
cache.get(key);
```

> I tried it and it's a bit slower than using the hashCode() method.

Do you have numbers how much slower?

> I used hashCode() to make the key String a bit shorter.

`hashCode()` is semantically incorrect. Two `Font`s or `Color`s hash codes may collide and be incorrectly mapped to the same cache entry. It's quite unlikely, but it's very bad practice to use `hashCode()` for something like this and we should avoid it.

sratz: > A I only used `hashCode()` to make the key String a bit shorter could we just not use the objects?

In practice I suppose we could, but that would rely on on its `toString()` implementation to return something unique and it's probably equally expensive.


Another problem I am not sure of, with the solution of holding on to the concrete references in `Key` is that we effectively preventing garbage collection of those `Font`s and `Color`s.

Using `WeakHashMap`s should solve that:

```java
Map<String, Map<Font, Map<Color, GCData>>> cache = new WeakHashMap<>();
GCData gcdata = cache.computeIfAbsent(string, k -> new WeakHashMap<>()).computeIfAbsent(font, k -> new WeakHashMap<>()).get(color);
```

Phillipus: > Yes, but without toString().

I force pushed this now and it's as fast as my previous tests (using [test snippet](url)).

> Another problem I am not sure of, with the solution of holding on to the concrete references in Key is that we effectively preventing garbage collection of those Fonts and Colors.

When the GC is destroyed the Cache is cleared with `textDataCache.release()` - is that sufficient?

sratz: > Actually, it's only the `Font` that is an object as we are storing an array of doubles for the foreground color.

For the `Font`:  We could also just hold on to the primitive (`long`) `font.handle.id` instead of the whole `Font` reference.

For the `Color`: `record#equals(Object)` does compare array members using `==`, but I guess this is OK here, it's probably not happening that we are dealing with two equal - but not same - `Color` references.

Phillipus: > For the Font: We could also just hold on to the primitive (long) font.handle.id instead of the whole Font reference.

I'm not sure if `font.handle` could be null. Here's `Font#hashCode()`

```java
public int hashCode() {
	return handle != null ? (int)handle.id : 0;
}
```

Phillipus: I've force pushed a new commit that uses `data.font.handle.id` for the font with null checks for `data.font` and `data.font.handle`. The key creation is a bit more verbose but now the cache key only contains primitives, not references to objects.

Phillipus: I've just noticed in the build failure reports:

```
warning: 'record' may become a restricted type name in a future release and may be unusable for type declarations or as the element type of an array
		private static record Key(String string, int alpha, long font, double[] fgColor) {}
```

iloveeclipse: > Is something in the build process not Java 17 compliant?

Well, ant still executed on Java 11 (see https://github.com/eclipse-platform/eclipse.platform.swt/issues/626) but the compiler should be able to compile Java 17 constructs if properly configured.

So far it looks like it is using plain `javac` task with almost no configuration for target/source/release ???
See https://github.com/eclipse-platform/eclipse.platform.swt/blob/2b9b6b4773991b9b15d534bffd660d1777a76238/bundles/org.eclipse.swt/buildSWT.xml#L298

If that is correct (and no other magic property is used), ant uses Java 11 and Java 11 uses Java 11 as default target, so it is correct to not being able to use records.

@HannesWell : how far are your efforts to get rid of ant in SWT?

Phillipus: > so it is correct to not being able to use records.

OK, thanks for that info.

So it seems that, at least for now, we cannot use a Java `record` for the cache key. A Java class could be used instead but then we'd have to supply our own `hashCode()` implementation.

HannesWell: > @HannesWell : how far are your efforts to get rid of ant in SWT?

Getting entirly rid of ANT in SWT will probably take some more time.
However I had a quick look and got the impression that at the moment the SWT java sources for all platforms are build twice with ant to verify that there are no compilation failures. But since the java code is already compiled with maven that seems useless for me now. So I would solve this particular problem by just removing the two extra ANT compilations.

I won't have time to check and work that out in detail before Sunday, probably even Monday, but I will do it with top priority then.

HannesWell: PR https://github.com/eclipse-platform/eclipse.platform.swt/pull/661 is merged.
You now should be able to use Java-17 language features. In the mentioned PR I verified that it works.

Phillipus: @HannesWell Thanks for that, this PR now builds and compiles.

@sratz What do you think? The cache key now consists only of the primitives of the string, alpha, font handle id and foreground color double[].

sratz (approved):
Looks good to me.

When merging, I suggest to keep the individual commits (not squash them) to see the changes in the history.

Phillipus:
> Looks good to me.
> 
> When merging, I suggest to keep the individual commits (not squash them) to see the changes in the history.

Thanks. I agree about the individual commits. At the very least we need to revert the revert.

Phillipus:
Will it be possible to merge this before 4.28 is released to avoid the regression caused by revert of https://github.com/eclipse-platform/eclipse.platform.swt/commit/9e996e825dbcaf2bc95a954cadadd24729b2ab8e making it into the build?

iloveeclipse:
> Will it be possible to merge this before 4.28 is released

I assume the question is more about "would it be possible to review this before 4.28?".
We can merge at any time - once we have agreement about the patch :-)

@sratz - you seem to have Mac - could you please look at this PR once again?

SyntevoAlex:
I could do that... eventually... but no promises when. Please proceed without me if you can.

lshanmug:
> To be clear, I don't want to force people to approve this PR. I just don't want the regression caused by the revert in [9e996e8](https://github.com/eclipse-platform/eclipse.platform.swt/commit/9e996e825dbcaf2bc95a954cadadd24729b2ab8e) to make it into Eclipse 4.28. So if it's not approved we could at least revert the revert (the first commit in this PR)

@Phillipus Can you please create a separate PR for the revert so that can be merged while commit2 is being reviewed?

Since it's a performance fix, what is the improvement we are seeing here? Possible to provide some numbers with a snippet similar to https://github.com/eclipse-platform/eclipse.platform.swt/issues/631#issuecomment-1524127112?
 
I found a old bug for the same issue- https://bugs.eclipse.org/bugs/show_bug.cgi?id=568777. Fixing it caused a severe performance [regression](https://bugs.eclipse.org/bugs/show_bug.cgi?id=569367) in the Eclipse editor. Can you please verify your fix in the Eclipse editor with "Show whitespace characters" turned on?

Phillipus:
@lshanmug  This PR fixes the issue described in #631. Due to the caching of single characters in the `GC` class if you set a different font, alpha or foreground color for a single chacracter in the same `GC` the original attributes are used instead. This PR uses a combination of those attributes as a cache. The snippets in #631 all work OK with this PR.

Running the Snippet https://github.com/eclipse-platform/eclipse.platform.swt/issues/631#issuecomment-1524127112 with this PR shows the same performance level from before the regression so it's all good again.

> Can you please verify your fix in the Eclipse editor with "Show whitespace characters" turned on?

Yes, I have done this as well.

vogella:
As we already have a revert commit, lets merge this one, sounds like we have a fast and working solution now.

vogella:
> > As we already have a revert commit, lets merge this one, sounds like we have a fast and working solution now.
> 
> @vogella Thanks, Lars. :-)

Thank you for your work here!

Issue #640, opened 2023-04-20
Remove NATIVE_STATS support


akurtakov:
Reasons are :
* broken for years
* no one from active committers seems to have used it for years
* requires rebuilding natives which too few people can do on their own
* systemtap/dtrace like tools provide far superior and easier to use profiling features
* documentation/examples for the usage is not existent (at least I am not aware of such)
* complicates the native headers and generation tool

akurtakov:
Initially brought to the surface by https://github.com/eclipse-platform/eclipse.platform.swt/pull/590

akurtakov:
Once new version of swt tools is available on update site, I'll push regenerated natives.

PR #651, opened 2023-04-28
Remove NATIVE_STATS generation and tool

akurtakov:
Reasons are :
* broken for years
* no one from active committers seems to have used it for years
* requires rebuilding natives locally to be used which too few people can do on their own
* systemtap/dtrace like tools provide far superior and easier to use profiling features
* documentation/examples for the usage is not existent (at least I am not aware of such)
* complicates the native headers and generation tool

Part of https://github.com/eclipse-platform/eclipse.platform.swt/issues/640


HannesWell:
Can you please check the readme files, the build scripts and make-files? IIRC they also contain references to NATIVE_STATS.

akurtakov:
> Can you please check the readme files, the build scripts and make-files? IIRC they also contain references to NATIVE_STATS.

I haven't found anything in readme files or build scripts. Makefiles are clean now. If you think of *.c and *.h files I would like to do that once new SWT tools is built and installed locally.

akurtakov:
@HannesWell What is the story with failing ppc64le verification build? Last I checked verification build was not rebuilding natives on this arch.

HannesWell:
> @HannesWell What is the story with failing ppc64le verification build? Last I checked verification build was not rebuilding natives on this arch.

Cannot tell immediately, but I can investigate it tomorrow evening.

SyntevoAlex:
I expected the same commit to contain re-generated JNI library sources.
Otherwise, the next committer to change something in JNI library will have these additional diffs.
Is there a reason why you didn't re-generate sources?

akurtakov:
> I expected the same commit to contain re-generated JNI library sources. Otherwise, the next committer to change something in JNI library will have these additional diffs. Is there a reason why you didn't re-generate sources?

https://github.com/eclipse-platform/eclipse.platform.swt/issues/640#issuecomment-1527131710 I'll do that tomorrow with the swt.tools build from p2 repo just to be 100% sure generation will be exactly the same for everyone .

akurtakov:
> Entire `NATIVE_ENTER` and `NATIVE_EXIT` generation could be cleaned up now, I think? See `NativesGenerator.generateEnterExitMacro()`

Actually, *_stats.h/c files makes not much sense anymore and they could be potentially removed (funtion names could go into main file) but that could be next step - if/when someone finds time for it.

SyntevoAlex:
ok

Issue #103, opened 2022-05-09
Some of the browser test fail on different platforms


laeubi:
I enabled the execution of tests on the[ three "latest" platforms of github](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources):

- **Windows:** windows-2022
- **Linux**: ubuntu-20.04
- **Mac**: macos-11

Unfortunately some of the browser tests fail on these

https://github.com/eclipse-platform/eclipse.platform.swt/runs/6346732956

akurtakov:
I'm aware of 2 browser test failures on Linux which you can see at https://download.eclipse.org/eclipse/downloads/drops4/I20220508-1800/testResults.php "swt.tests". Unfortunately never managed to find what's behind them when I looked. Hopefully will have some more time for this for this release.

laeubi:
> Unfortunately never managed to find what's behind them when I looked. Hopefully will have some more time for this for this release.

No problem, now they are more prominent there might be more interest in fixing them :-)

laeubi:
by the way do I get the output from the page right that we currently do not execute tests on windows?

akurtakov:
No, it means tests are still running for windows https://ci.eclipse.org/releng/job/ep424I-unit-win32-java11/ . I really can't tell you more but these are usually the slowest to run - quite likely as this is (shared?) VM with not enough power.

laeubi:
I think it would be interesting what the VM uses as windows OS, github offers

- windows-2019
- windows-2022

I could also enable windows-2019 if it helps?

Beside that I could fix 3/4 failing windows tests the last one is failing due it seems the browser is adding some more content and returning the live dom here, the tests itself seem a bit prone to changes in the sense of equal as they assume to get 'the same' html as the user provided back but obviously the same could be represented in many different ways in html in terms of case and additional white-spaces or even style attribute formatting.

So from the code perspective I think one either don't need to assume that one gets the html back that was set or we simply store the last text (what might then be outdated) or one would probably need more sophisticated comparison like https://jsoup.org/

laeubi:
It seem to only affect windows now for example here:
- https://github.com/eclipse-platform/eclipse.platform.swt/runs/12839370623

vogella:
cc @HeikoKlare who,I think I saw such a comment today,  indicated that he is interested in improving tests in SWT.

PR #653, opened 2023-04-28
Ignore semicolons in validation of Browser.getText tests #103

HeikoKlare:
The `org.eclipse.swt.tests.junitBrowser.Test_org_eclipse_swt_browser_Browser.test_getText*` tests expect the browser to return the same text as was set to it via `setText`, only ignoring formatting-related information such as capitalization and newline characters. In Windows tests, the browser returns a processed HTML document, which can also differ in other aspects while having the same semantics. One difference is the use of semicolons in style properties.

Since the tests only use basic HTML documents in which the semicolon has no meaning, the proposed change simply ignores semicolons when validating the expected result of Browser.getText.

A more sophisticated test could use the HTML DOM for comparison, but from my perspective the simple fix is sufficient as the comparison (1) is used only for very simple HTML documents right now and (2) should not be necessary in a more sophisticated way as SWT is not responsible for HTML processing of the browser but only for properly passing through API calls to the underlying browser.

Contributes to #103.


laeubi (approved):
This looks good and seem to fix some test, @HeikoKlare is this ready to be merged?

HeikoKlare:
Thanks @laeubi! Yes, this is ready to be merged.

PR #659, opened 2023-05-03
Regenerate bindings to no longer contain NATIVE_STATS

akurtakov:
Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/640


akurtakov:
Jenkins build succeeds and the code removed here was actually ifdefed for years - thus merging.

PR #660, opened 2023-05-05
Fix "no more handles" exception during Edge browser creation

HeikoKlare:
The Edge browser creation uses `Display.readAndDispath()` to process the OS events for browser instantiation. When there are other asynchronous executions scheduled while the browser is being initialized, they may be processed in between the processing of browser-instantiating OS events if too much time elapses between them. In case such an asynchronous execution changes a state that makes the browser instantiation fail, such as disposing the composite parent of the browser, an exception occurs.

In order to avoid the processing of asynchronous executions during browser instantiation, the change ensures that `readAndDisplay()` is only called when an OS event is present to be processed.

A drawback of this approach is that in case the browser instantiation takes a long time, for whatever reasons, the UI will not be responsive as UI events are not processed. I do not think that this situation is likely to happen but want to mention it as a potential drawback of the solution.


bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -241,13 +240,21 @@ static int callAndWait(String[] pstr, ToIntFunction<IUnknown> callable) {
...
 	completion.Release();
-	Display display = Display.getCurrent();
 	while (phr[0] == COM.S_OK && pstr[0] == null) {
-		if (!display.readAndDispatch()) display.sleep();
+		processNextOSMessage();
 	}
 	return phr[0];
 }
 
+private static void processNextOSMessage() {
```
iloveeclipse, line 249: Would be great to have a short comment explaining what the method is doing or why it is needed for Edge.

vogella:
Thanks @HeikoKlare this affected on of my clients in the past. I suggest to merge.

Destrolaric:
We also had to patch the issue behavior. It will be good to finally make it work properly inside the SWT. I would be glad if it would be merged.

HeikoKlare:
@Destrolaric can you already say whether this change makes your patch obsolete and fixes the behavior properly for you?

From my side, we can merge this PR. I have validated proper instantiation of the Edge browser on my system in an isolated SWT application as well as in our RCP product. Should someone else validate before merging?

lshanmug:
@deepika-u Can you please verify the patch?

deepika-u:
@HeikoKlare 

I have tried to recreate the problem of "No more handles" with below Snippet.

package org.eclipse.swt.bugs;

import org.eclipse.swt.*;
import org.eclipse.swt.browser.*;

public class NoMoreHandles {

		// TODO Auto-generated method stub

		createComposite(shell);

		shell.pack();
		shell.open();
		while (!shell.isDisposed()) {
				display.sleep();
		}
		display.dispose();
	}

	public static void createComposite(Composite parent) {
		parent.setLayout(new GridLayout(1, false));

		for (int i = 0; i <= 10; i++) {
			System.out.println("Iteration " + i);
			Composite c = new Composite(parent, SWT.NONE);
			Display.getCurrent().asyncExec(new Runnable() {

				@Override
				public void run() {
					c.dispose();
				}
			});
			Browser browser = null;
			if (!c.isDisposed()) {
				browser = new Browser(c, SWT.EDGE);
			}
			if (browser != null) {
				browser.dispose();
			}
		}
	}
}

When run seeing the below console output 
Iteration 0
Exception in thread "main" org.eclipse.swt.SWTError: No more handles [0x80070578]
	at org.eclipse.swt.SWT.error(SWT.java:4944)
	at org.eclipse.swt.browser.Edge.error(Edge.java:190)
	at org.eclipse.swt.browser.Edge.create(Edge.java:343)
	at org.eclipse.swt.browser.Browser.<init>(Browser.java:99)
	at org.eclipse.swt.bugs.NoMoreHandles.createComposite(NoMoreHandles.java:42)
	at org.eclipse.swt.bugs.NoMoreHandles.main(NoMoreHandles.java:16)

Then when i apply the patch, i am able to run it fine and not seeing any "No more handles".
So the patch looks fine for me. Please let me know if that verifies the fix and you want me to check anything else.

I have tried on below environment 
Eclipse SDK
Version: 2023-06 (4.28)
Build id: I20230507-1800
Java vendor: Eclipse Adoptium
Java runtime version: 20+36
Java version: 20

deepika-u:
@HeikoKlare 

Specifically tried for Edge browser with below snippet once again and edge browser window comes up properly after the patch is applied.

package org.eclipse.swt.bugs;

import org.eclipse.swt.*;
import org.eclipse.swt.browser.*;

public class NoMoreHandles {

		// TODO Auto-generated method stub

		createComposite(shell);

		shell.pack();
		shell.open();
		while (!shell.isDisposed()) {
				display.sleep();
		}
		display.dispose();
	}

	public static void createComposite(Composite parent) {
		//parent.setLayout(new GridLayout(1, false));

		for (int i = 0; i < 1; i++) {
			//for (int i = 0; i <= 10; i++) {
			System.out.println("Iteration " + i);
			Composite c = new Composite(parent, SWT.NONE);
			c.setLayout(new FillLayout());
			Display.getCurrent().asyncExec(new Runnable() {

				@Override
				public void run() {
					//c.dispose();
				}
			});
			Browser browser = null;
			if (!c.isDisposed()) {
				browser = new Browser(c, SWT.EDGE);
				browser.setSize(100, 100);
				browser.setUrl("whatismybrowser.com");
				//browser.setText("welcome");
			}
			if (browser != null) {
				//browser.dispose();
			}
		}
	}
}


So patch looks good to me.

niraj-modi (approved):
Approving, Thanks @deepika-u for verifying the patch.

HeikoKlare:
Great, thanks for verifying, @deepika-u!

deepika-u:
Verified this on below build and looks good to me.

Eclipse SDK
Version: 2023-06 (4.28)
Build id: I20230515-1800
Java vendor: Eclipse Adoptium
Java runtime version: 17.0.6+10
Java version: 17.0.6

PR #661, opened 2023-05-08
Remove redundant additional compilation by ANT

HannesWell:
Compiling the SWT sources with ANT is unnecessary because they are already compiled with Maven.
This saves build time, simplifies the build configuration and prevents problems because the ANT-compilation still required Java-11 (due to Nashorn).

@iloveeclipse FYI


vogella:
Thanks @HannesWell for simplifying the build all the time. This is really helpful, thanks for all your improvements.

Issue #644, opened 2023-04-26
SWT TableEditor : Caret is not shown for StyledText Control


raghucssit:
**Description**
If we use StyledText control in SWT TableEditor then caret is not visible inside Text Control. But for normal Text control it is visible.

**How to Reproduce**
Create a simple Table editor and use StyledText control for TableEditor.
Please use the attached snippet to reproduce.
[Snippet124withStyledText.txt](https://github.com/eclipse-platform/eclipse.platform.swt/files/11331049/Snippet124withStyledText.txt)

**Expected behavior**
Caret should be visible when user clicks on StyledText Control

**Screenshots**
![caret-missing](https://user-images.githubusercontent.com/10301131/234529177-6d4de53c-4600-4630-8b3e-8fad2fb64c50.png)

**Environment:**
- - [ ] All OS
- - [ ] Windows
- - [x] Linux
- - [x] macOS 
 
**Version**
 swt-3.121

Phillipus:
I tested this on Mac and I can see the caret when I click on the styled text control. However, I have to click twice to see the caret on the simple text controls.

raghucssit:
[TableEditorStyledTextExample.txt](https://github.com/eclipse-platform/eclipse.platform.swt/files/11381605/TableEditorStyledTextExample.txt)
I tried to add `Canvas` control as a table editor. Even on canvas `Caret` does not appear.
This hints that the issue is with `Table` instead of the `Control` which holds the `Caret`. Because i have added same control on a composite  and `Caret` appears.
I am trying to find root cause but so far no success,
@SyntevoAlex Any suggestion/hints ?

deepika-u:
> I tested this on Mac and I can see the caret when I click on the styled text control.

On windows 11 as well, we dont see any problem.

deepika-u:
@akurtakov 
Please have a look. Thanks

raghucssit:
> 
I have created a another simplest example uses only one cell and has canvas.
[SimplestExampleWithOnlyCanvas.txt](https://github.com/eclipse-platform/eclipse.platform.swt/files/11405355/SimplestExampleWithOnlyCanvas.txt)
I tried to find out when it was last worked. It was working on `Version: 2019-03 (4.11)
Build id: I20190307-0500`. In the next build (4.12) editor seems to be completely broken, i.e, editor itself did not render.
[...]

PR #663, opened 2023-05-09
SWT TableEditor: Caret is not shown for StyledText Control

iloveeclipse:
Add missing repaints for Caret that is created for an embedded Canvas.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/644

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Canvas.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Canvas.java
@@ -172,6 +172,11 @@ long gtk_commit (long imcontext, long text) {
 long gtk_draw (long widget, long cairo) {
 	if ((state & OBSCURED) != 0) return 0;
 	long result = super.gtk_draw (widget, cairo);
+	drawCaretInFocus(widget, cairo);
+	return result;
+}
+
+void drawCaretInFocus(long widget, long cairo) {
 	/*
 	 *  blink is needed to be checked as gtk_draw() signals sent from other parts of the canvas
 	 *  can interfere with the blinking state. This will ensure that we are only draw/redrawing the
@@ -183,7 +188,6 @@ long gtk_draw (long widget, long cairo) {
 		drawCaret(widget,cairo);
 		blink = false;
 	}
-	return result;
 }
 
 private void drawCaret(long widget, long cairo) {
@@ -223,7 +227,9 @@ private void drawCaret(long widget, long cairo) {
 
 		Cairo.cairo_fill(cairo);
 		Cairo.cairo_restore(cairo);
-		drawFlag = false;
+		if (caret.embeddedInto == null) {
+			drawFlag = false;
+		}
 	} else {
 		drawFlag = true;
 	}
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Caret.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Caret.java
@@ -17,6 +17,7 @@ package org.eclipse.swt.widgets;
 import org.eclipse.swt.*;
 import org.eclipse.swt.graphics.*;
 import org.eclipse.swt.internal.*;
+import org.eclipse.swt.internal.cairo.*;
 import org.eclipse.swt.internal.gtk.*;
 
 /**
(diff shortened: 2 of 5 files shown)


bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Caret.java
```diff
@@ -101,10 +113,24 @@ void createWidget (int index) {
 boolean drawCaret () {
 	if (parent == null) return false;
 	if (parent.isDisposed ()) return false;
-	GTK.gtk_widget_queue_draw(parent.handle);
+	if(embeddedInto != null && !embeddedInto.isDisposed() && isFocusCaret()) {
```
trancexpress, line 116: Missing empty space after `if`.

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet126.java
```diff
@@ -48,17 +48,25 @@ public static void main(String[] args) {
...
 		editor.grabHorizontal = true;
 		editor.setEditor(combo, item, 0);
+
 		editor = new TableEditor (table);
 		Text text = new Text (table, SWT.NONE);
 		text.setText("Text");
 		editor.grabHorizontal = true;
 		editor.setEditor(text, item, 1);
+
+		editor = new TableEditor (table);
```
trancexpress, line 58: What is this for?

tests/org.eclipse.swt.tests.gtk/ManualTests/org/eclipse/swt/tests/gtk/snippets/Issue644_CompleteTableEditor.java
```diff
@@ -100,51 +109,43 @@ public static void main(String[] args) {
 		display.dispose();
 	}
 
-	protected static void populateColumnName(int columnNum, TableColumn column) {
```
trancexpress, line 103: Don't really see a point in changing the switch? But probably you want to add the new method after the big switch methods, so that the history isn't as mangled as it is right now.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Canvas.java
```diff
@@ -223,7 +227,9 @@ private void drawCaret(long widget, long cairo) {
 
 		Cairo.cairo_fill(cairo);
 		Cairo.cairo_restore(cairo);
-		drawFlag = false;
+		if(caret.embeddedInto == null) {
```
trancexpress, line 230: Missing empty space after `if`.

trancexpress:
The canvas cells in the examples now also show a caret, one that can be seen but cannot be moved or used in any other way.

Should the new code be enabled only for a `StyledText`?

iloveeclipse:
> The canvas cells in the examples now also show a caret, one that can be seen but cannot be moved or used in any other way.

Because the code is not added in example to do that, there is no need for that to show if caret blinks or not. 

> Should the new code be enabled only for a `StyledText`?

StyledText is Canvas on steroids, and all the caret drawing happen in Canvas, so what's the point to fix only StyledText?

trancexpress:
> StyledText is Canvas on steroids, and all the caret drawing happen in Canvas, so what's the point to fix only StyledText?

What happens if I have a table cell with a canvas used to draw something? Will it also show a caret now?

iloveeclipse:
> What happens if I have a table cell with a canvas used to draw something? Will it also show a caret now?

If caret is created and set, it should be shown, what's the point to create and set a caret for canvas that shouldn't be shown?

trancexpress:
> If caret is created and set, it should be shown, what's the point to create and set a caret for canvas that shouldn't be shown?

Ah I see, I didn't notice the caret code added to the snippets. Makes sense then.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Canvas.java
```diff
@@ -183,7 +188,6 @@ long gtk_draw (long widget, long cairo) {
 		drawCaret(widget,cairo);
 		blink = false;
```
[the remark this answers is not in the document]
iloveeclipse: Intentional and also part of refactoring.
This part of the original method is called now from two different entry points, but what it does should be same for both cases.

tests/org.eclipse.swt.tests.gtk/ManualTests/org/eclipse/swt/tests/gtk/snippets/Issue644_CompleteTableEditor.java
iloveeclipse: The code is replaced with **one line**. Other switches were using indexes and I hate updating indexes every time because they do not have any meaning.

iloveeclipse:
Manually tested with our product (where it was reported for complex cell editors embedded in a Tree) and it seem to work as expected.

I will merge this PR now so we can see / test it before M3.

Issue #116, opened 2022-05-10
[win32] Test_org_eclipse_swt_dnd_DND fails


laeubi:
See https://github.com/eclipse-platform/eclipse.platform.swt/runs/6372153691

the test fails with:

> java.lang.AssertionError: Drop received other data as we dragged.
> 	at org.eclipse.swt.tests.win32.Test_org_eclipse_swt_dnd_DND.testImageTransfer(Test_org_eclipse_swt_dnd_DND.java:169)

HeikoKlare:
The test is still constantly failing, e.g.: https://github.com/eclipse-platform/eclipse.platform.swt/actions/runs/4862153146/jobs/8668569934

The reason is the following: 
The test first creates a 32bit (RGBA-)bitmap. Since the bitmap does not contain reasonable alpha values upon creation, the `Image.getImageData()` method assigns `255` as alpha value for every pixel:
https://github.com/eclipse-platform/eclipse.platform.swt/blob/8e8b0b72b8b2db9b74cbd1ac5e2dc2ac9244459d/bundles/org.eclipse.swt/Eclipse%20SWT/win32/org/eclipse/swt/graphics/Image.java#L1641-L1643
This behavior has been introduced with https://github.com/eclipse-platform/eclipse.platform.swt/commit/36091a0a057565596b97242c6230ea08bf07fabd to improve compatibility, even though alpha values should actually be stored in the `ImageData.alphaData` array rather than the `data` array.

When performing a drag-and-drop operation with this image, the data extraction in `ImageTransfer` treats these values as proper alpha values and puts them into the `ImageData.alphaData` array. In consequence, the dropped image has transparency type `TRANSPARENCY_ALPHA` while the original image has `TRANSPARENCY_NONE`.

I see basically two options to fix this:
1. Ensure that that the images before and after drag-and-drop operation via `ImageTransfer` are "completely" equal. The problem is that an image may be dropped from anywhere, so interpreting potentially present alpha values in the data array seems reasonable. In addition, dropped images are always considered to be device-independent bitmaps. In the test case, the original bitmap is device dependent. The only chance for improvement I see here is to completely avoid the situation that an image contains any alpha values within the `data` array instead of the `alphaData` array, which would be a follow up to https://github.com/eclipse-platform/eclipse.platform.swt/commit/36091a0a057565596b97242c6230ea08bf07fabd.
[...]

PR #668, opened 2023-05-12
Improve Test_org_eclipse_swt_dnd_DND.testImageTransfer #116

HeikoKlare:
The Windows test for DND image transfers expects transparency properties from the dropped image that cannot be guaranteed by the ImageData implementation for alpha channel values.

Since the current behavior of the ImageData implementation is valid, this change adapts the according test case by improving the assertion for the DND image transfer result. In addition, it adds multiple test cases to capture different types of ImageData to be transferred via drag and drop.

Contributes to #116.


jukzi (approved):
I have tested locally that the test does not fail on windows anymore with this change.

Issue #671, opened 2023-05-15
Browser tests are not executed for Edge browser


HeikoKlare:
**Describe the bug**
The browser tests `Test_org_eclipse_swt_browser_Browser` are currently only executed with flag `SWT.NONE`. On WIndows systems, this results in tests only being executed for the IE browser, but not for the Edge browser.

**Expected behavior**
Browser tests should be parameterized with relevant configurations to test. This particularly involves the Edge browser implementation on Windows.

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 

2. Additional OS info (e.g. OS version, Linux Desktop, etc)
irrelevant

3. JRE/JDK version
irrelevant
 
**Version since**
Probaby always

**Workaround (or) Additional context**
Regression tests were not run for the changes in #660, as the changes affect the Edge browser and tests are only execute for IE.

PR #672, opened 2023-05-15
Execute browser tests for Edge #671

HeikoKlare:
Browser tests were only executed for the default configuration of a system's browser using the `SWT.NONE` flag. Other configurations, such as using the Edge browser in Windows, were not tested.

This change parameterizes the browser tests to also execute them for the Edge browser on Windows. It also deactivates those tests for the Edge browser for which the implementation does (currently) not work. This allows to detect regressions when performing future changes to the Edge browser.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/671

### DIFF
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
@@ -81,6 +81,7 @@ import org.eclipse.swt.widgets.Event;
 import org.eclipse.swt.widgets.Shell;
 import org.eclipse.swt.widgets.Text;
 import org.junit.Before;
+import org.junit.BeforeClass;
 import org.junit.FixMethodOrder;
 import org.junit.Rule;
 import org.junit.Test;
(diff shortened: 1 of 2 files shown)


tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
```diff
@@ -206,6 +229,16 @@ public void tearDown() {
...
+	if (isEdge) {
+		// wait for and process pending events to properly cleanup Edge browser resources
+		do {
+			processUiEvents();
+			try {
+				Thread.sleep(100);
+			} catch (InterruptedException e) {
+			}
+		} while (Display.getCurrent().readAndDispatch());
+	}
```
HeikoKlare, line 241: This is necessary because otherwise some OS events that are somehow produced by one test occur and are processed by the next test and may interfere with the Edge browser instantation. I have invested quite some time to identify where the events come from and how to properly handle them, but I did not find a proper way to do so. If someone is able to identify the reasons for the events and how to properly process them, I would be glad to have a better solution than this,

vogella:
@HeikoKlare can this be merged?

jukzi:
```
creating browser took too long: 23469ms
java.lang.AssertionError: creating browser took too long: 23469ms
	at org.eclipse.swt.tests.junit.Test_org_eclipse_swt_browser_Browser.createBrowser(Test_org_eclipse_swt_browser_Browser.java:290)
	at org.eclipse.swt.tests.junit.Test_org_eclipse_swt_browser_Browser.setUp(Test_org_eclipse_swt_browser_Browser.java:173)

```

HeikoKlare:
I converted this back to draft due to the recent test failure and I will have another look at it.
Further increasing the waiting time between test cases would probably "solve"(/hide) the problem, but I'd prefer a solution that makes waiting times between test cases obsolete.

fedejeanne:
Test failures in Linux are unrelated #1564

PR #689, opened 2023-06-06
OleClientSite: use try-with-resource

jukzi:


deepika-u:
@jukzi 
This fix is for which issue? Does that also has recreate steps please? Can you confirm on this please.

jukzi:
@deepika-u i have no issue - it's just that theoretically if there would be an Exception before close then it could leak resource.

deepika-u:
Thanks for sharing the details.

PR #695, opened 2023-06-09
Run GH actions maven with 3.9.2

akurtakov:


akurtakov:
GH builds run but tests fail. Still a step forward thus merging.

PR #698, opened 2023-06-12
ImageLoader: Use try-with-resource

jukzi:
Doesn't try to fix anything. Just code and warnings reduced.


jukzi:
about to ignore unrelated failing tests

PR #701, opened 2023-06-13
MacGenerator: close InputStream after use

jukzi:


jukzi:
ignoring unrelated known test failures.

PR #719, opened 2023-07-02
SwtTestUtil: new method to wait for event

SyntevoAlex:
Was needed for #450, and feels useful overall.


SyntevoAlex:
In master now.

PR #721, opened 2023-07-03
Increase timeout when waiting for shell to become active in tests #616

HeikoKlare:
Some `Composite` tests introduced with #617 currently fail randomly because waiting for the shell to become active when initializing the test times out.

This change increases the timeout to eliminate slow hardware or bad runtime behavior as a potential reason for failing tests.

Contributes to #616.


HeikoKlare:
Failing test is unrelated and documented in #722.

HeikoKlare:
I'll merge to see whether increasing the timeout helps to avoid the random test failures mentioned in https://github.com/eclipse-platform/eclipse.platform.swt/issues/616#issuecomment-1616669535.

SyntevoAlex:
Is it expected that this PR does not change `test_setFocus_withVisibleAndInvisibleChild` test I reported in #616 ?

Issue #724, opened 2023-07-04
[cocoa] In tests, Shell sometimes doesn't activate


SyntevoAlex:
For example, see `test_setFocus_withVisibleAndInvisibleChild` failures:
https://github.com/eclipse-platform/eclipse.platform.swt/runs/14613072507
https://github.com/eclipse-platform/eclipse.platform.swt/runs/14690913997
https://github.com/eclipse-platform/eclipse.platform.swt/runs/14287265191
https://github.com/eclipse-platform/eclipse.platform.swt/runs/14286393799

Also see `test_Issue450_NoShellActivateOnSetFocus` failures:
https://download.eclipse.org/eclipse/downloads/drops4/I20230703-1800/testresults/html/org.eclipse.swt.tests_ep429I-unit-mac64-java17_macosx.cocoa.x86_64_17.html
https://download.eclipse.org/eclipse/downloads/drops4/I20230703-1800/testresults/html/org.eclipse.swt.tests_ep429I-unit-macM1-java17_macosx.cocoa.aarch64_17.html

Also see #721 where timeout was increased.

SyntevoAlex:
Turns out, on macOS, no amount of waiting helps.

Here it can be seen that all tests may fail even after waiting 30 sec:
https://github.com/eclipse-platform/eclipse.platform.swt/runs/14789121538

I noticed the following:
* Sometimes tests succeed
* Sometimes they fail, and all of them seem to fail together.

It seems that when a problem occurs, entire testing session is affected.

HeikoKlare:
Are you sure that this is an issue of the shell not becoming active or only of a missing `Activate` event? I'm just asking because https://github.com/eclipse-platform/eclipse.platform.swt/pull/725 replaced waiting for a shell by becoming the active one of the `Display` by waiting for the SWT event for activation.

sratz:
We've also had issues with `SWT.Activate` not firing on cocoa.

Waiting for `SWT.Show` instead fixed the problem.

HeikoKlare:
Thanks for the hint!

I can easily apply the fix to make the pre-existing focus tests running again, as they did not depend on the `Activate` event before. I am, however, unsure about the regression test for issue #450, i.e., I am unsure whether that test actually depends on awaiting that specific `Activate` event or if replacing it by waiting for `Show` would also be valid.

HeikoKlare:
[...]

PR #725, opened 2023-07-04
[cocoa] In tests, Shell sometimes doesn't activate in a timely manner

SyntevoAlex:
See #724


SyntevoAlex:
In master now. Tests only change, so I think it's good despite code freeze for M1.

PR #732, opened 2023-07-08
Issue #724: [cocoa] Shell.open() does not always cause Shell to activate

SyntevoAlex:
fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/724

### DIFF
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/SwtTestUtil.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/SwtTestUtil.java
@@ -396,26 +396,29 @@ public static boolean waitEvent(Runnable trigger, Control control, int swtEvent,
  * @return <code>true</code> if Shell became active within timeout
  */
 public static void waitShellActivate(Runnable trigger, Shell shell) {
-	final int timeout = 3000;
-	final long timeBeg = System.currentTimeMillis();
+	final int timeout = 1000;
+
+	Runnable trigger2 = () -> {
+		if (isCocoa) {
+			// Issue #731: if another app gains focus during entire JUnit session,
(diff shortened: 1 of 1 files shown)


SyntevoAlex:
In master now.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/SwtTestUtil.java
```diff
@@ -396,26 +396,29 @@ public static boolean waitEvent(Runnable trigger, Control control, int swtEvent,
...
-	final int timeout = 3000;
-	final long timeBeg = System.currentTimeMillis();
+	final int timeout = 1000;
+
+	Runnable trigger2 = () -> {
+		if (isCocoa) {
+			// Issue #731: if another app gains focus during entire JUnit session,
+			// newly opened Shells do not activate. The workaround is to activate
+			// explicitly.
+			shell.forceActive();
```
HeikoKlare, line 406: Taking a look at this change, I do not really understand why it should change anything.
In most cases, the `waitShellActivate` method is called for shells that are closed. But calling `forceActive` on a closed shell has no effect. Do I miss something?

SyntevoAlex: You are right, that was a mistake. It should go after original trigger. Thanks for the catch!

HeikoKlare: I give it a try in #741.

SyntevoAlex: Thanks!

Issue #722, opened 2023-07-03
Random failing Test_org_eclipse_swt_browser_Browser.*


HeikoKlare:
Different test cases of the `Test_org_eclipse_swt_browser_Browser` class randomly fail in GitHub Actions builds. So far I saw the following test cases affected:
* test_BrowserFunction_callback (e.g. https://github.com/eclipse-platform/eclipse.platform.swt/actions/runs/5442699357/jobs/9898632238)
* test_setUrl_remote (e.g. https://github.com/eclipse-platform/eclipse.platform.swt/actions/runs/5437689673/jobs/9888509738)

All tests seem to fail because of running into a timeout when waiting for a condition to be fulfilled (i.e., calling `waitForPassCondition`). Probably the tests fail because of slow hardware or specific runtime behavior due to which the condition takes longer than the rather small timeout (2 seconds) to be fulfilled.

A proper solution might be to increase the timeout as its value has no specific meaning and the timeout is only supposed to avoid non-terminating tests. I can provide a PR to test it.

jukzi:
still failing on I Build:
https://download.eclipse.org/eclipse/downloads/drops4/R-4.28-202306050440/testresults/html/org.eclipse.swt.tests_ep428I-unit-mac64-java17_macosx.cocoa.x86_64_17.html
java.lang.AssertionError:
```
Expected title and actual title do not match.
Expected: Website Title
Actual: file:///Users/genie.releng/workspace/AutomatedTests/ep428I-unit-mac64-java17/workarea/I20230605-0440/eclipse-testing/test-eclipse/Eclipse.app/Contents/Eclipse/plugins/org.eclipse.swt.tests_3.107.0.v20230512-1002/data/testWebsiteWithTitle.html
Test log:
PLUGIN_PATH: /Users/genie.releng/workspace/AutomatedTests/ep428I-unit-mac64-java17/workarea/I20230605-0440/eclipse-testing/test-eclipse/Eclipse.app/Contents/Eclipse/plugins/org.eclipse.swt.tests_3.107.0.v20230512-1002
URL: /Users/genie.releng/workspace/AutomatedTests/ep428I-unit-mac64-java17/workarea/I20230605-0440/eclipse-testing/test-eclipse/Eclipse.app/Contents/Eclipse/plugins/org.eclipse.swt.tests_3.107.0.v20230512-1002/data/testWebsiteWithTitle.html
TitleListener fired
[...]

PR #733, opened 2023-07-10
Improve shell activation assertions and remove warnings #724

HeikoKlare:
In addition to the test fix for ensuring shell activation when opening a shell on MacOS (#732), this PR makes the following improvements:
- Improve assertions when waiting for the shell to activate fails
- Remove exception declarations that became obsolete due to refactored shell activation logic and resulted in warnings

Contributes to https://github.com/eclipse-platform/eclipse.platform.swt/issues/724


HeikoKlare:
Failing tests are documented in #722,

PR #736, opened 2023-07-11
Issue #724: [cocoa] In tests, Shell sometimes doesn't activate

SyntevoAlex:
Previous patch didn't help to defeat test failures on GitHub. Add even more debugging code.

For debugging of https://github.com/eclipse-platform/eclipse.platform.swt/issues/724

### DIFF
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/SwtTestUtil.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/SwtTestUtil.java
@@ -18,6 +18,7 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
+import java.io.PrintStream;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.BooleanSupplier;
 
@@ -33,7 +34,13 @@ import org.eclipse.swt.graphics.RGB;
 import org.eclipse.swt.graphics.Rectangle;
 import org.eclipse.swt.layout.FillLayout;
 import org.eclipse.swt.layout.GridLayout;
-import org.eclipse.swt.widgets.*;
+import org.eclipse.swt.widgets.Canvas;
+import org.eclipse.swt.widgets.Control;
+import org.eclipse.swt.widgets.Display;
+import org.eclipse.swt.widgets.Label;
+import org.eclipse.swt.widgets.Listener;
+import org.eclipse.swt.widgets.Shell;
+import org.eclipse.test.Screenshots;
 
 public class SwtTestUtil {
 	/**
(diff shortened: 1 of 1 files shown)


HeikoKlare:
I assume this code was added for temporary debugging purposes only and not be kept permanently, wasn't it?
May we can have a tracking issue to ensure that the code will be removed after the issue to debug has been resolved?
Same applies to eclipse-platform/eclipse.platform.ui#933.

SyntevoAlex:
I actually intended that this code stays. Why not?

HeikoKlare:
With respect to this PR, it is particularly the usage of stdout for logging purposes. If the code is supposed to stay, I would at least propose to have the output logged with a proper logging mechanismus.

With respect to https://github.com/eclipse-platform/eclipse.platform.ui/pull/933: It is a code copy of logic within this repository. And here we already see what is bad about copies: The code in this repository has been modified as it contained a bug, while the code clone stays faulty. The logic should be extracted to a shared utility where it can be used from both repositories and maintained at one single place.

SyntevoAlex:
This PR only prints to console when something unexpected happened. I think that it's completely justified. When tests work as expected, nothing is expected to be printed.

As for https://github.com/eclipse-platform/eclipse.platform.ui/pull/933, I agree that it would be better to have code in one place, just it's hard to say which place could that be. Putting this code into SWT feels wrong, and using SWT-Tests jar in platform.ui tests also feels wrong.

HeikoKlare:
I would actually not propose to print to console anywhere (except for temporary, local testing purposes), since it has only drawbacks compared to using a proper logging framework (unflexible, unconfigurable etc.). But I won't insist on changing that here, so it's fine for me to keep it.

For the SWT testing utils, it could make sense to have a dedicated place for shared test utilities. But since the modification in https://github.com/eclipse-platform/eclipse.platform.ui/pull/933 is the only place of duplication I am aware of, it would be inappropriate to add another pluging or add further depedencies for that. Since the issue https://github.com/eclipse-platform/eclipse.platform.ui/issues/910 that was supposed to be debugged with the change in https://github.com/eclipse-platform/eclipse.platform.ui/pull/933 became obsolete as the causing PR has been reverted, I would simply propose to revert the commit: https://github.com/eclipse-platform/eclipse.platform.ui/pull/950

SyntevoAlex:
It wasn't wise to revert entire commit, for example, there is no sense in making screenshots on every test run. Also, the part that replaced `.setVisible()` with `shell.open()` was useful. According to logs, with my code the test was actually fixed, see

https://download.eclipse.org/eclipse/downloads/drops4/I20230716-1800/testresults/ep429I-unit-win32-java17_win32.win32.x86_64_17/org.eclipse.jface.tests.databinding.BindingTestSuite.txt
```
ControlObservableValueTest.testObserveFocus() start active shell: null
focus control (1): Text {}, c2? false
active shell (4): Shell {} [layout=FillLayout {type=SWT.HORIZONTAL}]
```

vs https://download.eclipse.org/eclipse/downloads/drops4/I20230721-1800/testresults/ep429I-unit-win32-java17_win32.win32.x86_64_17/org.eclipse.jface.tests.databinding.BindingTestSuite.txt
```
ControlObservableValueTest.testObserveFocus() start active shell: null
active shell (2): null
active shell (3): null
```

It can be seen that before my change (and after your revert), Shell fails to activate until `.setFocus()`, contrary to tests's intent.
With my test, `waitShellActivate()` contained `fail()` which didn't trigger so Shell did activate.

HeikoKlare:
The commit stated that it was for debbuging https://github.com/eclipse-platform/eclipse.platform.ui/issues/910, which became obsolete due to the reverted change of default `setFocus` behavior that caused the issue.

I understand that the test may still be broken (but does not reveal that by failing), so it's fine to adress that. But then I would expect it to be done in a commit/PR that addresses the according issue and in a way that avoids code duplications or reduces them to a minimum (e.g. no generalization over events to wait for, no platform specificity etc.). 
However, I do not really understand why the test is interested in the active shell at all. It is only printed to console, but nothing is asserted with respect to the shell being active. So either that can be removed from the test or debug output has to be replaced with proper assertions. Same applies to printed information with respect to the focus control and the observed value.

PR #739, opened 2023-07-12
Wait for SWT.Show event in Composite focus tests #616

HeikoKlare:
The tests for different focus settings of visible and invisible child controls added with #617 were implemented and refactored to wait for the shell to be activated after being opened.
Since the activation event cannot be reliably expected at least on MacOS and since activation is not even necessary for the tests, but having the shell shown is sufficient, this change adapts the awaited event to `SWT.Show`.

Contributes to https://github.com/eclipse-platform/eclipse.platform.swt/issues/616

This is based on the hint from @sratz in https://github.com/eclipse-platform/eclipse.platform.swt/issues/724#issuecomment-1625196057.


HeikoKlare:
Obviously, the assumption that focus is properly set as soon as the shell is shown was wrong. So the `setFocus` tests have to wait for shell activation.

PR #740, opened 2023-07-12
Explicitly activate shell after openining in setFocus tests #616

HeikoKlare:
The tests for different focus settings of visible and invisible child controls added with #617 were implemented and refactored to wait for the shell to be activated after being opened.
Since the activation event cannot be reliably expected at least on MacOS and since expecting activation by `Shell.open()` is nothing the test is supposed to cover, this change makes the tests explicitly activate the shell after opening it to avoid unrelated test failures.

Contributes to https://github.com/eclipse-platform/eclipse.platform.swt/issues/616


SyntevoAlex:
It probably makes more sense to fix the mistake you found in #732. I was instructed to switch to other tasks, so I'll leave this investigation to you for now.

HeikoKlare:
Should be obsolete du to #741.

PR #741, opened 2023-07-12
Correct order for triggering shell event and forcing active in test #724

HeikoKlare:
The test utility to wait for a shell to activate introduced in #732 uses a workaround for MacOS to always enforce activation. The current behavior may have no effect because forcing activation before actually making the shell visible has no effect.

This change reorders the statements so that first the activation-triggering action is executed before enforcing activation for the situation that is does not work properly.

Contributes to https://github.com/eclipse-platform/eclipse.platform.swt/issues/724


HeikoKlare:
I am not completely sure whether this change finally solves the test issues of #724. But since it cannot break anything, I merge to see whether subsequent builds succeed or still have sporadic failures.

PR #747, opened 2023-07-14
Don't set foreground color if color is null

Christopher-Hermann:
Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/938

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CLabel.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CLabel.java
@@ -1026,4 +1026,11 @@ private String[] splitString(String text) {
 	} while (pos != -1);
 	return lines;
 }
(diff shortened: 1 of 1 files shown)


mickaelistria (changes requested):
The specification of `GC.setForeground()` states
```
 * @exception IllegalArgumentException <ul>
 *    <li>ERROR_NULL_ARGUMENT - if the color is null</li>
```
We cannot change API contracts, so it's necessary as per specification to send the exception in case the color is null. This change breaks the API contract.
Fix should take place in consumer instead.

Christopher-Hermann:
Does this also applies for Controls? JavaDoc for Control.setForeground states: Sets the receiver's foreground color to the color specified by the argument, or to the default system color for the control if the argument is null.

Nevertheless, you are right, in both cases the API contract is broken. I will look for another place. Thanks for the hint.

Christopher-Hermann:
Fixed in https://github.com/eclipse-platform/eclipse.platform.ui/pull/939

Issue #662, opened 2023-05-09
Collapsed Table Columns can not be Re-Expanded on Mac


jwflicker:
**Describe the bug**
Using any SWT Table, such as on Preference page General > Network Connections, use the mouse to change the width of a table column to its minimum size.  Afterwards, there is no way to re-expand that table column.

**To Reproduce**
Navigate to preference page General > Network Connections
Using the hold the mouse down on the border between two column headers and drag the width down all the way to the left.
After wards, try to make that column visible... on Windows there is an affordance you can use to hold the mouse down and drag to the right to expand the hidden column's width... 
(or) 
Steps to reproduce the behavior.

**Expected behavior**
On Windows there is a way to place the mouse at the edge of the hidden column, hold the mouse down and drag to re-size it and make it visible again.  Expect something similar on Mac.

**Screenshots**

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [ ] Windows
- - [ ] Linux
- - [X] macOS 

2. Additional OS info (e.g. OS version, Linux Desktop, etc)
Apple M1 Max
Mac OS Ventura 13.3.1
Eclipse IDE for Eclipse Committers (includes Incubating components)
 - Version: 2022-03 (4.23.0)
 - Build id: 20220310-1457

3. JRE/JDK version
eclipse.buildId=4.23.0.I20220308-0310
java.version=11.0.16.1
java.vendor=Eclipse Adoptium
BootLoader constants: OS=macosx, ARCH=aarch64, WS=cocoa, NL=en_US
 
**Version since**
Reproducible on older versions and other jvms

**Workaround (or) Additional context**
No known workarounds

elsazac:
@jwflicker

 I could recreate the problem in Mac OS 13.3.1 aarch64.
In Mac OS there is an option to expand the hidden columns but seems its not working, when the column headers are overlapping on each other.
<img width="346" alt="Screenshot 2023-05-11 at 10 41 09 AM" src="https://github.com/eclipse-platform/eclipse.platform.swt/assets/122080218/343d4b63-2dab-48ee-895b-3355c38e412c">

Build information listed below:
Eclipse SDK
Version: 2023-06 (4.28)
Build id: I20230509-1800
OS: Mac OS X, v.13.3.1, aarch64 / cocoa
Java vendor: Eclipse Adoptium
Java runtime version: 20+36
Java version: 20

deepika-u:
[...]

PR #760, opened 2023-08-03
Enabling default value to spacing to avoid table collapsing in Mac.

elsazac:
Fixes :https://github.com/eclipse-platform/eclipse.platform.swt/issues/662

The table columns once collapsed one over the other, doesn't  reexpand to bring back to the original form.This is due to the very small spacing.width value of intercell spacing.This is a fix inorder to bring back the default spacing.width value (17.0) to avoid the collapsing problem.

cc/ @lshanmug


lshanmug:
@elsazac Is this issue specific to aarch64? Do we know what would be the spacing.width default value for x86_64?

lshanmug (approved):
LGTM

Phillipus:
This has cause a regression with showing horizontal scroll bars. Please see my comments in https://github.com/eclipse-platform/eclipse.platform.swt/issues/662

Phillipus:
Because of the regressions documented in #662 could we please revert this commit for Eclipse 4.29 to allow time to investigate the problems? As it is, we are seeing issues with Table and Tree column widths in our RCP app which are causing UI problems.

PR #769, opened 2023-08-15
Change the message of the Exception in ResourceTracker

fedejeanne:
See https://github.com/eclipse-platform/eclipse.platform/issues/528#issuecomment-1678484049

Change the error message in the `ResourceTracker` to inform about possible false-positives _i.e._ cases where it is **not** a leak. 

Contributes to https://github.com/eclipse-platform/eclipse.platform/issues/528

## How to reproduce the error
Randomly throw an Exception right after the tracker is initialized (at the bottom of `initNonDisposeTracking`)

```java
void initNonDisposeTracking() {
	...
	// throw an exception every once in a while
	if (Math.random() > 0.99) SWT.error(SWT.ERROR_UNSPECIFIED);	
}
```

And you will see some exceptions saying: `SWT Resource was not properly disposed`.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Resource.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Resource.java
@@ -168,8 +168,16 @@ void initNonDisposeTracking() {
 	// Avoid performance costs of having '.finalize()' when not tracking.
 	if (nonDisposedReporter == null) return;
 
-	// Capture a stack trace to help investigating the leak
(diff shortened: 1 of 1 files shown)


SyntevoAlex (changes requested):
I object to this PR because I suspect that the problem does not exist, and therefore this PR introduces incorrect assumption that it does exist.

I would like to get answers to questions here:
https://github.com/eclipse-platform/eclipse.platform/issues/528#issuecomment-1680540627
https://github.com/eclipse-platform/eclipse.platform/issues/528#issuecomment-1680646316

before proceeding.

fedejeanne:
See https://github.com/eclipse-platform/eclipse.platform/issues/528#issuecomment-1682410261

Issue #677, opened 2023-05-18
[macOS Silicon] Table and Tree item height gets smaller when setting the font


Phillipus:
This occurs on  Mac Silicon (aarch64) only, presumably because SWT on that platform uses a later Apple SDK. I've tested on Mac Mini M1 with Monterey and Ventura.

Setting a table or tree font to the same (or different) font reduces the item height.

1. Create a simple table with some rows
2. The item height of the rows is 25:
<img width="612" alt="Before" src="https://github.com/eclipse-platform/eclipse.platform.swt/assets/600504/8e3583c5-7121-4e52-8666-b1346de46359">

3. Set the table's font to its current font (`table.setFont(table.getFont())`)
4. The item height of the rows is now 18:
<img width="612" alt="After" src="https://github.com/eclipse-platform/eclipse.platform.swt/assets/600504/3a293388-68a5-430a-945d-cf1973f90755">

£xpected behaviour: should use the height of the target SDK.  On aarch64 this is the 25 pixel item height. On Intel it is 18.

Phillipus:
Snippet to reproduce:

```java
import static org.eclipse.swt.events.SelectionListener.widgetSelectedAdapter;

import org.eclipse.jface.layout.GridDataFactory;
import org.eclipse.swt.SWT;
import org.eclipse.swt.layout.GridData;
import org.eclipse.swt.layout.GridLayout;
import org.eclipse.swt.widgets.Button;
import org.eclipse.swt.widgets.Display;
import org.eclipse.swt.widgets.Label;
import org.eclipse.swt.widgets.Shell;
import org.eclipse.swt.widgets.Table;
import org.eclipse.swt.widgets.TableItem;

public class TableFont {

    public static void main(String[] args) {
        Display display = new Display();
        Shell shell = new Shell(display);
        shell.setText("Table Font");
        shell.setLayout(new GridLayout());
        
        Table table = new Table(shell, SWT.BORDER | SWT.V_SCROLL | SWT.H_SCROLL);
        GridDataFactory.create(GridData.FILL_HORIZONTAL).applyTo(table);
        
        for(int i = 0; i < 11; i++) {
            TableItem item = new TableItem(table, 0);
            item.setText("Item " + i);
        }
        
        Label label1 = new Label(shell, SWT.NONE);
        Label label2 = new Label(shell, SWT.NONE);
        updateLabels(table, label1, label2);
        
        Button button = new Button(shell, SWT.PUSH);
[...]

Issue #452, opened 2022-10-26
Crash with org.eclipse.swt.browser using Safari on macOS Ventura


dstainhauser:
**Describe the bug**
Eclipse and RCP runtimes crash on macOS 13 Ventura using the org.eclipse.swt.browser.

**To Reproduce**
On macOS 13 Ventura, Open the Internal web browser, enter https://google.com or any other url. It crashes
Same happens on a RCP build using the Safari browser.

**Expected behavior**
Using equo.chromium works on Ventura. It worked with previous macOS versions. 

**Screenshots**

> System Integrity Protection: enabled
> 
> Crashed Thread:        0  Dispatch queue: com.apple.main-thread
> 
> Exception Type:        EXC_BREAKPOINT (SIGTRAP)
> Exception Codes:       0x0000000000000002, 0x0000000000000000
> 
> Termination Reason:    Namespace SIGNAL, Code 5 Trace/BPT trap: 5
> Terminating Process:   exc handler [2742]
> 
> Thread 0 Crashed::  Dispatch queue: com.apple.main-thread
> 0   ???                           	    0x54ae678506d3 ???
> 1   JavaScriptCore                	    0x7ff8288d33b8 llint_entry + 119182
> 2   JavaScriptCore                	    0x7ff8288d33b8 llint_entry + 119182
> 3   JavaScriptCore                	    0x7ff8288d33b8 llint_entry + 119182
> 4   JavaScriptCore                	    0x7ff8288d33b8 llint_entry + 119182
> 5   JavaScriptCore                	    0x7ff8288d33b8 llint_entry + 119182
> 6   JavaScriptCore                	    0x7ff8288d4251 llint_entry + 122919
> 7   JavaScriptCore                	    0x7ff8288d33b8 llint_entry + 119182
> 8   JavaScriptCore                	    0x7ff8288d33b8 llint_entry + 119182
> 9   JavaScriptCore                	    0x7ff8288d33b8 llint_entry + 119182
> 10  JavaScriptCore                	    0x7ff8288b6027 vmEntryToJavaScript + 216
> 11  JavaScriptCore                	    0x7ff829057161 JSC::Interpreter::executeCall(JSC::JSGlobalObject*, JSC::JSObject*, JSC::CallData const&, JSC::JSValue, JSC::ArgList const&) + 545
> 1

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [] All OS
- - [ ] Windows
[...]

PR #771, opened 2023-08-18
Fixing the Item height for Table and Tree when the default font is set

elsazac:
Fixes: #677  

A new function named as `getInset()` is created in `Table.java`. This function calculates the inset value of a cell in a table for aarch64 architecture. The inset value of a row in a table is calculated by subtracting the height of a cell from a frame.This calculated inset value is added to the height in `setItemHeight()` function.

The inset value of 1 is retained for x86_64 architecture.

/cc @lshanmug

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/List.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/List.java
@@ -50,6 +50,9 @@ public class List extends Scrollable {
 
 	static final int CELL_GAP = 1;
 
+	/* Vertical cell padding for list item */
+	static final int VERTICAL_CELL_PADDING= 8;
+
 /**
  * Constructs a new instance of this class given its parent
  * and a style value describing its behavior and appearance.
@@ -1179,7 +1182,7 @@ void setFont (NSFont font) {
 	super.setFont (font);
 	double ascent = font.ascender ();
 	double descent = -font.descender () + font.leading ();
-	((NSTableView)view).setRowHeight ((int)Math.ceil (ascent + descent) + 1);
+	((NSTableView)view).setRowHeight ((int)Math.ceil (ascent + descent) + VERTICAL_CELL_PADDING);
 	setScrollWidth();
 }
 
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Table.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Table.java
@@ -93,6 +93,9 @@ public class Table extends Composite {
 	static final int TEXT_GAP = 2;
 	static final int CELL_GAP = 1;
 
+	/* Vertical cell padding for table item */
+	static final int VERTICAL_CELL_PADDING= 8;
+
 /**
  * Constructs a new instance of this class given its parent
  * and a style value describing its behavior and appearance.
@@ -2822,7 +2825,7 @@ void setItemHeight (Image image, NSFont font, boolean set) {
 	if (font == null) font = getFont ().handle;
 	double ascent = font.ascender ();
 	double descent = -font.descender () + font.leading ();
-	int height = (int)Math.ceil (ascent + descent) + 1;
+	int height = (int)Math.ceil (ascent + descent) + VERTICAL_CELL_PADDING;
 	Rectangle bounds = image != null ? image.getBounds () : imageBounds;
 	if (bounds != null) {
 		imageBounds = bounds;
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Tree.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Tree.java
@@ -109,6 +109,9 @@ public class Tree extends Composite {
 	static final int TEXT_GAP = 2;
 	static final int CELL_GAP = 1;
 
+	/* Vertical cell padding for tree item */
+	static final int VERTICAL_CELL_PADDING= 8;
+
 /**
  * Constructs a new instance of this class given its parent
  * and a style value describing its behavior and appearance.
@@ -3170,7 +3173,7 @@ void setItemHeight (Image image, NSFont font, boolean set) {
 	if (font == null) font = getFont ().handle;
 	double ascent = font.ascender ();
 	double descent = -font.descender () + font.leading ();
-	int height = (int)Math.ceil (ascent + descent) + 1;
+	int height = (int)Math.ceil (ascent + descent) + VERTICAL_CELL_PADDING;
 	Rectangle bounds = image != null ? image.getBounds () : imageBounds;
 	if (bounds != null) {
 		imageBounds = bounds;


Phillipus:
Hi, I tested the patch. If there are no items the heights are too small:

s

```java
import org.eclipse.jface.layout.GridDataFactory;
import org.eclipse.swt.SWT;
import org.eclipse.swt.widgets.Table;
import org.eclipse.swt.widgets.Tree;

public class FontTest {

        shell.setText("Tree & Table Font");
        
        Tree tree = new Tree(shell, SWT.BORDER | SWT.V_SCROLL | SWT.H_SCROLL);
        GridDataFactory.create(GridData.FILL_BOTH).applyTo(tree);
        tree.setLinesVisible(true);
        
        Table table = new Table(shell, SWT.BORDER | SWT.V_SCROLL | SWT.H_SCROLL);
        GridDataFactory.create(GridData.FILL_BOTH).applyTo(table);
        table.setLinesVisible(true);
        
        shell.setSize(500, 450);
        shell.open();
        
        tree.setFont(tree.getFont());
        table.setFont(table.getFont());
        
        while(!shell.isDisposed()) {
            if(!display.readAndDispatch()) display.sleep();
        }
        
        display.dispose();
    }
}
```

elsazac:
@Phillipus 

Can you test this patch?

Phillipus:
> @Phillipus
> 
> Can you test this patch?

Hi @elsazac,

I've tested the patch on Mac M1 Ventura 13.6 and it does fix the row height problem testing with the Snippet in #771 and also in our RCP application.

I remain puzzled, though, how and where the initial row height is set (25 pixels) before setting the font of the table or tree.

elsazac:
> @elsazac The problem also occurs for the `List` component. I've added a Snippet to the original issue [here](https://github.com/eclipse-platform/eclipse.platform.swt/issues/677#issuecomment-1746844294). This will require a similar fix at line 1182 here:
> 
> https://github.com/eclipse-platform/eclipse.platform.swt/blob/17dcab01f437e21052189dc5cd394856868ad197/bundles/org.eclipse.swt/Eclipse%20SWT/cocoa/org/eclipse/swt/widgets/List.java#L1178-L1184

@Phillipus I have made a similar change in List as well.
Could you test it?

Phillipus:
> @Phillipus I have made a similar change in List as well. Could you test it?

Hi @elsazac - I tested and the fix works OK, thank-you.

But before committing this patch I have a doubt concerning what should be the true initial row height. See comment here - https://github.com/eclipse-platform/eclipse.platform.swt/issues/677#issuecomment-1750292461

elsazac:
> > @Phillipus I have made a similar change in List as well. Could you test it?
> 
> Hi @elsazac - I tested and the fix works OK, thank-you.
> 
> But before committing this patch I have a doubt concerning what should be the true initial row height. See comment here - [#677 (comment)](https://github.com/eclipse-platform/eclipse.platform.swt/issues/677#issuecomment-1750292461)

@Phillipus I have responded in the linked issue. Can you review and help progress this PR?

merks:
It looks like this current incarnation sprinkles the appropriate magic numbers to fix the problem.  I think https://github.com/eclipse-platform/eclipse.platform.swt/issues/677#issuecomment-1846595468 answers the outstanding problem.

@Phillipus Are there any outstanding concerns with this PR?

Phillipus:
> @Phillipus Are there any outstanding concerns with this PR?

@merks Yes.

The larger item height is seen on macOS `aarch64` but not on `x86_64` when Eclipse or an RCP app is run from its launcher. As explained by @sravanlakkimsetti in https://bugs.eclipse.org/bugs/show_bug.cgi?id=574932#c2:

>  This is because eclipse launcher SDK version
>  On x86_64 SDK used is 10.14
>  On aarch64 SDK used is 11
>  But the way forward is to use latest SDK and update SWT to MacOS 11 SDK.

However, I recently installed a new JDK on an Intel Mac (x86_64) that has been compiled against Mac SDK 11.1 and when launching a child Eclipse, or our RCP app, the item height is also bigger. This is because when running a child Eclipse it uses the `java` binary, not the launcher, and this one has been compiled using 11.1 SDK.

So the inconsistency is due not to OS architecture but on the linked SDK version of the launcher or the `java` binary and perhaps we should consider this needs to be addressed by rebuilding the `x86_64` launcher against MacOS 11 SDK:

>  But the way forward is to use latest SDK and update SWT to MacOS 11 SDK.

I don't know how that is done (is it the equinox binaries project?)

If this is done, however, note that Eclipse and RCP apps will not be able to run on macOS < 11 (before Big Sur)

merks:
I see!  Thank you for the detailed information!!  Magic numbers never do feel quite right.  😱

sravanlakkimsetti:
> I don't know how that is done (is it the equinox binaries project?)

I prefer not to have different code for x86 and arm implementation. My suggestion would be to use same masOS sdk for both versions of launcher. 

[Here is the section from equinox project](https://github.com/eclipse-equinox/equinox/blob/master/features/org.eclipse.equinox.executable.feature/library/cocoa/build.sh#L66-L74) where we switch between macos 11 and macos 10.14 sdk for arm and intel architectures

Removing this if condition and and rebuilding launcher should make make both versions of launcher work in the same way.

If you need to upgrade to latest version of macos sdk you'll need to work with foundation to get it upgraded on the [macos build machine](https://ci.eclipse.org/releng/computer/nc1ht-macos11-arm64)

Phillipus:
@sravanlakkimsetti Thanks for that information!

I notice that you did already do this here:

https://github.com/eclipse-equinox/equinox/commit/0dad4635cce80cdc9a16764b8470b56ee421d058

but reverted this:

https://github.com/eclipse-equinox/equinox/commit/4db53c06417fc1c72955a1477c4f9cf658e45f91

Did you revert because it was no longer needed die to https://github.com/eclipse-platform/eclipse.platform.swt/issues/452 fixing itself?

Phillipus:
Opened https://github.com/eclipse-equinox/equinox/issues/473

sravanlakkimsetti:
> @sravanlakkimsetti Thanks for that information!
> 
> I notice that you did already do this here:
> 
> [eclipse-equinox/equinox@0dad463](https://github.com/eclipse-equinox/equinox/commit/0dad4635cce80cdc9a16764b8470b56ee421d058)
> 
> but reverted this:
> 
> [eclipse-equinox/equinox@4db53c0](https://github.com/eclipse-equinox/equinox/commit/4db53c06417fc1c72955a1477c4f9cf658e45f91)
> 
> Did you revert because it was no longer needed due to #452 fixing itself?

Not exactly. The problem was slightly complicated. At that time the supported java version was built with 10.14. When we moved to Macos 11, Eclipse had one behavior and the child, eclipse launched using run menu, had different behavior. For this reason we wanted to wait till java moved to Macos 11 before we move. 

Now that java also moved to Macos 11 we can move mow.  Lets reinstate https://github.com/eclipse-equinox/equinox/commit/0dad4635cce80cdc9a16764b8470b56ee421d058

Phillipus:
> Lets reinstate https://github.com/eclipse-equinox/equinox/commit/0dad4635cce80cdc9a16764b8470b56ee421d058

Could you do that?

sravanlakkimsetti:
> > Lets reinstate [eclipse-equinox/equinox@0dad463](https://github.com/eclipse-equinox/equinox/commit/0dad4635cce80cdc9a16764b8470b56ee421d058)
> 
> Could you do that?

Sure will take it up

Phillipus:
> Now that java also moved to Macos 11

For the Temurin version of Java this was a very recent move.

Temurin Java 17 moved to SDK 11.1 with version [jdk-17.0.9+9](https://github.com/adoptium/temurin17-binaries/releases/tag/jdk-17.0.9%2B9) - October 19 2023

Temurin Java 21 moved to SDK 11.1 with version [jdk-21.0.1+12](https://github.com/adoptium/temurin21-binaries/releases/tag/jdk-21.0.1%2B12) - October 24 2023

Phillipus:
Something to be aware of regarding using SDK 11 on the macOS Intel launcher in the future:

This patch will need to change from:

```java
int height = (int)Math.ceil (ascent + descent) + ((OS.IS_X86_64) ? 1 : 8);
```
to

```java
int height = (int)Math.ceil (ascent + descent) + 8;
```

And this means that an end user will also see the increased item height if they are using an older JDK that was compiled with 10.14.

Phillipus:
It turns out that the Mac x86_64 launcher has been built with SDK 13 since around Eclipse 4.30. So we don't need to wait.

@elsazac Could you update your patch by removing the `OS.IS_X86_64` condition so it applies to both x86_64 and aarch64, please?

elsazac:
> @elsazac Could you update your patch by removing the `OS.IS_X86_64` condition so it applies to both x86_64 and aarch64, please?

done.

Phillipus:
> done.

Thanks. I tested this on both x86_64 and aarch64 and it works OK.

However, a user will see discrepancies in Item height if they are launching a child Eclipse/RCP app using an older Java version (one compiled with Mac SDK 10.14). I guess in this case they should update to latest version.

I'm happy with this PR now. I don't know who should commit it.

sravanlakkimsetti (approved):

Phillipus:
Is this ready to be merged now?

Phillipus:
Be aware that because of the increased item height in Lists, Tables and Trees, text is not vertically cantered in the `List` component (run this [Snippet](https://github.com/eclipse-platform/eclipse.platform.swt/issues/677#issuecomment-1746844294) to see) nor in Cell Editors (`TableEditor` and `TreeEditor`). It looks kind of ugly but I guess that would need further investigation (I tried but failed)

Phillipus:
@elsazac In three places a value of 8 is added (it was 1). Do you think it would be a good idea to create a static constant in each class for this? Something like the existing `static final int CELL_GAP = 1` but with a new name? And add a comment to explain what it is for?

elsazac:
> @elsazac In three places a value of 8 is added (it was 1). Do you think it would be a good idea to create a static constant in each class for this? Something like the existing `static final int CELL_GAP = 1` but with a new name maybe `CELL_PADDING`)? And add a comment to explain what it is for?

I have added the changes and squashed the commits. Please have a look.

merks:
Nit picky detail... The comment `Row height for xxx item` is not really a correct/accurate description.  This is padding that is added to the height as determined by the font metrics, so `Row height padding for ...`

elsazac:
> Nit picky detail... The comment `Row height for xxx item` is not really a correct/accurate description. This is padding that is added to the height as determined by the font metrics, so `Row height padding for ...`

I have updated the comments.

merks:
Yes, that's a better more descriptive name too!  😄

Phillipus:
Thanks @elsazac for your work on this, and thanks @merks for the feedback and merge. :-)

PR #778, opened 2023-08-21
Disable unstable test_setUrl_local() on Mac

iloveeclipse:
See https://github.com/eclipse-platform/eclipse.platform.swt/issues/722


HeikoKlare (approved):
Makes sense to disable the test until it is fixed and keep the according issue open to track the (temporarily) disabled test.

Issue #687, opened 2023-06-05
Line dashes on Windows 200% display scale (hi-dpi) are half-size


Phillipus:
This affects Windows only, the problem is not seen on Mac or Linux.

Dashes on lines drawn on an SWT `GC` using `setLineDash` are half-size on Windows 200% scaling.

- Windows 10 or 11
- Java 17
- Any version of SWT

Snippet:

```java
import org.eclipse.swt.graphics.Color;
import org.eclipse.swt.layout.FillLayout;
import org.eclipse.swt.widgets.Display;
import org.eclipse.swt.widgets.Shell;

public class LineDashesSWT {

    public static void main(String[] args) {
        Display display = new Display();
        Shell shell = new Shell(display);
        shell.setLayout(new FillLayout());
        shell.setSize(220, 180);
        shell.setBackground(new Color(null, 255, 255, 255));

        
        shell.addPaintListener(e -> {
            e.gc.setLineDash(new int[] { 3 });
            e.gc.drawRectangle(10, 10, 180, 80);
            e.gc.drawLine(10, 50, 190, 50);
        });
        
        shell.open();

        while(!shell.isDisposed()) {
            if(!display.readAndDispatch()) {
                display.sleep();
            }
        }
    }
}
```
On Windows 100% scaling (this is how it should render):

![win1x](https://github.com/eclipse-platform/eclipse.platform.swt/assets/600504/7c0afa73-2fc0-45d8-93e0-dcf6895d965d)

On Windows 200% scaling:

![win2x](https://github.com/eclipse-platform/eclipse.platform.swt/assets/600504/36c95526-061d-4cb6-b0df-3d9c8be51132)

Compare with Mac and Linux:

Linux 100%:

![linux1x](https://github.com/eclipse-platform/eclipse.platform.swt/assets/600504/b71db41a-3009-499d-8210-bd4ab70232c5)

Linux 200%:

![linux2x](https://github.com/eclipse-platform/eclipse.platform.swt/assets/600504/1ea88e53-a2b7-4b08-a6cc-cfd4ea893161)

Mac 100%:

![mac1x](https://github.com/eclipse-platform/eclipse.platform.swt/assets/600504/602e3ba0-03e7-494c-872b-abb2b55e9e18)

Mac 200%:

![mac2x](https://github.com/eclipse-platform/eclipse.platform.swt/assets/600504/4e11574c-7ed3-4c10-a97b-5e04a54425af)

[...]

PR #781, opened 2023-08-24
Adjust line dashes on hi-dpi device

Phillipus:
- See https://github.com/eclipse-platform/eclipse.platform.swt/issues/687

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
@@ -3580,6 +3580,9 @@ public int getInterpolation() {
 public LineAttributes getLineAttributes () {
 	LineAttributes attributes = getLineAttributesInPixels();
 	attributes.width = DPIUtil.autoScaleDown(drawable, attributes.width);
+	if(attributes.dash != null) {
+		attributes.dash = DPIUtil.autoScaleDown(drawable, attributes.dash);
+	}
 	return attributes;
 }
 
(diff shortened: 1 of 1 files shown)


HeikoKlare (commented):
Thanks for that HiDPI scaling improvement!
From what I have tested so far, the fix works quite well. I have two remarks:
1. I would propose to improve the scaling calculation as mentioned in the detailed comments to avoid unnecessary computation. That also fixes one unnecessary source of rounding error. And it also allows to completely remove the `DPIUtil` extensions.
2. Shouldn't the `getLineDashes()` method also be adapted? The analogous `getLineWidth()` method also scales down the value. I would expect that the following is true, but is is not: `gc.setLineDash(gc.getLineDash()).equals(gc.getLineDash())`. You can also see this when testing the `CTable` with dashes applied to the underlying GC. You can reproduce it with the following initialization added to the snipped given in the associated issue #687.

```java
CTable table = new CTable(shell, SWT.BORDER) {
	@Override
	void onPaint(org.eclipse.swt.widgets.Event event) {
		event.gc.setLineDash(new int[] { 3 });
		super.onPaint(event);
	};
};
table.setLayoutData(new GridData(SWT.FILL, SWT.FILL, true, true));
table.setLinesVisible(true);
for (int col = 0; col < 3; col++) {
	CTableColumn column = new CTableColumn(table, SWT.NONE);
	column.setWidth(50);
}
```

Phillipus:
@HeikoKlare That's a lot to process. Would it be better if you created a PR?

HeikoKlare:
@Phillipus If you like, I can also provide a PR. Since you did the initial work for the improvement, I just wanted to give feedback to improve the PR, but I can also propose the changes as a whole in a separate PR. Just let me know.

HeikoKlare (approved):
Thanks for incorporating the suggestions, @Phillipus! From my side, the changes look good now and in my manual tests they have worked as expected.

Since the `GC` class is lacking tests, I have proposed the addition of some tests in #786, which also includes tests that serve as regression tests for this PR (in particular the symmetry of the get and set functions).

PR #785, opened 2023-08-29
CTabFolder: Add 'showSelectedImage' property

shubhamWaghmare-sap:
- Added new property "showSelectedImage" to CTabFolder parallel to existing property "showUnselectedImage"

- Image rendering in drawSelected method updated in CTabFolderRenderer with handling the new property "showSelectedImage"

- Added standard padding for text of tabs when tab icon is either not available or is hidden (both for selected and unselected tabs)

- Pre-requisite for [issue](https://github.com/eclipse-platform/eclipse.platform.ui/issues/735) and pull request : https://github.com/eclipse-platform/eclipse.platform.ui/pull/1071

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
@@ -196,6 +196,7 @@ public class CTabFolder extends Composite {
 	int[] gradientPercents;
 	boolean gradientVertical;
 	boolean showUnselectedImage = true;
+	boolean showSelectedImage = true;
 
 	// close, min/max and chevron buttons
 	boolean showClose = false;
@@ -1365,6 +1366,18 @@ public boolean getUnselectedCloseVisible() {
 public boolean getUnselectedImageVisible() {
 	checkWidget();
 	return showUnselectedImage;
+}
+/**
+ * Returns <code>true</code> if an image appears
+ * in selected tabs.
+ *
+ * @return <code>true</code> if an image appears in selected tabs
+ *
+ * @since 3.125
+ */
+public boolean getSelectedImageVisible() {
+	checkWidget();
+	return showSelectedImage;
 }
 /**
  * Return the index of the specified tab or -1 if the tab is not
@@ -3691,6 +3704,25 @@ public void setUnselectedImageVisible(boolean visible) {
 	if (showUnselectedImage == visible) return;
 	// display image on unselected items
 	showUnselectedImage = visible;
+	updateFolder(REDRAW);
+}
+/**
+ * Specify whether the image appears on selected tabs.
+ *
+ * @param visible <code>true</code> makes the image appear
+ *
+ * @exception SWTException <ul>
+ *    <li>ERROR_WIDGET_DISPOSED - if the receiver has been disposed</li>
+ *    <li>ERROR_THREAD_INVALID_ACCESS - if not called from the thread that created the receiver</li>
+ * </ul>
+ *
+ * @since 3.125
+ */
+public void setSelectedImageVisible(boolean visible) {
+	checkWidget();
+	if (showSelectedImage == visible) return;
+	// display image on selected items
+	showSelectedImage = visible;
 	updateFolder(REDRAW);
 }
 /**
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
@@ -98,6 +98,7 @@ public class CTabFolderRenderer {
 	static final int ITEM_LEFT_MARGIN = 4;
 	static final int ITEM_RIGHT_MARGIN = 4;
 	static final int INTERNAL_SPACING = 4;
+	static final int TABS_WITHOUT_ICONS_PADDING = 14;
 	static final int FLAGS = SWT.DRAW_TRANSPARENT | SWT.DRAW_MNEMONIC | SWT.DRAW_DELIMITER;
 	static final String ELLIPSIS = "..."; //$NON-NLS-1$
 	private static final String CHEVRON_ELLIPSIS = "99+"; //$NON-NLS-1$
@@ -318,10 +319,11 @@ public class CTabFolderRenderer {
 					Image image = item.getImage();
 					if (image != null && !image.isDisposed()) {
 						Rectangle bounds = image.getBounds();
-						if ((state & SWT.SELECTED) != 0 || parent.showUnselectedImage) {
+						if (((state & SWT.SELECTED) != 0 && parent.showSelectedImage)
+								|| ((state & SWT.SELECTED) == 0 && parent.showUnselectedImage)) {
 							width += bounds.width;
 						}
-						height =  bounds.height;
+						height = bounds.height;
 					}
 					String text = null;
 					if ((state & MINIMUM_SIZE) != 0) {
@@ -355,9 +357,17 @@ public class CTabFolderRenderer {
 							gc.setFont(gcFont);
 						}
 					}
+
+					width += getTextPadding(item, state) * 2;
+
 					if (parent.showClose || item.showClose) {
 						if ((state & SWT.SELECTED) != 0 || parent.showUnselectedClose) {
-							if (width > 0) width += INTERNAL_SPACING;
+							if (((state & SWT.SELECTED) != 0 && parent.showSelectedImage)
+									|| ((state & SWT.SELECTED) == 0 && parent.showUnselectedImage)) {
+								if (width > 0) width += INTERNAL_SPACING;
+							} else {
+								if (width > 0) width -= INTERNAL_SPACING;
+							}
 							width += computeSize(PART_CLOSE_BUTTON, SWT.NONE, gc, SWT.DEFAULT, SWT.DEFAULT).x;
 						}
 					}
@@ -368,6 +378,27 @@ public class CTabFolderRenderer {
 		width = trim.width;
 		height = trim.height;
 		return new Point(width, height);
+	}
+
+	/**
+	 * Returns padding for the text of a tab when image is not available or is hidden.
+	 *
+	 * @param item CTabItem
+	 * @param state current state
+	 *
+	 */
+	private int getTextPadding(CTabItem item, int state) {
+		CTabFolder parent = item.getParent();
+		Image image = item.getImage();
+		String text = item.getText();
+
+		if (text != null && parent.getMinimumCharacters() != 0) {
+			if (image == null || image.isDisposed() || ((state & SWT.SELECTED) != 0 && !parent.showSelectedImage)
+					|| ((state & SWT.SELECTED) == 0 && !parent.showUnselectedImage))
+				return TABS_WITHOUT_ICONS_PADDING;
+		}
+
+		return 0;
 	}
 
 	/**
@@ -1385,7 +1416,7 @@ public class CTabFolderRenderer {
 			int xDraw = x - trim.x;
 			if (parent.single && (parent.showClose || item.showClose)) xDraw += item.closeRect.width;
 			Image image = item.getImage();
-			if (image != null && !image.isDisposed()) {
+			if (image != null && !image.isDisposed() && parent.showSelectedImage) {
 				Rectangle imageBounds = image.getBounds();
 				// only draw image if it won't overlap with close button
 				int maxImageWidth = rightEdge - xDraw - (trim.width + trim.x);
@@ -1400,6 +1431,7 @@ public class CTabFolderRenderer {
 			}
 
 			// draw Text
+			xDraw += getTextPadding(item, state);
 			int textWidth = rightEdge - xDraw - (trim.width + trim.x);
 			if (!parent.single && item.closeRect.width > 0) textWidth -= item.closeRect.width + INTERNAL_SPACING;
 			if (textWidth > 0) {
@@ -1613,6 +1645,7 @@ public class CTabFolderRenderer {
 				}
 			}
 			// draw Text
+			xDraw += getTextPadding(item, state);
 			int textWidth = x + width - xDraw - (trim.width + trim.x);
 			if (parent.showUnselectedClose && (parent.showClose || item.showClose)) {
 				textWidth -= item.closeRect.width + INTERNAL_SPACING;
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_custom_CTabFolder.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_custom_CTabFolder.java
@@ -468,6 +468,18 @@ public void test_iconWrappedOnNextLine() {
 	}
 }
 
+@Test
+public void test_selectedImageVisible() {
+	createTabFolder(null);
+
+	ctabFolder.setSelectedImageVisible(true);
+	assertTrue(ctabFolder.getSelectedImageVisible());
+
+	ctabFolder.setSelectedImageVisible(false);
+	assertFalse(ctabFolder.getSelectedImageVisible());
+
+}
+
 private void processEvents() {
 	Display display = shell.getDisplay();
 


bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
```diff
@@ -1406,7 +1411,9 @@ void drawSelected(int itemIndex, GC gc, Rectangle bounds, int state ) {
 				}
 			}
 
-			// draw Text
+			// draw Text			
+			//Add some padding before text, if image is not visible
+			if(!parent.showSelectedImage) xDraw += INTERNAL_SPACING;
```
sratz, line 1416: is this not covered by the condition above already where spacing is added ?

sratz (changes requested):

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
```diff
@@ -3694,6 +3707,25 @@ public void setUnselectedImageVisible(boolean visible) {
...
+ * Specify whether the image appears on selected tabs.
+ *
+ * @param visible <code>true</code> makes the image appear
+ *
+ * @exception SWTException <ul>
+ *    <li>ERROR_WIDGET_DISPOSED - if the receiver has been disposed</li>
+ *    <li>ERROR_THREAD_INVALID_ACCESS - if not called from the thread that created the receiver</li>
+ * </ul>
+ *
+ * @since 3.124
```
HeikoKlare, line 3720: I think we need a version bump to 3.125 because of API extension here, don't we? 3.124.0 and 3.124.100 won't contain this API.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
HeikoKlare: As far as I understand, the calculation above is for the size of the tab and this is for the placement of the text within it. Still I do not really like that the code is explained via so many comments (even though it is better compared having no explanation). I would prefer to have methods like `addTextPadding(...)` that consistently adds the padding whereever needed (i.e., both in the size calculation as well as in the text placement). That both avoids inconsistencies in the code and makes the code more comprehensible (and the comments obsolete).

HeikoKlare (commented):
Thanks for this extension.
I guess the naming like `getSelectedImageVisible()` instead of `isSelectedImageVisible()` is because of consistency with the existing API, isn't it? Apart from that, I only have some minor comments.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
shubhamWaghmare-sap: Yes. We will bump it in a new change.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
```diff
@@ -318,9 +318,14 @@ protected Point computeSize (int part, int state, GC gc, int wHint, int hHint) {
 					Image image = item.getImage();
 					if (image != null && !image.isDisposed()) {
 						Rectangle bounds = image.getBounds();
-						if ((state & SWT.SELECTED) != 0 || parent.showUnselectedImage) {
-							width += bounds.width;
+						//If image is not visible, tab width should not include image width
```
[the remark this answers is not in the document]
shubhamWaghmare-sap: Sure. We'll re-format it. 
Also, we are working on extracting this logic into a separate method  `getTextPadding(...) ` for uniform padding calculation when image is not visible. This will simplify the code.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
```diff
@@ -318,9 +318,14 @@ protected Point computeSize (int part, int state, GC gc, int wHint, int hHint) {
...
+						//Additionally, tab width is increased by standard padding (to add padding to the text) 
+						if ((state & SWT.SELECTED) != 0){
+							width += parent.showSelectedImage ? bounds.width : INTERNAL_SPACING;
```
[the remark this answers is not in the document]
shubhamWaghmare-sap: @sratz We are working on extracting this logic into a separate method  `getTextPadding(...) ` for uniform padding calculation when image is not visible. This will simplify the code.

@HeikoKlare The new method `getTextPadding(...) ` will simplify it. We would be modifying the logic to increase the tab width by standard padding, for both selected and unselected tab.

Regarding the padding in line 344, it does not consider whether the image is visible or not. If image is not visible for a tab, there is very small space before the text, which makes it difficult to distinguish between tabs. Thus additional padding is to be added when image of a tab is not visible.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
shubhamWaghmare-sap: @sratz  As pointed out by @HeikoKlare, to add padding before the text, the width of the tab should also be increased otherwise the text would run out of the tab. Thus this change is to shift (pad) text by a standard width and same width size is added to the tab in the condition above.

@HeikoKlare We are working on extracting this logic into a separate method getTextPadding(...)  for uniform padding calculation when image is not visible.

HeikoKlare:
@shubhamWaghmare-sap Looks like you have addressed the review comments. Are you still working on this PR or it is already worth to re-review?

shubhamWaghmare-sap:
> @shubhamWaghmare-sap Looks like you have addressed the review comments. Are you still working on this PR or it is already worth to re-review?

@HeikoKlare Yes. We have incorporated the review comments in this PR. It is ready to be re-reviewed now. Thanks. 
We do have some open questions like a better name for the method `getTextPadding(...)` added in `CTabFolderRenderer` and the need for text padding for unselected tabs as well, when image is hidden or not available.

Hoping to get answers via discussions on PR. 👍

A short overview of changes:

- Text padding added in both selected and unselected tab when image is hidden or not available
- In case there is no text for the tab, no padding is added.
- Refactoring.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
```diff
@@ -355,6 +356,7 @@ protected Point computeSize (int part, int state, GC gc, int wHint, int hHint) {
 							gc.setFont(gcFont);
 						}
 					}
+					width += getTextPadding(item, state);
```
HeikoKlare, line 359: I would suggest to apply the padding two times to the overall width, so that the text is centered within the tab (as the padding is applied one time to the x offset of the text).

```suggestion
					width += getTextPadding(item, state) * 2;
```

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
```diff
@@ -355,6 +356,7 @@ protected Point computeSize (int part, int state, GC gc, int wHint, int hHint) {
...
 					if (parent.showClose || item.showClose) {
 						if ((state & SWT.SELECTED) != 0 || parent.showUnselectedClose) {
 							if (width > 0) width += INTERNAL_SPACING;
```
HeikoKlare, line 362: To me, it feels strange if this internal spacing is also applied when no images are shown, as it leads to a large padding right to the text, especially in case the padding has a significantly higher value.

```suggestion
						if (((state & SWT.SELECTED) != 0 && parent.showSelectedImage)
								|| ((state & SWT.SELECTED) == 0 && parent.showUnselectedImage)) {
							if (width > 0) width += INTERNAL_SPACING;
						}
						if ((state & SWT.SELECTED) != 0 || parent.showUnselectedClose) {
```

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
```diff
@@ -370,6 +372,27 @@ protected Point computeSize (int part, int state, GC gc, int wHint, int hHint) {
...
+	 */
+	private int getTextPadding(CTabItem item, int state) {
+		CTabFolder parent = item.getParent();
+		Image image = item.getImage();
+		String text = item.getText();
+
+		if (text != null && parent.getMinimumCharacters() != 0) {
+			if (image == null || image.isDisposed() || ((state & SWT.SELECTED) != 0 && !parent.showSelectedImage)
+					|| ((state & SWT.SELECTED) == 0 && !parent.showUnselectedImage))
+				return INTERNAL_SPACING;
```
HeikoKlare, line 390: See general comment: I propose to define the padding when not showing icons separately.
```suggestion
				return TABS_WITHOUT_ICONS_PADDING;
```

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_custom_CTabFolder.java
```diff
@@ -468,6 +468,20 @@ public void test_iconWrappedOnNextLine() {
...
+@Test
+public void test_selectedImageVisible() {
+	createTabFolder(null);
+
+	ctabFolder.setSelectedImageVisible(true);
+	assertTrue( ctabFolder.getSelectedImageVisible() );
+
+	ctabFolder.setSelectedImageVisible(false);
+	assertFalse( ctabFolder.getSelectedImageVisible());
+
```
HeikoKlare, line 480: Please correct formatting here. There are strange whitespaces that will be corrected with auto-formatting.

HeikoKlare (commented):
Thanks for the improvements to this PR and sorry for the late reply.

In general, I am fine with the changes. I only have two concerns with respect the padding:
1. I would propose to separate the padding for the case that icons are not shown from the existing `INTERAL_SPACING` (in a variable such as `TABS_WITHOUT_ICONS_PADDING`) as they apply to different behaviors and should thus not influence each other.
2. I would even be in favor of having a bigger spacing when no icons are shown. This seems to require 1. to work properly (see my detailed comments).

To me, a value like 14 for such a `TABS_WITHOUT_ICONS_PADDING` feels good :
image
But that may be a matter of personal preference, it might also be smaller but at least should be somehow independent from the existing spacing.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
shubhamWaghmare-sap: The INTERNAL_SPACING here is applied in case the _close_ is visible for tabs. I believe this has to be there irrespective of the current change.
Also, the spacing is applied **only** when no images are shown. The primary objective of applying the padding is to have proper visual separation between tabs when icons/images are not visible.

I believe there is a gap in my understanding.
@HeikoKlare Could you please elaborate a little in what you wanted to convey via this suggested change?

HeikoKlare: Sorry for the confusion. It's not a gap in your understanding, just something must have went wrong when I added the suggestion. The actual suggestion should be different. Let me explain it textually: The code here does two things: first, it adds the `INTERNAL_SPACING` to the width; seconds, it adds the size of the close button. The latter must _always_ be done, that is correct. But the former one is, from my understanding, supposed to ensure that between the text and the close button there is some spacing. However, when icons are disabled there is already a large gap and adding this addition spacing makes the padding to the left and right of the text uneven.

This is a screenshot with additional spacing (current state):
image

And this without (my proposal):
image

It is only a small different, but it makes the text look slightly uncentered (because it actually is uncentered).

One note: my explanation is conflicting with your statement:
> Also, the spacing is applied only when no images are shown.

Maybe I missed something in the code, but my understand is that this code will always be excuted when the tab to be drawn has a close icon.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
shubhamWaghmare-sap: In that case, code is reverted to adding the the padding twice uniformly to the tab width. 
Also, in case close button is visible for unselected tab, the close button size is still added to tab width. If we remove the close button size, the text gets truncated and the padding would need to be increased.

With the goal to center the text within tab, close button size wouldn't matter much.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
shubhamWaghmare-sap: @HeikoKlare Thanks for clarifying. The code is now updated accordingly and the tab text is now centered.

`Before`
image

`After`
image
> Also, the spacing is applied only when no images are shown.

>> Maybe I missed something in the code, but my understand is that this code will always be excuted when the tab to be drawn has a close icon.

You are right. The code here gets executed when tab has a close icon. I misunderstood it before.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
```diff
@@ -355,9 +357,15 @@ protected Point computeSize (int part, int state, GC gc, int wHint, int hHint) {
...
+
+					width += getTextPadding(item, state) * 2;
+
 					if (parent.showClose || item.showClose) {
 						if ((state & SWT.SELECTED) != 0 || parent.showUnselectedClose) {
-							if (width > 0) width += INTERNAL_SPACING;
+							if (((state & SWT.SELECTED) != 0 && parent.showSelectedImage)
+									|| ((state & SWT.SELECTED) == 0 && parent.showUnselectedImage)) {
+								if (width > 0) width += INTERNAL_SPACING;
+							}
```
HeikoKlare, line 368: This is rather nitpicky, but it feels to me like there needs to be less spacing here when no images are shown:
```suggestion
							} else {
								if (width > 0) width -= INTERNAL_SPACING;
							}
```
Actually, it is not about `INTERNAL_SPACING` to be subtracted but about the margin of the close button (which is equal to `INTERNAL_SPACING = 4`). This feels more natural to me, as I would expect the distance between left end of the tab and the beginning of the text should be equal to the distance between the end of the text and the beginning of the icon (and not until the beginning of the icon's margin).

HeikoKlare (commented):
The changes look good to me now. I have one nitpicky finding with respect to the text padding left.
@shubhamWaghmare-sap please also squash all your commits into one in preparation of merging this PR.
@sratz you have an open change request. Have your concerns be addressed or do you want to re-review?

Finally, I have one comment left but only for the sake of documentation, as I do not have a solution for that, so it should not affect this PR:
To me the behavior when showing the close button for unselected tabs does not yet feel right for me. The padding and text placement calculation is equal to when the close button is shown (for the selected tab). This is fine for the case when hovering over an unselected tab, but does not feel right for the other tabs (consider the "Tip of the Day" tab in the following screenshot). But since only one of these cases can be optimized without having tab headers jumping around, I do not see a better solution for this.
image

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
shubhamWaghmare-sap: Incorporated the suggestion and following is how it looks like after the change:
image

HeikoKlare (approved):
Thanks for this great contribution. Content-wise, the changes are ready to be merged from my side.

@shubhamWaghmare-sap Can you please squash the commits in this PR into one?

@sratz Can we remove your change request or is there anything left to do?

sratz (approved):

sratz:
Looks good.

@shubhamWaghmare-sap: Can you rebase on top of master?

shubhamWaghmare-sap:
@sratz @HeikoKlare 
Thanks for all the comments and suggestions and approving the changes.
I have squashed all commits and re-based on top of master.

HeikoKlare:
@shubhamWaghmare-sap Thank for this contribution! Looking forward to having the Platform UI PR merged and the feature available in Eclipse.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
```diff
@@ -1366,6 +1367,18 @@ public boolean getUnselectedImageVisible() {
 	checkWidget();
 	return showUnselectedImage;
 }
+/**
+ * Returns <code>true</code> if an image appears
+ * in selected tabs.
+ *
+ * @return <code>true</code> if an image appears in selected tabs
+ *
+ * @since 3.125
```
fedejeanne, line 1376: Shouldn't the Manifest and the Pom also be updated in this case?

HeikoKlare: The required version bump originally was part of this PR: https://github.com/eclipse-platform/eclipse.platform.swt/commit/cf4691d7d546f70b038a0c9b9bf74fe5603c219f#diff-6485b884863fd4f54a20976acef942fe0dd257eb4e72ca25533fd84b7b6fe2e6

But it obviously got lost. Interesting that the API tools baseline comparison did not complain.

I provided #944 to have the faulty version fixed soon.

HeikoKlare: Fixed

iloveeclipse: Here is related snippet from build.
https://ci.eclipse.org/releng/job/eclipse.platform.swt/job/PR-785/12/consoleText
@laeubi : I don't see API tools reporting anything except there is no baseline??

```
[2023-12-14T11:04:05.054Z] [INFO] --- tycho-source:4.0.5-SNAPSHOT:plugin-source (plugin-source) @ org.eclipse.swt ---
[2023-12-14T11:04:05.054Z] [INFO] 
[2023-12-14T11:04:05.054Z] [INFO] --- tycho-packaging:4.0.5-SNAPSHOT:package-plugin (default-package-plugin) @ org.eclipse.swt ---
[2023-12-14T11:04:05.311Z] [INFO] Building jar: /home/jenkins/agent/workspace/eclipse.platform.swt_PR-785/eclipse.platform.swt/bundles/org.eclipse.swt/target/org.eclipse.swt-3.124.300-SNAPSHOT.jar
[2023-12-14T11:04:05.577Z] [INFO] 
[2023-12-14T11:04:05.577Z] [INFO] --- tycho-p2:4.0.5-SNAPSHOT:p2-metadata-default (default-p2-metadata-default) @ org.eclipse.swt ---
[2023-12-14T11:04:05.578Z] [INFO] No baseline version MavenProject: org.eclipse.swt:org.eclipse.swt:3.124.300-SNAPSHOT @ /home/jenkins/agent/workspace/eclipse.platform.swt_PR-785/eclipse.platform.swt/bundles/org.eclipse.swt/pom.xml
[2023-12-14T11:04:05.578Z] [INFO] 
[2023-12-14T11:04:05.578Z] [INFO] --- tycho-p2-extras:4.0.5-SNAPSHOT:compare-version-with-baselines (compare-attached-artifacts-with-release) @ org.eclipse.swt ---
[2023-12-14T11:04:05.578Z] [INFO] Adding repository https://download.eclipse.org/eclipse/updates/4.30/R-4.30-202312010110
[2023-12-14T11:04:05.835Z] [INFO] 
[2023-12-14T11:04:05.835Z] [INFO] --------< org.eclipse.swt:org.eclipse.swt.fragments.localbuild >--------
```

laeubi: >  don't see API tools reporting anything except there is no baseline??

SWT projects have no comparable baseline because of the way they are build. This might change when LSF is enabled (or submodules) and we have one repository as then it would be possible to transform them into regular (tycho) projects.

merks:
Is it an unintended side effect that the tabs are so much wider now?

image

It doesn't look very nice...

HeikoKlare:
> Is it an unintended side effect that the tabs are so much wider now?

Yes. it is.

> It doesn't look very nice...

True.

The larger padding is intended to applied to tab folders that usually contain images but that may be disabled (such as tab folder of Eclipse views: eclipse-platform/eclipse.platform.ui#1071). But it currently applies to all tabs for which showing images is disabled _or_ that have no image assigned:

https://github.com/eclipse-platform/eclipse.platform.swt/blob/96692a51e54863a4670d88a78f0ab12043a07fa3/bundles/org.eclipse.swt/Eclipse%20SWT%20Custom%20Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java#L396-L398

The latter is true for all existing tab folders without images, such as the one in the screenshot. So the functionality added with this PR has to be improved, either by only applying the padding when images are explicitly disabled or by adding a propery for a tab folder that defines whether the new behavior (no images but larger padding) shall be enabled.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
HannesWell: Some work and expermients needs to be done, but I have the intention to do that during the Christmas new-years holidays.

HeikoKlare: Great to hear that you will work on that, @HannesWell, but don't forget to also use holidays as holidays 🙂

sratz:
> either by only applying the padding when images are explicitly disabled or by adding a propery for a tab folder that defines whether the new behavior (no images but larger padding) shall be enabled.

Alternatively: Would it make sense to expose

> TABS_WITHOUT_ICONS_PADDING

as a method rather than a constant (with backwards-compatible default value), such that the subclass in platform.ui can control (and potentially fine-tune via preferences) this beheavior?

HeikoKlare:
> Alternatively: Would it make sense to expose
> 
> > TABS_WITHOUT_ICONS_PADDING
> 
> as a method rather than a constant (with backwards-compatible default value), such that the subclass in platform.ui can control (and potentially fine-tune via preferences) this behavior?

Making the value configurable (potentially also via preferences) sound like a good idea to me. I was not aware that platform.ui has a specialization of this renderer, so providing the padding value via method and overwriting it in platform.ui would be possible. It could also be made "dynamically" configurable via some setter logic. Then you would not need to make a subclass for overwriting the value but can just do that by instantiating the renderer.

PR #787, opened 2023-08-29
Add a snippet to demonstrate GC transformation effects

HeikoKlare:
The snippet can be used to show how a pattern of multiple lines is printed with different line widths and different kinds of transformations applied to the GC (scaling and rotation).

### DIFF
--- a/examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet381.java
+++ b/examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet381.java
@@ -1,0 +1,210 @@
+/*******************************************************************************
+ * Copyright (c) 2023 Vector Informatik GmbH and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ *******************************************************************************/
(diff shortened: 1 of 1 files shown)


fedejeanne:
## In Mac
* _macOS Ventura (13.2.1)_
* M1 Chip

# Summary
* Tested in 2 Displays: integrated display and external (Dell) display
* I did "runs" with the following _Rotations_: 45°, 90°, 135°, 180°, 225°, 270°, 315°, 360° and 405°
* Before this PR, there are singularities (the arrow doesn't show) at rotations starting at 45° and increases of 90° _i.e_: 45°, 135°, 225°, 315° and 405°.
* Those singularities only happen with `Line Width = 0` --> _i.e._ starting at `Line Width = 0.1` **there are no singularities even before applying this PR**
* There are no singularities after applying this PR, regardless of the `Line Width` 🥇 
* _Use anti aliasing_ doesn't make any difference
* Applying _Horizontal / Vertical Scaling_ doesn't make any difference (I used a factor of **2.0** for each, both separately and simultaneously _i.e._ I did 3 runs)

# Some screenshots

## In "the big display": DELL U2518D (25'', 3840x2160)
### Before this PR
Bildschirm­foto 2023-11-30 um 08 10 29

### After this PR

Bildschirm­foto 2023-11-30 um 08 11 36

## In "the small integrated Display": Integrated Retina Display (13.3'', 2560x1600)

### Before this PR
Bildschirm­foto 2023-11-30 um 08 20 47

### After this PR
Bildschirm­foto 2023-11-30 um 08 23 35

HeikoKlare:
Provided via #789.

PR #789, opened 2023-08-29
Correct GC offset calculation #788

### LINKED ISSUE #788: GC draws at faulty positions when applying a rotating transformation
### Describe the bug
The `GC` class provides functionality to apply a transformation (translation, scaling, rotation) to the graphics canvas.
This functionality is broken when applying rotations to the GC, leading to
- completely lost draw operations (e.g. with 45° rotation)
- off-by-one errors in positions
- off-by-one errors in lines widths

It particularly occurs when drawing lines of od line width (including the default width of 0), which also includes lines that are supposed to have an off line width but are considered as even because of rounding errors during calculation.

There are also minor issues with applying a line width even if no transformation is applied in case non-horizontal/-vertical lines are drawn: they are drawn longer than they should be.

### To Reproduce
I have provided a PR containing `Snippet381` that allows to simulate the behavior of different transformations applied to a GC: #787
It allows to simulate different transformations of a GC to which the following pattern of lines is drawn. It also marks the "anchor point", i.e., the pixel representing (0,0), as a reference point in red.
Default

A large amount of issues with the functionality can occur. Some examples are as follows and can be reproduced with the given settings in the snippet:
1. Applying a transformation with a rotation of 45° to lines with line width 0 results in nothing being drawn (Line Width: 0, Rotation: 45).

2. Applying a transformation with a rotation of 90° to lines with line width 1 results in off-by-one errors (Line Width: 1, Rotation: 90).

3. Applying a transformation with a rotation of 360° to lines with line width 1 results in off-by-one errors (Line Width: 1, Rotation: 360).

4. Applying a transformation with a rotation of 90° to lines with line width 4 and 5 results in diagonal lines of faulty width (Line Width: 4/5, Rotation: 90)
Width 4 + 90
Width 5 + 90
5. Applying no transformation to lines with line width 5 results in diagonal lines being printed too long (Line Width: 5)
diagonal_too_long

### Expected behavior
When applying rotations to a GC, draw operations shall lead to proper results. This particularly means that:
1. 45° rotations of the GC are handled properly
2. Other rotations do not lead to off-by-one errors in positions and line widths. E.g., drawing something with 90° rotation should lead to the same result as without rotation but only rotated by 90°.

### Bug cause
The cause of the issue is rooted in the way SWT handles coordinates. My interpretation of what I have seen so far is as follows. SWT uses a continuous coordinate system and applies it to the discrete pixel raster. I.e., the coordinate (0,0) represents the upper left edge of the "first" pixel. The "real" coordinate of that pixel is (0.5,0.5). SWT applies different interpretations of coordinates depending on the current line width. With an even line width, such as 2, coordinates are taken as they are. For example, a horizontal line starting at (1,1) will stretch the pixel row above and below y-coordinate 1. With an odd line width, such as 1, coordinates are virtually moved by (0.5, 0.5), so that a horizontal line starting at (1,1) will "virtually" start at (1.5, 1.5), such that it stretched the row below y-coordinate 1.
Line Widths

This offset of 0.5 applies to the space to which the user applies draw operations. However, it is realized by applying an offset to the GC's drawing pane before applying some draw operations, such as drawing a line (see, e.g., [here](https://github.com/eclipse-platform/eclipse.platform.swt/blob/4daf005a457fb50b929719cb5b3616e55cc2e5c4/bundles/org.eclipse.swt/Eclipse%20SWT/win32/org/eclipse/swt/graphics/GC.java#L1655)). When a transformation is applied to the GC, the offset has to be transformed from the "user space" to the "GC drawing space". This is done by an according offset calculation in `checkGC`:
https://github.com/eclipse-platform/eclipse.platform.swt/blob/4daf005a457fb50b929719cb5b3616e55cc2e5c4/bundles/org.eclipse.swt/Eclipse%20SWT/win32/org/eclipse/swt/graphics/GC.java#L326-L345
The current offset calculation only works properly when applied to translate and scale operations, but not when applied to rotate operation. For example, a 45° rotation results in a calculated scaling value of 0, which results in an offset of `infinity`, so that nothing is drawn anymore.

### Additional Data

**Environment:**
- - [x] All OS
- - [ ] Windows
- - [ ] Linux
- - [ ] macOS 

I have seens the behavior on Windows, but since the GC implementation for the offset calculation is the same on all system, I expect the issue to exist on all systems. However, it would be interesting if someone using MacOS or Linux could use the provided snippet to see if the same issues exist.

Windows 11

3. JRE/JDK version
JDK 17
 
**Version since**
At least since 4.15, but the code exists in its current state for more than 15 years.

HeikoKlare:
The offset calculation in the `GC` when having an odd line width is erroneous in case rotation transformations have been applied to the `GC`. With a rotation of 45°, the calculation even suffers from singularities that result in effectively nothing being drawn. With other transformation properties, off-by-one error occur.

This change inverts the offset calculation: Instead of performing a forward transformation of a dummy point whose result is applied to the expected offset, the inverse transformation is applied to the expected offset. This avoids singularities and imprecision in the calculation and ensures that when applying rotations to the `GC`, the drawing figures are as expected.

The snippet in #787 allows to play around with the offset behavior in different transformation scenarios and to compare it with the existing behavior if once started with the existing and once with the code proposed in this PR.

The PR includes the fix for all three SWT implementations and provides a regression test for the existing singularity that fails without the fix. It also adds a snippet for demonstrating the original bug and the effects of this change.

Fixes #788.


fedejeanne:
I'll try to take a look at it today/tomorrow :)

fedejeanne:
## In Mac
* _macOS Ventura (13.2.1)_
* M1 Chip

# Summary
* Tested in 2 Displays: integrated display and external (Dell) display
* I did "runs" with the following _Rotations_: 45°, 90°, 135°, 180°, 225°, 270°, 315°, 360° and 405°
* Before this PR, there are singularities (the arrow doesn't show) at rotations starting at 45° and increases of 90° _i.e_: 45°, 135°, 225°, 315° and 405°.
* Those singularities only happen with `Line Width = 0` --> _i.e._ starting at `Line Width = 0.1` **there are no singularities even before applying this PR**
* There are no singularities after applying this PR, regardless of the `Line Width` 🥇 
* _Use anti aliasing_ doesn't make any difference
* Applying _Horizontal / Vertical Scaling_ doesn't make any difference (I used a factor of **2.0** for each, both separately and simultaneously _i.e._ I did 3 runs)

# Some screenshots

## In "the big display": DELL U2518D (25'', 3840x2160)
### Before this PR
Bildschirm­foto 2023-11-30 um 08 10 29

### After this PR

Bildschirm­foto 2023-11-30 um 08 11 36

## In "the small integrated Display": Integrated Retina Display (13.3'', 2560x1600)

### Before this PR
Bildschirm­foto 2023-11-30 um 08 20 47

### After this PR
Bildschirm­foto 2023-11-30 um 08 23 35

fedejeanne:
# In Linux
* Ubuntu 20.04.2 LTS (Yes, it's time for an upgrade)
* GNOME 3.36.8

# Summary
Same as [for Mac](https://github.com/eclipse-platform/eclipse.platform.swt/pull/789#issuecomment-1833273509) except for the last item:
* The check-boxes Draw horizontal / vertical line do show up in the snippet :heavy_check_mark: 

# Some screenshots
## In "the big display": DELL U2518D (25'', 3840x2160)
* On the left: **Before** this PR
* On the right: **After** this PR

With `Line Width == 0`
Screenshot from 2023-11-30 11-13-44

With `Line Width > 0`
Screenshot from 2023-11-30 11-14-23

## In "the small display": HP (17'', 1920x1080)
* On the left: **Before** this PR
* On the right: **After** this PR

With `Line Width == 0`
Screenshot from 2023-11-30 11-18-13

With `Line Width > 0`
Screenshot from 2023-11-30 11-18-54

fedejeanne:
> The check-boxes Draw horizontal / vertical line didn't show up in the snippet (they do in Linux though).

Scratch that, I just restarted the snippet and they are being shown in Mac too.

Bildschirm­foto 2023-11-30 um 11 23 13

All A-OK ✅

fedejeanne (approved):
Linux and Mac look good

HeikoKlare:
I integrated the commit with the snippet to visualize the original issue and the fix originally provided in #787 into this PR.

Since no objections have been made to this PR, I will merge it now.

PR #801, opened 2023-09-05
Tree.java: correctly escape & in default Tooltip

Wittmaxi:
Fix for
https://github.com/eclipse-platform/eclipse.platform.ui/issues/238
https://github.com/eclipse-platform/eclipse.platform.ui/issues/1075

grafik

When TreeColumn's are too small, a ToolTip is displayed on hover to reveal the whole Label of an Item (The WIN32-Docs call them [In-Place Tooltips](https://learn.microsoft.com/en-us/windows/win32/controls/using-tooltip-contro)).
These ToolTips pass through the Composite-ToolTipText-generator which escapes all & in a Text - allowing to reuse Text from Buttons as Tooltips. This is a well documented Feature of setToolTipText. Since TreeItem's Label does not have Mnemonics, we do not assume that the & should require escaping by the User.

This Commit Escapes all & in default Tooltips that are handled by the default mechanism.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
@@ -5608,7 +5608,7 @@ String toolTipText (NMTTDISPINFO hdr) {
 			}
 			//TEMPORARY CODE
 			if (isCustomToolTip ()) text = " ";
-			if (text != null) return text;
+			if (text != null) return text.replace("&", "&&");
 		}
 	}
 	return super.toolTipText (hdr);


Wittmaxi:
I am not sure wether using RegExes is an okay thing to do here.
I do not expect this method to be called a lot, hence why I am okay sacrificing a bit of Performance for better code clarity.

Phillipus:
Is that the right place to be changing the contents of the tooltip string? Something tells me that it's not, and perhaps ampersand escaping should be done further up the stack, or by the client.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
```diff
@@ -5639,7 +5639,8 @@ String toolTipText (NMTTDISPINFO hdr) {
 		if (findCell (pt.x, pt.y, item, index, cellRect, itemRect)) {
 			String text = null;
 			if (index [0] == 0) {
-				text = item [0].text;
+				text = new String(item [0].text);
+				text = text.replaceAll("&", "&&");
```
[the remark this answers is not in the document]
Wittmaxi: Thank you for the feedback, I agree with you and changed my code accordingly

Phillipus:
> @Phillipus Escaping the `&` definitely can't be done by the client since this ToolTip is automatically generated from the Text. I also don't think explicitly setting a ToolTip is a good idea since this ToolTip is only shown when the TreeColumn is too small - property that is checked for before drawing the ToolTip

Yeah, I can see the problem. It's just that when I look at the `String toolTipText (NMTTDISPINFO hdr)` code it seems tome to be the wrong place to actually change the string. But maybe there is no other place to change it.

Is this also happening on Mac and Linux?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
```diff
@@ -5639,7 +5639,8 @@ String toolTipText (NMTTDISPINFO hdr) {
 		if (findCell (pt.x, pt.y, item, index, cellRect, itemRect)) {
 			String text = null;
 			if (index [0] == 0) {
-				text = item [0].text;
+				text = new String(item [0].text);
```
Phillipus, line 5642: Is creating a new String necessary?

Wittmaxi: I have addressed this issue.

HeikoKlare:
I came across this again and again and always thought about alternative solutions but found that it's probably the best and most central place where we can escape the string. @Phillipus do you still have any concerns or should we merge this?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
```diff
@@ -5601,7 +5601,7 @@ String toolTipText (NMTTDISPINFO hdr) {
 		if (findCell (pt.x, pt.y, item, index, cellRect, itemRect)) {
 			String text = null;
 			if (index [0] == 0) {
-				text = item [0].text;
+				text = item [0].text.replace("&", "&&");
```
akurtakov: Could it be already escaped "&&" and thus become "&&&&"?

Phillipus: This only replaces the text in the first column of a `TreeColumn` it would also need to do it in line 5607 or once on `text` in line 5611.

Phillipus:
>  @Phillipus do you still have any concerns or should we merge this?

It only works for the tooltip in the first column of a Tree that uses columns. So that needs to be addressed.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
HeikoKlare: > This only replaces the text in the first column of a `TreeColumn` it would also need to do it in line 5607 or once on `text` in line 5611.

Good point. I have adapted the change to apply the escaping to the final text instead of the individual text retrievals. Now it is applied to every column.

> Could it be already escaped "&&" and thus become "&&&&"?

The tooltip is created from the original string presented in the cell. I don't see how that could already be escaped, as in that case the string with doubles `&` would then already be shown in the table cell, wouldn't it?

HeikoKlare:
fwiw, I used the snippet below for testing.
When testing I also found another issue: when reordering the table columns, the tooltips of the non-reordered cells are shown. I <s>will</s> have created a follow-up PR for that:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/3366

```java
	Tree t = new Tree (shell, 0);
	t.setHeaderVisible(true);
	t.setLinesVisible(true);
	TreeColumn col1 = new TreeColumn(t, 0);
	col1.setWidth(30);
	col1.setMoveable(true);
	TreeColumn col2 = new TreeColumn(t, 0);
	col2.setWidth(30);
	col2.setMoveable(true);
	new TreeItem (t, 0).setText(new String [] {"A & B", "C & D"});
	new TreeItem (t, 1).setText(new String [] {"E & F", "G & H"});

	shell.open ();
	}
	display.dispose ();
}
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
akurtakov: OK, I was just pointing to the most common escaping bug.

HeikoKlare: No guarantee that I did not miss something, it's just my current understanding of how it works 🙂 And of course I tried once with with cell contents that are "escaped" already but then the double `&` is simply shown in the cell.

PR #826, opened 2023-09-29
Code Cleanup: Java 16 instanceof pattern matching

elsazac:
Code Cleanup: Java 16 instanceof pattern matching

Performs code cleanup in pattern matching using
instanceof operator according to Java 16 guidelines, making the code more concise.


HeikoKlare (approved):
Thanks for the refactoring! The changes look good. I have added the required version bump for `org.eclipse.swt.examples`.

Issue #828, opened 2023-10-05
Shell.getBounds() doesn't match actual size on X11


ptziegler:
**Describe the bug**
The size of a shell reported by `getBounds()` doesn't match its actual size that is shown on the screen. 

**To Reproduce**

Consider the following code snippet:
```
Display display = Display.getDefault();
Shell shell = new Shell(display);
Composite composite = new Composite(shell, SWT.NONE);

shell.setLayout(new FillLayout());
shell.setSize(400, 500);
shell.open();
System.out.println("Shell: " + shell.getBounds());
System.out.println("Composite: " + composite.getBounds());

while (!shell.isDisposed()) {
	if (!display.readAndDispatch()) {
		display.sleep();
	}
}
```

On Linux, both the shell and the client area will be of size (400, 500). On Windows, the client area is "only" (384, 461). Reason being that the decorations of the shell are taken into consideration when determining the "usable" area. 
Note that the size of the client area of the shell is accurate, in both cases. Meaning that on Windows, `setSize()` describes the size of the shell and on Linux, the size of the client area (with the shell being larger than specified).

**Expected behavior**

The shell should report its real size. Consequently, its client area should also be smaller than its size.

**Screenshots**

A variant of a shell with size (400, 400), to highlight the size of the client area (a perfect square).
![image](https://github.com/eclipse-platform/eclipse.platform.swt/assets/70652978/d2b071de-5f53-432b-8934-e28d5ba5f86b)

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [ ] Windows
- - [x] Linux
- - [ ] macOS 

2. Additional OS info (e.g. OS version, Linux Desktop, etc)

Kubuntu 22.04
GTK 3.24.33

3. JRE/JDK version

Java 17.0.8.1+1-Ubuntu-0ubuntu122.04

**Version since**
Eclipse 4.30 (I20231004-1800)

**Workaround (or) Additional context**

We rely on the result of `getBounds()` to take a screenshot of a given widget. Because this returns the wrong size, the lower edge of the widget is cut off by ~29 pixels (or generally, the insets of the decorations). This works fine on Windows, but on Linux, you get visual artifacts as shown below:
 
![image](https://github.com/eclipse-platform/eclipse.platform.swt/assets/70652978/9440a1e5-4ae2-4b5e-aec8-daa3fb9582bf)

ptziegler:
[...]

PR #829, opened 2023-10-06
Create small test case to illustrate #828

ptziegler:
This test should pass on Windows, but fail on Linux.
Because there, the client area matches the size of the shell, even though it should be smaller due to the decorations.

I've tested a variant where I replaced the calculation in `getBoundsInPixel()` with [gdk_window_get_frame_extents](https://docs.gtk.org/gdk3/method.Window.get_frame_extents.html). In principle, this works, though unreliably.


akurtakov:
As the issue has been closed I'm closing this PR too. Please reopen with details if I misunderstood.

PR #846, opened 2023-10-30
Fixing the spacing.width value to enable re-collapsing of table columns

elsazac:
Fixes : https://github.com/eclipse-platform/eclipse.platform.swt/issues/662

The default intercell spacing value that is obtained from Cocoa for width is 17. This patch provides a change for `CELL_GAP` value. For width, `CELL_GAP` is replaced with `HCELL_GAP` with value 17 (default value obtained from Cocoa). For Height (existing value is retained which is 1) `CELL_GAP` is replaced with `VCELL_GAP` with value 1.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Table.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Table.java
@@ -651,6 +651,9 @@ void createItem (TableColumn column, int index) {
 	display.addWidget (headerCell, column);
 	column.nsColumn = nsColumn;
 	nsColumn.setWidth(0);
+	nsColumn.setMinWidth(2);
+	/* Maintaining the minimum width at the lowest value of 2, which guarantees the visibility of the drag cursor
+	and the possibility of table columns re-collapsing */
 	System.arraycopy (columns, index, columns, index + 1, columnCount++ - index);
 	columns [index] = column;
 	for (int i = 0; i < itemCount; i++) {


elsazac:
@Phillipus 

Could you test this patch?

Phillipus:
I tested this now. It seems to fix the original column resizing issue but has introduced a different problem. If you run [Snippet 77](https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet77.java) before and after the patch there is now a gap between the start of the table cell and the start of the cell text:

Before the patch:

before

And after the patch:

after

elsazac:
Yes, this is not a new issue due to the patch. The reason for the gap we observe is that the horizontal spacing has increased from 1 to 17, with 17 being the default `spacing.width` value obtained  from cocoa. Prior to this, the value was 1, which was a very small value, explaining why the gap was not apparent. Now the horizontal spacing is increased to 17. So the gap is visible.

> I tested this now. It seems to fix the original column resizing issue but has introduced a different problem. If you run [Snippet 77](https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet77.java) before and after the patch there is now a gap between the start of the table cell and the start of the cell text:

Phillipus:
> Yes, this is not a new issue due to the patch. The reason for the gap we observe is that the horizontal spacing has increased from 1 to 17, with 17 being the default `spacing.width` value obtained from cocoa. Prior to this, the value was 1, which was a very small value, explaining why the gap was not apparent. Now the horizontal spacing is increased to 17. So the gap is visible.
> 

Compare with Windows. We can see that column text and cell text align:

windows

With the proposed patch the column text and cell texts no longer align.

@tmssngr Will this affect your use of tables on Mac? WDYT?

Phillipus:
The gap is also present on the right side of the table cell, and both gaps are very obvious if there is an editable control in the cell. I definitely think this gap is wrong and is a regression.

elsazac:
> The gap is also present on the right side of the table cell, and both gaps are very obvious if there is an editable control in the cell. Also, because of this extra gap, some table cells' contents are clipped (you can see this in some tables when they are used in Eclipse's Preferences.)

Could you point me to which specific table in the preferences are you referring to where it got clipped?
I tried to check myself but couldn't find one such occurrence. In one of the occurrences  the text is not missing as seen in the below table.
 I am referring to Preference>Java>Editor>Templates
 
image

Phillipus:
> Could you point me to the specific table in the preferences

I was just trying to point out some real-life examples.

The point is that a gap of 17 pixels each side of the cell control is a high price to pay for dragging columns, particularly if a table does not have columns or is not draggable. Also, this gap is not present on Windows or Linux. It might be better to see if there is another way to achieve this by limiting the extent of the drag operation.

elsazac:
@Phillipus 

Could you test this one?
I have added a minimum width for column so that beyond the width value drag operation cannot happen.

Phillipus:
HI, I tested this but I see a couple of issues:

1. The minimum width is too big in some cases. In our app we create columns with fixed pixel widths to show an icon:

This is before the patch:

before

And after the patch:

after

2. If you have a cell editing control like a combo it now overflows the cell width:

control

But the patch seems to be going in the right direction. Maybe some more adjustments are needed?

elsazac:
> HI, I tested this but I see a couple of issues:
> 
> 1. The minimum width is too big in some cases. In our app we create columns with fixed pixel widths to show an icon:

> 2. If you have a cell editing control like a combo it now overflows the cell width:
> 

If possible could you provide a snippet to test this?

Phillipus:
I think this is the wrong approach and adding column spacing is causing side-effects.

See my comment [here](https://github.com/eclipse-platform/eclipse.platform.swt/issues/662#issuecomment-1685989095).

Compare the behaviour on Windows and Linux, they do not have additional spacing and this means you can resize the column to zero spacing which is expected behaviour.

The real issue on Mac is that the resizing cursor and hot spot is not active if the column is at zero width. See the original description in https://github.com/eclipse-platform/eclipse.platform.swt/issues/662#issue-1701104695 Perhaps this approach could be investigated?

elsazac:
> The real issue on Mac is that the resizing cursor and hot spot is not active if the column is at zero width. See the original description in [#662 (comment)](https://github.com/eclipse-platform/eclipse.platform.swt/issues/662#issue-1701104695) Perhaps this approach could be investigated?

But I suppose the resizing cursor is not directly controlled by SWT code but is instead coming from the native library.

Also if you could look at Mac applications, there is minimum width set for tables beyond which dragging cannot happen.

Phillipus:
> Also if you could look at Mac applications, there is minimum width set for tables beyond which dragging is not possible.

But SWT tries to maintain a level of cross-platform behaviour where  possible. For example, we have Tables that use a column that is fixed at 18 pixels wide to show an icon. I think that setting a minimum width breaks existing behaviour on Mac and may break other apps column widths.

elsazac:
> 2. If you have a cell editing control like a combo it now overflows the cell width:
> 
> control

@Phillipus 
I couldn't reproduce this particular issue you mentioned here. So could you provide me the snippet that you used to test table containing combo.

tmssngr:
@Phillipus What exactly (which patch) should I try?

Phillipus:
> @Phillipus What exactly (which patch) should I try?

@tmssngr This one - https://github.com/eclipse-platform/eclipse.platform.swt/pull/846/files

But I disagree with this approach. This patch is adding a minimum width of 51 pixels to a column which, in our product at least, messes up table columns. In some cases we set a fixed width of less than 51 pixels. I would prefer to leave things as they are (and keep consistent bahviour in line with Windows and Linux).

elsazac:
I have reduced the width value to the very minimum so that drag cursor is still visible.

elsazac:
@tmssngr If you could give a try to the new patch, it would be great.

tmssngr:
> > @Phillipus What exactly (which patch) should I try?
> 
> @tmssngr This one - https://github.com/eclipse-platform/eclipse.platform.swt/pull/846/files
> 
> But I disagree with this approach. This patch is adding a minimum width of 51 pixels to a column which, in our product at least, messes up table columns. In some cases we set a fixed width of less than 51 pixels. I would prefer to leave things as they are (and keep consistent bahviour in line with Windows and Linux).

Sorry for the delay. I didn't notice any change for tables in SmartGit by introducing this line
```java
nsColumn.setMinWidth(0);
```

Phillipus:
> Sorry for the delay. I didn't notice any change for tables in SmartGit by introducing this line
> 
> ```java
> nsColumn.setMinWidth(0);
> ```

The PR changes it so it should be:

```java
nsColumn.setMinWidth(2);
```

elsazac:
This is the behaviour I see with and without patch.

If you try on a sample snippet like [Snippet 3](https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet3.java)

Without patch :

https://github.com/eclipse-platform/eclipse.platform.swt/assets/122080218/af924f62-5c9b-49b2-9521-3a10e86e317f



With Patch :


https://github.com/eclipse-platform/eclipse.platform.swt/assets/122080218/1b6fc71d-7dc1-4d6c-8b14-527eb5d2cb6e

elsazac:
This is a Mac only issue, so the PR is for Mac alone. This fixes the issue https://github.com/eclipse-platform/eclipse.platform.swt/issues/662
@SyntevoAlex @sratz 
Can any one of you test and confirm please.

(I have added the behaviour before and after the patch in my previous comment)

SyntevoAlex:
Unfortunately my queue is more than full. I don't think I can have a look in the next month.

elsazac:
> But SWT tries to maintain a level of cross-platform behaviour where possible. For example, we have Tables that use a column that is fixed at 18 pixels wide to show an icon. I think that setting a minimum width breaks existing behaviour on Mac and may break other apps column widths.

I agree 100% that SWT should strive to maintain cross platform behaviour wherever possible. For that, the usual practice in my opinion is to first leverage / exploit the existing interfaces from the operating system (in this case NSTableView and NSTablecolumn, ref: https://developer.apple.com/documentation/appkit/nstableview and https://developer.apple.com/documentation/appkit/nstablecolumn) and then customize missing characteristics / behaviors using SWT extension code. I don't see any documented OS API through which we can modify the drag cursor's behaviour. From this, it is clear that the drag cursor is created and controlled fully by the OS, and cannot be designed from the user space.

I have also reviewed behavior of a number of native Mac applications (for example, the finder window) and examined the approach they have taken and their external behaviour. It is evident that the tables have a minimum width defined below which dragging is impossible!

In the absence of adequate platform support, this is exactly what I am trying to achieve here- fix a minimum width for the drag cursor. This, in my opinion, is a reasonable thing to do and a pragmatic approach, if not comprehensive. Having said that, I'm open to suggestions or recommendations if someone has a better idea!

tmssngr:
Below are screenshots from the Finder:
Screenshot 2023-12-14 at 11 13 13
and Activity Monitor:
Screenshot 2023-12-14 at 11 14 14
with the columns resized to their minimum sizes. As can be seen, they depend on the column, so I assume they were set by the developer. I'm somewhat surprised that `TableColumn` does not have a `setMinimumWidth` method. Maybe it makes sense to add that?

elsazac:
@tmssngr - I am not sure I understood you properly. Are you agreeing to my assertion or refuting? Unfortunately it is not clear from your description. Can you clarify?

> with the columns resized to their minimum sizes. As can be seen, they depend on the column, so I assume they were set by the developer.

Yes. the minimum width can be set by the developer using the `NSTableColumn.setMinWidth` API. It is the developer’s choice to set the value individually or on a per column basis, or with a constant value or derived from the length of the text that the column holds.

>  I’m somewhat surprised that TableColumn does not have a setMinimumWidth method. Maybe it makes sense to add that?

it is already there under cocoa library: https://github.com/eclipse-platform/eclipse.platform.swt/blob/27feab24022af12f31912000999f9717c3b7e803/bundles/org.eclipse.swt/Eclipse%20SWT%20PI/cocoa/org/eclipse/swt/internal/cocoa/NSTableColumn.java#L65-L67

Do you mean exposing this as a public API under TableColumn class?

tmssngr:
Yes, I meant it might make sense to create a new `TableColumn.setMinimumWidth()` method exposing the `NSTableColumn.setMinWith` functionality (using `int` instead of `double`). Of course, this only makes sense to be considered if at least Linux or Windows also support setting the minimum column width.
If this is no option, setting the minimum width on macOS only would be a workaround to be able to resize the columns.

merks:
I wonder if this problem also applies for trees with columns, not just tables?

I recalled that in some application that I programmatically set the column width to zero to hide the column.  Doing a search in the SDK workspace, I think there are a number of places calling setWidth(0) for such a purpose:

image

Indeed this one appears to be about hiding a table column:

https://github.com/eclipse-platform/eclipse.platform/blob/9aa3ddbffcdf831d4e014f290bd14d056dc67bec/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/memory/provisional/AbstractAsyncTableRendering.java#L226-L234

So I too would be concerned that simply setting a minimum for all table columns, but only on mac, could be problematic in general. 

I'm not sure what constructively to suggest.   It's clearly a nasty problem that a user can effectively collapse a column completely but then has no mechanism for opening the column back up.  But it's also kind of inconceivable that this can just be a general problem on mac where the only solution is to set a minimum width on a column.  Even if we provide a setMinimum method on column that works across the win, mac, and linux, the application programmer would have to call it for all columns always to ensure that mac works properly.  That hardly seems a good solution either.  I think there really needs to be a better way, but I can't think of one...

elsazac:
based on the conversation so far, I see these options
 - set default minimum width for columns of tables and trees.
 - set default minimum width for columns of tables and trees with an API provided to change that.
 - don't touch the minimum width but provide API for changing it.
 - land as is : this PR is scoped to address the linked issue- "Collapsed Table Columns can not be Re-Expanded on Mac"
 
 while my preference is to address the stated problem and build incrementally in subsequent PRs , I am fine with any of these options. Please let me know if we have consensus.

elsazac:
@tmssngr 
Thanks for agreeing to the proposal

@merks 
>So I too would be concerned that simply setting a minimum for all table columns, but only on mac, could be problematic in general.

I recognise your concern, but our options are limited. all the four options I suggested above boil down to setting the minimum width using the field assignment or through an (new)API invocation. And the whole thing is applicable only for Mac. other platforms do not suffer from this problem.

In the best interest of resolving the reported issue, I suggest we go with current change and introduce APIs for additional customisation. if there are better proposals I am open to suggestions.

tmssngr:
Would it be technically possible on Windows and Linux, too, to set the minimum column width?

merks:
Providing new API is fine, if it works across the OSes.  But, as I showed you with my search, there are places in the Platform itself that rely on setting the width to zero to make the column not visible.  If your change breaks that in order to fix some issue it's just changing the problem from one problem to another problem.  So you need to test whether those other places still work.  I.e., can one programmatically set the width to zero if you've set the minimum to  something greater than zero hidden inside of some implementation detail specific to Mac?  

Of course if there is new API where one must explicitly set the minimum, but it seems to me that then the problem will exist for everyone except those setting the minimum and where do we plan, in the Platform SDK code base to set such a minimum?

merks:
> TBH, I would rather leave things as they are, until a better solution is found.

Yes, I think to leaving a long-standing bug as a bug is  more appropriate than causing regressions for somethings that works now.  

I could imagine that one solution might be to detect when the **user** has resized the column to zero and, only in that case pop, it back to some minimum, and only on mac.  But I have no idea how to implement that (and I still find it inconceivable that the underlying OS widgets would have such a bug intrinsically).

tmssngr:
> Providing new API is fine, if it works across the OSes. But, as I showed you with my search, there are places in the Platform itself that rely on setting the width to zero to make the column not visible.

Let's name it correctly: parts of the Eclipse platform rely on this bug.
IMHO, the correct way would be to use `TableColumn.dispose()` to hide it and creating a new TableColumn to show it. But this would require changing legacy code which no one wants to do.

> I could imagine that one solution might be to detect when the **user** has resized the column to zero and, only in that case pop, it back to some minimum, and only on mac. But I have no idea how to implement that (and I still find it inconceivable that the underlying OS widgets would have such a bug intrinsically).

I wouldn't limit this to macOS only, but it sounds like a reasonable workaround.

tmssngr:
I've tested it on Windows and resizing causes ugly flickering because the column jumps between user-size and minimum size while dragging. Unfortunately, the resize event does not indicate whether the user still drags or the resize is finished.
screencast

elsazac:
closing this PR in light of the above comments. If a better solution is found, I'll attempt looking into it more.

PR #853, opened 2023-11-04
Leverage Tycho-pomless

HannesWell:


HannesWell:
Verified that the number of modules and executed tests (o.e.swt.tests plus swt.tests platform-specific) is the same with and without this change, so all looks good.
I noticed there are no platform-specific tests for cocoa executed (before and after), but since no tests are defined this is isn't surprising.

PR #863, opened 2023-11-07
Wayland qt kde clipboard fix

the-snowwhite:
I now know the buzzwords are:

> "Eclipse support for Wayland clipboard content types (RFC-1341  types)"

https://github.com/eclipse-platform/eclipse.platform.swt/issues/851#issuecomment-1794539321

However the main motivation here is:

` "Being able to use Eclipse comfortably with a LTS based linux distro in a wayland session in  2023"`

The first commit already has an open pr #861 the second has  been tested to work via this method #858
and then cherry picked from R_29_maintenance branch onto master without any issues.

PR #862, opened 2023-11-07
Remove blank Javadoc

jukzi:
This commit cleans up Javadoc that does not add information. It resolves ecj warnings:
 `Javadoc: Description expected after ...`
It helps to prevent future empty javadoc by disabling missingJavaDoc warnings. This resolves
 `Javadoc: Missing ...`

The modification is a result of regular expression search&replace:

in files `*.java`
 `^[\s]*\*[\s]*(@return|@param[\s]*[^\s]+|@throws[\s]*[^\s]+)\R([\s]*\*[\s]*@|[\s]*\*/\R)`
 ->`$2`
 `^([\s]*\*[\s]*\R)([\s]*\*/\R)`
 ->`$2`
 `^[\S\t ]*/\*\*\R[\s]*\*/\R`
 ->``

in files `org.eclipse.jdt.core.prefs`
 `org\.eclipse\.jdt\.core\.compiler\.problem\.missingJavadoc(Comments|Tags)\=[^\s]*`
 ->`org\.eclipse\.jdt\.core\.compiler\.problem\.missingJavadoc$1\=ignore`

PR #867, opened 2023-11-09
test: https://github.com/eclipse-tycho/tycho/issues/3019

the-snowwhite:
by "placing output.. = bin/" in build.properties

This commit seems to do no difference to the maven test build as expected.
But makes it possible to see if the step:
`ant -f eclipse.platform.swt/bundles/org.eclipse.swt/buildSWT.xml tag_projects`
builds correctly afterwards in:
https://github.com/eclipse-platform/eclipse.platform.swt/pull/865
https://github.com/eclipse-platform/eclipse.platform.swt/pull/863

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -92,7 +92,7 @@ pipeline {
 						git config --global user.email 'eclipse-releng-bot@eclipse.org'
 						git config --global user.name 'Eclipse Releng Bot'
 						
-						ant -f eclipse.platform.swt/bundles/org.eclipse.swt/buildSWT.xml check_preprocessing
+						ant -f eclipse.platform.swt/bundles/org.eclipse.swt/buildSWT.xml check_sources_and_update_sha1_file -DTAG=HEAD
 						ant -f eclipse.platform.swt/bundles/org.eclipse.swt/buildSWT.xml new_build_with_create_file -DTAG=HEAD
 					'''
 				}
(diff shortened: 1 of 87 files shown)


the-snowwhite:
This doesn't seem to be the solution as:
`ant -f eclipse.platform.swt/bundles/org.eclipse.swt/buildSWT.xml tag_projects`
still fails with:

    06:29:43  BUILD FAILED
    06:29:43  /home/jenkins/agent/workspace/eclipse.platform.swt_PR-867/eclipse.platform.swt/bundles/org.eclipse.swt/buildSWT.xml:309: Unable to determine new tag

iloveeclipse:
Could you please revert your recent change and force push here, to see if that fixes the build?

the-snowwhite:
Got the same failure
So the next logical step in this test would be to revert:
https://github.com/eclipse-platform/eclipse.platform.swt/pull/862
and
https://github.com/eclipse-platform/eclipse.platform.swt/pull/866
ok ?

akurtakov:
My feeling is that the build scripts can't go beyond the "z" for new tags. Last one is https://github.com/eclipse-platform/eclipse.platform.swt/releases/tag/v4963r3z so this build failure has nothing to do with reverting and appying patches.

laeubi:
The mentioned bug is only relevant for **Tycho Api Tools Check** how is it supposed to fix any **Custom Ant Build Scripts**? It would really help to try understand the failure instead of blaming everything on random Tycho issues.

the-snowwhite:
As it does not make sense to revert a commit inside same commit I do not test reverting this commit here.
Anyway I this this PR has made its point so I have created an issue instead:
I close this PR now.

Issue #868, opened 2023-11-09
ant -f eclipse.platform.swt/bundles/org.eclipse.swt/buildSWT.xml tag_projects fails with: 


the-snowwhite:
BUILD FAILED
    /home/jenkins/agent/workspace/eclipse.platform.swt_PR-867/eclipse.platform.swt/bundles/org.eclipse.swt/buildSWT.xml:309: Unable to determine new tag

**Describe the bug**
Releng builder jobs currently fail so that none of  recent merged PR's get built.

**To Reproduce**
Trigger a build.

**Expected behavior**
Build sicceds

**Screenshots** / build log:
https://ci.eclipse.org/releng/job/eclipse.platform.swt/view/change-requests/job/PR-863/lastFailedBuild/console

**Environment:**
Build system

akurtakov:
For the record:
```
ant -f buildSWT.xml tag_projects
Buildfile: /home/akurtakov/git/eclipse.platform.swt/bundles/org.eclipse.swt/buildSWT.xml

get_version:
Java 15 has removed Nashorn, you must provide an engine for running JavaScript yourself. GraalVM JavaScript currently is the preferred option.

BUILD FAILED
/home/akurtakov/git/eclipse.platform.swt/bundles/org.eclipse.swt/buildSWT.xml:322: Unable to create javax script engine for javascript

Total time: 0 seconds
```
This makes it even harder to test on local machine unless one goes extra hoops to supply rhino to the build.

the-snowwhite:
https://github.com/eclipse-platform/eclipse.platform.swt/pull/867#issuecomment-1803238345

the-snowwhite:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1528

the-snowwhite:
@akurtakov 
Seems like your suspensions are right on the spot:
The culprit must be right here:
https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/bundles/org.eclipse.swt/buildSWT.xml#L299
ascii 97 is a and ascii 122 is z
So perhaps the solution would be to create a wrap around from z to a ?

the-snowwhite:
Or perhaps better from z to za

mickaelistria:
[...]

PR #869, opened 2023-11-09
Test if incrementing tag suffix with wrap around to az azz fixes build.

the-snowwhite:


akurtakov:
Honestly I totally dislike the idea of even more javascript in the build scripts. It's a move in the wrong direction IMHO.

the-snowwhite:
Please do not merge this commit as there still are issues to be discussed around it.

the-snowwhite:
replaced by https://github.com/eclipse-platform/eclipse.platform.swt/commit/9a71c28273aa67256979430a38ae0bd2c4140c08

the-snowwhite:
This is a long term solution however it means more java code in the ant build.

mickaelistria:
Like @akurtakov , I think a long term solution must involve simplification, not extra javascript code.
Solution can be to just stop committing or to just suffix with timestamp instead of maintaining a suffix generation algorithm which seems not so necesary.

the-snowwhite:
I agree that simplification is to prefer rather than more java script.
However Why not suffix with commit id's instead of timestamps ?
That said I know nothing about for what reasons the current tag system was implemented and what it's good for
other than the build speedup mentioned in the ant build script.

mickaelistria:
> However Why not suffix with commit id's instead of timestamps ?

timestamps allow version to be comparable and to more easily favor the latest and greatest.

the-snowwhite:
Superseeded by:
https://github.com/eclipse-platform/eclipse.platform.swt/commits/master
https://github.com/eclipse-platform/eclipse.platform.swt/commit/5e283d68b0d92e4a72d1db14a2a5367d9d3dd3bc

PR #870, opened 2023-11-09
Test single letter increment of tag to fix build.

the-snowwhite:


bundles/org.eclipse.swt/buildSWT.xml
```diff
@@ -291,19 +291,27 @@
...
+	            var found = false;
+	            var suffix = 'a';
+	            while (!found) {
+	                var t = swt_version + suffix;
+	                if (!tags.match(t)) {
+	                    project.setProperty("swt_new_tag", t);
+	                    found = true;
+	                } else {
+	                    if (suffix === 'z') {
+	                        suffix = 'a'; // Reset to 'a' after 'z'
```
mickaelistria, line 308: In existing code, if we've reached the case of 'z', it is most likely that the case of 'a' (`i=97`) and all other letters already reported that the suffix is already used.
You would basically want to write an algorithm that iterates through the sequences a..z,aa...az, ba...bz, ..., za...zz, aaa...aaz, aba...abz, ..., aza,...,azz, baa...baz and so on. It's a lot of smartness to put in a build scripts, particularly if we don't really need the tags.
Maybe just a timestamp suffix would be enough?

the-snowwhite: Does github support numbers in suffixes ?

mickaelistria: Git and GitHub don't have a concept of suffix, it's just some characters in the tag name; and yes, they do support numbers as characters in tag names.

the-snowwhite: We would'nt have this problem if the v4963r3 part of the tags were sure to be updated before 26 commits,
What mechanism updates this part or the tag ?

the-snowwhite:
This build seems stuck.

PR #872, opened 2023-11-09
Make revision for natives to be 4 too

akurtakov:
Work towards fixing


akurtakov:
Merge as previous build failed in commit+tag which verification build doesn't test.

PR #878, opened 2023-11-10
Chromium BrowserFactory: fix deprecated constructor warning since 9

jukzi:


akurtakov (changes requested):
SWT doesn't ship Chromium impl at all thus this whole section should be deleted.

jukzi:
@akurtakov can you please prepare such a PR, thanks.

akurtakov:
I'll try to remember that for next cycle.

Issue #851, opened 2023-11-03
Cannot copy from KDE Applications to Eclipse in Wayland session


the-snowwhite:
Closing as this has been fixed in latest gtk3 and changing swt gtk3 clipboard implementation has to potential to destabilize things and gtk4 port has it working with the new approach already.

_Originally posted by @akurtakov in https://github.com/eclipse-platform/eclipse.platform.swt/issues/421#issuecomment-1369523287_

akurtakov:
Interesting. I have tried coping from Eclipse to Kate  (both running under wayland) and it worked but kate to eclipse didn't.

the-snowwhite:
@akurtakov  Yes thanks that is exactly my point.
I'm running kde neon where it is not possible to copy either way between kate and eclipse in a wayland session.
So I recompiled the newest gtk version: 3.24.38 (from mantic) and installed.
That made it possible to copy from eclipse to kate, but not the other way.

the-snowwhite:
I have also just tested Kubuntu 22.10 and Fedora KDE Plasma Desktop 38, same issue.

akurtakov:
I would not have time to look into this one any time soon so we need someone to step up and work on a fix.

the-snowwhite:
OK ... I'm not experienced with eclipse code so I do not know what more I can do right now.
Thanks for acknowledging the Issue :-)

the-snowwhite:
Inspired by a [comment in this thread](https://bugs.kde.org/show_bug.cgi?id=431100#c11) I figured out a small workaround:
install the wl-clipboard package and the run this command before pasting to eclipse:

    wl-paste -n -t text/plain | wl-copy -t STRING

the-snowwhite:
In hope of the KDE-Team being able to provide a wayland lecgacy solution like Gnome-gtk I have opened a bug repport there:
https://bugs.kde.org/show_bug.cgi?id=476522

mickaelistria:
Thanks for the report, this comment helps "the issue essentially boils down to Eclipse not supporting Wayland clipboard content types (RFC-1341 types).". Do you think you can help with that?

the-snowwhite:
@mickaelistria Testing out my bugfix is now completed with success 
https://github.com/eclipse-platform/eclipse.platform.swt/discussions/858 
The initial test commit is currently residing in this branch: (dont know if It would be possible to backport the fix)

[...]

PR #879, opened 2023-11-10
Fix wrong implementation of TEXT_PLAIN_UTF8 transfer feature

iloveeclipse:
1) The utf-8 content should be same as in UTF8_STRING, it just should use different label/type id.
2) The change in ClipboardProxy.setData() was non functional and makes no sense, it only leaks memory/wastes time

See https://github.com/eclipse-platform/eclipse.platform.swt/pull/863 See https://github.com/eclipse-platform/eclipse.platform.swt/issues/851

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Drag and Drop/gtk/org/eclipse/swt/dnd/ClipboardProxy.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Drag and Drop/gtk/org/eclipse/swt/dnd/ClipboardProxy.java
@@ -169,9 +169,6 @@ boolean setData(Clipboard owner, Object[] data, Transfer[] dataTypes, int clipbo
 				System.arraycopy(entries, 0, tmp, 0, entries.length);
 				tmp[entries.length] = entry;
 				entries = tmp;
-				TransferData tdata = new TransferData();
-				tdata.type = typeIds[j];
-				transfer.javaToNative(data[i], tdata);
 			}
 		}
 
--- a/bundles/org.eclipse.swt/Eclipse SWT Drag and Drop/gtk/org/eclipse/swt/dnd/TextTransfer.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Drag and Drop/gtk/org/eclipse/swt/dnd/TextTransfer.java
@@ -13,8 +13,6 @@
  *******************************************************************************/
 package org.eclipse.swt.dnd;
 
-import java.nio.charset.*;
-
 import org.eclipse.swt.internal.*;
 import org.eclipse.swt.internal.gtk.*;
 
@@ -92,11 +90,10 @@ public void javaToNative (Object object, TransferData transferData) {
 		transferData.pValue = ctext[0];
 		transferData.result = 1;
 	}
-	if (transferData.type == UTF8_STRING_ID) {
+	if (transferData.type == UTF8_STRING_ID || transferData.type == TEXT_PLAIN_UTF8_ID) {
 		long pValue = OS.g_malloc(utf8.length);
 		if (pValue ==  0) return;
 		C.memmove(pValue, utf8, utf8.length);
-		transferData.type = UTF8_STRING_ID;
 		transferData.format = 8;
 		transferData.length = utf8.length - 1;
 		transferData.pValue = pValue;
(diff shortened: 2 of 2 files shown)


iloveeclipse:
@the-snowwhite : please test this implementation with your original use case in #851.

the-snowwhite:
@iloveeclipse 
Your statement here is a misunderstanding.(verified bu the eclipse debugger)
> 1. The utf-8 content should be same as in UTF8_STRING, it just should use different label/type id.
> 
>     2. The change in ClipboardProxy.setData() was non functional and makes no sense, it only leaks memory/wastes time
> 
> 
> See #863 See #851

I have fully tested the functionality of https://github.com/eclipse-platform/eclipse.platform.swt/pull/863 wia functional testing in the eclipse debugger running in a wayland session.
with only the changes in TextTransfer.java the debugger responded to no breakpoints in javaToNative (in TextTransfer.java)
and no data was presented to the wayland clipboard.

Then after introducing the 3 lines in ClipboardProxy.java the debugger now stopped in javaToNative and data was presented to the wayland clipboard.

Please suggest functional code changes in functional `code ` instead of abstract programming lingo the is hard for me to understand.
I'm an intuition based  functional procedural programmer with a gift for code and bugfixes but i do not understand OOP or OOP lingo.
OK ?

the-snowwhite:
> @the-snowwhite : please test this implementation with your original use case in #851.

OK I hope it works I like the much more simple style than what chatgpt4 beaten with a (stick and large scissors) could provide...

iloveeclipse:
> I have fully tested the functionality of #863 wia functional testing in the eclipse debugger running in a wayland session.
> with only the changes in TextTransfer.java the debugger responded to no breakpoints in javaToNative (in TextTransfer.java)
> and no data was presented to the wayland clipboard.

What I'm asking you now: please test **this patch** with Wayland! Not what you might have tested before in some debug session!

**If** it is not working under Wayland, it means, the code you've added in `ClipboardProxy.setData()` should only be called under Wayland session (if it the right place to do it at all, which I highly doubt here), because now under X11 I see that we call this code **three times** and it is not needed at all under X11!

The use case (as far as I could understand it, because it is not clearly described) was:

> running in both gtk v3.24.33 and gtk v3.24.38 it is possible to paste into eclipse from all applications that only present
RFC-1341 types (so far only text/plain;charset=utf-8) in the clipboard under a wayland session.
I have also just tested pasting the danish special characters åæø and that went fine also.

If I understood it right, the text copied to clipboard from Eclipse should be able paste to "kate" editor under Wayland.

Note: it is not the `ClipboardProxy.setData()` that should call javaToNative(), but `ClipboardProxy.getFunc()`,  at least under X11, so if under Wayland it is differently, you should look why `ClipboardProxy.getFunc()` doesn't call `javaToNative` or why `ClipboardProxy.getFunc()` is not callled.

I have no Wayland, so I can't check, but current code state is definitely wrong.

@mickaelistria, @akurtakov, @trancexpress : could you please check this PR & use case, if you have Wayland?

If this PR doesn't work under Wayland, I would propose a PR reverting #863 for RC1 as it definitely does wrong things on X11.

the-snowwhite:
@iloveeclipse 

> What I'm asking you now: please test this patch with Wayland! Not what you might have tested before in some debug session!
Yes i'm just having some issues as my main eclipse installation for SWT building has crapped out showing over 861 errors
and not being able to run the clipboard example .
Luckily I have a Kubuntu 22.04 VM that has a working Eclipse build setup, however now I'm having some issues installing wayland so it's slow progress. I hope your PR works as its a much more simple solution (stated more verbosely in other comment).

the-snowwhite:
@iloveeclipse 
Btw the audience for the paste fix is ubuntu 22.04 LTS users that have gtkv3.23.33(the one without the gtk paste fix).

mickaelistria:
@the-snowwhite So do I get it right you successfully tested that patch and are OK with it being merged?

the-snowwhite:
@mickaelistria
I close this PR assuming https://github.com/eclipse-platform/eclipse.platform.swt/pull/879
Will be able to act as a replacement after though testing in 2023-12_R

iloveeclipse:
@the-snowwhite : once again: can you **clearly** answer to the question: **have you successfully tested that patch and are OK with it being merged**?

@mickaelistria : can **you** please test this patch - as obviously you were test the original one?

the-snowwhite (changes requested):
@iloveeclipse 
The title is mis leading
This is a functional change for an already functional verified and merged PR
Will you please wait til after the freeze ?

iloveeclipse:
> Will you please wait til after the freeze ?

No, because original patch is NOT OK.

Can you **please** answer to the question **have you successfully tested that patch and are OK with it being merged?**

the-snowwhite:
@iloveeclipse 
@mickaelistria 
The original intention of testing required that the buxfix was able to run on 2012-09_R So next logical test step will be
to cherrypick the https://github.com/eclipse-platform/eclipse.platform.swt/pull/879 to r_29_Maintanance and test via the
https://github.com/eclipse-platform/eclipse.platform.swt/discussions/858
procedure.
However if that works next logical step will be I have to figure out which branch corresponds to the most recent binaries
created by the I-build (I think @mickaelistria  gave me the info yesterday ?).
And then download the binaries from: here ?
https://www.eclipse.org/downloads/packages/2023-03m3
and test the 3 binaries I can produce in the plugin folder ...
oh wait the filename is 
eclipse-inst-jre-linux64.tar.gz
Hmmm will ruin my current 2012-09 installs ?
Ill start on the test procedure but give me a few hours or so
OK ?

mickaelistria (approved):
Using the ClipboardExample
Without patches from #863 , I can copy from StyledText to Qt widget (I'm using Carla application as example of a KDE/Qt application) but I cannot copy from Carla/Qt/Kde... to StyledText.
With current master including patches from #863, I can copy/paste in both directions (that what I tested before merging)
With this patch, I can also copy in both directions.

So I'm totally in favor of merging this patch, which achieves the same result with simpler code and removing some extra unnecessary processing.

Thanks @iloveeclipse for the polishing here.

iloveeclipse:
Thanks @mickaelistria, merging.

PR #880, opened 2023-11-10
Revert changes from #863

iloveeclipse:
Revert changes from https://github.com/eclipse-platform/eclipse.platform.swt/pull/863 made to fix https://github.com/eclipse-platform/eclipse.platform.swt/issues/851 for following reasons:

1) The change does unneeded things on X11 and might affect X11 copy/paste behavior in unpredictable way (the code transfers bits from Java to native clipboard 3x times more and at unexpected time)
2) The change was added too late (during M3 stabilization week) and is not fixing any regression we've observed in 4.30.
3) The change was not properly reviewed / tested.

So to not affect 4.30 stability, I prefer to revert the change completely *if* the alternative proposal in https://github.com/eclipse-platform/eclipse.platform.swt/pull/879 that fixes points above doesn't work on Wayland.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Drag and Drop/gtk/org/eclipse/swt/dnd/ClipboardProxy.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Drag and Drop/gtk/org/eclipse/swt/dnd/ClipboardProxy.java
@@ -169,9 +169,6 @@ boolean setData(Clipboard owner, Object[] data, Transfer[] dataTypes, int clipbo
 				System.arraycopy(entries, 0, tmp, 0, entries.length);
 				tmp[entries.length] = entry;
 				entries = tmp;
-				TransferData tdata = new TransferData();
-				tdata.type = typeIds[j];
-				transfer.javaToNative(data[i], tdata);
 			}
 		}
 
--- a/bundles/org.eclipse.swt/Eclipse SWT Drag and Drop/gtk/org/eclipse/swt/dnd/TextTransfer.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Drag and Drop/gtk/org/eclipse/swt/dnd/TextTransfer.java
@@ -13,8 +13,6 @@
  *******************************************************************************/
 package org.eclipse.swt.dnd;
 
-import java.nio.charset.*;
-
 import org.eclipse.swt.internal.*;
 import org.eclipse.swt.internal.gtk.*;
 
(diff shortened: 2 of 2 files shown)


iloveeclipse:
@mickaelistria, @akurtakov : FYI.

mickaelistria (changes requested):
I'm against reverting this patch. This has proven useful as it allows to copy paste between Eclipse and KDE apps on Wayland, which was impossible so far; so it's a great improvement.
If you can demonstrate it actually causes a regression on X11, I'd be OK to revert, but reverting a bugfix "just in case" is not the best way to move ahead.

> The change was added too late (during M3 stabilization week)

Alex and I have approved it and tested it and assumed it's viable for M3 and important enough to not delay it further.
The fact that it was merged in stabilization week per se is not a good argument for a revert.

> The change was not properly reviewed / tested.

We tested it, and while the code could have been better, it was evaluated as good enough.

I hope I can try https://github.com/eclipse-platform/eclipse.platform.swt/pull/879 soon and it works as well so we can just merge it and close this chapter.

iloveeclipse:
> If you can demonstrate it actually causes a regression on X11, I'd be OK to revert, but reverting a bugfix "just in case" is not the best way to move ahead.

Put breakpoint into `TextTransfer.javaToNative()` and see how many times it is hit on X11 with/without patch.

iloveeclipse:
> This has proven useful as it allows to copy paste between Eclipse and KDE apps on Wayland, which was impossible so far; so it's a great improvement.

X11 port is proven to be stable, and Wayland support is proven to be not nice in many different areas. So honestly, I don't see why do we are rushing to have this patch in next release if we can get it (tested / fixed) in 4.31 in 3 months. 

On the other side, I don't want to put in danger many X11 users that wait for a stable 4.30 release with a patch that affects very basic functionality and is not thoroughly tested during release cycle. 

To avoid the revert, either https://github.com/eclipse-platform/eclipse.platform.swt/pull/879 should be confirmed to work under Wayland, or **at least** a guard should be added around the questionable code in `ClipboardProxy.setData()` to not be executed under X11.

the-snowwhite:
@mickaelistria 
@iloveeclipse 
I have just got to being able to test https://github.com/eclipse-platform/eclipse.platform.swt/pull/879
With the swt clipboard example app in a wayland session with: org.eclipse.swt.internal.gtk.version=3.24.33
I was fully able to copy and paste to from the qt/KDE based app kate and æøå also gave no problems.
the part of testing with R_29 was neglected as that probably was what led to ClipboardProxy.java being modified. 
Thank you both for helping to get an elegant wayland clipboard fix in place for 2023-12 :-1:  :-)

the-snowwhite (changes requested):
@mickaelistria 
I have now tested this PR a second time and while TextTransfer copy past fully works as expected
RTFTransfer and HtmlTransfer 
do not provide an uft-8 variant as I think the PR this is supposed to replace does.
Rerunning the exercise of testing the R_29 backport on this simplefied patch gives no meaning as the former PR was initiallw developed on R_29_Maintenance and needed the 3 added lines in ClipboardProxy.java
I would like to put this albeit much more simple and elegant solution on halt until after
2023_R

mickaelistria:
@the-snowwhite That still doesn't answer my question on https://github.com/eclipse-platform/eclipse.platform.swt/pull/879 : do you think we should merge https://github.com/eclipse-platform/eclipse.platform.swt/pull/879 now or not?

iloveeclipse:
Closing in favour of https://github.com/eclipse-platform/eclipse.platform.swt/pull/879

Issue #882, opened 2023-11-11
TreeItem.getItemCount() and TreeItem.getItem(int) have linear complexity on Linux


basilevs:
The org.eclipse.swt.widgets.Tree is slow on Linux. In particular `TreeItem.getItemCount()` and `TreeItem.getItem(int)` have linear complexity (the time of execution is proportional to the number of children).

**To Reproduce**

```java
import org.eclipse.swt.SWT;
import org.eclipse.swt.layout.FillLayout;
import org.eclipse.swt.widgets.Shell;
import org.eclipse.swt.widgets.Tree;
import org.eclipse.swt.widgets.TreeItem;

public class TreeTest {

	private Shell shell = new Shell();
	{
		shell.setLayout(new FillLayout());
		shell.open();
	}


	public void getItemShouldHaveConstantTime() {
		for (int i = 0; i < 1000; i++) {
			measureGetItemTime(100); // warmup
		}
		double elapsed_10 = measureGetItemTime(10);
		double elapsed_100000 = measureGetItemTime(100000);
		double ratio = elapsed_100000 / elapsed_10;

		System.out.printf("Time for 10 elements: %f ns\nTime for 100000 elements: %f ns\nRatio: %f\n",elapsed_10, elapsed_100000, ratio);
	}

	private long measureGetItemTime(int count) {
		Tree tree = createTree();
		try {
			TreeItem subject = new TreeItem(tree, SWT.NONE);

			createChildren(subject, count);
			readAndDispatch();

			// warmup
			for (int i = 0; i < 1000; i++) {
				System.nanoTime();
				subject.getItem(count - 1);
			}

			long start = System.nanoTime();
			for (int i = 0; i < 10000; i++) {
				subject.getItem(count - 1);
			}
			long stop = System.nanoTime();
			long elapsed = stop-start;
			return elapsed;
		} finally {
			tree.dispose();
		}
	}

	private void createChildren(TreeItem subject, int count) {
		for (int i = 0; i < count; i++) {
			new TreeItem(subject, SWT.NONE);
		}
	}

	private Tree createTree() {
		Tree result = new Tree(shell, SWT.NONE);
		return result;
	}

	private void readAndDispatch() {
		for (int i = 0; i < 10; i++ ) {
			while(shell.getDisplay().readAndDispatch()) {}
		}
	}


}
```

This test wrapped in a product: 
[...]

PR #883, opened 2023-11-11
Speed up TreeItem.getItemCount() and TreeItem.getItemCount() #882

basilevs:
This alleviates #882, by caching results of heavy operations.

On Linux, methods `TreeItem.getItemCount()` and `TreeItem.getItem(int)`  take time proportional to child count (have linear asymptotic complexity). This causes quadratic execution time in JFace `TreeViewer.reveal()` method, which uses these to eagerly populate expanded virtual Tree nodes.

This PR demonstrates the problem with corresponding JUnit tests and suggests an incomplete solution, that relies on caching.
An alternative solution is discussed in #882 to reimplement SWT GTK child tracking  in a manner similar to that of Mac OS implementation, so this proposal should not be merged, if anyone is ready to implement a better and more complete solution.


basilevs:
I've a straightforward caching of item count helps in `getItemCount()` but not in `getItem()`. 
`getItem()` asks GTK to fetch the n-th child and `org.eclipse.swt.internal.gtk.GTK.gtk_tree_model_iter_nth_child(long, long, long, int)` has linear execution time again.

So at best I've reduced time in Jface case of interest by half, not by a thousand times

basilevs:
I'm seeing test results matching a previous commit in CI.

basilevs:
The immediate problem is solved, but in a strange way. There is a suggestion to abolish GTK queires for children entirely and track parent-child relationship in Java code like MacOS implementation does. I'm not ready to work on that, as scope is much greater, so I publish my current hackish solution for review and discussion.

basilevs:
Sorry, I've forgot to test the JFace reveal test from https://github.com/eclipse-platform/eclipse.platform.ui/issues/649 Back to draft.

basilevs:
JFace reveal is still quadratic:
```
1153 ms to process 10000 items
79621 ms to process 100000 items
The time to reveal scales as O(n^1.84) as the size of the model grows from 10000 to 100000
```
My tests probably do not adequately represent JFace interaction with Tree.

basilevs:
I give up for now.

Issue #885, opened 2023-11-11
builds of PR to R_29_Maintenance branch fail


the-snowwhite:
**Describe the bug**
builds of PR to R_29_Maintenance branch fails with the error message:

    [ERROR] Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:4.0.2:compare-version-with-baselines
    (compare-attached-artifacts-with-release) on project org.eclipse.swt: Only qualifier changed for 
    (org.eclipse.swt/3.124.100.v20231111-0040). Expected to have bigger x.y.z than what is available in baseline 
    (3.124.100.v20230825-1346)

running the command:

    mvn clean verify --batch-mode -DforkCount=0 -Dcompare-version-with-baselines.skip=false -Dmaven.compiler.failOnWarning=true -Dmaven.test.failure.ignore=true -Dmaven.test.error.ignore=true

**To Reproduce**
A standalone SWT snippet to reproduce the behavior (preferable)
(or) 
Steps to reproduce the behavior.
[Perform a PR to R_29_maintenance branch](https://github.com/eclipse-platform/eclipse.platform.swt/pull/881)

**Expected behavior**
The build succedes

**Environment:**

https://ci.eclipse.org/releng/job/eclipse.platform.swt/view/change-requests/job/PR-881/1/consoleFull
 
**Workaround (or) Additional context**
#858 

Update:
**The purpose of fixing this issue is**

1.being able to enable successful PR builds on any branch
2. Fetching the build artifacts.
3. Closing the PR without a merge.

merks:
You must increase the version in the MANIFEST.MF.   For the maintenance stream it would be add 1 to the third segment for this change.

the-snowwhite:
I can find 2 MANIFEST.MF files in
/home/mib/git/eclipse.platform.swt/bundles/org.eclipse.swt
    
    /home/mib/git/eclipse.platform.swt/bundles/org.eclipse.swt/target/MANIFEST.MF
   /home/mib/git/eclipse.platform.swt/bundles/org.eclipse.swt/META-INF/MANIFEST.MF

only

    /home/mib/git/eclipse.platform.swt/bundles/org.eclipse.swt/target/MANIFEST.MF

contains: 1346
So that must be the correct one.
@merks 
Thanks  Do you know how often this number needs to be increased (like for each new PR) ?

merks:
[...]

PR #886, opened 2023-11-11
fix: #885 make R_29 builds succeed again.

the-snowwhite:
This PR is not to be merged.
It is a placeholder for solving https://github.com/eclipse-platform/eclipse.platform.swt/issues/888
The purpose is only for testing what can make the builder deliver build artifacts that then can be downloaded.
In this specific case for testing a bugfix by copying 3 produced binaries into a working Standard Eclipse PHP 2023-09
Which it what the R_29 maintenance branch works on.
The cherry picked version of this commit is after approval then supposed to be able to be merged onto R_29_Maintenance or any other Maintenance branch exactly when the branch is no longer required to deliver binaries
to the stream.

### DIFF
--- a/bundles/org.eclipse.swt/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.swt/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-Name: %pluginName
 Bundle-Vendor: %providerName
 Bundle-SymbolicName: org.eclipse.swt; singleton:=true
-Bundle-Version: 3.124.100.qualifier
+Bundle-Version: 3.124.101.qualifier
 Bundle-ManifestVersion: 2
 Bundle-Localization: plugin
 DynamicImport-Package: org.eclipse.swt.accessibility2


bundles/org.eclipse.swt/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-Name: %pluginName
 Bundle-Vendor: %providerName
 Bundle-SymbolicName: org.eclipse.swt; singleton:=true
-Bundle-Version: 3.124.100.qualifier
+Bundle-Version: 3.125.100.qualifier
```
merks, line 5: Recall I said increment the third segment but here you incremented the second segment.

You'll probably want  to read this:

https://wiki.eclipse.org/Version_Numbering

merks (changes requested):
It's incorrectly incremented.

the-snowwhite:
The purpose of fixing the build is not to afterwards enable a PR merge with new functionality.
The Purpose is enabling the #858 method.
That means a PR does not lead to merges but only to being able to fetch build artifacts after a successful build.
Then the PR is closed again without a merge.
I do not see anything in https://wiki.eclipse.org/Version_Numbering that covers that use case.
Am I wrong ?

the-snowwhite:
@merks 
Also 3.124.200 is already taken See here:
https://ci.eclipse.org/releng/job/eclipse.platform.swt/view/change-requests/job/PR-883/lastSuccessfulBuild/artifact/eclipse.platform.swt.binaries/bundles/org.eclipse.swt.gtk.linux.x86_64/target/

So the last logical option is:
3.124.110 ?

merks:
I thought explaining +1 on the third segment was pretty clear.  I was on an IPhone so it wasn't easy to tell you exactly what the number should be, given I didn't know the actual value.  

If the value is x.y.z.qualifier, then the next one would be x.y.z+1.qualifier and here z = 100.  The reason why the development stream does +100 service segment increments is to leave a large gap for many+1 increments to that service segment in a service stream.  That article I linked tries to explain that, but you saw little value there. You first incremented the second segment, not the third and now propose incrementing the third segment by 10.    😕 

Note that you do generate a significant volume of emails.  It's somewhat tempting to leave answers to someone else, so that I don't end up with a long chain of follow up emails.  No offense intended.  It's just an observation and something to consider.

the-snowwhite:
> I thought explaining +1 on the third segment was pretty clear. I was on an IPhone so it wasn't easy to tell you exactly what the number should be, given I didn't know the actual value.
> 
> If the value is x.y.z.qualifier, then the next one would be x.y.z+1.qualifier and here z = 100. The reason why the development stream does +100 service segment increments is to leave a large gap for many+1 increments to that service segment in a service stream. That article I linked tries to explain that, but you saw little value there. You first incremented the second segment, not the third and now propose incrementing the third segment by 10. 😕
> 
> Note that you do generate a significant volume of emails. It's somewhat tempting to leave answers to someone else, so that I don't end up with a long chain of follow up emails. No offense intended. It's just an observation and something to consider.

Sorry for the clutter in your inbox generated by me.
I do not know how your notification system works.
But i only get e-mails when someone mentions me.
Do you get more that that ?

Im not sure if altering the only way to go.
`Bundle-Version: 3.124.100.qualifier`

Finding a way to skip the mvn commands the require that would be better.
For this purpose only the last 2 mvn commands in the build log are needed.

the-snowwhite:
`Bundle-Version: 3.124.101.qualifier`
Is what is in the current PR and it has been running for hours ....
https://github.com/eclipse-platform/eclipse.platform.swt/pull/886/files

laeubi:
Just note that PRs are not the way to get an artifact for local build, you should instead build it locally using maven+ant and the native build toolchain.

For that purpose look into what the `Jenkins` file does to get some hints on how to compile locally, there you then can increase version (for test) like you desire and only need to rebuild small portions of the code, so no need to trigger a full build.

the-snowwhite:
@laeubi 
please look at the new feature request #888 
Im now closing this PR

Issue #888, opened 2023-11-11
How can I make it possible to always be able to make a pull request to any branch and only fetch the build artifacts without any merge ?


the-snowwhite:
How can I make it possible to always be able to make a pull request to any branch and only fetch the build artifacts without any need to merge (and then close the PR after a build success)?

Based upon coming up with [a better answer for this question](https://github.com/eclipse-platform/eclipse.platform.swt/discussions/858)
[Than the fallback solution here:](https://github.com/eclipse-platform/eclipse.platform.swt/discussions/858#discussioncomment-7538739)

[The new answer is here:](https://github.com/eclipse-platform/eclipse.platform.swt/discussions/858#discussioncomment-7542214)

The current implementation of this new feature is is currently stuck in [this issue](https://github.com/eclipse-platform/eclipse.platform.swt/issues/885) by [this blocked pull request:](https://github.com/eclipse-platform/eclipse.platform.swt/pull/886/files#diff-6485b884863fd4f54a20976acef942fe0dd257eb4e72ca25533fd84b7b6fe2e6)
here:
https://github.com/eclipse-platform/eclipse.platform.swt/pull/886#pullrequestreview-1726125380

Is it possible to update [this book](https://wiki.eclipse.org/Version_Numbering) to cover an always successful build where only the build artifacts are required and no merge is performed ?
 
Or is some sort of alternative needed ?

the-snowwhite:
https://github.com/eclipse-platform/eclipse.platform.swt/pull/886#issuecomment-1806947866

the-snowwhite:
All I need is for the builder to skip the mvn -check command that requires an update of the.
`Bundle-Version: 3.124.100.qualifier`

only the last 2 mvn commands are required to generate build artifacts without a merge:

In the swt.binaries folder

    mvn --batch-mode -Pbuild-individual-bundles -DforceContextQualifier=zzz -Dcompare-version-with-baselines.skip=true -Dmaven.compiler.failOnWarning=true install

In the main swt folder:

    mvn --batch-mode -V -U -DforkCount=1 -Dcompare-version-with-baselines.skip=false -Dmaven.compiler.failOnWarning=true --fail-at-end -DskipNativeTests=false -DfailIfNoTests=false clean install

mickaelistria:
you should build SWT locally to get what you want. Looking in the Jenkinsfile will show you concrete commands to build SWT.
[...]

PR #889, opened 2023-11-12
This PR is not to be merged.It is a placeholder for solving #888

the-snowwhite:
Update:
The purpose is making a Maintenance branch builder change functionality
so that it can deliver build artifacts  that then can be downloaded.
To test bugfixes under a freeze period.

A specific example could be
Testing a bugfix by copying 3 produced
binaries into a working Standard Eclipse PHP 2023-09.
Without the need to have a working compiler available.

The cherry picked version of this commit is after approval then supposed
to be able to be merged onto a Maintenance branch exactly when the
branch is no longer required to deliver binaries
to the stream.
Or
It can be delivered by
dependabot[bot]

### DIFF
--- a/bundles/org.eclipse.swt/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.swt/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-Name: %pluginName
 Bundle-Vendor: %providerName
 Bundle-SymbolicName: org.eclipse.swt; singleton:=true
-Bundle-Version: 3.124.100.qualifier
+Bundle-Version: 3.124.101.qualifier
 Bundle-ManifestVersion: 2
 Bundle-Localization: plugin
 DynamicImport-Package: org.eclipse.swt.accessibility2
--- a/bundles/org.eclipse.swt/pom.xml
+++ b/bundles/org.eclipse.swt/pom.xml
@@ -21,7 +21,7 @@
     </parent>
     <groupId>org.eclipse.swt</groupId>
     <artifactId>org.eclipse.swt</artifactId>
-    <version>3.124.100-SNAPSHOT</version>
+    <version>3.124.101-SNAPSHOT</version>
     <packaging>eclipse-plugin</packaging>
 
     <properties>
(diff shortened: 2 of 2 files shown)


the-snowwhite:
Hmmm
I created the commit at 0759
But adjusted the timestamp to when I thought I would be able to make the PR foo I adjusted the time stamp to 0800
`<forceContextQualifier>v20231112-0800</forceContextQualifier>`

I get the error message:
09:04:35  [ERROR] Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:4.0.2:compare-version-with-baselines (compare-attached-artifacts-with-release) on project org.eclipse.swt: Only qualifier changed for (org.eclipse.swt/3.124.100.v20231112-0759). Expected to have bigger x.y.z than what is available in baseline (3.124.100.v20230825-1346) 

    You
    Could the reason for the error be that the qualifier was set to v20231112-0800  but the commit was created at 0759 and the build also started at 0759 ? 
    ChatGPT
    ChatGPT

    Yes, the timing discrepancy you've described could indeed be a contributing factor to the error you're encountering, particularly if your build process or versioning strategy involves generating or checking qualifiers based on timestamps.

    In Eclipse and Maven projects, qualifiers are often used to append build timestamps or other unique identifiers to the version number. If your build process is designed to generate a qualifier based on the current timestamp, and if there's a mismatch between the expected qualifier and the actual time of the build or commit, this could lead to issues.

    In your case:

        The qualifier is set to v20231112-0800, which suggests that the build or version is intended for 08:00 on November 12, 2023.
        However, the commit and build started at 07:59, which is a minute earlier than the timestamp in the qualifier.

    This mismatch might cause the build process to fail, especially if the build system or versioning checks are strict about the correlation between the build timestamp and the qualifier. The build system might be expecting the qualifier to match the actual build time, and when it doesn't, it throws an error.

    To resolve this issue, you could:

        Ensure Accurate Timing: Make sure the qualifier accurately reflects the time of the build. This might involve adjusting your build scripts or processes to correctly generate or assign the qualifier based on the actual build time.

        Review Versioning Strategy: Revisit your versioning strategy to ensure it aligns with your build process. If relying on timestamps is causing issues, consider whether there's a more reliable method you can use.

        Adjust Build Configuration: If necessary, modify your build configuration to be more forgiving about timing mismatches, though this should be done with caution to avoid future versioning issues.

    Remember, consistent and accurate versioning is crucial for effective software development, especially in complex projects with multiple dependencies.

mickaelistria:
> This PR is not to be merged.

You can create the PR as draft or use "Convert to draft" after creation then.
PRs are not a place for sandbox work, please work on your own branch and submit PRs only when you intend your code to be merged, as a regular PR is ready, as a draft if you need further collaboration to happen before the code is ready.

The development and workflows we implement here are specific. ChatGPT won't be very helpful because its training reflects relatively standard/popular data; reading and understanding existing doc and code (you can find interesting link with regular search engines) is a better way to find solution to specific problems.

PR #903, opened 2023-11-26
[GTK] Optimize TreeItem constructor #882

basilevs:
This eliminates a single hotspot. The full complexity for tree building is still quadratic.

Profile before the fix:
Screenshot 2023-11-26 at 21 18 45

Profile after the fix:
Screenshot 2023-11-26 at 21 22 34

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Tree.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Tree.java
@@ -191,7 +191,7 @@ TreeItem _getItem (long iter) {
 		parentIter = OS.g_malloc (GTK.GtkTreeIter_sizeof ());
 		GTK.gtk_tree_model_get_iter (modelHandle, parentIter, path);
 	}
-	items [id] = new TreeItem (this, parentIter, SWT.NONE, indices [indices.length -1], false);
+	items [id] = new TreeItem (this, parentIter, SWT.NONE, indices [indices.length -1], iter);
 	GTK.gtk_tree_path_free (path);
 	if (parentIter != 0) OS.g_free (parentIter);
 	return items [id];
@@ -199,11 +199,14 @@ TreeItem _getItem (long iter) {
 
 TreeItem _getItem (long parentIter, int index) {
 	long iter = OS.g_malloc (GTK.GtkTreeIter_sizeof ());
-	GTK.gtk_tree_model_iter_nth_child(modelHandle, iter, parentIter, index);
-	int id = getId (iter, true);
-	OS.g_free (iter);
-	if (items [id] != null) return items [id];
-	return items [id] = new TreeItem (this, parentIter, SWT.NONE, index, false);
+	try {
+		GTK.gtk_tree_model_iter_nth_child (modelHandle, iter, parentIter, index);
+		int id = getId (iter, true);
+		if (items [id] != null) return items [id];
+		return items [id] = new TreeItem (this, parentIter, SWT.NONE, index, iter);
+	} finally {
+		OS.g_free (iter);
+	}
 }
 
 void reallocateIds(int newSize) {
@@ -3532,7 +3535,7 @@ void setItemCount (long parentIter, int count) {
 		OS.g_free (iters);
 	} else {
 		for (int i=itemCount; i<count; i++) {
-			new TreeItem (this, parentIter, SWT.NONE, itemCount, true);
+			new TreeItem (this, parentIter, SWT.NONE, itemCount, 0);
 		}
 	}
 	if (!isVirtual) setRedraw (true);
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/performance/Test_org_eclipse_swt_widgets_Tree.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/performance/Test_org_eclipse_swt_widgets_Tree.java
@@ -1,0 +1,359 @@
+/*******************************************************************************
+ * Copyright (c) 2023, 2023 IBM Corporation and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Vasili Gulevich - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.swt.tests.junit.performance;
+
+import static org.junit.Assert.assertTrue;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Deque;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Objects;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.Consumer;
+import java.util.function.IntFunction;
+
+import org.eclipse.swt.SWT;
+import org.eclipse.swt.graphics.Color;
+import org.eclipse.swt.graphics.Font;
+import org.eclipse.swt.layout.FillLayout;
+import org.eclipse.swt.tests.junit.SwtTestUtil;
+import org.eclipse.swt.widgets.Control;
+import org.eclipse.swt.widgets.Shell;
+import org.eclipse.swt.widgets.Tree;
+import org.eclipse.swt.widgets.TreeItem;
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TestName;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameters;
+
+@RunWith(Parameterized.class)
+public class Test_org_eclipse_swt_widgets_Tree {
+	enum Shape {
+		BINARY {
+			@Override
+			void buildTree(Tree tree, int size, Consumer<TreeItem> initializeItem) {
+				tree.setItemCount(2);
+				{
+					int leftChildCount = size / 2;
+					tree.getItem(0).setData("childCount", leftChildCount - 1);
+					tree.getItem(1).setData("childCount", size - leftChildCount - 1);
+				}
+				boolean virtual = (tree.getStyle() & SWT.VIRTUAL) != 0;
+				if (virtual) {
+					tree.addListener(SWT.SetData, event -> {
+						TreeItem item = (TreeItem) event.item;
+						initializeItem.accept(item);
+						createBinaryChildren(item, initializeItem);
+					});
+				} else {
+					createBinaryBranch(tree.getItem(0), initializeItem);
+					createBinaryBranch(tree.getItem(1), initializeItem);
+				}
+			}
+
+			private void createBinaryBranch(TreeItem item, Consumer<TreeItem> initializeItem) {
+				initializeItem.accept(item);
+				for (TreeItem child : createBinaryChildren(item, initializeItem)) {
+					createBinaryBranch(child, initializeItem);
+				}
+			}
+
+			@Override
+			protected TreeItem lastItem(Tree tree) {
+				return lastItem(tree.getItem(1));
+			}
+
+			private List<TreeItem> createBinaryChildren(TreeItem item, Consumer<TreeItem> initializeItem) {
+				int childCount = (int) item.getData("childCount");
+				int leftChildCount = childCount / 2;
+				int rightChildCount = childCount - leftChildCount;
+				ArrayList<TreeItem> result = new ArrayList<>();
+				if (leftChildCount > 0) {
+					TreeItem left = new TreeItem(item, SWT.NONE);
+					left.setData("childCount", leftChildCount - 1);
+					result.add(left);
+				}
+				if (rightChildCount > 0) {
+					TreeItem right = new TreeItem(item, SWT.NONE);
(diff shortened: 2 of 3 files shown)


SyntevoAlex:
I would prefer some numbers here in the commit message.
For example:

when inserting 100_000 items into the Tree using this code
```java
Tree tree = new Tree(shell, SWT.VIRTUAL);
for (int i = 0; i < 100_000; i++)
{
    TreeItem item = new TreeItem(tree, 0);
    item.setText("item#" + i);
}
```
the speedup on my machine is 99.99sec -> 88.88sec

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Tree.java
```diff
@@ -191,19 +191,23 @@ TreeItem _getItem (long iter) {
...
+	items [id] = new TreeItem (this, parentIter, SWT.NONE, indices [indices.length -1], iter);
 	GTK.gtk_tree_path_free (path);
 	if (parentIter != 0) OS.g_free (parentIter);
 	return items [id];
 }
 
 TreeItem _getItem (long parentIter, int index) {
-	long iter = OS.g_malloc (GTK.GtkTreeIter_sizeof ());
-	GTK.gtk_tree_model_iter_nth_child(modelHandle, iter, parentIter, index);
-	int id = getId (iter, true);
```
SyntevoAlex, line 203: Please do not change code style (here, spaces and return on same line).
While I myself find SWT's code style weird, it's still the code style that already exists.

SyntevoAlex (changes requested):
With these small issues fixed, I am willing to merge this PR.

SyntevoAlex:
Please also rename commit to have `[GTK]` prefix, because it only affects GTK. Also, the commit message is poor in the usual way: it answers `what` instead of `why`. A better commit message header would be:
`[GTK] Tree optimization: speedup TreeItem constructor by using known iterator`

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/AllWidgetTests.java
```diff
@@ -33,6 +33,7 @@
 		Test_org_eclipse_swt_widgets_List.class, Test_org_eclipse_swt_widgets_Text.class,
 		Test_org_eclipse_swt_widgets_ScrollBar.class, Test_org_eclipse_swt_widgets_Sash.class,
 		Test_org_eclipse_swt_widgets_Tree.class, Test_org_eclipse_swt_widgets_TabFolder.class,
+		org.eclipse.swt.tests.junit.performance.Test_org_eclipse_swt_widgets_Tree.class,
```
SyntevoAlex, line 36: It is not appropriate to put performance tests into the main tests package.
Unit tests should run as fast as possible, and performance tests are slow by design.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/performance/Test_org_eclipse_swt_widgets_Tree.java
```diff
@@ -0,0 +1,359 @@
...
+		Tree result = new Tree(shell, virtual ? SWT.VIRTUAL : SWT.NONE);
+		shell.layout();
+		result.setRedraw(false);
+		shape.buildTree(result, size, initialize);
+		result.setRedraw(true);
+		return result;
+	}
+
+	/** Ensure that given function grows within acceptable polynomial degree */
+	private void assertMaximumDegree(double maximumDegree, IntFunction<Double> function) {
```
SyntevoAlex, line 300: Asserting on times is unreliable.
If system experience sudden slowdown for a second, tests will fail.
It's fine however if tests are designed to be run manually only.

The other problem is that you're asserting complexity for ALL platforms, not just GTK.
To me it makes sense to convert JUnit test to a manual snippet.

SyntevoAlex (commented):
Now that this commit has more issues, I would rather remove it from PR for now.

SyntevoAlex:
With small fixes, commit with the fix is good to merge.
As for the commit with tests, I would rather remove it from PR for now.

I have used up my (rather small) time to look at someone else's PRs this week, so I could only review the next batch in a few days. Feel free to ask someone else to review, but it's not going to be too easy because there're very few active members currently.

If you polish the first commit and don't introduce new problems, I can merge it when I see it.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/performance/Test_org_eclipse_swt_widgets_Tree.java
basilevs: I'm trying to fix a performance problem. Note that I do not assert the absolute time and instead check the ratios. There is no way to prevent performance regression besides running the tests in CI. I specifically enabled these tests on CI to be able to have a single point of reference and universal source of timings for the process of this review. The tests are designed to be run on CI.

Checking all the platforms is intentional, as the problem I'm trying to fix affects only Linux and tests should demonstrate it. No platform should exhibit quadratic behaviors, so they also serve as regression tests.

There are already manual snippets in the repository testing performance, but they failed to find the problem in question. I assume nobody runs them.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/AllWidgetTests.java
basilevs: Where should I put these tests to run them on CI?

SyntevoAlex: To my knowledge, there is no such place currently.
You can introduce new GitHub job maybe. But frankly I would rather not have it at all. Once problems are fixed, they are unlikely to come back. Yet you are going to burn hours of CPU time over years.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Tree.java
basilevs: Fixed

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/AllWidgetTests.java
basilevs: Can leave them in for now and disable before merge to master?

SyntevoAlex: You can keep whatever you want on your branch.
You can create another PR for testing.
If you want *this* PR to be merged, remove tests.

basilevs: Removed

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/performance/Test_org_eclipse_swt_widgets_Tree.java
basilevs: Test is now removed from suites and stays passive next to other abandoned performance tests.

SyntevoAlex:
You still have spaces changed in `_getItem`.

SyntevoAlex:
Also, a nitpick: you added unneeded noise in commit message.
Instead of this
```
Before the change:
jfaceReveal[Shape: STAR, virtual: true]
Execution time should grow as 1.100000 degree polynom.
Time for 10000 elements: 620682717.000000 ns
Time for 100000 elements: 62804214489.000000 ns
Ratio: 10.118570
Degree: 2.005119

After the change:
jfaceReveal[Shape: STAR, virtual: true]
Execution time should grow as 1.100000 degree polynom.
Time for 10000 elements: 501378662.000000 ns
Time for 100000 elements: 52781899817.000000 ns
Ratio: 10.527353
Degree: 2.022319
```

I would write
```
In test jfaceReveal[Shape: STAR, virtual: true]:
10_000 elements: 620_682_717ns -> 501_378_662ns
100_000 elements: 62_804_214_489ns -> 52_781_899_817ns
```

Can you see how much easier it is to understand without the noise?
But this is just a rant, I would be pleased to have it changed, but this is not required.

basilevs:
> You still have spaces changed in `_getItem`.
Added spaces before opening parentheses. Hard to spot. Is there a formatter configuration? I could not find one in contribution guide.

basilevs:
> Also, a nitpick: you added unneeded noise in commit message. Instead of this
> I would write

New message:
```
[GTK] Tree optimization: speedup TreeItem constructor by using known
iterator #882

This change eliminates a single hotspot. The overall complexity of
tree building and element reveal is still quadratic.

Performance improvement is proven by running
org.eclipse.swt.tests.junit.performance.Test_org_eclipse_swt_widgets_Tree.jfaceReveal()

In test jfaceReveal[Shape: STAR, virtual: true]:
10_000 elements: 620_682_717ns -> 501_378_662ns
100_000 elements: 62_804_214_489ns -> 52_781_899_817ns (-16%)
```

SyntevoAlex:
Unfortunately you still have spaces changed.
I don't know about formatter configuration. Use some git client that can show edited spaces well. I use SmartGit.

SyntevoAlex (approved):

basilevs:
> Unfortunately you still have spaces changed.

Fixed two more.

SyntevoAlex:
Oh well. So that you're not upset by spaces, I'll forget it in this commit.
I will however ask you to not change codestyle unnecessarily next time.

SyntevoAlex:
Once CI finishes testing this PR, I will merge it.

iloveeclipse:
Just in case: master is not open yet for merging.

SyntevoAlex:
Ahh, I always forget about that. Thanks for reminding.
Do I understand it correctly that we'll need to wait until 4.30 is released?

basilevs:
@SyntevoAlex thanks for the review.

@iloveeclipse  I've considered merging this in #902 meanwhile, but confused everyone with workflow discussion, so I'm not sure anymore 🤷

SyntevoAlex:
I hope I didn't upset you too much with all the requests. Now that you understand SWT things a bit better, next PR will likely require fewer changes.

SyntevoAlex:
@iloveeclipse while you're around, can I ask you about `assert` usage in SWT?
This PR introduces another `assert` which is the... third! Do you have any ideas why it was avoided before?

basilevs:
> @iloveeclipse while you're around, can I ask you about `assert` usage in SWT? This PR introduces another `assert` which is the... third! Do you have any ideas why it was avoided before?

https://github.com/eclipse-platform/eclipse.platform.swt/blob/518769da51ce22f8242f3feaed522dcf58ebfc4e/bundles/org.eclipse.swt/Readme.md?plain=1#L36-L47

SyntevoAlex:
@basilevs you can now preserve this PR's git branch as is until it's merged.
You can now create another PR (but I won't review it until a few days after because I've hit my limit).
You can rebase the other PR on top of this branch (so that it will contain copies of commits from this PR).
When this PR is merged, the other PR will become based on master.

iloveeclipse:
> @iloveeclipse while you're around, can I ask you about `assert` usage in SWT? This PR introduces another `assert` which is the... third! Do you have any ideas why it was avoided before?

I can't say why it wasn't used but I personally prefer not to use it too, so I would prefer we would not introduce it. I've not seen yet code that would really need that ancient C feature that was by mistake added to Java.

Thinking more about it, I believe SWT was created before this mistake happened to Java.

SyntevoAlex:
Why do you consider it a mistake?
I myself like to add asserts whenever appropriate, and they actually catch quite a few bugs,

iloveeclipse:
In most places I saw an assert it was a lazy way to say "actually we should check that and somehow handle possible wrong flow, but let it go wrong in production code, it can't really happen". So either it can't happen and assert is not needed, or the code will run with wrong state in production with side effects we haven't tested, because in tests we use assertions and flow is different anyway. So either one uses defensive code style, and no assertions needed because all possible code paths are handled by regular code flow, or one uses assertions.

SyntevoAlex:
Yes I can understand this logic.

However, for every possible assert location, let's estimate the likelihood of things going wrong.
If likelihood is high, then proper error handling code is needed. But handling errors properly is difficult, so there's a cost with it.
If likelihood is low, yet still your gut feeling is worried, that's a good place for assert. It's better than nothing, and cheap enough.
If likelihood is next to zero, then it's fine to not add any handling.

So my point is that there are many cases where paying the cost for proper handling is not justified because the problem seems to be very unlikely, yet still I'm not 100% sure.

laeubi:
If I assume/assert something can't happen I throw `AssertionError` with a message why I think it can't happen, that is independent of if assertions are enabled or not if throwing an exception is not appropriate. Right after that I most of the time delete it and handle it in a proper way... because according to [Murphy's law](https://en.wikipedia.org/wiki/Murphy%27s_law) "Anything that can go wrong will go wrong, and at the worst possible time." :-)

SyntevoAlex:
master is not open. Please rebase your PR onto master so that tests are run, then I can merge it.

SyntevoAlex:
In the future, don't merge master into your branch. Instead, rebase your branch onto master.

SyntevoAlex:
In master now. Thanks for your contribution!

You can now rebase your other branch onto master.

basilevs:
Next optimization step is https://github.com/eclipse-platform/eclipse.platform.swt/pull/904

PR #907, opened 2023-11-27
Fixes https://github.com/eclipse-platform/eclipse.platform.releng.agg…

deepika-u:
…regator/issues/1557

@niraj-modi : Can you review and merge this pr please.


bundles/org.eclipse.swt/Eclipse SWT PI/common/org/eclipse/swt/internal/Library.java
```diff
@@ -13,15 +13,10 @@
...
-import java.net.JarURLConnection;
-import java.net.URL;
-import java.net.URLConnection;
-import java.nio.file.Files;
-import java.nio.file.StandardCopyOption;
-import java.util.jar.Attributes;
+import java.io.*;
+import java.net.*;
+import java.nio.file.*;
+import java.util.jar.*;
```
niraj-modi, line 20: Undo above change, in SWT we usually to specific imports:
- To keep things straightforward and avoid any possible class name collision

niraj-modi (changes requested):
Version changes look good, but we should remove the organize import stuff

niraj-modi (approved):
Looks good.

PR #904, opened 2023-11-26
[GTK] Speedup TreeItem retrieval by removing redundant iterations #882

basilevs:
This is a follow-up to #903
This is a second optimization for #882.

TreeItem and Tree have been checking indexes for bounds during indexed access and traversal (`GTK.gtk_tree_model_iter_n_children()`) . Such checks are slow as node size computation is an O(N) operation.

As Tree already does iteration of O(N) complexity when retrieving a TreeItem by its index, and such iteration detects out-of-bounds problems, preliminary model bounds check is redundant. Also, bounds check is avoidable during iteration over a node - instead of an index, GTK iterator applies directly.

Performance improvement proved by running `org.eclipse.swt.tests.junit.performance.Test_org_eclipse_swt_widgets_Tree.jfaceReveal()`.

In test jfaceReveal[Shape: STAR, virtual: true]:
10_000 elements: 1_523_338_044ns -> 654_307_337ns (-57%)
100_000 elements: 262_129_718_282ns -> 92_328_764_928ns (-65%)

<details>
<summary>
Raw measurements
</summary>
Timings before the change:

```
setForeground[Shape: STAR, virtual: false]
Execution time should grow as 1.300000 degree polynom. 
Time for 10000 elements: 435754849.000000 ns
Time for 100000 elements: 32930055467.000000 ns
Ratio: 7.557014
Degree: 1.878350

jfaceReveal[Shape: STAR, virtual: false]
Execution time should grow as 2.000000 degree polynom. 
Time for 10000 elements: 282842310.000000 ns
Time for 100000 elements: 17676761285.000000 ns
Ratio: 6.249688
Degree: 1.795858

showItem[Shape: STAR, virtual: false]
Execution time should grow as 1.900000 degree polynom. 
Time for 10000 elements: 280539616.000000 ns
Time for 100000 elements: 17282639987.000000 ns
Ratio: 6.160499
Degree: 1.789616

build[Shape: STAR, virtual: false]
Execution time should grow as 1.100000 degree polynom. 
Time for 10000 elements: 628988573.000000 ns
Time for 100000 elements: 61514983876.000000 ns
Ratio: 9.779984
Degree: 1.990338

secondTraverse[Shape: STAR, virtual: false]
Execution time should grow as 1.200000 degree polynom. 
Time for 10000 elements: 3803257.000000 ns
Time for 100000 elements: 18826476.000000 ns
Ratio: 0.495009
Degree: 0.694613

traverse[Shape: STAR, virtual: false]
[...]

PR #908, opened 2023-11-27
[GTK] Optimize TreeItem.getItem(int) by caching TreeItem.getItems() #882

basilevs:
DO NOT MERGE - this is a demonstration, not an actual solution. See below.

This is a follow up to #903 and #904.

This PR eliminates explicit iterations over `Tree` from execution profile when `TreeItem.getItems()` is called before `TreeItem.getItem(int)`. This is the case for JFace lazy TreeViewer. 

Performance improvement is proven by running relevant tests from
org.eclipse.swt.tests.junit.performance.Test_org_eclipse_swt_widgets_Tree:

In test jfaceReveal[Shape: STAR, virtual: true]:
10_000 elements:  218_583_468ns -> 141_061_117ns
100_000 elements: 20_872_245_796ns ->  10_761_449_641ns (-48%)
In test dispose[Shape: STAR, virtual: true]:
10_000 elements: 5_637_088ns -> 6_222_363ns
100_000 elements: 62_422_153ns -> 56_442_156ns

Profile before the change:
Screenshot 2023-11-26 at 21 22 34
Profile after the change:
Screenshot 2023-11-27 at 14 09 00

The solution here is ideologically incorrect, as it targets a particular way of tree traversal, specific to JFAce lazy tree. However, a more general and self-consistent approach, that moves ownership of a TreeItem from a Tree to a parent item, prototyped in https://github.com/eclipse-platform/eclipse.platform.swt/compare/master...basilevs:eclipse.platform.swt:issue_882_gtk_direct_access is a much larger change. I suggest not to merge this PR and just use it as a starting point for discussion on removal of ID system from SWT GTK Tree.
The removal itself is done in #918

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Tree.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Tree.java
@@ -181,6 +181,7 @@ void _addListener (int eventType, Listener listener) {
 }
 
 TreeItem _getItem (long iter) {
+	if (iter == 0) return null;
 	int id = getId (iter, true);
 	if (items [id] != null) return items [id];
 	long path = GTK.gtk_tree_model_get_path (modelHandle, iter);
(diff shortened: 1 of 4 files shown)


akurtakov:
What is the status here ? I see https://github.com/eclipse-platform/eclipse.platform.swt/pull/918 and https://github.com/eclipse-platform/eclipse.platform.swt/pull/904 related. Would you please clarify and close unneeded ones?

basilevs:
This is needed for review of #918 and subsequent related PRs. Review of #918 is blocked by #904, my only non-draft PR.

basilevs:
Closing to prevent accidental merge, as a review is not going to happen anytime soon.

Issue #848, opened 2023-11-02
Distorted Buttons since 2022-03


jukzi:
This problem occurs when we are in the "light"theme, in "dark" theme it works fine. The problem is there since eclipse version 2022-03. So till 2021-12 it was working fine. The changes responsible for the problem is the because the api "wmColorChild" was overrriden in the "Button" class in order to fix the bugzilla bug 
https://bugs.eclipse.org/bugs/show_bug.cgi?id=577042

_Originally posted by @SudhindraKKulkarni in https://github.com/eclipse-pde/eclipse.pde/issues/727#issuecomment-1790272732_
         
![image](https://github.com/eclipse-platform/eclipse.platform.swt/assets/51790620/ba8cbc86-d673-4a3e-ae48-d7f6811bfeed)

jukzi:
@SudhindraKKulkarni as you have already analysed the problem, can you provide a fix, please?

SyntevoAlex:
Reproducible with this snippet on my Win11:
```java
final Display display = new Display();
final Shell shell = new Shell(display);
shell.setLayout (new GridLayout (1, true));

Label label = new Label(shell, 0);
label.setText(
	"1. Run on Windows\n" +
	"2. Press Alt\n" +
	"3. Issue 848: Buttons disappear"
);

TabFolder tabFolder = new TabFolder(shell, 0);

Composite composite = new Composite(tabFolder, 0);
composite.setLayout(new GridLayout (1, true));

TabItem tabItem = new TabItem(tabFolder, 0);
tabItem.setControl(composite);
tabItem.setText("TabPage");

Button button1 = new Button(composite, SWT.PUSH);
button1.setText("&Button1");

Button button2 = new Button(composite, SWT.PUSH);
button2.setText("B&utton2");
button2.setEnabled(false);

shell.pack();
shell.open();

while (!shell.isDisposed()) {
	if (!display.readAndDispatch()) {
		display.sleep();
	}
}

display.dispose();
```

SyntevoAlex:
I plan to investigate and fix this within a week.

SyntevoAlex:
The fix is available at https://github.com/eclipse-platform/eclipse.platform.swt/pull/911

SyntevoAlex:
I think that there's no need to rush this fix into upcoming 4.30 release:
* The problem is rare (only affects light theme and only affects buttons inside `TabFolder`)
* The problem has low importance (it's funny but doesn't really block any workflows)

PR #911, opened 2023-11-27
Issue #848: [win32] Button inside TabFolder disappears when pressing Alt

SyntevoAlex:
fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/848


SyntevoAlex:
If there are no objections in few next days until master is open, I will merge it.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Button.java
```diff
@@ -1107,7 +1107,24 @@ LRESULT wmColorChild (long wParam, long lParam) {
 		// Button has "transparent" portions which need to be filled with
 		// parent's (and not Button's) background. For example, SWT.PUSH
 		// button ~2px transparent area around the button.
-		return parent.wmColorChild(wParam, lParam);
+
+		// Issue 848: wmColorChild() contains a hack that paints entire
```
jukzi, line 1111: Do you mean org.eclipse.swt.widgets.Control.wmColorChild(long, long)?
```suggestion
		// Issue 848: Control.wmColorChild.() contains a hack that paints entire
```
i don't see a recent change there and wonder why the error is a recent regression then. Also the bug report is about light theme, so i wonder why THEME_BACKGROUND does influence light theme.

SyntevoAlex: Comment improved.

The regression happened because I used `parent.wmColorChild()` (right here in the code you're looking at), which contains the unexpected hack.

SyntevoAlex: Light theme is important because `Control.findBackgroundControl()` has condition `background != -1` which prevents `THEME_BACKGROUND` hack.
So it's not about light theme, but about NOT setting any colors.

SyntevoAlex:
good to merge?

jukzi (approved):
lgtm, not tested

SyntevoAlex:
In master now.

Issue #913, opened 2023-11-28
Introduce SWT.Disposed event supplementing SWT.Dispose


tmssngr:
**Is your feature request related to a problem? Please describe.**
There is an event `Dispose` that is sent when a control is disposed. Example: If a resource is created, that is shared among some controls, and needs to be disposed after using, one might tend to create code like (see a full example snippet at the bottom)
```java
final Font font = new Font(composite.getDisplay(), "Courier New", 11, 0);
composite.addListener(SWT.Dispose, event -> font.dispose());

final Label label = new Label(composite, SWT.NONE);
label.setFont(font);
...
```
But this is wrong, because when the shell is disposed (click [x]), the `Dispose` listener is sent at the beginning of the disposing process in `Widget.release(boolean)`, before it proceeds to child controls, **before any child control has been disposed**. In the example, the `Dispose` listener to dispose the font is called before `label` gets disposed! Hence, the label will held a disposed font for a short time. Often this is not a problem, because nothing needs that font in the mean time, but sometimes this causes hard to reproduce crashes, because, e.g. disposing a `Shell` with a `Tree` control disposes a tooltip, resulting in a `FocusIn` event to a not yet disposed child, causing a color change on a `StyledText`. The `StyledText` control can't detect that, because during the disposing process, no parent's `isDisposed` method will return `true`.

**Describe the solution you'd like**
Currently, there is no event sent **after** the control has been really disposed, including all children. Hence I suggest to introduce an `Disposed` event (mind the additional `d` at the end) and send it at the end of `Widget.release(boolean)`.

**Describe alternatives you've considered**
In the `SWT.Dispose` listener one can delay the font disposal with `event.display.asyncExec(() -> font.dispose());`, but this will run much later (maybe too late), if at all. We tend to avoid `display.asyncExec()` to get the events into the right order, because this easily becomes a nightmare.

**Full example snippet of problematic code**
```java
import org.eclipse.swt.*;
import org.eclipse.swt.custom.*;
import org.eclipse.swt.graphics.*;
import org.eclipse.swt.layout.*;
import org.eclipse.swt.widgets.*;

public class ParentDisposeBeforeChildren {

	public static void main(String[] args) {
		final Display display = new Display();
		display.addListener(SWT.Skin, event -> {
			if (event.type == SWT.Skin) {
				if (event.widget instanceof StyledText st) {
					updateSelectionBackground(st, st.isFocusControl());
				}
			}
		});

[...]

PR #914, opened 2023-11-28
SWT: update documentation for Dispose event

tmssngr:
This is related to issue #913, but #913 should be kept open.


SyntevoAlex:
Documentation only change, looks good to me.

Issue #697, opened 2023-06-09
`SIGSEGV` in `swt_fixed_resize`


torokati44:
**Describe the bug**

Segmentation fault.

```
#0  0x00007ffadd922a22 in _gtk_css_shadows_value_get_extents (border=0x7ffaeedf9940, shadows=0x0)
    at ../../../gtk/gtkcssshadowsvalue.c:327
#1  gtk_css_gadget_allocate (gadget=0x7ffaea33ca70, allocation=0x7ffaeedf9a10, baseline=-1, out_clip=0x7ffaeedf9a20)
    at ../../../gtk/gtkcssgadget.c:792
#2  0x00007ffadda79ee0 in gtk_scrolled_window_size_allocate (widget=0x7ffaea33bb10, allocation=0x7ffaeedf9b00)
    at ../../../gtk/gtkscrolledwindow.c:3420
#3  0x00007ffaddb42830 in gtk_widget_size_allocate_with_baseline
    (widget=0x7ffaea33bb10, allocation=<optimized out>, baseline=<optimized out>) at ../../../gtk/gtkwidget.c:6179
#4  0x00007ffabbe5221d in  ()
    at <redacted>/org.eclipse.osgi/340/0/.cp/libswt-pi3-gtk-4958r2.so
#5  0x00007ffaddb42830 in gtk_widget_size_allocate_with_baseline
    (widget=0x7ffaea33b0f0, allocation=<optimized out>, baseline=<optimized out>) at ../../../gtk/gtkwidget.c:6179
#6  0x00007ffabbe527d2 in swt_fixed_resize ()
    at <redacted>/org.eclipse.osgi/340/0/.cp/libswt-pi3-gtk-4958r2.so
#7  0x00007ffabbe4e76c in Java_org_eclipse_swt_internal_gtk_OS_swt_1fixed_1resize ()
    at <redacted>/org.eclipse.osgi/340/0/.cp/libswt-pi3-gtk-4958r2.so
#8  0x00007ffae220cb48 in  ()
#9  0x00007ffadf207988 in  ()
#10 0x00007ffaed0da576 in type_node_conforms_to_U
    (support_prerequisites=0, support_interfaces=1, iface_node=<optimized out>, node=<optimized out>) at ../../../gobject/gtype.c:3618
#11 g_type_check_instance_is_a (type_instance=type_instance@entry=0x7ffadf1d76f8, iface_type=<optimized out>)
[...]

PR #915, opened 2023-11-28
[GTK] Don't use deprecated gtk_style_context_invalidate()

SyntevoAlex:
GTK sources say:
  Deprecated: 3.12: Style contexts are invalidated automatically.

https://www.eclipse.org/swt/faq.php#gtkstartup says:
  Eclipse/SWT 4.16.x: GTK+ 3.20.0 and its dependencies
  Eclipse/SWT 4.15.x: GTK+ 3.14.0 and its dependencies

This means that it's been at least 3 years since SWT requires GTK 3.20. Sounds safe to stop using the deprecated function now.

Hopefully fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/697 I couldn't reproduce it in SWT snippet, so I didn't test this with SWT. But when I tested in native snippet, removing `gtk_style_context_invalidate()` changed things a bit and crash no longer occurred.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT PI/gtk/library/gtk3.c
+++ b/bundles/org.eclipse.swt/Eclipse SWT PI/gtk/library/gtk3.c
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2022 IBM Corporation and others.
+ * Copyright (c) 2000, 2023 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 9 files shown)


SyntevoAlex:
Unfortunately it didn't fix #697.
Still, let's have this code cleanup.
In master now.

iloveeclipse:
Note: there were no one successful gerrit build for this PR:

image

This (among other issues) caused errors in building SWT GTK related code, see https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1608.

I assume version updates are missing for GTK fragments, but I'm not sure. I'm looking into this.

iloveeclipse:
> https://download.eclipse.org/eclipse/downloads/drops4/I20231129-0600/buildlogs/reporeports/reports/versionChecks.html shows that both swt host and fragments have version updates.

OK, the diff shows that bytecode change is related to this PR. The new version is what we actually should use. So if the versions are bumped, why do we have the problem?

iloveeclipse:
The API errors are reported because classes & sources changed in 
- eclipse.platform.swt.binaries/bundles/org.eclipse.swt.gtk.linux.aarch64/pom.xml
- eclipse.platform.swt.binaries/bundles/org.eclipse.swt.gtk.linux.ppc64le/pom.xml
- eclipse.platform.swt.binaries/bundles/org.eclipse.swt.gtk.linux.x86_64/pom.xml

Here the excerpt:

```
3.  eclipse.platform.swt.binaries/bundles/org.eclipse.swt.gtk.linux.x86_64/pom.xml
   no-classifier: different
      org/eclipse/swt/internal/gtk3/GTK3.class: different
      org/eclipse/swt/widgets/Control.class: different
      org/eclipse/swt/widgets/Display.class: different
      org/eclipse/swt/widgets/Text.class: different
      org/eclipse/swt/widgets/ToolTip.class: different
      org/eclipse/swt/widgets/Tracker.class: different
      org/eclipse/swt/widgets/Widget.class: different
   classifier-sources: different
      library/gtk3.c: different
      library/gtk3_stats.h: different
      org/eclipse/swt/internal/gtk3/GTK3.java: different
      org/eclipse/swt/widgets/Control.java: different
      org/eclipse/swt/widgets/Display.java: different
      org/eclipse/swt/widgets/Text.java: different
      org/eclipse/swt/widgets/ToolTip.java: different
      org/eclipse/swt/widgets/Tracker.java: different
      org/eclipse/swt/widgets/Widget.java: different
    The main artifact has been replaced with the baseline version.
    The following attached artifacts have been replaced with the baseline version: [sources]
```

iloveeclipse:
I will now touch the 3 fragments, if that is the way to go.

PR #917, opened 2023-11-29
Use pattern matching in ExceptionStash

akurtakov:


akurtakov:
Hopefully this helps with https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1608

PR #919, opened 2023-11-29
Update Jenkinsfile to increase timeouts

iloveeclipse:
See https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1608

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -41,8 +41,8 @@ pipeline {
 	options {
 		skipDefaultCheckout() // Specialiced checkout is performed below
 		timestamps()
-		timeout(time: 180, unit: 'MINUTES')
-		buildDiscarder(logRotator(numToKeepStr:'5'))
+		timeout(time: 240, unit: 'MINUTES')
+		buildDiscarder(logRotator(numToKeepStr:'10'))
 		disableConcurrentBuilds(abortPrevious: true)
 	}
 	agent {
@@ -77,7 +77,7 @@ pipeline {
 				}
 				dir ('eclipse.platform.swt.binaries') {
 					checkout([$class: 'GitSCM', branches: [[name: 'refs/heads/master']],
-						extensions: [[$class: 'CloneOption', timeout: 120, noTags: false ]],
+						extensions: [[$class: 'CloneOption', timeout: 180, noTags: false ]],
 						userRemoteConfigs: [[url: 'https://github.com/eclipse-platform/eclipse.platform.swt.binaries.git']]
 					])
 					sh 'git remote set-url --push origin git@github.com:eclipse-platform/eclipse.platform.swt.binaries.git'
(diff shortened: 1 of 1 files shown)


iloveeclipse:
@akurtakov : WDYT? Would avoid the issue we had with overcrouded jenkins.

akurtakov:
Let me start a new build forcing natives rebuild. IIRC it was less than an hour.

akurtakov:
@HannesWell I started a build with forceNative https://ci.eclipse.org/releng/job/eclipse.platform.swt/job/master/308/parameters/ but it wasn't respected and detected as if natives rebuild is not needed. Would you please look into that?

iloveeclipse:
@akurtakov : the Jenkinsfile checks the last commit and if it is from `HEAD commit author: eclipse-releng-bot@eclipse.org` it will not build natives. So the build not doing anything is expected.

akurtakov:
> @akurtakov : the Jenkinsfile checks the last commit and if it is from `HEAD commit author: eclipse-releng-bot@eclipse.org` it will not build natives. So the build not doing anything is expected.

But it shouldn't do that when forceNativeBuilds is passed.

HannesWell:
> > @akurtakov : the Jenkinsfile checks the last commit and if it is from `HEAD commit author: eclipse-releng-bot@eclipse.org` it will not build natives. So the build not doing anything is expected.
> 
> But it shouldn't do that when forceNativeBuilds is passed.

That's both right. This is a a not so nice work-around to prevent endless build loops and I would be more than happy if that could be handled more elegant (probably that has to be discussed with the infra team).
However I will fix the Jenkins-Pipeline tonight. 
For now you can _Replay_ the Pipeline from the Jenkins UI and un-comment the problematic check.

But what makes me wonder, shouldn't be the commit of this PR be the last one?

akurtakov:
In the rare cases when we hit such natives mismatch  it would be nice if we can retrigger a full rebuild. Another case is if/when we update infrastructure (new Mac/Win/Linux builders with new native libraries.)

basilevs:
Are there any recommendations to PR authors given the situation? Should I retrigger failed builds somehow, or should I refrain to prevent overload?

akurtakov:
> Are there any recommendations to PR authors given the situation? Should I retrigger failed builds somehow, or should I refrain to prevent overload?

I can give only generic one - use the resources with caution. E.g. I see 5 open PRs in swt repo from you . Are they all needed? I strongly recommend for related work to have 1 (2 at best) PRs  so there is less pressure on builders.

With the above said - opening new stream is always tough period as everyone tries to push what he worked on during RCs freeze period combine that with releng changes and mass changes that are supposed to go in for M1 and infra has tough period . Usually after the first week things calm down.

akurtakov:
I'm closing this one, if the build didn't complete in 3h waiting one more is not helpful.

Issue #922, opened 2023-11-30
[macOS] Possible NPE in Menu._setVisible


tmssngr:
**Describe the bug**
SmartGit users report following exception:
```
java.lang.NullPointerException
	at org.eclipse.swt.widgets.Menu._setVisible(SourceFile:271)
	at org.eclipse.swt.widgets.Display.runPopups(SourceFile:4490)
	at org.eclipse.swt.widgets.Display.readAndDispatch(SourceFile:3982)
...
```
It is following code:
```java
		// Hold on to window in case it is disposed while the popup is open.
		window.retain();
		// NSMenu processes events on its own while the popup is open.
		display.sendPreExternalEventDispatchEvent ();
		NSEvent nsEvent = NSEvent.otherEventWithType(OS.NSApplicationDefined, location, 0, 0.0, window.windowNumber(), window.graphicsContext(), (short)0, 0, 0);
		NSMenu.popUpContextMenu(nsMenu, nsEvent, shell.view);
		display.sendPostExternalEventDispatchEvent (); // << here NPE
		window.release();
```

**To Reproduce**
Not reproducible reliably.

**Expected behavior**
No exception.

**Screenshots**

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [ ] Windows
- - [ ] Linux
- - [x] macOS

mickaelistria:
Closed with #923

PR #923, opened 2023-11-30
 #922: Possible NPE in Menu._setVisible

tmssngr:


SyntevoAlex:
@lshanmug could you please review?
I refrain from that due to conflict of interests (Thomas is my boss)

merks (approved):
This changing is just making it look like it does for Windows:

image

So that seems perfectly fine.

PR #925, opened 2023-11-30
[Build] Build commits from releng-bot if natives build is enforced

HannesWell:
And add a not to the description of 'forceNativeBuilds' parameter that if enabled and 'skipCommit' is not set, the then build binaries will be pushed to the master branch.

FYI @akurtakov and @iloveeclipse

Requested in https://github.com/eclipse-platform/eclipse.platform.swt/pull/919#issuecomment-1833269324.


akurtakov (approved):
Looks good.

PR #931, opened 2023-12-05
Revert "[GTK] Don't use deprecated gtk_style_context_invalidate()"

mickaelistria:
Reverts eclipse-platform/eclipse.platform.swt#915
This has caused some test failures, see https://github.com/eclipse-platform/eclipse.platform.ui/issues/1351
Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/1351


akurtakov:
I'll start an I-build now in order to get tests in better shape.

PR #936, opened 2023-12-12
#935 Adjusted computation of horizontal toolbar size to account for inner control

### LINKED ISSUE #935: Combobox in Toolbar cut off
**Describe the bug**
In a horizontal Toolbar the lower border of a Combobox is cut off.

**To Reproduce**
 - Open ControlExample 
 - Open Toolbar Tab
 - Enable ComboChild

**Expected behavior**
Combobox is displayed correctly

**Screenshots**
cut off combobox

**Environment:**

- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 

 --
3. JRE/JDK version
 Java 17
 
**Version since**
3.124 (probably way before)

**Fix will be contributed soon**

mh2207yatta:
Fixes #935

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/ToolBar.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/ToolBar.java
@@ -251,6 +251,12 @@ void clearSizeCache(boolean changed) {
 			OS.SendMessage (handle, OS.TB_GETITEMRECT, count - 1, rect);
 			width = Math.max (width, rect.right);
 			height = Math.max (height, rect.bottom);
+			// If a Separator has a control, its height might exceed the height of the Items
+			for (ToolItem item: items) {
+				if (item != null && item.getControl() != null) {
+					height = Math.max (height, item.getControl().getBoundsInPixels().height);
+				}
+			}
 		}
 		OS.SetWindowPos (handle, 0, 0, 0, oldWidth, oldHeight, flags);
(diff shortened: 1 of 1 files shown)


HeikoKlare:
I addition to my validation of this PR that I have posted in the according issue #935, I have one further finding:

Each view can have its own toolbar. This toolbar is placed (right aligned) in the same row as the view tab headers in case there is sufficient space. Otherwise, the toolbar is placed below the tab headers.
However, with this change, the toolbar will _always_ be placed below the tab header.

This is how it looked before:
image

And this is how it looks now:
image

iloveeclipse (changes requested):
See last comment, rendering is broken.

ShahzaibIbrahim:
> I addition to my validation of this PR that I have posted in the according issue #935, I have one further finding:
> 
> Each view can have its own toolbar. This toolbar is placed (right aligned) in the same row as the view tab headers in case there is sufficient space. Otherwise, the toolbar is placed below the tab headers. However, with this change, the toolbar will _always_ be placed below the tab header.
> 
> This is how it looked before: image
> 
> And this is how it looks now: image

@HeikoKlare is there a reproducible snippet for this? I didn't totally understand this layouting issue.

HeikoKlare:
> is there a reproducible snippet for this? I didn't totally understand this layouting issue.

I have to admit that I have no snippet I can share to test because the screenshot is from our tools which has a mechnism to embed the editor toolbar into the stack (just like it's done for views). What I would do to create a snippet/example is to add a combo box to the toolbar of a view. With the state of this PR, the view toolbar should then always be placed below instead of next to the tab headers.

PR #944, opened 2023-12-16
Bump org.eclipse.swt version due to new API

HeikoKlare:
Follow-up for #785.

### DIFF
--- a/bundles/org.eclipse.swt/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.swt/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-Name: %pluginName
 Bundle-Vendor: %providerName
 Bundle-SymbolicName: org.eclipse.swt; singleton:=true
-Bundle-Version: 3.124.300.qualifier
+Bundle-Version: 3.125.0.qualifier
 Bundle-ManifestVersion: 2
 Bundle-Localization: plugin
 DynamicImport-Package: org.eclipse.swt.accessibility2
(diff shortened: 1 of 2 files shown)


laeubi:
@HeikoKlare I suspect this now causes the build to fail see:

https://ci.eclipse.org/releng/job/Builds/job/I-build-4.31/45/console

It might be that a bump in swt also needs bumps in the fragments here:

https://github.com/eclipse-platform/eclipse.platform.swt.binaries/blob/master/bundles/org.eclipse.swt.gtk.linux.x86_64/META-INF/MANIFEST.MF

@akurtakov can you probably advice here? I can check if I can make Tycho more patient here (as the fragment still matches the range) but wonder what was the general contract here in the past.

laeubi:
I have now further checked this and it is indeed the problem [with this p2.inf](https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/bundles/org.eclipse.swt/META-INF/p2.inf) the fragment is pulled but needs the strict version:

https://github.com/eclipse-platform/eclipse.platform.swt/blob/f19db1416d82929923ffc3dd37ed1b03e522d4f8/bundles/org.eclipse.swt/META-INF/p2.inf#L10

so now this no longer matches what leads to the problem it not found anymore for the consumers (jface), so version bumps are required for all fragments as well!

iloveeclipse:
> might be that a bump in swt also needs bumps in the fragments here

Yes.

HeikoKlare:
@laeubi Thanks for fixing the fragments versions. It would be nice to have an automated check for this implicit constraint. Maybe it will become easier to have that once the fragments are part of this repo: eclipse-platform/eclipse.platform.swt.binaries#2

Issue #945, opened 2023-12-18
Unexpected padding in CTabFolders without images


HeikoKlare:
**Describe the bug**
The headers in CTabFolders always have a large padding if they either do not have a usable image assigned or if they are configured to now show icons. The latter case was the intention of #785, but as a side effect it also applies to when simply no image is provided (like in all existing CTabFolders without images). 

The reason is that the extra padding is also applied when the image of a tab is null or disposed:
https://github.com/eclipse-platform/eclipse.platform.swt/blob/f19db1416d82929923ffc3dd37ed1b03e522d4f8/bundles/org.eclipse.swt/Eclipse%20SWT%20Custom%20Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java#L396-L398

**To Reproduce**
The behavior can be seen when opening the CTabFolder in the `CustomControlsExample`.

Before:
![image](https://github.com/eclipse-platform/eclipse.platform.swt/assets/755472/c20b06b9-393a-4c23-bde4-d55ac42cec00)

After:
![image](https://github.com/eclipse-platform/eclipse.platform.swt/assets/755472/812eba47-f8f2-4a6f-9fb4-0d224d007727)

**Expected behavior**
CTabFolders without images shall be shown without extra padding like before #785.

**Screenshots**
The issue can, e.g., be seen in the plug-in settings. Originally posted in a comment of the original PR: https://github.com/eclipse-platform/eclipse.platform.swt/pull/785#issuecomment-1858819916

Before:
![image](https://github.com/eclipse-platform/eclipse.platform.swt/assets/755472/fa344695-036b-49b3-8f22-ebc490fdca9b)

After:
![image](https://github.com/eclipse-platform/eclipse.platform.swt/assets/755472/5fc9f591-0b4b-43ef-9d0d-d33d945385f1)

**Version since**
Since #785 / SWT 3.125.

HeikoKlare:
@shubhamWaghmare-sap do you know why you decided to not only apply the padding when the setting says to do so (i.e., `showSelectedImage == true`) but also when no image is attached to the TabItem at all? If there is no specific reason, simply removing the application of the extra padding in case there is no image assigned (or if it is diposed) should solve the issue.

At least for me, changing the above posted snipped as follows solves the issue for me:
```java
if (((state & SWT.SELECTED) != 0 && !parent.showSelectedImage) 
[...]

PR #947, opened 2023-12-19
Apply large padding only when image for tab item shall not be shown #945

HeikoKlare:
Currently, all headers in CTabFolders without an image have a large text padding. This applies to both the case where no image is assigned to a tab item and the case where images shall not be shown as the properties showSelectedImage and showUnselectedImage are both false. While the latter is intended behavior to switch between image-based tab folders and image-less, padding-based tab folders (see #785), the former unintentionally breaks with the existing UI experience for CTabFolders that do not use images.

This change makes the identification of whether large text padding shall be applied explicit by factoring it out into a method (which may be replaced by a different implementation later on). It reduces the cases in which the large text padding is applied to the intended case (showSelectedImage = showUnselectedImage = false) and restores the behavior for all existing use cases.

Before this change:
image

After this change (like before #785):
image

The "before" state can now only be achieved when the tab folder has showSelectedImage=showUnselectedImage=false (which will never be the case for existing CTabFolders as showSelectedImage=true by default).

Contributes to https://github.com/eclipse-platform/eclipse.platform.swt/issues/945

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
@@ -362,8 +362,7 @@ public class CTabFolderRenderer {
 
 					if (parent.showClose || item.showClose) {
 						if ((state & SWT.SELECTED) != 0 || parent.showUnselectedClose) {
-							if (((state & SWT.SELECTED) != 0 && parent.showSelectedImage)
-									|| ((state & SWT.SELECTED) == 0 && parent.showUnselectedImage)) {
+							if (!applyLargeTextPadding(parent)) {
 								if (width > 0) width += INTERNAL_SPACING;
 							} else {
 								if (width > 0) width -= INTERNAL_SPACING;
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
```diff
@@ -362,8 +362,7 @@ protected Point computeSize (int part, int state, GC gc, int wHint, int hHint) {
 
 					if (parent.showClose || item.showClose) {
 						if ((state & SWT.SELECTED) != 0 || parent.showUnselectedClose) {
-							if (((state & SWT.SELECTED) != 0 && parent.showSelectedImage)
-									|| ((state & SWT.SELECTED) == 0 && parent.showUnselectedImage)) {
+							if (!applyLargeTextPadding(parent)) {
```
sratz, line 365: this part deals with the regular padding (`INTERNAL_SPACING`). is correct to use the `applyLargeTextPadding()` condition here?

HeikoKlare: Yes, this case distinction was introduced in #785. Before, the `INTERNAL_SPACING` was always added here. It is necessary in order to have some space between the text and the close icon. With large text padding, however, there already is sufficient space between the text and the close icon. Adding further spacing makes the text looks uncentered. In addition, the close icon already has some padding equal to `INTERNAL_SPACING`, which makes it look even more uncentered. Thus, we decided to not apply the additional spacing but subtract it in that case (see https://github.com/eclipse-platform/eclipse.platform.swt/pull/785/files#r1422065827).

sratz: Ah, right. There are so many dimensions to this whitespace logic, it's hard to keep track of all of them.

sratz (approved):

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
HeikoKlare: True. A refactoring could help to separate the different functionalities and improve comprehensibility, but that would be quite some effort and may be prone to errors...

Thanks for the review, @sratz!

PR #958, opened 2024-01-04
Use high resolution images with proper scaling whenever available

HeikoKlare:
In the current implementation, images with some other scaling than 100% are only loaded when an image is available that exactly fits to the desired zoom level. In all other cases, the 100% image is used as a fallback.

With this change, the directly fitting or next larger available image of 150% or 200% is used. Only if none of them exists, the 100% image is used as a fallback. This improves the quality of images when a display scaling of more than 100% is used and no image is provided for exactly that scaling. To simplify the changes, DPIUtil methods returning multiple values were changed to not use parameters to pass returns values and some unnecessary case distinctions have been cleaned up.

The change is applied to the Windows and Linux/GTK implementations. Mac only requires handling for 100%/<s>200$</s>200% and already has that implemented.

## How to reproduce
This is easy to test with, e.g., the eclipse.sdk product on a screen with 125% or 175% scaling. Do not use 150% or 200% as for those factors exactly matching images/icons exist in many cases.

Ensure that SWT autoscaling is set to "quarter" or "exact" (and not to default "integer200"):
`-Dswt.autoScale=exact`
`-Dswt.autoScale=quarter`

### Before change at 175% scaling
image

### After change at 175% scaling
image

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
@@ -14,6 +14,8 @@
  *******************************************************************************/
 package org.eclipse.swt.internal;
 
+import java.util.function.*;
+
 import org.eclipse.swt.*;
 import org.eclipse.swt.graphics.*;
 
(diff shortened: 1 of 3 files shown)


HannesWell:
> Mac only requires handling for 100%/200$ and already has that implemented.

I know Apple is expensive and charges some parts separately, but charging for extra for scaling seems new. 😮

But besides the typo this sounds like a great enhancement for the scaling issues, although I cannot say much about the details.

HeikoKlare:
I have also verified the change on Linux: it works as expected, but actually it already worked as expected before. I found that GTK seems to only report 100% or 200% scaling and handles every other scale value in between by internal rescaling from a 200% rendering, such that the proposed change has currently no effect (and is not even necessary) on Linux.
It would be great if there is anyone with experience in Linux/GTK who can confirm, so that I can drop the additional image scaling functionality for Linux and make this PR a pure Windows improvement. @akurtakov maybe you can give some advice here?

HeikoKlare:
Thanks for the assessment, @akurtakov! Then we can keep the code as is for Linux as well.

HeikoKlare:
I have verified again for Windows:
1. With default scaling method (swtAutoscale=integer200) everything stays the same, i.e., 100% images are used (unscaled) for 100%-150% scaling and 200% images are used (unscaled) for 175% and above.
2. With "proper" scaling (swtAutoscale=quarter/exact) the sizes of the images stay the same (so no regressions w.r.t. layout) but the next-higher-resolution one is used and downscaled, so that images become sharper.

PR #956, opened 2024-01-02
Move SWT native binaries in this repository stored in Git LFS

HannesWell:
Configure this Git-repository to store the SWT native binaries, i.e. all *.dll, *.so and *jnilib files, using Git's Large File Storage (LFS) and move the platform-specific SWT fragments from the
eclipse.platform.swt.binaries repository into the binaries folder in this repository.

The platform specific projects are (minimally adjusted) copies of the corresponding projects from
https://github.com/eclipse-platform/eclipse.platform.swt.binaries/tree/9cdcee51a79a3f72e10e074b87326f80d3d6ada2

Only leave out the 'forceQualifier.txt' since the main bundle only drives the qualifier.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt.binaries/issues/2

Pending tasks before this can be submitted:
- [ ] Prepare a PR to clear the master branch of https://github.com/eclipse-platform/eclipse.platform.swt.binaries and add a note that it is effectively archived, similar to for example https://github.com/eclipse-platform/eclipse.platform.runtime. Don't mark the entire repository as archived to allow subsequent maintenance work.
- [x] Prepare a PR for https://github.com/eclipse-platform/eclipse.platform.releng.aggregator to remove the `eclipse.platform.swt.binaries` sub-module and adjust the Jenkins pipeline to perform an lfs-pull for sub-modules.
  - https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1700
- [x] Prepare adjustments of nightly I-Builds to perform lfs-pulls for sub-modules too.
  - Probably done with https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1700#issuecomment-1880011656
- [x] Prepare adjustments of documentation about how to set up an SWT workspace in https://www.eclipse.org/swt/git.php.
Remove the task to clone swt.binaries (or at least mention that is not necessary with the latest master so that those that work on older states can still set up a SWT workspace). Mention the requirement for the jgit.lfs plugin.
  - https://github.com/eclipse-platform/www.eclipse.org-swt/pull/11
- [x] Report that LFS usage is ready for submission and ask the EF-IT team to purchase extra LFS storage and bandwith as already discussed in https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/3076

PR #959, opened 2024-01-05
[Build] Remove unused ant tasks and properties

HannesWell:
Assume that natives are only build in the Jenkins pipeline or in a Maven build using `-Dnative=${os}.${ws}.${arch}`.

The "SWT Classpath Builder.launch" is not necessary anymore since Oomph does the same and the .classpath file suitable for the running platform can also be easily selected and renamed manually.
Or @iloveeclipse do you use that launch config or are aware of a wider user base?


iloveeclipse:
> do you use that launch config or are aware of a wider user base?

I don't think I've intentionally used that at all.
@trancexpress : did you used it?

As far as I can remember, we use shell scripts for natives build. I believe that must be also documented somewhere in the SWT repo.

HannesWell:
I just rechecked it and the `SWT Classpath Builder` launch is mentioned nowhere.

> As far as I can remember, we use shell scripts for natives build. I believe that must be also documented somewhere in the SWT repo.

Do you mean we as Eclipse SWT project or are you referring to your company?
Because currently it is a convoluted mixture of shell-scripts and ant-tasks that are called by maven or groovy directly in the Jenkins to build the natives for all platforms.
But you can also build the binaries only for the running platform via a simple maven build from the swt.binaries repo (or this repo once #956 is completed: `mvn clean verify -Dnative=${ws}.${os}.${arch}`
For windows for example `mvn clean verify -Dnative=win32.win32.x86_64`

iloveeclipse:
@HannesWell : may be I misunderstood your question. I don't use Oomph, so if the deleted launch was there to create proper platform dependent .classpath files, I still need that.

HannesWell:
> @HannesWell : may be I misunderstood your question. I don't use Oomph, so if the deleted launch was there to create proper platform dependent .classpath files, I still need that.

OK, I just checked again what the launch config is actually doing and it is not used to to rename any of the prepared .classpath_${os} files to .classpath but to build org.eclipse.swt_<version>.jar locally using the (also removed) `build.xml` file.
But I checked the result and the content of the jar is very incomplete, e.g. there are no class files and the metadata are also outdated. Since these data are obviously not maintained, I suggest to continue with the removal. One can still build a jar locally with maven.
Sorry for the confusion.

HannesWell:
I'm going to submit this now so that in case of unexpected I-build problems I have time tomorrow to fix them.

PR #969, opened 2024-01-11
[Mac x86_64] Fix text clipping in Trees

Phillipus:
Now that the Mac x86_64 launcher and the latest versions of Mac x86_64 Java are built using a later SDK we see text truncated in Tree widgets as we did in Mac aarch64.

The fix is to remove the check and apply the extra gap for both x86_64 and aarch64.

See https://github.com/eclipse-equinox/equinox/issues/473


Phillipus:
It turns out that the Mac x86_64 launcher has been built using SDK 13 since Eclipse 4.30 so the tree clipping is present in that version. We should try to get this fixed for 4.31.

Phillipus:
I'll give this a couple more days if anyone would like to review. Note that this is on latest x86_64 Mac version.

PR #973, opened 2024-01-12
[Build] Collect native sources in Jenkins pipeline with Java script

HannesWell:
in order to replace ANT tasks with JavaScript (which require Java-11) in Jenkins pipeline.
Stash native sources not-zipped to save the zip and unzip steps.

Additionally move all remaining ANT tasks to run a native build in a local Maven for the running platform completely to the
maven-antrun-plugin configuration.

Furthermore move declaration of tools to the common Jenkins pipeline section to make them usable in all stages.
Since https://github.com/eclipse-platform/eclipse.platform.swt/pull/633 the JDK on the native-build-agent's PATH is not used anymore when building the native binaries and thus it is not required anymore to keep the PATH untouched.

Part of
- https://github.com/eclipse-platform/eclipse.platform.swt/issues/513
- https://github.com/eclipse-platform/eclipse.platform.swt/issues/626

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -64,6 +64,10 @@ pipeline {
 	agent {
 		label 'centos-latest'
 	}
+	tools {
+		jdk 'openjdk-jdk17-latest'
+		maven 'apache-maven-latest'
+	}
 	environment {
 		MAVEN_OPTS = "-Xmx4G"
 		PR_VALIDATION_BUILD = "true"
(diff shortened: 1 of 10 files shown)


HannesWell:
To compile the native binaries locally with maven it is sufficient to run
```
mvn clean generate-resources -Dnative=${ws}.${os}.${arch} -DskipTests
```
If you also want to build the jars you have to run the build up to a later phase like `verify`.

If that takes to long for you, you can just directly run two commands to collect the native sources and compile them as it is implemented in the `build-native-binaries`, without using ant.

chirontt:
Thanks, but it should be the `process-resources` phase to invoke, not `generate-resources` like your above command. So for my WoA .dll production, the command series should be:
```
mvn clean process-resources -Dnative=win32.win32.aarch64 -DskipTests
```
The caveat: there must already exist the `swtBuildQualifier.txt` file in `bundles\org.eclipse.swt\target` directory. Not sure how this file was generated there in the first place, or whether it could be skipped...

And my ultimate goal is to cross-compile from a Windows x64 host to generate the binaries for the Arm64 target, so the command becomes:
```
mvn clean process-resources -Dnative=win32.win32.aarch64 -DskipTests -Dtargets="x64_arm64 all install" -DSWT_JAVA_HOME=\path\to\the\arm64-jdk
```
Thanks for your work to simplify these SWT build scripts.

HannesWell:
> Thanks, but it should be the `process-resources` phase to invoke, not `generate-resources` like your above command. 

That's right, sorry for the confusion.

So for my WoA .dll production, the command series should be (running on a WoA box):
> 
> ```
> cd binaries\org.eclipse.swt.win32.win32.aarch64
> mvn clean process-resources -Dnative=win32.win32.aarch64 -DskipTests
> ```
> 
> The caveat: there must already exist the `swtBuildQualifier.txt` file in `bundles\org.eclipse.swt\target` directory. Not sure how this file was generated there in the first place, or whether it could be skipped...
> 

If you just run the mvn command from the swt repository root, then that file is generated (that's done when the `org.eclipse.swt` project is built).

If you want to reduce your build time you could use the `-pl` and `-am` option of Maven:
`clean process-resources -pl org.eclipse.swt:org.eclipse.swt.win32.win32.x86_64 -am -Dnative=win32.win32.x86_64 -DskipTests`


> And my ultimate goal is to cross-compile from a Windows x64 host to generate the binaries for the Arm64 target, so the command becomes:
> 
> ```
> mvn clean process-resources -Dnative=win32.win32.aarch64 -DskipTests -Dtargets="x64_arm64 all install" -DSWT_JAVA_HOME=\path\to\the\arm64-jdk
> ```

That should work yes. In order to avoid the need to specify the targets on the CLI you could simply create a org.eclipse.swt.win32.win32.aarch64 fragment as copy from the existing windows fragment basically replacing all x86_64 by aarch64 plus the adjustments of the make files (but you probably know them better than I).
But then you probably still need the `SWT_JAVA_HOME` argument.

HannesWell:
I thought about building only the natives again and found an even faster solution.
@chirontt please see https://github.com/eclipse-platform/eclipse.platform.swt/pull/991 if you are interested.

PR #974, opened 2024-01-12
Fix erroneous <p> tags in JavaDoc of BidiUtil, Callback and Compatibility

HannesWell:
Required to enable https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1533.


HannesWell:
According to this answer SO answer https://stackoverflow.com/a/5681796/14542697, `<ul>` sections must not reside inside `<p>`.

PR #975, opened 2024-01-13
[Build] Move tools declaration to common Jenkins pipeline section

HannesWell:
to make it usable in all stages.
Since https://github.com/eclipse-platform/eclipse.platform.swt/pull/633 the JDK on the native-build-agent's PATH is not used anymore to build the natives and therefore it is not required anymore to keep the PATH untouched.


HannesWell:
Result looks good, but I'm about to incorporate this into https://github.com/eclipse-platform/eclipse.platform.swt/pull/973, where it is actually needed, thus closing.

PR #981, opened 2024-01-16
Remove P2Directory elements moved to Platform container from SWT-setup

HannesWell:
With https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1719 the required jgit.lfs Plug-in is now installed for all projects in the `Platform` container. Keeping it in the SWT setup is just a duplication.


HannesWell:
The setup is not tested by the verification build, thus submitting it now.

PR #982, opened 2024-01-16
[Build] Use JDK in JAVA_HOME to run the Java source scripts

HannesWell:
This increases the probability that the same jvm running the maven build is used to run the java source scripts.
Maven's launch script by default uses the java executable referenced in $JAVA_HOME/bin first and only then searches the PATH environment variable.
Using the java exe at JAVA_HOME also simplifies to control which jvm is actually used. For example when launching a Maven build from the Eclipse IDE JAVA_HOME is set to the jvm running that build by Eclipse automatically, but PATH is not modified.

This would also have fixed https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1724.

### DIFF
--- a/binaries/pom.xml
+++ b/binaries/pom.xml
@@ -120,7 +120,7 @@
 								</goals>
 								<configuration>
 									<target>
-										<exec executable="java" dir="${swtMainProject}" failonerror="true">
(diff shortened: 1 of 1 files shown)


HannesWell:
The test failures are very likely unrelated and I have created https://github.com/eclipse-platform/eclipse.platform.swt/issues/983 to resolve it.

PR #984, opened 2024-01-17
Use the ECJ generated compiler logs to parse compiler problems

laeubi:
Reading a structured XML is more reliable than parsing the logfiles.

For SWT we need to make sure that the logs are actually produced and located at the given location, so this is just a first try.


HannesWell:
> For SWT we need to make sure that the logs are actually produced and located at the given location, so this is just a first try.


The SWT build has no magic regarding complication anymore so actually everything should be at the default location, but I don't know where that is.

At least Jenkins says no `ECJ` warnings:
https://ci.eclipse.org/releng/job/eclipse.platform.swt/job/PR-984/1/

But for the current master there are no `Java Compiler` warnings either.
So this looks fine.

For confirmation you could temporarily add another commit where you introduce a warning and check if that is displayed as expected.

PR #988, opened 2024-01-19
[Build] Further streamlining and simplifications of SWT build pipeline

HannesWell:
Increment the Library revision also for enforced native binary re-builds. Besides allowing simplifications of the build, this ensures
the git-based qualifier of the org.eclipse.swt and consequently its native fragments (which use the same qualifier) is updated in that case and a subsequently 'manually' enforced qualifier update is not necessary. Furthermore letter-suffixes for revision git tags are not necessary too.

And remove superfluous steps.


HannesWell:
> Use the org.eclipse.cbi.maven.plugins:eclipse-winsigner-plugin Maven plug-in to sign the built binaries instead of bash-scripts.

I now dropped this idea again, since it took much longer to sign the binaries even with a single goal execution maven build (about a minute) compared to just curl-ing the binaries to sign as was already done (a few seconds).

PR #991, opened 2024-01-19
Add  Maven launch-config to build native binaries of running platform

HannesWell:
In the SWT maven build local building of the native binaries can be activated by setting the system property `native` to `${ws}.${os}.${arch}`.
Consequently the full command for a complete Maven build of SWT that also builds the natives is (with the SWT-repo root as working directory):
`mvn clean verify -Dnative=${ws}.${os}.${arch}`

This full build takes a few minutes and if one skips the test execution, by appending `-DskipTests`, the complete build terminates earlier.

The fastest way to build the native binaries for a platform with Maven is to only run the execution of the `maven-antrun-plugin` that contains the local natives build. On the CLI this would be (with the SWT-repo root as working directory):
`mvn clean antrun:run@build-native-binaries -f binaries/org.eclipse.swt.${ws}.${os}.${arch} -Dnative=${ws}.${os}.${arch}`

This PR adds a Maven launch configuration (requires [Eclipse m2e](https://github.com/eclipse-m2e/m2e-core)) to the `org.eclipse.swt` project to perform such native binaries only build for the currently running platform. The launch-config uses the latter Maven command and infers the `${ws}`, `${os}` and `${arch}` placeholder using the launch variables `${target.ws}`, `${target.os}` and `${target.arch}`. With that one can just build the natives without further configuration after the necessary prerequisites are installed as described in the corresponding articles in https://www.eclipse.org/swt/docs.php.

### DIFF
--- a/Build-SWT-native-binaries-for-running-platform.launch
+++ b/Build-SWT-native-binaries-for-running-platform.launch
@@ -1,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<launchConfiguration type="org.eclipse.m2e.Maven2LaunchConfigurationType">
+    <intAttribute key="M2_COLORS" value="0"/>
+    <booleanAttribute key="M2_DEBUG_OUTPUT" value="false"/>
+    <stringAttribute key="M2_GOALS" value="clean antrun:run@build-native-binaries"/>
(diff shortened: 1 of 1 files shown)


lathapatil:
For local testing what are the replacement commands or steps using maven build
[6th step here](https://www.eclipse.org/swt/jnigen.php) may be outdated by now

PR #918, opened 2023-11-29
[GTK] Optimize TreeItem constructor by removing ID #882

basilevs:
The ID system tracking association of every TreeItem with GTK modelentry had been using GTK.gtk_tree_store_set() to persist a key in GTK model. This operation has linear execution  time over model size.

IDs were replaced with tree paths and strong references, which, if implemented right, have logarithmic execution time on balanced trees and constant execution time on wide trees.

Performance improvement is proven by running org.eclipse.swt.tests.junit.performance.Test_org_eclipse_swt_widgets_Tree.jfaceReveal()

In test jfaceReveal[Shape: STAR, virtual: true]:
10_000 elements:  141_061_117ns -> 4_369_889ns
100_000 elements: 10_761_449_641ns -> 181_898_611ns (-98%)

PR #996, opened 2024-01-22
[GTK] Add missing spaces for consistent style

basilevs:
Add missing space after identifier, but before an opening parenthesis or bracket. The fix is done by replacing all occurrences of `\w\(` and `\w\[` regular expressions. Only GTK Tree and TreeItem files are touched.

This fixes some of style defects mentioned in  https://github.com/eclipse-platform/eclipse.platform.swt/pull/903

Rationale: The existing style inconsistencies make it harder to keep consistent style in  https://github.com/eclipse-platform/eclipse.platform.swt/pull/918

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Tree.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Tree.java
@@ -41,21 +41,21 @@ import org.eclipse.swt.internal.gtk4.*;
  * </p><p>
  * Here is an example of using a <code>Tree</code> with style <code>VIRTUAL</code>:</p>
  * <pre><code>
- *  final Tree tree = new Tree(parent, SWT.VIRTUAL | SWT.BORDER);
- *  tree.setItemCount(20);
- *  tree.addListener(SWT.SetData, new Listener() {
- *      public void handleEvent(Event event) {
+ *  final Tree tree = new Tree (parent, SWT.VIRTUAL | SWT.BORDER);
+ *  tree.setItemCount (20);
+ *  tree.addListener (SWT.SetData, new Listener () {
+ *      public void handleEvent (Event event) {
  *          TreeItem item = (TreeItem)event.item;
- *          TreeItem parentItem = item.getParentItem();
+ *          TreeItem parentItem = item.getParentItem ();
  *          String text = null;
  *          if (parentItem == null) {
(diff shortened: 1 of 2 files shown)


iloveeclipse:
1) I'm not sure this style is default one in Eclipse and SWT, I believe the opposite is the case.
2) Changing style by hand is error prone, next edit will not consider **this** special style and we have disconnect again - therefore one should make sure formatter settings are committed for project and that should be followed.
3) Not sure why only two files are changed.
4) Git blame will show now your commit for almost every line, because it is not just leading/trailing whitespece change.

In sum : I dislike this change and wish I wouldn't see it 1:(

basilevs:
> 1. I'm not sure this style is default one in Eclipse and SWT, I believe the opposite is the case.

@SyntevoAlex requests this style in #903
 
>     2. Changing style by hand is error prone, next edit will not consider **this** special style and we have disconnect again - therefore one should make sure formatter settings are committed for project and that should be followed.

This style is so exotic, that I don't think it can be expressed in Eclipse Formatter terms (and there is no formatter file for SWT to begin with).

>   3. Not sure why only two files are changed.

The method (to update the style) I use is not universal and I need only these to files to be consistent to proceed with #918

>     4. Git blame will show now your commit for almost every line, because it is not just leading/trailing whitespece change.

Sure, just like any other style fix.
 
> In sum : I dislike this change and wish I wouldn't see it 1:(

Same here.

iloveeclipse:
> Same here.

So let's close this PR.

basilevs:
> > Same here.
> 
> So let's close this PR.

Should I then format the files with default Eclipse Formatter? I still need to somehow apply my change for #918, which have to be consistent with something.

basilevs:
> I haven't checked your patch, sorry. I personally prefer to use "most convenient" coding style for changed code, that a) consistent and b) matches common understanding about format used in the project in question. Also I usually try to separate "cleanup" from "functional" changes.

It does not matter what my patch is. The most convenient style is not consistent with current SWT codebase. And this discussion shows that there is no common understanding. This MR was a separate cleanup change.

SyntevoAlex:
I agree that code style in SWT is somewhat exotic.

I would suggest to not touch what already exists, unless you're rewriting most of function's code.
If you're changing something slightly, keep to existing style, whatever it is.

SyntevoAlex:
In new code, you could pick from any of the styles that are widely represented in SWT.
For example, putting spaces between braces are seen in both ways in SWT:
`void foo ()`
`void foo()`

PR #997, opened 2024-01-23
Remove Chromium left overs

jukzi:


jukzi:
https://github.com/eclipse-platform/eclipse.platform.swt/pull/878#pullrequestreview-1724502824

akurtakov (approved):
Thanks @jukzi ! This looks good. Sorry that I totally forgot about it.

akurtakov:
It would be nice if you can cleanup the Cocoa implementation too https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/bundles/org.eclipse.swt/Eclipse%20SWT%20Browser/cocoa/org/eclipse/swt/browser/BrowserFactory.java .

jukzi:
@akurtakov what about https://github.com/eclipse-platform/www.eclipse.org-swt/blob/61bcc86539be17d94bb8c9f6142e3f6411d1c938/faq.php#L914 ... ? there are also many references to chromium

akurtakov:
> @akurtakov what about https://github.com/eclipse-platform/www.eclipse.org-swt/blob/61bcc86539be17d94bb8c9f6142e3f6411d1c938/faq.php#L914 ... ? there are also many references to chromium

This needs editing too. One may say some info could be kept for users of old versions but I believe that telling people how to make use of ancient full of CVEs chromium is disservice to everyone so IMO it's best to be entirely removed.

Issue #715, opened 2023-06-29
[Windows Hi-DPI] Program icons have black background


Phillipus:
Windows 10/11
Eclipse 4.28
Windows hi-dpi scaling at 200%.

Some program icons in Package Explorer and Project Explorer have a black background:

![explorer](https://github.com/eclipse-platform/eclipse.platform.swt/assets/600504/81df6d55-90bf-4329-85cd-2de55ec73365)

Sometimes, the icons render correctly.

I can't reproduce this with a stand-alone SWT snippet so it seems to suggest that it's caused by Eclipse doing things in a certain order, or something else affects it in the UI.

See also https://github.com/eclipse-platform/eclipse.platform.swt/issues/185

deepika-u:
Hi @Phillipus,
For me this looks to be similar issue reported in pr #409. Can you please also review that issue if that might give you some clue towards this problem.
Thanks

Phillipus:
> Hi @Phillipus, For me this looks to be similar issue reported in pr #409. Can you please review that issue if that might give you some clue towards this problem. Thanks

Hi @deepika-u I tested running a child Eclipse with the SWT PR (and the PR in Platform UI) but, unfortunately, the program icons still have a black background.

I think it's related to https://github.com/eclipse-platform/eclipse.platform.swt/issues/185

Phillipus:
> I think it's related to #185

Actually, it's not a recent regression. The problem exists in Eclipse 4.8 (June 2018)

merks:
Yes, this is quite an old problem.  BTW, I think these come from org.eclipse.swt.program.Program.getPrograms() and from there via org.eclipse.swt.program.Program.getImageData() which works it's way tout of org.eclipse.ui.internal.misc.ExternalProgramImageDescriptor.getImageData() eventually.   Unfortunately org.eclipse.swt.tests.win32.snippets.SWTIssue185.createImages(Display, String...) does not not reproduce the problem.

Phillipus:
> ....does not not reproduce the problem.

And I can't reproduce it in a stand-alone SWT snippet. And the infuriating thing is (in terms of debugging) is that *sometimes* the icons render OK!

Phillipus:
In the `org.eclipse.swt.program.Program` class:


```java
public ImageData getImageData () {
	if (extension != null) {
		SHFILEINFO shfi = new SHFILEINFO ();
[...]

PR #999, opened 2024-01-24
[Win32] Remove buggy icon retrieval from file extensions #715

HeikoKlare:
This is an alternative solution to #998 for #715. In my opinion, this solution is better than the one in #998, in particular because
* it reduces code (complexity) rather than increasing it,
* it only affects the `Program` class and not the central `ImageData` class, thus reducing the risk of side effects, and
* it aligns the Windows implementation of the `Program` class with the ones for the other OSes.

Under certain conditions, program icons loaded on Windows via GDI+ have empty mask data, even though the original icon has proper mask data. As a result, these icons are printed with a black instead of a transparent background. This only happens when retrieving icons via the program's file extension but works properly when using the (also available) full icon path.

With this change, the shortcut functionality retrieving an icon from a program's file extension is removed, so that the icon is always retrieved from its full path. This fixes the issue with missing mask data and reduces code.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/715

Before:
image

After:
image


Phillipus:
Just to confirm I have tested this and so +1

HeikoKlare:
Further (discussion on) verification of this PR happened in #715.

deepika-u:
Verified this issue and now icons show up as below
image

Icon are very clear now, checked on below build
Eclipse SDK
Version: 2024-03 (4.31)
Build id: I20240219-1800
Java vendor: Eclipse Adoptium
Java runtime version: 17.0.6+10
Java version: 17.0.6

PR #1007, opened 2024-01-26
[Build] Unify make targets for native binaries build

HannesWell:

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT PI/win32/library/make_win32.mak
+++ b/bundles/org.eclipse.swt/Eclipse SWT PI/win32/library/make_win32.mak
@@ -113,7 +113,7 @@ swt_awt.res:
 swt_wgl.res:
 	rc $(RCFLAGS) -DSWT_ORG_FILENAME=\"$(WGL_LIB)\" -r -fo swt_wgl.res swt_wgl.rc
 
-install:
+install: all
 	copy *.dll "$(OUTPUT_DIR)"
 
 clean:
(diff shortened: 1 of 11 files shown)


bundles/org.eclipse.swt/Eclipse SWT PI/win32/library/make_win32.mak
```diff
@@ -113,7 +113,7 @@ swt_awt.res:
 swt_wgl.res:
 	rc $(RCFLAGS) -DSWT_ORG_FILENAME=\"$(WGL_LIB)\" -r -fo swt_wgl.res swt_wgl.rc
 
-install:
+install: all
```
HannesWell, line 116: For other platforms the make files look similar and I had the impression that this means when invoking the `install` make-target call `all` before.
But it looks like this does not work on Windows.
@akurtakov I believe you are also familiar with C and maybe with make? Can you help on this one?

akurtakov: It should work as you described it. I have no idea why this doesn't happen on Windows. I haven't done anything with native compilers on Windows .

HannesWell: The problem was that in the `build.bat` native-binary built script for windows called the `shift` command to strip of the first passed argument and then passed all remaining arguments to the make file:
`nmake -f make_win32.mak %1 %2 %3 %4 %5 %6 %7 %8 %9`

But since the first argument always had to be `x86_64`, but then the value was never read I removed it with this change.
But this had the consequence, that when calling `build.bat install clean` the make file `make_win32.mak` was called without the first argument, consequently only invoked the `clean` target.

PR #1008, opened 2024-01-28
Link to java sources into native fragments to compile them in fragments

HannesWell:
Previously within the IDE workspace, the SWT java-sources for a platform have been compiled within the org.eclipse.swt project using a .classpath file specially crafted for each supported window-system/platform. Therefore when checking out the SWT sources in the IDE one had to rename the '.classpath_${ws}' template for the desired platform to '.classpath' to use it locally.
With this change this renaming is not necessary anymore.

Furthermore it allows PDE API-Tools to remove the built-in special handling of the org.eclipse.swt setup because now the API baseline check within the IDE can happen simply in the fragment as it exists in the baseline.


HannesWell:
This change would be much simpler, if we could link the (direct) content of the selected folder directly together using a link-filter.
Currently it seems that only the link target is made available.

iloveeclipse:
Not sure about that, right now because of API tooling. It shows for the new fragment project ~ 2400 new warnings of type "X illegally references field Y". May be because of API filters missing that were set on original project.

Also now I see API error on SWT: 
`The minor version should be the same for version 3.125.0, since no new APIs have been added since version 3.124.200	MANIFEST.MF	/org.eclipse.swt/META-INF	Version Numbering Problem`

HannesWell:
On my Windows Computer all fragments are error free if I import them in the IDE and it looks like they all compile the sources.

Have you tried it with deleting and re-creating the baseline? This usually helps for me if it went crazy.

> Also now I see API error on SWT:
> `The minor version should be the same for version 3.125.0, since no new APIs have been added since version 3.124.200 MANIFEST.MF /org.eclipse.swt/META-INF Version Numbering Problem`

I got that one as well, and added a filter for it.

But I don't understand the build-failure. I only see warnings, no errors. Have to investigate that later.

HannesWell:
But besides this API-Tooling problems, do you see general problems with this approach?

iloveeclipse:
Assuming API tooling would work, it could work :)
I also haven't checked how debugging would work while debugging IDE from IDE.

Haven't tried, but it looks like one could also open and compile & run API tools on all fragment projects in one workspace - this was not possible before. Would be useful for PR's that add API in fragments.

HannesWell:
> Assuming API tooling would work, it could work :) 

Great. :)
The build now succeeds. It failed because the org.eclipse.swt plugin exported the package `org.eclipse.swt.internal` with `swt.tools.spies` as friend:
`org.eclipse.swt.internal;x-friends:="org.eclipse.ui,org.eclipse.swt.tools.spies"`
But all the fragments don't do that. I now just added a suppression of the `restriction` warnings for the `Sleak` class.

>I also haven't checked how debugging would work while debugging IDE from IDE.

I just tried it and at least debugging an IDE launched from my workspace seems to work well as usually, just as if the java sources were really located in the fragments.

> Haven't tried, but it looks like one could also open and compile & run API tools on all fragment projects in one workspace - this was not possible before. Would be useful for PR's that add API in fragments.

It is possible, at least on my Windows computer to import the fragments for all platforms and they all compile well.
I also added an API error locally (removed a method) and got the expected API error, but only for the windows fragment. I haven't investigated it yet why this is the case. Maybe it is because the `Eclipse-PlatformFilter` lets the resolution of the fragment fail or maybe it is because the resource effectively exists multiple times in the workspace.

HannesWell:
> Maybe it is because the `Eclipse-PlatformFilter` lets the resolution of the fragment fail or maybe it is because the resource effectively exists multiple times in the workspace.

One thing that was of course missing (at least with the Oomph setup) is that the fragments of other platforms are not added to the baseline. I now have adjusted the setup to `includeAllPlatforms`. With that only (at least after I had listed all fragments explicitly before, but I hope it just works without that), the platform-specific fragments were in the baseline.
But an error was still only at the windows fragment for me.

HannesWell:
> Not sure about that, right now because of API tooling. It shows for the new fragment project ~ 2400 new warnings of type "X illegally references field Y". May be because of API filters missing that were set on original project.

@iloveeclipse I see about 2400 API warnings with such message in my workspace. Could it be that you have set up the API-tools to be more strict?
I have provisioned my workspace using Oomph, but I see no special preference for that so it probably is the default.

But looking at `org.eclipse.swt\.settings\org.eclipse.pde.api.tools.prefs` I see a lot of preferences set to ERROR, so we probably want to replicate that for the fragments with this.

But for org.eclipse.swt we could probably just disable the baseline checks if the bundle has no java classes.

HannesWell:
> But looking at `org.eclipse.swt\.settings\org.eclipse.pde.api.tools.prefs` I see a lot of preferences set to ERROR, so we probably want to replicate that for the fragments with this.

I have now adapted the preferences shared by all fragments to the one of `o.e.swt` and now get those warnings as errors as expected.
For `o.e.swt` I have now disabled API-tools. Since the bundle is empty I think thats the better solution than adding hundes/thounds of lines of API filters.

And mentioning filters, only for the o.e.swt.win32.win32.x86_64 fragment I had to add about 3000 lines of API filters in order to silence all errors and I wonder if it is really sensible to make API problems errors just to suppress them with filters.
Why not just ignore the problem in the first place?
Especially for problems like `Referencing a type, field or method tagged '@noreference'` that can never be an actual problem since SWT does not have any dependencies so it can only be a reference to an internal class.
With that the vast majority of errors is gone.
In this case all the `Restrictions` problems don't make sense as errors:
grafik

And in general I think PDE-APITools should not have issues with internal usage/references/overrides/... of internal classes. But maybe this is a side-effect of the special handling of SWT (https://github.com/eclipse-pde/eclipse.pde/issues/1084).

HannesWell:
For me this PR is ready and as far as I can test it (on Windows) everything works well.
Everybody interested, please review and try out this change.
If no objections are raised I plan to submit this tonight.

iloveeclipse:
> For `o.e.swt` I have now disabled API-tools.

But we still have API checks enabled & errors reported for fragments in IDE & command line build?

Could you also please update PR description to explain **what** the change does (it currently explains only how bad it was and how good it will be, but doesn't explain the actual change). This would simplify understanding of the changes.

@SyntevoAlex : would be nice if you could review this one too, because you seem to work with SWT on a completely different setup.
@Phillipus : could you please check this PR too?

Phillipus:
Well, this makes life easier. :-)

My use case when testing latest SWT on Mac is to test against Mac aarch64 (Apple Silicon) and x86_64 (Intel) targets. I can do this by switching the `Architecture` from `aarch64` to `x86_64` on the `Environment` tab of my `.target` file, and changing the target JDK to match as well. I have both the `org.eclipse.swt.cocoa.macosx.aarch64` and the `org.eclipse.swt.cocoa.macosx.x86_64` binary projects open in my workspace.

So I tested the above scenario and it works. Looks good to me, many thanks!

HeikoKlare (approved):
This is a great improvement and simplification, thank you! To me, the new structure makes more sense, as it is easier to set up and as it reflects the final artifact structure already during development time.

I went through the changes and tried them in both the IDE and a Tycho build (on Windows), and I did not find any issues. API errors are still reported in the IDE for the fragments (at least for the Windows fragment that I have checked). One thing I thought about is whether there could be some better place for the OS-specific classpath files than in the `binaries` parent folder, since this folder is usually not present in your workspace, which makes interaction with these files a little more "complicated". But since I do not consider this a relevant issue, I did not think further about where to better place them and it is fine more me as is.

So in short: I appreciate these changes; go from my side.

Phillipus (approved):
Adding this +1 comment here so that the my name has a tick bit in the list of reviewers. :-)

HannesWell:
Thank you all for testing.
I just pushed an update to remove the `CLASSPATH_ATTR_LIBRARY_PATH_ENTRY` classpath entry also for GTK/Linux. For Windows/Mac it was already done before (oversaw it for Linux). Additionally I also updated the Readme-files and removed the section about the `.classpath` renaming.

> Could you also please update PR description to explain **what** the change does (it currently explains only how bad it was and how good it will be, but doesn't explain the actual change). This would simplify understanding of the changes.

Did that in the update as well.


> One thing I thought about is whether there could be some better place for the OS-specific classpath files than in the `binaries` parent folder, since this folder is usually not present in your workspace, which makes interaction with these files a little more "complicated". But since I do not consider this a relevant issue, I did not think further about where to better place them and it is fine more me as is.

It is correct that the folder is often not present, but you can still modified the files through the linked folders. The changes will then be made in the original file and all other native fragments will see them to. This way consistent preferences are ensured.
But in general I think the preferences don't change often.
With https://github.com/eclipse-platform/eclipse.platform/issues/89 this could be done in a more standard way.


> PS
> Does this PR help to report API issues on SWT in Jenkins in any way, or do we need another ticket for that?

Not yet. Enabling API checks in the CI is done in https://github.com/eclipse-platform/eclipse.platform.swt/pull/1011.
But to answer your other question (just for completes since the others said it too), in the IDE it works at least for the fragment of the current platform. Maybe we find out later how to enable it for all platforms.

Phillipus:
Thanks for your work on this!

binaries/.classpath_win32
```diff
@@ -4,11 +4,7 @@
 	<classpathentry kind="src" path="Eclipse SWT/win32"/>
 	<classpathentry kind="src" path="Eclipse SWT/common"/>
 	<classpathentry kind="src" path="Eclipse SWT PI/common"/>
-	<classpathentry kind="src" path="Eclipse SWT PI/win32">
-		<attributes>
-			<attribute value="org.eclipse.swt.win32.win32.x86_64" name="org.eclipse.jdt.launching.CLASSPATH_ATTR_LIBRARY_PATH_ENTRY"/>
-		</attributes>
```
HeikoKlare, line 10: @HannesWell I just found that because of removing this attribute it is not possible to start a standalone SWT snippet from the IDE anymore. With this attribute, the `java.library.path` refers to the fragment containing the SWT dlls, which is missing without it. Was there a specific reason why you removed the attribute? Or can we reinsert it or provide some other way to run SWT standalone snippets easily?

Phillipus: Same on Mac.

Phillipus: @HannesWell @HeikoKlare  If the attribute is re-instated then on Mac there needs to be two `.classpath_cocoa` files - one for `aarch64` and one for `x86_64`.

HeikoKlare: Yes, the change would need to be done in every fragment's classpath file.

HannesWell: Indeed. When implementing this change I only tested it with an Eclipse application and for an OSGi runtime this was not necessary, so I assumed it was a left-over from old times.
But testing it now with a standalone SWT snippet I also see it fails, sorry for the disturbance.

But at the same time this was a welcome simplification since it allowed to re-use the same .classpath file for all macos platforms. At the `.classpath` file for Linux/GTK only referenced the x86_64 arch project so it will probably not have worked for development on all platforms.
To make it work we would probably have provide individual .classpath files for each fragment.
I have opened https://github.com/eclipse-platform/eclipse.platform.swt/pull/1031 to investigate if we can use a variable for that, but it currently seems not to work.

SyntevoAlex:
> @SyntevoAlex : would be nice if you could review this one too, because you seem to work with SWT on a completely different setup.

Indeed my setup is different, to an extent where I don't use any of the modified files. I use hand-crafted IDEA project to work with SWT.

elsazac:
I was working on an old branch of mine. It looks like I can't work on the branch now due to the change. Master branch works fine but I can see a lot of errors in the workspace right now with the old branch. I attempted rebasing, but it doesn't work.  Does anyone else have this problem? Any advice on how to make the branch functional? 

These are the errors I am getting in the workspace with old branch:

image

Phillipus:
> I attempted rebasing, but it doesn't work. 

Which branch / PR is it? What doesn't work?

If you can rebase it, that should solve problems.

PR #1009, opened 2024-01-28
Move Build-native-binaries launch to o.e.swt

HannesWell:
The repository root is not necessarily imported in each workspace and using -Dtycho.mode=maven supresses the TP resolution and thus speeds up the native-binaries only build.


laeubi (changes requested):
Please don't use `-Dtycho.mode=maven` it is deprecated and will be removed:
- https://github.com/eclipse-tycho/tycho/issues/676

If you don't need Tycho then simply do not use Tycho in all other cases dependency resolution is performed only if a mojo requests it.

HannesWell:
> If you don't need Tycho then simply do not use Tycho in all other cases dependency resolution is performed only if a mojo requests it.

Reverted that part, but looking at the build looks it seems that at least some level of dependency resolution happens although only the `clean` goal and `antrun:run` are invoked. When only invoking clean, then the resolution does not happen.

laeubi:
> but looking at the build looks it seems that at least some level of dependency resolution happens although only the `clean` goal and `antrun:run` are invoked

https://maven.apache.org/plugins/maven-antrun-plugin/run-mojo.html

Attributes:

- Requires a Maven project to be executed.
- **Requires dependency resolution** of artifacts in `scope: test`.
- The goal is thread-safe and supports parallel builds.

HannesWell:
> - Requires dependency resolution of artifacts in scope: test.

Indeed. But for this specific case it would be convenient to suppres dependeny resolution in general, but as far as I know this is not possible with Maven.

laeubi:
If you want to run ant without maven dependencies why not running it as ant ;-)

laeubi:
Thinking more about it if the project really does not need any dependency you probably can just configure it to use an empty target platform.

HannesWell:
> If you want to run ant without maven dependencies why not running it as ant ;-)

The advantage of running the maven goal is that it is also used in the full build and can just be re-used and also reads the environment and acts accordingly.

> Thinking more about it if the project really does not need any dependency you probably can just configure it to use an empty target platform.

That could work. But in the end this launch-config is intended as simple way for those that build the natives only occasionally.
If one wants to speed up this 20~30s build further one can call the build script for the desired platform with the necessary environment parameters. So I would not further complicate this, if it can't be handled within the launch config.

PR #1010, opened 2024-01-28
[Build] Run Maven build in parallel and only up until the verify phase

HannesWell:
All the fragments can build in parallel so this should speed-up the build.

### DIFF
--- a/tests/org.eclipse.swt.tests.win32/META-INF/MANIFEST.MF
+++ b/tests/org.eclipse.swt.tests.win32/META-INF/MANIFEST.MF
@@ -6,7 +6,8 @@ Bundle-Version: 3.108.100.qualifier
 Bundle-Vendor: Eclipse.org
 Bundle-Localization: plugin
 Require-Bundle: org.junit;bundle-version="4.12.0",
- org.eclipse.swt
+ org.eclipse.swt,
+ org.eclipse.swt.tests
 Eclipse-BundleShape: dir
 Bundle-RequiredExecutionEnvironment: JavaSE-17
 Automatic-Module-Name: org.eclipse.swt.tests.win32
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -63,6 +63,7 @@ jobs:
       with:
        run: >-
         mvn --batch-mode -V -U
+        --threads 1C
         -DforkCount=1
         -Dnative=${{ matrix.config.native }}
         -Dcompare-version-with-baselines.skip=true
@@ -71,7 +72,7 @@ jobs:
         --fail-at-end
         -DskipNativeTests=false
         -DfailIfNoTests=false
-        clean install
+        clean verify
     - name: Performance tests
       if: contains(github.event.pull_request.labels.*.name, 'performance')
       uses: coactions/setup-xvfb@b6b4fcfb9f5a895edadc3bc76318fae0ac17c8b3 # v1.0.1
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -299,7 +299,7 @@ pipeline {
 					dir('eclipse.platform.swt') {
 						sh '''
 							mvn clean verify \
-								--batch-mode -DforkCount=0 \
+								--batch-mode --threads 1C -DforkCount=0 \
 								-Dcompare-version-with-baselines.skip=false -Dmaven.compiler.failOnWarning=true \
 								-Dorg.eclipse.swt.tests.junit.disable.test_isLocal=true \
 								-Dmaven.test.failure.ignore=true -Dmaven.test.error.ignore=true
--- a/tests/org.eclipse.swt.tests.cocoa/META-INF/MANIFEST.MF
+++ b/tests/org.eclipse.swt.tests.cocoa/META-INF/MANIFEST.MF
@@ -6,7 +6,8 @@ Bundle-Version: 3.108.100.qualifier
 Bundle-Vendor: Eclipse.org
 Bundle-Localization: plugin
 Require-Bundle: org.junit;bundle-version="4.12.0",
- org.eclipse.swt
+ org.eclipse.swt,
+ org.eclipse.swt.tests
 Eclipse-BundleShape: dir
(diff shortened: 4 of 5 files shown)


HannesWell:
Looks like the test become flaky if run in parallel. Does anybody has experience with that?
Furthermore the runtime gain does not look as big as expected on the first runs.

iloveeclipse:
I assume parallel tests do not run in different VNC clients? I don't know how you run **UI** tests in parallel, probably on different displays on Linux but no idea about Windows. Usually UI tests are very sensible, as all of UI inputs/outputs are "singletons" (clipboard, keyboard, mouse, application focus), so UI tests running in parallel on same machine most likely will be affected by that.

HeikoKlare:
You would need to ensure that all bundles executing UI tests (namely `org.eclipse.swt.tests` and `org.eclipse.swt.tests.win32`) are not executed in parallel, for the reasons that @iloveeclipse mentioned. The tests in each bundle are still executed sequentially. The Linux and Macs builds are not be affected, because the OS-specific Mac tests are empty and the ones for GTK do not test any UI.

Maybe you know that @laeubi also did some according experiments for Platform UI (see https://github.com/eclipse-platform/eclipse.platform.ui/issues/921).

laeubi:
It all depends on the test, e.g. if you have tests that use a shared resource (like clipboard) one might want to write them in a way that it synchronize on that resource. e.g Junit5 has [@ResourceLock](https://junit.org/junit5/docs/5.3.0/api/org/junit/jupiter/api/parallel/ResourceLock.html). 

For different bundles surefire has a special settings [reactorConcurrencyLevel](https://tycho.eclipseprojects.io/doc/latest/tycho-surefire-plugin/test-mojo.html#reactorConcurrencyLevel) that limits the number of parallel executing reactor projects that use tests.

HannesWell:
Thank you for all the input.

> For different bundles surefire has a special settings [reactorConcurrencyLevel](https://tycho.eclipseprojects.io/doc/latest/tycho-surefire-plugin/test-mojo.html#reactorConcurrencyLevel) that limits the number of parallel executing reactor projects that use tests.

Unfortunately `org.eclipse.swt.tests` uses Maven-surefire to run the tests. So even if we would configure the `org.eclipse.swt.tests.win32` accordingly it would not prevent concurrent test execution.

One hack might be to have an explicit dependency of `org.eclipse.swt.tests.win32` to `org.eclipse.swt.tests`.
This way they cannot be built together.

HannesWell:
> One hack might be to have an explicit dependency of `org.eclipse.swt.tests.win32` to `org.eclipse.swt.tests`.
> This way they cannot be built together.

Looks like this worked (at least in the last pass).
Its not nice, but I would not be concerned to much about such such dependency for a test. I just wonder if we should add it to the other platform-specific tests just in case they also use resources in the future?

laeubi:
> One hack might be to have an explicit dependency of org.eclipse.swt.tests.win32 to org.eclipse.swt.tests.
This way they cannot be built together.

You mean in parallel ;-)

HannesWell:
> I just wonder if we should add it to the other platform-specific tests just in case they also use resources in the future?

Added it now for all.

tests/org.eclipse.swt.tests.win32/META-INF/MANIFEST.MF
```diff
@@ -6,7 +6,8 @@ Bundle-Version: 3.108.100.qualifier
 Bundle-Vendor: Eclipse.org
 Bundle-Localization: plugin
 Require-Bundle: org.junit;bundle-version="4.12.0",
- org.eclipse.swt
+ org.eclipse.swt,
+ org.eclipse.swt.tests
```
HeikoKlare, line 10: The problem with such a dependency is that it only makes sense in the context of a parallel Maven build. It does something else than what these dependency specifications are supposed to be used for, so no one except the ones involved in this PR will know that (and probably even they will forget), so there will be people trying to understand the necessity for this dependency and eventually even remove it.

I would at least suggest to add a comment in the Manifest that explains the necessity for the dependency.

laeubi: > I would at least suggest to add a comment in the Manifest that explains the necessity for the dependency.

Manifest can't have comments. Another alternative would be additional.bundles in the `build.properties` or a dependency in the `pom.xml` or a `p2.inf`

HeikoKlare: > Manifest can't have comments

Just out of interest: did that change at some point in time? Because in a former project, several years ago, I had comments in a Manifest (but I do not remember the syntax).

HannesWell: > > Manifest can't have comments
> 
> Just out of interest: did that change at some point in time? Because in a former project, several years ago, I had comments in a Manifest (but I do not remember the syntax).

It didn't change. But an OSGi framework is supposed to ignore Headers it doesn't understand. So you can have a `Comment: Some remark` header as long as all characters are permitted.

In general I agree with your concern, that's why I added an explanation in the commit message (that is then hopefully found).
But moving it to the build.properties is probably the best solution here.
Will prepare a follow-up tonight.

laeubi: A header is not a comment... manifests are meant to be processed by tools not humans that why [the specification](https://docs.oracle.com/en/java/javase/20/docs/specs/jar/jar.html#manifest-specification) is rather "odd" and its wide use makes its near to impossible to change (e.g. the stupid no more than 72 bytes (**not characters**) rule).

HeikoKlare: Thanks for clarification. Now I remember that we used a `Comment: ...` header, knowing that it was actually a misuse.

HannesWell: Created https://github.com/eclipse-platform/eclipse.platform.swt/pull/1017.

Issue #1013, opened 2024-01-29
No more handle exceptions with the Edge browser


vogella:
Using the Edge browser on Windows still results in a the no more handlers exception even after the commit https://github.com/eclipse-platform/eclipse.platform.swt/commit/ed8f82127b6e5f7b6d2526c0e027cde80b4e5eab from @HeikoKlare 

Before the commit, the error was in Edge.java:343, now it occurs in Edge.java:362.

https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/bundles/org.eclipse.swt/Eclipse%20SWT%20Browser/win32/org/eclipse/swt/browser/Edge.java#L362C1-L363C1

It seems to happen if the user starts multiple time the same application in different workspaces. I will try to provide a test case to reproduce.

vogella:
See https://github.com/eclipse-platform/eclipse.platform.swt/issues/339

HeikoKlare:
I am sorry that Edge instantiation does still not work properly. We did not have issues in productive usage recently, but we do not use it intensively either (usually only one instance). Some remarks on this:
* Even though a `no more handles" error is shown, that is probably not the reason. This error is thrown if _anything_ unexpected happens during WebView2 instantiation.
* I once tried to parameterize the browser tests to also be executed for Edge (#672). But I have not been successful yet because of some interference between the test cases and their used Edge instances: it looked as if some OS events were not processed properly (see https://github.com/eclipse-platform/eclipse.platform.swt/pull/672/files#r1202546545).

I could image that the problem you describe, @vogella, is related to the one that can be seen in the tests.
What I tried to do in my previous fix attempt was to ensure that asynchronous OS events for browser instantiation are processed properly. Taking a fresh look at the code, I would first have a look at the involved `Edge#callAndWait()` implementations again: https://github.com/eclipse-platform/eclipse.platform.swt/blob/2cdfd9705676cfcc15dc6c2cfb80057fdcbdf3fc/bundles/org.eclipse.swt/Eclipse%20SWT%20Browser/win32/org/eclipse/swt/browser/Edge.java#L212-L247
From my understanding, the OS event processing is supposed to trigger the callback setting the results (`ppv`/`phr`). However, the `callAndWait` performs a release on the callback even before doing the event processing, so maybe the problem is something pretty simple like an interface pointer being released too early.
[...]

PR #1015, opened 2024-01-30
Distinguish Edge instantiation error due to invalid thread access #1013

HeikoKlare:
Instead of always showing a "no more handles" error, with this change an error due to invalid thread access is explicitly logged as such.


vogella:
LGTM

PR #1016, opened 2024-01-30
Avoid callback release before its usage during Edge instantiation #1013

HeikoKlare:
The handle to a newly instantiated Edge browser is provided via a callback. This callback may be either be invoked synchronously when creating a WebView2 controller (happens on first creation of an Edge browser instance in an application) or asynchronously via a processed OS event (from second creation of an Edge browser instance onwards). The callback is currently released before the OS events are processed, which may cause issues when instantiating more than one Edge browser in a single application.

With this change, the callback is only released when it is not required and may not be called anymore. A test instantiating multiple (Edge) browsers in a single application is added to ensure that the asynchronous callback invocation is executed in a test.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
@@ -221,10 +221,12 @@ static int callAndWait(long[] ppv, ToIntFunction<IUnknown> callable) {
 	});
 	ppv[0] = 0;
 	phr[0] = callable.applyAsInt(completion);
-	completion.Release();
+	// "completion" callback may be called asynchronously,
+	// so keep processing next OS message that may call it
 	while (phr[0] == COM.S_OK && ppv[0] == 0) {
 		processNextOSMessage();
 	}
+	completion.Release();
 	return phr[0];
 }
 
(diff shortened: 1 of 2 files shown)


vogella (commented):
Looks really good, I have a similar test implementation pending in my workspace but could not upload it, due to a missing ssh key on this particular machine. Planning to merge tomorrow.

Thanks @HeikoKlare

HeikoKlare:
Thanks, @vogella! It don't expect the test case to be a proper reproducer for the issue in #1013, but at least it is a scenario that has not been properly tested before. I tried to provide a reproducer for #1013, but it is quite some effort as it would require the test to start different applications. I tried to implement a test case that emulates the behavior within the same application, but there is a limitation in the SWT Edge implementation due to its WebView2 environment being static and tied to the display it was instantiated with. When you try to run multiple browsers in multiple threads, you need to use multiple displays, but then WebView2 instantiation fails as the WebView 2 environment has been created with a different thread and its display.

So in case you can are able to provide a test case to reproduce #1013, that would still be great.

laeubi:
> When you try to run multiple browsers in multiple threads, you need to use multiple displays, but then WebView2 instantiation fails as the WebView 2 environment has been created with a different thread and its display.

Be aware that Windows allows multiple `Display`s (while Linux does not) so this is actually something that could happen under Windows/SWT!

HeikoKlare:
> Be aware that Windows allows multiple `Display`s (while Linux does not) so this is actually something that could happen under Windows/SWT!

Exactly, thanks for pointing out! That's why I consider it a limitation that the WebView2 environment is static. Whenever you try to create an Edge browser from another thread than the environment (and thus also in another display for which that is the UI thread), it will fail with error 0x802a000c (invalid thread access).

PR #1018, opened 2024-01-31
Support Edge instances in different displays #1013

HeikoKlare:
The current implementation of the Edge browser creates a WebView2Environment on first creation of an WebView2 / Edge browser instance. On every further creation of a WebView2 instance, this environment is used. In case the WebView2 instance is created for a different display, i.e., within a different thread, the instantiation fails as the WebView2Environment has been created in a different thread.

To support the creation of WebView2 instances in different displays, this change replaces the static WebView2Environment with one environment for every display. The WebView2 instances are created for the environment belonging to the display for which the current instantiation is requested. An according regression test is added.

A documentation for using multiple WebView2 environments can be found here: https://learn.microsoft.com/en-us/microsoft-edge/webview2/concepts/process-model?tabs=csharp#multiple-environment-objects


vogella (approved):
Thanks @HeikoKlare 

I suggest to merge this also today, so that we have test all your fixes in tomorrows I-build

HeikoKlare:
> I suggest to merge this also today, so that we have test all your fixes in tomorrows I-build

Sure, just note that "test" precisely means "manual test" because there are currently only three tests executed against the Edge browser (since I did not manage to make #672 running successfully yet).
At least I applied the commit of #672 to this PR to execute all the browser tests against the Edge browser as well, and only `test_setUrl_local()` fails (but also does without this PR).

I made slight changes after your review, in 52a73be. They concern the following:
- Store the environment an Edge instance belongs to within that instance, so that race conditions when accessing the global map on browser disposal are avoided
- Use `waitForPassCondition()` in tests instead of doing busy waiting without a timeout
- Dispose the additional displays in the test

There is one thing to consider left: the logic for clearing cookies (provided statically via `Browser.clearSessions()`) will clear the cookies of the environment for which the current thread is the UI thread. This is bit strange, as one might expect a static method to clear _all_ cookies of all environments. And if you call the method on an object rather than the class, it is even more misleading as you might have `browser1` and `browser2` with different displays and call `browser1.clearSessions()` from the UI thread of `browser2`, resulting in the coockies of _only_ `browser2` to be cleared.
In my opinion, this is rather an existing API problem (`clearSessions()` being a static method), which we cannot fix easily. But I still wanted to point out that behavior.

PR #1022, opened 2024-02-01
remove Snippet157 (java.applet.Applet deprecated for removal)

jukzi:
The type Applet has been deprecated since version 9 and marked for removal


Phillipus (commented):
Hi, it will need to also be removed from here:

https://github.com/eclipse-platform/eclipse.platform.swt/blob/3241d044e43b3a48deb65a4b964cc80c08774562/examples/org.eclipse.swt.snippets/Snippets.md?plain=1#L251

jukzi:
> Hi, it will need to also be removed from here:

thanks for the hint, i found even more references to it using workspace search.

Phillipus (approved):
Yes, those refs are everywhere. :-)

PR #1026, opened 2024-02-02
Expose cancel state of FileDialog and DirectoryDialog

HeikoKlare:
The `FileDialog` and `DirectoryDialog` implementations do currently not provide any way to identify that the dialog was canceled. The dialogs return a null string in case no file or folder was selected, but is not possible to identify whether this was intended (via a cancel action) or if there was some unexpected error while the dialog was open (such as too long file paths on Windows).

This change introduces the `openDialog()` method as an alternative to the existing `open()` method. This new method returns an optional String, which is only present if a file or directory was successfully selected. An empty Optional indicates a cancellation by the user and an actual error is realized as an SWTException. The dialogs evaluate the operating system's response code for the system's dialog control to identify the dialogs result state.

This PR is a result of the proposal by @laeubi in https://github.com/eclipse-platform/eclipse.platform.swt/pull/1005#issuecomment-1921684747. Supercedes and thus closes #1005.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/DirectoryDialog.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/DirectoryDialog.java
@@ -15,6 +15,8 @@
 package org.eclipse.swt.widgets;
 
 
+import java.util.*;
+
 import org.eclipse.swt.*;
 import org.eclipse.swt.internal.ole.win32.*;
 import org.eclipse.swt.internal.win32.*;
@@ -122,10 +124,37 @@ public String getMessage () {
  *    <li>ERROR_THREAD_INVALID_ACCESS - if not called from the thread that created the dialog</li>
  * </ul>
  */
-public String open() {
+public String open () {
+	try {
+		return openDialog().orElse(null);
+	} catch (SWTException e) {
+		if (e.code == SWT.ERROR_INVALID_RETURN_VALUE) {
+			return null;
+		}
+		throw e;
+	}
+}
+
+/**
+ * Makes the dialog visible and brings it to the front of the display.
+ * Equal to {@link DirectoryDialog#open()} but also exposes for state information like user cancellation.
+ *
+ * @return an Optional that either contains the absolute path of the selected directory
+ *         or is empty in case the dialog was canceled
+ *
+ * @exception SWTException <ul>
+ *    <li>ERROR_WIDGET_DISPOSED - if the dialog has been disposed</li>
+ *    <li>ERROR_THREAD_INVALID_ACCESS - if not called from the thread that created the dialog</li>
+ *    <li>ERROR_INVALID_RETURN_VALUE - if the dialog was not cancelled and did not return a valid path</li>
+ * </ul>
+ *
+ * @since 3.126
+ */
+public Optional<String> openDialog () {
 	this.directoryPath = null;
 
 	long [] ppv = new long [1];
+	int hr = COM.E_FAIL;
 	if (COM.CoCreateInstance(COM.CLSID_FileOpenDialog, 0, COM.CLSCTX_INPROC_SERVER, COM.IID_IFileOpenDialog, ppv) == OS.S_OK) {
 		IFileDialog fileDialog = new IFileDialog(ppv[0]);
 
@@ -162,7 +191,8 @@ public String open() {
 		Display display = parent.getDisplay();
 		long hwndOwner = parent.handle;
 		display.externalEventLoop = true;
-		if (fileDialog.Show(hwndOwner) == OS.S_OK) {
+		hr = fileDialog.Show(hwndOwner);
+		if (hr == OS.S_OK) {
 			if (fileDialog.GetResult(ppv) == OS.S_OK) {
 				IShellItem psi = new IShellItem(ppv[0]);
 				if (psi.GetDisplayName(OS.SIGDN_FILESYSPATH, ppv) == OS.S_OK) {
@@ -182,7 +212,12 @@ public String open() {
 		fileDialog.Release();
 	}
 
-	return directoryPath;
+	if (hr == COM.S_OK) {
+		return Optional.ofNullable(directoryPath);
+	} else if (hr == OS.HRESULT_FROM_WIN32(OS.ERROR_CANCELED)) {
+		return Optional.empty();
+	}
+	throw new SWTException(SWT.ERROR_INVALID_RETURN_VALUE);
 }
 
 /**
--- a/binaries/org.eclipse.swt.cocoa.macosx.aarch64/META-INF/MANIFEST.MF
+++ b/binaries/org.eclipse.swt.cocoa.macosx.aarch64/META-INF/MANIFEST.MF
@@ -3,7 +3,7 @@ Fragment-Host: org.eclipse.swt;bundle-version="[3.125.100,4.0.0)"
 Bundle-Name: %fragmentName
 Bundle-Vendor: %providerName
 Bundle-SymbolicName: org.eclipse.swt.cocoa.macosx.aarch64; singleton:=true
-Bundle-Version: 3.125.100.qualifier
+Bundle-Version: 3.126.0.qualifier
 Bundle-ManifestVersion: 2
 Bundle-Localization: fragment
 Export-Package: 
--- a/binaries/org.eclipse.swt.cocoa.macosx.x86_64/META-INF/MANIFEST.MF
+++ b/binaries/org.eclipse.swt.cocoa.macosx.x86_64/META-INF/MANIFEST.MF
@@ -3,7 +3,7 @@ Fragment-Host: org.eclipse.swt;bundle-version="[3.125.100,4.0.0)"
 Bundle-Name: %fragmentName
 Bundle-Vendor: %providerName
 Bundle-SymbolicName: org.eclipse.swt.cocoa.macosx.x86_64; singleton:=true
-Bundle-Version: 3.125.100.qualifier
+Bundle-Version: 3.126.0.qualifier
 Bundle-ManifestVersion: 2
 Bundle-Localization: fragment
 Export-Package: 
--- a/binaries/org.eclipse.swt.gtk.linux.aarch64/META-INF/MANIFEST.MF
+++ b/binaries/org.eclipse.swt.gtk.linux.aarch64/META-INF/MANIFEST.MF
@@ -3,7 +3,7 @@ Fragment-Host: org.eclipse.swt;bundle-version="[3.125.100,4.0.0)"
 Bundle-Name: %fragmentName
 Bundle-Vendor: %providerName
 Bundle-SymbolicName: org.eclipse.swt.gtk.linux.aarch64; singleton:=true
-Bundle-Version: 3.125.100.qualifier
+Bundle-Version: 3.126.0.qualifier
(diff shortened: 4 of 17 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/DirectoryDialog.java
```diff
@@ -182,7 +202,13 @@ public String open() {
...
 	}
 
-	return directoryPath;
+	if (hr == COM.S_OK) {
+		return FileDialogResult.createForSuccessfulSelection(directoryPath);
+	} else if (hr == OS.HRESULT_FROM_WIN32(OS.ERROR_CANCELED)) {
+		return FileDialogResult.createForUserCancellation();
+	} else {
+		return FileDialogResult.createForUnkownError();
+	}
```
HeikoKlare, line 211: fwiw: this is not a switch because the cancel state HRESULT is not a constant but a calculated value.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/widgets/FileDialogResult.java
```diff
@@ -0,0 +1,62 @@
+package org.eclipse.swt.widgets;
+
+/**
+ * A representation of the result of file dialogs such as the {@link FileDialog}
+ * or {@link DirectoryDialog}.
+ *
+ * @since 3.125
+ */
+public class FileDialogResult {
```
laeubi, line 9:
```suggestion
public final class FileDialogResult {
```

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/widgets/FileDialogResult.java
```diff
@@ -0,0 +1,62 @@
...
+		 */
+		SELECTED,
+		/**
+		 * The user has canceled the dialog
+		 */
+		CANCELED,
+		/**
+		 * An unspecified error occurred while the dialog was open
+		 */
+		UNKNOWN_ERROR
```
laeubi, line 22:
```suggestion
		ERROR
```

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/widgets/FileDialogResult.java
```diff
@@ -0,0 +1,62 @@
...
+
+	static FileDialogResult createForSuccessfulSelection(String path) {
+		return new FileDialogResult(path, State.SELECTED);
+	}
+
+	static FileDialogResult createForUserCancellation() {
+		return new FileDialogResult(null, State.CANCELED);
+	}
+
+	static FileDialogResult createForUnkownError() {
```
laeubi, line 58:
```suggestion
	static FileDialogResult createForError(SWTException exception) {
```

I would simply add a `SWTException` here and allow to query that, implementations can then pass additional details in the exception, it could be printed with a stacktrace to track the location of it and so on

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/widgets/FileDialogResult.java
```diff
@@ -0,0 +1,62 @@
...
+	private FileDialogResult(String path, State state) {
+		this.path = path;
+		this.state = state;
+	}
+
+	/**
+	 * {@return a string describing the absolute path of the selected file or folder
+	 * or <code>null</code> if the dialog was canceled or if there was an error}
+	 */
+	public String getSelectedPath() {
```
laeubi, line 38:
```suggestion
	public Optional<String> getPath() throws SWTException {
```

this should throw exception when passed in an error state and an empty optional if the operation was canceled, that way one can simply always call this method and don't need to fiddle around will `null` checks.

also "Selected" is superfluous here as there is no other path anyways.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/DirectoryDialog.java
```diff
@@ -123,9 +123,28 @@ public String getMessage () {
...
+ *         or a state describing that the dialog was cancelled or an error occurred
+ *
+ * @exception SWTException <ul>
+ *    <li>ERROR_WIDGET_DISPOSED - if the dialog has been disposed</li>
+ *    <li>ERROR_THREAD_INVALID_ACCESS - if not called from the thread that created the dialog</li>
+ * </ul>
+ *
+ * @since 3.125
+ */
+public FileDialogResult show() {
```
laeubi, line 143:
```suggestion
public FileDialogResult selectFile() {
```

with the above suggestion about optional one can use

`dialog.selectFile().getPath().ifPresent(path-> ... do something here if not canceled and not an error ...)`

laeubi (commented):
Just some suggestions/thoughts ...

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/widgets/FileDialogResult.java
HeikoKlare: Fixed in bff9b11

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/widgets/FileDialogResult.java
HeikoKlare: I like the idea of returning an Optional. I have implemented that together with the simplified name in bff9b11.
I don't think we should thrown an `SWTException` here, as the errors that may occur are no SWT-internal fails, but something related to semantics of the dialog, so that returning a proper path was not possible (like a too long pathname on Windows). So they are still kind of "valid results", at least in terms of the contract between caller (opening the dialog) and callee (the dialog), which is why I would not express them in terms of exceptions at all.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/DirectoryDialog.java
HeikoKlare: I now changed the name to `askSelectFile()` in bff9b11. To me `selectFile()` suggests that the caller does some selection, but it only asks for selection by the user.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/widgets/FileDialogResult.java
laeubi: The main point for me is that "error" does not mean anything, what should I show to the user? And why is not the dialog already handle this (e.g. does not allow me to select a patch that is too long?).

So this case seems really "exceptional" for me and I don't want to have the boilerplate to check each and every call if it might be of an unknown error, stead I assume when the method returns everything is normale so either:

1. I get a (valid) path
2. I get nothing (user canceled)
3. I get an Exception (something went wrong)

e.g. "NO MORE HANDLES" or a like it also not something by SWT so SWTExceptions (for me) are more a way to signal an native error one can't forsee or handle in client code (like here).

HeikoKlare:
Thanks for all the input, @laeubi! I tend to completely agree with your point of view.

In particular, I took a look at the kinds of "errors" that may happen and found that the currently used OS APIs only seem to provide "OK" and "Cancel", and maybe something else if there _really_ is a error (that may then reasonable be thrown as an SWTException). I took the according information from here:
Windows: https://learn.microsoft.com/en-us/windows/win32/api/shobjidl_core/nf-shobjidl_core-imodalwindow-show
Linux/GTK: https://docs.gtk.org/gtk3/method.NativeDialog.run.html
Mac: https://developer.apple.com/documentation/appkit/nssavepanel/1525357-runmodal
And on Windows, I manually tried to produce errors that have been delivered as such in the "old" file dialog implementation (for the result types, see [here](https://learn.microsoft.com/en-us/windows/win32/api/commdlg/nf-commdlg-commdlgextendederror)), but all these kinds of errors are now already handled by the dialog itself, so that you cannot even close the dialog pressing "Ok" with, e.g., an invalid file name. My contribution was driven by the fact that our RCP product still uses SWT with the "old" Windows file dialog implementation, where we had to explicitly handle different kinds of error cases, but this does not seem to be necessary with the current implementation anyway.

So I changed the implementation to only return an Optional result and throw an exception in other cases in 9641b01.
One drawback is that this complicates the realization of the existing `open()` method. To be API preserving, it has to distinguish between existing SWTExceptions and those potentially thrown due to an error result (i.e., not OK and not Cancel) of the dialog. So either this information has to be wrapped in a result object of some private method or it has to be extracted from the SWTException. I followed the latter approach to not bloat up the code, but it is actually a bit unclean.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/DirectoryDialog.java
```diff
@@ -182,7 +212,12 @@ public String open() {
 		fileDialog.Release();
 	}
 
-	return directoryPath;
+	if (hr == COM.S_OK) {
+		return Optional.of(directoryPath);
```
laeubi, line 216:
```suggestion
		return Optional.ofNullable(directoryPath);
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/DirectoryDialog.java
```diff
@@ -123,9 +125,36 @@ public String getMessage () {
...
+ *
+ * @exception SWTException <ul>
+ *    <li>ERROR_WIDGET_DISPOSED - if the dialog has been disposed</li>
+ *    <li>ERROR_THREAD_INVALID_ACCESS - if not called from the thread that created the dialog</li>
+ *    <li>ERROR_INVALID_RETURN_VALUE - if the dialog was not cancelled and did not return a valid path</li>
+ * </ul>
+ *
+ * @since 3.125
+ */
+public Optional<String> askSelectDirectory() {
```
laeubi, line 153:
```suggestion
public Optional<String> openDirectory() {
```

What do you think about something like this? askSelect sounds a bit strange and that way probabbly if it also starts with "open" users that are used to the old method might discover the new one.

laeubi (approved):
It looks fine for me just a bit unsure about the naming but maybe others like to give advice.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/DirectoryDialog.java
merks: Yes, I think something that starts with open so that it's sorted to be adjacent in documentation or in proposals....

HeikoKlare: Agreed. Maybe just `openDialog()`? Since the call does not open a file/directory, but a dialog (to choose one).

laeubi: I would be fine with openDialog() as well

PR #1030, opened 2024-02-05
Disable "fail on warning" in GitHub Actions and Jenkins builds

HeikoKlare:
The GitHub Actions and Jenkins build specifications currently contain the option `-Dmaven.compiler.failOnWarning=true`. Since warnings were defined to be ignored in the parent POM, this did not have any effect. The configuration to ignore warnings has been removed from the parent POM, so that the builds now fail as the code actually contains warnings.

This change removes the "fail on warning" option. Warnings are still considered via Jenkins quality gates.

See eclipse-platform/eclipse.platform.releng.aggregator#1774


HannesWell (dismissed):
Disabling failOnWarning makes sense to me, but we should then activate the quality gate, similar like it is done for example in PDE:

https://github.com/eclipse-pde/eclipse.pde/blob/d5c48d832150b036c20d61fcfcbe539ed96e85d3/Jenkinsfile#L42C6-L42C219

Although the exact settings are controversially discussed: https://github.com/eclipse-platform/eclipse.platform/pull/1046

HannesWell:
> I agree. Maybe we can have that as a direct follow-up PR? In case we need some experiments or discussion on the quality gate configuration, it will not block the failOnWarning disablement from being merged. Currently this makes all builds fail due to [eclipse-platform/eclipse.platform.releng.aggregator#1774](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1774), 

I'm also fine to do that as a follow up, if you prefer to discuss the exact settings first. But I would also be fine to apply only the weaker settings now, that only require no new warnings for the Compiler (which also includes API errors) and discuss later to make it more strict. Your call :)


> and since warnings have simply been ignored before, this option never had any effect.

Only for those warnings being ignored, not the once that were not ignored.

HeikoKlare:
> Only for those warnings being ignored, not the once that were not ignored.

True. Then removing the configuration option without enabling the quality gate would actually weaken the checks. I have added a quality gate configuration, which, from my understanding, should be rather weak. So we can see if that works fine and then apply a stricter one later on.

HannesWell (approved):
Looks good. Thank you!

The current quality-gate failure is obviously expected and will vanish for all following PRs.
So lets submit this and work on fixing the warnings. 🚀

PR #1032, opened 2024-02-05
Add back exports to the org.eclipse.swt host

tjwatson:
Some projects, like BND tools, will avoid using require-bundle at great lengths. This becomes problematic when the APIs they want to use do not version there `Export-Package` properly.

Many Eclipse projects, including SWT, do not version their API packages.  They only version their individual bundles with respect to semantic versioning.

For BND they resorted to using `Import-Package` with the `bundle-version` and `bundle-symbolic-name` of the host. This "worked" because the SWT API packages were defined in the `Export-Package` header of the host bundle. But now with PR #1008 this changed and the host bundle no longer defines the exported packages. The end result is BND tools is now broken and has many unresolvable bundles.

Until this is sorted out, this PR simply adds back the `Export-Package` header of the SWT host. The longer term solution should be to properly version SWT API packages so that consumers of the API can properly use `Import-Package` to depend on the API.

### DIFF
--- a/bundles/org.eclipse.swt/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.swt/META-INF/MANIFEST.MF
@@ -6,6 +6,22 @@ Bundle-Version: 3.125.0.qualifier
 Bundle-ManifestVersion: 2
 Bundle-Localization: plugin
 DynamicImport-Package: org.eclipse.swt.accessibility2
+Export-Package: 
+ org.eclipse.swt,
+ org.eclipse.swt.accessibility,
+ org.eclipse.swt.awt,
+ org.eclipse.swt.browser,
+ org.eclipse.swt.custom,
+ org.eclipse.swt.dnd,
+ org.eclipse.swt.events,
+ org.eclipse.swt.graphics,
+ org.eclipse.swt.internal,
+ org.eclipse.swt.internal.image;x-internal:=true,
+ org.eclipse.swt.layout,
+ org.eclipse.swt.opengl,
+ org.eclipse.swt.printing,
+ org.eclipse.swt.program,
+ org.eclipse.swt.widgets
 Eclipse-ExtensibleAPI: true
 Bundle-RequiredExecutionEnvironment: JavaSE-17
 Automatic-Module-Name: org.eclipse.swt


HannesWell (approved):
Looks good, thank you Tom.

HannesWell:
The quality-gate errors in Jenkins are somewhat expected, since it was setup newly for SWT and has to settle first.

laeubi:
> For BND they resorted to using `Import-Package` with the `bundle-version` and `bundle-symbolic-name` of the host.

So effectively they using `Require-Bundle` but just giving it another name ;-)

(yes I know that there are still slight differences)

> The longer term solution should be to properly version SWT API packages so that consumers of the API can properly use Import-Package to depend on the API.

an immediate solution would be to simply always export all packages with the bundle version.

iloveeclipse:
This change adds now 19 API errors on org.eclipse.swt.gtk.linux.x86_64, saying basically that all workspace types from `org.eclipse.swt.internal` new and should be tagged with `since 3.125` version.

Changing export to `org.eclipse.swt.internal;x-internal:=true,` fixes that. 
Note, originally it was `- org.eclipse.swt.internal;x-friends:="org.eclipse.ui,org.eclipse.swt.tools.spies",`

I'm alone who sees the problem? Or is this Linux specific?

merks:
I see this now:

image

HannesWell:
> Changing export to `org.eclipse.swt.internal;x-internal:=true,` fixes that.
> Note, originally it was `- org.eclipse.swt.internal;x-friends:="org.eclipse.ui,org.eclipse.swt.tools.spies",`

Yes adding back `org.eclipse.swt.internal` as public package was wrong, I oversaw that in the review.

Changing it to `org.eclipse.swt.internal;x-internal:=true,` should be fine. `org.eclipse.swt.tools.spies` already suppresses the restriction warnings and `org.eclipse.ui` seems not to use `swt.internal` (anymore):
https://github.com/search?q=repo%3Aeclipse-platform%2Feclipse.platform.ui%20path%3A%2F%5Ebundles%5C%2Forg%5C.eclipse%5C.ui%5C%2F%2F%20org.eclipse.swt.internal&type=code

If non one of you adds the `x-internal` directive back, I can do it this evening.

tjwatson:
> This change adds now 19 API errors on org.eclipse.swt.gtk.linux.x86_64, saying basically that all workspace types from `org.eclipse.swt.internal` new and should be tagged with `since 3.125` version.
> 
> Changing export to `org.eclipse.swt.internal;x-internal:=true,` fixes that. Note, originally it was `- org.eclipse.swt.internal;x-friends:="org.eclipse.ui,org.eclipse.swt.tools.spies",`
> 
> I'm alone who sees the problem? Or is this Linux specific?

Sorry about that.  I was told the `x-friends` was no longer needed for spies.  I didn't notice that it was still needed for `org.eclipse.ui`.  I mistakenly removed all of `x-friends`.  It should still make `org.eclipse.ui` a friend.

iloveeclipse:
Most likely this change also causes following issue: https://github.com/eclipse-platform/eclipse.platform.ui/issues/1654

laeubi:
> The end result is BND tools is now broken and has many unresolvable bundles.
> Until this is sorted out, this PR simply adds back the Export-Package header of the SWT host. 

@tjwatson can you explain what component complains here? As I'm currently looking into the resolver I'm also looking what the spec says and [there one finds](https://docs.osgi.org/specification/osgi.core/8.0.0/framework.module.html#framework.module.exportpackage):

> The Framework will automatically associate each package export definition with the following attributes:
> 
> - bundle-symbolic-name - The bundle symbolic name of the exporting bundle. **In the case of a fragment bundle**, this is the host bundle's symbolic name.
> - bundle-version - The bundle version of the exporting bundle. **In the case of a fragment bundle**, this is the host bundle's version.

For me it looks like the "complainer" here is not respecting that rule and only looks at packages from the host but not consider those contributed by fragments.

tjwatson:
I remember that part of the specification, but I don't believe it to be implementable. At one point I thought I removed that from the core specification.  No framework implements that behavior, because I believe it to be unimplementable.  And clearly there is no TCK for it.

PR #1033, opened 2024-02-05
Fix some of the new warnings

HannesWell:


bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
```diff
@@ -5277,7 +5277,6 @@ public void syncExec (Runnable runnable) {
  */
 public <T, E extends Exception> T syncCall(SwtCallable<T, E> callable) throws E {
 	Objects.nonNull(callable);
-	@SuppressWarnings("unchecked")
```
HannesWell, line 5280: Actually I'm not sure if this is really sane to be removed.

HeikoKlare: I don't think so. When I place this kind of statement somewhere in SWT code, it (correctly) gives an "unchecked" warning. I don't know about any proper way to instantiate a generically typed array in a type-safe way in Java. Since this array is only used as a means to pass some value from within the lambda outside, what about replacing the array with something that is actually supposed to pass a reference, like an `AtomicReference`? It also provides `setPlain()` and `getPlain()` to avoid the synchronization overhead.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
```diff
@@ -5277,7 +5277,6 @@ public void syncExec (Runnable runnable) {
...
 	T[] t = (T[]) new Object[1];
```
laeubi, line 5280:
```suggestion
	Object[] t = new Object[1];
```

and cat the return value as with the exception?

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
HannesWell: I noticed that `unchecked` warnings were suppressed for the entire `Display` class and therefore this suppression was flagged as unnecessary. But I removed the class-level suppression so it came back. It was only suppressed at the class-level for mac-os.

But I also applied your suggestion and used AtomicReference using set/getPlain. Thanks for that.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
HannesWell: Obsolete with applying Heiko's suggestion and using AtomicReference.

HannesWell:
Since I noticed some inconsistencies between the platform in suppressed warnings, I went one step further and resolved all suppressed-warnings where reasonable possible.
The remaining suppression are harder to fix, so I left them out.
The remaining problems are all either due to the usage of deprecated code or API leakage warnings.

HannesWell:
> Looks good. Thank you for cleaning up!

Thanks for your review. :)

PR #1038, opened 2024-02-06
Mark exported org.eclipse.swt.internal package as internal

iloveeclipse:
Fixes few API errors reported after 2ca0b37ecbeff105797d5fce72675bf306ad5282

### DIFF
--- a/bundles/org.eclipse.swt/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.swt/META-INF/MANIFEST.MF
@@ -15,7 +15,7 @@ Export-Package:
  org.eclipse.swt.dnd,
  org.eclipse.swt.events,
  org.eclipse.swt.graphics,
- org.eclipse.swt.internal,
+ org.eclipse.swt.internal;x-internal:=true,
  org.eclipse.swt.internal.image;x-internal:=true,
  org.eclipse.swt.layout,
  org.eclipse.swt.opengl,


bundles/org.eclipse.swt/META-INF/MANIFEST.MF
```diff
@@ -15,7 +15,7 @@ Export-Package:
  org.eclipse.swt.dnd,
  org.eclipse.swt.events,
  org.eclipse.swt.graphics,
- org.eclipse.swt.internal,
+ org.eclipse.swt.internal;x-internal:=true,
```
tjwatson, line 18: I think we still need `; x-friends:="org.eclipse.ui"` for to make ui a friend.  At least that is what the fragments do for this package.

HannesWell: When searching `org.eclipse.ui` I can't see any usage of swt.internal anymore:
https://github.com/search?q=repo%3Aeclipse-platform%2Feclipse.platform.ui%20path%3A%2F%5Ebundles%5C%2Forg%5C.eclipse%5C.ui%5C%2F%2F%20org.eclipse.swt.internal&type=code

I already considered to remove it from the fragments too, but haven't done it yet.

iloveeclipse: I don't see any warnings on `org.eclipse.ui` and looking on the code I don't think there is a need for that, it uses only public SWT type..

Do you see any warnings?

tjwatson: Thanks for checking.  Then I agree, making x-internal is correct.

HannesWell:
Thank you Andrey for this fix.

PR #1045, opened 2024-02-11
Compile SWT natives for Windows on Arm64 (WoA).

chirontt:
This PR contains changes to the `build.bat` file, to support compiling the SWT natives for the WoA platform, and associated files to produce the SWT binaries module's jar/zip files for WoA.

On a WoA box, run the following commands to produce the SWT natives (`swt*.dll`) for WoA:
```batch
mvn clean process-resources -Dnative=win32.win32.aarch64
```
and the `swt*.dll` files for WoA will be created in the current directory.

Cross-compiling between the x64 and Arm64 platforms, where either of them can be host or target, is also possible, after setting up the `TARGET_ARCH` environment variable with correct "target architecture" value, and with the `SWT_JAVA_HOME` property pointing to an available JDK of the target architecture.

For example, to cross-compile on an x64 host and produce SWT natives for Arm64 target, run the following commands:
```batch
set TARGET_ARCH=arm64
mvn clean process-resources -Dnative=win32.win32.aarch64 -DSWT_JAVA_HOME=\path\to\arm64_jdk
```
and the `swt*.dll` files for WoA will be created in the current directory. Note the above `SWT_JAVA_HOME` property in the command line which points to an available Arm64 JDK of the target architecture. If needed, an Arm64 JDK for WoA can be downloaded from [Microsoft](https://learn.microsoft.com/en-us/java/openjdk/download), [Liberica](https://bell-sw.com/pages/downloads/), or [Zulu](https://www.azul.com/downloads/?package=jdk#download-openjdk).

Similarly, to cross-compile on an Arm64 host and produce SWT natives for x64 target, run the following commands:
```batch
set TARGET_ARCH=x64
cd binaries\org.eclipse.swt.win32.win32.x86_64
mvn clean process-resources -Dnative=win32.win32.x86_64 -DSWT_JAVA_HOME=\path\to\x64_jdk
```
and the `swt*.dll` files for x64 will be created in the current directory.

Note that for cross-compiling between x64 and Arm64 to work, install the MSVC compiler version 2022, with mandatory build tools for both platforms included in the installation. Further more, there must exist a JDK installation of the `target` platform available on the host machine, so that the `SWT_JAVA_HOME` variable can point to it during cross-compiling and linking.

Once the SWT natives for Arm64 are generated, the SWT binaries module's files can be produced with the following build commands in current directory:
```
mvn clean verify -DskipTests=true
```
and the module's jar/zip files will be produced in the `target` directory:
```
org.eclipse.swt.win32.win32.aarch64-<version>.jar
org.eclipse.swt.win32.win32.aarch64-<version>-sources.jar
swt-<version>-win32-win32-aarch64.zip
```

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -51,6 +51,18 @@ spec:
 
 /** Returns the download URL of the JDK against whoose C headers (in the 'include/' folder) and native libaries the SWT natives are compiled.*/
 def getNativeJdkUrl(String os, String arch){ // To update the used JDK version update the URL template below
+	if('win32'.equals(os) && 'aarch64'.equals(arch)) {
+		// Temporary workaround until there are official Temurin GA releases for Windows on ARM that can be consumed through JustJ
+		dir("${WORKSPACE}/repackage-win32.aarch64-jdk") {
+			sh """
+				curl -L 'https://github.com/adoptium/temurin17-binaries/releases/download/jdk17u-2024-02-07-14-14-beta/OpenJDK17U-jdk_aarch64_windows_hotspot_2024-02-07-14-14.zip' > jdk.zip
+				unzip -q jdk.zip jdk-17.0.11+1/include/** jdk-17.0.11+1/lib/**
+				cd jdk-17.0.11+1
+				tar -czf ../jdk.tar.gz include/ lib/
+			"""
+		}
+		return "file://${WORKSPACE}/repackage-win32.aarch64-jdk/jdk.tar.gz"
+	}
 	return "https://download.eclipse.org/justj/jres/17/downloads/20230428_1804/org.eclipse.justj.openjdk.hotspot.jre.minimal.stripped-17.0.7-${os}-${arch}.tar.gz"
 }
 
@@ -177,7 +189,7 @@ pipeline {
 				axes {
 					axis {
 						name 'PLATFORM'
-						values 'cocoa.macosx.aarch64' , 'cocoa.macosx.x86_64', 'gtk.linux.aarch64', 'gtk.linux.ppc64le', 'gtk.linux.x86_64', 'win32.win32.x86_64'
+						values 'cocoa.macosx.aarch64' , 'cocoa.macosx.x86_64', 'gtk.linux.aarch64', 'gtk.linux.ppc64le', 'gtk.linux.x86_64', 'win32.win32.aarch64', 'win32.win32.x86_64'
 					}
 				}
 				stages {
--- a/binaries/org.eclipse.swt.win32.win32.aarch64/META-INF/MANIFEST.MF
+++ b/binaries/org.eclipse.swt.win32.win32.aarch64/META-INF/MANIFEST.MF
@@ -1,0 +1,34 @@
+Manifest-Version: 1.0
+Fragment-Host: org.eclipse.swt;bundle-version="[3.125.100,4.0.0)"
+Bundle-Name: %fragmentName
+Bundle-Vendor: %providerName
+Bundle-SymbolicName: org.eclipse.swt.win32.win32.aarch64; singleton:=true
+Bundle-Version: 3.126.0.qualifier
+Bundle-ManifestVersion: 2
+Bundle-Localization: fragment
+Export-Package: 
+ org.eclipse.swt,
+ org.eclipse.swt.accessibility,
+ org.eclipse.swt.awt,
+ org.eclipse.swt.browser,
+ org.eclipse.swt.custom,
+ org.eclipse.swt.dnd,
+ org.eclipse.swt.events,
+ org.eclipse.swt.graphics,
+ org.eclipse.swt.layout,
+ org.eclipse.swt.opengl,
+ org.eclipse.swt.printing,
+ org.eclipse.swt.program,
+ org.eclipse.swt.widgets,
+ org.eclipse.swt.internal; x-friends:="org.eclipse.ui",
+ org.eclipse.swt.internal.image; x-internal:=true,
+ org.eclipse.swt.ole.win32,
+ org.eclipse.swt.internal.gdip; x-internal:=true,
+ org.eclipse.swt.internal.ole.win32; x-internal:=true,
+ org.eclipse.swt.internal.win32; x-internal:=true,
+ org.eclipse.swt.internal.opengl.win32; x-internal:=true
+Eclipse-PlatformFilter: (& (osgi.ws=win32) (osgi.os=win32) (osgi.arch=aarch64))
+SWT-WS: win32
+SWT-OS: win32
+SWT-Arch: aarch64
+Automatic-Module-Name: org.eclipse.swt.win32.win32.aarch64
--- a/binaries/org.eclipse.swt.win32.win32.aarch64/build.properties
+++ b/binaries/org.eclipse.swt.win32.win32.aarch64/build.properties
@@ -1,0 +1,41 @@
+###############################################################################
+# Copyright (c) 2023, 2024 Tue Ton and others.
+# All rights reserved. This program and the accompanying materials
+# are made available under the terms of the Eclipse Public License v1.0
+# which accompanies this distribution, and is available at
+# http://www.eclipse.org/legal/epl-v10.html
+#
+# Contributors:
+#     Tue Ton - initial API and implementation
+#     Hannes Wellmann - Leverage Tycho pomless
+###############################################################################
+custom = true
+bin.includes = .,*.dll,fragment.properties
+bin.excludes = library/
+source.. = \
+	../legal_files/win32.win32,\
+	../../bundles/org.eclipse.swt/Eclipse SWT/win32,\
+	../../bundles/org.eclipse.swt/Eclipse SWT/common,\
+	../../bundles/org.eclipse.swt/Eclipse SWT PI/common,\
+	../../bundles/org.eclipse.swt/Eclipse SWT PI/win32,\
+	../../bundles/org.eclipse.swt/Eclipse SWT OLE Win32/win32,\
+	../../bundles/org.eclipse.swt/Eclipse SWT Accessibility/win32,\
+	../../bundles/org.eclipse.swt/Eclipse SWT Accessibility/common,\
+	../../bundles/org.eclipse.swt/Eclipse SWT AWT/win32,\
+	../../bundles/org.eclipse.swt/Eclipse SWT AWT/common,\
+	../../bundles/org.eclipse.swt/Eclipse SWT Drag and Drop/win32,\
+	../../bundles/org.eclipse.swt/Eclipse SWT Drag and Drop/common,\
+	../../bundles/org.eclipse.swt/Eclipse SWT Printing/win32,\
+	../../bundles/org.eclipse.swt/Eclipse SWT Printing/common,\
+	../../bundles/org.eclipse.swt/Eclipse SWT Program/win32,\
+	../../bundles/org.eclipse.swt/Eclipse SWT Program/common,\
+	../../bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common,\
+	../../bundles/org.eclipse.swt/Eclipse SWT Browser/common,\
+	../../bundles/org.eclipse.swt/Eclipse SWT Browser/win32,\
+	../../bundles/org.eclipse.swt/Eclipse SWT OpenGL/win32,\
+	../../bundles/org.eclipse.swt/Eclipse SWT OpenGL/common
+output.. = bin/
+
+pom.model.property.os=win32
+pom.model.property.ws=win32
+pom.model.property.arch=aarch64
--- a/bundles/org.eclipse.swt/Eclipse SWT PI/win32/library/build.bat
+++ b/bundles/org.eclipse.swt/Eclipse SWT PI/win32/library/build.bat
@@ -13,7 +13,6 @@
 @rem ***************************************************************************
 
 @rem The original build.bat source is located in /org.eclipse.swt/Eclipse SWT PI/win32/library/build.bat. It is copied during various build(s).
-@rem Typically it's not ran directly, instead it is reached by build.xml's build_libraries target found in eclipse.platform.swt\bundles\org.eclipse.swt.win32.win32.x86*
 
 @echo off
 echo
@@ -51,15 +50,44 @@ IF NOT EXIST "%MSVC_HOME%" (
 @rem Check for a usable JDK
 IF "%SWT_JAVA_HOME%"=="" CALL :ECHO "'SWT_JAVA_HOME' was not provided"
 IF NOT EXIST "%SWT_JAVA_HOME%" (
-    CALL :ECHO "WARNING: x64 Java JDK not found. Please set SWT_JAVA_HOME to the JDK directory containing the intended JDK native headers."
+    CALL :ECHO "WARNING: 64-bit Java JDK not found. Please set SWT_JAVA_HOME to the JDK directory containing the intended JDK native headers."
 )
 
-@rem -----------------------
-set PROCESSOR_ARCHITECTURE=AMD64
-IF "x.%OUTPUT_DIR%"=="x." set OUTPUT_DIR=..\..\..\org.eclipse.swt.win32.win32.x86_64
+@REM Compose host architecture string for MSVC
+IF "%PROCESSOR_ARCHITECTURE%"=="AMD64" (
+  SET HOST_ARCH=x64
+) ELSE IF "%PROCESSOR_ARCHITECTURE%"=="ARM64" (
+  SET HOST_ARCH=arm64
+) ELSE (
+  CALL :ECHO "ERROR: Unknown host architecture: %PROCESSOR_ARCHITECTURE%."
+  EXIT /B 1
+)
+
+@REM %TARGET_ARCH% may be specified by the caller for cross-compiling.
+@REM If not, build for builder machine's architecture
+IF "%TARGET_ARCH%"=="" (
+  SET TARGET_ARCH=%HOST_ARCH%
+)
+
+@REM Compose build argument for MSVC
+IF "%TARGET_ARCH%"=="%HOST_ARCH%" (
+  SET BUILD_ARCH=%TARGET_ARCH%
+) ELSE (
+  SET BUILD_ARCH=%HOST_ARCH%_%TARGET_ARCH%
+)
+
+@REM Select build's output directory (if not specified) based on target arch
+IF "%TARGET_ARCH%"=="x64" (
+  IF "x.%OUTPUT_DIR%"=="x." SET OUTPUT_DIR=..\..\..\org.eclipse.swt.win32.win32.x86_64
+) ELSE IF "%TARGET_ARCH%"=="arm64" (
+  IF "x.%OUTPUT_DIR%"=="x." SET OUTPUT_DIR=..\..\..\org.eclipse.swt.win32.win32.aarch64
+) ELSE (
+  CALL :ECHO "ERROR: Unknown target architecture: %TARGET_ARCH%."
+  EXIT /B 1
+)
 
 set CFLAGS=-DJNI64
-call "%MSVC_HOME%\VC\Auxiliary\Build\vcvarsall.bat" x64
+call "%MSVC_HOME%\VC\Auxiliary\Build\vcvarsall.bat" %BUILD_ARCH%
 
 @rem if call to vcvarsall.bat (which sets up environment) silently fails, then provide advice to user.
 WHERE cl
--- a/pom.xml
+++ b/pom.xml
@@ -172,6 +172,11 @@
                   <ws>win32</ws>
                   <arch>x86_64</arch>
                 </environment>
+                <environment>
+                  <os>win32</os>
+                  <ws>win32</ws>
+                  <arch>aarch64</arch>
+                </environment>
               </environments>
             </configuration>
           </plugin>
--- a/binaries/org.eclipse.swt.win32.win32.aarch64/.project
+++ b/binaries/org.eclipse.swt.win32.win32.aarch64/.project
@@ -1,0 +1,117 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
+	<name>org.eclipse.swt.win32.win32.aarch64</name>
+	<comment></comment>
+	<projects>
+	</projects>
+	<buildSpec>
+		<buildCommand>
+			<name>org.eclipse.jdt.core.javabuilder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
+		<buildCommand>
+			<name>org.eclipse.pde.ManifestBuilder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
+		<buildCommand>
+			<name>org.eclipse.pde.SchemaBuilder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
+		<buildCommand>
+			<name>org.eclipse.pde.api.tools.apiAnalysisBuilder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
+	</buildSpec>
+	<natures>
+		<nature>org.eclipse.jdt.core.javanature</nature>
+		<nature>org.eclipse.pde.PluginNature</nature>
+		<nature>org.eclipse.pde.api.tools.apiAnalysisNature</nature>
+	</natures>
+	<linkedResources>
+		<link>
+			<name>.classpath</name>
+			<type>1</type>
+			<locationURI>PARENT-1-PROJECT_LOC/.classpath_win32</locationURI>
+		</link>
+		<link>
+			<name>.settings/.api_filters</name>
+			<type>1</type>
+			<locationURI>PROJECT_LOC/.settings/.api_filters</locationURI>
+		</link>
+		<link>
+			<name>.settings</name>
+			<type>2</type>
+			<locationURI>PARENT-1-PROJECT_LOC/.settings</locationURI>
+		</link>
+		<link>
+			<name>Eclipse SWT</name>
+			<type>2</type>
+			<locationURI>SWT_HOST_PLUGIN/Eclipse%20SWT</locationURI>
+		</link>
+		<link>
+			<name>Eclipse SWT Accessibility</name>
+			<type>2</type>
+			<locationURI>SWT_HOST_PLUGIN/Eclipse%20SWT%20Accessibility</locationURI>
+		</link>
+		<link>
+			<name>Eclipse SWT AWT</name>
+			<type>2</type>
+			<locationURI>SWT_HOST_PLUGIN/Eclipse%20SWT%20AWT</locationURI>
+		</link>
+		<link>
+			<name>Eclipse SWT Browser</name>
+			<type>2</type>
+			<locationURI>SWT_HOST_PLUGIN/Eclipse%20SWT%20Browser</locationURI>
+		</link>
+		<link>
+			<name>Eclipse SWT Custom Widgets</name>
+			<type>2</type>
+			<locationURI>SWT_HOST_PLUGIN/Eclipse%20SWT%20Custom%20Widgets</locationURI>
+		</link>
+		<link>
+			<name>Eclipse SWT Drag and Drop</name>
+			<type>2</type>
+			<locationURI>SWT_HOST_PLUGIN/Eclipse%20SWT%20Drag%20and%20Drop</locationURI>
+		</link>
+		<link>
+			<name>Eclipse SWT OLE Win32</name>
+			<type>2</type>
+			<locationURI>SWT_HOST_PLUGIN/Eclipse%20SWT%20OLE%20Win32</locationURI>
+		</link>
+		<link>
+			<name>Eclipse SWT OpenGL</name>
+			<type>2</type>
+			<locationURI>SWT_HOST_PLUGIN/Eclipse%20SWT%20OpenGL</locationURI>
+		</link>
+		<link>
+			<name>Eclipse SWT PI</name>
+			<type>2</type>
+			<locationURI>SWT_HOST_PLUGIN/Eclipse%20SWT%20PI</locationURI>
+		</link>
+		<link>
+			<name>Eclipse SWT Printing</name>
+			<type>2</type>
+			<locationURI>SWT_HOST_PLUGIN/Eclipse%20SWT%20Printing</locationURI>
+		</link>
+		<link>
+			<name>Eclipse SWT Program</name>
+			<type>2</type>
+			<locationURI>SWT_HOST_PLUGIN/Eclipse%20SWT%20Program</locationURI>
+		</link>
+		<link>
+			<name>Eclipse SWT WebKit</name>
+			<type>2</type>
+			<locationURI>SWT_HOST_PLUGIN/Eclipse%20SWT%20WebKit</locationURI>
+		</link>
+	</linkedResources>
+	<variableList>
+		<variable>
+			<name>SWT_HOST_PLUGIN</name>
+			<value>$%7BPARENT-2-PROJECT_LOC%7D/bundles/org.eclipse.swt</value>
+		</variable>
+	</variableList>
+</projectDescription>
--- a/binaries/org.eclipse.swt.win32.win32.aarch64/.settings/.api_filters
+++ b/binaries/org.eclipse.swt.win32.win32.aarch64/.settings/.api_filters
@@ -1,0 +1,603 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<component id="org.eclipse.swt.win32.win32.aarch64" version="2">
+    <resource path="Eclipse SWT Accessibility/common/org/eclipse/swt/accessibility/AccessibleActionListener.java" type="org.eclipse.swt.accessibility.AccessibleActionListener">
+        <filter id="576720909">
+            <message_arguments>
+                <message_argument value="SWTEventListener"/>
+                <message_argument value="AccessibleActionListener"/>
+            </message_arguments>
+        </filter>
+    </resource>
+    <resource path="Eclipse SWT Accessibility/common/org/eclipse/swt/accessibility/AccessibleAttributeListener.java" type="org.eclipse.swt.accessibility.AccessibleAttributeListener">
+        <filter id="576720909">
+            <message_arguments>
+                <message_argument value="SWTEventListener"/>
+                <message_argument value="AccessibleAttributeListener"/>
+            </message_arguments>
+        </filter>
+    </resource>
+    <resource path="Eclipse SWT Accessibility/common/org/eclipse/swt/accessibility/AccessibleControlListener.java" type="org.eclipse.swt.accessibility.AccessibleControlListener">
+        <filter id="576720909">
+            <message_arguments>
+                <message_argument value="SWTEventListener"/>
+                <message_argument value="AccessibleControlListener"/>
+            </message_arguments>
+        </filter>
+    </resource>
+    <resource path="Eclipse SWT Accessibility/common/org/eclipse/swt/accessibility/AccessibleEditableTextListener.java" type="org.eclipse.swt.accessibility.AccessibleEditableTextListener">
+        <filter id="576720909">
+            <message_arguments>
+                <message_argument value="SWTEventListener"/>
+                <message_argument value="AccessibleEditableTextListener"/>
+            </message_arguments>
+        </filter>
+    </resource>
+    <resource path="Eclipse SWT Accessibility/common/org/eclipse/swt/accessibility/AccessibleHyperlinkListener.java" type="org.eclipse.swt.accessibility.AccessibleHyperlinkListener">
+        <filter id="576720909">
+            <message_arguments>
+                <message_argument value="SWTEventListener"/>
+                <message_argument value="AccessibleHyperlinkListener"/>
+            </message_arguments>
+        </filter>
+    </resource>
+    <resource path="Eclipse SWT Accessibility/common/org/eclipse/swt/accessibility/AccessibleListener.java" type="org.eclipse.swt.accessibility.AccessibleListener">
+        <filter id="576720909">
+            <message_arguments>
+                <message_argument value="SWTEventListener"/>
+                <message_argument value="AccessibleListener"/>
+            </message_arguments>
+        </filter>
+    </resource>
+    <resource path="Eclipse SWT Accessibility/common/org/eclipse/swt/accessibility/AccessibleTableCellListener.java" type="org.eclipse.swt.accessibility.AccessibleTableCellListener">
+        <filter id="576720909">
+            <message_arguments>
+                <message_argument value="SWTEventListener"/>
+                <message_argument value="AccessibleTableCellListener"/>
+            </message_arguments>
+        </filter>
+    </resource>
+    <resource path="Eclipse SWT Accessibility/common/org/eclipse/swt/accessibility/AccessibleTableListener.java" type="org.eclipse.swt.accessibility.AccessibleTableListener">
+        <filter id="576720909">
+            <message_arguments>
+                <message_argument value="SWTEventListener"/>
+                <message_argument value="AccessibleTableListener"/>
+            </message_arguments>
+        </filter>
+    </resource>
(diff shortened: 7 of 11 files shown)


bundles/org.eclipse.swt/Eclipse SWT PI/win32/library/os_custom.c
```diff
@@ -219,6 +227,10 @@ BOOL Validate_SetPreferredAppMode(const BYTE* functionPtr)
 		(functionPtr[0x00] == 0x8B) && (functionPtr[0x01] == 0x05) &&   // mov     eax,dword ptr [uxtheme!g_preferredAppMode]
 		(functionPtr[0x06] == 0x87) && (functionPtr[0x07] == 0x0D) &&   // xchg    ecx,dword ptr [uxtheme!g_preferredAppMode]
 		(functionPtr[0x0C] == 0xC3);                                    // ret
+#elif defined(_M_ARM64)
+	/* Not implemented yet */
+	functionPtr; /* to prevent: warning C4100: 'functionPtr': unreferenced formal parameter */
+	return FALSE;
```
hmartinez82, line 233: You should return TRUE here and SWT on Windows on ARM should require Windows 10 1903 or higher.

Or we can implement the disassembly hack that is done for x64 with ARM64 code.

If you need to understand why they did hacky detection, see
https://github.com/mintty/mintty/issues/983#issuecomment-611868330
and
https://stackoverflow.com/a/58547831/827532

chirontt: Yes, return TRUE solves the problem of scrollbars not in dark mode, on my WoA box! The scrollbars now show dark color correctly:
image

I will fix the code later, thanks. Does that mean we don't need any disassembly hack for WoA then?

hmartinez82: If we state that SWT on Windows on ARM require Windows 1903+ then we can return TRUE there and not do the assembly hack.

hmartinez82: Also. I personally would've not done that assembly hack. If you look at one of the links above, this is way less hacky:
```
                auto ord135 = GetProcAddress(hUxtheme, MAKEINTRESOURCEA(135));
                if (g_buildNumber < 18334)
                    _AllowDarkModeForApp = reinterpret_cast<fnAllowDarkModeForApp>(ord135);
                else
                    _SetPreferredAppMode = reinterpret_cast<fnSetPreferredAppMode>(ord135);
```
In the Validate_AllowDarkModeForWindowWithTelemetryId() all they would have to do was to just query the build version of Windows to decide if TRUE or FALSE stating that uxtheme has the expected function on ordinal 135.

SyntevoAlex:
All of the above comments are wrong.
Shame on you, you should have read commit message for commit where code was introduced.

SyntevoAlex:
This is (in short way) explained in code with code comment:
```
	/*
	 * In next Windows version, some other function can end up having this ordinal.
	 * Compare function's code to known signature to make sure.
	 */
```

hmartinez82:
So we need an ARM64 equivalent of the hack? I can try looking into it.

SyntevoAlex:
Right. One would need to dissasemble ARM windows DLLs and decide which part of it uniquely identifies desired functions.

SyntevoAlex:
Or you can just return FALSE for now. It can be supported later.

SyntevoAlex:
If you choose to return FALSE, please add another code comment explaining why.

hmartinez82:
For  `AllowDarkModeForWindowWithTelemetryId`() good news, that `0xA91E` is also there for ARM64. `pacibsp` is the first instruction. This will be a bit tricker than x64 because of the way that `0xA91E` is encoded. I'm leaving this here as doc.
```
00007FFB16C33210 D503237F             pacibsp  
00007FFB16C33214 A9BE53F3             stp         x19,x20,[sp,#-0x20]!  
00007FFB16C33218 F9000BF5             str         x21,[sp,#0x10]  
00007FFB16C3321C A9BF7BFD             stp         fp,lr,[sp,#-0x10]!  
00007FFB16C33220 910003FD             mov         fp,sp  
00007FFB16C33224 53001C34             uxtb        w20,w1  
00007FFB16C33228 D29523C1             mov         x1,#0xA91E  
00007FFB16C3322C AA0003F5             mov         x21,x0  
00007FFB16C33230 9400EEB6             bl          #GetPropW (07FFB16C6ED08h)
```

bundles/org.eclipse.swt/Eclipse SWT PI/win32/library/build.bat
```diff
@@ -51,15 +50,33 @@ IF NOT EXIST "%MSVC_HOME%" (
...
+IF NOT "%XC_HOST_TARGET%"=="" (
+  @rem Accept the already-set variable value for host_target, for manual cross-compile
+  CALL :ECHO "Manual cross-compile of host_target: %XC_HOST_TARGET%"
+  SET BUILD_ARCH=%XC_HOST_TARGET%
+  @rem With the knowledge that the OUTPUT_DIR is also already set
+  CALL :ECHO "Cross-compile output dir: %OUTPUT_DIR%"
+  IF "x.%OUTPUT_DIR%"=="x." (
+    CALL :ECHO "ERROR: OUTPUT_DIR environment variable must be set for manual cross-compile."
+    EXIT /B 1
+  )
```
SyntevoAlex, line 66: I think that this code can be simplified.

Realistically I don't expect any new archs in upcoming years, so there will be just two - amd64 and arm64. So I suggest to have just one new variable that would override `%PROCESSOR_ARCHITECTURE%`. For example, `COMPILE_ARCH` could be the only new variable. If it's empty, it falls back to current processor arch:
```
IF "%COMPILE_ARCH%"=="" SET COMPILE_ARCH=%PROCESSOR_ARCHITECTURE%
```
this way, you only use one var instead of two, and there's less code.

SyntevoAlex (commented):
I didn't review Java specific build files because I have no experience with these.

bundles/org.eclipse.swt/Eclipse SWT PI/win32/library/build.bat
chirontt: Doing it this way means the cross-compiling idea is abandoned. Is this what you really want?

The required OUTPUT_DIR variable must be set, either internally within this `build.bat` file, or externally set by the caller of this `build.bat` script. Perhaps we can make this `build.bat` script smarter, in working out the required OUTPUT_DIR value from the COMPILE_ARCH (or XC_HOST_TARGET) value, by parsing either `x64_arm64` or `arm64_x64` and working out the needed OUTPUT_DIR value.

bundles/org.eclipse.swt/Eclipse SWT PI/win32/library/os_custom.c
chirontt: I've updated the PR with new commit returning TRUE for Arm64, so no need for any further implementation. Windows on Arm64 is really only Windows 11 going forward, so there's no need for any checking for Windows 10 1903+ in Arm64 because Windows 10 is no longer supported in Arm64 by Microsoft.

hmartinez82: I think that what @SyntevoAlex is saying is that since these three functions are undocumented (they are not even exported by name, only by ordinal) they could be moved to another ordinal completely or even removed in the future builds of Windows 11 , thus it's unsafe to just return TRUE without checking first if the function is indeed the one expected at a given ordinal.

hmartinez82:
> `AllowDarkModeForWindow()`:
> 
> ```
> 00007FFB16C32450 D503237F             pacibsp  
> 00007FFB16C32454 A9BE53F3             stp         x19,x20,[sp,#-0x20]!  
> 00007FFB16C32458 F9000BF5             str         x21,[sp,#0x10]  
> 00007FFB16C3245C A9BF7BFD             stp         fp,lr,[sp,#-0x10]!  
> 00007FFB16C32460 910003FD             mov         fp,sp  
> 00007FFB16C32464 53001C33             uxtb        w19,w1  
> 00007FFB16C32468 D29523C1             mov         x1,#0xA91E  
> 00007FFB16C3246C AA0003F5             mov         x21,x0  
> 00007FFB16C32470 9400F226             bl          #GetPropW (07FFB16C6ED08h)
> ...
> ```
> 
> `AllowDarkModeForWindowWithTelemetryId():`
> 
> ```
> 00007FFB16C33210 D503237F             pacibsp  
> 00007FFB16C33214 A9BE53F3             stp         x19,x20,[sp,#-0x20]!  
> 00007FFB16C33218 F9000BF5             str         x21,[sp,#0x10]  
> 00007FFB16C3321C A9BF7BFD             stp         fp,lr,[sp,#-0x10]!  
> 00007FFB16C33220 910003FD             mov         fp,sp  
> 00007FFB16C33224 53001C34             uxtb        w20,w1  
> 00007FFB16C33228 D29523C1             mov         x1,#0xA91E  
> 00007FFB16C3322C AA0003F5             mov         x21,x0  
> 00007FFB16C33230 9400EEB6             bl          #GetPropW (07FFB16C6ED08h)
> ...
> ```
> 
> `SetPreferredAppMode()`:
> 
> ```
> 00007FFB16C33E10 D503237F             pacibsp  
> 00007FFB16C33E14 F81F0FF3             str         x19,[sp,#-0x10]!  
> 00007FFB16C33E18 A9BF7BFD             stp         fp,lr,[sp,#-0x10]!  
> 00007FFB16C33E1C 910003FD             mov         fp,sp  
> 00007FFB16C33E20 D00007E8             adrp        x8,wil::details::g_enabledStateManager+40h (07FFB16D31000h)  
> 00007FFB16C33E24 B94BD913             ldr         w19,[x8,#0xBD8]  
> 00007FFB16C33E28 2A0003E1             mov         w1,w0  
> 00007FFB16C33E2C 912F6100             add         x0,x8,#0xBD8  
> 00007FFB16C33E30 97FF7910             bl          _InterlockedExchange (07FFB16C12270h)  
> 00007FFB16C33E34 2A1303E0             mov         w0,w19  
> 00007FFB16C33E38 A8C17BFD             ldp         fp,lr,[sp],#0x10  
> 00007FFB16C33E3C F84107F3             ldr         x19,[sp],#0x10  
> 00007FFB16C33E40 D50323FF             autibsp  
> 00007FFB16C33E44 D65F03C0             ret  
> 00007FFB16C33E48 B0000090             adrp        xip0,GetBufferedPaintBitsEx+7320h (07FFB16C44000h)
> ...
> ```

@SyntevoAlex any recommendations here? Since the instructions have fixed width, I was going to just compare the expected opcodes at a certain offset from the start of the functions. 
For `AllowDarkModeForWindow()` check for `D29523C1             mov         x1,#0xA91E`
For `AllowDarkModeForWindowWithTelemetryId()` check for `D29523C1             mov         x1,#0xA91E`
For ``SetPreferredAppMode()` check for `912F6100             add         x0,x8,#0xBD8`

Any other suggestion?

bundles/org.eclipse.swt/Eclipse SWT PI/win32/library/os_custom.c
SyntevoAlex: Yes, the functions are still NOT documented, so they still need to be checked.

bundles/org.eclipse.swt/Eclipse SWT PI/win32/library/build.bat
SyntevoAlex: Cross-compiled arm64 is the same as native-compiled arm64. Therefore, if arm64 is compiled, output dirs will be for arm64.

SyntevoAlex:
I think that suggested checks are reasonable.

hmartinez82:
@chirontt here are the verifications to be used instead of just returning `TRUE` in the `_M_ARM64` sections:
For `Validate_AllowDarkModeForWindow()`:
```
#ifdef _M_ARM64
	if (*reinterpret_cast<const uint32_t*>(&functionPtr[0x18]) == 0xD29523C1) // mov x1,#0xA91E
	{
		return TRUE;
	}

	return FALSE;
```

For `AllowDarkModeForWindowWithTelemetryId()`:
```
#ifdef _M_ARM64
	if (*reinterpret_cast<const uint32_t*>(&functionPtr[0x18]) == 0xD29523C1) // mov x1,#0xA91E
	{
		return TRUE;
	}

	return FALSE;
```

For `SetPreferredAppMode()`:
```
#ifdef _M_ARM64
        if (*reinterpret_cast<const uint32_t*>(&functionPtr[0x1C]) == 0x912F6100) // add x0,x8,#0xBD8
	{
		return TRUE;
	}

    return FALSE;
```

bundles/org.eclipse.swt/Eclipse SWT PI/win32/library/build.bat
chirontt: Nope, on an x64 box, passing 'arm64' to initialize and compile the C code for Arm64 will certainly fail to work (I've tried it). For cross-compile, MSVC must know from which host to which target explicitly, hence it requires 'x64_arm64' on an x64 box, or 'arm64_x64' on an Arm64 box, for cross-compiling.

bundles/org.eclipse.swt/Eclipse SWT PI/win32/library/os_custom.c
chirontt: @SyntevoAlex @hmartinez82 I don't understand any of the proposed changes to the `os_custom.c` file.

I'm thinking of removing this `os_custom.c` file from this PR, so that you guys can submit it in a separate new PR with your proposed changes for Arm64. And my PR will wait for your new PR to be submitted first.

hmartinez82: @chirontt I can do that, but I don't know how to compile this.  I have no Maven or Java experience :( 
That's why I copied the code here. I can try to explain to you were those numbers 0x18 and 0x1C are coming from:

The validation check for x64, was done by looking at the machine code of those three functions, `AllowDarkModeForWindow()`, `AllowDarkModeForWindowWithTelemetryId()`, `SetPreferredAppMode()`. 
This was done by someone one Visual Studio, in an x64 machine.
With the disassembly of the code, one can see which bytes are at each address after the start of the function (the start of the function is the value returned by GetProcAddress()

Here's `SetPreferredAppMode()` for instance on x64:
image

As you can see, you have to open the disassembly view to be able to Step In the function call (because we don't have the source code for uxtheme.dll), but it's doable. 
The numbers circled in blue in the picture above are being tested in this line for instance https://github.com/eclipse-platform/eclipse.platform.swt/pull/1045/files#diff-d6068e15e4497052c1129ba671192ebce3920a189d39a48d0cb66e151019f9b8R227 in `Validate_SetPreferredAppMode()`

hmartinez82: @SyntevoAlex @chirontt https://github.com/eclipse-platform/eclipse.platform.swt/pull/1048

bundles/org.eclipse.swt/Eclipse SWT PI/win32/library/build.bat
SyntevoAlex: ok, so the expected value for `BUILD_ARCH` is for example `x64_arm64`, I didn't catch that initially. I think that it still makes sense to compose `BUILD_ARCH` from `%PROCESSOR_ARCHITECTURE%` and requested target arch, because the code would be easier to understand. What do you think?

```
REM Compose host architecture string for MSVC
IF "%PROCESSOR_ARCHITECTURE%"=="AMD64" (
  SET HOST_ARCH=x64
) ELSE IF "%PROCESSOR_ARCHITECTURE%"=="ARM64" (
  SET HOST_ARCH=arm64
) ELSE (
  CALL :ECHO "ERROR: Unknown host architecture: %PROCESSOR_ARCHITECTURE%."
  EXIT /B 1
)

REM %TARGET_ARCH% may be specified by the caller for cross-compiling.
REM If not, build for builder machine's architecture
IF "%TARGET_ARCH%"=="" (
  SET  TARGET_ARCH=%HOST_ARCH%
)

REM Compose build argument for MSVC
IF "%TARGET_ARCH%"=="%HOST_ARCH%" (
  SET BUILD_ARCH=%TARGET_ARCH%
) ELSE (
   SET BUILD_ARCH=%HOST_ARCH%_%TARGET_ARCH%
)

REM Select build directory based on target arch
IF "%TARGET_ARCH%"=="x64" (
  IF "x.%OUTPUT_DIR%"=="x." SET OUTPUT_DIR=..\..\..\org.eclipse.swt.win32.win32.x86_64
) ELSE IF "%TARGET_ARCH%"=="arm64" (
  IF "x.%OUTPUT_DIR%"=="x." SET OUTPUT_DIR=..\..\..\org.eclipse.swt.win32.win32.aarch64
) ELSE (
  CALL :ECHO "ERROR: Unknown target architecture: %TARGET_ARCH%."
  EXIT /B 1
)
```

SyntevoAlex: Thinking about it some more, I figured that my approach has an even bigger advantage: user only specifies what he wants, and doesn't have to care about how to make this happen. Script will figure that it's cross-compiling on its own.

SyntevoAlex: This, in turn, will simplify builder scripts: that will just say "build that and that", regardless of what they're running on.

chirontt:
I've restored the `os_custom.c` file to its base version before this PR change, effectively removing it from this PR, as the file is now handled by PR #1048

SyntevoAlex:
#1048 was merged.
Please squash branch (thus removing any changes in `os_ciustom.c`) and rebase it.
I'm still waiting for your reply regarding script improvement.

bundles/org.eclipse.swt/Eclipse SWT PI/win32/library/build.bat
chirontt: OK, your approach is a bit verbose, but it's quite clear in intention, so I have no problem with it. And current Maven build scripts would still work with it without change (this was my intention all along.) And doing cross-compiling has to be a _manual_ operation anyway, that the user need to explicitly specify an extra environment variable in order to trigger it, so I won't expect the current automated build scripts to do this cross-compiling, but the facility is there for automated cross-compile by the build scripts if required.

I'll merge this with the PR and squash it. And the PR's description need be changed too, to reflect the updated contents.

Another thing: I plan to apply these changes of `build.bat`, to the [build.bat](https://github.com/eclipse-equinox/equinox/blob/master/features/org.eclipse.equinox.executable.feature/library/win32/build.bat) of the Equinox project, to generate the Equinox launcher native for Arm64. If you have different idea (for the Equinox project), let me know in advance, thanks.

chirontt:
Branch squashed and rebased to latest master; PR's description updated also, to correctly describe the updated contents.

To summarize the description, if you're on an x64 box and want to cross-compile to produce the SWT natives for Arm64, here are the commands:
```batch
set TARGET_ARCH=arm64
mvn clean process-resources -Dnative=win32.win32.aarch64 -DSWT_JAVA_HOME=\path\to\arm64_jdk
```
If you're authorized, then you could commit the generated `swt*.dll` binaries to the repo.

SyntevoAlex (approved):
Looks good to me.
I guess we'll merge it when master is open again. Probably there's no rush to merge it ASAP?

chirontt:
Branch rebased to latest master; ready to merge.

HannesWell:
Thanks for working on this.

> If you're authorized, then you could commit the generated `swt*.dll` binaries to the repo.

The dll's should not be submitted manually only the Jenkins build should do that. With the current build pipeline the manually submitted would then probably be deleted on the next change that requires recompilation.
I'm currently on vacation but when back next week I can help you with setting up the build-pipeline for that, given the infrastructure is ready.

laeubi:
> Branch rebased to latest master; ready to merge.

> I'm currently on vacation but when back next week I can help you with setting up the build-pipeline for that, given the infrastructure is ready.

@jonahgraham can you tell about the state of the hardware here?

binaries/org.eclipse.swt.win32.win32.aarch64/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,34 @@
+Manifest-Version: 1.0
+Fragment-Host: org.eclipse.swt;bundle-version="[3.125.0,4.0.0)"
+Bundle-Name: %fragmentName
```
laeubi, line 3: The `fragment.properties` file seem missing that defines this values.

laeubi (commented):
This module needs to be mentioned in https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/binaries/pom.xml as well to be included in the build.

binaries/org.eclipse.swt.win32.win32.aarch64/build.properties
```diff
@@ -0,0 +1,40 @@
...
+	../../bundles/org.eclipse.swt/Eclipse SWT Printing/win32,\
+	../../bundles/org.eclipse.swt/Eclipse SWT Printing/common,\
+	../../bundles/org.eclipse.swt/Eclipse SWT Program/win32,\
+	../../bundles/org.eclipse.swt/Eclipse SWT Program/common,\
+	../../bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common,\
+	../../bundles/org.eclipse.swt/Eclipse SWT Browser/common,\
+	../../bundles/org.eclipse.swt/Eclipse SWT Browser/win32,\
+	../../bundles/org.eclipse.swt/Eclipse SWT OpenGL/win32,\
+	../../bundles/org.eclipse.swt/Eclipse SWT OpenGL/common
+src.includes = about.html,about_files/
```
laeubi, line 36: These files are missing as well.

chirontt: The `build.properties` and `MANIFEST.MF` files are not really needed to produce the SWT natives for this PR; they are needed so that the Maven build script would work with the following command to generate the SWT natives:
```
mvn clean process-resources -Dnative=win32.win32.aarch64
```
because the build script requires them to be there, hence they are included in this PR.

The above files are really part of my second PR that will actually produce the `win32.aarch64` module's jar and zip files which will include these SWT natives. I was waiting for this PR to be merged, so that the SWT natives would be produced and checked in, before submitting the second PR. But now you've mentioned these files, and to avoid confusion, I will submit the second PR quickly, to complete my work for Windows on Arm64 support for this SWT repo.

laeubi: I don't think it makes sense to have both PRs split... can we not just use one PR and maybe just comment out the module if it should not been build yet by maven?

chirontt: #1083 submitted, to complement this PR.

jonahgraham:
> @jonahgraham can you tell about the state of the hardware here?

I have asked for a timeline on https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/4276#note_1889851

binaries/org.eclipse.swt.win32.win32.aarch64/build.properties
chirontt: The separate PRs serve distinct aims: this PR is for producing the `SWT natives` for Windows Arm64, while the other one (#1083) is for building the SWT jar/zip files. The second one is dependent on this PR to be merged first and be generated the official `swt*.dll` files, because the second PR would fail to build if the `*.dll` files are missing.

But if you insist, I can merge the two PRs into this one, but will leave the SWT jar/zip files not built, and someone will have to modify it later to get it built by Maven, once the `swt*.dll` are generated and checked in to the repo.

laeubi: I  would say lets wait for hannes, but the checkked in binaries afaik are only for developer convenience now and are rebuild anyways

HannesWell: Yes please be so kind and combine both changes in PR. Otherwise we can't verify that everything is working.
And without the native build Jenkins agents we can't submit any binary as they would get deleted or out of sync.
If you prefer you can still have both changes in separated commits in this PR.

Btw. there have been changes in the configuration of the fragments recently that you probably should pick up.

chirontt: I've merged PR #1083 to this PR, rebased to the latest master branch and updated this branch according.

HannesWell:
Thank you @chirontt for updating this PR in general it looks good.
I have updated the api-filters and adjusted the Jenkins pipeline to also build natives for win32.aarch64.
[Now that there is a dedicated agent for that](https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/4276#note_1938461) I'll check whats necessary to make the natives build.
We can probably land that soon.

One thing I'm not sure about is the handling of the `WebView2Loader.dll`. That's the only dll in a windows fragment that's not build by ourself but comes from Microsoft. I'll have to check if there is a arm specific version we can add or if we can use the same.

hmartinez82:
@HannesWell There's one for ARM64 too. We used it at work. You can download it from https://developer.microsoft.com/en-us/microsoft-edge/webview2?form=MA13LH#download

chirontt:
I'm not sure if the `WebView2Loader.dll` file is needed anymore in SWT. I've read somewhere that it's no longer used in SWT, but I can't find the reference anymore.

I've used my own build of Eclipse 2024-03 SDK in my Windows Arm64 box, and it certainly doesn't contain this `WebView2Loader.dll` file (for Arm64), but I have no problem so far with my Eclipse IDE, like browsing Javadocs view which may contain hyperlinks which will show external pages properly.

Jenkinsfile
```diff
@@ -51,6 +51,10 @@ spec:
 
 /** Returns the download URL of the JDK against whoose C headers (in the 'include/' folder) and native libaries the SWT natives are compiled.*/
 def getNativeJdkUrl(String os, String arch){ // To update the used JDK version update the URL template below
+	//TODO: remove this workaround as soon as Temurin provides aarch64 JDKs for Windows
+	if('win32'.equals(os) && 'aarch64'.equals(arch)){
+		arch = 'x86_64'
+	}
```
chirontt, line 57: @HannesWell I don't think this "aarch64 replacement" with the x86_64 equivalent will work; it may work for the compiling step, but will certainly fail during linking due to mismatched native libs. If it does work during linking, which I very much doubt it, then the resulting `.dll` files won't work at runtime on a Windows Arm64 box.

hmartinez82: @chirontt How did you get it working on your side? Did you use Microsoft's OpenJDK?

pom.xml
```diff
@@ -170,7 +170,6 @@
                 <environment>
                   <os>win32</os>
                   <ws>win32</ws>
-                  <arch>x86_64</arch>
```
Phillipus, line 173: Does this actually work? I thought that `<arch>` was a mandatory tag? If so you'd have to declare a whole new `<environment>` entry.

Jenkinsfile
HannesWell: > @HannesWell I don't think this "aarch64 replacement" with the x86_64 equivalent will work; it may work for the compiling step, but will certainly fail during linking due to mismatched native libs. If it does work during linking, which I very much doubt it, then the resulting `.dll` files won't work at runtime on a Windows Arm64 box.

Yes. As we can see in the logs of the latest build you anticipated it right and it does not even build.
But this was just temporary to reach the the part of the pipeline where the new native agent is contacted and to test its very basic functionality (quasi a hello-world). But all I see looks very promising. We even reached to point that the native build complains that the JDK-libs don't match the architecture (the anticipated error):

```
   Creating library swt-awt-win32-4965r4.lib and object swt-awt-win32-4965r4.exp
swt_awt.obj : error LNK2019: unresolved external symbol __imp_JAWT_GetAWT referenced in function Java_org_eclipse_swt_awt_SWT_1AWT_getAWTHandle
C:\Users\genie.releng.rie8t-win11-arm\jenkins_agent\workspace\eclipse.platform.swt_PR-1045\jdk.resources\lib\jawt.lib : warning LNK4272: library machine type 'x64' conflicts with target machine type 'ARM64'
LINK : error LNK1218: warning treated as error; no output file generated
NMAKE : fatal error U1077: 'link @templrf' : return code '0x10b0'
Stop.
```
I'll continue tomorrow and adjust the pipeline to use the JDK installed at the agent (which I assume is a WoA one) for compilation.

Eventually Adoptium should provide also JDKs for WoA (not only for SWT but for WoA support of Eclipse in general)
- https://github.com/eclipse-justj/justj/issues/26
- https://github.com/adoptium/adoptium-support/issues/616

pom.xml
HannesWell: Probably not, but I'll test that when the native build is working.

Jenkinsfile
chirontt: @hmartinez82 Yes, I had used the Microsoft's OpenJDK build for Arm64 for over a year without any issues, but lately I've started using the beta version of OpenJDK Temurin build for Arm64 here: https://github.com/adoptium/temurin17-binaries/releases/tag/jdk17u-2024-02-07-14-14-beta

@HannesWell You can use the Microsoft OpenJDK build for Arm64 [here](https://learn.microsoft.com/en-us/java/openjdk/download) as a temporary solution for Arm64, while waiting for the official OpenJDK Temurin Arm64 release. But if you insist on using a Temurin OpenJDK build, you can try a beta version of it for Arm64 mentioned above.

pom.xml
chirontt: @Phillipus what you suggest is quite logical, but here's the mystery: if I add a new `<environment>` entry for the `aarch64` Windows, the build for the repo will fail (when building in either x64 and/or Arm64 boxes), and I can't work out why. Perhaps @HannesWell can help debugging it.

So eventually I came up with the current solution: remove that `<arch>x86_64</arch>` line so that the POM file is applicable for both environments.

akurtakov: I would guess that win32/win32/aarch64 triplet has to be defined in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/da0494c0d177b4e5f8d0d41a713c5359b2a2fa04/eclipse-platform-parent/pom.xml#L244 first.

HannesWell: > I would guess that win32/win32/aarch64 triplet has to be defined in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/da0494c0d177b4e5f8d0d41a713c5359b2a2fa04/eclipse-platform-parent/pom.xml#L244 first.

Good point, created https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1917 for that.

I also updated the p2.inf to require the native fragment through P2 at a WoA and the local target-config activated on Windows in a profile. With the latter I'm not sure if the change mentioned above would be necessary just for SWT. In general it probably is.

Jenkinsfile
HannesWell: > lately I've started using the beta version of OpenJDK Temurin build for Arm64 here: https://github.com/adoptium/temurin17-binaries/releases/tag/jdk17u-2024-02-07-14-14-beta

Thanks for that hint. I used that for the moment in order to verify it works and it does.
We had the first successful build of the SWT native binaries for Windows on ARM. 🎉

HannesWell:
The build of the native binaries for Windows on ARM is now working compiling against a beta jdk-17 win32.aarch64.
On Jenkins the build is also passing (there are just some new warnings that make the build unstable) and the GH-workflows just fail because there are no .dll files in the win32.aarch64 fragment. So it looks like this is shaping up well.

What's the opinion of the other committers about using that to build the SWT natives for WoA for some time until Temurin provides official releases for them? Do you think that would be a ok or should we wait until there is an official GA release for those binaries? I'm torn on this one. It is not the cleanest solution, but it would allow more testing of the WoA jdk17, which would maybe help the temurin maintainers to gain more confidence about it.

> I'm not sure if the `WebView2Loader.dll` file is needed anymore in SWT. I've read somewhere that it's no longer used in SWT, but I can't find the reference anymore.

I recently run the build without the `WebView2Loader.dll` for the existing `win32.x86_64` fragment accidentally and some test-failures happened. I have to admit I don't now why such native library isn't just shipped with the WebView2, but at the moment it seems necessary.
But of course it would be nice to get rid of it. If you are interested you can create a PR where you remove the `WebView2Loader.dll` and we'll see if we can make it work.

pom.xml
chirontt: Yes, this `p2.inf` file is the cause of my build problems of this repo for over a year! Thanks @HannesWell for the fix! Who would have thought that an (seemingly) unimportant file in an obscure corner of the repo would have such an impact on the build? With this fix for `p2.inf`, my builds for this repo in both x64 and aarch64 Windows boxes now succeed! (with that new `win32/win32/aarch64` triplet added to the pom.xml)

> I would guess that win32/win32/aarch64 triplet has to be defined in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/da0494c0d177b4e5f8d0d41a713c5359b2a2fa04/eclipse-platform-parent/pom.xml#L244 first.

Nope, this is for stand-alone build of this repo, nothing to do with the aggregator yet. In fact, my plan would be to change the aggregator repo as the very last step, after changing all submodules first for WoA (like, this SWT submodule, then equinox, eclipse.platform, etc.)

SyntevoAlex:
By "mostly identical" I mean that I'm convinced that DLLs compiled with one JDK or another JDK will work the same way.

chirontt:
>>    I'm not sure if the WebView2Loader.dll file is needed anymore in SWT. I've read somewhere that it's no longer used in SWT, but I can't find the reference anymore.
>
> But of course it would be nice to get rid of it. If you are interested you can create a PR where you remove the WebView2Loader.dll and we'll see if we can make it work.

Sorry, I take back my comment about the (no)need for `WebView2Loader.dll` file; its very requirement was added about 3 years ago to support MS Edge browser in SWT:

https://github.com/eclipse-platform/eclipse.platform.swt/blob/103b10edaf3fc07dbb69d180ac80e9e7827b5ce9/bundles/org.eclipse.swt/Eclipse%20SWT%20Browser/win32/org/eclipse/swt/browser/Edge.java#L31

So we need to add the `WebView2Loader.dll` for WoA to this PR, but I see that you've already raised a separate PR #1137 for it.

HannesWell:
> > > I'm not sure if the WebView2Loader.dll file is needed anymore in SWT. I've read somewhere that it's no longer used in SWT, but I can't find the reference anymore.
> > 
> > 
> > But of course it would be nice to get rid of it. If you are interested you can create a PR where you remove the WebView2Loader.dll and we'll see if we can make it work.
> 
> Sorry, I take back my comment about the (no)need for `WebView2Loader.dll` file; its very requirement was added about 3 years ago to support MS Edge browser in SWT:

Yes, exactly. I looked into WebView2Loader topic again yesterday but didn't have the time to answer here, yet.
I also tried to statically link the required code into the swt-win32.dll but didn't succeed (but I'm not experienced in that area and cannot even tell if that is possible). So I found no better solution than just adding the loader-dll for arm64 too.



> @HannesWell There's one for ARM64 too. We use it at work. You can download it from https://www.nuget.org/packages/Microsoft.Web.WebView2/1.0.2365.46 . The NuGet package is just a zip file, inside it you'll find `build\native\arm64\WebView2Loader.dll`

Thank you @hmartinez82 for that hint. I have now I have now added the arm64 variant of the `WebView2Loader.dll` from https://nuget.info/packages/Microsoft.Web.WebView2/1.0.1150.38 to match https://github.com/eclipse-platform/eclipse.platform.swt/pull/1137.

In order to get IP/License clearance I created
- https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/14072

Once that is approved this should be ready for submission.

chirontt:
@HannesWell Thanks for adding the `WebView2Loader.dll` to the branch. I've also added the `WebView2_LICENSE.txt` file, copied from the `x86_64` counterpart.

HannesWell:
> @HannesWell Thanks for adding the `WebView2Loader.dll` to the branch. I've also added the `WebView2_LICENSE.txt` file, copied from the `x86_64` counterpart.

That's a good catch. And that gave me the idea that it would be better to share the legal-files for the fragments of a platform in a common location. I therefore created https://github.com/eclipse-platform/eclipse.platform.swt/pull/1144.
Once that is merged, this PR can be further simplified.

chirontt:
With #1144 merged, I've rebased and updated this PR's branch to bring it up to date.

pom.xml
HannesWell: > Yes, this `p2.inf` file is the cause of my build problems of this repo for over a year! Thanks @HannesWell for the fix! Who would have thought that an (seemingly) unimportant file in an obscure corner of the repo would have such an impact on the build?

Yes sometimes it is difficult and the tiny details make the difference. I'm currently working on making that workaround obsolete but it is not yet fully ready. But I'm glad it is working now. :)

HannesWell (approved):
> With #1144 merged, I've rebased and updated this PR's branch to bring it up to date.

Thank you.
I have just pushed another small update for the Jenkins file to have the work-around for the different source of the `win32.aarch64` JDK located near the URL definition of the JDKs used by default.

Unfortunately I could not yet get more details about an estimation when Temurin will officially support WoA (see https://github.com/adoptium/adoptium-support/issues/616#issuecomment-2009123349), so it looks like we have to continue with that workaround until it is available and justj can adopted it.
But as it was discussed, it shouldn't have an impact.

> In order to get IP/License clearance to add the `WebView2Loader.dll` I created
> 
>     * https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/14072

The IP clearance was granted so everything is finally ready.

I did a final pass over the changes and all looks good to me (although I'm not an expert in configuring MSVC, but AFAICT it looks good).

Unfortunately there is currently an issue with the Eclipse build infrastructure that prevents the build from succeeding. But once that is resolved and we have a green build I think this is ready for submission.

HannesWell:
> Unfortunately there is currently an issue with the Eclipse build infrastructure that prevents the build from succeeding. But once that is resolved and we have a green build I think this is ready for submission.

The infrastructure issue has been resolved and all builds had passed.
After I submitted this PR the new natives finally have been built successfully at the master branch (the Jenkins instance had one or two hick-ups and a out-of-schedule I-build occupied a lot of resources) and SWT now has a win32.aarch64 fragment with native binaries.

Thank you @chirontt for this nice contribution and everyone that helped.

Please note that `o.e.swt.win32.aarch64` is not yet included in the `org.eclipse.e4.rcp` feature and will therefore not yet be included in the [Eclipse 4.32 I-builds update-site](https://download.eclipse.org/eclipse/updates/4.32-I-builds/):
https://github.com/eclipse-platform/eclipse.platform.ui/blob/3f00c8b2d69ac6f06a803c046a61a9a104ba5790/features/org.eclipse.e4.rcp/feature.xml#L227-L271

This should only be done once the Equinox launcher is available and and I-build for WoA is at least prepared.
Further progress can be tracked in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/577 or the linked more specific issues.

hmartinez82:
Where can one download the ARM64 binaries?

akurtakov:
At some point they should be visible in I-builds pages e.g. https://download.eclipse.org/eclipse/downloads/drops4/I20240329-0530/#SWT but not yet there. 
@HannesWell is an entry point to see what/how/where is used to publish to this page.

chirontt:
> Please note that o.e.swt.win32.aarch64 is not yet included in the org.eclipse.e4.rcp feature and will therefore not yet be included in the [Eclipse 4.32 I-builds update-site](https://download.eclipse.org/eclipse/updates/4.32-I-builds/):
https://github.com/eclipse-platform/eclipse.platform.ui/blob/3f00c8b2d69ac6f06a803c046a61a9a104ba5790/features/org.eclipse.e4.rcp/feature.xml#L227-L271

Right, they are contained in my PR (for WoA) for that `eclipse.platform.ui` repo: eclipse-platform/eclipse.platform.ui#1787

Please review it.

chirontt:
> Where can one download the ARM64 binaries?

They are now generated and stored in this repo:
https://github.com/eclipse-platform/eclipse.platform.swt/tree/master/binaries/org.eclipse.swt.win32.win32.aarch64

PR #1054, opened 2024-02-20
[gtk] Make Fonts GC disposable

laeubi:
# Current State

Currently handling fonts in SWT is hard and cumbersome, because one can't use the usual java techniques users are forced to either explicitly manage fonts (what maybe retain them longer as needed) or even create and dispose them for individual widgets (what hurts performance). This also leads to a lot of boilerplate code e.g. installing dispose handlers or having managed caches or local resource managers as well as requires documentation who 'owns' a font for example set on a label. Especially for newcommers or seldom users of SWT this is hard to get right and hinders them to use SWT correctly as java is usually managing objects automatically.

# What this changes

This now introduces the concept of [Cleaners added to Java 9+](https://docs.oracle.com/javase%2F9%2Fdocs%2Fapi%2F%2F/java/lang/ref/Cleaner.html) that allows automatic management of external resources that get automatically disposed when the JVM decides that an object is only phantom reachable and eligible for garbage collection.

I also used this to have a `Handle` act as a proxy to the native method calls so these are much more readable and the grouped together even related calls or required transformations to reduce code duplication.

# Further considerations

This itself is just meant as an initial starting point, we actually should extends this concept to any native resource handled by SWT (e.g. Images) so it become much more convenient for users to use these. I just choosing a font here as the only managed resource is a long pointer.

I also like to encourage users of other platforms (Win/Mac) to use this as a hint to prototype something similar like this as well so we can evaluate how it works out.

Also this will not replace `dispose()` or make it completely redundant as of course a user can always better dispose a resource directly if it is known that it is no longer in use. Also it might not be suitable for any "root objects" like a `Display` that is like a `Thread` in java some kind of UI root that always needs to be disposed explicitly.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Font.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Font.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2017 IBM Corporation and others.
+ * Copyright (c) 2000, 2024 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -10,6 +10,7 @@
  *
  * Contributors:
  *     IBM Corporation - initial API and implementation
+ *     Christoph Läubrich - introduce automatic resource handling for fonts
  *******************************************************************************/
 package org.eclipse.swt.graphics;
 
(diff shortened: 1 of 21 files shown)


laeubi:
I adjusted the title a bit now to make clear I don't want to **replace** the current disposal of resources, but make them GC disposable (that is for example it would be possible to hold Weak/Soft/PhantomReferences of items).

laeubi:
> could this apply to any resource, eg also Colors, Cursor, Image...?

Yes this basically could (and should) be applied everywhere we are holding a native resource that must be disposed when the object is disposed. (Colors don't need to be disposed anyways but thats a different topic).

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Font.java
```diff
@@ -35,6 +36,73 @@
...
+		public long pointer;
+
+		Handle(long handle, Font font, boolean owner) {
+			this.pointer = handle;
+			// IMPORTANT: never hold a reference to the font here! it is only used to unify
+			// the ignoreNonDisposed handling and register our handle with the cleaner!
+			if (owner) {
+				cleaner.register(font, this);
+			}
+			font.ignoreNonDisposed();
```
fedejeanne, line 55: Shouldn't this be done only if it's `owner == false` _i.e._ inside an `else` ?

akurtakov:
@laeubi Do you plan to finish this one? It already has conflicts with master.

akurtakov:
Looks like no interest in this one. Closing.

Issue #1059, opened 2024-02-22
OOME copy to Clipboard


jukzi:
Copying a lot of text for example from the Console costs much more memory then the plain ASCII Text is big, since it tries to copy formatted text and can cause OOME like

```
java.lang.OutOfMemoryError: Java heap space
	at java.base/java.util.Arrays.copyOf(Arrays.java:3541)
	at java.base/java.lang.AbstractStringBuilder.ensureCapacityInternal(AbstractStringBuilder.java:242)
	at java.base/java.lang.AbstractStringBuilder.append(AbstractStringBuilder.java:587)
	at java.base/java.lang.StringBuilder.append(StringBuilder.java:179)
	at org.eclipse.swt.dnd.HTMLTransfer.javaToNative(HTMLTransfer.java:108)
	at org.eclipse.swt.dnd.Clipboard.GetData(Clipboard.java:640)
	at org.eclipse.swt.dnd.Clipboard$1.method3(Clipboard.java:551)
	at org.eclipse.swt.internal.ole.win32.COMObject.callback3(COMObject.java:93)
	at org.eclipse.swt.internal.win32.OS.PeekMessage(Native Method)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:3655)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine$5.run(PartRenderingEngine.java:1151)
	at org.eclipse.core.databinding.observable.Realm.runWithDefault(Realm.java:339)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine.run(PartRenderingEngine.java:1042)
	at org.eclipse.e4.ui.internal.workbench.E4Workbench.createAndRunUI(E4Workbench.java:152)
	at org.eclipse.ui.internal.Workbench.lambda$3(Workbench.java:639)
	at org.eclipse.ui.internal.Workbench$$Lambda/0x0000014c901c7378.run(Unknown Source)
	at org.eclipse.core.databinding.observable.Realm.runWithDefault(Realm.java:339)
	at org.eclipse.ui.internal.Workbench.createAndRunWorkbench(Workbench.java:546)
	at org.eclipse.ui.PlatformUI.createAndRunWorkbench(PlatformUI.java:173)
	at org.eclipse.ui.internal.ide.application.IDEApplication.start(IDEApplication.java:152)
[...]

PR #1061, opened 2024-02-22
Catch OOME during copy HTML to Clipboard #1059

jukzi:
Copied HTML is much bigger then plain text. Fall back to copy at least the plain text.

https://github.com/eclipse-platform/eclipse.platform.swt/issues/1059

### DIFF
--- a/bundles/org.eclipse.swt/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.swt/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-Name: %pluginName
 Bundle-Vendor: %providerName
 Bundle-SymbolicName: org.eclipse.swt; singleton:=true
-Bundle-Version: 3.125.0.qualifier
+Bundle-Version: 3.125.100.qualifier
 Bundle-ManifestVersion: 2
 Bundle-Localization: plugin
 DynamicImport-Package: org.eclipse.swt.accessibility2
--- a/binaries/org.eclipse.swt.cocoa.macosx.aarch64/META-INF/MANIFEST.MF
+++ b/binaries/org.eclipse.swt.cocoa.macosx.aarch64/META-INF/MANIFEST.MF
@@ -1,9 +1,9 @@
 Manifest-Version: 1.0
-Fragment-Host: org.eclipse.swt;bundle-version="[3.125.0,4.0.0)"
+Fragment-Host: org.eclipse.swt;bundle-version="[3.125.100,4.0.0)"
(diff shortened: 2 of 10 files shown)


jukzi:
i have no clue how this is related or how it can be fixed: build fails with
```
Only qualifier changed for (org.eclipse.swt.gtk.linux.ppc64le/3.125.0.v20240305-1200). Expected to have bigger x.y.z than what is available in baseline (3.125.0.v20240227-1638) 
Only qualifier changed for (org.eclipse.swt.gtk.linux.x86_64/3.125.0.v20240305-1200). Expected to have bigger x.y.z than what is available in baseline (3.125.0.v20240227-1638)
```

merks:
I'm not show one coordinates the fragment's version with the host's version

image

But the error message sounded like it was complaining about a fragment's version.

bundles/org.eclipse.swt/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-Name: %pluginName
 Bundle-Vendor: %providerName
 Bundle-SymbolicName: org.eclipse.swt; singleton:=true
-Bundle-Version: 3.125.0.qualifier
+Bundle-Version: 3.125.200.qualifier
```
iloveeclipse, line 5: `0 -> 200` is wrong. Should be `0 -> 100`.
But also all fragments need to be updated.
Please do it in dedicated PR.

laeubi:
Only qualifier changed usually means you need to touch the `forceQualifier` instead of increment the version because you changed something indirectly.

iloveeclipse:
In SWT the fragment versions should match SWT version, so that is OK here.

merks:
> Only qualifier changed usually means you need to touch the `forceQualifier` instead of increment the version because you changed something indirectly.

I generally have only ever done that in features; mostly for features that include 3rd party bundles which are becoming rare.  After all the forceQualifier edit is not different than any other actual edit in the source project, right?

laeubi:
> > Only qualifier changed usually means you need to touch the `forceQualifier` instead of increment the version because you changed something indirectly.
> 
> I generally have only ever done that in features; mostly for features that include 3rd party bundles which are becoming rare. After all the forceQualifier edit is not different than any other actual edit in the source project, right? So the force is only for the first variation of the problem, i.e., compiler change but not actual other change in the bundle.

SWT is special as the actual code is not in the bundle but in the fragments, but the fragment only include specific parts, so it could happen that you change a "specialized part" and it will be picked automatic, or you change a generic part that might require manual intervention.

PR #1063, opened 2024-02-26
introduce static helper method GC#create

### LINKED ISSUE #955: Implement java.lang.AutoCloseable in class org.eclipse.swt.graphics.GC
I suggest to implement the java.lang.AutoCloseable interface in org.eclipse.swt.graphics.GC so that the try-with-resources Java feature can be used. 

Here a concrete example in org.eclipse.jface.dialogs.Dialog where the GC class is used. 
```
protected void initializeDialogUnits(Control control) {
	// Compute and store a font metric
	GC gc = new GC(control);
	gc.setFont(JFaceResources.getDialogFont());
	fontMetrics = gc.getFontMetrics();
	gc.dispose();
}
```
gc.dispose() is not called when JFaceResources.getDialogFont() throws an exception. 

Wouldn't it be much nicer if the try-with-resource approach would be used here:
```
protected void initializeDialogUnits(Control control) {
	// Compute and store a font metric
	try (GC gc = new GC(control)) {
	  gc.setFont(JFaceResources.getDialogFont());
	  fontMetrics = gc.getFontMetrics();
	}
}
```
Did you already discuss this topic? Are there any drawbacks of implementing the AutoCloseable interface in GC?

tobiasmelcher:
which takes care of creation and disposal of the GC object

Fixes: https://github.com/eclipse-platform/eclipse.platform.swt/issues/955

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/AnimatedProgress.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/AnimatedProgress.java
@@ -209,9 +209,9 @@ public synchronized void start() {
 	final Runnable [] timer = new Runnable [1];
 	timer [0] = () -> {
 		if (!active) return;
-		GC gc = new GC(AnimatedProgress.this);
-		paintStripes(gc);
-		gc.dispose();
+		try (var gc = GC.create(AnimatedProgress.this)) {
+			paintStripes(gc);
+		}
 		display.timerExec (SLEEP, timer [0]);
 	};
 	display.timerExec (SLEEP, timer [0]);
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/AnimatedProgress.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/AnimatedProgress.java
@@ -209,9 +209,10 @@ public synchronized void start() {
 	final Runnable [] timer = new Runnable [1];
 	timer [0] = () -> {
 		if (!active) return;
-		GC gc = new GC(AnimatedProgress.this);
-		paintStripes(gc);
-		gc.dispose();
+		try (AutoDisposableGC autoGC = GC.create(AnimatedProgress.this)) {
+			GC gc = autoGC.gc();
+			paintStripes(gc);
+		}
 		display.timerExec (SLEEP, timer [0]);
 	};
 	display.timerExec (SLEEP, timer [0]);
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
@@ -3835,42 +3835,25 @@ boolean updateItems() {
 }
 
 boolean updateItems (int showIndex) {
-	GC gc = new GC(this);
-	if (!single && !mru && showIndex != -1) {
-		// make sure selected item will be showing
-		int firstIndex = showIndex;
-		if (priority[0] < showIndex) {
-			int maxWidth = getRightItemEdge(gc) - getLeftItemEdge(gc, CTabFolderRenderer.PART_BORDER);
-			int width = 0;
-			int[] widths = new int[items.length];
-			for (int i = priority[0]; i <= showIndex; i++) {
-				int state = CTabFolderRenderer.MINIMUM_SIZE;
-				if (i == selectedIndex) state |= SWT.SELECTED;
-				widths[i] = renderer.computeSize(i, state, gc, SWT.DEFAULT, SWT.DEFAULT).x;
-				width += widths[i];
-				if (width > maxWidth) break;
-			}
-			if (width > maxWidth) {
-				width = 0;
-				for (int i = showIndex; i >= 0; i--) {
-					int state = CTabFolderRenderer.MINIMUM_SIZE;
-					if (i == selectedIndex) state |= SWT.SELECTED;
-					if (widths[i] == 0) widths[i] = renderer.computeSize(i, state, gc, SWT.DEFAULT, SWT.DEFAULT).x;
-					width += widths[i];
-					if (width > maxWidth) break;
-					firstIndex = i;
-				}
-			} else {
-				firstIndex = priority[0];
-				for (int i = showIndex + 1; i < items.length; i++) {
+	boolean changed = false;
+	try(var gc = GC.create(this)) {
+		if (!single && !mru && showIndex != -1) {
+			// make sure selected item will be showing
+			int firstIndex = showIndex;
+			if (priority[0] < showIndex) {
+				int maxWidth = getRightItemEdge(gc) - getLeftItemEdge(gc, CTabFolderRenderer.PART_BORDER);
+				int width = 0;
+				int[] widths = new int[items.length];
+				for (int i = priority[0]; i <= showIndex; i++) {
 					int state = CTabFolderRenderer.MINIMUM_SIZE;
 					if (i == selectedIndex) state |= SWT.SELECTED;
 					widths[i] = renderer.computeSize(i, state, gc, SWT.DEFAULT, SWT.DEFAULT).x;
 					width += widths[i];
-					if (width >= maxWidth) break;
+					if (width > maxWidth) break;
 				}
-				if (width < maxWidth) {
-					for (int i = priority[0] - 1; i >= 0; i--) {
+				if (width > maxWidth) {
+					width = 0;
+					for (int i = showIndex; i >= 0; i--) {
 						int state = CTabFolderRenderer.MINIMUM_SIZE;
 						if (i == selectedIndex) state |= SWT.SELECTED;
 						if (widths[i] == 0) widths[i] = renderer.computeSize(i, state, gc, SWT.DEFAULT, SWT.DEFAULT).x;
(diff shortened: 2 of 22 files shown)


HannesWell:
Hello @tobiasmelcher, thank you for this PR. Sorry for my late reply but I just started my almost three week of vacation when you created this PR and when I came back a full TODO list was ready to be worked off.

> Do we have to introduce the static method GC#drawOn in all platform specific GC implementations?

Yes we do, but you can just import/open the fragments for the other platforms and they will compile in your workspace. For you it is probably sufficient to only have one arch per OS since they only differ in the embedded native binaries, but the code is the same for all archs on one OS.

> I had to introduce a new Runnable interface which passes the GC as argument. Is there already an existing interface I could re-use?

Yes there is already `SWTRunnable`. But as I have just written in https://github.com/eclipse-platform/eclipse.platform.swt/issues/955#issuecomment-2016446861, I think this is not the optimal solution, which I only noticed when reviewing your changes to apply the new method (applying it immediately can therefore be very helpful 👍🏽 )

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
```diff
@@ -3835,94 +3835,95 @@ boolean updateItems() {
 }
 
 boolean updateItems (int showIndex) {
-	GC gc = new GC(this);
```
HannesWell, line 3838: I'm not sure if we should apply the new method in places where it causes so large format changes and consequently break the git-blame.
IIRC it is a relatively important source of information in SWT.
@iloveeclipse can you tell, if git-blame should be preserved.

For only a few lines it is fine, I guess.
I think the most interesting applications are outside of SWT where it is already used often in try-finally blocks.

HannesWell (changes requested):
Thanks for the update. In general it looks good, but I added some remarks below.

But the most important missing part is the implementation for the linux- and mac-platform as written in my previous comment.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/AnimatedProgress.java
```diff
@@ -195,9 +195,9 @@ public synchronized void start() {
 	final Runnable [] timer = new Runnable [1];
 	timer [0] = () -> {
 		if (!active) return;
-		GC gc = new GC(AnimatedProgress.this);
-		paintStripes(gc);
-		gc.dispose();
+		try(var gc = GC.create(AnimatedProgress.this)) {
```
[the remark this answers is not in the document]
tobiasmelcher: If I run the Format action on the given file, everything is changed. The diff is super huge and I assume that this is also not wanted. I am a little irritated here; what should I now do? Format only the changed parts of the source via Format Block?

tobiasmelcher:
Thanks a lot @HannesWell  for all the feedback. Could you please take a look and review my latest change eb1dc57 ? As already mentioned, I had some troubles with the formatting; I executed then the format action only on the changed coding parts. Is that ok?

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/GC.java
```diff
@@ -59,7 +59,22 @@
  * @see <a href="http://www.eclipse.org/swt/examples.php">SWT Examples: GraphicsExample, PaintExample</a>
  * @see <a href="http://www.eclipse.org/swt/">Sample code and further information</a>
  */
-public final class GC extends Resource {
+public sealed class GC extends Resource {
```
HannesWell, line 62: It would be good to also update the doc of this class (where it mentions `GC.dispose()`) and recommend to use the new factory, ideally with the little example also used in the Image class.
And I think Windows95 and Windows98 are not the most relevant target OS anymore. Maybe that just should be generalized.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -164,9 +164,9 @@ public final class Image extends Resource implements Drawable {
  * as shown in the following example:
  * <pre>
  *    Image i = new Image(device, width, height);
- *    GC gc = new GC(i);
- *    gc.drawRectangle(0, 0, 50, 50);
- *    gc.dispose();
+ *    try (var gc = GC.create(i)) {
+ *      gc.drawRectangle(0, 0, 50, 50));
+ *    }
```
HannesWell, line 169: Please apply this doc update also to the Image implementations for Linux and Mac.

HannesWell (commented):
> As already mentioned, I had some troubles with the formatting; I executed then the format action only on the changed coding parts. Is that ok?

Yes, that's IMO the perfect decision.

Could you please also update the doc of the `GC` class to mention the new `GC.create()` factory (for all platforms).

With that I think this change is complete (I'll mark it as ready for review), but I would like give others some time to review.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -59,7 +59,21 @@
...
+		Closeable(Drawable drawable, int style) {
+			super(drawable, style);
+		}
+		@Override
+		public void close() {
+			dispose();
+		}
+	}
+
+	public static GC.Closeable create(Drawable drawable) {
```
HannesWell, line 74: The new method needs some java-doc on all platforms. The existing single argument constructor of GC is probably a good source of inspiration.
The existing GC  constructor could also be adjusted to recommend this method for safety.

What's important is to add the tag `@since 3.126` in the Java-doc in order to make the API-tools aware this method was just added.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -59,7 +59,21 @@
  * @see <a href="http://www.eclipse.org/swt/examples.php">SWT Examples: GraphicsExample, PaintExample</a>
  * @see <a href="http://www.eclipse.org/swt/">Sample code and further information</a>
  */
-public final class GC extends Resource {
+public sealed class GC extends Resource {
+
+	public static final class Closeable extends GC implements AutoCloseable {
```
HannesWell, line 64: This new class needs some java-doc on all platforms. I think a short text that mentions that this is an auto-closable extension of the GC class is sufficient. The create method could be linked with `@see`.
What's important is to add the tag `@since 3.126` in the Java-doc in order to make the API-tools aware this method was just added.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
tobiasmelcher: Thanks for the hint. I hope it is fine with eb1dc57

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
HannesWell: I just stumbled upon this again and thought if it would be better to name this
`AutoDisposable`? Since it's an extension of `AutoClosable` that just calls `dispose()`.

laeubi: Nit-pick this does not extends `AutoClosable` (what is an interface) but implements it.

It also feels like `AutoDisposable` actually should be an interface so others can implement it as well.

And then it probabbly should become `AutoDisposableGC` here...

tobiasmelcher: renamed to `AutoDisposableGC` with [326fd48]

tobiasmelcher:
Can anyone give guidance how to solve the build error:

```
Error:  Failed to execute goal org.eclipse.tycho:tycho-apitools-plugin:5.0.2:verify (verify) on project org.eclipse.swt.gtk.linux.ppc64le: There are API errors:
Error:  The field org.eclipse.swt.graphics.GC.handle has been added to a class
Error:  META-INF/MANIFEST.MF:1 The major version should be incremented in version 3.133.0, since API breakage occurred since version 3.132.0
```

I assume that changing the final class GC to a sealed class introduces the API breakage? So, the current idea and approach will not work?

merks:
The message says `The field org.eclipse.swt.graphics.GC.handle has been added to a class` which seems bogus, so I kind of doubt the issue is related to sealed.  For starters, I would rebase this on master before anything else.

merks:
These are pretty clear and indicate you have the wrong values, so change that as well.

image

tobiasmelcher:
Should I really change to 4.0.0 ? I haven't changed GC.handle. The message irritates me. 

```
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.eclipse.tycho:tycho-apitools-plugin:5.0.2:verify (verify) on project org.eclipse.swt.cocoa.macosx.aarch64: There are API errors:
The field org.eclipse.swt.graphics.GC.handle has been added to a class
META-INF/MANIFEST.MF:1 The major version should be incremented in version 3.133.0, since API breakage occurred since version 3.132.0
org.eclipse.swt.graphics.GC.handle declared as non-API type org.eclipse.swt.internal.cocoa.NSGraphicsContext
`´`

**Comment by HeikoKlare** [comment]
> Should I really change to 4.0.0 ? I haven't changed GC.handle. The message irritates me.

There is definitely no need for bumping to 4.0.0 here. I also don't think it related to making the class sealed, as we did something similar with the `Point`/`Rectangle` classes several releases ago and did not have this kind of error: https://github.com/eclipse-platform/eclipse.platform.swt/pull/1711
It might be fine to just add an API filter to silence the error.

One thing I would like mention is that making a `final` class `sealed` will probably somehow break compatibility with consumers using Kotlin (see https://github.com/eclipse-platform/eclipse.platform.swt/pull/1711#issuecomment-2715777755). In my opinion, that's a flaw of the Kotlin language, still we might see issues being raised for SWT from Kotlin developers using SWT and instantiating a GC.

**Comment by merks** [comment]
It looks more like an API tools bug.  I can't reproduce the error in my IDE so I'm not sure what we need to add to suppress the message.    I can reproduce an error in the IDE only if I add a new public field:

image

@laeubi @HannesWell 

Any ideas?

**Comment by merks** [comment]
It might/probably would work to add a filter like this:

image

Like this:

````
    <resource path="Eclipse SWT/cocoa/org/eclipse/swt/graphics/GC.java" type="org.eclipse.swt.graphics.GC">
        <filter comment="Bogus problem" id="336658481">
            <message_arguments>
                <message_argument value="org.eclipse.swt.graphics.GC"/>
                <message_argument value="handle"/>
            </message_arguments>
        </filter>
    </resource>
```

tobiasmelcher:
Could we let the GC class implement AutoCloseable? So, there would be no need to change from final to sealed? Do you see any issues by doing it this way?

merks:
> Could we let the GC class implement AutoCloseable?

I wondered that as well.  That seems simpler to me...

HeikoKlare:
> Could we let the GC class implement AutoCloseable? So, there would be no need to change from final to sealed? Do you see any issues by doing it this way?

I would expect a bunch of resource leak warnings because of all existing creators of `GC` instances not closing them.

I am not sure if that was discussed already, but an alternative to an `AutoClosable` might be a kind of `execute` method that just takes a `Drawable` and a lambda that is executed on a temporary GC for that drawable. Then you could wrap everything inside that method instead of adapting the type hierarchy.
The `AutoClosable` solution is in my opinion definitely more elegant, so just consider it an alternative in exploring the design space which might avoid some of the drawbacks with the `AutoClosable` solution.

tobiasmelcher:
>>I am not sure if that was discussed already, but an alternative to an AutoClosable might be a kind of execute method that >>just takes a Drawable and a lambda that is executed on a temporary GC for that drawable. Then you could wrap >>everything inside that method instead of adapting the type hierarchy.

The very first proposal was implemented in this way.
https://github.com/eclipse-platform/eclipse.platform.swt/compare/d4fb76fdc63bf1d38b2b834dad679f7dc6f04e29..37519e2ade4acb878e50c8a78a9931524707854f

Heiko, I get from your comment that implementing AutoClosable in GC would lead to a big list of "resource leak warnings" which is not what we want?

HeikoKlare:
> The very first proposal was implemented in this way. https://github.com/eclipse-platform/eclipse.platform.swt/compare/d4fb76fdc63bf1d38b2b834dad679f7dc6f04e29..37519e2ade4acb878e50c8a78a9931524707854f

I see. Then I don't want to open up that discussion again.

> Heiko, I get from your comment that implementing AutoClosable in GC would lead to a big list of "resource leak warnings" which is not what we want?

Yes, every call to `new GC(...)` will give you this:
image

In the Eclipse platform repos, this would already lead to around 130 of such warnings.

akurtakov:
Is there still interest in this one? It conflicts with master.

BeckerWdf:
> Is there still interest in this one? It conflicts with master.

@tobiasmelcher  is on leave this week. Let's see what he says when he's back.

tobiasmelcher:
I now resolved the conflicts; let's see if the build is fine.

Issue #131, opened 2022-05-11
[Multi-Monitor] SWT applications to support and work with different zoom values per Shell.


niraj-modi:
Copying @SyntevoAlex comment from https://github.com/eclipse-platform/eclipse.platform.swt/issues/62#issuecomment-1116997677

To my understanding, the biggest blocker in current design is that in Windows, DPI is assigned per top-level window (Shell in terms of SWT) , whereas methods in DPIUtil do not use Shell as argument, so it uses a single DPI value for all monitors. With such design, we can only resort to supporting some scenarios at cost of other scenarios.

A proper fix would be to teach DpiUtil to consider Shell in question. Unfortunately, that will require massive (but probably not too complex) changes.

If someone volunteers to do that, I could assist to a reasonable extent.

akoch-yatta:
Hello,

I am currently working on the HiDPI issues on Windows and got most of it to a decent state. The issue I am currently struggling with is how to adjust Fonts properly, when there are different zoom levels. I think for the system fonts (font == null) I found a proper way to handle it, but when an explicit font is set, I do not see a good solution. The main problem is how the make sure, that the font is applied correctly to each control depending on the monitor zoom. Having it refreshed when the zoom change is happening is fine, but when it is set in between, e.g. as by the CSS SWT Engine, that does not work properly.

1. The obvious idea was to adapt the place, where FontData is used to create a font. In `Font.init(...)` `device.computePixels` is called, which uses `LOGPIXELSY` to calculate the scale factor. 
`
	int pixels = -(int)(0.5f + (height * OS.GetDeviceCaps(hDC, OS.LOGPIXELSY) / 72f));
`
As this uses the primary monitor, this won't work in a monitor setup with different zoom levels. As there is only one instance of `Display` even with multiple monitors I don't see a way to handle it there.

2. Second idea was to adapt `Control::setFont` to resize fonts if necessary. That does not sound great as well. When called, there is already a font created and should be set for this control and this could lead to the creation of a new one that needs to be managed. Additionally one would need to recalculate the font size taking the zoom of the primary monitor and the one of the controls monitor into consideration, which sounds rather complicated.

Anyone has another idea for an approach about that topic or am I missing a basic concept here?

HeikoKlare:
This issue has been resolved with the "monitor-specific UI scaling" feature and the according improvements done for it. With that feature, shells adapt their zoom to the current monitor:
[...]

Issue #130, opened 2022-05-11
[Multi-Monitor] Consistently fire 'SWT.ZoomChanged' event when application switches between multiple monitors.


niraj-modi:
#125 to fix the problem on Single monitor setup.

Raised this issue, to consistently fire 'SWT.ZoomChange' event when SWT application is moved between multiple monitor.

Once #125 is fixed, we need to test with multiple monitors and investigate a fix if required.

HeikoKlare:
This issue has been resolved with the "monitor-specific UI scaling" feature and the according improvements done for it. With that feature, shells adapt their zoom to the current monitor:
- https://eclipse.dev/eclipse/news/news.html?file=4.34/platform.html#rescale-on-runtime-preference

Remaining issues with that feature are tracked in this umbrella:
- https://github.com/eclipse-platform/eclipse.platform.swt/issues/1961

Issue #128, opened 2022-05-11
[Multi-Monitor] SWT should pick the right zoom value based on the monitor it's launched.


niraj-modi:
When we are in multi-monitor setup:
- Double check if SWT picks the right zoom for the monitor it's launched.

Here are some pointers:
- If required, we need to fix this in Display#getDeviceZoom()
- Additionally we may also update Shell() constructor to pick the correct zoom from the monitor it's launched.

HeikoKlare:
This issue has been resolved with the "monitor-specific UI scaling" feature and the according improvements done for it. With that feature, shells adapt their zoom to the current monitor:
- https://eclipse.dev/eclipse/news/news.html?file=4.34/platform.html#rescale-on-runtime-preference

Remaining issues with that feature are tracked in this umbrella:
- https://github.com/eclipse-platform/eclipse.platform.swt/issues/1961

Issue #125, opened 2022-05-11
Consistently fire 'SWT.ZoomChange' event on OS zoom changes with single monitor setup.


niraj-modi:
Currently SWT DPIUtil doesn't update it's caches zoom value after "SWT.ZoomChanged" event is fired.

This issues is to discuss/investigate on how to fire consistently fire 'SWT.ZoomChange' event for single monitor setup.

HeikoKlare:
This issue has been resolved with the "monitor-specific UI scaling" feature and the according improvements done for it. With that feature, shells adapt their zoom to the current monitor:
- https://eclipse.dev/eclipse/news/news.html?file=4.34/platform.html#rescale-on-runtime-preference

Remaining issues with that feature are tracked in this umbrella:
- https://github.com/eclipse-platform/eclipse.platform.swt/issues/1961

PR #1064, opened 2024-02-26
Rescaling of widgets on DPI zoom changes for basic controls (win32)

akoch-yatta:
Hello everyone,

**DISCLAIMER:** We are currently testing it in more complex application to identify blockers, that could result in changes in the PR. **BUT**, except from the ongoing testing the feature set in this PR includes everything we target for this first PR.
We used the branch nmodi/DynamicDPI_4909 as source of ideas and a starting point for our approach.

We are working on improving the HiDPI behaviour in SWT in the implementation for Windows for quite some time now together with @HeikoKlare. We try to split the adaptions up into smaller contributions to make the PRs easier to review and to discuss. This is the smallest set we managed to bundle having a usuable state of SWT for a subset of widgets. 
Due to this, the PR is not yet suitable to provide a proper scaling for the Eclipse IDE as a whole - some controls will not be scaled correctly or you will experience side effects.

**How to test**
As said above, the PR is just the first step, so testing it with the Eclipse IDE is not really a good idea (except to rule out side effect when **not** activating the feature. Best way to play around with the PR is to use the _ControlExample_ with the VM arguments: _-Dswt.autoScale.updateOnRuntime=true -Dswt.autoScale=quarter_. Still expect Glitches in the UI like cut off texts after DPI changes.

In the following I try to give a structured insight in our approach:

**Limitations:** (that will be covered separately in other PRs)
- With this PR there is still a limitation for one zoom level per application (`DPIUtil.deviceZoom`, see #131)
- Widgets in the common-package (e.g. CTabFolder) are not yet contained in this PR (except Item)
- There are multiple static attributes or attributes that e.g. are initialized on startup with the primary monitor zoom level and are currently not updated
 
**Opt-in feature configuration:**
The new behaviour is implemented as opt-in feature activated with the _swt.autoScale.updateOnRuntime_ flag. The logic how the DPI level is calculated is unchanged. In Windows using it with _-Dswt.autoScale=quarter_ could make sense, because the recent Windows versions usually propose scaling in 25% steps.

**DPI zoom update propagation:**
The whole propagation is integrated into the event handling triggered by `Control::WM_DPICHANGED`. The event triggers a listener in `Shell::handleZoomEvent` that propagates it into a custom compoanont with a lifecycle managed by `DPIZoomChangeRegistry`. We evaluated multiple ideas for that, like
- Adding a method directly inside _Widget_, that takes care of neccessary updates. This method would be needed to be overriden by sublassed to add additional update logic. Two major drawback for that were:
	- Exposing an additional method into the API of Widget (it must be public to be accessible from all neccessary loocations)
	- Unclear how to integrate this (Windows specific) behaviour into the common widgets, like CTabFolder
- Propagating the SWT event for a zoom change through each widget in the hierarchy and bind the update behaviour via listener (similar to `Shell::handleZoomEvent` for the "main" event). Two major drawback for that were:
	- How to register the listener? Doing it in the constructor leads to problems to distinguish if a listener must be added or not. If a Composite would be instantiated directly, you would need to register it, but not, when a subclass of it was instantiated. We didn't came a with a reliable solution to cover all issues regarding this approach.
	- How to structure the callbacks? One implementation per class that is individually called or having the listeners match the class hierarchy of the widget to delegate the event directly in the listener. Both approaches had not solvable drawback either caused by similar issues like in point 1 or by having unsorted handling of events.

In the end we decided to implement an independent workflow to have full control over the order how the updates are handled. The execution order is starting with the most general class in the class hierarchy, e.g. if a Composite is updated, the handlers are executed like (Widget -> Control -> Scrollable -> Composite). This way each class can provide a handler that only takes care for the attributes/method it adds, but the order is reliable and consistent.

**Fonts:**
With fonts there are two main challenges: How to scale fonts and how to make sure fonts are used in the correct scale?
_Scaling fonts_
The main issue arises from the conversion of pixels to points of a font. The conversion is dependent on the primary monitor zoom (see `Device::computePoints(LOGFONT logFont, long hFont)`). Therefor we need additional informartion to adapt this conversion to the correct target zoom. We achieve this by adding a package protected attribute `zoomLevel` to fonts, that is used to correct this conversion. According to our testing we see this approach works reliable.
_Identify the correct scaled variant of a font_
As a font (_instance_) can and is usually reused for many controls, we must assume these controls will be rendered on monitors with different zoom levels. Each font (_instance_) stores the handle that links to the font resource in the OS. In case a control is affected by a zoom level change, this change must be applied to the assigned font as well -> _Solution:_ update the font to a scaled variant and pass it to the control. This would be sufficient, if SWT would have the full control of font creation. But fonts are almost always created (and commonly cached) outside of SWT and we cannot assume, that a font is scaled to the correct size when passed to a control. If we would not consider this as well, e.g. a scaled font can be replaced by a non-scaled variant at any time. To cover both scenarios we adapt fonts in two places:
- Zoom level change of the Control will update the font (if != null) as well
- Update the font (usually via _setFont_) must check for the correct zoom level and replace it by a proper one, if necessary.

Both scenarios make use of (1) `Font:scaleFor(zoomLevel)` to derive scaled font variants from each font and (2) caching them in `ScalingFontRegistry` to be able to reuse scaled variants of a font that is used for multiple controls.

_Note:_ In default or if _swt.autoScale.updateOnRuntime_ flag is set to false DefaultSWTFontRegistry is used instead of `ScalingFontRegistry`

**Images:**
There was already automatic scaling in Image. This was adapted to cover additional issues we faced when testing. To reuse an image for multiple zoom levels at the same time, there will be additional adaptions necessary, according to our testing so far.

Any feedback is highly appreciated. We hope we can provide a proper starting point for improving the HiDPI behaviour  in Windows and are planning to continue with this topic in the following week.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
@@ -280,7 +280,7 @@ private static ImageData autoScaleImageData (Device device, final ImageData imag
  * Returns a new rectangle as per the scaleFactor.
  */
 public static Rectangle autoScaleBounds (Rectangle rect, int targetZoom, int currentZoom) {
-	if (deviceZoom == 100 || rect == null || targetZoom == currentZoom) return rect;
+	if (rect == null || targetZoom == currentZoom) return rect;
 	float scaleFactor = ((float)targetZoom) / (float)currentZoom;
 	Rectangle returnRect = new Rectangle (0,0,0,0);
 	returnRect.x = Math.round (rect.x * scaleFactor);
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
@@ -241,6 +241,11 @@ public static ImageData autoScaleImageData (Device device, final ImageData image
 	return autoScaleImageData(device, imageData, scaleFactor);
 }
 
+
+public static ImageData autoScaleImageData (Device device, final ElementAtZoom<ImageData> elementAtZoom, int targetZoom) {
+	return autoScaleImageData(device, elementAtZoom.element(), targetZoom, elementAtZoom.zoom());
+}
+
 private static ImageData autoScaleImageData (Device device, final ImageData imageData, float scaleFactor) {
 	// Guards are already implemented in callers: if (deviceZoom == 100 || imageData == null || scaleFactor == 1.0f) return imageData;
 	int width = imageData.width;
@@ -301,6 +306,10 @@ public static ImageData autoScaleUp (Device device, final ImageData imageData) {
 	return autoScaleImageData(device, imageData, 100);
 }
 
+public static ImageData autoScaleUp (Device device, final ElementAtZoom<ImageData> elementAtZoom) {
+	return autoScaleImageData(device, elementAtZoom.element(), elementAtZoom.zoom());
+}
+
 public static int[] autoScaleUp(int[] pointArray) {
 	if (deviceZoom == 100 || pointArray == null) return pointArray;
 	float scaleFactor = getScalingFactor ();
@@ -322,6 +331,14 @@ public static int[] autoScaleUp(Drawable drawable, int[] pointArray) {
 public static int autoScaleUp (int size) {
 	if (deviceZoom == 100 || size == SWT.DEFAULT) return size;
 	float scaleFactor = getScalingFactor ();
+	return Math.round (size * scaleFactor);
+}
+
+/**
+ * Auto-scale up int dimensions to match the given zoom level
+ */
+public static int autoScaleUp (int size, int zoom) {
+	float scaleFactor = getScalingFactor (zoom);
 	return Math.round (size * scaleFactor);
 }
 
@@ -402,10 +419,18 @@ public static Rectangle autoScaleUp (Drawable drawable, Rectangle rect) {
  * @return float scaling factor
  */
 private static float getScalingFactor () {
+	return getScalingFactor(deviceZoom);
+}
+
+/**
+ * Returns scaling factor from the given device zoom
+ * @return float scaling factor
+ */
+private static float getScalingFactor (int shellDeviceZoom) {
 	if (useCairoAutoScale) {
 		return 1;
 	}
-	return deviceZoom / 100f;
+	return shellDeviceZoom / 100f;
 }
 
 /**
@@ -422,12 +447,12 @@ public static int mapDPIToZoom (int dpi) {
 /**
  * Compute the DPI value value based on the zoom.
  *
- * @return zoom
+ * @return DPI
  */
-public static int mapZoomToDPI (int dpi) {
-	double zoom = (double) dpi / 100 * DPI_ZOOM_100;
-	int roundedZoom = (int) Math.round (zoom);
-	return roundedZoom;
+public static int mapZoomToDPI (int zoom) {
+	double dpi = (double) zoom / 100 * DPI_ZOOM_100;
+	int roundedDpi = (int) Math.round (dpi);
+	return roundedDpi;
 }
 
 /**
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
@@ -420,6 +420,17 @@ public static int mapDPIToZoom (int dpi) {
 }
 
 /**
+ * Compute the DPI value value based on the zoom.
+ *
+ * @return zoom
+ */
+public static int mapZoomToDPI (int dpi) {
+	double zoom = (double) dpi / 100 * DPI_ZOOM_100;
+	int roundedZoom = (int) Math.round (zoom);
+	return roundedZoom;
+}
+
+/**
  * Represents an element, such as some image data, at a specific zoom level.
  *
  * @param <T> type of the element to be presented, e.g., {@link ImageData}
@@ -491,6 +502,10 @@ private static <T> ElementAtZoom<T> getElementAtZoom(Function<Integer, T> elemen
 		}
 	}
 	return null;
+}
+
+public static int getNativeDeviceZoom() {
+	return nativeDeviceZoom;
 }
 
 public static int getDeviceZoom() {
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
@@ -44,6 +44,7 @@ public class DPIUtil {
 	private static enum AutoScaleMethod { AUTO, NEAREST, SMOOTH }
 	private static AutoScaleMethod autoScaleMethodSetting = AutoScaleMethod.AUTO;
 	private static AutoScaleMethod autoScaleMethod = AutoScaleMethod.NEAREST;
+	private static boolean autoScaleOnRuntime = false;
 
 	private static String autoScaleValue;
 	private static boolean useCairoAutoScale = false;
@@ -84,6 +85,16 @@ public class DPIUtil {
 	 * <a href="https://bugs.eclipse.org/493455">bug 493455</a>.
 	 */
 	private static final String SWT_AUTOSCALE_METHOD = "swt.autoScale.method";
+
+	/**
+	 * System property to enable to scale the applicaiton on runtime
+	 * when a DPI change is detected.
+	 * <ul>
+	 * <li>"true": the application is scaled on DPI changes</li>
+	 * <li>"false": the application will remain in its initial scaling</li>
+	 * </ul>
+	 */
+	private static final String SWT_AUTOSCALE_UPDATE_ON_RUNTIME = "swt.autoScale.updateOnRuntime";
 	static {
 		autoScaleValue = System.getProperty (SWT_AUTOSCALE);
 
@@ -95,6 +106,9 @@ public class DPIUtil {
 				autoScaleMethod = autoScaleMethodSetting = AutoScaleMethod.SMOOTH;
 			}
 		}
+
+		String updateOnRuntimeValue = System.getProperty (SWT_AUTOSCALE_UPDATE_ON_RUNTIME);
+		autoScaleOnRuntime = Boolean.parseBoolean(updateOnRuntimeValue);
 	}
 
 /**
@@ -533,6 +547,10 @@ public static int getZoomForAutoscaleProperty (int nativeDeviceZoom) {
 		zoom = Math.max ((nativeDeviceZoom + 25) / 100 * 100, 100);
 	}
 	return zoom;
+}
+
+public static boolean isAutoScaleOnRuntimeActive() {
+	return autoScaleOnRuntime;
 }
 
 /**
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Device.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Device.java
@@ -264,6 +264,10 @@ int computePixels(float height) {
 }
 
 float computePoints(LOGFONT logFont, long hFont) {
+	return computePoints(logFont, hFont, -1);
+}
+
+float computePoints(LOGFONT logFont, long hFont, int currentFontDPI) {
 	long hDC = internal_new_GC (null);
 	int logPixelsY = OS.GetDeviceCaps(hDC, OS.LOGPIXELSY);
 	int pixels = 0;
@@ -284,7 +288,14 @@ float computePoints(LOGFONT logFont, long hFont) {
 		pixels = -logFont.lfHeight;
 	}
 	internal_dispose_GC (hDC, null);
-	return pixels * 72f / logPixelsY;
+	float adjustedZoomFactor = 1.0f;
+	if (currentFontDPI > 0) {
+		// as Device::computePoints will always return point on the basis of the
+		// primary monitor zoom, a custom zoomFactor must be calculated if the font
+		// is used for a different zoom level
+		adjustedZoomFactor *= (float) logPixelsY / (float) currentFontDPI;
+	}
+	return adjustedZoomFactor * pixels * 72f / logPixelsY;
 }
 
 /**
@@ -905,6 +916,7 @@ protected void release () {
 		fontCollection = 0;
 		Gdip.GdiplusShutdown (gdipToken[0]);
 	}
+	SWTFontProvider.disposeFontRegistry(this);
 	gdipToken = null;
 	scripts = null;
 	logFonts = null;
@@ -946,5 +958,4 @@ void setEnableAutoScaling(boolean value) {
 protected int getDeviceZoom () {
 	return DPIUtil.mapDPIToZoom ( _getDPIx ());
 }
-
 }
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Font.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Font.java
@@ -15,6 +15,7 @@ package org.eclipse.swt.graphics;
 
 
 import org.eclipse.swt.*;
+import org.eclipse.swt.internal.*;
 import org.eclipse.swt.internal.win32.*;
 
 /**
@@ -49,11 +50,17 @@ public final class Font extends Resource {
 	 */
 	public long handle;
 
+	/**
+	 * The zoom in % of the standard resolution used for conversion of point height to pixel height
+	 * (Warning: This field is platform dependent)
+	 */
+	private int zoom;
 /**
  * Prevents uninitialized instances from being created outside the package.
  */
 Font(Device device) {
 	super(device);
+	this.zoom = extractZoom(this.device);
 }
 
 /**
@@ -78,6 +85,14 @@ Font(Device device) {
  */
 public Font(Device device, FontData fd) {
 	super(device);
+	this.zoom = extractZoom(this.device);
+	init(fd);
+	init();
+}
+
+private Font(Device device, FontData fd, int zoom) {
+	super(device);
+	this.zoom = zoom;
 	init(fd);
 	init();
 }
@@ -114,6 +129,7 @@ public Font(Device device, FontData[] fds) {
 	for (FontData fd : fds) {
 		if (fd == null) SWT.error(SWT.ERROR_INVALID_ARGUMENT);
 	}
+	this.zoom = extractZoom(this.device);
 	init(fds[0]);
 	init();
 }
@@ -145,6 +161,7 @@ public Font(Device device, FontData[] fds) {
 public Font(Device device, String name, int height, int style) {
 	super(device);
 	if (name == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
+	this.zoom = extractZoom(this.device);
 	init(new FontData (name, height, style));
 	init();
 }
@@ -152,6 +169,7 @@ public Font(Device device, String name, int height, int style) {
 /*public*/ Font(Device device, String name, float height, int style) {
 	super(device);
 	if (name == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
+	this.zoom = extractZoom(this.device);
 	init(new FontData (name, height, style));
 	init();
 }
@@ -195,7 +213,8 @@ public FontData[] getFontData() {
 	if (isDisposed()) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
 	LOGFONT logFont = new LOGFONT ();
 	OS.GetObject(handle, LOGFONT.sizeof, logFont);
-	return new FontData[] {FontData.win32_new(logFont, device.computePoints(logFont, handle))};
+	float heightInPoints = device.computePoints(logFont, handle, DPIUtil.mapZoomToDPI(zoom));
+	return new FontData[] {FontData.win32_new(logFont, heightInPoints)};
 }
 
 /**
@@ -218,6 +237,13 @@ void init (FontData fd) {
 	LOGFONT logFont = fd.data;
 	int lfHeight = logFont.lfHeight;
 	logFont.lfHeight = device.computePixels(fd.height);
+
+	int primaryZoom = extractZoom(device);
+	if (zoom != primaryZoom) {
+		float scaleFactor = 1f * zoom / primaryZoom;
+		logFont.lfHeight *= scaleFactor;
+	}
+
 	handle = OS.CreateFontIndirect(logFont);
 	logFont.lfHeight = lfHeight;
 	if (handle == 0) SWT.error(SWT.ERROR_NO_HANDLES);
@@ -250,6 +276,13 @@ public String toString () {
 	return "Font {" + handle + "}";
 }
 
+private static int extractZoom(Device device) {
+	if (device == null) {
+		return DPIUtil.getNativeDeviceZoom();
+	}
+	return DPIUtil.mapDPIToZoom(device._getDPIx());
+}
+
 /**
  * Invokes platform specific functionality to allocate a new font.
  * <p>
@@ -268,6 +301,7 @@ public String toString () {
  */
 public static Font win32_new(Device device, long handle) {
 	Font font = new Font(device);
+	font.zoom = extractZoom(font.device);
 	font.handle = handle;
 	/*
 	 * When created this way, Font doesn't own its .handle, and
@@ -278,4 +312,75 @@ public static Font win32_new(Device device, long handle) {
 	return font;
 }
 
+/**
+ * Invokes platform specific functionality to allocate a new font.
+ * <p>
+ * <b>IMPORTANT:</b> This method is <em>not</em> part of the public
+ * API for <code>Font</code>. It is marked public only so that it
+ * can be shared within the packages provided by SWT. It is not
+ * available on all platforms, and should never be called from
+ * application code.
+ * </p>
+ *
+ * @param device the device on which to allocate the font
+ * @param handle the handle for the font
+ * @param zoom zoom in % of the standard resolution
+ * @return a new font object containing the specified device and handle
+ *
+ * @noreference This method is not intended to be referenced by clients.
+ * @since 3.126
+ */
+public static Font win32_new(Device device, long handle, int zoom) {
+	Font font = win32_new(device, handle);
+	font.zoom = zoom;
+	return font;
+}
+
+/**
+ * Invokes platform specific private constructor to allocate a new font.
+ * <p>
+ * <b>IMPORTANT:</b> This method is <em>not</em> part of the public
+ * API for <code>Font</code>. It is marked public only so that it
+ * can be shared within the packages provided by SWT. It is not
+ * available on all platforms, and should never be called from
+ * application code.
+ * </p>
+ *
+ * @param device the device on which to allocate the font
+ * @param fontData font data to create the font for
+ * @param zoom zoom in % of the standard resolution
+ * @return a new font object using the specified font data with the
+ * specified zoom as factor for the font data
+ *
+ * @noreference This method is not intended to be referenced by clients.
+ * @since 3.126
+ */
+public static Font win32_new(Device device, FontData fontData, int zoom) {
+	return new Font(device, fontData, zoom);
+}
+
+/**
+ * Used to receive a font for the given zoom in the context
+ * of the current configuration of SWT at runtime.
+ * <p>
+ * <b>IMPORTANT:</b> This method is <em>not</em> part of the public
+ * API for <code>Font</code>. It is marked public only so that it
+ * can be shared within the packages provided by SWT. It is not
+ * available on all platforms, and should never be called from
+ * application code.
+ * </p>
+ *
+ * @param font font to create a font for the given target zoom
+ * @param targetZoom zoom in % of the standard resolution
+ * @return a font matching the specified font and zoom in %
+ *
+ * @noreference This method is not intended to be referenced by clients.
+ * @since 3.126
+ */
+public static Font win32_new(Font font, int targetZoom) {
+	if (targetZoom == font.zoom) {
+		return font;
+	}
+	return SWTFontProvider.getFont(font.getDevice(), font.getFontData()[0], targetZoom);
+}
 }
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
@@ -980,7 +980,7 @@ public void drawImage (Image image, int srcX, int srcY, int srcWidth, int srcHei
 		 * the coordinates may be slightly off. The workaround is to restrict
 		 * coordinates to the allowed bounds.
 		 */
-		Rectangle b = image.getBoundsInPixels();
+		Rectangle b = image.getBounds(deviceZoom);
 		int errX = src.x + src.width - b.width;
 		int errY = src.y + src.height - b.height;
 		if (errX != 0 || errY != 0) {
@@ -996,8 +996,7 @@ public void drawImage (Image image, int srcX, int srcY, int srcWidth, int srcHei
 
 void drawImage(Image srcImage, int srcX, int srcY, int srcWidth, int srcHeight, int destX, int destY, int destWidth, int destHeight, boolean simple) {
 	/* Refresh Image as per zoom level, if required. */
-	srcImage.refreshImageForZoom ();
-
+	srcImage.handleDPIChange(DPIUtil.getDeviceZoom());
 	if (data.gdipGraphics != 0) {
 		//TODO - cache bitmap
 		long [] gdipImage = srcImage.createGdipImage();
@@ -4387,7 +4386,7 @@ public void setFillRule(int rule) {
 public void setFont (Font font) {
 	if (handle == 0) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
 	if (font != null && font.isDisposed()) SWT.error(SWT.ERROR_INVALID_ARGUMENT);
-	data.font = font != null ? font : data.device.systemFont;
+	data.font = font != null ? Font.win32_new(font, DPIUtil.getNativeDeviceZoom()) : data.device.systemFont;
 	data.state &= ~FONT;
 }
 
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
@@ -15,6 +15,7 @@ package org.eclipse.swt.graphics;
 
 
 import java.io.*;
+import java.util.*;
 
 import org.eclipse.swt.*;
 import org.eclipse.swt.internal.*;
@@ -111,6 +112,12 @@ public final class Image extends Resource implements Drawable {
 	 * the GC which is drawing on the image
 	 */
 	GC memGC;
+
+	/**
+	 * Base image data at given zoom in % of the standard resolution. It will be used for
+	 * scaled variants of this image
+	 */
+	private ElementAtZoom<ImageData> dataAtBaseZoom;
 
 	/**
 	 * ImageFileNameProvider to provide file names at various Zoom levels
@@ -248,6 +255,7 @@ public Image(Device device, Image srcImage, int flag) {
 	this.imageFileNameProvider = srcImage.imageFileNameProvider;
 	this.styleFlag = srcImage.styleFlag | flag;
 	this.currentDeviceZoom = srcImage.currentDeviceZoom;
+	this.dataAtBaseZoom = srcImage.dataAtBaseZoom;
 	switch (flag) {
 		case SWT.IMAGE_COPY: {
 			switch (type) {
@@ -285,7 +293,7 @@ public Image(Device device, Image srcImage, int flag) {
 			break;
 		}
 		case SWT.IMAGE_DISABLE: {
-			ImageData data = srcImage.getImageDataAtCurrentZoom();
+			ImageData data = srcImage.getImageData(srcImage.currentDeviceZoom);
 			PaletteData palette = data.palette;
 			RGB[] rgbs = new RGB[3];
 			rgbs[0] = device.getSystemColor(SWT.COLOR_BLACK).getRGB();
@@ -344,7 +352,7 @@ public Image(Device device, Image srcImage, int flag) {
 			break;
 		}
 		case SWT.IMAGE_GRAY: {
-			ImageData data = srcImage.getImageDataAtCurrentZoom();
+			ImageData data = srcImage.getImageData(srcImage.currentDeviceZoom);
 			PaletteData palette = data.palette;
 			ImageData newData = data;
 			if (!palette.isDirect) {
@@ -484,7 +492,8 @@ public Image(Device device, ImageData data) {
 	super(device);
 	if (data == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
 	currentDeviceZoom = DPIUtil.getDeviceZoom ();
-	data = DPIUtil.autoScaleUp (device, data);
+	this.dataAtBaseZoom = new ElementAtZoom<>(data, 100);
+	data = DPIUtil.autoScaleUp(device, this.dataAtBaseZoom);
 	init(data);
 	init();
 }
@@ -527,6 +536,7 @@ public Image(Device device, ImageData source, ImageData mask) {
 		SWT.error(SWT.ERROR_INVALID_ARGUMENT);
 	}
 	currentDeviceZoom = DPIUtil.getDeviceZoom ();
+	this.dataAtBaseZoom = new ElementAtZoom<>(applyMask(source, ImageData.convertMask(mask)), 100);
 	source = DPIUtil.autoScaleUp(device, source);
 	mask = DPIUtil.autoScaleUp(device, mask);
 	mask = ImageData.convertMask(mask);
@@ -590,7 +600,8 @@ public Image(Device device, ImageData source, ImageData mask) {
 public Image (Device device, InputStream stream) {
 	super(device);
 	currentDeviceZoom = DPIUtil.getDeviceZoom ();
-	ImageData data = DPIUtil.autoScaleUp(device, new ImageData(stream));
+	this.dataAtBaseZoom =  new ElementAtZoom<>(new ImageData (stream), 100);
+	ImageData data = DPIUtil.autoScaleUp(device, this.dataAtBaseZoom);
 	init(data);
 	init();
 }
@@ -631,7 +642,8 @@ public Image (Device device, String filename) {
 	super(device);
 	if (filename == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
 	currentDeviceZoom = DPIUtil.getDeviceZoom ();
-	ImageData data = DPIUtil.autoScaleUp(device, new ImageData(filename));
+	this.dataAtBaseZoom = new ElementAtZoom<>(new ImageData (filename), 100);
+	ImageData data = DPIUtil.autoScaleUp(device, this.dataAtBaseZoom);
 	init(data);
 	init();
 }
@@ -720,17 +732,18 @@ public Image(Device device, ImageDataProvider imageDataProvider) {
 }
 
 /**
- * Refresh the Image based on the zoom level, if required.
+ * Update zoom and refresh the Image based on the zoom level, if required.
+ *
+ * @param deviceZoom zoom in % of the standard resolution the image shall be scaled for
  *
  * @return true if image is refreshed
  */
-boolean refreshImageForZoom () {
+boolean handleDPIChange (int deviceZoom) {
 	boolean refreshed = false;
-	int deviceZoomLevel = DPIUtil.getDeviceZoom();
 	if (imageFileNameProvider != null) {
-		if (deviceZoomLevel != currentDeviceZoom) {
-			ElementAtZoom<String> filename = DPIUtil.validateAndGetImagePathAtZoom (imageFileNameProvider, deviceZoomLevel);
-			if (filename.zoom() == deviceZoomLevel) {
+		if (deviceZoom != currentDeviceZoom) {
+			ElementAtZoom<String> filename = DPIUtil.validateAndGetImagePathAtZoom (imageFileNameProvider, deviceZoom);
+			if (filename.zoom() == deviceZoom) {
 				/* Release current native resources */
 				destroy ();
 				initNative(filename.element());
@@ -745,28 +758,27 @@ boolean refreshImageForZoom () {
 				init ();
 				refreshed = true;
 			}
-			setCurrentDeviceZoom(deviceZoomLevel);
+			setCurrentDeviceZoom(deviceZoom);
 		}
 	} else if (imageDataProvider != null) {
-		if (deviceZoomLevel != currentDeviceZoom) {
-			ElementAtZoom<ImageData> data = DPIUtil.validateAndGetImageDataAtZoom (imageDataProvider, deviceZoomLevel);
+		if (deviceZoom != currentDeviceZoom) {
+			ElementAtZoom<ImageData> data = DPIUtil.validateAndGetImageDataAtZoom (imageDataProvider, deviceZoom);
 			/* Release current native resources */
 			destroy ();
 			ImageData resizedData = DPIUtil.autoScaleImageData (device, data.element(), data.zoom());
 			init(resizedData);
 			init();
 			refreshed = true;
-			setCurrentDeviceZoom(deviceZoomLevel);
+			setCurrentDeviceZoom(deviceZoom);
 		}
-	} else {
-		if (deviceZoomLevel != currentDeviceZoom) {
-			ImageData data = getImageDataAtCurrentZoom();
+	} else if (this.dataAtBaseZoom != null) { // Resizing is only possible with a cached base image
+		if (deviceZoom != currentDeviceZoom) {
+			ImageData resizedData = getImageData(deviceZoom);
 			destroy ();
-			ImageData resizedData = DPIUtil.autoScaleImageData(device, data, deviceZoomLevel, currentDeviceZoom);
 			init(resizedData);
 			init();
 			refreshed = true;
-			setCurrentDeviceZoom(deviceZoomLevel);
+			setCurrentDeviceZoom(deviceZoom);
 		}
 	}
 	return refreshed;
@@ -1134,7 +1146,7 @@ public boolean equals (Object object) {
 	if (object == this) return true;
 	if (!(object instanceof Image)) return false;
 	Image image = (Image) object;
-	if (device != image.device || transparentPixel != image.transparentPixel) return false;
+	if (device != image.device || transparentPixel != image.transparentPixel || currentDeviceZoom != image.currentDeviceZoom) return false;
 	if (imageDataProvider != null && image.imageDataProvider != null) {
 		return (styleFlag == image.styleFlag) && imageDataProvider.equals (image.imageDataProvider);
 	} else if (imageFileNameProvider != null && image.imageFileNameProvider != null) {
@@ -1228,7 +1240,9 @@ public Rectangle getBounds() {
 }
 
 Rectangle getBounds(int zoom) {
-	Rectangle bounds = getBoundsInPixels();
+	if (isDisposed()) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
+	// Read the bounds in pixels from native layer.
+	Rectangle bounds = getBoundsInPixelsFromNative();
 	if (bounds != null && zoom != currentDeviceZoom) {
 		bounds = DPIUtil.autoScaleBounds(bounds, zoom, currentDeviceZoom);
 	}
@@ -1256,6 +1270,10 @@ public Rectangle getBoundsInPixels() {
 	if (width != -1 && height != -1) {
 		return new Rectangle(0, 0, width, height);
 	}
+	return getBoundsInPixelsFromNative();
+}
+
+private Rectangle getBoundsInPixelsFromNative() {
 	switch (type) {
 		case SWT.BITMAP:
 			BITMAP bm = new BITMAP();
@@ -1338,6 +1356,19 @@ public ImageData getImageData (int zoom) {
 	} else if (imageFileNameProvider != null) {
 		ElementAtZoom<String> fileName = DPIUtil.validateAndGetImagePathAtZoom (imageFileNameProvider, zoom);
 		return DPIUtil.autoScaleImageData (device, new ImageData (fileName.element()), zoom, fileName.zoom());
+	}
+
+	// if a GC is initialized with an Image (memGC != null), the image data must not be resized, because it would
+	// be a destructive operation. Therefor, always the current image data must be returned
+	if (memGC != null) {
+		return getImageDataAtCurrentZoom();
+	}
+	if (this.dataAtBaseZoom == null) {
+		// Cache data at base zoom before resizing it.
+		this.dataAtBaseZoom = new ElementAtZoom<>(getImageData(this.currentDeviceZoom), this.currentDeviceZoom);
+	}
+	if (this.dataAtBaseZoom != null) {
+		return DPIUtil.autoScaleImageData(device, this.dataAtBaseZoom, zoom);
 	} else {
 		return DPIUtil.autoScaleImageData (device, getImageDataAtCurrentZoom (), zoom, currentDeviceZoom);
 	}
@@ -1654,7 +1685,7 @@ public int hashCode () {
 	if (imageDataProvider != null) {
 		return imageDataProvider.hashCode();
 	} else if (imageFileNameProvider != null) {
-		return imageFileNameProvider.hashCode();
+		return Objects.hash(imageFileNameProvider, styleFlag, transparentPixel, currentDeviceZoom);
 	} else {
 		return (int)handle;
 	}
@@ -1997,6 +2028,11 @@ static long [] init(Device device, Image image, ImageData i) {
 }
 
 static long [] init(Device device, Image image, ImageData source, ImageData mask) {
+	ImageData imageData = applyMask(source, mask);
+	return init(device, image, imageData);
+}
+
+private static ImageData applyMask(ImageData source, ImageData mask) {
 	/* Create a temporary image and locate the black pixel */
 	ImageData imageData;
 	int blackIndex = 0;
@@ -2067,8 +2103,9 @@ static long [] init(Device device, Image image, ImageData source, ImageData mask
 	}
 	imageData.maskPad = mask.scanlinePad;
 	imageData.maskData = mask.data;
-	return init(device, image, imageData);
+	return imageData;
 }
+
 void init(ImageData i) {
 	if (i == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
 	init(device, this, i);
@@ -2116,7 +2153,7 @@ public long internal_new_GC (GCData data) {
 		}
 		data.device = device;
 		data.image = this;
-		data.font = device.systemFont;
+		data.font = Font.win32_new(device.getSystemFont(), DPIUtil.getNativeDeviceZoom());
 	}
 	return imageDC;
 }
@@ -2264,4 +2301,27 @@ public static Image win32_new(Device device, int type, long handle) {
 	return image;
 }
 
+/**
+ * Invokes platform specific functionality to adapt an image for
+ * the correct zoom.
+ * <p>
+ * <b>IMPORTANT:</b> This method is <em>not</em> part of the public
+ * API for <code>Image</code>. It is marked public only so that it
+ * can be shared within the packages provided by SWT. It is not
+ * available on all platforms, and should never be called from
+ * application code.
+ * </p>
+ *
+ * @param image the image to adapt for the provided zoom
+ * @param targetZoom zoom in % of the standard resolution
+ * @return an image object equal to the specified image scaled to the provided targetZoom
+ *
+ * @noreference This method is not intended to be referenced by clients.
+ */
+public static Image win32_new(Image image, int targetZoom) {
+	if (targetZoom != image.currentDeviceZoom) {
+		image.handleDPIChange(targetZoom);
+	}
+	return image;
+}
 }
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/DPIZoomChangeHandler.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/DPIZoomChangeHandler.java
@@ -1,0 +1,21 @@
+/*******************************************************************************
+ * Copyright (c) 2024 Yatta Solutions and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Yatta Solutions - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.swt.internal;
+
+import org.eclipse.swt.widgets.*;
+
+@FunctionalInterface
+public interface DPIZoomChangeHandler {
+	public void handleDPIChange(Widget widget, int newZoom, float scalingFactor);
+}
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/DPIZoomChangeRegistry.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/DPIZoomChangeRegistry.java
@@ -1,0 +1,59 @@
+/*******************************************************************************
+ * Copyright (c) 2024 Yatta Solutions and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Yatta Solutions - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.swt.internal;
+
+import java.util.*;
+import java.util.Map.*;
+
+import org.eclipse.swt.widgets.*;
+
+public class DPIZoomChangeRegistry {
+
+	private static Map<Class<? extends Widget>, DPIZoomChangeHandler> dpiZoomChangeHandlers =  new TreeMap<>(
+			(o1, o2) -> {
+	            if(o1.isAssignableFrom(o2)) {
+	            	return -1;
+	            }
+	            if(o2.isAssignableFrom(o1)) {
+	            	return 1;
+	            }
+	            return o1.getName().compareTo(o2.getName());
+	        });
+
+	/**
+	 * Calling this method will propagate the zoom change to all registered handlers that are responsible for the
+	 * class of the provided widget or one of its super classes or interfaces. Usually there will be multiple handlers
+	 * called per widget. To have a reliable and consistent execution order, the handler responsible for the most
+	 * general class in the class hierarchy is called first, e.g. if a {@code Composite} is updated, the handlers are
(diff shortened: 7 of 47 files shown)


HeikoKlare:
I want to emphasize that this PR is supposed to be an enabler for providing a complete multi-monitor HiDPI support for Windows. It aims to touch as few existing functionality as possible and to put most additional logic behind the new `swt.autoScale.updateOnRuntime` flag, to reduce the risk of regressions. This will allow to address smaller issues and features in dedicated, subsequent PRs.
However, the earlier we can merge this PR, the easier and faster the work on these subsequent issues will be. Still, this PR manifests the essential architectural additions, which is why feedback and agreement on them is rather important. Thus, I would like to encourage every interested person to give feedback to the general architecture for the HiDPI support as proposed by @akoch-yatta, in particular regarding the way the HiDPI update logic is provided and regarding the necessary addition of platform-specific API.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
```diff
@@ -394,6 +420,17 @@ private static float getScalingFactor () {
 	return deviceZoom / 100f;
```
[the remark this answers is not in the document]
akoch-yatta: Yes, makes sense. I probably missed that.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Widget.java
```diff
@@ -49,6 +49,8 @@
  * @see <a href="http://www.eclipse.org/swt/">Sample code and further information</a>
  */
 public abstract class Widget {
+
+	private int currentDeviceZoom;
```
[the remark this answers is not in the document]
akoch-yatta: You refer to the naming of the variable, right?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
```diff
@@ -379,7 +380,9 @@ public class Display extends Device implements Executor {
 	long hwndMessage, messageProc;
 
 	/* System Resources */
+	@Deprecated(since = "3.125.0")
 	LOGFONT lfSystemFont;
+	@Deprecated(since = "3.125.0")
```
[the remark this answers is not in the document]
akoch-yatta: I would love to delete that. I just wanted to avoid to remove anything non-private

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MenuItem.java
```diff
@@ -1140,8 +1144,9 @@ LRESULT wmDrawChild (long wParam, long lParam) {
 		* the item is in a menu bar.
 		*/
 		int x = (parent.style & SWT.BAR) != 0 ? MARGIN_WIDTH * 2 : struct.left;
+		int y = (parent.style & SWT.BAR) != 0 ? MARGIN_HEIGHT * 2 : struct.top + MARGIN_HEIGHT;
 		Image image = getEnabled () ? this.image : new Image (display, this.image, SWT.IMAGE_DISABLE);
-		gc.drawImage (image, DPIUtil.autoScaleDown(x), DPIUtil.autoScaleDown(struct.top + MARGIN_HEIGHT));
+		gc.drawImage (image, DPIUtil.autoScaleDown(x), DPIUtil.autoScaleDown(y));
```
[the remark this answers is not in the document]
akoch-yatta: I'll take that out. It targets a rather specific scenario. I'll create a separate PR with a reproducer for that

akoch-yatta:
> In addition to the fine-grained comments I have made, I have some general remarks:
> 
> * The propagation of DPI change events in any `Control` instance sets the static `currentDeviceZoom` of `DPIUtil` as well as the containing shell's native zoom level. What is the contract regarding these two values? I am a little confused about what I can expect from the shell's native zoom and from DPIUtil's current device zoom whenever I retrieve it. My understanding is: as soon as the DPI of any control changes, this value is updated. Is it guaranteed that it does not conflict with the pending DPI updates of all other controls?

Storing this native zoom per Shell is only required for the fonts to be scaled correctly to "real" monitor zoom levels instead of the artificial one that is stored as deviceZoom in SWT. 
As far as I know and understood the behaviour of the Windows API for the message that is handled here, it will be sent per top-window only. That should always be the Shell here. I could overwrite the WM_DPICHANGED method in Shell to have all update logic contained there instead of in Control. Although I cannot think of a real scenarios, where a conflicting update could appear.

> * I found some changes that may fix some issues independent from the goals of this PR or may even be completely unrelated. It would be great if we can extract every plain refactoring or fix of independent issues from this PR, as it will make reviewing easier. Currently, it is quite hard to see which changes are safe (do only apply when activating the newly introduced flag) and which have to be considered very thoroughly as they may affect existing functionality. I know it is a trade-off whether refactoring should only be merged with the new functionality that requires them or independent from it, but in this case I feel a little lost in reviewing everything together and feeling confident w.r.t. not producing regressions by merging the PR.

I removed those you mentioned. There are still some adaptions included besides the basic update behaviour. Handling of fonts,  calls regarding image bounds, replace theming OS calls with DPI aware alternative ot the replacement of icons of text to dpi aware alternative. Taking out the theming changes would be easy, at is one separate commit anyway, but would of course lead to weird glitches when theming is used. Taking out the other parts could be more tricky and I am unsure about the effect or the effort required for that.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/ScalingFontRegistry.java
```diff
@@ -0,0 +1,181 @@
...
+			fontHandleMap.put(scaledFont.handle, container);
+			fontKeyMap.put(scaledFont.getFontData()[0], container);
+		}
+		return scaledFont;
+	}
+
+	private int computeZoomFactor(FontData fontData) {
+		long hDC = display.internal_new_GC(null);
+		int pixelsAtPrimaryMonitorZoom = -(int) (0.5f + (fontData.height * OS.GetDeviceCaps(hDC, OS.LOGPIXELSY) / 72f));
+		display.internal_dispose_GC(hDC, null);
```
HeikoKlare, line 165: This looks like a direct code copy of what is implemented in `computePixels()`. In addition, there is quite some magic in the formula (in particular the 72), which should be explained to make it comprehensible.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Shell.java
```diff
@@ -298,8 +300,28 @@ public Shell (Display display, int style) {
...
+	int [] dpiX = new int [1];
+	int [] dpiY = new int [1];
+	int shellZoomFactor;
+	int shellNativeZoomFactor;
+	if (parent != null) {
+		shellZoomFactor = parent.getZoomFactor();
+		shellNativeZoomFactor = parent.getNativeZoomFactor();
+	} else {
+		Monitor monitor = getMonitor();
+		OS.GetDpiForMonitor (monitor.handle, 0, dpiX, dpiY);
```
HeikoKlare, line 313: This method was introduced with Windows 8.1 and will thus not return usable results for older Windows versions: https://learn.microsoft.com/en-US/windows/win32/api/shellscalingapi/nf-shellscalingapi-getdpiformonitor

You can have a look at `Display.getMonitor(long)`, which also uses that method.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Shell.java
```diff
@@ -298,8 +300,28 @@ public Shell (Display display, int style) {
...
+		int mappedDPIZoom = DPIUtil.mapDPIToZoom(dpiX[0]);
```
HeikoKlare, line 314: This looks like exactly the implementation of `Display.getMonitor(long)`, so that I would expect `getMonitor().getZoom()` to return what is calculated as `mappedDPIZoom` here. Is there any difference?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/ScalingFontRegistry.java
```diff
@@ -0,0 +1,181 @@
...
+ * API for SWT. It is marked public only so that it can be shared
+ * within the packages provided by SWT. It is not available on all
+ * platforms, and should never be called from application code.
+ * </p>
+ * @see <a href="http://www.eclipse.org/swt/">Sample code and further information</a>
+ * @noreference This class is not intended to be referenced by clients
+ */
+public class ScalingFontRegistry implements SWTFontRegistry {
+	private class ScaledFontContainer {
+		private Font baseFont;
```
HeikoKlare, line 36: I would find it helpful to have some short explanation of the characteristics of this base font. Does it have a specific scale value? It is simply the one the font was first initialized with?

HeikoKlare (commented):
In general, the changes look sound to me. Even though I still have a bunch of comments, they should be rather local. This review is almost complete, but currently excludes `Font`, `Image` and `GC`, which I will have a look at separately.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Text.java
```diff
@@ -79,6 +79,9 @@ public class Text extends Scrollable {
 	static final int IDI_SEARCH_DARKTHEME = 103;
 	static final int IDI_CANCEL_DARKTHEME = 104;
 
+	// Icon Size at default (100) zoom level
+	static int ICON_SIZE_AT_100;
```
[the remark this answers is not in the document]
akoch-yatta: That was probably used by a change I removed from this PR, I'll remove that as well

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
```diff
@@ -311,6 +327,14 @@ public static int autoScaleUp (int size) {
 	return Math.round (size * scaleFactor);
 }
 
+/**
+ * Auto-scale up int dimensions to match the given zoom level
+ */
+public static int autoScaleUp (int size, int shellDeviceZoom) {
```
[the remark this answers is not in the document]
akoch-yatta: sure, makes sense

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Font.java
```diff
@@ -195,7 +213,18 @@ public FontData[] getFontData() {
 	if (isDisposed()) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
 	LOGFONT logFont = new LOGFONT ();
 	OS.GetObject(handle, LOGFONT.sizeof, logFont);
-	return new FontData[] {FontData.win32_new(logFont, device.computePoints(logFont, handle))};
+	float heightInPoints = device.computePoints(logFont, handle);
+	int primaryZoom = DPIUtil.mapDPIToZoom (device._getDPIx());
```
HeikoKlare, line 217: I would expect this functionality to be encapsulated in the `Device` class and there is already `Device.getDeviceZoom()` that implements this. Can that method be reused (maybe by increasing visibility to be package private)?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Font.java
```diff
@@ -195,7 +213,18 @@ public FontData[] getFontData() {
...
+	float adjustedZoomFactor;
+	if (zoomFactor != primaryZoom) {
+		// as Device::computePoints will always return point on the basis of the
+		// primary monitor zoom, a custom zoomFactor must be calculated if the font
+		// is used for a different zoom level
+		adjustedZoomFactor = 1.0f * primaryZoom / zoomFactor;
+	} else {
+		adjustedZoomFactor = 1.0f;
+	}
```
HeikoKlare, line 226: This seems to be a premature optimization (assuming `zoomFactor != 0`) , or do I miss something?
```suggestion
	float adjustedZoomFactor = 1.0f * primaryZoom / zoomFactor;
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -112,6 +113,16 @@ public final class Image extends Resource implements Drawable {
...
 
+	/**
+	 * Base image data at zoom factor stored in baseDataZoomFactor
+	 */
+	private ImageData dataAtBaseZoomFactor;
+
+	/**
+	 * Zoom level for image data stored in <em>dataAtBaseZoomFactor</em>
+	 */
+	private int dataBaseZoomFactor;
```
HeikoKlare, line 124: Since these values belong together, they be encapsulated into an object/record. There is already the `ElementAtZoom` record in `DPIUtil` for this purpose. Maybe that can be reused and, e.g., moved to some more general place than the `DPIUtil` class, since that record combining some element with a zoom level seems to be of general interest in SWT.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -527,6 +542,8 @@ public Image(Device device, ImageData source, ImageData mask) {
 		SWT.error(SWT.ERROR_INVALID_ARGUMENT);
 	}
 	currentDeviceZoom = DPIUtil.getDeviceZoom ();
+	this.dataAtBaseZoomFactor =  applyMask(source, ImageData.convertMask(mask));
```
HeikoKlare, line 545: Please correct formatting of that line.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1227,8 +1260,24 @@ public Rectangle getBounds() {
...
+ *
+ * @return a rectangle specifying the image's bounds in pixel for the given zoom level.
+ *
+ * @exception SWTException <ul>
+ *    <li>ERROR_GRAPHIC_DISPOSED - if the receiver has been disposed</li>
+ *    <li>ERROR_INVALID_IMAGE - if the image is not a bitmap or an icon</li>
+ * </ul>
+ * @noreference This method is not intended to be referenced by clients.
+ */
+public Rectangle getBounds(int zoom) {
```
HeikoKlare, line 1278: By using `getBoundsInPixelsFromNative()` instead of `getBoundsInPixels()`, there is no dispose check anymore, so the defined contract for this method to throw an exception in case the receiver is disposed will probably not hold (and maybe even lead to unexpected behavior when trying to calculate the bounds for a disposed image).

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -760,15 +783,25 @@ boolean refreshImageForZoom () {
...
+				init();
+				refreshed = true;
+				currentDeviceZoom = deviceZoomLevel;
+			}
 		}
 	}
+	if (refreshed) {
+		// Reset cached width and height to -1, if the image is changed
+		width = height = -1;
+	}
```
HeikoKlare, line 804: Isn't it already a bug in the current implementatation that these values are not reset upon rescaling the image?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -760,15 +783,25 @@ boolean refreshImageForZoom () {
...
-			ImageData resizedData = DPIUtil.autoScaleImageData(device, data, deviceZoomLevel, currentDeviceZoom);
-			init(resizedData);
-			init();
-			refreshed = true;
-			currentDeviceZoom = deviceZoomLevel;
+			if (this.dataAtBaseZoomFactor == null && memGC == null) {
+				// Cache data at base zoom before refresh.
+				this.dataAtBaseZoomFactor = getImageData(currentDeviceZoom);
+				this.dataBaseZoomFactor = this.currentDeviceZoom;
+			}
```
HeikoKlare, line 790: Can you explain how this case can happen? I would expect the data for base zoom to always be initialized for an image, i.e., to be an invariant of an instance of this class.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -968,6 +968,7 @@ public void drawImage (Image image, int srcX, int srcY, int srcWidth, int srcHei
 	}
 	if (image == null) SWT.error (SWT.ERROR_NULL_ARGUMENT);
 	if (image.isDisposed()) SWT.error(SWT.ERROR_INVALID_ARGUMENT);
+	image.handleDPIChange(DPIUtil.getDeviceZoom());
```
HeikoKlare, line 971: Why is this call necessary? At the end, this method calls another method that also does the rescaling. I am concerned that this change will have an unknown impact on the "hack" applied later on in this method.

HeikoKlare (commented):
I have now also taken a look at `Font`, `Image` and `GC` and in general their changes also look sound to me.

In addition to my detailed comments, I have a general comment on tests: since this PR introduces new functionality, which is in parts also publicly available, I would expect the contracts of the new API to be tested. I would consider the following kinds of added tests relevant:
* Tests for the `SWTFontRegistry` implementations
* Tests for new public methods, such as `Font.scaleFor()`, `Image.handleDPIChange()` etc.
* At least rudimentary tests for rescaling widgets. E.g., for every kind of widget create a shell with an instance of that widget, rescale it, and check that important properties of that widget are changed (such as bounds, font size etc.). With "rudimentary" I mean that I would see priority in the tests to run through the complete rescaling logic to find potential errors (throwing exception etc.) without requiring anyone to run some SWT application with the activated `updateOnRuntime` option and having monitors with different scalings to find potential severe errors. The assertion of correct functionality would be of lower priority for me, as an incorrectly scaled font would be less severe than complete crashes of the rescaling logic leaving an application in an inconsistent scaling state.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/DPIZoomChangeHandler.java
```diff
@@ -0,0 +1,21 @@
...
+ * Contributors:
+ *     Yatta Solutions - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.swt.internal;
+
+import org.eclipse.swt.widgets.*;
+
+@FunctionalInterface
+public interface DPIZoomChangeHandler {
+	public void handleDPIChange(Widget widget, int newZoomFactor, float scalingFactor);
```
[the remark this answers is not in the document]
akoch-yatta: Reason is you never need the oldZoomFactor anywhere, but the scalingFactor. So, as a result, I would need to add the scalingFactor calculation in multiple places. But, I understand the reasoning of having two parameters with the same unit instead of two different units. I would not argue strongly against it.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/ScalingFontRegistry.java
```diff
@@ -0,0 +1,181 @@
...
+ * <p>
+ * <b>IMPORTANT:</b> This class is <em>not</em> part of the public
+ * API for SWT. It is marked public only so that it can be shared
+ * within the packages provided by SWT. It is not available on all
+ * platforms, and should never be called from application code.
+ * </p>
+ * @see <a href="http://www.eclipse.org/swt/">Sample code and further information</a>
+ * @noreference This class is not intended to be referenced by clients
+ */
+public class ScalingFontRegistry implements SWTFontRegistry {
```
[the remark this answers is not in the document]
akoch-yatta: No, that was not really intentional

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Font.java
```diff
@@ -78,6 +85,14 @@ public final class Font extends Resource {
  */
 public Font(Device device, FontData fd) {
 	super(device);
+	this.zoomFactor = extractZoomFactor(this.device);
+	init(fd);
+	init();
```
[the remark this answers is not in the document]
akoch-yatta: No, that doesn't work. device can be passed as null and this.device is null at that point, so there is will be null pointers doing that. this.device will be initialized in the Resource super constructor -> I would need to duplicate that in the extractZoomFactor method to handle the nullable device in this case, but I don't think that is the better solution

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
```diff
@@ -3407,6 +3412,16 @@ public boolean setFocus () {
...
+	checkWidget ();
+	if (font != null) {
+		if (font.isDisposed()) error(SWT.ERROR_INVALID_ARGUMENT);
+		_setFont(font.scaleFor(getShell().getNativeZoomFactor()));
+	} else {
+		_setFont(font);
+	}
+}
+
+private void _setFont (Font font) {
```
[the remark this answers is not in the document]
akoch-yatta: There are multiple usages of this pattern in SWT, I just adapted to them, e.g. Control::_setBackground or Button::_setImage. 
But, for this place, you are indeed correct, that I don't need a separate method anymore. Over some refactorings in the past weeks, other usages where removed, so I will inline it again.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
```diff
@@ -2445,19 +2453,21 @@ public Cursor getSystemCursor (int id) {
  */
 @Override
 public Font getSystemFont () {
+	return getSystemFont(getPrimaryMonitor().getZoom());
```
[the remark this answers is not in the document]
akoch-yatta: As the passed parameter won't be used from the default font factory anyway, this would be only an issue when the "swt.autoScale.updateOnRuntime" flag is set to true, which should not be possible for Windows 7 or 8 anyway due to most of the dpi dependent OS calls that were introduced with Windows 10.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Shell.java
akoch-yatta: see below

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Shell.java
akoch-yatta: No, not anymore, you are correct. I replace it with getMonitor().getZoom()

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Font.java
akoch-yatta: i can directly use it. after some change I ended up with the same implementation as getDeviceZoom without noticing it.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: This happens when `Image::win32_new(...)` is used with a given handle, e.g. in `Display::getSystemImage`

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
akoch-yatta: actually, it is only necessary because of that hack later. But having a second look on it, I was able to simplify the whole call by replacing `getBoundsInPixels()` with `getBounds(deviceZoom) `

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: But that is the method where the image is rescaled. Where would you expect the reset of width and height?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: So adding a dispose check there should be sufficient?

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
```diff
@@ -95,6 +106,11 @@ private static enum AutoScaleMethod { AUTO, NEAREST, SMOOTH }
 				autoScaleMethod = autoScaleMethodSetting = AutoScaleMethod.SMOOTH;
 			}
 		}
+
+		String updateOnRuntimeValue = System.getProperty (SWT_AUTOSCALE_UPDATE_ON_RUNTIME, "false");
+		if (Boolean.TRUE.toString().equalsIgnoreCase(updateOnRuntimeValue)) {
+			autoScaleOnRuntime = true;
+		}
```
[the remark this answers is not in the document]
akoch-yatta: That is not in the API of Boolean, isn't it?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/ScalingFontRegistry.java
akoch-yatta: It is the first one to be initialized, so usually it will always be scaled to the primary monitor zoom, unless its zoom is changed and a completely new font is created and used.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/ScalingFontRegistry.java
```diff
@@ -0,0 +1,181 @@
...
+		if (fetchSystemParametersInfo(info, targetZoomFactor)) {
+			LOGFONT logFont = info.lfMessageFont;
+			hFont = OS.CreateFontIndirect(logFont);
+		}
+		if (hFont == 0)
+			hFont = OS.GetStockObject(OS.DEFAULT_GUI_FONT);
+		if (hFont == 0)
+			hFont = OS.GetStockObject(OS.SYSTEM_FONT);
+		return hFont;
+	}
```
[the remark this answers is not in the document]
akoch-yatta: Hmm, yes sure. I am just not sure, where to put that common code. Any recommendations?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/ScalingFontRegistry.java
akoch-yatta: My fault. I was confused, why I couldn't used `computePixels(...)` due to visibility reasons, but could just do so by casting the `Display `to `Device`

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: I would like to postpone that to our PR when we refactored the Image to work properly with multiple zoom levels in parallel.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/CoolItem.java
```diff
@@ -533,21 +533,25 @@ void setSizeInPixels (int width, int height) {
...
+
 	if ((parent.style & SWT.VERTICAL) != 0) {
 		cx = height;
 		cyChild = width;
+		rbBand.cyMaxChild = Math.max(rbBand.cyMaxChild, width);
 		cxIdeal = Math.max (0, height - parent.getMargin (index));
 	} else {
 		cx = width;
 		cyChild = height;
+		rbBand.cyMaxChild = Math.max(rbBand.cyMaxChild, height);
```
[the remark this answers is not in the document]
akoch-yatta: I removed it for this PR. It is discarding some minor rescaling in some scenarios, but I couldn't come up with a good scenario immediately, so we can postpone it.

akoch-yatta:
> I have now also taken a look at `Font`, `Image` and `GC` and in general their changes also look sound to me.
> 
> In addition to my detailed comments, I have a general comment on tests: since this PR introduces new functionality, which is in parts also publicly available, I would expect the contracts of the new API to be tested. I would consider the following kinds of added tests relevant:
> 
> * Tests for the `SWTFontRegistry` implementations
> * Tests for new public methods, such as `Font.scaleFor()`, `Image.handleDPIChange()` etc.
> * At least rudimentary tests for rescaling widgets. E.g., for every kind of widget create a shell with an instance of that widget, rescale it, and check that important properties of that widget are changed (such as bounds, font size etc.). With "rudimentary" I mean that I would see priority in the tests to run through the complete rescaling logic to find potential errors (throwing exception etc.) without requiring anyone to run some SWT application with the activated `updateOnRuntime` option and having monitors with different scalings to find potential severe errors. The assertion of correct functionality would be of lower priority for me, as an incorrectly scaled font would be less severe than complete crashes of the rescaling logic leaving an application in an inconsistent scaling state.

Definitely valid points. we will have a look at separate tests for all mentioned components. Still, I want to raise the point, whether we want to wait for those tests with this PR or contribute them with a separate PR afterwards. As a lot more contributions are waiting for this PR, I would like to avoid to many changes to pile up locally.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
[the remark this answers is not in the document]
HeikoKlare: Sorry for te confusion, the correct suggestion is as follows.
```suggestion
		autoScaleValue = Boolean.parseBoolean(updateOnRuntimeValue);
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Font.java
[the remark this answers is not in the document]
HeikoKlare: I see. Thanks for the explanation.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Font.java
HeikoKlare: Sounds to me as if there is something broken in general:
- `Display` overwrites that method even though `Device` annotates that method as `@nooverride`
- `Device.getDeviceZoom()` does not seem to be used at all
- `Display` reimplemented the method, but the code here uses the `Device` implementation to retrieve the zoom. So if there is a need to retrieve the value as calulcated in `Device`, there probably should be some (additional) method in `Device` providing that value even if you have an instance of `Display`.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
HeikoKlare: Might be okay, but just to properly understand: is there a specific reason to postpone it and not do it now?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
HeikoKlare: Okay, shouldn't/couldn't this functionality then be part of `getImageData(int)`?
And why is it okay that nothing done when `dataAtBaseZoomFactor == null` and `memGC != null`?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
HeikoKlare: I rephrase my question: why is resetting `width` and `height` necessary when the image is refreshed now (i.e., after this PR) and was not necessary before? My understanding is that this method still initialized the image with data for a different zoom factor, like before. So the `width` and `height` will usually change and are recalculated upon calling `getBoundsInPixels()`.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
HeikoKlare: Yes, I think so.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/DPIZoomChangeHandler.java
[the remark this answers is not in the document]
HeikoKlare: I understand, then I would expect
1. a documentation of that method with an explanation of its parameters,
2. a proper naming to distinguish the two kinds of "factors". Even though I know we have discussed something different before: DPIUtil always simply uses "zoom" for zoom factors, so maybe that's the common term to distinguish from actual "factors"?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/ScalingFontRegistry.java
HeikoKlare: I actually meant to find it helpful to have this information _in the code_, to ease comprehension when making modifications to the code :-)

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1338,6 +1392,9 @@ public ImageData getImageData (int zoom) {
 	} else if (imageFileNameProvider != null) {
 		ElementAtZoom<String> fileName = DPIUtil.validateAndGetImagePathAtZoom (imageFileNameProvider, zoom);
 		return DPIUtil.autoScaleImageData (device, new ImageData (fileName.element()), zoom, fileName.zoom());
+	}
+	else if (this.dataAtBaseZoomFactor != null) {
```
HeikoKlare, line 1396: nitpicky: formatting issue
```suggestion
	} else if (this.dataAtBaseZoomFactor != null) {
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: Ah okay, sorry I missed your point before. I need to have a short look at it. I don't remember the scenario out of my head

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: Answer to the last question: we discussed that some weeks ago with the PaintExample. In this case we will have destructive modifications, If we would scale it. That ist why we excluded it for now.

HeikoKlare:
> Still, I want to raise the point, whether we want to wait for those tests with this PR or contribute them with a separate PR afterwards.

With respect to the tests for widget rescaling, it would be okay for me to have them in a follow-up PR. For the existing API extensions, however, I would be in favor of having tests now, as we do not know when that functionality may be affected by other changes and then we want to find potential regressions immediately and not only once tests are added.

HeikoKlare:
I propose to merge this PR as soon as the review comments have been addressed to allow @akoch-yatta to provide further PRs to contribute and finalize multi-monitor HiDPI support on Windows based on this initial PR.
In case anyone has concerns w.r.t. this PR, and in particular to the architecture for providing multi-monitor HiDPI support, please let us know until Wednesday (27th March). In case remaining improvements have been finalized until then, we plan to merge this afterwards.

@niraj-modi @iloveeclipse Maybe one of you as Windows code owners has an opinion on these changes?

This PR is also related to issues #62, #125, #128, #130. #131 and we should check and document how they may be affected or even solved.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: As discussed, I moved the caching into `getImageData(int)` and refactored the caller a bit

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Device.java
```diff
@@ -264,6 +265,15 @@ int computePixels(float height) {
...
 float computePoints(LOGFONT logFont, long hFont) {
+	return computePoints(logFont, hFont, -1);
+}
+
+float computePoints(LOGFONT logFont, long hFont, int currentFontZoomFactor) {
+	float fontDPI = DPIUtil.mapZoomToDPI(currentFontZoomFactor);
+	return computePoints(logFont, hFont, fontDPI);
+}
+
+private float computePoints(LOGFONT logFont, long hFont, float currentFontDPI) {
```
HeikoKlare, line 276: Shouldn't `currentFontDPI` be an integer? In particular, `DPIUtil.mapZoomToDPI()`, which delivers the value passed to this method, also returns an integer.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -758,11 +771,10 @@ boolean refreshImageForZoom () {
 			refreshed = true;
 			setCurrentDeviceZoom(deviceZoomLevel);
 		}
-	} else {
+	} else if (this.dataAtBaseZoom != null) { // Resizing is only possible with a cached base image
```
fedejeanne, line 774: There are 2 constructors that do **not** initialize `dataAtBaseZoom` (you can see them by marking the field as `final`). Have those 2 cases been tested?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Caret.java
```diff
@@ -148,7 +152,8 @@ public Font getFont () {
 	checkWidget();
 	if (font == null) {
 		long hFont = defaultFont ();
-		return Font.win32_new (display, hFont);
+		final Font newFont = Font.win32_new (display, hFont, getZoom());
+		return newFont;
```
fedejeanne, line 156: Why create it (even as `final`) just to return it in the next line? Why not simply:

```suggestion
		return Font.win32_new (display, hFont, getZoom());
```

?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/DPIZoomChangeRegistry.java
```diff
@@ -0,0 +1,47 @@
...
+	            if(o1.isAssignableFrom(o2)) {
+	            	return -1;
+	            }
+	            if(o2.isAssignableFrom(o1)) {
+	            	return 1;
+	            }
+	            return o1.getName().compareTo(o2.getName());
+	        });
+
+	public static <E> void applyChange(Widget widget, int newZoom, float scalingFactor) {
```
fedejeanne, line 34: Add the JavaDoc here and mention explicitly that this method may end up invoking several `DPIZoomChangeHandler`s (mention the order in which they will be invoked too). This is an important detail.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/CoolBar.java
```diff
@@ -1198,4 +1199,47 @@ LRESULT wmNotifyChild (NMHDR hdr, long wParam, long lParam) {
 	}
 	return super.wmNotifyChild (hdr, wParam, lParam);
 }
+
+private static void handleDPIChange(Widget widget, int newZoom, float scalingFactor) {
```
fedejeanne, line 1203: I don't think the usage of `var` on the variable of this method has any advantages. In fact it only makes it a bit more difficult to know the type of the variables.
IMO `var` should be used only where the type doesn't really play a role but these variables are either being passed to other methods or used inside calculations inside this method, so it's useful to know at first glance of which type the are. Additionally, it introduces another risk: imagine someone changes the return type of a method that lands in one of these `var`-variables (say `CoolBar::getWrapIndices`) and there is an overloaded method which happens to work with that new type. If that happens then you will inadvertently end up calling another method.

tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/tests/win32/graphics/ImageWin32Tests.java
```diff
@@ -0,0 +1,62 @@
...
+import org.junit.Before;
+import org.junit.Test;
+
+/**
+ * Automated Tests for class org.eclipse.swt.graphics.Image
+ * for Windows specific behavior
+ *
+ * @see org.eclipse.swt.graphics.Image
+ */
+public class ImageWin32Tests {
```
fedejeanne, line 34: This actually only runs in Windows without any further changes (like using `Assume`)?

fedejeanne (changes requested):

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/DefaultSWTFontRegistry.java
```diff
@@ -0,0 +1,82 @@
...
+/**
+ * This class is used in the win32 implementation only to support
+ * scaling of fonts in multiple DPI zoom level.
+ */
+public final class DefaultSWTFontRegistry implements SWTFontRegistry {
+	private static FontData KEY_SYSTEM_FONTS = new FontData();
+	private Map<FontData, Font> fontsMap = new HashMap<>();
+	private Device device;
+
+	public DefaultSWTFontRegistry(Device device) {
```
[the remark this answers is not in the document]
akoch-yatta: Is there a convention for that? I didn't find that at any class in the internal package. But I don't mind adding it.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/DefaultSWTFontRegistry.java
```diff
@@ -0,0 +1,82 @@
...
+import java.util.*;
+
+import org.eclipse.swt.graphics.*;
+import org.eclipse.swt.internal.win32.*;
+
+/**
+ * This class is used in the win32 implementation only to support
+ * scaling of fonts in multiple DPI zoom level.
+ */
+public final class DefaultSWTFontRegistry implements SWTFontRegistry {
```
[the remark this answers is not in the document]
akoch-yatta: see above

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: Yes, but I will have a look at possible automated tests for it

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/DefaultSWTFontRegistry.java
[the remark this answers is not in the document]
HeikoKlare: I am not sure what the semantics of a `@noreference` within an internal package should be. Clients cannot reference that class/method anyway.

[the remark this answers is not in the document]
fedejeanne: That's true, I was aiming for something else though: no one (not even internal code) should instantiate this class without going through the factory-method `org.eclipse.swt.internal.SWTFontProvider.newFontRegistry(Device)`. One exception to this rule are the tests (`org.eclipse.swt.internal.DefaultSWTFontRegistryTests`).

You could achieve this by:
1. Make the constructor package-protected 
2. (Optionally) Mark the **class** with `@noinstantiate` or mark the constructor with `@noreference`. In both cases you should mention the factory-method in the JavaDoc

If point nr. 2. is an overkill, at least 1. should be used.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/DefaultSWTFontRegistry.java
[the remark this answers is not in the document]
fedejeanne: Granted, this one might be an overkill: it should be well known after 50+ years of OOP that one programs against interfaces and not its subclasses, he. Scratch my suggestion :-)

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/DefaultSWTFontRegistry.java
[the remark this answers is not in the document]
fedejeanne: > > 1. Making the constructor package-protected
> 
> This is basically reasonable, but due to the lack of possibility to write proper unit tests by test fragments, making the constructor `public` is is the trade-off to achieve testability here.

Good point: reducing the visibility lets the code still compile but the tests fail (at least locally they do) 👍 



> This would only make sense as a "workaround" for proposal 1 not being possible. But in general, even though it does not make sense to instantiate the registry, I do not see a reason from a design/architectural perspective to disallow the instantiation.

The reason would be "Defensive programming": if you don't need it, don't offer it. If you offer it, you have to maintain it. Also, the registry has to be disposed at some point to avoid resource leaking. The `SWTFontProvider` takes care of that. If this is going to be left for the user-code to do then at least it should be mentioned that the `dispose` method needs to be called when instantiating the class directly.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: I had a look a the existing tests and would say, that the `test_getImageData_XXX `Tests in `Test_org_eclipse_swt_graphics_Image `cover most of it. I added one test for an additional constructor.

fedejeanne: Hm, those tests don't walk into the changed `else`-block and neither does the new test.

I also debugged all tests in the package and none of them reach the `else` (neither in your branch nor in the `master` branch). Alos, lots of them fail when I run them locally, maybe I'm doing something wrong? 

Could you please double check that:
1. Any test is reaching the `else`-block before applying your PR and
2. The test(s) that reached the `else` clock in `master` also pass in your branch?

This would at least mean that adding the extra condition to the else didn't break anything (or at least increase the chances that nothing broke because of the change).

fedejeanne (commented):
Requested changes have been made, thank you!

fedejeanne (approved):
LGTM

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: Just to summarize our discussion: as this behaviour is windows specific, I added a test in `ImageWin32Tests`, that is running into this block. In the current master, I didn't find a proper and real scenario, where I could run into this block.

tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/tests/win32/graphics/ImageWin32Tests.java
fedejeanne: FTR it does. OS-specific tests are divided in different projects: 
* `org.eclipse.swt.tests.win32`
* `org.eclipse.swt.tests.gtk`
* `org.eclipse.swt.tests.cocoa`

fedejeanne:
Thank you @akoch-yatta and @HeikoKlare for implementing, reviewing and testing this! This will be a very important stepping stone for future work 💪 

@akurtakov would you like to give this a final review before it gets merged? We plan to merge next Tuesday so it's in time for M2. Any comments on the overall architecture of this PR is also very much welcome :-)

Of course, others are invited to review and comment on this PR too.

akurtakov:
Unfortunately, my win32 port knowledge is close to zero and I don't even have access to windows to try it. As per https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/CODEOWNERS it's @niraj-modi and @iloveeclipse that should be able to do the review for you.

fedejeanne:
Thank you @akurtakov for the hint!

@niraj-modi / @iloveeclipse , it would be great if you could give us your input.

We would like to merge this PR on Tuesday next week and make it to the M2 deadline. If there's something we can do to make it easier for you to review/test, please let us know. It might be that we are "too close" to the issue and we missed something 😅

iloveeclipse:
I'm neither expert in HiDPI nor Windows API's.
@SyntevoAlex is a far better candidate for a reviewer role here.

fedejeanne:
> @SyntevoAlex is a far better candidate for a reviewer role here.

Thank you for the hint! @SyntevoAlex , [my offer](https://github.com/eclipse-platform/eclipse.platform.swt/pull/1064#issuecomment-2066347415) stands :-)

> We would like to merge this PR on Tuesday next week and make it to the M2 deadline. If there's something we can do to make it easier for you to review/test, please let us know. It might be that we are "too close" to the issue and we missed something 😅

SyntevoAlex:
Unfortunately I do not have time to review/handle this.
The sad truth is, SWT is currently super low on active maintainers. There's around zero for Windows.

akurtakov:
@fedejeanne From all previous replies on the topic - you seem to be the most qualified and active SWT on windows contirbutor so it's up to you when/what/how to merge.

fedejeanne:
> @fedejeanne From all previous replies on the topic - you seem to be the most qualified and active SWT on windows contirbutor so it's up to you when/what/how to merge.

Then we are in deep trouble, @akurtakov 😅 (see the last sentence [here](https://github.com/eclipse-platform/eclipse.platform.swt/discussions/1180#discussion-6519789))

HeikoKlare:
To clarify what we have done so far to ensure that this is valuable but also "safe" contribution (i.e., not producing any regressions) even with few to no experienced contributors available for deeper review:
* Stripped down this initial contribution to just set the basis for providing holistic multi-monitor HiDPI support by follow-up contributions, so that as few existing code is affected as possible by this initial PR.
* Placed most of the added code behind a [release feature toggle](https://martinfowler.com/articles/feature-toggles.html) to not be executed at all unless the added ini flag (serving as the feature toggle) is activated.
* Thoroughly reviewed in particular those changes affecting existing code (and being executed even with the feature toggle turned off).
* Manually tested this contribution together with the contributions to be made as follow-ups in an SDK product and also in our RCP-based product.

Our query for reviews particularly aimed at giving everyone enough chance to comment on the overall architecture of how multi-monitor HiDPI support will be provided for Windows. If everyone is fine with trusting us (in particular @akoch-yatta) in doing our attempt to provide multi-monitor HiDPI support properly and thoroughly, that's of course fine for us as well.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Font.java
```diff
@@ -49,11 +50,17 @@ public final class Font extends Resource {
 	 */
 	public long handle;
 
+	/**
+	 * The zoom in % of the standard resolution used for conversion of point height to pixel height
+	 * (Warning: This field is platform dependent)
+	 */
+	int zoom;
```
HeikoKlare, line 57: I don't see any access to this field from outside the class.
```suggestion
	private int zoom;
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Font.java
```diff
@@ -278,4 +312,74 @@ public static Font win32_new(Device device, long handle) {
...
+ * <b>IMPORTANT:</b> This method is <em>not</em> part of the public
+ * API for <code>Font</code>. It is marked public only so that it
+ * can be shared within the packages provided by SWT. It is not
+ * available on all platforms, and should never be called from
+ * application code.
+ * </p>
+ *
+ * @param font font to create a font for the given target zoom
+ * @param targetZoom zoom in % of the standard resolution
+ * @return a new font object containing the specified device and handle
```
HeikoKlare, line 374: There is no handle specified -> copy&paste mistake?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Font.java
```diff
@@ -278,4 +312,74 @@ public static Font win32_new(Device device, long handle) {
...
+ * API for <code>Font</code>. It is marked public only so that it
+ * can be shared within the packages provided by SWT. It is not
+ * available on all platforms, and should never be called from
+ * application code.
+ * </p>
+ *
+ * @param device the device on which to allocate the font
+ * @param fontData font data to create the font for
+ * @param zoom zoom in % of the standard resolution
+ * @return a new font object containing the specified device and handle
```
HeikoKlare, line 352: There is no handle specified -> copy&paste mistake?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/DPIZoomChangeRegistry.java
```diff
@@ -0,0 +1,59 @@
...
+	 * Calling this method will propagate the zoom change to all registered handlers that are responsible for the
+	 * class of the provided widget or one of its super classes or interfaces. Usually there will be multiple handlers
+	 * called per widget. To have a reliable and consistent execution order, the handler responsible for the most
+	 * general class in the class hierarchy is called first, e.g. if a {@code Composite} is updated, the handlers are
+	 * executed like ({@code Widget} -> {@code Control} -> {@code Scrollable} -> {@code Composite}). Each handler
+	 * should only take care to update the attributes the class, it is registered for, adds to the hierarchy.
+	 *
+	 * @param widget widget the zoom change shall be applied to
+	 * @param newZoom zoom in % of the standard resolution to be applied to the widget
+	 * @param scalingFactor
```
HeikoKlare, line 44: Empty tag

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -720,13 +732,14 @@ public Image(Device device, ImageDataProvider imageDataProvider) {
...
 /**
- * Refresh the Image based on the zoom level, if required.
+ * Update zoom and refresh the Image based on the zoom level, if required.
+ *
+ * @param deviceZoomLevel zoom level the image shall be scaled for
  *
  * @return true if image is refreshed
  */
-boolean refreshImageForZoom () {
+boolean handleDPIChange (int deviceZoomLevel) {
```
HeikoKlare, line 741: nit: this seems to be the only method using the term "zoomLevel" instead of "zoom"

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/SWTFontRegistry.java
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * This class is used in the win32 implementation only to support
+ * re-usage of fonts.
+ * <p>
+ * <b>IMPORTANT:</b> This class is <em>not</em> part of the public
+ * API for SWT. It is marked public only so that it can be shared
+ * within the packages provided by SWT. It is not available on all
+ * platforms, and should never be called from application code.
+ * </p>
+ * @see <a href="http://www.eclipse.org/swt/">Sample code and further information</a>
```
HeikoKlare, line 27: Is this `@see` tag intended?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/SWTFontRegistry.java
```diff
@@ -0,0 +1,54 @@
...
+	 * are managed by the font registry and should not be disposed of externally.
+	 *
+	 * @param fontData the data used to create the font
+	 * @param zoom zoom in % of the standard resolution to determine the appropriate font
+	 * @return the font best suited for the specified zoom
+	 */
+	Font getFont(FontData fontData, int zoom);
+
+	/**
+	 * Disposes of all fonts managed by the font registry.
```
HeikoKlare, line 51:
```suggestion
	 * Disposes all fonts managed by the font registry.
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Shell.java
```diff
@@ -298,8 +300,24 @@ public Shell (Display display, int style) {
...
+		shellZoom = DPIUtil.getZoomForAutoscaleProperty(mappedDPIZoom);
+		shellNativeZoom = mappedDPIZoom;
+	}
+	this.setZoom(shellZoom);
+	this.setNativeZoom(shellNativeZoom);
+
 	reskinWidget();
 	createWidget ();
+
+	addListener(SWT.ZoomChanged, this::handleZoomEvent);
```
HeikoKlare, line 320: nit: wouldn't it be better to add the listener only when `autoScaleOnRuntime` is true instead of checking that in the `handleZoomEvent` method?

tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/internal/ScalingSWTFontRegistryTests.java
```diff
@@ -0,0 +1,94 @@
...
+		assertEquals("Pixel height must be doubled between 100% and 200% zoom factor", heightFont100 * 2, heightFont200);
+	}
+
+	@Test
+	public void fontsAreCached() {
+		FontData fontData = new FontData(TEST_FONT, 10, SWT.NORMAL);
+		Font font100_1 = fontRegistry.getFont(fontData, 100);
+		FontData fontData2 = new FontData(TEST_FONT, 10, SWT.NORMAL);
+		Font font100_2 = fontRegistry.getFont(fontData2, 100);
+		assertTrue("Fonts for same font data and zoom factor must be reused", font100_1 == font100_2);
```
HeikoKlare, line 68: nit (not necessary to be fixed in this PR): `assertSame()` (and `assertNotSame()` at other places) would produce better failure output

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2116,7 +2153,7 @@ public long internal_new_GC (GCData data) {
 		}
 		data.device = device;
 		data.image = this;
-		data.font = device.systemFont;
+		data.font = Font.win32_new(device.getSystemFont(), this.currentDeviceZoom);
```
HeikoKlare, line 2156: Just for my understanding: at most places fonts are rescaled to the native device zoom. Why is the current device zoom used in this case?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/DefaultSWTFontRegistry.java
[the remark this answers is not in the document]
HeikoKlare: Anything left to do here?

HeikoKlare (commented):
Thanks for the recent changes, @akoch-yatta! I have taken a final look at them and only have minor comments that should be easy and fast to answer and/or address. All the changes look still as sound to me as in the previous review. My review focus was on finding potential regressions in existing code. Added functionality is kept internal as far as possible, so even if follow-up PRs require adaptations, they should be easy to realize as we do not fix too much public API with this PR.

I have tested the changes manually, also with a focus on finding regressions in existing functionality. I did not face any regressions yet, neither in a "default" setup with 100% monitor scaling, nor with other monitor scalings, both with `-Dswt.autoScale=quarter` and also with the flag unset (i.e., with `-Dswt.autoScale=integer200`). In particular, I have compared the l look (in particular the sizes and sharpness) of UI elements with and without this PR and they look exactly the same.

So, as soon as my final remarks are processed, there is a go from my side for merging this PR.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
```diff
@@ -44,6 +44,7 @@ public class DPIUtil {
 	private static enum AutoScaleMethod { AUTO, NEAREST, SMOOTH }
 	private static AutoScaleMethod autoScaleMethodSetting = AutoScaleMethod.AUTO;
 	private static AutoScaleMethod autoScaleMethod = AutoScaleMethod.NEAREST;
+	public static boolean autoScaleOnRuntime = false;
```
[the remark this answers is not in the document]
akoch-yatta: No, wie need it in some places (e.g. SWTFontProvider) to know, whether this is activated or not. Als alternative I could make the field private and make it only readable via getter.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Font.java
```diff
@@ -218,6 +237,13 @@ void init (FontData fd) {
 	LOGFONT logFont = fd.data;
 	int lfHeight = logFont.lfHeight;
 	logFont.lfHeight = device.computePixels(fd.height);
+
+	int primaryZoom = DPIUtil.mapDPIToZoom (device._getDPIx());
```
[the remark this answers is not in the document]
akoch-yatta: Good point

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Font.java
akoch-yatta: You are right. All accesses were refactored out over the last weeks.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Font.java
akoch-yatta: yes, that was not correctly adapted

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Shell.java
akoch-yatta: Good point

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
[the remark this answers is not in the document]
HeikoKlare: I see. Hiding it behind a getter would be good, as currently anyone could change this value at any point in time (even concurrently and/or by accident).

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/custom/CommonWidgetsDPIChangeHandlers.java
```diff
@@ -0,0 +1,51 @@
...
+ * <p>
+ * <b>IMPORTANT:</b> This class is <em>not</em> part of the public
+ * API for SWT. It is marked public only so that it can be shared
+ * within the packages provided by SWT. It is not available on all
+ * platforms, and should never be called from application code.
+ * </p>
+ * @see <a href="http://www.eclipse.org/swt/">Sample code and further information</a>
+ * @noreference This class is not intended to be referenced by clients
+ */
+public class CommonWidgetsDPIChangeHandlers {
```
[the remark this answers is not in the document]
akoch-yatta: I moved it over for this PR

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
HeikoKlare: Resolved as changed to native zoom

HeikoKlare (approved):
Thank you for the latest updates, @akoch-yatta! From my side, everything is fine now, and we also have the approval of @fedejeanne. So I will merge this now to have the changes in 2024-06 M2.

The failing MacOS build is unrelated:
* It fails in the same way for other PRs: https://github.com/eclipse-platform/eclipse.platform.swt/actions/runs/8799682907
* This PR only affects Win32 code, except for `DPIUtil` in which only extensions have been made. In addition, the extension were already present in previous builds of this PR, in which MacOS verification still worked and succeeded, e.g., https://github.com/eclipse-platform/eclipse.platform.swt/actions/runs/8750612935

HeikoKlare:
Just saw that the aarch64 Windows build fails on Jenkins. @akoch-yatta do you know whether that may be related to the changes in the native code you made or is it a problem with the pipeline?

HeikoKlare:
Yes, it runs into a timeout in the step building natives: https://ci.eclipse.org/releng/job/eclipse.platform.swt/view/change-requests/job/PR-1064/58/pipeline-console/?selected-node=333

Can you refer to any other PR with the same problem? I did not find any other PR compiling the binaries, as they do not perform any changes to native code.

akoch-yatta:
E.g. in for PR 1188, see https://ci.eclipse.org/releng/job/eclipse.platform.swt/job/PR-1188/1/pipeline-graph/

Isn't just the node for the aarch64 build offline?
image

HeikoKlare:
You are right, it is probably because of the Jenkins node being offline.

I am concerned that if we still merged, the Jenkins builds on `master` will not succeed and may not deploy the new binaries to the repo, such that the Java code becomes incompatible with the provided binaries.
@akurtakov @HannesWell maybe one of you knows if that assumption correct (i.e., whether we should not merge this as long as the Windows ARM binaries compilation does not work) and/or whether we can restart the Jenkins node (https://ci.eclipse.org/releng/computer/rie8t%2Dwin11%2Darm64/)?

iloveeclipse:
> whether we can restart the Jenkins node (https://ci.eclipse.org/releng/computer/rie8t%2Dwin11%2Darm64/)?

Create ticket here: https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues

HannesWell:
> I am concerned that if we still merged, the Jenkins builds on master will not succeed and may not deploy the new binaries to the repo, such that the Java code becomes incompatible with the provided binaries.

That's all correct and therefore should be avoided.

HeikoKlare:
Thanks for your quick replies, @iloveeclipse @HannesWell! That's what I've expected.
I've created a ticket for the issue: https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/4587

HeikoKlare:
> Thank you for the latest updates, @akoch-yatta! From my side, everything is fine now, and we also have the approval of @fedejeanne. So I will merge this now to have the changes in 2024-06 M2.
> 
> The failing MacOS build is unrelated:
> 
> * It fails in the same way for other PRs: https://github.com/eclipse-platform/eclipse.platform.swt/actions/runs/8799682907
> * This PR only affects Win32 code, except for `DPIUtil` in which only extensions have been made. In addition, the extension were already present in previous builds of this PR, in which MacOS verification still worked and succeeded, e.g., https://github.com/eclipse-platform/eclipse.platform.swt/actions/runs/8750612935

The Jenkins build has been fixed. GH Actions builds for MacOS and Linux fail (for different reasons), but both related to infrastructure issues and unrelated to the changes in this PR (see explanation above for MacOS, now there is also a signing issue on Linux).

Phillipus:
Are these changes entirely opt-in by specifying VM arguments? I looked at some of the code and I see fundamental changes that don't look optional. Can we expect possible regressions in Eclipse 4.32?

HeikoKlare:
Most of the changes are optional (to be activated via flag), but some changes had to be made in existing code. It is unavoidable to adapt existing code in order to provide multi-monitor HiDPI support for Windows, but we have, of course, thoroughly reviewed in particular those changes that are not opt-in to avoid regressions. If you still find regressions, please report them.

PR #1074, opened 2024-03-01
Enable automatic version bumps on milestone creation

akurtakov:
Work towards
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1844


akurtakov:
I'm merging this one as I can't easily test it as modification will require changes in parent pom in aggregator module to test it fully. It should be enough from what I can say.

Issue #1052, opened 2024-02-16
SelectionListener on SWT table control behaves differently on Mac and Windows


tobias-melcher:
**Describe the bug**
A selectionListener implementation on the SWT table control created by org.eclipse.jface.text.contentassist.CompletionProposalPopup behaves differently on Mac and Windows.

**To Reproduce**
A standalone SWT snippet to reproduce the behavior:
```
import java.util.ArrayList;
import java.util.List;

import org.eclipse.jface.bindings.keys.KeyStroke;
import org.eclipse.jface.dialogs.MessageDialog;
import org.eclipse.jface.fieldassist.ContentProposal;
import org.eclipse.jface.fieldassist.ContentProposalAdapter;
import org.eclipse.jface.fieldassist.IContentProposal;
import org.eclipse.jface.fieldassist.TextContentAdapter;
import org.eclipse.swt.SWT;
import org.eclipse.swt.events.KeyAdapter;
import org.eclipse.swt.events.KeyEvent;
import org.eclipse.swt.events.SelectionAdapter;
import org.eclipse.swt.layout.GridData;
import org.eclipse.swt.layout.GridLayout;
import org.eclipse.swt.widgets.Composite;
import org.eclipse.swt.widgets.Control;
import org.eclipse.swt.widgets.Display;
import org.eclipse.swt.widgets.Shell;
import org.eclipse.swt.widgets.Table;
import org.eclipse.swt.widgets.TableItem;
import org.eclipse.swt.widgets.Text;

public class Snippet1 {
	static String lastSelectedTableItemText = null;
	static Table proposalTable = null;

	private static class MyContentProposal extends ContentProposal {
		public MyContentProposal(String string) {
			super(string);
		}

		@Override
		public String getContent() {
			if (proposalTable != null) {
				System.out.println(lastSelectedTableItemText);
				MessageDialog.openInformation(Display.getDefault().getActiveShell(), "lastSelectedTableItemText",
						lastSelectedTableItemText);
			}
			return super.getContent();
		}
	}

	public static void main(String[] args) throws Exception {
		Display display = new Display();
		final Shell shell = new Shell(display);
		shell.setText("Snippet 1");

		var text = new Text(shell, SWT.NONE);
		text.setLayoutData(new GridData(GridData.FILL_HORIZONTAL));
		text.setText("text");

		KeyStroke keyStroke = KeyStroke.getInstance("Ctrl+Space");
[...]

PR #1079, opened 2024-03-04
fix SWT table/tree control SelectionListener on Mac

tobiasmelcher:
SelectionEvent.item in callback method widgetSelected now contains the current selected item

Fixes: https://github.com/eclipse-platform/eclipse.platform.swt/issues/1052

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Table.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Table.java
@@ -3477,6 +3477,9 @@ void tableViewSelectionDidChange (long id, long sel, long aNotification) {
 
 @Override
 void tableViewSelectionIsChanging (long id, long sel, long aNotification) {
+	if (keyDown) {
+		return;
+	}
 	didSelect = true;
 	sendSelection();
 }
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Table.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Table.java
@@ -3477,9 +3477,9 @@ void tableViewSelectionDidChange (long id, long sel, long aNotification) {
 
 @Override
 void tableViewSelectionIsChanging (long id, long sel, long aNotification) {
-	if (keyDown) {
-		return;
-	}
+	// tableViewSelectionIsChanging is called when pressing ARROW_DOWN, ARROW_UP key
+	// don't run sendSelection because it would then gather the "old" incorrect selected row
+	if (keyDown) return;
 	didSelect = true;
 	sendSelection();
 }
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Tree.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Tree.java
@@ -96,6 +96,7 @@ public class Tree extends Composite {
 
 	/* Used to control drop feedback when DND.FEEDBACK_EXPAND and DND.FEEDBACK_SCROLL is set/not set */
 	boolean shouldExpand = true, shouldScroll = true;
+	boolean keyDown;
 
 	final int nativeItemHeight;
 
@@ -2108,7 +2109,9 @@ boolean isTrim (NSView view) {
 @Override
 void keyDown(long id, long sel, long theEvent) {
 	ignoreSelect = preventSelect = false;
+	keyDown = true;
 	super.keyDown(id, sel, theEvent);
+	keyDown = false;
 }
 
 @Override
@@ -2463,6 +2466,9 @@ void outlineViewSelectionDidChange (long id, long sel, long notification) {
 
 @Override
 void outlineViewSelectionIsChanging (long id, long sel, long notification) {
+	if (keyDown) {
+		return;
+	}
 	didSelect = true;
 	sendSelection ();
 }
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Tree.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Tree.java
@@ -97,7 +97,6 @@ public class Tree extends Composite {
 	/* Used to control drop feedback when DND.FEEDBACK_EXPAND and DND.FEEDBACK_SCROLL is set/not set */
 	boolean shouldExpand = true, shouldScroll = true;
 	boolean keyDown;
-
 	final int nativeItemHeight;
 
 	static int NEXT_ID;
@@ -2466,9 +2465,9 @@ void outlineViewSelectionDidChange (long id, long sel, long notification) {
 
 @Override
 void outlineViewSelectionIsChanging (long id, long sel, long notification) {
-	if (keyDown) {
-		return;
-	}
+	// outlineViewSelectionIsChanging is called when pressing ARROW_DOWN, ARROW_UP key
+	// don't run sendSelection because it would then gather the "old" incorrect selected row
+	if (keyDown) return;
 	didSelect = true;
 	sendSelection ();
 }


bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/SelectionIsChangingMarkers.java
BeckerWdf, line 1: Copyright Header is missing

BeckerWdf:
> Only qualifier changed for (org.eclipse.swt/3.125.0.v20240304-0929). Expected to have bigger x.y.z than what is available in baseline (3.125.0.v20240227-1638)
> 
> says that plugin version number must be increased. I think to 3.125.100

Pls also adapt the version number.

merks:
I think I saw the swt version incremented for another change so probably a rebase will help.

lshanmug:
@Phillipus unable to add you as reviewer, hence tagging.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/widgets/TypedListener.java
```diff
@@ -14,9 +14,10 @@
 package org.eclipse.swt.widgets;
 
 
-import org.eclipse.swt.internal.SWTEventListener;
 import org.eclipse.swt.*;
 import org.eclipse.swt.events.*;
+import org.eclipse.swt.internal.*;
+import org.eclipse.swt.internal.cocoa.*;
```
BeckerWdf, line 20: you are not allow to use "cocoa" here.

BeckerWdf:
```
[540](https://ci.eclipse.org/releng/job/eclipse.platform.swt/job/PR-1079/5/pipeline-console/?start-byte=0&selected-node=211#log-540)
09:33:36  [ERROR] Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:4.0.6:compile (default-compile) on project org.eclipse.swt.gtk.linux.ppc64le: Compilation failure: Compilation failure: 
[541](https://ci.eclipse.org/releng/job/eclipse.platform.swt/job/PR-1079/5/pipeline-console/?start-byte=0&selected-node=211#log-541)
09:33:36  [ERROR] /home/jenkins/agent/workspace/eclipse.platform.swt_PR-1079/eclipse.platform.swt/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/widgets/TypedListener.java:[261] 
[542](https://ci.eclipse.org/releng/job/eclipse.platform.swt/job/PR-1079/5/pipeline-console/?start-byte=0&selected-node=211#log-542)
09:33:36  [ERROR] 	e.index = tree.getSelectedRow();
[543](https://ci.eclipse.org/releng/job/eclipse.platform.swt/job/PR-1079/5/pipeline-console/?start-byte=0&selected-node=211#log-543)
09:33:36  [ERROR] 	               ^^^^^^^^^^^^^^
[544](https://ci.eclipse.org/releng/job/eclipse.platform.swt/job/PR-1079/5/pipeline-console/?start-byte=0&selected-node=211#log-544)
09:33:36  [ERROR] The method getSelectedRow() is undefined for the type Tree
```

BeckerWdf:
I manually tested the PR with the Snippet from https://github.com/eclipse-platform/eclipse.platform.swt/issues/1052#issuecomment-1949067613

I could reproduce the bug without this PR and see that the bug is gone with this PR. So functionally this looks good to me.
But I am not an SWT expert. Maybe @lshanmug can review the code and say if it's fine.

tobiasmelcher:
I tried to implement a unit test which verifies the fix; the test should run through the changed code in

Table#tableViewSelectionIsChanging (long id, long sel, long aNotification)
and then in
TypedListener#handleEvent (Event e).

My goal is to write a stable test which does not use a Robot class. My experience is that Robot tests are unstable and run into issues when tests are executed in parallel. 
What is the testing strategy in the SWT project? And how are the test implemented?

The test code below does not work because I don't know how to send a ARRAY_DOWN key down event properly so that the mentioned two method calls are executed. 

Can you please help me here? Thanks.

```
public class TestsUnitTableWidgetSelectionListener {

	private TableItem lastSelectedItem = null;

	@Test
	public void tableItemInSelectionChangeEventOnKeyDown() throws Exception {
		Table table = new Table(shell, SWT.NONE);
		new TableItem(table, SWT.NONE).setText("first");
		new TableItem(table, SWT.NONE).setText("second");
		new TableItem(table, SWT.NONE).setText("third");
		table.addSelectionListener(new SelectionAdapter() {
			@Override
			public void widgetSelected(SelectionEvent e) {
				super.widgetSelected(e);
				lastSelectedItem = (TableItem) e.item;
			}
		});
		table.setSelection(0);
		table.setFocus();

		GridLayout layout = new GridLayout(1, true);
		shell.setLayout(layout);
		shell.setLayoutData(new GridData(GridData.FILL_BOTH));
		shell.open();
		while (!display.readAndDispatch()) {
			display.sleep();
		}
		table.setFocus();

		// how to send ARROW_DOWN key event in table control
		// so that next entry is selected and table selectionListener widgetSelected is called

		/*  // this doesn't work
		Event event = new Event();
		event.keyCode = SWT.ARROW_DOWN;
		table.notifyListeners(SWT.KeyDown, event);
		*/

		// try to call keyDown directly
		//table.keyDown(display.application.id, OS.sel_sendEvent_, 0);

		while (!display.readAndDispatch()) {
			display.sleep();
		}
		assertNotNull(lastSelectedItem);
		assertEquals("second", lastSelectedItem.getText());
	}
}
```

lshanmug:
@tobiasmelcher Is this a new issue with the recent OS versions?
I believe `tableSelectionIsChanging` is not expected to be called for keyboard - pls see doc - https://developer.apple.com/documentation/appkit/nstableviewdelegate/1530812-tableviewselectionischanging?language=objc. This seems like a Mac bug as it's being called now when selection is changed by keyboard.

I didn't fully understand the change in your PR, can you please explain? Why is setData being used?

tobiasmelcher:
> @tobiasmelcher Is this a new issue with the recent OS versions? I believe `tableSelectionIsChanging` is not expected to be called for keyboard - pls see doc - https://developer.apple.com/documentation/appkit/nstableviewdelegate/1530812-tableviewselectionischanging?language=objc. This seems like a Mac bug if it's being called now when selection is changed by keyboard and the widget.selectedRow() is not returning the correct selected row.
> 
> I didn't fully understand the change in your PR, can you please explain? Why is setData being used?

Sorry, I cannot tell you if this is a new issue. 

https://developer.apple.com/documentation/appkit/nstableviewdelegate/1530812-tableviewselectionischanging?language=objc. This seems like a Mac bug if it's being called now when selection is changed by keyboard and the widget.selectedRow() is not returning the correct selected row.
> I fully agree to your statement. The docu clearly indicates that tableviewselectionischanging is not called for keyboard events. But this is right now the case.

I tried to find a solution without modifying the native code (because I am not capable of fixing/changing the native code).
As `widget.selectedRow ();` is returning the wrong table index when called in tableViewSelectionIsChanging, the code in my change is moving the getIndex logic to a later point in time when the correct index is returned. This is the case after `sendSelectionEvent (SWT.Selection, event, false); ` is called when inside `TypedListener#handleEvent`. As the code inside TypedListener needs to be platform independent, I cannot directly access the SWT Table or Tree instance from the Mac implementation and then query the actual index. I therefore store the info via `setData` and using a callback method of new class `SelectionIsChangingEventUpdater`. 
Maybe, there is a more elegant solution which I currently don't see. Can you please help?

Thanks and best regards,
Tobias

lshanmug:
@tobiasmelcher
AFAIU `tableViewSelectionIsChanging` is not expected to send the selection event if selection changes by keyboard events. `tableViewSelectionDidChange` should send the selection event in that case. 
Since this is not happening, I think we can try to emulate this behavior by not calling `sendSelection()` in `tableViewSelectionIsChanging` if it is coming from a KeyDown.
I see there is keyDown property already available in Table and being set to true on Keydown. We can use that and just return from `tableViewSelectionIsChanging` and let `tableViewSelectionDidChange` send the selection event with the correct index.
```
tableViewSelectionIsChanging() {
  if (keyDown) return;
  // existing code
}
```
Please try it out and see if this works. Similar changes will be needed in Tree in that case.

tobiasmelcher:
> @tobiasmelcher AFAIU `tableViewSelectionIsChanging` is not expected to send the selection event if selection changes by keyboard events. `tableViewSelectionDidChange` should send the selection event in that case. Since this is not happening, I think we can try to emulate this behavior by not calling `sendSelection()` in `tableViewSelectionIsChanging` if it is coming from a KeyDown. I see there is keyDown property already available in Table and being set to true on Keydown. We can use that and just return from `tableViewSelectionIsChanging` and let `tableViewSelectionDidChange` send the selection event with the correct index.
> 
> ```
> tableViewSelectionIsChanging() {
>   if (keyDown) return;
>   // existing code
> }
> ```
> 
> Please try it out and see if this works. Similar changes will be needed in Tree in that case.

thanks so much for guiding me into the right direction. The updated change is so small and is working so nicely on my machine.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Table.java
```diff
@@ -3477,6 +3477,9 @@ void tableViewSelectionDidChange (long id, long sel, long aNotification) {
 
 @Override
 void tableViewSelectionIsChanging (long id, long sel, long aNotification) {
+	if (keyDown) {
```
lshanmug, line 3480: Can we make it `if (keyDown) return;` to keep the style consistent across?

lshanmug: Also, will be useful to add a comment here on why this check is required.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Tree.java
```diff
@@ -2463,6 +2466,9 @@ void outlineViewSelectionDidChange (long id, long sel, long notification) {
 
 @Override
 void outlineViewSelectionIsChanging (long id, long sel, long notification) {
+	if (keyDown) {
```
lshanmug, line 2469: see comments in Table.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Table.java
tobiasmelcher: sure, I will do that. Should I add a link to the developer.apple site or is this forbidden?

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Tree.java
```diff
@@ -2108,7 +2108,9 @@ boolean isTrim (NSView view) {
 @Override
 void keyDown(long id, long sel, long theEvent) {
 	ignoreSelect = preventSelect = false;
+	keyDown = true;
 	super.keyDown(id, sel, theEvent);
+	keyDown = false;
```
tobiasmelcher, line 2113: the same is also done in Table#keyDown(). I am asking me if we should better call ```keyDown = false``` in a finally block for the case when an exception is thrown in super.keyDown. Can we ensure that no exception is thrown in super.keyDown() ?

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Table.java
lshanmug: A short description of why the code is requried would help. Links to specific doc in apple developer site change sometimes or the content changes then the link may not be useful.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Tree.java
lshanmug: super.keyDown() sends the key down event and calls the keydown listeners. It's possible that there are exceptions thrown in the listener code. They have to be handled in the listener or it would be break the execution.

lshanmug (approved):
LGTM. 
@tobiasmelcher We can merge if it looks good from your testing POV.

PR #1083, opened 2024-03-05
Add SWT binaries module for Windows on Arm64 platform.

chirontt:
Follow-up on PR #1045.

With this PR and the above #1045, the SWT binaries module's files can be produced with the following build commands:
```
mvn clean verify -DskipTests=true
```
and the module's `jar/zip` files will be produced in the `target` directory:
```
org.eclipse.swt.win32.win32.aarch64-<version>.jar
org.eclipse.swt.win32.win32.aarch64-<version>-sources.jar
swt-<version>-win32-win32-aarch64.zip
```


chirontt:
Closing this PR, as it's been merged with PR #1045.

PR #1091, opened 2024-03-09
Increment the version of org.eclipse.swt.tools.base

merks:
Force a new version so that publishing to Maven Central can update the version without specialized remapping rules.

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1882


merks:
This is specifically to be able to remove this mapping rule hack:

image

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/f8bde409fa0267b917464fc197e4f2a0a25366ec/eclipse.platform.releng/publish-to-maven-central/SDK4Mvn.aggr#L18

Issue #1093, opened 2024-03-09
.api_description in SWT is empty in 4.31 RC2, API analysis broken


iloveeclipse:
See https://github.com/eclipse-pde/eclipse.pde/issues/1187.
Looks like we don't properly build SWT bundles since some late 4.31 build.
The `.api_description` files are empty in SWT OS specific bundles, which breaks API analysis.

- SDK-I20240202-1800 OK (org.eclipse.swt.win32.win32.x86_64_3.125.0.v20240201-1628.jar)
- SDK-I20240209-1800 Bad (org.eclipse.swt.win32.win32.x86_64_3.125.0.v20240208-1757.jar)

This is a big problem, because we will be unable to properly analyse API for 4.32!

@laeubi , @HannesWell , @merks : can we create 4.32 RC2a?

laeubi:
> Looks like we don't properly build SWT bundles since some late 4.31 build.
> The `.api_description` files are empty in SWT OS specific bundles, which breaks API analysis.

Why is this not discovered by I-Builds?

merks:
You could build a repo just for use as a baseline, but rebuilding SimRel, EPP, and all that is just not realistic at this late state of the game.  I think the publishing to Maven is done already too.  Can we even do a branch build at this point?

laeubi:
What I wonder most is why this was not discovered earlier?
Does it mean that here:

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/f8bde409fa0267b917464fc197e4f2a0a25366ec/eclipse-platform-parent/pom.xml#L130

it is not pointing to the "real" RC2?

merks:
The real one is this:

https://download.eclipse.org/justj/?file=eclipse/updates/4.31/R-4.31-202402290520

But I expect they have identical content.

merks:
Another concern I have about a respin is that it will (I assume) produce IUs and artifacts with identical version as the current ones, so there would be one variant of the artifact with API descriptions and one without, though they would look like the same artifact.  If the description-less version is already in the shared bundle pool, you'd not get a new one to be downloaded.  Also, depending on what repositories are visible, you might get either one being downloaded non-deterministically; these have been out in the wild for quite some time now, so could be in many other p2 repositories. 

This seems like a bit of a disaster.  😱  How can we do proper development with an API baseline for 4.31? 😭

merks:
@iloveeclipse 

[...]

PR #1099, opened 2024-03-11
[DO NOT MERGE] Change breaks SWT API by adding/removing annotations

iloveeclipse:
The only purpose is to see if that will fail Jenkins / Github validation.

The `@noreference` is supposed to be read either from Java sources (in IDE) or from *generated* .api_description files from compiled SWT native fragments.

See https://github.com/eclipse-platform/eclipse.platform.swt/issues/1093

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/cairo/org/eclipse/swt/graphics/Path.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cairo/org/eclipse/swt/graphics/Path.java
@@ -50,7 +50,6 @@ public class Path extends Resource {
 	 * platforms and should never be accessed from application code.
 	 * </p>
 	 *
-	 * @noreference This field is not intended to be referenced by clients.
 	 */
 	public long handle;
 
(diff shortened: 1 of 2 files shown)


laeubi:
@iloveeclipse for SWT we have no API checks at Jenkins/Github...

See:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/1011

iloveeclipse:
> we have no API checks at Jenkins/Github...

OMG. I assumed it was merged already.

So in sum: we don't have API checks in SDK build and also no checks in SWT repo. Perfect...

laeubi:
> > we have no API checks at Jenkins/Github...
> 
> OMG. I assumed it was merged already.
> 
> So in sum: we don't have API checks in SDK build and also no checks in SWT repo. Perfect...

We seem not to have ones in Ibuild either... but that how it worked for years now, so it must be perfect no? ... but we are working on it, just that you are not wondering about it.

akurtakov:
FWIW https://download.eclipse.org/eclipse/downloads/drops4/I20240310-1800/apitools/analysis/html/index.html used to be only source for api checks (besides workspace warnings/errors) before Mickael integrated it in the verification builds.

laeubi:
> FWIW https://download.eclipse.org/eclipse/downloads/drops4/I20240310-1800/apitools/analysis/html/index.html used to be only source for api checks (besides workspace warnings/errors) before Mickael integrated it in the verification builds.

Yes but I can't see an SWT bundle there...

PR #1109, opened 2024-03-13
API Filters on Mac

Phillipus:
API filters

PR #1110, opened 2024-03-13
Make all API leak-issues errors and fail build on API warnings

HannesWell:


iloveeclipse:
Not sure if you need https://github.com/eclipse-platform/eclipse.platform.swt/pull/1109 first. I haven't time to check it and also can't validate it because I have no Mac.

HannesWell:
> Not sure if you need #1109 first. I haven't time to check it and also can't validate it because I have no Mac.

Yes. For the moment I have also included the alternative solution given with https://github.com/eclipse-platform/eclipse.platform.swt/pull/1111 and the build says that then all seems to work.

HannesWell:
Great. This now passes. So we can make API-tools absolutely strict and (hopefully) never miss an issue in the future anymore.

PR #1111, opened 2024-03-13
Fix API warnings in macosx fragments

HannesWell:
For cocoa specific implementations of
- o.e.swt.accessibility.Accessible tag methods already tagged as @noreference also with @nooverride so that API tools considers them as internal and does not complain about non-API types being referenced in their signature.

- o.e.swt.graphics.GCData tag all fields as @noreference so that API tools does not complain about non-API types being used in public fields of this actually internal class (tagged as @noreference).

Suppress issue about forbidden extension of Scrollable by Composite.

### DIFF
--- a/binaries/org.eclipse.swt.cocoa.macosx.aarch64/.settings/.api_filters
+++ b/binaries/org.eclipse.swt.cocoa.macosx.aarch64/.settings/.api_filters
@@ -248,6 +248,108 @@
             </message_arguments>
         </filter>
     </resource>
+    <resource path="Eclipse SWT/cocoa/org/eclipse/swt/graphics/GCData.java" type="org.eclipse.swt.graphics.GCData">
+        <filter id="338940029">
+            <message_arguments>
+                <message_argument value="org.eclipse.swt.graphics.GCData"/>
+                <message_argument value="bg"/>
+            </message_arguments>
+        </filter>
+        <filter id="338940029">
+            <message_arguments>
+                <message_argument value="org.eclipse.swt.graphics.GCData"/>
+                <message_argument value="clipPath"/>
+            </message_arguments>
+        </filter>
+        <filter id="338940029">
+            <message_arguments>
+                <message_argument value="org.eclipse.swt.graphics.GCData"/>
+                <message_argument value="fg"/>
+            </message_arguments>
+        </filter>
+        <filter id="338940029">
+            <message_arguments>
+                <message_argument value="org.eclipse.swt.graphics.GCData"/>
+                <message_argument value="flippedContext"/>
+            </message_arguments>
+        </filter>
+        <filter id="338940029">
+            <message_arguments>
+                <message_argument value="org.eclipse.swt.graphics.GCData"/>
+                <message_argument value="inverseTransform"/>
+            </message_arguments>
+        </filter>
+        <filter id="338940029">
+            <message_arguments>
+                <message_argument value="org.eclipse.swt.graphics.GCData"/>
+                <message_argument value="layoutManager"/>
+            </message_arguments>
+        </filter>
+        <filter id="338940029">
+            <message_arguments>
+                <message_argument value="org.eclipse.swt.graphics.GCData"/>
+                <message_argument value="paintRect"/>
+            </message_arguments>
+        </filter>
+        <filter id="338940029">
+            <message_arguments>
+                <message_argument value="org.eclipse.swt.graphics.GCData"/>
+                <message_argument value="path"/>
+            </message_arguments>
+        </filter>
+        <filter id="338940029">
+            <message_arguments>
+                <message_argument value="org.eclipse.swt.graphics.GCData"/>
+                <message_argument value="size"/>
+            </message_arguments>
+        </filter>
+        <filter id="338940029">
+            <message_arguments>
+                <message_argument value="org.eclipse.swt.graphics.GCData"/>
+                <message_argument value="textContainer"/>
+            </message_arguments>
+        </filter>
+        <filter id="338940029">
+            <message_arguments>
+                <message_argument value="org.eclipse.swt.graphics.GCData"/>
+                <message_argument value="textStorage"/>
+            </message_arguments>
+        </filter>
+        <filter id="338940029">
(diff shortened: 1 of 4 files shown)


bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/GCData.java
```diff
@@ -51,19 +51,30 @@ public final class GCData {
 	public int fillRule = SWT.FILL_EVEN_ODD;
 	public Image image;
 
+	/** @noreference This field is not intended to be referenced by clients. */
```
HannesWell, line 54: This should IMHO also be implied by the `@noreference` of this class respectively public non-api types should not be an issue.

HannesWell:
Looks like even with this approach we have to add filters, but they are about previously API fields having become non-API.
I think the advantage in contrast to suppressing the leakage warning is that these filters are temporary and can be removed after the next release.

HannesWell:
Thanks to @Phillipus who tested this PR on a real Mac and verified all API issues on Mac are really fixed with this:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/1109#issuecomment-1997953001

Phillipus:
> Thanks to @Phillipus who tested this PR on a real Mac and verified all API issues on Mac are really fixed with this:
> 
>     * [API Filters on Mac #1109 (comment)](https://github.com/eclipse-platform/eclipse.platform.swt/pull/1109#issuecomment-1997953001)

When you merge this, ping me and I'll double check it.

binaries/org.eclipse.swt.cocoa.macosx.aarch64/.settings/.api_filters
```diff
@@ -248,6 +248,108 @@
             </message_arguments>
         </filter>
     </resource>
+    <resource path="Eclipse SWT/cocoa/org/eclipse/swt/graphics/GCData.java" type="org.eclipse.swt.graphics.GCData">
```
iloveeclipse, line 251: All new filters for GCData are reported as "no longer used" for me.

iloveeclipse:
@HannesWell : beside the unneeded `GCData` filters the rest looks good.
Please double check that, remove if you also see the same & feel free to merge.

HannesWell:
> @HannesWell : beside the unneeded `GCData` filters the rest looks good. Please double check that, remove if you also see the same & feel free to merge.

I used to get those warnings as well, but when I ensured the active workspace target and API-baseline target are properly resolved to one mac environment and the API-baseline is refreshed and after a clean+build they vanished.

Furthermore I have currently also included this change in [#1110 ](https://github.com/eclipse-platform/eclipse.platform.swt/pull/1110), where the build is configured to fail on API warnings and the build passes.
And in https://github.com/eclipse-platform/eclipse.platform.swt/pull/1109#issuecomment-1997953001 @Phillipus said he does not have any warning with this.
So as far as I can tell this is all save and sound.

Phillipus:
> And in https://github.com/eclipse-platform/eclipse.platform.swt/pull/1109#issuecomment-1997953001 @Phillipus said he does not have any warning with this.

I just double-checked with latest I-build and still no warnings or errors.

Phillipus:
@HannesWell Since 7479681c I am now getting errors for SWT such as `AccessibleActionListener extends non-API type SWTEventListener`. I'm using Eclipse 4.31 as IDE. Before this commit they were warnings. What should I do?

merks:
I think you must use Eclipse SDK 4.32 IDE because.  Also, I had problem with missing links to the api filters folder that @iloveeclipse helped me resolve. I ended up having to delete the projects from my workspace an importing them again....

Phillipus:
> I think you must use Eclipse SDK 4.32 IDE 

Hmmm...tricky. Before now I could use the current version of the IDE and import the SWT projects and simply ignore any warnings. Now each time I get the SWT projects I have to set these errors to warnings for the projects settings.

merks:
yes, there was a problem that the API baseline details are missing from the 4.31 release of SWT so they are hard coded in PDE but of course only for 4.32.  It's a tangled web we weave....

Phillipus:
My setup is I have Eclipse 4.31 installed on 6 machines (different OS and architectures) and pull and test the latest SWT project commits and ignore any API warnings. What I don't want to do is also install yet another version of the IDE on 6 machines. ;-) I guess I'll have to keep my own copy of `org.eclipse.pde.api.tools.prefs` for the SWT projects?

Phillipus:
>  I ended up having to delete the projects from my workspace an importing them again....

Strange. I did that and I no longer see the errors. False alarm?

HannesWell:
> yes, there was a problem that the API baseline details are missing from the 4.31 release of SWT so they are hard coded in PDE but of course only for 4.32. It's a tangled web we weave....

Indeed. For the next release cycle this will hopefully be stable again.
But the error desribed should not depend on the baseline but only on the current workspace. So unless the error persists it really could be a false alarm.


> My setup is I have Eclipse 4.31 installed on 6 machines (different OS and architectures) and pull and test the latest SWT project commits and ignore any API warnings.

That's a large setup 😮

> What I don't want to do is also install yet another version of the IDE on 6 machines. ;-) I guess I'll have to keep my own copy of `org.eclipse.pde.api.tools.prefs` for the SWT projects?

That's relatable. Making the warnings errors was meant to help developers to notice them early and not only in the CI that fails on API warnings since #1110.
But if it continues to be a problem I can also make the leakage issues warnings again. As said, the CI should then catch new errors.

Phillipus:
I think it was a one-off. I don't have an API baseline set so I shouldn't see the errors, right?

HannesWell:
> I think it was a one-off. I don't have an API baseline set so I shouldn't see the errors, right?

It dependents on the kind of error. If it said something changed in a wrong way, then a missing baseline might be a problem (but actually without a baseline such issues are actually, falsely, not raised).

> errors for SWT such as `AccessibleActionListener extends non-API type SWTEventListener`. 

That exact issues should not depend on the baseline, but who knows what API-tools is doing in its inner-most core. 😅

PR #1106, opened 2024-03-13
Revert 1101

iloveeclipse:
1) Revert "Replace internal SWTEventListener directly with java.util.EventListener"

This reverts commit bd306acb742a18bb7a234ae493a2b02257beaad2.
Replacing SWTEventListener directly with java.util.EventListener breaks binary compatibility with bundles compiled against any older SWT version.
- See discussion on https://github.com/eclipse-platform/eclipse.platform.swt/pull/1101

2) Added filters for SWTEventListener used in SWT API
- See https://github.com/eclipse-platform/eclipse.platform.swt/pull/1011
- See https://github.com/eclipse-platform/eclipse.platform.swt/pull/1101

PR #1119, opened 2024-03-15
Tag API leaking Windows OLE methods/fields as non-api

HannesWell:
Instead of just suppressing the resource-leakage warnings, tag the methods/fields that leak non-API types as non-api themself.
This is an alternative solution to https://github.com/eclipse-platform/eclipse.platform.swt/pull/1106#issuecomment-1994229955 to 
get rid of the resource-leakage warnings in the Windows OLE part of SWT.
I think that's the more natural way to resolve this issue and has the advantage that we only temporarily have to add filters to suppress the error that a method/field became non-api.

@iloveeclipse or anybody else, what do you think about that?


HannesWell:
Does anybody have an opinion on this?

akurtakov:
@HeikoKlare What do you think of this one?

HeikoKlare:
Sounds reasonable to me. I fully agree with:

> I think that's the more natural way to resolve this issue and has the advantage that we only temporarily have to add filters to suppress the error that a method/field became non-api.

It will allow us to get rid of most of the API filter for the Windows fragments, which would be really great. And I don't see any risk either, because if someone unexpectedly relies on those methods being API, they will still exist and will still be referencable.

The current failures seem to be caused by the temporary filters for the API removals missing in the aarch64 fragment. @HannesWell do you want to add them so that we may even merge this for the upcoming release and can get rid of all the filters already in December?

HannesWell:
This is an old PR. 

> The current failures seem to be caused by the temporary filters for the API removals missing in the aarch64 fragment. @HannesWell do you want to add them so that we may even merge this for the upcoming release and can get rid of all the filters already in December?

Just done it. And yes, I assume we could then remove them soon.

PR #1132, opened 2024-03-21
Updated bundles with comparator errors

trancexpress:
See: eclipse-platform/eclipse.platform.releng.aggregator#1923


iloveeclipse (approved):
Changes are OK, only waiting for jenkins build to confirm that.

PR #1144, opened 2024-03-27
Reference legal files in native fragments to ensure consistency

HannesWell:
Having a shared base for all fragments of a platform ensures consistent legal files among the fragments of that platform.
This also reduce duplication of legal files in the repository and therefore simplifies future adjustments.


HannesWell:
Without touching the org.eclipse.swt project and therefore having the qualifier of all native fragments unchanged compared to the last I-build I see only for the `org.eclipse.swt.gtk.linux.ppc64le` fragment the following comparator difference:

```
 [INFO] --- tycho-p2:4.0.7-SNAPSHOT:p2-metadata-default (default-p2-metadata-default) @ org.eclipse.swt.gtk.linux.ppc64le ---
 [WARNING] MavenProject: org.eclipse.swt:org.eclipse.swt.gtk.linux.ppc64le:3.126.0-SNAPSHOT @ /home/jenkins/agent/workspace/eclipse.platform.swt_PR-1144/eclipse.platform.swt/binaries/org.eclipse.swt.gtk.linux.ppc64le/.polyglot.META-INF: baseline and build artifacts have same version but different contents
    no-classifier: different
       about.html: different
       about_files/mpl-v20.txt: not present in baseline
       about_files/webkit-bsd.txt: not present in baseline
    classifier-sources: different
       about.html: different
       about_files/mpl-v20.txt: not present in baseline
       about_files/webkit-bsd.txt: not present in baseline
```

So it looks like this has already helped to find and resolve inconsistencies.
I'll enforce a qualifier-update of org.eclipse.swt (and consequently of all its fragments) in order to avoid I-build failures when this is submitted.

akurtakov:
This kind of changes are more than welcome. Thanks for them!

Issue #1145, opened 2024-03-28
DnDExample not working under wayland/Gtk 3


akurtakov:
**Describe the bug**
No drag start is even detected

**To Reproduce**
Run DNDExample from swt repository and make sure you use GDK_BACKEND=wayland

**Expected behavior**
DND operations work

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [ ] Windows
- - [x] Linux
- - [ ] macOS 

2. Additixonal OS info (e.g. OS version, Linux Desktop, etc)
Fedora 40, Gnome on wayland

akurtakov:
Some more investigation pointed that the issue is not there for Text, StyledText and Combo but happens for everything else.

PR #1149, opened 2024-03-29
Make gtk_gesture* bindings non-dynamic

akurtakov:
These are available since Gtk 3.14 and are used unconditionally thus don't have to be dynamic.
That would give a bit better backtraces when crashing during DnD.


akurtakov:
Prep step to make investigating https://github.com/eclipse-platform/eclipse.platform.swt/issues/1145 easier.

PR #1150, opened 2024-03-30
[Build|Win32] Simplify build.bat for SWT's windows native binaries

HannesWell:
and remove elements specific to old build-clusters that are unused now.

The win32.aarch64 binaries are already build with MSVC 2022 and I have already requested to update MSVC on the x86_64 windows build machine to 2022 as well:
https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/4444

Once that is completed I think we should simply require require MSVC 2022 to build the SWT native binaries for Windows for all architectures.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT PI/win32/library/build.bat
+++ b/bundles/org.eclipse.swt/Eclipse SWT PI/win32/library/build.bat
@@ -1,5 +1,5 @@
 @rem ***************************************************************************
-@rem Copyright (c) 2000, 2021 IBM Corporation and others.
+@rem Copyright (c) 2000, 2024 IBM Corporation and others.
 @rem
 @rem This program and the accompanying materials
 @rem are made available under the terms of the Eclipse Public License 2.0
@@ -18,10 +18,6 @@
 echo
 echo INFO Starting build of binaries. Detailed system setup instructions can be found in /Readme.Win32.md
 
-@rem SWT_BUILDDIR defaults to "W:" on the SWT Windows native build infrastructure machine.
-IF "x.%SWT_BUILDDIR%"=="x." set "SWT_BUILDDIR=W:"
-echo SWT build dir: %SWT_BUILDDIR%
-
 @rem Specify VisualStudio Edition: 'Community', 'Enterprise', 'Professional' etc.
 IF "x.%MSVC_EDITION%"=="x." set "MSVC_EDITION=auto"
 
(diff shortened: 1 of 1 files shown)


SyntevoAlex:
Why break something that works?
To me, it's actually good that any suitable compiler can be used.

SyntevoAlex:
I mean, native libraries are also built on developer's machines, and some of them may happen to have VS2019 but not VS2022. That's even more likely if they're also working on something that uses VS2019 and said codebase wasn't updated to VS2022 yet.

HannesWell:
> I mean, native libraries are also built on developer's machines, and some of them may happen to have VS2019 but not VS2022. That's even more likely if they're also working on something that uses VS2019 and said codebase wasn't updated to VS2022 yet.

OK. It was intended to help developers to indicate that MSVC 2022 is used to perform the compilation of the published artifacts as soon as https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/4444 is completed.
But if you say 2019 is so much more common and it is simpler to keep it as it I'm fine with that too. If anything is used that works in 2017/19 but not in 2022 it will be noticed in the CI, although I don't think SWT uses the latest features anyway.

Btw. I have introduced your auto-search for MSVC for the Equinox build too in https://github.com/eclipse-equinox/equinox/pull/581, but in a bit stripped down variant.
I applied the same simplifications with this PR as well. It does not change any functional behavior but saves a few lines of scripts that don't have a noticeable performance impact.

bundles/org.eclipse.swt/Eclipse SWT PI/win32/library/build.bat
```diff
@@ -108,9 +99,6 @@ GOTO :EOF
 @rem Find Visual Studio
 @rem %1 = path template with '$MSVC_VERSION$' and '$MSVC_EDITION$' tokens
 :FindVisualStudio
-	@rem Early return if already found
-	IF NOT "%MSVC_HOME%"=="" GOTO :EOF
```
SyntevoAlex, line 112: Why did you remove this?
I understand that for "auto" it will now do a lot more checking and produce a lot more output, finally selecting *last* found valid VS instead of the first one

bundles/org.eclipse.swt/Eclipse SWT PI/win32/library/build.bat
```diff
@@ -152,14 +134,6 @@ GOTO :EOF
 	@rem If the folder isn't there, then skip it without printing errors
 	IF NOT EXIST "%TESTED_VS_PATH%" GOTO :EOF
 
-	@rem Try this path
-	CALL :TryToUseVisualStudio "%TESTED_VS_PATH%"
-GOTO :EOF
-
-@rem Test Visual Studio and set '%MSVC_HOME%' on success
-@rem %1 = tested path
-:TryToUseVisualStudio
```
SyntevoAlex, line 161: The point of separate function here was for code clarity, to distinguish from 1/2/3 searching layers

bundles/org.eclipse.swt/Eclipse SWT PI/win32/library/build.bat
HannesWell: Less lines of code and thus a simpler script.

> I understand that for "auto" it will now do a lot more checking and produce a lot more output, 

It checks a few directories more, but I bet nobody will ever notice those few milliseconds (if at all) saved.
And no there is no additional print-out as you can see in the docs.

> finally selecting last found valid VS instead of the first one

This was not intended and only happened because I accidentally also removed the last `IF NOT "%MSVC_HOME%"=="" GOTO :EOF` check too. But that's restored.

bundles/org.eclipse.swt/Eclipse SWT PI/win32/library/build.bat
HannesWell: The three layers are still distinguished altough there are actually only two layers, one for the Version and one for the edition. The last one only assembles the path and checks if its's valid. And having that very last part in a separate method did not seem to improve the readability for me but still added more code.
If you want to have a method named `TryToUseVisualStudio` I would suggest to rename `FindVisualStudio3` to it, so that FindVisualStudio1 iterates over the versions, FindVisualStudio2 over the editon and the last tries to use that combination.

SyntevoAlex: It's a small issue and I'll leave it to you. Still, I find it not worthy to add diffs to patch over a matter of tastes.

bundles/org.eclipse.swt/Eclipse SWT PI/win32/library/build.bat
SyntevoAlex: > But that's restored

Not sure what you mean, patch still removes this condition.

HannesWell: Yes, but only for the first two functions, but the third one will still ensure that the first match found is used, but with less code.

bundles/org.eclipse.swt/Eclipse SWT PI/win32/library/build.bat
HannesWell: > It's a small issue and I'll leave it to you. Still, I find it not worthy to add diffs to patch over a matter of tastes.

OK, restored that part.

HannesWell:
Now that Visual Studio 2022 is also installed for the Win10-x86_64 build agent (see https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/4444) and the build passes this can be submitted.
I verified that for both Archs supported on Windows MSVC 2022 is used.

PR #1151, opened 2024-03-30
fix missing win32.aarch64 fragment in various p2.inf

chirontt:
Especially for the `org.eclipse.swt` fragment host, the `win32.aarch64` fragment entry in its `p2.inf` file would help in resolving the SWT classes required during compile of other modules (like `eclipse.platform`, `eclipse.platform.ui`, etc.) for the Windows on Arm64 platform.

### DIFF
--- a/bundles/org.eclipse.swt/META-INF/p2.inf
+++ b/bundles/org.eclipse.swt/META-INF/p2.inf
@@ -34,3 +34,8 @@ requires.7.namespace = org.eclipse.equinox.p2.iu
 requires.7.name = org.eclipse.swt.cocoa.macosx.aarch64
 requires.7.range = [$version$,$version$]
 requires.7.filter = (&(osgi.os=macosx)(osgi.ws=cocoa)(osgi.arch=aarch64)(!(org.eclipse.swt.buildtime=true)))
+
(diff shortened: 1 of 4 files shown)


HannesWell (commented):
Indeed I forgot to change that in #1045, don't know why. So thank you for providing this fix!

There is only one build failure:
```
[ERROR] Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:4.0.7:compare-version-with-baselines (compare-attached-artifacts-with-release) on project org.eclipse.swt.examples.ole.win32:
Only qualifier changed for (org.eclipse.swt.examples.ole.win32/3.109.0.v20240330-2225). Expected to have bigger x.y.z than what is available in baseline (3.109.0.v20231214-1947) -> [Help 1]
```

This can be fixed by simplify changing the Bundle-Version of `org.eclipse.swt.examples.ole.win32` to `3.109.100.qualifier`.

chirontt:
> This can be fixed by simplify changing the Bundle-Version of org.eclipse.swt.examples.ole.win32 to 3.109.100.qualifier.

Yes, PR updated, thanks.

HannesWell (approved):
Thanks!

PR #1158, opened 2024-04-02
Make requirement of o.e.swt for swt.win32.aarch64 temporarily optional

HannesWell:
As long org.eclipse.swt.win32.aarch64 is not included in the 'eclipse' repository it cannot be mandatory.

See https://github.com/eclipse-platform/eclipse.platform.ui/pull/1789#issuecomment-2031249051.


HannesWell:
@chirontt can you please add to the umbrella TODO-List that this change has to be reverted once the equinox.launcher is available?

merks (approved):
Thanks.

chirontt:
> @chirontt can you please add to the umbrella TODO-List that this change has to be reverted once the equinox.launcher is available?

Sure, done. I think you should create a draft PR for this upcoming reversion task, and I'll add it to the TODO list too.

HannesWell:
As intended, this was now reverted via https://github.com/eclipse-platform/eclipse.platform.swt/pull/1222.

Issue #1161, opened 2024-04-03
SWT Shell do not honor setBounds/setLayout on a new user settings on RHEL 9.2


raghucssit:
**Background**
Eclipse Application is migrated to RHEL 9.2 from RHEL 7.9

**What is the problem?**
Any SWT Shell with style constant `SWT.DIALOG_TRIM` on a new user settings creates cropped window/dialog.
It does not honor neither `setLayout` nor `setBounds`. i.e. size of the window is always fixed irrespective of content in it.
The static size obsorved is width=205, height=228.
Once the application/dialog is launched it creates file `.swt/trims.prefs` under user home. After this file exists shell starts behaving fine.
So to replicate this issue please `rename or delete the .swt folder`. This folder does not exist when new user is logged in to RHEL.
You can use the attached project: 
[simple-plugin.zip](https://github.com/eclipse-platform/eclipse.platform.swt/files/14858803/simple-plugin.zip)
Please see below video for more details:

https://github.com/eclipse-platform/eclipse.platform.swt/assets/10301131/d9fba577-113a-498a-9541-cc73761cd7a5

**Expected behavior**
Shell should appear as per the layout or bounds set by client.

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [ ] Windows
- - [ x] Linux
- - [ ] macOS 

2. Additional OS info (e.g. OS version, Linux Desktop, etc)
Operating System: Red Hat Enterprise Linux 9.2
KDE Plasma Version: 5.27.4
KDE Frameworks Version: 5.105.0
Qt Version: 5.15.3
Kernel Version: 5.14.0-284.25.1.el9_2.x86_64 (64-bit)
Graphics Platform: X11
 
**Version since**
 Eclipse 4.31
org.eclipse.swt-3.126.0.v20240321-1329

raghucssit:
This has been reported already here.
https://github.com/eclipse-platform/eclipse.platform.swt/issues/770

**Part 1:**
The important thing to note here is the word: **sometimes**.
This is a problem with the high performance of the machine 😊.
In a low performance computer once the window is open it takes few milliseconds to java to reach next native call to read the size of the shell. By this time window has fully populated and native call responds with the actual size set.
But on high performance machine java jumps to native call much faster than the window opens. This results in native call responding back with default window size GDK.gdk_window_get_frame_extents (window, rect) : GdkRectangle {0, 0, 1, 1}.

[...]

PR #1165, opened 2024-04-07
Use known width and height of the shell instead of asking GDK.gdk_window_get_frame_extents (window, rect).

raghucssit:
For GDK.gdk_window_get_frame_extents (window, rect) to return correct values shell should be open. Once SWT asks GTK to show the shell via Native calls it takes sometime to open the shell.
On high performance machines adjustTrip is asking GTK to give it width and height before it has completely opened the shell. This results in returning default size.(which is {0,0,1,1})
So We can use the height and width which we already know when opening the shell.

 Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/1161

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Shell.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Shell.java
@@ -479,20 +479,45 @@ void addToolTip (ToolTip toolTip) {
 	toolTips = newToolTips;
 }
 
-void adjustTrim () {
+/**
+ * Resize the shell based on the trim width and height. Trim width and height
+ * adjustments are read from '$HOME/.swt/trims.prefs' for each trim style. If
+ * such a file does not exist then the trim values are queried from GTK.
+ *
+ * Note that we are trying to resize the shell which is already open. Shell
+ * already has width and height, Which is either set by setSize()/setBounds() or
+ * size is calculated based on layout and content of the shell. Without setting
+ * bounds/layout shell won't display content which is created on it.
+ *
+ * @param widthHint  Current width of the shell that is already open.
+ * @param heightHint Current height of the shell that is already open.
+ */
+void adjustTrim (int widthHint, int heightHint) {
 	if (display.ignoreTrim) return;
 	GtkAllocation allocation = new GtkAllocation ();
 	GTK.gtk_widget_get_allocation (shellHandle, allocation);
 	int width = allocation.width;
 	int height = allocation.height;
-	GdkRectangle rect = new GdkRectangle ();
 
+	//Ask Gtk for actual window frame extents
+	GdkRectangle rect = new GdkRectangle ();
 	if (!GTK.GTK4) {
 		long window = gtk_widget_get_window (shellHandle);
 		GDK.gdk_window_get_frame_extents (window, rect);
+
+		// Gtk returns default window dimensions as {0,0,1,1} if the window is not
+		// ready. If GTK returns window dimensions we can use it, if not, we will use
+		// "requested" shell size as returned by gtk_window_resize() in setVisible().
+		// See https://github.com/eclipse-platform/eclipse.platform.swt/issues/1161
+		if (rect.width > 1 || rect.height > 1) {
+			widthHint = rect.width;
+			heightHint = rect.height;
+		}
 	}
-	int trimWidth = Math.max (0, rect.width - width);
-	int trimHeight = Math.max (0, rect.height - height);
+
+
+	int trimWidth = Math.max (0, widthHint - width);
+	int trimHeight = Math.max (0, heightHint - height);
 	/*
 	* Bug in GTK.  gdk_window_get_frame_extents() fails for various window
 	* managers, causing a large incorrect value to be returned as the trim.
@@ -2936,6 +2961,8 @@ public void setVisible (boolean visible) {
 			GTK.gtk_widget_show (shellHandle);
 			GTK3.gtk_window_resize(shellHandle, init_width[0], init_height[0]);
 			resizeBounds (init_width[0], init_height[0], false);
+			oldWidth = init_width[0];
+			oldHeight = init_height[0];
 		} else {
 			GTK.gtk_widget_show (shellHandle);
 		}
@@ -2975,7 +3002,7 @@ public void setVisible (boolean visible) {
 			if (!iconic) {
 				update (true, true);
 				if (isDisposed ()) return;
-				adjustTrim ();
+				adjustTrim (oldWidth, oldHeight);
 			}
 		}
 		mapped = true;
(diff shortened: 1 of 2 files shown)


raghucssit:
Test Failure has started from here:
https://ci.eclipse.org/releng/job/eclipse.platform.swt/job/master/508/

iloveeclipse:
The problem explanation makes sense for me.

However, I have my concerns regarding the patch.

- https://docs.gtk.org/gdk3/method.Window.get_frame_extents.html is supposed to do: `Obtains the bounding box of the window, **including** window manager titlebar/borders if any.`
- https://docs.gtk.org/gdk3/method.Window.resize.html doesn't say anything if the size is with or without trims: `Resizes window; for toplevel windows, asks the window manager to resize the window`

We replac also **unconditionally** code that used shell size including trims received (in good case) from window manager with **some** size we **ask** to set (but don't know the actual size as effect of the call).

What about to use the new arguments that you've added to the `adjustTrim()` **conditionally**, only if `gdk_window_get_frame_extents()`  returns garbage values (like in the error case we have)?

raghucssit:
> * doesn't say anything if the size is with or without trims: `Resizes window; for toplevel windows, asks the window manager to resize the window`

Yes. We can ask GTK for size if it returns default then we can use our passed values.

Similar problem was existed in the past. I found it was worked around like this,
https://github.com/eclipse-platform/eclipse.platform.swt/blob/5b437ee2bd8cdda7083a9f2ceeadb172a583ed37/bundles/org.eclipse.swt/Eclipse%20SWT/gtk/org/eclipse/swt/widgets/Shell.java#L2317

raghucssit:
> The problem explanation makes sense for me.
> 
> However, I have my concerns regarding the patch.
> 
> * https://docs.gtk.org/gdk3/method.Window.get_frame_extents.html is supposed to do: `Obtains the bounding box of the window, **including** window manager titlebar/borders if any.`
> * https://docs.gtk.org/gdk3/method.Window.resize.html doesn't say anything if the size is with or without trims: `Resizes window; for toplevel windows, asks the window manager to resize the window`
> 
> We replac also **unconditionally** code that used shell size including trims received (in good case) from window manager with **some** size we **ask** to set (but don't know the actual size as effect of the call).
> 
> What about to use the new arguments that you've added to the `adjustTrim()` **conditionally**, only if `gdk_window_get_frame_extents()` returns garbage values (like in the error case we have)?

I have pushed the change on this suggestion.
Now we use known width and height if gtk returns default values.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Shell.java
```diff
@@ -479,20 +479,42 @@ void addToolTip (ToolTip toolTip) {
 	toolTips = newToolTips;
 }
 
-void adjustTrim () {
+/**
+ * Resize the shell based on the trim width and height. Trim width and height
+ * adjustments are read from 'user/.swt.trims.prefs' for each trim style. If
```
iloveeclipse, line 484: `from 'user/` -> `from '$HOME/`

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Shell.java
```diff
@@ -479,20 +479,42 @@ void addToolTip (ToolTip toolTip) {
...
+ * such a file does not exist then it uses default values initialized.
```
iloveeclipse, line 485: `then it uses default values initialized.` -> `then the trim values are queried from GTK`.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Shell.java
```diff
@@ -479,20 +479,42 @@ void addToolTip (ToolTip toolTip) {
...
+	GdkRectangle rect = new GdkRectangle ();
 	if (!GTK.GTK4) {
 		long window = gtk_widget_get_window (shellHandle);
 		GDK.gdk_window_get_frame_extents (window, rect);
 	}
-	int trimWidth = Math.max (0, rect.width - width);
-	int trimHeight = Math.max (0, rect.height - height);
+
+	// Gtk returns default window dimensions as {0,0,1,1} if the window is not
+	// ready. If Gtk returns window dimensions we can use it. See 1161
```
iloveeclipse, line 510: `If Gtk returns window dimensions we can use it.` 
->  
`If GTK returns window dimensions we can use it, if not, we will use "requested" shell size as returned by gtk_window_resize() in setVisible()`

`See 1161` -> `See https://github.com/eclipse-platform/eclipse.platform.swt/issues/1161`

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Shell.java
```diff
@@ -479,20 +479,42 @@ void addToolTip (ToolTip toolTip) {
...
+	if (rect.width > 1 || rect.height > 1) {
```
iloveeclipse, line 511: I believe this new block including comment should go **inside** the `if (!GTK.GTK4)` block above, as the bad values are returned only in this case.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Shell.java
```diff
@@ -479,20 +479,42 @@ void addToolTip (ToolTip toolTip) {
...
+ *
+ * Note that we are trying to resize the shell which is already open. Shell
+ * already has width and height, Which is either set by setSize()/setBounds() or
+ * size is calculated based on layout and content of the shell. Without setting
+ * bounds/layout shell won't display content which is created on it.
+ *
+ * @param curWidth  Current width of the shell that is already open.
+ * @param curHeight Current height of the shell that is alreday open.
+ */
+void adjustTrim (int curWidth, int curHeight) {
```
iloveeclipse, line 495: `curWidth, curHeight` - I believe we should name it `widthHint`and `heightHint`, as those are not necessarily the correct sizes (depends on actual trim)

iloveeclipse (changes requested):
I believe the patch is better now, see some improvement comments below.

raghucssit:
All review comments fixed.

iloveeclipse (approved):
Thanks, looks good now.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Shell.java
```diff
@@ -479,20 +479,45 @@ void addToolTip (ToolTip toolTip) {
 	toolTips = newToolTips;
 }
 
-void adjustTrim () {
+/**
+ * Resize the shell based on the trim width and height. Trim width and height
+ * adjustments are read from '$HOME/.swt.trims.prefs' for each trim style. If
```
iloveeclipse, line 484: `$HOME/.swt.trims.prefs` -> `$HOME/.swt/trims.prefs`

Issue #1171, opened 2024-04-13
Update documentation about how to build the SWT binaries


fedejeanne:
See https://github.com/eclipse-platform/eclipse.platform.swt/pull/1167#issuecomment-2052499527

The `Readme.md` needs to mention the new Maven configuration and the Windows/Linux readmes don't need to mention the Ant builds.

In general, re-read all documentation and fix inconsistencies/outdated info

fedejeanne:
It's been a year and I didn't find the time to get to it. 

If anyone wants to take over, feel free to do so.

PR #1167, opened 2024-04-11
Fix Maven launch config to build SWT binaries

fedejeanne:
The path was broken since the projects got moved:

image

But now the path can found:

image

Also refresh the workspace after running:

image

Adapt the documentation accordingly.

PR #1127, opened 2024-03-21
Update OUTPUT_DIR path in build.sh

lathapatil:
Fixes : https://github.com/eclipse-platform/eclipse.platform.swt/issues/1086

PR #1174, opened 2024-04-15
[Documentation] Update instructions to build SWT binaries locally

fedejeanne:
Contributes to #1171 

See https://github.com/eclipse-platform/eclipse.platform.swt/pull/1167#issuecomment-2052499527

Move some sections of the documentation to the general `Readme.md` file and change some outdated/unclear documentation in Windows.

FYI I provided 2 commits for clarity but I will squash them before merging.

### DIFF
--- a/bundles/org.eclipse.swt/Readme.Linux.md
+++ b/bundles/org.eclipse.swt/Readme.Linux.md
@@ -4,40 +4,24 @@ The SWT Linux port is based on the GTK.
 
 The natives are build using make and (by default) gcc.
 
-# Development overview
-SWT is made up of 2 parts, Java and the natives (written in C).
-
-* **Java Part**
-If you only make changes in the Java part or need to configure SWT for running
-snippets, then you don't need to recompile the natives, instead, you can just
-use the pre-compiled libraries in the binary repository.
-
-* **Natives Part**
-Make sure the binary project for your platform is imported and open in your
-workspace.
-
 ## Building Natives
 
-### Building 64bit natives for GTK3
+### Building 64bit natives for GTK
 
 You need to install the following on your system:
 
 * make
 * gcc
-* GTK+ development files (gtk3-devel)
-* Java 17 JDK
+* GTK3/GTK4 development files (see https://www.gtk.org/docs/installations/linux/)
+* Java 17 JDK (https://adoptium.net/marketplace/?version=17)
 * (optional) Webkit for GTK development files (webkit2gtk3-devel)
 
 ### Building and Testing locally
 
 Import and build the SWT project `bundles/org.eclipse.swt` and `binaries/org.eclipse.swt.gtk.linux.${arch}` in Eclipse.
 
-To build SWT natives run the `Build-SWT-native-binaries-for-running-platform` Maven launch configuration.
-Alternatively run the `build.sh` script in `binaries/org.eclipse.swt.gtk.linux.${arch}/bin/library` directory. To
-use the locally build natives run the build script with the 'install' argument
-(`./build.sh install`) or --help to see more options.
+See **Building and Testing locally** in [Readme.md](Readme.md) for further instructions.
 
-You can find additional information on GTK SWT development can be found in
-Eclipse Wiki.
+Additional information on GTK SWT development can be found in Eclipse Wiki:
 
 https://wiki.eclipse.org/SWT/Devel/Gtk/Dev_guide
--- a/bundles/org.eclipse.swt/Readme.Win32.md
+++ b/bundles/org.eclipse.swt/Readme.Win32.md
@@ -2,39 +2,14 @@
 
 The document is structured so that you only need to read up to the point that you require, with advanced topics at the bottom.
 
-The SWT windows port is based on the Win32 api.
+The SWT windows port is based on the Win32 API.
 
-The natives are compiled via an ant build.xml script with the Microsoft Visual Studio (2019) C++ compiler.
+The natives are compiled using the Microsoft Visual Studio (2019) C++ compiler.
 
-# Development overview
-SWT is made up of 2 parts, Java and the natives (written in C).
+## Building Natives.
 
-* **Java Part**  
-If you only make changes in the Java part or need to configure SWT for running snippets,
-then you don't need to recompile the natives, instead, you can just use the pre-compiled
-libraries in the binary repository.
-
-* **Natives Part**  
-(Make sure the binary project for your platform is imported and open in your workspace).
-
-## Configure workspace to run snippets.
-
-* Download & install Java *JDK* (Last tested on JDK9) http://www.oracle.com/technetwork/java/javase/downloads/index.html
-* Download and install Eclipse. Either the "Eclipse IDE for Eclipse Committers" or a recent integration build:
-  https://download.eclipse.org/eclipse/downloads/
-* (Optionally) install CDT from marketplace if you want to work on C/Native parts of SWT.
-* (if not already installed) installed EGit integration.
-* Open the git perspective. Add the following two repositories, which can be found on: (use git protocol)
-	- https://github.com/eclipse-platform/eclipse.platform.swt
-* From swt repo, import most projects (except Cocoa/Gtk, root 'swt' project, once)
-* And import the project that reflects your platform.
-* In the snippet project, add the swt project as dependency to launch snippets.
-* You should be able to run snippets now. (e.g Snippet1).
-
-## (Advanced) Building Natives.
-
-### Building just 64bit natives with no Webkit
-(Last tested on Win10 64 bit & Java 11. May 2019):
+### Building 64bit natives
+(Last tested on Win10 64 bit & Java 17. April 2024):
 
 You need to install the following on your system:  
 
@@ -42,37 +17,14 @@ You need to install the following on your system:
   https://visualstudio.microsoft.com/vs/community/
   Either select the components "MSVC C++-x64/x86-Buildtools" and "Windows 10 SDK"
   or the workload "Desktop development with C++" which includes the required components and some more.
-* (Optional) Install Cygwin
-* Install Java 11. Oracle JDK or IBM JDK:  
-  http://www.oracle.com/technetwork/java/javase/overview/java8-2100321.html  
-  https://developer.ibm.com/javasdk/downloads/  
+* Java 17 JDK (https://adoptium.net/marketplace/?version=17)
 
 ### Building and Testing locally
 
-In the binary git repo, in the windows project, you can build the natives via the build.xml. To do so:  
+Import and build the SWT project `bundles/org.eclipse.swt` and `binaries/org.eclipse.swt.win32.win32.${arch}` in Eclipse.
 
-* In Eclipse, in the binary project org.eclipse.swt.win32.win32.x86_64, right click on build.xml:  
-* Run As -> Ant Build ...
-* On the Targets tab check the build_libraries target  (it should be the only one check for just the natives).
-* On the JRE tab select "Run in the same JRE as the workspace"
-* On the Refresh tab check "Refresh resources upon completion" to refresh your workspace after running the build; this ensures that Eclipse will pick up the fresh binaries
-* Press the "Run" button to begin.
-* If compile fails, inspect build log for warnings, they can point out issues in your configuration.
+See **Building and Testing locally** in [Readme.md](Readme.md) for further instructions.
 
-To test if you binaries actually get used, you could try to checkout an older binary repo commit with old
-bindings. Then try to run a snippet and get the 'missing natives' error. Then if you build the natives
-and the snippet(s) work, it shows that you've build them successfully.
-
-### Optional additional configuration
-* Install and configure Cygwin SSH Server on Windows. (Make sure 'openssh' package is installed).
-  See: https://docs.oracle.com/cd/E24628_01/install.121/e22624/preinstall_req_cygwin_ssh.htm#EMBSC281
-* Configure the machine for password-less SSH authentication with the Hudson machine.
-  For more details you can refer: http://users.cecs.anu.edu.au/~xzhang/pubDoc/IT/SSH%20without%20password%20from%20Windows.htm  
-  Sharing some key steps below(which I recall):
-  - Generate the 'dsa' public/private key from your "swtbuild" account from windows machine.
-  - Now login to the Hudson machine with "swtbuild" account.
-  - Copy the public keys and register then on the Hudson machine.. this should enable password-less authentication.
-  
 ## (More advanced) (Optional) Building Webkit
 Note, I.E is the default backend. But Webkit is also a possible alternative Browser backend.
 
@@ -99,25 +51,3 @@ Steps:
   - `SWT_BUILD_DIR\Webkit\r72896`  
   - `SWT_BLUID_DIR\Webkit\SupportLibrary`  
   - (Note: SWT's Webkit support exists for SWT 32bit on Windows, so Webkit SDKs are consumed only by the SWT 32bit build process)
-
-## Custom Java paths and 32 bit builds
-
-You can specify a specific java sdk via:  
-  `set SWT_JAVA_HOME=<YOUR_PATH>`
-
-To do a 32 bit build, install a 32bit java & specify it's path or put it into:
-`SWT_BUILD_DIR\Java\Oracle\jdk1.8.0-latest\x86`
-
-For building windows native build for 32bit, you need to convert the source to 32bit first, by running 'replace.64.to.32' target.  
-Once sources are converted to 32bit you can run the 'build_libraries' target for ${workspace_loc:/org.eclipse.swt.win32.win32.x86/build.xml}
-
-## Hudson
-(Optional for local setup)  
-Now you can point the Windows hudson job to this machine and trigger a native build.  
-Note: For testing purpose from hudson, temporarily turn the nativeChanges flag to 'true' to force a native build compilation.
-
-# Footnotes
-[1]  https://bugs.eclipse.org/bugs/show_bug.cgi?id=526802
-
-
-
--- a/bundles/org.eclipse.swt/Readme.macOS.md
+++ b/bundles/org.eclipse.swt/Readme.macOS.md
@@ -82,13 +82,19 @@ More information on how to use the generator is available [here](https://www.ecl
 this information into this readme.
 
 
-Compiling native libs
----------------------
+## Building Natives
 
-This is done from within the binaries project.
+### Building 64bit natives for Cocoa
 
-    cd <eclipse.platform.swt>/binaries/org.eclipse.swt.cocoa.macosx.x86_64
-    mvn clean package -Dnative=cocoa.macosx.x86_64
+You need to install the following on your system:
 
+* XCode Command Line Tools (already present in XCode)
+* Java 17 JDK (https://adoptium.net/marketplace/?version=17)
+
+### Building and Testing locally
+
+Import and build the SWT project `bundles/org.eclipse.swt` and `binaries/org.eclipse.swt.cocoa.macosx.${arch}` in Eclipse.
+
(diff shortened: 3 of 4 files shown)


HannesWell (commented):
Thank you @fedejeanne for this.
I have added some comments below how this could be further enhanced.

bundles/org.eclipse.swt/Readme.Win32.md
```diff
@@ -6,29 +6,24 @@ The SWT windows port is based on the Win32 api.
...
+* Add/clone the following repository: https://github.com/eclipse-platform/eclipse.platform.swt
+* From that repository, import all the projects except
+  * Everything has to do with 'Cocoa' or 'Gtk'
+  * The root 'swt' project
+  * The root 'binaries' project
+  * The root 'bundles' project
+  * Any 'bin' project
+* Make sure you imported the projects that reflects your platform _e.g._ `win32`
+* Make sure you imported the fragment that reflects your architecture _e.g._ `org.eclipse.swt.win32.win32.x86_64`
+* In the snippet project, rename the file `.classpath_win32` to `.classpath` and refresh the project.
```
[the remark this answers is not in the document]
fedejeanne: Are you sure?
I imported the stuff on my own into an existing repository (_i.e._ NOT from a fresh "contributor" installation like it's mention in [Contributing.md](https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md#creating-an-eclipse-development-environment)) and the dependency the `swt`-project was missing.

bundles/org.eclipse.swt/Readme.Linux.md
```diff
@@ -32,7 +20,8 @@ You need to install the following on your system:
 
 Import and build the SWT project `bundles/org.eclipse.swt` and `binaries/org.eclipse.swt.gtk.linux.${arch}` in Eclipse.
 
-To build SWT natives run the `Build-SWT-native-binaries-for-running-platform` Maven launch configuration.
+See **Building and Testing locally** in [Readme.md](Readme.md)
+
 Alternatively run the `build.sh` script in `binaries/org.eclipse.swt.gtk.linux.${arch}/bin/library` directory. To
```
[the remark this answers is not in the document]
fedejeanne: 👍

fedejeanne (commented):
Thanks for the tips, @HannesWell , I changed some stuff based on them, WDYT? 

I can squash all my commits before merging.

bundles/org.eclipse.swt/Readme.Win32.md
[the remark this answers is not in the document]
HannesWell: > Are you sure? 

I'm quite sure, yes. :) This became obsolete with https://github.com/eclipse-platform/eclipse.platform.swt/pull/1008.
But maybe we are not talking about the same?

> I changed the part about "_Downlad and install Eclipse_" to use the pre-configured IDE mentioned in Contributing.md and moved it to the general `Readme.md`. This simplifies things a lot.

I'll try to have a look at that change tomorrow. It's too late now.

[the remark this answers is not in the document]
HeikoKlare: Maybe the confusion is that this is about the platform-specific classpath files for the `swt.snippets` project and not for the `swt` project. Only the latter ones have been replaced by central classpath files linked into the SWT fragment projects, haven't they?
There is also still the according Oomph setup task:
https://github.com/eclipse-platform/eclipse.platform.swt/blob/6ebc3e2b57267c7b104a7ab3e2b19d706b6b83b4/bundles/org.eclipse.swt.tools/Oomph/platformSwt.setup#L50-L53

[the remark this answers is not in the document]
fedejeanne: @HannesWell I changed the text and removed it altogether anyway, so there's no need to rename any files while following the instructions in the (new) section *Running the snippets*, which is now in `Readme.md`

HannesWell (commented):
I just had another pass over the documentation and since there were many small points to further enhance (mainly in the already existing doc, but it IMHO fits into this change) I just applied the suggestions by myself and pushed them in a separate commit.
Please incoporate them as you see fit.
A short overview is
- Removed outdated information about Hudson builds.
- Remove links and mentioning of specific JDK vendors. From my impression there is not only one source for JDK binaries anymore and I assume that everyone that wants to contribute to SWT has already set up a JDK anyways.
If we want to provide a link, I suggest to use a neutral place like:
https://adoptium.net/marketplace/?version=17
- There is no help for Windows and Mac so I removed its mentioning.

Thank you for these updates. They are definitely due.

bundles/org.eclipse.swt/Readme.Win32.md
```diff
@@ -6,29 +6,24 @@ The SWT windows port is based on the Win32 api.
...
 * Download and install Eclipse. Either the "Eclipse IDE for Eclipse Committers" or a recent integration build:
   https://download.eclipse.org/eclipse/downloads/
 * (Optionally) install CDT from marketplace if you want to work on C/Native parts of SWT.
-* (if not already installed) installed EGit integration.
-* Open the git perspective. Add the following two repositories, which can be found on: (use git protocol)
-	- https://github.com/eclipse-platform/eclipse.platform.swt
-* From swt repo, import most projects (except Cocoa/Gtk, root 'swt' project, once)
-* And import the project that reflects your platform.
-* In the snippet project, add the swt project as dependency to launch snippets.
+* Install EGit.
```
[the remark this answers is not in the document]
HannesWell: Since you have already solved this in a different way, but in general would have been simpler if we would just write that one should import `Eclipse projects` at specific locations, which would already exclude for example the `binaries` project which is just Maven project.

bundles/org.eclipse.swt/Readme.Linux.md
```diff
@@ -32,12 +20,20 @@ You need to install the following on your system:
...
+
+Alternatively run the `build.sh` script under `org.eclipse.swt\Eclipse SWT PI\gtk\library\` directory. 
+To use the locally build natives run the script with the 'install' and `clean` arguments (`./build.sh clean install`)
+or --help to see more options.
+
+The JDK against which the SWT natives are built must be specified in the `SWT_JAVA_HOME` environment variable:
+  ```
+  SWT_JAVA_HOME=<YOUR_JDK_PATH>
+  export SWT_JAVA_HOME
+  ```
```
HannesWell, line 33: I'm not so much a Linux export, but from the existing build scripts I think that's the right thing to do.
Nevertheless it would be great if someone with more knowledge could confirm this.

bundles/org.eclipse.swt/Readme.macOS.md
```diff
@@ -82,13 +82,30 @@ More information on how to use the generator is available [here](https://www.ecl
...
 
-This is done from within the binaries project.
+### Building 64bit natives for Cocoa
 
-    cd <eclipse.platform.swt>/binaries/org.eclipse.swt.cocoa.macosx.x86_64
-    mvn clean package -Dnative=cocoa.macosx.x86_64
+You need to install the following on your system:
 
+* TBD
+* Java 17 JDK
```
HannesWell, line 92: @HeikoKlare or @Phillipus can you name the tools that have to be installed on a mac in order to be able to build SWT locally for mac?

Phillipus: @HannesWell i just have Java and Maven.

HannesWell: So on Mac everything to compile C code is already installed out of the box?

Phillipus: Good point. I have Apple's `Command Line Tools` installed which is kind of a lightweight version of `Xcode`. So I would add that one of these needs to be installed as a requirement.

bundles/org.eclipse.swt/Readme.Linux.md
HeikoKlare: Yes, it is correct to set `SWT_JAVA_HOME` (this way).

Still, the instruction do not work as the whole setup relies on being executed in a properly configured binary fragment project and not within the general `org.eclipse.swt` project:
* The default `OUTPUT_DIR` specification in `build.sh` assumes that the `build.sh` is _not_ executed from within the `org.eclipse.swt` project but from without the symlink created for the binaries fragment project, see #1127.
* There are further files required (like `make_common.mak`), which will only be available when properly linked together in a OS-/WS-/Arch-specific fragment project.

So the instruction have to be rewritten to be executed from within a binaries fragment, but then you need to describe as well how to set up that fragment (create the symlink etc, which is automatically done by Eclipse or the Maven build, respectively). Since it is very easy to create the binaries via Maven, I am not sure whether it is worth to describe all that manual setup (and keep it up to date), as that is already done by Maven (and thus documented in the Maven configuration). Whoever does not want to use Maven can just replicate by hand what Maven does (as that's what has to be described in these instructions otherwise).

bundles/org.eclipse.swt/Readme.Win32.md
```diff
@@ -2,77 +2,38 @@
...
+(Last tested on Win10 64 bit & Java 17. April 2024):
 
 You need to install the following on your system:  
 
 * Microsoft Visual Studio - Community and Windows 10 SDK:  (1*)
   https://visualstudio.microsoft.com/vs/community/
   Either select the components "MSVC C++-x64/x86-Buildtools" and "Windows 10 SDK"
   or the workload "Desktop development with C++" which includes the required components and some more.
+* Java 17 JDK
 * (Optional) Install Cygwin
```
HeikoKlare, line 21: I would remove the reference to Cygwin, even when marked as "optional".

bundles/org.eclipse.swt/Readme.Win32.md
```diff
@@ -2,77 +2,38 @@
...
-  - Generate the 'dsa' public/private key from your "swtbuild" account from windows machine.
-  - Now login to the Hudson machine with "swtbuild" account.
-  - Copy the public keys and register then on the Hudson machine.. this should enable password-less authentication.
-  
+Import and build the SWT project `bundles/org.eclipse.swt` and `binaries/org.eclipse.swt.win32.win32.${arch}` in Eclipse.
+
+See **Building and Testing locally** in [Readme.md](Readme.md)
+
+Alternatively run the `build.bat` script under `org.eclipse.swt\Eclipse SWT PI\win32\library\` directory.
+To use the locally build natives run the script with the 'install' and `clean` arguments (`.\build.bat clean install`).
```
HeikoKlare, line 30: Does not work for the same reasons as described for Linux (except the `OUTPUT_DIR` problem).

bundles/org.eclipse.swt/Readme.macOS.md
HeikoKlare: I can confirm that `Command Line Tools` are sufficient. I am not sure whether it is worth to note that if you want to build the binaries with gcc, you need to install that on MacOS by hand. By default `cc` will execute a `clang` compiler on MacOS.

bundles/org.eclipse.swt/Readme.macOS.md
```diff
@@ -82,13 +82,30 @@ More information on how to use the generator is available [here](https://www.ecl
...
-Happy development!
\ No newline at end of file
+### Building and Testing locally
+
+Import and build the SWT project `bundles/org.eclipse.swt` and `binaries/org.eclipse.swt.cocoa.macosx.${arch}` in Eclipse.
+
+See **Building and Testing locally** in [Readme.md](Readme.md)
+
+Alternatively run the `build.sh` script under `org.eclipse.swt\Eclipse SWT PI\cocoa\library\` directory.
+To use the locally build natives run the script with the 'install' and `clean` arguments (`./build.sh clean install`).
```
HeikoKlare, line 101: Does not work, see explanation for Linux.

bundles/org.eclipse.swt/Readme.md
```diff
@@ -3,12 +3,32 @@ org.eclipse.swt
...
+snippets then you don't need to recompile the natives and you can just
+use the pre-compiled libraries in the binary repository instead.
+
+* **Natives Part**
+
+Make sure the binary project for your platform is imported and open in your workspace,
+_for example_
+  * `binaries/org.eclipse.swt.win32.win32.x86_64` for Windows or
+  * `binaries/org.eclipse.swt.gtk.linux.x86_64` for Linux or
+  * `binaries/org.eclipse.swt.cocoa.macosx.x86_64` for Mac
```
HeikoKlare, line 23: Maybe it is worth to also make some note on `aarch64`, as currently these examples only refer to fragments for environments with x86 architecture but ARM is becoming more and more common.

HeikoKlare (commented):
If I am not mistaken, the descriptions how to build the binaries by hand (directly executing the `build.sh` / `build.bat` scripts), do not work as described. I would propose to completely remove these descriptions, as there is a Maven configuration that does everything required and replication all the information in a documentation leads to unnecessary duplications that needs to be kept consistent.

HannesWell:
> Since it is very easy to create the binaries via Maven, I am not sure whether it is worth to describe all that manual setup (and keep it up to date), as that is already done by Maven (and thus documented in the Maven configuration). 

Just as a quick-reply for now: 
Yes, I agree. But to make it simpler to find the relevant part in the pom I suggest to also add a perma-link to the configuration of the corresponding antrun-mojo configuration.
Of course this might also change in the future, but if it is forgotten to update the link (which actually can be done after the fact since it contains the commit Id :/), then it is at least easier to find to up-to-date info.

bundles/org.eclipse.swt/Readme.Linux.md
fedejeanne: I removed all the manual steps in d556545966b1beefd598027b20675dfa34cde7e6.

@HannesWell we can add a permanent link or other further improvements after this PR ;-)

bundles/org.eclipse.swt/Readme.macOS.md
fedejeanne: Removed in d556545966b1beefd598027b20675dfa34cde7e6

bundles/org.eclipse.swt/Readme.md
fedejeanne: Mentioned in d556545966b1beefd598027b20675dfa34cde7e6

fedejeanne:
@HannesWell , @HeikoKlare , thank you for the comments, I addressed almost all of them in d556545966b1beefd598027b20675dfa34cde7e6 and made some other minor changes. I am also in favor of simplifying these instructions as much as possible and let the "hackers" figure out the alternative ways (in case they don't want to run Maven) on their own.

I'm positive that the current state of these instructions is good enough to compile the binaries in the majority of the setups.

Once this PR is good to go, I will squash the commits.

@HannesWell We can add a permanent link to the desired commit after merging this PR if you want to. Just to be sure: you want to link this line (but not in master), right? 

https://github.com/eclipse-platform/eclipse.platform.swt/blob/979d3f1580b4786aee1efe046da8a6f836300804/binaries/pom.xml#L93

How would you phrase it in the documentation (in the follow-up PR)?

fedejeanne:
The failing tests are unrelated

SyntevoAlex:
Sorry, but inverse changes are actually needed.

You are removing steps for manual building, but that's exactly what I was always doing, and I was pretty much committer number one for a few years.

I was running steps manually because I'm not using Eclipse for developing SWT. And when I run maven from IDEA, it downloads unneeded stuff (tycho etc) for some 20 minutes before actually building.

HeikoKlare:
> You are removing steps for manual building, but that's exactly what I was always doing, and I was pretty much committer number one for a few years.

There seems to be some confusion here, as the proposed change removes nothing but redundant and outdated documentation. You can still do a "manual build" (i.e., do by hand what Maven does automatically). Having an explanation of that process in the documentation would be nothing else than writing down the according Maven step in natural language:
https://github.com/eclipse-platform/eclipse.platform.swt/blob/979d3f1580b4786aee1efe046da8a6f836300804/binaries/pom.xml#L92-L134

So as @HannesWell proposed (https://github.com/eclipse-platform/eclipse.platform.swt/pull/1174#issuecomment-2068594910), adding a link to the Maven configuration in the documentation for those who want to perform the steps by hand would be good.

SyntevoAlex:
I still think that instead of removing documentation, it should be fixed, so removing is half step in wrong direction. It is not redundant, because parsing maven files is hard for people who never dealt with maven before (such as me).

I'm merely commenting and do not intend to hold this PR.

merks:
I think the point is that while the actual logic as spelled out in the pom.xml may evolve in the future (logic that is well tested by the builds and by the other developers) the documentation will forever need to evolve manually in lockstep.  That's likely to be forgotten or not to happen.  I image that most people will be happy if they can invoke Maven without further thought about the details, and even if that takes 20 minutes the first time to fill the cache, that's probably for most people better tradeoff than trying to read and to follow more detailed documentation instructions.

So I do understand your concerns, I think this is a little like with Oomph where it's must easier to maintain something that's regularly and automatically used, than it is to maintain documentation that's rarely read, rarely used, and error prone to follow exactly to the letter...

fedejeanne:
@SyntevoAlex I agree with @HannesWell , @HeikoKlare and @merks here: pointing to the Maven goal that does the trick (in a follow-up PR) is enough here so I wouldn't hold this PR just to add the "manual" steps which are likely to be outdated in the near future anyway.

So as soon as @HeikoKlare and @HannesWell give me the green light (I hope I addressed every one of your comments), I'm squashing + merging :-)

bundles/org.eclipse.swt/Readme.Linux.md
```diff
@@ -4,40 +4,29 @@ The SWT Linux port is based on the GTK.
...
 Import and build the SWT project `bundles/org.eclipse.swt` and `binaries/org.eclipse.swt.gtk.linux.${arch}` in Eclipse.
 
-To build SWT natives run the `Build-SWT-native-binaries-for-running-platform` Maven launch configuration.
-Alternatively run the `build.sh` script in `binaries/org.eclipse.swt.gtk.linux.${arch}/bin/library` directory. To
-use the locally build natives run the build script with the 'install' argument
-(`./build.sh install`) or --help to see more options.
+The JDK against which the SWT natives are built must be specified in the `SWT_JAVA_HOME` environment variable:
+  ```
+  export SWT_JAVA_HOME=<YOUR_JDK_PATH>
+  ```
```
HeikoKlare, line 26: Setting `SWT_JAVA_HOME` should not be necessary when using the Maven build.

HeikoKlare:
> I still think that instead of removing documentation, it should be fixed, so removing is half step in wrong direction. It is not redundant, because parsing maven files is hard for people who never dealt with maven before (such as me).

I understand your point that it is harder to parse Maven files than reading documentation that explains the Maven files in comparably simple language (even though that does not change anything w.r.t. it being redundant). However, maintaining the documentation is even harder, as only very few people will need that documentation (as stated by @merks). In my opinion, it even leads to the opposite: the people that may be interested in compiling the natives will run across all the information on how to execute the build.sh/build.bat and be tempted to follow that approach instead of simply executing the Maven build. So the documentation would have to be structured very well to lead the people in the proper direction, i.e., to just use the simple, Maven-based solution and understand that all the information about the manual build is irrelevant for them.

bundles/org.eclipse.swt/Readme.Linux.md
fedejeanne: Removed it from all 3 platforms in 74bda8a9bb2fd3618d2c3b98c41fb22fe9074947

fedejeanne:
Addressed https://github.com/eclipse-platform/eclipse.platform.swt/pull/1174#discussion_r1576022056 and also added the perma-link from https://github.com/eclipse-platform/eclipse.platform.swt/pull/1174#issuecomment-2071705795 in 74bda8a9bb2fd3618d2c3b98c41fb22fe9074947

HannesWell (commented):
Hoping that it will be a good compromise I made more suggestions for improvements below.

Issue #1162, opened 2024-04-04
StyledText IME (Japanese) error 


tueterich:
IllegalArgumentException at org.eclipse.swt.custom.DefaultContent.getLineAtOffset(575). 
[ime_exception.txt](https://github.com/eclipse-platform/eclipse.platform.swt/files/14865622/ime_exception.txt)

This happens when you press CTRL+BACKSPACE when the IME Popup is open and some text is already present in a StyledText Control.

![screenshot_ime](https://github.com/eclipse-platform/eclipse.platform.swt/assets/15818853/b7891a7e-0749-4955-8b3e-bd05461c3aac)

Steps to reproduce:
- Launch the attached Snippet and focus the StyledText Control
- Switch Keyboard Settings to Japanese (Microsoft-IME)
- Activate Hiragana or any other Option from the IME Menu
- Type something and choose a suggestion from the IME Popup
- Type something else so the IME Popup opens again
- Press CTRL+BACKSPACE while the Popup is open

There are some other cases where the exception is thrown when using the IME but I don't know how to reproduce them:
[ime_exception_2.txt](https://github.com/eclipse-platform/eclipse.platform.swt/files/14872705/ime_exception_2.txt)

OS: Win 11

Tested with:
Oracle JDK 8 / SWT 4.919 
and 
Corretto JDK 21 and SWT 4.963

Snippet:
[StyledTextIMEErrorSnippet.txt](https://github.com/eclipse-platform/eclipse.platform.swt/files/14872723/StyledTextIMEErrorSnippet.txt)

SyntevoAlex:
Full exception text from attached file:
```
Exception in thread "main" java.lang.IllegalArgumentException: Argument not valid
	at org.eclipse.swt.SWT.error(SWT.java:4576)
	at org.eclipse.swt.SWT.error(SWT.java:4510)
	at org.eclipse.swt.SWT.error(SWT.java:4481)
	at org.eclipse.swt.custom.DefaultContent.error(DefaultContent.java:682)
	at org.eclipse.swt.custom.DefaultContent.getLineAtOffset(DefaultContent.java:575)
	at org.eclipse.swt.custom.StyledTextRenderer.getTextLayout(StyledTextRenderer.java:1141)
	at org.eclipse.swt.custom.StyledTextRenderer.getTextLayout(StyledTextRenderer.java:871)
[...]

PR #1176, opened 2024-04-16
fix IllegalArgumentException when deleting text while IME open #1162

jukzi:
If text is deleted while the IME assistent is open the start of the open IME has to be adjusted accordingly

https://github.com/eclipse-platform/eclipse.platform.swt/issues/1162


tueterich:
@jukzi Do you think its possible that this fix is going to make it into SWT Version 4.919?

jukzi:
i just submited it.

Issue #1182, opened 2024-04-18
Scrollable.getScrollbarsMode returns SWT.SCROLLBAR_OVERLAY even when overlay scrolling is disabled on Linux/GTK3


wiresketch:
**Describe the bug**
On Linux/GTK3 overlay scrolling can be disabled by setting the GTK_OVERLAY_SCROLLING=0 environment variable. However the Scrollable.getScrollbarsMode() still returns SWT.SCROLLBAR_OVERLAY in this case. Scrollable.gtk_draw method already checks this environment variable on GTK3 using the OS.GTK_OVERLAY_SCROLLING_DISABLED constant populated from it. So it would only be logical to use the same kind of check in Scrollable.getScrollbarsMode() to return a consistent result.

**To Reproduce**

[ScrollbarsModeOverlayTest.java.txt](https://github.com/eclipse-platform/eclipse.platform.swt/files/15027110/ScrollbarsModeOverlayTest.java.txt)

Run the attached snippet with and without GTK_OVERLAY_SCROLLING=0 setting. In both cases the result of Scrollable.getScrollbarsMode() will be printed as SWT.SCROLLBAR_OVERLAY.
  
**Expected behavior**
With GTK_OVERLAY_SCROLLING=0 the result of Scrollable.getScrollbarsMode() should be SWT.NONE

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [ ] Windows
- - [x] Linux
- - [ ] macOS

iloveeclipse:
@wiresketch : do you want to provide a patch?

iloveeclipse:
The patch I have in mind would be:
```
diff --git a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Scrollable.java b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Scrollable.java
index 0157962..e19aad6 100644
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Scrollable.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Scrollable.java
@@ -306,3 +306,5 @@
 	if (GTK.gtk_scrolled_window_get_overlay_scrolling(scrolledHandle)) {
-		return SWT.SCROLLBAR_OVERLAY;
[...]

PR #1183, opened 2024-04-18
Fix Scrollable.getScrollbarsMode to be consistent #1182

wiresketch:
Fix Scrollable.getScrollbarsMode to return SWT.NONE when GTK_OVERLAY_SCROLLING=0 setting is present for Linux/GTK3.

Fixes: #1182

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Scrollable.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Scrollable.java
@@ -303,6 +303,11 @@ public ScrollBar getHorizontalBar () {
  */
 public int getScrollbarsMode () {
 	checkWidget();
+	if (!GTK.GTK4) {
+		if (OS.GTK_OVERLAY_SCROLLING_DISABLED) {
+			return SWT.NONE;
+		}
+	}
 	if (GTK.gtk_scrolled_window_get_overlay_scrolling(scrolledHandle)) {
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Scrollable.java
```diff
@@ -303,6 +303,11 @@ public ScrollBar getHorizontalBar () {
  */
 public int getScrollbarsMode () {
 	checkWidget();
+	if (!GTK.GTK4) {
```
iloveeclipse, line 306: Is this different on GTK4?

wiresketch: On GTK4 GTK_OVERLAY_SCROLLING variable is not supported anymore. For reference see [GTK3 environment variables:](https://docs.gtk.org/gtk3/running.html#environment-variables) and [GTK4 environment variables](https://docs.gtk.org/gtk4/running.html#environment-variables) where GTK_OVERLAY_SCROLLING variable is not listed anymore.

iloveeclipse (approved):

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Scrollable.java
iloveeclipse: thanks

PR #1187, opened 2024-04-19
[win][edge] add focus delegate to browser

### LINKED ISSUE #1139: Browser control with SWT.EDGE does not emit Activate/Decactivate events and does not activate parts in an RCP application
**Describe the bug**
In an Eclipse RCP application with a Browser control with SWT.EDGE in a view, that view does not get activated when another view is activate and a left click in the view with the browser is performed. It works as expected when SWT.NONE is used instead of SWT.EDGE.

An event listener on SWT.Activate and SWT.Deactivate is never called for a Browser with SWT.EDGE, and even if I don't know if this is related, it would be my guess.

**To Reproduce**
There is an example RCP application at https://github.com/pms1/swt-browser-activate that can be used to reproduce the problem. It opens 3 views, one with Browser SWT.NONE, one with Browser SWT.EDGE and one with a button. Clicking into those views shows that the view with the EDGE browser never becomes activated and that no events are emitted (debug output on System.out).

**Expected behavior**
Browser with SWT.EDGE behaves like a Browser with SWT.NONE regarding activation and deactivation.

**Screenshots**
n.a.

**Environment:**
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 

Windows 10 22H2 (Build 19045.4170)
WebView2 runtime 122.0.2365.92

3. JRE/JDK version
Java 17.0.10 (version included in Eclipse 2024-03)
 
**Version since**
Observed in SWT 3.125 from Eclipse 2024-03 (4.31.0). Not checked with older versions.

n.a.

eneufeld:
Currently when the ICoreWebView2Controller gets the focus this is not reflected on the browser widget.
This adds a GotFocus handler and forces the focus on the browser widget.

Fixes #1139

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
@@ -426,6 +426,9 @@ public void create(Composite parent, int style) {
 	handler = newCallback(this::handleMoveFocusRequested);
 	controller.add_MoveFocusRequested(handler, token);
 	handler.Release();
+	handler = newCallback(this::handleGotFocus);
+	controller.add_GotFocus(handler, token);
+	handler.Release();
 	if (webView_2 != null) {
 		handler = newCallback(this::handleDOMContentLoaded);
 		webView_2.add_DOMContentLoaded(handler, token);
@@ -783,6 +786,11 @@ int handleNewWindowRequested(long pView, long pArgs) {
 			inNewWindow = false;
 		}
 	});
+	return COM.S_OK;
+}
+
+int handleGotFocus(long pView, long pArg) {
+	this.browser.forceFocus();
 	return COM.S_OK;
 }
 


akurtakov:
@fedejeanne  Maybe this one is of interest for you? IMO IE days are counted so fixes/improvements like this one are critical for Windows port.

fedejeanne:
@akurtakov thanks for the heads-up! I'll try to review this today/tomorrow and hopefully learn a lot from it :-) . Would you say this is an "easy" PR (keep in mind, I'm still a newbie when it comes to reviewing and testing PRs in SWT)

akurtakov:
It looks quite small and isolate so testing and verifying it looks "easy". Furthermore Edge not being default (IIRC) makes it even safer.

eneufeld:
hi @fedejeanne ,

thank you for testing this. I also retested in more detail and I can reproduce your behavior.
The activation fails quite randomly for me and I can't pinpoint what is causing this. Going from plain swt to the browser content seems to be more stable than switching between the browser contents.
Unfortunately I have no idea about the Edge internals and this looks like some weird behavior of Edge that it doesn't trigger an event.
I have no real idea where to look here. If anybody has a clue where to debug this I could try to take a look.

PR #1201, opened 2024-04-30
Adapt GitHub Actions configuration to execute MacOS on x86_64 image

HeikoKlare:
<s>GitHub Actions changed their latest images for MacOS to use the AArch64 instead of the x86_64 architecture. The SWT tests are currently failing because they still assume `macos-latest` to be an x86_64 architecture and therefore use the according SWT binaries.

With this change, the proper SWT binaries (AArch64) are used for the MacOS image.

As an alternative, we could use the `macos-latest-large` image, which uses latest x86_64, but since for Mac devices AArch64 is becoming kind of the new normal, I would propose to stick to `macos-latest` and use the according binaries instead. Feel free to share your opinion on this.</s>

GitHub Actions changed their latest images for MacOS to use the aarch64 instead of the x86_64 architecture. The SWT tests are currently failing because they still assume macos-latest to be an x86_64 architecture and therefore use the according SWT binaries.

With this change, the used GitHub actions runner is changed from macos-latest to macos-latest-large to ensure that still the x86_64 build is used. This is supposed to restore the existing purpose of the configuration. Changing to or adding an aarch64 runner for MacOS can be done subsequently.

See also https://github.com/actions/runner-images

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -31,7 +31,7 @@ jobs:
         config: 
           - { name: Linux, os: ubuntu-latest, native: gtk.linux.x86_64 }
           - { name: Windows, os: windows-latest, native: win32.win32.x86_64 }
-          - { name: MacOS, os: macos-latest, native: cocoa.macosx.x86_64 }
+          - { name: MacOS, os: macos-latest-large, native: cocoa.macosx.x86_64 }
     name: Verify ${{ matrix.config.name }} with Java-${{ matrix.java }}
     steps:
     - name: checkout swt


laeubi:
@HeikoKlare what do you think about _extending_ the matrix to test _both_ macOS variants?

HeikoKlare:
> what do you think about _extending_ the matrix to test _both_ macOS variants?

Sounds reasonable. From my understanding, resources for standard runners are unlimited for public repos, so adding another configuration to the matrix should not be a problem.

The build still fails with the new configuration because the `jnilib` files seem to be incorrect (they were initially marked as incompatible because of being x86_64 instead of aarch64):
```
java.lang.UnsatisfiedLinkError: 
2024-04-30T09:37:45.9205340Z Could not load SWT library. Reasons: 
2024-04-30T09:37:45.9307200Z 	no swt-cocoa-4965r8 in java.library.path: /Users/runner/Library/Java/Extensions:/Library/Java/Extensions:/Network/Library/Java/Extensions:/System/Library/Java/Extensions:/usr/lib/java:.
2024-04-30T09:37:45.9409700Z 	no swt-cocoa in java.library.path: /Users/runner/Library/Java/Extensions:/Library/Java/Extensions:/Network/Library/Java/Extensions:/System/Library/Java/Extensions:/usr/lib/java:.
2024-04-30T09:37:45.9511850Z 	no swt in java.library.path: /Users/runner/Library/Java/Extensions:/Library/Java/Extensions:/Network/Library/Java/Extensions:/System/Library/Java/Extensions:/usr/lib/java:.
2024-04-30T09:37:45.9614120Z 	Can't load library: /Users/runner/work/eclipse.platform.swt/eclipse.platform.swt/tests/org.eclipse.swt.tests/target/.swt/lib/macosx/aarch64/libswt-cocoa-4965r8.jnilib
2024-04-30T09:37:45.9716200Z 	Can't load library: /Users/runner/work/eclipse.platform.swt/eclipse.platform.swt/tests/org.eclipse.swt.tests/target/.swt/lib/macosx/aarch64/libswt-cocoa.jnilib
2024-04-30T09:37:45.9818000Z 	Can't load library: /Users/runner/work/eclipse.platform.swt/eclipse.platform.swt/tests/org.eclipse.swt.tests/target/.swt/lib/macosx/aarch64/libswt.jnilib
2024-04-30T09:37:45.9920630Z 	/Users/runner/work/eclipse.platform.swt/eclipse.platform.swt/tests/org.eclipse.swt.tests/target/.swt/lib/macosx/aarch64/libswt-cocoa-4965r8.jnilib: dlopen(/Users/runner/work/eclipse.platform.swt/eclipse.platform.swt/tests/org.eclipse.swt.tests/target/.swt/lib/macosx/aarch64/libswt-cocoa-4965r8.jnilib, 0x0001): tried: '/Users/runner/work/eclipse.platform.swt/eclipse.platform.swt/tests/org.eclipse.swt.tests/target/.swt/lib/macosx/aarch64/libswt-cocoa-4965r8.jnilib' (not a mach-o file), '/System/Volumes/Preboot/Cryptexes/OS/Users/runner/work/eclipse.platform.swt/eclipse.platform.swt/tests/org.eclipse.swt.tests/target/.swt/lib/macosx/aarch64/libswt-cocoa-4965r8.jnilib' (no such file), '/Users/runner/work/eclipse.platform.swt/eclipse.platform.swt/tests/org.eclipse.swt.tests/target/.swt/lib/macosx/aarch64/libswt-cocoa-4965r8.jnilib' (not a mach-o file)
```

Looks as if they were not checked out from LFS properly. For testing purposes, I change the matrix configuration to use `macos-latest-large` with x86_64 to see if that works again. Then we can extend the matrix with the aarch64 configuration.

HeikoKlare:
Switching back to the x86_64 runner for MacOS solves the build issues and restores the existing state, i.e., the intent of the existing configuration to run on all three major OS with x86_64 architecture. I propose to merge the PR this way to restore the existing configuration and get the builds running again.

Subsequently, we can add another configuration for aarch64 to the matrix, as proposed by @laeubi. It seems to require some additional configuration, as the initial test run with that runner did not succeed.

akurtakov:
Fully agreed. Let's get things to non-broken state with priority.

PR #1203, opened 2024-04-30
Reduce Win32 FontRegistries visibility and move tests inside fragment

HeikoKlare:
The classes DefaultSWTFontRegistry and ScalingSWTFontRegistry are currently public only to make them testable from a test bundle. There is currently no pre-configured way to implement unit tests for classes that are not directly accessible via public API.

With this change, the SWT Win32 fragments are extended by a configuration for providing fragment-internal unit tests executed with plain Maven surefire, which can access the classes under test even with reduced visibility. Due to current limitations in Tycho (see https://github.com/eclipse-tycho/tycho/issues/3803), an according temporary extension to the Maven configuration for the binaries projects is made. This configuration is applied to the tests for the DefaultSWTFontRegistry and ScalingSWTFontRegistry, such that their visibilities can properly be reduced to package visibility.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/internal/PlatformSpecificExecution.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/internal/PlatformSpecificExecution.java
@@ -1,0 +1,28 @@
+package org.eclipse.swt.internal;
+
+import static org.junit.Assume.assumeTrue;
+
+import java.net.*;
+
+public final class PlatformSpecificExecution {
+	private PlatformSpecificExecution()  {
+	}
+
+	public static void assumeIsFittingPlatform() {
+		assumeTrue("test is specific for Windows", isFittingOS());
+		assumeTrue("architecture of platform does not match", isFittingArchitecture());
+	}
+
+	private static boolean isFittingOS() {
+		return Library.os().equals("win32");
+	}
+
+	private static boolean isFittingArchitecture() {
+		Class<?> thisClass = PlatformSpecificExecution.class;
+		String thisClassResourcePath = thisClass.getName().replace('.', '/')  + ".class";
+		URL thisClassURL = thisClass.getClassLoader().getResource(thisClassResourcePath); //$NON-NLS-1$
+		return thisClassURL.toString().contains(Library.arch());
+	}
+
+}
+
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 bin/
+bin_test/
 *.log
 target/
 /.project
--- a/binaries/.classpath_win32
+++ b/binaries/.classpath_win32
@@ -25,5 +25,11 @@
 	<classpathentry kind="src" path="Eclipse SWT Browser/win32"/>
 	<classpathentry kind="src" path="Eclipse SWT OpenGL/win32"/>
 	<classpathentry kind="src" path="Eclipse SWT OpenGL/common"/>
+	<classpathentry kind="src" output="bin_test" path="Eclipse SWT Tests/win32">
+		<attributes>
+			<attribute name="test" value="true"/>
+		</attributes>
+	</classpathentry>
+	<classpathentry kind="con" path="org.eclipse.jdt.junit.JUNIT_CONTAINER/5"/>
 	<classpathentry kind="output" path="bin"/>
 </classpath>
--- a/binaries/org.eclipse.swt.win32.win32.aarch64/.project
+++ b/binaries/org.eclipse.swt.win32.win32.aarch64/.project
@@ -107,6 +107,12 @@
 			<type>2</type>
 			<locationURI>SWT_HOST_PLUGIN/Eclipse%20SWT%20WebKit</locationURI>
 		</link>
+		<link>
+			<name>Eclipse SWT Tests</name>
+			<type>2</type>
+			<!-- Once Tycho supports custom variables, SWT_HOST_PLUGIN should be used here: https://github.com/eclipse-tycho/tycho/issues/3820 -->
+			<locationURI>PARENT-2-PROJECT_LOC/bundles/org.eclipse.swt/Eclipse%20SWT%20Tests</locationURI>
+		</link>
 	</linkedResources>
 	<variableList>
 		<variable>
--- a/binaries/org.eclipse.swt.win32.win32.x86_64/.project
+++ b/binaries/org.eclipse.swt.win32.win32.x86_64/.project
@@ -107,6 +107,12 @@
 			<type>2</type>
 			<locationURI>SWT_HOST_PLUGIN/Eclipse%20SWT%20WebKit</locationURI>
 		</link>
+		<link>
+			<name>Eclipse SWT Tests</name>
+			<type>2</type>
+			<!-- Once Tycho supports custom variables, SWT_HOST_PLUGIN should be used here: https://github.com/eclipse-tycho/tycho/issues/3820 -->
+			<locationURI>PARENT-2-PROJECT_LOC/bundles/org.eclipse.swt/Eclipse%20SWT%20Tests</locationURI>
+		</link>
 	</linkedResources>
 	<variableList>
 		<variable>
--- a/binaries/pom.xml
+++ b/binaries/pom.xml
@@ -45,6 +45,25 @@
 	<build>
 		<plugins>
 			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-surefire-plugin</artifactId>
+				<dependencies>
+					<dependency>
+						<groupId>org.apache.maven.surefire</groupId>
+						<artifactId>surefire-junit-platform</artifactId>
+						<version>3.2.5</version>
+					</dependency>
+				</dependencies>
+				<executions>
+					<execution>
+						<id>execute-tests</id>
+						<goals>
+							<goal>test</goal>
+						</goals>
+					</execution>
+				</executions>
+			</plugin>
+			<plugin>
 				<groupId>org.eclipse.tycho</groupId>
 				<artifactId>tycho-apitools-plugin</artifactId>
 				<version>${tycho.version}</version>
--- a/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/internal/DefaultSWTFontRegistryTests.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/internal/DefaultSWTFontRegistryTests.java
@@ -22,12 +22,18 @@ import org.eclipse.swt.graphics.FontData;
 import org.eclipse.swt.widgets.Display;
 import org.junit.After;
 import org.junit.Before;
+import org.junit.BeforeClass;
 import org.junit.Test;
 
 public class DefaultSWTFontRegistryTests {
 	private static String TEST_FONT = "Helvetica";
 	private Display display;
 	private SWTFontRegistry fontRegistry;
+
+	@BeforeClass
+	public static void assumeIsFittingPlatform() {
+		PlatformSpecificExecution.assumeIsFittingPlatform();
+	}
 
 	@Before
 	public void setUp() {
--- a/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/internal/ScalingSWTFontRegistryTests.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/internal/ScalingSWTFontRegistryTests.java
@@ -23,11 +23,17 @@ import org.eclipse.swt.graphics.FontData;
 import org.eclipse.swt.widgets.Display;
 import org.junit.After;
 import org.junit.Before;
+import org.junit.BeforeClass;
 import org.junit.Test;
 
 public class ScalingSWTFontRegistryTests {
 	private static String TEST_FONT = "Helvetica";
 	private SWTFontRegistry fontRegistry;
+
+	@BeforeClass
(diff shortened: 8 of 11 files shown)


binaries/org.eclipse.swt.win32.win32.x86_64/build.properties
```diff
@@ -35,8 +35,14 @@ source.. = \
 	../../bundles/org.eclipse.swt/Eclipse SWT Browser/win32,\
 	../../bundles/org.eclipse.swt/Eclipse SWT OpenGL/win32,\
 	../../bundles/org.eclipse.swt/Eclipse SWT OpenGL/common
+source.test = \
```
laeubi, line 38: Test folders should usually not be part of the `source.xxx` /  `bin.xxx` of PDE...

binaries/org.eclipse.swt.win32.win32.x86_64/build.properties
```diff
@@ -35,8 +35,14 @@ source.. = \
...
+	../../bundles/org.eclipse.swt/Eclipse SWT Tests/win32
 output.. = bin/
+output.test = bin_test/
 
 pom.model.property.os=win32
 pom.model.property.ws=win32
 pom.model.property.arch=x86_64
+
+additional.bundles = junit-jupiter-api,\
```
laeubi, line 47: For test folder use a JUNIT Classpath container instead of additional bundles (what server a completely different purpose)

HeikoKlare: I wanted to ask you exactly what you think is the proper approach here, as additional.bundles were only my first attempt to get things running. I have added this as a means to make Tycho have the JUnit dependencies at compile-/test-time. I've already specified a JUnit classpath container in the `.classpath`, but that's not considered for the Tycho build, is it? At least Tycho complains about missing JUnit dependencies when compiling the test sources. Probably it would be better to add the JUnit dependency directly to the Maven build?

binaries/org.eclipse.swt.win32.win32.x86_64/build.properties
HeikoKlare: I think I am not yet sure how to properly make Tycho compile the test classes. With this configuration, Tycho properly generates test classes in a separate output folder to be used for executing the maven-surefire-plugin. What would be the proper approach to achieve that?

binaries/org.eclipse.swt.win32.win32.x86_64/build.properties
laeubi: > I've already specified a JUnit classpath container in the .classpath, but that's not considered for the Tycho build, is it? 

This is supported since [Tycho 2.4](https://github.com/eclipse-tycho/tycho/blob/main/RELEASE_NOTES.md#support-resolving-of-junit-classpath-container)

> At least Tycho complains about missing JUnit dependencies

Maybe a bug or misconfiguration, at least we have some test for this [here](https://github.com/eclipse-tycho/tycho/tree/main/tycho-its/projects/compiler.junitcontainer) that *should* prove it works.

> Probably it would be better to add the JUnit dependency directly to the Maven build?

While it maybe works, the purpose of Tycho is to not need special maven configuration so it might be a good opportunity to find out why it does not work and either improve Tycho or [the documentation](https://tycho.eclipseprojects.io/doc/latest/TestingBundles.html#maven-surefire-plugin) (sadly it misses the part about eclipse setup currently).

HeikoKlare: Then something does obviously not work here as expected:
* Tycho does not consider the test source folder specified via the `.classpath`. If I understand correctly, it should automatically use those test source folder to compile test classes.
* When specifying the test source and output folder manually in the build.properties, Tycho complain about missing JUnit dependencies when compiling them.

I guess that both the specification of test source and output folder in the build.properites as well as the additional specification of JUnit dependencies relate to the same root cause, due to which the information from the `.classpath` file are not considered correctly. I will try to figure out what is special in the SWT setup compared to the Tycho integration tests for that behavior.

binaries/org.eclipse.swt.win32.win32.x86_64/build.properties
laeubi: it will jsut work without that, as it is configured in the `.classpath` already.

binaries/org.eclipse.swt.win32.win32.x86_64/build.properties
laeubi: it might be because classpath is a linked file, can you try to copy the file as a regular one into the project? I recently added support for linked files, but that for sure not respected everywhere at the moment.

HeikoKlare: Thank you, @laeubi! The linked files are actually the problem here. It affects both the linked `.classpath` file as well as the linked test source folder.
* When replacing the linked `.classpath` with the same file directly placed in the project, Tycho properly considers the test source folder specified in the `.classpath` file. Then it only complains that it does not find that (linked) folder:
`[WARNING] Source directory C:\dev\eclipse\platform-only\git\eclipse.platform.swt\binaries\org.eclipse.swt.win32.win32.x86_64\Eclipse SWT Tests\win32 does not exist`
* When also replacing the linked source test folder with the same folder directly placed in the project, test source files are compiled correctly and tests are executed.

laeubi: okay then this needs some adjustment in Tycho first, I'll try to take a look. As always if one want to 
provide an [integration-test](https://github.com/eclipse/tycho/wiki#providing-an-integration-test) to demonstrate the issue that would be helpful.

HeikoKlare: I have already created an according integration test (on https://github.com/HeikoKlare/tycho/tree/junit5-with-linked-files) and will provide a describing issue and a PR with that test to the Tycho project tomorrow.

binaries/org.eclipse.swt.win32.win32.x86_64/build.properties
HeikoKlare: That's correct. Only the .classpath configuration is currently not considered because the .classpath file is not considered due to being a linked resource.

binaries/org.eclipse.swt.win32.win32.x86_64/build.properties
HeikoKlare: Tycho issue is now documented along with a reproducer PR: https://github.com/eclipse-tycho/tycho/issues/3803

HeikoKlare:
I have restructured the custom configuration for running the tests via Tycho as a temporary solution until https://github.com/eclipse-tycho/tycho/issues/3803 has been addressed. The solution has the following main properties to make it easy to replace with an in-build Tycho solution once available:
* Makes the Maven build behave exactly as if the Tycho issue was resolved (i.e., as if `.classpath` and the test source folder were no linked resources)
* Only affects Maven configuration (no PDE configuration, such as with `source` and `output` specification or `additional.bundles` in the `build.properties`)

I consists of the following additions:
* An active-by-default profile to the `pom.xml` of the binaries folder, which adds the necessary JUnit dependencies. I preferred the profile over directly adding the configuration to a `pom.xml` in the win32 fragments, as they have auto-generated POMs which I do not want to replace with fixed POMs just for a temporary solution.
* Configures the `testSourceDirectory` by making it configurable via a Maven property that is set via the win32 fragments' `build.properties`.

I have added comments to all additions that are part of the temporary solution to easily remove them later on.

@laeubi What do you think about this temporary solution? Since the colleagues working on Windows HiDPI support have a rather urgent demand for being able to provide this kind of tests, I would like to integrate this (or some other kind of temporary solution) to not depend on when the Tycho feature can/will be realized.

For whomever is wondering that it might make sense to directly start with JUnit 5 instead of JUnit 4: I want to keep this initial restructuring PR as simple as possible (and the moved tests are JUnit 4) and will provide a follow-up that migrates to JUnit 5.

laeubi:
@HeikoKlare if it works I won't object in general, but maybe it also works for your college to simply not branch from master but from your brach and add items on top of it?

Then it should work to already develop / test things and later on only needs a rebase on master if we (hopefully soon) have the final solution.

HeikoKlare:
> if it works I won't object in general, but maybe it also works for your college to simply not branch from master but from your brach and add items on top of it?

In general, I would prefer that over having an intermediate solution in master, However, the tests come together with several functional additions. These functional additions depend on each other and we want / need to merge them into master in small increments to ensure continuous validation and minimize the risk of regressions. Merging all of them in a separate branch would lead to a large merge into master afterwards with a too high risk of producing regressions.

bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/internal/PlatformSpecificExecution.java
```diff
@@ -0,0 +1,28 @@
...
+	private PlatformSpecificExecution()  {
+	}
+
+	public static void assumeIsFittingPlatform() {
+		assumeTrue("test is specific for Windows", isFittingOS());
+		assumeTrue("architecture of platform does not match", isFittingArchitecture());
+	}
+
+	private static boolean isFittingOS() {
+		return System.getProperty("os.name").toLowerCase().startsWith("win");
```
laeubi, line 17: I somehow remmber we already have a way in SWT to determine the platform.

merks: org.eclipse.swt.tests.junit.SwtTestUtil.isWindows

HeikoKlare: Yes, there are at least two ways to determine the platform, but unfortunately none of them is applicable here:
1. The platform for which the currently accessed fragment as stored in `Platform.PLATFORM`. At usual runtime, you have the fragment matching your OS loaded (as defined by the platform filter of the fragment), so you can use this to determine the platform you are running. When executing these tests, however, you will always access the `Platform` class from the fragment that you are currently processing, so `Platform.PLATFORM` will always be "win32" in this case, even if you are running on Linux or MacOS.
2. The `SwtTestUtil` (as referred to by Ed). But they are placed in the downstream test plug-in and not in the SWT plug-in itself.

But since I have moved this util class to the same package as the `Library` class to retrieve the system architecture from the `Library` class, I can now also retrieve the OS from that class (which does the same as the `SwtTestUtil`). I have adapted this method accordingly in 91a97ca.

HeikoKlare:
Thanks to @laeubi's fast improvement for Tycho, the tests inside the SWT fragments now run properly without custom Maven configuration. I was able to update / simplify this PR accordingly in b256ce7. So we can merge this without adding any temporary solution.

The only thing remaining is that the test source folder linked in the project description cannot use the `SWT_HOST_PLUGIN` variable, as Tycho does not support custom variables yet: https://github.com/eclipse-tycho/tycho/issues/3820
I consider this acceptable and something that can be improved if/once Tycho supports variables.

HannesWell:
> There is currently no pre-configured way to implement unit tests for classes that are not directly accessible via public API.

Just as a quick remark without knowing if this has already been considered: The 'PDE-way' to achieve this is to make the test a fragment of the tested plugin and place the test in the same package. Or is this not possible due to the special structure of the SWT bundles/fragments?

laeubi:
> Or is this not possible due to the special structure of the SWT bundles/fragments?

A fragment can't have another fragment as a host.

HeikoKlare:
> But any way, I just wanted to make sure all options were on the table. In general it is not bad to have the tests in the Plug-in project (like it is common in Maven projects),

We have discussed different options on how to realize such tests in a separate discussion thread (https://github.com/eclipse-platform/.github/discussions/203). Seems like I forgot to reference the discussion in this PR.

> I think it would just be good to have all (windows-)tests at one place.

The Windows tests (as well as potentially other OS tests or even common tests) can/will be at two places:
1. The (common or OS-specific) test plug-in, testing the SWT plug-in's API (already existing now)
2. "Internal" tests for abstraction layers within the plug-in/fragment that require testing but are not part of public API (like the ones in this PR).

HannesWell:
> > But any way, I just wanted to make sure all options were on the table. In general it is not bad to have the tests in the Plug-in project (like it is common in Maven projects),
> 
> We have discussed different options on how to realize such tests in a separate discussion thread ([eclipse-platform/.github#203](https://github.com/eclipse-platform/.github/discussions/203)). Seems like I forgot to reference the discussion in this PR.

Sorry I have missed that (I'm not subscribed for all events in the platform and platform.ui repo).
Thank you for your elaboration.

laeubi:
> Of course, but it could just use org.eclipse.swt just like it requires it now.

While this technically works, I'm not sure the tooling (IDE + Build) is really prepared for that, lets assume we Have `Host`, `Fragement` and `Test`, where `Fragment` and `Test` are bot fragments to the `Host`, to allow Test to access Fragment classes we must make sure they are always used together and there is some way for Test to require `Fragment` + `Host` so the compiler can see the classes.

HeikoKlare:
> > Of course, but it could just use org.eclipse.swt just like it requires it now.
> 
> While this technically works, I'm not sure the tooling (IDE + Build) is really prepared for that, lets assume we Have `Host`, `Fragement` and `Test`, where `Fragment` and `Test` are bot fragments to the `Host`, to allow Test to access Fragment classes we must make sure they are always used together and there is some way for Test to require `Fragment` + `Host` so the compiler can see the classes.

I tried this as an alternative to this PR with the initial, temporary solution (when Tycho support was linked .classpath files was missing) but did not find a proper way to enable the test fragment to access classes of the Windows fragment. I did not invest too much time but it did not feel as good as the solution of having the unit tests directly within the fragments anyway.

PR #1204, opened 2024-04-30
Extend GitHub actions configuration matrix for MacOS AArch64 build

HeikoKlare:
GitHub provides aarch64 runners for MacOS and has recently made them even the default runners for MacOS. In order to also validate on aarch64 in addition to x86_64, this change extends the build configuration matrix to also be executed on a MacOS aarch64 runner.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -31,7 +31,8 @@ jobs:
         config: 
           - { name: Linux, os: ubuntu-latest, native: gtk.linux.x86_64 }
           - { name: Windows, os: windows-latest, native: win32.win32.x86_64 }
-          - { name: MacOS, os: macos-latest-large, native: cocoa.macosx.x86_64 }
+          - { name: MacOS x86, os: macos-latest-large, native: cocoa.macosx.x86_64 }
+          - { name: MacOS ARM, os: macos-latest, native: cocoa.macosx.aarch64 }
     name: Verify ${{ matrix.config.name }} with Java-${{ matrix.java }}
     steps:
     - name: checkout swt
--- a/pom.xml
+++ b/pom.xml
@@ -100,12 +100,7 @@
                 <environment>
                   <os>macosx</os>
                   <ws>cocoa</ws>
-                  <arch>x86_64</arch>
-                </environment>
-                <environment>
-                  <os>macosx</os>
-                  <ws>cocoa</ws>
-                  <arch>aarch64</arch>
+                  <arch>${os.arch}</arch>
                 </environment>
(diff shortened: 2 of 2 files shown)


laeubi:
@HeikoKlare I got the expression that we need two profiles [here](https://github.com/eclipse-platform/eclipse.platform.swt/blob/aab3bfaeaa2d0823aecdff210d1b8f3c542bc435/pom.xml#L83) one for x86 and one for aarch mac  that only contain the matching environment, but that's only a guess at the moment.

HeikoKlare:
> I got the expression that we need two profiles [here](https://github.com/eclipse-platform/eclipse.platform.swt/blob/aab3bfaeaa2d0823aecdff210d1b8f3c542bc435/pom.xml#L83) one for x86 and one for aarch mac that only contain the matching environment, but that's only a guess at the moment.

It looks like your guess is completely correct: when I remove the x86 environment from the MacOS profile on my Mac (aarch64), the correct binaries are used and properly loaded.

HeikoKlare:
I have now simply parametrized the OS value used for the target platform environment with the OS value given by the Maven build property. Since they follow the same pattern, this is far less verbose than adding multiple profiles for different architectures.

HannesWell:
> I have now simply parametrized the OS value used for the target platform environment with the OS value given by the Maven build property. Since they follow the same pattern, this is far less verbose than adding multiple profiles for different architectures.

Do you have a reference for the values of os.arch in Maven or does Maven just use the `os.arch` value provided by the JVM?
If that's the case I'm not sure if the values really always follow the same pattern?
At least under Windows I see in my Eclipse Configuration
```
*** System properties:
os.arch=amd64
```
Maybe the values are fine on Mac only?

HeikoKlare:
> Do you have a reference for the values of os.arch in Maven or does Maven just use the `os.arch` value provided by the JVM?

Yes, Maven takes the values from Java system properties: https://maven.apache.org/guides/introduction/introduction-to-profiles.html#os

> If that's the case I'm not sure if the values really always follow the same pattern?

I don't think they do, so you are right that this will probably not work for every case / every OS. I also found that on Windows this "simplification" I did for Mac does not work because of the architecture "amd64" not matching. However, I expect this to already be the only exception, as that's the only architecture for which SWT has a special case in it's library handling as well:

https://github.com/eclipse-platform/eclipse.platform.swt/blob/b58f379a6b35e7eb1a2650af9344b085d2898285/bundles/org.eclipse.swt/Eclipse%20SWT%20PI/common/org/eclipse/swt/internal/Library.java#L69-L73

Issue #127, opened 2022-05-11
Dynamically adapt SWT on 'SWT.ZoomChanged' event on single monitor.


niraj-modi:
This bug is to adapt SWT/Eclipse dynamic  to changing OS zoom values:
It will require adapting all the SWT widgets to work on the new zoom value.

Note: This is a big ticket item and it will require lot of changes in SWT, here the branch where i could partially achieve this:
https://git.eclipse.org/c/platform/eclipse.platform.swt.git/?h=nmodi%2FDynamicDPI_4909

Since this was partially working then and hence we opted to restart Eclipse on 'SWT.ZoomChanged' event:
https://www.eclipse.org/eclipse/news/4.9/platform.php#zoom-changed

HeikoKlare:
This issue has been resolved with the "monitor-specific UI scaling" feature and the according improvements done for it. With that feature, shells adapt their zoom to the current monitor:
- https://eclipse.dev/eclipse/news/news.html?file=4.34/platform.html#rescale-on-runtime-preference

Remaining issues with that feature are tracked in this umbrella:
- https://github.com/eclipse-platform/eclipse.platform.swt/issues/1961

Issue #62, opened 2022-04-21
Display class should not have the zoom


MarcelBoerner:
When using the swts Display class it has different methods which take care about the zoom. The Display class should not have the zooms, only the devices should have the zoom. When you call the getMonitor method of the display class, the zoom to the devices is set to the ones of the primary monitor on windows versions above 8.1:
![image](https://user-images.githubusercontent.com/104011194/164384351-f4418098-994b-4706-8962-df31ed98a02a.png)
As you see this can not happen since the else case can not be reached on windows 8.1. And even if the dpi is used you do not get the results which are helping. When you have dpi unaware, the dpi is always 96 pixels, if you have dpi aware application mode than you get the one from the primary monitor and if you have the per monitor (v2) you get the correct dpi. But the last mode can not be used since the dpi awareness only works on the main window and its direct child. The other controls like tabs and so on are not taking part in the per monitor awareness.

sravanlakkimsetti:
@niraj-modi

MarcelBoerner:
As Addition i checked the fixes done in the past like: 
[Per Monitor HIDPI Fix for event notification](https://bugs.eclipse.org/bugs/show_bug.cgi?id=515502) 

I tried it with the per monitorv2 feature "on" in the javaw.exe.manifest:
` <dpiAwareness xmlns="http://schemas.microsoft.com/SMI/2016/WindowsSettings">PerMonitorV2</dpiAwareness>
`

The application is in permonitorv2 mode:
![image](https://user-images.githubusercontent.com/104011194/165030158-28c90793-799e-469b-9431-ec27fdb57ea1.png)

So I expected that a zoom changed event is fired when moving my application from one monitor to the other.
My Primary monitor is at 3840x2160 @150% scaling and the second 3840x2160 @175% scaling.
The application was started at the secondary monitor:

When moving the application from secondary monitor to primary monitor i get no event:
![image](https://user-images.githubusercontent.com/104011194/165031064-5b7666a5-16af-4cd8-b992-764387874de4.png)

When moving the application from primary monitor to secondary monitor, i get an event:
![image](https://user-images.githubusercontent.com/104011194/165031141-33969c09-2783-41ba-8af7-bd40854a34de.png)

[...]

PR #1209, opened 2024-05-03
Adapt GetSystemMetrics call to be DPI dependent

ShahzaibIbrahim:
**NOTE: these changes affect only _Windows_**

## What it does

Writing GetSystemMetrics method in Widget class so that the child controls can use it to get System Metrics to be DPI dependent. There are some visible affected areas (before and after zooming to 200 from 100) for which screenshots are attached. 

The OS.getSystemMetrics call had a slight issue in some of the controls i.e. Combo, List and Table. The values returned for scroll i.e. `OS.SM_CXVSCROLL` were independent of DPI. So moving the shell from smaller zoom level to higher zoom level we can see the visible difference in the scroll bar which is not visible in the latter part. 

If a scrollbar is shown or not could perhaps be a side effect in some scenarios, but the main reason for that change is that the correct value for scroolbar sizes are used when the size for the table is calculated. You'll see the effect in the following screenshot:

image

Both are started on 100% and moved to 200% (That is when the screenshots were made). The left one uses OS.GetSystemMetrics, the right one uses OS.GetSystemMetricsForDpi. You'll see the left table is smaller, because it still uses the scrollbar sizes for the primary monitor (=OS.GetSystemMetrics)

## HOW TO TEST **(Windows only!)**

- Run the ControlExample 
- Add these lines in your vm arguments  
`-Dswt.autoScale=quarter`
`-Dswt.autoScale.updateOnRuntime=true`
- Open the SWT Controls view
- Switch to the tab "Combo"/"List"/"Table" 
- Change the size to 50x50 from Size section
- See if gives more spacing to widget itself rather than scrollbar than before 

## EXPECTED BEHAVIOUR

None of the widgets should look out of order when changing the zoom level.

Contributes to #62 and #127:

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Button.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Button.java
@@ -54,24 +54,13 @@ public class Button extends Control {
 	ImageList imageList;
 	boolean ignoreMouse, grayed, useDarkModeExplorerTheme;
 	static final int MARGIN = 4;
-	static final int CHECK_WIDTH, CHECK_HEIGHT;
+	private final int checkWidth, checkHeight;
 	static final int ICON_WIDTH = 128, ICON_HEIGHT = 128;
 	static /*final*/ boolean COMMAND_LINK = false;
 	static final char[] STRING_WITH_ZERO_CHAR = new char[] {'0'};
 	static final long ButtonProc;
 	static final TCHAR ButtonClass = new TCHAR (0, "BUTTON", true);
 	static {
-		long hBitmap = OS.LoadBitmap (0, OS.OBM_CHECKBOXES);
-		if (hBitmap == 0) {
-			CHECK_WIDTH = OS.GetSystemMetrics (OS.SM_CXVSCROLL);
-			CHECK_HEIGHT = OS.GetSystemMetrics (OS.SM_CYVSCROLL);
-		} else {
-			BITMAP bitmap = new BITMAP ();
-			OS.GetObject (hBitmap, BITMAP.sizeof, bitmap);
-			OS.DeleteObject (hBitmap);
-			CHECK_WIDTH = bitmap.bmWidth / 4;
-			CHECK_HEIGHT =  bitmap.bmHeight / 3;
-		}
 		WNDCLASS lpWndClass = new WNDCLASS ();
 		OS.GetClassInfo (0, ButtonClass, lpWndClass);
 		ButtonProc = lpWndClass.lpfnWndProc;
@@ -119,6 +108,17 @@ public class Button extends Control {
  */
 public Button (Composite parent, int style) {
 	super (parent, checkStyle (style));
+	long hBitmap = OS.LoadBitmap (0, OS.OBM_CHECKBOXES);
+	if (hBitmap == 0) {
+		checkWidth = getSystemMetrics(OS.SM_CXVSCROLL);
+		checkHeight = getSystemMetrics (OS.SM_CYVSCROLL);
+	} else {
+		BITMAP bitmap = new BITMAP ();
+		OS.GetObject (hBitmap, BITMAP.sizeof, bitmap);
+		OS.DeleteObject (hBitmap);
+		checkWidth = bitmap.bmWidth / 4;
+		checkHeight =  bitmap.bmHeight / 3;
+	}
 }
 
 void _setImage (Image image) {
@@ -315,11 +315,11 @@ int computeLeftMargin () {
 	int width = 0, height = 0, border = getBorderWidthInPixels ();
 	if ((style & SWT.ARROW) != 0) {
 		if ((style & (SWT.UP | SWT.DOWN)) != 0) {
-			width += OS.GetSystemMetrics (OS.SM_CXVSCROLL);
-			height += OS.GetSystemMetrics (OS.SM_CYVSCROLL);
+			width += getSystemMetrics (OS.SM_CXVSCROLL);
+			height += getSystemMetrics (OS.SM_CYVSCROLL);
 		} else {
-			width += OS.GetSystemMetrics (OS.SM_CXHSCROLL);
-			height += OS.GetSystemMetrics (OS.SM_CYHSCROLL);
+			width += getSystemMetrics (OS.SM_CXHSCROLL);
+			height += getSystemMetrics (OS.SM_CYHSCROLL);
 		}
 	} else {
 		if ((style & SWT.COMMAND) != 0) {
@@ -373,7 +373,7 @@ int computeLeftMargin () {
 						flags = OS.DT_CALCRECT | OS.DT_WORDBREAK;
 						rect.right = wHint - width - 2 * border;
 						if (isRadioOrCheck()) {
-							rect.right -= CHECK_WIDTH + 3;
+							rect.right -= checkWidth + 3;
 						} else {
 							rect.right -= 6;
 						}
@@ -392,8 +392,8 @@ int computeLeftMargin () {
 				OS.ReleaseDC (handle, hDC);
 			}
 			if (isRadioOrCheck()) {
-				width += CHECK_WIDTH + extra;
-				height = Math.max (height, CHECK_HEIGHT + 3);
+				width += checkWidth + extra;
+				height = Math.max (height, checkHeight + 3);
 			}
 			if ((style & (SWT.PUSH | SWT.TOGGLE)) != 0) {
 				width += 12;  height += 10;
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Shell.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Shell.java
@@ -1063,12 +1063,12 @@ Point getMaximumSizeInPixels () {
 	int width = Math.min (Integer.MAX_VALUE, maxWidth);
 	int trim = SWT.TITLE | SWT.CLOSE | SWT.MIN | SWT.MAX;
 	if ((style & SWT.NO_TRIM) == 0 && (style & trim) != 0) {
-		width = Math.min (width, OS.GetSystemMetrics (OS.SM_CXMAXTRACK));
+		width = Math.min (width, getSystemMetrics (OS.SM_CXMAXTRACK));
 	}
 	int height = Math.min (Integer.MAX_VALUE, maxHeight);
 	if ((style & SWT.NO_TRIM) == 0 && (style & trim) != 0) {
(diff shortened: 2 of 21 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Table.java
```diff
@@ -1479,10 +1479,10 @@ public void clearAll () {
 	int border = getBorderWidthInPixels ();
 	width += border * 2;  height += border * 2;
 	if ((style & SWT.V_SCROLL) != 0) {
-		width += OS.GetSystemMetrics (OS.SM_CXVSCROLL);
+		height += getSystemMetrics (OS.SM_CYHSCROLL);
```
fedejeanne, line 1482: Shouldn't this still be `width += ...` ?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Text.java
```diff
@@ -976,7 +976,7 @@ void fixAlignment () {
 	* style bits instead.
 	*/
 //	if ((style & SWT.BORDER) != 0 && (style & SWT.FLAT) != 0) {
-//		return OS.GetSystemMetrics (OS.SM_CXBORDER);
```
fedejeanne, line 979: Since when has this code been there and commented out?
Is it safe to remove or does someone count on it? (check the git logs for clues)

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Widget.java
```diff
@@ -2650,4 +2650,14 @@ void setZoom(int zoom) {
 private static void handleDPIChange(Widget widget, int newZoom, float scalingFactor) {
 	widget.setZoom(newZoom);
 }
+
+int getSystemMetrics(int nIndex) {
+	if (OS.WIN32_BUILD >= OS.WIN32_BUILD_WIN10_1607) {
+		return OS.GetSystemMetricsForDpi(nIndex, DPIUtil.mapZoomToDPI(getZoom()));
+	}  else {
+		return OS.GetSystemMetrics(nIndex);
+	}
```
fedejeanne, line 2659: Move outside `else`

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Widget.java
```diff
@@ -2650,4 +2650,14 @@ void setZoom(int zoom) {
 private static void handleDPIChange(Widget widget, int newZoom, float scalingFactor) {
 	widget.setZoom(newZoom);
 }
+
+int getSystemMetrics(int nIndex) {
+	if (OS.WIN32_BUILD >= OS.WIN32_BUILD_WIN10_1607) {
```
fedejeanne, line 2655: A comment here would help understanding this `if`

fedejeanne (changes requested):
I see that this PR changes (all?) calls to `OS.GetSystemMetrics` for a call to a new intermediate method in `Widget`: why not simply change `OS.GetSystemMetrics` and put the new logic in there?

The request for changes is only because of the change in `Table`.

fedejeanne:
@ShahzaibIbrahim please also improve the commit message:
* Remove the links from the first line (they are in the body anyway)
* Add the text "Fixes" or "Contributes to" appropriately for the linked issues

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Text.java
ShahzaibIbrahim: The logs says it has been last change in 2009. Should I remove it or keep it as is?

fedejeanne: Remove it

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Table.java
```diff
@@ -1479,10 +1479,10 @@ public void clearAll () {
 	int border = getBorderWidthInPixels ();
 	width += border * 2;  height += border * 2;
 	if ((style & SWT.V_SCROLL) != 0) {
-		width += OS.GetSystemMetrics (OS.SM_CXVSCROLL);
+		width += getSystemMetrics (OS.SM_CYHSCROLL);
```
fedejeanne, line 1482: I think the parameter is still wrong
```suggestion
		width += getSystemMetrics (OS.SM_CXVSCROLL);
```

fedejeanne (changes requested):
One call is still wrong (look at the parameter)

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Table.java
ShahzaibIbrahim: Sorry for that.

fedejeanne (changes requested):
Thank you for fixing my findings. I took a deeper look and found some other stuff I would like your input on:
1. I noticed there are still some calls inside the `widgets` and the `win32` package that are **not** using the new method, namely these 3 on top
image
   The first one (in the `OS` class) is probably OK because it asks for the value of `SM_IMMENABLED` which I assume is the same regardless of the current DPI settings (please confirm/rebut) 
2. What would happen if one removes the `if (OS.WIN32_BUILD >= OS.WIN32_BUILD_WIN10_1607)` and always calls `OS.GetSystemMetricsForDpi(...)`? Would the application crash in older Windows versions?

NOTE: The request for changes is only because point nr 1.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Button.java
```diff
@@ -54,24 +54,13 @@ public class Button extends Control {
 	ImageList imageList;
 	boolean ignoreMouse, grayed, useDarkModeExplorerTheme;
 	static final int MARGIN = 4;
-	static final int CHECK_WIDTH, CHECK_HEIGHT;
+	final int CHECK_WIDTH, CHECK_HEIGHT;
```
fedejeanne, line 57: Make them `private` and rename them to `checkWidth` and `checkHeight`

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Button.java
```diff
@@ -119,6 +108,17 @@ public class Button extends Control {
  */
 public Button (Composite parent, int style) {
 	super (parent, checkStyle (style));
+	long hBitmap = OS.LoadBitmap (0, OS.OBM_CHECKBOXES);
+	if (hBitmap == 0) {
+		CHECK_WIDTH = OS.GetSystemMetrics(OS.SM_CXVSCROLL);
+		CHECK_HEIGHT = OS.GetSystemMetrics (OS.SM_CYVSCROLL);
```
fedejeanne, line 114: Shouldn't these call the new method: `getSystemMetrics(...)` ?

fedejeanne (dismissed):

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Shell.java
```diff
@@ -125,7 +125,7 @@ public class Shell extends Decorations {
 	ToolTip [] toolTips;
 	long hwndMDIClient, lpstrTip, toolTipHandle, balloonTipHandle, menuItemToolTipHandle;
 	int minWidth = SWT.DEFAULT, minHeight = SWT.DEFAULT, maxWidth = SWT.DEFAULT, maxHeight = SWT.DEFAULT;
-	private int nativeZoom;
+	public int nativeZoom;
```
fedejeanne, line 128: Why public?

fedejeanne (commented):
Thank you for addressing my change requests, @ShahzaibIbrahim . Would it be possible to rebase in `origin/master` so I can checkout your PR and take a deeper look at it?

fedejeanne (commented):
Changes look better now, I see that all the widgets go through the new method now ✔️ 

image

Since there has been some changes since the beginning of this PR, I would like to validate the behavior of the code myself. Some brief "How to test" and also the "expected (observable?) behavior" sections in the description of this PR would help.

fedejeanne:
@ShahzaibIbrahim thank you for adding a _How to test_. I tested in **Windows 10** and I see no difference before and after this PR. 

## Before (Table, size 100x100)

Zoom = 100
image

Zoom = 200
image

## After

Zoom = 100
image

Zoom = 200
image

Maybe I'm testing it wrong?

ShahzaibIbrahim:
> Maybe I'm testing it wrong?

No you have tested it correctly. I just forgot to mention that you need to add these vm arguments while running the example.

-Dswt.autoScale=quarter

fedejeanne:
I tested it again and I still see no difference between before and after this PR. Moreover, the status **before** this PR looks OK (I see both scroll bars when moving to the monitor with _zoom = 200_. Can you please confirm?

## Before (Table size 100x100)

Zoom = 100
image

Zoom = 200
According to the initial description of the bug, this image should show a missing horizontal scrollbar, right?

image


## After (Table size 100x100)

Zoom = 100
image

Zoom = 200
image

ShahzaibIbrahim:
> I tested it again and I still see no difference between before and after this PR. Moreover, the status **before** this PR looks OK (I see both scroll bars when moving to the monitor with _zoom = 200_. Can you please confirm?

Please check the scroll with 50x50 size and you will see the difference visibly. I have also added this in How to test section.

akoch-yatta:
> I tested it again and I still see no difference between before and after this PR. Moreover, the status **before** this PR looks OK (I see both scroll bars when moving to the monitor with _zoom = 200_. Can you please confirm?

If a scrollbar is shown or not could perhaps be a side effect in some scenarios, but the main reason for that change is that the correct value for scroolbar sizes are used when the size for the table is calculated. You'll see the effect in the following screenshot:

image

Both are started on 100% and moved to 200% (That is when the screenshots were made). The left one uses `OS.GetSystemMetrics`, the right one uses `OS.GetSystemMetricsForDpi`. You'll see the left table is smaller, because it still uses the scrollbar sizes for the primary monitor (=`OS.GetSystemMetrics`)

iloveeclipse:
This change caused severe regression in the compare editor, see https://github.com/eclipse-platform/eclipse.platform/issues/1489.

PR #1214, opened 2024-05-06
Implementation of GC to enable multi zoom level support for win32

amartya4256:
## Addressed issues

## Requires
* [x] https://github.com/eclipse-platform/eclipse.platform.swt/pull/1258

## Unlocks
* https://github.com/eclipse-platform/eclipse.platform.swt/pull/1236
* https://github.com/eclipse-platform/eclipse.platform.swt/pull/1247
* https://github.com/eclipse-platform/eclipse.platform.swt/pull/1243

## Description

This pull request is a base pull request for several following PRs which implement multi zoom level support for different components (resources, widgets, etc). 

In this PR, the focus is on GC to enable it to use the zoom level information which is added to widget. Also, it provides zoom level information in GCData which can later be used by resources like Image, Region, etc which don't have the zoom information since they are not a widget. Also, GC now uses the DPIUtil methods with zoom for getting the size as per the zoom level. This zoom in GC is derived from the GCData where we add the zoom information.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
@@ -455,8 +455,9 @@ void checkGC(int mask) {
  * </ul>
  */
 public void copyArea (Image image, int x, int y) {
-	x = DPIUtil.autoScaleUp(drawable, x);
-	y = DPIUtil.autoScaleUp(drawable, y);
+	int deviceZoom = getZoom();
+	x = DPIUtil.autoScaleUp(drawable, x, deviceZoom);
+	y = DPIUtil.autoScaleUp(drawable, y, deviceZoom);
 	copyAreaInPixels(image, x, y);
 }
 
@@ -511,12 +512,13 @@ public void copyArea (int srcX, int srcY, int width, int height, int destX, int 
  * @since 3.1
  */
 public void copyArea (int srcX, int srcY, int width, int height, int destX, int destY, boolean paint) {
-	srcX = DPIUtil.autoScaleUp(drawable, srcX);
-	srcY = DPIUtil.autoScaleUp(drawable, srcY);
-	width = DPIUtil.autoScaleUp(drawable, width);
-	height = DPIUtil.autoScaleUp(drawable, height);
-	destX = DPIUtil.autoScaleUp(drawable, destX);
-	destY = DPIUtil.autoScaleUp(drawable, destY);
+	int deviceZoom = getZoom();
+	srcX = DPIUtil.autoScaleUp(drawable, srcX, deviceZoom);
+	srcY = DPIUtil.autoScaleUp(drawable, srcY, deviceZoom);
+	width = DPIUtil.autoScaleUp(drawable, width, deviceZoom);
+	height = DPIUtil.autoScaleUp(drawable, height, deviceZoom);
+	destX = DPIUtil.autoScaleUp(drawable, destX, deviceZoom);
+	destY = DPIUtil.autoScaleUp(drawable, destY, deviceZoom);
 	copyAreaInPixels(srcX, srcY, width, height, destX, destY, paint);
 }
 
@@ -757,10 +759,11 @@ void disposeGdip() {
  * </ul>
  */
 public void drawArc (int x, int y, int width, int height, int startAngle, int arcAngle) {
-	x = DPIUtil.autoScaleUp(drawable, x);
-	y = DPIUtil.autoScaleUp(drawable, y);
-	width = DPIUtil.autoScaleUp(drawable, width);
-	height = DPIUtil.autoScaleUp(drawable, height);
+	int deviceZoom = getZoom();
+	x = DPIUtil.autoScaleUp(drawable, x, deviceZoom);
+	y = DPIUtil.autoScaleUp(drawable, y, deviceZoom);
+	width = DPIUtil.autoScaleUp(drawable, width, deviceZoom);
+	height = DPIUtil.autoScaleUp(drawable, height, deviceZoom);
 	drawArcInPixels(x, y, width, height, startAngle, arcAngle);
 }
 
@@ -840,10 +843,11 @@ void drawArcInPixels (int x, int y, int width, int height, int startAngle, int a
  * @see #drawRectangle(int, int, int, int)
  */
 public void drawFocus (int x, int y, int width, int height) {
-	x = DPIUtil.autoScaleUp (drawable, x);
-	y = DPIUtil.autoScaleUp (drawable, y);
-	width = DPIUtil.autoScaleUp (drawable, width);
-	height = DPIUtil.autoScaleUp (drawable, height);
+	int deviceZoom = getZoom();
+	x = DPIUtil.autoScaleUp (drawable, x, deviceZoom);
+	y = DPIUtil.autoScaleUp (drawable, y, deviceZoom);
+	width = DPIUtil.autoScaleUp (drawable, width, deviceZoom);
+	height = DPIUtil.autoScaleUp (drawable, height, deviceZoom);
 	drawFocusInPixels(x, y, width, height);
 }
 
@@ -918,8 +922,9 @@ void drawFocusInPixels (int x, int y, int width, int height) {
  * </ul>
  */
 public void drawImage (Image image, int x, int y) {
-	x = DPIUtil.autoScaleUp(drawable, x);
-	y = DPIUtil.autoScaleUp(drawable, y);
+	int deviceZoom = getZoom();
+	x = DPIUtil.autoScaleUp(drawable, x, deviceZoom);
+	y = DPIUtil.autoScaleUp(drawable, y, deviceZoom);
 	drawImageInPixels(image, x, y);
 }
 
@@ -971,9 +976,9 @@ public void drawImage (Image image, int srcX, int srcY, int srcWidth, int srcHei
 	if (image == null) SWT.error (SWT.ERROR_NULL_ARGUMENT);
 	if (image.isDisposed()) SWT.error(SWT.ERROR_INVALID_ARGUMENT);
 
-	Rectangle src = DPIUtil.autoScaleUp(drawable, new Rectangle(srcX, srcY, srcWidth, srcHeight));
-	Rectangle dest = DPIUtil.autoScaleUp(drawable, new Rectangle(destX, destY, destWidth, destHeight));
-	int deviceZoom = DPIUtil.getDeviceZoom();
+	int deviceZoom = getZoom();
+	Rectangle src = DPIUtil.autoScaleUp(drawable, new Rectangle(srcX, srcY, srcWidth, srcHeight), deviceZoom);
+	Rectangle dest = DPIUtil.autoScaleUp(drawable, new Rectangle(destX, destY, destWidth, destHeight), deviceZoom);
 	if (deviceZoom != 100) {
 		/*
 		 * This is a HACK! Due to rounding errors at fractional scale factors,
@@ -1640,10 +1645,11 @@ void drawBitmapColor(Image srcImage, int srcX, int srcY, int srcWidth, int srcHe
  * </ul>
  */
 public void drawLine (int x1, int y1, int x2, int y2) {
-	x1 = DPIUtil.autoScaleUp (drawable, x1);
-	x2 = DPIUtil.autoScaleUp (drawable, x2);
-	y1 = DPIUtil.autoScaleUp (drawable, y1);
-	y2 = DPIUtil.autoScaleUp (drawable, y2);
+	int deviceZoom = getZoom();
+	x1 = DPIUtil.autoScaleUp (drawable, x1, deviceZoom);
+	x2 = DPIUtil.autoScaleUp (drawable, x2, deviceZoom);
+	y1 = DPIUtil.autoScaleUp (drawable, y1, deviceZoom);
+	y2 = DPIUtil.autoScaleUp (drawable, y2, deviceZoom);
 	drawLineInPixels(x1, y1, x2, y2);
 }
 
@@ -1692,10 +1698,11 @@ void drawLineInPixels (int x1, int y1, int x2, int y2) {
  * </ul>
  */
 public void drawOval (int x, int y, int width, int height) {
-	x = DPIUtil.autoScaleUp (drawable, x);
-	y = DPIUtil.autoScaleUp (drawable, y);
-	width = DPIUtil.autoScaleUp (drawable, width);
-	height = DPIUtil.autoScaleUp (drawable, height);
+	int deviceZoom = getZoom();
+	x = DPIUtil.autoScaleUp (drawable, x, deviceZoom);
+	y = DPIUtil.autoScaleUp (drawable, y, deviceZoom);
+	width = DPIUtil.autoScaleUp (drawable, width, deviceZoom);
+	height = DPIUtil.autoScaleUp (drawable, height, deviceZoom);
 	drawOvalInPixels(x, y, width, height);
 }
 
@@ -1768,8 +1775,9 @@ public void drawPath (Path path) {
  * @since 3.0
  */
 public void drawPoint (int x, int y) {
-	x = DPIUtil.autoScaleUp (drawable, x);
-	y = DPIUtil.autoScaleUp (drawable, y);
+	int deviceZoom = getZoom();
+	x = DPIUtil.autoScaleUp (drawable, x, deviceZoom);
+	y = DPIUtil.autoScaleUp (drawable, y, deviceZoom);
 	drawPointInPixels(x, y);
 }
 
@@ -1802,7 +1810,7 @@ void drawPointInPixels (int x, int y) {
  */
 public void drawPolygon (int[] pointArray) {
 	if (pointArray == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
-	drawPolygonInPixels(DPIUtil.autoScaleUp(drawable, pointArray));
+	drawPolygonInPixels(DPIUtil.autoScaleUp(drawable, pointArray, getZoom()));
 }
 
 void drawPolygonInPixels(int[] pointArray) {
@@ -1850,7 +1858,7 @@ void drawPolygonInPixels(int[] pointArray) {
  * </ul>
  */
 public void drawPolyline (int[] pointArray) {
-	drawPolylineInPixels(DPIUtil.autoScaleUp(drawable, pointArray));
+	drawPolylineInPixels(DPIUtil.autoScaleUp(drawable, pointArray, getZoom()));
 }
 
 void drawPolylineInPixels(int[] pointArray) {
@@ -1903,10 +1911,11 @@ void drawPolylineInPixels(int[] pointArray) {
  * </ul>
  */
 public void drawRectangle (int x, int y, int width, int height) {
-	x = DPIUtil.autoScaleUp (drawable, x);
-	y = DPIUtil.autoScaleUp (drawable, y);
-	width = DPIUtil.autoScaleUp (drawable, width);
-	height = DPIUtil.autoScaleUp (drawable, height);
+	int deviceZoom = getZoom();
+	x = DPIUtil.autoScaleUp (drawable, x, deviceZoom);
+	y = DPIUtil.autoScaleUp (drawable, y, deviceZoom);
+	width = DPIUtil.autoScaleUp (drawable, width, deviceZoom);
+	height = DPIUtil.autoScaleUp (drawable, height, deviceZoom);
 	drawRectangleInPixels(x, y, width, height);
 }
 
@@ -1962,7 +1971,7 @@ void drawRectangleInPixels (int x, int y, int width, int height) {
  */
 public void drawRectangle (Rectangle rect) {
 	if (rect == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
-	rect = DPIUtil.autoScaleUp(drawable, rect);
+	rect = DPIUtil.autoScaleUp(drawable, rect, getZoom());
 	drawRectangleInPixels(rect.x, rect.y, rect.width, rect.height);
 }
 
@@ -1988,12 +1997,13 @@ public void drawRectangle (Rectangle rect) {
  * </ul>
  */
 public void drawRoundRectangle (int x, int y, int width, int height, int arcWidth, int arcHeight) {
-	x = DPIUtil.autoScaleUp (drawable, x);
-	y = DPIUtil.autoScaleUp (drawable, y);
-	width = DPIUtil.autoScaleUp (drawable, width);
-	height = DPIUtil.autoScaleUp (drawable, height);
-	arcWidth = DPIUtil.autoScaleUp (drawable, arcWidth);
-	arcHeight = DPIUtil.autoScaleUp (drawable, arcHeight);
+	int deviceZoom = getZoom();
+	x = DPIUtil.autoScaleUp (drawable, x, deviceZoom);
+	y = DPIUtil.autoScaleUp (drawable, y, deviceZoom);
+	width = DPIUtil.autoScaleUp (drawable, width, deviceZoom);
+	height = DPIUtil.autoScaleUp (drawable, height, deviceZoom);
+	arcWidth = DPIUtil.autoScaleUp (drawable, arcWidth, deviceZoom);
+	arcHeight = DPIUtil.autoScaleUp (drawable, arcHeight, deviceZoom);
 	drawRoundRectangleInPixels(x, y, width, height, arcWidth, arcHeight);
 }
 
@@ -2085,8 +2095,9 @@ void drawRoundRectangleGdip (long gdipGraphics, long pen, int x, int y, int widt
  * </ul>
  */
 public void drawString (String string, int x, int y) {
-	x = DPIUtil.autoScaleUp(drawable, x);
-	y = DPIUtil.autoScaleUp(drawable, y);
+	int deviceZoom = getZoom();
+	x = DPIUtil.autoScaleUp(drawable, x, deviceZoom);
+	y = DPIUtil.autoScaleUp(drawable, y, deviceZoom);
 	drawStringInPixels(string, x, y, false);
 }
 
@@ -2118,8 +2129,9 @@ public void drawString (String string, int x, int y) {
  * </ul>
  */
 public void drawString (String string, int x, int y, boolean isTransparent) {
-	x = DPIUtil.autoScaleUp(drawable, x);
-	y = DPIUtil.autoScaleUp(drawable, y);
+	int deviceZoom = getZoom();
+	x = DPIUtil.autoScaleUp(drawable, x, deviceZoom);
+	y = DPIUtil.autoScaleUp(drawable, y, deviceZoom);
 	drawStringInPixels(string, x, y, isTransparent);
 }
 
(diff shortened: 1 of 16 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Shell.java
```diff
@@ -125,7 +125,20 @@ public class Shell extends Decorations {
...
+	 * <b>IMPORTANT:</b> This field is <em>not</em> part of the SWT
+	 * public API. It is marked public only so that it can be shared
+	 * within the packages provided by SWT. It is not available on all
+	 * platforms and should never be accessed from application code.
+	 * </p>
+	 *
+	 * @noreference This field is not intended to be referenced by clients.
+	 * @since 3.126
+	 */
+	public int nativeZoom;
```
fedejeanne, line 141: Don't do this, this allows writing into the field.
If you really need to share across packages then elevate the visibility of `getNativeZoom()` and add the `@noreference` to it so it's clear that it is only intended for internal usage

```suggestion
	private int nativeZoom;
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Widget.java
```diff
@@ -52,7 +52,20 @@
...
+	 * <b>IMPORTANT:</b> This field is <em>not</em> part of the SWT
+	 * public API. It is marked public only so that it can be shared
+	 * within the packages provided by SWT. It is not available on all
+	 * platforms and should never be accessed from application code.
+	 * </p>
+	 *
+	 * @noreference This field is not intended to be referenced by clients.
+	 * @since 3.126
+	 */
+	public int zoom;
```
fedejeanne, line 68: Dito, make it private and provide a (wider) getter
```suggestion
	private int zoom;
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -139,7 +139,7 @@ public final class Image extends Resource implements Drawable {
 	/**
 	 * Attribute to cache current device zoom level
 	 */
-	private int currentDeviceZoom = 100;
+	int currentDeviceZoom = 100;
```
fedejeanne, line 142: Why? Is this necessary for follow-up PRs? If it is, keep in mind that this allows writing into it so maybe you better provide a getter insted

```suggestion
	private int currentDeviceZoom = 100;
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GCData.java
```diff
@@ -47,6 +48,8 @@ public final class GCData {
 	public float[] lineDashes;
 	public float lineMiterLimit = 10;
 	public int alpha = 0xFF;
+	public int deviceZoom;
+	public int nativeDeviceZoom;
```
fedejeanne, line 52: Letting these 2 be public can be dangerous since they can be set separately, which contradicts what the new method `setNativeDeviceZoom` does.
Make them private and provide getters instead

fedejeanne (dismissed):
Hi @amartya4256 , thank you for your contribution.

I noticed some problems with the visibility of some fields (the most serious one is allowing to set the `deviceZoom` and `nativeDeviceZoom` separately.

Could you please look into it? 
Thank you.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Shell.java
amartya4256: Introducing a public getter/setter will change the API and it is a Win32 specific implementation so we should not do that.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Widget.java
amartya4256: Introducing a public getter/setter will change the API and it is a Win32 specific implementation so we should not do that.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
amartya4256: Introducing a public getter/setter will change the API and it is a Win32 specific implementation so we should not do that.

laeubi:
I just wanted to note that (platform dependent) _package private_ or even _public_ fields / methods are quite common, they should just be documented in this case as platform dependent.

fedejeanne:
> I just wanted to note that (platform dependent) _package private_ or even _public_ fields / methods are quite common, they should just be documented in this case as platform dependent.

@laeubi Is it really necessary to specify that a _package private_ field/method inside a fragment is platform-dependent? I mean the fragment is already platform-dependent and the field/method can't be seen outside the fragment. Or am I missing something?

laeubi:
@fedejeanne I have only seen it on public field, so maybe in my sentence it should read as an XOR so:

> Either the field is _package private_ **or instead** public but clearly documented to be platform dependent.

:-)

tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/graphics/GCWin32Tests.java
```diff
@@ -0,0 +1,76 @@
...
+	public void tearDown() {
+		DPIUtil.setDeviceZoom(initialZoom);
+	}
+
+	@Test
+	public void gcZoomLevelMustChangeOnShellZoomChange() {
+		int zoom = DPIUtil.getDeviceZoom();
+		Shell shell = new Shell(display);
+		shell.addListener(SWT.Paint, event -> {
+			assertEquals("GCData must have a zoom level equal to the actual zoom level of the widget/shell", shell.zoom, event.gc.getGCData().deviceZoom);
```
fedejeanne, line 48: Don't use assertions inside listeners, they will only be printed to the output. If you change this to `assertTrue(false)` then this test still passes.

tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/graphics/GCWin32Tests.java
```diff
@@ -0,0 +1,76 @@
...
+		});
+		shell.addListener(SWT.ZoomChanged, event -> {
+			assertEquals("GCData must have a zoom level equal to the actual zoom level of the widget/shell on zoomChanged event", shell.zoom, event.gc.getGCData().deviceZoom);
```
fedejeanne, line 51: Dito

HeikoKlare:
> The tests are broken

Note that new test classes in a test plug-in usually need to be added to the plug-ins test suite. In the win32 SWT tests, this is the `AllWin32Tests` test suite. If you do so for the new `GCWin32Tests`, the PR build will fail on Windows because of the test failures mentioned by @fedejeanne.

If you cannot provide reasonable, sufficient tests via public API only, you need to add the test cases in the win32 fragments, as prepared for the `DefaultSWTFontRegistry` and `ScalingSWTFontRegistry` in #1203.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GCData.java
```diff
@@ -57,4 +60,13 @@ public final class GCData {
...
 	public boolean focusDrawn;
+
+	int getNativeDeviceZoom() {
+		return nativeDeviceZoom;
+	}
+
+	public void setNativeDeviceZoom(int nativeDeviceZoom) {
+		this.nativeDeviceZoom = nativeDeviceZoom;
+		this.deviceZoom = DPIUtil.getZoomForAutoscaleProperty(nativeDeviceZoom);
+	}
```
HeikoKlare, line 71: As discussed, adding a public method to a platform-dependent, public class is only allowed when added to all three implementations. Or do I miss anything why that's not necessary here, @akoch-yatta @amartya4256?

amartya4256: The whole class of GCData is only applicable to windows. So I think it should be okay to have a public method inside. I and @fedejeanne had a talk about this change before. Let me know what you think.

amartya4256: Here it reads as follows:
`/**
 * Instances of this class are descriptions of GCs in terms
 * of unallocated platform-specific data fields.
 * <p>
 * <b>IMPORTANT:</b> This class is <em>not</em> part of the public
 * API for SWT. It is marked public only so that it can be shared
 * within the packages provided by SWT. It is not available on all
 * platforms, and should never be called from application code.
 * </p>
 * @noreference This class is not intended to be referenced by clients
 */
public final class GCData {`

akoch: I think @HeikoKlare is correct about this remark. Although the GCData is heavily different on all OS, the difference is only on attribute level. We discussed about adding the method to simplify the usage (by setting two attributes via one call), but we mustn't do that as a publich method.

iloveeclipse: If I remember right, API checks are not enabled for SWT in Jenkins, but in the IDE. @HannesWell : please correct me if I'm wrong. So the correct way to review SWT PR that might change API is to import PR in the IDE and clean build SWT there.

HeikoKlare: If I am not mistaken, API checks in CI / Jenkins have been recently enabled: https://github.com/eclipse-platform/eclipse.platform.swt/pull/1011 
But these checks only validate versions and not the "compatibility" between the different OS implementations, do they? So if the minor of SWT has already been bumped in the current development cycle and if you then add public API only for a single OS implementation, I would not expect any automated check to complain. This error will only be caught "automatically" when a consumer of that public API method is added in code that is compiled for all operating systems. Then this will result in a compiler error because of the according method not existing in the class for that specific OS.

fedejeanne: So if I read this correctly, the only way to see if we're breaking the API these days is to:
1. Use the newly added code (or all the public API for that matter) in client code, _e.g._ in tests inside platform-independent plugins like `org.eclipse.swt.tests`
2. Compile for all platforms --> Jenkins

(+1 for TDD and full test coverage )

And also: the I-Builds would **not** break if one forgets to add the methods in some platforms.

right, @HeikoKlare ?

@iloveeclipse I didn't quite get what you meant by compiling it inside the IDE: does this help to catch inconsistencies in the public API across platforms or merely breaking API changes in the current platform?

iloveeclipse: > I didn't quite get what you meant by compiling it inside the IDE: does this help to catch inconsistencies in the public API across platforms or merely breaking API changes in the current platform?

The later, sorry for confusion. The consistency between platform API's is supposed to be reviewed manually by the reviewer.

> If I am not mistaken, API checks in CI / Jenkins have been recently enabled: https://github.com/eclipse-platform/eclipse.platform.swt/pull/1011

Good. I'm not sure it was only done for Linux or for every platform.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GCData.java
```diff
@@ -57,4 +60,13 @@ public final class GCData {
...
 	public boolean focusDrawn;
+
+	int getNativeDeviceZoom() {
+		return nativeDeviceZoom;
+	}
+
+	void setNativeDeviceZoom(int nativeDeviceZoom) {
+		this.nativeDeviceZoom = nativeDeviceZoom;
+		this.deviceZoom = DPIUtil.getZoomForAutoscaleProperty(nativeDeviceZoom);
+	}
```
HeikoKlare, line 71: What is the purpose of these methods?
- `getNativeDeviceZoom()` is rather obsolete as the field for which it serves as an accessor is public anyway.
- `setNativeDeviceZoom()` defines some protocol for setting the `nativeDeviceZoom` along with the `deviceZoom`. While this is a good idea, this only really makes sense if the class provides a proper encapsulation and enforces the protocol for setting the values. As both fields are public, this is not the case here. There is one caller of this method in `GC`, but the same logic is reimplenmented in `Control` as that class cannot access this method for visibility reasons. I would expect the responsibility for correctly setting these values to either be at the class' consumer or at the class itself but not scattered across both.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -133,6 +134,13 @@ public final class GC extends Resource {
  */
 public GC(Drawable drawable) {
 	this(drawable, SWT.NONE);
+	if (drawable instanceof Control) {
+		data.setNativeDeviceZoom(((Control) drawable).getShell().nativeZoom);
+	} else if(drawable instanceof Widget) {
+		data.deviceZoom = ((Widget) drawable).zoom;
```
HeikoKlare, line 140: Why is it okay here that `nativeDeviceZoom` is unset? This will result in `getDeviceZoom()` returning some fallback value (based on `DPIUtil`). If I see correctly, this is used by the `setFont` method, which then creates a font with a scale value vased on that fallback value. My naive expectation would be that this will eventually fail when having multiple shells on monitors with different scale values and using both Controls (with a properly set native zoom) and Widgets (with a fallback value). Even if that behavior is intended, it is rather incomprehensible why is makes sense to use a fallback value in some cases.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GCData.java
HeikoKlare, line 52: One rather general question: why do these values have to be stored in the `GCData`? In my understanding, these values can be retrieved from a GC's drawable. So isn't this just some kind of cache? Then if this caching is really necessary (e.g., for performance reasons), is it really necessary to cache the values in `GCData` instead of `GC` if they are only used there? The current implementation also seems to require that a `Control` changes the zoom values in `GCData`, but wouldn't it be better to have a `GC` registered as a listener for zoom change events at the `Control`, so that it can update these information on its own (just like we discussed for any other kind of class that caches some values that need to updated upon zoom changes).

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -133,6 +134,13 @@ public final class GC extends Resource {
  */
 public GC(Drawable drawable) {
 	this(drawable, SWT.NONE);
+	if (drawable instanceof Control) {
```
HeikoKlare, line 137: Why is it okay to make the initialization of the `(native)DeviceZoom` values only in one of the `GC` constructors, but not in the other?

HeikoKlare (dismissed):
I have some remarks and questions regarding the current design for which I need some clarification before I can approve this. The essential problem I have here is with respect to responsibilities. My understanding is that in the proposed implementation `Controls` are reponsible for updating zoom data used by a `GC` indirectly via changing values in `GCData`. This leads to visibility problems, duplicated logic for setting these values across package borders etc.. Since these values only seem to be a cache of information provided by the underlying `Drawable` of a `GC`, I do not understand why (1) this caching is necessary and (2) why `GC` cannot simply listen for zoom change events at the `Control` and then update the data itself but requires the `Control` to take care of it.
And even if the information really have to be provided to `GCData` (e.g., because they will be required by other `Resources`, as mentioned in the PR description), I would still expect the `GC` to be responsible for updating them and not the `Controls`. Then you should also get rid of the current visibility problems.

Please also extend the commit description with information about
- why these changes are necessary, and
- what they do,

so that people are able to understand the changes only by the commit, even without the PR.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Shell.java
fedejeanne, line 141: Is this field still necessary? According to https://github.com/eclipse-platform/eclipse.platform.swt/pull/1229/files#diff-5566416573176154bc5feeca2a6faf274ff547a9dec2d305fef52de47d8e7ea5R128, it is not.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
```diff
@@ -3563,6 +3563,12 @@ public void setMenu (Menu menu) {
 	this.menu = menu;
 }
 
+void setNativeZoomForGCData(GCData data) {
+	int nativeDeviceZoom = getShell().getNativeZoom();
+	data.nativeDeviceZoom = nativeDeviceZoom;
+	data.deviceZoom = DPIUtil.getZoomForAutoscaleProperty(nativeDeviceZoom);
+}
```
fedejeanne, line 3570: This is a case of "[feature envy](https://refactoring.guru/smells/feature-envy)"

If you address [Heiko's comment](https://github.com/eclipse-platform/eclipse.platform.swt/pull/1214/files#r1599589995) and manage to remove the fields from `GCData` altogether then this method wouldn't be necessary. 

WDYT, would that be possible?

fedejeanne (changes requested):
It seems that some fields added here will not be necessary in future PRs (see `Shell` and `Widget`), please remove them.

Also, the method `GCData::setNativeZoomForGCData` suffers from a code smell, it would be nice if it's not necessary or if it can be done differently.

Have the tests been addressed yet?

amartya4256:
> It seems that some fields added here will not be necessary in future PRs (see `Shell` and `Widget`), please remove them.
> 
> Also, the method `GCData::setNativeZoomForGCData` suffers from a code smell, it would be nice if it's not necessary or if it can be done differently.
> 
> Have the tests been addressed yet?

I updated this PR with respect to the PR #1229. It should look more consistent now.

tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/graphics/GCWin32Tests.java
```diff
@@ -0,0 +1,101 @@
...
+
+	@Before
+	public void setUp() {
+		changeAutoScaleOnRuntime(true);
+		display = Display.getDefault();
+	}
+
+	@After
+	public void tearDown() {
+		changeAutoScaleOnRuntime(false);
```
fedejeanne, line 44: It would be better to store the old value in `setup()` and set it back in `tearDown()` instead of assuming that `false` was the initial value.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -3887,6 +3909,7 @@ long identity() {
 
 void init(Drawable drawable, GCData data, long hDC) {
 	int foreground = data.foreground;
+	this.data = data;
```
fedejeanne, line 3912: Why move it up?

tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/graphics/GCWin32Tests.java
amartya4256: We want it to be falsified intentionally otherwise it will mess up with other tests. The tests which need autoscaling can turn it on in their setup.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GCData.java
```diff
@@ -47,6 +47,7 @@ public final class GCData {
 	public float[] lineDashes;
 	public float lineMiterLimit = 10;
 	public int alpha = 0xFF;
+	public int nativeDeviceZoom;
```
HeikoKlare, line 50: Is `deviceZoom` the correct word here? The value is not tied to the device but to the consumer (widget or rather it's shell), isn't it? This applies to all field, parameter, method ... names in this PR.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Widget.java
```diff
@@ -2647,6 +2647,11 @@ void notifyDisposalTracker() {
 	}
 }
 
+GC getNewGC(long hDC, GCData data) {
```
HeikoKlare, line 2650: This is not a getter but a creator for a new GC. So it should be rather called `createNewGC` or the like.

tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/graphics/GCWin32Tests.java
HeikoKlare: Of course, isolating tests is a proper goal that needs to be ensured by an according teardown implementation. What @fedejeanne means is that you assume a specific default value for `autoScaleOnRuntime`. More precisely, you assume that setting the `autoScaleOnRuntime` to `false` restores the `default` state, but you do not know whether at some point in time the default may change to `autoScaleOnRuntime=true`. There might be two ways to restore the proper value:
1. There is some central place where the default value is defined
2. You have to store the original value during `setup` and restore it during `teardown`.

tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/graphics/GCWin32Tests.java
```diff
@@ -0,0 +1,101 @@
...
+		gc.getGCData().lineWidth = 10;
+		assertEquals("DPIUtil calls with getDeviceZoom should scale to the right value", gc.getGCData().lineWidth, gc.getLineWidth() * scalingFactor, 0);
+	}
+
+	void changeAutoScaleOnRuntime(boolean value) {
+        try {
+            Field autoScaleOnRuntimeField = DPIUtil.class.getDeclaredField("autoScaleOnRuntime");
+            autoScaleOnRuntimeField.setAccessible(true);
+            autoScaleOnRuntimeField.setBoolean(null, value);
+            autoScaleOnRuntimeField.setAccessible(false);
```
HeikoKlare, line 94: Unfortunately, using reflection in tests is not a good idea in most cases. With this, you assume that simply writing a private field does not conflict with any protocol that is usally ensures (and tested) via public API. So it would be good to think about whether you have or can provide a "direct" and properly documented/tested way of setting this value that you can use within the test. For example, you could have a package protected method to change this value that is then accessed via unit tests, which are placed in the same plug-in and package.

HeikoKlare (commented):
In general, these changes now look fine to me. With the preparation in #1229, my recent concerns have been addressed.
I have some remaining comments on naming and test methodology, all of them being minor.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Widget.java
akoch-yatta: That will be changed/removed after our discussion yesterday anyway

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GCData.java
amartya4256: Let's name it nativeZoom. It contains the nativeZoom of the monitor on which GC is going to paint.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Widget.java
amartya4256: So I'll wait for the finalizing of what needs to be added and then directly add those changes (directly using the monitor zoom).
@HeikoKlare for your context, this method is in the parent of all the consumers who were creating the GC and were supposed to pass their native zoom in ther GCData. For now they just call this method which is defined in their parents which inserts the nativeZoom more like an interceptor and then creates a new GC with it. But we'll change it later.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
amartya4256: moved back

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -5120,4 +5158,12 @@ private static int sin(int angle, int length) {
 	return (int)(Math.sin(angle * (Math.PI/180)) * length);
 }
 
+private int getNativeDeviceZoom() {
+	return data.nativeZoom != 0 ? data.nativeZoom : DPIUtil.getNativeDeviceZoom();
```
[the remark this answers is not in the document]
amartya4256: You are right. The implementation of extractZoom was put, later on and with the current implementation, we don't need getNativeZoom method since it makes sure that the GCData would always have a nativeZoom.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -4582,8 +4618,9 @@ void setLineAttributesInPixels (LineAttributes attributes) {
 		}
 		if (changed) {
 			float[] newDashes = new float[dashes.length];
+			int deviceZoom = getDeviceZoom();
```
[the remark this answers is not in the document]
amartya4256: There's a loop below and this value is used in the loop. Just wanted to make sure that the value is consistent throughout the loop and doesnt change due to some simultaneous event.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -3886,6 +3908,7 @@ long identity() {
 }
 
 void init(Drawable drawable, GCData data, long hDC) {
+	this.data = data;		// Need to set GCData first because it is needed for getDeviceZoom calls
```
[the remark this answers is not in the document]
amartya4256: What do you recommend? @HeikoKlare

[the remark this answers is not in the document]
amartya4256: Resolved

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
fedejeanne: As discussed in the call: leave it at the end (_i.e._ revert this change) and adapt the usages of `this.data` to avoid using it before the `init` method has finished.

HeikoKlare (commented):
The changes look good now. Only the "actual" commit of this PR seems to contain the changes that have been extracted into #1258. Could you please rebase this PR onto its dependencies (i.e., the PRs that need to be merged before)?

HeikoKlare:
@amartya4256 merging the required PR introduced a conflict in this PR, so please resolve and rebase on master once you find the time

amartya4256:
> @amartya4256 merging the required PR introduced a conflict in this PR, so please resolve and rebase on master once you find the time

This PR now depends on https://github.com/eclipse-platform/eclipse.platform.swt/pull/1258 because of splitting the OS bindings to a separate PR. I have rebased it to the PR and I would update the description.

HeikoKlare (commented):
The changes themselves still look good.
@amartya4256 Could you please improve the commit message with some more details on _how_ GC is changed to enhance multi-zoom support?

fedejeanne (commented):
LGTM. I'm going to test it tomorrow and merge it.

PR #1219, opened 2024-05-07
Remove macos-large image from GitHub Actions configuration

HeikoKlare:
The macos-large-* images currently produce costs which is why we are asked not to use them.

See mail by Frederic Gurr.


.github/workflows/maven.yml
```diff
@@ -31,7 +31,6 @@ jobs:
         config: 
           - { name: Linux, os: ubuntu-latest, native: gtk.linux.x86_64 }
           - { name: Windows, os: windows-latest, native: win32.win32.x86_64 }
-          - { name: MacOS x86, os: macos-latest-large, native: cocoa.macosx.x86_64 }
```
laeubi, line 34: can't we use macos-13 instead?

HeikoKlare:
Closed in favor of #1220.

PR #1229, opened 2024-05-14
Store nativeDeviceZoom in widgets instead of zoom for win32

amartya4256:
## Addressed issues
* https://github.com/eclipse-platform/eclipse.platform.swt/issues/127

## Unlocks
* https://github.com/eclipse-platform/eclipse.platform.swt/pull/1214

## Description

This contribution is to provide the nativeDeviceZoom to all the widgets so that it can be used later on for e.g. font scaling and more. Currently, the nativeDeviceZoom is only available via the shell, which is not sufficient later when there is no shell available, e.g. GC.

In this following PR https://github.com/eclipse-platform/eclipse.platform.swt/pull/1214 , we had discussed that instead of storing zoom for widgets, we should store nativeDeviceZoom, to make it more accessible where shell is not present. Because in some cases, there was no nativeDeviceZoom present because of the previous implementation and we discussed if a fallback might be necessary. This implementation makes sure that in those scenarios, nativeDeviceZoom would always be available for further usage.

The consumers can now have deviceZoom wherever necessary by using the utility method to translate nativeDeviceZoom to deviceZoom, which is available in DPIUtil.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
@@ -137,9 +137,9 @@ public final class Image extends Resource implements Drawable {
 	private int styleFlag = SWT.IMAGE_COPY;
 
 	/**
-	 * Attribute to cache current device zoom level
+	 * Attribute to cache current native zoom level
 	 */
-	private int currentDeviceZoom = 100;
+	private int currentNativeZoom = 100;
 
 	/**
 	 * width of the image
@@ -161,7 +161,7 @@ public final class Image extends Resource implements Drawable {
  */
 Image (Device device) {
 	super(device);
-	currentDeviceZoom = DPIUtil.getDeviceZoom ();
+	currentNativeZoom = DPIUtil.getNativeDeviceZoom();
 }
 
 /**
@@ -201,7 +201,7 @@ Image (Device device) {
  */
 public Image(Device device, int width, int height) {
 	super(device);
-	currentDeviceZoom = DPIUtil.getDeviceZoom ();
+	currentNativeZoom = DPIUtil.getNativeDeviceZoom();
 	width = DPIUtil.autoScaleUp (width);
 	height = DPIUtil.autoScaleUp (height);
 	init(width, height);
@@ -254,7 +254,7 @@ public Image(Device device, Image srcImage, int flag) {
 	this.imageDataProvider = srcImage.imageDataProvider;
 	this.imageFileNameProvider = srcImage.imageFileNameProvider;
 	this.styleFlag = srcImage.styleFlag | flag;
-	this.currentDeviceZoom = srcImage.currentDeviceZoom;
+	currentNativeZoom = srcImage.currentNativeZoom;
 	this.dataAtBaseZoom = srcImage.dataAtBaseZoom;
 	switch (flag) {
 		case SWT.IMAGE_COPY: {
@@ -293,7 +293,7 @@ public Image(Device device, Image srcImage, int flag) {
 			break;
 		}
 		case SWT.IMAGE_DISABLE: {
-			ImageData data = srcImage.getImageData(srcImage.currentDeviceZoom);
+			ImageData data = srcImage.getImageData(srcImage.getZoom());
 			PaletteData palette = data.palette;
 			RGB[] rgbs = new RGB[3];
 			rgbs[0] = device.getSystemColor(SWT.COLOR_BLACK).getRGB();
@@ -352,7 +352,7 @@ public Image(Device device, Image srcImage, int flag) {
 			break;
 		}
 		case SWT.IMAGE_GRAY: {
-			ImageData data = srcImage.getImageData(srcImage.currentDeviceZoom);
+			ImageData data = srcImage.getImageData(srcImage.getZoom());
 			PaletteData palette = data.palette;
 			ImageData newData = data;
 			if (!palette.isDirect) {
@@ -459,7 +459,7 @@ public Image(Device device, Image srcImage, int flag) {
 public Image(Device device, Rectangle bounds) {
 	super(device);
 	if (bounds == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
-	currentDeviceZoom = DPIUtil.getDeviceZoom ();
+	currentNativeZoom = DPIUtil.getNativeDeviceZoom();
 	bounds = DPIUtil.autoScaleUp (bounds);
 	init(bounds.width, bounds.height);
 	init();
@@ -491,7 +491,7 @@ public Image(Device device, Rectangle bounds) {
 public Image(Device device, ImageData data) {
 	super(device);
 	if (data == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
-	currentDeviceZoom = DPIUtil.getDeviceZoom ();
+	currentNativeZoom = DPIUtil.getNativeDeviceZoom();
 	this.dataAtBaseZoom = new ElementAtZoom<>(data, 100);
 	data = DPIUtil.autoScaleUp(device, this.dataAtBaseZoom);
 	init(data);
@@ -535,7 +535,7 @@ public Image(Device device, ImageData source, ImageData mask) {
 	if (source.width != mask.width || source.height != mask.height) {
 		SWT.error(SWT.ERROR_INVALID_ARGUMENT);
 	}
-	currentDeviceZoom = DPIUtil.getDeviceZoom ();
+	currentNativeZoom = DPIUtil.getNativeDeviceZoom();
 	this.dataAtBaseZoom = new ElementAtZoom<>(applyMask(source, ImageData.convertMask(mask)), 100);
 	source = DPIUtil.autoScaleUp(device, source);
 	mask = DPIUtil.autoScaleUp(device, mask);
@@ -599,7 +599,7 @@ public Image(Device device, ImageData source, ImageData mask) {
  */
 public Image (Device device, InputStream stream) {
 	super(device);
-	currentDeviceZoom = DPIUtil.getDeviceZoom ();
+	currentNativeZoom = DPIUtil.getNativeDeviceZoom();
 	this.dataAtBaseZoom =  new ElementAtZoom<>(new ImageData (stream), 100);
 	ImageData data = DPIUtil.autoScaleUp(device, this.dataAtBaseZoom);
 	init(data);
@@ -641,7 +641,7 @@ public Image (Device device, InputStream stream) {
 public Image (Device device, String filename) {
 	super(device);
 	if (filename == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
-	currentDeviceZoom = DPIUtil.getDeviceZoom ();
+	currentNativeZoom = DPIUtil.getNativeDeviceZoom();
 	this.dataAtBaseZoom = new ElementAtZoom<>(new ImageData (filename), 100);
 	ImageData data = DPIUtil.autoScaleUp(device, this.dataAtBaseZoom);
 	init(data);
@@ -680,9 +680,9 @@ public Image (Device device, String filename) {
 public Image(Device device, ImageFileNameProvider imageFileNameProvider) {
 	super(device);
 	this.imageFileNameProvider = imageFileNameProvider;
-	currentDeviceZoom = DPIUtil.getDeviceZoom ();
-	ElementAtZoom<String> fileName = DPIUtil.validateAndGetImagePathAtZoom (imageFileNameProvider, currentDeviceZoom);
-	if (fileName.zoom() == currentDeviceZoom) {
+	currentNativeZoom = DPIUtil.getNativeDeviceZoom();
+	ElementAtZoom<String> fileName = DPIUtil.validateAndGetImagePathAtZoom (imageFileNameProvider, getZoom());
+	if (fileName.zoom() == getZoom()) {
 		initNative (fileName.element());
 		if (this.handle == 0) init(new ImageData (fileName.element()));
 	} else {
@@ -724,26 +724,28 @@ public Image(Device device, ImageFileNameProvider imageFileNameProvider) {
 public Image(Device device, ImageDataProvider imageDataProvider) {
 	super(device);
 	this.imageDataProvider = imageDataProvider;
-	currentDeviceZoom = DPIUtil.getDeviceZoom ();
-	ElementAtZoom<ImageData> data =  DPIUtil.validateAndGetImageDataAtZoom(imageDataProvider, currentDeviceZoom);
+	currentNativeZoom = DPIUtil.getNativeDeviceZoom();
+	ElementAtZoom<ImageData> data =  DPIUtil.validateAndGetImageDataAtZoom(imageDataProvider, getZoom());
 	ImageData resizedData = DPIUtil.autoScaleImageData(device, data.element(), data.zoom());
 	init (resizedData);
 	init();
 }
 
 /**
- * Update zoom and refresh the Image based on the zoom level, if required.
+ * Update zoom and refresh the Image based on the native zoom level, if required.
  *
- * @param deviceZoom zoom in % of the standard resolution the image shall be scaled for
+ * @param nativeZoom native zoom in % of the monitor on which the image is painted
  *
  * @return true if image is refreshed
  */
-boolean handleDPIChange (int deviceZoom) {
+boolean handleDPIChange (int nativeZoom) {
+	int newZoom = DPIUtil.getZoomForAutoscaleProperty(nativeZoom);
+	int oldZoom = this.getZoom();
 	boolean refreshed = false;
 	if (imageFileNameProvider != null) {
-		if (deviceZoom != currentDeviceZoom) {
-			ElementAtZoom<String> filename = DPIUtil.validateAndGetImagePathAtZoom (imageFileNameProvider, deviceZoom);
-			if (filename.zoom() == deviceZoom) {
+		if (newZoom != oldZoom) {
+			ElementAtZoom<String> filename = DPIUtil.validateAndGetImagePathAtZoom (imageFileNameProvider, newZoom);
+			if (filename.zoom() == newZoom) {
 				/* Release current native resources */
 				destroy ();
 				initNative(filename.element());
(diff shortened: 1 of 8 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -161,7 +161,7 @@ public final class Image extends Resource implements Drawable {
  */
 Image (Device device) {
 	super(device);
-	currentDeviceZoom = DPIUtil.getDeviceZoom ();
+	currentDeviceZoom = DPIUtil.getNativeDeviceZoom();
```
HeikoKlare, line 164: I think the field should not be called `currentDeviceZoom` anymore, as it suggests it was the actually used device zoom and not the "native" one.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -738,10 +738,11 @@ public Image(Device device, ImageDataProvider imageDataProvider) {
  *
  * @return true if image is refreshed
  */
-boolean handleDPIChange (int deviceZoom) {
+boolean handleDPIChange (int nativeZoom) {
```
HeikoKlare, line 741: Javadoc needs to be adapted when changing parameter names

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2257,14 +2258,18 @@ public void setBackground(Color color) {
 }
 
 private void setCurrentDeviceZoom(int newZoomFactor) {
```
HeikoKlare, line 2260: This method does not set the current device zoom anymore, but the native zoom, so please rename it.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Shell.java
```diff
@@ -2650,19 +2645,8 @@ LRESULT WM_WINDOWPOSCHANGING (long wParam, long lParam) {
...
-	return nativeZoom;
-}
-
-void setNativeZoom(int nativeZoom) {
-	this.nativeZoom = nativeZoom;
-}
-
 private void handleZoomEvent(Event event) {
-	float scalingFactor = 1f * event.detail / getZoom();
+	float scalingFactor = 1f * DPIUtil.getZoomForAutoscaleProperty(event.detail) / getZoom();
```
HeikoKlare, line 2649: Could you please assign the `event.detail` value to a variable with a self-explaining name to make this more comprehensive?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Widget.java
```diff
@@ -2639,15 +2639,19 @@ void notifyDisposalTracker() {
 /**
  * The current DPI zoom level the widget is scaled for
  */
```
HeikoKlare, line 2641: Please adapt Javadoc

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
```diff
@@ -4885,20 +4885,18 @@ LRESULT WM_DESTROY (long wParam, long lParam) {
...
 		Event event = new Event();
 		event.type = SWT.ZoomChanged;
 		event.widget = this;
-		event.detail = newSWTZoom;
+		event.detail = nativeZoom;
 		event.doit = true;
 
 		if (DPIUtil.isAutoScaleOnRuntimeActive()) {
 			DPIUtil.setDeviceZoom (nativeZoom);
-			getShell().setNativeZoom(nativeZoom);
```
HeikoKlare, line 4901: Just out of curiosity: why was it sufficient to set the native but not the current zoom here and why is it okay to completely remove it?

HeikoKlare (changes requested):
In general, the changes look sound and storing the native zoom for which the actually used value can always be calculated on demand absolutely makes sense.

Currently, there are several naming and consistency issues, as identifiers have not been renamed, Javadoc has not been adapted etc.
Please have a look that all the identifiers (variable, field, method names) still properly capture their meaning.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
```diff
@@ -4885,20 +4885,18 @@ LRESULT WM_DESTROY (long wParam, long lParam) {
 LRESULT WM_DPICHANGED (long wParam, long lParam) {
 	// Map DPI to Zoom and compare
 	int nativeZoom = DPIUtil.mapDPIToZoom (OS.HIWORD (wParam));
-	int newSWTZoom = DPIUtil.getZoomForAutoscaleProperty (nativeZoom);
-	int oldSWTZoom = getShell().getZoom();
+	int oldNativeZoom = getShell().getNativeZoom();
 
 	// Throw the DPI change event if zoom value changes
-	if (newSWTZoom != oldSWTZoom) {
+	if (nativeZoom != oldNativeZoom) {
```
akoch-yatta, line 4891: Just a remark about this line, so everyone is aware of it. Before the deviceZoom was checked, so for the default value for swt.autoScale (integer200) a zoom change from 125 to 150 was converted to newSWTZoom and oldSWTZoom  both being 100, so no zoom event was propagated. Now, as the native zoom is used, this condition will be true and the event will be propagated. As fonts are based on the native zoom, they will be adapted. I this this behaviour is more consistent, but wanted to point it out.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2256,15 +2257,19 @@ public void setBackground(Color color) {
 	device.internal_dispose_GC(hDC, null);
 }
 
-private void setCurrentDeviceZoom(int newZoomFactor) {
-	if (this.currentDeviceZoom != newZoomFactor) {
-		this.currentDeviceZoom = newZoomFactor;
+private void setCurrentNativeZoom(int newZoomFactor) {
+	if (this.getZoom() != newZoomFactor) {
```
fedejeanne, line 2261: Can you explain how does this work? Specifically: why do you now compare using `getZoom()` instead of the current device zoom (it was `currentDeviceZoom` before).
It seems a bit odd that ones compares against the output of a method instead of the previously stored value but I assume it has something to do with rounding errors?

Also, please rename the parameter to `deviceZoom` since it's going to end up being stored in `this.currentDeviceZoom` (make the names "match (better)")

fedejeanne (changes requested):
Changes look good, I only have one question and one minor remark for the method `Image::setCurrentNativeZoom`

Also, just for my understanding (recap of "the big picture"), this PR "merely": 
* Replaces the previously cached device zoom in `Image` (`currentDeviceZoom`) by a calculated zoom (`Image::getZoom()` --> `DPIUtil::getZoomForAutoscaleProperty(...)`)
* Propagates the native device zoom instead of the (current) device zoom in GC
* Adapts `Image::handleDPIChange(int)` to use the newly propagated native device zoom

Did I miss anything?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
amartya4256: It was current device zoom earlier and now it is current NATIVE zoom. getZoom method converts native zoom to device zoom. For image's internal usage only device zoom is relevant. In this case we only wanna create a new handle for the image when the device zoom chnages (i.e. if the native zoom changes from 100 to 125, the device zoom would still stay 100, in this case we would end up creating 2 handles for 100 whioch is inefficient). Also the setCurrentNativeZoom sets the height and width of image to -1, making it need to call init method again to be usable. 

For this PR, this is sufficient since in the later GC and Image PRs, we would extend to the current native zoom to provide the native zoom to GCData when creating a new GC for a drawable of type Image. In further Image PR, you'd see this behaviour changing because then we would not need to set current native zoom anymore for the image but rather initial native zoom only.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -724,24 +724,25 @@ public Image(Device device, ImageFileNameProvider imageFileNameProvider) {
...
  *
  * @return true if image is refreshed
  */
-boolean handleDPIChange (int deviceZoom) {
+boolean handleDPIChange (int nativeZoom) {
+	int deviceZoom = DPIUtil.getZoomForAutoscaleProperty(nativeZoom);
 	boolean refreshed = false;
 	if (imageFileNameProvider != null) {
-		if (deviceZoom != currentDeviceZoom) {
+		if (deviceZoom != getZoom()) {
```
HeikoKlare, line 745: Is `deviceZoom` correct here? It's not the zoom of the device but of a monitor, isn't is? Maybe you could also change `deviceZoom` to `newZoom` and assign `getZoom()` to some `oldZoom` variable to make it easier to understand what happens here.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
HeikoKlare: Still the name `zoomFactor` seems to be inappropriate. And assigning `getZoom()` to some `oldZoom` variable would make the code more comprehensible.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
HeikoKlare: Thanks for pointing that out! Interesting observation and probably something that would have produced bugs later on.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
HeikoKlare: Thanks for the explanation. That makes the changes of this PR even more valuable in terms of reducing responsibilities for propagating zoom changes.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Widget.java
```diff
@@ -2637,17 +2637,21 @@ void notifyDisposalTracker() {
...
+}
+
+void setNativeZoom(int zoom) {
+	this.nativeZoom = zoom;
 }
 
-void setZoom(int zoom) {
-	this.zoom = zoom;
+int getZoom() {
+	return DPIUtil.getZoomForAutoscaleProperty(getNativeZoom());
```
HeikoKlare, line 2651: Even though it is generally good practice to access fields via accessors, I think in this case we could directly access `nativeZoom` instead of using its accessor. Since this method will be called really often, having direct access to the field and not requiring the JIT to inline the access may be reasonable (but could also be considered a premature optimization).

HeikoKlare (commented):
Thanks for updating the PR. I only have minor comments left.

Note that my review is currently based on GitHub UI, as I don't have a Windows laptop available. So I will not "formally" approve the PR as I would like to validate the changes in Eclipse for that.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
amartya4256: no we calculate the device zoom from native zoom at line no. 742
and getZoom gets the devicezoom calculated from the currentNativeZoom.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1348,7 +1350,7 @@ public ImageData getImageData() {
 public ImageData getImageData (int zoom) {
 	if (isDisposed()) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
 
-	if (zoom == currentDeviceZoom) {
+	if (zoom == getZoom()) {
```
fedejeanne, line 1353: Similar to https://github.com/eclipse-platform/eclipse.platform.swt/pull/1229#discussion_r1601379823, please extract `getZoom()` to a variable called `oldZoom` to improve readability. It makes sense in this case since the method receives a (new/targeted) `zoom` as parameter and the method `getZoom()` retrieves the current/soon-to-be-old zoom.

fedejeanne, line 1353: Similar to https://github.com/eclipse-platform/eclipse.platform.swt/pull/1229#discussion_r1601379823, please extract all calls to `getZoom()` inside this method to a variable called `oldZoom` to improve readability. It makes sense in this case since the method receives a (new/targeted) `zoom` as parameter and the method `getZoom()` retrieves the current/soon-to-be-old zoom.

fedejeanne (changes requested):
Thank you for addressing my previous comment @amartya4256 , I just have one final change request to improve readability. After that, this PR is good to go from my side.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2256,15 +2258,20 @@ public void setBackground(Color color) {
 	device.internal_dispose_GC(hDC, null);
 }
 
-private void setCurrentDeviceZoom(int newZoomFactor) {
-	if (this.currentDeviceZoom != newZoomFactor) {
-		this.currentDeviceZoom = newZoomFactor;
+private void setCurrentNativeZoom(int newZoom) {
+	int oldZoom = this.getZoom();
+	if (oldZoom != newZoom) {
```
HeikoKlare, line 2263: The naming is now confusing, because `oldZoom` referes to a "corrected" zoom value, whereas `newZoom` is a native zoom value. Maybe `newZoom` should be called `newNativeZoom` instead? In addition, the comparison `oldZoom != new[Native]Zoom` probably requires a comment to understand why native and non-native zoom values are compared.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
```diff
@@ -4885,20 +4885,18 @@ LRESULT WM_DESTROY (long wParam, long lParam) {
 LRESULT WM_DPICHANGED (long wParam, long lParam) {
 	// Map DPI to Zoom and compare
 	int nativeZoom = DPIUtil.mapDPIToZoom (OS.HIWORD (wParam));
```
HeikoKlare, line 4887: Maybe this could be called `newNativeZoom`?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Widget.java
```diff
@@ -52,7 +52,19 @@
...
+	 * <p>
+	 * <b>IMPORTANT:</b> This field is <em>not</em> part of the SWT
+	 * public API. It is marked public only so that it can be shared
+	 * within the packages provided by SWT. It is not available on all
+	 * platforms and should never be accessed from application code.
+	 * </p>
+	 *
+	 * @noreference This field is not intended to be referenced by clients.
+	 */
+	protected int nativeZoom;
```
HeikoKlare, line 67: Can't this be package protected (for the scope of this PR) and increased in visibility once required?

HeikoKlare (commented):
In general, this PR looks good and sound to me. I only have some nitpicky comments regarding naming and one regarding visibilities.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
amartya4256: I have made changes in this method to fix the names and usage. However, in a bigger picture it is a stale method and would be removed in the image PR. The comparision should be between old device zoom and new device zoom, because we dont wanna create many handles for the same device zoom. (Since multiple native zooms can have same device zoom)

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
amartya4256: in this case we would call it currentZoom since here we are not setting the new zoom but requesting imageData at a particular zoom. I'd propose to call it currentZoom. The logic is like if the zoom is equal to currentZoom of the Image then return the imageData of the image otherwise do some caling and other stuff.

amartya4256: already addressed in the comment above :)

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
HeikoKlare: Two further remarks on this:
1. Even if a method is supposed to be removed in the future, current changes to it still need to be sound and of good quality. No one can guarantee that a currently intended subsequent change will happen, so we can usually only review and decide based on what exists and what is done in this particular PR.
2. I understand the goal of having only one handle per device zoom, but it still looks wrong that an image does not update its native zoom just because the actually used zoom calculated from the native zoom did not change. In my opinion, the necessity not to update something when the native zoom changed without affecting the actual zoom has to be done in some recalculation logic but not when setting the native zoom value.

HeikoKlare:
@amartya4256 I don't see my previous comment (https://github.com/eclipse-platform/eclipse.platform.swt/pull/1229#discussion_r1611590128) addressed yet. Could you please check before I re-review the PR?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
amartya4256: I have resolved this now. Also made sure that it wont have any side effects with generatign new handles.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2256,15 +2258,20 @@ public void setBackground(Color color) {
 	device.internal_dispose_GC(hDC, null);
 }
 
-private void setCurrentDeviceZoom(int newZoomFactor) {
-	if (this.currentDeviceZoom != newZoomFactor) {
-		this.currentDeviceZoom = newZoomFactor;
+private void setCurrentNativeZoom(int newNativeZoom) {
+	int oldZoom = this.getZoom();
+	if (oldZoom != newNativeZoom) {
```
HeikoKlare, line 2263: I still don't see my comment (https://github.com/eclipse-platform/eclipse.platform.swt/pull/1229#discussion_r1611590128) properly addressed: the current implementation does still not always update the `currentNativeZoom` with the `newNativeZoom` when this method is called, which I find confusing. This setter is only called when the current and old current zoom values differ anyway, so couldn't the `currentNativeZoom` simply always be updated when this method is called?
In addition, the modification of the condition (now comparing against the new native zoom instead of the new current zoom) changes when the `height` and `width` value will be reset. Is this functional change intended? If yes, the comment needs to be updated.

HeikoKlare (approved):
To me, this looks good now. Thank you!

amartya4256:
> Thank you for addressing my previous comment @amartya4256 , I just have one final change request to improve readability. After that, this PR is good to go from my side.

@fedejeanne can you approved the changes? :)

fedejeanne (approved):

HeikoKlare:
The changes are basically a refactoring, which we have reviewed at code level. Still, I have now also done a final manual validation. To this end, I have started an Eclipse SDK with different primary monitor scaling values (100%, 150%, 200%), opened different views, editors and dialogs, and compared the results of the state before this PR and after this PR side-by-side. As expected, they were exactly the same in all cases.

Issue #1228, opened 2024-05-14
[Mac Sonoma] "WARNING: Secure coding is not enabled for restorable state" warning message


Phillipus:
**The issue**
When running Eclipse or an RCP app on macOS Sonoma a warning message is written to the console:

`eclipse[4647:170325] WARNING: Secure coding is not enabled for restorable state! Enable secure coding by implementing NSApplicationDelegate.applicationSupportsSecureRestorableState: and returning YES.`

I've Googled this and it seems that this could lead to a vulnerabilty and needs to be addressed. For example see here:

https://stackoverflow.com/questions/77283578/sonoma-and-nsapplicationdelegate-applicationsupportssecurerestorablestate?rq=2

I'm reporting this against SWT because SWT provides the native implementations of Mac classes such as `NSApplicationDelegate`.

@lshanmug what do you think? Can SWT implement `applicationSupportsSecureRestorableState` in `NSApplicationDelegate`?

**To Reproduce**
The only way to see this warning message is to run Eclipse on Mac Sonoma from the command line so the message is written to the console in the Terminal app:

1. Open the Mac Terminal app and run the following command:
`Eclipse.app/Contents/MacOS/eclipse`
2. The warning message is written to the terminal window

**Expected behavior**
The warning message should not appear.

**Environment:**
macOS Sonoma 14.5
MacBook M2 aarch64
Temurin JDK 17.0.11+9
Eclipse 4.31 and latest 4.32 I builds

Phillipus:
Actually, after debugging this some more, the warning message appears before any SWT components are created. It occurs in equinox:


https://github.com/eclipse-equinox/equinox/tree/master/features/org.eclipse.equinox.executable.feature/library/cocoa

Phillipus:
Could someone transfer this issue to https://github.com/eclipse-equinox/equinox please? This is not SWT.

fedejeanne:
You could just copy it there since the issue is new and no discussion really happened :)

Phillipus:
Copied to https://github.com/eclipse-equinox/equinox/issues/630

Phillipus:
Re-opening this. PR for SWT incoming...

Phillipus:
[...]

PR #1231, opened 2024-05-15
Fix warning message on Mac about secure coding not enabled

Phillipus:
- See https://github.com/eclipse-platform/eclipse.platform.swt/issues/1228

- On macOS 14 and later a warning message is written to console:

"WARNING: Secure coding is not enabled for restorable state! Enable secure coding by implementing NSApplicationDelegate.applicationSupportsSecureRestorableState: and returning YES."

- As recommended by Apple, this change adds a new selector for applicationSupportsSecureRestorableState and returns a 1 value (YES) in the callback in the Display class

- However, this only takes care of the NSApplicationDelegate created in the Display class. The warning message will still appear when a splash screen is created in Equinox JNI code so a fix there is also required. See https://github.com/eclipse-equinox/equinox/issues/630

- This implementation for SWT can be tested by using the -noSplash argument when launching Eclipse

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT PI/cocoa/org/eclipse/swt/internal/cocoa/OS.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT PI/cocoa/org/eclipse/swt/internal/cocoa/OS.java
@@ -156,8 +156,9 @@ public class OS extends C {
 	public static final long sel_setStyle = Selector.sel_setStyle.value;
 	public static final int NSTableViewStylePlain = 4;
 
-    /** 14.0 selector */
-    public static final long sel_setClipsToBounds_ = Selector.sel_setClipsToBounds_.value;
+	/** 14.0 selectors */
+	public static final long sel_setClipsToBounds_ = Selector.sel_setClipsToBounds_.value;
+	public static final long sel_applicationSupportsSecureRestorableState_ = Selector.sel_applicationSupportsSecureRestorableState_.value;
 
 	/* AWT application delegate. Remove these when JavaRuntimeSupport.framework has bridgesupport generated for it. */
 	public static final long class_JRSAppKitAWT = objc_getClass("JRSAppKitAWT");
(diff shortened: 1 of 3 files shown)


lshanmug (approved):
@Phillipus Changes look good. Should we merge for 4.32 RC1?

Phillipus:
> @Phillipus Changes look good. Should we merge for 4.32 RC1?

@lshanmug  The `applicationSupportsSecureRestorableState` selector was added in macOS 12 (see [here](https://developer.apple.com/documentation/appkit/nsapplicationdelegate/3762521-applicationsupportssecurerestora)) so I guess the selector is simply ignored on earlier macOS? So I think it's safe to merge.

But as mentioned above, equinox project also need to apply a fix in their native Mac code. I create a PR for that case [here](https://github.com/eclipse-equinox/equinox/pull/632).

lshanmug:
> > @Phillipus Changes look good. Should we merge for 4.32 RC1?
> 
> @lshanmug The `applicationSupportsSecureRestorableState` selector was added in macOS 12 (see [here](https://developer.apple.com/documentation/appkit/nsapplicationdelegate/3762521-applicationsupportssecurerestora)) so I guess the selector is simply ignored on earlier macOS? So I think it's safe to merge.

Yes, the selector code will not be called for earlier versions.

> 
> But as mentioned above, equinox project also need to apply a fix in their native Mac code. I create a PR for that case [here](https://github.com/eclipse-equinox/equinox/pull/632).

The SWT PR can go in anyways correct?

Phillipus:
> The SWT PR can go in anyways correct?

Yes, I think so. If Eclipse is started with the `-noSplash` option then the SWT selector is invoked and there is no warning message, as desired.

One thing I don't understand is where is the singleton `NSApplication` created? Equinox references `[NSApplication sharedApplication]` [here](https://github.com/eclipse-equinox/equinox/blob/master/features/org.eclipse.equinox.executable.feature/library/cocoa/eclipseCocoa.c) and SWT references it in the `Display class`.

lshanmug:
> 
> One thing I don't understand is where is the singleton `NSApplication` created? Equinox references `[NSApplication sharedApplication]` [here](https://github.com/eclipse-equinox/equinox/blob/master/features/org.eclipse.equinox.executable.feature/library/cocoa/eclipseCocoa.c) and SWT references it in the `Display class`.

`NSApplication.sharedApplication` itself creates the `NSApplication` instance for the first time and return it for future calls. Please see  [doc](https://developer.apple.com/documentation/appkit/nsapplication/1428360-sharedapplication).

PR #1232, opened 2024-05-15
Fix for Mac drawing bug issue 1012 part 2

Phillipus:
- OS.objc_msgSend(id, OS.sel_setClipsToBounds_, true) needs to be applied in two more cases

- See https://github.com/eclipse-platform/eclipse.platform.swt/issues/1012

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Widget.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Widget.java
@@ -2101,6 +2101,11 @@ boolean setMarkedText_selectedRange (long id, long sel, long string, long range)
 void setNeedsDisplay (long id, long sel, boolean flag) {
 	if (flag && !isDrawing()) return;
 	NSView view = new NSView(id);
+	/*
+	* Since macOS 14 the clipsToBounds property of NSView has to be set to true
+	* See https://developer.apple.com/documentation/macos-release-notes/appkit-release-notes-for-macos-14
+	*/
+	OS.objc_msgSend(id, OS.sel_setClipsToBounds_, true);
 	if (flag && display.isPainting.containsObject(view)) {
 		NSMutableArray needsDisplay = display.needsDisplay;
 		if (needsDisplay == null) {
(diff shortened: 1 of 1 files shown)


Phillipus:
This applies the fix in two more places in the `Widget` class and fixes the blank areas seen [here](https://github.com/eclipse-platform/eclipse.platform.swt/issues/1012#issuecomment-2112345916).

However, I wonder if we are playing whack-a-mole with this and the real fix is to call `OS.objc_msgSend(id, OS.sel_setClipsToBounds_, true)` in one place:

```java
public NSView(long id) {
	super(id);
	OS.objc_msgSend(id, OS.sel_setClipsToBounds_, true);
}
```

But as explained in https://github.com/eclipse-platform/eclipse.platform.swt/pull/1081#issuecomment-1984309289 we can't add it there.

@lshanmug WDYT?

Phillipus:
Actually, can someone merge this so we can get it in for 4.32?  Without this fix, users are still going to see blank areas on Mac when using a JDK linked to Mac SDK 14.

Perhaps we can look at implementing it in `NSView` after 4.32?

lshanmug:
> Actually, can someone merge this so we can get it in for 4.32? Without this fix, users are still going to see blank areas on Mac when using a JDK linked to Mac SDK 14.
> 
> Perhaps we can look at implementing it in `NSView` after 4.32?

@Phillipus Ideally the call to `setClipsToBounds` should be done in at the time of initialization to cover all cases. But, this change looks good for 4.32.

lshanmug (approved):

Phillipus:
> @Phillipus Ideally the call to `setClipsToBounds` should be done in at the time of initialization to cover all cases. But, this change looks good for 4.32.

@lshanmug Thanks for your feedback and the merge!

Phillipus:
> Ideally the call to setClipsToBounds should be done in at the time of initialization to cover all cases. But, this change looks good for 4.32.

@lshanmug I opened https://github.com/eclipse-platform/eclipse.platform.swt/issues/1235 to track that.

lshanmug:
> > Ideally the call to setClipsToBounds should be done in at the time of initialization to cover all cases. But, this change looks good for 4.32.
> 
> @lshanmug I opened #1235 to track that.

@Phillipus Thanks for the fix! I'm also thinking how to address the issue, will take a look next week and update.

PR #1236, opened 2024-05-17
Implementation of Image to enable Multi Zoom Level Support for Win32

amartya4256:
## Addressed issues

## Requires

## Unlocks
* [ ] https://github.com/eclipse-platform/eclipse.platform.swt/pull/1265

## Description

This pull request is based on the implementations of PR #1214. It extends to the native zoom which is provided within widget and propagated to GC. In this PR, we provide a map of different handles to Image for different device zoom levels as required by different monitors. The method win32_getHandle provides the appropriate handle for the scaled image as per requested by the client. The clients avail the zoom level information from the GCData and Widget (based on PR #1214 ).

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Drag and Drop/win32/org/eclipse/swt/dnd/DragSource.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Drag and Drop/win32/org/eclipse/swt/dnd/DragSource.java
@@ -491,7 +491,7 @@ private void drag(Event dragEvent) {
 	hwndDrag = 0;
 	topControl = null;
 	if (image != null) {
-		imagelist = new ImageList(SWT.NONE);
+		imagelist = new ImageList(SWT.NONE, DPIUtil.getZoomForAutoscaleProperty(nativeZoom));
 		imagelist.add(image);
 		topControl = control.getShell();
 		/*
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
@@ -300,7 +300,7 @@ public Cursor(Device device, ImageData source, int hotspotX, int hotspotY) {
 		if (hMask == 0) SWT.error(SWT.ERROR_NO_HANDLES);
 	} else {
 		ImageData mask = source.getTransparencyMask();
-		long [] result = Image.init(this.device, null, source, mask);
+		long [] result = Image.init(this.device, null, source, mask, null);	// Since the image is null, the device zoom can be null
 		hBitmap = result[0];
 		hMask = result[1];
 	}
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
@@ -139,7 +139,7 @@ public final class Image extends Resource implements Drawable {
 	/**
 	 * Attribute to cache current native zoom level
 	 */
-	private int currentNativeZoom = 100;
+	private int initialNativeZoom = 100;
 
 	/**
 	 * width of the image
@@ -156,12 +156,14 @@ public final class Image extends Resource implements Drawable {
 	 */
 	static final int DEFAULT_SCANLINE_PAD = 4;
 
+	private HashMap<Integer, Long> zoomLevelToHandle = new HashMap<>();
+
 /**
  * Prevents uninitialized instances from being created outside the package.
  */
 Image (Device device) {
 	super(device);
-	currentNativeZoom = DPIUtil.getNativeDeviceZoom();
+	initialNativeZoom = DPIUtil.getNativeDeviceZoom();
 }
 
 /**
@@ -201,7 +203,7 @@ Image (Device device) {
  */
 public Image(Device device, int width, int height) {
 	super(device);
-	currentNativeZoom = DPIUtil.getNativeDeviceZoom();
+	initialNativeZoom = DPIUtil.getNativeDeviceZoom();
 	width = DPIUtil.autoScaleUp (width);
 	height = DPIUtil.autoScaleUp (height);
 	init(width, height);
@@ -254,7 +256,7 @@ public Image(Device device, Image srcImage, int flag) {
 	this.imageDataProvider = srcImage.imageDataProvider;
 	this.imageFileNameProvider = srcImage.imageFileNameProvider;
 	this.styleFlag = srcImage.styleFlag | flag;
-	currentNativeZoom = srcImage.currentNativeZoom;
+	initialNativeZoom = srcImage.initialNativeZoom;
 	this.dataAtBaseZoom = srcImage.dataAtBaseZoom;
 	switch (flag) {
 		case SWT.IMAGE_COPY: {
@@ -269,7 +271,7 @@ public Image(Device device, Image srcImage, int flag) {
 					long hOldSrc = OS.SelectObject(hdcSource, srcImage.handle);
 					BITMAP bm = new BITMAP();
 					OS.GetObject(srcImage.handle, BITMAP.sizeof, bm);
-					handle = OS.CreateCompatibleBitmap(hdcSource, rect.width, bm.bmBits != 0 ? -rect.height : rect.height);
+					setHandleForZoomLevel(OS.CreateCompatibleBitmap(hdcSource, rect.width, bm.bmBits != 0 ? -rect.height : rect.height), this.getZoom());
 					if (handle == 0) SWT.error(SWT.ERROR_NO_HANDLES);
 					long hOldDest = OS.SelectObject(hdcDest, handle);
 					OS.BitBlt(hdcDest, 0, 0, rect.width, rect.height, hdcSource, 0, 0, OS.SRCCOPY);
@@ -284,7 +286,7 @@ public Image(Device device, Image srcImage, int flag) {
 					transparentPixel = srcImage.transparentPixel;
 					break;
 				case SWT.ICON:
-					handle = OS.CopyImage(srcImage.handle, OS.IMAGE_ICON, rect.width, rect.height, 0);
+					setHandleForZoomLevel(OS.CopyImage(srcImage.handle, OS.IMAGE_ICON, rect.width, rect.height, 0), getZoom());
 					if (handle == 0) SWT.error(SWT.ERROR_NO_HANDLES);
 					break;
 				default:
@@ -348,7 +350,7 @@ public Image(Device device, Image srcImage, int flag) {
 					offset++;
 				}
 			}
-			init (newData);
+			init (newData, getZoom());
 			break;
 		}
 		case SWT.IMAGE_GRAY: {
@@ -412,7 +414,7 @@ public Image(Device device, Image srcImage, int flag) {
 					}
 				}
 			}
-			init (newData);
+			init (newData, getZoom());
 			break;
 		}
 		default:
@@ -459,8 +461,8 @@ public Image(Device device, Image srcImage, int flag) {
 public Image(Device device, Rectangle bounds) {
 	super(device);
 	if (bounds == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
-	currentNativeZoom = DPIUtil.getNativeDeviceZoom();
-	bounds = DPIUtil.autoScaleUp (bounds);
+	initialNativeZoom = DPIUtil.getNativeDeviceZoom();
+	bounds = DPIUtil.autoScaleUp (bounds, getZoom());
 	init(bounds.width, bounds.height);
 	init();
 }
@@ -491,10 +493,10 @@ public Image(Device device, Rectangle bounds) {
 public Image(Device device, ImageData data) {
 	super(device);
 	if (data == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
-	currentNativeZoom = DPIUtil.getNativeDeviceZoom();
+	initialNativeZoom = DPIUtil.getNativeDeviceZoom();
 	this.dataAtBaseZoom = new ElementAtZoom<>(data, 100);
 	data = DPIUtil.autoScaleUp(device, this.dataAtBaseZoom);
-	init(data);
+	init(data, getZoom());
 	init();
 }
 
@@ -535,12 +537,12 @@ public Image(Device device, ImageData source, ImageData mask) {
 	if (source.width != mask.width || source.height != mask.height) {
 		SWT.error(SWT.ERROR_INVALID_ARGUMENT);
 	}
-	currentNativeZoom = DPIUtil.getNativeDeviceZoom();
+	initialNativeZoom = DPIUtil.getNativeDeviceZoom();
 	this.dataAtBaseZoom = new ElementAtZoom<>(applyMask(source, ImageData.convertMask(mask)), 100);
 	source = DPIUtil.autoScaleUp(device, source);
 	mask = DPIUtil.autoScaleUp(device, mask);
 	mask = ImageData.convertMask(mask);
-	init(this.device, this, source, mask);
+	init(this.device, this, source, mask, getZoom());
 	init();
 }
 
@@ -599,10 +601,10 @@ public Image(Device device, ImageData source, ImageData mask) {
  */
 public Image (Device device, InputStream stream) {
 	super(device);
-	currentNativeZoom = DPIUtil.getNativeDeviceZoom();
+	initialNativeZoom = DPIUtil.getNativeDeviceZoom();
 	this.dataAtBaseZoom =  new ElementAtZoom<>(new ImageData (stream), 100);
 	ImageData data = DPIUtil.autoScaleUp(device, this.dataAtBaseZoom);
-	init(data);
+	init(data, getZoom());
 	init();
 }
 
@@ -641,10 +643,10 @@ public Image (Device device, InputStream stream) {
 public Image (Device device, String filename) {
 	super(device);
 	if (filename == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
-	currentNativeZoom = DPIUtil.getNativeDeviceZoom();
+	initialNativeZoom = DPIUtil.getNativeDeviceZoom();
 	this.dataAtBaseZoom = new ElementAtZoom<>(new ImageData (filename), 100);
 	ImageData data = DPIUtil.autoScaleUp(device, this.dataAtBaseZoom);
-	init(data);
+	init(data, getZoom());
 	init();
 }
 
@@ -680,14 +682,18 @@ public Image (Device device, String filename) {
 public Image(Device device, ImageFileNameProvider imageFileNameProvider) {
 	super(device);
 	this.imageFileNameProvider = imageFileNameProvider;
-	currentNativeZoom = DPIUtil.getNativeDeviceZoom();
+	initialNativeZoom = DPIUtil.getNativeDeviceZoom();
 	ElementAtZoom<String> fileName = DPIUtil.validateAndGetImagePathAtZoom (imageFileNameProvider, getZoom());
 	if (fileName.zoom() == getZoom()) {
-		initNative (fileName.element());
-		if (this.handle == 0) init(new ImageData (fileName.element()));
+		long handle = initNative (fileName.element(), getZoom());
+		if (handle == 0) {
+			init(new ImageData (fileName.element()), getZoom());
+		} else {
+			setHandleForZoomLevel(handle, getZoom());
+		}
 	} else {
 		ImageData resizedData = DPIUtil.autoScaleImageData (device, new ImageData (fileName.element()), fileName.zoom());
-		init(resizedData);
+		init(resizedData, getZoom());
 	}
 	init();
 }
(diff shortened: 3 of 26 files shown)


amartya4256:
@fedejeanne @HeikoKlare @akoch-yatta I have created the PR for Image. You can review it now.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/ImageList.java
```diff
@@ -22,17 +22,19 @@ public class ImageList {
 	long handle;
 	int style, refCount;
 	Image [] images;
+	final int currentDeviceZoom;
```
akoch-yatta, line 25: I think we should rename it to "zoom". That seems more consistent.

bundles/org.eclipse.swt/Eclipse SWT Drag and Drop/win32/org/eclipse/swt/dnd/DragSource.java
```diff
@@ -491,7 +491,7 @@ private void drag(Event dragEvent) {
 	hwndDrag = 0;
 	topControl = null;
 	if (image != null) {
-		imagelist = new ImageList(SWT.NONE);
+		imagelist = new ImageList(SWT.NONE, nativeZoom);
```
akoch-yatta, line 494: I think we need to pass DPIUtil.getZoomForAutoscaleProperty(nativeZoom) here

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -725,68 +731,62 @@ public Image(Device device, ImageDataProvider imageDataProvider) {
...
+	init (resizedData, getCurrentDeviceZoom());
 	init();
 }
 
 /**
  * Update zoom and refresh the Image based on the native zoom level, if required.
  *
- * @param nativeZoom native zoom in % of the monitor on which the image is painted
+ * @param image the image to get the handle of
+ * @param deviceZoomLevel native zoom in % of the monitor on which the image is painted
```
akoch-yatta, line 744: its not the native zoom anymore, is it?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2327,9 +2345,9 @@ public static Image win32_new(Device device, int type, long handle) {
  * @noreference This method is not intended to be referenced by clients.
  */
 public static Image win32_new(Image image, int targetZoom) {
-	if (targetZoom != image.getZoom()) {
-		image.handleDPIChange(targetZoom);
+	if (targetZoom != image.getCurrentDeviceZoom()) {
```
akoch-yatta, line 2348: win32_new looks like a pointless operation now. Is that still use somewhere and does it still makes sense? It doesn't change anything for the image, right?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2108,9 +2136,10 @@ private static ImageData applyMask(ImageData source, ImageData mask) {
 	return imageData;
 }
 
-void init(ImageData i) {
+
+void init(ImageData i, Integer zoomLevel) {
```
akoch-yatta, line 2140: As we call it zoom or nativeZoom anywhere else, we should call it zoom here as well

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2014,24 +2033,33 @@ else if (i.alphaData != null) {
...
 	}
 	return result;
 }
 
-static long [] init(Device device, Image image, ImageData source, ImageData mask) {
+private void setHandleForZoomLevel(long handle, int height, int width, Integer zoomLevel) {
+	if (this.handle == 0) {
+		this.handle = handle;	// Set handle for default zoom level
+	}
+	if (zoomLevel != null && handleMap.get(zoomLevel) == null) {
```
akoch-yatta, line 2055: Can't you just use handleMap.contains(zoomLevel) for the second check?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1809,7 +1828,7 @@ static ImageData directToDirect(ImageData src, int newDepth, PaletteData newPale
 	return img;
 }
 
-static long [] init(Device device, Image image, ImageData i) {
+static long [] init(Device device, Image image, ImageData i, Integer zoomLevel) {
```
akoch-yatta, line 1831: As we call it zoom or nativeZoom anywhere else, we should call it zoom here as well

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1699,7 +1717,8 @@ void init(int width, int height) {
 	}
 	type = SWT.BITMAP;
 	long hDC = device.internal_new_GC(null);
-	handle = OS.CreateCompatibleBitmap(hDC, width, height);
+	setHandleForZoomLevel(OS.CreateCompatibleBitmap(hDC, width, height), height, width, getCurrentDeviceZoom());
+	handleMap.put(getCurrentDeviceZoom(), handle);
```
akoch-yatta, line 1721: Isn't setHandleForZoomLevel adding it already to the handleMap?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -948,13 +949,19 @@ void initNative(String filename) {
...
 		}
 	}
+	return handle;
 }
 
 long [] createGdipImage() {
+	return createGdipImage(this.getCurrentDeviceZoom());
+}
+
+long [] createGdipImage(Integer zoomLevel) {
```
akoch-yatta, line 963: As we call it zoom or nativeZoom anywhere else, we should call it zoom here as well

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2259,20 +2288,9 @@ public void setBackground(Color color) {
...
-	if (oldZoom != newZoom) {
-		this.currentNativeZoom = newZoom;
-		// width and height are tied to the current device zoom
-		// they must be reset the the zoom factor changes
-		width = height = -1;
-	}
-}
-
-private int getZoom() {
+private int getCurrentDeviceZoom() {
```
akoch-yatta, line 2291: Let's revert the renaming back to getZoom()

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
amartya4256: In the image it is native zoom because we need that for internal_new_GC

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
amartya4256: Its called for setImage in all the handleDPI. Doesnt change anything. We can also re set the image instsad in the callback

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: But the param deviceZoomLevel  is not equal to the nativeZoom, but to DPIUtil.getZoomForAutoscaleProperty(nativeZoom), right?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
```diff
@@ -942,19 +942,25 @@ void createDisplay (DeviceData data) {
 }
 
 static long create32bitDIB (Image image) {
+	return create32bitDIB (image, DPIUtil.getDeviceZoom());
+}
+
+static long create32bitDIB (Image image, int nativeZoomLevel) {
+	int zoomLevel = DPIUtil.getZoomForAutoscaleProperty(nativeZoomLevel);
```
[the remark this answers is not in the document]
amartya4256: Aah I had removed this. Probably got reverted. Thanks for pointing out

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: Yes, because this call is now only creating a new handle internally. I think that method is not needed with your changes anymore. Resetting it in the callbacks should be enough then

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
amartya4256: Oh Sorry never mind I thought it was some other method. yes you are right. I'll update the jdoc comment

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
amartya4256: containsKey, yes good idea

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
amartya4256: yeah, a miss. Sorry

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/ImageList.java
```diff
@@ -22,17 +22,19 @@ public class ImageList {
 	long handle;
 	int style, refCount;
 	Image [] images;
+	final int zoom;
 
-public ImageList (int style) {
-	this (style, 32, 32);
+public ImageList (int style, int currentZoomLevel) {
```
akoch-yatta, line 27: Rename the parameter to zoom as well for consistency. See the other constructor as well

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
```diff
@@ -1960,7 +1965,7 @@ public int getIconDepth () {
 	};
 }
 
-ImageList getImageList (int style, int width, int height) {
+ImageList getImageList (int style, int width, int height, int currentZoomLevel) {
```
akoch-yatta, line 1968: Rename the parameter to zoom as well for consistency. See all the other image List methods as well

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -724,69 +730,63 @@ public Image(Device device, ImageFileNameProvider imageFileNameProvider) {
...
-	ImageData resizedData = DPIUtil.autoScaleImageData(device, data.element(), data.zoom());
-	init (resizedData);
+	ImageData resizedData = DPIUtil.autoScaleImageData(device, data.element(), getZoom(), data.zoom());
+	init (resizedData, getZoom());
 	init();
 }
 
 /**
  * Update zoom and refresh the Image based on the native zoom level, if required.
  *
```
akoch-yatta, line 742: * <b>IMPORTANT:</b> This method is <em>not</em> part of the public
 * API for <code>Image</code>. It is marked public only so that it
 * can be shared within the packages provided by SWT. It is not
 * available on all platforms, and should never be called from
 * application code.
 * </p>
 *
 * @param image the image to adapt for the provided zoom
 * @param targetZoom zoom in % of the standard resolution
 * @return an image object equal to the specified image scaled to the provided targetZoom

and

is missing in this comment

And remove the @since then

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -724,69 +730,63 @@ public Image(Device device, ImageFileNameProvider imageFileNameProvider) {
...
-				init(resizedData);
-				init ();
-				refreshed = true;
-			}
-			setCurrentNativeZoom(nativeZoom);
-		}
-	} else if (imageDataProvider != null) {
-		if (newZoom != oldZoom) {
-			ElementAtZoom<ImageData> data = DPIUtil.validateAndGetImageDataAtZoom (imageDataProvider, newZoom);
+public static Long win32_getHandle (Image image, int deviceZoomLevel) {
```
akoch-yatta, line 749: Lets rename deviceZoomLevel to zoom for consistency

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -724,69 +730,63 @@ public Image(Device device, ImageFileNameProvider imageFileNameProvider) {
...
+			image.dataAtBaseZoom = new ElementAtZoom<>(image.getImageData(image.getZoom()), image.getZoom());
+		}
+		if (image.dataAtBaseZoom != null) {
+			ImageData resizedData = image.getImageData(deviceZoomLevel);
+			image.init(resizedData, deviceZoomLevel);
+			image.init();
 		}
 	}
-	return refreshed;
+	return image.handleMap.get(deviceZoomLevel);
```
akoch-yatta, line 785: Is there a scenario, where this could be null?

amartya4256: Yes it is always checked in the block above and if not available, it is added.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/TreeColumn.java
```diff
@@ -536,7 +536,7 @@ void setImage (Image image, boolean sort, boolean right) {
 			hdItem.mask &= ~OS.HDI_IMAGE;
 			hdItem.fmt &= ~OS.HDF_IMAGE;
 			hdItem.fmt |= OS.HDF_BITMAP;
-			hdItem.hbm = image.handle;
+			hdItem.hbm = Image.win32_getHandle(image, parent.getZoom());
```
akoch-yatta, line 539: Can't we use getZoom() directly without the parent?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Table.java
```diff
@@ -7368,4 +7368,4 @@ private static void handleDPIChange(Widget widget, int newZoom, float scalingFac
 	 table.fixCheckboxImageListColor (true);
 	 table.settingItemHeight = false;
 }
-}
+}
```
akoch-yatta, line 7371: Can you remove that change?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
```diff
@@ -1984,13 +1989,13 @@ ImageList getImageList (int style, int width, int height) {
...
 
-	ImageList list = new ImageList (style, width, height);
+	ImageList list = new ImageList (style, width, height, zoom);
 	imageList [i] = list;
 	list.addRef();
 	return list;
 }
 
-ImageList getImageListToolBar (int style, int width, int height) {
+ImageList getImageListToolBar (int style, int width, int height, int currentZoomLevel) {
```
akoch-yatta, line 1998: Can you rename all _currentZoomLevel_ into _currentZoom_ or even _zoom_ here for consistency?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
```diff
@@ -942,19 +942,24 @@ void createDisplay (DeviceData data) {
 }
 
 static long create32bitDIB (Image image) {
+	return create32bitDIB (image, DPIUtil.getDeviceZoom());
+}
+
+static long create32bitDIB (Image image, int zoomLevel) {
```
akoch-yatta, line 948: Can you rename zoomLevel into zoom for consistency?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/ImageList.java
```diff
@@ -443,4 +445,4 @@ public int size () {
 	return result;
 }
 
-}
```
akoch-yatta, line 446: Please remove that change

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/ImageList.java
```diff
@@ -22,17 +22,19 @@ public class ImageList {
...
+	final int zoom;
 
-public ImageList (int style) {
-	this (style, 32, 32);
+public ImageList (int style, int zoom) {
+	this (style, 32, 32, zoom);
 }
 
-public ImageList (int style, int width, int height) {
+public ImageList (int style, int width, int height, int currentZoomLevel) {
```
akoch-yatta, line 31: Rename currentZoomLevel to zoom for consistency

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2308,28 +2331,4 @@ public static Image win32_new(Device device, int type, long handle) {
...
- * @noreference This method is not intended to be referenced by clients.
- */
-public static Image win32_new(Image image, int targetZoom) {
-	if (targetZoom != image.getZoom()) {
-		image.handleDPIChange(targetZoom);
-	}
-	return image;
-}
-}
+}
```
akoch-yatta, line 2334: Please remove that change on the last line

bundles/org.eclipse.swt/Eclipse SWT Drag and Drop/win32/org/eclipse/swt/dnd/DragSource.java
amartya4256: Done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/ImageList.java
amartya4256: Done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Table.java
amartya4256: Done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/ImageList.java
amartya4256: Done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
```diff
@@ -942,19 +942,24 @@ void createDisplay (DeviceData data) {
 }
 
 static long create32bitDIB (Image image) {
+	return create32bitDIB (image, DPIUtil.getDeviceZoom());
```
akoch-yatta, line 945: The method without zoom is not used anywhere anymore, right? We could remove that, couldn't we?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Button.java
```diff
@@ -1552,7 +1552,7 @@ private static void handleDPIChange(Widget widget, int newZoom, float scalingFac
 	}
 	// Refresh the image
 	if (button.image != null) {
-		button._setImage(Image.win32_new(button.image, newZoom));
+		button._setImage(button.image);
```
fedejeanne, line 1555: Just for my understanding: Why set it to "what was already the current value" (is my assumption correct?) instead of using the zoom? Does this mean that the button now only has the image in the original size (scaling factor of `1`) and the consumers should rescale it?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/TableItem.java
```diff
@@ -1276,14 +1276,6 @@ private static void handleDPIChange(Widget widget, int newZoom, float scalingFac
...
 		return;
 	}
-	Image[] images = tableItem.images;
-	if (images != null) {
-		for (Image innerImage : images) {
-			if (innerImage != null) {
-				Image.win32_new(innerImage, newZoom);
-			}
-		}
-	}
```
fedejeanne, line 1286: Where will be the inner images be notified about the change when this is removed?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/ToolItem.java
```diff
@@ -1231,24 +1231,13 @@ private static void handleDPIChange(Widget widget, int newZoom, float scalingFac
...
-		Image.win32_new(image, newZoom);
-
-		Image disabledImage = item.getDisabledImage();
-		if (disabledImage != null && !disabledImage.isDisposed()) {
-			Image.win32_new(disabledImage, newZoom);
-		}
-
-		Image hotImage = item.getHotImage();
-		if (hotImage != null) {
-			Image.win32_new(hotImage, newZoom);
```
fedejeanne, line 1251: Why is this code not necessary anymore?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/TreeItem.java
```diff
@@ -1821,14 +1821,6 @@ private static void handleDPIChange(Widget widget, int newZoom, float scalingFac
...
 		return;
 	}
-	Image[] images = treeItem.images;
-	if (images != null) {
-		for (Image innerImage : images) {
-			if (innerImage != null) {
-				Image.win32_new(innerImage, newZoom);
-			}
-		}
-	}
```
fedejeanne, line 1831: Where will these images be created then?

tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/graphics/ImageWin32Tests.java
```diff
@@ -37,19 +37,15 @@ public void setUp() {
 	}
 
 	@Test
-	public void imageMustBeRescaledOnZoomChange() {
+	public void imageShouldHaveDimesionAsPerZoomLevel() {
```
fedejeanne, line 40:
```suggestion
	public void testImageShouldHaveDimesionAsPerZoomLevel() {
```

tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/graphics/ImageWin32Tests.java
```diff
@@ -37,19 +37,15 @@ public void setUp() {
...
 		int zoom = DPIUtil.getDeviceZoom();
 		Image image = new Image(display, 10, 10);
 
 		try {
 			ImageData baseImageData = image.getImageData(zoom);
 			assertEquals("Width should equal the initial width on the same zoom", 10, baseImageData.width);
-			Image scaledImage = Image.win32_new(image, zoom);
-			ImageData scaledImageData = scaledImage.getImageData(zoom*2);
+			ImageData scaledImageData = image.getImageData(zoom*2);
```
fedejeanne, line 47: Extract the "magic number" `2` to a variable called `scalingFactor` or something similar.

fedejeanne (changes requested):
Thank you for this contribution! I have some questions (for my own understanding) and some minor improvement proposals.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/ImageList.java
```diff
@@ -46,7 +48,7 @@ public int add (Image image) {
 		index++;
 	}
 	if (count == 0) {
-		Rectangle rect = image.getBoundsInPixels();
+		Rectangle rect = DPIUtil.autoScaleBounds(image.getBounds(), zoom, 100);
```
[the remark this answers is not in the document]
amartya4256: Sounds reasonable. What do you say @akoch-yatta ?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/ImageList.java
```diff
@@ -22,17 +22,19 @@ public class ImageList {
 	long handle;
 	int style, refCount;
 	Image [] images;
+	final int zoom;
```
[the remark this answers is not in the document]
amartya4256: Yes, done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Button.java
amartya4256: the setImage method calls the OS at the low level to set the image again which has some callbacks. Since the zoom level has changed, on the callback the handle extracted by win32_getHandle method gets the handle for the specified zoom level from the imge object. And the zoomed handle of the image is set for the button.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/TableItem.java
amartya4256: Earlier, we were changing the handle of the image on the zoom level change. The method win32_new used to call handleDPIchange which used to set the new zoomed handle in the image. As of now, we have completely changed the way this all happens. Now we dont need to reset the handle of the image on the zoom event, but rather when the image is drawn, the appropriate handle is obtained using the win32_getHandle method where the client passes the current zoom. So, we don't need to do anything on the dpiChanged event but rather it is automatically zoomed in the drawing phase when the widget containing the image is redrawn after the dpi changed event. Since image is a resource, it can only be drawn by widgets or GC. And all the clients which consume the handle of the image now use win32_getHandle method instead to obtain the right handle.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/TreeItem.java
amartya4256: Its not creating the image. it is just calling the handleDpiChange method to set the right handle to the image as per the zoom level, and since the way we obtain the right handle of the image has changed as per the previous explanation, this can be removed.

tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/graphics/ImageWin32Tests.java
amartya4256: Done

tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/graphics/ImageWin32Tests.java
amartya4256: Done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
```diff
@@ -300,7 +301,7 @@ public Cursor(Device device, ImageData source, int hotspotX, int hotspotY) {
 		if (hMask == 0) SWT.error(SWT.ERROR_NO_HANDLES);
 	} else {
 		ImageData mask = source.getTransparencyMask();
-		long [] result = Image.init(this.device, null, source, mask);
+		long [] result = Image.init(this.device, null, source, mask, DPIUtil.getDeviceZoom());
```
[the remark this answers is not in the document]
amartya4256: In the beginning when we don't have any zoom info, we set it to this one.
But in this specific case, this zoom information is never utilized in the init method. because all this does is to set the handle for the zoom level in the map. but since the passed image is null here, and that block is never entered, it does nothing.

The most important thing is the imageData in this case. because the imageData is obtained using image.getImageData in the MenuSpyHandler and that gets the imageData at zoom level 100 (also in points basically). But overall the argument DPIUtil.getDeviceZoom() doesnt play any roll in this call here. We can also nullify it here.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
amartya4256: Removed

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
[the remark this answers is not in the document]
fedejeanne: If it doesn't play any role and it can be nullified then please do it and add a small comment in the code explaining that the parameter will be ignored anyway.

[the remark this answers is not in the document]
amartya4256: Hmm, I'll have to change the data type of the parameter to Integer in this case to make it nullable though. Is that okay?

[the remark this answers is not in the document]
amartya4256: Done

fedejeanne (approved):
Better, it works now. Thanks!

amartya4256:
Fixed the issue with the icon of the shell. But I have also created a follow up issue which addresses the problem in the depth but it doesn't have to be the part of this PR because it addresses the usability of a method outside its scope.
https://github.com/orgs/vi-eclipse/projects/1/views/1?filterQuery=assignee%3Aamartya4256+47&pane=issue&itemId=67443386
FYI @akoch-yatta @fedejeanne @HeikoKlare

PR #1239, opened 2024-05-17
Updating the control example by adding option to make Table and Tree Widgets editable

ShahzaibIbrahim:
Updating the example by adding option to make Table and Tree Widgets editable

This commit is part of testing the custom controls for multi-monitor support.

**HOW TO TEST**

* Run the _Runtime Workspace_ launch configuration
* Open the _SWT Controls_ view
* Switch to the tab "Tree"/"Table"
* Check the "Editable" option under "Other"
* Edit the content of the tree/table

Contributes to #62 and #127

### DIFF
--- a/examples/org.eclipse.swt.examples/src/org/eclipse/swt/examples/controlexample/TableTab.java
+++ b/examples/org.eclipse.swt.examples/src/org/eclipse/swt/examples/controlexample/TableTab.java
@@ -17,6 +17,7 @@ package org.eclipse.swt.examples.controlexample;
 import static org.eclipse.swt.events.SelectionListener.widgetSelectedAdapter;
 
 import org.eclipse.swt.SWT;
+import org.eclipse.swt.custom.TableEditor;
 import org.eclipse.swt.events.SelectionListener;
 import org.eclipse.swt.graphics.Color;
 import org.eclipse.swt.graphics.Font;
@@ -24,9 +25,11 @@ import org.eclipse.swt.graphics.FontData;
 import org.eclipse.swt.graphics.Image;
 import org.eclipse.swt.graphics.Point;
 import org.eclipse.swt.graphics.RGB;
+import org.eclipse.swt.graphics.Rectangle;
 import org.eclipse.swt.layout.GridData;
 import org.eclipse.swt.layout.GridLayout;
 import org.eclipse.swt.widgets.Button;
+import org.eclipse.swt.widgets.Control;
 import org.eclipse.swt.widgets.Event;
 import org.eclipse.swt.widgets.Group;
 import org.eclipse.swt.widgets.Item;
@@ -35,6 +38,7 @@ import org.eclipse.swt.widgets.MenuItem;
 import org.eclipse.swt.widgets.Table;
 import org.eclipse.swt.widgets.TableColumn;
 import org.eclipse.swt.widgets.TableItem;
+import org.eclipse.swt.widgets.Text;
 import org.eclipse.swt.widgets.Widget;
 
 class TableTab extends ScrollableTab {
@@ -49,7 +53,8 @@ class TableTab extends ScrollableTab {
 	Button noScrollButton, checkButton, fullSelectionButton, hideSelectionButton;
 
 	/* Other widgets added to the "Other" group */
-	Button multipleColumns, moveableColumns, resizableColumns, headerVisibleButton, sortIndicatorButton, headerImagesButton, linesVisibleButton, subImagesButton;
+	Button multipleColumns, moveableColumns, resizableColumns, headerVisibleButton, sortIndicatorButton,
+			headerImagesButton, linesVisibleButton, subImagesButton, editableButton;
 
 	/* Controls and resources added to the "Colors and Fonts" group */
 	static final int ITEM_FOREGROUND_COLOR = 3;
@@ -248,6 +253,8 @@ class TableTab extends ScrollableTab {
 		headerImagesButton.setText (ControlExample.getResourceString("Header_Images"));
 		subImagesButton = new Button (otherGroup, SWT.CHECK);
 		subImagesButton.setText (ControlExample.getResourceString("Sub_Images"));
+		editableButton = new Button(otherGroup, SWT.CHECK);
+		editableButton.setText(ControlExample.getResourceString("Editable"));
 
 		/* Add the listeners */
 		linesVisibleButton.addSelectionListener (widgetSelectedAdapter(event -> setWidgetLinesVisible ()));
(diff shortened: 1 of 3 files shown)


fedejeanne (changes requested):
@ShahzaibIbrahim thank you for this contribution!

It looks good, just one detail: you don't need to add the controls to the _SWT **Custom** Controls_ view since they are already in the _SWT Controls_ view.

Please base your PR on top of `master` instead of `akoch-master-33` to improve testability,

I changed the "How to test" in the description of this PR since manual testing is also testing

examples/org.eclipse.swt.examples/src/org/eclipse/swt/examples/controlexample/TreeTab.java
```diff
@@ -279,6 +389,7 @@ void createExampleGroup () {
 		imageTreeGroup.setLayoutData (new GridData (SWT.FILL, SWT.FILL, true, true));
 		imageTreeGroup.setText (ControlExample.getResourceString("Tree_With_Images"));
 	}
+    static int column = -1;
```
fedejeanne, line 392: Move up, to where the other fields are declared (before the constructor)

fedejeanne (changes requested):
Last detail and then this PR is OK

fedejeanne (approved):

ShahzaibIbrahim:
> Unfortunately, you have now applied some auto-formatting to the complete classes, so the commit is full of unrelated formatting changes. Please revert those formatting changes and only format the code that you have edited/added.

@HeikoKlare I have applied formatting now only on my changes also changed the commit message

examples/org.eclipse.swt.examples/src/org/eclipse/swt/examples/controlexample/TreeTab.java
```diff
@@ -258,6 +263,108 @@ void createOtherGroup () {
...
+	/**
+	 * Make Tree Editable
+	 */
+	void makeTreeContentEditable() {
+		// Tree 1
+		final TreeEditor editor = new TreeEditor(tree1);
+		editor.horizontalAlignment = SWT.LEFT;
+		editor.grabHorizontal = true;
+		// Add double click listener to the tree
+		tree1.addListener(SWT.MouseDoubleClick, new Listener() {
```
HeikoKlare, line 278: `Listener` is a functional interface, so you can use a lambda instead of providing an anonymous implementation of that interface.

examples/org.eclipse.swt.examples/src/org/eclipse/swt/examples/controlexample/TreeTab.java
```diff
@@ -258,6 +263,108 @@ void createOtherGroup () {
...
+						final String oldText = item.getText();
+
+						// Create a text field to edit the item text
+						final Text text = new Text(tree1, SWT.NONE);
+						text.setText(oldText);
+						text.selectAll();
+						text.setFocus();
+
+						// Add a focus out listener to commit changes on focus lost
+						text.addListener(SWT.FocusOut, new Listener() {
```
HeikoKlare, line 295: Like above: `Listener` is a functional interface, so you can use a lambda instead of providing an anonymous implementation of that interface.

examples/org.eclipse.swt.examples/src/org/eclipse/swt/examples/controlexample/TreeTab.java
```diff
@@ -258,6 +263,108 @@ void createOtherGroup () {
...
+	void makeTreeContentEditable() {
+		// Tree 1
+		final TreeEditor editor = new TreeEditor(tree1);
+		editor.horizontalAlignment = SWT.LEFT;
+		editor.grabHorizontal = true;
+		// Add double click listener to the tree
+		tree1.addListener(SWT.MouseDoubleClick, new Listener() {
+			@Override
+			public void handleEvent(Event event) {
+				if (editableButton.getSelection()) {
```
HeikoKlare, line 281: I would prefer an early return here instead of having the whole method indented.

examples/org.eclipse.swt.examples/src/org/eclipse/swt/examples/controlexample/TreeTab.java
```diff
@@ -258,6 +263,108 @@ void createOtherGroup () {
...
+								if (e.keyCode == SWT.CR || e.keyCode == SWT.KEYPAD_CR) {
+									item.setText(text.getText());
+									text.dispose(); // Dispose the text field after editing
+								}
+							}
+						});
+
+						// Edit the text field on double click
+						editor.setEditor(text, item);
+					}
```
HeikoKlare, line 364: This looks like the exact same code as in the handler for other tree. Can it be moved to a separate method that is reused in both cases?

examples/org.eclipse.swt.examples/src/org/eclipse/swt/examples/controlexample/TreeTab.java
```diff
@@ -258,6 +263,108 @@ void createOtherGroup () {
 		resizableColumns.addSelectionListener (widgetSelectedAdapter(event -> setColumnsResizable ()));
 		headerImagesButton.addSelectionListener (widgetSelectedAdapter(event -> recreateExampleWidgets ()));
 		subImagesButton.addSelectionListener (widgetSelectedAdapter(event -> recreateExampleWidgets ()));
+		editableButton.addSelectionListener(widgetSelectedAdapter(event -> makeTreeContentEditable()));
+	}
+
+	/**
+	 * Make Tree Editable
+	 */
```
HeikoKlare, line 271: Please remove this comment, as it just repeats the method's name.

examples/org.eclipse.swt.examples/src/org/eclipse/swt/examples/controlexample/TableTab.java
```diff
@@ -258,6 +271,75 @@ void createOtherGroup () {
 		resizableColumns.addSelectionListener (widgetSelectedAdapter(event -> setColumnsResizable ()));
 		headerImagesButton.addSelectionListener (widgetSelectedAdapter(event -> recreateExampleWidgets ()));
 		subImagesButton.addSelectionListener (widgetSelectedAdapter(event -> recreateExampleWidgets ()));
+		editableButton.addSelectionListener(widgetSelectedAdapter(event -> makeTableContentEditable()));
+	}
+
+	/**
+	 * Make Tree Editable
+	 */
+	void makeTableContentEditable() {
```
HeikoKlare, line 280: Please transfer the comments I made for the `TreeTab` to this class as well (if they apply here). In addition, the comment is inconsistent (probably due to copy/paste), but since it is obsolete, you can simply remove it.

examples/org.eclipse.swt.examples/src/org/eclipse/swt/examples/controlexample/TableTab.java
```diff
@@ -60,9 +68,12 @@ class TableTab extends ScrollableTab {
 	static final int CELL_FONT = 8;
 	static final int HEADER_FOREGROUND_COLOR = 9;
 	static final int HEADER_BACKGROUND_COLOR = 10;
-	Color itemForegroundColor, itemBackgroundColor, cellForegroundColor, cellBackgroundColor, headerForegroundColor, headerBackgroundColor;
+	Color itemForegroundColor, itemBackgroundColor, cellForegroundColor, cellBackgroundColor, headerForegroundColor,
+			headerBackgroundColor;
 	Font itemFont, cellFont;
 
+	static int column = -1;
```
HeikoKlare, line 75: Why is this static field necessary? The value does only seem to be used in the mouse click handler, but there is no need to store it across multiple invocations of that handler. In case there is reason for storing this state across multiple method invocations, why does it need to be static?

examples/org.eclipse.swt.examples/src/org/eclipse/swt/examples/controlexample/TableTab.java
```diff
@@ -258,6 +271,75 @@ void createOtherGroup () {
...
+		editor.horizontalAlignment = SWT.LEFT;
+		editor.grabHorizontal = true;
+		// Listener to activate editor
+		table1.addMouseListener(new MouseAdapter() {
+			@Override
+			public void mouseDoubleClick(MouseEvent event) {
+				if (editableButton.getSelection()) {
+					Control oldEditor = editor.getEditor();
+					if (oldEditor != null)
+						oldEditor.dispose();
```
HeikoKlare, line 292: Please always put if/do/while/for statements into blocks.

HeikoKlare (commented):
I have some further comments on code complexity and code style.

examples/org.eclipse.swt.examples/src/org/eclipse/swt/examples/controlexample/TableTab.java
HeikoKlare: @ShahzaibIbrahim I still do not see any need to this field. Can you please explain why it is necessary?

ShahzaibIbrahim: @HeikoKlare We use this column field to identify which column is selected for editing and then add the text. See the line 307 to 311.

HeikoKlare: But isn't that value calculated anew everytime the listener is called (in lines 294 to 300)? Then why store it in a field?

examples/org.eclipse.swt.examples/src/org/eclipse/swt/examples/controlexample/TreeTab.java
```diff
@@ -258,6 +262,64 @@ void createOtherGroup () {
...
+		});
+
+		// Tree 2
+		final TreeEditor editor2 = new TreeEditor(tree2);
+		editor2.horizontalAlignment = SWT.LEFT;
+		editor2.grabHorizontal = true;
+
+		tree2.addListener(SWT.MouseDoubleClick, event -> {
+			treeDoubleClickListener(editor2, event);
+		});
```
HeikoKlare, line 286: Also, the configuration of the tree editor and adding the listener is repeated code, which could be combined in separate `makeTreeEditable(Tree)` method.

examples/org.eclipse.swt.examples/src/org/eclipse/swt/examples/controlexample/TableTab.java
HeikoKlare: My question is why the value needs to be stored outside the lambda at all. Why not simply define a variable `int column = -1` at the beginning of the lamda? What is the scenario in which the value stored in the previous execution of the lambda needs to be available in a subsequent execution of it?

ShahzaibIbrahim: I tried that too but again I needed to use this field into another lambda (modify listener) inside (308 to 310) and I ran to the same problem again.

ShahzaibIbrahim: For Context:

image

HeikoKlare: You can still assign the calculated value to a local final variable that is accessed from within the lambda.

To make it more clean, you can even extract the logic for finding the column containing the point into a separate method and assign the result to a local final variable, i.e., replace lines 294 to 300 with:
```java
final int column = findColumnContainingPoint(item, point);
```
and add an according method:
```java
	private static int findColumnContainingPoint(TableItem item, Point point) {
		for (int i = 0; i < item.getParent().getColumnCount(); i++) {
			Rectangle rect = item.getBounds(i);
			if (rect.contains(point)) {
				return i;
			}
		}
		return -1;
	}
```

HeikoKlare (approved):
Thank you for the contribution and for addressing my comments! The code looks good now. And the changes still work as expected.

HeikoKlare:
Merging, since freeze period check only fails because of M1 promotion calendar entry, but merging is actually allowed (and the change does not affect production code anyway).

Issue #843, opened 2023-10-26
Internal Browser not working on Linux


Phillipus:
**Describe the bug**
The internal Browser is not rendering on Linux.

**To Reproduce**
Install latest Eclipse on Ubuntu 23.04 (X11 not Wayland) and do one of the following:
- Open up the internal Browser and try to view a URL
- Open the Welcome page to try to view the content
- Hover over some Java to invoke the tooltip
- Run a basic SWT Snippet to launch a Browser component and set the URL

Either the browser is blank, or sometimes using the Internal Browser View in Eclipse you might see a little of the content. It also seems to be hanging because there is a long delay closing Eclipse.

**Expected behavior**
Browser component should render on Linux.

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [ ] Windows
- - [x] Linux
- - [ ] macOS 

2. Additional OS info (e.g. OS version, Linux Desktop, etc)

Ubuntu 23.04
GTK 3.24.37

or:

Linux Mint 21.2
GTK 3.24.33

5. JRE/JDK version

Temurin-17.0.8+7 (build 17.0.8+7)
 
**Version since**
Tested on Eclipse 4.27 and later

**Workaround (or) Additional context**
None

akurtakov:
Do you say it works with Wayland for you?
What is the webkitgtk version you have ? On fedora 39 with webkit2gtk4.1-2.42.1 internal web browser works just fine with both x11 and wayland for me.

Phillipus:
> Do you say it works with Wayland for you?

I've now tested it with Wayland on Ubuntu and it works. But not working with X11.

I can't test Wayland on Linux Mint because that distro only supports X11.

> What is the webkitgtk version you have

On Ubuntu I see these versions:

- libwebkit2gtk-4.0-37
- libwebkit2gtk-4.1-0
- libwebkitgtk-6.0-4

iloveeclipse:
And which webkit is loaded by SWT?

akurtakov:
> > What is the webkitgtk version you have
> 
> On Ubuntu I see these versions:
> 
>     * libwebkit2gtk-4.0-37
>     * libwebkit2gtk-4.1-0
>     * libwebkitgtk-6.0-4

[...]

PR #1242, opened 2024-05-21
Update bug_report.md for Linux users

iloveeclipse:
Added note for Linux users about crash reporting/webkit

See https://github.com/eclipse-platform/eclipse.platform.swt/issues/843


HannesWell:
Alternatively or additionally https://github.com/eclipse-platform/eclipse.platform.swt/issues/843 could be pinned to get more attention if one visit's SWT's issues.

PR #1243, opened 2024-05-21
Implementation of FontMetrics to enable multi zoom level support for win32

amartya4256:
## Addressed issues

## Requires

## Unlocks
* https://github.com/eclipse-platform/eclipse.platform.swt/pull/1244

## Description

This pull request is based on the implementations of PR #1214. It provides the native zoom to the font metrics which can further be utilized by GC and Text Layout using Public APIs, since fonts are scaled to native zoom. (based on PR #1214 ).

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/FontMetrics.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/FontMetrics.java
@@ -43,6 +43,8 @@ public final class FontMetrics {
 	 */
 	public TEXTMETRIC handle;
 
+	private int nativeZoom;
+
 /**
  * Prevents instances from being created outside the package.
  */
@@ -95,7 +97,7 @@ public boolean equals (Object object) {
  * @return the ascent of the font
  */
 public int getAscent() {
-	return DPIUtil.autoScaleDown(handle.tmAscent - handle.tmInternalLeading);
+	return DPIUtil.scaleDown(handle.tmAscent - handle.tmInternalLeading, getZoom());
 }
 
 /**
@@ -118,7 +120,7 @@ public double getAverageCharacterWidth() {
  */
 @Deprecated
 public int getAverageCharWidth() {
-	return DPIUtil.autoScaleDown(handle.tmAveCharWidth);
+	return DPIUtil.scaleDown(handle.tmAveCharWidth, getZoom());
 }
 
 /**
@@ -130,7 +132,7 @@ public int getAverageCharWidth() {
  * @return the descent of the font
  */
 public int getDescent() {
-	return DPIUtil.autoScaleDown(handle.tmDescent);
+	return DPIUtil.scaleDown(handle.tmDescent, getZoom());
 }
 
 /**
@@ -145,7 +147,7 @@ public int getDescent() {
  * @see #getLeading
  */
 public int getHeight() {
-	return DPIUtil.autoScaleDown(handle.tmHeight);
+	return DPIUtil.scaleDown(handle.tmHeight, getZoom());
 }
 
 /**
@@ -170,6 +172,10 @@ public int getLeading() {
 	 * value here:
 	 */
 	return getHeight() - getAscent() - getDescent();
+}
+
+private int getZoom() {
+	return DPIUtil.getZoomForAutoscaleProperty(nativeZoom);
 }
 
 /**
@@ -211,6 +217,29 @@ public int hashCode() {
 public static FontMetrics win32_new(TEXTMETRIC handle) {
 	FontMetrics fontMetrics = new FontMetrics();
 	fontMetrics.handle = handle;
+	return fontMetrics;
+}
+
+/**
+ * Invokes platform specific functionality to allocate a new font metrics.
+ * <p>
+ * <b>IMPORTANT:</b> This method is <em>not</em> part of the public
+ * API for <code>FontMetrics</code>. It is marked public only so that
+ * it can be shared within the packages provided by SWT. It is not
+ * available on all platforms, and should never be called from
+ * application code.
+ * </p>
+ *
+ * @param handle the <code>TEXTMETRIC</code> containing information about a font
+ * @param nativeZoom the native zoom of the monitor for which Font Metrics is created
+ * @return a new font metrics object containing the specified <code>TEXTMETRIC</code>
+ *
+ * @noreference This method is not intended to be referenced by clients.
+ */
+public static FontMetrics win32_new(TEXTMETRIC handle, int nativeZoom) {
+	FontMetrics fontMetrics = new FontMetrics();
+	fontMetrics.handle = handle;
+	fontMetrics.nativeZoom = nativeZoom;
 	return fontMetrics;
 }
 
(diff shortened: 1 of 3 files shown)


amartya4256:
@fedejeanne @HeikoKlare @akoch-yatta You can review this now.

fedejeanne (commented):
Changes look OK so far

fedejeanne:
> Changes look OK so far

@amartya4256 build is failing though

merks:
The build says "Killed" which is a problem that is driving me crazy in other projects too.:
```
05:15:00  /home/jenkins/agent/workspace/eclipse.platform.swt_PR-1243/eclipse.platform.swt@tmp/durable-eb24fb20/script.sh.copy: line 2:   422 Killed                  mvn clean verify --batch-mode --threads 1C -V -U -e -DforkCount=0 -Papi-check -Dcompare-version-with-baselines.skip=false -Dorg.eclipse.swt.tests.junit.disable.test_isLocal=true -Dmaven.test.failure.ignore=true -Dmaven.test.error.ignore=true
```
I'll start a new build and hope for the best.

merks:
This time the buidl was good...

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/FontMetrics.java
```diff
@@ -208,10 +214,11 @@ public int hashCode() {
  *
  * @noreference This method is not intended to be referenced by clients.
  */
-public static FontMetrics win32_new(TEXTMETRIC handle) {
+public static FontMetrics win32_new(TEXTMETRIC handle, int nativeZoom) {
```
akoch-yatta, line 217: We probably cannot directly extend that? I would propose to create a new method, delegate from the old one with DPIUtil.getNativeDeviceZoom() and mark the old one as deprecated. What do you think @HeikoKlare?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/TextLayout.java
```diff
@@ -2121,7 +2121,7 @@ public FontMetrics getLineMetrics (int lineIndex) {
 	lptm.tmHeight = DPIUtil.autoScaleUp(getDevice(), ascentInPoints + descentInPoints);
 	lptm.tmInternalLeading = DPIUtil.autoScaleUp(getDevice(), leadingInPoints);
 	lptm.tmAveCharWidth = 0;
-	return FontMetrics.win32_new(lptm);
+	return FontMetrics.win32_new(lptm, DPIUtil.getNativeDeviceZoom());
```
HeikoKlare, line 2124: This looks wrong to me: why is it valid to use the zoom from non-multi-monitor-aware `DPIUtil`? Won't this fail in the multi-monitor use case?

fedejeanne:
FYI the newly added API violation...

```
[Check warning on line 2124 in eclipse.platform.swt/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/TextLayout.java](https://github.com/eclipse-platform/eclipse.platform.swt/pull/1243/files#annotation_22593304264)
```

... can be ignored since the method has been deprecated in this PR but its usage is removed in #1244

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/FontMetrics.java
```diff
@@ -194,6 +200,7 @@ public int hashCode() {
 }
 
 /**
+ * * @deprecated Use {@link #win32_new(TEXTMETRIC, int)} instead.
```
fedejeanne, line 203: Put it at the end and remove the extra `*`

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/FontMetrics.java
```diff
@@ -208,10 +215,34 @@ public int hashCode() {
  *
  * @noreference This method is not intended to be referenced by clients.
  */
+@Deprecated
```
fedejeanne, line 218:
```suggestion
@Deprecated(forRemoval = true, since = "2024-09")
```

fedejeanne (changes requested):
Please improve the javadoc and push again so the tests will run. I see some failed checks, let's see if they go away ;)

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/FontMetrics.java
akurtakov: There is no need to deprecate methods marked with noreference as it's platform specific and thus not part of SWT API, feel free to remove it directly.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/FontMetrics.java
akurtakov: There is no need to deprecate methods marked with noreference as it's platform specific and thus not part of SWT API, feel free to remove it directly.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/FontMetrics.java
amartya4256: Done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/FontMetrics.java
amartya4256: Done

fedejeanne: @akurtakov would it be wrong to mark it as deprecated for removal? Please don't get me wrong, I prefer to remove it too but deprecating was also an option (https://github.com/eclipse-platform/eclipse.platform.swt/pull/1243/files#r1609580640)

akurtakov: This method doesn't exist in Cocoa and Gtk port thus deprecating it will not buy you/us anything. While technically not wrong to mark unused "private"(implementation specific) method it just leaves us with garbage in the codebase that too often people forget to collect later.

akurtakov: It's worth mentioning that javadoc published in doc bundles and help center are built on Linux thus this deprecation message and notification will not end up in "user visible" location.

fedejeanne (approved):
I tested the code and it's looking good, I couldn't find any regressions.

As soon as the checks run green and [this matter](https://github.com/eclipse-platform/eclipse.platform.swt/pull/1243#discussion_r1638056010) is resolved, this PR is good to go.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/FontMetrics.java
fedejeanne: Ok, so I guess it's no use to deprecate it. Thank you for clarifying this! @amartya4256 you may remove the deprecated method and push your changes :)

fedejeanne: @akurtakov fyi we decided to remove the deprecation notice and leave the method because the only remaining usage is going to be removed in #1244 anyway and the method will be removed there.

PR #1247, opened 2024-05-22
Implementation of Subclasses of Widgets to enable multi zoom level support for win32

amartya4256:
## Addressed issues

## Requires

## Description

This pull request is based on the implementations of PR #1214. It uses the native zoom of widget to do the scaling using the DPIUtil methods with zoom level. The Event class also uses DPIUtil in a few methods which are used by the win32 widget classes. This contribution adds the reference of the zoom level fom the widget from those calls in these methods within Event.

The test assertion is purposely commented out since the access to the widget zoom is no longer available outside the package. We might want to create a test fragment for the package itself so that its easy to test package protected members.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/WidgetWin32Tests.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/WidgetWin32Tests.java
@@ -1,0 +1,25 @@
+package org.eclipse.swt.widgets;
+
+import static org.junit.Assert.assertEquals;
+
+import org.eclipse.swt.*;
+import org.eclipse.swt.internal.*;
+import org.junit.*;
+
+public class WidgetWin32Tests extends Win32AutoscaleTestBase {
+
+	@Test
+	public void testWidgetZoomShouldChangeOnZoomLevelChange() {
+		int zoom = DPIUtil.getDeviceZoom();
+		int scaledZoom = zoom * 2;
+
+		Button button = new Button(shell, SWT.PUSH);
+		button.setBounds(0, 0, 200, 50);
+		button.setText("Widget Test");
+		button.setBackground(shell.getDisplay().getSystemColor(SWT.COLOR_CYAN));
+		shell.open();
+		assertEquals("The initial zoom is wrong", zoom, button.getZoom()); //pre-condition
+		changeDPIZoom(scaledZoom);
+		assertEquals("The Zoom Level should be updated for button on zoom change event on its shell", scaledZoom, button.getZoom());
+	}
+}
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/CoolItem.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/CoolItem.java
@@ -185,9 +185,10 @@ protected void checkSubclass () {
  */
 public Point computeSize (int wHint, int hHint) {
 	checkWidget ();
-	wHint = (wHint != SWT.DEFAULT ? DPIUtil.autoScaleUp(wHint) : wHint);
-	hHint = (hHint != SWT.DEFAULT ? DPIUtil.autoScaleUp(hHint) : hHint);
-	return DPIUtil.autoScaleDown(computeSizeInPixels(wHint, hHint));
+	int zoom = getZoom();
+	wHint = (wHint != SWT.DEFAULT ? DPIUtil.autoScaleUp(wHint, zoom) : wHint);
+	hHint = (hHint != SWT.DEFAULT ? DPIUtil.autoScaleUp(hHint, zoom) : hHint);
+	return DPIUtil.scaleDown(computeSizeInPixels(wHint, hHint), zoom);
 }
 Point computeSizeInPixels (int wHint, int hHint) {
 	int index = parent.indexOf (this);
@@ -222,7 +223,7 @@ void destroyWidget () {
  */
 public Rectangle getBounds () {
 	checkWidget ();
-	return DPIUtil.autoScaleDown(getBoundsInPixels());
+	return DPIUtil.scaleDown(getBoundsInPixels(), getZoom());
 }
 
 Rectangle getBoundsInPixels () {
@@ -382,7 +383,7 @@ public void setControl (Control control) {
  */
 public Point getPreferredSize () {
 	checkWidget ();
-	return DPIUtil.autoScaleDown(getPreferredSizeInPixels());
+	return DPIUtil.scaleDown(getPreferredSizeInPixels(), getZoom());
 }
 
 Point getPreferredSizeInPixels () {
@@ -413,7 +414,8 @@ Point getPreferredSizeInPixels () {
  */
 public void setPreferredSize (int width, int height) {
 	checkWidget ();
-	setPreferredSizeInPixels(DPIUtil.autoScaleUp(width), DPIUtil.autoScaleUp(height));
+	int zoom = getZoom();
+	setPreferredSizeInPixels(DPIUtil.autoScaleUp(width, zoom), DPIUtil.autoScaleUp(height, zoom));
 }
 
 void setPreferredSizeInPixels (int width, int height) {
@@ -462,7 +464,7 @@ void setPreferredSizeInPixels (int width, int height) {
 public void setPreferredSize (Point size) {
 	checkWidget ();
 	if (size == null) error(SWT.ERROR_NULL_ARGUMENT);
-	size = DPIUtil.autoScaleUp(size);
+	size = DPIUtil.autoScaleUp(size, getZoom());
 	setPreferredSizeInPixels(size.x, size.y);
 }
 
@@ -481,7 +483,7 @@ public void setPreferredSize (Point size) {
  */
 public Point getSize () {
 	checkWidget ();
-	return DPIUtil.autoScaleDown(getSizeInPixels());
+	return DPIUtil.scaleDown(getSizeInPixels(), getZoom());
 }
 
 Point getSizeInPixels() {
@@ -523,7 +525,8 @@ Point getSizeInPixels() {
  */
 public void setSize (int width, int height) {
 	checkWidget ();
-	setSizeInPixels(DPIUtil.autoScaleUp(width), DPIUtil.autoScaleUp(height));
+	int zoom = getZoom();
+	setSizeInPixels(DPIUtil.autoScaleUp(width, zoom), DPIUtil.autoScaleUp(height, zoom));
 }
 
 void setSizeInPixels (int width, int height) {
@@ -589,7 +592,7 @@ void setSizeInPixels (int width, int height) {
 public void setSize (Point size) {
 	checkWidget ();
 	if (size == null) error(SWT.ERROR_NULL_ARGUMENT);
-	size = DPIUtil.autoScaleUp(size);
+	size = DPIUtil.autoScaleUp(size, getZoom());
 	setSizeInPixels(size.x, size.y);
 }
 
@@ -608,7 +611,7 @@ public void setSize (Point size) {
  */
 public Point getMinimumSize () {
 	checkWidget ();
-	return DPIUtil.autoScaleDown(getMinimumSizeInPixels());
+	return DPIUtil.scaleDown(getMinimumSizeInPixels(), getZoom());
 }
 
 Point getMinimumSizeInPixels () {
@@ -641,7 +644,8 @@ Point getMinimumSizeInPixels () {
  */
 public void setMinimumSize (int width, int height) {
 	checkWidget ();
-	setMinimumSizeInPixels(DPIUtil.autoScaleUp(width), DPIUtil.autoScaleUp(height));
+	int zoom = getZoom();
+	setMinimumSizeInPixels(DPIUtil.autoScaleUp(width, zoom), DPIUtil.autoScaleUp(height, zoom));
 }
 
 void setMinimumSizeInPixels (int width, int height) {
@@ -691,7 +695,7 @@ void setMinimumSizeInPixels (int width, int height) {
 public void setMinimumSize (Point size) {
 	checkWidget ();
 	if (size == null) error (SWT.ERROR_NULL_ARGUMENT);
-	size = DPIUtil.autoScaleUp(size);
+	size = DPIUtil.autoScaleUp(size, getZoom());
 	setMinimumSizeInPixels(size.x, size.y);
 }
 
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
@@ -2944,8 +2944,9 @@ boolean isValidThread () {
 public Point map (Control from, Control to, Point point) {
 	checkDevice ();
 	if (point == null) error (SWT.ERROR_NULL_ARGUMENT);
-	point = DPIUtil.autoScaleUp(point);
-	return DPIUtil.autoScaleDown(mapInPixels(from, to, point));
+	int zoom = getZoomLevelForMapping(from, to);
+	point = DPIUtil.autoScaleUp(point, zoom);
+	return DPIUtil.scaleDown(mapInPixels(from, to, point), zoom);
 }
 
 Point mapInPixels (Control from, Control to, Point point) {
@@ -2990,9 +2991,10 @@ Point mapInPixels (Control from, Control to, Point point) {
  */
 public Point map (Control from, Control to, int x, int y) {
 	checkDevice ();
-	x = DPIUtil.autoScaleUp(x);
-	y = DPIUtil.autoScaleUp(y);
-	return DPIUtil.autoScaleDown(mapInPixels(from, to, x, y));
+	int zoom = getZoomLevelForMapping(from, to);
+	x = DPIUtil.autoScaleUp(x, zoom);
+	y = DPIUtil.autoScaleUp(y, zoom);
+	return DPIUtil.scaleDown(mapInPixels(from, to, x, y), zoom);
 }
 
 Point mapInPixels (Control from, Control to, int x, int y) {
@@ -3006,6 +3008,15 @@ Point mapInPixels (Control from, Control to, int x, int y) {
 	point.y = y;
 	OS.MapWindowPoints (hwndFrom, hwndTo, point, 1);
 	return new Point (point.x, point.y);
+}
+
+private int getZoomLevelForMapping(Control from, Control to) {
+	if (from != null && from.isDisposed()) error (SWT.ERROR_INVALID_ARGUMENT);
+	if (to != null && to.isDisposed()) error (SWT.ERROR_INVALID_ARGUMENT);
+	if (to != null) {
+		return to.getZoom();
+	}
+	return from.getZoom();
 }
 
 /**
@@ -3047,8 +3058,9 @@ Point mapInPixels (Control from, Control to, int x, int y) {
 public Rectangle map (Control from, Control to, Rectangle rectangle) {
 	checkDevice ();
 	if (rectangle == null) error (SWT.ERROR_NULL_ARGUMENT);
-	rectangle = DPIUtil.autoScaleUp(rectangle);
-	return DPIUtil.autoScaleDown(mapInPixels(from, to, rectangle));
+	int zoom = getZoomLevelForMapping(from, to);
+	rectangle = DPIUtil.autoScaleUp(rectangle, zoom);
+	return DPIUtil.scaleDown(mapInPixels(from, to, rectangle), zoom);
 }
 
 Rectangle mapInPixels (Control from, Control to, Rectangle rectangle) {
@@ -3095,11 +3107,12 @@ Rectangle mapInPixels (Control from, Control to, Rectangle rectangle) {
  */
 public Rectangle map (Control from, Control to, int x, int y, int width, int height) {
 	checkDevice ();
-	x = DPIUtil.autoScaleUp(x);
-	y = DPIUtil.autoScaleUp(y);
-	width = DPIUtil.autoScaleUp(width);
-	height = DPIUtil.autoScaleUp(height);
-	return DPIUtil.autoScaleDown(mapInPixels(from, to, x, y, width, height));
+	int zoom = getZoomLevelForMapping(from, to);
+	x = DPIUtil.autoScaleUp(x, zoom);
+	y = DPIUtil.autoScaleUp(y, zoom);
+	width = DPIUtil.autoScaleUp(width, zoom);
+	height = DPIUtil.autoScaleUp(height, zoom);
+	return DPIUtil.scaleDown(mapInPixels(from, to, x, y, width, height), zoom);
 }
 
 Rectangle mapInPixels (Control from, Control to, int x, int y, int width, int height) {
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Label.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Label.java
@@ -553,7 +553,8 @@ void wmDrawChildImage(DRAWITEMSTRUCT struct) {
 	int height = struct.bottom - struct.top;
 	if (width == 0 || height == 0) return;
 
-	Rectangle imageRect = DPIUtil.autoScaleBounds(image.getBounds(), getZoom(), 100);
+	int zoom = getZoom();
+	Rectangle imageRect = DPIUtil.autoScaleBounds(image.getBounds(), zoom, 100);
 
 	int x = 0;
 	if ((style & SWT.CENTER) != 0) {
@@ -566,7 +567,7 @@ void wmDrawChildImage(DRAWITEMSTRUCT struct) {
 	data.device = display;
 	GC gc = createNewGC(struct.hDC, data);
 	Image image = getEnabled () ? this.image : new Image (display, this.image, SWT.IMAGE_DISABLE);
-	gc.drawImage (image, DPIUtil.autoScaleDown(x), DPIUtil.autoScaleDown(Math.max (0, (height - imageRect.height) / 2)));
+	gc.drawImage (image, DPIUtil.scaleDown(x, zoom), DPIUtil.scaleDown(Math.max (0, (height - imageRect.height) / 2), zoom));
 	if (image != this.image) image.dispose ();
 	gc.dispose ();
 }
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
@@ -418,6 +418,7 @@ LRESULT CDDS_ITEMPOSTPAINT (NMTVCUSTOMDRAW nmcd, long wParam, long lParam) {
 	if ((style & SWT.FULL_SELECTION) != 0 || findImageControl () != null || ignoreDrawSelection || explorerTheme) {
 		OS.SetBkMode (hDC, OS.TRANSPARENT);
 	}
+	int zoom = getZoom();
 	boolean selected = isItemSelected (nmcd);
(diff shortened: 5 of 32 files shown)


amartya4256:
@HeikoKlare @fedejeanne @akoch-yatta Please have a look.

tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/graphics/WidgetWin32Tests.java
```diff
@@ -0,0 +1,73 @@
...
+		Shell shell = new Shell(display);
+
+		Button button = new Button(shell, SWT.PUSH);
+		button.setBounds(0, 0, 200, 50);
+		button.setText("Widget Test");
+		button.setBackground(display.getSystemColor(SWT.COLOR_CYAN));
+		shell.open();
+		changeZoomLevel(scaledZoom, shell);
+		// TODO: Zoom level not accessible from this test, Would be fixed later.
+//		assertEquals("The Zoom Level should be updated for button on zoom change event on its shell", scaledZoom, button.getZoom());
```
HeikoKlare, line 49: Please fix this :-)

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Widget.java
```diff
@@ -1232,6 +1232,7 @@ boolean sendMouseEvent (int type, int button, long hwnd, long lParam) {
 boolean sendMouseEvent (int type, int button, int count, int detail, boolean send, long hwnd, long lParam) {
 	if (!hooks (type) && !filters (type)) return true;
 	Event event = new Event ();
+	event.widget = this;
```
HeikoKlare, line 1235: This change seems to be unrelated to this PR. Is this some independent bug? Can we extract that change?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/TreeItem.java
```diff
@@ -375,7 +375,7 @@ public Color getBackground (int index) {
  */
 public Rectangle getBounds () {
 	checkWidget ();
-	return DPIUtil.autoScaleDown(getBoundsInPixels());
+	return DPIUtil.scaleDown(getBoundsInPixels(), parent.getZoom());
```
HeikoKlare, line 378: Why is `getZoom()` called on the parent at some places (e.g., `TreeItem`, `TreeColum`, `Tracker`, ...) instead of just calling `getZoom()` on the widget iself?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
```diff
@@ -7928,8 +7928,9 @@ LRESULT wmNotifyHeader (NMHDR hdr, long wParam, long lParam) {
 							GC gc = getNewGC(nmcd.hdc, data);
 							Rectangle imageBounds = DPIUtil.autoScaleBounds(columns[i].image.getBounds(), this.getZoom(), 100);
 							int y = Math.max (0, (nmcd.bottom - imageBounds.height) / 2);
-							gc.drawImage (columns[i].image, DPIUtil.autoScaleDown(x), DPIUtil.autoScaleDown(y));
-							x += imageBounds.width + 12;
+							Rectangle scaledBounds = DPIUtil.autoScaleBounds(columns[i].image.getBounds(), getZoom(), 100);
```
HeikoKlare, line 7931: I do not understand the necessity for the `scaledBounds`. Aren't they calculated in exactly the same way in which `imageBounds` is calculated?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
```diff
@@ -757,24 +757,24 @@ LRESULT CDDS_ITEMPOSTPAINT (NMTVCUSTOMDRAW nmcd, long wParam, long lParam) {
...
 							//int y1 = rect.top + (index == 0 ? (getItemHeight () - size.y) / 2 : 0);
-							int y1 = rect.top + DPIUtil.autoScaleUp((getItemHeight () - size.y) / 2);
+							int y1 = rect.top + DPIUtil.autoScaleUp((getItemHeight () - size.y) / 2, getZoom());
 							int x1 = Math.max (rect.left, rect.left - inset + 1);
 							GCData data = new GCData();
 							data.device = display;
-							GC gc = getNewGC(hDC, data);
-							gc.setClipping (DPIUtil.autoScaleDown(new Rectangle(x1, rect.top, rect.right - x1, rect.bottom - rect.top)));
-							gc.drawImage (image, 0, 0, bounds.width, bounds.height, DPIUtil.autoScaleDown(x1), DPIUtil.autoScaleDown(y1), size.x, size.y);
+							GC gc = GC.win32_new (hDC, data);
```
HeikoKlare, line 771: This seems to revert changes of a previous commit. Is that intended?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/TaskItem.java
```diff
@@ -462,12 +462,9 @@ public void setProgressState (int progressState) {
...
 	this.shell = shell;
-	shell.addListener (SWT.Dispose, new Listener () {
-		@Override
-		public void handleEvent (Event event) {
-			if (isDisposed ()) return;
-			dispose ();
-		}
+	shell.addListener (SWT.Dispose, event -> {
+		if (isDisposed ()) return;
+		dispose ();
```
HeikoKlare, line 467: Please revert this unrelated change.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
```diff
@@ -3007,6 +3007,14 @@ Point mapInPixels (Control from, Control to, int x, int y) {
 	return new Point (point.x, point.y);
 }
 
+private int getZoomLevelForMapping(Control from, Control to) {
+	return Optional.ofNullable(to)
+            .map(Control::getZoom)
+            .orElseGet(() -> Optional.ofNullable(from)
+                    .map(Control::getZoom)
+                    .orElse(0));
```
HeikoKlare, line 3015: Two question on this:
1. Does it make sense to default the zoom value to "0"? In case this really happens, the results will be completely unusable, so I assume you do not expect this case to occur at all. Wouldn't it be better to thrown an exception then to make explicit what goes wrong if that case really happens?
2. I am not sure whether using stream operations is a good idea here. It produces quite some overhead and the "simple" implementation would, in my opinion, be even easier to understand, as would look something like this, wouldn't it:
```java
if (to != null) {
	return to.getZoom();
}
return from.getZoom();
```

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/widgets/Event.java
```diff
@@ -357,13 +359,26 @@ public void setBounds (Rectangle rect) {
 	this.height = rect.height;
 }
 
+private int getDeviceZoom() {
```
HeikoKlare, line 362: This should just be `getZoom()`, shouldn't it?

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/widgets/Event.java
```diff
@@ -334,15 +334,17 @@ public Rectangle getBounds () {
 	return new Rectangle (x, y, width, height);
 }
 Rectangle getBoundsInPixels () {
-	return DPIUtil.autoScaleUp(getBounds());
+	int deviceZoom = getDeviceZoom();
+	return DPIUtil.autoScaleUp(getBounds(), deviceZoom);
```
HeikoKlare, line 338: You can simply inline the method call instead of assigning it a variable (as the method call is self-explaining) like you do in all the other classes.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/widgets/Event.java
```diff
@@ -357,13 +359,26 @@ public void setBounds (Rectangle rect) {
...
 
+private int getDeviceZoom() {
+	if (widget != null) {
+		return widget.getZoom();
+	} else if (item != null) {
+		return item.getZoom();
+	} else if (gc != null && gc.getGCData() != null) {
+		return DPIUtil.getZoomForAutoscaleProperty(gc.getGCData().nativeDeviceZoom);
+	} else {
+		return DPIUtil.getDeviceZoom();
```
HeikoKlare, line 370: Do you know if/when this case can happen? Using `DPIUtil.getDeviceZoom()` should be something we only do for reasons of compatibility but not because it's actually needed, shouldn't it? I would expect the complete Windows SWT implementation to not use `DPIUtil.getDeviceZoom()` after HiDPI support has been completed at all.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
amartya4256: Yes you are right. There has been a mistake here.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/widgets/Event.java
amartya4256: Reverted changes for Event, since it is in common. @akoch-yatta has a separate PR for this.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/widgets/Event.java
amartya4256: Reverted changes for Event, since it is in common. @akoch-yatta has a separate PR for this.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/widgets/Event.java
amartya4256: Reverted changes for Event, since it is in common. @akoch-yatta has a separate PR for this.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/widgets/Event.java
```diff
@@ -376,4 +376,4 @@ void setLocationInPixels (int x, int y) {
 public String toString () {
 	return "Event {type=" + type + " " + widget + " time=" + time + " data=" + data + " x=" + x + " y=" + y + " width=" + width + " height=" + height + " detail=" + detail + "}";  //$NON-NLS-1$//$NON-NLS-2$ //$NON-NLS-3$ //$NON-NLS-4$ //$NON-NLS-5$ //$NON-NLS-6$ //$NON-NLS-7$
 }
-}
+}
```
HeikoKlare, line 379: Please revert this unrelated/unnecessary change.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
```diff
@@ -3008,6 +3008,13 @@ Point mapInPixels (Control from, Control to, int x, int y) {
 	return new Point (point.x, point.y);
 }
 
+private int getZoomLevelForMapping(Control from, Control to) {
+	if (to != null) {
+		return to.getZoom();
+	}
+	return from.getZoom();
```
HeikoKlare, line 3015: At other places, the `from` and `to` values are checked and throw report proper SWT error, such as 
```java
if (from != null && from.isDisposed()) error (SWT.ERROR_INVALID_ARGUMENT);
```
Should this be done here as well?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
```diff
@@ -7928,8 +7928,8 @@ LRESULT wmNotifyHeader (NMHDR hdr, long wParam, long lParam) {
 							GC gc = createNewGC(nmcd.hdc, data);
 							Rectangle imageBounds = DPIUtil.autoScaleBounds(columns[i].image.getBounds(), this.getZoom(), 100);
 							int y = Math.max (0, (nmcd.bottom - imageBounds.height) / 2);
-							gc.drawImage (columns[i].image, DPIUtil.autoScaleDown(x), DPIUtil.autoScaleDown(y));
-							x += imageBounds.width + 12;
+							x += imageBounds.width + DPIUtil.autoScaleUp(12, getZoom());
+							gc.drawImage (columns[i].image, DPIUtil.scaleDown(x, getZoom()), DPIUtil.scaleDown(y, getZoom()));
```
HeikoKlare, line 7932: Is it inteded that the two statements have been swapped? That seems to modify the behavior.

tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/graphics/WidgetWin32Tests.java
```diff
@@ -0,0 +1,59 @@
...
+import org.eclipse.swt.widgets.Control;
+import org.eclipse.swt.widgets.Display;
+import org.eclipse.swt.widgets.Event;
+import org.eclipse.swt.widgets.Shell;
+import org.eclipse.swt.widgets.WidgetTestUtil;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+public class WidgetWin32Tests {
```
HeikoKlare, line 18: Why is this class placed in the `graphics` instead of the `widgets` package? The `WidgetTestUtil` also seems to be a symptom of this potential misplacement.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
amartya4256: Yeah right. it is used in map methods which uses mapInPixels, so it could have the same effect.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
amartya4256: hmm seems like a mistake but just thinking if adding to x after using it even has any effect. Seems weird.

tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/graphics/WidgetWin32Tests.java
amartya4256: Just wanted to have all these tests for multi zoom level in one package. But if its not a problem then I can move them to separate packages.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
amartya4256: Yes I tested it. Seems consitent. You are right.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/CoolItem.java
```diff
@@ -495,7 +495,7 @@ Point getSizeInPixels() {
 	rect.left -= margins.cxLeftWidth;
 	rect.right += margins.cxRightWidth;
 	if (!parent.isLastItemOfRow (index)) {
-		rect.right += (parent.style & SWT.FLAT) == 0 ? CoolBar.SEPARATOR_WIDTH : 0;
+		rect.right += (parent.style & SWT.FLAT) == 0 ? DPIUtil.autoScaleUp(CoolBar.SEPARATOR_WIDTH, getZoom())  : 0;
```
HeikoKlare, line 498: This changes the existing behavior and rather seems like a fix for an existing bug than a necessity in terms of multi-zoom support (also see my other detailed comment on changed behavior). I would propose to address such existing bugs and the adaptation for multi-zoom support (of which the latter is the goal of this PR) in different PRs to separate concerns and ease review and discussion.

HeikoKlare (commented):
My previous concern almost seem to be addressed, but the recent change seems to introduce faulty behavior (see my detailed comment).
In addition, there are some places where the scaling behavior of constants has been changed. I guess the calculation was erroneous before, but then those changes should be kept separate from this PR which is only supposed to introduce multi-zoom support and should not fix other bugs as a side effect.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Label.java
```diff
@@ -137,9 +137,9 @@ static int checkStyle (int style) {
 	if ((style & SWT.SEPARATOR) != 0) {
 		int lineWidth = OS.GetSystemMetrics (OS.SM_CXBORDER);
 		if ((style & SWT.HORIZONTAL) != 0) {
-			width = DEFAULT_WIDTH;  height = lineWidth * 2;
+			width = DPIUtil.autoScaleUp(DEFAULT_WIDTH, getZoom());  height = lineWidth * 2;
```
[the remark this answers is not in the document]
amartya4256: Basically I have removed the scaling from everywhere where a scaling was not applicable earlier.

tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/widgets/WidgetWin32Tests.java
```diff
@@ -0,0 +1,53 @@
...
+		Button button = new Button(shell, SWT.PUSH);
+		button.setBounds(0, 0, 200, 50);
+		button.setText("Widget Test");
+		button.setBackground(display.getSystemColor(SWT.COLOR_CYAN));
+		shell.open();
+		changeZoomLevel(scaledZoom, shell);
+		assertEquals("The Zoom Level should be updated for button on zoom change event on its shell", scaledZoom, button.getZoom());
+	}
+
+	public void changeZoomLevel(int zoomLevel, Control widget) {
```
fedejeanne, line 43: Does this one really need to be `public`?

tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/widgets/WidgetWin32Tests.java
```diff
@@ -0,0 +1,53 @@
...
+		display = Display.getDefault();
+	}
+
+	@After
+	public void tearDown() {
+		DPITestUtil.setAutoScaleOnRunTime(autoScaleOnRuntime);
+	}
+
+	@Test
+	public void widgetZoomShouldChangeOnZoomLevelChange() {
```
fedejeanne, line 29: Please name the test methods with the prefix `test`, it makes it easier to spot tests in the call hierarchy view.

fedejeanne (changes requested):
@amartya4256 have Heiko's comments been addressed? I think they have, If they have been addressed then please explicitly answer in the threads and mark them as "resolved".

I noticed the tests are not being executed, can it be?

Also, please name the tests with the prefix "test" so one can spot them easier in the call hierarchy view.

tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/graphics/WidgetWin32Tests.java
amartya4256: Done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Widget.java
amartya4256: It was not needed anymore. Removed it.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/TreeItem.java
amartya4256: Changed back to getZoom()

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
amartya4256: Nope. Probably because of rebasing. Removed it.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/TaskItem.java
amartya4256: Done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
amartya4256: Adapted as requested

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/widgets/Event.java
amartya4256: Done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
```diff
@@ -7928,8 +7928,8 @@ LRESULT wmNotifyHeader (NMHDR hdr, long wParam, long lParam) {
 							GC gc = createNewGC(nmcd.hdc, data);
 							Rectangle imageBounds = DPIUtil.autoScaleBounds(columns[i].image.getBounds(), this.getZoom(), 100);
 							int y = Math.max (0, (nmcd.bottom - imageBounds.height) / 2);
-							gc.drawImage (columns[i].image, DPIUtil.autoScaleDown(x), DPIUtil.autoScaleDown(y));
-							x += imageBounds.width + 12;
+							gc.drawImage (columns[i].image, DPIUtil.scaleDown(x, getZoom()), DPIUtil.scaleDown(y, getZoom()));
+							x += imageBounds.width + DPIUtil.autoScaleUp(12, getZoom());
```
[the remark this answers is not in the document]
amartya4256: Reverted to the previous state. I think a separate PR is not needed for now.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/CoolItem.java
amartya4256: Removed DPIUTil calls for all the places where it wasnt there earlier

tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/widgets/WidgetWin32Tests.java
amartya4256: nope. removed the whole method since it is not needed anymore. Extending to the base test for the method

tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/widgets/WidgetWin32Tests.java
amartya4256: Done

tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/widgets/WidgetWin32Tests.java
```diff
@@ -0,0 +1,53 @@
...
+import static org.junit.Assert.assertEquals;
+
+import org.eclipse.swt.SWT;
+import org.eclipse.swt.internal.DPITestUtil;
+import org.eclipse.swt.internal.DPIUtil;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+public class WidgetWin32Tests {
```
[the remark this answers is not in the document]
amartya4256: Moved it to the other class where Andy had implemented similar tests in #1272

bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/WidgetWin32Tests.java
```diff
@@ -0,0 +1,25 @@
...
+		int zoom = DPIUtil.getDeviceZoom();
+		int scaledZoom = zoom * 2;
+
+		Button button = new Button(shell, SWT.PUSH);
+		button.setBounds(0, 0, 200, 50);
+		button.setText("Widget Test");
+		button.setBackground(shell.getDisplay().getSystemColor(SWT.COLOR_CYAN));
+		shell.open();
+		changeDPIZoom(scaledZoom);
+		assertEquals("The Zoom Level should be updated for button on zoom change event on its shell", scaledZoom, button.getZoom());
```
fedejeanne, line 22: Nit: I would explicitly check for this to make the test clearer and more future-proof:

```suggestion
		assertEquals("The initial zoom is wrong", zoom, button.getZoom()); //pre-condition
		changeDPIZoom(scaledZoom);
		assertEquals("The Zoom Level should be updated for button on zoom change event on its shell", scaledZoom, button.getZoom());
```

fedejeanne (changes requested):
Thank you for addressing the comments @amartya4256 , it's looking way better now.

I just have one Nit-Picky comment about the test (with code proposal) and a proposal regarding the commit text (hence the request for changes)

I see in the body of the commit text that:

```
This contribution removes the dependency to the static deviceZoom in DPIUtil for all widgets in the win32 implementation. All calls to DPIUtil method are replaced with calls passing the zoom of the widget directly.
```

How about these calls still in the code? I assume the one in `Display` is still OK (please confirm/rebut) but shouldn't the other ones be removed (in future PRs)?

image

Proposal: reword the commit to:
```
Multi zoom level for Widgets for win32

Do not use DPIUtil::getDeviceZoom to scale the widgets, use DPIUtil::autoScaleUp/scaleDown instead and pass the zoom of the widget to them.
```

If my assumption is correct, that's all that this PR does, right?

Small remark (for a future PR): "autoScaleUp" should probably be renamed to "scaleUp" in order to remain consistent with "scaleDown", which also accepts the zoom as a parameter.

That being said, since I don't expect any "breaking" changes based on my comments, I'm going to test your code for regressions :-)

akoch-yatta:
> How about these calls still in the code? I assume the one in `Display` is still OK (please confirm/rebut) but shouldn't the other ones be removed (in future PRs)?
The call into Taskbar will be adapted in https://github.com/eclipse-platform/eclipse.platform.swt/pull/1269
The calls in the tests are used to store (and later restore) this attribute for now. We should evaluate removing that when we really merged all PR, because any call to DPIUtils without passing the zoom, will still implicitly use hte deviceZoom attribute.
The call in Display is adapted in the Image PR, I think so https://github.com/eclipse-platform/eclipse.platform.swt/pull/1236

fedejeanne:
> > How about these calls still in the code? I assume the one in `Display` is still OK (please confirm/rebut) but shouldn't the other ones be removed (in future PRs)?
> 
> The call into Taskbar will be adapted in #1269. The calls in the tests are used to store (and later restore) this attribute for now. We should evaluate removing that when we really merged all PRs, because any call to DPIUtils without passing the zoom, will still implicitly use hte deviceZoom attribute. The call in Display is adapted in the Image PR, I think, so #1236

Great, thank you @akoch-yatta !

@amartya4256 you should only change the commit text then and add the extra assertion in the test and then this PR is good to go. My testing went well: I didn't find any regressions :-)

fedejeanne (approved):

fedejeanne:
> As soon as the checks are green (specially `Tests / Build`) we can merge this PR. Thank you all for your input!

That one passed and "only" the `Jenkins` check failed with oh-so-annoying: `403 Killed` inside `script.sh.copy`. Last re-run, I'm merging afterwards.

PR #1251, opened 2024-05-23
Remove package protected methods consuming or producing pixel values in Event

akoch-yatta:
In this contribution the DPIUtil dependency from org.eclipse.swt.widgets.Event is removed including all package protected pixel consuming or returning methods. As (at least) Windows receivces HiDPI support for multiple zoom levels, this is a prerequisite to contribute to this feature. As a consequence all callers inside SWT that still used the pixel dependent methods are adapted to only use the public API.

Contributes to #62 and #131

## Requires

- [x]  https://github.com/eclipse-platform/eclipse.platform.swt/pull/1247

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Text.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Text.java
@@ -2504,8 +2504,9 @@ int untranslateOffset (int offset) {
 
 @Override
 void updateMenuLocation (Event event) {
-	Point point = display.mapInPixels (this, null, getCaretLocationInPixels ());
-	event.setLocationInPixels(point.x, point.y + getLineHeightInPixels ());
+	Point pointInPixels = display.mapInPixels (this, null, getCaretLocationInPixels ());
+	int zoom = getZoom();
+	event.setLocation(DPIUtil.scaleDown(pointInPixels.x, zoom), DPIUtil.scaleDown(pointInPixels.y + getLineHeightInPixels (), zoom));
 }
 
 @Override
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
@@ -646,9 +646,9 @@ LRESULT CDDS_ITEMPOSTPAINT (NMTVCUSTOMDRAW nmcd, long wParam, long lParam) {
 								}
 							}
 						}
-						Rectangle boundsInPixels = new Rectangle (cellRect.left, cellRect.top, cellRect.right - cellRect.left, cellRect.bottom - cellRect.top);
-						event.setBoundsInPixels (boundsInPixels);
-						gc.setClipping (DPIUtil.scaleDown (boundsInPixels, zoom));
+						Rectangle bounds = DPIUtil.scaleDown(new Rectangle (cellRect.left, cellRect.top, cellRect.right - cellRect.left, cellRect.bottom - cellRect.top), getZoom());
+						event.setBounds (bounds);
+						gc.setClipping (bounds);
 						sendEvent (SWT.EraseItem, event);
 						event.gc = null;
 						int newTextClr = data.foreground;
@@ -864,7 +864,7 @@ LRESULT CDDS_ITEMPOSTPAINT (NMTVCUSTOMDRAW nmcd, long wParam, long lParam) {
 						}
 					}
 				}
-				event.setBoundsInPixels(new Rectangle(itemRect.left, itemRect.top, itemRect.right - itemRect.left, itemRect.bottom - itemRect.top));
+				event.setBounds(DPIUtil.scaleDown(new Rectangle(itemRect.left, itemRect.top, itemRect.right - itemRect.left, itemRect.bottom - itemRect.top), getZoom()));
 				RECT cellRect = item.getBounds (index, true, true, true, true, true, hDC);
 				int cellWidth = cellRect.right - cellRect.left;
 				int cellHeight = cellRect.bottom - cellRect.top;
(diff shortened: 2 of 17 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
```diff
@@ -3729,7 +3729,8 @@ boolean hitTestSelection (long hItem, int x, int y) {
 	int state = (int)OS.SendMessage (handle, OS.TVM_GETITEMSTATE, hItem, OS.TVIS_SELECTED);
 	int detail = (state & OS.TVIS_SELECTED) != 0 ? SWT.SELECTED : 0;
 	Event event = sendMeasureItemEvent (item, order [index [0]], hDC, detail);
-	if (event.getBoundsInPixels ().contains (x, y)) result = true;
+	int zoom = getZoom();
+	if (event.getBounds ().contains (DPIUtil.scaleDown(x, zoom), DPIUtil.scaleDown(y, zoom))) result = true;
```
HeikoKlare, line 3733: I am not sure how relevant that is, but couldn't this slightly change behavior because of different rounding? In general, I would expect data loss when scaling up the bounds instead of scaling down the position to check for being within the bounds.

HeikoKlare (commented):
In general, the changes look good to me. I only have two remarks:
1. Please check my comment w.r.t. potentially changed behavior by different rounding.
2. This should best be merged after #1247, as otherwise we will have many places in the widget implementations still using `autoScale` methods while whatever is touched by this PR already uses "scale" methods based on an actual zoom.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
akoch-yatta: There is a second place, where this is done. I don't expect it to have a real measurable effect by the user, but I will adapt it, so the bounds a upscaled to don't change the logic at all.

HeikoKlare: I agree that there will probably not be a (noticable) effect, but 
1. we never know ;-)
2. we should keep refactoring separate from (potential) functional changes.

HeikoKlare (approved):

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Text.java
```diff
@@ -2514,7 +2514,8 @@ int untranslateOffset (int offset) {
 @Override
 void updateMenuLocation (Event event) {
 	Point point = display.mapInPixels (this, null, getCaretLocationInPixels ());
```
fedejeanne, line 2516: Nit: I would rename this one to `pointInPixels` so it's easier to see that you're adding pixels with pixels 2 lines below in the call to `event.setLocation(...)`

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Widget.java
```diff
@@ -1137,7 +1137,8 @@ void reskinWidget() {
 boolean sendDragEvent (int button, int x, int y) {
 	Event event = new Event ();
 	event.button = button;
-	event.setLocationInPixels(x, y); // In Pixels
+	int zoom = getZoom();
+	event.setLocation(DPIUtil.scaleDown(x, zoom), DPIUtil.scaleDown(y, zoom)); // In Pixels
```
fedejeanne, line 1141: The comment is now outdated, right?
```suggestion
	event.setLocation(DPIUtil.scaleDown(x, zoom), DPIUtil.scaleDown(y, zoom));
```

fedejeanne (dismissed):
Looks quite good, I only have 2 small remarks.

Also, for my own understanding, one could boil down the code changes from this PR to:
* Replace calls to `Event::getXInPixels` to `Event::getX` (where `X` is one of `Bounds` or `Location`). The callers need to scale the parameters **up** (providing the right `zoom`).
* Replace calls to `Event::setXInPixels` to `Event::setX` (where `X` is one of `Bounds` or `Location`). The callers need to scale the parameters **down** (providing the right `zoom`).

Is that correct or did I miss anything?

fedejeanne (approved):

akoch-yatta:
> Looks quite good, I only have 2 small remarks.
> 
> Also, for my own understanding, one could boil down the code changes from this PR to:
> 
> * Replace calls to `Event::getXInPixels` to `Event::getX` (where `X` is one of `Bounds` or `Location`). The callers need to scale the parameters **up** (providing the right `zoom`).
> * Replace calls to `Event::setXInPixels` to `Event::setX` (where `X` is one of `Bounds` or `Location`). The callers need to scale the parameters **down** (providing the right `zoom`).
> 
> Is that correct or did I miss anything?

That is correct, yes. It is shifting the responsibility to convert from pixels to points and vice versa out of Event to the caller side by doing exactly as you have described.

fedejeanne (dismissed):
The code looks fine but I've tested this and found a possible bug.

The slider is not scaled properly when moving to the 200% monitor. Here you can see it: in white you see the application with the code in `master` and in black you see the one with the code of this PR:

image

### How to reproduce
* Open the SWT Controls and select the tab "**Slider**"
* Move to the 200% monitor

Moving back and forth between the 100% monitor and the 200% monitors does not fix the issue.

This issue is only visible when opening the **Slider** tab in the **SWT Controls** example. Other sliders are scaled properly so it might just be that the example code does something wrong

Can you please look into that?

fedejeanne (approved):

fedejeanne:
FTR rebasing on `master` fixed https://github.com/eclipse-platform/eclipse.platform.swt/pull/1251#pullrequestreview-2132540229

PR #1253, opened 2024-05-23
Unify test and example JDT compiler preferences and link preference files

HannesWell:
This PR consists of three commits to unify JDT compiler settings of example and test projects.
In case of conflicts the stricter variant is chosen, i.e. 'error' over 'warning' and 'warning' over 'info' or not to ignore a problem instead of ignoring it.
Exceptions are
1. Problems about unused elements because tests and examples have some unused elements for good reasons and only warnings be suppressed. Since the build fails on new warnings it is hard to introduce new ones.
2. Problems about indirect static access are set to info because some examples are better to read if static members are accessed through sub-classes.

Additionally fix or suppress all new warnings.

In the third commit the now equal JDT compiler settings files are linked from a shared location in example and test projects.
The first two commits prepare that attempting to make the unification simpler to understand.

All this is based on the discussion in https://github.com/eclipse-platform/eclipse.platform.swt/pull/1238#issuecomment-2125733970. But in contrast to that PR I named the folder containing the shared settings `.settings_shared` instead of `_global` since I think the former is more precise.

In a subsequent PR the preferences for JDT UI could be unified as well in a similar way.


examples/.settings_shared/org.eclipse.jdt.core.prefs
```diff
@@ -99,23 +98,23 @@ org.eclipse.jdt.core.compiler.problem.unlikelyCollectionMethodArgumentType=warni
...
-org.eclipse.jdt.core.compiler.problem.unnecessaryTypeCheck=ignore
+org.eclipse.jdt.core.compiler.problem.unnecessaryTypeCheck=warning
 org.eclipse.jdt.core.compiler.problem.unqualifiedFieldAccess=ignore
 org.eclipse.jdt.core.compiler.problem.unstableAutoModuleName=warning
-org.eclipse.jdt.core.compiler.problem.unusedDeclaredThrownException=ignore
-org.eclipse.jdt.core.compiler.problem.unusedDeclaredThrownExceptionExemptExceptionAndThrowable=enabled
-org.eclipse.jdt.core.compiler.problem.unusedDeclaredThrownExceptionIncludeDocCommentReference=enabled
+org.eclipse.jdt.core.compiler.problem.unusedDeclaredThrownException=warning
+org.eclipse.jdt.core.compiler.problem.unusedDeclaredThrownExceptionExemptExceptionAndThrowable=disabled
+org.eclipse.jdt.core.compiler.problem.unusedDeclaredThrownExceptionIncludeDocCommentReference=disabled
```
fedejeanne, line 106: Didn't you fix these in the code? If you did, you could leave these 2 `enabled`

examples/.settings_shared/org.eclipse.jdt.core.prefs
```diff
@@ -99,23 +98,23 @@ org.eclipse.jdt.core.compiler.problem.unlikelyCollectionMethodArgumentType=warni
...
 org.eclipse.jdt.core.compiler.problem.unusedDeclaredThrownExceptionWhenOverriding=disabled
 org.eclipse.jdt.core.compiler.problem.unusedExceptionParameter=ignore
 org.eclipse.jdt.core.compiler.problem.unusedImport=error
 org.eclipse.jdt.core.compiler.problem.unusedLabel=warning
-org.eclipse.jdt.core.compiler.problem.unusedLocal=error
+org.eclipse.jdt.core.compiler.problem.unusedLocal=warning
 org.eclipse.jdt.core.compiler.problem.unusedObjectAllocation=ignore
 org.eclipse.jdt.core.compiler.problem.unusedParameter=ignore
-org.eclipse.jdt.core.compiler.problem.unusedParameterIncludeDocCommentReference=enabled
+org.eclipse.jdt.core.compiler.problem.unusedParameterIncludeDocCommentReference=disabled
```
fedejeanne, line 114: This one also is now more lenient. Was that intended?

If I understand this correctly (see [COMPILER_PB_UNUSED_PARAMETER_INCLUDE_DOC_COMMENT_REFERENCE](https://github.com/eclipse-jdt/eclipse.jdt.core/blob/master/org.eclipse.jdt.core/model/org/eclipse/jdt/core/JavaCore.java#L557:L571)), this change means that parameters mentioned as `@param` in the JavaDoc are **not** mandatory _i.e._ there could be wrong documentation (a parameter is documented but it doesn't exist). Is that correct?

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet131.java
```diff
@@ -39,7 +39,7 @@ public static void main (String [] args) {
 		while (!menu.isDisposed () && menu.isVisible ()) {
 			if (!display.readAndDispatch ()) display.sleep ();
 		}
-		while (display.readAndDispatch()); // needed, to get the selection event, which is fired AFTER the menu is hidden
+		while (display.readAndDispatch()) // needed, to get the selection event, which is fired AFTER the menu is hidden
```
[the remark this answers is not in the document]
HannesWell: Yes. The semicolon is superfluous and a warning was added about an empty expression (which it effectively is, although the formatting implies something else).

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet155.java
```diff
@@ -21,15 +21,14 @@
...
-import java.awt.Graphics;
-import java.awt.Dimension;
 
 import org.eclipse.swt.*;
-import org.eclipse.swt.widgets.*;
+import org.eclipse.swt.awt.*;
 import org.eclipse.swt.layout.*;
-import org.eclipse.swt.awt.SWT_AWT;
+import org.eclipse.swt.widgets.*;
+import org.eclipse.swt.widgets.Composite;
```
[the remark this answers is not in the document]
HannesWell: Looks like this explicit import is required to distinguish the SWT Canvas from `java.awt.Composite`.
If I delete it, there is a Compiler-error.

examples/.settings_shared/org.eclipse.jdt.core.prefs
HannesWell: If I understand it correctly it is the other way round: If this key is set to `enabled` it means that a `@param` java-doc tag counts towards the usages of a parameter. Thus just having a parameter documented would disable an unused-warning even if the parameter is not used in the code.

I think the text in the UI also confirms this:
grafik

The settings
- [`unusedDeclaredThrownExceptionIncludeDocCommentReference`](https://github.com/eclipse-jdt/eclipse.jdt.core/blob/6eb38a2431cb33876687312fddec8298f4ee9f6b/org.eclipse.jdt.core/model/org/eclipse/jdt/core/JavaCore.java#L952)
- [`unusedDeclaredThrownExceptionExemptExceptionAndThrowable`](https://github.com/eclipse-jdt/eclipse.jdt.core/blob/6eb38a2431cb33876687312fddec8298f4ee9f6b/org.eclipse.jdt.core/model/org/eclipse/jdt/core/JavaCore.java#L978)

are similar.
At the other hand [`unusedDeclaredThrownExceptionWhenOverriding`](https://github.com/eclipse-jdt/eclipse.jdt.core/blob/6eb38a2431cb33876687312fddec8298f4ee9f6b/org.eclipse.jdt.core/model/org/eclipse/jdt/core/JavaCore.java#L931) is different because setting it to `enabled` is more strict. Consequently unchecking it in the preference UI leads to the value `enabled`. I have to update that.

examples/.settings_shared/org.eclipse.jdt.core.prefs
HannesWell: Yes I did, the values are just counterintuitive as described in https://github.com/eclipse-platform/eclipse.platform.swt/pull/1253#discussion_r1614405951.

HeikoKlare (commented):
Thank you making this cleanup and "centralization" of settings, @HannesWell!

The changes look sound to me. I have to admit that I did not make a detailed comparison of every existing settings configuration with the new shared one, but checking the commit trace in the PR indicates that it is done "correctly", and the new settings make sense to me anyway.

My only comment is w.r.t. the removed semicolon in one of the snippets, which actually changes the behavior (see my comment on that).

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet131.java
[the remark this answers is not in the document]
HannesWell: Yes, you are absolutely right.
Sorry, I somehow interpreted that control structure as a do-while loop, while skimming over all the warnings and didn't understand what @fedejeanne meant.
As you suggested I replaced the semicolon with empty braces.

examples/.settings_shared/org.eclipse.jdt.core.prefs
fedejeanne: > If I understand it correctly it is the other way round: If this key is set to enabled it means that a @param java-doc tag counts towards the usages of a parameter. Thus just having a parameter documented would disable an unused-warning even if the parameter is not used in the code.
>
> I think the text in the UI also confirms this:

You're right, thank you for the screenshot, it's clearer to me.

> At the other hand [unusedDeclaredThrownExceptionWhenOverriding](https://github.com/eclipse-jdt/eclipse.jdt.core/blob/6eb38a2431cb33876687312fddec8298f4ee9f6b/org.eclipse.jdt.core/model/org/eclipse/jdt/core/JavaCore.java#L931) is different because setting it to enabled is more strict. Consequently unchecking it in the preference UI leads to the value enabled. I have to change that.

👍

fedejeanne (approved):
LGTM

HeikoKlare (approved):

HannesWell:
Thanks.

@iloveeclipse, @akurtakov is this PR save to submit at the moment? It just affects the SWT tests and examples, but they are still published in the eclipse/updates repository.

akurtakov:
@HannesWell I would prefer to wait for next stream. There are literally couple more builds and I don't want to have to deal with some build breakage due to this.

HannesWell:
Quality-gate fails because API-filters are unused, like on master. In order to unblock the PRs depending I'll leave that for another PR (not in the computer anymore).
Unless somebody does it earlier, I can do it tomorrow evening.

PR #1256, opened 2024-05-29
Remove methods producing pixel values in GlyphMetrics

akoch-yatta:
In this contribution the DPIUtil dependency from org.eclipse.swt.graphics.GlyphMetrics is removed including all package protected pixel producing methods. As (at least) Windows receivces HiDPI support for multiple zoom levels, this is a prerequisite to contribute this feature. As a consequence all callers inside SWT that still used the pixel dependent methods are adapt to only use the public API.

Contributes to #62 and #131


HeikoKlare (approved):
Looks good, in particular since all but one method were even completely unused. Thank you!

PR #1258, opened 2024-06-03
added native bindings for win32

amartya4256:
## Addressed issues
* https://github.com/eclipse-platform/eclipse.platform.swt/issues/127

## Unlocks
* https://github.com/eclipse-platform/eclipse.platform.swt/pull/1214

## Description

This PR provides the access to windows API which are later needed for the following PRs for the support of Multi Zoom Level.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT PI/win32/org/eclipse/swt/internal/win32/OS.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT PI/win32/org/eclipse/swt/internal/win32/OS.java
@@ -497,6 +497,9 @@ public class OS extends C {
 	public static final int FR_PRIVATE = 0x10;
 	public static final int FSHIFT = 0x4;
 	public static final int FVIRTKEY = 0x1;
+	public static final int GA_PARENT = 0x1;
+	public static final int GA_ROOT = 0x2;
+	public static final int GA_ROOTOWNER = 0x3;
 	public static final int GCP_REORDER = 0x0002;
 	public static final int GCP_GLYPHSHAPE = 0x0010;
 	public static final int GCP_CLASSIN = 0x00080000;
@@ -2709,6 +2712,8 @@ public static final native int FillRect (long hDC, RECT lprc, long hbr);
 public static final native int GdiSetBatchLimit (int dwLimit);
 public static final native int GetACP ();
 public static final native long GetActiveWindow ();
+/** @param hWnd cast=(HWND) */
+public static final native long GetAncestor (long hWnd, int gaFlags);
 /** @param hDC cast=(HDC) */
 public static final native int GetBkColor (long hDC);
 public static final native long GetCapture ();
--- a/bundles/org.eclipse.swt/Eclipse SWT PI/win32/library/os.c
+++ b/bundles/org.eclipse.swt/Eclipse SWT PI/win32/library/os.c
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2023 IBM Corporation and others.
+ * Copyright (c) 2000, 2024 IBM Corporation and others.
  *
  * This program and the accompanying materials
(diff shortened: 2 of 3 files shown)


bundles/org.eclipse.swt/Eclipse SWT PI/win32/org/eclipse/swt/internal/win32/OS.java
```diff
@@ -2709,6 +2710,8 @@ public static int HRESULT_FROM_WIN32(int x) {
 public static final native int GdiSetBatchLimit (int dwLimit);
 public static final native int GetACP ();
 public static final native long GetActiveWindow ();
+/** @param hWnd cast=(HWND) */
```
fedejeanne, line 2713: If I read the [MSDN documentation correctly](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getancestor#parameters), the 2nd parameter could be one of `GA_PARENT` (`1`), `GA_ROOT` (`2`) or `GA_ROOT_OWNER` (`3`). I would specify this in the documentation or at least make it clear that only `GA_ROOT` is available.

fedejeanne (changes requested):
The documentation could be improved.

Other than that, I'm not an expert on JNI calls so I cannot comment on the changes in the C-files. I jut approved the workflow, we should see if all checks pass.

bundles/org.eclipse.swt/Eclipse SWT PI/win32/org/eclipse/swt/internal/win32/OS.java
```diff
@@ -497,6 +497,7 @@ public class OS extends C {
 	public static final int FR_PRIVATE = 0x10;
 	public static final int FSHIFT = 0x4;
 	public static final int FVIRTKEY = 0x1;
+	public static final int GA_ROOT = 0x2;
```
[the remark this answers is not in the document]
HeikoKlare: From what I have seen in the `OS` class so far, not all predefined argument / flag values are provided together with a method accepting them, but they are added on-demand. Maybe this was by accident and actually it does no harm to define all predefined argument values together with a method. So it is of course okay to do so, but I would not expect it to be mandatory, as it would only add constants that are unused.

HeikoKlare (approved):
Changes look good to me. Added access to Win32 API seems to be consistent with its documentation: https://learn.microsoft.com/de-de/windows/win32/api/winuser/nf-winuser-getancestor.

bundles/org.eclipse.swt/Eclipse SWT PI/win32/org/eclipse/swt/internal/win32/OS.java
amartya4256: I am adding other parameters as well.

amartya4256: There are methods like GetCharacterPlacement and GetDIBColorTable, where they have paramters like dwFlag or gwFlag and they are not the part of documentation. I think I should follow the exisiting convention and not mention them in the docs since there flags are avaialble above anyways with the same name.

fedejeanne: 👍

fedejeanne (approved):
LGTM now, thank you!

HeikoKlare (approved):

HeikoKlare:
After triggering like 15 rebuilds, the `master` build finally succeeded (https://ci.eclipse.org/releng/job/eclipse.platform.swt/job/master/586/) and pushed the new binaries based on the changes in this PR (https://github.com/eclipse-platform/eclipse.platform.swt/commit/567a8b40f4913e08cc4f878e4ccc9d0b45981634).

PR #1263, opened 2024-06-05
Remove API filter for 4.32 release

HeikoKlare:
For the 4.32 release, an API filter for the TypedListener being annotated with `@noextend` has been introduced (see https://github.com/eclipse-platform/eclipse.platform.swt/pull/1101). After having switched to the 4.33 stream, this filter now became obsolete and is thus removed with this change.

This resolves the warning for an unused API filter:
image
I would actually expect that we can remove the other API filters as well, but since the build validation currently only complains about the `TypedListener`, this change only removes exactly that filter.

@HannesWell I just took over from you (https://github.com/eclipse-platform/eclipse.platform.swt/pull/1253#issuecomment-2150956767), as we would like to merge some finalized PRs tomorrow and would thus like to resolve the new warnings before.

### DIFF
--- a/binaries/org.eclipse.swt.cocoa.macosx.aarch64/.settings/.api_filters
+++ b/binaries/org.eclipse.swt.cocoa.macosx.aarch64/.settings/.api_filters
@@ -550,12 +550,4 @@
             </message_arguments>
         </filter>
     </resource>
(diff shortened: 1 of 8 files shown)


HannesWell (approved):
Thank you! 
Looks good.

PR #1264, opened 2024-06-06
Update tree item height correctly on DPI change

akoch-yatta:
With the current implementation of the handler on DPI zoom changes, not all rescaling cases are covered correctly. If the item height was set at least once programmatically (currently done in Tree::setItemHeight), it will never be done by the OS anymore. With the current implementation a double scale up can happen (once scaled up from the OS and once from the handler), see:

image

To make it failsafe a new internal flag is used to evaluate whether the item height was set programmatically or not.

## How to reproduce

1. Adapt the VM argument of e.g. the Eclipse IDE by adding
```
-Dswt.autoScale=quarter
```
2. Have two monitors with different zoom, e.g. 100% and 150%
3. Start the IDE on 100%
4. Open the Outline view
5. Move it to the monitor with 150%  -> The item height will be too big as in the screenshot

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
@@ -8280,11 +8280,12 @@ private static void handleDPIChange(Widget widget, int newZoom, float scalingFac
 		tree.imageList = null;
 	}
 
-	if (tree.hooks(SWT.MeasureItem)) {
-		// with the measure item hook, the height must be programmatically recalculated
-		var itemHeight = tree.getItemHeightInPixels();
-		tree.setItemHeight(Math.round(itemHeight * scalingFactor));
-	}
+	// if the item height was set at least once programmatically with TVM_SETITEMHEIGHT,
+	// the item height of the tree is not managed by the OS anymore e.g. when the zoom
+	// on the monitor is changed, the height of the item will stay at the fixed size.
+	// Resetting it will re-enable the default behavior again
+	tree.setItemHeight(-1);
+
 	for (TreeColumn treeColumn : tree.getColumns()) {
 		DPIZoomChangeRegistry.applyChange(treeColumn, newZoom, scalingFactor);
 	}


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
```diff
@@ -4347,7 +4348,11 @@ void setItemCount (int count, long hParent) {
 /*public*/ void setItemHeight (int itemHeight) {
 	checkWidget ();
 	if (itemHeight < -1) error (SWT.ERROR_INVALID_ARGUMENT);
-	OS.SendMessage (handle, OS.TVM_SETITEMHEIGHT, itemHeight, 0);
+	long newHeight = OS.SendMessage (handle, OS.TVM_SETITEMHEIGHT, itemHeight, 0);
```
fedejeanne, line 4351: According to the [documentation of TVM_SETITEMHEIGHT](https://learn.microsoft.com/en-us/windows/win32/controls/tvm-setitemheight#return-value):
> # Return value
> Returns the previous height of the items, in pixels.

So this one should be renamed to `oldHeight`.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
```diff
@@ -4347,7 +4348,11 @@ void setItemCount (int count, long hParent) {
...
+	// if the item height was set at least once programmatically, it will never be
+	// done by the OS anymore. To handle resizing after DPI changes, this must be
+	// tracked with the itemHeightSetProgrammatically-flag
```
fedejeanne, line 4354: It's kind of confusing what this comment wants to say, can you rephrase it?

If I interpret the changes correctly, the `if` in line 8288 can only work if the method `setItemHeight(int)` has been called at least twice, right?

akoch-yatta: Sure, I will rephrase it a bit

But why twice?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
akoch-yatta: I updated the branch. The solution is now much simpler as before. The behavior can just be reset by setting it to -1.

PR #1267, opened 2024-06-07
Ignore Changes to macOS .DS_Store Files

BeckerWdf:


BeckerWdf:
Why is the build failing?
I don't understand the logs.

PR #1268, opened 2024-06-07
Init nativeZoom of Tracker on creation

akoch-yatta:
As one constructor of Tracker is not calling the super constructor the nativeZoom attribute must be set here as well. As it only has the display as reference, the zoom of the display (currently equal to primary monitor zoom) will be used.

Contributes to #62 and #131


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tracker.java
```diff
@@ -142,6 +142,7 @@ public Tracker (Display display, int style) {
 	}
 	this.style = checkStyle (style);
 	this.display = display;
+	this.nativeZoom = display.getDeviceZoom();
```
HeikoKlare, line 145:
```suggestion
	this.nativeZoom = display.getNativeDeviceZoom();
```

fedejeanne: 👍 Makes sense.

In that case:
* The `@since` tags could also be removed (since it's a protected method)
* `org.eclipse.swt.internal.DPIUtil.setDeviceZoom(int)` should also be renamed to `setNativeDeviceZoom`.

@akoch-yatta propose a separate PR for those changes please.

fedejeanne (approved):
The changes look good. Since I propose to do a separate PR for @HeikoKlare 's comment, this PR is good to go on my account.

@HeikoKlare if you agree, this PR can be merged.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tracker.java
HeikoKlare: > * The `@since` tags could also be removed (since it's a protected method)

Which method do you mean?

HeikoKlare:
> @HeikoKlare if you agree, this PR can be merged.

No, the implementation is wrong (see my comment), so it should not be merged.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tracker.java
fedejeanne: I meant the `@since` in `org.eclipse.swt.widgets.Display.getDeviceZoom()`

HeikoKlare: Adding protected methods requires a `@since` tag, so that one is correct. Maybe you are mixing up _package protected_ and _protected_ methods. Only the former don't need that tag as they are "internal".

akoch-yatta: I'm confused. 
There is not getNativeDeviceZoom() in Display (or Device). getDeviceZoom() is the one that is returning the (native) zoom of the primary method.

HeikoKlare: I am so sorry for introducing this confusion 🙁 I mixed up the methods in `Display` with those in `DPIUtil`...

My comments w.r.t. the `@since` are still valid, they properly refer to the `Display` class.

HeikoKlare (approved):
The change is fine, thank you! My comment was invalid as I mixed up methods from the `Display` and `DPIUtil` classes.

PR #1269, opened 2024-06-07
[win32] Init nativeZoom of taskbar on creation

akoch-yatta:
As the constructor of TaskBar is not calling the super constructor the nativeZoom attribute must be set here as well. As it only has the display as reference, the zoom of the display (currently equal to primary monitor zoom) will be used

Contributes to #62 and #131

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/TaskBar.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/TaskBar.java
@@ -65,6 +65,7 @@ public class TaskBar extends Widget {
 
 TaskBar (Display display, int style) {
 	this.display = display;
+	this.nativeZoom = display.getDeviceZoom();
 	createHandle ();
 	reskinWidget ();
 }
@@ -166,14 +167,14 @@ IShellLink createShellLink (MenuItem item) {
 				ImageData data;
 				if (item.hBitmap != 0) {
 					Image image2 = Image.win32_new (display, SWT.BITMAP, item.hBitmap);
-					data = image2.getImageData (DPIUtil.getDeviceZoom ());
+					data = image2.getImageData (getZoom());
 					/*
 					 * image2 instance doesn't own the handle and shall not be disposed. Make it
 					 * appear disposed to cause leak trackers to ignore it.
 					 */
 					image2.handle = 0;
 				} else {
-					data = image.getImageData (DPIUtil.getDeviceZoom ());
+					data = image.getImageData (getZoom());
 				}
 				ImageLoader loader = new ImageLoader ();
 				loader.data = new ImageData [] {data};


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/TaskBar.java
```diff
@@ -65,6 +65,7 @@ public class TaskBar extends Widget {
 
 TaskBar (Display display, int style) {
 	this.display = display;
+	this.nativeZoom = display.getDeviceZoom();
```
HeikoKlare, line 68:
```suggestion
	this.nativeZoom = display.getNativeDeviceZoom();
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/TaskBar.java
```diff
@@ -166,14 +167,14 @@ IShellLink createShellLink (MenuItem item) {
 				ImageData data;
 				if (item.hBitmap != 0) {
 					Image image2 = Image.win32_new (display, SWT.BITMAP, item.hBitmap);
-					data = image2.getImageData (DPIUtil.getDeviceZoom ());
+					data = image2.getImageData (getZoom());
```
HeikoKlare, line 170: Can't the zoom of the `MenuItem` passed to the method be used instead of using the zoom of the `TaskBar`, which is simply the one of the primary monitor? I would expect this to usually use the wrong scale value.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/TaskBar.java
fedejeanne: See https://github.com/eclipse-platform/eclipse.platform.swt/pull/1268#discussion_r1658595251

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/TaskBar.java
fedejeanne: I thought `getZoom()` gets the current zoom (`nativeZoom` located in `Widget`), isn't that so? 

The way I understood it is that every time an event of the type `SWT.ZoomChanged` is triggered, this ends up calling `org.eclipse.swt.widgets.Widget.handleDPIChange(Widget, int, float)`, which sets `nativeZoom`.

On the other hand, I don't see any special handler registration in `TaskBar` as other widgets have (_e.g._ `ExpandBar`).

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/TaskBar.java
HeikoKlare: My comment was invalid (see https://github.com/eclipse-platform/eclipse.platform.swt/pull/1268#discussion_r1660574063).

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/TaskBar.java
akoch-yatta: Hmm, yes, I think in this context using the zoom from menu item makes probably one sense. I will have a second look at it

HeikoKlare: I am not sure if (and under which conditions) `TaskBar` receives any zoom change events from the OS. Maybe this happens when the primary monitor or its DPI settings change? Anyway, if no zoom change happens, the `TaskBar` is initialized with a `nativeZoom` given by `display.getDeviceZoom()` (see the other change in this PR). That one always originates from the primary monitor.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/TaskBar.java
fedejeanne: So no need to change anything here, right @HeikoKlare ?

HeikoKlare: Not at this place, thus I had resolved this conversation. But why did you resolve my other comment? That one requires some adaptation.

fedejeanne: > ... But why did you resolve my other comment? That one requires some adaptation.

Sloppy fingers, sorry!

akurtakov:
Is this PR still relevant?

akoch-yatta:
No, I think this PR is not relevant anymore. It is practically unused anyway and the change could perhaps brake something. As we are not aware of any issues with it, I will close this PR.

PR #1277, opened 2024-06-12
Snippets: reduce warnings

jukzi:
* The static method should be accessed directly
* The static field should be accessed directly
* discouraged access

### DIFF
--- a/examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet123.java
+++ b/examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet123.java
@@ -28,6 +28,7 @@ package org.eclipse.swt.snippets;
  * http://www.eclipse.org/swt/snippets/
  */
 import org.eclipse.swt.*;
+import org.eclipse.swt.internal.*;
 import org.eclipse.swt.internal.ole.win32.*;
 import org.eclipse.swt.internal.win32.*;
 import org.eclipse.swt.layout.*;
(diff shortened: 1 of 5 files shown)


HannesWell:
> * The static field should be accessed directly

In https://github.com/eclipse-platform/eclipse.platform.swt/pull/1253 actually I disabled that warning assuming it was written like that to increase readability.

jukzi:
> In #1253 actually I disabled that warning

I see problem marker for this in workspace? disabling would be OK for me too, but seeing a Problem is not appropriate when it can just be fixed with quickfix.
image
I don't see a difference in readability.

HannesWell:
> > In #1253 actually I disabled that warning
> 
> I see problem marker for this in workspace? disabling would be OK for me too, but seeing a Problem is not appropriate when it can just be fixed with quickfix. 

Aren't these info-level markers issued by Sonar-Lint? But these could also be info-markers issued by JDT (haven't checked it).

> I don't see a difference in readability.

I haven't looked in detail into the snippets and don't have a strong opinion. I just kept it as it is when unifying the preferences because I assumed the original authors found it better the way it is.
Since these snippets are for consumers of SWT I think the code-style should be a good example.
If we think qualifying the static access always with the declaring class I'm fine either.

But then the marker-level should be increased to warning to not add new issues.

jukzi:
> Aren't these info-level markers issued by Sonar-Lint? 

I don't even have that installed and you can look in your workspace's Problems view.

> But then the marker-level should be increased to warning to not add new issues.

Well it was your commit that set org.eclipse.jdt.core.compiler.problem.indirectStaticAccess=info on all related projects.
Before it was ignore/warning depending on project.
What i do not understand is why info markers are not visible in jenkins. And why you modify warning levels without looking in the Problems view  ;-) and why you submitted with +1 new warning, so did even ignore jenkins warnings.
Anyway i think your "info" level is fair as static access is not any real problem. And linking the settings was a good idea.

PR #1278, opened 2024-06-12
Multi Zoom Support for region for win32

amartya4256:
## Addressed issues

## Requires

## Description

This pull request is based on the implementations of PR #1214. It extends to the native zoom which is provided within widget and propagated to GC. In this PR, the multiple zoom level functionality is attained by using a map to maintain the handle of the region scale as per zoom level. The handle can then be obtained by the method win32_getHandle by passing the zoom information from the client.

contributes to #62 and #127

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/graphics/RegionWin32Tests.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/graphics/RegionWin32Tests.java
@@ -1,0 +1,52 @@
+/*******************************************************************************
+ * Copyright (c) 2024 Yatta Solutions
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Yatta Solutions - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.swt.graphics;
+
+import static org.junit.Assert.assertEquals;
+
+import org.eclipse.swt.internal.*;
+import org.eclipse.swt.internal.win32.*;
+import org.junit.jupiter.api.*;
+
+class RegionWin32Tests extends Win32AutoscaleTestBase {
+
+	@Test
+	public void testRegionMustBeScaledOnHandleOfScaledZoomLevel() {
+		int zoom = DPIUtil.getDeviceZoom();
+		int scalingFactor = 2;
+
+		Region region = new Region(display);
+		region.add(0, 0, 5, 10);
+		region.subtract(0,0,1,1);
+		region.translate(0, 5);
+		region.intersect(1,1,1,1);
+
+		long handle = Region.win32_getHandle(region, zoom);
+		long scaledRegionHandle = Region.win32_getHandle(region, zoom * scalingFactor);
+
+		RECT rect = new RECT();
+		OS.GetRgnBox(handle, rect);
+		Rectangle bounds = new Rectangle(rect.left, rect.top, rect.right - rect.left, rect.bottom - rect.top);
+
+		rect = new RECT();
+		OS.GetRgnBox(scaledRegionHandle, rect);
+		Rectangle scaledBounds = new Rectangle(rect.left, rect.top, rect.right - rect.left, rect.bottom - rect.top);
+
+		assertEquals("scaled region's height should be double of unscaled region", bounds.height * scalingFactor, scaledBounds.height);
+		assertEquals("scaled region's width should be double of unscaled region", bounds.width * scalingFactor, scaledBounds.width);
+		assertEquals("scaled region's x position should be double of unscaled region", bounds.x * scalingFactor, scaledBounds.x);
+		assertEquals("scaled region's y position should be double of unscaled region", bounds.y * scalingFactor, scaledBounds.y);
+	}
+
+}
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Region.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Region.java
@@ -14,6 +14,9 @@
 package org.eclipse.swt.graphics;
 
 
+import java.util.*;
+import java.util.stream.*;
+
 import org.eclipse.swt.*;
 import org.eclipse.swt.internal.*;
 import org.eclipse.swt.internal.win32.*;
@@ -33,19 +36,11 @@ import org.eclipse.swt.internal.win32.*;
  */
 public final class Region extends Resource {
 
-	/**
-	 * the OS resource for the region
-	 * (Warning: This field is platform dependent)
-	 * <p>
-	 * <b>IMPORTANT:</b> This field is <em>not</em> part of the SWT
-	 * public API. It is marked public only so that it can be shared
-	 * within the packages provided by SWT. It is not available on all
-	 * platforms and should never be accessed from application code.
-	 * </p>
-	 *
-	 * @noreference This field is not intended to be referenced by clients.
-	 */
-	public long handle;
+	private int initialZoom;
+
+	private HashMap<Integer, Long> zoomToHandle = new HashMap<>();
+
+	private List<Operation> operations = new ArrayList<>();
 
 /**
  * Constructs a new empty region.
@@ -84,7 +79,9 @@ public Region () {
  */
 public Region (Device device) {
 	super(device);
-	handle = OS.CreateRectRgn (0, 0, 0, 0);
+	initialZoom = DPIUtil.getDeviceZoom();
+	long handle = OS.CreateRectRgn (0, 0, 0, 0);
+	zoomToHandle.put(initialZoom, handle);
 	if (handle == 0) SWT.error(SWT.ERROR_NO_HANDLES);
 	init();
 }
@@ -107,13 +104,8 @@ public Region (Device device) {
 public void add (int[] pointArray) {
 	if (isDisposed()) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
 	if (pointArray == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
-	addInPixels(DPIUtil.autoScaleUp(pointArray));
-}
-
-void addInPixels (int[] pointArray) {
-	long polyRgn = OS.CreatePolygonRgn(pointArray, pointArray.length / 2, OS.ALTERNATE);
-	OS.CombineRgn (handle, handle, polyRgn, OS.RGN_OR);
-	OS.DeleteObject (polyRgn);
+	final Operation operation = new OperationWithArray(Operation::add, Arrays.copyOf(pointArray, pointArray.length));
+	storeAndApplyOperationForAllHandles(operation);
 }
 
 /**
@@ -133,8 +125,8 @@ void addInPixels (int[] pointArray) {
 public void add (Rectangle rect) {
 	if (isDisposed()) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
 	if (rect == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
-	rect = DPIUtil.autoScaleUp(rect);
-	addInPixels(rect.x, rect.y, rect.width, rect.height);
+	final Operation operation = new OperationWithRectangle(Operation::add, new Rectangle(rect.x, rect.y, rect.width, rect.height));
+	storeAndApplyOperationForAllHandles(operation);
 }
 
 /**
@@ -157,14 +149,8 @@ public void add (Rectangle rect) {
  */
 public void add (int x, int y, int width, int height) {
 	if (isDisposed()) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
-	addInPixels(DPIUtil.autoScaleUp(x), DPIUtil.autoScaleUp(y), DPIUtil.autoScaleUp(width), DPIUtil.autoScaleUp(height));
-}
-
-void addInPixels (int x, int y, int width, int height) {
-	if (width < 0 || height < 0) SWT.error(SWT.ERROR_INVALID_ARGUMENT);
-	long rectRgn = OS.CreateRectRgn (x, y, x + width, y + height);
-	OS.CombineRgn (handle, handle, rectRgn, OS.RGN_OR);
-	OS.DeleteObject (rectRgn);
+	final Operation operation = new OperationWithRectangle(Operation::add, new Rectangle(x, y, width, height));
+	storeAndApplyOperationForAllHandles(operation);
 }
 
 /**
@@ -186,7 +172,8 @@ public void add (Region region) {
 	if (isDisposed()) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
 	if (region == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
 	if (region.isDisposed()) SWT.error(SWT.ERROR_INVALID_ARGUMENT);
-	OS.CombineRgn (handle, handle, region.handle, OS.RGN_OR);
+	final Operation operation = new OperationWithRegion(Operation::add, region);
+	storeAndApplyOperationForAllHandles(operation);
 }
 
 /**
@@ -204,11 +191,11 @@ public void add (Region region) {
  */
 public boolean contains (int x, int y) {
 	if (isDisposed()) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
-	return containsInPixels(DPIUtil.autoScaleUp(x), DPIUtil.autoScaleUp(y));
+	return containsInPixels(DPIUtil.autoScaleUp(x, initialZoom), DPIUtil.autoScaleUp(y, initialZoom));
 }
 
 boolean containsInPixels (int x, int y) {
-	return OS.PtInRegion (handle, x, y);
+	return OS.PtInRegion (getHandleForInitialZoom(), x, y);
 }
 
 /**
@@ -229,14 +216,15 @@ boolean containsInPixels (int x, int y) {
 public boolean contains (Point pt) {
 	if (isDisposed()) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
 	if (pt == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
-	pt = DPIUtil.autoScaleUp(pt);
-	return containsInPixels(pt.x, pt.y);
+	Point p = DPIUtil.autoScaleUp(pt, initialZoom);
+	return containsInPixels(p.x, p.y);
 }
 
 @Override
 void destroy () {
-	OS.DeleteObject(handle);
-	handle = 0;
+	zoomToHandle.values().forEach(handle -> OS.DeleteObject(handle));
+	zoomToHandle.clear();
+	operations.clear();
 }
 
 /**
@@ -254,7 +242,7 @@ public boolean equals (Object object) {
 	if (this == object) return true;
 	if (!(object instanceof Region)) return false;
 	Region rgn = (Region)object;
-	return handle == rgn.handle;
+	return getHandleForInitialZoom() == rgn.getHandleForInitialZoom();
 }
 
 /**
@@ -272,12 +260,12 @@ public boolean equals (Object object) {
  */
 public Rectangle getBounds () {
 	if (isDisposed()) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
-	return DPIUtil.autoScaleDown(getBoundsInPixels());
+	return DPIUtil.scaleDown(getBoundsInPixels(), initialZoom);
 }
 
 Rectangle getBoundsInPixels() {
 	RECT rect = new RECT();
-	OS.GetRgnBox(handle, rect);
+	OS.GetRgnBox(getHandleForInitialZoom(), rect);
 	return new Rectangle(rect.left, rect.top, rect.right - rect.left, rect.bottom - rect.top);
 }
 
@@ -293,7 +281,7 @@ Rectangle getBoundsInPixels() {
  */
 @Override
 public int hashCode () {
-	return (int)handle;
+	return (int)getHandleForInitialZoom();
 }
 
 /**
@@ -315,8 +303,8 @@ public int hashCode () {
 public void intersect (Rectangle rect) {
 	if (isDisposed()) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
 	if (rect == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
-	rect = DPIUtil.autoScaleUp(rect);
-	intersectInPixels(rect.x, rect.y, rect.width, rect.height);
+	final Operation operation = new OperationWithRectangle(Operation::intersect, new Rectangle(rect.x, rect.y, rect.width, rect.height));
+	storeAndApplyOperationForAllHandles(operation);
 }
 
 /**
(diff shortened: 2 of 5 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Region.java
```diff
@@ -47,6 +49,8 @@ public final class Region extends Resource {
 	 */
 	public long handle;
 
+	private int initalZoom;
```
fedejeanne, line 52:
```suggestion
	private int initialZoom;
```

fedejeanne (changes requested):
The code doesn't compile. 

Also, there's a typo.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Region.java
amartya4256: Yeah it needed some cleaning but its all fixed now

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Region.java
```diff
@@ -614,8 +621,29 @@ void translateInPixels (int x, int y) {
...
+	translateInPixels(p.x, p.y);
+}
+
+private HashMap<Integer, Long> handleMap = new HashMap<>();
+
+/**
+ * Gets the handle for the region scaled at required zoom level
+ *
+ * @param region the region to be scaled
+ *
```
akoch-yatta, line 634: The comment is missing the `@noreference` and the please to not use part

amartya4256: Done

fedejeanne (changes requested):
I tested the `Snippet219` and when I move from the 100% monitor to the 200% monitor I see that the region (content) is resized but the shell isn't 

100% 
image

200% (ignore the background, it's my wallpaper)
image

Can you please look into that @amartya4256 ?

laeubi:
> ignore the background, it's my wallpaper

In such case I tend to open a text editor to get a white screen for the screenshots :-)

vogella:
Here is another example with cut-off text in a shell using the notification API (which basically it display a shell with some content)

image

SWT Snippet (requires org.eclipse.jface.notifications as MANIFEST.MF dependency:

````

import org.eclipse.jface.notifications.AbstractNotificationPopup;
import org.eclipse.jface.widgets.WidgetFactory;
import org.eclipse.swt.SWT;
import org.eclipse.swt.widgets.Composite;

public class Snippet081NotificationPopup {

	private Display display = new Display();
	private Shell shell = WidgetFactory.shell(SWT.NONE).layout(new FillLayout(SWT.VERTICAL)).create(display);

		new Snippet081NotificationPopup().start();
	}

	private void start() {

		shell.open();

		NotificationFactory.showNotification(shell,
				"This is text 1. This is text 2. This is text 3. This is text 4 This is text 5. This is text 6. This is text 7. This is text 8");
		while (!shell.isDisposed()) {
				display.sleep();
		}
		display.dispose();
	}

	private class NotificationFactory {

		public static void showNotification(Shell shell, String message) {
			NotificationPopUp popup = new NotificationFactory.NotificationPopUp(shell.getDisplay(), shell, message);
			popup.open();
		}

		private static class NotificationPopUp extends AbstractNotificationPopup {

			private String fText;

			public NotificationPopUp(Display display, Shell shell, String text) {
				super(display);
				this.fText = text;
				setParentShell(shell);
				setDelayClose(20000);
			}

			@Override
			protected String getPopupShellTitle() {
				return "Warnung";
			}

			@Override
			protected void createContentArea(Composite parent) {
				Label label = new Label(parent, SWT.WRAP);
				label.setBackground(parent.getBackground());
				label.setText(fText);
				GridData gridData = new GridData(GridData.FILL_BOTH);
//			adjustLabelHeight(parent.getDisplay(), gridData);
				label.setLayoutData(gridData);
			}
//
//		private void adjustLabelHeight(Display display, GridData gridData) {
//			int dpi = display.getDPI().x; // Assume uniform DPI scaling for both x and y
//			// 125 % scaling
//			if (dpi >= 120 && dpi < 144) { // 150% scaling is typically around 144 DPI
//				gridData.heightHint = 80;
//			}
//			// Zwischen 150 % und 200 % scaling
//			if (dpi >= 144 && dpi < 192) { // 150% scaling is typically around 144 DPI
//				gridData.heightHint = 100;
//			}
//		}
		}
	}
}

////

**Comment by vogella** [comment]
I moved the shell not correctly sized example to https://github.com/eclipse-platform/eclipse.platform.swt/issues/62

**Comment by fedejeanne** [comment]
@amartya4256 just to be on the same page: the error I reported was there **before** this PR.

This behavior is in the current `master` branch (I switch from the 100% monitor on the left to the 200% monitor on the right):

Snippet219_multi_monitor

**Comment by amartya4256** [comment]
> @amartya4256 just to be on the same page: the error I reported was there **before** this PR.
> 
> This behavior is in the current `master` branch (I switch from the 100% monitor on the left to the 200% monitor on the right):
> 
> Snippet219_multi_monitor [ Snippet219_multi_monitor ](https://private-user-images.githubusercontent.com/2205684/341044204-58cc8d43-2a94-4d49-bc54-25df1fbafe09.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTg4ODIyODQsIm5iZiI6MTcxODg4MTk4NCwicGF0aCI6Ii8yMjA1Njg0LzM0MTA0NDIwNC01OGNjOGQ0My0yYTk0LTRkNDktYmM1NC0yNWRmMWZiYWZlMDkuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI0MDYyMCUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNDA2MjBUMTExMzA0WiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9YjE5ZGQ1MzFkYWE2MzRkODQ0MjcwZTJlOGNiZjRiNTRjZjM0NDA1YjcxMWJlZjdlNTg1MjM0YTIzNjRmZTI1NyZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QmYWN0b3JfaWQ9MCZrZXlfaWQ9MCZyZXBvX2lkPTAifQ.6sCNMqVcQn4QAeVMXLSvA_4N5zT5wg8-3K7mATue6aQ) [ ](https://private-user-images.githubusercontent.com/2205684/341044204-58cc8d43-2a94-4d49-bc54-25df1fbafe09.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTg4ODIyODQsIm5iZiI6MTcxODg4MTk4NCwicGF0aCI6Ii8yMjA1Njg0LzM0MTA0NDIwNC01OGNjOGQ0My0yYTk0LTRkNDktYmM1NC0yNWRmMWZiYWZlMDkuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI0MDYyMCUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNDA2MjBUMTExMzA0WiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9YjE5ZGQ1MzFkYWE2MzRkODQ0MjcwZTJlOGNiZjRiNTRjZjM0NDA1YjcxMWJlZjdlNTg1MjM0YTIzNjRmZTI1NyZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QmYWN0b3JfaWQ9MCZrZXlfaWQ9MCZyZXBvX2lkPTAifQ.6sCNMqVcQn4QAeVMXLSvA_4N5zT5wg8-3K7mATue6aQ)

You can retest now.

File: bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/graphics/RegionWin32Tests.java [hunk 3]
```diff
@@ -0,0 +1,36 @@
...
+ *     Yatta Solutions - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.swt.graphics;
+
+import static org.junit.Assert.assertEquals;
+
+import org.eclipse.swt.internal.*;
+import org.junit.*;
+
+public class RegionWin32Tests extends Win32AutoscaleTestBase {
```
akoch-yatta, line 21: Migrate the test to junit 5, e.g. see GCWin32Tests and reduce the class visibility to package protected.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Region.java
```diff
@@ -603,8 +644,39 @@ void translateInPixels (int x, int y) {
...
 	if (isDisposed()) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
 	if (pt == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
-	pt = DPIUtil.autoScaleUp(pt);
-	translateInPixels(pt.x, pt.y);
+	operations.add(new Operation(OperationType.TRANSLATE, ArgumentType.POINT, new Point(pt.x, pt.y)));
+	Point p = DPIUtil.autoScaleUp(pt, initialZoom);
+	translateInPixels(p.x, p.y);
+}
+
+private HashMap<Integer, Long> handleMap = new HashMap<>();
```
akoch-yatta, line 652: Please move that up to the other attributes

bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/graphics/RegionWin32Tests.java
amartya4256: Done I guess

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Region.java
amartya4256: Okay

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_widgets_Shell.java
```diff
@@ -727,7 +727,7 @@ public void a_test_setRegion() {
 	Shell shell2 = new Shell(display, SWT.NO_TRIM);
 	assertNull(":d:", shell2.getRegion());
 	shell2.setRegion(region);
-	assertEquals(":e:", region.handle, shell2.getRegion().handle);
+	assertTrue(":e:", region.equals(shell2.getRegion()));
```
HeikoKlare, line 730: Why not just use `assertEquals` for the regions?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Region.java
```diff
@@ -616,7 +676,121 @@ public void translate (Point pt) {
...
+	enum OperationType {
+		ADD, SUBTRACT, INTERSECT, TRANSLATE
+	}
+	enum ArgumentType {
+		REGION, RECTANGLE, POINTS, POINT
+	}
+
+	OperationType operationType;
+	ArgumentType argumentType;
+	Object data;
```
HeikoKlare, line 692: These should probably be private.

HeikoKlare (changes requested):
I like the proposed way of scaling regions by tracking the operations and applying them for a different scaling on demand rather than scaling a region itself.
However, I now see a duplication of all operations on regions, once implemented in the `Region` class and once in the `Operation` class. I would expect that everything related to executing an operation is moved to the `Operation` class and thus completely removed from the `Region` class. So I would expect all the `add`/`addInPixels`, `translate`/... methods to only just use the `Operation` class for applying an operation.

bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/graphics/RegionWin32Tests.java
fedejeanne: It's done, you're using `jupiter` packages which is JUnit**5**

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Region.java
amartya4256: Yes right

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_widgets_Shell.java
amartya4256: Yes right haha

amartya4256:
@HeikoKlare I have adapted the Region class as per you asked and separated the changes between the 2 commits. Please have a look and let me know if it looks fien to you.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Region.java
```diff
@@ -616,7 +629,158 @@ public void translate (Point pt) {
...
+	}
+	enum ArgumentType {
+		REGION, RECTANGLE, POINTS, POINT
+	}
+
+	private OperationType operationType;
+	private ArgumentType argumentType;
+	private Object data;
+
+	Operation(OperationType operationType, ArgumentType argumentType, Object data) {
```
HeikoKlare, line 647: There is an implicit dependency between the value of `argumentType` and the type of `data`, which is not explicitly documented but just comes from how `data` is cast depending on the `argumentType`. This is hard to comprehend and error prone.

The best solution would be to use polymorphism and implement the logic in multiple classes according to the type of argument. It would be okay for me to follow a simpler solution here and simply remove the `ArgumentType` and switch over the type of `data` (and validate in the constructor that `data` is of any valid type). That is not perfect, but definitely better than the current implicit dependency between the method arguments.

HeikoKlare (changes requested):
Overall, the PR looks quite good, but there are two major points that need to be addressed, one functional and one design error, for which you find more detailed information in the comments.
- Functional: operations are not applied to all already created handles
- Design: OO violation by implicit dependency between argument types in the `Operation` class.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Region.java
```diff
@@ -204,11 +197,11 @@ public void add (Region region) {
  */
 public boolean contains (int x, int y) {
 	if (isDisposed()) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
-	return containsInPixels(DPIUtil.autoScaleUp(x), DPIUtil.autoScaleUp(y));
+	return containsInPixels(DPIUtil.autoScaleUp(x, initialZoom), DPIUtil.autoScaleUp(y, initialZoom));
 }
 
 boolean containsInPixels (int x, int y) {
-	return OS.PtInRegion (handle, x, y);
+	return OS.PtInRegion (win32_getHandle(this, initialZoom), x, y);
```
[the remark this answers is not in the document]
amartya4256: Done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Region.java
```diff
@@ -616,7 +629,158 @@ public void translate (Point pt) {
 @Override
 public String toString () {
 	if (isDisposed()) return "Region {*DISPOSED*}";
-	return "Region {" + handle + "}";
+	return "Region " + zoomToHandle;
```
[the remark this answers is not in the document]
amartya4256: Alright

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Region.java
amartya4256: Alright, I have redesigned the Operations implementation. Please have a look again.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Region.java
```diff
@@ -107,13 +104,9 @@ public Region (Device device) {
...
-	addInPixels(DPIUtil.autoScaleUp(pointArray));
-}
-
-void addInPixels (int[] pointArray) {
-	long polyRgn = OS.CreatePolygonRgn(pointArray, pointArray.length / 2, OS.ALTERNATE);
-	OS.CombineRgn (handle, handle, polyRgn, OS.RGN_OR);
-	OS.DeleteObject (polyRgn);
+	final Operation operation = new Operation(OperationType.ADD, ArgumentType.POINTS, Arrays.copyOf(pointArray, pointArray.length));
+	operations.add(operation);
+	operation.apply(win32_getHandle(this, initialZoom), initialZoom);
```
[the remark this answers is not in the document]
amartya4256: Extracted and applied for all the handles

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Region.java
```diff
@@ -133,8 +126,11 @@ void addInPixels (int[] pointArray) {
 public void add (Rectangle rect) {
 	if (isDisposed()) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
 	if (rect == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
-	rect = DPIUtil.autoScaleUp(rect);
-	addInPixels(rect.x, rect.y, rect.width, rect.height);
+	final Operation operation = new Operation(OperationType.ADD, ArgumentType.RECTANGLE, new Rectangle(rect.x, rect.y, rect.width, rect.height));
+	operations.add(operation);
+	operation.apply(win32_getHandle(this, initialZoom), initialZoom);
+//	Rectangle r = DPIUtil.autoScaleUp(rect, initialZoom);
+//	addInPixels(r.x, r.y, r.width, r.height);
```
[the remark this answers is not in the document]
amartya4256: Done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Region.java
```diff
@@ -616,7 +623,204 @@ public void translate (Point pt) {
...
-	return "Region {" + handle + "}";
+	return "Region {" + zoomToHandle.entrySet().stream().map(entry -> entry.getValue() + "(zoom:" + entry.getKey() + ")").collect(Collectors.joining(","));
 }
 
+@FunctionalInterface
+private interface OperationStrategy {
+	void apply(Operation operation, long handle, int zoom);
+}
+
+abstract class Operation {
```
HeikoKlare, line 634: Could be defined `private`.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Region.java
```diff
@@ -616,7 +623,204 @@ public void translate (Point pt) {
...
+	OperationStrategy operationStrategy;
```
HeikoKlare, line 635: Better have this field `private` and add a constructor setting the field, which is then used as the super constructor for all subclasses.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Region.java
```diff
@@ -254,7 +242,7 @@ public boolean equals (Object object) {
 	if (this == object) return true;
 	if (!(object instanceof Region)) return false;
 	Region rgn = (Region)object;
-	return handle == rgn.handle;
+	return getHandleForInitialZoom() == win32_getHandle(rgn, initialZoom);
```
HeikoKlare, line 245:
```suggestion
	return getHandleForInitialZoom() == rgn.getHandleForInitialZoom();
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Region.java
```diff
@@ -107,13 +104,8 @@ public Region (Device device) {
...
 	if (isDisposed()) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
 	if (pointArray == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
-	addInPixels(DPIUtil.autoScaleUp(pointArray));
-}
-
-void addInPixels (int[] pointArray) {
-	long polyRgn = OS.CreatePolygonRgn(pointArray, pointArray.length / 2, OS.ALTERNATE);
-	OS.CombineRgn (handle, handle, polyRgn, OS.RGN_OR);
-	OS.DeleteObject (polyRgn);
+	final Operation operation = new OperationWithArray(Operation::add, Arrays.copyOf(pointArray, pointArray.length));
```
HeikoKlare, line 107: That's nice 👍  I was not aware that it is possible to pass a method reference to an abstract method that is properly the bound in the context of the instantiated subclass.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Region.java
```diff
@@ -616,7 +623,204 @@ public void translate (Point pt) {
...
+
+	@Override
+	void intersect(long handle, int zoom) {
+		Rectangle bounds = getScaledRectangle(zoom);
+		intersectInPixels(handle, bounds.x, bounds.y, bounds.width, bounds.height);
+	}
+
+	@Override
+	void translate(long handle, int zoom) {
+		// No implementation required
```
HeikoKlare, line 679: I would prefer to throw an `OperationNotSupportedException` (in all the unimplemented methods), so that in case someone starts using this method, they are informed about the missing implementation rather than doing nothing.

HeikoKlare (commented):
The redesigned implementation looks really good. I only have nipicky comments left.

I have tested the patch and did not find any regressions when using regions (when starting an SDK product with a single zoom value).

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Region.java
amartya4256: Ahh yes, a miss

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Region.java
amartya4256: Done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Region.java
amartya4256: Done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Region.java
amartya4256: It is a check exception and it will change the ssignature of all the methods. I already tried it and I also needed to add the throws in all the client calls. Do we not have an alternative?

HeikoKlare (approved):
Thanks for addressing my comments! The changes look good now.

amartya4256:
> Thanks for addressing my comments! The changes look good now.

Thanks, I have rebased the branch to the master as well.

HeikoKlare:
> I tested the `Snippet219` and when I move from the 100% monitor to the 200% monitor I see that the region (content) is resized but the shell isn't
> 
> 100% image
> 
> 200% (ignore the background, it's my wallpaper) image

I have tested the behavior with the current PR state: it works fine and as expected.

PR #1283, opened 2024-06-17
[win32][fix] apply dpi change to menus attribute for Decorations in windows

akoch-yatta:
**Note:** issue can only occur in windows with flag swt.autoScale.updateOnRuntime set to true

This commit applies dpi changes to all menus in this menus attribute in Decorations as well in windows, so pop-up menus are no longer opened on the wrong location.

## Requires

- [x] https://github.com/eclipse-platform/eclipse.platform.swt/pull/1251

## How to reproduce

1. Have two monitors with different zoom settings
2. Start IDE
3. Move Java Editors on both of the monitors
4. Right-click on multiple positions in the editor on the non-primary monitor -> in some places they are opened at the mouse location in some places not

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/DPIZoomChangeRegistry.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/DPIZoomChangeRegistry.java
@@ -44,6 +44,9 @@ public class DPIZoomChangeRegistry {
 	 * @param scalingFactor factor as division between new zoom and old zoom, e.g. 1.5 for a scaling from 100% to 150%
 	 */
 	public static void applyChange(Widget widget, int newZoom, float scalingFactor) {
+		if (widget == null) {
+			return;
+		}
 		for (Entry<Class<? extends Widget>, DPIZoomChangeHandler> entry : dpiZoomChangeHandlers.entrySet()) {
 			Class<? extends Widget> clazz = entry.getKey();
 			DPIZoomChangeHandler handler = entry.getValue();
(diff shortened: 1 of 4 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/DPIZoomChangeRegistry.java
```diff
@@ -44,6 +44,9 @@ public class DPIZoomChangeRegistry {
 	 * @param scalingFactor factor as division between new zoom and old zoom, e.g. 1.5 for a scaling from 100% to 150%
 	 */
 	public static void applyChange(Widget widget, int newZoom, float scalingFactor) {
+		if (widget == null) {
+			return;
+		}
```
fedejeanne, line 49: Is this really needed? I mean is some call passing `null` as a parameter?

fedejeanne (changes requested):
The code looks OK, just a minor question about the `null`-check.

I tested this and the menu looks like this when most of the shell is on the 200% monitor but I right-click on the side of the 100% monitor:

image

Is this expected? Can it be fixed?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/DPIZoomChangeRegistry.java
akoch-yatta: Yes, decorations.menus contains null values in some scenarios. And it is definitely a source for possible NPE in the future, so I think fetching nulls here is better

akoch-yatta:
> The code looks OK, just a minor question about the `null`-check.
> 
> I tested this and the menu looks like this when most of the shell is on the 200% monitor but I right-click on the side of the 100% monitor:
> 
> image
> 
> Is this expected? Can it be fixed?

In most cases it will always look as expected now. The only exception is, when you have two monitor with different zoom and stretch **one** Shell over **both** monitors. When the menus will always be scaled according to  the zoom the Shell is assigned to from Windows.

fedejeanne (approved):

PR #1284, opened 2024-06-18
Scaling combo text according to zoom level

ShahzaibIbrahim:
Registering the comboDpiChange handler and applying change manually to
 text, image and arrow

Contributes to https://github.com/eclipse-platform/eclipse.platform.swt/issues/62 and https://github.com/eclipse-platform/eclipse.platform.swt/issues/127

## HOW TO TEST **(Windows only!)**

- Run the `CustomControlExample` with the following **VM Arguments**
```
-Dswt.autoScale=quarter
```
- Switch to the tab "CCombo"
- See if scales properly

## EXPECTED BEHAVIOUR

Combo list, text and arrow should be properly scaled.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CCombo.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CCombo.java
@@ -2025,4 +2025,20 @@ public boolean traverse(int event){
 	}
 	return super.traverse(event);
 }
+
+/**
+ * The method accepts a combo and a callback which takes
+ * all the child of the CCombo as the argument and executes it.
+ * All children are refreshed after the execution of the callback.
+ *
+ * @noreference This method is not intended to be referenced by clients.
+ * @param combo the Combo to get the children widget from
+ * @param childUpdater the callback which works with the child widgets
+ */
+public static void updateAndRefreshChildren(CCombo combo, Consumer<Widget> childUpdater) {
+	childUpdater.accept(combo.text);
+	childUpdater.accept(combo.list);
+	childUpdater.accept(combo.arrow);
+	childUpdater.accept(combo.popup);
+}
 }
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/CommonWidgetsDPIChangeHandlers.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/CommonWidgetsDPIChangeHandlers.java
@@ -33,6 +33,7 @@ public class CommonWidgetsDPIChangeHandlers {
 	public static void registerCommonHandlers() {
 		DPIZoomChangeRegistry.registerHandler(CommonWidgetsDPIChangeHandlers::handleItemDPIChange, Item.class);
 		DPIZoomChangeRegistry.registerHandler(CommonWidgetsDPIChangeHandlers::handleStyledTextDPIChange, StyledText.class);
+		DPIZoomChangeRegistry.registerHandler(CommonWidgetsDPIChangeHandlers::handleCComboDPIChange, CCombo.class);
 	}
 
 	private static void handleItemDPIChange(Widget widget, int newZoom, float scalingFactor) {
@@ -51,9 +52,16 @@ public class CommonWidgetsDPIChangeHandlers {
 			return;
 		}
 
+
 		StyledText.updateAndRefreshCarets(styledText, caretToRefresh -> {
 			DPIZoomChangeRegistry.applyChange(caretToRefresh, newZoom, scalingFactor);
 			Caret.win32_setHeight(caretToRefresh, styledText.getLineHeight());
 		});
+	}
+	private static void handleCComboDPIChange(Widget widget, int newZoom, float scalingFactor) {
+		if (!(widget instanceof CCombo combo)) {
+			return;
+		}
+		CCombo.updateAndRefreshChildren(combo, childWidget -> DPIZoomChangeRegistry.applyChange(childWidget, newZoom, scalingFactor));
 	}
 }


fedejeanne (changes requested):
@ShahzaibIbrahim the code looks OK but I tested and found out that the elements of the list are displayed far away from the combo.

image

Can you please look into that?

fedejeanne (dismissed):
I tested it and it looks fine. I just have a question regarding the moving of the class outside of the package `internal`

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/custom/CommonWidgetsDPIChangeHandlers.java
```diff
@@ -11,9 +11,10 @@
  * Contributors:
  *     Yatta Solutions - initial API and implementation
  *******************************************************************************/
-package org.eclipse.swt.internal;
+package org.eclipse.swt.custom;
```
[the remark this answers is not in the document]
ShahzaibIbrahim: Since we were using some of the package protected fields (also will use them in upcoming PRs) and also we wanted to move this new class to win32 packages anyway.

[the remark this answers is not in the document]
fedejeanne: This class doesn't add any value to "normal" users.
Move it back to internal and change the handleCComboDPIChange instead so that it delegates everything to CCombo (1 call) instead of doing 4 calls and passing the fields of CCombo (that's a code smell: https://refactoring.guru/smells/inappropriate-intimacy)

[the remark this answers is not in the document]
akoch-yatta: What should it delegate to CCombo? CCombo is a common widget for all OS, we e.g. cannot use the windows specifics like the DPIZoomChangeRegistry in CCombo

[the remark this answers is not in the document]
ShahzaibIbrahim: As discussed in the meeting, we have delegated everything to CCombo by creating adapt method and passing the lamba function to apply changes internally in CCombo class, @fedejeanne @akoch-yatta

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/CommonWidgetsDPIChangeHandlers.java
```diff
@@ -43,4 +45,12 @@ private static void handleItemDPIChange(Widget widget, int newZoom, float scalin
 			item.setImage(image);
 		}
 	}
-}
+	private static void handleCComboDPIChange(Widget widget, int newZoom, float scalingFactor) {
+		if (!(widget instanceof CCombo)) {
+			return;
+		}
+		CCombo combo = (CCombo) widget;
```
akoch-yatta, line 52:
```suggestion
		if (!(widget instanceof CCombo combo)) {
			return;
		}
```
a bit shorter variant

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CCombo.java
```diff
@@ -2025,4 +2025,13 @@ public boolean traverse(int event){
 	}
 	return super.traverse(event);
 }
+/**
+ * @since 3.126
```
akoch-yatta, line 2029: some javadoc remarks:
- I think the since version must be 3.127
- the javadoc should contain a description what the method is doing (passing all widgets managed by the ccombo to the specified consumer)
- I think the method name should be more explanatory, but currently dont have a good idea myself

akoch-yatta: and what was the argument of static vs. not static again? @fedejeanne @HeikoKlare

HeikoKlare: > * I think the method name should be more explanatory, but currently dont have a good idea myself

We may align with the name for `StyledText` used in #1294, which is `updateAndRefreshCarets`, so maybe more general `updateAndRefreshChildren`?
In addition, the parameter `consumer` should have a more specific name like `childUpdater`.

> and what was the argument of static vs. not static again?

Non-static would mean that this becomes public API. I would propose to add a `@noreference` and keep it static to not have it as part of public API. Then you can also remove the `@since` tag.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/CommonWidgetsDPIChangeHandlers.java
```diff
@@ -43,4 +45,12 @@ private static void handleItemDPIChange(Widget widget, int newZoom, float scalin
...
+
+		CCombo.adapt(combo, w -> DPIZoomChangeRegistry.applyChange(w, newZoom, scalingFactor));
```
HeikoKlare, line 54: Would be good to use a more specific name than `w`, like `childWidget`.

HeikoKlare (commented):
In general, the changes look good. The comments of Andreas and me should be rather easy to address.

HeikoKlare:
@fedejeanne please check if you still request changes or may drop your change request

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CCombo.java
```diff
@@ -2025,4 +2025,20 @@ public boolean traverse(int event){
...
 }
+
+/**
+ * The method accepts a combo and a callback which takes
+ * all the child of the CCombo as the argument and executes it.
+ * All children are refreshed after the execution of the callback.
+ *
+ * @noreference This method is not intended to be referenced by clients.
+ * @param combo
+ * @param childUpdater
```
HeikoKlare, line 2036: Please add documentation for the parameters

ShahzaibIbrahim: @HeikoKlare Added!

fedejeanne (approved):

Issue #1288, opened 2024-06-19
Exception when creating GC from Image inside a thread on Windows


ddscharfe:
**Describe the bug**
A `GC` can no longer be created from an `Image` inside a thread on Windows. This used to work in older versions and still works under Linux. According to my understanding of the `GC` javadoc this is a valid use case, but I might be wrong about that.

In commit f809372 (see #62 and #127) [org.eclipse.swt.graphics.Image](https://github.com/eclipse-platform/eclipse.platform.swt/blob/f80937230f0caf83f0c38fd6536654b301b8934d/bundles/org.eclipse.swt/Eclipse%20SWT/win32/org/eclipse/swt/graphics/Image.java), line 2156
`data.font = device.systemFont;`
was changed to
`data.font = Font.win32_new(device.getSystemFont(), DPIUtil.getNativeDeviceZoom());`

`device.getSystemFont()` leads to calling `Display::checkDevice()` which asserts that the current thread is the display thread, which is not the case if the GC is created from another thread.

**To Reproduce**
```Java
import org.eclipse.swt.graphics.GC;
import org.eclipse.swt.graphics.Image;
import org.eclipse.swt.widgets.Display;

public class Test {

	public static void main(String[] args) {
		new Display();

		Thread thread = new Thread(() -> {
			Image image = new Image(null, 100, 100);
			GC gc = new GC(image); // <-- throws SWTException
			gc.dispose();
			image.dispose();
		});
		thread.start();

		try {
			thread.join();
		} catch (InterruptedException e) {
			e.printStackTrace();
		}
	}
}
```

**Expected behavior**
According to the JavaDoc of `GC`, I'd expect no exception, as long as the drawable (which is an `Image`) is created in the same thread.

**Stack Trace**
```
Exception in thread "Thread-0" org.eclipse.swt.SWTException: Invalid thread access
	at org.eclipse.swt.SWT.error(SWT.java:4922)
	at org.eclipse.swt.SWT.error(SWT.java:4837)
	at org.eclipse.swt.SWT.error(SWT.java:4808)
	at org.eclipse.swt.widgets.Display.error(Display.java:1286)
	at org.eclipse.swt.widgets.Display.checkDevice(Display.java:863)
	at org.eclipse.swt.widgets.Display.getPrimaryMonitor(Display.java:2245)
	at org.eclipse.swt.widgets.Display.getSystemFont(Display.java:2459)
[...]

PR #1289, opened 2024-06-19
Fix for #1288: No exception when creating GC with image from another thread

akoch-yatta:
This commit fixes an issue when creating a GC from an Image using a device bound to another thread as the GC is currently created in.

### DIFF
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_GC.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_GC.java
@@ -23,8 +23,11 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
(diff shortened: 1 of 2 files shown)


ddscharfe:
Tested under Windows 11, I don't get an exception anymore

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_GC.java
```diff
@@ -767,6 +770,31 @@ public void test_drawLine_noSingularitiesIn45DregreeRotation() {
...
+		} catch(Exception e) {
+			exceptionReference.set(e);
+		}
+	});
+	thread.start();
+	try {
+		thread.join();
+	} catch (InterruptedException e) {
+		fail("Thread was interrupted");
+	}
```
fedejeanne, line 794: No need to catch this one, throw it

fedejeanne (changes requested):

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_GC.java
akoch-yatta: Done

fedejeanne (approved):

PR #1294, opened 2024-06-21
Multi zoom level support for caret

amartya4256:
## Addressed issues

## Requires
* [x] https://github.com/eclipse-platform/eclipse.platform.swt/pull/1247

## Description

This PR allows the caret to scale on a DPI change event on styled text. It uses the line height of the styled text to set on caret. The height and width field in the caret are storing data in points now and not in pixels. We do so to avail the scaling refernece. Also we use SWT.DEFAULT to ignore setting of width. It is useful when we set the new height on DPI change since the preferred width of the caret is loadeed from preferences on the Caret creation.

Contributes to #62 and #127

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/StyledText.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/StyledText.java
@@ -19,6 +19,7 @@ package org.eclipse.swt.custom;
 
 
 import java.util.*;
+import java.util.function.*;
 import java.util.stream.*;
 
 import org.eclipse.swt.*;
@@ -10857,6 +10858,27 @@ void updateSelection(int startOffset, int replacedLength, int newLength) {
 		}).flatMapToInt(p -> IntStream.of(p.x, p.y - p.x))
 		.toArray(), true, false);
 	setCaretLocations();
+}
+
+/**
+ * The method accepts a StyledText and a callback which takes
+ * all the carets of the StyledText as the argument and executes it.
+ * The caret is refreshed after the execution of the callback.
+ *
+ * @param styledText the StyledText to get the carets from
+ * @param caretUpdater the callback which works with the carets
+ *
+ * @noreference This method is not intended to be referenced by clients.
+ */
+public static void updateAndRefreshCarets(StyledText styledText, Consumer<Caret> caretUpdater) {
+	caretUpdater.accept(styledText.getCaret());
+	caretUpdater.accept(styledText.defaultCaret);
+	for (Caret caret : styledText.carets) {
+		caretUpdater.accept(caret);
+	}
+	styledText.updateCaretVisibility();
+	styledText.setCaretLocations();
+
 }
 
 }
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/CommonWidgetsDPIChangeHandlers.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/CommonWidgetsDPIChangeHandlers.java
@@ -13,6 +13,7 @@
  *******************************************************************************/
 package org.eclipse.swt.internal;
 
+import org.eclipse.swt.custom.*;
 import org.eclipse.swt.graphics.*;
 import org.eclipse.swt.widgets.*;
 
@@ -31,6 +32,7 @@ public class CommonWidgetsDPIChangeHandlers {
 
 	public static void registerCommonHandlers() {
 		DPIZoomChangeRegistry.registerHandler(CommonWidgetsDPIChangeHandlers::handleItemDPIChange, Item.class);
+		DPIZoomChangeRegistry.registerHandler(CommonWidgetsDPIChangeHandlers::handleStyledTextDPIChange, StyledText.class);
 	}
 
 	private static void handleItemDPIChange(Widget widget, int newZoom, float scalingFactor) {
@@ -42,5 +44,16 @@ public class CommonWidgetsDPIChangeHandlers {
 		if (image != null) {
 			item.setImage(image);
 		}
+	}
+
+	private static void handleStyledTextDPIChange(Widget widget, int newZoom, float scalingFactor) {
+		if (!(widget instanceof StyledText styledText)) {
+			return;
+		}
+
+		StyledText.updateAndRefreshCarets(styledText, caretToRefresh -> {
+			DPIZoomChangeRegistry.applyChange(caretToRefresh, newZoom, scalingFactor);
+			Caret.win32_setHeight(caretToRefresh, styledText.getLineHeight());
+		});
 	}
 }
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Caret.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Caret.java
@@ -127,15 +127,15 @@ public Rectangle getBounds () {
 Rectangle getBoundsInPixels () {
 	if (image != null) {
 		Rectangle rect = image.getBoundsInPixels ();
-		return new Rectangle (x, y, rect.width, rect.height);
+		return new Rectangle (getXInPixels(), getYInPixels(), rect.width, rect.height);
 	}
 	if (width == 0) {
 		int [] buffer = new int [1];
 		if (OS.SystemParametersInfo (OS.SPI_GETCARETWIDTH, 0, buffer, 0)) {
-			return new Rectangle (x, y, buffer [0], height);
+			return new Rectangle (getXInPixels(), getYInPixels(), buffer [0], getHeightInPixels());
 		}
 	}
-	return new Rectangle (x, y, width, height);
+	return new Rectangle (getXInPixels(), getYInPixels(), getWidthInPixels(), getHeightInPixels());
 }
 
 /**
@@ -185,10 +185,6 @@ public Image getImage () {
  */
 public Point getLocation () {
 	checkWidget();
-	return DPIUtil.scaleDown(getLocationInPixels(), getZoom());
-}
-
-Point getLocationInPixels () {
 	return new Point (x, y);
 }
 
@@ -230,10 +226,26 @@ Point getSizeInPixels () {
 	if (width == 0) {
 		int [] buffer = new int [1];
 		if (OS.SystemParametersInfo (OS.SPI_GETCARETWIDTH, 0, buffer, 0)) {
-			return new Point (buffer [0], height);
+			return new Point (buffer [0], getHeightInPixels());
 		}
 	}
-	return new Point (width, height);
+	return new Point (getWidthInPixels(), getHeightInPixels());
+}
+
+private int getWidthInPixels() {
+	return DPIUtil.autoScaleUp(width, getZoom());
+}
+
+private int getHeightInPixels() {
+	return DPIUtil.autoScaleUp(height, getZoom());
+}
+
+private int getXInPixels() {
+	return DPIUtil.autoScaleUp(x, getZoom());
+}
+
+private int getYInPixels() {
+	return DPIUtil.autoScaleUp(y, getZoom());
 }
 
 /**
@@ -293,7 +305,7 @@ void killFocus () {
 void move () {
 	moved = false;
 	setCurrentCaret(this);
-	if (!OS.SetCaretPos (x, y)) return;
+	if (!OS.SetCaretPos (getXInPixels(), getYInPixels())) return;
 	resizeIME ();
 }
 
@@ -354,15 +366,15 @@ void resize () {
 	long hwnd = parent.handle;
 	OS.DestroyCaret ();
 	long hBitmap = image != null ? Image.win32_getHandle(image, getZoom()) : 0;
-	int width = this.width;
-	if (image == null && width == 0) {
+	int widthInPixels = this.getWidthInPixels();
+	if (image == null && widthInPixels == 0) {
 		int [] buffer = new int [1];
 		if (OS.SystemParametersInfo (OS.SPI_GETCARETWIDTH, 0, buffer, 0)) {
-			width = buffer [0];
+			widthInPixels = buffer [0];
 		}
 	}
-	OS.CreateCaret (hwnd, hBitmap, width, height);
-	OS.SetCaretPos (x, y);
+	OS.CreateCaret (hwnd, hBitmap, widthInPixels, getHeightInPixels());
+	OS.SetCaretPos (getXInPixels(), getYInPixels());
 	OS.ShowCaret (hwnd);
 	move ();
 }
@@ -395,11 +407,6 @@ void restoreIMEFont () {
  */
 public void setBounds (int x, int y, int width, int height) {
 	checkWidget();
-	int zoom = getZoom();
-	setBoundsInPixels(DPIUtil.autoScaleUp(x, zoom), DPIUtil.autoScaleUp(y, zoom), DPIUtil.autoScaleUp(width, zoom), DPIUtil.autoScaleUp(height, zoom));
-}
-
-void setBoundsInPixels (int x, int y, int width, int height) {
 	boolean samePosition = this.x == x && this.y == y;
 	boolean sameExtent = this.width == width && this.height == height;
 	if (samePosition && sameExtent && isCurrentCaret()) return;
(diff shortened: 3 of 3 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/custom/CommonWidgetsDPIChangeHandlers.java
```diff
@@ -11,9 +11,11 @@
  * Contributors:
  *     Yatta Solutions - initial API and implementation
  *******************************************************************************/
-package org.eclipse.swt.internal;
+package org.eclipse.swt.custom;
```
fedejeanne, line 14: Probably OK but let's wait for this discussion to be solved: https://github.com/eclipse-platform/eclipse.platform.swt/pull/1284/files#r1647473087

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/custom/CommonWidgetsDPIChangeHandlers.java
```diff
@@ -43,4 +46,25 @@ private static void handleItemDPIChange(Widget widget, int newZoom, float scalin
...
 		}
 	}
+
+	private static void handleStyledTextDPIChange(Widget widget, int newZoom, float scalingFactor) {
+		if (!(widget instanceof StyledText)) {
+			return;
+		}
+		StyledText styledText = (StyledText) widget;
+
+
```
fedejeanne, line 56:
```suggestion

```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Caret.java
```diff
@@ -354,15 +362,15 @@ void resize () {
 	long hwnd = parent.handle;
 	OS.DestroyCaret ();
 	long hBitmap = image != null ? Image.win32_getHandle(image, getZoom()) : 0;
-	int width = this.width;
+	int width = this.getWidthInPixels();
```
fedejeanne, line 365:
```suggestion
	int widthInPixels = this.getWidthInPixels();
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Caret.java
```diff
@@ -414,6 +414,14 @@ void setBoundsInPixels (int x, int y, int width, int height) {
 		resized = true;
 		if (isVisible && hasFocus ()) resize ();
 	}
+	setBoundsInPixels(DPIUtil.autoScaleUp(x, zoom), DPIUtil.autoScaleUp(y, zoom), sameExtent);
+}
+
+void setBoundsInPixels (int x, int y, boolean sameExtent) {
```
fedejeanne, line 420:
```suggestion
private void setBoundsInPixels (int xInPixels, int yInPixels, boolean sameExtent) {
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Caret.java
```diff
@@ -431,25 +439,21 @@ void setBoundsInPixels (int x, int y, int width, int height) {
...
-	setBoundsInPixels (rect.x, rect.y, rect.width, rect.height);
+	setBounds(rect.x, rect.y, rect.width, rect.height);
 }
 
 void setFocus () {
 	long hwnd = parent.handle;
 	long hBitmap = 0;
 	if (image != null) hBitmap = Image.win32_getHandle(image, getZoom());
-	int width = this.width;
+	int width = this.getWidthInPixels();
```
fedejeanne, line 449:
```suggestion
	int widthInPixels = this.getWidthInPixels();
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Caret.java
```diff
@@ -544,8 +548,8 @@ public void setLocation (int x, int y) {
 }
 
 void setLocationInPixels (int x, int y) {
```
fedejeanne, line 550:
```suggestion
private void setLocationInPixels (int xInPixels, int yInPixels) {
```

fedejeanne (changes requested):
Some minor suggestions to improve readability.

Also, please check the whole class and see if you can:
1. Reduce the visibility of some of the methods with `package protected` (default) visibility
2. Rename some other parameters in methods `xyzInPixels` so they are also called `xInPixels` / `yInPixels` etc

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/custom/CommonWidgetsDPIChangeHandlers.java
amartya4256: orite

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/custom/CommonWidgetsDPIChangeHandlers.java
amartya4256: okay

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Caret.java
amartya4256: Sure

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Caret.java
amartya4256: Okay

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Caret.java
fedejeanne: Done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/custom/CommonWidgetsDPIChangeHandlers.java
```diff
@@ -43,4 +46,24 @@ private static void handleItemDPIChange(Widget widget, int newZoom, float scalin
...
 	}
+
+	private static void handleStyledTextDPIChange(Widget widget, int newZoom, float scalingFactor) {
+		if (!(widget instanceof StyledText)) {
+			return;
+		}
+		StyledText styledText = (StyledText) widget;
+
+		DPIZoomChangeRegistry.applyChange(styledText.getCaret(), newZoom, scalingFactor);
+		styledText.getCaret().setSize(SWT.DEFAULT, styledText.getLineHeight());
```
fedejeanne, line 57: Isn't this superfluous? I mean the `for`-loop below probably does it too, right?

fedejeanne: Ok

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/CommonWidgetsDPIChangeHandlers.java
```diff
@@ -43,4 +46,15 @@ private static void handleItemDPIChange(Widget widget, int newZoom, float scalin
 			item.setImage(image);
 		}
 	}
+
+	private static void handleStyledTextDPIChange(Widget widget, int newZoom, float scalingFactor) {
+		if (!(widget instanceof StyledText styledText)) {
+			return;
+		}
+
+		StyledText.updateAndRefreshCarets(styledText, x -> {
```
HeikoKlare, line 55: Please use a better variable name than `x`, like `caretToRefresh`.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Caret.java
```diff
@@ -414,6 +414,14 @@ void setBoundsInPixels (int x, int y, int width, int height) {
 		resized = true;
 		if (isVisible && hasFocus ()) resize ();
 	}
+	setBoundsInPixels(DPIUtil.autoScaleUp(x, zoom), DPIUtil.autoScaleUp(y, zoom), sameExtent);
+}
+
+private void setBoundsInPixels (int xInPixels, int yInPixels, boolean sameExtent) {
+	boolean samePosition = this.xInPixels == xInPixels && this.yInPixels == yInPixels;
+	if (samePosition && sameExtent && isCurrentCaret()) return;
```
HeikoKlare, line 422: Isn't this check more expensive than simply setting the (unchanged) values? Can't the values just be xet in the `setBounds` method?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Caret.java
```diff
@@ -414,6 +414,14 @@ void setBoundsInPixels (int x, int y, int width, int height) {
 		resized = true;
 		if (isVisible && hasFocus ()) resize ();
 	}
+	setBoundsInPixels(DPIUtil.autoScaleUp(x, zoom), DPIUtil.autoScaleUp(y, zoom), sameExtent);
+}
+
+private void setBoundsInPixels (int xInPixels, int yInPixels, boolean sameExtent) {
```
HeikoKlare, line 420: The method name is misleading. It does not set the bounds but only the x/y values (not width/height).

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Caret.java
```diff
@@ -396,15 +404,7 @@ void restoreIMEFont () {
...
 	int zoom = getZoom();
-	setBoundsInPixels(DPIUtil.autoScaleUp(x, zoom), DPIUtil.autoScaleUp(y, zoom), DPIUtil.autoScaleUp(width, zoom), DPIUtil.autoScaleUp(height, zoom));
-}
-
-void setBoundsInPixels (int x, int y, int width, int height) {
-	boolean samePosition = this.x == x && this.y == y;
 	boolean sameExtent = this.width == width && this.height == height;
-	if (samePosition && sameExtent && isCurrentCaret()) return;
-	this.x = x;
-	this.y = y;
```
HeikoKlare, line 407: Why must this be deferred to the `setBoundsInPixels` method call at the method's end instead of setting those values here?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Caret.java
```diff
@@ -396,15 +404,7 @@ void restoreIMEFont () {
 public void setBounds (int x, int y, int width, int height) {
 	checkWidget();
 	int zoom = getZoom();
-	setBoundsInPixels(DPIUtil.autoScaleUp(x, zoom), DPIUtil.autoScaleUp(y, zoom), DPIUtil.autoScaleUp(width, zoom), DPIUtil.autoScaleUp(height, zoom));
-}
-
-void setBoundsInPixels (int x, int y, int width, int height) {
-	boolean samePosition = this.x == x && this.y == y;
 	boolean sameExtent = this.width == width && this.height == height;
-	if (samePosition && sameExtent && isCurrentCaret()) return;
```
HeikoKlare, line 405: Why is it reasonable to remove this check? Now every time nothing has changed the caret will be resized (running into the else block of the subsequent condition). Is that correct?

HeikoKlare (changes requested):
I have some comments, the most important one about a non-API-conforming change that needs to be addressed.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/CommonWidgetsDPIChangeHandlers.java
amartya4256: Alright

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Caret.java
amartya4256: Its basically the previous implementation. The only thing that is different is that we extracted the setting of width and height from here because earlier they were set in pixels and now we set them in points, hence they are set in the setBounds method instead of setBoundsInPixels. And about the check since it was there before, I haven't bothered changing the logic. However we still need to check for extent and currentCaret as per the previous logic since they can be same even if the position is not the same. Overall this check is O(1),so I wont call it expensive.

Here, we are not setting the x and y in setBounds methods cause they are in pixels. However, another solution would be to also store x and y in points and then we can get rid of all the inPixels method. But the checks should still stay.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Caret.java
amartya4256: I think it looks a bit weird on the git diff. It all shows up like that because I changed the parameters to this method (removed width and height). Now that height and width are set in points, we dont need them here. However, if you look in the method down below: https://github.com/eclipse-platform/eclipse.platform.swt/pull/1294/files#r1656807098, the check is still there.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Caret.java
amartya4256: It is set in setBoundsInPixels where it was set earlier indeed.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Caret.java
amartya4256: So now we set height, width, x and y all in pixels

HeikoKlare (commented):
The changes look good now. In particular, I did not find any faulty direct access to x, y, width and height not being replaced by the new value-in-pixel getters.
I have also successfully tested the changes in an SDK product at 100%, 150% and 200%.

I have only minor remarks left.

HeikoKlare:
@fedejeanne anything left to do from your side?

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/StyledText.java
```diff
@@ -10859,4 +10860,25 @@ void updateSelection(int startOffset, int replacedLength, int newLength) {
...
 
+/**
+ * The method accepts a StyledText and a callback which takes
+ * all the carets of the StyledText as the argument and executes it.
+ * The caret is refreshed after the execution of the callback.
+ *
+ * @param styledText the StyledText to get the carets from
+ * @param consumer the callback which works with the carets
+ *
+ * @since 3.127
```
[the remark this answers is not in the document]
amartya4256: Alright

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/StyledText.java
```diff
@@ -10859,4 +10860,25 @@ void updateSelection(int startOffset, int replacedLength, int newLength) {
...
+ */
+public static void updateAndRefreshCarets(StyledText styledText, Consumer<Caret> consumer) {
```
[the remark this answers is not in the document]
amartya4256: Done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Caret.java
```diff
@@ -230,10 +226,26 @@ Point getSizeInPixels () {
...
+
+private int getWidthInPixels() {
+	return DPIUtil.autoScaleUp(width, getZoom());
+}
+
+private int getHeightInPixels() {
+	return DPIUtil.autoScaleUp(height, getZoom());
+}
+
+private int getXinPixels() {
```
fedejeanne, line 243:
```suggestion
private int getXInPixels() {
```

fedejeanne: Last one and then it's ready to merge

fedejeanne (changes requested):

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Caret.java
amartya4256: Done

fedejeanne (approved):

PR #1296, opened 2024-06-21
[Build] Reduce max JVM heap-size to avoid executor kills

HannesWell:


Jenkinsfile
```diff
@@ -96,7 +96,7 @@ pipeline {
 		maven 'apache-maven-latest'
 	}
 	environment {
-		MAVEN_OPTS = "-Xmx4G"
+		MAVEN_OPTS = "-Xmx3G"
```
merks, line 99: Still a very big number on a node limited to 4g.

HannesWell: That's right.
I'll experiment with lower numbers.

In general I think its is better to be at the lower limit than at the upper, since the GC can maybe try hard to free heap memory. And OOMErrors of the JVM can be analyzed through a heap-dump more easy (to find possible memory 'leaks'), which is probably impossible/harder if the executor/step gets killed due to excessive memory demand.

HannesWell:
After a few more experiments with explicitly set heap-sizes even with `-Xmx1G` the build succeeded. While setting `-Xmx8G -Xms4G` lead to the 'known' obscure `<some number> Killed` error messages.
My explanation for this is that since the max heap limit was set to 4GiB the JVM respectively GC exhausted that limit, became lazy to collect garbage in the heap and just grew the heap. As soon as the memory demand of the JVM in total (which needs besides the heap also native memory for class data, string constants, etc. and its own execution) plus the demand of other programs running exceeded the 4GiB of the executor the script step was canceled respectively the maven execution killed.

Since by default the VM argument `-XX:MaxRAMPercentage` is set to 25(%) of the RAM available in total, just removing the explicit 
`MAVEN_OPTS` is equivalent to specify `-Xmx1G`. One can verify that by passing `-XX:+PrintFlagsFinal`.

In case the implicitly set `-Xmx1G` is not sufficient after all, the Maven JVM will now throw an `OutOfMemoryError` (which at least can be dumped and analyzed quite easy) and we can still double the heap to `-Xmx2G` and should still leave sufficient memory for others.

I have also investigated where the previous value was coming from and it was set from the very beginning when the Jenkinsfile was created. I assume it was set without any specific need just to be 'on the save side' of heap demand.

Hopefully this (surprisingly simple) fix will make these obscure `<some number> Killed` aborts a relic of the past.
If anyone still encounters them (or encounters OOMErrors) please let me know.

jukzi:
You may consider to have same settings in the same file as in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1798/files

HannesWell:
> You may consider to have same settings in the same file as in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1798/files

A discussion about exact these settings in fact lead to this change. But at the moment these settings are
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/cc85a7a285e7e893345763cc6f8df51957546a47/.mvn/jvm.config#L2

But I don't think it is necessary to always enable `-XX:+PrintFlagsFinal` and since the defaults seem to be fine, I would like to just continue with that and keep it simple. If this needs more fine tuning in the future we can still add explicit VM arguments again or add a `.mvn/jvm.config` file. But from all I know, local builds don't had such problems in the past (nowadays computers usually have way more than 4GiB of RAM).

PR #1301, opened 2024-06-25
GC Data image scaling Bugfix

amartya4256:
This PR fixes the image scaling bug in GC data. The image was scaled with the native zoom of the GC Data instead of the zoom. This leads to the the problem of no image drawn but just a white background. The image is only visible when the zoom and nativeZoom are the same. This behavior can be seen if the IDE is started at i.e. 125% zoom level on the default monitor. Hence, in this case the zoom is set to 100 and nativeZoom is 125. And due to wrong scaling, the lineNumberRulerColumn and minimize and maximize button images are not visible but rather just a white background is visible. 

image

This fix provides the right scaling to the image in the GC Data using the zoom but not the nativeZoom level as per the standard.

Contributes to #62 and #127

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
@@ -3930,7 +3930,7 @@ void init(Drawable drawable, GCData data, long hDC) {
 	}
 	Image image = data.image;
 	if (image != null) {
-		data.hNullBitmap = OS.SelectObject(hDC, Image.win32_getHandle(image, data.nativeZoom));
+		data.hNullBitmap = OS.SelectObject(hDC, Image.win32_getHandle(image, DPIUtil.getZoomForAutoscaleProperty(data.nativeZoom)));
 		image.memGC = this;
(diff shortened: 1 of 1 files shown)


deepika-u:
Inspite of applying the pr, i am still not able to see the line numbers and maximize/minimize buttons.

Also there is a white background covering the widgets at many places. To make it more visible i have changed to dark theme to showcase the same issue as below
I am not able to see 
- line numbers though enabled, 
- '+' sign next to imports, 
- similarly maximize/minimize and so on....

image

fedejeanne:
The fix works for me, merging.

image

@deepika-u your WS seems quite "full". Which application did you start to test it? Which WS did you use? I started `Runtime Workspace` (app) with a fresh workspace and added a project and a java class to it.

deepika-u:
Sorry it works, i applied pr but missed to take child eclipse to test it. That was my miss. Thanks fede.

fedejeanne:
> Sorry it works, i applied pr but missed to take child eclipse to test it. That was my miss. Thanks fede.

... been there, done that (https://github.com/eclipse-platform/eclipse.platform.ui/pull/1916#issuecomment-2196358903) 😅

My pleasure! 👍

PR #1303, opened 2024-06-25
Make win32 embedded tests package protected

merks:
Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/1298


HeikoKlare (commented):
Unfortunately, I think this will not work as package-protected classes will not be picked as test classes by a JUnit 4 runner. GitHub also reports the tests as removed as they don't appear as executed in the logs:
image

My understanding is that a migration to JUnit 5 is required to allow the visibility reduction.

vogella:
In the past we removed the exported API from test packages. Tests are not API (which was confirmed a longer time ago in the PMC IIRC (before Ed joined).

merks:
Then let's abandon this.

PR #1309, opened 2024-06-26
Removal of region constructor with handle param

amartya4256:
This commit removes the constructor and the following method win32_new and hereby confirm that none of the clients consume them.

contributes to #62 and #127

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Region.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Region.java
@@ -90,17 +90,6 @@ public Region (Device device) {
 }
 
 /**
- * Constructs a new region given a handle to the operating
- * system resources that it should represent.
- *
- * @param handle the handle for the result
- */
-Region(Device device, int handle) {
-	super(device);
-	this.handle = handle;
-}
-
-/**
  * Adds the given polygon to the collection of polygons
  * the receiver maintains to describe its area.
  *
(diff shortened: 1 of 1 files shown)


HeikoKlare (approved):
The removals looks good to me. Only the commit message seems a little broken: "[...] and the following method win_32 [...]". Would be great if you could fix that.

As an additional note on this PR: is does not only simplify the existing code, it will also simplify and improve the Windows HiDPI improvements for the `Region` class (#1278).

Even though the removed static method is tagged with `@noreference` (and even only part of an OS-specific implementation) and should thus be safe to remove, it would be great to have this confirmed by someone more experienced. Maybe @akurtakov you can give an advice whether this removal is okay?

akurtakov:
> The removals looks good to me. Only the commit message seems a little broken: "[...] and the following method win_32 [...]". Would be great if you could fix that.
> 
> As an additional note on this PR: is does not only simplify the existing code, it will also simplify and improve the Windows HiDPI improvements for the `Region` class (#1278).
> 
> Even though the removed static method is tagged with `@noreference` (and even only part of an OS-specific implementation) and should thus be safe to remove, it would be great to have this confirmed by someone more experienced. Maybe @akurtakov you can give an advice whether this removal is okay?

This removal is fine as this method is only available on windows, has been properly annotated with noreference and will actually break cross platformness of the codebase if anyone has used it and even that would be quite hard as one would have to know the C pointer in order to use it proper.

HeikoKlare:
Thank you for the quick response and confirmation!

PR #1311, opened 2024-06-26
Parse swt.autoScale.updateOnRuntime only in win32

akoch-yatta:
The swt.autoScale.updateOnRuntime has currently only effect in the win32 implementation. Therefor parsing of the flag will only be done on win32. This does not fix any open issue currently, but will make sure, **IF** this flag is used in common code, relying on it will result in a unintended execution on GTK or cocoa.
This PR is a prerequisite to get rid of the "Restart on DPI Change" dialog in WorkbenchWindow with the flag being true on win32


HeikoKlare (approved):
Makes sense and looks good.

PR #1315, opened 2024-06-28
LineHeaderCodeMining not rendered properly with DPIUtil#deviceZoom=200

tobiasmelcher:
LineHeaderCodeMinings are not properly rendered on my windows machine when DPIUtil#deviceZoom is set to value 200.

Here a screenshot how it looks like:
vertical_line_indent_1
The height of line 7 and 13 is not correct from my point of view, the height is twice as high as it should be. Over these 2 lines the minings are rendered. 

After applying the changes in this pull request, it looks much better:
vertical_line_indent_2

TextLayout#getBounds() calls DPIUtil.autoScaleUp for the lineVerticalIndent value which then makes the height of the given line too high. 
Removing the DPIUtil.autoScaleUp for the lineVerticalIndent is not a solution, text rendering is then destroyed for all editor lines. 
It would be great if an expert could take a look into this issue and maybe find a much better solution than mine. It feels more like a workaround what I have done here.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/StyledTextRenderer.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/StyledTextRenderer.java
@@ -21,6 +21,7 @@ import java.util.List;
 
 import org.eclipse.swt.*;
 import org.eclipse.swt.graphics.*;
+import org.eclipse.swt.internal.*;
 import org.eclipse.swt.widgets.*;
 
 /**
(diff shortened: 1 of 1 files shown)


akoch-yatta:
Thanks for bringing that topic up. As we are currently working on a better HiDPI support in windows, I'll try to have a look at it this week.

BeckerWdf:
> Thanks for bringing that topic up. As we are currently working on a better HiDPI support in windows, I'll try to have a look at it this week.

@akoch-yatta: Any new on this. Will you provide a fix for the september-release?

akoch-yatta:
I had a look at it this morning, and https://github.com/eclipse-platform/eclipse.platform.swt/pull/1381 should be a fix for it, see
image
 vs 
 image

BeckerWdf:
Fixed with https://github.com/eclipse-platform/eclipse.platform.swt/pull/1381

PR #1316, opened 2024-07-02
Scaling of Text Icon for multi zoom level (win 32)

amartya4256:
This contribution provides the ability to scaling the special icons of Text in win 32, i.e., Search and Cancel icon. The scaled icon is obtained at the draw time in the method Text.WM_DRAWITEM. We get rid of the fields hIconSearch and hIconCancel and we maintain a hashmap of size to icon for both the icons. We use the size as the key since the size of the icon is based on the rectangle it is drawn in. And since the size of the rectangle is determined from the OS directly, the size can be non-uniform sometimes. So we maintain the hashmap for size to handle instead of the zoom.

image

### How to Test
* Start an eclipse workspace with the arguments metioned below and open the error log view.
```
        -Xms256m
        -Xmx2048m
```
* There's a text bar in the view containing a search icon to the left. You need to type in some text in the text bar to make the cancel icon appear to the right.
* Now move around the icon to monitors with different zoom levels. The icons should scale to the respective zoom levels.

contributes to #62 and #127

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
@@ -271,8 +271,8 @@ public class Display extends Device implements Executor {
 	boolean textUseDarkthemeIcons = false;
 
 	/* Custom icons */
-	long hIconSearch;
-	long hIconCancel;
+	private HashMap<Integer, Long> sizeToSearchIconHandle = new HashMap<>();
+	private HashMap<Integer, Long> sizeToCancelIconHandle = new HashMap<>();
 
 	/* Focus */
 	int focusEvent;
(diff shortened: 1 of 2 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
```diff
@@ -1188,6 +1188,24 @@ static Image createIcon (Image image) {
 	return Image.win32_new (device, SWT.ICON, hIcon);
 }
 
+long getTextSearchIcon(int size) {
+	if (!sizeToSearchIconHandle.containsKey(size)) {
+		int searchIconResource = textUseDarkthemeIcons ? Text.IDI_SEARCH_DARKTHEME : Text.IDI_SEARCH;
+	    long iconHandle = OS.LoadImage (OS.GetLibraryHandle (), searchIconResource, OS.IMAGE_ICON, size, size, 0);
```
fedejeanne, line 1194: You should check for `0` like `Test::addIcons()` was doing, in case there are no more handles,

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Text.java
```diff
@@ -2770,7 +2753,10 @@ LRESULT WM_DRAWITEM (long wParam, long lParam) {
 		int state = OS.GetKeyState (OS.VK_LBUTTON) < 0 ? OS.PBS_PRESSED : OS.PBS_HOT;
 		OS.DrawThemeBackground (display.hButtonThemeAuto (), struct.hDC, OS.BP_PUSHBUTTON, state, rect, null);
 	}
-	long hIcon = (struct.CtlID == SWT.ICON_SEARCH) ? display.hIconSearch : display.hIconCancel;
+	int width = rect.right - rect.left;
+	int height = rect.bottom - rect.top;
+	int size = width <= height ? width : height;
```
fedejeanne, line 2758:
```suggestion
	int size = Math.min(width, height);
```

fedejeanne (changes requested):
Also simplify the commit text, there are too many details for my taste. Is there no issue to reference in the commit?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Text.java
amartya4256: Alright

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
amartya4256: Done

fedejeanne:
How do I test this?

fedejeanne:
FTR: I don't know why the git bot says that this PR removes 35 tests. I checked the logs, they are running.

PR #1320, opened 2024-07-03
Corrected TextLayout to produce right number of lines when '\r\n' sequence comes.

### LINKED ISSUE #184: TextLayout Produces Wrong Lines for '\r\n'
**Descritption**

When having a sequence of `\r\n\r\n' and a style range that intersects the line break, TextLayout produces too many lines. 

The sequence `some text\r\n\r\nmore text? should produce a single empty line between the two text lines, but it produces two empty lines if the style range is set appropriately.

**To Reproduce**
Run the example below:

<details>
<summary>Example Snippet</summary>

```
/ * Copyright (c) 2000, 2016 IBM Corporation and others.
 *
 *
 *
 * Contributors:
 *     IBM Corporation - initial API and implementation

/*
 * TextLayout example snippet: text with underline and strike through
 *
 * For a list of all SWT example snippets see
 * http://www.eclipse.org/swt/snippets/
 *
 * @since 3.1
 */
import org.eclipse.swt.*;

public class Snippet256a {

	private static FontData[] getFontData(Font font, int style) {
		FontData[] fontDatas = font.getFontData();
		for (FontData fontData : fontDatas) {
			fontData.setStyle(style);
		}
		return fontDatas;
	}

		final Shell shell = new Shell(display, SWT.SHELL_TRIM | SWT.DOUBLE_BUFFERED);
		shell.setText("Underline, Strike Out");
		String text = "Here is \r\n\r\nsome text that is \r\nunderlined or struck out or both.";
		FontData[] fontData = getFontData(shell.getFont(), SWT.BOLD);
		Font font = new Font(shell.getDisplay(), fontData);
		final TextLayout layout = new TextLayout(display);
		layout.setText(text);
		TextStyle style1 = new TextStyle(font, null, null);
		layout.setStyle(style1, 0, 10);

		int[] offsets = layout.getLineOffsets();
		System.out.println(offsets.length);
		shell.addListener(SWT.Paint, event -> {
			Point point = new Point(10, 10);
			int width = shell.getClientArea().width - 2 * point.x;
			layout.setWidth(width);
			layout.draw(event.gc, point.x, point.y);
		});
		shell.open();
		while (!shell.isDisposed()) {
				display.sleep();
		}
		layout.dispose();
		display.dispose();
	}
}

```

</details>

**Environment:**
- Select the platforms on which the behavior is seen.
[] All OS
[x] Windows
[ ] Linux
[ ] macOS

deepika-u:
Corrected TextLayout to produce right number of lines when '\r\n' sequence comes.

Fixes #184

@sravanlakkimsetti 
@niraj-modi 
@merks 
Can one of you review this fix please.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/TextLayout.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/TextLayout.java
@@ -42,7 +42,7 @@ import org.eclipse.swt.internal.win32.*;
  */
 public final class TextLayout extends Resource {
 	Font font;
-	String text, segmentsText;
+	String text, segmentsText, originalText;
 	int lineSpacingInPoints;
 	int ascent, descent;
 	int alignment;
@@ -309,7 +309,7 @@ public TextLayout (Device device) {
 	styles[0] = new StyleItem();
 	styles[1] = new StyleItem();
 	stylesCount = 2;
-	text = ""; //$NON-NLS-1$
+	text = originalText = ""; //$NON-NLS-1$
 	long[] ppv = new long[1];
 	OS.OleInitialize(0);
 	if (COM.CoCreateInstance(COM.CLSID_CMultiLanguage, 0, COM.CLSCTX_INPROC_SERVER, COM.IID_IMLangFontLink2, ppv) == OS.S_OK) {
@@ -2724,8 +2724,16 @@ private int getScaledVerticalIndent() {
  */
 public TextStyle getStyle (int offset) {
 	checkLayout();
-	int length = text.length();
+	int length = originalText.length();
 	if (!(0 <= offset && offset < length)) SWT.error(SWT.ERROR_INVALID_RANGE);
+	int crCount = 0, subStringLength = originalText.subSequence(0, offset+1).length();
+	for (int crIndex = 0; crIndex < subStringLength; crIndex++) {
+		if (originalText.charAt(crIndex) == '\r') {
+			crCount++;
+		}
+	}
+	offset = offset - crCount;
+
 	for (int i=1; i<stylesCount; i++) {
 		if (styles[i].start > offset) {
 			return styles[i - 1].style;
@@ -2956,15 +2964,11 @@ StyleItem[] merge (long items, int itemCount) {
 			linkBefore = false;
 		}
 		char ch = segmentsText.charAt(start);
-		switch (ch) {
-			case '\r':
-			case '\n':
-				item.lineBreak = true;
-				break;
-			case '\t':
-				item.tab = true;
-				break;
+		if (ch == '\r' && start + 1 < end) {
+			ch = segmentsText.charAt(start + 1);
 		}
+		item.lineBreak = ch == '\n';
+		item.tab = ch == '\t';
 		if (itemLimit == -1) {
 			nextItemIndex = itemIndex + 1;
 			OS.MoveMemory(scriptItem, items + nextItemIndex * SCRIPT_ITEM.sizeof, SCRIPT_ITEM.sizeof);
@@ -3455,6 +3459,22 @@ public void setStyle (TextStyle style, int start, int end) {
 	int length = text.length();
 	if (length == 0) return;
 	if (start > end) return;
+
+	int startCount = 0;
+	int endCount = 0;
+	for (int crIndex = originalText.indexOf('\r', 0); crIndex >= 0; crIndex = originalText.indexOf('\r', crIndex + 1)) {
+		if (crIndex < start) {
+			++startCount;
+		} else if (crIndex <= end) {
+			++endCount;
+		} else {
+			break;
+		}
+	}
+	endCount = endCount + startCount;
+	start -= startCount;
+	end -= endCount;
+
 	start = Math.min(Math.max(0, start), length - 1);
 	end = Math.min(Math.max(0, end), length - 1);
 	int low = -1;
@@ -3568,6 +3588,8 @@ public void setTabs (int[] tabs) {
  */
 public void setText (String text) {
 	checkLayout();
+	this.originalText = text;
+	text = text.replace("\r", "");
 	if (text == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
 	if (text.equals(this.text)) return;
 	freeRuns();
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_TextLayout.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_TextLayout.java
@@ -547,6 +547,60 @@ public void test_setStyle() {
 	assertNull(layout.getStyle(4));
 	assertNull(layout.getStyle(5));
 	layout.dispose();
+
+	layout = new TextLayout (display);
+	layout.setText("\rabc\ndef\r\rghi\n\njkl\r\nmno\r\n\r\npqr\r");
+	layout.setStyle(s1, 0, 5);//abcd
+	layout.setStyle (s2, 7, 10);//fg
+	layout.setStyle (s3, 11, 18);//hijkl
+	layout.setStyle (s4, 19, 24);//mno
+	assertEquals(s1, layout.getStyle(0));
+	assertEquals(s1, layout.getStyle(1));
+	assertEquals(s1, layout.getStyle(3));
+	assertEquals(s1, layout.getStyle(4));
+	assertEquals(s1, layout.getStyle(5));
+
+	assertEquals(s2, layout.getStyle(7));
+	assertEquals(s2, layout.getStyle(8));
+	assertEquals(s2, layout.getStyle(9));
+	assertEquals(s2, layout.getStyle(10));
+
+	assertEquals(s3, layout.getStyle(12));
+	assertEquals(s3, layout.getStyle(13));
+	assertEquals(s3, layout.getStyle(14));
+	assertEquals(s3, layout.getStyle(15));
+
+	assertEquals(s4, layout.getStyle(19));
+	assertEquals(s4, layout.getStyle(20));
+	assertEquals(s4, layout.getStyle(22));
+	assertEquals(s4, layout.getStyle(23));
+	assertEquals(s4, layout.getStyle(24));
+
+	layout.setStyle (null, 0, 3);//abc
+	assertNull(layout.getStyle(0));
+
+	layout.setStyle(s1, 0, 9);//abcdef
+	assertEquals(s1, layout.getStyle(8));
+
+	layout.setStyle (s2, 1, 24);//abcdefghijklmno
+	assertEquals(s2, layout.getStyle(7));
+	assertEquals(s2, layout.getStyle(8));
+	assertEquals(s2, layout.getStyle(14));
+	assertEquals(s2, layout.getStyle(18));
+	assertEquals(s2, layout.getStyle(22));
+	assertEquals(s2, layout.getStyle(23));
+	assertEquals(s2, layout.getStyle(24));
+
+	layout.setStyle (s3, 0, 30);//abcdefghijklmnopqr
+	assertEquals(s3, layout.getStyle(0));
+	assertEquals(s3, layout.getStyle(1));
+	assertEquals(s3, layout.getStyle(30));
+
+	layout.setStyle (s4, 1, 30);//abcdefghijklmnopqr
+	assertEquals(s4, layout.getStyle(1));
+	assertEquals(s4, layout.getStyle(29));
+	assertEquals(s4, layout.getStyle(30));
+	layout.dispose();
 }
 
 @Test
@@ -1217,6 +1271,33 @@ public void test_Bug579335_win32_StyledText_LongLine() {
 			image.dispose();
 		if (font != null)
 			font.dispose();
+	}
+}
+
+@Test
+public void test_getLineCount() {
+	// Skipping this test for GTK and Cocoa as this pr #1320 applies only for
+	// windows platform.
+	// Without pr #1320(master) : Number of lines: 11 (for Cocoa it is : 10)
+	// Expected : Number of lines: 7 (for Cocoa it is : 10)
+	TextLayout layout = new TextLayout(display);
+	try {
+		String text = "\rabc\ndef\r\rghi\n\njkl\r\nmno\r\n\r\npqr\r";
+		layout.setText(text);
+		int lineCount = layout.getLineCount();
+
+		int expected = 0;
+		if (SwtTestUtil.isWindows) {
+			expected = 7;
+		} else if (SwtTestUtil.isCocoa) {
+			expected = 10;
+		} else if (SwtTestUtil.isGTK) {
+			expected = 11;
+		}
+
+		assertEquals(expected, lineCount);
+	} finally {
+		layout.dispose();
 	}
 }
 }
--- a/tests/org.eclipse.swt.tests/ManualTests/org/eclipse/swt/tests/manual/Issue184_CarriageReturnHandled.java
+++ b/tests/org.eclipse.swt.tests/ManualTests/org/eclipse/swt/tests/manual/Issue184_CarriageReturnHandled.java
@@ -1,0 +1,69 @@
+/*******************************************************************************
+ * Copyright (c) 2024 IBM and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.swt.tests.manual;
+
+import org.eclipse.swt.*;
+import org.eclipse.swt.graphics.*;
+import org.eclipse.swt.widgets.*;
+
+public class Issue184_CarriageReturnHandled {
+
+	private static FontData[] getFontData(Font font, int style) {
+		FontData[] fontDatas = font.getFontData();
+		for (FontData fontData : fontDatas) {
+			fontData.setStyle(style);
+		}
+		return fontDatas;
+	}
+
+	public static void main(String[] args) {
+		Display display = new Display();
+		final Shell shell = new Shell(display, SWT.SHELL_TRIM | SWT.DOUBLE_BUFFERED);
+		shell.setText("Underline, Strike Out");
+		String text1 = "\r\nde\nep\rika\r\rudaya\n\ngiri\r";
+
+		FontData[] fontData = getFontData(shell.getFont(), SWT.BOLD);
+		Font font = new Font(shell.getDisplay(), fontData);
+
+		FontData[] fontData1 = getFontData(shell.getFont(), SWT.ITALIC | SWT.BOLD);
+		Font font1 = new Font(shell.getDisplay(), fontData1);
+
+		FontData[] fontData2 = getFontData(shell.getFont(), SWT.BOLD);
+		Font font2 = new Font(shell.getDisplay(), fontData2);
+
+		final TextLayout layout = new TextLayout(display);
+		layout.setText(text1);
+
+		TextStyle style1 = new TextStyle(font, null, null);
+		layout.setStyle(style1, 3, 7); // eep in bold
+
+		TextStyle style2 = new TextStyle(font1, null, null);
+		layout.setStyle(style2, 12, 18); // udaya in bold
+
+		TextStyle style3 = new TextStyle(font2, null, null);
+		layout.setStyle(style3, 21, 24); // iri in bold
+
+		shell.addListener(SWT.Paint, event -> {
+			Point point = new Point(10, 10);
+			int width = shell.getClientArea().width - 2 * point.x;
+			layout.setWidth(width);
+			layout.draw(event.gc, point.x, point.y);
+		});
+		shell.setSize(400, 300);
+		shell.open();
+		while (!shell.isDisposed()) {
+			if (!display.readAndDispatch())
+				display.sleep();
+		}
+		layout.dispose();
+		display.dispose();
+	}
+}


deepika-u:
Anyone testing this pr can use the below snippet.
simple test case is with text1.
complex test case is with text2.

```
package org.eclipse.swt.bugs4;

/*
 * TextLayout example snippet: text with underline and strike through
 *
 * For a list of all SWT example snippets see
 * http://www.eclipse.org/swt/snippets/
 *
 * @since 3.1
 */
import org.eclipse.swt.*;

public class Issue184 {

	private static FontData[] getFontData(Font font, int style) {
		FontData[] fontDatas = font.getFontData();
		for (FontData fontData : fontDatas) {
			fontData.setStyle(style);
		}
		return fontDatas;
	}

		final Shell shell = new Shell(display, SWT.SHELL_TRIM | SWT.DOUBLE_BUFFERED);
		shell.setText("Underline, Strike Out");
		//String text = "Here is 1.\r\n\r\nsome text that is 2.\r\nunderlined or 3.\rstruck out or 4.\nboth.";
		String text1 = "0123456789\r\n\r\n0123456789\r\n0123456789\r0123456789\n0123456789";
		String text2 = "\r\n0123456789\r\n\r\n01\r\n456789\r\n0123456789\r0123456789\n0123456789";

		FontData[] fontData = getFontData(shell.getFont(), SWT.BOLD);
		Font font = new Font(shell.getDisplay(), fontData);

		FontData[] fontData1 = getFontData(shell.getFont(), SWT.ITALIC | SWT.BOLD);
		Font font1 = new Font(shell.getDisplay(), fontData1);

		FontData[] fontData2 = getFontData(shell.getFont(), SWT.BOLD);
		Font font2 = new Font(shell.getDisplay(), fontData2);

		final TextLayout layout = new TextLayout(display);
		//layout.setText(text1);
		layout.setText(text2);

		TextStyle style1 = new TextStyle(font, null, null);
		//layout.setStyle(style1, 0, 16);
		layout.setStyle(style1, 5, 16);

		TextStyle style2 = new TextStyle(font1, null, null);
		layout.setStyle(style2, 21, 24);

		TextStyle style3 = new TextStyle(font2, null, null);
		layout.setStyle(style3, 27, 31);
		//layout.setStyle(style3, 25, 29);

		int[] offsets = layout.getLineOffsets();
		System.out.println(offsets.length);
		shell.addListener(SWT.Paint, event -> {
			Point point = new Point(10, 10);
			int width = shell.getClientArea().width - 2 * point.x;
			layout.setWidth(width);
			layout.draw(event.gc, point.x, point.y);
		});
		shell.setSize(400, 300);
		shell.open();
		while (!shell.isDisposed()) {
				display.sleep();
		}
		layout.dispose();
		display.dispose();
	}
}
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/TextLayout.java
```diff
@@ -2952,8 +2952,11 @@ StyleItem[] merge (long items, int itemCount) {
 			linkBefore = false;
 		}
 		char ch = segmentsText.charAt(start);
+		if(ch == '\r') {
+			ch = segmentsText.charAt(start+1);
+		}
 		switch (ch) {
-			case '\r':
+			//case '\r':
```
elsazac, line 2959: You can remove this line , if commented out

deepika-u: We will retain this as people looking at this part of code should get to know that there was earlier implementation which was the cause of this extra line break which is '\r' case and the respective handling of indexes is all about the whole fix proposed in this pr.

laeubi: It would be better to write an explicit comment instead e.g.
```suggestion
			// '\r': is already handled above
```

on the other hand I wonder if the switch is here of any use... e.g. you can write:

item.lineBreak = ch=='\n'
item.tab = ch=='\t'

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/TextLayout.java
```diff
@@ -2952,8 +2952,11 @@ StyleItem[] merge (long items, int itemCount) {
 			linkBefore = false;
 		}
 		char ch = segmentsText.charAt(start);
+		if(ch == '\r') {
+			ch = segmentsText.charAt(start+1);
```
laeubi, line 2956: is a boundcheck required here maybe? Just assume a single `\r` at end of line would read bejond the `end`

deepika-u: yes agree

deepika-u: i am checking on further if the string in full is just a "\r", what needs to be done.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/TextLayout.java
deepika-u: done with the suggestions.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/TextLayout.java
deepika-u: Done.

deepika-u:
@niraj-modi @merks
Can you also please review this?

@laeubi : Do you have any additional comments on this fix? or are we good to merge this?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/TextLayout.java
```diff
@@ -42,8 +42,8 @@
  */
 public final class TextLayout extends Resource {
 	Font font;
-	String text, segmentsText;
-	int lineSpacingInPoints;
+	String text, segmentsText, originalText;
+	int lineSpacingInPoints, previousCountNextLine = 0, previousEnd = -1;
```
merks, line 46: Initializing to 0 is pointless and actually generates more byte code.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/TextLayout.java
```diff
@@ -2952,15 +2952,11 @@ StyleItem[] merge (long items, int itemCount) {
...
 		char ch = segmentsText.charAt(start);
-		switch (ch) {
-			case '\r':
-			case '\n':
-				item.lineBreak = true;
-				break;
-			case '\t':
-				item.tab = true;
-				break;
+		if(ch == '\r' && (start+1 < end)) {
```
merks, line 2955: The formatting of `if(` is poor.  The () around the < condition is redundant and odd that it's not used for the == test

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/TextLayout.java
```diff
@@ -3451,6 +3447,28 @@ public void setStyle (TextStyle style, int start, int end) {
 	int length = text.length();
 	if (length == 0) return;
 	if (start > end) return;
+	int countNextLine = 0, noStyleCountNextLine = 0, loop = 0;
+	if (previousEnd < 0)
+		loop = previousEnd+1;
+	else
+		loop = previousEnd;
+	for(int i = loop; i < end; i++) {
```
merks, line 3455: Here `for(` is not nice format.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/TextLayout.java
deepika-u: Updated initialization as part of constructor.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/TextLayout.java
deepika-u: corrected it.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/TextLayout.java
deepika-u: corrected it.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/TextLayout.java
```diff
@@ -3564,6 +3584,13 @@ public void setTabs (int[] tabs) {
  */
 public void setText (String text) {
 	checkLayout();
+
+	this.originalText = text;
+	String[] strings = text.split("\r");
+	text = "";
+	for(int i = 0; i < strings.length; i++)
```
elsazac, line 3591: The formatting of this for loop could also be improved.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/TextLayout.java
```diff
@@ -3564,6 +3584,13 @@ public void setTabs (int[] tabs) {
  */
 public void setText (String text) {
 	checkLayout();
+
+	this.originalText = text;
+	String[] strings = text.split("\r");
+	text = "";
+	for (int i = 0; i < strings.length; i++)
```
Wittmaxi, line 3591: This can be done using a map reduction. Anyway, please add braces to the for loop

[reference](https://www.google.com/url?sa=t&source=web&cd=&ved=2ahUKEwjQk7ChiquHAxXHzAIHHVg0BosQFnoECA8QAQ&url=https%3A%2F%2Fwww.baeldung.com%2Fjava-stream-reduce&usg=AOvVaw0nCMK38jusEkWmmenr6Mzu&opi=89978449)

merks: It's this a complicated and poorly performance way of doing test.replace("\r", "")?

Wittmaxi: @merks yes, looks like it. Your approach is the best to me!

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/TextLayout.java
```diff
@@ -3451,6 +3449,28 @@ public void setStyle (TextStyle style, int start, int end) {
 	int length = text.length();
 	if (length == 0) return;
 	if (start > end) return;
+	int countNextLine = 0, noStyleCountNextLine = 0, loop = 0;
+	if (previousEnd < 0)
+		loop = previousEnd+1;
+	else
+		loop = previousEnd;
+	for (int i = loop; i < end; i++) {
+		if (originalText.charAt(i) == '\r' && originalText.charAt(i+1) =='\n')
```
merks, line 3458: Is i + 1 necessarily < end?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/TextLayout.java
deepika-u: let me test and update

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/TextLayout.java
deepika-u: let me check it.

deepika-u: you are correct, when string ends with \r, it is resulting OOB exception. So corrected it.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/TextLayout.java
deepika-u: Thanks @merks for your feedback, after updating this i found a better way of fixing the issue. So new change is updated, you may please review now.

deepika-u:
@niraj-modi 
Can you also please review this? so that it can be merged before M3.

@merks @laeubi @Wittmaxi  : Do you have any additional comments on this fix? or are we good to merge this?

deepika-u:
@merks @laeubi @Wittmaxi 
Can one of you review the changes please.

merks:
@deepika-u 

Could you please squash this to a single commit?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/TextLayout.java
```diff
@@ -3451,6 +3447,19 @@ public void setStyle (TextStyle style, int start, int end) {
...
+
+	startString = startString.charAt(stringLength-1) == '\r' ? startString.substring(0,stringLength-1) : startString;
+	int startCount = (int) startString.chars().filter(ch -> ch == '\r').count();
+
+	String endString = originalText.substring(0, end+1);
+	int endCount = (int) endString.chars().filter(ch -> ch == '\r').count();
+
+	start = start == 0 ? start : start - startCount;
+	end = end - endCount;
+
```
merks, line 3462: I think the performance of setStyle is rather critical and this block of code looks scary both in complexity and in performance.  Several strings are created, only to be scanned with a stream and an lamda, just for counting the number of \r and very often there won't be one at all.

I think it would be far better to use java.lang.String.indexOf(int, int) to look for \r rather than to create a bunch of Strings, streams, and lambdas.

E.g., something like this **untested** code:
```
		int startCount = 0;
		int endCount = 0;
		for (int crIndex = originalText.indexOf('\r', 0); crIndex >= 0; crIndex = originalText.indexOf('\r',
				crIndex + 1)) {
			if (crIndex <= start) {
				++startCount;
			} else if (crIndex <= end) {  // Not sure if end offset is inclusive...
				++endCount;
			} else {
				break;
			}
		}

		start -= startCount;
		end -= endCount;
```

I also wonder how a \r not followed by a \n is handled overall...

Given the complexity here, a unit test would seem like a very good thing...

merks (changes requested):

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/TextLayout.java
deepika-u: As per your suggestion i have the same fix with some additional changes. I have uploaded the newer version with  java.lang.String.indexOf(int, int) being used. 
@merks : You may want to review now.

Also attached the snippet in manual tests.
Also attaching the excel sheet i have tried to test this fix(for your reference).
[Issue184_21stAug2024.xlsx](https://github.com/user-attachments/files/16687533/Issue184_21stAug2024.xlsx)

tests/org.eclipse.swt.tests/ManualTests/org/eclipse/swt/tests/manual/Issue184_CarriageReturnHandled.java
```diff
@@ -0,0 +1,81 @@
...
+		Font font = new Font(shell.getDisplay(), fontData);
+
+		FontData[] fontData1 = getFontData(shell.getFont(), SWT.ITALIC | SWT.BOLD);
+		Font font1 = new Font(shell.getDisplay(), fontData1);
+
+		FontData[] fontData2 = getFontData(shell.getFont(), SWT.BOLD);
+		Font font2 = new Font(shell.getDisplay(), fontData2);
+
+		final TextLayout layout = new TextLayout(display);
+		//layout.setText(text);
```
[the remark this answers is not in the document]
deepika-u: Those are the various test cases tried, if required anyone can rerun depending on their need so i have retained.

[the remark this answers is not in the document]
deepika-u: > Looks good, but why is there this much commented-out code?

Apart from the manual test case, the fix is reviewed and approved? do you mean that?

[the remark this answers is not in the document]
sravanlakkimsetti: Deepika please remove the comment code and update the PR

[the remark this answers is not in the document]
deepika-u: Done.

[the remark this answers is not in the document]
Wittmaxi: Yes, looks good now :)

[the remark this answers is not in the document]
deepika-u: Thanks alot for the review.

deepika-u:
@merks : All review suggestions are implemented, can you review once again please?

merks:
Wouldn't it be better for there to be actual tests that run with every build?

deepika-u:
> Wouldn't it be better for there to be actual tests that run with every build?

Apart from this manual test(tests/org.eclipse.swt.tests/ManualTests/org/eclipse/swt/tests/manual/Issue184_CarriageReturnHandled.java) which is already added, please let me know what else needs to be done, so that this can be taken forward. Can you point me to a sample kind of instance? to refer.

deepika-u:
> Wouldn't it be better for there to be actual tests that run with every build?

@merks : Can you please check now?

deepika-u:
@merks : I have added some junits for the fix. Can you check now please?

BeckerWdf:
is this only an issue on Windows? May the macOS and Linux implementation have the same issue?

deepika-u:
> is this only an issue on Windows? May the macOS and Linux implementation have the same issue?

@BeckerWdf  : Atleast from the parent #184 => this is marked as only windows issue and i have implemented the fix only for windows. In the past i remember vaguely but when tested with mac,  the behavior on mac Vs behavior on windows (before the fix) were different. So took it up as windows specific issue.

BeckerWdf:
> > is this only an issue on Windows? May the macOS and Linux implementation have the same issue?
> 
> @BeckerWdf : Atleast from the parent #184 => this is marked as only windows issue and i have implemented the fix only for windows. In the past i remember vaguely but when tested with mac, the behavior on mac Vs behavior on windows (before the fix) were different. So took it up as windows specific issue.

Thanks.

deepika-u:
@merks : 
I have added some junits for this fix. **Can you make some time** to move this forward? I know you are super busy these days but yeah please need your inputs this.

deepika-u:
@merks 
We have crossed the M2 milestone for 4.34 and entered into M3 already. Can you find some time to check on this please so that atleast for this release, its available for customers.

merks (approved):
It looks okay to me.

That being said, I would prefer that at least one other person deeply involved in SWT reviews this too.

merks:
@HeikoKlare 

So you know who else might be good to review this?

deepika-u:
@sratz @jukzi @laeubi @HeikoKlare 
Can one of you review this please. 
@merks thanks alot for your time.
But he feels if anyone else also reviews it would be great. - Thanks alot in advance, hoping this would go in 4.34.

tests/org.eclipse.swt.tests/ManualTests/org/eclipse/swt/tests/manual/Issue184_CarriageReturnHandled.java
```diff
@@ -0,0 +1,68 @@
+package org.eclipse.swt.tests.manual;
```
HeikoKlare, line 1: Please add missing copyright header

tests/org.eclipse.swt.tests/ManualTests/org/eclipse/swt/tests/manual/Issue184_CarriageReturnHandled.java
```diff
@@ -0,0 +1,68 @@
...
+		TextStyle style1 = new TextStyle(font, null, null);
+		layout.setStyle(style1, 3, 7); // eep in bold
+
+		TextStyle style2 = new TextStyle(font1, null, null);
+		layout.setStyle(style2, 12, 18); // udaya in bold
+
+		TextStyle style3 = new TextStyle(font2, null, null);
+		layout.setStyle(style3, 21, 24); // iri in bold
+
+		int[] offsets = layout.getLineOffsets();
```
HeikoKlare, line 52: This variable is unused and thus produces a warning. It should be safe to just delete this line of code.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_TextLayout.java
```diff
@@ -547,6 +547,60 @@ public void test_setStyle() {
 	assertNull(layout.getStyle(4));
 	assertNull(layout.getStyle(5));
 	layout.dispose();
+
+	layout = new TextLayout (display);
```
HeikoKlare, line 551: The test with these changes succeeds on master, so it's not a regression test for the changes made in this PR. Can you adapt the test so that it serves as a regression test for the addressed issue?

HeikoKlare (commented):
I have not worked on TextLayout so far. Thus, unfortunately, I cannot say with high confidence whether the overall change is okay. Still, from what I understand, the conceptual change and the code for it looks sound to me. I've tested the proposed change with the snippet in the original issue and it indeed fixes it.

I have remarks regarding the automated and manual test. In particular the automated test does not seem to be an adequate regression test, as the same test succeeds on `master`, thus it does not seem to cover the addressed issue.

HeikoKlare:
And one additional remark: since the `TextLayout` is very essential and we do not seem to have completely confident reviews so far, I am not sure whether we should bring this kind of change in M3. Maybe it makes more sense to bring in the next M1 to have better implicit testing until it is finally released? The issue has been there for quite a long time without higher complains that I am not sure whether the gain outweighs the risk of regression.

tests/org.eclipse.swt.tests/ManualTests/org/eclipse/swt/tests/manual/Issue184_CarriageReturnHandled.java
deepika-u: Updated.

tests/org.eclipse.swt.tests/ManualTests/org/eclipse/swt/tests/manual/Issue184_CarriageReturnHandled.java
deepika-u: Removed.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_TextLayout.java
deepika-u: Here we have not changed the behavior of the api's. The original behavior of the api's still work the same. We are internally detecting, correcting and passing the corrected content in specific cases only. We dint even add new api's. It is only the look and feel by which you can make out if some extra lines are expected or not expected varies. - is my understanding.

deepika-u:
> And one additional remark: since the `TextLayout` is very essential and we do not seem to have completely confident reviews so far, I am not sure whether we should bring this kind of change in M3. Maybe it makes more sense to bring in the next M1 to have better implicit testing until it is finally released? The issue has been there for quite a long time without higher complains that I am not sure whether the gain outweighs the risk of regression.

Thanks for your opinion. Since it is already tested now by various people, i feel we are safe to merge this. Considering the M3 deadline is 8th Nov(2 weeks far and not away like 2days or so) and we can give it a try. If something is breaking we'll revert it immediately and rectify to make it to next M1? Atleast this risk we can take to improve SWT - i feel so in particularly such long standing issues when they passed all the stages of fixing, review, successfully verified.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_TextLayout.java
HeikoKlare: Okay, I mistakenly expected the added test to be a regression test for the fix because of https://github.com/eclipse-platform/eclipse.platform.swt/pull/1320#issuecomment-2396659597

Still, this change is no refactoring but affects the behavior of public APIs (e.g., `TextLayout#draw(...)` behaves differently). I guess that, e.g., also `TextLayout#getBounds()` now behaves differently, so wouldn't it be possible to have a regression test comparing the bounds of two configurations that should actually result in the bounds but did not do before this change (or vice versa)?

deepika-u: @HeikoKlare 
As per your suggestion i have added new junit tests with `TextLayout#getLineCount()` and `TextLayout#getBounds()`.
Since this change is only with respect to windows, i have excluded mac, gtk for the tests to run.

I guess this makes it all complete as we discussed. If i missed out any, please do let me know.

Can you review it and take it forward please?

deepika-u: Please hold on, let me cross check and update you again.

deepika-u: Hi @HeikoKlare 
Sorry for the confusion caused.
I am successfull in adding only TextLayout#getLineCount() junit. I think we can go ahead with this pr now.

I have tried even TextLayout#getBounds() but had to withdraw as the behavior is not consistent across.
With the sample input taken, 
- On master, getBounds() is returning {0, 0, 30, 220} where width and height are 30, 220 respectively. [junit is failing]
- On the branch where i committed my changes(fix_for_issue184), getBounds() is returning {0, 0, 43, 140}. [junit is passing]
- To cross check, i have pulled the pr 1320 to local, getBounds() is returning {0, 0, 43, 140}. [junit is passing]
- But on the build machine, getBounds() is returning {0, 0, 34, 105}. [junit is failing because it is getting 34 but expecting 43]. -> https://github.com/eclipse-platform/eclipse.platform.swt/actions/runs/12312890737/job/34365830603?pr=1320
Neither able to understand the reason for the failure(to fix) nor able to replicate the problem locally to debug further.

This completes my analysis on this pr. Can you please review now?

deepika-u: @HeikoKlare : Need your inputs here please?

HeikoKlare: Thank you for adding the regression test. That one, testing `getLineCount()`, makes even more sense to me for testing the fix than testing the changed result of `getBounds()`. 👍 

The PR is ready to go from my side.
Just a small final remark regarding the added test: I am just wondering whether it may not be even better to use different expected line count values for the different platforms in the test instead of excluding the test for GTK and Cocoa. Since the functionality is not broken or non-existent on those other platform but just produces expected different results, testing for those other results seems to be more comprehensible to me.

deepika-u: @HeikoKlare 
Thanks for your time. I have updated them now, can you take a look now. Assuming everything is done, can you take this forward please.

HeikoKlare:
This looks good now and we also have Ed's approval, thus merging.

deepika-u:
Verified this issue on below build.
Eclipse SDK
Version: 2025-03 (4.35)
Build id: I20250107-1800
Java vendor: Eclipse Adoptium
Java runtime version: 23.0.1+11
Java version: 23.0.1

Number of lines: 7

jukzi:
This PR introduced a regression https://github.com/eclipse-platform/eclipse.platform.swt/issues/1714, please fix.
org.eclipse.swt.graphics.TextLayout.getRanges() tries to return only ranges with style!= null. TextLayoutgetStyle(range) for such a range should always return not null.

PR #1328, opened 2024-07-08
fix boundary condition in StyledText#getTextBounds

tobiasmelcher:
StyledText#getTextBounds returns 0x7fffffff as x value if input parameters start and end have the same value and line at the given offset is empty. This looks not correct from my point of view. 
This has the effect that code minings in org.eclipse.jface.text.examples.codemining.CodeMiningDemo are not drawn correctly. Two LineHeaderCodeMinings to be drawn on the same line are then drawn on top of each other (see screenshot at https://github.com/eclipse-platform/eclipse.platform.ui/pull/2024).

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/StyledText.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/StyledText.java
@@ -4892,7 +4892,7 @@ public Rectangle getTextBounds(int start, int end) {
 	Rectangle rect;
 	int y = getLinePixel(lineStart);
 	int height = 0;
-	int left = 0x7fffffff, right = 0;
+	int left = Integer.MAX_VALUE, right = 0;
 	for (int i = lineStart; i <= lineEnd; i++) {
 		int lineOffset = content.getOffsetAtLine(i);
 		TextLayout layout = renderer.getTextLayout(i);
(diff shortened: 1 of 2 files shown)


iloveeclipse:
Any chance for a test case / reproducer?

tobiasmelcher:
> Any chance for a test case / reproducer?

I added a unit test in Test_org_eclipse_swt_custom_StyledText and a standalone application org.eclipse.swt.examples.texteditor.mining.CodeMining which shows the problem scenario.

Running org.eclipse.swt.examples.texteditor.mining.CodeMining without the fix in StyledText#getTextBounds results in two overlapping minings:
two_minings_overlap

With the fix, the two minings are rendered properly:
two_minings_not_overlap

I know that I introduced a cycle by adding a dependency to org.eclipse.jface.text in org.eclipse.swt.examples. I just wanted to show you the scenario in a reproducible app. Should I remove the CodeMining application from the examples plugin once you were able to reproduce it?

examples/org.eclipse.swt.examples/src/org/eclipse/swt/examples/texteditor/mining/CodeMining.java
```diff
@@ -0,0 +1,137 @@
+package org.eclipse.swt.examples.texteditor.mining;
```
iloveeclipse, line 1: I believe we want header comments on all files.

examples/org.eclipse.swt.examples/META-INF/MANIFEST.MF
```diff
@@ -21,6 +21,9 @@ Export-Package: org.eclipse.swt.examples.accessibility,
  org.eclipse.swt.examples.layoutexample,
  org.eclipse.swt.examples.paint,
  org.eclipse.swt.examples.texteditor
-Require-Bundle: org.eclipse.swt
+Require-Bundle: org.eclipse.swt,
+ org.eclipse.ui,
+ org.eclipse.jface.text,
```
iloveeclipse, line 26: better to not add cyclic dependencies. 
In this case I would add the example in jface repo.

tobiasmelcher: ok, dependency to jface and test app now removed with [ded8838]

BeckerWdf (approved):

Issue #1286, opened 2024-06-19
Cosmetic issues in rendering when 'line spacing' option is set


sratz:
Cosmetic issues when 'line spacing' is set under _Preferences -> General -> Editors -> Text Editors -> Line spacing_.

### Line spacing 0:

![image](https://github.com/eclipse-platform/eclipse.platform.swt/assets/14908423/87e54a2d-406e-4f67-89a6-4a12f9d7e092)

![image](https://github.com/eclipse-platform/eclipse.platform.swt/assets/14908423/b01eb560-dd99-4d9c-80a0-a76e9b504044)

![image](https://github.com/eclipse-platform/eclipse.platform.swt/assets/14908423/d70fb825-e25f-452d-93e7-71abfba7bc74)

### Line spacing >0 (extreme value for demo purposes):

![image](https://github.com/eclipse-platform/eclipse.platform.swt/assets/14908423/f5683360-8055-40ed-bb01-b9239392d46b)

![image](https://github.com/eclipse-platform/eclipse.platform.swt/assets/14908423/efa43a89-0d61-41a4-b079-c7c0af643669)

![image](https://github.com/eclipse-platform/eclipse.platform.swt/assets/14908423/4da65143-6b54-4dc2-a08f-8513abda43cd)

### Compare VS Code:

![image](https://github.com/eclipse-platform/eclipse.platform.swt/assets/14908423/f0222ff1-faea-45e0-95e7-ec68e58f0d6e)

![image](https://github.com/eclipse-platform/eclipse.platform.swt/assets/14908423/b529013c-9d1f-4175-b6ac-2e28360d1ad9)

### To summarize:
When line spacing > 0 is enabled, there are several issues:
1. The spacing is not symmetrical, the text is not centered
2. The caret is not using the full row height.
3. There exist some custom carets in https://github.com/eclipse-platform/eclipse.platform.ui/blob/954bc581051c3f0a5ebcacc1d85492d2acd85707/bundles/org.eclipse.ui.workbench.texteditor/src/org/eclipse/ui/texteditor/AbstractTextEditor.java#L6508-L6516 referring to issue https://bugs.eclipse.org/bugs/show_bug.cgi?id=118612 which needs to be addressed also.
4. The text blue selection is not using the full row height.

1 and 2 can be fixed in `StyledText` / `StyledTextRenderer`.
3 in platform.ui.
4 I am not sure.

PR #1329, opened 2024-07-08
selection background drawn with the line spacing for multiline selection

tobiasmelcher:
Line selection background is drawn without the line spacing value in the StyledText control. For a multiline selection, a zebra effect occurs which looks questionable from my point of view. In this change, the line spacing height is included when drawing the selection background. 

Before with line spacing value 60%:
multiline_selection_line_spacing_before

After with line spacing value 60%:
multiline_selection_line_spacing_after

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/TextLayout.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/TextLayout.java
@@ -498,6 +498,9 @@ public void draw(GC gc, int x, int y, int selectionStart, int selectionEnd, Colo
 					rect.y += pt.y;
 					if (fixedLineMetrics != null) rect.height = fixedLineMetrics.height;
 					rect.height = Math.max(rect.height, ascent + descent);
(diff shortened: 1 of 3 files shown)


BeckerWdf:
contributes to https://github.com/eclipse-platform/eclipse.platform.swt/issues/1286

BeckerWdf:
if nobody objects I will merge this today EoB.

BeckerWdf:
@sratz: Do you have more insights into that? Do you see problems / side effects with this change?

sratz:
> @sratz: Do you have more insights into that? Do you see problems / side effects with this change?

This change looks good.

By not extending the selection of the last line to the bottom, this avoids many other minor visual issues (e.g. if there is also an other highlight applied on the last line, such as a marked occurrence or a search result).

However: This change only fixes it for Win32. What about macOS and Linux? Can the same fix be applied there as well?

tobiasmelcher:
the latest patch request contains implementations for windows, linux and mac.

Windows:
windows

Linux:
linux

Mac:
mac

sratz:
I missed to merge this change for the M3 deadline :(

Any chance this could go in for RC1?

The coding is rather safe, especially in the default 'line spacing 0' case, in which no behavior is changed at all.

merks:
@sratz 

It is fixing a bug and I trust in your good judgement, so I will defer the decision to you with a +1 from me (the PMC).

sratz:
> @sratz
> 
> It is fixing a bug and I trust in your good judgement, so I will defer the decision to you with a +1 from me (the PMC).

Thanks!

PR #1330, opened 2024-07-09
Re-calculating the indent size for Tree on zoom change

ShahzaibIbrahim:
The method calculateAndApplyIndentSize in `Tree ` class is called again and scaled according to the zoom level when moving the shell across different zoom level screens. 

Before the fix: (100 -> 200)

image

After the fix:  (100 -> 200)

image

## HOW TO TEST

```
-Dswt.autoScale=quarter
```
- Switch to the tab "Tree"
- See if the indentation looks normal

## EXPECTED BEHAVIOUR

The list should be indented properly.

Note: There is a visible issue with other zoom level changes as well. i.e. 200 -> 100

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
@@ -5371,7 +5371,7 @@ public void setTopItem (TreeItem item) {
  * In a Tree without imageList, the indent also controls the chevron (glyph) size.
  */
 private void calculateAndApplyIndentSize() {
-	int indent = DPIUtil.autoScaleUpUsingNativeDPI(DEFAULT_INDENT);
+	int indent = DPIUtil.scaleUp(DEFAULT_INDENT, nativeZoom);
 	OS.SendMessage(handle, OS.TVM_SETINDENT, indent, 0);
 }
 
@@ -8297,6 +8297,7 @@ private static void handleDPIChange(Widget widget, int newZoom, float scalingFac
 		DPIZoomChangeRegistry.applyChange(item, newZoom, scalingFactor);
 	}
 
+	tree.calculateAndApplyIndentSize();
 	tree.updateOrientation();
 	tree.setScrollWidth();
 	// Reset of CheckBox Size required (if SWT.Check is not set, this is a no-op)


HeikoKlare (commented):
There are some further usages of `DPIUtil.autoScaleUpUsingNativeDPI()` in the Win32 implementation. Are they all problematic and need to be fixed?

ShahzaibIbrahim:
> There are some further usages of `DPIUtil.autoScaleUpUsingNativeDPI()` in the Win32 implementation. Are they all problematic and need to be fixed?

@HeikoKlare It's not problematic generally but problematic in this case as it tries to take native zoom of primary monitor into consideration all the time.

HeikoKlare:
Alright, that's fine for this PR. Still, the calls to `DPIUtil.autoScaleUpUsingNativeDPI()` seem incompatible with multi-monitor HiDPI support to me. So it might make sense to check whether they are still valid.

ShahzaibIbrahim:
> Alright, that's fine for this PR. Still, the calls to `DPIUtil.autoScaleUpUsingNativeDPI()` seem incompatible with multi-monitor HiDPI support to me. So it might make sense to check whether they are still valid.

I have checked and verified it. It uses the native DPI of current monitor where shell is opened, as expected.

HeikoKlare (changes requested):
> I have checked and verified it. It uses the native DPI of current monitor where shell is opened, as expected.

This statement is not correct. The method does not use the _current_ monitor's native zoom, but the one that is stored in `DPIUtil`. That value only depends on the last DPI change and not on "current" monitor. If you have two shells, there cannot even be a "current" monitor from the `DPIUtils` point of view. As a consequence, this change does not work as expected. You can reproduce it as follows:
1. Have two monitors with different DPI settings.
2. Start an Eclipse product (`DPIUtil.nativeZoom` is set to the zoom of the monitor on which the workbench shell is opened).
3. Take any view of the workbench and drag&drop it to the other monitor (now `DPIUtil.nativeZoom` is changed to the native zoom of that other monitor)
4. Open any view in the original workbench containing a tree (such as the Project Explorer). This will result in broken indents, because the native zoom of the other monitor is taken:

This is how it should look:
image

And this is how it actually looks:
image

From a conceptual point of view, `DPIUtil.autoScaleUpUsingNativeDPI` cannot work with in a multi-monitor setup for above mentioned reasons. Thus I expect that all usages need to be replaced with calls that properly take into account the _current_ monitor and its DPI settings.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
```diff
@@ -5371,7 +5371,7 @@ public void setTopItem (TreeItem item) {
  * In a Tree without imageList, the indent also controls the chevron (glyph) size.
  */
 private void calculateAndApplyIndentSize() {
-	int indent = DPIUtil.autoScaleUpUsingNativeDPI(DEFAULT_INDENT);
+	int indent = DPIUtil.autoScaleUp(DEFAULT_INDENT);
```
akoch-yatta, line 5374: That won't work in multi monitor setup with different zoom levels. You always need to use the (scaleUp) method and passing the zoom to it, in this case the nativeZoom from the Tree

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
```diff
@@ -5371,7 +5371,7 @@ public void setTopItem (TreeItem item) {
  * In a Tree without imageList, the indent also controls the chevron (glyph) size.
  */
 private void calculateAndApplyIndentSize() {
-	int indent = DPIUtil.autoScaleUpUsingNativeDPI(DEFAULT_INDENT);
+	int indent = DPIUtil.scaleUp(DEFAULT_INDENT, getZoom());
```
akoch-yatta, line 5374: we need to keep using the native zoom via the nativeZoom attribute here instead of getZoom() -> Have a look the control example without "-Dswt.autoScale=quarter" on 150% and you'll notice the difference.

HeikoKlare (approved):
Thank you for the correction. For me, it works as expected now.

@akoch-yatta is this also ready to be merged from your side?

akoch-yatta:
> Thank you for the correction. For me, it works as expected now.
> 
> @akoch-yatta is this also ready to be merged from your side?

Yes, looks good for me as well

PR #1332, opened 2024-07-09
[Build] Use general 'native.builder' agent labels and simplify definition of native build agent for gtk.linux.x86_64

HannesWell:


akurtakov:
Keep these simplifications coming - this is exactly the work needed to bring some sanity in the releng side.

HannesWell:
The initial idea to fully define all build dependencies for `linux.x86_64` in the pipeline and avoid the need for custom docker-images didn't work out, but at least this simplifies the configuration a little bit.

Furthermore this now also ensures that the workspace of the native build agents is cleaned up after the build.

PR #1336, opened 2024-07-10
Return empty instead of null if FileDialog fails to open

kysmith-csg:
For Cocoa if `NSSavePanel` or `NSOpenPanel` can't be created, return an empty Optional instead of `null` to prevent a `NullPointerException` in `FileDialog.open()`.

Fixes an issue with #588.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/FileDialog.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/FileDialog.java
@@ -383,7 +383,7 @@ public Optional<String> openDialog () {
 	fullPath = null;
 	if ((style & SWT.SAVE) != 0) {
 		NSSavePanel savePanel = NSSavePanel.savePanel();
-		if (savePanel == null) return null;
+		if (savePanel == null) error(SWT.ERROR_INVALID_RETURN_VALUE);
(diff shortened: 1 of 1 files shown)


mickaelistria:
I think an exception in that case is the proper behavior.

HeikoKlare:
Returning an empty Option is invalid, as it violates the method's contract. The method has been introduced to be able to identify that the user has canceled the dialog (which is indicated by an empty Optional).

I see two options to fix the problem:
1. Correct the `open()` implementation to consider that `openDialog()` may return null:
https://github.com/eclipse-platform/eclipse.platform.swt/blob/08e9fa0dbada988b5716911337362610301f98cc/bundles/org.eclipse.swt/Eclipse%20SWT/cocoa/org/eclipse/swt/widgets/FileDialog.java#L358
2. Throw an exception (as proposed by @mickaelistria)

Actually, I think solution (1) is the "correct" bugfix, as it restores the method's behavior from before #1026, where `null` was "properly" returned by the `open()` method in this situation (rather than having a `NullPointerException`).

laeubi (approved):
Throwing an exception for this case sounds correct as it allows the caller to potentially handle this (exceptional) case if desired.

HeikoKlare (approved):
Looks good, thank you! The current solution also restores the behavior  (`open()` returning null in the addressed case) from before #1026 👍

PR #1337, opened 2024-07-10
 Swing Port from SWTSwing

fedejeanne:
This is a first attempt to integrate the existing SWTSwing (https://github.com/nu11ptr/SWTSwing).

Current status:
* The fragment itself compiles but it introduces a **breaking API change** by using a `CGC` in `Drawable` as opposed to using handles (`long`s) like the other fragments do
* Some snippets run
* `StyledText` and `Browser` do not work so I simply commented them out
* I still haven't managed to start the plugin "cleanly", I have to replace some other fragment with the code from this plugin. This can probably be done better if we follow the same approach as in the GTK port for windows _i.e_ remove other fragments from the launch configuration, change the platform of the other fragments to something other than your current platform and change the platform of this plugin to the one that matches your current system.


fedejeanne:
Wrong repo

PR #1339, opened 2024-07-11
Refactoring autoScaleUp with zoom to scaleUp

amartya4256:
This PR contributes to the renaming of methods DPIUtil.autoScaleUp containing int zoom as an argument to DPIUtil.scaleUp, since the scaling is not done automatically but based on the zoom level which is provided.

Contributes to #62 and #127

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
@@ -257,8 +257,8 @@ public static Rectangle autoScaleDown(Rectangle rect) {
 public static Rectangle scaleDown(Rectangle rect, int zoom) {
 	if (zoom == 100 || rect == null) return rect;
 	Rectangle scaledRect = new Rectangle (0,0,0,0);
-	Point scaledTopLeft = DPIUtil.scaleDown(new Point (rect.x, rect.y), zoom);
-	Point scaledBottomRight = DPIUtil.scaleDown(new Point (rect.x + rect.width, rect.y + rect.height), zoom);
+	Point scaledTopLeft = scaleDown(new Point (rect.x, rect.y), zoom);
+	Point scaledBottomRight = scaleDown(new Point (rect.x + rect.width, rect.y + rect.height), zoom);
 
 	scaledRect.x = scaledTopLeft.x;
 	scaledRect.y = scaledTopLeft.y;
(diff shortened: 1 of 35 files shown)


bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
```diff
@@ -455,26 +455,26 @@ public static Point autoScaleUp(Point point, int zoom) {
...
 }
 
-public static Rectangle autoScaleUp(Rectangle rect, int zoom) {
+public static Rectangle scaleUp(Rectangle rect, int zoom) {
 	if (zoom == 100 || rect == null) return rect;
 	Rectangle scaledRect = new Rectangle(0,0,0,0);
-	Point scaledTopLeft = DPIUtil.autoScaleUp (new Point(rect.x, rect.y), zoom);
-	Point scaledBottomRight = DPIUtil.autoScaleUp (new Point(rect.x + rect.width, rect.y + rect.height), zoom);
+	Point scaledTopLeft = DPIUtil.scaleUp (new Point(rect.x, rect.y), zoom);
+	Point scaledBottomRight = DPIUtil.scaleUp (new Point(rect.x + rect.width, rect.y + rect.height), zoom);
```
HeikoKlare, line 477: nitpicky:
```suggestion
	Point scaledTopLeft = scaleUp (new Point(rect.x, rect.y), zoom);
	Point scaledBottomRight = scaleUp (new Point(rect.x + rect.width, rect.y + rect.height), zoom);
```

HeikoKlare (commented):
Change is sound. It's hard to check every line of code, but I guess you simply applied the IDE refactoring?

As a follow-up, I propose to also align `autoScaleImageData` and `autoScaleBounds`, so that the naming is consistent throughout `DPIUtil`.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
amartya4256: Done, also refactored the similar thing in a few more places.

amartya4256:
@HeikoKlare Should I also do the changes for autoScaleBounds in the same PR? Since its just refactoring and done by the IDE anyways.

HeikoKlare:
> Should I also do the changes for autoScaleBounds in the same PR? Since its just refactoring and done by the IDE anyways.

Let's keep the PR's scope as small as possible and have separate PRs for other (similar) refactorings. That's easier to review and allows to revert less changes (if necessary).

PR #1345, opened 2024-07-15
Rename auto scale image data

amartya4256:
This PR renames the method autoScaleImageData to scaleImageData in DPIUtil for the variants of the method which accept a target zoom since it sounds more relevant to the action.

Contributes to #62 and #127


fedejeanne (dismissed):
Since `DPIUtil` is in `common`, you need to adapt the Mac and Linux fragments too (hence the compilation errors)

amartya4256:
> Since `DPIUtil` is in `common`, you need to adapt the Mac and Linux fragments too (hence the compilation errors)

The changes has been made. Let's wait for the pipelines. @fedejeanne

fedejeanne (approved):

PR #1347, opened 2024-07-17
Adding disable/gray image props in ImageData when rescaled

ShahzaibIbrahim:
Disabled image do not work when rescaled. Writing a new method `copyImageDataForDisabledOrGrayImages` and calling it before init image in win32_getHandle. 

## HOW TO TEST

Put the following snippet into the snippet folder to test the effect:

```
 * Copyright (c) 2015, 2018 IBM Corporation and others.
 *
 *
 *
 * Contributors:
 *     IBM Corporation - initial API and implementation

import org.eclipse.swt.*;

public class Snippet382 {
	private static final String IMAGE_100 = "eclipse16.png";
	private static final String IMAGE_150 = "eclipse24.png";
	private static final String IMAGE_200 = "eclipse32.png";
	private static final String IMAGES_ROOT = "bin/org/eclipse/swt/snippets/";

	private static final String IMAGE_PATH_100 = IMAGES_ROOT + IMAGE_100;
	private static final String IMAGE_PATH_150 = IMAGES_ROOT + IMAGE_150;
	private static final String IMAGE_PATH_200 = IMAGES_ROOT + IMAGE_200;

		final ImageFileNameProvider filenameProvider = zoom -> {
			switch (zoom) {
			case 100:
				return IMAGE_PATH_100;
			case 150:
				return IMAGE_PATH_150;
			case 200:
				return IMAGE_PATH_200;
			default:
				return null;
			}
		};
		final ImageDataProvider imageDataProvider = zoom -> {
			switch (zoom) {
			case 100:
				return new ImageData (IMAGE_PATH_100);
			case 150:
				return new ImageData (IMAGE_PATH_150);
			case 200:
				return new ImageData (IMAGE_PATH_200);
			default:
				return null;
			}
		};

		final Display display = new Display ();
		final Shell shell = new Shell (display);
		shell.setText("Snippet367");
		shell.setLayout (new GridLayout (3, false));
		Listener l = new Listener() {
			@Override
			public void handleEvent(Event e)  {
				if (e.type == SWT.Paint) {
					GC mainGC = e.gc;
					GCData gcData = mainGC.getGCData();
					final Image imageWithFileNameProvider = new Image (display, filenameProvider);
					final Image disabledImageWithFileNameProvider = new Image (display,imageWithFileNameProvider, SWT.IMAGE_DISABLE);
					final Image greyImageWithFileNameProvider = new Image (display,imageWithFileNameProvider, SWT.IMAGE_GRAY);

					final Image imageWithDataProvider = new Image (display, imageDataProvider);
					final Image disabledImageWithDataProvider = new Image (display,imageWithDataProvider, SWT.IMAGE_DISABLE);
					final Image greyImageWithDataProvider = new Image (display,imageWithDataProvider, SWT.IMAGE_GRAY);

					final Image imageWithData = new Image (display, IMAGE_PATH_100);
					final Image disabledImageWithData = new Image (display,imageWithData, SWT.IMAGE_DISABLE);
					final Image greyImageWithData = new Image (display,imageWithData, SWT.IMAGE_GRAY);

					try {
						drawImages(mainGC, gcData, "Normal",40, imageWithFileNameProvider);
						drawImages(mainGC, gcData, "Disabled",80, disabledImageWithFileNameProvider);
						drawImages(mainGC, gcData, "Greyed",120, greyImageWithFileNameProvider);

						drawImages(mainGC, gcData, "Normal",160, imageWithDataProvider);
						drawImages(mainGC, gcData, "Disabled",200, disabledImageWithDataProvider);
						drawImages(mainGC, gcData, "Greyed",240, greyImageWithDataProvider);

						drawImages(mainGC, gcData, "Normal",280, imageWithDataProvider);
						drawImages(mainGC, gcData, "Disabled",320, disabledImageWithData);
						drawImages(mainGC, gcData, "Greyed",360, greyImageWithData);
					} finally {
						mainGC.dispose ();
					}
				}
			}

			private void drawImages(GC mainGC, GCData gcData, String text, int y, final Image imageWithFileNameProvider) {
				gcData.nativeZoom = 100;
				mainGC.drawText(text, 0, y);
				mainGC.drawImage(imageWithFileNameProvider, 50, y);
				gcData.nativeZoom = 150;
				mainGC.drawImage(imageWithFileNameProvider, 100, y);
				gcData.nativeZoom = 200;
				mainGC.drawImage(imageWithFileNameProvider, 150, y/2);
			}
		};
		shell.addListener(SWT.Paint, l);




		shell.setSize(600, 600);
		shell.open ();
		}
		display.dispose ();
	}

}

```

image


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -737,6 +737,150 @@ public Image(Device device, ImageDataProvider imageDataProvider) {
...
+
+/**
+ * Upon zoom update, this method copies the image data and add disabled or grayed image depending on the flag
+ *
+ * @param device the device on which to create the image
+ * @param data the image data to create the image from (must not be null)
+ * @param flag the style, either <code>IMAGE_COPY</code>, <code>IMAGE_DISABLE</code> or <code>IMAGE_GRAY</code>
+ * @return new image data with valid styling properties as per the flag
+ */
+private ImageData copyImageDataForDisabledOrGrayImages(Device device, ImageData data, int flag) {
```
akoch-yatta, line 749: Isn't that mostly duplicated what is in one of the Image constructor? I see there are some differences here and there, but it should be possible to extract the logic for IMAGE_DISABLE into one method and use it it both places and do the same for IMAGE_GRAY

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -685,7 +685,7 @@ public Image(Device device, ImageFileNameProvider imageFileNameProvider) {
 	initialNativeZoom = DPIUtil.getNativeDeviceZoom();
 	ElementAtZoom<String> fileName = DPIUtil.validateAndGetImagePathAtZoom (imageFileNameProvider, getZoom());
 	if (fileName.zoom() == getZoom()) {
-		long handle = initNative (fileName.element(), getZoom());
+		long handle = initNative (fileName.element(), getZoom(), SWT.IMAGE_COPY);
```
akoch-yatta, line 688: Hmm, we're not really wanting to copy something here. I would e.g. use SWT.DEFAULT (-1) here and change the default case in the switch in  copyImageDataForDisabledOrGrayImages like
```
default: {
	returnImageData = data;
	break;
}
```
I'm not 100% sure about that, but using SWT.IMAGE_COPY (just from the semantics) sounds not correct either.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -763,20 +907,22 @@ public static Long win32_getHandle (Image image, int zoom) {
...
+			long handle = image.initNative(imageCandidate.element(), zoom, image.styleFlag);
+			if (handle == 0) image.init(imageData, zoom);
 			image.init();
 		} else {
-			ImageData resizedData = DPIUtil.autoScaleImageData (image.device, new ImageData (imageCandidate.element()), zoom, imageCandidate.zoom());
-			image.init(resizedData, zoom);
-			image.init ();
+			ImageData resizedData = DPIUtil.autoScaleImageData (image.device, imageData, zoom, imageCandidate.zoom());
+			ImageData newData = image.copyImageDataForDisabledOrGrayImages(image.getDevice(), resizedData, image.styleFlag);
+			image.init( newData, zoom);
```
akoch-yatta, line 919: remove the whitespace after (

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
ShahzaibIbrahim: You are right, but I wanted to stay away from default. It could give the users a wrong impression if they use some wrong flag. It's just my opinion.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -785,14 +829,15 @@ public static Long win32_getHandle (Image image, int zoom) {
 		}
 		if (image.dataAtBaseZoom != null) {
 			ImageData resizedData = image.getImageData(zoom);
-			image.init(resizedData, zoom);
+			ImageData newData = image.copyImageDataForDisabledOrGrayImages(image.getDevice(), resizedData, image.styleFlag);
```
akoch-yatta, line 832: I don't think it is neccessary to apply that here as well. As the image data is directly used from the existing image, I think it should already be greyed or disabled and only needs to be rescaled

ShahzaibIbrahim: The imageData is not being directly used from the existing image (it is recalculated and not stored in as a variable). `getImageData()` method is calling `autoScaleImageData()` and it's returning a new instance of ImageData where we lose the gray/disabled property again. You can try it out by commenting the line I added.

akoch-yatta: Ah yes, you are right

fedejeanne (changes requested):
I tested it and it has the expected behavior ✔️ 

There's only one small typo (missing "e") and I would like to have the snippet also contributed. You need to set the name of the dialog properly though: `shell.setText("Snippet382");`

examples/org.eclipse.swt.snippets/Snippets.md
```diff
@@ -213,6 +213,7 @@ To contribute a new snippet, [create a snippet contribution as a pull request](h
...
 - [draw an image scaled to half size and double size](https://github.com/eclipse-platform/eclipse.platform.swt/tree/master/examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet355.java) – [(preview)](https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/examples/org.eclipse.swt.snippets/previews/Snippet355.png "Preview for Snippet 355")
 - [draw an image at various zoom/dpi levels](https://github.com/eclipse-platform/eclipse.platform.swt/tree/master/examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet367.java) – [(preview)](https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/examples/org.eclipse.swt.snippets/previews/Snippet367.png "Preview for Snippet 367")
+- [draw a disabled/gray image at various zoom/dpi levels](https://github.com/eclipse-platform/eclipse.platform.swt/tree/master/examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet382.java) – [(preview)](https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/examples/org.eclipse.swt.snippets/previews/Snippet382.png "Preview for Snippet 382")
```
fedejeanne, line 216:
```suggestion
- [draw a disabled/grayed image at various zoom levels](https://github.com/eclipse-platform/eclipse.platform.swt/tree/master/examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet382.java) – [(preview)](https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/examples/org.eclipse.swt.snippets/previews/Snippet382.png "Preview for Snippet 382")
```

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet382.java
```diff
@@ -0,0 +1,124 @@
...
+				mainGC.drawImage(imageWithFileNameProvider, 50, y);
+				gcData.nativeZoom = 150;
+				mainGC.drawImage(imageWithFileNameProvider, 100, (int) (y/1.5));
+				gcData.nativeZoom = 200;
+				mainGC.drawImage(imageWithFileNameProvider, 150, y/2);
+			}
+		};
+		shell.addListener(SWT.Paint, l);
+
+		shell.setSize(600, 600);
```
fedejeanne, line 116: Less unnecessary borders
```suggestion
		shell.setSize(400, 500);
```

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet382.java
```diff
@@ -0,0 +1,124 @@
...
+import org.eclipse.swt.graphics.*;
+import org.eclipse.swt.layout.*;
+import org.eclipse.swt.widgets.*;
+
+/**
+ * Snippet to test the use of the DPI-aware Image constructors with disabled and gray images
+ * Work in progress in https://bugs.eclipse.org/399786
+ * <p>
+ * For a list of all SWT example snippets see
+ * http://www.eclipse.org/swt/snippets/
```
fedejeanne, line 26:
```suggestion
 * <p>
 * For a list of all SWT example snippets see
 * http://www.eclipse.org/swt/snippets/
 * </p>
```

fedejeanne (changes requested):
I corrected some stuff and created a new screenshot without borders that you can use:

image

fedejeanne (approved):

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -738,6 +627,159 @@ public Image(Device device, ImageDataProvider imageDataProvider) {
...
+		case SWT.IMAGE_GRAY: {
+			ImageData newData = applyGrayImageData(data, data.height, data.width);
+			returnImageData = newData;
+			break;
+		}
+		default: {
+			returnImageData = data;
+			break;
+		}
+	}
```
HeikoKlare, line 656: Wouldn't this implementation be simpler and also sufficient?
```suggestion
	switch (flag) {
		case SWT.IMAGE_DISABLE: {
			return applyDisableImageData(device, data, data.height, data.width);
		}
		case SWT.IMAGE_GRAY: {
			return applyGrayImageData(data, data.height, data.width);
		}
		default: {
			return data;
		}
	}
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -738,6 +627,159 @@ public Image(Device device, ImageDataProvider imageDataProvider) {
 	init();
 }
 
+
+/**
+ * Upon zoom update, this method copies the image data and add disabled or grayed image depending on the flag
```
HeikoKlare, line 632: This method is independent from zoom updates, isn't it? So the documentation should not refer to consumers but only to what the method is supposed to do.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -738,6 +627,159 @@ public Image(Device device, ImageDataProvider imageDataProvider) {
...
+ *
+ * @param device the device on which to create the image
+ * @param data the image data to create the image from (must not be null)
+ * @param flag the style, either <code>IMAGE_DISABLE</code> or <code>IMAGE_GRAY</code>
+ * @return new image data with valid styling properties as per the flag
+ */
+private ImageData copyImageDataForDisabledOrGrayImages(Device device, ImageData data, int flag) {
```
HeikoKlare, line 639: I found the changes hard to comprehend at first sight, because the method does not do what it's name suggests: it does not copy data (only) for disabled or gray images, but transforms the data in these two cases and simply returns the original data in other cases. So maybe this should be somethink like `adaptImageDataIfDisabledOrGray()`?

HeikoKlare, line 639: Why does this method take a device and a flag? Can't this information be taken from the image (`this`) itself? For the device parameter, I see no caller using another value than the device of the image, and for the flag parameter it is only one case where `SWT.Default` is passed explicitly, even though the image's default `SWT.IMAGE_COPY` should lead to the same result here.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -738,6 +627,159 @@ public Image(Device device, ImageDataProvider imageDataProvider) {
 	init();
 }
 
+
+/**
+ * Upon zoom update, this method copies the image data and add disabled or grayed image depending on the flag
+ *
+ * @param device the device on which to create the image
+ * @param data the image data to create the image from (must not be null)
+ * @param flag the style, either <code>IMAGE_DISABLE</code> or <code>IMAGE_GRAY</code>
```
HeikoKlare, line 636: This documentation is incorrect. The flag may have any value, just that the method has some specific behavior for these two values.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -738,6 +627,159 @@ public Image(Device device, ImageDataProvider imageDataProvider) {
 	init();
 }
 
+
+/**
+ * Upon zoom update, this method copies the image data and add disabled or grayed image depending on the flag
+ *
+ * @param device the device on which to create the image
```
HeikoKlare, line 634: This documentation is incorrect, as the method does not create any image at all.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -738,6 +627,159 @@ public Image(Device device, ImageDataProvider imageDataProvider) {
...
+ * @param data the image data to create the image from (must not be null)
```
HeikoKlare, line 635: This documentation is incorrect, as the method does not create any image.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -786,14 +830,15 @@ public static Long win32_getHandle (Image image, int zoom) {
...
+			ImageData newData = image.copyImageDataForDisabledOrGrayImages(image.getDevice(), resizedData, image.styleFlag);
+			image.init(newData, zoom);
 			image.init();
 		}
 	}
 	return image.zoomLevelToHandle.get(zoom);
 }
 
-long initNative(String filename, int zoom) {
+long initNative(String filename, int zoom, int imageStyle) {
```
HeikoKlare, line 841: Can you please explain a concrete scenario in which passing the `imageStyle` is necessary compared to just using the existing style of the image?

HeikoKlare (commented):
From a functional perspective, the changes look sound to me. Still the change seem to introduce some unnecessary complexity (or I do not understand the necessity for the complexity so far).

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
ShahzaibIbrahim: Yes, that's more clear.

ShahzaibIbrahim: I passed device and a flag because it could be from another object of an image, the win32_getHandle method passes different object of the Image which we get from GC. 

For the second part, initially I did use the SWT.IMAGE_COPY as you suggested but andy had a good point in his comment above that we are not actually copying anything and this flag is kind of gives different idea.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
ShahzaibIbrahim: I am kind of  confused here, shouldn't I just mention in docs what values should be passed to this method for correct usage?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
ShahzaibIbrahim: As I mentioned in the comment above, `win32_getHandle `has a different object for an image. We need `imageStyle `of that particular image object which is different from `this`

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
HeikoKlare: Maybe I miss something, but the calls of the method I see always looks like this: `image.copyImageDataForDisabledOrGrayImages(image.getDevice(), resizedData, image.styleFlag)`
So you extract the device and the flags out of the image on which you call the method. Only difference is the `initNative` method, but there the device is also retrieved from `this` and the flag is always retrieved from the image (i.e., `this`) by the caller, expect for one caller using `SWT.DEFAULT`, but that could also simply reduce to the default flag of the image begin `SWT.IMAGE_COPY`. So I do not understand why there parameter are necessary at all and why simply using `this.getDevice()` and `this.styleFlag` is not sufficient.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
HeikoKlare: I was just confused because I found it unnecessary complex. But it's of course okay as it is and rather a matter of taste.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
HeikoKlare: Yes, but the statement "the device on which to create the image" implies that the method creates an image, but it does nothing like that.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
HeikoKlare: I don't get that one. Yes ,`win32_getHandle` calls this method for another image, but the metohd is then executed on exactly that instance, so you can simply use the values from that instance.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
ShahzaibIbrahim: Sorry this comment was for this review: 
I am kind of confused here, shouldn't I just mention in docs what values should be passed to this method for correct usage?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
ShahzaibIbrahim: Now I get your point, you're right. We could just use the `this` since it was called by that image instances.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
HeikoKlare: That's correct, but you call the method also with other flags and that's also a correct usage, isn't it? The behavior is as follows: for these two flags, the image applies changes to the given image data, for other flags it does not (but still they are valid).

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
fedejeanne: @ShahzaibIbrahim as a rule of thumb: if the method is `static` then pass the state as parameters and if it's not then use the internal state (fields) :-)

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -738,6 +627,157 @@ public Image(Device device, ImageDataProvider imageDataProvider) {
 	init();
 }
 
+
+/**
+ * This method copies the image data and add disabled or grayed image depending on the flag
+ * @param data the image data from which original data is copied
+ *
+ * @return new image data with valid styling properties as per the flag
+ */
```
HeikoKlare, line 636: Please either revise the comment again, as it is not fitting after the recent API changes anymore, or simply remove it (as the method is private and the name+implementation is easy to comprehend anyway).
- "the flag" is unspecified, as the method does not except a flag. That would have to be "this image's flags" or the like
- data for disabled/grayed is not added but just overwrites the "normal" image data, doesn't it? So the statement about that the method "copies the image data" is not corret, as it is either copied and overwrites it or simply returns the original one (without performing a copy).

I would be in favor of simply having this documentation removed.

HeikoKlare (commented):
Thank you for the recent improvements. The PR looks much better now. Changes are easy to comprehend and look sound. I only have one comment remaining w.r.t. documentation, but since it is about a private method that is easy to comprehend by its name and implementation, I propose to simply remove the potentially irritating documentation.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
ShahzaibIbrahim: I will remove it then

iloveeclipse:
Commit ba97568f5e037755c23b067c6587325610b6b189 broke compilation of snippets project on Linux. Please next time exclude platform specific sources in the Snippets project from classpath on the platforms where the source can't compile.

I've created https://github.com/eclipse-platform/eclipse.platform.swt/pull/1372.

@HeikoKlare:  I wonder if we can make the `org.eclipse.swt.snippets` compilable with Tycho (with some extra config?) that would apply OS specific .classpath fixes to .classpath. I'm not a maven expert, but it would work with good old ant for sure :-)

laeubi:
I once opened some requests to JDT to support class folders with different settings so one not need copy around classpath settings or other hacks:
- https://github.com/eclipse-jdt/eclipse.jdt.core/issues/265
- https://github.com/eclipse-jdt/eclipse.jdt.core/issues/642

Also see this related Tycho issues:
- https://github.com/eclipse-tycho/tycho/issues/632
- https://github.com/eclipse-tycho/tycho/issues/2256
- https://github.com/eclipse-tycho/tycho/issues/253

In the end such issues are that seldom that no one really cared to work on that.

HeikoKlare:
Thank you for the fix, @iloveeclipse! I missed that the snippet uses OS-specific functionality. I guess that it also breaks on MacOS. I will check that as a follow-up.

It would of course be great to have automated checks for this kind of problems in the build / Tycho. But I am not sure whether it is worth the effort (as indicated by Christoph's comment) or if it would not be even better to introduce some (better) way of declaring and statically checking (already in the IDE) what is part of the SWT API. E.g., having warnings for access of classes/methods marked as `@noreference` would have made the problem obvious in the IDE (by a warning marker) and in the builds (by introducing an additional warning).

HeikoKlare:
MacOS is also affected. I've created #1373 to fix that.

laeubi:
> E.g., having warnings for access of classes/methods marked as @noreference would have made the problem obvious in the IDE (by a warning marker) and in the builds (by introducing an additional warning).

I think this would usually happen if we enable API analysis, but here again, the question is how much effort one want to invest compared to how often are snippets added.

I usually simply have an own project in my workspace named "swt-test" that I use to write a snippet then I copy it over (e.g. for a bugreport or for permanent usage), that works "good enough" at least :-)

HeikoKlare:
Fully agree that it's a matter of ROI and if this was only relevant for snippets, I agree that only low effort would be worth the benefit in the rather seldom case of having changes/additions to snippets.

I have to admit that I just understood @iloveeclipse's comment completely: the snippet project is not part of the Tycho build at all. I was not aware of that. Doing that should be quite easy and thus a quick win (maybe as an intermediate solution until the requests mentioned by @laeubi for JDT/Tycho have been processed). I will have a look at that.

HeikoKlare:
> I have to admit that I just understood @\iloveeclipse's comment completely: the snippet project is not part of the Tycho build at all. I was not aware of that. Doing that should be quite easy and thus a quick win (maybe as an intermediate solution until the requests mentioned by @\laeubi for JDT/Tycho have been processed). I will have a look at that.

Works fine, but requires at least:
- https://github.com/eclipse-tycho/tycho/issues/253

@laeubi can you estimate how difficult it would be to implement that in Tycho?

PR #1356, opened 2024-07-22
Initialize Win32 Display thread with correct OS DPI awareness 

HeikoKlare:
The runtime auto-scaling for multi-monitor HiDPI support on Windows requires the system's DPI awareness to be set to PerMonitorV2. This change makes the Display class initialize its UI thread with the according DPI awareness.

Contributes to https://github.com/eclipse-platform/eclipse.platform.swt/issues/62

Contributes to https://github.com/eclipse-platform/eclipse.platform.swt/issues/131

<s>⚠️  Depends on and has to be merged after #1355</s>

### How to test

You need to start an application with a different DPI awareness mode than "Per Monitor V2", e.g., by changing the JVMs manifest, with the runtime autoscaling activated (VM argument `-Dswt.autoScale.updateOnRuntime=true`).
I have added a manifest with DPI awareness set to "System" to my JVM and started SWT Snippet 106 on a 100% monitor and moved to a 150% monitor. Here is how it looks like.

Without this change:
image

With this change:
image

To avoid confusion, the shrunken size is because of the default scaling method `integer200` which "corrects" the widgets sizes to 100% scaling. This is existing SWT behavior. With scaling mode set to `quarter`, it looks like this:
image

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
@@ -582,6 +582,29 @@ public Display () {
  */
 public Display (DeviceData data) {
 	super (data);
+	initializeProperDPIAwareness();
+}
+
+private void initializeProperDPIAwareness() {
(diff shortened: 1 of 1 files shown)


akoch-yatta:
Change looks good to me. I tested it a bit and it worked as expected as well.

PR #1342, opened 2024-07-12
Replace dpiutil scale calls by scale with zoom calls

amartya4256:
Depends on #1339

PR #1357, opened 2024-07-23
Remove and replace DPIUtil.autoScaleUpUsingNativeDpi

amartya4256:
This contribution replaces the DPIUtil.autoScaleUpUsingNativeDPI calls
from clients in win32 with DPIUtil.scaleUp utilizing the zoom
available within the clients. It additionally removes the method
DPIUtil.autoScaleUpUsingNativeDPI since it's not used anymore.

Note: Only the last PR is relevant. Depends on #1342


HeikoKlare (approved):
Looks good, thank you!

PR #1358, opened 2024-07-23
Assign Windows runtime rescaling to Display instead of DPIUtil

HeikoKlare:
Currently, the runtime rescaling of shells/widgets on Windows is tied to a property in the DPIUtil class. Since that class is internal, the property can only be set via a VM argument at startup but cannot be changed or at least retrieved during runtime from outside the SWT bundle.

This change reduces the DPIUtil property to a global setting that may be used to set the behavior for a whole application (in particular an SWT application) but move the actual state handling for runtime rescaling behavior to the Display class. This allow to retrieve and change the value directly on the Display, on which also the DPI awareness for the UI thread is initialized. As a side effect, it improves testability as a properly configured display can be used instead of changing the global state of DPIUtil.


bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
```diff
@@ -6832,4 +6832,30 @@ static long windowProc(long id, long sel, long arg0, long arg1, long arg2, long
...
+/**
+ * Activate or deactivates rescaling of shells at runtime whenever the DPI
+ * scaling of the shell's monitor changes. This is only safe to call as long a
+ * no shell has been created for this display. When changing the value after a
+ * shell has been created for this display, the effect is undefined.
+ * <p>
+ * <b>Note:</b> This functionality is only available on Windows. Calling this
+ * method on other operating system will have no effect.
+ *
+ * @param activate whether rescaling shell be activated or deactivated
```
akoch-yatta, line 6855: I would say either 'sh**a**ll' or 'rescaling of shells' 🤔

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
```diff
@@ -6832,4 +6832,30 @@ static long windowProc(long id, long sel, long arg0, long arg1, long arg2, long
...
+ * shell's monitor changes is activated for this device}
+ *
+ * @since 3.127
+ */
+public boolean isRescalingAtRuntime() {
+	return false;
+}
+
+/**
+ * Activate or deactivates rescaling of shells at runtime whenever the DPI
```
akoch-yatta, line 6847: Activate**s**

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
```diff
@@ -6832,4 +6832,30 @@ static long windowProc(long id, long sel, long arg0, long arg1, long arg2, long
...
+ * scaling of the shell's monitor changes. This is only safe to call as long a
```
akoch-yatta, line 6848: as long a**s**

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
```diff
@@ -138,6 +138,7 @@ public class Display extends Device implements Executor {
 	EventTable eventTable, filterTable;
 	boolean useOwnDC;
 	boolean externalEventLoop; // events are dispatched outside SWT, e.g. TrackPopupMenu or DoDragDrop
+	boolean rescalingAtRuntime;
```
akoch-yatta, line 141: as this field is exposed via API, we should make it private, shouldn't we? Especially as set setter contains important logic

akoch-yatta (commented):
I like inlining that logic into Display. Especially as it makes it possible to change the setting on runtime

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
```diff
@@ -6832,4 +6832,30 @@ static long windowProc(long id, long sel, long arg0, long arg1, long arg2, long
 static boolean isActivateShellOnForceFocus() {
 	return "true".equals(System.getProperty("org.eclipse.swt.internal.activateShellOnForceFocus", "true")); //$NON-NLS-1$
 }
+
+/**
+ * {@return whether rescaling of shells at runtime when the DPI scaling of a
+ * shell's monitor changes is activated for this device}
```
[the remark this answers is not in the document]
HeikoKlare: Yes, makes sense. I thought we might leave that out because the method always correctly returns "false", but I don't see why we should not add an according hint here as well. I've added it with 784ffc7.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
HeikoKlare: It was supposed to be "shall". I just mixed up the two words 😉 
Fixed with 784ffc7.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
HeikoKlare: Corrected in 784ffc7.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
HeikoKlare: Corrected in 784ffc7.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
HeikoKlare: Totally makes sense. I just forgot about that because all other fields in the class are package protected, but making it private is totally correct. Fixed in 784ffc7.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
```diff
@@ -6832,4 +6832,30 @@ static long windowProc(long id, long sel, long arg0, long arg1, long arg2, long
...
+public boolean isRescalingAtRuntime() {
+	return false;
+}
+
+/**
+ * Activate or deactivates rescaling of shells at runtime whenever the DPI
+ * scaling of the shell's monitor changes. This is only safe to call as long a
+ * no shell has been created for this display. When changing the value after a
+ * shell has been created for this display, the effect is undefined.
+ * <p>
```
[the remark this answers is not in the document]
HeikoKlare: The closing `</p>` is optional in HTML and we usually leave it out in Javadoc.

PR #1361, opened 2024-07-24
[win32] Use SWTFontProvider in GC fonts

akoch-yatta:
This commit adapts how to fetch fonts in GC in the win32 implementation. Now it always uses the internal SWTFontProvider to fetch system and non-system fonts in the correct scaling.

Currently the systemFont was not scaled at all (although it is almost used nowhere in this context anyway) and in general in makes sense to use the SWTFontProvider anyhow for that

Contributes to #62 and #131


HeikoKlare (approved):
Thank you! Change looks sound.

PR #1362, opened 2024-07-24
Create CTabFolder images with ImageDataProvider

akoch-yatta:
This commit changes the createButtonImage method in CTabFolder. Until now the requested image data were created once and are scaled to other zoom settings with an AutoScaleImageDataProvider, which is a destructive operation when scaling it to a smaller zoom. With this change the whole logic is moved into an ImageDataProvider, so each time ImageData for a specific zoom value is requested, it will be created from scratch for this zoom value.

If you have the rescaling enabled in win32 switching from 200% to 100% would shrink the image to something like
image

With the change it looks like 
image

This is affected a common widget, so it must be tested on all OS. As I currently don't have a working Linux or MacOS available, I cannot do it myself.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
@@ -17,8 +17,6 @@ import org.eclipse.swt.*;
 import org.eclipse.swt.accessibility.*;
 import org.eclipse.swt.events.*;
 import org.eclipse.swt.graphics.*;
-import org.eclipse.swt.internal.*;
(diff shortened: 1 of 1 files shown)


HeikoKlare (approved):
Change looks sound and works fine on MacOS and Linux as well. Tested with multi-monitor setups using combinations of 100%/200% and also fractional scaling on Ubuntu with 100%/200% as well as 100%/150% setups.

deepika-u:
I verified on below environment 
-------------------------------- 
Eclipse SDK
Version: 2024-09 (4.33)
Build id: I20240822-0100
Java vendor: Eclipse Adoptium
Java runtime version: 19.0.2+7
Java version: 19.0.2

I tried shifting between 100% to 200% and 200% to 175%, 150%,125% and dont see any problem on windows 11.

PR #1363, opened 2024-07-24
Only consider rescaling at runtime enabled when DPI awareness mode fits

HeikoKlare:
Activating rescaling at runtime requires a proper DPI awareness mode to be set. Currently, if setting the DPI awareness mode fails, rescaling may still be activated if the user requested to. With this change, setting the rescaling mode of a Display ensures that the correct DPI awareness for the UI thread is set and, in case an error occurs, the rescaling mode is not changed. It also adapts some faulty constants and provides according test cases for setting the rescaling behavior.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/DisplayWin32Test.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/DisplayWin32Test.java
@@ -1,0 +1,55 @@
+package org.eclipse.swt.widgets;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+import org.eclipse.swt.internal.*;
+import org.eclipse.swt.internal.win32.*;
+import org.junit.jupiter.api.*;
+
+public class DisplayWin32Test {
+
+	private Display display;
+
+	@BeforeAll
+	public static void assumeIsFittingPlatform() {
+		PlatformSpecificExecution.assumeIsFittingPlatform();
+	}
+
+	@BeforeEach
+	public void createDisplay() {
+		display = new Display();
+	}
+
+	@AfterEach
+	public void destroyDisplay() {
+		display.dispose();
+	}
+
+	@Test
+	public void setRescaleAtRuntime_activate() {
+		display.setRescalingAtRuntime(true);
+		assertTrue(display.isRescalingAtRuntime());
+		assertEquals(OS.DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2, OS.GetThreadDpiAwarenessContext());
+	}
+
+	@Test
+	public void setRescaleAtRuntime_deactivate() {
+		display.setRescalingAtRuntime(false);
+		assertFalse(display.isRescalingAtRuntime());
+		assertEquals(OS.DPI_AWARENESS_CONTEXT_SYSTEM_AWARE, OS.GetThreadDpiAwarenessContext());
+	}
+
+	@Test
+	public void setRescaleAtRuntime_toggling() {
+		display.setRescalingAtRuntime(false);
+		assertFalse(display.isRescalingAtRuntime());
+		assertEquals(OS.DPI_AWARENESS_CONTEXT_SYSTEM_AWARE, OS.GetThreadDpiAwarenessContext());
+		display.setRescalingAtRuntime(true);
+		assertTrue(display.isRescalingAtRuntime());
+		assertEquals(OS.DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2, OS.GetThreadDpiAwarenessContext());
+		display.setRescalingAtRuntime(false);
+		assertFalse(display.isRescalingAtRuntime());
+		assertEquals(OS.DPI_AWARENESS_CONTEXT_SYSTEM_AWARE, OS.GetThreadDpiAwarenessContext());
+	}
+
+}
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
@@ -6856,9 +6856,11 @@ public boolean isRescalingAtRuntime() {
  * method on other operating system will have no effect.
  *
  * @param activate whether rescaling shall be activated or deactivated
+ * @return whether activating or deactivating the rescaling was successful
  * @since 3.127
  */
-public void setRescalingAtRuntime(boolean activate) {
+public boolean setRescalingAtRuntime(boolean activate) {
 	// not implemented for Cocoa
+	return false;
 }
 }
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
@@ -6290,10 +6290,12 @@ public boolean isRescalingAtRuntime() {
  * method on other operating system will have no effect.
  *
  * @param activate whether rescaling shall be activated or deactivated
+ * @return whether activating or deactivating the rescaling was successful
  * @since 3.127
  */
-public void setRescalingAtRuntime(boolean activate) {
+public boolean setRescalingAtRuntime(boolean activate) {
 	// not implemented for GTK
+	return false;
 }
 
 }
--- a/bundles/org.eclipse.swt/Eclipse SWT PI/win32/org/eclipse/swt/internal/win32/OS.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT PI/win32/org/eclipse/swt/internal/win32/OS.java
@@ -369,8 +369,8 @@ public class OS extends C {
 	public static final short DMDUP_SIMPLEX = 1;
 	public static final short DMDUP_VERTICAL = 2;
 	public static final short DMDUP_HORIZONTAL = 3;
-	public static final int DPI_AWARENESS_CONTEXT_UNAWARE = 16;
-	public static final int DPI_AWARENESS_CONTEXT_SYSTEM_AWARE = 17;
+	public static final int DPI_AWARENESS_CONTEXT_UNAWARE = 24592;
+	public static final int DPI_AWARENESS_CONTEXT_SYSTEM_AWARE = 24593;
 	public static final int DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE = 18;
 	public static final int DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2 = 34;
 	public static final int DSTINVERT = 0x550009;
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
@@ -5272,38 +5272,42 @@ public boolean isRescalingAtRuntime() {
  * method on other operating system will have no effect.
  *
  * @param activate whether rescaling shall be activated or deactivated
+ * @return whether activating or deactivating the rescaling was successful
  * @since 3.127
  */
-public void setRescalingAtRuntime(boolean activate) {
-	rescalingAtRuntime = activate;
-	// dispose a existing font registry for the default display
-	SWTFontProvider.disposeFontRegistry(this);
-	setProperDPIAwareness();
+public boolean setRescalingAtRuntime(boolean activate) {
+	int desiredApiAwareness = activate ? OS.DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2 : OS.DPI_AWARENESS_CONTEXT_SYSTEM_AWARE;
+	if (setDPIAwareness(desiredApiAwareness)) {
+		rescalingAtRuntime = activate;
+		// dispose a existing font registry for the default display
+		SWTFontProvider.disposeFontRegistry(this);
+		return true;
+	}
+	return false;
 }
 
-private void setProperDPIAwareness() {
-	long desiredDpiAwareness = OS.DPI_AWARENESS_CONTEXT_SYSTEM_AWARE;
+private boolean setDPIAwareness(int desiredDpiAwareness) {
 	if (OS.WIN32_BUILD < OS.WIN32_BUILD_WIN10_1607) {
 		System.err.println("***WARNING: the OS version does not support setting DPI awareness.");
-		return;
-	}
-	if (rescalingAtRuntime) {
-		desiredDpiAwareness = OS.DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2;
-		// Auto scaling on runtime requires DPI awareness mode "Per Monitor V2"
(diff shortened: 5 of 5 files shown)


bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
```diff
@@ -6290,6 +6290,7 @@ public boolean isRescalingAtRuntime() {
  * method on other operating system will have no effect.
  *
  * @param activate whether rescaling shall be activated or deactivated
+ * @return whether activating or deactivating the rescaling was successful
  * @since 3.127
  */
 public void setRescalingAtRuntime(boolean activate) {
```
fedejeanne, line 6296:
```suggestion
public boolean setRescalingAtRuntime(boolean activate) {
```

bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/DisplayWin32Test.java
```diff
@@ -0,0 +1,55 @@
...
+		display.setRescalingAtRuntime(true);
+		assertTrue(display.isRescalingAtRuntime());
+		assertEquals(OS.DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2, OS.GetThreadDpiAwarenessContext());
+	}
+
+	@Test
+	public void setRescaleAtRuntime_deactivate() {
+		display.setRescalingAtRuntime(false);
+		assertFalse(display.isRescalingAtRuntime());
+		assertEquals(OS.DPI_AWARENESS_CONTEXT_SYSTEM_AWARE, OS.GetThreadDpiAwarenessContext());
```
fedejeanne, line 39: (see comment below about "default value")

bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/DisplayWin32Test.java
```diff
@@ -0,0 +1,55 @@
...
+		display = new Display();
+	}
+
+	@AfterEach
+	public void destroyDisplay() {
+		display.dispose();
+	}
+
+	@Test
+	public void setRescaleAtRuntime_activate() {
```
fedejeanne, line 29: NIT: I know it's not a rule but I'd rather have the tests named `test...blablabla`, it helps me to see very easily which methods are tests when using the call hierarchy.

fedejeanne (changes requested):
I think the current state would brake the API (the methods in Mac/Linux have a different return value than the one for Windows).

The rest is not critical.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
HeikoKlare: Thank you! Fixed with 1a5d396.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
```diff
@@ -6856,6 +6856,7 @@ public boolean isRescalingAtRuntime() {
  * method on other operating system will have no effect.
  *
  * @param activate whether rescaling shall be activated or deactivated
+ * @return whether activating or deactivating the rescaling was successful
  * @since 3.127
  */
 public void setRescalingAtRuntime(boolean activate) {
```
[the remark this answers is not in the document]
HeikoKlare: Thank you! Fixed with https://github.com/eclipse-platform/eclipse.platform.swt/commit/1a5d39686f4741e5f34780a2229e36824e53fba0.

bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/DisplayWin32Test.java
```diff
@@ -0,0 +1,55 @@
...
+		display.setRescalingAtRuntime(false);
+		assertFalse(display.isRescalingAtRuntime());
+		assertEquals(OS.DPI_AWARENESS_CONTEXT_SYSTEM_AWARE, OS.GetThreadDpiAwarenessContext());
+	}
+
+	@Test
+	public void setRescaleAtRuntime_toggling() {
+		display.setRescalingAtRuntime(false);
+		assertFalse(display.isRescalingAtRuntime());
+		assertEquals(OS.DPI_AWARENESS_CONTEXT_SYSTEM_AWARE, OS.GetThreadDpiAwarenessContext());
```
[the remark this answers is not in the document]
HeikoKlare: There is no default value. With rescaling at runtime deactivated, the DPI awareness is set to "system" (as that's the mode for which the existing HiDPI support was developed). For that reason, the OS class currently already makes the process DPI aware at startup.
Currently, it would be possible to start the application with a custom DPI awareness (which already has prefence over the old "DPI aware" property) and then calling `setRescaleAtRuntime(false)` for the main display would change that custom DPI awareness to "system", but I do not see why anyone should want the application to run in an unsupported DPI awareness mode at all.

bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/DisplayWin32Test.java
HeikoKlare: Can you give me a use case in which that would be helpful, i.e., why do I want to see test methods in the call hierarchy? I am always annoyed if classes have test methods starting with `test` (which is usually because these classes were originally written for JUnit 3) as it just makes the names longer and less readable without providing any benefit.

fedejeanne: Sure thing:

image

In this case one can spot the last test (in green) easily and the other 3 (in red) not so easily

bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/DisplayWin32Test.java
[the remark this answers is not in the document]
fedejeanne: I see you point but this test is testing 2 things right now:
1. Changing the mode
2. What's the default mode

And since there is no default mode (just a de-facto default that is given by the JVM / OS) then it's not correct to test for it and enforce it in a test. There's nothing to win and something to loose: the test will unnecessarily break if the de-facto default changes in the future.

If testing that the (de-facto) default is "System" is important then it should be done in another test method.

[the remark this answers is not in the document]
HeikoKlare: Please read my previous comment. The test does not rely on a default (of the JVM/OS/whatever), it explicitly sets the values to be tested from the beginning.

bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/DisplayWin32Test.java
HeikoKlare: I think that we have an XY problem here: I see that I can spot tests that way more easily. But I do not see _why_ I should ever want to do that. I never had a need for that.

fedejeanne: Not sure what _XY problem_ is but identifying the pieces of test code more easily helps me focus on the pieces of code that are using a method and the ones that are testing it. 

I personally (often) discover the meaning of methods by looking at _how_ it is used (production code) and -in an ideal world- discover what their contracts are by looking at the tests (provided that they are unit tests and that they have been created by following the TDD paradigm). 

Long story short, it helps me differentiate types of usages and prioritize where to focus my attention better.

bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/DisplayWin32Test.java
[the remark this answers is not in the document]
fedejeanne: Ah, I see what you mean: by calling `setRescaleAtRuntime(false)` you're already setting the DPI awareness to `OS.DPI_AWARENESS_CONTEXT_SYSTEM_AWARE` ("system").

All good, thank you for the clarification.

bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/DisplayWin32Test.java
HeikoKlare: I am sorry, but I still don't get it. A generic "xyz helps be prioritize" will not allow me to understand use cases, in which this could be important. As mentioned: I have never felt a need for this information. In addition, I have never read about anyone propose this `test` prefix to be used a convention for whatever benefits and I am not aware of any convention for the Eclipse project but I only found it as a leftover of JUnit 3 times (and that's why there are even automated ways of getting rid of them, e.g., https://docs.openrewrite.org/recipes/java/testing/cleanup/removetestprefix).

Please provide some proper sources justifying your proposal. Otherwise, sorry to say, I will not adapt this PR to conform to a single person's personal preference.

fedejeanne: You never read about anyone proposing to preserve the "test" because there was no need for it: they were already there. If anything, one should read about the need to remove them (_i.e._ to change the status quo). If you ever start that discussion, please ping me ;-) 

You're reading about it now, from me.

"Leftovers" don't mean "bad", it's just legacy and legacy is not always bad. In fact, I have a use for it, even if you don't.

There are now conventions for it because, again, there was no explicit need to change/preserve the status quo.

An OpenRewrite recipe is neither a convention nor an enforcer, it's just a tool. One needs to choose when to use that tool. Conventions are specified by teams, not tools.

Again, this is not a blocker and I don't want to keep you from merging this PR. I'm just asking you to please do not get rid of something that is **useful** (to me). 

I'm not sure where do you draw the line between a "personal preference" and a "well-founded ask", but I hope you see my ask as the latter. After all, I'm not simply writing "_please don't remove them because I like them_", I'm writing "_please don't remove them because I **need** them, I find them useful_".

HeikoKlare: I don't think this discussion leads to anything. As long as you are not willing to explain the reasons for your request, I am totally lost in finding a way to properly respond to it. Still I cannot leave some of the recent statements uncommented.

> You never read about anyone proposing to preserve the "test" because there was no need for it: they were already there. If anything, one should read about the need to remove them (_i.e._ to change the status quo). If you ever start that discussion, please ping me ;-)

The introduction of JUnit 4 made the `test` prefix obsolete, which was required in JUnit 3 times. JUnit 4 was released in 2006. Even though many projects migrated later than that, the general statement that the prefixes "were already there" is not true for almost 20 years anymore.

> You're reading about it now, from me.
> 
> "Leftovers" don't mean "bad", it's just legacy and legacy is not always bad. In fact, I have a use for it, even if you don't.
> 
> There are no conventions for it because, again, there was no explicit need to change/preserve the status quo.
> 
> An OpenRewrite recipe is neither a convention nor an enforcer, it's just a tool. One needs to choose when to use that tool. Conventions are specified by teams, not tools.

I don't see anything I wrote conflicting with this statement. Indeed, conventions are specified by people that find them useful (in a specific context). And I have already stated that I am not aware of any convention that we are following for this project. In addition, an OpenRewrite recipe is not a self purpose. Someone has written it for the sake of getting rid of test prefixes in an actual context (nor in _every_ context, of course).

> Again, this is not a blocker and I don't want to keep you from merging this PR. I'm just asking you to please do not get rid of something that is **useful** (to me).
>
> I'm not sure where do you draw the line between a "personal preference" and a "well-founded ask", but I hope you see my ask as the latter. After all, I'm not simply writing "_please don't remove them because I like them_", I'm writing "_please don't remove them because I **need** them, I find them useful_".

Maybe you are mistaken about what this PR does: I am not changing names of existing test method. I just add new test methods without that prefix, as I do not see any value in them. I never asked another PR to remove the `test` prefix in their changed/added test method. _You_ ask me to change this, without giving me the chance to understand _why_. You still did not respond to my question of explaining me (by concrete example) why your proposal is useful.

To conclude this:
- If you want to have a convention (or at least a discussion about a need for it) for this project, please start an "official" discussion in the GitHub organization, so that everyone has a chance for sharing an opinion. If others also feel the need for such a prefix, we might agree on a convention that I will gladly follow.
- If you want to convince me, explain me why your proposal makes sense (at a concrete scenario and not only because "you need it").
- Otherwise, in case there are no further objections, I propose to close this discussion and proceed with the essence of this PR and target its merge.

fedejeanne: 👍 

Proceed.

Change the name of the class too then. It's new and JUnit doesn't need the suffix `Test` anymore either.

HeikoKlare: > Change the name of the class too then. It's new and JUnit doesn't need the suffix `Test` anymore either.

I would really prefer to stay professional and have a serious discussion instead of making sulky statements or making fun of other people's opinions (not sure which of them was the intention here).

fedejeanne: > > Change the name of the class too then. It's new and JUnit doesn't need the suffix `Test` anymore either.
> 
> I would really prefer to stay professional and have a serious discussion instead of making sulky statements or making fun of other people's opinions (not sure which of them was the intention here).

I was neither trying to make fun of your opinion nor trying to be sulky.

But I see I didn't exactly hit the nail in politeness either and that my comment can be very easily misread, so please allow to me rephrase:
_it's worth considering changing the name of the class too since it falls into the same category as the name of the methods: no convention that dictates `Test` is necessary as a suffix and no need for it because of JUnit_.

Furthermore, I found out that there is one trick in Eclipse that I can use to cover my use case (distinguishing Test sources in the _hierarchy call_ view): **filters**.

image

image

With that, I can easily get rid of all calls from within _test code_, so I'm all set.

image

On my account this PR is ready to be merged (it has been since the very beginning actually, hence the "NIT" in my 1st comment) and this discussion is starting to go sideways so there's no need to prolong it.

I thank you for your patience and I sincerely value your opinion.

Should me mark this conversation as _resolved"_? ✔️

HeikoKlare: Yes, we can mark this as resolved. In case we need to change anything related to tests, let's do it in a follow-up to have the productive code changes in now.

Why didn't I take the last proposal serious? `org.eclipse.swt.widgets.DisplayWin32` neither sounds like a reasonable FQN for a class nor does it indicate in any way that it is a test class. It might have been okay if the class was placed in a separate package with a test indication given by some package in the FQN, but this is not the case, so it would not make any sense.

PR #1364, opened 2024-07-25
[win32] Refactor toolbar dpi callback

akoch-yatta:
This PR mainly is targeted on refactoring the existing callback for ToolBar and ToolItem after a DPI change in the win32 implementation. 
To achieve that a method to scale an ImageList was added. Scaled variants of an image list are saved and reused similar to the other Resources.
The refactoring uses the scalable ImageLists to reuse the existing toolbar items and only resets the ImageList with a correctly scaled one. Due to the limitations of the win32 implementation all items must be removed and re-added via the win32-API to make shrinking of toolbars possible. Additionally, the commit removes the now unused callback in ToolItem.

All calls to the ImageList handle are replaced to use the added method.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/ImageList.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/ImageList.java
@@ -14,6 +14,8 @@
 package org.eclipse.swt.internal;
 
 
+import java.util.*;
+
 import org.eclipse.swt.*;
 import org.eclipse.swt.graphics.*;
 import org.eclipse.swt.internal.win32.*;
@@ -23,6 +25,10 @@ public class ImageList {
 	int style, refCount;
 	Image [] images;
 	private final int zoom;
+	private int width, height;
+	private int flags;
+
+	private HashMap<Integer, Long> zoomToHandle = new HashMap<>();
 
 public ImageList (int style, int zoom) {
 	this (style, 32, 32, zoom);
@@ -30,9 +36,12 @@ public ImageList (int style, int zoom) {
 
 public ImageList (int style, int width, int height, int zoom) {
 	this.style = style;
-	int flags = OS.ILC_MASK | OS.ILC_COLOR32;
-	if ((style & SWT.RIGHT_TO_LEFT) != 0) flags |= OS.ILC_MIRROR;
-	handle = OS.ImageList_Create (width, height, flags, 16, 16);
+	this.flags = OS.ILC_MASK | OS.ILC_COLOR32;
+	if ((style & SWT.RIGHT_TO_LEFT) != 0) this.flags |= OS.ILC_MIRROR;
+	this.height = height;
+	this.width = width;
+	handle = OS.ImageList_Create (width, height, this.flags, 16, 16);
+	zoomToHandle.put(zoom, handle);
 	images = new Image [4];
 	this.zoom = zoom;
 }
@@ -51,7 +60,7 @@ public int add (Image image) {
 		Rectangle rect = DPIUtil.scaleBounds(image.getBounds(), zoom, 100);
 		OS.ImageList_SetIconSize (handle, rect.width, rect.height);
 	}
-	set (index, image, count);
+	setForAllHandles(index, image, count);
 	if (index == images.length) {
 		Image [] newImages = new Image [images.length + 4];
 		System.arraycopy (images, 0, newImages, 0, images.length);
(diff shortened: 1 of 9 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/ImageList.java
```diff
@@ -23,6 +25,9 @@ public class ImageList {
 	int style, refCount;
 	Image [] images;
 	private final int zoom;
+	int width, height;
```
HeikoKlare, line 28: Can these values be private?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/ImageList.java
```diff
@@ -445,4 +457,22 @@ public int size () {
 	return result;
 }
 
+public static long win32_getHandle(ImageList imageList, int zoom) {
+	if(!imageList.zoomToHandle.containsKey(zoom)) {
+		int scaledWidth = DPIUtil.scaleUp(DPIUtil.scaleDown( imageList.width,imageList.zoom), zoom);
+		int scaledHeight = DPIUtil.scaleUp(DPIUtil.scaleDown( imageList.height,imageList.zoom), zoom);
```
HeikoKlare, line 463: whitespace seems to be broken here, e.g.
`DPIUtil.scaleDown( imageList.width,imageList.zoom)` -> `DPIUtil.scaleDown(imageList.width, imageList.zoom)`

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/ImageList.java
```diff
@@ -445,4 +457,22 @@ public int size () {
...
+		long handle = OS.ImageList_Create (scaledWidth, scaledHeight, OS.ILC_MASK | OS.ILC_COLOR32, 16, 16);
```
HeikoKlare, line 464: Shouldnt't this use the flags of the original list?
```suggestion
		long handle = OS.ImageList_Create (scaledWidth, scaledHeight, imageList.flags, 16, 16);
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/ImageList.java
```diff
@@ -445,4 +457,22 @@ public int size () {
 	return result;
 }
 
+public static long win32_getHandle(ImageList imageList, int zoom) {
```
HeikoKlare, line 460: Shouldn't the `getHandle()` method be removed in favor of the `win32_getHandle()` method? It should even be possible to replace the existing `getHandle()` method with one accepting a zoom parameter, as this class is internal and its interface may be changed.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/ToolBar.java
```diff
@@ -1742,75 +1745,59 @@ private static void handleDPIChange(Widget widget, int newZoom, float scalingFac
...
+	var seperatorWidth = new int[toolItems.length];
+	int itemCount = toolItems.length;
+
+	// Remove and re-add all button the let Windows resize the tool bar
+    ToolItem[] items = new ToolItem[itemCount];
+    TBBUTTON[] buttondata = new TBBUTTON[itemCount];
+    for (int i = itemCount - 1; i >= 0; i--) {
+    	ToolItem item = toolItems[i];
+    	if ((item.style & SWT.SEPARATOR) != 0 && item.getControl() != null) {
+    		// Take note of widths of separators with control, so they can be resized
```
HeikoKlare, line 1758: There seem to be some inconsistencies in the formatting of this file.

HeikoKlare (commented):
I have not tested the proposal yet, but I have taken a first look at the code. So far, I only have minor comments.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/ImageList.java
akoch-yatta: Yes, I made all added variables private

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/ImageList.java
akoch-yatta: Yes, I adapted that in the latest commit

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/ImageList.java
akoch-yatta: Good point. We store the flags in the image list now and use that for the creation of the scaled variant

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/ImageList.java
akoch-yatta: Yes, definitely. I moved the new logic into getHandle and adapted all calls accordingly

HeikoKlare (approved):
The change looks sound to me. I've also tested it (in combination with https://github.com/akoch-yatta/eclipse.platform.ui/commit/384e390ac8e9e4e4a912695880aaef864de07897 to properly see effects on the main toolbar) and found it working fine.

@fedejeanne maybe you could also have a look since parts of the change slightly affect the behavior also when using existing HiDPI support, so it would be good to have an additional review.

fedejeanne (approved):
LGTM, just 1 nit

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/ImageList.java
```diff
@@ -23,16 +25,23 @@ public class ImageList {
 	int style, refCount;
 	Image [] images;
 	private final int zoom;
+	private int width, height;
+	private int flags;
```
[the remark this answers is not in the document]
akoch-yatta: I adapted that in [ee0b8d](https://github.com/eclipse-platform/eclipse.platform.swt/commit/ee0b8d7f887dc269d53e69efa13857c4a9508031)

PR #1368, opened 2024-07-26
[win32] Forward ZoomChanged to all child widgets

akoch-yatta:
This PR adapt how the DPI change is propagated into the DPI update mechanism. In comes down to the following:

## Automatic rescaling is not active
It is exactly as in the beginning -> A ZoomChanged event is send for the affected control, wenn the zoom is changed

## Automatic rescaling is active
The update is no longer propagation via a ZoomChanged event that was used by a listener in a Shell to propagate this change into the DPIZoomChangeRegistry, but it is directly propagated into DPIZoomChangeRegistry in the WM_DPICHANGED callback.
Additionally, for each widget, that is handled in DPIZoomChangeRegistry for a dpi change a ZoomChanged event is created and sent to the widget. That enables creators of custom widget to hook into the update and refresh the state of the affected widget, e.g. refresh a widget, when it is rendering its content without a layout manager. 

Contributes to #62 and #131

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
@@ -4900,28 +4900,32 @@ LRESULT WM_DESTROY (long wParam, long lParam) {
 LRESULT WM_DPICHANGED (long wParam, long lParam) {
 	// Map DPI to Zoom and compare
 	int newNativeZoom = DPIUtil.mapDPIToZoom (OS.HIWORD (wParam));
-	int oldNativeZoom = getShell().nativeZoom;
-
-	// Throw the DPI change event if zoom value changes
-	if (newNativeZoom != oldNativeZoom) {
-		Event event = new Event();
-		event.type = SWT.ZoomChanged;
-		event.widget = this;
-		event.detail = newNativeZoom;
-		event.doit = true;
-
-		if (getDisplay().isRescalingAtRuntime()) {
+	if (getDisplay().isRescalingAtRuntime()) {
+		int oldNativeZoom = nativeZoom;
+		if (newNativeZoom != oldNativeZoom) {
 			DPIUtil.setDeviceZoom (newNativeZoom);
-		}
 
-		notifyListeners(SWT.ZoomChanged, event);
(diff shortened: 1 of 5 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
```diff
@@ -4900,28 +4900,32 @@ LRESULT WM_DESTROY (long wParam, long lParam) {
...
 
-		if (getDisplay().isRescalingAtRuntime()) {
 			RECT rect = new RECT ();
 			COM.MoveMemory(rect, lParam, RECT.sizeof);
 			this.setBoundsInPixels(rect.left, rect.top, rect.right - rect.left, rect.bottom-rect.top);
+			return LRESULT.ZERO;
+		}
+	} else {
+		int newZoom = DPIUtil.getZoomForAutoscaleProperty (newNativeZoom);
+		int oldZoom = DPIUtil.getZoomForAutoscaleProperty (nativeZoom);
```
HeikoKlare, line 4918: This will change the behavior for the existing HiDPI support (without rescaling at runtime): in default mode `integer200`, a "zoom changed" event will not be generated for every native zoom change anymore, but only when the "corrected" one (100, 200 etc.) changes. In my opinion, this is a breaking API change (since consumers will receive the event in different cases than before), but in any way I propose to have such a change performed in a separate PR to be discussed there.

HeikoKlare (commented):
Overall, the change looks good. I have only one concern regarding a modification of existing behavior.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
akoch-yatta: Actually that is restoring the initial implementation. We adapted that in [77c1da](https://github.com/eclipse-platform/eclipse.platform.swt/commit/77c1da8956b7868e95eb45a3816ad732ceffda84#diff-f0c798cff8b637e98ce90c54208151a6a98e83f0c22c4e2427598957e108889c) some weeks ago

HeikoKlare: Interesting, thank you for the pointer! Then it makes sense to take this chance and restore the existing behavior.

HeikoKlare (approved):
The change is sound and ready to be merged from my side.

@fedejeanne can you also review this please?

fedejeanne (approved):
LGTM

PR #1369, opened 2024-07-26
Replace DPIUtil.get(Native)DeviceZoom for win32

amartya4256:
Now that we have the zoom level information avaialble for the element, we can use the right zoom information instead of DPIUtil.get(Native)DeviceZoom method calls since this information might not always be right.


HeikoKlare (approved):
The changes look sound.

@akoch-yatta are the changes also okay from your side?

akoch-yatta:
> The changes look sound.
> 
> @akoch-yatta are the changes also okay from your side?

Looks good to me as well

PR #1374, opened 2024-07-29
[win32] Prevent exception on toolbar DPI change

akoch-yatta:
This commit prevents an ArrayIndexOutOfBounds exception, when the commandIds of toolbar buttons are not in line with button count, e.g. there is an id higher than the button count.

Contributes to #62 and #131

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/ToolBar.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/ToolBar.java
@@ -14,6 +14,8 @@
 package org.eclipse.swt.widgets;
 
 
+import java.util.*;
+
 import org.eclipse.swt.*;
 import org.eclipse.swt.graphics.*;
 import org.eclipse.swt.internal.*;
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/ToolBar.java
```diff
@@ -1752,37 +1754,33 @@ private static void handleDPIChange(Widget widget, int newZoom, float scalingFac
...
 	toolBar.setHotImageList(toolBar.getHotImageList());
-
 	OS.SendMessage(toolBar.handle, OS.TB_BUTTONSTRUCTSIZE, TBBUTTON.sizeof, 0);
-	for (int i = 0; i < buttondata.length; i++) {
-		TBBUTTON button = buttondata[i];
-		if (button != null) {
-			OS.SendMessage(toolBar.handle, OS.TB_ADDBUTTONS, 1, button);
-			ToolItem item = items[i];
-
+	for (ToolItemData itemData : buttondata.values()) {
```
HeikoKlare, line 1780: This will not result in the correct order, will it? There is no guarantee that the `values()` are iterated in an order conforming to a standard comparison for the according keys. If I understand correctly, the keys of the map are not even used except for ordering. So couldn't you simply use a LIFO structure (like a stack) in which you put the ToolItemData to achieve exactly what is intended here?

akoch-yatta: Actually I would read the Javadoc like is is doing that "The collection's iterator returns the values in ascending order of the corresponding keys". Nevertheless, as it is really pure LIFO, Stack sounds like the better approach. I updated the PR accordingly.

HeikoKlare: That's interesting. I didn't expect that behavior of a map, but I understand it the same way as you do. So the previous implementation was probably correct, still a stack seems to reflect the intention even better. Thank you for sharing the insights 🙂

HeikoKlare (approved):
The changes are sound are would be ready to be merged from my side.

@akoch-yatta is the PR still in draft mode by intent or by accident?

akoch-yatta:
> The changes are sound are would be ready to be merged from my side.
> 
> @akoch-yatta is the PR still in draft mode by intent or by accident?

That is intentional. I found one scenario, where some images are not correctly displayed after rescaling. I am investigating that.

PR #1376, opened 2024-07-31
Add early return to ToolBar::handleDPIChange in case it has no items

fedejeanne:
If the toolbar has no items then the method still executes and sends 2 unnecessary `OS` calls:
* `OS.SendMessage(toolBar.handle, OS.TB_BUTTONSTRUCTSIZE, TBBUTTON.sizeof, 0);`
* `OS.SendMessage(toolBar.handle, OS.TB_AUTOSIZE, 0, 0);`

## How to test
* Modify the `Snippet18` and remove all items in the toolbar
* Run the modified `Snippet18` with the vm arguments: `-Dswt.autoScale.updateOnRuntime=true`

### Expected results
* Same visual result as the code in `master`
* The OS calls should not happen.


akoch-yatta:
> Change looks sound to me. Would be good to also have your opinion, @akoch-yatta.

I just had one scenario in my mind that could be a problem, I need to try that out.

PR #1379, opened 2024-08-02
Fix typo README

Olexandr88:
Update link

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,13 +108,6 @@ Contact the project developers via the project's "dev" list.
 
 - <https://accounts.eclipse.org/mailing-list/platform-dev>
 
-Search for bugs:
-----------------
-
-Github is used to track current issues
-
-- <https://github.com/issues?user=eclipse-platform>
-
 Create a new bug:
 -----------------
 
--- a/README.md
+++ b/README.md
@@ -111,9 +111,9 @@ Contact the project developers via the project's "dev" list.
 Search for bugs:
 ----------------
 
-SWT used to track ongoing development and issues in Bugzilla .
+Github is used to track current issues
 
-- <https://bugs.eclipse.org/bugs/buglist.cgi?component=SWT&product=Platform>
+- <https://github.com/issues?user=eclipse-platform>
 
 Create a new bug:
 -----------------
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -113,7 +113,7 @@ Search for bugs:
 
 SWT used to track ongoing development and issues in Bugzilla .
 
-- <https://bugs.eclipse.org/bugs/buglist.cgi?product=Platform&component=SWT>
+- <https://bugs.eclipse.org/bugs/buglist.cgi?component=SWT&product=Platform>
```
laeubi, line 116: We are now using Github issues...

Wittmaxi, line 116: Both links lead to the same page, what does this change fix?

Olexandr88: Yes, I completely agree with you, but in my opinion the second link is more correct

laeubi (changes requested):
Both links are equivalent (and both are wrong), but none of them are "typo".

If you interested in changing the readme please

1. Change Bugzilla references to github reference
2. Include a useful description in the commit message

README.md
Phillipus: Either the link should be removed, or the title and rubric changed to mention that BugZilla is no longer used and this is the link to old (archived) bugs.

README.md
```diff
@@ -111,9 +111,9 @@ Contact the project developers via the project's "dev" list.
 Search for bugs:
 ----------------
 
-SWT used to track ongoing development and issues in Bugzilla .
+Github is used to track current issues
 
-- <https://bugs.eclipse.org/bugs/buglist.cgi?product=Platform&component=SWT>
+- <https://github.com/issues?user=eclipse-platform>
```
Phillipus, line 116: The URL for SWT should be `https://github.com/eclipse-platform/eclipse.platform.swt/issues`

But is there any point in this? This README file is on the front page of the GH with the "Issues" menu right at the top of the page.

Olexandr88: So I think it doesn't make sense. Then you can remove this section entirely, as one of the reviewers advised me to do

merks:
At some point someone will ask you to squash your commits.  Amending a commit and force pushing it is typically better.  Here we have 3 commits which no one will want to merge as is.

PR #1385, opened 2024-08-06
Replacing the usage of getBoundsInPixels with getBounds

ShahzaibIbrahim:
Initial native image must be the one that will be used by a widget, thats why using getBounds with initialNativeZoom. This PR just remove the usage of deprecated method getBoundsInPixels wherever possible. This doesn't fix anything broken.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
@@ -466,7 +466,7 @@ void copyAreaInPixels(Image image, int x, int y) {
 	if (image == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
 	if (image.type != SWT.BITMAP || image.isDisposed()) SWT.error(SWT.ERROR_INVALID_ARGUMENT);
 	/* Copy the bitmap area */
-	Rectangle rect = image.getBoundsInPixels();
+	Rectangle rect = image.getBounds(data.nativeZoom);
 	long memHdc = OS.CreateCompatibleDC(handle);
 	long hOldBitmap = OS.SelectObject(memHdc, Image.win32_getHandle(image, getZoom()));
 	OS.BitBlt(memHdc, 0, 0, rect.width, rect.height, handle, x, y, OS.SRCCOPY);
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Table.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Table.java
@@ -7293,7 +7293,7 @@ LRESULT wmNotifyToolTip (NMTTCUSTOMDRAW nmcd, long lParam) {
 					if (pinfo.iSubItem != 0) x -= gridWidth;
 					Image image = item.getImage (pinfo.iSubItem);
 					if (image != null) {
-						Rectangle rect = image.getBoundsInPixels ();
+						Rectangle rect = image.getBoundsInPixels();
 						RECT imageRect = item.getBounds (pinfo.iItem, pinfo.iSubItem, false, true, false, false, hDC);
 						Point size = imageList == null ? new Point (rect.width, rect.height) : imageList.getImageSize ();
 						int y = imageRect.top + Math.max (0, (imageRect.bottom - imageRect.top - size.y) / 2);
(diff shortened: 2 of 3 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Table.java
```diff
@@ -7293,7 +7293,7 @@ LRESULT wmNotifyToolTip (NMTTCUSTOMDRAW nmcd, long lParam) {
 					if (pinfo.iSubItem != 0) x -= gridWidth;
 					Image image = item.getImage (pinfo.iSubItem);
 					if (image != null) {
-						Rectangle rect = image.getBoundsInPixels ();
+						Rectangle rect = image.getBoundsInPixels();
```
HeikoKlare, line 7296: Is it intended that this is only a formatting change?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -466,7 +466,7 @@ void copyAreaInPixels(Image image, int x, int y) {
 	if (image == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
 	if (image.type != SWT.BITMAP || image.isDisposed()) SWT.error(SWT.ERROR_INVALID_ARGUMENT);
 	/* Copy the bitmap area */
-	Rectangle rect = image.getBoundsInPixels();
+	Rectangle rect = image.getBounds(data.nativeZoom);
```
HeikoKlare, line 469: This modifies the behavior, doesn't it?
Let's assume `getBoundsInPixels()` returned `(100, 100, 100, 100)` and let's assume that `data.nativeZoom=125`. Then `getBounds(data.nativeZoom)` will scale up from 100 to 125, i.e., it will return `(125, 125, 125, 125)`. Do I miss something?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Table.java
ShahzaibIbrahim: formatting leftover

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
akoch-yatta: Depends what the native zoom of the image is. If the native zoom of the image is 125, then it will return (100, 100, 100, 100). If the native zoom of the image is 100, then yes. Probably would be could to look out for a snippet, where this is used and play around with that.

HeikoKlare: I do not understand that. Why should it return (100, 100, 100, 100) when the native zoom of the image is 125? Within `getBounds(int)`, the "corrected" zoom is used according to `DPIUtil.getZoomForAutoscaleProperty(initialNativeZoom)`, so even if the native zoom of the image is 125, a value of 100 is used for scaling, isn't it?

akoch-yatta: So you mean the call should be getBounds(DPI.Util.getZoomForAutoscaleProperty(data.nativeZoom)) to be in the same zoom restrictions as Image uses internally?

akoch-yatta: What I meant for example is the scenario was with mode exact or quarter, where the Image was e.g. created on the primary monitor with 125%. Then getBounds(125) would be equal to getZoom() from the image, so it would return  (100, 100, 100, 100)

HeikoKlare: That's true. I had the "default" case in mind (i.e., scaling mode `integer200`) in which I still think that the behavior might change.

ShahzaibIbrahim: > That's true. I had the "default" case in mind (i.e., scaling mode `integer200`) in which I still think that the behavior might change.

You are right, the behavior changes. I tested it with the snippet 215, where image.getBoundsInPixels() returns Rectangle {0, 0, 5120, 1440} but later scaled due to native zoom (which is 200) to final value  image.getBounds(data.nativeZoom) : Rectangle {0, 0, 10240, 2880}

HeikoKlare:
@ShahzaibIbrahim can this one be closed? It seems as if all of the places touched in this PR have meanwhile been revised anyway.

ShahzaibIbrahim:
> @ShahzaibIbrahim can this one be closed? It seems as if all of the places touched in this PR have meanwhile been revised anyway.

Yes, we can close this one.

PR #1393, opened 2024-08-08
[win32] Prevent event on tree column resize

akoch-yatta:
This PR prevents the creation of a SWT.Resize event when tree columns are resized after a zoom change. Therefore, the ignoreColumnResize attribute is added to Tree to copy the behavior of Table in the same scenario. This attribute is used during the resizing because of the zoom change to disable the creation of the resize event.

This will prevent JFace TreeColumnLayouts to break when using ColumnWeighData to configure column sizes. They would be invalidated when a SWT.Resize occurs.

As the resizing of the column after a zoom change is not really a resizing but just an adjusting to the new environment, preventing a SWT.Resize event seems like the way to go.

You can test the different behaviour of the PR with Snippet027TreeLayout


HeikoKlare (approved):
Change is sound and works as expected. It can only affect behavior when rescaling at runtime is activated. It aligns with #1394 for tables.

Issue #213, opened 2022-06-20
Welcome page has missing color and styling with Edge browser in Eclipse


niraj-modi:
**Describe the bug**
This bug is to support improved Edge based Welcome page in Eclipse, original bugzilla bug report:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=578700

**To Reproduce**
When Eclipse is launched with Edge browser:
- Welcome page has missing color and styling.

**Expected behavior**
Welcome page should show color and styling similar to embedded IE.

**Screenshots**
https://bugzillaattachments.eclipsecontent.org/bugs/attachment.cgi?id=285442

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 

2. Additional OS info (e.g. OS version, Linux Desktop, etc)

3. JRE/JDK version
 
**Version since**
 4.19

**Workaround (or) Additional context**
 Use embedded IE instead of Edge.

thegreystone:
This one would be very interesting for https://github.com/openjdk/jmc. The embedded IE is old enough that most new views in Mission Control (that are using various popular Java Script visualization frameworks) do not work on Windows. Not being able to load local resources is a huge problem.

https://github.com/MicrosoftEdge/WebView2Feedback/issues/456

thegreystone:
Also, see https://bugs.openjdk.org/browse/JMC-7118.

viragpurnam:
Hi, we have a similar requirement. We have designed some views using Java Script. These views work absolutely fine on Mac and Linux platforms but not on Windows. Unfortunately, most of the users are on windows and they are not able to use these views. (These views are the main feature of the tool). ** Could you please take this bug on priority? ** This will surely help many developers. Thanks. Looking forward for a positive response.

laeubi:
>  Could you please take this bug on priority?

You can contact the Eclipse-WG and offer some funding for this particular issue getting fixed:

https://eclipseide.org/working-group/#members

> This will surely help many developers. 

... or some of the "many developers" can take a look and provide a PR to fix this.

Phillipus:
Correct me if I'm wrong, but my understanding is that calling `Browser.setText("some html")` where the html text includes references to local files (CSS and JS) is not permitted by Edge, so Welcome page doesn't display. Therefore this can't be fixed in Eclipse and SWT.

[...]

PR #1378, opened 2024-08-02
Fix Edge Browser Deadlock issue for instantiation during a WebView Callback

amartya4256:
This contribution fixes Edge browser for win32 to handle the deadlock issue during the instantiation of a new edge browser object amidst a webview callback.

The problem with webview is that it can't execute tasks simultaneously and all the requests are serialized. Hence, a new request to the webview inside its callback leads to a deadlock since, both the tasks wait for each other to execute. The solution is to schedule the Webview calls during the callback asynchronously and let the control in the callback move on.

contributes to #669 

## Scope not covered
* This implementation doesn't fix the evaluate calls during the webview callback and would be addressed in a separate issue.

## Pre-requisites
- #1395 

## Testing
- Change the method `org.eclipse.swt.browser.BrowserFactory.createWebBrowser` to: 

```java
WebBrowser createWebBrowser (int style) {
	return new Edge ();
}
```

PR #1395, opened 2024-08-08
Handle local resource references in Edge#setText()

amartya4256:
This contribution fixes Edge browser for win32 to allow serving webpages using setText method where the source code may refer to local resources.

The problem with setText in the modern browser is that webView2 navigates to a page "about:blank" (which is not the part of the intranet) [[reference](https://learn.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.winforms.webview2.navigatetostring?view=webview2-dotnet-1.0.2535.41#remarks)] and sets the html to the DOM. In case the html contains reference to local resource, the browsers throws error "can't load local resource" because of security reasons. This contribution fixes it by navigating edge to a temporary file created in the temporary folder of the OS and hence allowing it to load the local resources, and then sets the html to the DOM directly.

**Note: We only need the file to get a concrete local address instead of "about:local" for setText method. We don't perform any I/O to the file itself.**

To test the implementation, you need to set the default browser to Edge. To do it in the source code, change org.eclipse.swt.browser.BrowserFactory.createWebBrowser with the following source code: 
`return new Edge();`

**Scope not covered:** 
* This standalone implementation makes the javadoc tooltip open in an edge browser. (Would be covered in a follow up PR, part of eclipse.jdt.ui)
* The caching of intro page (Welcome page) is also not covered by this PR. (Would be covered in a follow up PR, part of eclipse.platform)

contributes to #213

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
@@ -13,8 +13,11 @@
  *******************************************************************************/
 package org.eclipse.swt.browser;
 
+import java.io.*;
 import java.net.*;
 import java.nio.charset.*;
+import java.nio.file.*;
+import java.nio.file.Path;
 import java.time.*;
 import java.util.*;
 import java.util.function.*;
@@ -43,6 +46,13 @@ class Edge extends WebBrowser {
 	static final String DATA_DIR_PROP = "org.eclipse.swt.browser.EdgeDataDir";
 	static final String LANGUAGE_PROP = "org.eclipse.swt.browser.EdgeLanguage";
 	static final String VERSIONT_PROP = "org.eclipse.swt.browser.EdgeVersion";
+	/**
+	 * The URI_FOR_CUSTOM_TEXT_PAGE is the path to a temporary html file which is used
+	 * by Edge browser to navigate to for setting html content in the
+	 * DOM of the browser to enable it to load local resources.
+	 */
+	private static final URI URI_FOR_CUSTOM_TEXT_PAGE = setupAndGetLocationForCustomTextPage();
+	private static final String ABOUT_BLANK = "about:blank";
 
 	private record WebViewEnvironment(ICoreWebView2Environment environment, ArrayList<Edge> instances) {
 		public WebViewEnvironment(ICoreWebView2Environment environment) {
@@ -65,6 +75,8 @@ class Edge extends WebBrowser {
 	boolean inNewWindow;
 	HashMap<Long, LocationEvent> navigations = new HashMap<>();
 	private boolean ignoreFocus;
+
+	private String lastCustomText;
 
 	static {
 		NativeClearSessions = () -> {
@@ -172,6 +184,18 @@ class Edge extends WebBrowser {
 			CookieResult = hr >= COM.S_OK;
 		};
 	}
+
+private static URI setupAndGetLocationForCustomTextPage() {
+	URI absolutePath;
+	try {
+		Path tempFile = Files.createTempFile("base", ".html");
+		absolutePath = tempFile.toUri();
+		tempFile.toFile().deleteOnExit();
+	} catch (IOException e) {
+		absolutePath = URI.create(ABOUT_BLANK);
+	}
+	return absolutePath;
+}
 
 static String wstrToString(long psz, boolean free) {
 	if (psz == 0) return "";
@@ -447,6 +471,8 @@ public void create(Composite parent, int style) {
 		handler.Release();
 	}
 
+	addProgressListener(ProgressListener.completedAdapter(__ -> writeToDefaultPathDOM()));
+
 	IUnknown hostDisp = newHostObject(this::handleCallJava);
 	long[] hostObj = { COM.VT_DISPATCH, hostDisp.getAddress(), 0 }; // VARIANT
 	webView.AddHostObjectToScript("swt\0".toCharArray(), hostObj);
@@ -559,7 +585,11 @@ public String getText() {
 public String getUrl() {
 	long ppsz[] = new long[1];
 	webView.get_Source(ppsz);
-	return wstrToString(ppsz[0], true);
+	return getExposedUrl(wstrToString(ppsz[0], true));
+}
+
+private String getExposedUrl(String url) {
+	return isLocationForCustomText(url) ? ABOUT_BLANK : url;
 }
 
 int handleCloseRequested(long pView, long pArgs) {
@@ -625,7 +655,7 @@ int handleNavigationStarting(long pView, long pArgs, boolean top) {
 	long[] ppszUrl = new long[1];
 	int hr = args.get_Uri(ppszUrl);
 	if (hr != COM.S_OK) return hr;
-	String url = wstrToString(ppszUrl[0], true);
+	String url = getExposedUrl(wstrToString(ppszUrl[0], true));
 	long[] pNavId = new long[1];
 	args.get_NavigationId(pNavId);
 	LocationEvent event = new LocationEvent(browser);
@@ -666,7 +696,7 @@ int handleSourceChanged(long pView, long pArgs) {
 	long[] ppsz = new long[1];
 	int hr = webView.get_Source(ppsz);
 	if (hr != COM.S_OK) return hr;
-	String url = wstrToString(ppsz[0], true);
+	String url = getExposedUrl(wstrToString(ppsz[0], true));
 	browser.getDisplay().asyncExec(() -> {
 		if (browser.isDisposed()) return;
 		LocationEvent event = new LocationEvent(browser);
@@ -954,15 +984,30 @@ public void stop() {
 	webView.Stop();
 }
 
-@Override
-public boolean setText(String html, boolean trusted) {
-	char[] data = new char[html.length() + 1];
-	html.getChars(0, html.length(), data, 0);
-	return webView.NavigateToString(data) == COM.S_OK;
+private boolean isLocationForCustomText(String location) {
+		try {
+			return URI_FOR_CUSTOM_TEXT_PAGE.equals(new URI(location));
+		} catch (URISyntaxException e) {
+			return false;
+		}
+}
+
+private void writeToDefaultPathDOM() {
+	if(lastCustomText != null && getUrl().equals(ABOUT_BLANK)) {
+		boolean test = jsEnabled;
+		jsEnabled = true;
+		execute("document.open(); document.write(`" + lastCustomText + "`); document.close();");
+		jsEnabled = test;
+		this.lastCustomText = null;
+	}
 }
 
 @Override
-public boolean setUrl(String url, String postData, String[] headers) {
+public boolean setText(String html, boolean trusted) {
+	return setWebpageData(URI_FOR_CUSTOM_TEXT_PAGE.toASCIIString(), null, null, html);
+}
+
+private boolean setWebpageData(String url, String postData, String[] headers, String html) {
 	// Feature in WebView2. Partial URLs like "www.example.com" are not accepted.
 	// Prepend the protocol if it's missing.
 	if (!url.matches("[a-z][a-z0-9+.-]*:.*")) {
@@ -970,6 +1015,9 @@ public boolean setUrl(String url, String postData, String[] headers) {
 	}
 	int hr;
 	char[] pszUrl = stringToWstr(url);
+	if(isLocationForCustomText(url)) {
+		this.lastCustomText = html;
+	}
 	if (postData != null || headers != null) {
 		if (environment2 == null || webView_2 == null) {
 			SWT.error(SWT.ERROR_NOT_IMPLEMENTED, null, " [WebView2 version 88+ is required to set postData and headers]");
@@ -1002,6 +1050,11 @@ public boolean setUrl(String url, String postData, String[] headers) {
 		hr = webView.Navigate(pszUrl);
 	}
 	return hr == COM.S_OK;
+}
+
+@Override
+public boolean setUrl(String url, String postData, String[] headers) {
+	return setWebpageData(url, postData, headers, null);
 }
 
 }
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
@@ -527,7 +527,6 @@ public void test_isLocationForCustomText_setUrlAfterDisposedThrowsSwtException()
 
 @Test
 public void test_isLocationForCustomText_isSetUrlNotCustomTextUrlAfterSetText() {
-	assumeFalse("behavior is not (yet) supported by Edge browser", isEdge);	// Edge doesn't fire a changed event if the URL is the same as the previous location.
 	String url = getValidUrl();
 	AtomicBoolean locationChanged = new AtomicBoolean(false);
 	browser.addLocationListener(changedAdapter(event -> {
@@ -544,7 +543,6 @@ public void test_isLocationForCustomText_isSetUrlNotCustomTextUrlAfterSetText() 
 
 @Test
 public void test_isLocationForCustomText_isFirstSetTextURLStillCustomTextUrlAfterSetUrl() {
-	assumeFalse("behavior is not (yet) supported by Edge browser", isEdge);	// Edge doesn't fire a changed event if the URL is the same as the previous location.
 	AtomicBoolean locationChanged = new AtomicBoolean(false);
 	browser.addLocationListener(changedAdapter(event -> locationChanged.set(true)));
 	String url = getValidUrl();
@@ -582,7 +580,6 @@ public void test_isLocationForCustomText_isSetUrlNotCustomTextUrl() {
 
 @Test
 public void test_isLocationForCustomText() {
-	assumeFalse("behavior is not (yet) supported by Edge browser", isEdge);	// Edge doesn't fire a changed event if the URL is the same as the previous location.
 	AtomicBoolean locationChanged = new AtomicBoolean(false);
 	browser.addLocationListener(changedAdapter(e -> locationChanged.set(true)));
 	browser.setText("Hello world");
@@ -601,7 +598,6 @@ public void test_LocationListener_changing() {
 }
 @Test
 public void test_LocationListener_changed() {
-	assumeFalse("behavior is not (yet) supported by Edge browser", isEdge);
 
 	AtomicBoolean changedFired = new AtomicBoolean(false);
 	browser.addLocationListener(changedAdapter(e ->	changedFired.set(true)));
@@ -612,8 +608,6 @@ public void test_LocationListener_changed() {
 }
 @Test
 public void test_LocationListener_changingAndOnlyThenChanged() {
-	assumeFalse("behavior is not (yet) supported by Edge browser", isEdge);
-
 	// Test proper order of events.
 	// Check that 'changed' is only fired after 'changing' has fired at least once.
 	AtomicBoolean changingFired = new AtomicBoolean(false);
@@ -657,8 +651,6 @@ public void test_LocationListener_changingAndOnlyThenChanged() {
 
 @Test
 public void test_LocationListener_then_ProgressListener() {
-	assumeFalse("behavior is not (yet) supported by Edge browser", isEdge);
-
 	AtomicBoolean locationChanged = new AtomicBoolean(false);
 	AtomicBoolean progressChanged = new AtomicBoolean(false);
 	AtomicBoolean progressChangedAfterLocationChanged = new AtomicBoolean(false);
@@ -752,8 +744,6 @@ public void test_LocationListener_ProgressListener_cancledLoad () {
 @Test
 /** Ensue that only one changed and one completed event are fired for url changes */
 public void test_LocationListener_ProgressListener_noExtraEvents() {
-	assumeFalse("behavior is not (yet) supported by Edge browser", isEdge);
-
 	AtomicInteger changedCount = new AtomicInteger(0);
 	AtomicInteger completedCount = new AtomicInteger(0);
 
@@ -860,7 +850,6 @@ public void test_OpenWindowListener_open_ChildPopup() {
 /** Validate event order : Child's visibility should come before progress completed event */
 @Test
 public void test_OpenWindow_Progress_Listener_ValidateEventOrder() {
-	assumeFalse("behavior is not (yet) supported by Edge browser", isEdge);
 
 	AtomicBoolean windowOpenFired = new AtomicBoolean(false);
 	AtomicBoolean childCompleted = new AtomicBoolean(false);
@@ -1105,7 +1094,6 @@ public void test_TitleListener_event() {
 	assertTrue(errMsg, passed);
 }
 
-
 @Test
 public void test_setText() {
 	String expectedTitle = "Website Title";
@@ -1199,16 +1187,16 @@ private void validateTitleChanged(String expectedTitle, Runnable browserSetFunc)
 	browserSetFunc.run();
 	shell.open();
 
-	boolean hasFinished = waitForPassCondition(() -> actualTitle.get().length() != 0
-			&& !actualTitle.get().contains("about:blank")); // Windows sometimes does 2 loads, one "about:blank", and one actual load.
-	boolean passed = hasFinished && actualTitle.get().equals(expectedTitle);
+	boolean passed = waitForPassCondition(() -> actualTitle.get().equals(expectedTitle));
 	String errMsg = "";
-	if (!hasFinished)
-		errMsg = "Test timed out. TitleListener not fired";
-	else if (!actualTitle.get().equals(expectedTitle)) {
-		errMsg = "\nExpected title and actual title do not match."
-				+ "\nExpected: " + expectedTitle
-				+ "\nActual: " + actualTitle;
+	if (!passed) {
+		if (actualTitle.get().length() == 0) {
+			errMsg = "Test timed out. TitleListener not fired";
+		} else {
+			errMsg = "\nExpected title and actual title do not match."
+					+ "\nExpected: " + expectedTitle
+					+ "\nActual: " + actualTitle;
+		}
 	}
 	assertTrue(errMsg + testLog.toString(), passed);
 }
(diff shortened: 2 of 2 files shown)


Phillipus:
Isn't this the same as https://github.com/eclipse-platform/eclipse.platform.swt/pull/1378?

amartya4256:
> Isn't this the same as #1378?

The PR got quite complicated so I split it into 2 PRs. This is specifically for fixing the set text behaviour and #1378 is for fixing the deadlock issue.

fedejeanne (commented):
I didn't have time to **test** this again but the overall state of the code LGTM. All my comments from  #1378 have been addressed and I see them addressed in this PR too.

I can't approve it though since I haven't had time to properly test this new PR and I can't know for sure if splitting the commits of #1378 had some side-effects.

HeikoKlare (commented):
In general, it's great to see that with this fix Edge's `setText()` works properly, so even Javadoc, Java element information, Welcome page etc. can be displayed.

My central concern is about the introduction of a public `BASE_URI` as a replacement for the generally used `about:blank`. Conceptually, I would like to have this hidden as an implementation detail of the browser. I.e., there should not be any magic URL used when setting custom contents via `setText()` against which consumers compare (like they now do for `about:blank`. From what I have seen so far, the information demand of consumers is to decide whether a page with some custom text is shown. To this end, the browser could provide an according interface providing the required information while hiding how this is implemented for a specific browser. For WebKit and IE that might be done with the `about:blank` page, Edge might use a custom base URL, and that implementation detail may even be changed as desired or required without affecting consumers relying on that interface.

As a concrete proposal, some additional method for `Browser` like an `isLocationForCustomText(String location)` (may be called differently, accept a URL instead, ...) could be provided. This method could either match the provided location for some known strings (`about:blank`, BASE_URI, starting with `data:` or the like) or delegate the decision to the used `WebBrowser` so that only the specific location for that browser implementation is considered.

Another benefit of that solution is that we could add such a method in a separate PR, adapt the known consumers to use that method, and then have this PR merged for adapting the Edge browser. That would avoid that merging this PR in the current state, Java element links in JDT would be broken when using Edge as the default browser until JDT is adapted to also match the new magic BASE_URI: https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/caf5a4b29592e6d3911a2491aa6b811b52d66f74/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/viewsupport/JavaElementLinks.java#L599-L611

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
```diff
@@ -1045,6 +1033,17 @@ public void test_TitleListener_event() {
...
+public void test_isLocationForCustomText() {
+	String expectedTitle = "Website Title";
+	Runnable browserSetFunc = () -> {
+		String html = "<html><title>Website Title</title><body>Html page with custom title</body></html>";
+		boolean opSuccess = browser.setText(html);
+		assertTrue("Expecting setText() to return true", opSuccess);
+	};
+	validateTitleChanged(expectedTitle, browserSetFunc);
+	assertTrue("Browser didn't navigate to the implicit custom text location", browser.isLocationForCustomText(browser.getUrl()));
+}
```
HeikoKlare, line 1046: Can't the test method be already implemented like this in the original contribution in #1409 rather than changing it in this PR directly after introducing it? Or is there a reason why it must be implemented differently before the changes of this PR?

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -907,21 +935,34 @@ public void stop() {
...
-public boolean setText(String html, boolean trusted) {
-	char[] data = new char[html.length() + 1];
-	html.getChars(0, html.length(), data, 0);
-	return webView.NavigateToString(data) == COM.S_OK;
+public URI getBaseUri() {
+	return BASE_URI;
+}
+
+private void writeToDefaultPathDOM() {
+	if(html != null && URI.create(getUrl()).equals(BASE_URI)) {
```
HeikoKlare, line 943: I guess that this should use `isLocationForCustomText()`.

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -31,6 +34,18 @@ class Edge extends WebBrowser {
...
+	static {
+		URI absolutePath;
+		try {
+			Path tempFile = Files.createTempFile("base", ".html");
+			absolutePath = tempFile.toUri();
+			tempFile.toFile().deleteOnExit();
+		} catch (IOException e) {
+			absolutePath = URI.create("about:blank");
+		}
+		BASE_URI = absolutePath;
```
HeikoKlare, line 46: Are there reasons for writing the value to a temporary local variable instead of directly assigning the value to field?

HeikoKlare (commented):
In order to validate this solution, I've read the according Microsoft documentation for these use cases of Edge browser: https://learn.microsoft.com/en-us/microsoft-edge/webview2/concepts/working-with-local-content

With that information, this still seems to be a proper solution for the problem. I only have some minor comments left.

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
amartya4256: Yes, changed it.

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
amartya4256: image
Asks to remove final

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
amartya4256: Done

HeikoKlare: Not this change seems to be gone completely. Isn't is relevant anymore?

amartya4256: I have moved it all to #1409 since it just adds an extra check of cutom tect url on setText which is relevant to the previous PR.

HeikoKlare: I see. Then please see my comment on #1409:
https://github.com/eclipse-platform/eclipse.platform.swt/pull/1409#discussion_r1740776474

amartya4256: Addressed

sratz:
AFAIK all the other browsers consistently use `about:blank` as the location for text written via `setText()`.

That means:
- `Browser.getext()` returns 'about:blank'
- `LocationEvent#location` contains 'about:blank'
And there are many many code pieces where 'about:blank' is checked.

With this approach this would change, meaning that all consumers would have to adapt and do something like

```diff
-if ("about:blank".equals(event.location)) {
+if (browser.isLocationForCustomText(event.location)) {
```

I was wondering: To maximize compatibility, could we instead hide this fact and fake `getUrl()` to return 'about:blank' and handle this case internally?

HeikoKlare:
> I was wondering: To maximize compatibility, could we instead hide this fact and fake `getUrl()` to return 'about:blank' and handle this case internally?

Interesting idea. I would expect that we have already considered that, but I am not sure anymore. This would of course also involve the location listeners to be provided with the faked location. @amartya4256, do you have any insights on this?

> With this approach this would change, meaning that all consumers would have to adapt and do something like
> 
> ```diff
> -if ("about:blank".equals(event.location)) {
> +if (browser.isLocationForCustomText(event.location)) {
> ```

That's true. For some essential repos, there are already follow-up PRs to do this:
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/2211
- https://github.com/eclipse-platform/eclipse.platform/pull/1521
- https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1597

In my opinion, independent of whether the proposal of faking the URL behavior works or not, this adaptation should be done. Currently, all consumers rely on the usage of magic values, which will forever be unchangeable. Hiding this behind an API improves the design and makes the behavior exchangeable.

amartya4256:
> > I was wondering: To maximize compatibility, could we instead hide this fact and fake `getUrl()` to return 'about:blank' and handle this case internally?
> 
> Interesting idea. I would expect that we have already considered that, but I am not sure anymore. This would of course also involve the location listeners to be provided with the faked location. @amartya4256, do you have any insights on this?
> 
> > With this approach this would change, meaning that all consumers would have to adapt and do something like
> > ```diff
> > -if ("about:blank".equals(event.location)) {
> > +if (browser.isLocationForCustomText(event.location)) {
> > ```
> 
> That's true. For some essential repos, there are already follow-up PRs to do this:
> 
> * [Adapted the client to use new Browser API eclipse.platform.ui#2211](https://github.com/eclipse-platform/eclipse.platform.ui/pull/2211)
> * [Adapted the client to use new Browser API eclipse.platform#1521](https://github.com/eclipse-platform/eclipse.platform/pull/1521)
> * [Adapted the client to use new Browser API eclipse-jdt/eclipse.jdt.ui#1597](https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1597)
> 
> In my opinion, independent of whether the proposal of faking the URL behavior works or not, this adaptation should be done. Currently, all consumers rely on the usage of magic values, which will forever be unchangeable. Hiding this behind an API improves the design and makes the behavior exchangeable.

I agree with @HeikoKlare. The decision to provide an API for this was taken upon the fact that it would abstract away the behaviour of browser, hiding the URLs that clients have to hard code at the moment which in my opinion is a bad practice. It also opens up possibilities for integrating browsers in the future which mgiht have a different behaviour. We can't expect every browser to behave the same, which we saw already between Edge and IE. Also, the way Webkit is implemented, it seems to force itself to comply with how IE behaves (Async to Sync, Open Window events, ABOUT_BLANK url, etc) and that should not be the case. We should have the API for such things where the web browsers can take decisions on their own or execute in their own fashion and provide the appropriate result accordingly.

As mentioned, we already have the PRs to adapt the clients to use the API and new PRs would be following up for WebKit's internal usage of this API as well. As an idea, we can also do the same refactoring for IE.

HeikoKlare:
> Maybe one can combine the approach having a constant for `about:blank` and adding a hint to the new method where it is relevant, let browsers consistently report this as the location and have a a (default) implemented API method `Browser#isLocationForCustomText(event.location)`... that way implementers can move to the new methods more smoothly.

I agree. That's why I asked whether Sebastian's proposal is feasible, so that with that approach we might achieve an implementation of Edge browser that works with existing client implementations for now, but still provide a path to improve the design, i.e., migrate from comparison against magic value `about:blank` to usage of `Brower#isLocationForCustomText()`. The only point I do not understand is the constant for `about:blank`. You would need to migrate to usage of that constant as well, but in that case you could (and should) directly migrate to the new method. Am I missing something here?

amartya4256:
> ntation of Edge browser that works with existing client implementations for now, but still provide a path to improve the design, i.e., migrate from comparison against magic value `about:blank` to usage of `Brower#isLocationForCustomText()`. The only point I do not understand is the constant for `about:b`

It is feasible, but only getUrl would not be enough. All the callbacks in Edge where the event contains a location needs to be validated and mutated if it is a custom text url

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -31,6 +34,18 @@ class Edge extends WebBrowser {
...
+	static {
+		URI absolutePath;
+		try {
+			Path tempFile = Files.createTempFile("base", ".html");
+			absolutePath = tempFile.toUri();
+			tempFile.toFile().deleteOnExit();
+		} catch (IOException e) {
+			absolutePath = URI.create("about:blank");
+		}
+		URI_FOR_CUSTOM_TEXT_PAGE = absolutePath;
```
HeikoKlare, line 46: Just saw that there is already a static initializer block, so it would be good to integrate this into the other (though keeping the two functionalities separated in separate methods called by that initializer).

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -616,6 +649,9 @@ int handleNavigationStarting(long pView, long pArgs, boolean top) {
 	int hr = args.get_Uri(ppszUrl);
 	if (hr != COM.S_OK) return hr;
 	String url = wstrToString(ppszUrl[0], true);
+	if(isLocationForCustomText(url)) {
+		url = ABOUT_BLANK;
+	}
```
HeikoKlare, line 651: This logic to replace the actual with the fake URL if necessary is replicated three times. Would be good to wrap it into some internal, self-explaining `getExposedUrl()` method or the like.

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -907,21 +941,40 @@ public void stop() {
 }
 
 @Override
-public boolean setText(String html, boolean trusted) {
-	char[] data = new char[html.length() + 1];
-	html.getChars(0, html.length(), data, 0);
-	return webView.NavigateToString(data) == COM.S_OK;
+public URI getUriForCustomText() {
+	return URI.create(ABOUT_BLANK);
+}
```
HeikoKlare, line 946: This is unnecessary and can be removed.
```suggestion
```

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -907,21 +941,40 @@ public void stop() {
...
+
+private boolean isLocationForCustomText(String location) {
+	return URI_FOR_CUSTOM_TEXT_PAGE.equals(URI.create(location));
```
HeikoKlare, line 949: Is there a guarantee that `URI.create()` succeedson the location? I am concerned that this may result in problems like #1455

So maybe we could also do a simple string comparison?

HeikoKlare (commented):
I've tested the revised proposal and it now seems to work fine even without any changes of clients. In particular, I have tested the JDT Java element informations, which now work as expected.

@sratz This PR now implements your proposal from https://github.com/eclipse-platform/eclipse.platform.swt/pull/1395#issuecomment-2333673494
What do you think about the revised changes?

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
amartya4256: I had introduced uri comparision cause the path can have different notations
like // and / or /../../ and many more. URI comparision seems to handle that. but yes uri always needs a protocol in the beginning (file://, html://, etc) and the error that I saw in #1455 seems to occur when the location starts like /something/....

But i am pretty sure string comparision might be a bit difficult. we need a better way to check this.

laeubi: YOu should normalize the URI and catch exception, if its not a parseable URI just return false.

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
amartya4256: Done

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -64,112 +74,131 @@ public WebViewEnvironment(ICoreWebView2Environment environment) {
...
-		NativeClearSessions = () -> {
-			ICoreWebView2CookieManager manager = getCookieManager();
-			if (manager == null) return;
+		URI_FOR_CUSTOM_TEXT_PAGE = setupAndGetLocationForCustomTextPage();
+		setupCookieAndSessionHandlers();
+	}
 
-			long[] ppv = new long[1];
-			int hr = callAndWait(ppv, completion -> manager.GetCookies(null, completion));
+private static void setupCookieAndSessionHandlers() {
```
HeikoKlare, line 84: Can you please align the intendentation of the refactored methods with the existing indentation, so that not all this code is marked as modified? The formatting change makes reviewing and history lookups unnecessarily difficult.

amartya4256: In the file itself the methods don't have any indentation while the static block does. If I add indentation to the method it will look different from the other methods.

amartya4256: And I moved the code in the static block about cokkie and session to a method setupCookieAndSessionHandlers so that's why it appears moved.

HeikoKlare: Okay, indentation is inconsistent even now, I see.
Taking another look: why do we need the static initializer at all? Why not have
```java
private static final URI URI_FOR_CUSTOM_TEXT_PAGE = createUriForCustomText();
```
and the generation in that method of the URI? Is there some technical limitation is do not see?

amartya4256: Oh yeah. I just didn't think about it. Just made the change.

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
amartya4256: Okay maybe one stupid question: what to do when the protocol is missing. In the Unix filesystem, it could be that we receive paths startign with "/"
One solution would to always prepend html:// or file:// if a protocol is missing the way its done in in setWebPage:

`if (!url.matches("[a-z][a-z0-9+.-]*:.*")) {
		url = "http://" + url;
	}`
	
but again for checking the path for custom text (in isLocationForCustomText) it makes more sense to prepend a file:// instead of html://
Regarding normalizing, URI seems to take care of that implicitly, the only problem I forsee is when the protocol is missing.

laeubi: The API claims that the location is always a valid URI, so anything passed in here that is not an URI must return false.

HeikoKlare: > One solution could be to always prepend html:// or file:// if the protocol is missing, just the way its done in setWebPageData:
> 
> `if (!url.matches("[a-z][a-z0-9+.-]*:.*")) { url = "http://" + url; }`
> 
> but again, for checking the path for custom text (in isLocationForCustomText) it makes more sense to prepend a file:// instead of html://

Why is there any need to modify the location string? Currently, this is only produced internally by retrieving the current browser URL. And this should always have a protocol prefix (or be some magic string like "about:blank", shouldn't it? When I print out the locations passed to method in all the existing browser tests, this assumption is fulfilled. Maybe some specific handling for values that are, by intent, no URIs is necessary (like "about:blank")?
In case this method will be exposed at some point in time, you should only pass some URL that you have retrieved from the browser into it as well, which then should also have an according protocol prefix. In case someone passes any other string to the method (without a protocol prefix), I agree with @laeubi that this should simply return false. Is there anything speaking against this?

amartya4256: Alright then, I'll just handle the exception and return false here.

HeikoKlare: Looks sound to me. @laeubi does it reflect your proposal/expectation now?

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -906,22 +936,40 @@ public void stop() {
 	webView.Stop();
 }
 
+private boolean isLocationForCustomText(String location) {
+	try {
+		return URI_FOR_CUSTOM_TEXT_PAGE.equals(URI.create(location));
```
laeubi, line 941: I would rather use new URI, as URI.create is meant for situations where you are sure URI is valid.

HeikoKlare: Indeed, thank you for the hint! @amartya4256 can you please adapt the code?

sratz:
In general this looks good. Also, with this PR merged, my PR #1451 will become easier, since no more of those `data:text/html;` URLs will appear with the physical file URL being used.

Regarding hiding the internal URL and exposing `about:blank` for compatibility:
> @sratz This PR now implements your proposal from [#1395 (comment)](https://github.com/eclipse-platform/eclipse.platform.swt/pull/1395#issuecomment-2333673494)
> What do you think about the revised changes?

There is one more corner case to consider: The application layer consumers will now only see `about:blank` but they are still encouraged to not compare that literally, but to call `Browser#isLocationForCustomText(String)`.

That means, that this method now also needs to return `true` for an `about:blank` argument.


I have one more general remark:
Why is this HTML file a `static` one? Isn't this problematic when multiple `Edge` instance exist and all share the same file?

HeikoKlare:
> There is one more corner case to consider: The application layer consumers will now only see `about:blank` but they are still encouraged to not compare that literally, but to call `Browser#isLocationForCustomText(String)`.
> 
> That means, that this method now also needs to return `true` for an `about:blank` argument.

Currently, `Browser#setLocationForCustomText(String)` returns true exactly when passing `about:blank` into it. Still, we should revise that method (and the discussed provision of such information via the location change event) in subsequent PRs to improve the design. The result of this PR will (only) be that Edge behaves like the other browsers. A follow-up will be to enhance the Browser API to get rid of the comparison against "about:blank" in clients and to adapt the clients accordingly. In order words: this PR would work completely independent from the `Browser#isLocationForCustomText(String)` method.

> I have one more general remark: Why is this HTML file a `static` one? Isn't this problematic when multiple `Edge` instance exist and all share the same file?

In my understanding, this file is not used at all. It is just an empty dummy that allows to provide WebView with custom contents of the document to display. The file is never written, so in my opinion it makes sense to share it across all WebView instances.

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -549,7 +575,11 @@ public String getText() {
 public String getUrl() {
 	long ppsz[] = new long[1];
 	webView.get_Source(ppsz);
-	return wstrToString(ppsz[0], true);
+	return getExposedUrl(wstrToString(ppsz[0], true));
+}
+
+private String getExposedUrl(String url) {
+	return isLocationForCustomText(url) ? url = ABOUT_BLANK : url;
```
sratz, line 582:
```suggestion
	return isLocationForCustomText(url) ? ABOUT_BLANK : url;
```
(no need to also assign url)

sratz:
> Currently, `Browser#setLocationForCustomText(String)` returns true exactly when passing `about:blank` into it. Still, we should revise that method (and the discussed provision of such information via the location change event) in subsequent PRs to improve the design. The result of this PR will (only) be that Edge behaves like the other browsers. A follow-up will be to enhance the Browser API to get rid of the comparison against "about:blank" in clients and to adapt the clients accordingly. In order words: this PR would work completely independent from the `Browser#isLocationForCustomText(String)` method.

Oh, right, I was confused by the same name of the private method in `Edge` and the public method in `Browser`.

> In my understanding, this file is not used at all. It is just an empty dummy that allows to provide WebView with custom contents of the document to display. The file is never written, so in my opinion it makes sense to share it across all WebView instances.

Ah, right; we're only tricking Edge into thinking we are dealing with an actual HTML file on disk, but are then still manipulating the DOM via JavaScript.

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
amartya4256: Done

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java

HeikoKlare (approved):
I see all comments addressed and no outstanding complains with respect to this PR.

In case there are no objections, I plan to merge this by the end of the day to have it available for further testing in tomorrow's I-Build.

sratz:
This change broke working with the DOM in `ProgressListener`s.

Consider:

```java
public class Demo {

		Shell shell = new Shell();
		Browser browser = new Browser(shell, SWT.EDGE);
		browser.setText("""
				<html>
					<head>
						<script src=\"file://c:/test.js\"></script>
					</head>
					<body>
						<h1>Hello, World!</h1>
					</body>
				</html>
				""");
		browser.addProgressListener(ProgressListener.completedAdapter(event -> {
			String script = """
					var h1s = document.getElementsByTagName("h1");
					alert("Found h1s: " + h1s.length);
					"""; //$NON-NLS-1$
			browser.execute(script);
		}));

		shell.open();
		while (!shell.isDisposed()) {
				display.sleep();
			}
		}
		display.dispose();
	}

}
```

This should bring up a `Found h1s: 1` alert, but actually shows `Found h1s: 0`.

The debugging tools show:
 ```
VM1955:1 [Violation] Avoid using document.write().
VM1955:1 [Violation] Parser was blocked due to document.write(<script>)
VM1955:1 [Violation] Parser was blocked due to document.write(<script>)
```

Changing 
```diff
-execute("document.open(); document.write(`" + lastCustomText + "`); document.close();");
+execute("document.documentElement.innerHTML = `" + lastCustomText + "`;");
```
fixes the problem

PR #1397, opened 2024-08-09
Missing scrollbars in compare editor

ShahzaibIbrahim:
Scrollbars are missing on non default screen scaling [Blocker for 4.33 M3 build]. Regression occurred due to https://github.com/eclipse-platform/eclipse.platform.swt/pull/1209

Contributes to: https://github.com/eclipse-platform/eclipse.platform/issues/1489


merks:
I'll merge this because it was definitely broken before and discussions on the issue agree this is the right fix.

PR #1399, opened 2024-08-10
Switch to temurin JVM

akurtakov:
Openjdk one is stuck at 17.0.2 while temurin one is updated to newer builds.


laeubi:
@akurtakov are there any specific bugfixes we require for Java 17?

akurtakov:
No, the reason for that change is to run tests with an uptodate java 17 jvm.

PR #1401, opened 2024-08-12
Remove getZoomForAutoscaleProperty calls for win32

amartya4256:
This PR contributes to the removal of getZoomForAutoscaleProperty  calls for win32 which is used to have the right zoom level for the widgets based on what the native zoom of the monitor is. However, this method is only needed for images and the rest can be taken care of by the Layout Manager while using native zoom.

contributes #62 and #127


akurtakov:
This seems to have stalled. Is it still needed?

HeikoKlare:
Yes, this is still under investigation for Windows HiDPI support. @amartya4256 can you please add a short documentation to this PR explaining what it is supposed to do/evaluate?

amartya4256:
@HeikoKlare I have updated the description, but the work is on halt right now since we have to figure out the image scaling issue first.

amartya4256:
This PR is obsolete now. Closing it.

Issue #1317, opened 2024-07-03
Move TabFolderLayout into SWT and set it as a default in TabFolder


laeubi:
Today I debugged some issues with `TabFolder` and noticed that it at least behaves strange if tabs are added dynamically, after investigate a little bit more I tried to fix it with an own Layout as usual layout mangers (like `FillLayout`) get sometimes confused as `TabFolder` has more childs than are actually displayed.

Short after that I noticed that there are already some instances of `TabFolderLayout` in platform:
![grafik](https://github.com/eclipse-platform/eclipse.platform.swt/assets/1331477/ed71f6db-1a50-44cd-a960-a29feb278f43)

They all seem to be copies of each other (more or less) except the license header and when they where added, I therefore wonder if we should simply add that layout to SWT and even set it by default to `TabFolder`.

Kiiv:
Hi !

Seems a HUGE breaking change to me. What about "old" plugins steel using those classes ?

For example, after upgrading to Eclipse 2024-12, Subversive does not work anymore :
```
Opération de mise à jour échouée .
java.lang.NoClassDefFoundError: org/eclipse/compare/internal/TabFolderLayout
```

I will open an issue over-there but in the meantime I'll probably need to revert to Eclipse 2024-09 (and it's not easy :s).
Since Subversive, even if it's part of the Eclipse Foundation, does not seems alive, I have no much hope.

Any workaround ?

dogla:
Hi Kiiv,

> For example, after upgrading to Eclipse 2024-12, Subversive does not work anymore :
> ```
> java.lang.NoClassDefFoundError: org/eclipse/compare/internal/TabFolderLayout
> ```

I had the same problem.

To use eclipse 4.34 with a working subversive I opened the org.eclipse.compare_3.11.200.v20241024-1455.jar and added the TabFolderLayout.class from a previous eclipse version, in my case from 4.33, into the right subfolder.
After starting eclipse SVN worked as before.

It's a hack but it works for me.
So I can use the latest eclipse with an old subversive version, that depends on the removed org.eclipse.compare.internal.TabFolderLayout.

Rajagopalang:
@dogla Working fine your solution...I have added same TabFolderLayout.class from old version to org.eclipse.compare_3.11.200.v20241024-1455.jar... Thanks lot...

![image](https://github.com/user-attachments/assets/0c31b652-d651-48ae-ac27-11ce54de7fb5)

[...]

PR #1402, opened 2024-08-13
Moving TabFolderLayout into SWT.

deepika-u:
Moving TabFolderLayout into SWT.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/1317

This pr needs to be merged first and later https://github.com/eclipse-platform/eclipse.platform.ui/pull/2186

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/layout/TabFolderLayout.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/layout/TabFolderLayout.java
@@ -1,0 +1,64 @@
+/*******************************************************************************
+ * Copyright (c) 2000, 2005 IBM Corporation and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+
+package org.eclipse.swt.layout;
+
+import org.eclipse.swt.*;
+import org.eclipse.swt.graphics.*;
+import org.eclipse.swt.widgets.*;
+
+/**
+ * This layout controls the position and size of the children of a tab folder.
+ * In addition this layout can be used for any component that wants
+ * TabFolder-Like behaviour (e.g. only showing all component at the full size
+ * of the composite stacked onto each other)
+ *
+ * @since 3.128
+ */
+public class TabFolderLayout extends Layout {
+
+	@Override
+	protected Point computeSize (Composite composite, int wHint, int hHint, boolean flushCache) {
+		if (wHint != SWT.DEFAULT && hHint != SWT.DEFAULT)
+			return new Point(wHint, hHint);
+
+		Control [] children = composite.getChildren ();
+		int count = children.length;
+		int maxWidth = 0, maxHeight = 0;
+		for (int i=0; i<count; i++) {
+			Control child = children [i];
+			Point pt = child.computeSize (SWT.DEFAULT, SWT.DEFAULT, flushCache);
+			maxWidth = Math.max (maxWidth, pt.x);
+			maxHeight = Math.max (maxHeight, pt.y);
+		}
+
+		if (wHint != SWT.DEFAULT)
+			maxWidth= wHint;
+		if (hHint != SWT.DEFAULT)
+			maxHeight= hHint;
+
+		return new Point(maxWidth, maxHeight);
+
+	}
+
+	@Override
+	protected void layout (Composite composite, boolean flushCache) {
+		Rectangle rect= composite.getClientArea();
+
+		Control[] children = composite.getChildren();
+		for (Control c : children) {
+			c.setBounds(rect);
+		}
+	}
+}


bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/layout/TabFolderLayout.java
```diff
@@ -0,0 +1,68 @@
...
+
+import org.eclipse.swt.*;
+import org.eclipse.swt.graphics.*;
+import org.eclipse.swt.widgets.*;
+
+/**
+ * This layout controls the position and size
+ * of the children of a tab folder.
+ *
+ * @since 2.1
```
merks, line 25: Is this value correct?

deepika-u: I havent updated the content of the file actually. What should be the value ideally?

merks: It should be the current version of the bundle. If the bundle does currently have a 0 micro version the minor version should be incremented because the is new API and that new version should be used. I’m on an iPhone so it’s hard to check.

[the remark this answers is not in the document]
deepika-u: Done, thank you.

deepika-u:
@laeubi @merks : Can this pr be merged please? or anything else needs to be done?

merks:
Has this been rebased on master?

I see this as the current version of SWT:

image

laeubi:
> Has this been rebased on master?

Done!

akurtakov (changes requested):
This class totally lacks documentation compared to rest of SWT. From just looking at it I have zero clue what/why/how/when to use it. Please add proper documentation before it can be accepted in SWT.

laeubi:
> This class totally lacks documentation compared to rest of SWT.

Can you be more specific? 

> I have zero clue what/why/how/when to use it

e.g I see this 

```
This layout controls the position and size
of the children of a tab folder.
```

As explained in the referenced issues, this is somehow not often used directly (but not prohibited of course) if I look for example at `CTabFolderLayout` it has similar (quite basic) documentation.

So while better documentation is always good, I think its hard to ask the person that wants to help deduplication to do it, so I would more see this as something for all SWT/Platform commiters to enhance after we have deduplicated the code if there are anything left that is really unclear in its usage.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/layout/TabFolderLayout.java
```diff
@@ -0,0 +1,68 @@
...
+import org.eclipse.swt.graphics.*;
+import org.eclipse.swt.widgets.*;
+
+/**
+ * This layout controls the position and size
+ * of the children of a tab folder.
+ *
+ * @since 3.127
+ */
+class TabFolderLayout extends Layout {
```
laeubi, line 27: Can this really be package private?

laeubi (commented):
I think it would be good if this PR would include setting this layout in the `TabFolder` as a default layout, this would also make it more clear how/when/why use it.

If it should be package private I think it needs to reside in the same package as `TabFolder`

akurtakov:
Well, I still don't understand what the effects of using this one are. 
P.S. Don't even get me started on all C* widgets and friends - they should be deleted IMO.

akurtakov:
Also looking into the issue all the classes that are to be deduplicated are "internal" so they are not supposed to be generally useful which explains (to some extend) lack of documentation. I don't ask for big documentation but at least some basic one on what the effects of using it are.

akurtakov:
One more observation this seems to be used with CTabFolder from the platform.ui rather than TabFolder. What happens when used with TabFolder (some screenshots please)?  It should be clear whether it's for one or the other or both TabFolders and if the later maybe clarified in the name.

laeubi:
I use it for `TabFolder` (like reported on the issue), but the layout can be used for any component that wants TabFolder-Like behaviour (e.g. only showing all component at the full size of the composite stacked onto each other)

akurtakov:
> I use it for `TabFolder` (like reported on the issue), but the layout can be used for any component that wants TabFolder-Like behaviour (e.g. only showing all component at the full size of the composite stacked onto each other)

Consider putting such an explanation as a class comment :)

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/layout/TabFolderLayout.java
```diff
@@ -0,0 +1,68 @@
...
+
+	/*
```
akurtakov, line 29: Remove the see tag. It just generates clutter in the codebase.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/layout/TabFolderLayout.java
```diff
@@ -0,0 +1,68 @@
...
+		if (wHint != SWT.DEFAULT)
+			maxWidth= wHint;
+		if (hHint != SWT.DEFAULT)
+			maxHeight= hHint;
+
+		return new Point(maxWidth, maxHeight);
+
+	}
+
+	/*
```
akurtakov, line 56: Remove this comment with see tag only.

deepika-u: From the above conversations, updated the file accordingly.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/layout/TabFolderLayout.java
akurtakov: @deepika-u You haven't answered here? How is this supposed to be used from platform.ui?

deepika-u: Yeah totally agree when no modifier is mentioned it has defaulted to private. For this class to be available for platform.ui it needs to be public. It is done now.

deepika-u: Is anything being missing to be done from my side? Please do let me know.

akurtakov (approved):

deepika-u:
@akurtakov : Thanks alot. Now i'll try to focus on other repo changes for which this is the starter. PDE, Platform and ui accordingly.

laeubi (commented):
Thank you for driving this forward!

PR #1404, opened 2024-08-13
Edge: Implement ICoreWebView2Controller#add_AcceleratorKeyPressed()

### LINKED ISSUE #1403: Edge: implement ICoreWebView2Controller#add_AcceleratorKeyPressed() to receive basic key combinations (to allows e.g. an 'Element Info' popup to be closed via ESC key)
The Edge-based `Browser` implementation currently does not receive any key up/down events from the WebView2 engine.

The WebView2 offers no API to get all key events, but there is `ICoreWebView2Controller#add_AcceleratorKeyPressed()` which allows us to receive some (many) key events:
https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2controller?view=webview2-1.0.2592.51#add_acceleratorkeypressed

Without this, it is currently impossible to handle an ESCAPE key press in SWT, making it e.g. impossible to close an 'Code Element Information' popup using the ESCAPE key when using edge.

sratz:
The ICoreWebView2AcceleratorKeyPressedEvent does not give us all key events and is therefore much less complete than the IE.handleDOMEvent() equivalent.

But at least this allows us to react on VK_ESCAPE and similar things, allowing e.g. element info popups to be closed via ESC.

Resolves #1403

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
@@ -429,6 +429,9 @@ public void create(Composite parent, int style) {
 	handler = newCallback(this::handleGotFocus);
 	controller.add_GotFocus(handler, token);
 	handler.Release();
+	handler = newCallback(this::handleAcceleratorKeyPressed);
+	controller.add_AcceleratorKeyPressed(handler, token);
+	handler.Release();
 	if (webView_2 != null) {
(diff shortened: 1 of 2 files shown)


sratz:
Any objections to merge this for 2024-09 M3?

We are currently adopting Edge selectively in some places in our product and this has been one of the showstoppers.

akurtakov:
As Edge is not enabled by default for any users it's fine to get this change in.

sratz:
OK, I cleaned up the Javadoc a bit and will merge after the build succeeds.

PR #1405, opened 2024-08-14
Drawing custom menu items for Menu Bar **Dark Theme Only**

ShahzaibIbrahim:
Using gc to draw menu item, moving the responsibility from OS due to wrong scaling of menu bar horizontally in **Dark** Theme. Snippet 373 can be used to test menuItems with images. 

Additional points covered in this PR: 

- Changing the text color from a grayish tone to white.
- When ALT key is pressed mnemonics are underlined and work as a toggle, the behavior which was missing from dark theme previously.

Screenshot of the fix: 

image

## HOW TO TEST

- Run the `RunTimeWorkspace` with the following **VM Arguments**
```
-Dswt.autoScale=quarter
```
- Go to "Windows" -> "Preferences" -> "General" -> "Appearance" 
- Change theme to "Dark" 
- Click "Apply and Close" 
- Move the window from 100 -> 175 (or anything 100+) zoom level
- See if menu bar is properly aligned.

## SCENARIOS TESTED

autoscale.updateOnRuntime = false && autoscale=int200 (Prevent regressions)
100% (Primary), 175% (Second)
150% (Primary), 200% (Second)
175% (Primary), 125% (Second)
200% (Primary), 100% (Second)
200% (Primary), 150% (Second)
autoscale.updateOnRuntime = false && autoscale=quarter (Prevent regressions)
150% (Primary), 200% (Second)
175% (Primary), 125% (Second)
autoscale.updateOnRuntime = true && autoscale=int200
100% (Primary), 175% (Second)
150% (Primary), 200% (Second)
175% (Primary), 125% (Second)
200% (Primary), 100% (Second)
autoscale.updateOnRuntime = true && autoscale=quarter
150% (Primary), 200% (Second) 
175% (Primary), 125% (Second)

## EXPECTED BEHAVIOUR

The menu bar should be scaled properly.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MenuItem.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MenuItem.java
@@ -47,7 +47,12 @@ public class MenuItem extends Item {
 	/* Image margin. */
 	final static int MARGIN_WIDTH = 1;
 	final static int MARGIN_HEIGHT = 1;
-
+	private final static int LEFT_TEXT_MARGIN = 7;
+	private final static int IMAGE_TEXT_GAP = 3;
+	// There is a weird behavior in the Windows API with menus in OWENERDRAW mode that the returned
+	// value in wmMeasureChild is increased by a fixed value (in points) when wmDrawChild is called
+	// This static is used to mitigate this increase
+	private final static int WINDOWS_OVERHEAD = 6;
 	static {
 		DPIZoomChangeRegistry.registerHandler(MenuItem::handleDPIChange, MenuItem.class);
 	}
@@ -1121,10 +1126,32 @@ LRESULT wmCommandChild (long wParam, long lParam) {
 	return null;
 }
 
-LRESULT wmDrawChild (long wParam, long lParam) {
-	DRAWITEMSTRUCT struct = new DRAWITEMSTRUCT ();
-	OS.MoveMemory (struct, lParam, DRAWITEMSTRUCT.sizeof);
-	if (image != null) {
+@Override
+GC createNewGC(long hDC, GCData data) {
+	if (getDisplay().isRescalingAtRuntime()) {
+		return super.createNewGC(hDC, data);
+	} else {
+		data.nativeZoom = getMonitorZoom();
+		return GC.win32_new(hDC, data);
+	}
+}
+
+private int getMonitorZoom() {
+	return getMenu().getShell().getMonitor().zoom;
+}
+
+private int getMenuZoom() {
+	if (getDisplay().isRescalingAtRuntime()) {
+		return super.getZoom();
+	} else {
+		return DPIUtil.getZoomForAutoscaleProperty(getMonitorZoom());
+	}
+}
+
+LRESULT wmDrawChild(long wParam, long lParam) {
+	DRAWITEMSTRUCT struct = new DRAWITEMSTRUCT();
+	OS.MoveMemory(struct, lParam, DRAWITEMSTRUCT.sizeof);
+	if ((text != null || image != null)) {
 		GCData data = new GCData();
 		data.device = display;
 		GC gc = createNewGC(struct.hDC, data);
@@ -1134,14 +1161,50 @@ LRESULT wmDrawChild (long wParam, long lParam) {
 		* coordinate.  The fix is to ignore this value when
 		* the item is in a menu bar.
 		*/
-		int x = (parent.style & SWT.BAR) != 0 ? MARGIN_WIDTH * 2 : struct.left;
-		Image image = getEnabled () ? this.image : new Image (display, this.image, SWT.IMAGE_DISABLE);
-		int zoom = getZoom();
-		gc.drawImage (image, DPIUtil.scaleDown(x, zoom), DPIUtil.scaleDown(struct.top + MARGIN_HEIGHT, zoom));
-		if (this.image != image) image.dispose ();
-		gc.dispose ();
+		int x = (parent.style & SWT.BAR) == 0 ? MARGIN_WIDTH * 2 : struct.left;
+		int zoom = getMenuZoom();
+		Rectangle menuItemArea = null;
+		if (text != null) {
+			this.getParent().redraw();
+			int flags = SWT.DRAW_DELIMITER;
+			boolean isInactive = ((struct.itemState & OS.ODS_INACTIVE) != 0);
+			boolean isSelected = ((struct.itemState & OS.ODS_SELECTED) != 0);
+			boolean isNoAccel = ((struct.itemState & OS.ODS_NOACCEL) != 0);
+
+			String drawnText = "";
+			if(isNoAccel) {
+				drawnText = this.text.replace("&", "");
+			} else {
+				drawnText = this.text;
+				flags |= SWT.DRAW_MNEMONIC;
+			}
+			Rectangle menuItemBounds = this.getBounds();
+
+			int fillMenuWidth =  DPIUtil.scaleDown(menuItemBounds.width, zoom);
+			int fillMenuHeight = DPIUtil.scaleDown(menuItemBounds.height, zoom);
+			menuItemArea = new Rectangle(DPIUtil.scaleDown(x, zoom), DPIUtil.scaleDown(struct.top, zoom), fillMenuWidth, fillMenuHeight);
+
+			gc.setForeground(isInactive ? display.getSystemColor(SWT.COLOR_GRAY) : display.getSystemColor(SWT.COLOR_WHITE));
+			gc.setBackground(isSelected ? display.getSystemColor(SWT.COLOR_DARK_GRAY) : parent.getBackground());
+			gc.fillRectangle(menuItemArea);
+
+			int xPositionText = LEFT_TEXT_MARGIN + DPIUtil.scaleDown(x, zoom) + (this.image != null ? this.image.getBounds().width + IMAGE_TEXT_GAP : 0);
+			int yPositionText = DPIUtil.scaleDown(struct.top, zoom) + MARGIN_HEIGHT;
+			gc.drawText(drawnText, xPositionText, yPositionText, flags);
+		}
+		if (image != null) {
+			Image image = getEnabled() ? this.image : new Image(display, this.image, SWT.IMAGE_DISABLE);
+			int gap = (menuItemArea.height - image.getBounds().height) / 2;
+			gc.drawImage(image, LEFT_TEXT_MARGIN + DPIUtil.scaleDown(x, zoom), gap + DPIUtil.scaleDown(struct.top, zoom));
+			if (this.image != image) {
+				image.dispose();
+			}
+		}
+		gc.dispose();
 	}
-	if (parent.foreground != -1) OS.SetTextColor (struct.hDC, parent.foreground);
+	if (parent.foreground != -1) {
+		OS.SetTextColor(struct.hDC, parent.foreground);
+	}
 	return null;
 }
 
@@ -1151,6 +1214,9 @@ LRESULT wmMeasureChild (long wParam, long lParam) {
 
 	if ((parent.style & SWT.BAR) != 0) {
 		if (parent.needsMenuCallback()) {
+			Point point = calculateRenderedTextSize();
+			int menuZoom = getDisplay().isRescalingAtRuntime() ? super.getZoom() : getMonitorZoom();
+			struct.itemHeight = DPIUtil.scaleUp(point.y, menuZoom);
 			/*
 			 * Weirdness in Windows. Setting `HBMMENU_CALLBACK` causes
 			 * item sizes to mean something else. It seems that it is
@@ -1158,10 +1224,9 @@ LRESULT wmMeasureChild (long wParam, long lParam) {
 			 * if menu item has a mnemonic, it's always drawn at a fixed
 			 * position. I have tested on Win7, Win8.1, Win10 and found
 			 * that value of 5 works well in matching text to mnemonic.
-			 * NOTE: autoScaleUpUsingNativeDPI() is used to avoid problems
-			 * with applications that disable automatic scaling.
 			 */
-			struct.itemWidth = DPIUtil.scaleUp(5, nativeZoom);
+			int horizontalSpaceImage = this.image != null ? this.image.getBounds().width + IMAGE_TEXT_GAP: 0;
+			struct.itemWidth = DPIUtil.scaleUp(LEFT_TEXT_MARGIN + point.x - WINDOWS_OVERHEAD + horizontalSpaceImage, menuZoom);
 			OS.MoveMemory (lParam, struct, MEASUREITEMSTRUCT.sizeof);
 			return null;
 		}
@@ -1203,6 +1268,32 @@ LRESULT wmMeasureChild (long wParam, long lParam) {
 		OS.MoveMemory (lParam, struct, MEASUREITEMSTRUCT.sizeof);
 	}
 	return null;
+}
+
+private Point calculateRenderedTextSize() {
+	GC gc = new GC(this.getMenu().getShell());
+	String textWithoutMnemonicCharacter = getText().replace("&", "");
+	Point points = gc.textExtent(textWithoutMnemonicCharacter);
+	gc.dispose();
+
+	if (!getDisplay().isRescalingAtRuntime()) {
+		int primaryMonitorZoom = this.getDisplay().getDeviceZoom();
+		int adjustedPrimaryMonitorZoom = DPIUtil.getZoomForAutoscaleProperty(primaryMonitorZoom);
+		if (primaryMonitorZoom != adjustedPrimaryMonitorZoom) {
+			// Windows will use a font matching the native primary monitor zoom for calculating the size in pixels,
+			// GC will use the native primary monitor zoom to scale down from pixels to points in this scenario
+			// Therefore we need to make sure adjust the points as if it would have been scaled down by the
+			// native primary monitor zoom.
+			// Example:
+			// Primary monitor on 150% with int200: native zoom 150%, adjusted zoom 100%
+			// Pixel height of font in this example is 15px
+			// GC calculated height of 15px, scales down with adjusted zoom of 100% and returns 15pt -> should be 10pt
+			// this calculation is corrected by the following line
+			// This is the only place, where the GC needs to use the native zoom to do that, therefore it is fixed only here
+			points = DPIUtil.scaleDown(DPIUtil.scaleUp(points, adjustedPrimaryMonitorZoom), primaryMonitorZoom);
+		}
+	}
+	return points;
 }
 
 private static final class MenuItemToolTip extends ToolTip {
--- a/bundles/org.eclipse.swt/Eclipse SWT PI/win32/org/eclipse/swt/internal/win32/OS.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT PI/win32/org/eclipse/swt/internal/win32/OS.java
@@ -897,6 +897,7 @@ public class OS extends C {
 	public static final int MFS_CHECKED = 0x8;
 	public static final int MFS_DISABLED = 0x3;
 	public static final int MFS_GRAYED = 0x3;
+	public static final int MFT_OWNERDRAW = 0x100;
 	public static final int MFT_RADIOCHECK = 0x200;
 	public static final int MFT_RIGHTJUSTIFY = 0x4000;
 	public static final int MFT_RIGHTORDER = 0x2000;
@@ -1019,6 +1020,8 @@ public class OS extends C {
 	public static final int OBJ_PEN = 0x1;
 	public static final int OBM_CHECKBOXES = 0x7ff7;
 	public static final int ODS_SELECTED = 0x1;
+	public static final int ODS_NOACCEL = 0x0100;
+	public static final int ODS_INACTIVE = 0x80;
 	public static final int ODT_MENU = 0x1;
 	public static final int OIC_BANG = 0x7F03;
 	public static final int OIC_HAND = 0x7F01;
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Menu.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Menu.java
@@ -358,7 +358,7 @@ void createItem (MenuItem item, int index) {
 	info.fMask = OS.MIIM_ID | OS.MIIM_TYPE | OS.MIIM_DATA;
 	info.wID = item.id;
 	info.dwItemData = item.id;
-	info.fType = item.widgetStyle ();
+	info.fType = (style & SWT.BAR) != 0 && needsMenuCallback() ? OS.MFT_OWNERDRAW :  item.widgetStyle ();
 	info.dwTypeData = pszText;
 	boolean success = OS.InsertMenuItem (handle, index, true, info);
 	if (pszText != 0) OS.HeapFree (hHeap, 0, pszText);


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MenuItem.java
```diff
@@ -1161,7 +1178,8 @@ LRESULT wmMeasureChild (long wParam, long lParam) {
 			 * NOTE: autoScaleUpUsingNativeDPI() is used to avoid problems
 			 * with applications that disable automatic scaling.
 			 */
-			struct.itemWidth = DPIUtil.scaleUp(5, nativeZoom);
```
HeikoKlare, line 1164: This has set the left margin of an entry before. now the margin appears to be zero and has to be added at some point. The first element is drawn exactly at x coordinate 0, which looks a bit odd:
Screenshot 2024-08-19 130842

Here is a comparison with the existing spacing (on top). Not only the initial left margin is different, but also the other spacing seems to be different (larger than before):
Screenshot 2024-08-19 130854

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MenuItem.java
```diff
@@ -1121,27 +1124,38 @@ LRESULT wmCommandChild (long wParam, long lParam) {
...
 		int zoom = getZoom();
-		gc.drawImage (image, DPIUtil.scaleDown(x, zoom), DPIUtil.scaleDown(struct.top + MARGIN_HEIGHT, zoom));
-		if (this.image != image) image.dispose ();
-		gc.dispose ();
+		if (text != null) {
+			int flags = SWT.DRAW_MNEMONIC | SWT.DRAW_DELIMITER;
+			boolean isInactive = ((struct.itemState & OS.ODS_INACTIVE) != 0);
+			boolean isSelected = ((struct.itemState & OS.ODS_SELECTED) != 0);
+			gc.setForeground(isInactive? display.getSystemColor(SWT.COLOR_GRAY) :  display.getSystemColor(SWT.COLOR_WHITE));
+			gc.setBackground(isSelected? display.getSystemColor(SWT.COLOR_BLUE) : parent.getBackground());
```
HeikoKlare, line 1142: While in general I like the new colors more than the old ones (in particular the white instead of the gray text in a selected entry), the region that is colored is incorrect. Only the label is affected, but actually the background of the whole menu item should be drawn in that color, as it was done before. Maybe you can simply fill the whole menu item with that color first? And maybe it is also possible to use the old blue tone for that?
Here is how a selected item looks now:
Screenshot 2024-08-19 130943

In here is how it looked before:
Screenshot 2024-08-19 130933

You can also see that the text color has changed in general (from a grayish tone to white), which I personally prefer, but we have to consider that this is an additional change.

HeikoKlare (commented):
I tested the changes with different scale factors and auto-scale settings (quarter/integer200). The scaling results look fine, in constrast to the existing scaling behavior that was broken.

I have some detailed comments and a general question: are there snippets to test the behavior also in combination with images? Our default applications only have menus with text but it would also be interested to test the changes with images, so we should either have or provide a snippet with images for validation.

Please also improve the commit message according to the [recommendations](https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md#commit-message-recommendations), such that it explains the existing issue and more precisely describes the fix. E.g., it currently sounds as if the menu is always draw in a custom way by this change, but it only affects the behavior when using dark theme.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Device.java
```diff
@@ -645,7 +645,8 @@ public Color getSystemColor (int id) {
 		case SWT.COLOR_DARK_GREEN:			pixel = 0x00008000;  break;
 		case SWT.COLOR_YELLOW:				pixel = 0x0000FFFF;  break;
 		case SWT.COLOR_DARK_YELLOW:			pixel = 0x00008080;  break;
-		case SWT.COLOR_BLUE:				pixel = 0x00FF0000;  break;
+		case SWT.COLOR_LIGHT_BLUE:			pixel = 0x00FF9933;  break;
```
HeikoKlare, line 648: Adding a new system color for require adaptations of every OS implementation (i.e., also in the `Device` classes of MacOS and Linux). I am not sure whether we should really add a color for this use case. In particular, the color does not seem to have a specific meaning. In my Windows installation, the color currently used for the menu bar in dark mode does not correspond to this value, i.e., it may also depend on the OS version. Since it does not fit to the hightlight color of menu items either (see screenshot below) and since in light theme there is no similar hightlight color used as well (also see screenshot below).

image
image

I wonder whether we should not better just use one of the already defined color that we consider fitting. Since the hovering is also confirmed by opening the menu, a good visible color is not as important as, e.g., for menu items, where you want to see over which you currently hover. What about, for example, just using `COLOR_DARK_GRAY`? That would look as follows.
image

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MenuItem.java
```diff
@@ -1121,27 +1124,47 @@ LRESULT wmCommandChild (long wParam, long lParam) {
...
+		int x = (parent.style & SWT.BAR) == 0 ? MARGIN_WIDTH * 2 : struct.left;
 		int zoom = getZoom();
-		gc.drawImage (image, DPIUtil.scaleDown(x, zoom), DPIUtil.scaleDown(struct.top + MARGIN_HEIGHT, zoom));
-		if (this.image != image) image.dispose ();
-		gc.dispose ();
+		if (text != null) {
+			int flags = SWT.DRAW_MNEMONIC | SWT.DRAW_DELIMITER;
+			boolean isInactive = ((struct.itemState & OS.ODS_INACTIVE) != 0);
+			boolean isSelected = ((struct.itemState & OS.ODS_SELECTED) != 0);
+			int textMargin = 10; // left margin for x-coordinate spacing
```
HeikoKlare, line 1141: Please extract this into a static constant and give it a proper, self-explaining name that makes the comment obsolete (like `leftTextMargin`).

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MenuItem.java
```diff
@@ -1121,27 +1124,47 @@ LRESULT wmCommandChild (long wParam, long lParam) {
...
+
+            Rectangle bounds = this.getBounds();
+            gc.setForeground(isInactive? display.getSystemColor(SWT.COLOR_GRAY) :  display.getSystemColor(SWT.COLOR_WHITE));
+			gc.setBackground(isSelected? display.getSystemColor(SWT.COLOR_LIGHT_BLUE) : parent.getBackground());
+			int xMargin = DPIUtil.scaleDown(x + textMargin , zoom) + (this.image != null? this.image.getBounds().width: 0);
```
HeikoKlare, line 1146: Please fix broken formatting (this marker is only made exemplarily)

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MenuItem.java
```diff
@@ -1121,27 +1124,47 @@ LRESULT wmCommandChild (long wParam, long lParam) {
...
+            gc.fillRectangle( DPIUtil.scaleDown(x, zoom), DPIUtil.scaleDown(struct.top + MARGIN_HEIGHT, zoom),  DPIUtil.scaleDown(bounds.width + (image != null ? image.getBounds().width : 0), zoom),  DPIUtil.scaleDown(bounds.height, zoom));
+
+			gc.drawText(this.text, xMargin, yMargin, flags);
+		}
+		if (image != null) {
+			/*
+			 * Bug in Windows. When a bitmap is included in the menu bar, the HDC seems to
+			 * already include the left coordinate. The fix is to ignore this value when the
+			 * item is in a menu bar.
+			 */
```
HeikoKlare, line 1158: This comment is misplaced. Preivously, it belonged to a different line of code.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MenuItem.java
```diff
@@ -1121,27 +1124,47 @@ LRESULT wmCommandChild (long wParam, long lParam) {
...
+			Image image = getEnabled() ? this.image : new Image(display, this.image, SWT.IMAGE_DISABLE);
+
+			gc.drawImage(image, DPIUtil.scaleDown(x, zoom) , DPIUtil.scaleDown(struct.top + MARGIN_HEIGHT, zoom));
+			if (this.image != image)
+				image.dispose();
```
HeikoKlare, line 1163: Please enclose blocks in braces.

HeikoKlare (changes requested):
Thank you for the recent improvements. There are several improvements in the new version, that's great. In particular, this change also improves the visualization of menus with images. This is how it looked before:
image
And this is how it looks now:
image
Only problem is that the left margin is always applied to the text, even if an image is present. But in that case, the margin should be applied to the image (and maybe some additional small margin needs to be placed beetween image and text). This is, e.g., how it looks for the light theme:
image

One additional issue I found concerns the handling of mnemonics, i.e., when pressing Alt+$ for accesing menu item with menmonic $.
1. The visualization is broken: the background color of the menu items does not fit when pressing the "Alt" button (it is black while the rest of the menu is grayish). See the screenshot below.
2. Mnemonics are always displayed (underlined) and not only when Alt is pressed. This is, actually, existing behvior, but differs from default behavior, such as in light theme, where the mnemonics are only underlined when Alt is pressed. Maybe we could address this as a follow up, as no regression is introduced. On the contrary, in the current behavior, pressing Alt hides the mnemonics, so badly inverts the intended behavior.
3. Most severe issue: there is no reaction to key bindings for mnemonics. E.g., pressing Alt+F should open the "File" menu, but does nothing.

image

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Device.java
```diff
@@ -645,7 +645,6 @@ public Color getSystemColor (int id) {
 		case SWT.COLOR_DARK_GREEN:			pixel = 0x00008000;  break;
 		case SWT.COLOR_YELLOW:				pixel = 0x0000FFFF;  break;
 		case SWT.COLOR_DARK_YELLOW:			pixel = 0x00008080;  break;
-		case SWT.COLOR_LIGHT_BLUE:			pixel = 0x00FF9933;  break;
```
akurtakov: If you introduce such new API you shoud do it for Cocoa and GTK too.

ShahzaibIbrahim: We are now sticking to the default color that is already present i.e. COLOR_DARK_GRAY

HeikoKlare:
Something seems to be wrong with the PR now. No matter whether I rebase on #1349, I see two issues:
1. The left margin of each menu item is quite large 
image
2. There are unexpected menu items, like another "File" entry with only a single entry in the menu that opens up from it 
image

Can you please check that, @ShahzaibIbrahim?

HeikoKlare:
I retested and found that the faulty second "File" entry is because of a broken workbench.xmi on my side. Still, it shows some placement issue, as the text is cut off (hapenning for a "File" entry without a mnemonic, while the original "File" entry with mnemonic is not cut off). That may be investigated. You can easily reproduce by manually adding two menus (with at least one dummy menu item in each of them to have the menu shown) through the Model Spy:
image

HeikoKlare:
Please check your latest commit. Only considering the diff, it does not seem to be in the intended/final state. There is a `System.out.println()` statement and some some added magic number in addition to the essential change:
image

ShahzaibIbrahim:
> Please check your latest commit. Only considering the diff, it does not seem to be in the intended/final state. There is a `System.out.println()` statement and some some added magic number in addition to the essential change: image

Yes sorry, that was a leftover.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MenuItem.java
```diff
@@ -1151,17 +1176,12 @@ LRESULT wmMeasureChild (long wParam, long lParam) {
...
-			/*
-			 * Weirdness in Windows. Setting `HBMMENU_CALLBACK` causes
-			 * item sizes to mean something else. It seems that it is
-			 * the size of left margin before the text. At the same time,
-			 * if menu item has a mnemonic, it's always drawn at a fixed
-			 * position. I have tested on Win7, Win8.1, Win10 and found
-			 * that value of 5 works well in matching text to mnemonic.
-			 * NOTE: autoScaleUpUsingNativeDPI() is used to avoid problems
-			 * with applications that disable automatic scaling.
-			 */
```
HeikoKlare, line 1163: Isn't this relevant anymore, so that the width can be set to a different value?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MenuItem.java
```diff
@@ -1151,17 +1176,12 @@ LRESULT wmMeasureChild (long wParam, long lParam) {
...
-			struct.itemWidth = DPIUtil.scaleUp(5, nativeZoom);
+			GC gc = new GC (display);
+			String textWithoutMnemonicCharacter = getText().replace("&", "");
+			Point points = gc.textExtent(textWithoutMnemonicCharacter);
+			gc.dispose ();
```
HeikoKlare, line 1182: Would be good to put this into a self-explaining method, such as `calculateRenderedTextSize()`.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MenuItem.java
```diff
@@ -47,7 +47,8 @@ public class MenuItem extends Item {
 	/* Image margin. */
 	final static int MARGIN_WIDTH = 1;
 	final static int MARGIN_HEIGHT = 1;
-
+	final static int LEFT_TEXT_MARGIN = 15;
+	final static int ADDED_ITEM_WIDTH = 15;
```
HeikoKlare, line 51: How are these values derived? The margins look rather different to the existing implementation and to the light theme. In the existing implementation, a value of "5" was hardcoded for the left margin at some point, maybe that would fit as well?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MenuItem.java
```diff
@@ -1121,27 +1122,51 @@ LRESULT wmCommandChild (long wParam, long lParam) {
...
+				drawnText = this.text;
+				flags |= SWT.DRAW_MNEMONIC;
+			}
+			Rectangle bounds = this.getBounds();
+
+			gc.setForeground(isInactive ? display.getSystemColor(SWT.COLOR_GRAY) : display.getSystemColor(SWT.COLOR_WHITE));
+			gc.setBackground(isSelected? display.getSystemColor(SWT.COLOR_DARK_GRAY) : parent.getBackground());
+			gc.fillRectangle(DPIUtil.scaleDown(x, zoom), DPIUtil.scaleDown(struct.top, zoom), DPIUtil.scaleDown(bounds.width + (image != null ? image.getBounds().width : 0), zoom), DPIUtil.scaleDown(bounds.height, zoom));
+
+			int xMargin = DPIUtil.scaleDown(x + LEFT_TEXT_MARGIN, zoom) + (this.image != null ? this.image.getBounds().width : 0);
```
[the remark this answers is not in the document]
HeikoKlare: In addition, it seems to be incorrect to scale down the predefined values, as they are the desired, consistent values in points, aren't they? With the current implementations, the margin are inconsistent across different scale values (I would say half the size on 200% compared to 100%). This applies to all scaleDown operations in this change including constant values in points.

Example 100%:
image

Example 175%:
image

Note that the _left_ margin on 175% is near the expected one while the one at 100% is way to high (compared with existing behavior / light theme). The overall margins seem to be too large in both cases.

HeikoKlare (commented):
The proposed change works really great now! It fixes the existing issues with scaling the menu bar in dark theme and it also improves the appearance of the menu bar in dark theme overall.

I only have one behavioral concern left, which is about the statically defined margins that (1) seem to be inappropriate and (2) are scaled incorrectly. The other comments are regarding code style / design and rather nitpicky.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MenuItem.java
```diff
@@ -1121,27 +1122,51 @@ LRESULT wmCommandChild (long wParam, long lParam) {
...
+				drawnText = this.text.replace("&", "");
+			} else {
+				drawnText = this.text;
+				flags |= SWT.DRAW_MNEMONIC;
+			}
+			Rectangle bounds = this.getBounds();
+
+			gc.setForeground(isInactive ? display.getSystemColor(SWT.COLOR_GRAY) : display.getSystemColor(SWT.COLOR_WHITE));
+			gc.setBackground(isSelected? display.getSystemColor(SWT.COLOR_DARK_GRAY) : parent.getBackground());
+			gc.fillRectangle(DPIUtil.scaleDown(x, zoom), DPIUtil.scaleDown(struct.top, zoom), DPIUtil.scaleDown(bounds.width + (image != null ? image.getBounds().width : 0), zoom), DPIUtil.scaleDown(bounds.height, zoom));
```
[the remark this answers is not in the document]
ShahzaibIbrahim: I have now refactored and assigned them to more self-explanatory variables and then passed menuItemArea for fillRectangle. As for `bounds.width` I have refactored its name `menuItemBounds `to be more clear.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MenuItem.java
ShahzaibIbrahim: This information still seems relevant on explaining the difference between `left margin` and `item size`. Also we are back to using value of 5 again to match the margins between items in menu bar. I would suggest to keep this comment as is.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MenuItem.java
ShahzaibIbrahim: I am going back to the value 5 for item width as before (because of larger gap) and left margin value was derived as a result of testing against different values and finding out the closest to the light theme.

HeikoKlare (commented):
Code quality improvements are overall fine. The behavioral concerns in the last review have not been addressed yet. Please check them again.

In addition, testing Snippet 373 shows a misplacement of the image: 
image
image

One minor finding that may be addressed independently as it is not part of this PR:
the image is not scaled correctly when using `rescaleOnRuntime`:
image

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MenuItem.java
```diff
@@ -1121,27 +1123,56 @@ LRESULT wmCommandChild (long wParam, long lParam) {
...
+	if ((text != null || image != null)) {
 		GCData data = new GCData();
 		data.device = display;
 		GC gc = createNewGC(struct.hDC, data);
-		/*
-		* Bug in Windows.  When a bitmap is included in the
-		* menu bar, the HDC seems to already include the left
-		* coordinate.  The fix is to ignore this value when
-		* the item is in a menu bar.
-		*/
```
HeikoKlare, line 1136: Just saw that this comment still seems to be relevant, so would be good to revert its removal.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MenuItem.java
```diff
@@ -1121,27 +1123,56 @@ LRESULT wmCommandChild (long wParam, long lParam) {
...
+				flags |= SWT.DRAW_MNEMONIC;
+			}
+			Rectangle menuItemBounds = this.getBounds();
+
+			int fillMenuWidth =  DPIUtil.scaleDown(menuItemBounds.width, zoom);
+			int fillMenuHeight = DPIUtil.scaleDown(menuItemBounds.height, zoom);
+			Rectangle menuItemArea = new Rectangle(DPIUtil.scaleDown(x, zoom), DPIUtil.scaleDown(struct.top, zoom), fillMenuWidth, fillMenuHeight);
+
+			gc.setForeground(isInactive ? display.getSystemColor(SWT.COLOR_GRAY) : display.getSystemColor(SWT.COLOR_WHITE));
+			gc.setBackground(isSelected? display.getSystemColor(SWT.COLOR_DARK_GRAY) : parent.getBackground());
```
HeikoKlare, line 1156:
```suggestion
			gc.setBackground(isSelected ? display.getSystemColor(SWT.COLOR_DARK_GRAY) : parent.getBackground());
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MenuItem.java
```diff
@@ -1121,27 +1123,56 @@ LRESULT wmCommandChild (long wParam, long lParam) {
...
+			gc.fillRectangle(menuItemArea);
+
+
```
HeikoKlare, line 1159: Double empty line seems unintended.
```suggestion

```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MenuItem.java
```diff
@@ -1121,27 +1123,56 @@ LRESULT wmCommandChild (long wParam, long lParam) {
...
+			int xPosiitonText = DPIUtil.scaleDown(x, zoom) + LEFT_TEXT_MARGIN + (this.image != null ? this.image.getBounds().width + ADDED_IMAGE_WIDTH : 0);
```
HeikoKlare, line 1160: Please fix the typo in `xPosiitonText`.

HeikoKlare (commented):
Looks good now, thank you! Only some comments on documentation, formatting, typo left.

Could you please squash the commits into a single one?

HeikoKlare (approved):
This look good now, thank you!

Build failures are unrelated. Still we need to wait for them to be fixed until we can merge.

HeikoKlare (commented):
Unfortunately, when doing another final testing after rebasing, I found a further issue. To reproduce, do the following:
* Have a primary monitor with >100% scaling (e.g., 125%), have a secondary monitor with 100% scaling
* Start an application with PerMonitorV2 DPI awareness in dark mode
* Move the application from 125% monitor (on which it looks fine) to 100% monitor.

The sizes seem to be calculated properly, but the font is too small (probably 100% instead of 125%):
{5A84A883-8FE7-4D82-A6BC-CFAE977C2ED5}

@akoch-yatta @ShahzaibIbrahim can you please have a look? Maybe this is easy to resolve by using proper scale values at some place.

HeikoKlare (commented):
Unfortunately, it looks still almost the same. On 125% it became better/good, but with the primary monitor on 150% or above, I still see the issue. E.g., this is with primary monitor 175% on secondary monitor with 100%:
{A059FDB7-CBCC-4CDD-9505-CFCA3476717E}

akoch-yatta:
> Unfortunately, it looks still almost the same. On 125% it became better/good, but with the primary monitor on 150% or above, I still see the issue. E.g., this is with primary monitor 175% on secondary monitor with 100%: {A059FDB7-CBCC-4CDD-9505-CFCA3476717E}

@HeikoKlare Sorry for all that back and forth, we need to improve the test protocol on our side, so you don't have that much work retesting changes. It is not yet properly working on all scenarios, as you mentioned. We'll have a look on that tomorrow.

akoch-yatta:
In each of the following screenshot there is an adapted version of Snippet373 started with commit [2fe01bf](https://github.com/eclipse-platform/eclipse.platform.swt/commit/2fe01bfb14ea7ceb50c029267a073a0b06ff3ff9), 4 times each. The title gives the info of the configuration:

1. autoscaleOnRuntime active or not
2. autoscale mode: int200 or quarter 

-------------------------------------------------
Setup: Primary 125%, secondary 175%
Primary:
image
Secondary:
image

-------------------------------------------------
Setup: Primary 175%, secondary 125%
Primary:
image
Secondary:
image

-------------------------------------------------
Setup: Primary 150%, secondary 200%
Primary:
image
Secondary:
image

-------------------------------------------------
Setup: Primary 200%, secondary 100%
Primary:
image
Secondary:
image

HeikoKlare:
Is this supposed to be ready for another review? I have only taken a quick look into an SDK application, but spacing is still off. See this screenshot at 150% (top one is with current implementation, bottom one with this PR):
{124AEFF1-853E-4512-8926-1BB6DCFECDCC}

akoch-yatta:
Hmm, I cannot reproduce any vertical spacing issue, neither with quarter nor with int200. 
100% 
image
150% 
image
175%
image
200%
image

Regarding the horizontal indentation: I agree, that int direct comparision positioning could be improved. I tweaked some statics in https://github.com/eclipse-platform/eclipse.platform.swt/pull/1405/commits/cd3bb9bd972b495995e1e4757f38db37d79afadc to achieve the following result
100%
image
150%
image
200%
image

fedejeanne (approved):
The changes look fine to me and I didn't find any regressions/bugs while testing. 

I only have 2 nit-picky comments regarding readability but aside from that, this PR is good to go on my account.

fedejeanne (approved):
@ShahzaibIbrahim please squash your commits and this one is good to go on my account. @HeikoKlare should I merge or do you want to take another look?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MenuItem.java
```diff
@@ -1134,14 +1161,50 @@ LRESULT wmDrawChild (long wParam, long lParam) {
...
+			int fillMenuWidth =  DPIUtil.scaleDown(menuItemBounds.width, zoom);
+			int fillMenuHeight = DPIUtil.scaleDown(menuItemBounds.height, zoom);
+			menuItemArea = new Rectangle(DPIUtil.scaleDown(x, zoom), DPIUtil.scaleDown(struct.top, zoom), fillMenuWidth, fillMenuHeight);
+
+			gc.setForeground(isInactive ? display.getSystemColor(SWT.COLOR_GRAY) : display.getSystemColor(SWT.COLOR_WHITE));
+			gc.setBackground(isSelected ? display.getSystemColor(SWT.COLOR_DARK_GRAY) : parent.getBackground());
+			gc.fillRectangle(menuItemArea);
+
+			int xPositionText = LEFT_TEXT_MARGIN + DPIUtil.scaleDown(x, zoom) + (this.image != null ? this.image.getBounds().width + IMAGE_TEXT_GAP : 0);
+			int yPositionText = DPIUtil.scaleDown(struct.top , zoom) + MARGIN_HEIGHT;
```
HeikoKlare, line 1192:
```suggestion
			int yPositionText = DPIUtil.scaleDown(struct.top, zoom) + MARGIN_HEIGHT;
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MenuItem.java
```diff
@@ -1134,14 +1161,50 @@ LRESULT wmDrawChild (long wParam, long lParam) {
...
+			gc.drawText(drawnText, xPositionText, yPositionText, flags);
+		}
+		if (image != null) {
+			Image image = getEnabled() ? this.image : new Image(display, this.image, SWT.IMAGE_DISABLE);
+			int gap = (menuItemArea.height - image.getBounds().height)/2;
```
HeikoKlare, line 1197:
```suggestion
			int gap = (menuItemArea.height - image.getBounds().height) / 2;
```

HeikoKlare (approved):
I've retested this and did not find any issues anymore. In particular with existing HiDPI support (i.e., without per-monitor rescaling at runtime) and all combinations of settings (integer200 or quarter scaling, DPI awareness set to PerMonitorV2 or System, two monitors with different scalings between 100% and 200%).

The general appearance is also better than with the existing implementation (placement/spacing and highlight color). There is still some room for slight improvements on vertical alignment (with 100% is slightly off to the top while at 175% it is slightly off to the bottom), but that's rather nitpicky and, as said, the proposed state is still far better than the existing one.

@ShahzaibIbrahim can you please check the code formatting again (see, e.g., the comments I made) and squash the commits, so that we can merge this for M2?

fedejeanne (approved):

HeikoKlare:
> Is there a shortage?

https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/5131

HeikoKlare:
Jenkins is still unavailable, but (Jenkins) builds before latest update were successful and without new issues. Since [only formatting changes have been performed since then](https://github.com/eclipse-platform/eclipse.platform.swt/compare/372338d1058469845da8f9610f5b569554ca45bd..885fc638794b03091866673603f3546be0dd2692), this is still safe to merge.

PR #1409, opened 2024-08-16
Extend Browser API to check for custom text location

amartya4256:
The contribution provides an API to check if the location is the location where custom texts are loaded in a browser. By construct, "about:blank" is used for that purpose. This API can be adapted by the clients instead of explicitly defining the string and checking.

Contributes to #213

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Browser/common/org/eclipse/swt/browser/Browser.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Browser/common/org/eclipse/swt/browser/Browser.java
@@ -13,6 +13,9 @@
  *******************************************************************************/
 package org.eclipse.swt.browser;
 
+import java.net.*;
+import java.util.*;
+
 import org.eclipse.swt.*;
 import org.eclipse.swt.widgets.*;
 
@@ -1185,6 +1188,21 @@ public boolean setUrl (String url, String postData, String[] headers) {
 	checkWidget();
 	if (url == null) SWT.error (SWT.ERROR_NULL_ARGUMENT);
 	return webBrowser.setUrl (url, postData, headers);
+}
+
+/**
+ * Checks if the location supplied is the location for setting custom text in the browser.
+ * This means, when passing the browser URL to this method after a custom text has been set
+ * via {@link #setText(String)} will yield true as long as no further navigation
+ * to some other location is performed.
+ *
+ * @param location the URL to be checked, must not be null
+ *
+ * @since 3.128
+ */
+public boolean isLocationForCustomText(String location) {
+	checkWidget();
+	return URI.create(Objects.requireNonNull(location)).equals(webBrowser.getUriForCustomText());
 }
 
 /**
--- a/bundles/org.eclipse.swt/Eclipse SWT Browser/common/org/eclipse/swt/browser/WebBrowser.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Browser/common/org/eclipse/swt/browser/WebBrowser.java
@@ -13,6 +13,7 @@
  *******************************************************************************/
 package org.eclipse.swt.browser;
 
+import java.net.*;
 import java.util.*;
 import java.util.List;
 
@@ -36,6 +37,7 @@ abstract class WebBrowser {
 
 	static final String ERROR_ID = "org.eclipse.swt.browser.error"; // $NON-NLS-1$
 	static final String EXECUTE_ID = "SWTExecuteTemporaryFunction"; // $NON-NLS-1$
+	private static final URI URI_FOR_CUSTOM_TEXT_PAGE = URI.create("about:blank");
 
 	static List<String[]> NativePendingCookies = new ArrayList<> ();
 	static String CookieName, CookieValue, CookieUrl;
@@ -722,6 +724,10 @@ boolean sendKeyEvent (Event event) {
 
 public void setBrowser (Browser browser) {
 	this.browser = browser;
+}
+
+URI getUriForCustomText() {
+	return URI_FOR_CUSTOM_TEXT_PAGE;
 }
 
 public abstract boolean setText (String html, boolean trusted);
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
@@ -34,6 +34,7 @@ import java.lang.management.ThreadInfo;
 import java.lang.management.ThreadMXBean;
 import java.net.HttpURLConnection;
 import java.net.MalformedURLException;
+import java.net.URI;
 import java.net.URL;
 import java.nio.file.DirectoryStream;
 import java.nio.file.Files;
@@ -516,6 +517,80 @@ public void test_LocationListener_addAndRemove() {
 	};
 	for (int i = 0; i < 100; i++) browser.addLocationListener(listener);
 	for (int i = 0; i < 100; i++) browser.removeLocationListener(listener);
+}
+
+@Test
+public void test_isLocationForCustomText_setUrlAfterDisposedThrowsSwtException() {
+	Browser testBrowser = createBrowser(shell, SWT.NONE);
+	testBrowser.dispose();
+	assertThrows(SWTException.class, () -> testBrowser.isLocationForCustomText("about:blank"));
+}
+
+@Test
+public void test_isLocationForCustomText_isSetUrlNotCustomTextUrlAfterSetText() {
+	assumeFalse("behavior is not (yet) supported by Edge browser", isEdge);	// Edge doesn't fire a changed event if the URL is the same as the previous location.
+	URI url = getValidUrl();
+	AtomicBoolean locationChanged = new AtomicBoolean(false);
+	browser.addLocationListener(changedAdapter(event -> {
+			locationChanged.set(true);
+	}));
+
+	browser.setText("Custom text");
+	assertTrue("Time Out: The Browser didn't navigate to the URL", waitForPassCondition(locationChanged::get));
+	locationChanged.set(false);
+	browser.setUrl(url.toASCIIString());
+	assertTrue("Time Out: The Browser didn't navigate to the URL", waitForPassCondition(locationChanged::get));
+	assertEquals("Url is wrongly considered Custom Text Url", URI.create(browser.getUrl()), url);
+	assertFalse("The navigated URL is falsly indicated to be the custom text URL", browser.isLocationForCustomText(browser.getUrl()));
+}
+
+@Test
+public void test_isLocationForCustomText_isFirstSetTextURLStillCustomTextUrlAfterSetUrl() {
+	assumeFalse("behavior is not (yet) supported by Edge browser", isEdge);	// Edge doesn't fire a changed event if the URL is the same as the previous location.
+	AtomicBoolean locationChanged = new AtomicBoolean(false);
+	browser.addLocationListener(changedAdapter(event -> locationChanged.set(true)));
+	URI url = getValidUrl();
+	browser.setText("Custom text");
+	assertTrue(waitForPassCondition(locationChanged::get));
+	String firstUrl = browser.getUrl();
+	locationChanged.set(false);
+	browser.setUrl(url.toASCIIString());
+	assertTrue("Time Out: The Browser didn't navigate to the URL", waitForPassCondition(locationChanged::get));
+	assertTrue(browser.isLocationForCustomText(firstUrl));
+	assertFalse(browser.isLocationForCustomText(browser.getUrl()));
+}
+
+private URI getValidUrl() {
+	String pluginPath = System.getProperty("PLUGIN_PATH");
+	URI url;
+	// Depending on how the jUnit test is ran, (gui/maven/ant), url for local file needs to be acquired differently.
+	if (pluginPath != null) {
+		url = URI.create(pluginPath + "/data/testWebsiteWithTitle.html");
+	} else {
+		// used when ran from Eclipse gui.
+		url = URI.create(Test_org_eclipse_swt_browser_Browser.class.getClassLoader().getResource("testWebsiteWithTitle.html").toString());
+	}
+	return url;
+}
+
+@Test
+public void test_isLocationForCustomText_isSetUrlNotCustomTextUrl() {
+	AtomicBoolean locationChanged = new AtomicBoolean(false);
+	browser.addLocationListener(changedAdapter(event -> locationChanged.set(true)));
+	URI url = getValidUrl();
+	browser.setUrl(url.toASCIIString());
+	waitForPassCondition(locationChanged::get);
+	assertFalse("Url is wrongly considered Custom Text Url", browser.isLocationForCustomText(browser.getUrl()));
+}
+
+@Test
+public void test_isLocationForCustomText() {
+	assumeFalse("behavior is not (yet) supported by Edge browser", isEdge);	// Edge doesn't fire a changed event if the URL is the same as the previous location.
+	AtomicBoolean locationChanged = new AtomicBoolean(false);
+	browser.addLocationListener(changedAdapter(e -> locationChanged.set(true)));
+	browser.setText("Hello world");
+	assertTrue("Timeout: LocationListener.changing() event was never fired", waitForPassCondition(locationChanged::get));
+	assertTrue("Custom Text URI was not loaded on setText", browser.isLocationForCustomText(browser.getUrl()));
 }
 
 @Test
--- a/binaries/org.eclipse.swt.cocoa.macosx.aarch64/META-INF/MANIFEST.MF
+++ b/binaries/org.eclipse.swt.cocoa.macosx.aarch64/META-INF/MANIFEST.MF
@@ -1,9 +1,9 @@
 Manifest-Version: 1.0
-Fragment-Host: org.eclipse.swt;bundle-version="[3.127.0,4.0.0)"
+Fragment-Host: org.eclipse.swt;bundle-version="[3.128.0,4.0.0)"
 Bundle-Name: %fragmentName
 Bundle-Vendor: %providerName
 Bundle-SymbolicName: org.eclipse.swt.cocoa.macosx.aarch64; singleton:=true
-Bundle-Version: 3.127.100.qualifier
+Bundle-Version: 3.128.0.qualifier
 Bundle-ManifestVersion: 2
 Bundle-Localization: fragment
 Export-Package: 
--- a/binaries/org.eclipse.swt.cocoa.macosx.x86_64/META-INF/MANIFEST.MF
+++ b/binaries/org.eclipse.swt.cocoa.macosx.x86_64/META-INF/MANIFEST.MF
@@ -1,9 +1,9 @@
 Manifest-Version: 1.0
-Fragment-Host: org.eclipse.swt;bundle-version="[3.127.0,4.0.0)"
+Fragment-Host: org.eclipse.swt;bundle-version="[3.128.0,4.0.0)"
 Bundle-Name: %fragmentName
 Bundle-Vendor: %providerName
 Bundle-SymbolicName: org.eclipse.swt.cocoa.macosx.x86_64; singleton:=true
-Bundle-Version: 3.127.100.qualifier
+Bundle-Version: 3.128.0.qualifier
 Bundle-ManifestVersion: 2
 Bundle-Localization: fragment
 Export-Package: 
--- a/binaries/org.eclipse.swt.gtk.linux.aarch64/META-INF/MANIFEST.MF
+++ b/binaries/org.eclipse.swt.gtk.linux.aarch64/META-INF/MANIFEST.MF
@@ -1,9 +1,9 @@
 Manifest-Version: 1.0
-Fragment-Host: org.eclipse.swt;bundle-version="[3.127.0,4.0.0)"
+Fragment-Host: org.eclipse.swt;bundle-version="[3.128.0,4.0.0)"
 Bundle-Name: %fragmentName
 Bundle-Vendor: %providerName
 Bundle-SymbolicName: org.eclipse.swt.gtk.linux.aarch64; singleton:=true
-Bundle-Version: 3.127.100.qualifier
+Bundle-Version: 3.128.0.qualifier
 Bundle-ManifestVersion: 2
 Bundle-Localization: fragment
 Export-Package: 
--- a/binaries/org.eclipse.swt.gtk.linux.loongarch64/META-INF/MANIFEST.MF
+++ b/binaries/org.eclipse.swt.gtk.linux.loongarch64/META-INF/MANIFEST.MF
@@ -1,9 +1,9 @@
 Manifest-Version: 1.0
-Fragment-Host: org.eclipse.swt;bundle-version="[3.127.0,4.0.0)"
+Fragment-Host: org.eclipse.swt;bundle-version="[3.128.0,4.0.0)"
 Bundle-Name: %fragmentName
 Bundle-Vendor: %providerName
 Bundle-SymbolicName: org.eclipse.swt.gtk.linux.loongarch64; singleton:=true
-Bundle-Version: 3.127.100.qualifier
+Bundle-Version: 3.128.0.qualifier
 Bundle-ManifestVersion: 2
 Bundle-Localization: fragment
 Export-Package: 
--- a/binaries/org.eclipse.swt.gtk.linux.ppc64le/META-INF/MANIFEST.MF
+++ b/binaries/org.eclipse.swt.gtk.linux.ppc64le/META-INF/MANIFEST.MF
@@ -1,9 +1,9 @@
 Manifest-Version: 1.0
-Fragment-Host: org.eclipse.swt;bundle-version="[3.127.0,4.0.0)"
+Fragment-Host: org.eclipse.swt;bundle-version="[3.128.0,4.0.0)"
 Bundle-Name: %fragmentName
 Bundle-Vendor: %providerName
 Bundle-SymbolicName: org.eclipse.swt.gtk.linux.ppc64le;singleton:=true
-Bundle-Version: 3.127.100.qualifier
+Bundle-Version: 3.128.0.qualifier
 Bundle-ManifestVersion: 2
 Bundle-Localization: fragment
 Export-Package: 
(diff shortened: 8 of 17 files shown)


HeikoKlare (commented):
Seeing the implementation, I am wondering whether it would also be sufficient to just add a `showsCustomText()` method instead of the `isLocationForCustomText(String)` one. Then the functionality would be completely independent from the location and the browser handle the state of whether it shows custom text or some URI-based page completely internally and even better hide it from consumers.
@amartya4256 Can you check whether that's possible?

Edit: Please also validate against the WebKit implementation. It seems to already have the about:blank page encoded as kind of a base URI internally. It also has some comments about using about:blank to be consistent with IE, so maybe there is also some further chance to align. I.e., a comment states: 
https://github.com/eclipse-platform/eclipse.platform.swt/blob/97cf720f1a64c04de12998b8a50cf57c7189c5f8/bundles/org.eclipse.swt/Eclipse%20SWT%20WebKit/cocoa/org/eclipse/swt/browser/WebKit.java#L634-L637

amartya4256:
> Seeing the implementation, I am wondering whether it would also be sufficient to just add a `showsCustomText()` method instead of the `isLocationForCustomText(String)` one. Then the functionality would be completely independent from the location and the browser handle the state of whether it shows custom text or some URI-based page completely internally and even better hide it from consumers. @amartya4256 Can you check whether that's possible?
> 
> Edit: Please also validate against the WebKit implementation. It seems to already have the about:blank page encoded as kind of a base URI internally. It also has some comments about using about:blank to be consistent with IE, so maybe there is also some further chance to align. I.e., a comment states:
> 
> https://github.com/eclipse-platform/eclipse.platform.swt/blob/97cf720f1a64c04de12998b8a50cf57c7189c5f8/bundles/org.eclipse.swt/Eclipse%20SWT%20WebKit/cocoa/org/eclipse/swt/browser/WebKit.java#L634-L637

using showsCustomText() won't be possible everywhere since there are Url comparison before the location of the browser has changed. For example in location listener in the changing method which executes before the location is actually changed. Another option was to cache the state type on setText but we don't know where we land with back and forward calls so it cannot be fully implemented. Moreover, I find the current method fine to accept a location as an argument. If you want to check something explictly, that can be passed, otherwise browser.getUrl() can be passed.

bundles/org.eclipse.swt/Eclipse SWT Browser/common/org/eclipse/swt/browser/Browser.java
```diff
@@ -1187,6 +1189,17 @@ public boolean setUrl (String url, String postData, String[] headers) {
 	return webBrowser.setUrl (url, postData, headers);
 }
 
+/**
+ * Checks if the location supplied is the location for setting custom text in the browser
```
HeikoKlare, line 1193: Would be good to extend the documentation a bit on what "setting custom text" means. Somethin like:
```suggestion
 * Checks if the location supplied is the location for setting custom text in the browser. This means, when passing the browser URL to this method after a custom text has been set via {@link #setText(String)} will yield true as long as no further navigation to some other location is performed.
```

bundles/org.eclipse.swt/Eclipse SWT Browser/common/org/eclipse/swt/browser/Browser.java
```diff
@@ -1187,6 +1189,17 @@ public boolean setUrl (String url, String postData, String[] headers) {
...
+ *
+ * @param location the URL to be checked
+ *
+ * @since 3.127
+ */
+public boolean isLocationForCustomText(String location) {
+	return URI.create(location).equals(webBrowser.getBaseUri());
```
HeikoKlare, line 1200: I propose to have a null check for `location` here. Either specify what happen when passing null (such as always returning false for a null value), or explicitly specifying in the Javadoc that null is not allowed and additionally explicitly checking the parameter to have an early failure:
```suggestion
	return URI.create(Objects.requireNonNull(location)).equals(webBrowser.getBaseUri());
```

bundles/org.eclipse.swt/Eclipse SWT Browser/common/org/eclipse/swt/browser/WebBrowser.java
```diff
@@ -724,6 +726,10 @@ public void setBrowser (Browser browser) {
 	this.browser = browser;
 }
 
+public URI getBaseUri() {
```
HeikoKlare, line 729: Can this be package protected, since the only intended consumer (`Browser´) is in the same package? Please also improve the name according to field name (see comment on the field name for that).

In addition, please check whether this method can be used by the IE and WebKit implementations, which currently have their own `ABOUT_BLANK` field, which might (at least at some places) be reasonably replaced by this method.

bundles/org.eclipse.swt/Eclipse SWT Browser/common/org/eclipse/swt/browser/WebBrowser.java
```diff
@@ -36,6 +37,7 @@ abstract class WebBrowser {
 
 	static final String ERROR_ID = "org.eclipse.swt.browser.error"; // $NON-NLS-1$
 	static final String EXECUTE_ID = "SWTExecuteTemporaryFunction"; // $NON-NLS-1$
+	private static final URI BASE_URI = URI.create("about:blank");
```
HeikoKlare, line 40: Please change this to a more self-explaining name, such as `URI_FOR_CUSTOM_TEXT_PAGE`.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
```diff
@@ -518,6 +518,18 @@ public void changing(LocationEvent event) {
...
+
+
+@Test
+public void test_isLocationForCustomText() {
+	AtomicBoolean changingFired = new AtomicBoolean(false);
+	browser.addLocationListener(changingAdapter(e -> changingFired.set(true)));
+	shell.open();
+	browser.setText("Hello world");
+	waitForPassCondition(changingFired::get);
+	assertTrue("LocationListener.changing() event was never fired", browser.isLocationForCustomText(browser.getUrl()));
```
HeikoKlare, line 530: The assertion message is unrelated to what the assertion actually checks. I propose to make two assertions out of this:
1. one assertion checking that `LocationListener.changing()` has fired
2. one assertion checking the return value of the new method.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
```diff
@@ -518,6 +518,18 @@ public void changing(LocationEvent event) {
 	for (int i = 0; i < 100; i++) browser.removeLocationListener(listener);
 }
 
+
+
+@Test
+public void test_isLocationForCustomText() {
```
HeikoKlare, line 524: Please extend the tests to not only cover the happy path. I consider at least the following scenarios relevant (maybe there are even more):
- check that location after initialization is not the location for custom text
- check that location after moving to some "actual" URL is not considered a location for custom text
- check that when calling `setText()` first and then navigating to some other page via `setUrl()` results in the location not being considered the location for custom text
- check that when calling `setText()` twice and using the browser Url after the first call of `setText()` after the second call of `setText()` to check whether it is a location for custom text still yields true (this is for the case that some asynchronous navigation is performed while some consumer still wants to check the previous URL)
- validate what happens if calling that method after the browser has been disposed

HeikoKlare (commented):
The API extension in general looks good to me. I have some comments on naming, documentation and testing.

Please also adapt the commit message header (and PR name) to be more self-explaining. I.e., instead of "Added API for checking if BASE_URI" use something like "Extend Browser API to check for custom text location"

bundles/org.eclipse.swt/Eclipse SWT Browser/common/org/eclipse/swt/browser/Browser.java
amartya4256: Done

bundles/org.eclipse.swt/Eclipse SWT Browser/common/org/eclipse/swt/browser/Browser.java
amartya4256: Done

bundles/org.eclipse.swt/Eclipse SWT Browser/common/org/eclipse/swt/browser/WebBrowser.java
amartya4256: Made it package protected but I think refactoring of the browser specific usage can be done in a follow up PR?

bundles/org.eclipse.swt/Eclipse SWT Browser/common/org/eclipse/swt/browser/WebBrowser.java
amartya4256: Done

bundles/org.eclipse.swt/Eclipse SWT Browser/common/org/eclipse/swt/browser/Browser.java
HeikoKlare: Nothing changed in the specification, only the null check was added to the code.

bundles/org.eclipse.swt/Eclipse SWT Browser/common/org/eclipse/swt/browser/WebBrowser.java
HeikoKlare: That's also fine. But then it should be done directly after merging this change.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
HeikoKlare: I see, So probably the first proposed test case does not make any sense. No need to add a different kind of test for that. This should only be about testing the contract of the new `isLocationForCustomText()` method.

bundles/org.eclipse.swt/Eclipse SWT Browser/common/org/eclipse/swt/browser/Browser.java
amartya4256: Added in the parameter specification

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
amartya4256: Added the tests except for the 4th point. I don't get it. Can you explain a bit?

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
```diff
@@ -518,6 +519,98 @@ public void changing(LocationEvent event) {
...
+		url = URI.create(Test_org_eclipse_swt_browser_Browser.class.getClassLoader().getResource("testWebsiteWithTitle.html").toString());
+	}
+	AtomicBoolean changed = new AtomicBoolean(false);
+	browser.addLocationListener(changedAdapter(event -> {
+		if (URI.create(event.location).equals(url)) {
+			changed.set(true);
+		}
+	}));
+
+	browser.setText("Custom text");
```
HeikoKlare, line 548: Since this operation is asynchronous, you should wait for it to complete before calling `setUrl()`. Currently there is a race condition, as `setText()` may be completely processed before `setUrl()` is executed or not, thus potentially resulting in flaky test results in case of a regression.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
```diff
@@ -518,6 +519,98 @@ public void changing(LocationEvent event) {
 	for (int i = 0; i < 100; i++) browser.removeLocationListener(listener);
 }
 
+@Test
+public void test_setUrlAfterDisposed() {
+	Browser testBrowser = createBrowser(shell, SWT.NONE);
+	testBrowser.dispose();
+	assertThrows(SWTException.class, () -> testBrowser.setUrl("about:blank"));
```
HeikoKlare, line 526: This test is unrelated to the PR, isn't it? There should be a test for behavior of `isLocationForCustomURL()` after disposal instead/in addition.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
```diff
@@ -518,6 +519,98 @@ public void changing(LocationEvent event) {
...
+			changed.set(true);
+		}
+	}));
+
+	browser.setText("Custom text");
+	browser.setUrl(url.toASCIIString());
+	shell.open();
+	assertTrue("Time Out: The Browser didn't navigate to the URL", waitForPassCondition(changed::get));
+	assertTrue("Url is wrongly considered Custom Text Url", URI.create(browser.getUrl()).equals(url)
+			&& !browser.isLocationForCustomText(browser.getUrl()));
```
HeikoKlare, line 553: Please make two assertions out of this, to identify which of the calls leads to a failure. The URI comparison can even be replaced with an `assertEquals` to get better output in case of a failure.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
```diff
@@ -518,6 +519,98 @@ public void changing(LocationEvent event) {
...
 
+@Test
+public void test_setUrlAfterDisposed() {
+	Browser testBrowser = createBrowser(shell, SWT.NONE);
+	testBrowser.dispose();
+	assertThrows(SWTException.class, () -> testBrowser.setUrl("about:blank"));
+}
+
+@Test
+public void test_isSetUrlNotCustomTextUrlAfterSetText() {
```
HeikoKlare, line 530: Since all the added tests are dedicated to test behavior of `isLocationForCustomText()`, it would be good to have the test names always start with the tested method and append the use case to that. E.g., `test_isLocationForCustomText_falseAfterSetUrl`.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
```diff
@@ -518,6 +519,98 @@ public void changing(LocationEvent event) {
...
+	testLogAppend("PLUGIN_PATH: " + pluginPath);
+	URI url;
+	// Depending on how the jUnit test is ran, (gui/maven/ant), url for local file needs to be acquired differently.
+	if (pluginPath != null) {
+		url = URI.create(pluginPath + "/data/testWebsiteWithTitle.html");
+	} else {
+		// used when ran from Eclipse gui.
+		url = URI.create(Test_org_eclipse_swt_browser_Browser.class.getClassLoader().getResource("testWebsiteWithTitle.html").toString());
+	}
+	AtomicBoolean changed = new AtomicBoolean(false);
```
HeikoKlare, line 541: Would be nice to give these guard variables a more self-explaining name like `locationChanged`.

HeikoKlare (commented):
Thank you for adding the tests! In general, they look fine. My comments are mostly nitpicky style proposals, but some are also on potential race conditions. Anyway, all of them should be easy to resolve.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
```diff
@@ -518,6 +519,98 @@ public void changing(LocationEvent event) {
...
+public void test_setUrlAfterDisposed() {
+	Browser testBrowser = createBrowser(shell, SWT.NONE);
+	testBrowser.dispose();
+	assertThrows(SWTException.class, () -> testBrowser.setUrl("about:blank"));
+}
+
+@Test
+public void test_isSetUrlNotCustomTextUrlAfterSetText() {
+	String pluginPath = System.getProperty("PLUGIN_PATH");
+	testLogAppend("PLUGIN_PATH: " + pluginPath);
```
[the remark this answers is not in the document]
amartya4256: Its for debugging.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
amartya4256: Done

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
```diff
@@ -518,6 +519,98 @@ public void changing(LocationEvent event) {
...
+		url = URI.create(Test_org_eclipse_swt_browser_Browser.class.getClassLoader().getResource("testWebsiteWithTitle.html").toString());
+	}
+	shell.open();
+	browser.setText("Custom text");
+	waitForPassCondition(changed::get);
+	String firstUrl = browser.getUrl();
+	changed.set(false);
+	browser.setUrl(url.toASCIIString());
+	waitForPassCondition(changed::get);
+	assertTrue(browser.isLocationForCustomText(firstUrl));
```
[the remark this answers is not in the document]
amartya4256: Done

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
amartya4256: Done

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
```diff
@@ -518,6 +519,98 @@ public void changing(LocationEvent event) {
...
+	// Depending on how the jUnit test is ran, (gui/maven/ant), url for local file needs to be acquired differently.
+	if (pluginPath != null) {
+		url = URI.create(pluginPath + "/data/testWebsiteWithTitle.html");
+	} else {
+		// used when ran from Eclipse gui.
+		url = URI.create(Test_org_eclipse_swt_browser_Browser.class.getClassLoader().getResource("testWebsiteWithTitle.html").toString());
+	}
+	shell.open();
+	browser.setText("Custom text");
+	waitForPassCondition(changed::get);
```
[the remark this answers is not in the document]
amartya4256: Done

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
amartya4256: Done

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
amartya4256: Done

bundles/org.eclipse.swt/Eclipse SWT Browser/common/org/eclipse/swt/browser/Browser.java
```diff
@@ -1187,6 +1190,20 @@ public boolean setUrl (String url, String postData, String[] headers) {
 	return webBrowser.setUrl (url, postData, headers);
 }
 
+/**
+ * Checks if the location supplied is the location for setting custom text in the browser.
+ * This means, when passing the browser URL to this method after a custom text has been set
+ * via {@link #setText(String)} will yield true as long as no further navigation
+ * to some other location is performed.
+ *
+ * @param location the URL to be checked. Null values are not allowed and would lead to a NullPointerException.
```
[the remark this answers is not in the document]
amartya4256: Done

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
```diff
@@ -518,6 +519,98 @@ public void changing(LocationEvent event) {
...
+	String pluginPath = System.getProperty("PLUGIN_PATH");
+	testLogAppend("PLUGIN_PATH: " + pluginPath);
+	URI url;
+	// Depending on how the jUnit test is ran, (gui/maven/ant), url for local file needs to be acquired differently.
+	if (pluginPath != null) {
+		url = URI.create(pluginPath + "/data/testWebsiteWithTitle.html");
+	} else {
+		// used when ran from Eclipse gui.
+		url = URI.create(Test_org_eclipse_swt_browser_Browser.class.getClassLoader().getResource("testWebsiteWithTitle.html").toString());
+	}
```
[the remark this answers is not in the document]
amartya4256: Done

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
```diff
@@ -518,6 +519,86 @@ public void changing(LocationEvent event) {
...
+	AtomicBoolean locationChanged = new AtomicBoolean(false);
+	browser.addLocationListener(changedAdapter(event -> {
+			locationChanged.set(true);
+	}));
+
+	browser.setText("Custom text");
+	assertTrue("Time Out: The Browser didn't navigate to the URL", waitForPassCondition(locationChanged::get));
+	locationChanged.set(false);
+	browser.setUrl(url.toASCIIString());
+	shell.open();
```
HeikoKlare, line 542: Just wondering: it does not seem to matter when (or even it at all) the shell is opened. Still is seems weird that it is opened after all operations have already been executed. Maybe the shell should be opened at the beginning of the test to avoid confusion (at least it should be done at a consistent point in time throughout the tests)? Or is there a reason for doing it here?

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
```diff
@@ -518,6 +519,86 @@ public void changing(LocationEvent event) {
...
+	locationChanged.set(false);
+	browser.setUrl(url.toASCIIString());
+	assertTrue("Time Out: The Browser didn't navigate to the URL", waitForPassCondition(locationChanged::get));
+	assertTrue(browser.isLocationForCustomText(firstUrl));
+	assertFalse(browser.isLocationForCustomText(browser.getUrl()));
+}
+
+private URI getValidUrl() {
+	String pluginPath = System.getProperty("PLUGIN_PATH");
+	testLogAppend("PLUGIN_PATH: " + pluginPath);
```
HeikoKlare, line 568: I still do not understand the idea of this logging. The `testLogAppend()` seems to be used for writing to a local log that is evaluated only at very specifiy places (e.g. `validateTitleChanged()`). Is the value ever used in the related test cases? Otherwise, e.g., if this is only for local debugging purposes by manually inspecting the written variable in a debugger, you may also just set a breakpoint here and take a look at the original variable (`pluginPath`).
Don't get me wrong: writing to this log does not really hurt, but it leaves future maintainers future with the question what the intent of executing this operation was (as it not obviously useful).

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
```diff
@@ -1043,6 +1124,7 @@ public void test_setText() {
 		assertTrue("Expecting setText() to return true", opSuccess);
 	};
 	validateTitleChanged(expectedTitle, browserSetFunc);
+	assertTrue("Browser didn't navigate to the implicit custom text location", browser.isLocationForCustomText(browser.getUrl()));
```
HeikoKlare, line 1127: This assertion seems misplaced. The test is about validating the behavior of `setText()`, now it might fail because `isLocationForCustomText()` is buggy. So if this is a relevant scenario that is not already covered by the `test_isLocationForCustomText_*()` method, please add an according one instead of mixing up concerns here.

HeikoKlare (commented):
Thanks again for the recent update! I have some minor questions/comments left.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
amartya4256: It has been tested for setText in many isLocationForCustomText tests. So, I'll just remove this here.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
amartya4256: I just used one of the fragments which was generating the URL in one of the tests. To me, it has no use. So I can also remove it.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
amartya4256: It doesnt matter at all haha. Te tests even work without opening the shell. It's just for the visibility's sake but I think, if we don't open the shell, the tests run faster. Also we don't have those blinking windows on the monitor for the time the tests are running.

amartya4256: Removed the shell.open calls from new tests

HeikoKlare (approved):
From my side, this is ready to be merged once master branch is open again. Thank you!

PR #1418, opened 2024-08-21
Edge: Fix focus handling to emit SWT.FocusIn events

### LINKED ISSUE #1139: Browser control with SWT.EDGE does not emit Activate/Decactivate events and does not activate parts in an RCP application

### LINKED ISSUE #1417: Edge: Focus event is not correctly propagated
Consider the following situation in a Eclipse IDE:

- Set Edge as the SWT browser engine
- Open the 'Internal Web Browser' view and any other view, e.g. editor, next to each other.
- Click inside the editor
- Check (e.g. using PDE spies, Alt+Shift+F1) that the active part is the Java editor:
  image
- Now, click inside the 'Internal Web Browser'
- Check again using the spies.
  Expected: the active part should be the `WebBrowserView`, but it still is the editor:
  image

sratz:
Follow-up on #1187.
Fixes #1139.
Fixes #1417.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
@@ -63,6 +63,7 @@ class Edge extends WebBrowser {
 	static boolean inCallback;
 	boolean inNewWindow;
 	HashMap<Long, LocationEvent> navigations = new HashMap<>();
+	private boolean ignoreFocus;
 
 	static {
 		NativeClearSessions = () -> {
@@ -480,6 +481,7 @@ void browserDispose(Event event) {
 }
 
 void browserFocusIn(Event event) {
+	if (ignoreFocus) return;
 	// TODO: directional traversals
 	controller.MoveFocus(COM.COREWEBVIEW2_MOVE_FOCUS_REASON_PROGRAMMATIC);
 }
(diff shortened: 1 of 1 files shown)


HeikoKlare (approved):
Great to see this issue fixed. Thank you, @sratz!

I have tested the fix, in particular also with the scenario/reproduction described in https://github.com/eclipse-platform/eclipse.platform.swt/issues/1139, and found the solution to work fine.

I just wonder whether we could have a regression test for this, such as emulating a mouse event to set focus and check whether focus was properly set afterwards? So that we are able to identify potential regressions w.r.t. this behavior (even though browser tests are currently not executed for Edge, but hopefully we can soon enable them when the fixes currently in the pipeline are merged).

Maybe @N1k145 can also verify the fix, as they also faced the issue?

N1k145:
@HeikoKlare Works like a charm.
@sratz Thank you very much for the fix

PR #1421, opened 2024-08-26
[GTK4] Migrate DirectoryDialog from GtkFileChooser to GtkFileDialog

ptziegler:
The FileChooser has been marked as deprecated with GTK 4.10 with the FileDialog being its replacement. Using this new API poses a challenge, as it requires the usage of the AsyncReadyCallback mechanism in order to respond to the closure of the dialog.

Once the dialog has been opened via gtk_file_dialog_select_folder(), it is necessary to call gtk_file_dialog_select_folder_finish() from within the callback method. This method also returns the selected folder, which has to be cached until after the dialog has been closed.

Following native methods have been added to GTK4:
- gtk_file_dialog_new
- gtk_file_dialog_select_folder
- gtk_file_dialog_select_folder_finish
- gtk_file_dialog_set_initial_folder

PR #1422, opened 2024-08-27
Build GTK 4 binaries

laeubi:
Currently the GTK4 binaries are not build as part of the build, this has several drawbacks:

1) If anything is adjusted for GTK4 part it might break without notice 2) We have no GTK4 binaries by default

This enables *compilation* of the gtk4 parts to see at least everything can be compiled.

The following has to be checked:

- [x] do we have everything to compile for Jenkins Pod
- [x] do we have everything to compile for Github
- [ ] ...

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -19,7 +19,7 @@ def runOnNativeBuildAgent(String platform, Closure body) {
 	def final nativeBuildStageName = 'Build SWT-native binaries'
 	if (platform == 'gtk.linux.x86_64') {
 		podTemplate(inheritFrom: 'ubuntu-latest' /* inhert general configuration */, containers: [
-			containerTemplate(name: 'swtbuild', image: 'eclipse/platformreleng-centos-swt-build:8',
+			containerTemplate(name: 'swtbuild', image: 'eclipse/platformreleng-debian-swtnativebuild:12',
 				resourceRequestCpu:'1000m', resourceRequestMemory:'512Mi',
 				resourceLimitCpu:'2000m', resourceLimitMemory:'4096Mi',
 				alwaysPullImage: true, command: 'cat', ttyEnabled: true)
@@ -228,7 +228,15 @@ pipeline {
 										if (isUnix()){
 											sh '''
 												mkdir libs
-												sh build.sh install
+												if [ "${PLATFORM}" = "gtk.linux.aarch64" ]; then
+													sh build.sh -gtk3 install
+												elif [ "${PLATFORM}" = "gtk.linux.ppc64le" ]; then
+													sh build.sh -gtk3 install
+												elif [ "${PLATFORM}" = "gtk.linux.riscv64" ]; then
+													sh build.sh -gtk3 install
+												else
+													sh build.sh install 
+												fi
 												ls -1R libs
 											'''
 										} else {
--- a/binaries/pom.xml
+++ b/binaries/pom.xml
@@ -135,12 +135,10 @@
 														</exec>
 													</then>
 													<else>
-														<property name="gtk_version" value="3.0" />
 														<exec dir="${build_dir}" executable="sh" failonerror="true">
 															<arg line="build.sh"/>
 															<env key="SWT_JAVA_HOME" value="${SWT_JAVA_HOME}"/>
 															<env key="OUTPUT_DIR" value="${project.basedir}"/>
-															<env key="GTK_VERSION" value="${gtk_version}"/>
 															<env key="MODEL" value="${arch}"/>
 															<arg line="install clean"/>
 														</exec>
--- a/bundles/org.eclipse.swt/Eclipse SWT PI/gtk/library/build.sh
+++ b/bundles/org.eclipse.swt/Eclipse SWT PI/gtk/library/build.sh
@@ -280,14 +280,7 @@ func_build_gtk3 () {
 	fi
 }
 
-if [ "$1" = "-gtk-all" ]; then
-	shift
-	func_echo_plus "Note: When building multiple GTK versions, a cleanup is required (and automatically performed) between them."
-	func_clean_up
-	func_build_gtk4 "$@"
-	func_clean_up
-	func_build_gtk3 "$@"
-elif [ "$1" = "-gtk4" ]; then
+if [ "$1" = "-gtk4" ]; then
 	shift
 	func_build_gtk4 "$@"
 elif [ "$1" = "-gtk3" ]; then
@@ -295,7 +288,17 @@ elif [ "$1" = "-gtk3" ]; then
 	func_build_gtk3 "$@"
 elif [ "${GTK_VERSION}" = "4.0" ]; then
 	func_build_gtk4 "$@"
-elif [ "${GTK_VERSION}" = "3.0" -o "${GTK_VERSION}" = "" ]; then
-	export GTK_VERSION="3.0"
+elif [ "${GTK_VERSION}" = "3.0" ]; then
+	func_build_gtk3 "$@"
+else
+	if [ "$1" = "-gtk-all" ]; then
+		shift
+	fi
+	func_echo_plus "==== Building GTK3 + GTK 4 ===="
+	func_echo_plus "Note: When building multiple GTK versions, a cleanup is required (and automatically performed) between them."
+	func_clean_up
+	func_build_gtk4 "$@"
+	func_echo_plus "Note: When building multiple GTK versions, a cleanup is required (and automatically performed) between them."
+	func_clean_up
 	func_build_gtk3 "$@"
 fi


laeubi:
Linux already builds sucessufll on GH, but Mac complains about

```
/Applications/Xcode_15.2.app/Contents/Developer/usr/bin/make: invalid option -- g
/Applications/Xcode_15.2.app/Contents/Developer/usr/bin/make: invalid option -- a
Usage: make [options] [target] ...
Options:
  -b, -m                      Ignored for compatibility.
  -B, --always-make           Unconditionally make all targets.
  -C DIRECTORY, --directory=DIRECTORY
                              Change to DIRECTORY before doing anything.
  -d                          Print lots of debugging information.
  --debug[=FLAGS]             Print various types of debugging information.
  -e, --environment-overrides
                              Environment variables override makefiles.
  -f FILE, --file=FILE, --makefile=FILE
                              Read FILE as a makefile.
  -h, --help                  Print this message and exit.
  -i, --ignore-errors         Ignore errors from commands.
  -I DIRECTORY, --include-dir=DIRECTORY
                              Search DIRECTORY for included makefiles.
  -j [N], --jobs[=N]          Allow N jobs at once; infinite jobs with no arg.
  -k, --keep-going            Keep going when some targets can't be made.
  -l [N], --load-average[=N], --max-load[=N]
                              Don't start multiple jobs unless load is below N.
  -L, --check-symlink-times   Use the latest mtime between symlinks and target.
  -n, --just-print, --dry-run, --recon
                              Don't actually run any commands; just print them.
  -o FILE, --old-file=FILE, --assume-old=FILE
                              Consider FILE to be very old and don't remake it.
  -p, --print-data-base       Print make's internal database.
  -q, --question              Run no commands; exit status says if up to date.
  -r, --no-builtin-rules      Disable the built-in implicit rules.
  -R, --no-builtin-variables  Disable the built-in variable settings.
  -s, --silent, --quiet       Don't echo commands.
  -S, --no-keep-going, --stop
                              Turns off -k.
  -t, --touch                 Touch targets instead of remaking them.
  -v, --version               Print the version number of make and exit.
  -w, --print-directory       Print the current directory.
  --no-print-directory        Turn off -w, even if it was turned on implicitly.
  -W FILE, --what-if=FILE, --new-file=FILE, --assume-new=FILE
                              Consider FILE to be infinitely new.
  --warn-undefined-variables  Warn when an undefined variable is referenced.
  -N OPTION, --NeXT-option=OPTION
                              Turn on value of NeXT OPTION.

This program built for i386-apple-darwin11.3.0
Report bugs to <bug-make@gnu.org>
```

bundles/org.eclipse.swt/Eclipse SWT PI/gtk/library/gtk3.h
```diff
@@ -10,6 +10,7 @@
  *
  * Contributors:
  *     Syntevo - initial API and implementation
+ * DUMMY CHANGE
```
laeubi, line 13: Needs to be reverted before merge

laeubi:
Now Github succeed and Jenkins fails as expected:

```
11:31:19  gcc -shared -fPIC  -s -o libswt-cairo-gtk-4966r6.so swt.o cairo.o cairo_structs.o cairo_stats.o `pkg-config --libs-only-L cairo` -lcairo

11:31:19  gcc -O -Wall -fPIC -DSWT_VERSION=4966r6   -DLINUX -DGTK -I/home/jenkins/agent/workspace/eclipse.platform.swt_PR-1422/jdk.resources/include -I/home/jenkins/agent/workspace/eclipse.platform.swt_PR-1422/jdk.resources/include/linux -DJNI64 -Werror   -c -o swt_awt.o swt_awt.c

11:31:19  gcc -shared   -s -o libswt-awt-gtk-4966r6.so swt_awt.o -L/home/jenkins/agent/workspace/eclipse.platform.swt_PR-1422/jdk.resources/lib -ljawt

11:31:19  cp libswt-gtk-4966r6.so libswt-awt-gtk-4966r6.so libswt-pi3-gtk-4966r6.so libswt-cairo-gtk-4966r6.so libswt-atk-gtk-4966r6.so libswt-glx-gtk-4966r6.so libswt-webkit-gtk-4966r6.so /home/jenkins/agent/workspace/eclipse.platform.swt_PR-1422/libs

11:31:19  
GTK3 Build succeeded

11:31:19  
Note: When building multiple GTK versions, a cleanup is required (and automatically performed) between them.

11:31:19  
Cleaning up...

11:31:19  rm -f *.o *.so

11:31:19  
Building GTK4 bindings:

11:31:19  gcc -O -Wall -fPIC -DSWT_VERSION=4966r6   -DLINUX -DGTK -I/home/jenkins/agent/workspace/eclipse.platform.swt_PR-1422/jdk.resources/include -I/home/jenkins/agent/workspace/eclipse.platform.swt_PR-1422/jdk.resources/include/linux -DJNI64 -Werror -c swt.c

11:31:19  gcc -O -Wall -fPIC -DSWT_VERSION=4966r6   -DLINUX -DGTK -I/home/jenkins/agent/workspace/eclipse.platform.swt_PR-1422/jdk.resources/include -I/home/jenkins/agent/workspace/eclipse.platform.swt_PR-1422/jdk.resources/include/linux -DJNI64 -Werror   -c -o c.o c.c

11:31:19  gcc -O -Wall -fPIC -DSWT_VERSION=4966r6   -DLINUX -DGTK -I/home/jenkins/agent/workspace/eclipse.platform.swt_PR-1422/jdk.resources/include -I/home/jenkins/agent/workspace/eclipse.platform.swt_PR-1422/jdk.resources/include/linux -DJNI64 -Werror   -c -o c_stats.o c_stats.c

11:31:19  gcc -O -Wall -fPIC -DSWT_VERSION=4966r6   -DLINUX -DGTK -I/home/jenkins/agent/workspace/eclipse.platform.swt_PR-1422/jdk.resources/include -I/home/jenkins/agent/workspace/eclipse.platform.swt_PR-1422/jdk.resources/include/linux -DJNI64 -Werror `pkg-config --cflags gtk4 gtk4-x11 gtk4-unix-print` -DUSE_ASSEMBLER -c callback.c

11:31:19  Package gtk4 was not found in the pkg-config search path.

11:31:19  Perhaps you should add the directory containing `gtk4.pc'

11:31:19  to the PKG_CONFIG_PATH environment variable

11:31:19  Package 'gtk4', required by 'virtual:world', not found

11:31:19  Package 'gtk4-x11', required by 'virtual:world', not found

11:31:19  Package 'gtk4-unix-print', required by 'virtual:world', not found

11:31:23  gcc -shared -fPIC  -s -o libswt-gtk-4966r6.so swt.o c.o c_stats.o callback.o

11:31:23  gcc -O -Wall -fPIC -DSWT_VERSION=4966r6   -DLINUX -DGTK -I/home/jenkins/agent/workspace/eclipse.platform.swt_PR-1422/jdk.resources/include -I/home/jenkins/agent/workspace/eclipse.platform.swt_PR-1422/jdk.resources/include/linux -DJNI64 -Werror `pkg-config --cflags gtk4 gtk4-x11 gtk4-unix-print` -c os.c

11:31:23  Package gtk4 was not found in the pkg-config search path.

11:31:23  Perhaps you should add the directory containing `gtk4.pc'

11:31:23  to the PKG_CONFIG_PATH environment variable

11:31:23  Package 'gtk4', required by 'virtual:world', not found

11:31:23  Package 'gtk4-x11', required by 'virtual:world', not found

11:31:23  Package 'gtk4-unix-print', required by 'virtual:world', not found

11:31:23  In file included from os_structs.h:19,

11:31:23                   from os.c:20:

11:31:23  os.h:30:10: fatal error: gtk/gtk.h: No such file or directory

11:31:23   #include <gtk/gtk.h>

11:31:23            ^~~~~~~~~~~

11:31:23  compilation terminated.

11:31:23  make: *** [make_linux.mak:144: os.o] Error 1

11:31:23  
*** GTK4 Build failed, aborting further actions..
```

ptziegler:
Given that I wanted to have a look at some of those deprecation errors in the near future, what version of GTK4 do you plan on compiling against?

With https://github.com/eclipse-platform/eclipse.platform.swt/pull/1421, SWT would require at least require 4.10 but then the build would have to continue despite the deprecations. With older versions, the build would succeed but then the PR no longer builds as the new methods don't exist yet.

akurtakov:
As GTK4 is WIP it should be fine to require as new version as needed . By the time this is ready for prime time that version would have made it to mainstream distros.

laeubi:
Main goal is to build it with "what is available right now", if we have this working one can go step further to support a specific version. But still then one need to take care as using "the latest" might simply make it impossible for some linux distributions (e.g LTE) that not always have latest and greatest of libraries, but "latest stable" version.

laeubi:
Even though GTK4 was added here:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2259/files

and container build seem to have completed

https://ci.eclipse.org/releng/view/Miscellaneous/job/Build-Docker-images/732/

the build still complains:

> Package gtk4 was not found in the pkg-config search path.

@fredg02 anything else that needs to be done so the new container is visible? Can we do anything to verify the change is actually used? I didn't find anything in the log that indicates the actual used container has/version/date/....

iloveeclipse:
> Main goal is to build it with "what is available right now", if we have this working one can go step further to support a specific version.

We should have target Linux versions specified in the [project plan](https://www.eclipse.org/projects/project-plan.php?planurl=https://www.eclipse.org/eclipse/development/plans/eclipse_project_plan_4_33.xml), similar to Java versions. 
- RHEL 9
- SUSE 15 SP4
- Ubuntu 24.04
The minimal GTK4 version could be derived from there.

akurtakov:
> > Main goal is to build it with "what is available right now", if we have this working one can go step further to support a specific version.
> 
> We should have target Linux versions specified in the [project plan](https://www.eclipse.org/projects/project-plan.php?planurl=https://www.eclipse.org/eclipse/development/plans/eclipse_project_plan_4_33.xml), similar to Java versions.
> 
>     * RHEL 9
> 
>     * SUSE 15 SP4
> 
>     * Ubuntu 24.04
>       The minimal GTK4 version could be derived from there.

I would agree with such version targeting when we discuss making GTK 4 default one. Unfortunately we are very far from it so by the time this discussion comes we will most likely have RHEL 10(most likely 11) and Ubuntu 26.04. 
Speaking from history - when port to GTK 3 was done we had to give feedback to gtk developers, report bugs, test patches and several times target even development versions in order to manage to complete SWT on GTK3 port. Setting targets based on what current LTS distros ships will most likely make the port far more harder than it should be (and some SWT features even impossible).

akurtakov:
One such impossible feature on RHEL 9 would be Browser support as even though there is gtk4 there is no webkitgtk4.

laeubi:
That's why I said lets first get *something* compiled, then its easier to iterate, I tried to add some commands to find out actual installed versions for the github runners it is this versions:

```
libgtk-3-0/jammy-updates,jammy-security,now 3.24.33-1ubuntu2.2 amd64 [installed]
libgtk-3-common/jammy-updates,jammy-security,now 3.24.33-1ubuntu2.2 all [installed,automatic]
libgtk-3-dev/jammy-updates,jammy-security,now 3.24.33-1ubuntu2.2 amd64 [installed]
libgtk-4-1/jammy-updates,jammy-security,now 4.6.9+ds-0ubuntu0.22.04.2 amd64 [installed,automatic]
libgtk-4-bin/jammy-updates,jammy-security,now 4.6.9+ds-0ubuntu0.22.04.2 amd64 [installed,automatic]
libgtk-4-common/jammy-updates,jammy-security,now 4.6.9+ds-0ubuntu0.22.04.2 all [installed,automatic]
libgtk-4-dev/jammy-updates,jammy-security,now 4.6.9+ds-0ubuntu0.22.04.2 amd64 [installed]
```

laeubi:
Jenkins CentOS currently using

```
gtk3-devel-3.22.30-11.el8.aarch64
webkit2gtk3-jsc-2.38.5-1.el8.aarch64
webkit2gtk3-2.38.5-1.el8.aarch64
gtk-update-icon-cache-3.22.30-11.el8.aarch64
webkit2gtk3-jsc-devel-2.38.5-1.el8.aarch64
gtk2-2.24.32-5.el8.aarch64
gtk2-devel-2.24.32-5.el8.aarch64
gtk3-3.22.30-11.el8.aarch64
libcanberra-gtk3-0.30-18.el8.aarch64
webkit2gtk3-devel-2.38.5-1.el8.aarch64
```

so it seems docker file change has had no effect...

fredg02:
When you have found the right job to build the right docker image, I'd verify that it includes everything that is needed by running it locally. Once that is done, you can specify the sha256 of the image in your Jenkinsfile to make sure that the exact image is pulled.

fredg02:
As noted above, https://ci.eclipse.org/releng/job/Build-Docker-images/ is not building the Dockerfile that @laeubi modified.

laeubi:
@fredg02 

I changed the file here:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2259

And Jenkins config says it runs `build-centos_9_swt_build.sh` what includes:

```
set -e

pushd centos-gtk4-mutter/9-swtBuild
echo "Building Centos 9 swt build image"
docker build --pull -t eclipse/platformreleng-centos-swt-build:9 .
popd
```

do I miss something obvious?

laeubi:
@sravanlakkimsetti can you maybe shade some light here, the Jenkins file config and collection of shellscripts seems not easy to understand :-\

I even wonder if it would not be easier to have one Jeankinsfile (in git) that includes the shell calls directly instead of many different shell scripts with only a few line in it.

fredg02:
> @fredg02
> I changed the file here:
> https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2259

Ok. That was not obvious to me, based on the comments here.

laeubi:
Finally the docker images was updated and now we have GTK4 at the build node, this is the used GTK version and it shows a lot of deprecation warnings:

```
19:52:21  Name        : gtk4-devel
19:52:21  Version     : 4.12.3
19:52:21  Release     : 2.el9
19:52:21  Architecture: x86_64
19:52:21  Install Date: Wed 28 Aug 2024 05:27:02 PM UTC
19:52:21  Group       : Unspecified
19:52:21  Size        : 13102429
19:52:21  License     : LGPL-2.0-or-later
19:52:21  Signature   : RSA/SHA256, Mon 18 Dec 2023 09:49:21 PM UTC, Key ID 05b555b38483c65d
19:52:21  Source RPM  : gtk4-4.12.3-2.el9.src.rpm
19:52:21  Build Date  : Fri 08 Dec 2023 12:37:08 PM UTC
19:52:21  Build Host  : x86-04.stream.rdu2.redhat.com
19:52:21  Packager    : builder@centos.org
19:52:21  Vendor      : CentOS
19:52:21  URL         : https://www.gtk.org
19:52:21  Summary     : Development files for GTK
```

HannesWell:
> @sravanlakkimsetti can you maybe shade some light here, the Jenkins file config and collection of shellscripts seems not easy to understand :-\
> 
> I even wonder if it would not be easier to have one Jeankinsfile (in git) that includes the shell calls directly instead of many different shell scripts with only a few line in it.

That were exaclty my toughts as well when I came along the docker images in the releng-aggregator repo the last time, but I never had the time to do that. But I think there is no reason that speaks against doing the latter if the Job is maintained in the releng.aggregator repo. In order to not lose it again:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2271

Lets have further discussion about improvements of that process there.

laeubi:
@fredg02 can I somehow run my docker image on the native (e.g. `aarch64` machine) e.g. something like

```
podTemplate(inheritFrom: 'centos-latest' /* inhert general configuration */, containers: [
	containerTemplate(name: 'swtbuild', image: 'eclipse/platformreleng-debian-swtnativebuild:12',
		resourceRequestCpu:'1000m', resourceRequestMemory:'512Mi',
		resourceLimitCpu:'2000m', resourceLimitMemory:'4096Mi',
		alwaysPullImage: true, command: 'cat', ttyEnabled: true)
]) {
	node('native.builder-'gtk.linux.aarch64) { stage(nativeBuildStageName) { container('swtbuild') { body() } } }
}
```

HannesWell:
If the native build agent's have docker installed, we could maybe launch another docker node from that agent?

And we could maybe even build a docker-image for SWT on the fly using:
https://www.jenkins.io/doc/book/pipeline/docker/#building-containers

If that works as desired/hoped, we could even maintain the docker-image in this repo without the need to push it to docker-hub.

laeubi:
> If that works as desired/hoped, we could even maintain the docker-image in this repo without the need to push it to docker-hub.

One drawback is that building a docker image can take some time and requires download of a lot of stuff, so I think using a prebuild image from a registry can still be good. On the other hand on a "real" machine one can probably benefit from caching (in both cases).

Anyways as ther is no easy "out of the box" way I first will head on to make the lib available on the machines:

- https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/5110

laeubi:
This is now ready to be merged once master is open for the next stream.

mickaelistria:
If we can audit that this change doesn't modify any of the already built artifacts (only adds new ones) and that the behavior seems backward compatible; then could we consider a merge ASAP, before next stream starts even?
@laeubi are those assumptions supposed to hold true?
@akurtakov What do you think?

laeubi:
This will at least raise the GLIBC version requirements and therefore might not be good in current release phase.

Jenkinsfile
```diff
@@ -228,7 +228,15 @@ pipeline {
...
-												sh build.sh install
+												if [[ ${PLATFORM} == gtk.linux.aarch64 ]]; then
+													sh build.sh -gtk3 install
+												elif [[ ${PLATFORM} == gtk.linux.ppc64le ]]; then
+													sh build.sh -gtk3 install
+												elif [[ ${PLATFORM} == gtk.linux.riscv64 ]]; then
+													sh build.sh -gtk3 install
+												else
+													sh build.sh install 
+												fi
```
HannesWell, line 239: If the script becomes a Groovy String-template (by using three double-quotes instead of three single-quotes), this part can be much shorter and much more expressive
```suggestion
   sh build.sh ${ os == 'linux' && arch != 'x86_64' ? '-gtk3' : '' } install
```

laeubi: As the goal is to successively build for other architectures as well I choose a "longer" form so it is more clear what to remove if one enable a certain architecture. Hopefully the whole block can even be removed again soon.

laeubi:
Now the native build succeeds but for some reason the text execution failed in the initialize shell.

laeubi:
I think the problem is that SWT still uses centos-latest pod:

https://github.com/eclipse-platform/eclipse.platform.swt/blob/6ab4ccd26a5f3b076016f2f949eeb0397514f065/Jenkinsfile#L87

akurtakov:
From the log: 
```
version `GLIBC_2.34' not found (required by /home/jenkins/agent/workspace/eclipse.platform.swt_PR-1422/eclipse.platform.swt/tests/org.eclipse.swt.tests/target/.swt/lib/linux/x86_64/libswt-pi3-gtk-4967r9.so)
```
Please move to ubuntu-latest (maybe in separate PR and rebase this one on top of it so we deliver that part now independent of gtk4)

laeubi:
I have created
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/1624

chirontt:
@laeubi 
Using the latest of master branch, I try to (re)build the SWT natives for Linux x86_64, on my box with Ubuntu 2404, I get **tons** of deprecation warnings when compiling for GTK4, and the compile then fails due to the `-Werror` compile option turned on for both GTK 3 & 4 versions:
https://github.com/eclipse-platform/eclipse.platform.swt/blob/cfa74b494291a25ec1b452ddfbddd92401ec0a43/bundles/org.eclipse.swt/Eclipse%20SWT%20PI/gtk/library/make_linux.mak#L114

To get around the above compile failure for GTK4, I need to turn off the `-Werror` flag for GTK4, like:
```
ifeq ($(GTK_VERSION), 3.0)
CFLAGS += -Werror
endif
```
Do you want me to submit a PR for the above change?

chirontt:
FYI: here are the manual commands I use to (re)build the SWT natives for Linux x86_64 on my Ubuntu box:
```shell
cd binaries/org.eclipse.swt.gtk.linux.x86_64
mvn clean process-resources -Dnative=gtk.linux.x86_64
```

laeubi:
@chirontt I think the goal is more to resolve these warnings (@sratz has recently working on these) so any help is welcome.

In the meantime you mentioned workaround seems suitable for local compilation and we probably should mention this in the developer docs.

PR #1426, opened 2024-08-28
Use centos-9 to build the swt natives

laeubi:
Centos-8 is out of service and [the container even does not build anymore](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2263) therefore we can't get any updates.

This changes the build to use Centos 9 for building instead.

FYI @akurtakov @fredg02 @sravanlakkimsetti 

I'll cancel the first build and trigger one with forced binary compilation so we can see everything works fine, this should then be merged as soon as the master is open for next release so we can get enough time to test. This will also be a prerequisite for GTK4+ support.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -24,7 +24,7 @@ kind: Pod
 spec:
   containers:
   - name: "swtbuild"
-    image: "eclipse/platformreleng-centos-swt-build:8"
+    image: "eclipse/platformreleng-centos-swt-build:9"
     imagePullPolicy: "Always"
     resources:
       limits:


laeubi:
All native builds are green

grafik

merks:
Very good. 👍 thanks 🙏

HannesWell (approved):
To give some back-ground why it was at 8: It just was there when I migrated the many manually maintained single Jenkins Jobs into this single pipeline.
So I think updating is good.

The same should probably be done in Equinox:
https://github.com/eclipse-equinox/equinox/blob/4526f764f444a6217fb614835452cb8a3956f12d/Jenkinsfile#L25

laeubi:
It seems EF plans to drop centos completely and instead offering ubuntu-lts as default runners:

https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/1623

I therefore wonder if we should do the same for the native builds instead of aim for centos-9 ...

laeubi:
Another option would be using debian what is quite stable and the base for ubuntu under the hood.

The github actions already use (and compile) for "ubuntu-latest" what is https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md at the moment.

fredg02:
> It seems EF plans to drop centos completely and instead offering ubuntu-lts as default runners:
>
> https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/1623
>
> I therefore wonder if we should do the same for the native builds instead of aim for centos-9 ... what would be Ubuntu 20.04.6 LTS as far as I can see.

FTR: You are free to make this decision independently of what we offer as default runners, since you can create and define custom docker images for your builds and tests.

laeubi:
Yes but basically many projects depend on SWT so if we require to use a (non standard) template this will impact many projects CI setups. So if we can just tell them to switch to a new (default) runner label that seems easy enough.

akurtakov:
> Another option would be using debian what is quite stable and the base for ubuntu under the hood.

Let's concentrate on the GTK 4 enablement. We have centos 9 already (even this PR) and it serves the purpose of providing initial builds. The problem is what to do with jenkins verification builds which are at centos 8 too. IMO we should be switching them to ubuntu-latest image (if available). 
As a next step going to Debian is probably good idea but I would rather have it later so a 100-steps task doesn't become 150-steps task now.

laeubi:
> We have centos 9 already (even this PR) and it serves the purpose of providing initial builds.

The problem is that centos-9 kicks out a large range of distributions due to higher GLIBC version, so for me the first step looks like to upgrade our infra from centos-8 > whateverisnextminimum (and this is what this PR is actually aiming at).

The we can already use this (independent form GTK4) and then look what is required from whateverisnextminimum > GTK4, this will also decrease the diff between `master` + `GTK4` branch.

laeubi:
I'll close this for now even though it proves we *can* switch it probably raises the barriers too much and as centos-9 is using rolling release we have not enough control over the impact to the overall community. 

I therefore continuing first here:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2441

to establish a new stable baseline for GTK4

PR #1429, opened 2024-08-29
Need to print the exception message to give more information

tjwatson:
If running on a system that does not have the required glibc version then important information is lost to tell the user what is wrong. For example:

version 'GLIBC_2.34' not found (required by .../libswt-pi3-gtk-4966r6.so)

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT PI/gtk/org/eclipse/swt/internal/gtk/OS.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT PI/gtk/org/eclipse/swt/internal/gtk/OS.java
@@ -87,14 +87,24 @@ public class OS extends C {
 				Library.loadLibrary("swt-pi4");
 			} catch (Throwable e) {
 				System.err.println("SWT OS.java Error: Failed to load swt-pi4, loading swt-pi3 as fallback.");
-				Library.loadLibrary("swt-pi3");
+				try {
+					Library.loadLibrary("swt-pi3");
+				} catch (Throwable fallback) {
+					e.addSuppressed(fallback);
+					throw e;
+				}
 			}
 		} else {
 			try {
 				Library.loadLibrary("swt-pi3");
 			} catch (Throwable e) {
 				System.err.println("SWT OS.java Error: Failed to load swt-pi3, loading swt-pi4 as fallback.");
-				Library.loadLibrary("swt-pi4");
+				try {
+					Library.loadLibrary("swt-pi4");
+				} catch (Throwable fallback) {
+					e.addSuppressed(fallback);
+					throw e;
+				}
 			}
 		}
 	}


tjwatson:
I think this is necessary before #1426 is done otherwise it will be impossible for end users to know why Eclipse fails to start when they don't have the required glibc version.

HannesWell (approved):
I agree that this is imporant and it looks good.

> version 'GLIBC_2.34' not found (required by .../libswt-pi3-gtk-4966r6.so)

Is that the message of the exception that is printed with this?
If not, could you please provide an example?

tjwatson:
> Is that the message of the exception that is printed with this? If not, could you please provide an example?

Here is the sanitized exceptions after removing my system path information:

```
java.lang.UnsatisfiedLinkError: Could not load SWT library. Reasons: 
	Failed to load library (".../eclipse/configuration/org.eclipse.osgi/266/0/.cp/libswt-pi3-gtk-4966r6.so") /lib64/libc.so.6: version `GLIBC_2.34' not found (required by .../eclipse/configuration/org.eclipse.osgi/266/0/.cp/libswt-pi3-gtk-4966r6.so)
	Can't load swt-pi3-gtk
	Can't load swt-pi3
	Failed to load library (".../.swt/lib/linux/x86_64/libswt-pi3-gtk-4966r6.so") /lib64/libc.so.6: version `GLIBC_2.34' not found (required by .../.swt/lib/linux/x86_64/libswt-pi3-gtk-4966r6.so)
	Can't load .../.swt/lib/linux/x86_64/libswt-pi3-gtk.so
	Can't load .../.swt/lib/linux/x86_64/libswt-pi3.so
	Failed to load library (".../.swt/lib/linux/x86_64/libswt-pi3-gtk-4966r6.so") /lib64/libc.so.6: version `GLIBC_2.34' not found (required by .../.swt/lib/linux/x86_64/libswt-pi3-gtk-4966r6.so)

```

The message is a little weird because of all the different fallbacks SWT has for attempting to load its libraries.

bundles/org.eclipse.swt/Eclipse SWT PI/gtk/org/eclipse/swt/internal/gtk/OS.java
```diff
@@ -86,13 +86,15 @@ public class OS extends C {
...
 			} catch (Throwable e) {
+				System.err.println(e.getMessage());
 				System.err.println("SWT OS.java Error: Failed to load swt-pi4, loading swt-pi3 as fallback.");
 				Library.loadLibrary("swt-pi3");
 			}
 		} else {
 			try {
 				Library.loadLibrary("swt-pi3");
 			} catch (Throwable e) {
+				System.err.println(e.getMessage());
```
laeubi, line 97: As message could be null, should we check for that case or simply use
```suggestion
				System.err.println(e);
```

tjwatson: My preference would be to actually capture the original exception, try the fallback;  If that fails do the `println(e)` of the fallback exception and then throw the original exception to the caller.

akurtakov: Please amend this PR accordingly as we should merge the other one ASAP.

laeubi:
> The message is a little weird because of all the different fallbacks SWT has for attempting to load its libraries.

I wonder if Javas "suppressed exception" feature could better be used here? this was for sure not possible at the time this method was introduced, but should work with Java 17?

bundles/org.eclipse.swt/Eclipse SWT PI/gtk/org/eclipse/swt/internal/gtk/OS.java
tjwatson: Done, ready to merge assuming it builds successfully.

laeubi: Great, can you post a how such stacktrace would look like?

tjwatson:
This now records the error in the log like this:

[1725464596788.log](https://github.com/user-attachments/files/16873432/1725464596788.log)

laeubi (approved):
I think this should be sufficient, as with all native errors its hard to understand for the user but includes enough to find out for someone investigating the failure and much better than simply hide the cause!

Issue #1434, opened 2024-09-02
JavaDoc hover in java editor looks clipped on HiDPI screens


MikOfClassX:
JavaDoc hover in java editor looks clipped on HiDPI screens.

This happens on Win10/Win11, HiDPI monitor, when screen scaling is != 100%

![immagine](https://github.com/user-attachments/assets/74aef160-a23c-4a0c-aa0f-f9b311b5d541)

iloveeclipse:
Could be both HiDPI or Edge related. Are you using edge browser?

Could you please post text info shown if you open Help -> About -> Installation Details -> Configuration and type "SWT" in the filter box?

MikOfClassX:
Id: org.eclipse.e4.ui.css.swt, Version: 0.15.400.v20240321-1245, Location: reference:file:plugins/org.eclipse.e4.ui.css.swt_0.15.400.v20240321-1245.jar
Id: org.eclipse.e4.ui.css.swt.theme, Version: 0.14.400.v20240424-0956, Location: reference:file:plugins/org.eclipse.e4.ui.css.swt.theme_0.14.400.v20240424-0956.jar
Id: org.eclipse.e4.ui.swt.win32, Version: 1.2.300.v20240416-0658, Location: reference:file:plugins/org.eclipse.e4.ui.swt.win32_1.2.300.v20240416-0658.jar
Id: org.eclipse.e4.ui.workbench.addons.swt, Version: 1.5.500.v20240620-1945, Location: reference:file:plugins/org.eclipse.e4.ui.workbench.addons.swt_1.5.500.v20240620-1945.jar
Id: org.eclipse.e4.ui.workbench.renderers.swt, Version: 0.16.500.v20240727-1037, Location: reference:file:plugins/org.eclipse.e4.ui.workbench.renderers.swt_0.16.500.v20240727-1037.jar
Id: org.eclipse.e4.ui.workbench.swt, Version: 0.17.500.v20240807-0911, Location: reference:file:plugins/org.eclipse.e4.ui.workbench.swt_0.17.500.v20240807-0911.jar
Id: org.eclipse.swt, Version: 3.127.0.v20240821-1414, Location: reference:file:plugins/org.eclipse.swt_3.127.0.v20240821-1414.jar
[...]

PR #1436, opened 2024-09-03
Use correct DPIUtil call for pixel conversion in OleClientSite.setBounds

merks:
- This commit fixes an issue where a call to DPIUtil was scaling down from points to pixels instead of scaling up

https://github.com/eclipse-platform/eclipse.platform.swt/issues/1434


merks:
Oomph wrong branch.

Issue #414, opened 2022-09-29
slow StyledTextRenderer


jukzi:
On windows eclipse text editors sometimes lag when scrolling
I sampled the times and see that most times is spend in
TextLayout.getLineCount() which is called for every line in the handlePaint() loop. 

As far as i understand the number of lines won't change while drawing so it should be enough to calculate the number of lines once.

![image](https://user-images.githubusercontent.com/51790620/192950469-0def7097-0f06-4f33-aabf-f08911bf1da7.png)

jukzi:
Turns out lineCount() is the drawn lines per lines which is ==1 without linewrap. but the shape() will be done anyway to calculate the width of the line.
The slow text file contains multiple <STX><ETX> sequences per line which the default font "consolas" can not draw. Font "Courier New" does not show the performance problem even though it can't render those special chars neither.

i prefer the notepad++ way to (fast) draw meaningfull symbols instead of undrawable characters:

eclipse:
![image](https://user-images.githubusercontent.com/51790620/193021140-44e27fba-93d3-4187-b6e9-f006b6ed3a35.png)
notepad++
![image](https://user-images.githubusercontent.com/51790620/193021202-a24f9669-d232-4a34-aba7-e063d2f1348d.png)

tmssngr:
Could you please provide a short snippet where this is reproducible with pure SWT?

jukzi:
@tmssngr here is a pur swt Example:

```
package org.eclipse.swt.snippets;
import org.eclipse.swt.*;
import org.eclipse.swt.custom.*;
import org.eclipse.swt.graphics.*;
import org.eclipse.swt.layout.*;
import org.eclipse.swt.widgets.*;

public class SlowStyledTextRender414 {

  public static void main(String[] args) {
    Display display = new Display();
    Shell shell = new Shell(display);
    shell.setLayout(new FillLayout());

    StringBuilder b=new StringBuilder();
    for (int y=0;y<1000;y++) {
    	b.append(y+" ");
	    for (int x=0;x<100;x++) {
	    	b.append("\u0002msg\u0003 ");
	 //fast:    	b.append("2msg3 ");
	    }
	    b.append("\n");
    }
    String s=b.toString();
    StyledText t = new StyledText(shell, SWT.MULTI | SWT.BORDER | SWT.V_SCROLL);
    // faster: Text t = new Text(shell, SWT.MULTI | SWT.BORDER | SWT.V_SCROLL);
[...]

PR #1439, opened 2024-09-03
#414: slow StyledTextRenderer for unicode and unprintable control cha…

tmssngr:
…racters

This test replaces the characters 0x00-0x1F with some alternative character (by default '?') for rendering. This improves the performance significantly.

The previous pull request (https://github.com/eclipse-platform/eclipse.platform.swt/pull/1425) did not work.

Note, that when changing `replaceBelowSpaceCharacters` to return the commented unicode characters, the snippet will fail with a "No more handles" exception. I reckon, we best should stick to a US-ASCII character from the range 0x20<=..<0x7F.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/TextLayout.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/TextLayout.java
@@ -2857,6 +2857,14 @@ public boolean isDisposed () {
 	return device == null;
 }
 
+private static char replaceBelowSpaceCharacters(char chr) {
+	if (chr >= 0x20 || chr == '\t' || chr == '\r' || chr == '\n') {
+		return chr;
+	}
+
+	return '?'; // (char)(0x2400 + chr);
+}
+
 /*
  *  Itemize the receiver text
  */
@@ -2884,6 +2892,9 @@ StyleItem[] itemize () {
 	int[] pcItems = new int[1];
 	char[] chars = new char[length];
 	segmentsText.getChars(0, length, chars, 0);
+	for (int i = 0; i < length; i++) {
+		chars[i] = replaceBelowSpaceCharacters(chars[i]);
+	}
 	// enable font ligatures
 	scriptControl.fMergeNeutralItems = true;
 	/*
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/TextLayout.java
```diff
@@ -2854,6 +2854,13 @@ public boolean isDisposed () {
 	return device == null;
 }
 
+private static char replaceBelowSpaceCharacters(char chr) {
+	return chr >= ' '
+		? chr
+		//: (char)(0x2400 + chr);
+		: '?';
```
merks, line 2861: Given that \t\r\n all fall into this category, can this really be correct?

tmssngr: You are right. The bug was not directly visible, but caused weird effects when editing.

tmssngr:
My commit used the correct, approved email address, but I've created the pull request with my GitHub account. Don't know why eclipsefdn/eca complains.

iloveeclipse:
> My commit used the correct, approved email address, but I've created the pull request with my GitHub account. Don't know why eclipsefdn/eca complains.

This commit should only have committers / authors with ECA signed, but you see there two different identities : https://github.com/syntevo/eclipse.platform.swt/commit/176f48ac36cfb00fa5db757f0b6e3fb5b88f885d

ECA wants clarity regarding the code origin, so that **every** contribution is legal.

mihnita:
I wonder if it has something to do with the "Interpret ASCII control characters" setting
(under Settings -- Run/Debug -- Console)

I know that is only about the Console, but the console is implemented on top of StyledText.
So maybe the root cause is that processing.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/TextLayout.java
```diff
@@ -2854,6 +2854,14 @@ public boolean isDisposed () {
 	return device == null;
 }
 
+private static char replaceBelowSpaceCharacters(char chr) {
+	if (chr >= 0x20 || chr == '\t' || chr == '\r' || chr == '\n') {
```
mihnita, line 2858: Should we consider allowing more?
Other characters that are used in CLI applications:
* `\u0007` : BELL, makes a "beep" sound
* `\u0008` : BS, backspace, moves back one character
* `\u001B` : ESC, used for ANSI escape sequences (changing colors, cursor movement, clear screen, etc). This one is used quite a lot.

Since the Eclipse Console is built on top of the styled text widget not allowing ESC might break quite a lot of things.
(see https://eclipse.dev/eclipse/news/4.25/platform.php#debug-ansi-support)

PR #1440, opened 2024-09-03
Version bumps for 4.34 stream

deepika-u:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2299


akurtakov:
Fragments have to be bumped too when host version is bumped.

deepika-u:
> Fragments have to be bumped too when host version is bumped.

Done.

Issue #924, opened 2023-11-30
[Win32] Possible ERROR_GRAPHIC_DISPOSED error during painting


tmssngr:
**Describe the bug**
A couple of users reported following exception
```
org.eclipse.swt.SWTException: Graphic is disposed
	at org.eclipse.swt.SWT.error(SWT.java:4918)
	at org.eclipse.swt.SWT.error(SWT.java:4833)
	at org.eclipse.swt.SWT.error(SWT.java:4804)
	at org.eclipse.swt.graphics.GC.fillRectangleInPixels(GC.java:3003)
	at org.eclipse.swt.graphics.GC.fillRectangle(GC.java:2999)
	at smartgit.MH.a(SourceFile:347)
	at smartgit.ML.handleEvent(SourceFile:629)
	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:89)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:4257)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1066)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1090)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1075)
	at org.eclipse.swt.widgets.Composite.WM_PAINT(Composite.java:1536)
	at org.eclipse.swt.widgets.Control.windowProc(Control.java:4803)
	at org.eclipse.swt.widgets.Display.windowProc(Display.java:5043)
	at org.eclipse.swt.internal.win32.OS.DispatchMessage(Native Method)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:3649)
...
```
It looks like `Composite.WM_PAINT` creates an `Event` with a `gc` with `handle == 0`. This can happen in this part of the `Composite.WM_PAINT` code:
```java
				long [] phdc = new long [1];
				int flags = OS.BPBF_COMPATIBLEBITMAP;
				RECT prcTarget = new RECT ();
				OS.SetRect (prcTarget, ps.left, ps.top, ps.right, ps.bottom);
				long hBufferedPaint = OS.BeginBufferedPaint (hDC, prcTarget, flags, null, phdc);
				GCData data = new GCData ();
				data.device = display;
				data.foreground = getForegroundPixel ();
				Control control = findBackgroundControl ();
				if (control == null) control = this;
				data.background = control.getBackgroundPixel ();
[...]

PR #1441, opened 2024-09-03
#924 [Win32] Possible ERROR_GRAPHIC_DISPOSED error during painting

tmssngr:
This should cover the case that BeginBufferedPaint might fail.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Composite.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Composite.java
@@ -1521,30 +1521,32 @@ LRESULT WM_PAINT (long wParam, long lParam) {
 				RECT prcTarget = new RECT ();
 				OS.SetRect (prcTarget, ps.left, ps.top, ps.right, ps.bottom);
 				long hBufferedPaint = OS.BeginBufferedPaint (hDC, prcTarget, flags, null, phdc);
-				GCData data = new GCData ();
-				data.device = display;
-				data.foreground = getForegroundPixel ();
-				Control control = findBackgroundControl ();
-				if (control == null) control = this;
-				data.background = control.getBackgroundPixel ();
-				data.font = Font.win32_new(display, OS.SendMessage (handle, OS.WM_GETFONT, 0, 0), nativeZoom);
-				data.uiState = (int)OS.SendMessage (handle, OS.WM_QUERYUISTATE, 0, 0);
-				if ((style & SWT.NO_BACKGROUND) != 0) {
-					/* This code is intentionally commented because it may be slow to copy bits from the screen */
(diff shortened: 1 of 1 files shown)


tmssngr:
Where can I find what check failed?

tmssngr:
Can somebody please merge this request? I don't think that the failing GTK tests are caused by this patch.

BeckerWdf:
So this PR fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/924 ?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Composite.java
```diff
@@ -1521,30 +1521,32 @@ LRESULT WM_PAINT (long wParam, long lParam) {
...
-				data.font = Font.win32_new(display, OS.SendMessage (handle, OS.WM_GETFONT, 0, 0), nativeZoom);
-				data.uiState = (int)OS.SendMessage (handle, OS.WM_QUERYUISTATE, 0, 0);
-				if ((style & SWT.NO_BACKGROUND) != 0) {
-					/* This code is intentionally commented because it may be slow to copy bits from the screen */
-					//paintGC.copyArea (image, ps.left, ps.top);
-				} else {
-					RECT rect = new RECT ();
-					OS.SetRect (rect, ps.left, ps.top, ps.right, ps.bottom);
-					drawBackground (phdc [0], rect);
+				if (hBufferedPaint != 0 && phdc[0] != 0) {
```
BeckerWdf, line 1524: The diff looks a bit complicated. Do I get it right that the only change is to add the

if (hBufferedPaint != 0 && phdc[0] != 0)

condition?

tmssngr: Yes.

BeckerWdf:
I am not an expert on that topic and even do no work on windows. So somebody else should  have a look at it.

HeikoKlare:
In general, the change looks sound to me. I am just wondering whether it is necessary and thus whether we should merge this change now. We are already past M3 and even though I would consider the risk of the change as low in general, at the same the affected code is quite crucial. The issue does not seem to be critical: the report has been made a year ago and no one else complained so far. So would it be sufficient to target this for 2025-03 M1? Then we have the whole three months of 2025-03 development for implicit testing by developers using Windows.

tmssngr:
Would someone be brave enough to commit to master? Why does the Linux build fail? Shouldn't be caused by this PR.

HeikoKlare:
> Would someone be brave enough to commit to master? Why does the Linux build fail? Shouldn't be caused by this PR.

Yes, I've just rebased the branch onto master. Linux build failures are existing and documented here: #1564
I'm just curious about the reported new Maven issues, which is why I retriggered the Jenkins build.

tmssngr:
Thanks for merging.

PR #1444, opened 2024-09-04
Add support for riscv64

yuzibo:


yuzibo:
In my initial thoughts, I think this PR can be a start to support riscv64.

akurtakov:
There should be no harm from getting this one in. Pushing.

yuzibo:
Thank you very much!  Next I will look at how to support more about this

PR #1446, opened 2024-09-05
Make Edge browser available in ControlExample

HeikoKlare:
The SWT ControlExample does currently not provide a configuration option to use Edge browser on Windows (i.e., to activate the SWT.EDGE flag, like is provided for SWT.WEBKIT) already.
This change adds an according configuration option.

image

### DIFF
--- a/examples/org.eclipse.swt.examples/src/org/eclipse/swt/examples/controlexample/BrowserTab.java
+++ b/examples/org.eclipse.swt.examples/src/org/eclipse/swt/examples/controlexample/BrowserTab.java
@@ -48,6 +48,7 @@ class BrowserTab extends Tab {
 
 	/* Style widgets added to the "Style" group */
 	Button webKitButton;
+	Button edgeButton;
 
 	String errorMessage, lastText, lastUrl;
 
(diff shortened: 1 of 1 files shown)


akurtakov:
Having "checkbox" buttons for both webkit and edge seems wrong as there is no way one can request both. It should either be made "radio" buttons or some logic be added to disable the other when selecting one of them.

HeikoKlare:
I thought about that as well, but actually the `Browser` implementation allows both flags to be set at the same time, so I would expect that I can test what happens when doing so with the `ControlExample`. I agree that this is no reasonable use case, but the `ControlExample`, in my understanding, is for testing all possible configurations and for testing only "reasonable" ones.

HeikoKlare:
I changed them to be radio buttons, as some other options that are (implicitly) mututally exclusive are represented in the same way:
image

But it's someone strange that we have some options that are mutually exclusive without either complains (like runtime checks when settings conflicting styles) or even specifications (like Javadoc) about this. It all seems implicit. Consider the button styles (SWT.TOGGLE, SWT.PUSH, SWT.RADIO, etc.). I can pass all of them to a single button, but it will not behave as requested (like a toggle + push + radio button), but only like one of them that, by accident, had precedence. Of course, it is conceptually difficult/impossible to behave like a push+radio button. but it might be possible to behave like a toggle+push button. In any case, there is some implicit API (mutual exclusions, precedence of options) in here, and some lack of documentation.

sratz (approved):
LGTM.

I like the radio button approach better, since `ControlExample` should be a practical utility to see what's meaningfully possible and not necessarily to explore undefined behavior.

PR #1451, opened 2024-09-05
Edge: Improve handling of LocationListener#changed()

### LINKED ISSUE #1464: Edge: LocationListener#changed() is not called
**Describe the bug**
`LocationListener#changing()` is called before navigation, but `LocationListener#changed()` is not called consistently afterwards.

**Environment:**
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS

sratz:
Previously, the changed() method was not called consistently.

Move the firing of 'changed' LocationEvents in Edge from handleSourceChanged() to handleNavigationCompleted().

Fixes #1464.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
@@ -666,9 +666,10 @@ int handleNavigationStarting(long pView, long pArgs, boolean top) {
 		listener.changing(event);
 		if (browser.isDisposed()) return COM.S_OK;
 	}
+	// Save location and top for all events that use navigationId.
+	// will be eventually cleared again in handleNavigationCompleted().
+	navigations.put(pNavId[0], event);
 	if (event.doit) {
-		// Save location and top for all events that use navigationId.
-		navigations.put(pNavId[0], event);
 		jsEnabled = jsEnabledOnNextPage;
 		settings.put_IsScriptEnabled(jsEnabled);
 		// Register browser functions in the new document.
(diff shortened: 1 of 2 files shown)


bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -618,6 +618,17 @@ int handleNavigationStarting(long pView, long pArgs, boolean top) {
...
+	// NavigationStarting / NavigationCompleted is of the form
+	//   data:text/html;<base64_encode_of_the_whole_content>
+	// but ICoreWebView2.get_Source(long[]) returns
+	//   about:blank
+	// Other Browser implementations (IE, WebKit, Safari)
+	// always use 'about:blank' in the LocationEvents sent to LocationListeners
+	// Therefore, to maximize compatibility we also send 'about:blank' here:
+	if (url.startsWith("data:text/html;")) {
+		url = "about:blank";
+	}
```
HeikoKlare, line 631: We should check whether this conflicts with #1395

sratz: Yes, it probably does, with #1395 we won't reach this piece of code anymore.

I have added a comment in #1395.

sratz: That part is now obsolete, as no more data:text/html URLs occur with #1395 merged.

sratz:
Also added a test case for this scenario.

If there are no objections I'd like to merge this Today, so that it's available in the I-builds on Monday.

PR #1453, opened 2024-09-07
Correct faulty reference to listener constants for Widget

HeikoKlare:
Widget#addListener(int, Listener) currently states that event constants are defined in the SWTError class, but these constants are defined in the SWT class. This change fixes the faulty reference.

The faulty reference was introduced with:
- #1112

@HannesWell, I guess this was just by accident and not intended, right?


HannesWell (approved):
I don't think this was intended. Thanks for fixing it.
IIRC I copied large parts of the doc from existing methods, so there might be more places that need adjustments.

HeikoKlare:
Thank you, Hannes! This only seems to be an auto-completion accident. `<code>SWT</code>` was replaced with `{@link SWTError}`. But I'll keep my eyes open for potential further needs of adjustment.

Issue #1447, opened 2024-09-05
New find/replace overlay broken under wayland


akurtakov:
Trying to use new find/replace under Wayland is almost entirely impossible, It constantly moves and hides. Just watch the video

https://github.com/user-attachments/assets/9b8085fa-ea45-45ed-934b-e047c1501628

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [x ] Linux

2. Additional OS info (e.g. OS version, Linux Desktop, etc)
Fedora 40, Gnome on Wayland

4. JRE/JDK version
 Java 23

akurtakov:
@jonahgraham @sratz  That might be interesting for the IDE WG as it will be a problem for users on latest Linux distros

HeikoKlare:
I quickly reproduced the issue on WSL using Wayland. At least one reason seems to be this:
- #790

Setting the shell's location (the one of the find/replace overlay shell) does not work, as also documented here: https://github.com/eclipse-platform/eclipse.platform.swt/issues/790#issuecomment-1699000148
But the find/replace overlay implementation relies on this.

This means, the GTK `Shell` implementation does not behave API-conforming on Wayland. Isn't this an issue also in other use cases? 
And just out of curiosity: Is there a reason for not adding a (temporary) comment to `Shell`'s API that the method will not work on Wayland, given that this issue is known for quite some time now and given that we seem to want / claim SWT to be usable on Wayland?

HeikoKlare:
We show a notification popup to users on first usage of the find/replace overlay to switch back to the existing find/replace dialog. This would give them the chance to easily switch to existing UI (they can of course still switch the UI in the preferences later on).
Unfortunately, this notification popup does also not work properly because of the same issue. Popups also use the `Shell#setLocation()` functionality: https://github.com/eclipse-platform/eclipse.platform.ui/blob/31a902749e53ae095fb98c40459ba8cfc70d6d0d/bundles/org.eclipse.jface.notifications/src/org/eclipse/jface/notifications/AbstractNotificationPopup.java#L422-L423

tfonda-fbk:
For anyone facing this, the workaround mentioned by Heiko is to open Preferences > General > Editors > Text editors and untick "Use find/replace overlay".
@HeikoKlare thanks for fixing this issue. May I ask if I will receive the fix as a software update via the 2024-09 update site, or if I have to wait a few months for the next Eclipse release to get it?

merks:
[...]

Issue #790, opened 2023-08-30
Shell.setLocation doesn't work properly (or at all)


merks:
**Describe the bug**

Calling Shell.setLocation doesn't visibly move the shell to the specified location, and sometimes does move the shell at all.

**To Reproduce**

I modified this snippet:
```
/*******************************************************************************
 * Copyright (c) 2000, 2005 IBM Corporation and others.
 *
 * This program and the accompanying materials
 * are made available under the terms of the Eclipse Public License 2.0
 * which accompanies this distribution, and is available at
 * https://www.eclipse.org/legal/epl-2.0/
 *
 * SPDX-License-Identifier: EPL-2.0
 *
 * Contributors:
 *     IBM Corporation - initial API and implementation
 *******************************************************************************/
package org.eclipse.swt.snippets;

/*
 * Monitor example snippet: center a shell on the primary monitor
 *
 * For a list of all SWT example snippets see
 * http://www.eclipse.org/swt/snippets/
 *
 * @since 3.0
 */
import org.eclipse.swt.graphics.*;
import org.eclipse.swt.widgets.*;

public class Snippet120 {

public static void main (String [] args) {
	Display display = new Display ();
	Shell shell = new Shell (display);
	shell.setText("Snippet 120");
	shell.setSize (200, 200);
	Monitor primary = display.getPrimaryMonitor ();
	Rectangle bounds = primary.getBounds ();
	Rectangle rect = shell.getBounds ();
	System.out.println("primaryMnitorBounds=" + bounds);
	int x = bounds.x + (bounds.width - rect.width) / 2;
	int y = bounds.y + (bounds.height - rect.height) / 2;
	String gtkVersion = System.getProperty("org.eclipse.swt.internal.gtk.version");
	System.out.println("gtkVersion=" + gtkVersion);
	System.out.println("location=" + x + "," + y);
	shell.setLocation (x, y);
	System.out.println("location=" + shell.getLocation());
	shell.open ();
	while (!shell.isDisposed ()) {
		if (!display.readAndDispatch ()) display.sleep ();
	}
	display.dispose ();
}
}
```
Here we see it run and clearly the location is not in the center of the monitor though it says it's in the center:

![image](https://github.com/eclipse-platform/eclipse.platform.swt/assets/208716/8b22abb4-a62e-4797-a9c7-b91055b31a7a)

The problem with the Eclipse Installer is even worse, where I cannot move the installer as we see here the value after calling set location remains unchanged:

[...]

PR #1454, opened 2024-09-07
Correct contracts of Control#[get|set]Location() #790

HeikoKlare:
The `Control#getLocation()` and `Control#setLocation(...)` operations currently claim to always work (i.e., yield reasonable values or actually change the location). This specification is incorrect when executing these methods in shells on a platform that uses the Wayland protocol. 
Currently, there does not seem to be a fix to make the implementation fulfill the specification. In case we claim that SWT can be used on Wayland, the specification thus has to be adapted to reflect that behavior. This change adds an according warning to the specification in order to avoid that clients rely on the specification to be fulfilled.

Currently, most UI elements requiring a separate shell/dialog are not placed as expected when using a platform with Wayland.  In the Eclipse SDK, this includes the find/replace UIs (both the existing dialog and the new overlay), notification popups, quick type hierarchy, and other.
Having a warning in the specification of the broken methods is supposed to avoid further usage of those methods in problematic scenarios (to prevent avoidable regressions like #1447). 

#### Why in Control instead of Shell?
We might also add the warning only to `Shell#[get|set]Location()`, as it only applies to those methods executed on a shell. This But that would break the substitution principle as you might have a control at hand, not knowing that it's actually a shell that behaves differently. 

Contributes to https://github.com/eclipse-platform/eclipse.platform.swt/issues/790

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Control.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Control.java
@@ -1726,10 +1726,15 @@ public Object getLayoutData () {
 }
 
 /**
- * Returns a point describing the receiver's location relative
- * to its parent in points (or its display if its parent is null), unless
- * the receiver is a shell. In this case, the point is
- * relative to the display.
+ * Returns a point describing the receiver's location relative to its parent in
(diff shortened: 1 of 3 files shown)


HeikoKlare:
@akurtakov what do you think?
The only alternative I see it to state that SWT does not support platforms using Wayland, but that's obviously not something we would like to do...

Issue #1455, opened 2024-09-09
Test_org_eclipse_swt_browser_Browser fails in IBuilds


jukzi:
Test_org_eclipse_swt_browser_Browser
for example:

```
java.lang.IllegalArgumentException: Illegal character in opaque part at index 2: C:\Users\genie.releng\workspace\AutomatedTests\ep434I-unit-win32-java17/workarea/I20240908-1800/eclipse-testing/test-eclipse/eclipse/plugins/org.eclipse.swt.tests_3.107.600.v20240905-0755/data/testWebsiteWithTitle.html
at java.base/java.net.URI.create(URI.java:906)
at org.eclipse.swt.tests.junit.Test_org_eclipse_swt_browser_Browser.getValidUrl(Test_org_eclipse_swt_browser_Browser.java:568)
at org.eclipse.swt.tests.junit.Test_org_eclipse_swt_browser_Browser.test_isLocationForCustomText_isSetUrlNotCustomTextUrl(Test_org_eclipse_swt_browser_Browser.java:580)
at java.base/java.lang.reflect.Method.invoke(Method.java:568)
at org.eclipse.test.TracingSuite.runChild(TracingSuite.java:279)Caused by: java.net.URISyntaxException: Illegal character in opaque part at index 2: C:\Users\genie.releng\workspace\AutomatedTests\ep434I-unit-win32-java17/workarea/I20240908-1800/eclipse-testing/test-eclipse/eclipse/plugins/org.eclipse.swt.tests_3.107.600.v20240905-0755/data/testWebsiteWithTitle.html
at java.base/java.net.URI$Parser.fail(URI.java:2974)
at java.base/java.net.URI$Parser.checkChars(URI.java:3145)
at java.base/java.net.URI$Parser.parse(URI.java:3181)
at java.base/java.net.URI.<init>(URI.java:623)
at java.base/java.net.URI.create(URI.java:904)... 4 more
```
https://download.eclipse.org/eclipse/downloads/drops4/I20240908-1800/testresults/html/org.eclipse.swt.tests_ep434I-unit-win32-java17_win32.win32.x86_64_17.html

(but also other OS)

probably since recent https://github.com/eclipse-platform/eclipse.platform.swt/pull/1409
@HeikoKlare  @amartya4256

HeikoKlare:
Definitely introduced by #1409, I will have a look.

[...]

PR #1456, opened 2024-09-09
Remove erroneous assertion from browser test #1455

HeikoKlare:
The browser test case test_isLocationForCustomText_isSetUrlNotCustomTextUrlAfterSetText validates that after setting a browser URL this same URL is then returned by Browser#getURL(). However, the Browser API does not guarantee this behavior and the assertion is even unnecessary for the scenario to be tested, which is already covered by other assertions at the proper level of abstraction.

This change removes the unnecessary assertion.

Contributes to https://github.com/eclipse-platform/eclipse.platform.swt/issues/1455


HeikoKlare:
@amartya4256 do you agree that this assertion can be removed?

amartya4256:
> @amartya4256 do you agree that this assertion can be removed?

@HeikoKlare yes it is not a relevant assertion and can be removed.

PR #1459, opened 2024-09-09
GlyphMetrics.width not properly scaled up

tobiasmelcher:
LineContentCodeMinings are not properly rendered on windows with odd text scale factors like 125%, 150%, 250%, ... 

Here a screenshot of the CodeMiningDemo application located in eclipse.platform.ui\examples\org.eclipse.jface.text.examples.
codemining_zoom_150
The gap after the echoed code mining "echo" does not appear for text scale factor 100% and 200%.

The gap disappears by applying the changes in this pull request as shown in the next screenshot:
codemining_zoom_150_fixed

All `DPIUtil.scaleUp()` calls in class `org.eclipse.swt.graphics.TextLayout` use the result of `getZoom` as zoom level; but not the ones for `GlyphMetrics.width` which use the value `nativeZoom`. Is this intended? This pull request now uses also `getZoom` for `GlyphMetrics.width`.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/TextLayout.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/TextLayout.java
@@ -1861,7 +1861,7 @@ Rectangle getBoundsInPixels (int start, int end) {
 			int cx = 0;
 			if (run.style != null && run.style.metrics != null) {
 				GlyphMetrics metrics = run.style.metrics;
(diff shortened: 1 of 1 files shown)


BeckerWdf:
@HeikoKlare: What's your opinion on this PR?

HeikoKlare (commented):
Sounds reasonable to me. @akoch-yatta @amartya4256, what do you think about this? Do you know whether there was some reason to directly use the actual `nativeZoom` here?

akoch-yatta:
Sounds good to me as well. I tested it a bit and worked as expected. I think the original implementation was based on a misconception how the glyph metrics must be scaled.

PR #1462, opened 2024-09-10
Edge: Use display-relative coordinates in ContextMenuRequested

sratz:


sratz:
@akoch-yatta Thanks for catching this in https://github.com/eclipse-platform/eclipse.platform.swt/pull/1450#discussion_r1751957179

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -711,13 +711,21 @@ int handleContextMenuRequested(long pView, long pArgs) {
...
 	long[] locationPointer = new long[1];
 	args.get_Location(locationPointer);
-	POINT pt = new POINT();
-	OS.MoveMemory(pt, locationPointer, POINT.sizeof);
-	pt.x = DPIUtil.autoScaleDown(pt.x); // To Points
-	pt.y = DPIUtil.autoScaleDown(pt.y); // To Points
+	POINT ptWidgetRelative = new POINT();
+	OS.MoveMemory(ptWidgetRelative, locationPointer, POINT.sizeof);
+	ptWidgetRelative.x = DPIUtil.autoScaleDown(ptWidgetRelative.x); // To Points
+	ptWidgetRelative.y = DPIUtil.autoScaleDown(ptWidgetRelative.y); // To Points
```
HeikoKlare, line 717: We may take the change to correct the usage of `autoScaleDown` to use the proper context zoom. It should be something like this:
```suggestion
	ptWidgetRelative.x = DPIUtil.scaleDown(ptWidgetRelative.x, browser.getShell().nativeZoom); // To Points
	ptWidgetRelative.y = DPIUtil.scaleDown(ptWidgetRelative.y, browser.getShell().nativeZoom); // To Points
```

sratz: Hmm, since the `pt` information comes from the WebView2 C library, isn't the actual device zoom irrelevant here? We always have to map to **100%** pixel coordinates because that's what the SWT API works with?
l

sratz: Hmm, I can't get this to work correctly in all combinations of

- `-Dswt.autoScale=[false|exact]`
- `-Dswt.autoScale.updateOnRuntime=[false|true]`

`Menu.setLocation()`, calls `org.eclipse.swt.widgets.Widget.getZoom()`, which uses `org.eclipse.swt.internal.DPIUtil.getZoomForAutoscaleProperty(int)`, which depends on `swt.autoScale`.

`Menu.setLocationInPixels()` is not visible from the `.browser` package.

The coordinates returned from Edge for the `pt` are relative to the WebView2, but are not zoomed.

I can't really put my head around all this scaling / DPI craziness involved here :O.

Maybe there even is a problem in how these methods currently work?

 @akoch-yatta Would be great, if you could take a look here!

akoch-yatta: I had a look into it. The conversion between points and pixels is always a bit messy. I think there is something wrong in Control::toDisplay and Control::toControl, to cover all cases, but I need to have a deeper look into that topic

sratz: The problem is that Edge knows nothing about the `swt.autoScale` property, so we have to deal with 3 different coordinate systems here:
- Win32 POINTs received from Edge
- SWT Display coordinates (these are affected by the `swt.autoScale`)
- Pixels

So this has to be done in a 2-step manner:
- Up from Win32 POINTs to Pixel (enforcing the native zoom value, ignoring `swt.autoScale`)
- Down to SWT Display (respecting `swt.autoScale`)

With the latest force-push things work fine for me in all possible combinations I tried. I guess #1536 also helped.

sratz:
I'd also like to get this into M2 as the current behavior is completely broken with anything other than 100% OS zoom level.

HeikoKlare:
> I'd also like to get this into M2 as the current behavior is completely broken with anything other than 100% OS zoom level.

I agree that we should just merge this for M2. Maybe @amartya4256 can have another look into this? @akoch-yatta may also review the latest change after the merge (when he's back from vacation).

sratz:
Test failures are unrelated.

PR #1463, opened 2024-09-11
Edge: Follow-up to local resource handling in setText()

sratz:
Follow-up on #1395.

See https://github.com/eclipse-platform/eclipse.platform.swt/pull/1395#issuecomment-2343645550

### DIFF
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
@@ -1105,6 +1105,44 @@ public void test_setText() {
 	validateTitleChanged(expectedTitle, browserSetFunc);
 }
 
+/**
+ * Corner-case, probably only relevant on Edge, see
+ * https://github.com/eclipse-platform/eclipse.platform.swt/pull/1463
+ */
+@Test
+public void test_setTextContainingScript_applicationLayerProgressListenerMustSeeUpToDateDom() {
+	AtomicBoolean completed = new AtomicBoolean();
+	browser.addProgressListener(ProgressListener.completedAdapter(event -> {
+		String script = """
+				var h1s = document.getElementsByTagName("h1");
+				// extract the information from the DOM via the document's title
+				// since getText() afterwards does not necessarily return the updated DOM (platform-dependent)
+				document.title = "ProgressListener: Found " + h1s.length + " h1 tag(s)";
+				""";
+		browser.execute(script);
+		completed.set(true);
+	}));
+	AtomicReference<String> title = new AtomicReference<>();
+	browser.addTitleListener(event -> {
+		if (event.title.startsWith("ProgressListener: ")) {
+			title.set(event.title);
+		}
+	});
+	browser.setText("""
+			<html>
+				<head>
+					<script src=\"file:///does/not/really/needs/to/exist.js\"></script>
+				</head>
+				<body>
+					<h1>Hello, World!</h1>
+				</body>
+			</html>
+			""");
+	waitForPassCondition(completed::get);
+	waitForPassCondition(() -> title.get() != null);
+	assertEquals("ProgressListener: Found 1 h1 tag(s)", title.get());
+}
+
 @Test
 public void test_setUrl_local() {
 	assumeFalse("Test fails on Mac, see https://github.com/eclipse-platform/eclipse.platform.swt/issues/722", SwtTestUtil.isCocoa);
--- a/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
@@ -735,7 +735,7 @@ int handleDOMContentLoaded(long pView, long pArgs) {
 				return COM.S_OK;
 			});
 			webView.ExecuteScript(
-					stringToWstr("document.open(); document.write(`" + lastCustomText + "`); document.close();"),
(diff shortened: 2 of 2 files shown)


sratz:
> Are you able to run the browser tests using Edge with this change? When I execute the `Test_org_eclipse_swt_browser_Browser` with `-Dorg.eclipse.swt.browser.DefaultType=edge`, several tests fail/timeout.

Hmm, you are right, setting innerHTML does not evaluate javascript.

So open() + write() is then the only way.

The question is: Why is the DOM not immediately available in subsequent ProgressListeners?

sratz:
I think this needs to be a 2-step process.

The internal ProgressListener to manipulate the DOM needs to be separated from the ones of the application layer code.

We first need to wait for the DOM manipulation event to finish before sending out the downstream application code ProgressListeners

HeikoKlare:
I think the problem is that `Edge#execute()`, which is used to submit the JS operation to set the contents, operates asynchronously (it calls `ICoreWebView2#ExecuteScript()`):
https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2?view=webview2-1.0.2592.51#executescript

So yes, sending the progress event needs to performed after this asynchronous operation has been finished and not already when it has been submitted.

sratz:
Yep, that's it.

I've implemented this in the latest force-push.

With that, all browser tests are green and the Demo snippet works as expected.

laeubi:
May I ask why we use javascript in the first place and not https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2?view=webview2-1.0.2592.51#navigatetostring ?
This seems to be the natural choice for "setText"....

sratz:
> May I ask why we use javascript in the first place and not https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2?view=webview2-1.0.2592.51#navigatetostring ? This seems to be the natural choice for "setText"....

See #1395.

sratz:
I have added another commit to safeguard the data before passing it via a JavaScript string

laeubi:
> See https://github.com/eclipse-platform/eclipse.platform.swt/pull/1395.

That's odd and looks like a bug in the browser

sratz:
> > See #1395.
> 
> That's odd and looks like a bug in the browser

Sadly, it's not :(, see also (https://github.com/eclipse-platform/eclipse.platform.swt/issues/213#issuecomment-1586105245:

https://learn.microsoft.com/en-us/microsoft-edge/webview2/concepts/working-with-local-content?tabs=dotnetcsharp#additional-web-resources-1
>Loading local content by using the NavigateToString method doesn't make it possible for the resulting document to reference additional web resources such as CSS, image, or script files. The method only lets you specify the string content of the HTML document. To reference additional web resources from your HTML document, use one of the other approaches described in this article, or represent those additional web resources inline in the HTML document.

laeubi:
But is this not exactly what `setText` is supposed to do? I never seen external resource where allowed to load when using `setText` but I might be wrong. One should take into account that this is actually also for security.

HeikoKlare:
> In general, it's great to see that with this fix Edge's setText() works properly, so even Javadoc, Java element information, Welcome page etc. can be displayed. (https://github.com/eclipse-platform/eclipse.platform.swt/pull/1395#pullrequestreview-2235639395)

Many views in the Eclipse SDK require that `setText()` loads external resources (Welcome page, Javadoc, Java element info, help, ...). Until two days ago, all these things were broken in the SDK when using Edge.

In addition, I would propose not to further pollute this PR with such a discussion, but either have it at the original PR or start an separate issue/discussion about it.

HeikoKlare (commented):
This looks sound and works as expected. Thank you!
Can we add a regression test case (according to what the Demo class you have provided did)?

HeikoKlare:
Maybe something like:
```java
@Test
public void test_setText_progress() {
	String expectedBody = "<h1>Hello, World!</h1>";
	AtomicReference<String> textAfterProgressCompletion = new AtomicReference<>();
	browser.addProgressListener(ProgressListener.completedAdapter(event -> {
		textAfterProgressCompletion.set(browser.getText());
	}));
	browser.setText("""
		<html>
			<head>
				<script src=\"file://c:/test.js\"></script>
			</head>
			<body>
				%s
			</body>
		</html>
		""".formatted(expectedBody));
	assertTrue(waitForPassCondition(() -> textAfterProgressCompletion.get() != null && textAfterProgressCompletion.get().contains(expectedBody)));
}
```

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
```diff
@@ -1105,6 +1105,35 @@ public void test_setText() {
...
+				<head>
+					<script src=\"file:///does/not/really/needs/to/exist.js\"></script>
+				</head>
+				<body>
+					<h1>Initial Heading</h1>
+				</body>
+			</html>
+			""");
+	assertTrue(waitForPassCondition(() -> textAfterProgressCompletion.get() != null
+			&& textAfterProgressCompletion.get().contains("<h1>Hello, World!</h1>")));
```
HeikoKlare, line 1134: Not sure why the test is failing. Locally, it works fine (with both IE and Edge). Maybe improve the assertion to get better feedback? E.g. with the available Hamcrest matchers:
```suggestion
	assertThat(textAfterProgressCompletion.get(), containsString("<h1>Hello, World!</h1>"));
```

sratz: Yeah same here on Windows. However, on Linux it did fail also locally, because getText() behaves differently.

Tried another approach to exfiltrate the modified DOM content.

sratz: Finally all tests are green :)

PR #1445, opened 2024-09-04
Add linux riscv64 build with bundles

yuzibo:

PR #1469, opened 2024-09-14
Fix riscv64 MANIFEST.MF header field

Sakura286:
A blank after a header field is essential.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/pull/1445


HannesWell (approved):
Thank you.

PR #1475, opened 2024-09-19
 Add dependency to o.e.test to o.e.swt.tests

akurtakov:
Needed for Screenshots and TracingSuite classes.


akurtakov:
Merging with failing tests as the issue is known and discussed in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2360 and this test shouldn't cause breakage as seen(at least in theory). I-build will confirm that.

PR #1477, opened 2024-09-19
Destroy unused handles

amartya4256:
This PR contributes to the cleanup of those handles of Resources which were created for the zoom levels which are not applicable for the existing zoom levels of all the monitor. i.e., if there were 2 monitors with 100% and 200% zoom levels and the seconds monitor is switched to 100% as well. On the DPI_CHANGED event, we trigger a cleanup of handles of the resources using the resource tracker to remove the handles which were created for 200% zoom level since there doesn't exist any monitor with a zoom level of 100% anymore.

With this PR, we also refactor the resource tracker to allow it to track all the resources for win32 all the time but rather using the tracking flag to report errors, since that is the only use case of the tracking flag as of now. By allowing the Resource Tracker to track the resources, we can trigger the cleanup of all the relevant handles from a single point.

Contributes to #62 and #127

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Resource.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Resource.java
@@ -13,6 +13,7 @@
  *******************************************************************************/
 package org.eclipse.swt.graphics;
 
+import java.util.*;
 import java.util.function.*;
 
 import org.eclipse.swt.*;
@@ -121,6 +122,17 @@ Resource(Device device) {
 }
 
 void destroy() {
+}
+
+/**
+ * Destroy all handles of the resource which are not necessary for the given
+ * zoom levels. This method is supposed to be overridden by sub-classes that
+ * retain handles for different zoom levels.
+ *
+ * @param zoomLevels The zoom levels for which the handles are supposed to be
+ *                   retained.
+ */
+void destroyHandlesExcept(Set<Integer> zoomLevels) {
 }
 
 /**
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Device.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Device.java
@@ -14,10 +14,14 @@
 package org.eclipse.swt.graphics;
 
 
+import java.util.*;
+import java.util.concurrent.*;
+
 import org.eclipse.swt.*;
 import org.eclipse.swt.internal.*;
 import org.eclipse.swt.internal.gdip.*;
 import org.eclipse.swt.internal.win32.*;
+import org.eclipse.swt.widgets.*;
 
 /**
  * This class is the abstract superclass of all device objects,
@@ -55,6 +59,7 @@ public abstract class Device implements Drawable {
 	String[] loadedFonts;
 
 	volatile boolean disposed;
+	private Set<Resource> resourcesWithZoomSupport  = ConcurrentHashMap.newKeySet();
 
 	/*
 	* TEMPORARY CODE. When a graphics object is
@@ -914,6 +919,8 @@ protected void release () {
 		Gdip.GdiplusShutdown (gdipToken[0]);
 	}
 	SWTFontProvider.disposeFontRegistry(this);
+	resourcesWithZoomSupport.clear();
+	resourcesWithZoomSupport = null;
 	gdipToken = null;
 	scripts = null;
 	logFonts = null;
@@ -946,5 +953,25 @@ public void setWarnings (boolean warnings) {
  */
 protected int getDeviceZoom () {
 	return DPIUtil.mapDPIToZoom ( _getDPIx ());
+}
+
+void registerResourceWithZoomSupport(Resource resource) {
+	resourcesWithZoomSupport.add(resource);
+}
+
+/**
+ * Destroys the handles of all the resources in the resource tracker by
+ * identifying the zoom levels which is not valid for any monitor
+ *
+ * @noreference This method is not intended to be referenced by clients.
+ */
+public static void win32_destroyUnusedHandles(Display display) {
+	Set<Integer> availableZoomLevels = new HashSet<>();
+	for (Monitor monitor : display.getMonitors()) {
+	    availableZoomLevels.add(DPIUtil.getZoomForAutoscaleProperty(monitor.getZoom()));
+	}
+	for (Resource resource: ((Device) display).resourcesWithZoomSupport) {
+		resource.destroyHandlesExcept(availableZoomLevels);
+	}
 }
 }
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
@@ -164,6 +164,7 @@ public final class Image extends Resource implements Drawable {
 Image (Device device) {
 	super(device);
 	initialNativeZoom = DPIUtil.getNativeDeviceZoom();
+	this.device.registerResourceWithZoomSupport(this);
 }
 
 /**
@@ -209,6 +210,7 @@ public Image(Device device, int width, int height) {
 	height = DPIUtil.scaleUp (height, zoom);
 	init(width, height);
 	init();
+	this.device.registerResourceWithZoomSupport(this);
 }
 
 /**
@@ -311,6 +313,7 @@ public Image(Device device, Image srcImage, int flag) {
 			SWT.error(SWT.ERROR_INVALID_ARGUMENT);
 	}
 	init();
+	this.device.registerResourceWithZoomSupport(this);
 }
 
 /**
@@ -355,6 +358,7 @@ public Image(Device device, Rectangle bounds) {
 	bounds = DPIUtil.scaleUp (bounds, getZoom());
 	init(bounds.width, bounds.height);
 	init();
+	this.device.registerResourceWithZoomSupport(this);
 }
 
 /**
@@ -388,6 +392,7 @@ public Image(Device device, ImageData data) {
 	data = DPIUtil.autoScaleUp(device, this.dataAtBaseZoom);
 	init(data, getZoom());
 	init();
+	this.device.registerResourceWithZoomSupport(this);
 }
 
 /**
@@ -434,6 +439,7 @@ public Image(Device device, ImageData source, ImageData mask) {
 	mask = ImageData.convertMask(mask);
 	init(this.device, this, source, mask, getZoom());
 	init();
+	this.device.registerResourceWithZoomSupport(this);
 }
 
 /**
@@ -496,6 +502,7 @@ public Image (Device device, InputStream stream) {
 	ImageData data = DPIUtil.autoScaleUp(device, this.dataAtBaseZoom);
 	init(data, getZoom());
 	init();
+	this.device.registerResourceWithZoomSupport(this);
 }
 
 /**
@@ -538,6 +545,7 @@ public Image (Device device, String filename) {
 	ImageData data = DPIUtil.autoScaleUp(device, this.dataAtBaseZoom);
 	init(data, getZoom());
 	init();
+	this.device.registerResourceWithZoomSupport(this);
 }
 
 /**
@@ -586,6 +594,7 @@ public Image(Device device, ImageFileNameProvider imageFileNameProvider) {
 		init(resizedData, getZoom());
 	}
 	init();
+	this.device.registerResourceWithZoomSupport(this);
 }
 
 /**
@@ -625,6 +634,7 @@ public Image(Device device, ImageDataProvider imageDataProvider) {
 	ImageData resizedData = DPIUtil.scaleImageData(device, data.element(), getZoom(), data.zoom());
 	init (resizedData, getZoom());
 	init();
+	this.device.registerResourceWithZoomSupport(this);
 }
 
 private ImageData adaptImageDataIfDisabledOrGray(ImageData data) {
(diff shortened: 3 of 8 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Device.java
```diff
@@ -820,26 +826,30 @@ void new_Object (Object object) {
...
+		}
 	}
 }
 
 void printErrors () {
 	if (!DEBUG) return;
 	if (tracking) {
 		synchronized (trackingLock) {
-			if (objects == null || errors == null) return;
+			if (objects == null && errors == null) return;
```
fedejeanne, line 852: Why the change here?

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Resource.java
```diff
@@ -123,6 +124,10 @@ public Resource() {
 void destroy() {
 }
 
+void destroyHandlesExcept(Set<Integer> zoomLevels) {
```
fedejeanne, line 127: Add some documentation here and mention the fact that the default implementation is empty and therefore sub-classes should override this method if necessary.

fedejeanne (commented):
I only skimmed the code, I haven't tested it yet (I'll do it as soon as the PR is ready to be reviewed). I only have 2 minor remarks.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Device.java
amartya4256: Its because earlier we had both objects and errors guarded by the lock. But now, object is not guarded by the lock anymore and it would not be null anymore. So, we only wanna return already if both image and errors are null. i think even one check just for errors would also be fine. We don't need to check for objects since the method is called printErrors. Why print only objects in case there's no error? What do you say?

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Resource.java
amartya4256: Alright

fedejeanne: How about:

"Destroy all handles of the resource  which are <strong>not</strong> necessary for the given zoom levels. This method is supposed to be overridden by sub-classes that retain handles for different zoom levels."

For the body? 

The `@param` is ok.

amartya4256: But wer dont know what zoom levels these resources have handles of. So, the only way possible is to tell them to keep these zoom levels and discard others if they have.

amartya4256: alright

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Device.java
```diff
@@ -947,4 +953,24 @@ public void setWarnings (boolean warnings) {
 protected int getDeviceZoom () {
 	return DPIUtil.mapDPIToZoom ( _getDPIx ());
 }
+
+void registerResource(Resource resource) {
```
fedejeanne, line 957:
```suggestion
void registerResourceWithZoomSupport(Resource r) {
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Device.java
```diff
@@ -55,6 +58,7 @@ public abstract class Device implements Drawable {
 	String[] loadedFonts;
 
 	volatile boolean disposed;
+	private Vector<Resource> resources = new Vector<>();
```
fedejeanne, line 61:
```suggestion
	private Vector<Resource> resourcesWithZoomSupport = new Vector<>();
```

Also, why use `Vector` and not `Collection` / `List`? Do you need synchronization?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Device.java
fedejeanne: The change has been reverted.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Device.java
amartya4256: Yes

fedejeanne: Rename?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Device.java
fedejeanne: Rename?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Device.java
amartya4256: Done

fedejeanne (approved):
The changes look good and produce the expected results. Also, no regressions were found during the testing.

I re-triggered the builds because there were some test failures (for Browser!). If the tests pass in the next run then this PR is good to go on my account.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1186,16 +1196,31 @@ void destroy () {
...
 
+@Override
+void destroyHandlesExcept(Set<Integer> zoomLevels) {
+	zoomLevelToHandle.entrySet().removeIf(entry -> {
+		if (!zoomLevels.contains(entry.getKey()) && entry.getKey() != DPIUtil.getZoomForAutoscaleProperty(initialNativeZoom)) {
+			destroyHandle(entry.getValue());
+            return true;
+        }
+        return false;
+	});
```
HeikoKlare, line 1213: Please properly format code (here and at other similar places).

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1186,16 +1196,31 @@ void destroy () {
...
+		destroyHandle(handle);
 	}
 	zoomLevelToHandle.clear();
 	handle = 0;
 }
 
+@Override
+void destroyHandlesExcept(Set<Integer> zoomLevels) {
+	zoomLevelToHandle.entrySet().removeIf(entry -> {
+		if (!zoomLevels.contains(entry.getKey()) && entry.getKey() != DPIUtil.getZoomForAutoscaleProperty(initialNativeZoom)) {
```
HeikoKlare, line 1208: Extracting reused `entry.getKey()` into a self-explaining variable would improve readability.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Device.java
```diff
@@ -55,6 +58,7 @@ public abstract class Device implements Drawable {
 	String[] loadedFonts;
 
 	volatile boolean disposed;
+	private Vector<Resource> resourcesWithZoomSupport  = new Vector<>();
```
HeikoKlare, line 61: Why use outdated Vector instead of, e.g., a synchronized set?

HeikoKlare (commented):
I did some sanity checking by testing the change with an SDK product:
- Two monitors, the secondary at 100%, the primary one switched between multiple zoom values
- Once with `swt.autoScale=quarter` and once with default (`integer200`)

I have locally added some logging to see when resources are destroyed and it appears to happen when expected.

I have not conduced an in-depth code review but some minor comments that should be addressed before merging.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Device.java
amartya4256: Done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java

HeikoKlare: Fotmatting is still inconsistent.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
HeikoKlare: Pleas also adapt the other places with the same code consistently

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Device.java
HeikoKlare: I meant to use a synchronized set, not to have a set to which every access it synchronized (as that's quite similiar to using a vector).

amartya4256: Oh sorry, its done now

laeubi: I suppose everything has to happen in the UI thread so do we need synchronized (whatever) anyways?
That's actually one of the main points why UI frameworks use "display threads" to prevent the synchronization and possible deadlocks if multiple threads would operate on the same output device.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Device.java
```diff
@@ -55,6 +58,7 @@ public abstract class Device implements Drawable {
 	String[] loadedFonts;
 
 	volatile boolean disposed;
+	private Set<Resource> resourcesWithZoomSupport  = Collections.synchronizedSet(new HashSet<>());
```
laeubi, line 61: See general comment about `synchronized`, in any case 

```suggestion
	private Set<Resource> resourcesWithZoomSupport  = ConcurrentHashMap.newKeySet();
```

might be a better alternative these days.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -355,6 +358,7 @@ public Image(Device device, Rectangle bounds) {
 	bounds = DPIUtil.scaleUp (bounds, getZoom());
 	init(bounds.width, bounds.height);
 	init();
+	this.device.registerResourceWithZoomSupport(this);
```
laeubi, line 361: init seem to be called everywhere where this code is called, so maybe put it in the init method?!?

amartya4256: We only wanna register those resources which have multiple handles for different zoom levels. But some other classes like Cursor, Font, etc also call init() but they don't have multiple handles like Image, Path, Pattern, etc so it would be better to call it from the constructors. Also, some constructors like Image(Device device) do not call init() so it will make the registration of such resources inconsistent.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Device.java
HeikoKlare: Valid point. I did not challenge the original answer that synchronization is necessary here (see https://github.com/eclipse-platform/eclipse.platform.swt/pull/1477#discussion_r1783956150), but taking a look at the implementation I do not see any access to the data structure that can happen from outside the UI thread, as both resource instantiation and OS callback processing happen within the UI thread (i.e., the thread of the according `Device`/`Display` instance), don't they?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
fedejeanne: @amartya4256 I noticed that some of the new code uses spaces instead of tabs to indent. Maybe that's what Heiko means by "still inconsistent"?

HeikoKlare: > I noticed that some of the new code uses spaces instead of tabs to indent. Maybe that's what Heiko means by "still inconsistent"?

Yes, seems to be a mixture of spaces and tabs. E.g., this:
{DF3ACF41-FFEE-4B87-B72C-7560E8E0F39D}

@amartya4256 can you please adapt the inconsistent formatting at all remaining places?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
fedejeanne: Are you satisfied with the answer @laeubi ? May we mark this conversation as "resolved"?

fedejeanne:
Jenkins is still [offline](https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/5131) but the last checks were fine. Merging.

Any additional comments may be addressed in a follow-up PR.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
fedejeanne: They seem to be done.

I'll mark this as resolved. If there are any open issues, they can be addressed in a follow-up PR.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
fedejeanne: @laeubi since we would like to have this in M2 and the current approach LGTM, I merged this PR. 

We can continue this discussion in a follow-up PR if necessary.

akurtakov:
> Jenkins is still [offline](https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/5131) but the last checks were fine. Merging.
> 
> Any additional comments may be addressed in a follow-up PR.

The report you link to is not about Platform and Releng JIPPs so unless you open one specifically for them they will stay offline.

HeikoKlare:
The releng Jenkins has just been restarted.

HeikoKlare:
After Jenkins was back, I just triggered an SWT master build to verify that none of the PRs merged today produced any issues. It finished successfully: https://ci.eclipse.org/releng/job/eclipse.platform.swt/job/master/767/

Regarding reporting of the Jenkins issue, there was a comment that several instances were affected and restarted, so at least I assumed that the infrastructure team checks the whole list of instances:
{CE680D35-5546-4F83-9634-4323ADF91308}

Issue #1205, opened 2024-05-01
Shell close event listener not called on Command+Q on Mac


mwhapples:
On MacOSX when I press Command+Q for quit in an SWT application the shell closes without calling the close event listener. My Mac is a bit outdated, but I have been assured by work colleagues the issue still shows with newer Macs.

Steps to reproduce:
1. Launch Eclipse SWT Snippet99
2. Press Command+Q

Expected: The message box asking whether to close should appear.
Actual: The shell just closes.

SWT version tested: 3.125.0 (from maven central)

This bug seems to be specific to pressing the Command+Q keyboard shortcut, pressing the close button on screen works as expected.

tmssngr:
IIRC, you need to add a SWT.Dispose listener to the display to get a notification for Cmd+Q on MacOS.

mwhapples:
If I understand it correctly a dispose listener on the Display object will be called after the shell is closed, so attempting to cancel the dispose event will not prevent the shell being closed. Thus it does not seem to be a solution if a confirmation to quit dialog is wanted like in Snippet99 and the prompt is to appear when using Command+Q keyboard shortcut. If I have misunderstood and it does actually work as I would have hoped, then may be Snippet99 needs updating to include this as well.

mwhapples:
@elsazac thanks for the example of using a dispose listener. Whilst that does run the listener it seems like the setting of doit is being ignored, thus meaning the event cannot be cancelled. So even when pressing no in the example the window still closes. So I think there really is a bug here in SWT which needs fixing.

tmssngr:
Please provide a complete snippet how to reproduce it. Snippet99 has no dispose listener on the display.

mwhapples:
@tmssngr here is the code I tested with, attached and with the java extension changed to txt as GitHub will not let me attach a .java file. I renamed the close button to dispose for the display dispose listener so it is possible to identify which listener is being called. Note how even when pressing the no button the window closes and the app exits, so not replicating how Snippet99 works on other platforms or even serving the purpose of the snippet. I invoke java from the commandline outside any IDE or other environment.

[Example with display dispose listener](https://github.com/user-attachments/files/20141296/Snippet99Fix.txt)

elsazac:
[...]

PR #1484, opened 2024-09-23
Command+Q on Snippet99

elsazac:
Fixes: https://github.com/eclipse-platform/eclipse.platform.swt/issues/1205

Snippet99 requires modification as Command+Q is not implemented on that. Currently, the snippet demonstrates a message box pop-up when attempting to close the dialog window, but pressing Command+Q closes the window without triggering this behaviour. The snippet has been updated to implement Command+Q functionality and to add an SWT dispose listener to the display object.

### DIFF
--- a/examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet99.java
+++ b/examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet99.java
@@ -30,13 +30,26 @@ public static void main (String [] args) {
 	Display display = new Display ();
 	final Shell shell = new Shell (display);
 	shell.setText("Snippet 99");
-	shell.addListener (SWT.Close, event -> {
-		MessageBox messageBox = new MessageBox(shell, SWT.ICON_QUESTION | SWT.YES | SWT.NO);
-		messageBox.setText("Information");
-		messageBox.setMessage("Close the shell?");
-		messageBox.setButtonLabels(Map.of(SWT.YES, "Close"));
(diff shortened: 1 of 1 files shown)


Phillipus:
But this doesn't demonstrate what is being asked in https://github.com/eclipse-platform/eclipse.platform.swt/issues/1205

As said in [this comment](https://github.com/eclipse-platform/eclipse.platform.swt/issues/1205#issuecomment-2093195557):

> a dispose listener on the Display object will be called after the shell is closed, so attempting to cancel the dispose event will not prevent the shell being closed. 

So in this Snippet with the PR pressing Command+Q will open the dialog but choosing "No" has no effect.

elsazac:
> But this doesn't demonstrate what is being asked in #1205

> So in this PR pressing Command+Q will open the dialog but choosing "No" has no effect. i.e `event.doit = false` has no effect on the dispose event.

I just noticed that while the snippet shows the message box, it doesn't prevent the app from quitting when Command+Q is pressed. Since in macOS, Command+Q is handled at the OS level, even though the SWT.Dispose event listener can trigger the message box, it doesn't have control over stopping the closing of the shell process triggered by the system. So, the application closes right after the message box appears, regardless of the any other choice.

I tried implementing the keydown event for Command and Q, but this doesn't work either because macOS takes over this key combination before it reaches the SWT application.

Does anyone have any other suggestions on how to handle this situation effectively?

akurtakov:
Will this one be finished or it should be closed?

akurtakov:
Closing as there seems to be no interest in finishing it.

PR #1485, opened 2024-09-24
Add support riscv64 with p2.inf

yuzibo:
This can be enable after riscv64 fragment built by default


yuzibo:
Can we merge the PR now?

yuzibo:
> Can we merge the PR now?

In fact, I mean the PR does not block to build Eclipse on riscv64(maybe), but we are blocked by other errors. Building this again to see the reslut.

HannesWell:
> > Can we merge the PR now?
> 
> In fact, I mean the PR does not block to build Eclipse on riscv64(maybe), but we are blocked by other errors. Building this again to see the reslut.

This can be merged as soon as https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2394 is submitted.
Only with both changes the SWT (and Equinox) native artifacts end up in the p2 repository and can be fetched by builds for other repositories, which would fail if only this is submitted.

HannesWell (approved):
Looks good, Thank you.

Now that https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2394 is about to be submitted, this can be submitted too.

PR #1478, opened 2024-09-20
Remove obsolete API filter for GCData on MacOS

HeikoKlare:
Since recent PDE and Tycho changes, the API filters for GCData on MacOS are not necessary anymore and produce warnings due to being obsolete. This removes the according filters.

image

PR #1488, opened 2024-09-25
Switch back GitHub Actions workflow to use current Tycho version

HeikoKlare:
The Tycho version used in GitHub Actions workflow has been temporarily reverted to 4.0.8. This change switched back the Tycho version defined in the Tycho build itself.

Since recent PDE and Tycho changes that apply with this change of the Tycho version, the API filters for GCData and Display on MacOS are not necessary anymore and produce warnings due to being obsolete. Thus, this change also removes the according filters. The commit has been taken from as it semantically belongs to the Tycho version change (and is required to make the CI builds run successfully):
- #1478

<s>⚠️ This PR is supposed to be merged once the issue discussed in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2360 and reproduced via https://github.com/eclipse-tycho/tycho/pull/4293 has been resolved or once we want to have SWT use latest Tycho again to further investigate the failures.</s>

Supercedes and thus closes #1478

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -71,7 +71,6 @@ jobs:
         --fail-at-end
         -DskipNativeTests=false
         -DfailIfNoTests=false
-        '-Dtycho.version=4.0.8'
         clean install
     - name: Performance tests
       if: contains(github.event.pull_request.labels.*.name, 'performance')
--- a/binaries/org.eclipse.swt.cocoa.macosx.aarch64/.settings/.api_filters
+++ b/binaries/org.eclipse.swt.cocoa.macosx.aarch64/.settings/.api_filters
@@ -248,105 +248,11 @@
             </message_arguments>
         </filter>
     </resource>
-    <resource path="Eclipse SWT/cocoa/org/eclipse/swt/graphics/GCData.java" type="org.eclipse.swt.graphics.GCData">
-        <filter id="338940029">
-            <message_arguments>
-                <message_argument value="org.eclipse.swt.graphics.GCData"/>
-                <message_argument value="bg"/>
-            </message_arguments>
-        </filter>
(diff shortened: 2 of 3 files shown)


HeikoKlare:
I've experimentally changed this PR to use Tycho 4.0.10-SNAPSHOT, in order to validate that the reason for temporarily switching back to Tycho 4.0.8 (failures during testCompile on Linux/MacOS) has been resolved.

The builds now successfully pass that compilation step (@laeubi fyi).
They still fail, but for different, unrelated reasons:
- Linux: #1564
- macOS: #1478

I guess there are no plans for releasing Tycho 4.0.10 soon? So would it make sense to upgrade the whole platform to use 4.0.10-SNAPSHOT, or should we maybe at least apply it to the GH actions builds of SWT?

sratz:
> I guess there are no plans for releasing Tycho 4.0.10 soon?

There are: https://github.com/eclipse-tycho/tycho/discussions/2455#discussioncomment-11140479

laeubi:
> I guess there are no plans for releasing Tycho 4.0.10 soon? So would it make sense to upgrade the whole platform to use 4.0.10-SNAPSHOT, or should we maybe at least apply it to the GH actions builds of SWT?

We should use 4.0.10-SNAPSHOT for platform to getting more confident that there are no regressions, so if you like just go ahead and upgrade the aggregator for a while, I'll then take care to release the version before the code freeze.

HeikoKlare:
> We should use 4.0.10-SNAPSHOT for platform to getting more confident that there are no regressions, so if you like just go ahead and upgrade the aggregator for a while, I'll then take care to release the version before the code freeze.

Alright, done: https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2520

PR #1474, opened 2024-09-18
Build gtk.linux.riscv64 native SWT fragment

Sakura286:
With command

```shell
mvn verify -DskipTests=true -Dnative=gtk.linux.riscv64
```

riscv64 artifact can be generated at `binaries/org.eclipse.swt.gtk.linux.riscv64/target/`

PR #1490, opened 2024-09-25
Bump riscv64 version

Sakura286:
This PR upgrades riscv64 version from `3.127.100` to `3.128.0`


HannesWell:
Thank you. That was missing.
I added this commit to https://github.com/eclipse-platform/eclipse.platform.swt/pull/1474 so that we have a clean version when starting to built this fragment.

PR #1491, opened 2024-09-25
Set up workflow for automated version increments in pull-requests 

HannesWell:
Set up for this repository the workflow for automated version increments in pull-requests added via https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2352

Before this can work complete, the EF infra team has to add the required PAT of the Platform-bot to this repo as described in the linked PR and requested in
- https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/5020#note_2825941

Additionally this also adds the call of the checkMergeCommits workflow.


HannesWell:
Looks like the configuration of the `tycho-p2-extras-plugin:compare-version-with-baselines` is too strict and has to be adapted.
I'll look into this later.

HannesWell:
> Looks like the configuration of the `tycho-p2-extras-plugin:compare-version-with-baselines` is too strict and has to be adapted.

What ever was causing this problem, it doesn't happen anymore. Therefore I'm submitting this so we have the automated version-increment also in SWT.
But lets keep an eye on it, if there are are unexpected errors.

PR #1492, opened 2024-09-25
[Build] Replace use of coactions/setup-xvfb

HannesWell:
Similar to https://github.com/eclipse-equinox/equinox/pull/677.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -68,7 +68,6 @@ jobs:
         -DforkCount=1
         -Dnative=${{ matrix.config.native }}
         -Papi-check
-        -Dcompare-version-with-baselines.skip=true
         -Dtycho.baseline.replace=none
         --fail-at-end
         -DskipNativeTests=false
@@ -83,7 +82,6 @@ jobs:
        run: >-
         mvn --batch-mode -V -U -e
         -DforkCount=1
-        -Dcompare-version-with-baselines.skip=true
         --fail-at-end
         -DskipNativeTests=true
         -DfailIfNoTests=true
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -60,26 +60,24 @@ jobs:
       with:
         maven-version: 3.9.9
     - name: Build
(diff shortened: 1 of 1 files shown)


HannesWell:
The tests fail on Linux although XVFB is now available by default on the Linux runners:
https://github.com/actions/runner-images/blob/ubuntu22/20240922.1/images/ubuntu/Ubuntu2204-Readme.md

HannesWell:
@akurtakov do you have any idea why this fails with the preinstalled `xvfb` on `gtk.linux.x86_64-java17`?
According to the list of installed software it's available:
https://github.com/actions/runner-images/blob/ubuntu22/20240922.1/images/ubuntu/Ubuntu2204-Readme.md

Could there be a version problem?

laeubi:
See my comment here
- https://github.com/coactions/setup-xvfb/issues/21#issuecomment-2374498905

`xvfb` is *installed* that does not mean it works by simply doing nothing.... one has to use `xvfb-run <command>` (of course only on linux) and somehow make it work/skip on windows, so I think the calim is wrong that the action is no longer needed, it would just not be needed to *install* xvfb.

HannesWell:
> `xvfb` is _installed_ that does not mean it works by simply doing nothing.... one has to use `xvfb-run <command>` (of course only on linux) and somehow make it work/skip on windows, so I think the calim is wrong that the action is no longer needed, it would just not be needed to _install_ xvfb.

Now that you have said it totally makes sense and it's obvious that it wasn't working with the initial attempt 🤦🏽‍♂️. Thanks for pointing this.

In order to just prepend the xvfb-run command only on Linux I have used the 'ternay' operator for the GH action syntax:
`${{ startsWith(matrix.config.native, 'gtk') && 'xvfb-run' || '' }}`

With that the Linux build now succeeds again without any test-failure.

HannesWell:
Finally also the Windows workflow passes all tests. For some reasons, when using the `CMD` shell some tests fail. So I quoted all parameters that contain a dot so that powershell (the default shell for Windows) does not interpret them.

HannesWell:
Test failures in Jenkins are unrelated, submitting.

PR #1497, opened 2024-09-26
Refresh entire workspace after Maven launch to build SWT binaries

HannesWell:
This simplifies maintenance since the list of refreshed projects does not have to be kept in sync with the list of available native fragments.

It's already out-of-sync it wasn't updated when the Windows on ARM and Linux on RISC-V fragment were added recently.


HannesWell:
Test failure is unrelated.

PR #1506, opened 2024-10-07
[win32] Use DPI dependent OS API calls

akoch-yatta:
This commit replaces the OS calls for OpenThemeData with calls to the dpi dependent equivalent OpenThemeDataForDpi. Therefor the handling of loading/unloading of theme in Display is refactored to be able to manage multiple DPI dependent variants of a theme in multi zoom environments. Without this change some widget can look inconsistent in a multi-zoom scenario on the non-primary monitor. 

One example that is affected by that change can be found in the Manifest editor
Without PR
image
With PR
image


jukzi:
is it possible to add a junit test, that shows/tests the changed behaviour?

akoch-yatta:
> is it possible to add a junit test, that shows/tests the changed behaviour?

Hmm, depends a bit on what the test should achieve. A test for a realistic scenario would need two monitors with different zoom, which is not possible in the acccessible test environment. And even then, the usage of the replaced calls is only in drawing callback, so currently I don't see any helpful automated test here, we could add.  But I am of course open for ideas.

bundles/org.eclipse.swt/Eclipse SWT PI/win32/org/eclipse/swt/internal/win32/OS.java
```diff
@@ -4581,4 +4581,12 @@ public static int HRESULT_FROM_WIN32(int x) {
 public static final native boolean DuplicateHandle(long hSourceProcessHandle, long hSourceHandle, long hTargetProcessHandle,
 		long [] lpTargetHandle, int dwDesiredAccess, boolean b, int dwOptions);
 
+
+public static long OpenThemeData(long hwnd, char[] themeName, int dpi) {
+	if (OS.WIN32_BUILD >= OS.WIN32_BUILD_WIN10_1809) {
+		return OS.OpenThemeDataForDpi(hwnd, themeName, dpi);
+	} else {
+		return OS.OpenThemeData(hwnd, themeName);
```
fedejeanne, line 4589: Since these 2 methods are now replaced by your new method, it would be worth declaring them `private`.

fedejeanne (commented):
Changes look good and the fix work.

I only have a minor comment regarding making the replaced methods `private`.

Other than that, this PR is good to go on my account.

fedejeanne:
> Other than that, this PR is good to go on my account.

I mean of course after Jörg's comments have been addressed.

akoch-yatta:
> May be some test that checks drawing for single monitor with different dpi settings works as expected. Would that make sense, or are there already such tests? At least a snippet to reproduce the screenshot given in [#1506 (comment)](https://github.com/eclipse-platform/eclipse.platform.swt/pull/1506#issue-2569723414) to demonstrate the change in a manual test should be possible.

I added Snippet383 to test that behavior for Button, Tree & Table.
image

An automated test for that would be quite difficult, because the changes affect internal custom drawing behavior which is hard to test. I would only see something like comparing screenshots for that use case.

jukzi:
i confirm the snippet looks better with the Patch, even though it looks a bit different for me:
image

jukzi (approved):

fedejeanne (approved):
Ignoring random failing tests: https://github.com/eclipse-platform/eclipse.platform.swt/issues/1523

PR #1519, opened 2024-10-09
Make new GTK4 Dialog dynamic loaded for pre GTK 4.10

laeubi:
This is a draft to make the GTK4.10 dialog compile with older GTK versions

FYI @ptziegler @akurtakov 

this is not complete but in case someone wants to pick it up I wanted to share my progress here, with this changes it actually compiles again now.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT PI/gtk/library/gtk4.h
+++ b/bundles/org.eclipse.swt/Eclipse SWT PI/gtk/library/gtk4.h
@@ -22,4 +22,7 @@
 
 #include <gtk/gtk.h>
 
+// Hard-link code generated from GTK4.java to LIB_GTK
+#define GTK4_LOAD_FUNCTION(var, name) LOAD_FUNCTION_LIB(var, LIB_GTK, name)
+
 #endif /* INC_gtk4_H */
--- a/bundles/org.eclipse.swt/Eclipse SWT PI/gtk/library/gtk4.c
+++ b/bundles/org.eclipse.swt/Eclipse SWT PI/gtk/library/gtk4.c
@@ -662,7 +662,15 @@ JNIEXPORT jlong JNICALL GTK4_NATIVE(gtk_1file_1dialog_1get_1default_1filter)
 {
 	jlong rc = 0;
 	GTK4_NATIVE_ENTER(env, that, gtk_1file_1dialog_1get_1default_1filter_FUNC);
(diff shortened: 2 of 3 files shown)


bundles/org.eclipse.swt/Eclipse SWT PI/gtk/library/gtk4.h
```diff
@@ -22,4 +22,12 @@
 
 #include <gtk/gtk.h>
 
+// Hard-link code generated from GTK4.java to LIB_GTK
+#define GTK4_LOAD_FUNCTION(var, name) LOAD_FUNCTION_LIB(var, LIB_GTK, name)
+
+#ifndef GtkFileDialog
```
akurtakov, line 28: You should not need this define, the correct thing to do is to remove the @param like with cast=(GtkFileDialog *) and it will be dynamically figured (no type cast).

laeubi: done

akurtakov (approved):
I'm giving it +1 as it will allow compiling on 4.8 (as debian stable has) although runtime requirement will be 4.10+. Having newer than that build requirement is not feasible as it will introduce even newer glibc and etc. .

akurtakov:
Let's merge this one to allow compiling on debian image. We may need to bump https://github.com/eclipse-platform/eclipse.platform.swt/blob/e329edcf5ca69830fa6816f14958a558ea17e2cf/bundles/org.eclipse.swt/Eclipse%20SWT/gtk/org/eclipse/swt/widgets/Display.java#L526 to 4.10 which would mean that build machine wouldn't be able to smoke test things but as long as ubuntu-latest can - this is fine IMO. Also I believe that by the time gtk 4 is ready to be default 4.10 would be old enough (4.16 being the latest now). With the above said - if you agree do the change now but considering there is no easy way  to even try it now, this change can be done later if/when people start finding it a problem.

ptziegler:
Apologies for the late response. Generally, I agree with @akurtakov. I don't think the current GTK4 port is anywhere near a productive state and by the time it becomes stable, the bottom line of 4.10 will already be quite old.

This doesn't mean that we should always use the latest API, but at least for something that affects the UI like the dialogs, I believe that it is less of a headache to migrate it now, before it becomes widely used.

And when GTK4 becomes the default backend, I doubt that SWT will immediately drop GTK3, leaving it as an usable alternative for older platforms, at least for a few more releases.

PR #1522, opened 2024-10-09
Widget.getData: fix NPE

jukzi:
fix java.lang.NullPointerException: Cannot read the array length because "table" is null

As Widget.setData(Object) can be called with null argument even if KEYED_DATA is set.


jukzi:
ignoring random fail https://github.com/eclipse-platform/eclipse.platform.swt/issues/1523

PR #1527, opened 2024-10-10
Images not aligned properly in Expand Bar

ShahzaibIbrahim:
When adding image larger than the header size in expand bar, it was drawn outside of the bar on the top. This was due to wrong calculation and mixing of points and pixels. 

## HOW TO TEST

- Go to ExpandBarTab class and set any larger image on item

```
-Dswt.autoScale=quarter
```
- Switch to the tab "Expand Items"
- See if the image is within the bounds of the bar/header

## SCREENSHOTS

**Before Fix:**
image

**After Fix:**
image

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/ExpandItem.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/ExpandItem.java
@@ -43,6 +43,7 @@ public class ExpandItem extends Item {
 	boolean expanded, hover;
 	int x, y, width, height;
 	int imageHeight, imageWidth;
+	private static final int IMAGE_MARGIN = 8;
 	static final int TEXT_INSET = 6;
 	static final int BORDER = 1;
 	static final int CHEVRON_SIZE = 24;
@@ -183,9 +184,13 @@ private void drawChevron (long hDC, RECT rect) {
 
 void drawItem (GC gc, long hTheme, RECT clipRect, boolean drawFocus) {
 	long hDC = gc.handle;
-	int headerHeight = parent.getBandHeight ();
+	int headerHeightinPixels = getHeaderHeightInPixels();
+	int zoom = getZoom();
+	int imageHeightInPixels = DPIUtil.scaleUp(imageHeight, zoom);
+	int imageWidthInPixels = DPIUtil.scaleUp(imageWidth, zoom);
+
 	RECT rect = new RECT ();
-	OS.SetRect (rect, x, y, x + width, y + headerHeight);
+	OS.SetRect (rect, x, y, x + width, y + headerHeightinPixels);
 	if (hTheme != 0) {
 		OS.DrawThemeBackground (hTheme, hDC, OS.EBP_NORMALGROUPHEAD, 0, rect, clipRect);
 	} else {
@@ -194,14 +199,10 @@ void drawItem (GC gc, long hTheme, RECT clipRect, boolean drawFocus) {
 		OS.SelectObject (hDC, oldBrush);
 	}
 	if (image != null) {
-		int zoom = getZoom();
 		rect.left += ExpandItem.TEXT_INSET;
-		if (imageHeight > headerHeight) {
-			gc.drawImage (image, DPIUtil.scaleDown(rect.left, zoom), DPIUtil.scaleDown(rect.top + headerHeight - imageHeight, zoom));
-		} else {
-			gc.drawImage (image, DPIUtil.scaleDown(rect.left, zoom), DPIUtil.scaleDown(rect.top + (headerHeight - imageHeight) / 2, zoom));
-		}
-		rect.left += imageWidth;
+		int yInPoints = DPIUtil.scaleDown(rect.top + ((headerHeightinPixels - imageHeightInPixels) / 2), zoom);
+		gc.drawImage (image, DPIUtil.scaleDown(rect.left, zoom), yInPoints);
+		rect.left += imageWidthInPixels;
 	}
 	if (text.length () > 0) {
 		rect.left += ExpandItem.TEXT_INSET;
@@ -213,8 +214,7 @@ void drawItem (GC gc, long hTheme, RECT clipRect, boolean drawFocus) {
 			} else {
 				buffer = (RLE + text).toCharArray ();
 			}
-		}
-		else {
+		} else {
 			buffer = text.toCharArray ();
 		}
 		if (hTheme != 0) {
@@ -227,7 +227,7 @@ void drawItem (GC gc, long hTheme, RECT clipRect, boolean drawFocus) {
 	}
 	int chevronSize = ExpandItem.CHEVRON_SIZE;
 	rect.left = rect.right - chevronSize;
-	rect.top = y + (headerHeight - chevronSize) / 2;
+	rect.top = y;
 	rect.bottom = rect.top + chevronSize;
 	if (hTheme != 0) {
 		int partID = expanded ? OS.EBP_NORMALGROUPCOLLAPSE : OS.EBP_NORMALGROUPEXPAND;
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/ExpandItem.java
```diff
@@ -183,9 +183,14 @@ private void drawChevron (long hDC, RECT rect) {
...
-	int headerHeight = parent.getBandHeight ();
+	int headerHeightinPixels = getHeaderHeightInPixels();
+	int zoom = getZoom();
+	int imageHeightInPixels = DPIUtil.scaleUp(imageHeight, zoom);
+	int imageWidthInPixels = DPIUtil.scaleUp(imageWidth, zoom);
+
 	RECT rect = new RECT ();
-	OS.SetRect (rect, x, y, x + width, y + headerHeight);
+	OS.SetRect (rect, x, y, x + width, y + headerHeightinPixels);
+	//System.out.println("y = " + y + " Height = " + headerHeightinPixels + " Total = " + (y + headerHeightinPixels));
```
fedejeanne, line 193: Remove comment
```suggestion
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/ExpandItem.java
```diff
@@ -310,7 +310,13 @@ public int getHeaderHeight () {
 }
 
 int getHeaderHeightInPixels () {
-	return Math.max (parent.getBandHeight (), imageHeight);
+	int headerHeightInPixels = parent.getBandHeight ();
+	int zoom = getZoom();
```
fedejeanne, line 314: Inline

fedejeanne (changes requested):
The code looks quite good and it's easy to understand (just one minor nit-pick), but I tested it and found an issue.

## How to reproduce
- Change the **second** image of `org.eclipse.swt.examples.controlexample.ControlExample.imageLocations` from `openFolder.gif` to some other image that is way bigger (_e.g._ 300 x 300 or more)
- Run the `ControlsExample` and go to the `ExpandBar`
- Expand the upper section: "_What is your favorite button?_"

The text of the second section (1) and its image (2) are still in the same place as before but they should be moved a bit lower. 

image

I can't reproduce this error on `master`, I mean the text is still **too** low but it will be moved accordingly when expanding/collapsing the first section:

image

examples/org.eclipse.swt.examples/src/org/eclipse/swt/examples/controlexample/ControlExample.java
```diff
@@ -44,7 +44,7 @@ public class ControlExample {
 	static final int ciClosedFolder = 0, ciOpenFolder = 1, ciTarget = 2, ciBackground = 3, ciParentBackground = 4;
 	static final String[] imageLocations = {
 		"closedFolder.gif", 			//$NON-NLS-1$
-		"openFolder.gif", 				//$NON-NLS-1$
```
fedejeanne, line 47: Revert

examples/org.eclipse.swt.examples/src/org/eclipse/swt/examples/controlexample/ExpandBarTab.java
```diff
@@ -80,6 +80,7 @@ void createExampleWidgets () {
 		item.setHeight(composite.computeSize(SWT.DEFAULT, SWT.DEFAULT).y);
 		item.setControl(composite);
 		item.setImage(instance.images[ControlExample.ciClosedFolder]);
+		item.setExpanded(true);
```
fedejeanne, line 83: Revert

fedejeanne (commented):
Thank you for fixing the issue, @ShahzaibIbrahim ! I tested again and I only have 2 comments about changes that need to be reverted. 

I would also change the first line of the commit text to `Properly align big images in ExpandBar`.

Once that is fixed, the PR is good to go on my account.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/ExpandItem.java
```diff
@@ -310,7 +309,12 @@ public int getHeaderHeight () {
 }
 
 int getHeaderHeightInPixels () {
-	return Math.max (parent.getBandHeight (), imageHeight);
+	int headerHeightInPixels = parent.getBandHeight ();
+	int imageHeightInPixels = DPIUtil.scaleUp(imageHeight, getZoom());
+	if(imageHeightInPixels > headerHeightInPixels) {
+		headerHeightInPixels = imageHeightInPixels + headerHeightInPixels;
```
HeikoKlare, line 315: Is this really intended? It looks wrong to me that in case the image is higher than the band of the expand bar, the header his is _the sum_ of these two. Shouldn't it just be the image height then?

I guess that's also the reason why there is some unexpected margin at the bottom and top of the image within the header when the image is larger than the header (text) itself:
image

When "fixing" this, it looks like this:
image

I do not want to say that this is better / intended, but just adding the `headerHeightInPixels` for a margin seems wrong to me. Wouldn't it be better to define some explicit margin that is added to the image height to ensure that the image does not start/end at the bounds of the header?

examples/org.eclipse.swt.examples/src/org/eclipse/swt/examples/controlexample/ExpandBarTab.java
HeikoKlare, line 83: Why has this been added and removed from at some other place? Seems like an erroneous revert to me.

HeikoKlare (changes requested):

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/ExpandItem.java
HeikoKlare: You should test to see what also looks good in particular in border cases. Keep in mind that you should not apply the margin once the image height exceeds the header height but that you should ensure that the margin is always present, even if the image fits into the ordinary header height. Otherwise, an image with the height of the header will have no margin and an image with height+1 of the header will have margin of 10. So I would test what happens with according values for the margin in these border cases.

ShahzaibIbrahim: I have decided on 8 to be the best margin. Also I have handled the case to have the margin of at least 8 when the image height is equal or closer to the header height.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/ExpandItem.java
```diff
@@ -310,7 +310,15 @@ public int getHeaderHeight () {
...
 int getHeaderHeightInPixels () {
-	return Math.max (parent.getBandHeight (), imageHeight);
+	int headerHeightInPixels = parent.getBandHeight ();
+	int imageHeightInPixels = DPIUtil.scaleUp(imageHeight, getZoom());
+	int imageHeaderDiff = headerHeightInPixels - imageHeightInPixels;
+	if(imageHeaderDiff < 0) {
+		headerHeightInPixels = imageHeightInPixels + IMAGE_MARGIN;
+	} else if(imageHeaderDiff < IMAGE_MARGIN) {
+		headerHeightInPixels += (IMAGE_MARGIN - imageHeaderDiff);
+	}
```
HeikoKlare, line 320: Isnt' this equivalent to ... ?
```suggestion
	if(imageHeaderDiff < IMAGE_MARGIN) {
		headerHeightInPixels = imageHeightInPixels + IMAGE_MARGIN;
	}
```
Please also correct the code formatting here.

HeikoKlare (commented):
The recent changes have properly addressed my comment. In my opinion, the selected margin looks good:
image

And adding this margin does also not change exististing layout (i.e., does not unnecessarily introduce additional margins), at least when taking a look at the ControlExample (left is existing, right is with this PR):
image

I have just one remaining comment.

HeikoKlare (approved):

HeikoKlare:
Test failures unrelated and documented: #1564

PR #1528, opened 2024-10-10
Expand Bar is not scaled during DPI change

ShahzaibIbrahim:
When moving from one zoom level monitor to another zoom level (no matter what %), The expanded items and texts are not drawn at correct coordinates.

## HOW TO TEST
```
-Dswt.autoScale=quarter
```
- Switch to the tab "Expand Items"
- Move the window from 200 to 100 zoom level monitor (Producible in any combination of zoom levels)
- See if the "expanded" items are aligned to correct coordinates

## SCREENSHOTS

**Before Fix:**
image

**After Fix:**
image

## Requires
* [x] https://github.com/eclipse-platform/eclipse.platform.swt/pull/1527

Contributes to https://github.com/eclipse-platform/eclipse.platform.swt/issues/62 and https://github.com/eclipse-platform/eclipse.platform.swt/issues/127

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/ExpandItem.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/ExpandItem.java
@@ -539,7 +539,7 @@ private static void handleDPIChange(Widget widget, int newZoom, float scalingFac
 	if (item.height != 0 || item.width != 0) {
 		int newWidth = Math.round(item.width * scalingFactor);
 		int newHeight = Math.round(item.height * scalingFactor);
-		item.setBoundsInPixels(item.x, item.y, newWidth, newHeight, false, true);
+		item.setBoundsInPixels(item.x, item.y, newWidth, newHeight, true, true);
 	}
 }
 }
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/ExpandBar.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/ExpandBar.java
@@ -878,6 +878,7 @@ private static void handleDPIChange(Widget widget, int newZoom, float scalingFac
 	for (ExpandItem item : expandBar.getItems()) {
 		DPIZoomChangeRegistry.applyChange(item, newZoom, scalingFactor);
 	}
+	expandBar.layoutItems(0, true);
 	expandBar.redraw();
 }
 }


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/ExpandItem.java
```diff
@@ -539,7 +539,8 @@ private static void handleDPIChange(Widget widget, int newZoom, float scalingFac
 	if (item.height != 0 || item.width != 0) {
 		int newWidth = Math.round(item.width * scalingFactor);
 		int newHeight = Math.round(item.height * scalingFactor);
-		item.setBoundsInPixels(item.x, item.y, newWidth, newHeight, false, true);
+		item.setBoundsInPixels(item.x, item.y, newWidth, newHeight, true, true);
+		item.parent.layoutItems(0, true);
```
HeikoKlare, line 543: Why can't we execute this in `ExpandBar#handleDPIChange()`, so that it is only executed once per expand bar instead of updating all updated n times (with n being the number items)?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/ExpandItem.java
```diff
@@ -539,7 +539,8 @@ private static void handleDPIChange(Widget widget, int newZoom, float scalingFac
 	if (item.height != 0 || item.width != 0) {
 		int newWidth = Math.round(item.width * scalingFactor);
```
HeikoKlare, line 540: When testing this PR I found that this calculation only work with scale mode `quarter`. When using `integer200`, the `scalingFactor` is always 1, such that items are not scaled. It then looks like this:
image

Using `scalingFactor` does not seem to be the right thing here. @akoch-yatta do you know how we could fix tihs?

ShahzaibIbrahim: The scaling was broken even before and this change doesn't bring any regression. From my understanding we only care about int200 scale mode only if we are regressing something, should we try to fix it for int200?

HeikoKlare: Yes, that's not a regression of this PR. Still, the rescaling behavior will still be broken after this PR due to the mentioned bug. Currently, `int200` is even the more important case, since it's the default value and will be used with the experimental option for monitor-specific rescaling that we have provided. In the long term, the `quarter` option should of course be the default one, but for now it's not.

So it's fine for me to only address the repositioning in this PR and keep the resizing for later. We should just make sure that we have an issue to track that and address it as a follow-up task.

ShahzaibIbrahim: I have created a new issue in our backlog for resizing in int200 case, https://github.com/orgs/vi-eclipse/projects/1/views/1?pane=issue&itemId=86212846

HeikoKlare:
For failing tests see: #1564

Issue #1518, opened 2024-10-08
ScreenshotOnFailure doesn't work


jukzi:
the ScreenshotOnFailure TestWatcher produces only black images on failed tests because 
the `@After` `org.eclipse.swt.tests.junit.Test_org_eclipse_swt_widgets_Widget.tearDown()` calls `shell.dispose();` before Screenshot is taken

reproducable with the failing Test_org_eclipse_swt_custom_StyledText

jukzi:
see https://stackoverflow.com/questions/18008241/run-a-testwatcher-before-the-after

basilevs:
[Please configure Jenkins agent to run on an unlocked graphical Windows session](https://stackoverflow.com/questions/19441324/how-to-run-gui-tests-on-a-jenkins-windows-slave-without-remote-desktop-connectio?noredirect=1&lq=1)

basilevs:
@jukzi is this reproducible locally? Does it happen on Windows only?

jukzi:
> if some @After/https://github.com/before method fails or a rule such as
 @ExpectedException fails the test. In those cases, the behavior is
 unchanged: these screenshots are still taken after @after has run.

@tomaswolf thanks, i like that idea.

jukzi:
[SWTBotJunit4ClassRunner](https://github.com/eclipse-swtbot/org.eclipse.swtbot/blob/ed2034a65915f6095d3c2464ac43068b105ebd68/org.eclipse.swtbot.junit4_x/src/org/eclipse/swtbot/swt/finder/junit/SWTBotJunit4ClassRunner.java)
looks a bit like TracingSuite already used in AllBrowserTests. Maybe that TracingSuite could be enhanced to take screenshots also on fail and be used in other suits as well.

iloveeclipse:
@jukzi : some of the recent changes in this area broke a lot of UI & SWT tests.

Typical stack from https://download.eclipse.org/eclipse/downloads/drops4/I20241013-1800/testresults/html/org.eclipse.swt.tests_ep434I-unit-cen64-gtk3-java17_linux.gtk.x86_64_17.html

```
Cannot invoke "java.util.function.Supplier.get()" because "this.shell" is null

java.lang.NullPointerException: Cannot invoke "java.util.function.Supplier.get()" because "this.shell" is null
at org.eclipse.test.Screenshots$ScreenshotOnFailure.dispose(Screenshots.java:52)
[...]

PR #1532, opened 2024-10-11
abstract Widget tests: take screenshot before dispose

jukzi:
https://github.com/eclipse-platform/eclipse.platform.swt/issues/1518


jukzi:
needs https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2446

PR #1533, opened 2024-10-11
Revert "Extend Browser API to check for custom text location"

amartya4256:
This reverts commit 51b8d8f655c34c9a536a1694c2f8e6273c3c6d76. The API isLocationForCustomText is not needed for Browsers.

contributes to #213

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Browser/common/org/eclipse/swt/browser/Browser.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Browser/common/org/eclipse/swt/browser/Browser.java
@@ -13,9 +13,6 @@
  *******************************************************************************/
 package org.eclipse.swt.browser;
 
-import java.net.*;
-import java.util.*;
-
 import org.eclipse.swt.*;
 import org.eclipse.swt.widgets.*;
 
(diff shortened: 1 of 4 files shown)


amartya4256:
@HeikoKlare Do we need to do another version bump?

sratz:
This reverts public API introduced in the last release. It is not very likely anyone has adopted that yet but I don't think we can simply revert this now.

In fact, the API tools should suggest a major version bump for this change - not sure why this is not the case (or I am just missing it?)

HeikoKlare:
It has been introduced after the last release for M1, so it should be safe to revert now, shouldn't it?

sratz:
> It has been introduced after the last release for M1, so it should be safe to revert now, shouldn't it?

Ah, you're right. Yes, then it's perfectly fine to revert now without a version bump.

And yes, we should do so as early as possible to get this into M2.

sratz (approved):

HeikoKlare:
3 new issues of type `The method onFailure() from the type Screenshots is deprecated` introduced by method deprecation in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2446, so unrelated to this PR. Only occurs because no SWT master build as new reference happened since then.

HeikoKlare:
[Recent master build](https://ci.eclipse.org/releng/job/eclipse.platform.swt/job/master/762/) now shows the same number of issues than in this PR, but Jenkins still compares builds for this PR against previous master build:
{38C0DE4E-E7C1-431A-A7D0-1DCACB79ABFC}

PR #1536, opened 2024-10-14
set Location of menu wrt Display coordinate system

amartya4256:
This PR contributes to the following regression:

In a multi zoom setup (mine was left (primary 200%), right (secondary 175%) there is a regression (both with and without update on autoscale) in menus.
We need to treat Menus similar to Shells -> they are always positioned with an absolute value. Currently they are set with two different ways:

1. Old zoomed down location, e.g. in Widget::showMenu -> this still works
2. New zoomed location, e.g. Push menus like the Run configurations -> this will result in

image

however, the locations supplied to these methods are in display coordinate system and hence must be scaled with the right methods as reflected in the PR.


fedejeanne (approved):

fedejeanne:
Failing tests are unrelated

PR #1545, opened 2024-10-18
[PoC] Edge: Mouse events on DOM -> SWT mouse events

### LINKED ISSUE #1540: Edge Browser: Replace BrowserInformationControl dosn't work on hover
**Describe the bug**
When using Edge Browser as the default browser, the BrowserInformationControl for the JavaDoc hover tooltip doesn't replace with navihation buttons on hovering.

**To Reproduce**
- Use Edge browser as the default browser.
- Hover over a piece of code to open a javadoc tooltip.
- Hover over the tooltip.
- The navigation buttons won't appear and it wont be possible to navigate if you click on a link.
- The BrowserInformationControl is only replace if you hover to the bottom of the tooltip where the ends in the bottom. It is triggered by the styledText in the bottom and not the browser.

**Expected behavior**
The BrowserInformationControl should be replaced with the ones with navigation buttons with any action on the tooltip.

Note: With WebView2, it doesn't seem possible to capture hover events and one possible proposal could be to trigger this event on Mouse Click and other Mouse Events which can be captured.

**Screenshots**

**Environment:**
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 

3. JRE/JDK version
 
**Version since**
 Eclipse or SWT version since when the behavior is seen [e.g. 4.23]

Any known workarounds for the problem?

sratz:
Would resolve #1540.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
@@ -641,6 +641,76 @@ void setupBrowser(int hr, long pv) {
 	webView.AddHostObjectToScript("swt\0".toCharArray(), hostObj);
 	hostDisp.Release();
 
+	class MouseNavigationFunction extends BrowserFunction {
+		public MouseNavigationFunction (Browser browser, String name) {
+			super (browser, name);
+		}
+		@Override
+		public Object function (Object[] arguments) {
+			System.out.println(Arrays.toString(arguments));
+			Event newEvent = new Event ();
+			newEvent.widget = browser;
+			String eventType = (String) arguments[0];
+			int x = (int) Math.round((Double) arguments[1]);
+			int y= (int)  Math.round((Double) arguments[2]);
+			int button = (int) Math.round((Double) arguments[3]);
+
+			if (OS.GetKeyState (OS.VK_MENU) < 0) newEvent.stateMask |= SWT.ALT;
+			if (OS.GetKeyState (OS.VK_SHIFT) < 0) newEvent.stateMask |= SWT.SHIFT;
+			if (OS.GetKeyState (OS.VK_CONTROL) < 0) newEvent.stateMask |= SWT.CONTROL;
+
+			x =	DPIUtil.scaleUp(x, DPIUtil.getNativeDeviceZoom());
+			y = DPIUtil.scaleUp(y, DPIUtil.getNativeDeviceZoom());
+			x = DPIUtil.scaleDown(x, DPIUtil.getZoomForAutoscaleProperty(browser.getShell().nativeZoom)); //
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -641,6 +641,76 @@ void setupBrowser(int hr, long pv) {
 	webView.AddHostObjectToScript("swt\0".toCharArray(), hostObj);
 	hostDisp.Release();
 
+	class MouseNavigationFunction extends BrowserFunction {
+		public MouseNavigationFunction (Browser browser, String name) {
+			super (browser, name);
+		}
+		@Override
+		public Object function (Object[] arguments) {
+			System.out.println(Arrays.toString(arguments));
```
sratz, line 650: I receive the mouse events just fine, e.g. in the "internal web browser".

But for some reason I do not receive any events in the element info popup or the variable hover.

HeikoKlare: The mouse events only seem to be processed when then content is retrieved from an "ordinary" URL. Whenever you set contents via `setText` (such as in the Javadoc view, in the Java element info popup, in the help) it does not work. It does not depend on how the widget is embedded: in the help view, the first page is custom content (also with a custom context menu), which does not report mouse events, while when navigating to other entries in the help (like from the "see also" list), mouse events will be reported.

So it seems to be some problem of registering the browser function for "custom content" pages. Is it even possible to register browser functions to them?

sratz: That's what I thought as well at first, but it works perfectly fine with other `setText()`-based browser usages, so does the following example:

```java
public class Test {

		Shell shell = new Shell ();

		new Browser(shell, SWT.EDGE).setText("hi");

		shell.open ();
		while (!shell.isDisposed()) {
		}
		display.dispose ();
	}
}
```

I don't understand what the element info / hover does differently.

sratz: Oh, easy one... It's simply that those use `Browser.setJavascriptEnabled(false)`

sratz: https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/iwebview2settings#get_isscriptenabled:

> _This only affects scripts in the document; scripts injected with ExecuteScript will run even if script is disabled. It is true by default._

So we need to register those listeners on the `window`, not the `document`.

sratz: Hmm, no dice. Doesn't make a difference... I guess this script-based approach won't be possible universally :(

Issue #1523, opened 2024-10-09
Random "Too many leaked file descriptors" in Test_org_eclipse_swt_browser_Browser


jukzi:
for example

```
[...]

PR #1548, opened 2024-10-21
Allow to configure Edge userdata directory via Display data

amartya4256:
With this contribution, Edge browser uses the workspace directory for the userdata directory for the separation of the usage of userdata directory per workspace by the webview2Environment.

Depends on https://github.com/eclipse-platform/eclipse.platform.ui/pull/2434

contributes to https://github.com/eclipse-platform/eclipse.platform.swt/issues/1013

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
@@ -40,6 +40,7 @@ class Edge extends WebBrowser {
 
 	// Display.getData() keys
 	static final String APPLOCAL_DIR_KEY = "org.eclipse.swt.internal.win32.appLocalDir";
+	static final String EDGE_USER_DATA_FOLDER = "org.eclipse.swt.internal.win32.Edge.userDataFolder";
 	static final String EDGE_USE_DARK_PREFERED_COLOR_SCHEME = "org.eclipse.swt.internal.win32.Edge.useDarkPreferedColorScheme"; //$NON-NLS-1$
 
 	// System.getProperty() keys
@@ -473,13 +474,9 @@ WebViewEnvironment createEnvironment() {
 
 	// Gather customization properties
 	String browserDir = System.getProperty(BROWSER_DIR_PROP);
-	String dataDir = System.getProperty(DATA_DIR_PROP);
 	String browserArgs = System.getProperty(BROWSER_ARGS_PROP);
 	String language = System.getProperty(LANGUAGE_PROP);
-	if (dataDir == null) {
-		// WebView2 will append "\\EBWebView"
-		dataDir = (String)display.getData(APPLOCAL_DIR_KEY);
-	}
+	String dataDir = getDataDir(display);
 
 	// Initialize options
 	long pOpts = COM.CreateSwtWebView2Options();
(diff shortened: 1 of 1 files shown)


HeikoKlare:
> I'm not sure if **workspace** is the right place for user data of Edge browser that can be used on **workspace free** RCP applications.

I would rather propose to use the workspace directory whenever there is such a directory and in other cases use a different one (probably the ordinary user-data folder as used now).

In addition, SWT should not depend on the concept of a "workspace URL". There should rather be some "custom edge data folder", "display-specific edge data folder" or the like instead. Also, we cannot replace the usage of the existing system property, because people may still want to use that opportunity (in particular in plain SWT applicationS).

So what about using these values in decreasing preference:
- Use the value of the existing system property if provided?
- Use a value for some "custom edge data directory" defined as data in the display if specified (which can then be used by the Eclipse platform to define some specific value, may it workspace-dependent or not).
- Use the default value.

laeubi:
I'm not sure what the edge user data directory is actually storing, but from OSGi perspective there is

https://docs.osgi.org/javadoc/osgi.core/8.0.0/org/osgi/framework/Bundle.html#getDataFile-java.lang.String-

akurtakov:
> I'm not sure what the edge user data directory is actually storing, but from OSGi perspective there is
> 
> https://docs.osgi.org/javadoc/osgi.core/8.0.0/org/osgi/framework/Bundle.html#getDataFile-java.lang.String-

SWT should stay usable in non-OSGi environment too.

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -476,6 +477,9 @@ WebViewEnvironment createEnvironment() {
 	String dataDir = System.getProperty(DATA_DIR_PROP);
 	String browserArgs = System.getProperty(BROWSER_ARGS_PROP);
 	String language = System.getProperty(LANGUAGE_PROP);
+	if (dataDir == null) {
+		dataDir = (String) browser.getDisplay().getData(DISPLAY_USER_DATA_FOLDER);
+	}
```
fedejeanne, line 482: Extract the logic to calculate `dataDir` to its own method: `getDataDir(...)`.

Also, why use both `browser.getDisplay()` and `Display.getCurrent()` as the displays when fetching the data? (see lines below too)

laeubi:
> > I'm not sure what the edge user data directory is actually storing, but from OSGi perspective there is
> > https://docs.osgi.org/javadoc/osgi.core/8.0.0/org/osgi/framework/Bundle.html#getDataFile-java.lang.String-
> 
> SWT should stay usable in non-OSGi environment too.

As far as I understand the settings/property is setup by the one who creates the display, so at that point OSGI might or might not be used.

Beside that, such thing can even be crafted in a way that is opaque for non-OSGi users, e.g with having an activator that initializes a static field, in non-OSGi env this will simply not happen and any default can be used instead.

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
HeikoKlare: Fixed with 6509979.

HeikoKlare (approved):

fedejeanne (approved):
@HeikoKlare I extracted the method `getDataDir(Display)` and pushed the changes.

HeikoKlare:
Test failures are unrelated and already documented: 
- #1523

Issue #1122, opened 2024-03-18
Browser widgets using Edge flicker when resized causing flashes


kwinter108:
# Describe the bug
Browser widgets using the Edge browser flicker when resized. When included in a RCP Part and resized, the flicker can cause rapid flashing between a black background and the content displayed in the browser. The flashes created can exceed the threshold defined in [WCAG 2.1 - SC 2.3](https://www.w3.org/WAI/WCAG21/Understanding/seizures-and-physical-reactions) and may trigger seizures in individuals who have photosensitive seizure disorders.

# To Reproduce
## Option 1
1. Start the Eclipse application "_Runtime Workspace_" with this VM-Arg: `-Dorg.eclipse.swt.browser.DefaultType=Edge`
2. Resize the _JavaDoc_ view

### Video
**WARNING: This video may potentially trigger seizures for people with photosensitive epilepsy. Viewer discretion is advised.**
[flickering_javadoc_view_using_edge_as_browser](https://github.com/user-attachments/assets/9e7b42bc-d4ed-4ca6-b7c8-d0d1a8c10843)

## Option 2
1. Create a RCP application (the 'Create sample content' option can be used to speed up the remaining steps)
4. Create 2 Part Stacks in the same Part Sash Container.
5. Create a Part in each Part Stack and associate them to classes. One of the parts must contain a Browser widget using the SWT.EDGE style. Set the browser to any valid URL. The other Part can contain any widget as it only exists to allow the parts to be resized within the window.
6. Launch the application using the .product file.
7. Resize the application window.
8. Use the mouse to drag the boarder between the Parts to cause the Browser to resize.

### Video
**WARNING: This video may potentially trigger seizures for people with photosensitive epilepsy. Viewer discretion is advised.**
[Recording of issue](https://github.com/eclipse-platform/eclipse.platform.swt/assets/25162503/e71193f3-078a-4c16-a673-88c9306d481b)

# Expected behavior
Browser widgets using the Edge browser should resize smoothly without causing flickering.

# Environment
1. Select the platform(s) on which the behavior is seen:
*I have only been able to test this issue on a Windows machine. Given the nature of this issue, it is likely that it will occur in Linux and macOS as well*
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 

9. Additional OS info (e.g. OS version, Linux Desktop, etc)
+ Windows 10 Pro version 22H2
+ Windows 11 Pro version 23H2

[...]

PR #1551, opened 2024-10-22
Add processing of mouse movements in Edge browser

amartya4256:
This contribution captures the hover over edge browser and sends a mouse move event to its listener.

contributes to https://github.com/eclipse-platform/eclipse.platform.swt/issues/1540

## How to test
- Make sure your IDE is **not** using Edge as the default browser (at least until other issues have been solved)
- Start the _Runtime Workspace_ application with the start parameter `-Dorg.eclipse.swt.browser.DefaultType=Edge`
- Hover over some element (class name, method, etc) so that a JavaDoc pops up
- Move the cursor inside the JavaDoc pop-up --> The scrollbars and the navigation buttons should appear in the JavaDoc pop-up:

image

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
@@ -78,6 +78,9 @@ class Edge extends WebBrowser {
 	private boolean ignoreFocus;
 	private String lastCustomText;
 
+	private static record CursorPosition(Point location, boolean isInsideBrowser) {};
+	private CursorPosition previousCursorPosition = new CursorPosition(new Point(0, 0), false);
+
 	static {
 		NativeClearSessions = () -> {
 			ICoreWebView2CookieManager manager = getCookieManager();
@@ -654,6 +657,7 @@ void setupBrowser(int hr, long pv) {
 	browser.addListener(SWT.FocusIn, this::browserFocusIn);
 	browser.addListener(SWT.Resize, this::browserResize);
 	browser.addListener(SWT.Move, this::browserMove);
+	scheduleMouseMovementHandling();
 
 	containingEnvironment.instances().add(this);
 	// Sometimes when the shell of the browser is opened before the browser is
@@ -707,6 +711,52 @@ void browserResize(Event event) {
 	OS.GetClientRect(browser.handle, rect);
 	controller.put_Bounds(rect);
 	controller.put_IsVisible(true);
+}
+
+private void scheduleMouseMovementHandling() {
+	browser.getDisplay().timerExec(100, () -> {
+		if (browser.isDisposed()) {
+			return;
+		}
+		if (browser.isVisible() && hasDisplayFocus()) {
+			handleMouseMovement();
+		}
+		scheduleMouseMovementHandling();
+	});
+}
+
+private void handleMouseMovement() {
+	final Point currentCursorLocation = browser.getDisplay().getCursorLocation();
+	Point cursorLocationInControlCoordinate = browser.toControl(currentCursorLocation);
+	boolean isCursorInsideBrowser = browser.getBounds().contains(cursorLocationInControlCoordinate);
+	boolean hasCursorLocationChanged = !currentCursorLocation.equals(previousCursorPosition.location);
+
+	boolean mousePassedBrowserBorder = previousCursorPosition.isInsideBrowser != isCursorInsideBrowser;
+	boolean mouseMovedInsideBrowser = isCursorInsideBrowser && hasCursorLocationChanged;
+	if (mousePassedBrowserBorder) {
+		if (isCursorInsideBrowser) {
+			sendMouseEvent(cursorLocationInControlCoordinate, SWT.MouseEnter);
+		} else {
+			sendMouseEvent(cursorLocationInControlCoordinate, SWT.MouseExit);
+		}
+	} else if (mouseMovedInsideBrowser) {
+		sendMouseEvent(cursorLocationInControlCoordinate, SWT.MouseMove);
+	}
+	previousCursorPosition = new CursorPosition(currentCursorLocation, isCursorInsideBrowser);
+}
+
+private void sendMouseEvent(Point cursorLocationInControlCoordinate, int mouseEvent) {
+	Event newEvent = new Event();
+	newEvent.widget = browser;
+	Point position = cursorLocationInControlCoordinate;
+	newEvent.x = position.x;
+	newEvent.y = position.y;
+	newEvent.type = mouseEvent;
+	browser.notifyListeners(newEvent.type, newEvent);
+}
+
+private boolean hasDisplayFocus() {
+	return browser.getDisplay().getFocusControl() != null;
 }
 
 @Override


merks:
I would expect some type of checking the location of the pointer over some time interval to see it hasn't changed, i.e., that the user is actually hovering and not just moving the pointer across the control and you happen to catch that after the .5 second delay.  I'm not sure what a "normal" hover time is.   And shouldn't it produce a org.eclipse.swt.SWT.MouseHover event?

HeikoKlare:
> I would expect some type of checking the location of the pointer over some time interval to see it hasn't changed, i.e., that the user is actually hovering and not just moving the pointer across the control and you happen to catch that after the .5 second delay. I'm not sure what a "normal" hover time is. And shouldn't it produce a org.eclipse.swt.SWT.MouseHover event?

I agree to both:
- From my understanding, the proposal is actually not about "hovering" (i.e., _not_ moving the mouse for some time), but it's just about position tracking (identifying mouse movements and producing according mouse move, enter, and exit events). So there should be some position tracking, such that an event is only fired if there is an actual change in the position. I would also expect this to happen more often than 2 times a second, as you can have quite some mouse movement within 500ms, but I am also not sure what a reasonable value could be (maybe 100ms)?
- Other events like `MouseEnter` and `MouseExit` should be created as well, in particular because I guess for the issue to be addressed (missing mouse enter event for the hover information control), those are necessary events.

You should also have the code properly represent this. Currently the method states that it handles browser hover, but isn't it about something like "registring mouse movement tracking"?

amartya4256:
> * derstanding, the proposal is actually not about "hovering" (i.e., _not_ moving the mouse for some time), but it's just about position tracking (identifying mouse movements and producing according mouse move, enter, and exit events). So there should be some position tracking, such tha

After looking at how IE handles this, IE only sends out MouseEnter, Move and Exit events and not hover. If we have to replicate the behavior of IE, I think we should not add MouseHover: also AbstractHoverInformationControl doesn't have any case for MouseHover in its listener. But I would implement MouseEnter and MouseExit. I'll have to manage the state somehow so that I know what event to emit.

Moreover, I can try setting the check every 100 ms. But I also noticed that if i send out events too often, they are not processed, i.e. at rate < 10 ms. I'll try with 100 ms.

fedejeanne (changes requested):
@amartya4256 I took a brief look into this PR, I have some minor comments regarding the coding style and I also had an exception while testing.

This is the exception:
```java
org.eclipse.swt.SWTException: Failed to execute runnable (org.eclipse.swt.SWTError: No more handles [0x8007139f])
	at org.eclipse.swt.SWT.error(SWT.java:4922)
	at org.eclipse.swt.SWT.error(SWT.java:4837)
	at org.eclipse.swt.widgets.Synchronizer.runAsyncMessages(Synchronizer.java:135)
	at org.eclipse.swt.widgets.Display.runAsyncMessages(Display.java:4141)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:3757)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine$5.run(PartRenderingEngine.java:1151)
	at org.eclipse.core.databinding.observable.Realm.runWithDefault(Realm.java:339)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine.run(PartRenderingEngine.java:1042)
	at org.eclipse.e4.ui.internal.workbench.E4Workbench.createAndRunUI(E4Workbench.java:153)
	at org.eclipse.ui.internal.Workbench.lambda$3(Workbench.java:639)
	at org.eclipse.core.databinding.observable.Realm.runWithDefault(Realm.java:339)
	at org.eclipse.ui.internal.Workbench.createAndRunWorkbench(Workbench.java:546)
	at org.eclipse.ui.PlatformUI.createAndRunWorkbench(PlatformUI.java:173)
	at org.eclipse.ui.internal.ide.application.IDEApplication.start(IDEApplication.java:178)
	at org.eclipse.equinox.internal.app.EclipseAppHandle.run(EclipseAppHandle.java:208)
	at org.eclipse.core.runtime.internal.adaptor.EclipseAppLauncher.runApplication(EclipseAppLauncher.java:143)
	at org.eclipse.core.runtime.internal.adaptor.EclipseAppLauncher.start(EclipseAppLauncher.java:109)
	at org.eclipse.core.runtime.adaptor.EclipseStarter.run(EclipseStarter.java:439)
	at org.eclipse.core.runtime.adaptor.EclipseStarter.run(EclipseStarter.java:271)
	at org.eclipse.equinox.launcher.Main.invokeFramework(Main.java:668)
	at org.eclipse.equinox.launcher.Main.basicRun(Main.java:605)
	at org.eclipse.equinox.launcher.Main.run(Main.java:1481)
	at org.eclipse.equinox.launcher.Main.main(Main.java:1454)
Caused by: org.eclipse.swt.SWTError: No more handles [0x8007139f]
	at org.eclipse.swt.SWT.error(SWT.java:4948)
	at org.eclipse.swt.browser.Edge.error(Edge.java:221)
	at org.eclipse.swt.browser.Edge.setupBrowser(Edge.java:561)
	at org.eclipse.swt.browser.Edge.lambda$10(Edge.java:543)
	at org.eclipse.swt.browser.Edge.lambda$5(Edge.java:228)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:3752)
	at org.eclipse.jface.internal.text.html.BrowserInformationControl.setVisible(BrowserInformationControl.java:357)
	at org.eclipse.jface.text.AbstractInformationControlManager.showInformationControl(AbstractInformationControlManager.java:1240)
	at org.eclipse.jface.internal.text.StickyHoverManager.showInformationControl(StickyHoverManager.java:262)
	at org.eclipse.jface.internal.text.InformationControlReplacer.showInformationControl(InformationControlReplacer.java:150)
	at org.eclipse.jface.text.AbstractInformationControlManager.internalShowInformationControl(AbstractInformationControlManager.java:1134)
	at org.eclipse.jface.text.AbstractInformationControlManager.presentInformation(AbstractInformationControlManager.java:1120)
	at org.eclipse.jface.text.AbstractInformationControlManager.setInformation(AbstractInformationControlManager.java:431)
	at org.eclipse.jface.internal.text.InformationControlReplacer.computeInformation(InformationControlReplacer.java:116)
	at org.eclipse.jface.text.AbstractInformationControlManager.doShowInformation(AbstractInformationControlManager.java:1101)
	at org.eclipse.jface.text.AbstractInformationControlManager.showInformation(AbstractInformationControlManager.java:1091)
	at org.eclipse.jface.internal.text.InformationControlReplacer.replaceInformationControl(InformationControlReplacer.java:103)
	at org.eclipse.jface.text.AbstractInformationControlManager.replaceInformationControl(AbstractInformationControlManager.java:1268)
	at org.eclipse.jface.text.AbstractHoverInformationControlManager.replaceInformationControl(AbstractHoverInformationControlManager.java:794)
	at org.eclipse.jface.text.TextViewerHoverManager.replaceInformationControl(TextViewerHoverManager.java:297)
	at org.eclipse.jface.text.AbstractHoverInformationControlManager$1.lambda$0(AbstractHoverInformationControlManager.java:856)
	at org.eclipse.ui.internal.PendingSyncExec.run(PendingSyncExec.java:68)
	at org.eclipse.ui.internal.UILockListener.doPendingWork(UILockListener.java:166)
	at org.eclipse.swt.widgets.Synchronizer.runAsyncMessages(Synchronizer.java:132)
	... 22 more
```

## How I tested
- Added start parameter `-Dorg.eclipse.swt.browser.DefaultType=Edge`
- Created class and added JavaDoc
- Hovered over the name of the class 
  - JavaDoc appeared but it was empty ❌ 
  - Hovering turned the popup into a browser (buttons showed up) ✔️ 
  - The text was still blank ❌ 
  - Resizing the popup didn't redraw properly ❌ 

hover_javadoc_edge_bug

HeikoKlare:
> Moreover, I can try setting the check every 100 ms. But I also noticed that if i send out events too often, the replacement doesn't occur, i.e. at rate < 10 ms. I'll try with 100 ms; maybe because the schedule never gets an opportunity to execute? **Update: At 100 ms, it seems to not be able to trigger the replacement, but at 200 ms, it seems to work.**

This sounds quite strange. Why should the success of the event processing depend on the rate in which it is checked? Sounds to me as if we need to understand the root cause for this. Just trying out what value may work in some specific environment and using this value will not be sustainable. I would expect this to even work even if you schedule this repeatedly, without any delay, as I do not see a conceptual reason why it should not work.

> Also I looked into the code of IE. It doesn't have any implementation for checking if the user stays longer on the hover but it rather creates the Mouse event and sends it directly on the callback. The handling of the delay and replacement is completely handled by the AbstractHoverInformationControlManager (it delays the process by 1 second if the flag is set), but it is completely independent of Browser implementation. The browser is only responsible for sending the event on MouseMove.
> 
> eg: If we are just passing by the tooltip and the mouse exits the tooltip in less than one second then there wont be any replacement since the tooltip would close before the replacement which happens after a delay of 1 second. So, I dont think we need to implement that part. But rather just implementing sending out of the right event: MouseEnter, MouseExit and MouseMove should suffice.

That's all fine. Of course, we do not and even should not implement anything related to use-case specific delays (like when the hover information control is replaced) in SWT. It is really about separation of concerns:
- This implementation in SWT should be responsible for generating appropriate MouseMove/MoveEnter/MouseExit events. The time to wait until such an event may be generated must _not_ depend on a specific use case for which it might be relevant (like the replacement of the hover information control).
- The hover information control processes the MoveMove/MoveEnter events to identify if there is a hovering for more than 1 second. In that case, it replaces the widget. For this, it requires the according, timely events fired by the underlying browser widget.

amartya4256:
@HeikoKlare @merks Cleaned up the code a bit. Now, we have all the types of events being sent out as per the state of the cursor w.r.t. browser. Since this also limits the number of events being sent out, now it works with 100 ms as well. But you are right, there are 1000s of events generated in a second, it should be able to process all of them. I need to check what happens to the message at the handler, in case a lot of them are sent. Would keep you posted.

amartya4256:
> @amartya4256 I took a brief look into this PR, I have some minor comments regarding the coding style and I also had an exception while testing.
> 
> ## How I tested
> * Added start parameter `-Dorg.eclipse.swt.browser.DefaultType=Edge`
> * Created class and added JavaDoc
> * Hovered over the name of the class
>   
>   * JavaDoc appeared but it was empty ❌
>   * Hovering turned the popup into a browser (buttons showed up) ✔️
>   * The text was still blank ❌
>   * Resizing the popup didn't redraw properly ❌
> 

You have this issue because you are using Edge browser on your IDE and launching another sdk product with Edge browser there as well. This issue is related to this: https://github.com/eclipse-platform/eclipse.platform.swt/issues/1013 and here are the PRs for that, currently in progress: https://github.com/eclipse-platform/eclipse.platform.ui/pull/2434 & https://github.com/eclipse-platform/eclipse.platform.swt/pull/1548

If you set your default browser to IE in your working IDE and then try to launch the sdk product, you won't have this issue.

fedejeanne:
> You have this issue because you are using Edge browser on your IDE and launching another sdk product with Edge browser there as well.

I switched to IE in my IDE and it works better (the JavaDoc is there and it reacts to hover events) but there's still the flickering (see block below):

<details><summary><b>WARNING!!! Flickering</b></summary>
<p>

hover_javadoc_edge_flickering


</p>
</details>

fedejeanne:
BTW @amartya4256 I updated the description of this PR and added a "How to test". Feel free to improve it :-)

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -700,6 +704,60 @@ void browserResize(Event event) {
...
+			}
+			break;
+		}
+		handleBrowserHover();
+	});
+}
+
+private enum CursorPosition {
+	INSIDE, OUTSIDE
+}
```
HeikoKlare, line 745: I wonder whether it is worth to have an enum with only these two values, i.e., whether representing this via enums is more expressive/comprehensible than storing it in some `isCursorInsideBrowser` boolean.

In addition, I am not sure whether it is really necessary to store whether the cursor was inside the browser or whether we can just calculate it every 100ms, as it's only a cheap position-within-bounds check. That would reduce the number of instance fields that have to be managed thoroughly to not represent an inconsistent state.

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -645,6 +645,7 @@ void setupBrowser(int hr, long pv) {
 	browser.addListener(SWT.FocusIn, this::browserFocusIn);
 	browser.addListener(SWT.Resize, this::browserResize);
 	browser.addListener(SWT.Move, this::browserMove);
+	handleBrowserHover();
```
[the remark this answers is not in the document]
HeikoKlare: This method is not even so specific to hovering (which may be considered the action having the mouse over the control without movement for some amount of time). It's rather about the general processing of mouse movements within the browser control, see also https://github.com/eclipse-platform/eclipse.platform.swt/pull/1551#issuecomment-2437071832

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -700,6 +701,35 @@ void browserResize(Event event) {
...
+	browser.getDisplay().timerExec(500, () -> {
+		if (browser.isDisposed()) {
+			return;
+		}
+		if (!browser.isVisible() || !hasDisplayFocus()) {
+			handleBrowserHover();
+			return;
+		}
+		final Point cursorLocation = browser.getDisplay().getCursorLocation();
+		Point cursorLocationInControlCoordinate = browser.toControl(cursorLocation);
```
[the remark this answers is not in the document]
HeikoKlare: why?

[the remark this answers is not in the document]
fedejeanne: Oh, they are both needed (I missed the assignment `previousCursorLocation = cursorLocation;`).

Scratch my comment, thank you for the heads up.

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
fedejeanne: > In addition, I am not sure whether it is really necessary to store whether the cursor was inside the browser or whether we can just calculate it every 100ms...

Do you mean getting rid of the field `cursorPosition`?
If I read the code correctly, it is necessary to detect a `MouseEnter` / `MouseExit` situation. Calculating it on-the-fly would tell you were it is right now but you need to know where it was before in order to know if it entered/left an area.

HeikoKlare: Yes, I mean getting rid of `cursorPosition`. There is `previousCursorLocation` from which you can always derive the value of `cursorPosition`, so it only serves as a cache. And since the calculation is so cheap, I would challenge the performance benefit of caching the value compared to the implicit dependency between the `previousCursorLocation` and `cursorPosition` values, necessity to update the together and the resulting risk of inconsistencies/regressions.

fedejeanne: Ah, I understand what you mean now.

I would rather leave it as it is right now. The reason for that is that if one uses `previousCursorLocation` to calculate the `cursorPosition` then the semantics become a bit (more) difficult and one also has to move the instruction `previousCursorLocation = cursorLocation;` below the `switch` statement because of the added "temporal dependency" (`previousCursorLocation` needs to remain **previous** before it becomes the **current (for the next iteration)** cursor location). 

Right now the only problem I see with the field is that its semantics change inside the `switch` statement: at the beginning it's the "last known position" (from the previous iteration) and inside the `switch`, where it is assigned, it becomes the "current position". But I prefer it that way, it makes the code easier to read.

HeikoKlare:
@sratz as you have also been working on this issue (via #1545): do you any opinions on the proposed solution?

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -78,6 +78,9 @@ public WebViewEnvironment(ICoreWebView2Environment environment) {
 	private boolean ignoreFocus;
 	private String lastCustomText;
 
+	record CursorPosition(Point location, boolean isInsideBrowser) {};
```
fedejeanne, line 81: I'd make it private

fedejeanne (commented):
I'd make the `record` definition private since it's not needed anywhere else.

I can understand the code, I don't notice any performance change and it works as expected ✔️ 

The only thing missing is fixing the flickering (see https://github.com/eclipse-platform/eclipse.platform.swt/pull/1551#issuecomment-2440809292). I assume this is the exact same issue as https://github.com/eclipse-platform/eclipse.platform.swt/issues/1122 so I'd say we leave it "unsolved" (in this PR).

Other than that, this PR LGTM (approved!) 👍

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
HeikoKlare: Thanks! Yes, this should be private. Fixed with a76d981.

HeikoKlare:
> The only thing missing is fixing the flickering (see [#1551 (comment)](https://github.com/eclipse-platform/eclipse.platform.swt/pull/1551#issuecomment-2440809292)). I assume this is the exact same issue as #1122 so I'd say we leave it "unsolved" (in this PR).

Indeed, that issue remains, but affects every composite embedding an Edge browser. So it should be addressed separately, according to its documentation in #1122.

sratz:
> @sratz as you have also been working on this issue (via #1545): do you any opinions on the proposed solution?

I think this is a good first solution and better than the javascript approach (which does not work universally). 👍 

I was wondering: Currently we schedule a 100ms runnable for each Browser instance and even if the cursor is outside the Browser.

I see two possibile optimizations:
- Can we somehow detect that we enter or leave the browser and only do repeated scheduling if we are still the boundaries?
- Can we use a static runnable that we start once and add / remove Browser instances to which deals with all of them in one go?

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -78,6 +78,9 @@ public WebViewEnvironment(ICoreWebView2Environment environment) {
 	private boolean ignoreFocus;
 	private String lastCustomText;
 
+	private record CursorPosition(Point location, boolean isInsideBrowser) {};
```
laeubi, line 81:
```suggestion
	private static record CursorPosition(Point location, boolean isInsideBrowser) {};
```

sratz: Aren't records implicitly static?

HeikoKlare:
> I was wondering: Currently we schedule a 100ms runnable for each Browser instance and even if the cursor is outside the Browser.
> 
> I see two possibile optimizations:
> 
> * Can we somehow detect that we enter or leave the browser and only do repeated scheduling if we are within its boundaries?
> * Can we use a static runnable that we start once and add / remove Browser instances to which deals with all of them in one go?

Good point, I also thought about whether we can avoid that this runnable is executed all the time. But since once the mouse enters the browser widgets no mouse move events are fires any more, for which this is actually the workaround, we do not see any possibility of improving the detection of the mouse entering/leaving the browser. Otherwise that mechanism would completely address the current issue that we want to solve, as it's only about the entering event (and not the move event) at all.

Having just one static runnable to avoid this effort to increase by the number of open browser sounds reasonable. What about merging this PR as is to give it a try as soon as possible and I will improve the implementation to use a static runnable as a follow-up PR?

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
HeikoKlare: I think they are only implicitly static in specific situations, such as when defined locally (within a method). So I think adding the static modifier makes sense. I will update the PR accordingly. Thank you for the pointer!

HeikoKlare: Fixed in 77a5a81

HeikoKlare:
Okay, let's give this a try. I will prepare the promised follow-up PR as soon as possible.

Failing tests are unrelated and documented: #1523

HeikoKlare:
> * Can we use a static runnable that we start once and add / remove Browser instances to which deals with all of them in one go?

I have just started working on this but found that all the calculations need to be performed in a browser-instance-specific way (as each requires its own coordinate transformation and check for the mouse being inside the browser). Only the scheduling itself could be done in a "static" way. So I am not sure whether it is (1) worth the effort and additional complexity to have a central, WebViewEnvironment-specific scheduling of the movement tracking and whether it is (2) not even better to have this distributed across the browser instances so that the according events are also distributed over time instead of always having a batch of multiple move events at the same point in time. @sratz what do you think?

merks:
Avoid making it more complicated. I doubt performance is a concern.

sratz:
Agreed, let's test this out in the current state first.

PR #1530, opened 2024-10-10
Edge: Inherit dark theme from Eclipse

sratz:
Set ICoreWebView2Profile#put_PreferredColorScheme() based on the Eclipse theme, if this is dark.

The same way it is done in

  Display.setData(USE_DARKMODE_EXPLORER_THEME_KEY, ...)

via

  OS.SetPreferredAppMode(PreferredAppMode_...);

Fixes #1529

PR #1558, opened 2024-10-25
Edge: Dark mode support: Use LIGHT instead of AUTO

sratz:
In the scenario
- System: Dark
- Eclipse: Light

we need to explicitly set
- Edge: Light

as well, as with the previous
- Edge: Auto

it would inherit 'Dark' from the system, which is not desired.

Follow-up on 4d61d1dfcc879911bcb361e1af25b08ef19ed88e / #1530.


HeikoKlare (approved):

sratz:
Maybe when we go in the direction of https://github.com/eclipse-platform/eclipse.platform.ui/issues/2440 (inherit from system + auto-adapt theme on-the-fly) we need to use `AUTO` at some point.

sratz:
Test failures on Linux are unrelated.

PR #1562, opened 2024-10-28
Display.withCrLf() is producing wrong line breaks on Windows #1557

### LINKED ISSUE #1557: Display.withCrLf() is producing wrong line breaks on Windows
**Describe the bug**
Display.withCrLf() is producing wrong line breaks based on the heuristics in the Windows implementation. This happens only for Strings that contains broken formatting itself. But nevertheless, I think we can fix this issue by improving the method implementation. 

**To Reproduce**
See Test case
```        
	@Test
	public void testMixedLfAndCrflOnWindows() {
		if (!SWT.getPlatform().equals("win32")) {
			return;
		}

		String act = Display.withCrLf("First Line \n second line \r\n third line");

		assertEquals("First Line \r\n second line \r\n third line", act);
	}
```
**Expected behavior**
Expectation is that the method replaces all standalone \n with \r\n.
With the given example, \r\n results in \r\r\n

**Screenshots**

**Environment:**
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS

Evgeni-Kramerov:
Fixes issue #1557 

Changes functions so line breaks display correctly in Windows.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/1557

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
@@ -5215,79 +5215,65 @@ static String withCrLf (String string) {
 	int length = string.length ();
 	if (length == 0) return string;
 
-	/*
-	* Check for an LF or CR/LF and assume the rest of
-	* the string is formated that way.  This will not
-	* work if the string contains mixed delimiters.
-	*/
+	/* If there is no LF - return string */
 	int i = string.indexOf ('\n', 0);
 	if (i == -1) return string;
-	if (i > 0 && string.charAt (i - 1) == '\r') {
-		return string;
-	}
 
-	/*
-	* The string is formatted with LF.  Compute the
-	* number of lines and the size of the buffer
-	* needed to hold the result
-	*/
-	i++;
-	int count = 1;
-	while (i < length) {
-		if ((i = string.indexOf ('\n', i)) == -1) break;
-		count++; i++;
-	}
-	count += length;
+	/*Create String Builder for return string*/
+	StringBuilder result = new StringBuilder();
+	int j = 0;
 
-	/* Create a new string with the CR/LF line terminator. */
-	i = 0;
-	StringBuilder result = new StringBuilder (count);
+	/*If there is just LF, without CR before it,
+	add CR before LF*/
+
 	while (i < length) {
-		int j = string.indexOf ('\n', i);
-		if (j == -1) j = length;
-		result.append (string.substring (i, j));
-		if ((i = j) < length) {
-			result.append ("\r\n"); //$NON-NLS-1$
-			i++;
+		result.append(string.substring(j, i));
+		if(string.charAt(i - 1) == '\r'){
+			result.append('\n');
+		}
+		else {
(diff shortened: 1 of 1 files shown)


BeckerWdf:
can you please add the test-case provided in https://github.com/eclipse-platform/eclipse.platform.swt/issues/1557

> To Reproduce
> See Test case
> 
> 	@Test
> 	public void testMixedLfAndCrflOnWindows() {
> 		if (!SWT.getPlatform().equals("win32")) {
> 			return;
> 		}
> 
> 		String act = Display.withCrLf("First Line \n second line \r\n third line");
> 
> 		assertEquals("First Line \r\n second line \r\n third line", act);
> 	}

Christopher-Hermann:
There are failing test that are related to this change. Looks like some edge cases which results in out of bounds exception
String index out of range: -1
java.lang.StringIndexOutOfBoundsException: String index out of range: -1
	at java.base/java.lang.StringLatin1.charAt(StringLatin1.java:48)
	at java.base/java.lang.String.charAt(String.java:1517)
	at org.eclipse.swt.widgets.Display.withCrLf(Display.java:5231)
	at org.eclipse.swt.widgets.Text.setText(Text.java:2305)
	at org.eclipse.swt.tests.junit.Test_org_eclipse_swt_custom_CTabFolder.createTabFolder(Test_org_eclipse_swt_custom_CTabFolder.java:140)

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
```diff
@@ -5215,79 +5215,65 @@ static String withCrLf (String string) {
...
+		}
+		j = i + 1;
+		i = string.indexOf('\n', j);
+		if (i == -1) {
+			i = length;
+			result.append(string.substring(j));
 		}
 	}
-	return result.toString ();
+	return result.toString();
```
Christopher-Hermann, line 5244: I did not took a detailed look in the original coding, but according to ChatGTP, we could just use a regex to normalize the String. See the coding:
```
/**
     * Normalize line breaks in a string to Windows line breaks (\r\n)
     *
     * @param input The input string
     * @return The string with normalized line breaks
     */
    public static String normalizeLineBreaks(String input) {
        // Replace \r\n, \r, or \n with \r\n
        return input.replaceAll("(\r\n|\r|\n)", "\r\n");
    }
```

This seems pretty easy and clean. Maybe you can give it a try.

akurtakov:
@HeikoKlare Is this PR needed or it should be closed?

HeikoKlare:
I don't think this is needed anymore. It was superceded by:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/1708

That one seems to have caused a regression:
- https://github.com/eclipse-platform/eclipse.platform.swt/issues/1718

Which was then resolved via:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/1722

Even after that, the test showing the reported issue still succeeds:
- https://github.com/eclipse-platform/eclipse.platform.swt/issues/1557

So I think this is not necessary anymore. @Evgeni-Kramerov can we close this?

PR #1563, opened 2024-10-28
[GTK4] SIGSEGV in OS.gdk_display_get_default() during splash screen

ptziegler:
The static initializer in OS is calling "gdk_display_get_default()" via "isX11()" and therefore violates the requirement that "gtk_init()"/"gtk_init_check()" must be invoked before any other GTK+ function. While doing so works in GTK3, it is by no means a documented behavior, but rather something that works by pure chance [1].

With this change, the system property is moved from the initializer to the Display constructor, where the init function is explicity called. The change itself was made in an error being thrown while executing the JUnit tests on Wayland [2]. But because the display is created before those are executed, it shouldn't matter where this property is set.

[1] https://bugs.eclipse.org/bugs/show_bug.cgi?id=465863
[2] https://bugs.eclipse.org/bugs/show_bug.cgi?id=528968


ptziegler:
Even with this change, the IDE still gets stuck during startup. But at the very least, it doesn't end with a total crash.

ptziegler:
The browser tests also seem to fail on the main branch...

akurtakov:
Thanks!

Issue #1564, opened 2024-10-30
[Linux] 3 tests failing in Test_org_eclipse_swt_browser_Browser


fedejeanne:
## The problem
It is not possible to open a child browser with newer versions of WebKit.

## How to reproduce
The problem can be reproduced in **Ubuntu 24.04** (tested using WSL2) with **WebKit 2.46**

- Install the proper version of WebKit with this command:
```bash
VERSION=2.46.3-0ubuntu0.24.04.1 && sudo apt-get install libwebkitgtk-6.0-4=${VERSION} libjavascriptcoregtk-6.0-1=${VERSION}
```

- Run this snippet:
<details><summary>Snippet</summary>
<p>

```java
public class OpenChildBrowser {
	public static void main(String[] args) {
		Display display = new Display();
		final Shell shell = new Shell(display);
		shell.setLayout(new FillLayout());
		final Browser browser = new Browser(shell, SWT.NONE);
		Shell childShell = new Shell(shell, SWT.None);
		childShell.setText("Child shell");
		childShell.setLayout(new FillLayout());
		final Browser browserChild = new Browser(childShell, SWT.NONE);
		browser.addOpenWindowListener(event -> {
			System.out.println("Open Window Listener");
			event.browser = browserChild;
		});
		browserChild.addVisibilityWindowListener(VisibilityWindowListener.showAdapter(event -> {
			System.out.println("Open shell");
			childShell.open();
			browserChild.setText("Child Browser");
		}));
		shell.open();
		browser.setText("""
				<html>
					<script type='text/javascript'>
						var newWin = window.open('about:blank');
					</script>
					<body>
						This test uses javascript to open a new window.
					</body>
				</html>
				""");
		while (!shell.isDisposed()) {
			if (!display.readAndDispatch())
				display.sleep();
		}
		display.dispose();
	}
}
```
</p>
</details> 

### Expected result
A child window should be opened

### Actual result
No child window is opened

### Reverting back to a working version of WebKit
Version **2.44** of WebKit didn't have this problem. You can revert back to that version by running this command:

```bash
VERSION=2.44.0-2 && sudo apt-get install libwebkit2gtk-4.1-0=${VERSION} libjavascriptcoregtk-4.1-0=${VERSION}
```

# Original issue: failing tests

[...]

PR #1565, opened 2024-10-30
[win] De-register resources with zoom support from device

HeikoKlare:
Resources with zoom support are registered at their device so that handles for zoom levels that are not necessary anymore can be disposed. When these resources are destroyed, they are currently not de-registered form their device, potentially leading to resource leaks. This change makes the resources de-register themselves form their device when being destroyed.

Fixes regression from:
- #1477

Fixes https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1752

{73E6A5F0-38EC-48D0-AF43-F821B5B4FD56}


fedejeanne (changes requested):
Thank you for the fix!

fedejeanne (approved):

HeikoKlare:
Failing tests are Linux-specific and documented: #1564

PR #1566, opened 2024-10-31
[win32] possible wrong zoom set for system font

akoch-yatta:
This PR adapts the initialization of a new base system font in the ScalingSWTFontRegistry to ensure the correct zoom is set at initialization.

I am not aware of any real issue caused by a wrong zoom value, but not using the targetZoom here is a possible source of errors.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/ScalingSWTFontRegistry.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/ScalingSWTFontRegistry.java
@@ -90,10 +90,13 @@ final class ScalingSWTFontRegistry implements SWTFontRegistry {
 	private ScaledFontContainer getOrCreateBaseSystemFontContainer(Device device) {
 		ScaledFontContainer systemFontContainer = fontKeyMap.get(KEY_SYSTEM_FONTS);
 		if (systemFontContainer == null) {
-			int targetZoom = DPIUtil.mapDPIToZoom(device.getDPI().x);
-			long systemFontHandle = createSystemFont(targetZoom);
-			Font systemFont = Font.win32_new(device, systemFontHandle);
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/ScalingSWTFontRegistry.java
```diff
@@ -92,7 +92,7 @@ private ScaledFontContainer getOrCreateBaseSystemFontContainer(Device device) {
 		if (systemFontContainer == null) {
 			int targetZoom = DPIUtil.mapDPIToZoom(device.getDPI().x);
 			long systemFontHandle = createSystemFont(targetZoom);
-			Font systemFont = Font.win32_new(device, systemFontHandle);
+			Font systemFont = Font.win32_new(device, systemFontHandle, targetZoom);
```
HeikoKlare, line 95: Since the zoom-less method also extracts the zoom from the font's device in the same was as this method does, this change is a pure refactoring, isn't it?

It's of course fine and makes sense to me. Just want to understand whether this, in contrast to the PR title, isn't actually just a refactoring.

akoch-yatta: Yes and no. Your observation is correct, that it looks like a refactoring, but the effect is quite easy to reproduce in a int200 scenario. I had a look why it is like that and noticed, that device.getDPI() always returns the scaled down version of the primary monitor DPI -> means in a 125% int200 setting the targetZoom is 100. With my change, we would make sure the correct zoom is set, but we store a possibly unnecessary font, so I  could adapt the upper call as well.

HeikoKlare (approved):
The change is sound. In addition, since the affected font registry is only used when the experimental rescale-at-runtime feature is activated, there is no risk in this change.

HeikoKlare:
Failing tests are unrelated and documented: #1564

PR #1568, opened 2024-10-31
Win32: Fix compatibility with Windows Server 2012, Windows Server 2016, and Windows 8.1

### LINKED ISSUE #1252: UnsatisfiedLinkError of swt-win32-4965r8.dll on Windows Server 2008 R2
Hello,

Issue description: 

Recently we switched to the  latest Eclipse SDK release 
https://download.eclipse.org/releases/2024-06  version build [202405031000](https://download.eclipse.org/releases/2024-06/202405031000) 
and bumped with product crash on specific OS's:  Windows 8 / Windows Server 2008 
due to **UnsatisfiedLinkError of swt-win32-4965r8.dll**

PS:
Let us please, know if you reduce support of OS like Windows 8 / Windows 2008 RC  

The original issue in DBeaver product:
https://github.com/dbeaver/dbeaver/issues/34034

ShadelessFox:
This PR marks the following Win32 functions as _dynamic_:
- `OpenThemeDataForDpi`<sup>1</sup>
- `GetSystemMetricsForDpi`
- `GetThreadDpiAwarenessContext`
- `SetThreadDpiAwarenessContext`
- `SystemParametersInfoForDpi`

It's enough to make SWT run on Windows Server 2016, Windows Server 2012, and Windows 8.1.

The Java code that uses these functions **already checks** if the OS version is above the minimum required, so no actions should be taken there.

<sup>1</sup>Although the `OpenThemeDataForDpi` is [documented](https://learn.microsoft.com/en-us/windows/win32/api/uxtheme/nf-uxtheme-openthemedatafordpi) to be available since Windows Server 2016, it's not available there, or at least not in the LTS version.

Addresses and fixes #1252 and possibly many other issues related to running SWT apps on older versions of Windows.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT PI/win32/library/os.c
+++ b/bundles/org.eclipse.swt/Eclipse SWT PI/win32/library/os.c
@@ -3276,7 +3276,15 @@ JNIEXPORT jint JNICALL OS_NATIVE(GetSystemMetricsForDpi)
 {
 	jint rc = 0;
 	OS_NATIVE_ENTER(env, that, GetSystemMetricsForDpi_FUNC);
+/*
 	rc = (jint)GetSystemMetricsForDpi(arg0, arg1);
+*/
+	{
+		OS_LOAD_FUNCTION(fp, GetSystemMetricsForDpi)
+		if (fp) {
+			rc = (jint)((jint (CALLING_CONVENTION*)(jint, jint))fp)(arg0, arg1);
+		}
+	}
 	OS_NATIVE_EXIT(env, that, GetSystemMetricsForDpi_FUNC);
 	return rc;
 }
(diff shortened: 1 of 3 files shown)


Phillipus:
What happens when core SWT code calls these functions and they're not available? A crash?

HeikoKlare (approved):
Thank you for this contribution! It's great to have the compatibility issue addressed.

The proposal looks completely sound and works as expected. It is consistent with the other implementations of dynamically loaded functions. Only concern might be that in case you call a method that is not available there will be no error but you simply get `0` as a result, but I think it's fine to assume that calls to such a function need to be guarded with a version check (like is already done right now).

I would be in favor of merging this as soon as possible to have it in the upcoming release (and still tested for some days before that).

@Phillipus what is your opinion?

Phillipus:
@Phillipus what is your opinion?

As long as it doesn't break anything, fine.

HeikoKlare:
> As long as it doesn't break anything, fine.

I do not see a risk of breaking anything. It just aligns with other dynmically loaded functions and actually corrects how these functions are called: their calls have already been wrapped in version checks but you can never reach this code as platforms not fulfilling these version constraints will not even start the applications (as we know).

HeikoKlare:
I've retested different scenarios so that all affected calls have been executed. They all work as expected on my system (Windows 11 22631), which provides all those functions.
I have also re-checked all callers and as stated by @ShadelessFox, they are guarded by according version checks (and anyway, things could not get worse in those environments).

I will merge this so that we have some days/weeks of (implicit) testing. In case there are further remarks or concerns, feel free to comment and let's address them as a follow-up.

For failing tests see: #1564

Thanks again, @ShadelessFox!

PR #1569, opened 2024-11-01
[win32] workaround for ImageDataProvider implementation

akoch-yatta:
This commit adds a workaround for ImageDataProvider implementations that instantiate new Images inside the getImageData(zoom) implementation. This image can be instantiated with the wrong zoom in a multi zoom setting as Images still relay on the static zoom value in DPIUtil. This workaround should be replaced by a proper solution.

## How to reproduce
You can reproduce the behavior as follows:

1. Set up at least two monitors with different zoom values (e.g., 100% and 125%)
2. Start an Eclipse application with rescaling on runtime activated (and auto scale mode quarter).
3. Place the workbench window on the 100% monitor. Note: if you move it between both monitors now, the minimize/maximize icons will be rescaled correctly.
4. Move the workbench window from the 100% to the 125% monitor.
5. Change focus to another application on the 100% monitor.
6. Move the workbench window from the 125% to the 100% monitor.

Now the icons are scaled incorrectly for the active part stack. As soon as you switch focus to some other application again, the icons will be rescaled correctly.

**Important note:** This is not a permanent solution for this issue, but (due to being quite late in the current development cycle) meant as workaround for the 2024-12 release. A proper solution will require an API change, which is not possible now. I tried to implement the workaround as least invasive as possible

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
@@ -22,6 +22,7 @@ import org.eclipse.swt.internal.*;
 import org.eclipse.swt.internal.DPIUtil.*;
 import org.eclipse.swt.internal.gdip.*;
 import org.eclipse.swt.internal.win32.*;
+import org.eclipse.swt.widgets.*;
 
 /**
  * Instances of this class are graphics which have been prepared
@@ -819,7 +820,10 @@ public static Long win32_getHandle (Image image, int zoom) {
 			image.init(newData, zoom);
 		}
 	} else if (image.imageDataProvider != null) {
-		ElementAtZoom<ImageData> imageCandidate = DPIUtil.validateAndGetImageDataAtZoom (image.imageDataProvider, zoom);
+		ElementAtZoom<ImageData> imageCandidate;
+		try (StaticZoomUpdater unused = image.new StaticZoomUpdater(zoom)) {
+			imageCandidate = DPIUtil.validateAndGetImageDataAtZoom (image.imageDataProvider, zoom);
+		}
 		ImageData resizedData = DPIUtil.scaleImageData (image.device, imageCandidate.element(), zoom, imageCandidate.zoom());
 		ImageData newData = image.adaptImageDataIfDisabledOrGray(resizedData);
 		image.init(newData, zoom);
@@ -1443,7 +1447,10 @@ public ImageData getImageData (int zoom) {
 	if (zoom == currentZoom) {
 		return getImageDataAtCurrentZoom();
 	} else if (imageDataProvider != null) {
-		ElementAtZoom<ImageData> data = DPIUtil.validateAndGetImageDataAtZoom (imageDataProvider, zoom);
+		ElementAtZoom<ImageData> data;
+		try (StaticZoomUpdater unused = new StaticZoomUpdater(zoom)) {
+			data = DPIUtil.validateAndGetImageDataAtZoom (imageDataProvider, zoom);
+		}
 		return DPIUtil.scaleImageData (device, data.element(), zoom, data.zoom());
 	} else if (imageFileNameProvider != null) {
 		ElementAtZoom<String> fileName = DPIUtil.validateAndGetImagePathAtZoom (imageFileNameProvider, zoom);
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -819,7 +819,10 @@ public static Long win32_getHandle (Image image, int zoom) {
 			image.init(newData, zoom);
 		}
 	} else if (image.imageDataProvider != null) {
-		ElementAtZoom<ImageData> imageCandidate = DPIUtil.validateAndGetImageDataAtZoom (image.imageDataProvider, zoom);
+		ElementAtZoom<ImageData> imageCandidate;
+		try (StaticZoomUpdater ignored = new StaticZoomUpdater(zoom)) {
```
HeikoKlare, line 823: The name `ignored` confused me. I guess it is meant to indicate that the variable itself is "ignored", but that was not clear to me at first. Maybe rather call it `unused` or simply `__` (although I know there is some controversy on whether that's good practice).

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2397,4 +2403,25 @@ public static Image win32_new(Device device, int type, long handle) {
...
+			DPIUtil.setDeviceZoom(targetZoom);
+		}
+	}
+
+	@Override
+	public void close() {
+		if (currentNativeDeviceZoom != targetZoom) {
+			DPIUtil.setDeviceZoom(currentNativeDeviceZoom);
+		}
+	}
```
HeikoKlare, line 2425: Even though these methods should only have any side-effects in case the rescale at runtime functionality is enabled, maybe we can add some additional guard for this since
- we are late in the release cycle and I would like to avoid any unnecessary risk of regressions, and
- even though adding it would not be "clean" as should actually be unnecessary, the whole class is a temporary workaround to be removed again soon.

HeikoKlare (commented):
Workaround looks acceptable to me. We target the proper solution as a replacement for M1.

Can you please re-check the description for how to reproduce? I was not able to reproduce the behavior and I am not sure what point 6 is supposed to say. Should it be moving back to the 100% monitor?
> Move the workbench window from the 125% to the 125% monitor.

Still, if I do that, everything looks fine even on master.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: What kind of guard do you mean? something like `device instanceof Display display && display.isRescalingAtRuntime()`?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: Sure, I will change that one

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
HeikoKlare: Yes, I mean exactly that kind of guard.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: I updated it

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: I updated it

HeikoKlare (approved):
Thank you for the latest changes.

I was finally able to reproduce the issue (only in dark mode though).

This is how it looks when following the described scenario on master:
image

This is how it looks when following the described scenario on this change:
image

HeikoKlare:
Failing tests are documented and unrelated: #1564

PR #1572, opened 2024-11-05
Parameterize browser tests to allow executing them for Edge browser #671

HeikoKlare:
Browser tests were only executed for the default configuration of a system's browser using the SWT.NONE flag. Other configurations, such as using the Edge browser in Windows, were not tested.

This change parameterizes the browser tests to also allow executing them for Edge browser on Windows. Due to current issues with the tests in the CI, this change only adds the parameterization but does not add Edge as a parameter yet.

This prepares for finally activating automated tests for Edge browser and thus contributes to https://github.com/eclipse-platform/eclipse.platform.swt/issues/671

This is extracted from:
- #672


HeikoKlare:
Failing tests are documented and unrelated: #1564

PR #1574, opened 2024-11-06
[win32] Always use native zoom for OS theme calls

### LINKED ISSUE #1573: [Win11] Styling with SWT.Skin listener breaks look
**Describe the bug**
If I add a Skin listener and set the foreground color of a checkbox, it looks too compact.

**To Reproduce**
Run the following snippet on Windows 11 with 150% zoom level.
```java
import java.io.*;

import org.eclipse.swt.*;

public class HiDpiTest {
	public static void main(String[] args) throws IOException {
		display.addListener(SWT.Skin, event -> {
			if (event.widget instanceof Button button && (button.getStyle() & SWT.CHECK) != 0) {
				button.setForeground(display.getSystemColor(SWT.COLOR_BLACK));
			}
		});

		addCheckbox(shell);
		addCheckbox(shell).setSelection(true);
		addCheckbox(shell).setGrayed(true);

		final Button button = new Button(shell, SWT.PUSH);
		button.setText("Button");
		button.setLayoutData(new GridData(GridData.END, GridData.CENTER, false, false));

		shell.setSize(shell.computeSize(SWT.DEFAULT, SWT.DEFAULT));
		shell.open();
		while (!shell.isDisposed()) {
				display.sleep();
			}
		}
		display.dispose();
	}

	private static Button addCheckbox(Composite parent) {
		final Button checkbox = new Button(parent, SWT.CHECK);
		checkbox.setText("Checkbox");
		checkbox.setLayoutData(new GridData(GridData.BEGINNING, GridData.CENTER, false, false));
		return checkbox;
	}
}
```
The checkbox text is too close to the box:

With older SWT builds before commit 5dd0c286) it looks correctly:

**Environment:**
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 

Windows 11, 4k monitor, 150% zoom level

**Version since**
 Eclipse or SWT version since when the behavior is seen [e.g. 4.23]

No known workaround.

akoch-yatta:
This commit changes all DPI dependent OS theme calls to use the native zoom instead of using the zoom adjusted by the swt.autoScale setting. This is neccessary, because theme dependent elements like the checkbox for a button will be drawn by the OS and must be calculated with the native zoom.

Fixes #1573

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Button.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Button.java
@@ -1311,7 +1311,7 @@ private int getCheckboxTextOffset(long hdc) {
 	SIZE size = new SIZE();
 
 	if (OS.IsAppThemed ()) {
-		OS.GetThemePartSize(display.hButtonTheme(getZoom()), hdc, OS.BP_CHECKBOX, OS.CBS_UNCHECKEDNORMAL, null, OS.TS_TRUE, size);
+		OS.GetThemePartSize(display.hButtonTheme(nativeZoom), hdc, OS.BP_CHECKBOX, OS.CBS_UNCHECKEDNORMAL, null, OS.TS_TRUE, size);
 		result += size.cx;
 	} else {
 		result += DPIUtil.scaleUp(13, nativeZoom);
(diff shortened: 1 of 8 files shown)


HeikoKlare (approved):
Adapting these zoom values to be consistent with previous behavior of using the primary monitor zoom is reasonable, as also confirmed by the report in #1573.

HeikoKlare:
For failing tests, see #1564

PR #1575, opened 2024-11-06
For Menu.wmTimer use Menu.indexOf instead of MenuItem.index

merks:
- Remove MenuItem.index which is now unused.
- Remove unused MenuItem.MenuItem(Menu, Menu, int, int)

https://github.com/eclipse-platform/eclipse.platform.ui/issues/2483

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Menu.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Menu.java
@@ -1349,7 +1349,7 @@ LRESULT wmTimer (long wParam, long lParam) {
 		OS.GetCursorPos (pt);
 		if (selectedMenuItem != null && selectedMenuItem.parent != null) {
 			RECT rect = new RECT ();
-			boolean success = OS.GetMenuItemRect (0, selectedMenuItem.parent.handle, selectedMenuItem.index, rect);
+			boolean success = OS.GetMenuItemRect (0, selectedMenuItem.parent.handle, indexOf(selectedMenuItem), rect);
 			if (!success) return null;
 			if (OS.PtInRect (rect, pt)) {
 				// Mouse cursor is within the bounds of menu item
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MenuItem.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MenuItem.java
@@ -42,7 +42,7 @@ import org.eclipse.swt.internal.win32.*;
 public class MenuItem extends Item {
 	Menu parent, menu;
 	long hBitmap;
-	int id, accelerator, userId, index;
+	int id, accelerator, userId;
 	ToolTip itemToolTip;
 	/* Image margin. */
 	final static int MARGIN_WIDTH = 1;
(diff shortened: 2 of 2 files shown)


merks:
@HeikoKlare 

I wonder who could review this?

To reproduce the problem I used Navigate -> Open Setup -> Workspace and have this additional target, which you can copy and paste:
```
<setup.targlets:TargletTask
    xmi:version="2.0"
    xmlns:xmi="http://www.omg.org/XMI"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
    xmlns:setup.targlets="http://www.eclipse.org/oomph/setup/targlets/1.0"
    xsi:schemaLocation="http://www.eclipse.org/oomph/setup/targlets/1.0 https://raw.githubusercontent.com/eclipse-oomph/oomph/master/setups/models/SetupTarglets.ecore">
  <targlet name="EGit">
    <requirement
        name="org.eclipse.egit.feature.group"/>
    <repositoryList>
      <repository
          url="https://download.eclipse.org/egit/updates-stable-nightly"/>
    </repositoryList>
  </targlet>
</setup.targlets:TargletTask>
```

Perform the targlet task and In the launch configuration include the EGit feature.  Then in any plain text editor the Team submenu's items should all have tooltips that don't work without these changes because the index recorded in the MenuItem is simply wrong because dispose() is called on other menu items in the menu changing the index of other items

merks:
The build looks like a bit of a mess, but nothing looks related to my actual changes. Does anyone know what's going on?

merks:
@HeikoKlare 

I would prefer we get this in for m3.  

I don't actually think it needs any review.  The existing design is just fundamentally broken because the index of a menu item changes when items are inserted or disposed. So recording the index at the time of creation is just wrong.

Note that it's not even necessary to reproduce the original problem but simply to confirm that tool tips do show with the new logic, which is the case:

image

I'll add some comments and if you wish to review them, that's fine.  If not I will merge the changes later today.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MenuItem.java
```diff
@@ -136,16 +136,7 @@ public MenuItem (Menu parent, int style) {
 public MenuItem (Menu parent, int style, int index) {
 	super (parent, checkStyle (style));
 	this.parent = parent;
-	parent.createItem (this, (this.index = index));
```
merks, line 139: This looks confusingly like a removall, but note there is addition below:
```
parent.createItem (this, index);
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Menu.java
```diff
@@ -1349,7 +1349,7 @@ LRESULT wmTimer (long wParam, long lParam) {
 		OS.GetCursorPos (pt);
 		if (selectedMenuItem != null && selectedMenuItem.parent != null) {
 			RECT rect = new RECT ();
-			boolean success = OS.GetMenuItemRect (0, selectedMenuItem.parent.handle, selectedMenuItem.index, rect);
+			boolean success = OS.GetMenuItemRect (0, selectedMenuItem.parent.handle, indexOf(selectedMenuItem), rect);
```
merks, line 1352: The fundamental change here is to compute the correct index using the existing method for computing the index.

merks (commented):
Note the only potentially controversial change is the removal of the unused MenuItem constructor.

PR #1576, opened 2024-11-07
[win32] Fix for refreshing the background image

akoch-yatta:
This PR adapts the refreshing of a background image of a control on a DPI change. This must only be done, if the image was set directly on the control.


HeikoKlare (approved):
Thank you! Works as expected. And since it only affects code that is executed with the experimental rescaling feature enabled, there is no risk in merging.

HeikoKlare:
For failing tests, see: #1564

PR #1578, opened 2024-11-07
Edge: properly handle initialization failures and abortion

HeikoKlare:
There is currently only few handling for failures during initialization of an Edge browser / WebView instance. The Microsoft documentation provides further information on this, in particular:
- ERROR_INVALID_STATE indicates that multiple Edge instances with the same data folder but different environment options exist
- E_ABORT indicates an active abortion of the initialization process
- On any other non-OK return value than the above ones, the app should retry initialization of the instance

This change adds appropriate handling for these scenarios, consisting of uniform rollback logic when the initialization fails or is aborted and retry logic to make repeated creation attempts.

Contributes to https://github.com/eclipse-platform/eclipse.platform.swt/issues/1664

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
@@ -57,6 +57,8 @@ class Edge extends WebBrowser {
 	private static final URI URI_FOR_CUSTOM_TEXT_PAGE = setupAndGetLocationForCustomTextPage();
 	private static final String ABOUT_BLANK = "about:blank";
 
+	private static final int MAXIMUM_CREATION_RETRIES = 5;
+
 	private record WebViewEnvironment(ICoreWebView2Environment environment, ArrayList<Edge> instances) {
 		public WebViewEnvironment(ICoreWebView2Environment environment) {
 			this (environment, new ArrayList<>());
@@ -307,6 +309,10 @@ class WebViewProvider {
 		initializeWebView_13(webView);
 		webViewFuture.complete(webView);
 		return webView;
+	}
+
+	private void abortInitialization() {
+		webViewFuture.cancel(true);
 	}
 
 	private void initializeWebView_2(ICoreWebView2 webView) {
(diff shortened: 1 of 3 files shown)


HeikoKlare:
@ amartya4256 I cannot add you as a reviewer via GitHub tooling, but maybe you can still have a look at this one?

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -567,39 +573,65 @@ private String getDataDir(Display display) {
...
+		case COM.E_ABORT:
+			initializationRollback.run();
+			break;
+		default:
+			initializationRollback.run();
+			if (previousAttempts < MAXIMUM_CREATION_RETRIES) {
+				System.err.println(String.format("Edge initialization failed, retrying (attempt %d / %d)", previousAttempts + 1, MAXIMUM_CREATION_RETRIES));
+				createInstance(previousAttempts + 1);
+			} else {
+				System.err.println(String.format("Aborting Edge initialiation after %d retries", MAXIMUM_CREATION_RETRIES));
```
amartya4256, line 626: Wondering if we should also throw an error here since the initialization was not successful.

bundles/org.eclipse.swt/Eclipse SWT PI/win32/org/eclipse/swt/internal/win32/OS.java
```diff
@@ -457,6 +457,7 @@ public class OS extends C {
 	public static final int EN_CHANGE = 0x300;
 	public static final int EP_EDITTEXT = 1;
 	public static final int ERROR_FILE_NOT_FOUND = 0x2;
+	public static final int ERROR_INVALID_STATE = 0x139F;
```
amartya4256, line 460: isn't 0x8007139F also a COMExceptio and hence should be put in COM.java?
Refer these: https://github.com/MicrosoftEdge/WebView2Feedback/issues/2495, https://github.com/MicrosoftEdge/WebView2Feedback/issues/3008

amartya4256 (commented):
The changes look good to me. I have just added 2 suggestions. Also I came across these: https://github.com/MicrosoftEdge/WebView2Feedback/issues/2495, https://github.com/MicrosoftEdge/WebView2Feedback/issues/3008

It might be intersting for the error code 139f.

bundles/org.eclipse.swt/Eclipse SWT PI/win32/org/eclipse/swt/internal/win32/OS.java
HeikoKlare: `ERROR_INVALID_STATE == 0x139f` is an ordinary Windows error code (see https://learn.microsoft.com/en-us/windows/win32/debug/system-error-codes--4000-5999-), just like the other ordinary Windows error codes we have in `OS` (like `ERROR_NO_MORE_ITEMS`). So I would say it is correctly placed in `OS`.

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
HeikoKlare: Yes, makes sense. I adapted this to throw an error instead of just printing an error message with ca41f1e.

bundles/org.eclipse.swt/Eclipse SWT PI/win32/org/eclipse/swt/internal/win32/OS.java
amartya4256: Oh alright then its all good

HeikoKlare:
Thank you for your reviews!

PR #1582, opened 2024-11-09
[GTK4] Cleanup execution of non-blocking dialog calls

ptziegler:
This adds a new AsyncReadyCallback class which is used to handle the asynchronous execution of dialogs. The goal is provide a cleaner and more readable interface than what is currently available by SyncDialogUtil.

Note that this class currently simply wraps the call to SyncDialogUtil. But once all of the remaining dialogs (Color/Font/MessageDialog) have been migrated, it might make sense to remove this class entirely to avoid this additional indirection.

Follow-up to 2e61b4be73c51afe0264028534306fff874e2ea9

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/internal/AsyncReadyCallback.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/internal/AsyncReadyCallback.java
@@ -1,0 +1,38 @@
+/*******************************************************************************
+ * Copyright (c) 2024 Patrick Ziegler and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Patrick Ziegler - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.swt.internal;
+
+/**
+ * This class implements the GIO AsyncReadyCallback type and is used to
+ * transform an asynchronous {@code async} and synchronous {@code await}
+ * operation into a single synchronous {@code run} operation.
+ */
+public interface AsyncReadyCallback {
+	/**
+	 * This method is responsible for initializes the asynchronous operation
+	 *
+	 * @param callback The callback address to execute when the operation is
+	 *                 complete.
+	 */
+	void async(long callback);
+
+	/**
+	 * This method is called from within the callback function in order to
+	 * finish the executed operation and to return the result.
+	 *
+	 * @param result The generic, asynchronous function result.
+	 * @return The specific result of the operation.
+	 */
+	long await(long result);
+}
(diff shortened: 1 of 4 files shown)


ptziegler:
Motivation being that I started on migrating the FontDialog and I had a hard time understanding, what those two lambda expressions in SyncDialogUtil.run() are supposed to do. Hopefully this approach is a little bit more readable.

laeubi:
It might be good to delay the merge of this until we enabled (hopefully soon) the compilation of GTK4 so we do not run into issues with newer GTK API.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/internal/AsyncReadyCallback.java
```diff
@@ -0,0 +1,51 @@
...
+package org.eclipse.swt.internal;
+
+import org.eclipse.swt.widgets.*;
+
+/**
+ * This class implements the GIO AsyncReadyCallback type and is used to
+ * transform an asynchronous {@code async} and synchronous {@code await}
+ * operation into a single synchronous {@code run} operation.
+ */
+public abstract class AsyncReadyCallback {
```
laeubi, line 23: Maybe make this an interface and put the run method into SyncDialogUtils (or make it default implemented)

ptziegler: To be honest, I'd really like to get rid of the SyncDialogUtils class, which is why I didn't move the run method there. But the best way can also be decided at a later stage, I suppose...
In any case, this class is now an interface and the old run() method in SyncDialogUtils has been replaced.

ptziegler:
> It might be good to delay the merge of this until we enabled (hopefully soon) the compilation of GTK4 so we do not run into issues with newer GTK API.

Wouldn't this be more relevant for https://github.com/eclipse-platform/eclipse.platform.swt/pull/1583 given, that I don't change the native bindings here? Or would this then also execute the tests in a GTK4 environment?

akurtakov:
Would you please share how you envision getting rid of SyncDialogUtils? Without a bit longer term strategy info this PR looks like syntax change only.

ptziegler:
> Would you please share how you envision getting rid of SyncDialogUtils? Without a bit longer term strategy info this PR looks like syntax change only.

The SyncDialogUtils is currently only used for GTK4 and only for the dialogs.

image

It currently only defines the following methods:
- `static public long run(Display display, Consumer<Long> asyncOpen, Function<Long, Long> asyncFinish)`
- `static public int run(Display display, long handle, boolean isNativeDialog)`

The former is what's being replaced by this PR and the method that's used for the migrated GTK4 dialogs, while other one is used by the other (GTK4) dialogs.

The way the latter works is that it opens the dialog and then blocks until it is closed again. Afterwards, the dialog result is fetched. This approach doesn't really work with the new dialogs, as the result can only be fetched while the dialog closes and only within a callback method, which is why I initially added the other method.

This also means that once the other dialogs have been migrated, the old method is no longer needed and can be removed. With only one other method left, I would then simply move the callback mechanism into the AsyncReadCallback class and delete the SyncDialogUtil class alltogether, in order to remove this additional indirection.

akurtakov:
Thanks for the explanation. Sounds like a good plan.

PR #1586, opened 2024-11-11
Install libgtk-4-dev

laeubi:
As all build-maschines are now have libgtk-4-dev we should do the same for the github verification.


laeubi:
Jenkins is unaffected by this change and the testfailures seem also be present in master.

PR #1588, opened 2024-11-11
Fix for Flickering in Embedded Apps

amartya4256:
## What it does

---
### <TL;DR>
Drastically reduces the flickering when resizing embedded components by ignoring calls to `Control::setRedraw(boolean)` for controls with embedded children.

---

This PR contributes to fixing the flicker issue in SashLayout whenever an embedded app is present inside it. The fix disallows the setredraw on controls if they contain any embedded application as their descendant. There's an issue open https://bugs.eclipse.org/bugs/show_bug.cgi?id=558392 for a long time for the resizing of the sashlayout. The current way of performing layout on mouse move works fine for eclipse widgets by calling setRedraw to false which freezes the state of UI and performing a re-layout and then setting redraw to true leading to a smooth transition of the Sashlayout when its borders are dragged by mouse but it is not the same for external Embedded applications.

The embedded applications, e.g. webview 2, excel, etc, use a different renderer because of which whenever setRedraw is set to false, the whole portion where it was drawn goes blank white and when its set back, the apps are visible again, leading to a flicker effect.

One possible solution is to never allow this setredraw to be set back and forth when there's a child with embedded application present and visible. This PR implements the same approach disallowing setRedraw to be enforcedin this case. Even if it is not called from SashLayout, if a client sets it to false, it will lead to the disappearance of the application from the display,  which is not desirable. So, the approach is to only allow it when the embedded applications are either not visible or not present.

The solution is not perfect but it definitely feels better as compared to the flickering effect. It definitely brings back a shaky effect on the min/max button as mentioned in the bug above while resizing but only when embedded apps are visible. However, this effect is also visible in these scenarios by default: https://github.com/eclipse-platform/eclipse.platform.swt/issues/1122#issuecomment-2457665894 and also when you resize the shell.

Contributes to https://github.com/eclipse-platform/eclipse.platform.swt/issues/1122

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
@@ -16,15 +16,18 @@ package org.eclipse.swt.widgets;
 
 
 import java.util.*;
+import java.util.stream.*;
 
 import org.eclipse.swt.*;
 import org.eclipse.swt.accessibility.*;
+import org.eclipse.swt.browser.*;
 import org.eclipse.swt.events.*;
 import org.eclipse.swt.graphics.*;
 import org.eclipse.swt.internal.*;
 import org.eclipse.swt.internal.gdip.*;
 import org.eclipse.swt.internal.ole.win32.*;
 import org.eclipse.swt.internal.win32.*;
+import org.eclipse.swt.ole.win32.*;
 
 /**
  * Control is the abstract superclass of all windowed user interface classes.
@@ -3631,6 +3634,21 @@ boolean setRadioSelection (boolean value) {
  */
 public void setRedraw (boolean redraw) {
 	checkWidget ();
+
+	/*
+	 * Some embedded applications like webview2 (Edge), Excel, etc have a renderer
+	 * that will draw the whole application white when setting redraw to false and
+	 * will repaint it completely when setting it to true. This causes flickering,
+	 * which is why turning redraw off is not allowed in case this control has an
+	 * embedded application as a child.
+	 *
+	 * https://github.com/eclipse-platform/eclipse.platform.swt/issues/1122
+	 */
+	boolean isShown = isVisible() && !isDisposed();
+	if (!redraw && isShown && embedsWin32Control()) {
+		return;
+	}
+
 	/*
 	 * Feature in Windows.  When WM_SETREDRAW is used to turn
 	 * off drawing in a widget, it clears the WS_VISIBLE bits
@@ -3665,6 +3683,25 @@ public void setRedraw (boolean redraw) {
 			if (handle != topHandle) OS.SendMessage (handle, OS.WM_SETREDRAW, 0, 0);
 		}
 	}
+}
+
+private boolean embedsWin32Control () {
+	if (this instanceof Browser) {
+		// The Edge browser embeds webView2
+		return (getStyle() & SWT.EDGE) != 0;
+	}
+
+	if (this instanceof OleClientSite) {
+		// OLE objects are always embedded by windows
+		return true;
+	}
+
+	// This needs to be checked AFTER OleClientSite because OleClientSite itself is a Composite
+	if (this instanceof Composite comp) {
+		return Stream.of(comp.getChildren()).anyMatch(Control::embedsWin32Control);
+	}
+
+	return false;
 }
 
 /**


fedejeanne (approved):
It works as expected and the flickering is reduced.

@sratz would you like to take a look at it before I merge it?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
```diff
@@ -3631,6 +3656,20 @@ boolean setRadioSelection (boolean value) {
...
+	 * have a renderer that will draw the whole application white
+	 * when setting redraw to false and will repaint it completely
+	 * when setting it to true. This causes flickering, which is why
+	 * turning redraw on/off is disabled in these cases.
+	 *
+	 * https://github.com/eclipse-platform/eclipse.platform.swt/issues/1122
+	 */
+	if(hasEmbeddedChildren()) {
+		return;
+	}
```
HeikoKlare, line 3671: This should only be done when `redraw==false` shouldn't it? Otherwise you may set redrawing to false when no embedded control is open, then open an embedded control, and redrawing will then never be reenabled.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
```diff
@@ -3606,6 +3614,23 @@ boolean setRadioSelection (boolean value) {
 	return false;
 }
 
+private boolean hasEmbeddedChildren() {
+	for (Control ec : embeddedControls) {
+		if (!ec.isDisposed() && ec.isVisible() && ec.isChildOf(this)) return true;
+	}
+	return false;
+}
```
HeikoKlare, line 3622: Can't we calculate this on-the-fly (via the children) instead of managing a data structure storing those widgets?

HeikoKlare (changes requested):

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
```diff
@@ -711,6 +713,9 @@ void createWidget () {
 	if ((state & PARENT_BACKGROUND) != 0) {
 		setBackground ();
 	}
+	if((style & SWT.EMBEDDED) != 0) {
```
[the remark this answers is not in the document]
fedejeanne: > Do all 'embedded' controls actually have this problem?

AFAIK, yes. You can try embedding an Excel sheet, it'll flicker too.

> Is this workaround actually needed still? I removed it for testing and cannot see any problem.

Since the logic checks for the presence of `SWT.EMBEDDED`, if you remove it then Edge will flicker. I tried and it flickers (I had to restart the application after the change, hot-code replacement didn't work).

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
fedejeanne: Done. I had to move the check to `Composite` since `Control` doesn't know its children.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
fedejeanne: Done.

fedejeanne:
@HeikoKlare I added the changes you requested and they work as expected. I have currently 2 commits just in case it's necessary to revert back to a previous implementation. All commits will be squashed before merging.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Composite.java
```diff
@@ -1183,6 +1183,40 @@ void setResizeChildren (boolean resize) {
...
+private boolean hasEmbeddedChildren() {
+	for (Control child : getChildren()) {
+		if (!child.isDisposed() && child.isVisible()) {
+
+			if ((child.style & SWT.EMBEDDED) != 0)
+				return true;
+
+			// if it's a composite, check its children too
+			if (child instanceof Composite comp && comp.hasEmbeddedChildren())
+				return true;
```
HeikoKlare, line 1213: Can we have the expression blocks encapsulated in braces?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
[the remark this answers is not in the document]
HeikoKlare: My understanding of the original question (whether the _original_ workaround is still needed) asks for the flag being required as a workaround for the focus handling, not for identifying that for the control redrawing should not be disabled.\
So rephrasing the question: with this PR, does setting the embedded flag become a workaround just for identifying that redrawing should not be disabled for the browser control? If that's the case, we might think of other properties of all controls for which redrawing should not be disabled.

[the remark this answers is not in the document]
fedejeanne: I don't know if `SWT.EMBEDDED` is still needed for the browser to receive focus (according to @sratz, it isn't) but it is needed for this PR.

What other identifier would you propose instead?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
fedejeanne: Good point. How would you suggest to do it?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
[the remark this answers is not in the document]
HeikoKlare: > but it is needed for this PR.

That's not necessarily true. From a conceptual perspective, what is needed for this PR is some identifier for controls for which redrawing must not be disabled because they simply disappear when disabling redrawing (and then lead to flickering). The current assumption was that `SWT.EMBEDDED` is a valid identifier, but if Edge just uses this flag as some workaround (that we may remove now), it does not seem to be a proper identifier but is rather another workaround to attach some (rather random) flag to controls for identifying that redrawing should not be disabled for them. If I remember correctly, this embedded flag is also used for embedding Swing controls, isn't it? And is it correct to disable redrawing for them as well?

Might be that there currently is no proper identifier for those controls for which redrawing must not be disabled. Then a solution might be to add some method to `Control` to identify whether redrawing can be disabled, which can then be properly implemented for the affected controls like a browser using Edge (and other embeddings of native controls on Windows).

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
HeikoKlare: I would simply evaluate this in the `setRedraw()` method of `Control` instead of the one of `Composite`.

fedejeanne: How would you get the children? Can you provide a snippet?

HeikoKlare: I am afraid I do not understand the question, as I would do it like it is done now, just at a different place. Best may be to use a polymorphic implementation, i.e, something like:

In `Control`:
```java
boolean hasEmbeddedChildren() {
	if (isDisposed() || !isVisible()) {
		return false;
	}
	return (this.style & SWT.EMBEDDED) != 0;
}
```

In `Composite`:
```java
boolean hasEmbeddedChildren() {
	if (super.hasEmbeddedChildren()) {
		return true;
	}
        return Stream.of(getChildren()).anyMatch(Control::hasEmbeddedChildren);
}
```

fedejeanne: Got it. I added your snippets (with some slight modifications)

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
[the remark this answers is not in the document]
fedejeanne: I get what you guys mean now, thank you for the clarification. 

I changed the code so it doesn't use `SWT.EMBEDDED` but a newly created method instead.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Composite.java
fedejeanne: The code has been removed.

fedejeanne:
I changed the name to `win32_embedsControl` so it follows the same "standard" of other OS-specific methods like `Image::win32_getHandle`.

Mac and Linux implementations are also provided.

HeikoKlare (commented):
The updated version is semantically a bit strange: There are now `win32_*` methods within cocoa/gtk implementations, which does not feel correct.
I would either ...
- make the methods OS-independent, i.e., just let them yield whether the control contains an embedded native control. But that may not be correct for cocoa/gtk either, as there also native controls may be embedded, for which the method then returns false unexpectedly.
- place the method in the win32 implementation only. Then it will probably not be possible to use a polymorphic implementation but just some central logic with a case distinction over types, but since the type hierarchy is supposed to be sealed anyway, that may be fine here.

I would probably go with the latter solution to not unnecessarily bloat the cocoa/gtk implementations. 

And I would still use a proper name for the new method in any case. `embedsControl` makes me expect "true" as a result for every composite (at least for every composite actually having children). The name should be specific to what you can actually expect from that method.

fedejeanne:
@HeikoKlare would you mind coding it? I think it would be easier, since you probably have a very specific solution in mind :-)

You have write permissions in this PR and you were also added as co-author in a previous commit so you have green light 🟢

HeikoKlare (approved):

HeikoKlare (approved):
This now looks like the most local, least invasive solution for the issue we can achieve with current understanding.

PR #1591, opened 2024-11-12
[Win32] Use PrintWindow() with PW_RENDERFULLCONTENT flag

### LINKED ISSUE #373: [Edge] Browser with SWT.EDGE returns blank image when using print()
**Describe the bug**
In our app, we have the functionality of printing images from an internal browser window. We previously used Internet Explorer as an internal browser but we changed it to Edge by default for Windows systems. Previously this functionality correctly created the image of browser composite, but it doesn't work for Edge.
**To Reproduce**
```
        browser = new Browser(composite, SWT.EDGE);
        //any URL
        browser.setUrl(file.toURI().toURL().toString())
        Image image = new Image(Display.getDefault(), browser.getBounds());
        GC gc = new GC(image);
        try {
           browser.print(gc);
        } finally {
            gc.dispose();
        }
        ImageTransfer imageTransfer = ImageTransfer.getInstance();
        Clipboard clipboard = new Clipboard(Display.getCurrent());
        clipboard.setContents(new Object[] {image.getImageData()}, new Transfer[]{imageTransfer});

```
**Expected behavior**
Correct image with browser content shown on it.

**Environment:**

- - [X] Windows

Windows 10, Edge installed 
3. JRE/JDK version
java.vendor.version=Temurin-11.0.15+10
java.version=11.0.15
4. Version
found on:
org.eclipse.swt.browser.EdgeVersion=105.0.1343.27
 
Probably screenshots can be done via browser functionality. But I was not able to do that.

sratz:
Resolves #373.

I came up with this solution with some google research. It seems to work well, but I'm quite clueless about those Win32 APIs, so would be great if a Win32 expert could help out here.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT PI/win32/org/eclipse/swt/internal/win32/OS.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT PI/win32/org/eclipse/swt/internal/win32/OS.java
@@ -39,6 +39,7 @@ public class OS extends C {
 	/**
 	 * Values taken from https://en.wikipedia.org/wiki/List_of_Microsoft_Windows_versions
 	 */
+	public static final int WIN32_BUILD_WIN8_1 = 9600; // "Windows 8.1"
 	public static final int WIN32_BUILD_WIN10_1607 = 14393; // "Windows 10 August 2016 Update"
 	public static final int WIN32_BUILD_WIN10_1809 = 17763; // "Windows 10 October 2018 Update"
 	public static final int WIN32_BUILD_WIN10_2004 = 19041; // "Windows 10 May 2020 Update"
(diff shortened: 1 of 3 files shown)


DenisUngemach:
Hi, the undocumented flag PW_RENDERFULLCONTENT seems to be necessary for printWindow.
Also chromium does not really know why:
https://chromium.googlesource.com/chromium/src.git/+/62.0.3178.1/ui/snapshot/snapshot_win.cc
> Without that flag PrintWindow may not correctly capture what's actually onscreen.

sratz:
This appears to be working well.

Nobody in Platform is actually calling `Contro.print (GC)`, so it's hard to test that more intensively. 

My proposal would be to get that in in M1 and hope for (no negative) feedback from downstream consumers elsewhere.

HeikoKlare (commented):
The change sounds reasonable, in particular also considering the Chromium implementation, but I am no expert on this either. We also do not make any use of the `print` method in our product, so I cannot assist with any testing.

> My proposal would be to get that in in M1 and hope for (no negative) feedback from downstream consumers elsewhere.

I agree. Let's try to fix this as early as possible in a release cycle. What about explicitly pinging the people involved in #373 to test the fix and give feedback? Otherwise I think we might not get any (negative) feedback simply because of the lack of testing.

HeikoKlare (approved):
Let's give it a try.

PR #1593, opened 2024-11-13
Scaling of Color/Font Dialog not working properly after DPI change

ShahzaibIbrahim:
Creating a separate UI thread to set the DPI Awareness context to use UNAWARE GDI scale before opening the either dialogs since windows has issues with dialog scaling in AWARE context

## HOW TO TEST
```
-Dswt.autoScale=quarter
```
- Run the Runtimeworkspace
- Go to Window -> Preferences -> General -> Appearance -> Colors and Fonts
- Collapse **Basic** in the main window
- Double click on any option (to open either color or font dialog) 
- Move the dialog across multiple monitor 
- See if the dialog is scaling correctly on multiple zoom level monitors

## Before Fix

**Case: Primary monitor 200%, Secondary monitor 100%**

100%: 
image
image

200%:
image
image

## After Fix

100%:
image
image

200%:
image
image

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/ColorDialog.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/ColorDialog.java
@@ -233,54 +233,69 @@ public RGB open () {
 
 	display.externalEventLoop = true;
 	display.sendPreExternalEventDispatchEvent ();
-	/* Open the dialog */
-	boolean success = OS.ChooseColor (lpcc);
-	display.externalEventLoop = false;
-	display.sendPostExternalEventDispatchEvent ();
-
-	/* Clear the temporary dialog modal parent */
-	if ((style & (SWT.APPLICATION_MODAL | SWT.SYSTEM_MODAL)) != 0) {
-		display.setModalDialog (oldModal);
-	}
-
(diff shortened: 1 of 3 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/ColorDialog.java
```diff
@@ -184,6 +184,9 @@ public RGB open () {
 
 	/* Allocate the Custom Colors and initialize to white */
 	Display display = parent.display;
+	if(display.isRescalingAtRuntime()) {
+		OS.SetThreadDpiAwarenessContext(OS.DPI_AWARENESS_CONTEXT_UNAWARE_GDISCALED);
```
akoch-yatta, line 188: Two remarks: 

1. I would put that code directly before the place where the dialog is opened, directly above`boolean success = OS.ChooseColor (lpcc);`
2. You need to revert the change of the DPI awareness afterwards, otherwise everything created after it, will use this awareness as well. SetThreadDpiAwarenessContext returns the previous awareness context, that you can store and reapply after the change

Same comment applies to the font dialog

fedejeanne (approved):

fedejeanne:
Failing tests and checks are unrelated.

Issue #1592, opened 2024-11-13
[Edge] Browser.clearSessions() throws exception when no Edge is open


N1k145:
**Describe the bug**
Exception is thrown when `Browser.clearSessions()` is called before the Browser is rendered.
This is a new behaviour in the current I Build compared to the last release (2024-09)

```
!ENTRY org.eclipse.ui.browser 4 0 2024-11-13 12:51:38.734
!MESSAGE Not implemented [WebView2: cookie access requires a Browser instance]
!STACK 0
org.eclipse.swt.SWTError: Not implemented [WebView2: cookie access requires a Browser instance]
	at org.eclipse.swt.SWT.error(SWT.java:4947)
	at org.eclipse.swt.browser.Edge.getCookieManager(Edge.java:474)
	at org.eclipse.swt.browser.Edge.lambda$0(Edge.java:86)
	at org.eclipse.swt.browser.WebBrowser.clearSessions(WebBrowser.java:265)
	at org.eclipse.swt.browser.Browser.clearSessions(Browser.java:220)
	at org.eclipse.ui.internal.browser.BrowserViewer.<init>(BrowserViewer.java:240)
	at org.eclipse.ui.internal.browser.WebBrowserView.createPartControl(WebBrowserView.java:51)
	at org.eclipse.ui.internal.e4.compatibility.CompatibilityPart.createPartControl(CompatibilityPart.java:158)
```

**To Reproduce**
Change the Code of the BrowserViewer:
![grafik](https://github.com/user-attachments/assets/eb23c909-ed87-4edd-80e3-02b5e9f77ac9)
(Also change the catch block to report the exception)
Try to open the Internal Browser View

**Expected behavior**
No exception is thrown, like in the last release. 
When the call can not work, maybe log a warning/error instead

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 

2. Additional OS info (e.g. OS version, Linux Desktop, etc)
Edge Browser with `-Dorg.eclipse.swt.browser.DefaultType=edge`

3. JRE/JDK version
 org.eclipse.justj.openjdk.hotspot.jre.full.win32.x86_64_22.0.2.v20240802-1626

**Version since**
I20241112
Did not test earlier I Builds
2024-09 was fine

fedejeanne:
[...]

PR #1595, opened 2024-11-13
Store Edge instance in Environemnt record sync-ly

amartya4256:
This PR contributes to storing the edge instance on creation in the webenvironment record synchronously for the sake of the consistency of the record. Since the record can be read by the cookie manager for several purposes on client calls, we need to make sure the record is in a consistent state, i.e. Linking the Edge instances (completely initialized or not) to their webview environment in the create method (synchronously).

In the bug #1592 the error occurs because we wait for the browser to initialize before it can be added to the record. It is safe to add it synchronously in the record since, all the further API calls of Edge browser is guarded by waitForFutureToFinish method.

contributes to https://github.com/eclipse-platform/eclipse.platform.swt/issues/1592

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
@@ -568,6 +568,7 @@ private String getDataDir(Display display) {
 @Override
 public void create(Composite parent, int style) {
 	containingEnvironment = createEnvironment();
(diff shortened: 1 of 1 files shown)


amartya4256:
> I did test this with my actual product and it works now like before. Thank you very much for the quick fix @amartya4256

Since we are adding it early, we also need to remove the instance if the initialization fails. I have amended the commit which includes the removal on failure as well.

akurtakov:
As this change is in "opt-in" functionality, fixes regression and reporter verified it : PMC +1

HeikoKlare:
Thank you for the immediate approval! The failing tests are documented: #1564

PR #1603, opened 2024-11-20
Non system cursors adaption to DPI scale changes

ShahzaibIbrahim:
Creating a map to provide different handle w.r.t the current zoom of the display. The method win32_getHandle provides the appropriate handle for the scaled image as per requested by the client.

## HOW TO TEST
- Run the `Snippet92` (Image) or `Snippet119` (Icon) with the following **VM Arguments**
```
-Dswt.autoScale=quarter
```
- Run the example
- Move the mouse pointer to the window
- See if the mouse pointer is scaled to the current monitor zoom level. 
- Now move the window to another monitor with different zoom level (i.e. 200%)
- Move the mouse pointer to the window again
- See if the pointer still scaled correctly according to 200% zoom level

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
@@ -14,7 +14,10 @@
 package org.eclipse.swt.graphics;
 
 
+import java.util.*;
+
 import org.eclipse.swt.*;
+import org.eclipse.swt.internal.*;
 import org.eclipse.swt.internal.win32.*;
 
 /**
@@ -56,17 +59,30 @@ public final class Cursor extends Resource {
 	 * platforms and should never be accessed from application code.
 	 * </p>
 	 *
-	 * @noreference This field is not intended to be referenced by clients.
 	 */
-	public long handle;
+	private long handle;
+	/**
+	 * Attribute to cache current native zoom level
+	 */
+	private static final int DEFAULT_ZOOM = 100;
+
+	private HashMap<Integer, Long> zoomLevelToHandle = new HashMap<>();
 
 	boolean isIcon;
-
+	private final ImageData source;
+	private final ImageData mask;
+	private final int hotspotX;
+	private final int hotspotY;
 /**
  * Prevents uninitialized instances from being created outside the package.
  */
 Cursor(Device device) {
 	super(device);
+	this.source = null;
+	this.mask = null;
+	this.hotspotX = -1;
+	this.hotspotY = -1;
+	this.device.registerResourceWithZoomSupport(this);
 }
 
 /**
@@ -116,7 +132,7 @@ Cursor(Device device) {
  * @see #dispose()
  */
 public Cursor(Device device, int style) {
-	super(device);
+	this(device);
 	long lpCursorName = 0;
 	switch (style) {
 		case SWT.CURSOR_HAND: 		lpCursorName = OS.IDC_HAND; break;
@@ -184,6 +200,10 @@ public Cursor(Device device, int style) {
  */
 public Cursor(Device device, ImageData source, ImageData mask, int hotspotX, int hotspotY) {
 	super(device);
+	this.source = source;
+	this.mask = mask;
+	this.hotspotX = hotspotX;
+	this.hotspotY = hotspotY;
 	if (source == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
 	if (mask == null) {
 		if (source.getTransparencyType() != SWT.TRANSPARENCY_MASK) {
@@ -213,6 +233,7 @@ public Cursor(Device device, ImageData source, ImageData mask, int hotspotX, int
 	handle = OS.CreateCursor(hInst, hotspotX, hotspotY, source.width, source.height, sourceData, maskData);
 	if (handle == 0) SWT.error(SWT.ERROR_NO_HANDLES);
 	init();
+	this.device.registerResourceWithZoomSupport(this);
 }
 
 /**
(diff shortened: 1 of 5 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
```diff
@@ -59,9 +62,18 @@ public final class Cursor extends Resource {
 	 * @noreference This field is not intended to be referenced by clients.
 	 */
 	public long handle;
```
fedejeanne, line 64: Why not make this one `private` and let every consumer go through the new static method `win32_getHandle(...)`?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
```diff
@@ -319,6 +338,55 @@ public Cursor(Device device, ImageData source, int hotspotX, int hotspotY) {
...
+	} else {
+		ImageData source = DPIUtil.scaleImageData(cursor.device, cursor.source, zoom, DEFAULT_ZOOM);
+		if (cursor.isIcon) {
+			Cursor newCursor = new Cursor(cursor.device, source, cursor.hotspotX, cursor.hotspotY);
+			cursor.setHandleForZoomLevel(newCursor.handle, zoom);
+		} else {
+			ImageData mask = DPIUtil.scaleImageData(cursor.device, cursor.mask, zoom, DEFAULT_ZOOM);
+			Cursor newCursor = new Cursor(cursor.device, source, mask, cursor.hotspotX, cursor.hotspotY);
+			cursor.setHandleForZoomLevel(newCursor.handle, zoom);
+		}
```
fedejeanne, line 376:
```suggestion
		ImageData mask = cursor.isIcon ? DPIUtil.scaleImageData(cursor.device, cursor.mask, zoom, DEFAULT_ZOOM) : null;
		Cursor newCursor = new Cursor(cursor.device, source, mask, cursor.hotspotX, cursor.hotspotY);
		cursor.setHandleForZoomLevel(newCursor.handle, zoom);
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
```diff
@@ -319,6 +338,55 @@ public Cursor(Device device, ImageData source, int hotspotX, int hotspotY) {
...
+ * @param cursor the cursor to get the handle of
+ * @param zoom device zoom in % of the monitor on which the image is painted
+ *
+ * @return true if cursor is refreshed
+ *
+ * @noreference This method is not intended to be referenced by clients.
+ */
+public static Long win32_getHandle (Cursor cursor, int zoom) {
+	if (cursor.isDisposed()) {
+		return cursor.handle;
```
fedejeanne, line 359: Why when it's disposed? Shouldn't it be:
```suggestion
	if (!cursor.isDisposed()) {
		return cursor.handle;
```

?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
```diff
@@ -319,6 +338,55 @@ public Cursor(Device device, ImageData source, int hotspotX, int hotspotY) {
...
+ * can be shared within the packages provided by SWT. It is not
+ * available on all platforms, and should never be called from
+ * application code.
+ *
+ * Updates zoom and refresh the Cursor based on the native zoom level, if required.
+ *
+ * @param cursor the cursor to get the handle of
+ * @param zoom device zoom in % of the monitor on which the image is painted
+ *
+ * @return true if cursor is refreshed
```
fedejeanne, line 353:
```suggestion
 * Get the handle for a cursor given a zoom level.
 *
 * @param cursor the cursor .
 * @param zoom zoom level (in %) of the monitor the cursor is currently in.
 *
 * @return The handle of the cursor.
```

fedejeanne (commented):
I didn't test this, I only have some proposals regarding the coding style and JavaDoc. 

I also have 2 questions regarding returning a handle of a cursor that is already disposed and about making the field `handle` private.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
ShahzaibIbrahim: This suggestion won't work as we are using different constructor for both the conditions. It's not just about using the mask imageData

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
ShahzaibIbrahim: For non-disposed case we have the handling. We add the condition to return the handle even if the resource is disposed. Same as its done with Image.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
fedejeanne: Hm, you're right. I assumed that the constructors would be fairly similar but they are not.

Scratch my suggestion.

fedejeanne (commented):
Changes look good, they produce the expected result and I see no regressions either.

Please remove the extra (unnecessary) call to `registerResourceWithZoomSupport(this)` from the constructor, squash and rebase your commits and then this PR is good to go ✔️

fedejeanne (approved):
Test failures are unrelated

Issue #1606, opened 2024-11-21
[Gtk3] Bug: SWTException on Table.remove(...) for a row with focus


om-11-2024:
**Describe the bug**
`SWTException` on `Table.remove(...)` for a row with focus.

**To Reproduce**
A standalone SWT snippet to reproduce the behavior:
```java
import org.eclipse.swt.SWT;
import org.eclipse.swt.layout.*;
import org.eclipse.swt.widgets.*;

public class SwtBugSnippet {

  public static void main( String[] args ) {
    var display = Display.getDefault();
    var shell = new Shell( display );
    shell.setLayout( new GridLayout() );

    var button = new Button( shell, SWT.PUSH );
    button.setText( "GO" );

    var table = new Table( shell, SWT.VIRTUAL );
    var layoutData = new GridData( 300, 300 );
    layoutData.grabExcessHorizontalSpace = layoutData.grabExcessVerticalSpace = true;
    table.setLayoutData( layoutData );
    table.setItemCount( 2 );
    table.addListener( SWT.SetData, event -> {
      var item = ( TableItem ) event.item;
      item.setText( "Item #" + System.identityHashCode( item ) );
    } );

    shell.pack();
    shell.open();

    table.setFocus();

    button.addListener( SWT.Selection, e -> {
      table.clear( 0 );
      table.remove( new int[] { 0 } );
    } );

    while ( !shell.isDisposed() ) {
      if ( !display.readAndDispatch() ) {
        display.sleep();
      }
    }
  }
}
```
Just press the "GO" button and the program terminates after printing this to stderr:
```
Exception in thread "main" org.eclipse.swt.SWTException: Widget is disposed
	at org.eclipse.swt.SWT.error(SWT.java:4922)
	at org.eclipse.swt.SWT.error(SWT.java:4837)
	at org.eclipse.swt.SWT.error(SWT.java:4808)
	at org.eclipse.swt.widgets.Widget.error(Widget.java:565)
	at org.eclipse.swt.widgets.Widget.checkWidget(Widget.java:480)
	at org.eclipse.swt.widgets.TableItem.setText(TableItem.java:1363)
	at SwtBugSnippet.lambda$0(SwtBugSnippet.java:24)
	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:89)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:5855)
[...]

PR #1608, opened 2024-11-21
[Gtk3] fix for SWTException on Table.remove(...) with focused row #1606

om-11-2024:
Fix for `SWTException` on `Table.remove(...)` with a focused row.  
Cause: `gtk_list_store_remove(...)` for a focused row invokes `Table.cellDataProc(...)` with not-yet-updated `Table.items`.  
Fix: remove `TableItem` from `Table.items` before `gtk_list_store_remove(...)`.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/1606

This bug happens in Gtk3.  
I haven't tried to reproduce this bug in GTK4.

Java stacktrace:
```java
org.eclipse.swt.SWTException: Widget is disposed
	at org.eclipse.swt.SWT.error(SWT.java:4922)
	at org.eclipse.swt.SWT.error(SWT.java:4837)
	at org.eclipse.swt.SWT.error(SWT.java:4808)
	at org.eclipse.swt.widgets.Widget.error(Widget.java:597)
	at org.eclipse.swt.widgets.Widget.checkWidget(Widget.java:512)
	at org.eclipse.swt.widgets.TableItem.setText(TableItem.java:1363)
	at org.eclipse.swt.tests.gtk.Test_Gtk3_Table_Remove_Focused_Row.lambda$3(Test_Gtk3_Table_Remove_Focused_Row.java:68)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:5857)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1617)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1643)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1626)
	at org.eclipse.swt.widgets.Table.checkData(Table.java:289)
	at org.eclipse.swt.widgets.Table.cellDataProc(Table.java:227)
	at org.eclipse.swt.widgets.Display.cellDataProc(Display.java:995)
	at org.eclipse.swt.internal.gtk.GTK.gtk_list_store_remove(Native Method)
	at org.eclipse.swt.widgets.Table.remove(Table.java:2668)
	...
```

The main cause of the error is that:
 1. when a row with focus is deleted with `gtk_list_store_remove(...)`, then cell data function `Table.cellDataProc(...)` is called for a row after the removed one
 2. but inside `Table.cellDataProc(...)` `Table.items` isn't yet updated , therefore `Table.cellDataProc(...)` operates with `TableItem`, which is already disposed but not yet removed from `Table.items`

Inside `gtk_list_store_remove(...)` the row is removed from the `GtkTreeModel`, and only then `GtkTreeView` callbacks are invoked. It means that when `Table.cellDataProc(...)` is invoked, the row has already been removed in Gtk3.

The fix: remove `TableItem` from `Table.items` before `gtk_list_store_remove(...)` instead of after.

Os:  Ubuntu 24.04.1 LTS
Gtk: 3.24.41
Swt: 4.967.8

=================================================

Why `gtk_list_store_remove(...)` invokes `Table.cellDataProc(...)` only for a row with focus and not for ordinary rows?

The reason is that in gtk3 when a row with focus is deleted, the next row receives focus, which among other things creates `GtkCellAccessible` (a "cell with focus" for assistive technology in gtk3).  
Creation of `GtkCellAccessible` invokes cell data function `Table.cellDataProc(...)` - just like it happens when a standard cell in `GtkTreeView` gets rendered.

Here is the stacktrace in gtk3 code:
```c
apply_cell_attributes() at gtkcellarea.c:1257
g_hash_table_foreach() at ghash.c:2117
gtk_cell_area_real_apply_attributes() at gtkcellarea.c:1286
gtk_cell_area_box_apply_attributes() at gtkcellareabox.c:1310
_gtk_marshal_VOID__OBJECT_BOXED_BOOLEAN_BOOLEANv() at gtkmarshalers.c:5447
g_type_class_meta_marshalv() at gclosure.c:1062
_g_closure_invoke_va() at gclosure.c:897
signal_emit_valist_unlocked() at gsignal.c:3424
g_signal_emit_valist() at gsignal.c:3263
g_signal_emit() at gsignal.c:3583
gtk_cell_area_apply_attributes() at gtkcellarea.c:2373
gtk_tree_view_column_cell_set_cell_data() at gtktreeviewcolumn.c:2821
set_cell_data() at gtktreeviewaccessible.c:347
create_cell() at gtktreeviewaccessible.c:439
_gtk_tree_view_accessible_add_state() at gtktreeviewaccessible.c:2053
gtk_tree_view_real_set_cursor() at gtktreeview.c:13377
gtk_tree_view_row_deleted() at gtktreeview.c:9430
g_cclosure_marshal_VOID__BOXED() at gmarshal.c:1628
g_closure_invoke() at gclosure.c:834
signal_emit_unlocked_R() at gsignal.c:3888
signal_emit_valist_unlocked() at gsignal.c:3520
g_signal_emit_valist() at gsignal.c:3263
g_signal_emit() at gsignal.c:3583
gtk_tree_model_row_deleted() at gtktreemodel.c:1914
gtk_list_store_remove() at gtkliststore.c:1219
Java_org_eclipse_swt_internal_gtk_GTK_gtk_1list_1store_1remove()
```

The code that leads to execution of cell data function for a row with focus is in gtktreeview.c:
```
static void
gtk_tree_view_row_deleted (GtkTreeModel *model,
			   GtkTreePath  *path,
			   gpointer      data)
{
  ...
  /* If the cursor row got deleted, move the cursor to the next row */
  if (tree_view->priv->cursor_node &&
      (tree_view->priv->cursor_node == node ||
       (node->children && (tree_view->priv->cursor_tree == node->children ||
                           _gtk_rbtree_contains (node->children, tree_view->priv->cursor_tree)))))
    {
      ...
      cursor_changed = TRUE;
    }
  ...
  if (cursor_changed)
    {
      if (cursor_node)
        {
          ...
          gtk_tree_view_real_set_cursor (tree_view, cursor_path, CLEAR_AND_SELECT | CURSOR_INVALID);
          ...
        }
      ...
    }
  ...
}
```

### DIFF
--- a/tests/org.eclipse.swt.tests.gtk/JUnit Tests/org/eclipse/swt/tests/gtk/Test_Gtk3_Table_Remove_Focused_Row.java
+++ b/tests/org.eclipse.swt.tests.gtk/JUnit Tests/org/eclipse/swt/tests/gtk/Test_Gtk3_Table_Remove_Focused_Row.java
@@ -1,0 +1,93 @@
+package org.eclipse.swt.tests.gtk;
+
+import org.eclipse.swt.SWT;
+import org.eclipse.swt.internal.gtk.GTK;
+import org.eclipse.swt.layout.FillLayout;
+import org.eclipse.swt.widgets.Shell;
+import org.eclipse.swt.widgets.Table;
+import org.eclipse.swt.widgets.TableItem;
+import org.junit.After;
+import org.junit.Assume;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+public class Test_Gtk3_Table_Remove_Focused_Row {
+
+	private Shell shell;
+	private Table table;
+
+	@BeforeClass
+	public static void setUpClass() {
+		var isGtk3 = false;
+		if ("gtk".equals(SWT.getPlatform())) {
+			@SuppressWarnings("restriction")
+			var gtkMajorVersion = GTK.GTK_VERSION >>> 16;
+			isGtk3 = (gtkMajorVersion == 3);
+		}
+		Assume.assumeTrue("Test is only for Gtk3.", isGtk3);
+	}
+
+	@Before
+	public void setUp() {
+		shell = new Shell();
+		shell.setMinimumSize(400, 400);
+		shell.setLayout(new FillLayout());
+	}
+
+	@After
+	public void tearDown() {
+		if (table != null) {
+			table.dispose();
+		}
+		if (shell != null) {
+			shell.dispose();
+		}
+	}
+
+	@Test
+	public void test_remove_focused_row_remove_one() {
+		test_remove_focused_row_remove_method_arg(() -> table.remove(0));
+	}
+
+	@Test
+	public void test_remove_focused_row_remove_array() {
+		test_remove_focused_row_remove_method_arg(() -> table.remove(new int[] { 0 }));
+	}
+
+	@Test
+	public void test_remove_focused_row_remove_interval() {
+		test_remove_focused_row_remove_method_arg(() -> table.remove(0, 0));
+	}
+
+	private void test_remove_focused_row_remove_method_arg(Runnable removeRow0) {
+		table = new Table(shell, SWT.VIRTUAL);
+		table.setItemCount(2);
+		table.addListener(SWT.SetData, event -> {
+			var item = (TableItem) event.item;
+			item.setText("Item #" + System.identityHashCode(item) + " " + item.toString());
+		});
+
+		shell.pack();
+		shell.open();
+
+		processUiEvents();
+
+		// set focus on row[0]
+		table.setFocus();
+
+		processUiEvents();
+
+		table.clear(0);
+		removeRow0.run();
+
+		processUiEvents();
+	}
+
+	private void processUiEvents() {
+		while (table.getDisplay().readAndDispatch()) {
+			// continue to next event
+		}
+	}
+
+}
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Table.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Table.java
@@ -1124,11 +1124,11 @@ void destroyItem (TableItem item) {
 	}
 	if (index == itemCount) return;
 	long selection = GTK.gtk_tree_view_get_selection (handle);
+	System.arraycopy (items, index + 1, items, index, --itemCount - index);
+	items [itemCount] = null;
 	OS.g_signal_handlers_block_matched (selection, OS.G_SIGNAL_MATCH_DATA, 0, 0, 0, 0, CHANGED);
 	GTK.gtk_list_store_remove (modelHandle, item.handle);
 	OS.g_signal_handlers_unblock_matched (selection, OS.G_SIGNAL_MATCH_DATA, 0, 0, 0, 0, CHANGED);
-	System.arraycopy (items, index + 1, items, index, --itemCount - index);
-	items [itemCount] = null;
 	if (itemCount == 0) resetCustomDraw ();
 }
 
@@ -2664,11 +2664,11 @@ public void remove (int index) {
 	}
 	if (!disposed) {
 		long selection = GTK.gtk_tree_view_get_selection (handle);
+		System.arraycopy (items, index + 1, items, index, --itemCount - index);
+		items [itemCount] = null;
 		OS.g_signal_handlers_block_matched (selection, OS.G_SIGNAL_MATCH_DATA, 0, 0, 0, 0, CHANGED);
 		GTK.gtk_list_store_remove (modelHandle, iter);
 		OS.g_signal_handlers_unblock_matched (selection, OS.G_SIGNAL_MATCH_DATA, 0, 0, 0, 0, CHANGED);
-		System.arraycopy (items, index + 1, items, index, --itemCount - index);
-		items [itemCount] = null;
 	}
 	OS.g_free (iter);
 }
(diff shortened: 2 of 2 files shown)


om-11-2024:
FYI:
- `Table.removeAll()` is not affected.
   Internally `GTK.gtk_list_store_clear()`[calls `gtk_list_store_remove()` in a loop](https://gitlab.gnome.org/GNOME/gtk/-/blob/77ebdd85091833a7869ece48c3114fa6d9966321/gtk/gtkliststore.c?page=2#L1429-L1434)  (this is GTK3.24.41 - current GTK3 version for Ubuntu 24.04LTS) .  
  As a result, if some row has focus, then during `gtk_list_store_clear()` the focus moves to different rows and `Table.cellDataProc()` is invoked.  
   Fortunately, `Table.removeAll()` sets `items = new TableItem [4]` before calling `GTK.gtk_list_store_clear()`, as a result there are no cached+disposed `TableItem`s in `Table.items` when `Table.cellDataProc()` is executed => `SWTError` isn't thrown by `TableItem.setText()`.
- `Tree` is not affected, because `Tree.items` are never out-of-sync with rows in `GtkTreeView`. (that's because `Tree` uses different mapping scheme for items (e.g. it uses `ID_COLUMN`))

tests/org.eclipse.swt.tests.gtk/JUnit Tests/org/eclipse/swt/tests/gtk/Test_Gtk3_Table_Remove_Focused_Row.java
```diff
@@ -0,0 +1,93 @@
...
+
+	@BeforeClass
+	public static void setUpClass() {
+		var isGtk3 = false;
+		if ("gtk".equals(SWT.getPlatform())) {
+			@SuppressWarnings("restriction")
+			var gtkMajorVersion = GTK.GTK_VERSION >>> 16;
+			isGtk3 = (gtkMajorVersion == 3);
+		}
+		Assume.assumeTrue("Test is only for Gtk3.", isGtk3);
```
basilevs, line 28: Will this test fail on other platforms? If not, why should not it run there?

om-11-2024: The platform (as returned by  `org.eclipse.swt.SWT.getPlatform()` ) is "gtk".
All the test is in `tests/gtk/...`  are executed only on "gtk" platform.

This test is in `tests/gtk/...` because it tests a bug in the swt implementation for gtk.

For now swt uses gtk3 + switching to gtk4 is in progress.
In gtk4 `GtkTreeView` is [deprecated](https://docs.gtk.org/gtk4/section-tree-widget.html), which means that the code for `Table` for gtk4 will be rewritten.
That's why this test and this fix are only for gtk3.

om-11-2024: Additionally.  
In GTK3 this bug involves `GtkCellAccessible` from the so called Accessibility toolkit (ATK) used by GTK3.  
In GTK4 accessibility [implemented differently](https://blog.gtk.org/2020/10/21/accessibility-in-gtk-4/) and ATK isn't used anymore.  
As a result, even if SWT would continue to use deprecated `GtkTreeView` in GTK4, this bug would not happen there.

laeubi:
@om-11-2024 interesting observation, have you checked if a similar problem exits in the `Tree` widget?

akurtakov:
Issue is fixed in alternative way https://github.com/eclipse-platform/eclipse.platform.swt/pull/2472

om-11-2024:
@laeubi 

> interesting observation, have you checked if a similar problem exits in the Tree widget?

`Tree` doesn't suffer from this problem.  
That's because `Tree` implements the association of swt `Item`s and gtk rows differently (look at the code that gets `TableItem`/`TreeItem` in `cellDataProc(...)` methods in `Table`/`Tree`):
- in `Table` it's assumed that index of the gtk row == index of `TableItem` in `Table.items`
- in `Tree` each gtk row has unique id that is stored in `ID_COLUMN`column, this unique id is the index of the associated `TreeItem` in `Tree.items`

What causes the bug:
- when the focused row is removed, gtk makes the next row focused and invokes `cellDataProc(...)` for that next row
- inside `cellDataProc(...)` the code gets the associated `Item` while the field `items` isn't updated yet and contain old disposed `Item`s
  - for `Tree` everything works fine because index of `TreeItem` is taken from `ID_COLUMN` of the row
  - for `Table` we get a problem because `Table.items` contains an already removed and disposed `TableItem` at the index of the row

om-11-2024:
@akurtakov 
> Issue is fixed in alternative way #2472

#1606 is fixed. 
But, as I understand, this still has problems.

Let's see what happens:
1. inside `Table.remove()` the gtk row is removed by `GTK.gtk_list_store_remove(...)`
2. inside `gtk_list_store_remove(...)`: because we remove row with focus, gtk makes the next row focused and invokes `Table.cellDataProc(...)` on this next row
3. inside `Table.cellDataProc(...)` the code finds the associated `TableItem` as `Table.items[idx_of_gtk_row]`
   - was: since `Table.items` aren't updated yet, the `TableItem` at the index is the removed `TableItem` instead of the `TableItem` for the new row that gets the focus
   - #2472: same as previous, but now we check if `TableItem` is disposed
   - expected: the `TableItem` for the new row that get focus should be processed

In other words #2472 fixes #1606 because the disposed `TableItem` isn't processed in `Table.cellDataProc(...)`.  
But the correct `TableItem` (i.e. the row that gets the focus) is still not processed in `Table.cellDataProc(...)` - which may cause problems.

PR #1611, opened 2024-11-22
[GTK3] Test Tree crash in SetData event

basilevs:
The failure only happens in Ubuntu 24 (GTK 3.24.41)

This is a test for #678 

The fix is in #1612.

PR #1612, opened 2024-11-22
[GTK3] Do not crash in SetData event

### LINKED ISSUE #678: JVM crash in `Tree.cellDataProc` when calling `TreeItem.setImage`
**Describe the bug**

Virtual `Tree`  widget crashes JVM (through GTK3) when setting `Image` onto `TreeItem` in `Listener` to `SWT.SetData` .
This is a continuation of https://bugs.eclipse.org/bugs/show_bug.cgi?id=573090.

**To Reproduce**

```java
package swt_tree_crash;

import org.eclipse.swt.SWT;
import org.eclipse.swt.widgets.Event;
import org.eclipse.swt.widgets.Listener;
import org.eclipse.swt.widgets.Tree;
import org.eclipse.swt.widgets.TreeItem;

public class Main {

        Tree tree = new Tree(shell, SWT.VIRTUAL);
        tree.addListener(SWT.SetData, new Listener() {
            public void handleEvent(final Event e) {
            	TreeItem item = (TreeItem)e.item;
                item.setText(0, "A");
                item.setImage(new Image(display, 20, 20)); // <-- this is the critical line!
            }
        });
        tree.setItemCount(1);
        
        shell.setSize(400, 300);
        shell.open();

        while (!shell.isDisposed()) {
                display.sleep();
            }
        }
        display.dispose();
    }
}
```

**Expected behavior**

The application always shows a window with a tree widget in it, containing one item, with a small white icon.

**Screenshots**

Not applicable.

**Environment:**

- - [ ] All OS
- - [ ] Windows
- - [X] Linux
- - [ ] macOS 

Operating System: Fedora Linux 38
KDE Plasma Version: 5.27.4
KDE Frameworks Version: 5.105.0
Qt Version: 5.15.9
Kernel Version: 6.2.15-300.fc38.x86_64 (64-bit)
Graphics Platform: Wayland
Processors: 16 × AMD Ryzen 7 3800X 8-Core Processor
Memory: 31.2 GiB of RAM
Graphics Processor: AMD Radeon RX 550 Series

3. JRE/JDK version

JRE version: OpenJDK Runtime Environment Temurin-17.0.7+7 (17.0.7+7) (build 17.0.7+7)
Java VM: OpenJDK 64-Bit Server VM Temurin-17.0.7+7 (17.0.7+7, mixed mode, tiered, compressed oops, compressed class ptrs, g1 gc, linux-amd64)

**Version since**

Tested on Eclipse 4.27 (SWT 4.958).

Skipping the `TreeItem.setImage` call avoids the crash.

basilevs:
The crash happens when renderers are replaced during render. If replacement is postponed, the crash no longer happens. The failure only happens in Ubuntu 24 (GTK 3.24.41).

Fixes #678.
Is tested by #1611.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/TreeItem.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/TreeItem.java
@@ -1518,69 +1518,81 @@ public void setImage(int index, Image image) {
 		surface = imageList.getSurface(imageIndex);
 		pixbuf = ImageList.createPixbuf(surface);
 	}
-
-	int modelIndex = parent.columnCount == 0 ? Tree.FIRST_COLUMN : parent.columns [index].modelIndex;
-	long parentHandle = parent.handle;
-	long column = GTK.gtk_tree_view_get_column (parentHandle, index);
-	long pixbufRenderer = parent.getPixbufRenderer (column);
-	int [] currentWidth = new int [1];
-	int [] currentHeight= new int [1];
-	GTK.gtk_cell_renderer_get_fixed_size (pixbufRenderer, currentWidth, currentHeight);
-	if (!parent.pixbufSizeSet) {
-		if (image != null) {
-			int iWidth, iHeight;
-			if (DPIUtil.useCairoAutoScale()) {
-				iWidth = image.getBounds ().width;
-				iHeight = image.getBounds ().height;
-			} else {
-				iWidth = image.getBoundsInPixels ().width;
-				iHeight = image.getBoundsInPixels ().height;
-			}
-			if (iWidth > currentWidth [0] || iHeight > currentHeight [0]) {
-				GTK.gtk_cell_renderer_set_fixed_size (pixbufRenderer, iWidth, iHeight);
-				parent.pixbufSizeSet = true;
-				parent.pixbufHeight = iHeight;
-				parent.pixbufWidth = iWidth;
-				/*
-				 * Feature in GTK: a Tree with the style SWT.VIRTUAL has
-				 * fixed-height-mode enabled. This will limit the size of
-				 * any cells, including renderers. In order to prevent
-				 * images from disappearing/being cropped, we re-create
-				 * the renderers when the first image is set. Fix for
-				 * bug 480261.
-				 */
-				if ((parent.style & SWT.VIRTUAL) != 0) {
+	try {
+		int modelIndex = parent.columnCount == 0 ? Tree.FIRST_COLUMN : parent.columns [index].modelIndex;
+		long parentHandle = parent.handle;
+		long column = GTK.gtk_tree_view_get_column (parentHandle, index);
+		long pixbufRenderer = parent.getPixbufRenderer (column);
+		int [] currentWidth = new int [1];
+		int [] currentHeight= new int [1];
+		GTK.gtk_cell_renderer_get_fixed_size (pixbufRenderer, currentWidth, currentHeight);
+		if (!parent.pixbufSizeSet) {
+			if (image != null) {
+				int iWidth, iHeight;
+				if (DPIUtil.useCairoAutoScale()) {
+					iWidth = image.getBounds ().width;
+					iHeight = image.getBounds ().height;
+				} else {
+					iWidth = image.getBoundsInPixels ().width;
+					iHeight = image.getBoundsInPixels ().height;
+				}
+				if (iWidth > currentWidth [0] || iHeight > currentHeight [0]) {
+					GTK.gtk_cell_renderer_set_fixed_size (pixbufRenderer, iWidth, iHeight);
+					parent.pixbufSizeSet = true;
+					parent.pixbufHeight = iHeight;
+					parent.pixbufWidth = iWidth;
 					/*
-					 * Only re-create SWT.CHECK renderers if this is the first column.
-					 * Otherwise check-boxes will be rendered in columns they are not
-					 * supposed to be rendered in. See bug 513761.
+					 * Feature in GTK: a Tree with the style SWT.VIRTUAL has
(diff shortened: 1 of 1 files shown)


eselmeister:
+1

basilevs:
Converting to draft until an alternative of renderer reconfiguration  instead of recreation is investigated.

basilevs:
[The alternative](https://github.com/eclipse-platform/eclipse.platform.swt/pull/1743) is quite complex in comparison and does not avoid the asyncExec().

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/TreeItem.java
```diff
@@ -1518,69 +1518,81 @@ public void setImage(int index, Image image) {
...
+						 */
+						boolean check = modelIndex == Tree.FIRST_COLUMN && (parent.style & SWT.CHECK) != 0;
+						// Renderers can't be replaced during render on GTK 3.24.41
+						// https://github.com/eclipse-platform/eclipse.platform.swt/issues/678
+						getDisplay ().asyncExec (() -> {
+							// Do not perform request if it is no longer applicable
+							if (parent.isDisposed () || Math.max (1, parent.getColumnCount ()) <= index) return;
+							// On multiple resize requests, perform only the last one
+							if (parent.pixbufHeight != iHeight || parent.pixbufWidth != iWidth) return;
+							int modelIndexAsync = parent.columnCount == 0 ? Tree.FIRST_COLUMN : parent.columns [index].modelIndex;
```
zkxvh, line 1566: Is it possible that another column gets inserted at this index after this async task is scheduled but before it's actually executed?
If that's possible, then `createRenderers(...)` might not be executed for the original column.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/TreeItem.java
```diff
@@ -1518,69 +1518,81 @@ public void setImage(int index, Image image) {
...
+			 * which results in icons being set to 0. Fix is to compare only positive sizes.
+			 */
+			if (parent.pixbufWidth > Math.max(currentWidth [0], 0) || parent.pixbufHeight > Math.max(currentHeight [0], 0)) {
+				GTK.gtk_cell_renderer_set_fixed_size (pixbufRenderer, parent.pixbufWidth, parent.pixbufHeight);
+			}
 		}
-	} else {
+	
+		GTK.gtk_tree_store_set(parent.modelHandle, handle, modelIndex + Tree.CELL_PIXBUF, pixbuf, -1);
+		GTK.gtk_tree_store_set(parent.modelHandle, handle, modelIndex + Tree.CELL_SURFACE, surface, -1);
```
zkxvh, line 1586: This is interesting.  
So this line was missing before, which means that `TreeItem._getImage(int)` always returned `null`.  
And somehow `Tree` still worked.

zkxvh:
Hi @basilevs.  
Just FYI, I checked this PR with [this snippets](https://github.com/eclipse-platform/eclipse.platform.swt/pull/1743#issuecomment-2604085464) and (at least on my machine) this PR has no negative effects on how images gets rendered in the `Tree`: in all cases except one the images are rendered identically, in one case the code with this PR rendered an image, that was erroneously not rendered by the original code.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/TreeItem.java
basilevs: That's related to the bug with resizing you are trying to fix. `parent.pixbufSizeSet` is set only once per Tree, so the column index is not important 🤷 AFAIK, the only requirement for the column index - the column should exist.

And column may be deleted, so that's a concern worth investigating. Thanks.
The index can't be trusted here and might need to be clamped.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/TreeItem.java
```diff
@@ -1518,69 +1518,81 @@ public void setImage(int index, Image image) {
...
 					 */
-					boolean check = modelIndex == Tree.FIRST_COLUMN && (parent.style & SWT.CHECK) != 0;
-					parent.createRenderers(column, modelIndex, check, parent.style);
+					if ((parent.style & SWT.VIRTUAL) != 0) {
+						/*
+						 * Only re-create SWT.CHECK renderers if this is the first column.
+						 * Otherwise check-boxes will be rendered in columns they are not
+						 * supposed to be rendered in. See bug 513761.
+						 */
+						boolean check = modelIndex == Tree.FIRST_COLUMN && (parent.style & SWT.CHECK) != 0;
```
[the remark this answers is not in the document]
basilevs: Yes, I've failed to account for column deletion. Thanks.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/TreeItem.java
basilevs: It was not missing. I've moved it from below.

basilevs:
> I checked this PR with [this snippets](https://github.com/eclipse-platform/eclipse.platform.swt/pull/1743#issuecomment-2604085464)

basilevs:
> Should this one be looked at or #1743 is considered superior to it?

@akurtakov this PR contains a bug, and was marked as Draft until I have the time to fix it. https://github.com/eclipse-platform/eclipse.platform.swt/pull/1743 is indeed somewhat superior, but more invasive too.

akurtakov:
Alternative fix via https://github.com/eclipse-platform/eclipse.platform.swt/pull/3262

PR #1615, opened 2024-11-24
 Stop using SWTEventListener internally and mark it as deprecated  forRemoval

akurtakov:
As it:
* is internal
* makes api tools fail thanks to it
* serves no purpose since SWT no longer runs on Java ME

Internally SWT is moved to use java.util.EventListener and as it's the
superclass of SWTEventListener makes all the APIs that taken
SWTEventListener to keep working.


laeubi:
There already was an attempt here:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/1101

but there was some concerns raised.

akurtakov:
Thanks for the pointers @laeubi . The new versions keeps SWTEventListener (marked as deprecated for removal) and should be compatible for all the cases except:
* SomeSWTListener instanceof SWTEventListener
* TypedListener.getEventListener returning EventListener and not SWTEventListener

@Phillipus can you check whether this file works for you?

Phillipus:
I checked out the PR and get some errors in the Widget class:

```
protected void removeListener (int eventType, SWTEventListener listener) {
	removeTypedListener(eventType, listener);
}
``

**Comment by akurtakov** [comment]
> I checked out the PR and get some errors in the Widget class:
> 
> ```
> protected void removeListener (int eventType, SWTEventListener listener) {
> 	removeTypedListener(eventType, listener);
> }
> ```

I did the change for Gtk only, new version should have the changes for mac/win too.

File: bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Widget.java [hunk 1]
```diff
@@ -1039,7 +1039,7 @@ public void removeListener (int eventType, Listener listener) {
  * @noreference This method is not intended to be referenced by clients.
  * @nooverride This method is not intended to be re-implemented or extended by clients.
  */
-protected void removeListener (int eventType, SWTEventListener listener) {
+protected void removeListener (int eventType, EventListener listener) {
```
laeubi, line 1042: As it is a protected method of a public class I think this must stay and be deprecated.

Then we need a new `protected void removeTypedListener (int eventType, EventListener listener) {`

Phillipus:
I'm using Nebula Grid Widget and get:

```
java.lang.NoClassDefFoundError: org/eclipse/swt/internal/SWTEventListener
	at org.eclipse.nebula.jface.gridviewer.GridTableViewer.<init>(GridTableViewer.java:138)
```

Does Nebula need a change?

akurtakov:
> I'm using Nebula Grid and Gallery Widgets and get:
> 
> ```
> java.lang.NoClassDefFoundError: org/eclipse/swt/internal/SWTEventListener
> 	at org.eclipse.nebula.jface.gridviewer.GridTableViewer.<init>(GridTableViewer.java:138)
> ```
> 
> Does Nebula need a change? I'm using 3.1.1

The goal is to not need a change (for majority of cases). Please try with latest version.

Phillipus:
> The goal is to not need a change (for majority of cases). Please try with latest version.

Now I get:

```
java.lang.NoSuchMethodError: 'void org.eclipse.swt.widgets.TypedListener.<init>(org.eclipse.swt.internal.SWTEventListener)'
	at org.eclipse.nebula.widgets.grid.Grid.addTreeListener(Grid.java:995)
```

This is:

```java
addListener(SWT.Expand, new TypedListener(listener));
```

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/SWTEventListener.java
```diff
@@ -29,5 +29,6 @@
  * </p>
  * @noreference This interface is not intended to be referenced by clients.
  */
+@Deprecated(forRemoval=true, since="2025-03")
```
HannesWell, line 32: I think the since tag should represent the bundle version since which this is deprecated, which will probably be `3.129`.
Furthermore, since it's the hard to know for ordinary users when the API will actually be removed, I experimeted with a corresponding extension in PDE and I think that's helpful.
https://github.com/eclipse-pde/eclipse.pde/blob/026b764ff7b262ffb0ed5a3732ddd66ff4a89673/ui/org.eclipse.pde.core/src/org/eclipse/pde/core/plugin/PluginRegistry.java#L67

```suggestion
@Deprecated(forRemoval = true, since = "3.129 (removal in 2027-03 or later)")
```

HannesWell (commented):
As Christoph mentioned I tried the same in https://github.com/eclipse-platform/eclipse.platform.swt/pull/1101 and failed.
While I'm note happy about the current situation and would like to get rid of `SWTEventListener` too I think we have to be much more careful here because although it should not have been used, `SWTEventListener` is probably used in more places than we would like it happen. It seems it has become a de-factor API and we should treat it like this to not potentially break everything.


> - SomeSWTListener instanceof SWTEventListener

Furthermore assigning SWT-events of a specific type to a variable of type `SWTEventListener` does not work anymore too.
But in that case either a specific event type or the more general `EventListener` can be used immediately as an replacement.
And the same solution can be used for existing `instanceof` checks.
But of course there are no runtime errors like in the first case but the case will just always be false at runtime.


> - TypedListener.getEventListener returning EventListener and not SWTEventListener

That could be a bigger problem, because although it again should not have been used, it was necessary to use it because a lack of API added only two/three releases ago and this change is binary incompatible.

I have to check the details again.

> Does Nebula need a change? I'm using 3.1.1

I did a lot of migration in the past in the context of https://github.com/eclipse-platform/eclipse.platform.swt/pull/1101#issuecomment-1992443452 to use new APIs:
https://github.com/eclipse/nebula/pulls?q=is%3Apr+author%3AHannesWell+is%3Aclosed

But IIRC since the PRs were open for some time new code using the old APIs have been added, so another pass might be necessary. I'll check that.

HannesWell:
> > The goal is to not need a change (for majority of cases). Please try with latest version.
> 
> Now I get:
> 
> ```
> java.lang.NoSuchMethodError: 'void org.eclipse.swt.widgets.TypedListener.<init>(org.eclipse.swt.internal.SWTEventListener)'
> 	at org.eclipse.nebula.widgets.grid.Grid.addTreeListener(Grid.java:995)
> ```
> 

You are not using the latest version of Nebula are you?
Because this was changed a few months ago in
- https://github.com/eclipse/nebula/pull/583/files#diff-c3b4ffebdc875a02b8d0c4103d43376486a3dc50672e536164834cc6873ed2bd

But I don't know how often Nebula releases new versions.

Btw. with https://github.com/eclipse/nebula/pull/613, all references to SWT's `TypedListener` and `SWTEventListener` should be removed from Nebula. Certain use cases of `TypedListener` also required `SWTEventListener` as your error shows.
There is only one utility method left, which is deprecated and not necessary and used (in Nebula itself) anymore.

Phillipus:
> You are not using the latest version of Nebula are you?
Because this was changed a few months ago in

If "latest" means nightly build then no, because:

> The goal is to not need a change (for majority of cases).

It's OK with the nightly build of Nebula.

HannesWell:
> But I don't know how often Nebula releases new versions.

Looking at https://github.com/eclipse/nebula/releases, [V3.1.1](https://github.com/eclipse/nebula/releases/tag/V3.1.1) seems to be the latest release and it's from February and therefore does not include all the mentioned changes.
But https://projects.eclipse.org/projects/technology.nebula says there is a 3.2.0 release done after the mentioned changes were done.
But at https://download.eclipse.org/nebula/updates/ I can also only see 3.1.1.

@Phillipus can you try the latest master?

HannesWell:
> It's OK with the nightly build of Nebula.

Thanks for already trying. Then we should ask for a new Nebula release after https://github.com/eclipse/nebula/pull/613 has landed.

akurtakov:
I've added back the TypedListener constructor using SWTEventListener but marked it as deprecated forRemoval. That should fix this issue too. 
@Phillipus Huge thanks for baring with us to help getting some sense back into SWT . If we can't manage to get our tools (apitools) work with it we can just give up on getting new contributors right now.

HannesWell:
@akurtakov have you seen my comments in
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/1615#discussion_r1855456939
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/1615#discussion_r1855459834

bundles/org.eclipse.swt/Eclipse SWT Drag and Drop/common/org/eclipse/swt/dnd/DNDListener.java
```diff
@@ -13,17 +13,18 @@
...
+
 import org.eclipse.swt.widgets.*;
 
 
 class DNDListener extends TypedListener {
 	Widget dndWidget;
 /**
  * DNDListener constructor comment.
- * @param listener org.eclipse.swt.internal.SWTEventListener
+ * @param listener java.util.SWTEventListener
```
HannesWell, line 25:
```suggestion
 * @param listener java.util.EventListener
```

Or even better just remove the type from the java-doc or even the entire `@param` line as it has no value.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/internal/SWTEventListener.java
```diff
@@ -0,0 +1,34 @@
...
+ * implement SWTEventListener instead of java.util.EventListener.
+ * </p>
+ * <p>
+ * Note: java.util.EventListener is not part of CDC and CLDC.
+ * </p>
+ * @noreference This interface is not intended to be referenced by clients.
+ */
+@Deprecated(forRemoval=true, since="2025-03")
+public interface SWTEventListener extends EventListener {
+}
```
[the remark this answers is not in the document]
akurtakov: Not at all and is now removed.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/SWTEventListener.java
akurtakov: This has been discussed some time ago and decided to go with "human readable calendar form" (see https://github.com/eclipse-platform/eclipse.platform/blob/master/docs/Eclipse_API_Central_Deprecation_Policy.md#java-api ) as the bundle version is not easily translatable to release version and for some bundles it would even happen that same version will be shipped in multiple versions of the SDK thus making the 2 years timeframe calculation vague.

bundles/org.eclipse.swt/Eclipse SWT Drag and Drop/common/org/eclipse/swt/dnd/DNDListener.java
akurtakov: param removed.

akurtakov:
@Phillipus if we manage to get to a state where using your product as an experiment we can land this PR without any breakage in it that would be huge step.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Widget.java
akurtakov: Done, the method getting SWTEventListener is marked as deprecated forRemoval.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/SWTEventListener.java
akurtakov: If you have good idea how to improve it please bring your proposal to the next PMC call for a change. Till we get a change in guidelines I would prefer that we stick to them.

Phillipus:
Latest push to PR is working with existing Nebula Grid and Gallery widgets.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/SWTEventListener.java
HannesWell: > as the bundle version is not easily translatable to release version and for some bundles it would even happen that same version will be shipped in multiple versions of the SDK thus making the 2 years timeframe calculation vague.

Understand. The goal is the same and additionally it should be make simple for those that don't know our 2 year policy.

> If you have good idea how to improve it please bring your proposal to the next PMC call for a change.

Sure I'll bring this up in December. 👍🏽

akurtakov:
I plan to merge this one in once new stream opens if there are no identified issues.

Phillipus:
> I plan to merge this one in once new stream opens if there are no identified issues.

I'll test again when that happens. But I only use two of the Nebula widgets, Grid and Gallery, Wouldn't it make more sense for the Nebula committers to test all their widgets against these changes in SWT using automated tests?

akurtakov:
Ideally every downstream project (not only Nebula) should test with latest I-builds constantly so issues are reported and identified ASAP. I am very grateful to everyone that tests and reports timely and of course this works in the other direction too - early testers don't face (or at least less often) face regressions.

merks:
FYI, it looks like datechooser has not gotten any attention:

image

Builds are definitely not being built against the latest, though  making it configurable for that wouldn't be so much work...

akurtakov:
> FYI, it looks like datechooser has not gotten any attention:
> 
> image
> 
> Builds are definitely not being built against the latest, though making it configurable for that wouldn't be so much work...

I can not reproduce locally. I have this project (+deps ) open in my swt workspace and they compile just fine. What am I missing?

merks:
Probably/maybe related to the rwt thing:

image

I don't have time to investigate further now...

merks:
@HannesWell 

I've never seen this `bundle-version` thing before.  It looks like this would maybe make the plugin not work with the latest rwt-providing bundle:

image

I think RWT has not been adapted, so the code doesn't work there, and that there is confusion here when RWT is in the target platform...

akurtakov:
So what I understand is that Nebula fails when RWT (RAP?) is used instead of SWT aka not not related to SWT itself.

HannesWell:
> I've never seen this `bundle-version` thing before. It looks like this would maybe make the plugin not work with the latest rwt-providing bundle:

It's indeed not used often but according to the spec it legal:
https://docs.osgi.org/specification/osgi.core/8.0.0/framework.module.html#framework.module.importpackage
```
bundle-version - A version-range to select the bundle version of the exporting bundle.
The default value is [0.0.0, ∞). See Semantic Versioning.
In the case of a fragment bundle, the version is from the host bundle.
```

> I think RWT has not been adapted, so the code doesn't work there, and that there is confusion here when RWT is in the target platform...

Looks like they have copied the class...
https://github.com/eclipse-rap/org.eclipse.rap/blob/924a3e0bf8f0669308eba601b9ed08cf95b53b60/bundles/org.eclipse.rap.rwt/src/org/eclipse/swt/widgets/TypedListener.java#L66

Don't know why they do it but it's for sure their responsibility to update it.

wimjongman:
> > But I don't know how often Nebula releases new versions.
> 

I'm prepping a new release. Let me know if you need something else.

akurtakov:
> > > But I don't know how often Nebula releases new versions.
> 
> I'm prepping a new release. Let me know if you need something else.

Would you please run your full test suite against SWT patched with this PR and report back ?

merks:
@akurtakov 

Does the build produce an update site?

akurtakov:
> @akurtakov
> 
> Does the build produce an update site?

Nope. At some point I hope that every git repo will provide update site for usage but I start from the leaves and this one is kind of 'root'. In order for people to try it they have to build it themself .
P.S. It feels right to ask you about "update sites" - https://github.com/eclipse-pde/eclipse.pde/issues/1136 - would you please look into this one?

HannesWell:
> Does the build produce an update site?

If the build would create a P2 repo just for SWT the result could be archived by the Jenkins build. The URL to the corresponding build artifact can then be used just like any other URL for a p2-repo, e.g. in a TP or to install new software.
But it would increase the disk-storage demand on Jenkins and I don't know if the rare occasions where this is needed are worth it.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/widgets/TypedListener.java
```diff
@@ -63,11 +63,26 @@ public class TypedListener implements Listener {
  *
  * @noreference This method is not intended to be referenced by clients.
  */
+@Deprecated(forRemoval=true, since="2025-03")
```
HannesWell, line 66: With https://github.com/eclipse-platform/eclipse.platform.swt/pull/1616, this entire class is now deprecated, so we don't have to deprecate this single constructor separately anymore.

akurtakov:
> > > I plan to merge this one in once new stream opens if there are no identified issues.
> > 
> > 
> > I'll test again when that happens.
> 
> This merge is working with the Nebula Grid and Gallery widgets (3.1.1). Thank-you!

@Phillipus It wouldn't have been the case without you!

PR #1616, opened 2024-11-24
Deprecate TypedListener for removal to make it strongly internal later

HannesWell:
The class `TypedListener` is intended to be internal even if it's in a public package and because there were cases where it has to be used for custom widges it happened that it was still used. With the APIs introduced in https://github.com/eclipse-platform/eclipse.platform.swt/pull/1112 there are better solutions using only public APIs for the cases it was still used.

Together with https://github.com/eclipse-platform/eclipse.platform.swt/pull/1615 this is the continuation of https://github.com/eclipse-platform/eclipse.platform.swt/pull/1101#issuecomment-1995483340 to deprecate `TypedListener` for removal so that it can later be moved into an internal package to strengthen the statement that it's internal.

Eclipse Nebula also does not use 

- https://github.com/eclipse/nebula/pull/582
- https://github.com/eclipse/nebula/pull/583
- https://github.com/eclipse/nebula/pull/602
- https://github.com/eclipse/nebula/pull/613

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/DefaultContent.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/DefaultContent.java
@@ -749,6 +749,7 @@ public String getTextRange(int start, int length) {
  * </ul>
  */
 @Override
+@SuppressWarnings("removal")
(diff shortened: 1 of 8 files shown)


HannesWell:
It looks like there is no reference to this class in the Eclipse-SDK (anymore).

So if there are no objections about continuing with this I plan to submit this tomorrow evening.

PR #1617, opened 2024-11-24
don't modify topIndex in handleTextChanged if control has not the focus

tobiasmelcher:
fix split screen issue https://github.com/eclipse-platform/eclipse.platform.ui/issues/2142 by not setting StyledText#topIndex if the control has not the focus.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/StyledText.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/StyledText.java
@@ -6046,7 +6046,7 @@ void handleTextChanged(TextChangedEvent event) {
 	}
 	int firstLine = content.getLineAtOffset(lastTextChangeStart);
 	resetCache(firstLine, 0);
-	if (!isFixedLineHeight() && topIndex > firstLine) {
+	if (!isFixedLineHeight() && isFocusControl() && topIndex > firstLine) {
 		topIndex = firstLine;
 		if (topIndex < 0) {
 			// TODO: This logging is in place to determine why topIndex is getting set to negative values.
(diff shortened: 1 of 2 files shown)


tobiasmelcher:
@mickaelistria would you prefer that we delete the complete if-block "if (!isFixedLineHeight() && topIndex > firstLine) {" or is the approach of this pull request also fine?

vogella:
We should try to delete the code that we do not know why it is there (https://github.com/eclipse-platform/eclipse.platform.ui/issues/2142) early in the next release cycle to find if this causes any yet unknown issues.

mickaelistria:
This PR looks good.

> We should try to delete the code that we do not know why it is there

Here I think the reason for this code are clear enough: it can happen while editing some lines of non fixed height that the current position of the topIndex doesn't allow to see the whole first edited line. So I think it's best to keep it as proposed.

tobiasmelcher:
test error 
org.eclipse.swt.tests.junit.Test_org_eclipse_swt_custom_StyledText.test_notFixedLineHeightDoesntChangeLinePixelIfUnnecessary
might be related to this change? I need to check.

mickaelistria:
> might be related to this change? 

That's extremely probable. Note that this test is important, it may not be written properly so maybe tweaking it so it is doing the job better can help, but in any case, the behavior it does test (notFixedLineHeightDoesntChangeLinePixelIfUnnecessary is hopefully explicit enough) is important to be guaranteed by a test. If this test is failing then it means that whenever a line header annotation is printed, the content of the text widget moves accordingly and may even hide (whose position is supposed to be stable for comfortable usage)

BeckerWdf:
mh...
org.eclipse.swt.tests.junit.Test_org_eclipse_swt_custom_StyledText.test_notFixedLineHeightDoesntChangeLinePixelIfUnnecessary
still fails.

mickaelistria:
Can you reproduce the failure locally?

PR #1620, opened 2024-11-26
PatternWin32ManualTest: suppress Discouraged access warning

jukzi:
`The type 'DPIUtil' is not API (restriction on required project 'org.eclipse.swt.win32.win32.x86_64')`

### DIFF
--- a/tests/org.eclipse.swt.tests.win32/ManualTests/org/eclipse/swt/widgets/PatternWin32ManualTest.java
+++ b/tests/org.eclipse.swt.tests.win32/ManualTests/org/eclipse/swt/widgets/PatternWin32ManualTest.java
@@ -33,6 +33,7 @@ public class PatternWin32ManualTest {
 	private static Display display = Display.getDefault();
 
 	public static void main (String [] args) {
+		@SuppressWarnings("restriction")
 		int zoom = DPIUtil.getDeviceZoom();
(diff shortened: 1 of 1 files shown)


jukzi:
CI complains about 10 unnecessary `@SuppressWarnings("restriction")` - what has changed?
image

jukzi:
> CI complains about 10 unnecessary

created https://github.com/eclipse-platform/eclipse.platform.swt/issues/1648

akurtakov:
That's because the plugin is not enabled in maven build. If you look at https://github.com/eclipse-platform/eclipse.platform.swt/blob/3c1065b067e01526a303134b0f1b947af92ad229/pom.xml#L233 there is only one test bundle enabled.

jukzi:
Is it possible and good to add `org.eclipse.swt.tests.win32` there?

akurtakov:
From releng POV it would probably have to be added in a maven profile as IIRC it uses SWT methods that are internal for the win32 port. 
With the above said it's up to people with windows access to evaluate whether it would be beneficial  for GHA as Jenkins runs on Linux thus the module will be skipped there.

jukzi:
lets keep it simple by now: i know the warning is reduced in my workspace - lets merge?

akurtakov:
Merge it.

PR #1623, opened 2024-11-27
Updated bundles with comparator errors

trancexpress:
Changes to ecj are for handling of switch on String.

See: https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2595

### DIFF
--- a/binaries/org.eclipse.swt.gtk.linux.aarch64/forceQualifierUpdate.txt
+++ b/binaries/org.eclipse.swt.gtk.linux.aarch64/forceQualifierUpdate.txt
@@ -1,0 +1,1 @@
+https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2595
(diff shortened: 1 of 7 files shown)


akurtakov:
I am not sure that would do the trick. Please force update for org.eclipse.swt host too.

bundles/org.eclipse.swt/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-Name: %pluginName
 Bundle-Vendor: %providerName
 Bundle-SymbolicName: org.eclipse.swt; singleton:=true
-Bundle-Version: 3.129.0.qualifier
+Bundle-Version: 3.129.100.qualifier
```
akurtakov, line 5: SWT minor version has already been update so no version change needed

trancexpress:
3 tests in `org.eclipse.swt.tests.junit.Test_org_eclipse_swt_browser_Browser` fail, I assume that is not related to the changes here.

PR #1624, opened 2024-11-27
Use ubuntu-latest instead of outdated centos-8

laeubi:


Jenkinsfile
```diff
@@ -18,7 +18,7 @@
 def runOnNativeBuildAgent(String platform, Closure body) {
 	def final nativeBuildStageName = 'Build SWT-native binaries'
 	if (platform == 'gtk.linux.x86_64') {
-		podTemplate(inheritFrom: 'centos-latest' /* inhert general configuration */, containers: [
+		podTemplate(inheritFrom: 'ubuntu-latest' /* inhert general configuration */, containers: [
 			containerTemplate(name: 'swtbuild', image: 'eclipse/platformreleng-centos-swt-build:8',
```
akurtakov, line 22: What will the effect be of using eclipse/platformreleng-centos-swt-build:8 as containerTemplate?

laeubi: I'm not sure what the question is, according to the doc we only inherit the general config and then run inside container (what is `eclipse/platformreleng-centos-swt-build:8`) maybe @HannesWell  can tell more?

laeubi: Yes, the goal here is just to update the runtime to be prepared for new glibc

akurtakov: OK, if it works, let's merge it and retry gtk pr.

laeubi:
> OK, if it works, let's merge it and retry gtk pr.

I assume test failures are expected?!?

akurtakov:
> > OK, if it works, let's merge it and retry gtk pr.
> 
> I assume test failures are expected?!?

If we speak about the 3 long failing Browser tests - yes.

laeubi:
Yep

grafik

Jenkinsfile
HannesWell: > according to the doc we only inherit the general config and then run inside container (what is `eclipse/platformreleng-centos-swt-build:8`) maybe @HannesWell can tell more?

Exactly. Probably even the basic image is sufficient.
Let's find this out:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/1627

PR #1629, opened 2024-11-28
[GTK4] Migrate deprecated ColorChooser to ColorDialog

ptziegler:
The Color Chooser has been moved to the GTK3 class and all references inside the GTK4 blocks have been updated to use the new API.

Note that the GTK4 API doesn't provide the means to set a custom palette, meaning that the colors specified with setRGBS(RGB[]) are effectively ignored.


ptziegler:
GTK3:

ColorDialog_gtk3

GTK4:

ColorDialog_gtk4

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/ColorDialog.java
```diff
@@ -141,34 +158,46 @@ public RGB open () {
...
-				rgba.blue = (double) rgbS.blue / 255;
-				OS.memmove (gdkRGBAS + i * GdkRGBA.sizeof, rgba, GdkRGBA.sizeof);
-			}
+
+	if (GTK.GTK4) {
+		// TODO The ColorDialog API doesn't support setting a custom palette
+		// See https://gitlab.gnome.org/GNOME/gtk/-/issues/5936
+		// > The new GtkColorDialog api doesn't have this feature atm. We might
+		// > want to bring it back, but judging from application use of color
+		// > choosers, setting custom palettes is not very popular.
```
ptziegler, line 167: This is probably really bad. But as mentioned in the linked issue, as well as the [documentation](https://docs.gtk.org/gtk4/class.ColorDialog.html), there is simply no way to customize the palette.

Meaning this is probably something that would first have to be implemented, but I don't think my GTK knowledge is sufficient to do so. :/

akurtakov: I think we should merge this one as there is not much to be done from SWT side.

akurtakov:
Would you please rebase on master and remove the draft if you agree that it's time to get this one in.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/ColorDialog.java
```diff
@@ -131,7 +131,24 @@ public RGB[] getRGBs() {
...
+	if (handle == 0) error (SWT.ERROR_NO_HANDLES);
+	if (GTK.GTK4) {
+		// TODO The color dialog isn't opened a non-modal window. However, the
+		// "same" API is used for e.g. the Font Dialog, where everything works.
+		// So this looks like a bug within GTK...
+		// Error:
+		// g_object_new_valist: invalid object type 'SwtFixed' for value type 'GtkWindow'
+		// GtkDialog mapped without a transient parent. This is discouraged.
+		GTK4.gtk_color_dialog_set_modal(handle, true);
+		GTK4.gtk_color_dialog_set_title(handle, buffer);
```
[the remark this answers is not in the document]
ptziegler: Fixed. I was simply using the wrong parent handle.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/ColorDialog.java
ptziegler: Sounds good to me. I fear that contributing the palette customization directly to GTK is beyond my capabilities, so the current state has to suffice for now.

ptziegler:
> Would you please rebase on master and remove the draft if you agree that it's time to get this one in.

I only wanted to revisit the issue with the modal dialog but that has now been fixed. I don't see any (additional) warnings when trying out the snippets so I think this is done for now.

akurtakov:
There is one fundamental problem - the gtk_color_dialog calls are guarded with if( GTK4) blocks but they are introduced in gtk 4.10 so these symbols will simply fail to load (hard crash) if swt is run on gtk 4.0-4.8. 
I believe the ColorChooser API should stay in GTK class and be used till gtk 4.8. The ColorDialog API should be guarded using `if (GTK.GTK_VERSION >= OS.VERSION(4, 10, 0)) blocks`.

akurtakov:
Now I realize I missed the same problem in https://github.com/eclipse-platform/eclipse.platform.swt/pull/1583 so we will have to redo it too.

ptziegler:
> Now I realize I missed the same problem in #1583 so we will have to redo it too.

I was just about to ask about the Font/Dialog chooser. I can have a look at it over the weekend.

ptziegler:
I've also updated this PR to guard the ColorDialog API behind a "GTK 4.10" check. This should avoid the crash with older GTK4 versions.

akurtakov:
This one is good to go.

PR #1632, opened 2024-12-02
Feature/new dirty indicator bullet

schneidermic0:
Before this change: If a part is dirty (has unsaved changes), this will be indicated by an `*` in front of a tab's name. 

With this change, the `*` is removed and dirty parts are indicated by a bullet at the location of closing a tab. As soon as, the bullet is hovered, the "button" to close the tab is shown.

Furthermore, the rendering of the close button of tabs is lightly changed.

**Examples**

Tabs unchanged:
image

Tabs unchanged (hovered):
image

Tab "Test.java" (dirty; selected):
image

Tab "Test.java" (dirty; not selected):
image

Tab "Test.java" (dirty; hovered)
image

Complete flow:

To enable this behavior, changes in PR <tbd> are required, too.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
@@ -877,44 +877,59 @@ public class CTabFolderRenderer {
 		}
 	}
 
-	void drawClose(GC gc, Rectangle closeRect, int closeImageState) {
+	void drawClose(GC gc, Rectangle closeRect, int closeImageState, boolean showDirtyIndicator) {
 		if (closeRect.width == 0 || closeRect.height == 0) return;
 
-		// draw X with length of this constant
-		final int lineLength = 8;
-		int x = closeRect.x + Math.max(1, (closeRect.width-lineLength)/2);
-		int y = closeRect.y + Math.max(1, (closeRect.height-lineLength)/2);
-		y += parent.onBottom ? -1 : 1;
 		int originalLineWidth = gc.getLineWidth();
 		Color originalForeground = gc.getForeground();
-		switch (closeImageState & (SWT.HOT | SWT.SELECTED | SWT.BACKGROUND)) {
-			case SWT.NONE: {
-				drawCloseLines(gc, x, y , lineLength, false);
-				break;
+	    int state = closeImageState & (SWT.HOT | SWT.SELECTED | SWT.BACKGROUND);
+	    if (state == SWT.NONE) {
+	        if (showDirtyIndicator) drawDirtyIndicator(gc, closeRect, originalForeground, false);
+	        drawCloseButton(gc, closeRect, false);
+	    } else if (state == SWT.HOT || state == SWT.SELECTED) {
+			if (showDirtyIndicator) {
+				drawDirtyIndicator(gc, closeRect, originalForeground, true);
+				drawCloseButton(gc, closeRect, false);
+			} else {
+				drawCloseButton(gc, closeRect, true);
 			}
-			case SWT.HOT: {
-				drawCloseLines(gc, x, y , lineLength, true);
-				break;
-			}
-			case SWT.SELECTED: {
-				drawCloseLines(gc, x, y , lineLength, true);
-				break;
-			}
-			case SWT.BACKGROUND: {
-				int[] shape = new int[] {x,y, x+10,y, x+10,y+10, x,y+10};
-				drawBackground(gc, shape, false);
-				break;
-			}
-		}
-		gc.setLineWidth(originalLineWidth);
+	    } else if (state == SWT.BACKGROUND) {
+	        if (showDirtyIndicator)
+	        	drawDirtyIndicator(gc, closeRect, originalForeground, false);
+	        else
+		    	drawBackground(gc, closeRect, SWT.BACKGROUND);
+
+	    }
+	    gc.setLineWidth(originalLineWidth);
 		gc.setForeground(originalForeground);
 	}
 
-	private void drawCloseLines(GC gc, int x, int y, int lineLength, boolean hot) {
+	private void drawDirtyIndicator(GC gc, Rectangle closeRect, Color originalForeground, boolean hot) {
+		Color indicatorColor = hot ? getFillColor() : originalForeground;
+		drawCloseBackground(gc, closeRect, indicatorColor);
+	}
+
+	private void drawCloseBackground(GC gc, Rectangle closeRect, Color backgroundColor) {
+		Color originalBackground = gc.getBackground();
+		gc.setBackground(backgroundColor);
+		gc.setForeground(originalBackground);
+		gc.fillRoundRectangle(closeRect.x + 3, closeRect.y + 4, closeRect.width - 5, closeRect.height - 5, 4, 4);
+		gc.setBackground(originalBackground);
+	}
+
+
+	private void drawCloseButton(GC gc, Rectangle closeRect, boolean hot) {
 		if (hot) {
-			gc.setLineWidth(gc.getLineWidth() + 2);
-			gc.setForeground(getFillColor());
+			drawCloseBackground(gc, closeRect, parent.getDisplay().getSystemColor(SWT.COLOR_WIDGET_NORMAL_SHADOW));
+//			gc.setLineWidth(gc.getLineWidth() + 2);
+			gc.setForeground(gc.getBackground());
 		}
+		// draw X with length of this constant
+		final int lineLength = 6;
+		int x = closeRect.x + Math.max(1, (closeRect.width-lineLength)/2);
+		int y = closeRect.y + Math.max(1, (closeRect.height-lineLength)/2);
+		y += parent.onBottom ? -1 : 1;
+
 		gc.setLineCap(SWT.CAP_ROUND);
 		gc.drawLine(x, y, x + lineLength, y + lineLength);
 		gc.drawLine(x, y + lineLength, x + lineLength, y);
@@ -1463,7 +1478,7 @@ public class CTabFolderRenderer {
 					gc.setBackground(orginalBackground);
 				}
 			}
-			if (shouldDrawCloseIcon(item)) drawClose(gc, item.closeRect, item.closeImageState);
+			if (shouldDrawCloseIcon(item)) drawClose(gc, item.closeRect, item.closeImageState, item.showDirty);
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


schneidermic0:
Hi, 

I have opened this PR as draft, because I want to get your opinion on this change. The code can be improved at several places and current implementation has some shortcomings that need to be improved as soon as this approach shall be taken up.

Existing gaps (as of now):

- [ ] Dirty indicator is not moved to not, yet opened tab (i.e, if there are two tabs of the same file)
- [ ] In the tab overview and editor selection (Ctrl+E), dirty editors are still marked with `*`

iloveeclipse:
> Other editors like VScode use the same approach

Could you attach some screenshots from "other" editors? I haven't seen yet something like proposed here.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
```diff
@@ -877,44 +877,58 @@ void drawBody(GC gc, Rectangle bounds, int state) {
 		}
 	}
 
-	void drawClose(GC gc, Rectangle closeRect, int closeImageState) {
+	void drawClose(GC gc, Rectangle closeRect, int closeImageState, boolean showDirtyIndicator) {
```
iloveeclipse, line 880: I wonder if this code, and, in general, tab presentation (font style, color, text decoration, borders, close icon) could be delegated to a new callback interface that, if the callback is not set, would by default use current SWT implementation, and if set, would give the callback full control of the tab visuals.
This way SWT could keep its code independent of the application needs and application (like workbench renderers) could implement whatever possible.

I personally would like to see dirty tabs in bold red font and I don't want to see close buttons at all to save space. I've had that in my Extended VS Presentation plugin for Eclipse 3.x many years ago before e4 killed Presentation API.
That's of course much more as proposal here, but it shouldbe doable and it would allow 3rd parties to decide whether they want close buttons  turned into circles or disappear completely or just stay "old good" black crosses etc.

iloveeclipse: See https://web.archive.org/web/20170702182707/http://andrei.gmxhome.de/skins/examples.html

BeckerWdf:
> Hi,
> 
> I have opened this PR as draft, because I want to get your opinion on this change. The code can be improved at several places and current implementation has some shortcomings that need to be improved as soon as this approach shall be taken up.
> 
> Existing gaps (as of now):
> 
> * [ ]  Dirty indicator is not moved to not, yet opened tab (i.e, if there are two tabs of the same file)
> * [ ]  In the tab overview and editor selection (Ctrl+E), dirty editors are still marked with `*`

What about tabs of multi-page editors as e.g. the plugin.xml editor?

vogella:
> > Other editors like VScode use the same approach
> 
> Could you attach some screenshots from "other" editors? I haven't seen yet something like proposed here.

image

schneidermic0:
> Could you attach some screenshots from "other" editors? I haven't seen yet something like proposed here.

Here is a screenshot I took in VS Code:
image

laeubi:
> Looks awesome to me, thanks. The * is outside Eclipse not well understood as "I need saving".

I think this is maybe a biased opinion, at least on Debian / Gnome this is common and "well understood" (at least by me)

grafik

So as always there is not the one-and-only truth ;-)

> I personally would like to see dirty tabs in bold red font and I don't want to see close buttons at all to save space. 

Can't one change this via CSS styling.

iloveeclipse:
> Can't one change this via CSS styling.

Yes, but I don't also like native widgets to be additionally styled by e4 css for multiple reasons.

iloveeclipse:
> `CTabFolder` is not a native widget ;-)

But the rest of the UI is, and you can't have only tabs styled as of today.

BeckerWdf:
what I like with the proposed change is that the black circle is much bigger then the little tiny "*" we use today.
And also the width of the tab is not changing when the editor is dirty so it does not "flicker".

schneidermic0:
Thank you for the feedback so far. I've noticed some positive responses, but I’m uncertain about the rest. Should I proceed with improving this PR, or do we need more input from the community first? 

I can work on it when I find some time alongside my other tasks.

vogella:
@HannesWell and @akurtakov please bring this to the PMC and decide if this is desired

HannesWell:
> @HannesWell and @akurtakov please bring this to the PMC and decide if this is desired

We have discussed this topic in the open dev-call on Thursday:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/wiki/Eclipse-IDE-%E2%80%90-Developers-Community-Call#06th-december-2024

In general the feedback was that avoiding resizing is nice, but the exact 'dirty' symbol might not be liked by everyone and maybe alternatives could be considered. Furthermore the old and new behavior would ideally be configurable in a preference.

mickaelistria:
> what I like with the proposed change is that the black circle is much bigger then the little tiny "*" we use today.

For that particular part, just using another character such as `●` or even `✏️` char instead of `*` could do the job; or maybe even keeping the `*` but making it bold. I personally like emojis as they are very expressive and are undertood more easily; the drawback is that there is a trend of color-less and fade icons while emojis tend to be the opposite.

> And also the width of the tab is not changing when the editor is dirty so it does not "flicker".

I also find it interesting. However it is un

Another possibility that would fit both could be an overlay on the editor icon of the tab, but it might be more complex to implement.

schneidermic0:
@HannesWell Thanks for the update. Next step: I will add a preference so that users can enable the new behavior.

schneidermic0:
> For that particular part, just using another character such as ● or even ✏️ char instead of * could do the job; or maybe even keeping the * but making it bold. I personally like emojis as they are very expressive and are undertood more easily; the drawback is that there is a trend of color-less and fade icons while emojis tend to be the opposite.

Actually, using an emoji was my first attempt. Unfortunately this changed the size of the tab much more than the `*`. Therefore, I tried it with the change I proposed with this PR.

HeikoKlare:
I support the attempt of improving the indicator for unsaved documents. It is easy to miss the current indicator. So for me, it would be more about the goal of visibility than about tab header resizing, but it makes sense to combine the two of them, so any replacement should be one that does not change sizes.

I have no strong opinion on the actual solution to take. I would expect best visibility in a recolored icon (like the ones in Notepad++), but it will probably be hard to find a generic recoloring algorithm for any possible icon. Did you consider using some highlight indicator (like for the currently selected tab) to indicate unsaved tabs? Something like this:
image

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabItem.java
```diff
@@ -276,6 +277,20 @@ public boolean getShowClose() {
...
+ *
+ * @return <code>true</code> if the dirty indicatorn should be shown
+ *
+ * @since 3.129
+ */
+public boolean getShowDirty() {
+	checkWidget();
+	return showClose;
+}
+
```
fedejeanne, line 293: @schneidermic0 I know that you said that there is still a lot to improve but I think one of these may be a copy-paste bug so I'll bring it up (and more in the same method):

- Why `return showClose`? (_Copy-paste_ bug?)
- I'd call it `is...` and not `get...` (I don't know why `getShowClose()` uses `get...`). 
- There is a typo in the javadoc:

```suggestion
/**
 * Returns <code>true</code> to indicate that the dirty indicator should be shown.
 * Otherwise return <code>false</code>.
 *
 * @return <code>true</code> if the dirty indicator should be shown
 *
 * @since 3.129
 */
public boolean isShowDirty() {
	checkWidget();
	return showDirty;
}

```

fedejeanne (commented):
I have no strong opinions about the L&F, I like this proposal.

I made only 1 comment about a possible mistake (and small details in the same method).

BeckerWdf:
> Did you consider using some highlight indicator (like for the currently selected tab) to indicate unsaved tabs? Something like this:
> image

I would not use the same indicator (location) for different stuff (active/inactive vs. saved/unsaved). Indicators that only differ in color would also be an issue for color blind people.

HeikoKlare:
> I would not use the same indicator (location) for different stuff (active/inactive vs. saved/unsaved). Indicators that only differ in color would also be an issue for color blind people.

Indeed, thank you for pointing that out. That also makes recoloring of icons an inappropriate solution.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabItem.java
fedejeanne: +1 for changing it to `isShowDirty()`. I like that convention better than using `get...` for `boolean`s.

schneidermic0:
Just a heads-up for transparency: I haven't been finding time to work on this PR lately. I had hoped to continue working on it this afternoon, but other things popped up that needed my attention.

I still want to work on these changes, but I expect it will take a while until this PR is finished. Thanks for being patient with me.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabItem.java
schneidermic0: I have renamed it to `isDirtyShown()` with https://github.com/eclipse-platform/eclipse.platform.swt/pull/1632/commits/8f9835592ac738b47ed51eff8cebc9d2dce467b9

vogella:
@schneidermic0 any news?

schneidermic0:
Sorry, no real progress here from my side. :( 

I am afraid I don't expect to find time in the near future to continue. Shall I close this PR? Is there anybody who want to take this topic up?

vogella:
Please let it open, the new indicator would be really nice. I might send Claude or Gemini to help with this PR in the future.

schneidermic0:
Thanks @vogella 👍

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
laeubi: @iloveeclipse just as a note, one should be able to use CSS for this case to hide items or chow indicatiors and so on.

vogella:
Closing in favor of https://github.com/eclipse-platform/eclipse.platform.swt/pull/3141

Issue #840, opened 2023-10-18
The internal browser is outdated


wimjongman:
We cannot receive youtube anymore in the internal browser.

![image](https://github.com/eclipse-platform/eclipse.platform.ui/assets/180969/97db8923-4ecd-444d-8df9-a8b0836dc26f)

vogella:
Most likely Windows. I have seem that too. I think SWT still uses IE on Windows.

Phillipus:
Assuming you're on Windows you could use Edge by adding this to eclipse.ini (and make sure Edge Runtime is installed):

`-Dorg.eclipse.swt.browser.DefaultType=edge`

But you'll find other issues using Edge:

https://github.com/eclipse-platform/eclipse.platform.swt/issues?q=edge+browser

vogella:
@Phillipus can Eclipse use the internal Edge browser? I though that was not possible and requires an extra installation of Edge.

vogella:
@Phillipus awesome to hear that Windows 11 provides Edge. I think if SWT runs on Windows 11 we should in this case default to Edge instead of the unmaintained IE API

HeikoKlare:
The premature state of the Edge browser can also be seen in #672. I tried to enable the browser tests for Edge with that PR, but did not yet get it running properly. Still you can see in the PR how many tests have to be disabled for the Edge browser because either there is no API for the functionality or because it has not been implemented in SWT yet.

eliasbalasis:
I understand the problem and even though there is still a temporary workaround using `Dorg.eclipse.swt.browser.DefaultType=edge` at eclipse/lsp4e#820 it is not permanent and may start breaking going forward.

Unfortunately, I don't have enough familiarity to contribute, but I do need to stress the importance.

Phillipus:
Even if Edge-related bugs are fixed a major issue to address is https://github.com/eclipse-platform/eclipse.platform.swt/issues/213, and see [this comment](https://github.com/eclipse-platform/eclipse.platform.swt/issues/213#issuecomment-1773736860).

eliasbalasis:
It starts sounding like the tip of the iceberg.

eliasbalasis:
In relation to eclipse/lsp4e#820

On Eclipse 2023-09 with m2e 2.4.0
`-Dorg.eclipse.swt.browser.DefaultType=edge` is not needed anymore on Windows OS.

wimjongman:
[...]

Issue #1466, opened 2024-09-12
Edge: Roadmap to making edge the default browser on Windows


sratz:
## Background
The Internet Explorer engine on Windows is still the default engine to date.
This engine however is heavily out-of-date and does not support the latest HTML/JS features imperative for modern developments.

The Edge implementation is available via `SWT.EDGE` flag or via `-Dorg.eclipse.swt.browser.DefaultType=edge` vm arg.

## Goal
The goal should be to make the Edge implementation the default browser engine on Windows.

## Proposal
- Use this (umbrella) issue to keep track of open issues and to categorize them by severity
- Once we have a stable enough state with no blockers, test-drive the Edge implementation by temporarily making it the default in the Platform I-Builds.
- Based on this test drive:
  - Collect more bugs and issues
  - Revert back to IE, if critical

## Issues
- Edge related Issues / PRs: https://github.com/eclipse-platform/eclipse.platform.swt/labels/edge

### Blocker / Major
- #730
- #1122
- #1013
- #738
- #671
- #669
- #213
- #212
- #1464
- #1529
- #1540
- #1845

### Minor
- #1408
- #1640
- #1771
- #1848
- #373
- #1857

### Uncategorized
- #1375
- #20
- #1581
- #1592
- #1670
- #1664

### PDE
* https://github.com/eclipse-pde/eclipse.pde/issues/1509

### Out-of-scope (e.g. Edge-specific API)
- #207
- #1834

### Browser-agnostic feature requests
- #283

konczdev:
Can #1834 be added to the scope of this issue? Without this sites behind Azure AD authentication (typically in enterprise environment) cannot be loaded with the current Edge implementation.

sratz:
> Can [#1834](https://github.com/eclipse-platform/eclipse.platform.swt/pull/1834) be added to the scope of this issue? Without this sites behind Azure AD authentication (typically in enterprise environment) cannot be loaded with the current Edge implementation.

Azure AD auth never worked in IE, did it? 

Then I don't consider it as part of the scope or replacing IE with Edge, but as additional functional requirement to be tracked separately.

selundqma:
Using Eclipse 4.35 with the Edge implementation used as default, I see a delay between my dialog showing with the Browser component and the text appearing. Also a very short text shows the same delay.. like "HELLO". If I use the SWT.IE flag, the delay is gone.

HeikoKlare:
[...]

PR #1637, opened 2024-12-06
Use Edge/WebView2 as the default browser on Windows #1466

HeikoKlare:
This change replaces Internet Explorer as the default browser used on Windows with Edge/WebView2.

Internet Explorer is progressively taken out of support while Edge/WebView2 and replace by Edge/WebView2. It provides a more future-proof browser integration capability for SWT on Windows, with better capabilities, performance, and development tooling support.

This change consists of the following:
- Introduce SWT flag SWT.IE to be used for Browser to create an Internet Explorer instance
- Adapt mapping of existing VM argument for default browser to map "ie" to new Internet Explorer flag and use Edge as default
- Make SWT Browser instantiate Edge on Windows by default, i.e., unless the flag SWT.IE is specified
- Adapt the SWT ControlExample to represent the SWT.IE flag and properly switch between Edge, Internet Explorer and default
- Adapt Browser tests to consider to new default configuration

Contributes to https://github.com/eclipse-platform/eclipse.platform.swt/issues/1466
Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/840

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Browser/common/org/eclipse/swt/browser/Browser.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Browser/common/org/eclipse/swt/browser/Browser.java
@@ -182,10 +182,10 @@ static int checkStyle(int style) {
 				if (current.equalsIgnoreCase ("webkit")) { //$NON-NLS-1$
 					DefaultType = SWT.WEBKIT;
 					break;
+				} else if (current.equalsIgnoreCase ("ie") && "win32".equals (platform)) { //$NON-NLS-1$ //$NON-NLS-2$
+					DefaultType = SWT.IE;
 				} else if (current.equalsIgnoreCase ("edge") && "win32".equals (platform)) { //$NON-NLS-1$ //$NON-NLS-2$
 					DefaultType = SWT.EDGE;
-				} else if (current.equalsIgnoreCase ("ie") && "win32".equals (platform)) { //$NON-NLS-1$ //$NON-NLS-2$
-					DefaultType = SWT.NONE;
 					break;
 				}
 				index = newIndex + 1;
@@ -196,7 +196,7 @@ static int checkStyle(int style) {
 		}
 	}
 	/* If particular backend isn't specified, use the value from the system property. */
-	if ((style & (SWT.WEBKIT | SWT.EDGE)) == 0) {
+	if ((style & (SWT.WEBKIT | SWT.IE | SWT.EDGE)) == 0) {
 		style |= DefaultType;
 	}
 	if ("win32".equals (platform) && (style & SWT.EDGE) != 0) { //$NON-NLS-1$
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
@@ -3686,9 +3686,9 @@ public void setRedraw (boolean redraw) {
 }
 
 private boolean embedsWin32Control () {
-	if (this instanceof Browser) {
+	if (this instanceof Browser browser) {
 		// The Edge browser embeds webView2
-		return (getStyle() & SWT.EDGE) != 0;
+		return "edge".equals(browser.getBrowserType());
 	}
 
 	if (this instanceof OleClientSite) {
--- a/binaries/org.eclipse.swt.cocoa.macosx.aarch64/.settings/.api_filters
+++ b/binaries/org.eclipse.swt.cocoa.macosx.aarch64/.settings/.api_filters
@@ -456,6 +456,14 @@
             </message_arguments>
         </filter>
     </resource>
+    <resource path="Eclipse SWT/common/org/eclipse/swt/SWT.java" type="org.eclipse.swt.SWT">
+        <filter id="336658481">
+            <message_arguments>
+                <message_argument value="org.eclipse.swt.SWT"/>
+                <message_argument value="IE"/>
+            </message_arguments>
+        </filter>
+    </resource>
     <resource path="META-INF/MANIFEST.MF">
         <filter id="926941240">
             <message_arguments>
--- a/binaries/org.eclipse.swt.cocoa.macosx.x86_64/.settings/.api_filters
+++ b/binaries/org.eclipse.swt.cocoa.macosx.x86_64/.settings/.api_filters
@@ -456,6 +456,14 @@
             </message_arguments>
         </filter>
     </resource>
+    <resource path="Eclipse SWT/common/org/eclipse/swt/SWT.java" type="org.eclipse.swt.SWT">
+        <filter id="336658481">
+            <message_arguments>
+                <message_argument value="org.eclipse.swt.SWT"/>
+                <message_argument value="IE"/>
+            </message_arguments>
+        </filter>
+    </resource>
     <resource path="META-INF/MANIFEST.MF">
         <filter id="926941240">
             <message_arguments>
--- a/binaries/org.eclipse.swt.gtk.linux.aarch64/.settings/.api_filters
+++ b/binaries/org.eclipse.swt.gtk.linux.aarch64/.settings/.api_filters
@@ -456,6 +456,14 @@
             </message_arguments>
         </filter>
     </resource>
+    <resource path="Eclipse SWT/common/org/eclipse/swt/SWT.java" type="org.eclipse.swt.SWT">
+        <filter id="336658481">
+            <message_arguments>
+                <message_argument value="org.eclipse.swt.SWT"/>
+                <message_argument value="IE"/>
+            </message_arguments>
+        </filter>
+    </resource>
     <resource path="META-INF/MANIFEST.MF">
         <filter id="926941240">
             <message_arguments>
--- a/binaries/org.eclipse.swt.gtk.linux.loongarch64/.settings/.api_filters
+++ b/binaries/org.eclipse.swt.gtk.linux.loongarch64/.settings/.api_filters
@@ -456,6 +456,14 @@
             </message_arguments>
         </filter>
     </resource>
+    <resource path="Eclipse SWT/common/org/eclipse/swt/SWT.java" type="org.eclipse.swt.SWT">
+        <filter id="336658481">
+            <message_arguments>
+                <message_argument value="org.eclipse.swt.SWT"/>
+                <message_argument value="IE"/>
+            </message_arguments>
+        </filter>
+    </resource>
     <resource path="META-INF/MANIFEST.MF">
         <filter id="926941240">
             <message_arguments>
--- a/binaries/org.eclipse.swt.gtk.linux.ppc64le/.settings/.api_filters
+++ b/binaries/org.eclipse.swt.gtk.linux.ppc64le/.settings/.api_filters
@@ -456,6 +456,14 @@
             </message_arguments>
         </filter>
     </resource>
+    <resource path="Eclipse SWT/common/org/eclipse/swt/SWT.java" type="org.eclipse.swt.SWT">
+        <filter id="336658481">
+            <message_arguments>
+                <message_argument value="org.eclipse.swt.SWT"/>
+                <message_argument value="IE"/>
+            </message_arguments>
+        </filter>
+    </resource>
     <resource path="META-INF/MANIFEST.MF">
         <filter id="926941240">
             <message_arguments>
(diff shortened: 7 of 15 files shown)


bundles/org.eclipse.swt/Eclipse SWT Browser/common/org/eclipse/swt/browser/Browser.java
```diff
@@ -182,9 +182,9 @@ static int checkStyle(int style) {
 				if (current.equalsIgnoreCase ("webkit")) { //$NON-NLS-1$
 					DefaultType = SWT.WEBKIT;
 					break;
-				} else if (current.equalsIgnoreCase ("edge") && "win32".equals (platform)) { //$NON-NLS-1$ //$NON-NLS-2$
-					DefaultType = SWT.EDGE;
 				} else if (current.equalsIgnoreCase ("ie") && "win32".equals (platform)) { //$NON-NLS-1$ //$NON-NLS-2$
+					DefaultType = SWT.IE;
+				} else if (current.equalsIgnoreCase ("edge") && "win32".equals (platform)) { //$NON-NLS-1$ //$NON-NLS-2$
```
fedejeanne, line 187: This change breaks the following check (introduced by #1588) and therefore brings back the flickering:

https://github.com/amartya4256/eclipse.platform.swt/blob/7cee4b6bc7c8c2ef3059129b6a0dec765c5eb0b5/bundles/org.eclipse.swt/Eclipse%20SWT/win32/org/eclipse/swt/widgets/Control.java#L3691

The solution is to change the check in `Control`

```
if (this instanceof Browser) {
	// The Edge browser (default in Windows) embeds webView2
	return (getStyle() & SWT.IE) == 0;
}
```

fedejeanne (changes requested):

bundles/org.eclipse.swt/Eclipse SWT Browser/common/org/eclipse/swt/browser/Browser.java
```diff
@@ -182,9 +182,9 @@ static int checkStyle(int style) {
...
 					DefaultType = SWT.NONE;
```
sratz, line 188: Couldn't we also just set `SWT.EDGE` here explicitly instead of `NONE`?

bundles/org.eclipse.swt/Eclipse SWT Browser/common/org/eclipse/swt/browser/Browser.java
HeikoKlare: Thanks for the pointer! I have developed on a state without that PR, so I missed that usage of the flag. I will update the PR accordingly.

bundles/org.eclipse.swt/Eclipse SWT Browser/common/org/eclipse/swt/browser/Browser.java
HeikoKlare: Yes, we could. The result should be the same. I thought this could make explicit that Edge is default and ensure that we do not have different behavior depending on whether the flag is `NONE` or `EDGE`. But I agree that this should be ensured elsewhere and in case someone explicitly specifies "edge", we can/should also explicitly set the according flag. I will update the PR accordingly.

fedejeanne: In that case I would say:
- Get rid of the `else if (current.equalsIgnoreCase ("edge") ...` block altogether
- Change this a few lines below:

```java
if (DefaultType == SWT.DEFAULT) {
	DefaultType = SWT.EDGE; // was SWT.NONE
}
```

BTW that would also fix the flickering so there would be no need for this fix: https://github.com/eclipse-platform/eclipse.platform.swt/pull/1637#discussion_r1875768494

HeikoKlare: The code is OS independent, so we cannot simply set default to Edge. And since "edge" was (and still is) an explicit flag, we should have code documenting its purpose (even though it becomes kind of a no-op with this change).

bundles/org.eclipse.swt/Eclipse SWT Browser/common/org/eclipse/swt/browser/Browser.java
HeikoKlare: Fixed with 075d949

bundles/org.eclipse.swt/Eclipse SWT Browser/common/org/eclipse/swt/browser/Browser.java
HeikoKlare: Updated with 075d949

bundles/org.eclipse.swt/Eclipse SWT Browser/common/org/eclipse/swt/browser/Browser.java
fedejeanne: My pleasure.

One thing though: checking for `(getStyle() & (SWT.NONE | SWT.EDGE)) != 0` like 1b76c75e2f517c4265ca75df588e540aabbb61c7 does is not good enough because `x & SWT.NONE == 0` which means that `embedsWind32Control` will always return `false`.

## How to test
- Add one `sysout` in `Control::setRedraw`
```java
if (!redraw && isShown && embedsWin32Control()) {
	System.out.println("Embeds"); // add this one
	return;
}
```
- Start the _Runtime Application_ without any special parameters (use the default browser)
- Resize the _Javadoc_ view.

You'll see that the console does **not** print any `Embeds`

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
```diff
@@ -3688,7 +3688,7 @@ public void setRedraw (boolean redraw) {
 private boolean embedsWin32Control () {
 	if (this instanceof Browser) {
 		// The Edge browser embeds webView2
-		return (getStyle() & SWT.EDGE) != 0;
+		return (getStyle() & SWT.IE) == 0;
```
sratz, line 3691: Now that we always have the "EDGE" flag set, I think we should keep this check as it was before

bundles/org.eclipse.swt/Eclipse SWT Browser/common/org/eclipse/swt/browser/Browser.java
```diff
@@ -196,7 +196,7 @@ static int checkStyle(int style) {
 		}
 	}
 	/* If particular backend isn't specified, use the value from the system property. */
-	if ((style & (SWT.WEBKIT | SWT.EDGE)) == 0) {
+	if ((style & (SWT.WEBKIT | SWT.IE)) == 0) {
```
sratz, line 199:
```suggestion
	if ((style & (SWT.WEBKIT | SWT.EDGE | SWT.IE)) == 0) {
```

bundles/org.eclipse.swt/Eclipse SWT Browser/common/org/eclipse/swt/browser/Browser.java
HeikoKlare: I already did that with my recent commit, or do I misunderstand the comment?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
HeikoKlare: We do not always have the flag set. By default, the style flag will simply be `SWT.NONE`.

fedejeanne: In its current state (075d949dc7e5f3c27e68075e75a9126c82913767) this wouldn't work because `SWT.EDGE` is set `if (current.equalsIgnoreCase ("edge") && "win32".equals (platform))` which means one has to explicitly provide the parameter. Not providing any parameter still sets the style to `SWT.NONE`

bundles/org.eclipse.swt/Eclipse SWT Browser/common/org/eclipse/swt/browser/Browser.java
HeikoKlare: Fixed with 428a039.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
HeikoKlare: Sounds good. I've adapted the code with c0466d7.

sratz:
Changes look good.

But what's that weird

_"The field org.eclipse.swt.SWT.IE has been added to a class"_

API error about?

HeikoKlare:
> But what's that weird
> 
> _"The field org.eclipse.swt.SWT.IE has been added to a class"_
> 
> API error about?

Where do you see that error? The added field is an (accepted) API change for which I added an according compatibility filter.

sratz:
> > But what's that weird
> > _"The field org.eclipse.swt.SWT.IE has been added to a class"_
> > API error about?
> 
> Where do you see that error? The added field is an (accepted) API change for which I added an according compatibility filter.

I wanted to see what the filters were for and temporarily removed it and was wondering why that's the only occurrence of `SWT` in the API filters / why this was not needed before.

HeikoKlare:
> I wanted to see what the filters were for and temporarily removed it

The filter is for introduced breaking API change. Adding a public field to a class is actually a breaking change (but one that you can usually accept as the chance of any issue because of that are unlikely).

> and was wondering why that's the only occurrence of `SWT` in the API filters / why this was not needed before.

I think there is no other filter for the `SWT` class because for every other breaking API change, for which a filter was added, that filter was removed in the subsequent release cycle, because the API then became compatible with the baseline version (the previous release) again.

sratz (approved):

sratz:
> I think there is no other filter for the `SWT` class because for every other breaking API change, for which a filter was added, that filter was removed in the subsequent release cycle, because the API then became compatible with the baseline version (the previous release) again.

Makes sense, thanks for the clarification!

akurtakov:
This seem to have missed a version bump as can be seen at https://ci.eclipse.org/releng/job/eclipse.platform.swt/job/master/828/console and even fails https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2649 .
I really ask everyone to be far more careful in expecting and enhancing releng scripts to be more strict where they aren't as keeping things building becomes a full time job without any possibility to look into anything else for me.

fedejeanne:
@akurtakov I don't see the errors that you mention in the [logs](https://ci.eclipse.org/releng/job/eclipse.platform.swt/job/master/828/console), only the 3 know test failures: https://github.com/eclipse-platform/eclipse.platform.swt/issues/1564.

And I don't see the relation to https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2649 or https://github.com/eclipse-equinox/equinox/pull/716 in this PR either.

Could you please be more specific so next time we know what to look for?

HeikoKlare:
We missed these messages:

>  Caused by: org.eclipse.tycho.core.exceptions.VersionBumpRequiredException: Only qualifier changed for (org.eclipse.swt.examples/3.108.600.v20241210-1215). Expected to have bigger x.y.z than what is available in baseline (3.108.600.v20241014-1802)

Sorry for that and thank you for bumping the version, @akurtakov. Don't know why/how I missed that...
Seems to be quite hard to track around the failures we currently have in SWT builds. We should definitely find a solution for the Linux test failures, because they make so many checks fail unnecessarily. This is not meant to ask anyone specific to look at those tests, but honestly, if no one really cares about the issues those tests reveal, we may also think about simply deactivating those tests on Linux.

I see that the Windows tests are also failing today because of some unsuccessful cleanup related to this PR (as it is related to Edge being used), see https://ci.eclipse.org/releng/job/eclipse.platform.swt/job/master/828/console:
> C:\Users\genie.releng\workspace\AutomatedTests\ep435I-unit-win32-x86_64-java17\workarea\I20241211-0310\eclipse-testing\test.xml:132: Unable to delete file C:\Users\genie.releng\workspace\AutomatedTests\ep435I-unit-win32-x86_64-java17\workarea\I20241211-0310\eclipse-testing\test-eclipse\eclipse\update_configurator_folder\.metadata\.plugins\org.eclipse.swt\EBWebView\lockfile

Do you have a look at them, @fedejeanne @amartya4256?

PR #1645, opened 2024-12-11
Version bump for new stream

akurtakov:
Fixes broken build after
https://github.com/eclipse-platform/eclipse.platform.swt/pull/1637


akurtakov:
Test failures are not related to version bump in this commit.

PR #1646, opened 2024-12-11
Device getDPI scaling fix

### LINKED ISSUE #1671: [Win32] Broken compatibility with swt.autoScale=false
**Describe the bug**
Because of [serious painting problems on Windows](https://github.com/eclipse-platform/eclipse.platform.swt/issues/1596) we have to set `swt.autoScale` to `false` and use `display.getDPI()` to (indirectly) find out the zoom level. However, build c66545ba broke this compatibility.

**To Reproduce**
Run this snippet
```java
import org.eclipse.swt.*;

public class AutoScaleFalseTest {

		System.getProperties().putIfAbsent("swt.autoScale", "false");

		shell.addListener(SWT.Paint, event -> {
			final String deviseZoom = System.getProperty("org.eclipse.swt.internal.deviceZoom");
			event.gc.drawString(deviseZoom, 0, 0);
			final Point dpi = display.getDPI();
			event.gc.drawString(String.valueOf(dpi), 0, 30);
		});

		shell.setSize(400, 300);
		shell.open();

		while (!shell.isDisposed()) {
				display.sleep();
			}
		}

		display.dispose();
	}
}
```
on commit 40b4fc6c and on commit c66545ba on a system with 4k monitor set to 150% zoom level. The first will report correctly 144 dpi, the latter 96dpi.

**Expected behavior**
If `swt.autoScale` is set to `false`, no autoscaling must happen.

**Screenshots**
Commit 40b4fc6c:
40b4fc6c

Commit c66545ba:
c66545ba

**Environment:**
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 

We are Win11 24H2, but this should not have any influence.

Nothing known.

amartya4256:
This contribution fixes the device getDPI calculation logic by usibg the device zoom to be used for scaling down the DPI values instead of the native zoom.

contributes to #62 and #127

Fixes #1671

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Device.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Device.java
@@ -519,7 +519,7 @@ public Point getDPI () {
 	int dpiX = OS.GetDeviceCaps (hDC, OS.LOGPIXELSX);
 	int dpiY = OS.GetDeviceCaps (hDC, OS.LOGPIXELSY);
 	internal_dispose_GC (hDC, null);
-	return DPIUtil.scaleDown(new Point (dpiX, dpiY), getDeviceZoom());
+	return DPIUtil.scaleDown(new Point (dpiX, dpiY), DPIUtil.getZoomForAutoscaleProperty(getDeviceZoom()));
 }
 
 /**


jukzi:
Until CI and I Builds are green again please hold back any submit that does not contribute to that.

tmssngr:
This fixes #1671 for me. Thanks.

fedejeanne (approved):

akurtakov:
@fedejeanne I see more test failures and with different errors like https://ci.eclipse.org/releng/job/eclipse.platform.swt/job/PR-1646/1/testReport/junit/org.eclipse.swt.tests.junit/Test_org_eclipse_swt_browser_Browser/test_BrowserFunction_callback_with_boolean_browser_flags__0_/ . Would you please investigate whether it comes from this PR or some of the others just merged (I haven't investigated which one exactly caused it).

fedejeanne:
In this PR? I don't see it.
In any case, that is also one of the "usual suspects": #1676

Issue #1438, opened 2024-09-02
Improving Eclipse Icon Scaling by Supporting Vectorized Icons


Michael5601:
Eclipse currently relies on PNG files for displaying icons, requiring two different sizes to handle various zoom levels. For zoom factors between 100% and 174%, the original-size icon is used, while for zoom levels between 175% and 200%, a double-sized icon is displayed. This approach has several drawbacks:

1. **Quality Loss with Autoscaling**: Enabling autoscaling for icons using the flags `-Dswt.autoScale=quarter` or `-Dswt.autoScale=exact` can cause quality degradation, particularly on high-DPI displays or at high zoom levels, since PNGs do not scale well.

2. **Maintenance Challenges**: Maintaining multiple PNG versions in different sizes is not a sustainable solution, especially as higher DPI displays may require even more icon variations.

Currently, new icons are manually created by converting SVGs to PNGs, a process that is both time-consuming and inefficient. By fully utilizing SVGs, which support resolution-independent scaling, these problems can be effectively addressed.

Previous discussions have highlighted these issues, such as those documented in this [Eclipse bug report](https://bugs.eclipse.org/bugs/show_bug.cgi?id=568215).

My bachelor’s thesis explores how SVGs can be better integrated into Eclipse to improve icon scalability and reduce maintenance overhead. If you have any insights or suggestions, please feel free to reach out or leave a comment!

BeckerWdf:
> On the other hand it is no longer true that Eclipse only supports 2 sizes (actually it has before supported 1.5 as well) it is just what is the most common today, Eclipse also supports a scheme where one can specify a URL that uses MxN as part of its name and the one can support any zoom level.
Interesting. I didn't know this. Where is that documented?

> So for my usecase I finally ended up in having a script that automatically generates the icons from a SVG using Inkscape (what is the only descend SVG render I know about) and place them in the bundle, this increases the size a bit but gives the best, reproducible and portable result.

It that script publicly available. And if now would you mind sharing it?

laeubi:
> It that script publicly available. And if now would you mind sharing it?

It is not public and rather roughly put together, something similar is already available here:

https://github.com/eclipse-platform/eclipse.platform.images/tree/master/org.eclipse.images.renderer

in my case it just a bit more automated, it first calculates all NxM variants for desired zoomfactors, then check if the source SVG has > modification time than target (or target do not exits) and then creates the PNG in a corresponding folder structure recognized by Eclipse.

[...]

PR #1638, opened 2024-12-06
Rasterization of SVGs at Runtime for Eclipse Icons 

Michael5601:
Eclipse currently loads icons exclusively as raster graphics (e.g., `.png`), without support for vector formats like `.svg`. A major drawback of raster graphics is their inability to scale without degrading image quality. Additionally, generating icons of different sizes requires manually rasterizing SVGs outside Eclipse, leading to unnecessary effort and many icon files.

This PR introduces support for vector graphics in Eclipse, enabling SVGs to be used for icons. Existing PNG icons will continue to be loaded alongside SVGs, allowing the use of the new functionality without the need to replace all PNG files at once.

---

### **Key Features**

- **Example Images**:  
  Screenshots showcasing the new functionality can be found below. These screenshots were taken with 125% monitor-zoom and scaling enabled with the flag `-Dswt.autoScale=quarter`.
  
| PNG      | SVG      |
|---------------|---------------|
| PNG  | SVG  |
| PNG2  | SVG2  |
| PNG3  | SVG3  |
| PNG4  | SVG4  |

  

- **How It Works**:  
  - To use SVG icons, simply place the SVG file in the bundle and reference it in the `plugin.xml` and other necessary locations, as is done for PNGs. No additional configuration is required.  
  - At runtime, Eclipse uses the library JSVG to rasterize the SVG into a raster image of the desired size, eliminating the need for scaling. My analysis shows that JSVG is the most suitable Java library for this purpose.
  - You need to write the flag `-Dswt.autoScale=quarter` into your `eclipse.ini` file or into the run arguments of a new configuration.

- **Demonstration**:  
  This PR includes SVG versions of icons for the following bundles:  
  - `org.eclipse.search`  
  - `org.eclipse.ui`  
  - `org.eclipse.ui.ide`  
  - `org.eclipse.ui.ide.application`  
  - `org.eclipse.ui.editors`  
  - `org.eclipse.ui.navigator`  

Due to this the PR of Platform UI is very large. If preferred, I can limit the changes to `org.eclipse.search`, allowing SVG icons to be tested specifically in the search result window.
I did not delete any exisiting PNGs. 

---

### **Architecture**

Eclipse icons fall into three categories, all three adressed by this PR:

1. **Standard Icons**:  
   These is the most common type of icons. They are loaded, scaled if needed, and displayed in the UI.

2. **Composite Icons**:  
   Composite icons combine a base icon with up to four overlay icons. Each component is loaded separately and then combined for display.

[...]

PR #1647, opened 2024-12-11
Follow-Up Feature: Icon-Enablement with Rasterization of SVGs

Michael5601:
This draft outlines the follow-up functionality for the [PR](https://github.com/eclipse-platform/eclipse.platform.swt/pull/1638) (Feature Proposal: Rasterization of SVGs at Runtime for Eclipse Icons).

The commit 4e6abc7c21d10d293a68288170ef91e086d6fef1 contains the new functionality that extends the base functionality of the mentioned PR. All future changes to the PR will be performed to this draft.

The new functionality extends the current automatic customization of icons in the `Image(Device device, Image srcImage, int flag)` constructor. While all of the core functionality is implemented within SWT, some changes are required in Platform UI, which can be found in the following [Draft](https://github.com/eclipse-platform/eclipse.platform.ui/pull/2621). 

When the `Image` constructor is invoked, the new functionality attempts to create a graphically customized icon by passing a specific flag (e.g., `SWT.IMAGE_DISABLE`) to the rasterization functionality introduced in the base PR. This functionality is enhanced by a preprocessing step that automatically adds a filter to the SVG before rasterization. These filters are designed to ensure that the resulting icons resemble the current pre-created disabled/gray icons loaded at runtime. 

It is important to note that the automatic icon customization differs between GTK and Cocoa/Win32. As such, the icons generated with the new functionality may not align perfectly with the automatically customized GTK icons. However, this discrepancy is also present with the current pre-created icons, so the impact is minimal. Additionally, most icons are pre-created and not automatically customized at runtime. 

Below is a comparison between the current pre-created icons and the icons customized automatically at runtime using the new functionality. The color/saturation/brightness can be changed by adjusting the filter:

**Pre-created and scaled down to 125% device zoom:**
image

**Automatically customized at runtime with the new functionality:**
image

The following tasks need to be completed for this draft:

1. Decide whether the unification of the behavior across different operating systems should occur before merging the new functionality.
2. Merge the main feature upon which this draft is based.
3. Remove disabled icon paths from the `plugin.xml` files to ensure the automatic customization is triggered.
4. Write documentation (JavaDocs) for the new API.
5. Provide regression tests and a performance comparison between automatic customization and FileIO for pre-created icons. The possibility exists that the rasterization with pre-processing is faster than the additional FileIO to load the pre-created icons. 

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/1438.

### DIFF
--- a/bundles/org.eclipse.swt.svg/src/org/eclipse/swt/svg/JSVGRasterizer.java
+++ b/bundles/org.eclipse.swt.svg/src/org/eclipse/swt/svg/JSVGRasterizer.java
@@ -35,14 +35,29 @@ import java.awt.Graphics2D;
 import java.awt.RenderingHints.Key;
 import java.awt.image.BufferedImage;
 import java.awt.image.DataBufferInt;
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.OutputStream;
 import java.util.Map;
 
 import org.eclipse.swt.SWT;
 import org.eclipse.swt.graphics.ImageData;
 import org.eclipse.swt.graphics.PaletteData;
 import org.eclipse.swt.internal.image.SVGRasterizer;
+
+import javax.xml.parsers.DocumentBuilder;
+import javax.xml.parsers.DocumentBuilderFactory;
+import javax.xml.parsers.ParserConfigurationException;
+import javax.xml.transform.Transformer;
+import javax.xml.transform.TransformerException;
+import javax.xml.transform.TransformerFactory;
+import javax.xml.transform.dom.DOMSource;
+import javax.xml.transform.stream.StreamResult;
+import org.w3c.dom.Document;
+import org.w3c.dom.Element;
+import org.xml.sax.SAXException;
 
 import com.github.weisj.jsvg.SVGDocument;
 import com.github.weisj.jsvg.geometry.size.FloatSize;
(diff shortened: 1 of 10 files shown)


BeckerWdf:
> Below is a comparison between the current pre-created icons and the icons customized automatically at runtime using the new functionality. The color/saturation/brightness can be changed by adjusting the filter:
> 
> **Pre-created and scaled down to 125% device zoom:** image
> 
> **Automatically customized at runtime with the new functionality:** image

To be honest I almost cannot see a difference. So I think this should not be a problem.

Michael5601:
> To be honest I almost cannot see a difference. So I think this should not be a problem.

This is a good thing for me. :) We could utilize the new behaviour from this draft, which means the pre-created PNGs can be removed and we have a little improvement in the look of the icons without changing their appearance (gray tone) too much.

akurtakov:

Michael5601:
> What is the status of this one?

We are currently working on this [PR](https://github.com/eclipse-platform/eclipse.platform.ui/pull/2593) in Platform UI, so that SVGs can be also used by external consumers like `URLImageDescriptor` or `FileImageDescriptor` in Jface. With this also Composite Icons can be loaded by using SVGs. 

The icon disablement as proposed in this Draft needs further adjustments as it does not work in Cocoa yet and also new API needs to be added so that icons can be created with arbitrary Hue, Saturation and Brightness at runtime. Additionally we should first merge this [PR](https://github.com/eclipse-platform/eclipse.platform.swt/pull/1936) to unify the disablement algorithms for all OS.

Apart from that the functionality in this draft is working and allows the disablement for all icons by perprocessing SVGs before rasterization. Please note that due to the different implementation for cocoa it could happen that the changes proposed in this draft will be dismissed completely.

PR #1657, opened 2024-12-13
Disable test_TabTraversalOutOfBrowser for IE

fedejeanne:
In `class Test_org_eclipse_swt_browser_Browser`.


fedejeanne:
FTR the functionality in IE is not broken, just the text.

I opened the Welcome page with IE (`-Dorg.eclipse.swt.browser.DefaultType=ie`) and was able to navigate between elements using `Tab` / `Shift` + `Tab`

navigate_tab_key_ie

PR #1662, opened 2024-12-13
SIGSEGV in Tree.cellDataProc when calling TreeItem.setImage #678

zkxvh:
Fix and test for bug #678

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/678

Reproducing the crash:
- https://github.com/eclipse-platform/eclipse.platform.swt/issues/678#issue-1715955828
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/1611/
- https://github.com/eclipse-platform/eclipse.platform.swt/issues/678#issuecomment-2507994360

The cause of the crash is described here: https://github.com/eclipse-platform/eclipse.platform.swt/issues/678#issuecomment-2511711696  
In short, the problem was executing `Tree.createRenderers()` inside `TreeItem.setImage()`.  
The sequence of action leading to the crash was:  
in a `Tree` with `SWT.VIRTUAL` a `TreeItem` is rendered for the first time or after `clear()`  
- `Tree.cellDataProc()` is executed for the item and one of it's
renderers  
  - `Tree.checkData(item)` is called for the item  
    - `SWT.SetData` is invoked and executes `TreeItem.setImage()`  
      - `Tree.createRenderers()` executes and disposes the current renderer  
  - further actions in `Tree.cellDataProc()` that access the already-disposed renderer (they think it's alive)

How it's fixed:
1. set fixed height+width to pixbuf renderers with `GTK.gtk_cell_renderer_set_fixed_size`  
   This does 2 things:
     - from now on it makes images rendered by the renderer to be rendered with these height+width
     - from now on fixed row height calculation will return height of at least fixed height of the pixbuf
2. make `GtkTreeView` recompute it's internally stored fixed row height with re-setting `fixed_height_mode` of the GtkTreeView
3. make all existing tree rows update their height by invoking `gtk_tree_view_row_changed()` on each of the rows.

(2) and (3) can also be achieved by creating a copy of the current `GtkTreeModel` and setting it to the `GtkTreeView`.  
But this also resets current selected rows, focus and scroll position; that's why I chose re-setting `fixed_height_mode` + `gtk_tree_view_row_changed()` instead.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Tree.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Tree.java
@@ -4319,6 +4319,80 @@ void checkSetDataInProcessBeforeRemoval() {
 	}
 }
 
+boolean initPixbufSize(Image image) {
+	if (pixbufSizeSet || image == null) {
+		return false;
+	}
+	int iWidth, iHeight;
+	if (DPIUtil.useCairoAutoScale()) {
+		iWidth = image.getBounds().width;
+		iHeight = image.getBounds().height;
+	} else {
+		iWidth = image.getBoundsInPixels().width;
+		iHeight = image.getBoundsInPixels().height;
+	}
+	if (iWidth <= 0 || iHeight <= 0) {
+		return false;
+	}
+	pixbufSizeSet = true;
+	pixbufHeight = iHeight;
+	pixbufWidth = iWidth;
+	/*
+	 * Feature in GTK: a Tree with the style SWT.VIRTUAL has fixed-height-mode
+	 * enabled. This will limit the size of any cells, including renderers. In order
+	 * to prevent images from disappearing/being cropped, we update row heights when
+	 * the first image is set. Fix for bug 480261.
+	 */
+	if ((style & SWT.VIRTUAL) != 0) {
+		resetFixedRowHeight();
+	}
+	return true;
+}
+
+private void resetFixedRowHeight() {
+	long columnList = GTK.gtk_tree_view_get_columns(handle);
+	if (columnList == 0) {
+		return;
+	}
+
+	// set fixed width and height for all GtkCellRendererPixbufs
+	for (var colItem = columnList; colItem != 0; colItem = OS.g_list_next(colItem)) {
+		long column = OS.g_list_data(colItem);
+		var cellList = GTK.gtk_cell_layout_get_cells(column);
+		for (var cellItem = cellList; cellItem != 0; cellItem = OS.g_list_next(cellItem)) {
+			var renderer = OS.g_list_data(cellItem);
+			if (GTK.GTK_IS_CELL_RENDERER_PIXBUF(renderer)) {
+				GTK.gtk_cell_renderer_set_fixed_size(renderer, pixbufWidth, pixbufHeight);
+			}
+		}
+		OS.g_list_free(cellList);
+	}
+	OS.g_list_free(columnList);
+
+	// Executed in asyncExec because when this method is invoked from checkData(),
+	// then the "row_changed" signal is blocked (but we need it unblocked to invalidate
+	// existing rows).
+	display.asyncExec(() -> {
+		if (!isDisposed() && modelHandle != 0) {
+			// reset computed 'fixed_height' to '-1'
+			OS.g_object_set(handle, OS.fixed_height_mode, false, 0);
+			OS.g_object_set(handle, OS.fixed_height_mode, true, 0);
+
+			// to update height of the existing rows we need to invalidate them in gtk
+			// we do that by invoking gtk_tree_view_row_changed on each of them
+			long iter = OS.g_malloc(GTK.GtkTreeIter_sizeof());
+			if (GTK.gtk_tree_model_get_iter_first(modelHandle, iter)) {
+				int[] value = new int[1];
+				do {
+					GTK.gtk_tree_model_get(modelHandle, iter, ID_COLUMN, value, -1);
+					GTK.gtk_tree_store_set(modelHandle, iter, ID_COLUMN, value[0], -1);
+				} while (GTK.gtk_tree_model_iter_next(modelHandle, iter));
+			}
+			OS.g_free(iter);
+		}
+	});
+}
+
 private void throwCannotRemoveItem(int i) {
 	String message = "Cannot remove item with index " + i + ".";
 	throw new SWTException(message);
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Tree.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Tree.java
@@ -4319,17 +4319,17 @@ void checkSetDataInProcessBeforeRemoval() {
 	}
 }
 
-boolean initPixbufSize(Image image) {
+boolean initPixbufSize (Image image) {
 	if (pixbufSizeSet || image == null) {
 		return false;
 	}
 	int iWidth, iHeight;
-	if (DPIUtil.useCairoAutoScale()) {
-		iWidth = image.getBounds().width;
-		iHeight = image.getBounds().height;
+	if (DPIUtil.useCairoAutoScale ()) {
+		iWidth = image.getBounds ().width;
+		iHeight = image.getBounds ().height;
 	} else {
-		iWidth = image.getBoundsInPixels().width;
-		iHeight = image.getBoundsInPixels().height;
+		iWidth = image.getBoundsInPixels ().width;
+		iHeight = image.getBoundsInPixels ().height;
 	}
 	if (iWidth <= 0 || iHeight <= 0) {
 		return false;
@@ -4344,51 +4344,51 @@ boolean initPixbufSize(Image image) {
 	 * the first image is set. Fix for bug 480261.
 	 */
 	if ((style & SWT.VIRTUAL) != 0) {
-		resetFixedRowHeight();
+		resetFixedRowHeight ();
 	}
 	return true;
 }
 
-private void resetFixedRowHeight() {
-	long columnList = GTK.gtk_tree_view_get_columns(handle);
+private void resetFixedRowHeight () {
+	long columnList = GTK.gtk_tree_view_get_columns (handle);
 	if (columnList == 0) {
 		return;
 	}
 
 	// set fixed width and height for all GtkCellRendererPixbufs
-	for (var colItem = columnList; colItem != 0; colItem = OS.g_list_next(colItem)) {
-		long column = OS.g_list_data(colItem);
-		var cellList = GTK.gtk_cell_layout_get_cells(column);
-		for (var cellItem = cellList; cellItem != 0; cellItem = OS.g_list_next(cellItem)) {
-			var renderer = OS.g_list_data(cellItem);
-			if (GTK.GTK_IS_CELL_RENDERER_PIXBUF(renderer)) {
-				GTK.gtk_cell_renderer_set_fixed_size(renderer, pixbufWidth, pixbufHeight);
+	for (var colItem = columnList; colItem != 0; colItem = OS.g_list_next (colItem)) {
+		long column = OS.g_list_data (colItem);
+		var cellList = GTK.gtk_cell_layout_get_cells (column);
+		for (var cellItem = cellList; cellItem != 0; cellItem = OS.g_list_next (cellItem)) {
+			var renderer = OS.g_list_data (cellItem);
+			if (GTK.GTK_IS_CELL_RENDERER_PIXBUF (renderer)) {
+				GTK.gtk_cell_renderer_set_fixed_size (renderer, pixbufWidth, pixbufHeight);
 			}
 		}
-		OS.g_list_free(cellList);
+		OS.g_list_free (cellList);
 	}
-	OS.g_list_free(columnList);
+	OS.g_list_free (columnList);
 
 	// Executed in asyncExec because when this method is invoked from checkData(),
-	// then the "row_changed" signal is blocked (but we need it unblocked to invalidate
-	// existing rows).
-	display.asyncExec(() -> {
-		if (!isDisposed() && modelHandle != 0) {
+	// then the "row_changed" signal is blocked (but we need it unblocked to
+	// invalidate existing rows).
+	display.asyncExec ( () -> {
+		if (!isDisposed () && modelHandle != 0) {
 			// reset computed 'fixed_height' to '-1'
-			OS.g_object_set(handle, OS.fixed_height_mode, false, 0);
-			OS.g_object_set(handle, OS.fixed_height_mode, true, 0);
+			OS.g_object_set (handle, OS.fixed_height_mode, false, 0);
+			OS.g_object_set (handle, OS.fixed_height_mode, true, 0);
 
 			// to update height of the existing rows we need to invalidate them in gtk
 			// we do that by invoking gtk_tree_view_row_changed on each of them
-			long iter = OS.g_malloc(GTK.GtkTreeIter_sizeof());
-			if (GTK.gtk_tree_model_get_iter_first(modelHandle, iter)) {
+			long iter = OS.g_malloc (GTK.GtkTreeIter_sizeof ());
+			if (GTK.gtk_tree_model_get_iter_first (modelHandle, iter)) {
 				int[] value = new int[1];
 				do {
-					GTK.gtk_tree_model_get(modelHandle, iter, ID_COLUMN, value, -1);
-					GTK.gtk_tree_store_set(modelHandle, iter, ID_COLUMN, value[0], -1);
-				} while (GTK.gtk_tree_model_iter_next(modelHandle, iter));
+					GTK.gtk_tree_model_get (modelHandle, iter, ID_COLUMN, value, -1);
+					GTK.gtk_tree_store_set (modelHandle, iter, ID_COLUMN, value[0], -1);
+				} while (GTK.gtk_tree_model_iter_next (modelHandle, iter));
 			}
-			OS.g_free(iter);
+			OS.g_free (iter);
 		}
 	});
 }
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/TreeItem.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/TreeItem.java
@@ -1519,59 +1519,18 @@ public void setImage(int index, Image image) {
 		pixbuf = ImageList.createPixbuf(surface);
 	}
 
-	int modelIndex = parent.columnCount == 0 ? Tree.FIRST_COLUMN : parent.columns [index].modelIndex;
-	long parentHandle = parent.handle;
-	long column = GTK.gtk_tree_view_get_column (parentHandle, index);
-	long pixbufRenderer = parent.getPixbufRenderer (column);
-	int [] currentWidth = new int [1];
-	int [] currentHeight= new int [1];
-	GTK.gtk_cell_renderer_get_fixed_size (pixbufRenderer, currentWidth, currentHeight);
 	if (!parent.pixbufSizeSet) {
-		if (image != null) {
-			int iWidth, iHeight;
-			if (DPIUtil.useCairoAutoScale()) {
-				iWidth = image.getBounds ().width;
-				iHeight = image.getBounds ().height;
-			} else {
-				iWidth = image.getBoundsInPixels ().width;
-				iHeight = image.getBoundsInPixels ().height;
-			}
-			if (iWidth > currentWidth [0] || iHeight > currentHeight [0]) {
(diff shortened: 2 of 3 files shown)


zkxvh:
There were 3 failed tests in the build above.  
As I understand, the tests are for `Browser` widget and the code in this PR doesn't affect `Browser` any any way.

basilevs:
If you use asyncExec already, why do all the other changes? The problem was never the computation of heights, it was resource management. And resources are managed just fine as long as you defer their disposal. Therefore, my fix in #1612 is  less intrusive and achieves the same results at less cost (admittedly I have not tested if selection is lost, but it might).

basilevs (approved):

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Tree.java
```diff
@@ -4319,6 +4319,80 @@ void checkSetDataInProcessBeforeRemoval() {
...
+		if (!isDisposed() && modelHandle != 0) {
+			// reset computed 'fixed_height' to '-1'
+			OS.g_object_set(handle, OS.fixed_height_mode, false, 0);
+			OS.g_object_set(handle, OS.fixed_height_mode, true, 0);
+
+			// to update height of the existing rows we need to invalidate them in gtk
+			// we do that by invoking gtk_tree_view_row_changed on each of them
+			long iter = OS.g_malloc(GTK.GtkTreeIter_sizeof());
+			if (GTK.gtk_tree_model_get_iter_first(modelHandle, iter)) {
+				int[] value = new int[1];
```
basilevs, line 4385: [minor] unnecessary reallocation

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/TreeItem.java
```diff
@@ -1519,59 +1519,18 @@ public void setImage(int index, Image image) {
...
-		 * Bug 483112: We check to see if the cached value is greater than the size of the pixbufRenderer.
-		 * If it is, then we change the size of the pixbufRenderer accordingly.
-		 * Bug 489025: There is a corner case where the below is triggered when current(Width|Height) is -1,
-		 * which results in icons being set to 0. Fix is to compare only positive sizes.
+		 * Bug 483112: make sure width and height are set for the pixbufRenderer
 		 */
-		if (parent.pixbufWidth > Math.max(currentWidth [0], 0) || parent.pixbufHeight > Math.max(currentHeight [0], 0)) {
-			GTK.gtk_cell_renderer_set_fixed_size (pixbufRenderer, parent.pixbufWidth, parent.pixbufHeight);
-		}
+		GTK.gtk_cell_renderer_set_fixed_size (pixbufRenderer, parent.pixbufWidth, parent.pixbufHeight);
```
basilevs, line 1530: Why do we allow to decrease size now?

basilevs (changes requested):
Reducing the fixed size seems to be an error.

tests/org.eclipse.swt.tests.gtk/JUnit Tests/org/eclipse/swt/tests/gtk/Test_Gtk_Tree_Virtual_setImage.java
```diff
@@ -0,0 +1,185 @@
...
+		var color = shell.getDisplay().getSystemColor(SWT.COLOR_GREEN);
+		var gc = new GC(image);
+		gc.setForeground(color);
+		gc.setBackground(color);
+		gc.fillRectangle(0, 0, IMAGE_SIZE, IMAGE_SIZE);
+		gc.dispose();
+	}
+
+	private void createTree(boolean withCheckbox, int totalRows, int imgRowIdx, Supplier<Image> getImage) {
+		tree = new Tree(shell, SWT.VIRTUAL | SWT.BORDER | (withCheckbox ? SWT.CHECK : 0));
```
basilevs, line 136: [mino] Now, that a potential regression for setting an image of decreasing sizes in different columns is discovered, we may want some tests for multi-column case.

basilevs (changes requested):

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Tree.java
```diff
@@ -4319,6 +4319,80 @@ void checkSetDataInProcessBeforeRemoval() {
 	}
 }
 
+boolean initPixbufSize(Image image) {
```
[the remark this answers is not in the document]
zkxvh: done: https://github.com/eclipse-platform/eclipse.platform.swt/pull/1662/commits/df4c8675ba25dfa5da31a11b683c865817155a80

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Tree.java
```diff
@@ -4319,6 +4319,80 @@ void checkSetDataInProcessBeforeRemoval() {
...
+	if (pixbufSizeSet || image == null) {
```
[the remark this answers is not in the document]
zkxvh: As I understand from the swt code, for any `Tree`:
1. all images in the tree must have the same size
2. that fixed size is the size of the 1st image set for the tree

In the code this is implemented in `TreeItem.setImage(int index, Image image)` [here](https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/bundles/org.eclipse.swt/Eclipse%20SWT/gtk/org/eclipse/swt/widgets/TreeItem.java#L1508-L15189):
```java
	if (image != null) {
		ImageList imageList = parent.imageList;
		if (imageList == null) imageList = parent.imageList = new ImageList();
		int imageIndex = imageList.indexOf(image);
		// When we create a blank image surface gets created with dimensions 0, 0.
        // This call recreates the surface with correct dimensions
		long tempSurface = ImageList.convertSurface(image);
		Cairo.cairo_surface_destroy(tempSurface);
		if (imageIndex == -1) {
			imageIndex = imageList.add(image);
		}
		surface = imageList.getSurface(imageIndex);
		pixbuf = ImageList.createPixbuf(surface);
	}
```
Basically, `ImageList` has `width` and `height` which it initializes from the 1st added image, and then all subsequent added images are scaled to that size.

Here is a snippet that demonstrates how the images in a tree are scaled to the same size:
```java
import org.eclipse.swt.SWT;

public class TryColumnsAndImageSizes {

	private Display display;
	private Shell shell;
	private Composite buttonBox;
	private Tree tree;

	private Image[] images = new Image[3];

	public static void main (String[] args) {
		new TryColumnsAndImageSizes ().run ();
	}

	void run () {
		display = new Display ();
		shell = new Shell (display);
		shell.setLayout (new GridLayout ());

		buttonBox = new Composite (shell, 0);
		buttonBox.setLayout (new FillLayout ());

		addSetImageButton ("r50", 0, 20, display.getSystemColor (SWT.COLOR_RED));
		addSetImageButton ("g70", 1, 70, display.getSystemColor (SWT.COLOR_GREEN));
		addSetImageButton ("b90", 2, 120, display.getSystemColor (SWT.COLOR_BLUE));

		tree = new Tree (shell, SWT.VIRTUAL);
		tree.setLayoutData (new GridData (GridData.FILL_BOTH));
		for (int i = 1; i <= 3; i++) {
			var column = new TreeColumn (tree, SWT.LEFT);
			column.setText ("Column " + i);
			column.setWidth (200);
		}

		tree.addListener (SWT.SetData, e -> {
			TreeItem item = (TreeItem) e.item;
			item.setText (0, "a");
			item.setText (1, "b");
			item.setText (2, "c");
			var idx = tree.indexOf (item);
			for (var i = 0; i < images.length; i++) {
				if (idx == i + 2) {
					item.setImage (i, images[i]);
				}
			}
		});
		tree.setItemCount (10);

		shell.setSize (400, 300);
		shell.open ();

			if (!display.readAndDispatch ()) {
				display.sleep ();
			}
		}
		display.dispose ();
	}

	private void addSetImageButton (String label, int imageIdx, int imageSize, Color imageColor) {
		var button = new Button (buttonBox, SWT.PUSH);
		button.setText (label);
		button.addListener (SWT.Selection, e -> {
			images[imageIdx] = createImage (imageSize, imageColor);
			tree.clearAll (false);
		});
	}

	private Image createImage (int imageSize, Color color) {
		var image = new Image (display, imageSize, imageSize);
		var gc = new GC (image);
		gc.setForeground (color);
		gc.setBackground (display.getSystemColor (SWT.COLOR_WHITE));
		gc.setLineWidth (5);
		gc.fillRectangle (0, 0, imageSize, imageSize);
		gc.drawRectangle (0, 0, imageSize - 1, imageSize - 1);
		gc.dispose ();
		return image;
	}
}
```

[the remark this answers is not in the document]
zkxvh: As a result, the answer to this:
> Can't we monotonically increase the size for every new image?

is "no, because `Tree` in swt is designed to have images of the same size".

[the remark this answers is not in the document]
basilevs: Ok, as long as they are not cropped, we can leave this as is.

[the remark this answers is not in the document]
basilevs: > is "no, because `Tree` in swt is designed to have images of the same size".

It does not matter, as on each new image we could resize all previously added ones as necessary.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Tree.java
zkxvh: Sorry, but I don't understand.  
Unnecessary reallocation of what?  
As I can see, both `iter` and `value` are allocated once.  
Could you clarify it, please.

basilevs: I had read `int[] value = new int[1];` as being a part of loop iteration. I was wrong. Sorry about that.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/TreeItem.java
zkxvh: As I understand, in swt all images in a `Tree` have the same size. ([here](https://github.com/eclipse-platform/eclipse.platform.swt/pull/1662#discussion_r1886392687) is why I think so).  
As a result, all `GtkCellRendererPixbuf`s should have the same fixed width and height.  

Actually in `Tree.resetFixedRowHeight()` I added the code that iterates over all `GtkCellRendererPixbuf`s and set them fixed width and height.  
So maybe, `GTK.gtk_cell_renderer_set_fixed_size()` might even be redundant.  
I left this code here just in case (e.g. maybe there is code somewhere in swt that adds a new column and doesn't call `GTK.gtk_cell_renderer_set_fixed_size()` for it)

tests/org.eclipse.swt.tests.gtk/JUnit Tests/org/eclipse/swt/tests/gtk/Test_Gtk_Tree_Virtual_setImage.java
zkxvh: It seems to me that `Tree` doesn't support different images sizes in swt by design.  
In [this reply](https://github.com/eclipse-platform/eclipse.platform.swt/pull/1662#discussion_r1886392687) I included a snippet that demonstrates what happens when there are multiple columns.  
Please, note that images are scaled to fixed size inside `ImageList`, which is pretty old - that's why I think that `Tree` expects images of the same size by design.

zkxvh: BTW, I can add unit test(s) for multiple columns and different images sizes to `Test_Gtk_Tree_Virtual_setImage` no problem.  
I just don't know what additional cases you want to be checked.  
The current unit tests in `Test_Gtk_Tree_Virtual_setImage` are for the bug with row heights, that I noticed (and fixed in this PR) while I've been fixing crash in #678.

basilevs: The scenario is:
- for three sequences of images three in length each:
  - one monotonically increasing in size
  - one monotonically decreasing
  - and one non-monotonic
- test that behavior (heights, widths, exceptions, ideally cropping but that's probably too hard) is unchanged when:
  - adding images to different columns 
  - adding to same column
  - adding to a tree without columns

zkxvh:
@basilevs  

> If you use `asyncExec()` already, why do all the other changes? The problem was never the computation of heights, it was resource management. And resources are managed just fine as long as you defer their disposal. Therefore, my fix in #1612 is less intrusive and achieves the same results at less cost (admittedly I have not tested if selection is lost, but it might).
> 
> UPDATE: Oh, I see. There are less resources disposed this way.

Yes, the method with `fixed_height_mode` property doesn't delete and recreate existing renderers like `Tree.createRenderers()` did.

Also I tried to understand what the old code did (when it did not crash the application).  
[Here](https://github.com/eclipse-platform/eclipse.platform.swt/issues/678#issuecomment-2541779349) is more details, but in short:
- a `Tree` in swt has fixed size for all images
- the fixed image size is initialized from the 1st image set to a `Tree`
- when fixed image size is initialized for a `Tree` with `SWT.VIRTUAL`, then the (fixed) row heights should be updated to fit it.

`fixed_height_mode` + `gtk_tree_view_row_changed` is the least intrusive way to do the above that I was able to find in gtk.

---

**Note:** changing `fixed_height_mode` is fine here because `Tree.initPixbufSize()` is called only once per tree, which means the fixed row height is changed only once.  
But if fixed height is going to be updated multiple times, then it seems that you'll have to switch to `gtk_tree_view_set_model()`.  
It seems like gtk assumes that `fixed_height_mode` property should be set once - before any row is added to a tree.  
Why I think so:
- in [gtk_tree_view_set_fixed_height_mode](https://gitlab.gnome.org/GNOME/gtk/-/blob/30f02d8981c4246af854a570d5718f440367d701/gtk/gtktreeview.c#L8527) the signal handler for `"notify::sizing"` is added, but never removed. I.e. every time we set `fixed_height_mode` to `true` one more handler is attached to `sizing` property
- the fact that I had to use `gtk_tree_view_row_changed` to update the heights of the existing rows (had gtk supported changing `fixed_height_mode` it would have updated the height of the existing rows automatically).  

`gtk_tree_view_set_model()` can be used to update row heights any number of times, but it resets current selection, focus and scroll position of the tree.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Tree.java
zkxvh: no problem

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Tree.java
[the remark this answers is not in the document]
zkxvh: OK.  
Apparently, this is another issue.  
Just be careful, if you are going to increase fixed row size multiple times, then you'll probably have to replace `fixed_height_mode + gtk_tree_view_row_changed` to `gtk_tree_view_set_model()`. Here is more [details](https://github.com/eclipse-platform/eclipse.platform.swt/pull/1662#issuecomment-2545220394).

tests/org.eclipse.swt.tests.gtk/JUnit Tests/org/eclipse/swt/tests/gtk/Test_Gtk_Tree_Virtual_setImage.java
zkxvh: OK.  
As I understand, you are working on some task related to image cropping, multiple image sizes, and multiple columns.  
I suppose including those unit tests in the commit for this task I would be more appropriate.

basilevs: No, I literally tried to fix the very same problem - crash in VIRTUAL Tree. While I succeeded, your solution might be a bit better.

basilevs: No, I [fixed](https://github.com/eclipse-platform/eclipse.platform.swt/pull/1612) the very same problem - crash in VIRTUAL Tree, but your solution might be better.

zkxvh:
Converted to draft because it seems like this bug should be should be fixed in another place in the code.  

After fixing this bug by changing `TreeItem.setImage(...)` I get the same process crash in Eclipse RCP in a code that uses  `Tree` but with slightly different error message:
```
Bail out! Gtk:ERROR:../gtk/gtktreestore.c:597:gtk_tree_store_get_path: assertion failed: (G_NODE (iter->user_data)->parent != NULL)
**
Gtk:ERROR:../gtk/gtktreestore.c:597:gtk_tree_store_get_path: assertion failed: (G_NODE (iter->user_data)->parent != NULL)
```
Stacktrace is this:
```
	org.eclipse.swt.widgets.TreeItem.getParentItem(TreeItem.java:866)
	at org.eclipse.jface.viewers.TreeViewer.getParentItem(TreeViewer.java:228)
	at org.eclipse.jface.viewers.AbstractTreeViewer.getTreePathFromItem(AbstractTreeViewer.java:3046)
	at org.eclipse.jface.viewers.AbstractTreeViewer.internalFindItem(AbstractTreeViewer.java:211)
	at org.eclipse.jface.viewers.AbstractTreeViewer.internalFindItems(AbstractTreeViewer.java:188)
	at org.eclipse.jface.viewers.TreeViewer.replace(TreeViewer.java:435)
	at TestLazyContentProvider.updateElement(TestLazyContentProvider.java:100)
	at org.eclipse.jface.viewers.TreeViewer.virtualLazyUpdateWidget(TreeViewer.java:1001)
	at org.eclipse.jface.viewers.TreeViewer.lambda$1(TreeViewer.java:260)
	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:89)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:5855)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1529)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1555)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1538)
	at org.eclipse.swt.widgets.Tree.checkData(Tree.java:372)
	at org.eclipse.swt.widgets.Tree.cellDataProc(Tree.java:304)
	at org.eclipse.swt.widgets.Display.cellDataProc(Display.java:995)
	at org.eclipse.swt.internal.gtk.GTK.gtk_tree_store_clear(Native Method)
	at org.eclipse.swt.widgets.Tree.removeAll(Tree.java:2997)
	at org.eclipse.jface.viewers.TreeViewer.removeAll(TreeViewer.java:289)
	at org.eclipse.jface.viewers.AbstractTreeViewer.lambda$1(AbstractTreeViewer.java:1593)
	at org.eclipse.jface.viewers.StructuredViewer.preservingSelection(StructuredViewer.java:1392)
	at org.eclipse.jface.viewers.TreeViewer.preservingSelection(TreeViewer.java:367)
	at org.eclipse.jface.viewers.StructuredViewer.preservingSelection(StructuredViewer.java:1353)
	at org.eclipse.jface.viewers.AbstractTreeViewer.inputChanged(AbstractTreeViewer.java:1589)
	at org.eclipse.jface.viewers.ContentViewer.setInput(ContentViewer.java:282)
	at org.eclipse.jface.viewers.StructuredViewer.setInput(StructuredViewer.java:1636)
```
This is a very similar bug: again we get:
```
Display.cellDataProc(...)
  ↳ Tree.checkData(...)
    ↳ sendEvent(SWT.SetData, ...)
      ↳ a callback for SWT.SetData is invoked, it adds/removes columns/rows/renderers/other tree components
  ↳ the code in Display.cellDataProc(...) continues execution and crashes when it accesses removed/modified tree components
```

The problem again is that a `SWT.SetData` callback changes the structure of the tree, while `Display.cellDataProc(...)` (where the callback is called from) is a [cell data function in GTK terms](https://docs.gtk.org/gtk3/treeview-tutorial.html#cell-data-functions) which should only set renderers' properties like text, image, etc. and should not change tree structure.

Since callbacks for `SWT.SetData` can contain any code, I think the the better fix would be to change `Display.cellDataProc(...)` to postpone sending `SWT.SetData` until the method finishes.  
This would also fix all other similar bugs that crash the application because of `Tree`.

PR #1496, opened 2024-09-26
Refactor Image for storing ImageMetadata

amartya4256:
This PR contributes to the refactoring of `Image` class for **win32**. 

The goal is to store more than just the `handle` of the image for each zoom level. To do that, a new private class is created (it's called `ImageHandle`) and an image contains one instance of this new class for each zoom level. The class contains the following fields:
- `handle` 
- `height`
- `width`

contributes to #62 and #127

PR #1672, opened 2024-12-17
[win32] fix invalid icon bounds

akoch-yatta:
This commit fixes an issue in the win32 implementation that returned invalid bounds for an icon in some scenarios. This was introduced with #1496 

## How to reproduce

- Install the XML editors from WTP 
- Update site: http://download.eclipse.org/webtools/repository/latest/

image

- Open any XML file with the **XML Editor**

image

It is not 100% reproducible, but should lead to an exception most of the times without this PR


fedejeanne:
Test failures are unrelated #1676 , #1564

PR #1673, opened 2024-12-17
Add exception to setImageMetadataForHandle

amartya4256:
This contribution adds a check for imageMetadata for handle in case there already exists an entry in the zoomLeveltoImageHandle map for a zoom level and it must throw an exception in that case.

contributes to #62 and #127


fedejeanne:
Failing tests are unrelated #1564

PR #1674, opened 2024-12-17
Use InputStream.readAllBytes() in ImageLoader for GTK

HannesWell:


akurtakov:
Thanks for this change! Test failures on Jenkins are the long standing ones on GTK. Merging.

PR #1675, opened 2024-12-18
Enable fail on javadoc in SWT repo

akurtakov:
Bundle org.eclipse.swt itself is skipped as jni generation relies on what "looks like" javadoc tags.
Identified via https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2675


akurtakov:
Test failures are long standing ones on Gtk.

laeubi:
> Bundle org.eclipse.swt itself is skipped as jni generation relies on what "looks like" javadoc tags

Just in case, one can define custom tags like here:

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/a9f14779d474a03ca9bf232ae8919e0ed10c8ed8/eclipse-platform-parent/pom.xml#L585

not sure if it helps here, but might be something to consider.

Issue #1676, opened 2024-12-18
Tests fail because creating the Edge browser for the tests takes too long


fedejeanne:
**Describe the bug**
Creating the Edge browser for the tests in `Test_org_eclipse_swt_browser_Browser` takes too long.

```java
creating browser took too long: 23811ms
java.lang.AssertionError: creating browser took too long: 23811ms
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.assertTrue(Assert.java:42)
	at org.eclipse.swt.tests.junit.Test_org_eclipse_swt_browser_Browser.createBrowser(Test_org_eclipse_swt_browser_Browser.java:311)
	at org.eclipse.swt.tests.junit.Test_org_eclipse_swt_browser_Browser.setUp(Test_org_eclipse_swt_browser_Browser.java:184)
...
```

**To Reproduce**
It fails intermittently in GH Actions _e.g._ in this one: https://github.com/eclipse-platform/eclipse.platform.swt/actions/runs/12374507303/job/34537197042?pr=1672

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 

**Version since**
Ever since #672 has been merged and both Edge and IE are tested

PR #1680, opened 2024-12-18
Destroying unnecessary zoom level handles

ShahzaibIbrahim:
Add display change event triggered by display zoom changes, regardless of window presence on the monitor. 

When the zoom level is changed for a monitor where the window is not present, we still need to trigger the event to clear resources for the old zoom level on the non-active monitor.  

For example, if there are two monitors with zoom levels 100% and 200%, and the window is on the 200% zoom monitor, changing the zoom level of the 100% monitor to 150% should destroy the handles for the old 100% zoom level, as they are no longer needed.


fedejeanne (approved):

fedejeanne:

PR #1681, opened 2024-12-18
Disable test_TabTraversalOutOfBrowser

### LINKED ISSUE #1656: org.eclipse.swt.tests.junit.Test_org_eclipse_swt_browser_Browser.test_TabTraversalOutOfBrowser fails intermittently, both on IE and on Edge
The test fails locally and it can fail with both browsers on Windows.

I ran the test using the parameterized version of the test class `org.eclipse.swt.tests.junit.Test_org_eclipse_swt_browser_Browser` that can be found in #672 (I disabled all other tests in the class to speed up the testing), I swapped the parameters around and also enabled/disabled them, just for fun, and this is what I saw:

image
---

image
---

image
---

image
---

image
---

**Note: `[browser flags: 0]` means _Edge_ and `[browser flags: 524,288]` means _IE_**

The only combination that I couldn't get was: First Edge then IE and **both** failing... but I can't say that one won't fail either.

sratz:
Fixes #1656.


fedejeanne:

PR #1685, opened 2024-12-19
Add Null check before assigning coordinateMapper

### LINKED ISSUE #1686: MultiZoomCoordinateSystemMapper Overwritten in Device:init
**Describe the bug**
In the current implementation, the Display seems to never use MultiZoomCoordinateSystemMapper for its coordinateSystemMapper. This mapper is used when rescalingAtRuntime is active but eventually Device:init is called at some point which overwrites the coordinateSystemMapper to a SingleZoomCoordinateSystemMapper. In case of a multi zoom setup, this makes the UI to behave weirdly when you drag and drop a tab across the screens.

**To Reproduce**
- Start the SDK product.
- Drag an editor tab to another monitor with a different DPI
- The Editor will not open where the preview is pointing to but rather a different location (slightly scaled)
(My config was: Left monitor - primary - 175%, Right monitor - 100 %; Dragging from right to left)

**Expected behavior**
The top left corner of the dropped editor should be the same as where the cursor dropped it.

**Environment:**
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 

Introduced by #1630

amartya4256:
This contribution adds a null check before assigning the coordinate mapper in the Device::init since it might already be assigned in coordinateMapper.

As of the current implementation, the MultiZoomCoordinateMapper is assigned to coordinateMapper (if rescalingAtRuntime is active) in setRescalingAtRuntime and then it is overwritten in Device:init and always reset to SingleZoomCoordinateSystemMapper which is not the intended behaviour.

Fixes #1686 
contributes to #62 and #127

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
@@ -2772,7 +2772,9 @@ protected void init () {
 	// Field initialization happens after super constructor
 	controlByHandle = new HashMap<>();
 	this.synchronizer = new Synchronizer (this);
-	this.coordinateSystemMapper = new SingleZoomCoordinateSystemMapper();
+	if (this.coordinateSystemMapper == null) {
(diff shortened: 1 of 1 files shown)


fedejeanne:

PR #1688, opened 2024-12-20
Add check for provider

fedejeanne:
See https://github.com/eclipse-platform/eclipse.platform.swt/pull/1667#issuecomment-2556458475

Issue #1653, opened 2024-12-12
Github SWT Linux verification builds fail on ubuntu-latest


jukzi:
Each and every recent PR after https://github.com/eclipse-platform/eclipse.platform.swt/pull/1627 sees
![image](https://github.com/user-attachments/assets/de5ec44f-31cb-4927-a714-ae5e75c028c1)

```
Error:  Failed to execute goal org.apache.maven.plugins:maven-antrun-plugin:3.1.0:run (build-native-binaries) on project org.eclipse.swt.gtk.linux.x86_64: An Ant BuildException has occured: exec returned: 2
Error:  around Ant part ...<exec failonerror="true" dir="${build_dir}" executable="sh">... @ 24:73 in /home/runner/work/eclipse.platform.swt/eclipse.platform.swt/binaries/org.eclipse.swt.gtk.linux.x86_64/target/antrun/build-main.xml
Error:  -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.apache.maven.plugins:maven-antrun-plugin:3.1.0:run (build-native-binaries) on project org.eclipse.swt.gtk.linux.x86_64: An Ant BuildException has occured: exec returned: 2
around Ant part ...<exec failonerror="true" dir="${build_dir}" executable="sh">... @ 24:73 in /home/runner/work/eclipse.platform.swt/eclipse.platform.swt/binaries/org.eclipse.swt.gtk.linux.x86_64/target/antrun/build-main.xml
    at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute2 (MojoExecutor.java:333)
    at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute (MojoExecutor.java:316)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:212)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:174)
    at org.apache.maven.lifecycle.internal.MojoExecutor.access$000 (MojoExecutor.java:75)
    at org.apache.maven.lifecycle.internal.MojoExecutor$1.run (MojoExecutor.java:162)
    at org.apache.maven.plugin.DefaultMojosExecutionStrategy.execute (DefaultMojosExecutionStrategy.java:39)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:159)
[...]

PR #1690, opened 2024-12-20
Throw invalid argument exception if the type of the provider is not the one expected.

fedejeanne:
Addresses https://github.com/eclipse-platform/eclipse.platform.swt/pull/1688#discussion_r1893890540


fedejeanne:
Failures are unrelated: #1653 , #1676

PR #1692, opened 2024-12-22
Disable deprecation warning in GTK at github validataion

laeubi:
To unblock
- https://github.com/eclipse-platform/eclipse.platform.swt/issues/1653


laeubi:
All relevant checks have completed successful.

PR #1695, opened 2024-12-23
Using native zoom to compute size for width and height

ShahzaibIbrahim:
When autoScale is int200 the size compute for width and height is not scaled for the following buttons correctly during DPI change. Using native zoom instead of corrected zoom fixes the issue.

**Scenario**
Moving from Secondary (200) to Primary (150) Project -> Properties -> Save actions. (autoScale = Int200)

**Before Fix**
image

**After Fix**
image


HeikoKlare (commented):
Can you please explain why this is a reasonable fix?
If I am not mistaken, all size calculations/computations are currently based on scaling via the `deviceZoom`, not via the `nativeZoom`. So this change seems to introduce an inconsistency in how scaling of control sizes is performed.

HeikoKlare:
Superceded by:
- #1721

PR #1699, opened 2025-01-02
Refactored CoordinateSystemMapper and added tests

amartya4256:


HeikoKlare (commented):
Thank you for adding tests for the mapping calculations in `Display` and refactoring to the code accordingly.
In order to have the refactoring integrated as soon as possible (i.e., to have this PR finalized and merged), I have two questions @amartya4256:
- Since the `MultiZoomCoordinateSystemMapper` is still under development (and as long as that's the case some tests will probably always be failing), can we deactivate those test for now?
- Can we parameterize (or copy and adapt) the tests to also be applied to the `SingleZoomCoordinateSystemMapper`? Having regression tests for the existing HiDPI behavior and the mapping of coordinates in it would be great as well.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/SingleZoomCoordinateMapper.java
```diff
@@ -0,0 +1,91 @@
+package org.eclipse.swt.widgets;
+
+import org.eclipse.swt.*;
+import org.eclipse.swt.graphics.*;
+import org.eclipse.swt.internal.*;
+
+class SingleZoomCoordinateSystemMapper extends CoordinateSystemMapper {
```
HeikoKlare, line 7: Class has a different name than the file

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
```diff
@@ -5369,7 +5360,7 @@ public boolean setRescalingAtRuntime(boolean activate) {
 	int desiredApiAwareness = activate ? OS.DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2 : OS.DPI_AWARENESS_CONTEXT_SYSTEM_AWARE;
 	if (setDPIAwareness(desiredApiAwareness)) {
 		rescalingAtRuntime = activate;
-		coordinateSystemMapper = activate ? new MultiZoomCoordinateSystemMapper() : new SingleZoomCoordinateSystemMapper();
+		coordinateSystemMapper = activate ? new MultiZoomCoordinateSystemMapper(this, () -> this.getMonitors()) : new SingleZoomCoordinateSystemMapper(this);
```
HeikoKlare, line 5363:
```suggestion
		coordinateSystemMapper = activate ? new MultiZoomCoordinateSystemMapper(this, this::getMonitors) : new SingleZoomCoordinateSystemMapper(this);
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/CoordinateSystemMapper.java
```diff
@@ -0,0 +1,35 @@
...
+import org.eclipse.swt.graphics.*;
+
+abstract class CoordinateSystemMapper {
+
+	Display display;
+
+	CoordinateSystemMapper(Display display) {
+		super();
+		this.display = display;
+	}
```
HeikoKlare, line 12: Has this some benefit over making the class an interface and have this stored in the individual implementations?
Anyway, the field should be declared final and, in my opinion, as private with an according accessor.

bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/CoordinateSystemMapperTests.java
```diff
@@ -0,0 +1,138 @@
...
+
+	@Test
+	void translatePointInNoMonitorBackAndForthShouldBeTheSame() {
+		Point pt = new Point(5000, -400);
+		Point px = multiZoomCoordinateSystemMapper.translateToDisplayCoordinates(pt, 0);
+		assertEquals(pt, multiZoomCoordinateSystemMapper.translateFromDisplayCoordinates(px, 0));
+	}
+
+	@Test
+	@Disabled
```
HeikoKlare, line 60: Would be good to have a short reason text added to the annotations.

HeikoKlare (changes requested):
Thank you for the adaptations. Would it be possible to reuse / parameterize the tests for SingleZoomCoordinateSystemMapper (see https://github.com/eclipse-platform/eclipse.platform.swt/pull/1699#pullrequestreview-2537231547)?

I have some further comments and please add a copyright header to the three new classes.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
amartya4256: Yes makes sense

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/CoordinateSystemMapper.java
amartya4256: Can be either way. I just converted it to an abstract class because I wanted to keep the possibilty of any common implementations and field declarations open. As of now, only display is something common in both. What would you recommend?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/SingleZoomCoordinateMapper.java
amartya4256: Fixed

bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/CoordinateSystemMapperTests.java
amartya4256: Done

amartya4256:
> Thank you for adding tests for the mapping calculations in `Display` and refactoring to the code accordingly. In order to have the refactoring integrated as soon as possible (i.e., to have this PR finalized and merged), I have two questions @amartya4256:
> 
> * Can we parameterize (or copy and adapt) the tests to also be applied to the `SingleZoomCoordinateSystemMapper`? Having regression tests for the existing HiDPI behavior and the mapping of coordinates in it would be great as well.


I can adjust the tests to support that. The monitors need to be created differently because right now the monitors are created with the offset values for multi monitor. But other than that, it should be fine except for some tests specific to Gap which will not be valid for it.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/CoordinateSystemMapper.java
HeikoKlare: I would usually go for an interface if an abstract class only provides very few benefits. In this case, the abstract class only wraps a single value, which can easily be replicated in the concrete classes without introducing high redundancy. But I would say it is rather a matter of taste here, so no strong preference for one over the other. The only thing I would strongly recommend is to make the field final.

HeikoKlare:
> I can adjust the tests to support that. The monitors need to be created differently because right now the monitors are created with the offset values for multi monitor. But other than that, it should be fine except for some tests specific to Gap which will not be valid for it.

If that can be added really quickly, it would be a great win for regression testing. But if not, we should just leave it as is, since changes to the existing coordinate system mapper will be rather seldom (and we could also extend the test cases once changes need to be performed).

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/CoordinateSystemMapper.java
amartya4256: Alright I have updated the classes. Since before the refactoring it was implemented as an interface, I have decided to make it an interface as well. And added a final to displays.

HeikoKlare (approved):
Thank you for the latest update, including the extension of the tests to the single-zoom coordinate system mapper.

It looks fine to me now. I have just rebased on master and applied some formatting as indentations in the test class were broken.

Issue #1670, opened 2024-12-17
[Win][Edge]Test_org_eclipse_swt_browser_Browser.test_setUrl_local fails for Edge


fedejeanne:
**Describe the bug**
The test `Test_org_eclipse_swt_browser_Browser.test_setUrl_local` fails for the **Edge** browser in the I-BUILDs and for both IE and Edge when run locally. 

### Stack trace (I-BUILD)
```java
Multiple Failures (2 failures) org.junit.ComparisonFailure: Test log: PLUGIN_PATH: C:\Users\genie.releng\workspace\AutomatedTests\ep435I-unit-win32-x86_64-java21/workarea/I20241216-1800/eclipse-testing/test-eclipse/eclipse/plugins/org.eclipse.swt.tests_3.107.700.v20241216-1631 URL: C:\Users\genie.releng\workspace\AutomatedTests\ep435I-unit-win32-x86_64-java21/workarea/I20241216-1800/eclipse-testing/test-eclipse/eclipse/plugins/org.eclipse.swt.tests_3.107.700.v20241216-1631/data/testWebsiteWithTitle.html TitleListener fired expected:<[Website Title]> but was:<[c]> java.lang.RuntimeException: AwtScreenshot VM finished with exit code 1: Exception in thread "main" java.awt.AWTError: no screen devices at java.desktop/sun.awt.Win32GraphicsEnvironment.getDefaultScreenDevice(Win32GraphicsEnvironment.java:101) at java.desktop/java.awt.Robot.<init>(Robot.java:135) at org.eclipse.test.AwtScreenshot.main(AwtScreenshot.java:43)

org.opentest4j.MultipleFailuresError: Multiple Failures (2 failures)
org.junit.ComparisonFailure:
Test log:
PLUGIN_PATH: C:\Users\genie.releng\workspace\AutomatedTests\ep435I-unit-win32-x86_64-java21/workarea/I20241216-1800/eclipse-testing/test-eclipse/eclipse/plugins/org.eclipse.swt.tests_3.107.700.v20241216-1631
URL: C:\Users\genie.releng\workspace\AutomatedTests\ep435I-unit-win32-x86_64-java21/workarea/I20241216-1800/eclipse-testing/test-eclipse/eclipse/plugins/org.eclipse.swt.tests_3.107.700.v20241216-1631/data/testWebsiteWithTitle.html
TitleListener fired expected:<[Website Title]> but was:<[c]>
[...]

PR #1701, opened 2025-01-02
Edge: correct URL to local resource in tests #1670

HeikoKlare:
The URL used for the `test_setUrl_Local()` browser test case currently only consists of a file path when provided via the `PLUGIN_PATH` environment variable, e.g., during I-Build test execution. Edge does, however, require a valid URL with an according protocol prefix. Currently, the drive prefix of the file path is erroneously treated as the protocol.

This change properly converts the path to a URL.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/1670

Note that this change will not have any effect in the PR validation, but only in the I-Build test execution (or if locally executed with the `PLUGIN_PATH` variable specified. I locally executed the tests with that variable specified to validate that it works as expected.

### DIFF
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
@@ -762,7 +762,7 @@ private String getValidUrl() {
 	testLogAppend("PLUGIN_PATH: " + pluginPath);
 	// When test is run via Ant, URL needs to be acquired differently. In that case the PLUGIN_PATH property is set and used.
 	if (pluginPath != null) {
-		return pluginPath + "/data/testWebsiteWithTitle.html";
+		return Path.of(pluginPath, "data/testWebsiteWithTitle.html").toUri().toString();
 	} else {
 		// used when ran from Eclipse gui.
 		return Test_org_eclipse_swt_browser_Browser.class.getClassLoader().getResource("testWebsiteWithTitle.html").toString();


tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
```diff
@@ -762,7 +762,7 @@ private String getValidUrl() {
 	testLogAppend("PLUGIN_PATH: " + pluginPath);
 	// When test is run via Ant, URL needs to be acquired differently. In that case the PLUGIN_PATH property is set and used.
 	if (pluginPath != null) {
-		return pluginPath + "/data/testWebsiteWithTitle.html";
+		return "file://" + pluginPath + "/data/testWebsiteWithTitle.html";
```
merks, line 765: This feels like a hack to me.  If this truly a Path, one should properly convert it to a URI, e.g., `Path.of().toUri().toString()` so that `\` is properly handled and even things like ` ` are properly encoded.

HeikoKlare: Indeed, thank you! I improved that with 5423f29. Local execution works fine with that change.

merks (approved):
Yes, that looks cleaner.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
```diff
@@ -762,7 +762,7 @@ private String getValidUrl() {
 	testLogAppend("PLUGIN_PATH: " + pluginPath);
 	// When test is run via Ant, URL needs to be acquired differently. In that case the PLUGIN_PATH property is set and used.
 	if (pluginPath != null) {
-		return pluginPath + "/data/testWebsiteWithTitle.html";
+		return Path.of(pluginPath + "/data/testWebsiteWithTitle.html").toUri().toString();
```
laeubi, line 765:
```suggestion
		return Path.of(pluginPath , "data/testWebsiteWithTitle.html").toUri().toString();
```

HeikoKlare: Done: e25fc56

HeikoKlare:
I'll check tomorrow's I-Build to see whether the test failure is actually fixed there.

PR #1702, opened 2025-01-02
Switch to the Cleaner API

### LINKED ISSUE #1465: Avoid overriding finalize()
**Is your feature request related to a problem? Please describe.**

The current implementation to make [Resource.setNonDisposehandler(Consumer)](https://help.eclipse.org/latest/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Freference%2Fapi%2Forg%2Feclipse%2Fswt%2Fgraphics%2FResource.html) work, uses `finalize()` under the hood:

https://github.com/eclipse-platform/eclipse.platform.swt/blob/4d6832927eb8874dccb0c59eac5a761ce9ea025f/bundles/org.eclipse.swt/Eclipse%20SWT/common/org/eclipse/swt/graphics/Resource.java#L74-L81

This method is deprecated for removal since Java 18 as described in [JEP 421](https://openjdk.org/jeps/421).

**Describe the solution you'd like**

We should rewrite this functionality using [PhantomReferences](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/ref/PhantomReference.html) and a daemon thread polling from the shared [ReferenceQueue](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/ref/ReferenceQueue.html).

**Describe alternatives you've considered**

The cleaner API might also work:

https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/ref/Cleaner.html

otbutz:
Fixes #1465

* replaced the current implementation  that relies on `finalize()` by using a [Cleaner](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/ref/Cleaner.html) instead
* the `ResourceTracker` no longer holds a reference to the resource, as the Cleaner API mandates that:    
    > The cleaning action is invoked only after the associated object becomes phantom reachable, so it is important that the object implementing the cleaning action does not hold references to the object.
*  a flag is now used to detect whether the `nonDisposedReporter` should be invoked.
* the tracker is registered during `init()` to avoid false-negatives for resources throwing an exception in their constructor. e.g `new Cursor(display, -100)`

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Resource.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Resource.java
@@ -13,7 +13,11 @@
  *******************************************************************************/
 package org.eclipse.swt.graphics;
 
+import java.lang.ref.*;
+import java.lang.ref.Cleaner.*;
 import java.util.*;
+import java.util.concurrent.*;
+import java.util.concurrent.atomic.*;
 import java.util.function.*;
 
 import org.eclipse.swt.*;
@@ -44,41 +48,35 @@ import org.eclipse.swt.*;
 public abstract class Resource {
 
 	/**
-	 * Used to track not disposed SWT resource. A separate class allows
-	 * not to have the {@link #finalize} when tracking is disabled, avoiding
-	 * possible performance issues in GC.
+	 * Used to track not disposed SWT resource.
 	 */
-	private static class ResourceTracker {
+	private static final class ResourceTracker implements Runnable {
 		/**
-		 * Resource that is tracked here
+		 * Invokes {@link #run()} once the resource is eligible for GC
 		 */
-		private Resource resource;
+		private static final Cleaner cleaner = Cleaner.create(new ResourceTrackerThreadFactory());
 
 		/**
 		 * Recorded at Resource creation if {@link #setNonDisposeHandler} was
 		 * enabled, used to track resource disposal
 		 */
-		private Error allocationStack;
+		private final Error allocationStack;
 
 		/**
-		 * Allows to ignore specific Resources even if they are not disposed
-		 * properly, used for example for Fonts that SWT doesn't own.
+		 * Controls whether the {@link Resource#nonDisposedReporter} should be notified
 		 */
-		boolean ignoreMe;
+		private final AtomicBoolean reporting = new AtomicBoolean(false);
 
-		ResourceTracker(Resource resource, Error allocationStack) {
-			this.resource = resource;
+		ResourceTracker(Error allocationStack) {
 			this.allocationStack = allocationStack;
 		}
 
 		@Override
-		protected void finalize() {
-			if (ignoreMe) return;
+		public void run() {
+			if (!reporting.get()) return;
 			if (nonDisposedReporter == null) return;
 
-			// If the Resource is GC'ed before it was disposed, this is a leak.
-			if (!resource.isDisposed())
-				nonDisposedReporter.accept(allocationStack);
+			nonDisposedReporter.accept(allocationStack);
 		}
 	}
 
@@ -96,6 +94,11 @@ public abstract class Resource {
 	 * Used to track not disposed SWT resource
 	 */
 	private ResourceTracker tracker;
+
+	/**
+	 * Represents the {@link #tracker} registered as a cleaning action via the {@link ResourceTracker#cleaner}
+	 */
+	private Cleanable cleanable;
 
 	static {
 		boolean trackingEnabled = Boolean.getBoolean("org.eclipse.swt.graphics.Resource.reportNonDisposed"); //$NON-NLS-1$
@@ -142,6 +145,8 @@ void destroyHandlesExcept(Set<Integer> zoomLevels) {
  * This method does nothing if the resource is already disposed.
  */
 public void dispose() {
+	if (tracker != null) tracker.reporting.set(false);
+	if (cleanable != null) cleanable.clean();
 	if (device == null) return;
 	if (device.isDisposed()) return;
 	destroy();
@@ -164,30 +169,29 @@ public Device getDevice() {
 }
 
 void ignoreNonDisposed() {
-	if (tracker != null) {
-		tracker.ignoreMe = true;
-	}
+	if (tracker != null) tracker.reporting.set(false);
+	if (cleanable != null) cleanable.clean();
 }
 
 void init() {
 	if (device.tracking) device.new_Object(this);
+	if (tracker != null && tracker.reporting.compareAndSet(false, true)) {
+		cleanable = ResourceTracker.cleaner.register(this, tracker);
+	}
 }
 
 void initNonDisposeTracking() {
 	// Color doesn't really have any resource to be leaked, ignore.
 	if (this instanceof Color) return;
 
-	// Avoid performance costs of having '.finalize()' when not tracking.
+	// Avoid performance costs of gathering the current stack trace when not tracking.
 	if (nonDisposedReporter == null) return;
 
 	// Capture a stack trace to help investigating the leak
 	Error error = new Error("SWT Resource was not properly disposed"); //$NON-NLS-1$
 
-	// Allocate a helper class with '.finalize()' in it, it will do the actual
-	// work of detecting and reporting errors. This works because Resource
-	// holds the only reference to 'ResourceTracker' and therefore the tracker
-	// is only GC'ed when Resource itself is ready to be GC'ed.
-	tracker = new ResourceTracker(this, error);
+	// Create the tracker that will later be registered as a cleaning action for this resource.
+	tracker = new ResourceTracker(error);
 }
 
 /**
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Resource.java
```diff
@@ -87,6 +86,11 @@ protected void finalize() {
 	 */
 	Device device;
 
+	/**
+	 * Invokes {@link ResourceTracker#run()} once the resource is eligible for GC
+	 */
+	private static final Cleaner cleaner = Cleaner.create();
```
otbutz, line 92: We can pass our own ThreadFactory to have a nicer thread name, but IMHO this adds too much complexity for what it's worth.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Resource.java
```diff
@@ -165,29 +170,28 @@ public Device getDevice() {
...
 	Error error = new Error("SWT Resource was not properly disposed"); //$NON-NLS-1$
 
-	// Allocate a helper class with '.finalize()' in it, it will do the actual
-	// work of detecting and reporting errors. This works because Resource
-	// holds the only reference to 'ResourceTracker' and therefore the tracker
-	// is only GC'ed when Resource itself is ready to be GC'ed.
-	tracker = new ResourceTracker(this, error);
+	// Create and register the tracker to run as a cleaning action for this resource.
+	tracker = new ResourceTracker(error);
+	cleaner.register(this, tracker);
```
otbutz, line 194: `init()` looked like the more appropriate place for registering the tracker, but apparently it can be called multiple times per resource object. e.g `Image.refreshImageForZoom()`

iloveeclipse:
@SyntevoAlex : are you interested in a review of this PR?

SyntevoAlex:
Please proceed without me.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Resource.java
```diff
@@ -142,6 +138,7 @@ void destroyHandlesExcept(Set<Integer> zoomLevels) {
  * This method does nothing if the resource is already disposed.
  */
 public void dispose() {
+	if (tracker != null) tracker.reporting.set(false);
```
jonahgraham, line 141: And manually run the clean operation here so that it is cleared out of the cleaner list immediately

```suggestion
	if (tracker != null) tracker.reporting.set(false);
	if (cleanable != null) cleanable.clean(); // maybe after isDisposed check and maybe assign it null too?
```

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Resource.java
jonahgraham: FWIW it is really nice to have as many threads named when tracking down some problems. A while ago someone updated worker threads to have their name be that of the org.eclipse.core.runtime.jobs.Job running and that is super useful. 

I think you need to do this to give it a name:

```suggestion
	private static final Cleaner cleaner = Cleaner.create((r) -> new Thread(r, "SWTResourceTracker")); //$NON-NLS-1$
```

But the thread created by Cleaner.create() has a bunch of special properties (its an InnocuousThread) and it has higher than normal priority too.

Therefore, while I would love a name on it too, the default name of Cleaner-X (X is incrementing number) is probably great.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Resource.java
```diff
@@ -165,29 +162,29 @@ public Device getDevice() {
...
 	if (tracker != null) {
-		tracker.ignoreMe = true;
+		tracker.reporting.set(false);
 	}
 }
 
 void init() {
 	if (device.tracking) device.new_Object(this);
+	if (tracker != null && tracker.reporting.compareAndSet(false, true)) {
+		cleaner.register(this, tracker);
```
jonahgraham, line 172: Save the return value here to new field cleanable.

```suggestion
		cleanable = cleaner.register(this, tracker);
```

jonahgraham (commented):
I think it is great to see the removal of finalize. The not disposed reporting has been really useful. Having it be ready for the future when finalize is removed is the right step forward. Thanks for taking this on.

I have added some small suggestions/commentary.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Resource.java
otbutz: I'll try to mimic the built-in factory. It's a little sad that the API doesn't provide an easy way to use custom thread names without reinventing the wheel :frowning_face:

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Resource.java
```diff
@@ -44,41 +46,30 @@
...
-	 * possible performance issues in GC.
+	 * Used to track not disposed SWT resource.
 	 */
-	private static class ResourceTracker {
-		/**
-		 * Resource that is tracked here
-		 */
-		private Resource resource;
-
+	private static class ResourceTracker implements Runnable {
```
laeubi, line 51:
```suggestion
	private static final class ResourceTracker implements Runnable {
```

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Resource.java
laeubi: > Therefore, while I would love a name on it too, the default name of Cleaner-X (X is incrementing number) is probably great.

I also agree that in this case the default name might be enough.

That said, the following hack might work to set the thread name indirectly without a factory:

```
        cleaner.register(new Object(), new Runnable() {
            
            @Override
            public void run() {
                Thread.currentThread().setName("SWT-Cleaner-Thread");
            }
        }).clean();
```

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Resource.java
otbutz: Isn't the Runnable invoked directly in the UI thread when we do this? I mean it doesn't have that much overhead, but we can do this 100% lazy in the cleaner thread.

otbutz:
@jonahgraham I think the ResourceTrackerThreadFactory should be close enough to what the InnocuousThreadFactory does.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Resource.java
jonahgraham: The runnable is done in the UI thread in that case, but by doing it there the disposal is complete as the cleanable is removed from the cleaner and no new thread is needed at gc time either.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Resource.java
jonahgraham: > That said, the following hack might work to set the thread name indirectly without a factory:

I don't think this will work, or rather not completely because it will set the name of the thread when `new Object()` is garbage collected, but then that thread will be done and future cleanables will run in a new thread.

jonahgraham (approved):
I have tried out the patch on Linux and it looks good to me. I tested it both with the manual test org.eclipse.swt.tests.manual.Bug569752_DetectNonDisposedOsResources and by running the IDE (with a couple of dispose calls commented out to see the reporting in operation - and heap status on so I can trigger GC)

This looks good to me. As this code sits so completely core to everything SWT I would prefer if someone else approved too before it gets merged.

> @jonahgraham I think the ResourceTrackerThreadFactory should be close enough to what the InnocuousThreadFactory does.

It looks like it to me.

otbutz:
Sorry for the last minute change. I've moved the cleaner from the Resource class to the ResourceTracker class. The advantage of this is that the cleaner thread is not started until the first instance of a ResourceTracker is created, as specified in [JLS §12.4.1].(https://docs.oracle.com/javase/specs/jls/se23/html/jls-12.html#jls-12.4.1)

jonahgraham:
> Sorry for the last minute change. I've moved the cleaner from the Resource class to the ResourceTracker class. The advantage of this is that the cleaner thread is not started until the first instance of a ResourceTracker is created, as specified in [JLS §12.4.1](https://docs.oracle.com/javase/specs/jls/se23/html/jls-12.html#jls-12.4.1)

I did a recheck with the latest commit. It looks fine to me.

otbutz:
Missed a call to `Cleanable.clean()` within `ignoreNonDisposed()`.

HannesWell:
> This looks good to me. As this code sits so completely core to everything SWT I would prefer if someone else approved too before it gets merged.

I just had a brief look and generally this looks sane, but I can have a detailed second review tomorrow.

@otbutz is this an urgent change for you that should be available for the March release? If not we could postpone the submission to the beginning of the next release cycle to have more time to find potential, unexpected drawbacks.

In general @otbutz could you please squash all commits into one with a commit message that covers and if necessary explains the overall change?

otbutz:
Feel free to postpone. The current finalize-based implementation works fine, and it would be unwarranted to rush a replacement that hasn't been properly vetted.

iloveeclipse:
I've only rebased on latest SWT state.

iloveeclipse (approved):
LGTM.

PR #1707, opened 2025-01-06
Text in Dialog was cutoff due to invalid use of AdjustWindowRectEx

ShahzaibIbrahim:
AdjustWindowRectEx does not take DPI into account. So using AdjustWindowRectExForDpi and passing the zoom level to retrieve correct coordinates and scaling accordingly.

### How to Reproduce

- Run the Runtime Workspace with the following VM Arguments on 200% monitor

```
-Dswt.autoScale=quarter
```
- Add a class and create a **conditional breakpoint**
- Double click on Breakpoint to remove it
- Maximize the main IDE window
- The dialog will appear with the text cutoff
image

### **Note**

Replacing AdjustWindowRectEx in Decoration fixes the issue for dialogs but replacing all the non-dpi calls for AdjustWindowRectEx even though there is no visible difference (but slight difference in returned coordinates) right now but having it DPI dependent makes it consistent.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Widget.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Widget.java
@@ -2708,5 +2708,15 @@ int getSystemMetrics(int nIndex) {
 	return OS.GetSystemMetrics(nIndex);
 }
 
+boolean adjustWindowRectEx(RECT lpRect, int dwStyle, boolean bMenu, int dwExStyle) {
+	/*
+	 * DPI-dependent version of the method was introduced with Windows 10 Version 1607
+	 */
+	if (OS.WIN32_BUILD >= OS.WIN32_BUILD_WIN10_1607) {
+		return OS.AdjustWindowRectExForDpi (lpRect, dwStyle, bMenu, dwExStyle, DPIUtil.mapZoomToDPI(nativeZoom));
+	}
+	return OS.AdjustWindowRectEx(lpRect, dwStyle, bMenu, dwExStyle);
+}
+
 
 }
--- a/bundles/org.eclipse.swt/Eclipse SWT PI/win32/library/os.c
+++ b/bundles/org.eclipse.swt/Eclipse SWT PI/win32/library/os.c
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2024 IBM Corporation and others.
+ * Copyright (c) 2000, 2025 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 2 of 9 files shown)


HeikoKlare (commented):
One first comment on the new usage of the Windows API method: The method has been introduced for Windows 10 1607 (https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-adjustwindowrectexfordpi#requirements). This means:
- We should guard its calls by a version check and use the old (non-DPI-aware) method if that method is not met (like, e.g., for `getSystemMetrics(ForDpi)`, as done in `Widget#getSystemMetrics()`
- We should load the native method dymically to avoid a linker error on platforms that do not provide that method. This is to avoid the issue that we had before that SWT actually required rather update-to-date Windows versions due to rather new methods being statically linked to. For example, take a look at how `GetSystemMetricsForDpi` is called in `os.c`.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Widget.java
```diff
@@ -2708,5 +2708,15 @@ int getSystemMetrics(int nIndex) {
 	return OS.GetSystemMetrics(nIndex);
 }
 
+boolean adjustWindowRectEx(RECT lpRect, int dwStyle, boolean bMenu, int dwExStyle) {
+	/*
+	 * DPI-dependent version of the method was introduced with Windows 10 Version 1607
+	 */
+	if (OS.WIN32_BUILD >= OS.WIN32_BUILD_WIN10_1607) {
+		return OS.AdjustWindowRectExForDpi (lpRect, dwStyle, bMenu, dwExStyle, getZoom());
```
HeikoKlare, line 2716: Why does this use `getZoom()`?
- The method expects the DPI, not the zoom, so it probably needs to be converted?
- Why use the device zoom instead of the native zoom, like in the `getSystemMetrics()` method? Is that correct?

HeikoKlare (commented):
Replacing usages of that method with the DPI-dependent version makes sense.

I have updated the PR by merging the commits and fixing copy-paste errors in the documentation.

Still I am non-deterministically getting wrongly scaled versions of dialogs, e.g., this one completely missing the checkbox to not show that message again:
image
I also have a question regarding the DPI parameter of the newly adapted method, which may be related to the issue.

HeikoKlare (commented):
The change itself looks sound, but I cannot successfully validate that it has the intended effect. With the change, I now _always_ get a dialog that is cut off (not showing the checkbox):

image

HeikoKlare (approved):
Turns out that the native adaptation was erroneous (linking to a wrong library), so that the method always returned `false`. I've fixed that with 99cf55d.

For me, dialogs now show up in the correct size for all scenarios (that I have tested). E.g., in the originally documented scenario (200% monitor), it looks as expected:
image

@ShahzaibIbrahim can you please evaluate whether the change work as expected as well?

PR #1708, opened 2025-01-07
Display.withCrLf() is producing wrong line breaks on Windows

### LINKED ISSUE #1557: Display.withCrLf() is producing wrong line breaks on Windows

Christopher-Hermann:
Replacing all line breaks on windows with CrLf using regex, which is much safer.

Fixes #1557

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
@@ -5152,85 +5152,16 @@ String wrapText (String text, long handle, int width) {
 }
 
 static String withCrLf (String string) {
-
-	/* If the string is empty, return the string. */
-	int length = string.length ();
-	if (length == 0) return string;
-
-	/*
-	* Check for an LF or CR/LF and assume the rest of
-	* the string is formated that way.  This will not
-	* work if the string contains mixed delimiters.
-	*/
-	int i = string.indexOf ('\n', 0);
-	if (i == -1) return string;
-	if (i > 0 && string.charAt (i - 1) == '\r') {
+	if (string == null) {
 		return string;
 	}
-
-	/*
-	* The string is formatted with LF.  Compute the
-	* number of lines and the size of the buffer
-	* needed to hold the result
-	*/
-	i++;
-	int count = 1;
-	while (i < length) {
-		if ((i = string.indexOf ('\n', i)) == -1) break;
-		count++; i++;
-	}
-	count += length;
-
-	/* Create a new string with the CR/LF line terminator. */
-	i = 0;
-	StringBuilder result = new StringBuilder (count);
-	while (i < length) {
-		int j = string.indexOf ('\n', i);
-		if (j == -1) j = length;
-		result.append (string.substring (i, j));
-		if ((i = j) < length) {
-			result.append ("\r\n"); //$NON-NLS-1$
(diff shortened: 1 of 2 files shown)


Christopher-Hermann:
I developed the fix on MacOS, so I had no change to test the fix. Would be nice if someone on windows can retest the fix.

deepika-u:
@Christopher-Hermann 
I tested this pr with below test case so that it covers all the cases from your fix(regex).

```
	@Test
	public void test_mixedLfAndCrfl() {
		//Use text control for testing since Display.withCrLf() is package private
		Text text = new Text(shell, SWT.None);

		//text.setText("First Line \n second line \r\n third line");
                //text.setText("\r First \r\r Line \n second \n\n line \r\n third \r\n\r\n line \r");
		text.setText("\n First \r\r Line \n second \n\n line \r\n third \r\n\r\n line \n");
		assertEquals("\r\n First \r\n\r\n Line \r\n second \r\n\r\n line \r\n third \r\n\r\n line \r\n", text.getText());
	}
```

**Your fix works like charm.**
It is clean, simple and straight forward. A go for the fix from my side. 

Tested on below environment:
Eclipse SDK
Version: 2025-03 (4.35)
Build id: I20250107-1800
Java vendor: Eclipse Adoptium
Java runtime version: 23.0.1+11
Java version: 23.0.1

DenisUngemach:
I have tested the PR on a Text Control on windows, it works. 

One thing i want to mention. In this change a single \r will be replaced with a \r\n.
So we have this in the Text widget:

Before:
\r was shown as space
With PR:
\r is replaced with \r\n and shown as new line.

This is a change in the behavior.

Here are the uses:
\n (Line Feed):
on Unix systems 
\r\n (Carriage Return + Line Feed):
on Windows-systems
\r (Carriage Return):
was used as new line on old macOS-versions  (before OS X)

Please decide which behavior you wish. In my opinion both is fine and this is not that relevant.

Christopher-Hermann:
> Before: \r was shown as space With PR: \r is replaced with \r\n and shown as new line.

Good point. As \r was used as new lines in old macOS, in my opinion the new behavior is correct. But changing the behavior is of course never a good idea...

Any other opinions on that?

BeckerWdf:
> in older versions of macOS

how old are these "older versions" of macOS?

BeckerWdf:
> before OS X

Ah I see: "before OS X". This long time ago. I think we can merge this PR as it is.

jukzi:
`org.eclipse.swt.widgets.Display.withCrLf("\n ".repeat(100_000_000));` seems to be twice as slow now (https://github.com/eclipse-platform/eclipse.platform.swt/issues/1718), can you please fix?

PR #1711, opened 2025-01-08
Use monitor-aware coordinates in multi-zoom coordinate system

amartya4256:
This PR contributes to the use of MonitorAware Points and Rectangles for the translation between points and pixels coordinates in the Display Coordinate System. Since the Display Coordinate System can have different scales (zoom) in different area, it is designed to be not continuous in the points coordinates. Hence when we manipulate the coordinates of a Point or a Rectangle object, it might end up in a region which is between the 2 monitors in the point coordinate system, which we consider a gap. So, we need the context of the monitor on which those points and rectangles were created in the first place to evaluate the scaling factor. If the context is not available or the coordinates were updated to an irrelevant value, a fallback method tries to evaluate the right monitor for the coordinates and evaluates the scaled value with that.

Contributes to #62 and #127

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/CoordinateSystemMapperTests.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/CoordinateSystemMapperTests.java
@@ -14,6 +14,7 @@
 package org.eclipse.swt.widgets;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.util.function.*;
 import java.util.stream.*;
@@ -39,62 +40,104 @@ public class CoordinateSystemMapperTests {
 		return monitor;
 	}
 
-	void setupMonitors(CoordinateSystemMapper mapper) {
+	private void setupMonitors(CoordinateSystemMapper mapper) {
 		Rectangle boundsInPixelsForLeftMonitor = new Rectangle(0, 0, 2000, 2000);
 		Rectangle boundsInPixelsForRightMonitor = new Rectangle(2000, 0, 2000, 2000);
 		monitors = new Monitor[] { createMonitor(mapper, boundsInPixelsForLeftMonitor, 200),
 				createMonitor(mapper, boundsInPixelsForRightMonitor, 100) };
 	}
 
-	Stream<CoordinateSystemMapper> provideCoordinateSystemMappers() {
-		return Stream.of(new MultiZoomCoordinateSystemMapper(null, () -> monitors),
-				new SingleZoomCoordinateSystemMapper(null));
+	private Stream<CoordinateSystemMapper> provideCoordinateSystemMappers() {
+		return Stream.of(getMultiZoomCoordinateSystemMapper(), getSingleZoomCoordinateSystemMapper());
+	}
+
+	private MultiZoomCoordinateSystemMapper getMultiZoomCoordinateSystemMapper() {
+		return new MultiZoomCoordinateSystemMapper(null, () -> monitors);
+	}
+
+	private SingleZoomCoordinateSystemMapper getSingleZoomCoordinateSystemMapper() {
+		return new SingleZoomCoordinateSystemMapper(null);
 	}
 
 	@ParameterizedTest
 	@MethodSource("provideCoordinateSystemMappers")
 	void translatePointInNoMonitorBackAndForthShouldBeTheSame(CoordinateSystemMapper mapper) {
 		setupMonitors(mapper);
-		Point pt = new Point(5000, -400);
+		Point pt = createExpectedPoint(mapper, 5000, -400, monitors[0]);
 		Point px = mapper.translateToDisplayCoordinates(pt, monitors[0].getZoom());
 		assertEquals(pt, mapper.translateFromDisplayCoordinates(px, monitors[0].getZoom()));
 	}
 
-	@ParameterizedTest
-	@MethodSource("provideCoordinateSystemMappers")
-	@Disabled("Disabled due to current limitations of MultiZoomCoordinateSystemMapper")
-	void translatePointInGapBackAndForthShouldBeTheSame(CoordinateSystemMapper mapper) {
+	@Test
+	void translatePointInGapBackAndForthInSingleZoomShouldBeTheSame() {
+		SingleZoomCoordinateSystemMapper mapper = getSingleZoomCoordinateSystemMapper();
 		setupMonitors(mapper);
 		Point pt = new Point(1900, 400);
 		Point px = mapper.translateToDisplayCoordinates(pt, monitors[0].getZoom());
 		assertEquals(pt, mapper.translateFromDisplayCoordinates(px, monitors[0].getZoom()));
 	}
 
-	@ParameterizedTest
-	@MethodSource("provideCoordinateSystemMappers")
-	void translateRectangleInNoMonitorBackAndForthShouldBeTheSame(CoordinateSystemMapper mapper) {
+	@Test
+	void translatePointInGapBackAndForthInMultiZoomShouldEndInsideTheSameMonitor() {
+		MultiZoomCoordinateSystemMapper mapper = getMultiZoomCoordinateSystemMapper();
 		setupMonitors(mapper);
-		Rectangle rectInPts = new Rectangle(5000, -400, 200, 200);
-		Rectangle rectInPxs = mapper.translateToDisplayCoordinates(rectInPts, monitors[0].getZoom());
-		assertEquals(rectInPts, mapper.translateFromDisplayCoordinates(rectInPxs, monitors[0].getZoom()));
+		Point pt = new Point(1900, 400);
+		Point px = mapper.translateToDisplayCoordinates(pt, monitors[0].getZoom());
+		Point translatedPt = mapper.translateFromDisplayCoordinates(px, monitors[0].getZoom());
+		Point translatedPx = mapper.translateToDisplayCoordinates(translatedPt, monitors[0].getZoom());
+		assertEquals(new Point(translatedPt.x, translatedPt.y), translatedPx);
+		assertEquals(translatedPx, px);
 	}
 
 	@ParameterizedTest
 	@MethodSource("provideCoordinateSystemMappers")
-	@Disabled("Disabled due to current limitations of MultiZoomCoordinateSystemMapper")
-	void translateRectangleInGapBackAndForthShouldBeTheSame(CoordinateSystemMapper mapper) {
+	void translateRectangleInNoMonitorBackAndForthShouldBeTheSame(CoordinateSystemMapper mapper) {
+		setupMonitors(mapper);
+		Rectangle rectInPts = createExpectedRectangle(mapper, 5000, -400, 200, 200, monitors[0]);
+		Rectangle rectInPxs = mapper.translateToDisplayCoordinates(rectInPts, monitors[0].getZoom());
+		assertEquals(rectInPts, mapper.translateFromDisplayCoordinates(rectInPxs, monitors[0].getZoom()));
+	}
+
+	@Test
+	void translateRectangleInGapBackAndForthInSingleZoomShouldBeTheSame() {
+		SingleZoomCoordinateSystemMapper mapper = getSingleZoomCoordinateSystemMapper();
 		setupMonitors(mapper);
 		Rectangle rectInPts = new Rectangle(1800, 400, 100, 100);
 		Rectangle rectInPxs = mapper.translateToDisplayCoordinates(rectInPts, monitors[0].getZoom());
 		assertEquals(rectInPts, mapper.translateFromDisplayCoordinates(rectInPxs, monitors[0].getZoom()));
 	}
 
-	@ParameterizedTest
-	@MethodSource("provideCoordinateSystemMappers")
-	@Disabled("Disabled due to current limitations of MultiZoomCoordinateSystemMapper")
-	void translateRectangleInGapPartiallyInRightBackAndForthShouldBeTheSame(CoordinateSystemMapper mapper) {
+	@Test
+	void translateRectangleInGapBackAndForthInMultiZoomShouldBeInMonitorBounds() {
+		MultiZoomCoordinateSystemMapper mapper = getMultiZoomCoordinateSystemMapper();
 		setupMonitors(mapper);
-		Rectangle rectInPts = new Rectangle(1950, 400, 100, 100);
+		Rectangle rectInPts = new Rectangle(1800, 400, 100, 100);
+		Rectangle rectInPxs = mapper.translateToDisplayCoordinates(rectInPts, monitors[0].getZoom());
+		Rectangle rectInPtsTranslated = mapper.translateFromDisplayCoordinates(rectInPxs, monitors[0].getZoom());
+		boolean isInsideMonitor = false;
+		for (Monitor monitor : monitors) {
+			if (monitor.getClientArea().intersects(rectInPtsTranslated)) {
+				isInsideMonitor = true;
+				break;
+			}
+		}
+		assertTrue(isInsideMonitor, "The translated rectangle in points is inside the monitor bounds in points");
+	}
+
+	@Test
+	void translateRectangleInGapPartiallyInRightBackAndForthInSingleZoomShouldBeTheSame() {
+		SingleZoomCoordinateSystemMapper mapper = getSingleZoomCoordinateSystemMapper();
+		setupMonitors(mapper);
+		Rectangle rectInPts = new Rectangle(1950, 400, 150, 100);
+		Rectangle rectInPxs = mapper.translateToDisplayCoordinates(rectInPts, monitors[0].getZoom());
+		assertEquals(rectInPts, mapper.translateFromDisplayCoordinates(rectInPxs, monitors[0].getZoom()));
+	}
+
+	@Test
+	void translateRectangleInGapPartiallyInRightBackAndForthInMultiZoomShouldBeInside() {
+		MultiZoomCoordinateSystemMapper mapper = getMultiZoomCoordinateSystemMapper();
+		setupMonitors(mapper);
+		Rectangle rectInPts = new MonitorAwareRectangle(1950, 400, 150, 100, monitors[1]);
 		Rectangle rectInPxs = mapper.translateToDisplayCoordinates(rectInPts, monitors[0].getZoom());
 		assertEquals(rectInPts, mapper.translateFromDisplayCoordinates(rectInPxs, monitors[0].getZoom()));
 	}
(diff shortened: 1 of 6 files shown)


fedejeanne:
@amartya4256 I see this PR is a WiP (based on the commit text), which is fine for me, but please add some description to the PR (the issue that is fixed, intended result, how to test it, etc).

For the moment I can only do an initial review of the coding standards and so on.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Point.java
```diff
@@ -41,7 +41,7 @@
  * @see <a href="http://www.eclipse.org/swt/">Sample code and further information</a>
  */
 
-public final class Point implements Serializable {
```
fedejeanne, line 44: Would it be possible to wrap `Point` instead of extending it? Opening a class may have undesired consequences in consumers. 

If it's not possible, have you already considered simply adding the field `Monitor` to `Point`?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MultiZoomCoordinateSystemMapper.java
```diff
@@ -93,122 +93,148 @@ public Rectangle mapMonitorBounds(Rectangle rect, int zoom) {
...
+		if(monitor != null) {
+			if (monitor.getClientArea().contains(x, y)) {
+				return monitor;
+			} else {
+				Monitor containingMonitor = getContainingMonitorForPoints(x, y);
+				if (containingMonitor != null) {
+					return containingMonitor;
+				} else {
+					return monitor;
+				}
```
fedejeanne, line 191: Extract to method and call it twice

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MultiZoomCoordinateSystemMapper.java
```diff
@@ -255,11 +281,32 @@ private Point getPointFromPixels(Monitor monitor, int x, int y) {
 		int zoom = getApplicableMonitorZoom(monitor);
 		int mappedX = DPIUtil.scaleDown(x - monitor.clientX, zoom) + monitor.clientX;
 		int mappedY = DPIUtil.scaleDown(y - monitor.clientY, zoom) + monitor.clientY;
-		return new Point(mappedX, mappedY);
+		Point pt = new MonitorAwarePoint(mappedX, mappedY, monitor);
+		return pt;
```
fedejeanne, line 285:
```suggestion
		return new MonitorAwarePoint(mappedX, mappedY, monitor);
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
```diff
@@ -5324,5 +5324,4 @@ private boolean setDPIAwareness(int desiredDpiAwareness) {
 	}
 	return true;
 }
-
```
fedejeanne, line 5327: Undo

fedejeanne (commented):
I played around a bit with it and this is what I found 
## When _rescaling at runtime_ is **deactivated** 
- Moving the workbench window between 2 monitors with different zoom rescales the menu bar according to the zoom of the monitor. 

IIRC this is expected so I do **not** consider it a bug.

## When _rescaling at runtime_ is **activated** 
No findings

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Point.java
fedejeanne: Class sealed

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MultiZoomCoordinateSystemMapper.java
amartya4256: I think I cant do it cause its not completely identical. Have a look again?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MultiZoomCoordinateSystemMapper.java
amartya4256: Done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MultiZoomCoordinateSystemMapper.java
fedejeanne: Good point. You can still make it more readable though:

```java
private Monitor getValidMonitorIfApplicable(int x, int y, Monitor monitor) {
	if (monitor != null) {
		if (monitor.getClientArea().contains(x, y)) {
			return monitor;
		}

		Monitor containingMonitor = getContainingMonitorForPoints(x, y);
		return containingMonitor != null ? containingMonitor : monitor;
	}
	Monitor containingMonitor = getContainingMonitorForPoints(x, y);
	return containingMonitor != null ? containingMonitor : monitorSupplier.get()[0];
}
```

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Point.java
fedejeanne: Now that I think about it, sealing the class forces you to name the allowed subclasses in `Point` and, more importantly, to make them part of the API (they are now **public** and belong to **common**).

Isn't there a better way?

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Point.java
```diff
@@ -16,6 +16,8 @@
 
 import java.io.*;
 
+import org.eclipse.swt.internal.*;
```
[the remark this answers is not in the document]
amartya4256: Done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MultiZoomCoordinateSystemMapper.java
amartya4256: Yes that's better. Done

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Point.java
amartya4256: If we have to work with sealed classes we have to do it with either the mapper or the monitor aware classes. And in a discussion with @HeikoKlare , we came to a conclusion that, if we introduce this as an API, we can later use it everywhere and can in some way get rid of the DPIUtils completely. In other words, introducing them as an API is more futureproof

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/MonitorAwareRectangle.java
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Instances of this class represent {@link org.eclipse.swt.graphics.Rectangle}
+ * objects along with the context of the monitor in relation to which they are
+ * placed on the display. The monitor awareness makes it easy to scale and
+ * translate the rectangles between pixels and points
+ *
+ * @since 3.129
+ */
+public final class MonitorAwareRectangle extends Rectangle {
```
HeikoKlare, line 26: These new classes should be tagged with: 
`@noreference This class is not intended to be referenced by clients`

Just like, e.g., `GCData`. We would not like to have this as completely public API (at least for now).

HeikoKlare (commented):
No testing and in-depth review from my side so far, but at least one important point from my side: please mark the new classes as "not API" (i.e., "noreference"), as we should not make them public API (at least for now).

Please also improve the commit message to shortly explain the reasons for the change and what it does, so that people can later on generally understand what the commit does. Please also add that commit description to the PR description, in the based case with some how-to-test information.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MultiZoomCoordinateSystemMapper.java
```diff
@@ -255,11 +261,15 @@ private Point getPointFromPixels(Monitor monitor, int x, int y) {
...
 	}
 
 	private int getApplicableMonitorZoom(Monitor monitor) {
 		return DPIUtil.getZoomForAutoscaleProperty(monitor.zoom);
 	}
 
+
+
+
+
```
fedejeanne, line 274:
```suggestion
```

fedejeanne:
I tested and found this:

## When _rescaling at runtime_ is deactivated
- The size of the checkboxes and radio buttons in the _Preferences_ reacts to zoom changes. 

This is on the 100% monitor:
image

To reproduce:
- Start Eclipse in 200% monitor
- Open _Preferences_ (it should open in the 200% monitor)
- Move _Preferences_ dialog to the 100% monitor

## When _rescaling at runtime_ is activated

- It is still difficult to (re)attach views when the workbench window spans across 2 monitors **(Corner case!)**

reattach_views_rescaling_enabled

In the video you can see that:
- I wasn't able to detach the editor in my 1st attempt
- Once detached, I could only reattach it on the 200% monitor (on the left). For your reference, the border between both monitors is right in the middle of the "Navigate back" (arrow)
image

## My setup
- 100% Monitor on the **right**
- 200% **Primary** monitor on the **left**. 

## Prioritization
I can live with both issues so I suggest we merge this PR and leave the findings for another release. @HeikoKlare WDYT?

HeikoKlare:
Is the issue with rescaling at runtime being deactivated really an issue introduced by this PR? I don't see how this could be related, so would be good to check that. Otherwise it might be an independent regression for which we need to open an issue and address separately as soon as possible.

amartya4256:
When the rescaling at runtime is deactivated it still uses the old singlezoom mapper. So I don't think that this PR could have caused this issue.

fedejeanne:
It's not --> https://github.com/vi-eclipse/Eclipse-Platform/issues/191

HeikoKlare:
@amartya4256 can you please update the commit message as requested in https://github.com/eclipse-platform/eclipse.platform.swt/pull/1711#pullrequestreview-2560442073?

> Please also improve the commit message to shortly explain the reasons for the change and what it does, so that people can later on generally understand what the commit does. Please also add that commit description to the PR description, in the based case with some how-to-test information.

bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/CoordinateSystemMapperTests.java
```diff
@@ -60,16 +69,27 @@ void translatePointInNoMonitorBackAndForthShouldBeTheSame(CoordinateSystemMapper
 		assertEquals(pt, mapper.translateFromDisplayCoordinates(px, monitors[0].getZoom()));
 	}
 
-	@ParameterizedTest
-	@MethodSource("provideCoordinateSystemMappers")
-	@Disabled("Disabled due to current limitations of MultiZoomCoordinateSystemMapper")
-	void translatePointInGapBackAndForthShouldBeTheSame(CoordinateSystemMapper mapper) {
+    @Test
+	void translatePointInGapBackAndForthInSingleZoomShouldBeTheSame() {
```
HeikoKlare, line 73: Please fix inconsistent indentation

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/MonitorAwarePoint.java
```diff
@@ -0,0 +1,55 @@
...
+ *******************************************************************************/
+package org.eclipse.swt.graphics;
+
+import org.eclipse.swt.widgets.*;
+
+/**
+ * Instances of this class represent {@link org.eclipse.swt.graphics.Point}
+ * objects along with the context of the monitor in relation to which they are
+ * placed on the display. The monitor awareness makes it easy to scale and
+ * translate the points between pixels and points
```
HeikoKlare, line 22:
```suggestion
 * translate the points between pixels and points.
```

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/MonitorAwarePoint.java
```diff
@@ -0,0 +1,55 @@
...
+	 *
+	 * @since 3.129
+	 */
+	public MonitorAwarePoint(int x, int y, Monitor monitor) {
+		super(x, y);
+		this.monitor = monitor;
+	}
+
+	/**
+	 * Gets the monitor with whose context the instance is created
```
HeikoKlare, line 46:
```suggestion
	 * {@return the monitor with whose context the instance is created}
```

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/MonitorAwareRectangle.java
```diff
@@ -0,0 +1,57 @@
...
+	 *
+	 * @since 3.129
+	 */
+	public MonitorAwareRectangle(int x, int y, int width, int height, Monitor monitor) {
+		super(x, y, width, height);
+		this.monitor = monitor;
+	}
+
+	/**
+	 * Gets the monitor with whose context the instance is created
```
HeikoKlare, line 48:
```suggestion
	 * {@return the monitor with whose context the instance is created}
```

HeikoKlare (approved):
Thank you, @amartya4256!
The changes now look fine to me. I have also tested the changes and found the behavior of the multi-zoom coordinate mapping much better now. It's only corner cases (right at the border between two monitors with different scalings) that sometimes behave a bit "unexpected", but at least it behaves rather "consistently" (e.g., regarding a preview of a detach operation and the final result). And those corner cases are not easy to resolve anyway, so it's fine like it is. In particular, dragging things to another monitor now behaves eben better than without monitor-specific scaling, where, for example, a detach operation moved to another monitor with different zoom value leads to unexpected positioning.

In addition, this change is safe in the sense that it only affected the multi-zoom coordinate system mapper, which is used by the experimental monitor-specific scaling mode on Windows. So no effects regarding existing behavior of "ordinary" functionality are to be expected.

With recent change 7a58782, I adapted the equals/hashcode implementations for Point/Rectangle and MonitorAwarePoint/MonitorAwareRectangle, as otherwise they would not take the monitor into account.

nedtwigg:
First off, I *love* this new pace of innovation! I don't mind regressions or breaking changes when the code is moving this fast, so great to see SWT come back to life!

This change makes it impossible to instantiate a `Point` or `Rectangle` from Kotlin, due to a `Sealed types cannot be instantiated` error. It is possible to work around this by writing a factory method in java and replacing all `new Point(x, y)` with `PointFactory.create(x, y)`.

But if the code is only used in the `MultiZoomCoordinateSystemMapper`, it seems like it would be better to contain this complexity there. Dunno if there's been any benchmarking, but I would expect keeping `Point` and `Rectangle` as monomorphic would help a fair bit.

Would you be open to making this geometric base classes back how they were, or are these `MonitorAware` versions here to stay?

HeikoKlare:
Thank you for your feedback! 

First of all, the `MonitorAware` versions are supposed to stay. They are only accessed in `MultiZoomCoordinateSysteMapper` but are also "used" outside. More precisely, a shell's methods return monitor-aware points/rectangles that may be processed by clients (changing their values) and then passed to any other method accepting points/rectangles in a shell again. It's is necessary to carry the monitor the original point/rectangle was created for belongs to, as otherwise it is not possible to identify the monitor it belongs to. The underlying issue is that some years ago a decision was made to use logical coordinates in SWT also at the display level (i.e., across all monitors), which is conceptually incompatible with monitor-specific scaling of the UI as it relies on a linear relation between logical points in pixels but with different scaling for every monitor, this is not univerally given anymore. The introduction of the monitoir-aware points/rectangles was the best way of dealing with the limitatiations without unacceptable glitched and without requiring to rewrite consumers that we found so far.

Regarding the issue with Kotlin, I have to admit that I do not understand the actual issue. The structure of, e.g., Point and MonitorAwarePoint is nothing special, just a sealed hierarchy of two classes. If Kotlin is not able to instantiate those types, my first impression is that Kotlinm must be incapable of properly processing sealed types at all (but I gues that's not the case). Do you know more about the reasons for Kotlin complaining here?

ptziegler:
> Regarding the issue with Kotlin, I have to admit that I do not understand the actual issue. The structure of, e.g., Point and MonitorAwarePoint is nothing special, just a sealed hierarchy of two classes. If Kotlin is not able to instantiate those types, my first impression is that Kotlinm must be incapable of properly processing sealed types at all (but I gues that's not the case). 

From what I understand, Kotlin treats sealed types as an abstract classes:
https://kotlinlang.org/docs/sealed-classes.html#constructors
> A sealed class itself is always an [abstract class](https://kotlinlang.org/docs/classes.html#abstract-classes), and as a result, can't be instantiated directly.

HeikoKlare:
Thank you for those insights, @ptziegler! I wasn't aware of that behavior/decision in Kotlin. If I do not miss anything, that means changing a final type to a sealed type is source- and binary-compatible for Java consumers, but for Kotlin consumers it might break. I don't think this is anything you need to or should considers when evolving Java APIs but rather consider it a limitation in Kotlin, in accordance with:

> So my guess is that Kotlin does a 1-to-1 mapping from sealed Java classes to sealed Kotlin classes, while ignoring the fact that sealed Java classes don't need to be abstract.

So maybe this is something that should be reported to JetBrains?

HeikoKlare:
I have to admit that I do not understand that proposal. It conflicts with what I tired to explain in https://github.com/eclipse-platform/eclipse.platform.swt/pull/1711#issuecomment-2717008752
With the proposed change, points and rectangles returned by the methods of a shell, modified outside of it, and then passed to that (or another) shell again, will not carry information about the monitor this point or rectangle belongs to anymore. That leads to positioning issues when using multiple monitors of different zoom, which have been addressed by this PR. Performing the proposed change would significantly degrade the behavior of the monitor-specific scaling support on Windows again.

And in my opinion, this is still without any proper need. We have a correct usage of sealed types and a source- and binary-compatible change made to use it. I am quite surprised that Kotlin seems to be incompatible with Java code regarding sealed types. But from my point of view, this pushes the responsibility for fixing this at Kotline side. Even if we found a solution here, this situation can occur again if Kotlin is not adapted. Maybe there is a misunderstanding on my side, but this is what I understood from the explanations so far.

nedtwigg:
Kotlin incompatibility doesn't bother me at all, I'd like to drop that completely. My problem is that geometric equality is broken.

- Two points with the same coordinates might not be equal to each other, depending on whether one has a monitor implicitly attached.
- The way that a monitor should be passed along for arithmetic is not clear
- To me this is a big smell that there is some more internal approach which is possible. I tried again in #1905. If this approach is okay I think it could be cleaned up, but I tried to demonstrate that the tests can pass without changing the inheritance hierarchy of geometry.

PR #1713, opened 2025-01-09
Obtain ImageData for any requested zoom value, not only autoscaled values

amartya4256:
This contribution helps retrieve the imageData for the requested zoom by searching for the zoom in the zoomToHandleMap followed by (if not available) scaling the handle at the zoom nearest to it while getting rid of only searching among the autoscaled zoom values.

Contributes to #62 and #127

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
@@ -1359,10 +1359,10 @@ public ImageData getImageData() {
  */
 public ImageData getImageData (int zoom) {
 	if (isDisposed()) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
-	int currentZoom = getZoom();
-	if (zoom == currentZoom) {
-		return getImageDataAtCurrentZoom();
-	} else if (imageProvider != null) {
+	if (zoomLevelToImageHandle.containsKey(zoom)) {
+		return zoomLevelToImageHandle.get(zoom).getImageData();
+	}
+	if (imageProvider != null) {
 		return imageProvider.getImageData(zoom);
 	}
 
@@ -1371,7 +1371,9 @@ public ImageData getImageData (int zoom) {
 	if (memGC != null) {
 		return getImageDataAtCurrentZoom();
 	}
-	return DPIUtil.scaleImageData (device, getImageMetadata(currentZoom).getImageData(), zoom, currentZoom);
+	TreeSet<Integer> availableZooms = new TreeSet<>(zoomLevelToImageHandle.keySet());
+	int closestZoom = Optional.ofNullable(availableZooms.higher(zoom)).orElse(availableZooms.lower(zoom));
+	return DPIUtil.scaleImageData (device, getImageMetadata(closestZoom).getImageData(), zoom, closestZoom);
 }
 
 /**
--- a/tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/graphics/ImageWin32Tests.java
+++ b/tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/graphics/ImageWin32Tests.java
@@ -15,6 +15,7 @@ package org.eclipse.swt.graphics;
 
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotEquals;
 
 import org.eclipse.swt.internal.DPIUtil;
 import org.eclipse.swt.widgets.Display;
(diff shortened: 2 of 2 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1359,10 +1359,10 @@ public ImageData getImageData() {
  */
 public ImageData getImageData (int zoom) {
 	if (isDisposed()) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
-	int currentZoom = getZoom();
-	if (zoom == currentZoom) {
-		return getImageDataAtCurrentZoom();
-	} else if (imageProvider != null) {
+	if (zoomLevelToImageHandle.get(zoom) != null) {
```
HeikoKlare, line 1362:
```suggestion
	if (zoomLevelToImageHandle.containsKey(zoom)) {
```

tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/graphics/ImageWin32Tests.java
```diff
@@ -36,6 +37,25 @@ public void setUp() {
...
+		int zoom1 = 125;
+		int zoom2 = 150;
+		ImageData imageDataAtZoom1 = image.getImageData(125);
+		ImageData imageDataAtZoom2 = image.getImageData(150);
+		assertNotEquals("ImageData::height should not be the same for imageData at different zoom levels within the same autoScale zoom bounds", imageDataAtZoom1.height, imageDataAtZoom2.height);
+		assertNotEquals("ImageData::width should not be the same for imageData at different zoom levels within the same autoScale zoom bounds", imageDataAtZoom1.width, imageDataAtZoom2.width);
+		int correctedZoomForZoom1 = DPIUtil.getZoomForAutoscaleProperty(zoom1);
+		int correctedZoomForZoom2 = DPIUtil.getZoomForAutoscaleProperty(zoom2);
+		assertEquals("Zoom1 and Zoom2 should have the same corrected zoom", correctedZoomForZoom1, correctedZoomForZoom2);
+		ImageData imageDataAtCorrectedZoom = image.getImageData(correctedZoomForZoom1);
```
HeikoKlare, line 52: In case we leave this test in, I propose to not rely on specific behavior depending on auto scaling properties. I would rather limit the test to the first part, only testing whether the scaled data has the expected size (no matter which auto scaling mode we are using).

HeikoKlare (commented):
It seems to me as if I had misunderstood the original intent of this change. It does not seem to be an actual bugfix but rather an improvement regarding which image data (more particularly, the image data at which scaling) to use for generating the one at the requested zoom level of. Is that correct?

Then of course the kind of test I proposed and is provided in this PR is not a regression test for the change. And probably it's something that we do not even need to test in that way at all.

tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/graphics/ImageWin32Tests.java
amartya4256: Alright

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java

tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/graphics/ImageWin32Tests.java
amartya4256: Done

amartya4256:
> It seems to me as if I had misunderstood the original intent of this change. It does not seem to be an actual bugfix but rather an improvement regarding which image data (more particularly, the image data at which scaling) to use for generating the one at the requested zoom level of. Is that correct?
> 
> Then of course the kind of test I proposed and is provided in this PR is not a regression test for the change. And probably it's something that we do not even need to test in that way at all.

In the previous implementation in case of integer 200 scaling, we only allowed the clients to retrieve the imageData of corrected zoom since our initial expectations were that we don't allow images to be created at, for example, the quarter zoom if the quarter scaling is disabled. But since we know that in some cases like Program icons we can get handles of image at these zoom levels, it doesnt make sense to restrict clients from requesting imagedata at certain zoom levels. In other words, this is indeed a feature enhancement as we are now allowing to store any zoom level in the handles map and retrieve them on demand.

Hence, I would not consider this change as a bugfix but rather an improvement. But the tests do make sense to me since it validates the behaviour of getImageData at random zoom levels. It would not be considered a regression test but a feature test.

PR #1716, opened 2025-01-11
[WiP] Change monitor-specific scaling setting before Display creation

HeikoKlare:
⚠️ This PR currently only serves as basis for discussions

The method `Display#setRescalingAtRuntime()` allows to activate the monitor-specific scaling mechanism on Windows. This must, however, actually be activated before the Display is instantiated, like it is done during the Display constructor execution if the VM argument for global activation of monitor-specific scaling is specified. In consequence, calling `#setRescalingAtRuntime()` does not take the full intended effect.

This change deprecated the method (along with its according accessor `#isRescalingAtRuntime()`) and adds a new method to activate monitor-specific scaling on the Display class instead of an instance to be used for subsequent displays to be instantiated.


HeikoKlare:
Closing in favor of #1719 and #1720

PR #1717, opened 2025-01-12
Fix maven group id to match the actual deployment

merks:
- Also fix test bundles and any eclipse-plugin packaging simply for consistency.

### DIFF
--- a/bundles/org.eclipse.swt/pom.xml
+++ b/bundles/org.eclipse.swt/pom.xml
@@ -19,7 +19,7 @@
         <version>4.35.0-SNAPSHOT</version>
         <relativePath>../../</relativePath>
     </parent>
-    <groupId>org.eclipse.swt</groupId>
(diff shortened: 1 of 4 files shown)


merks:
@HannesWell 

The build looks like a bit of a mine field.   I assume none of this is related to my changes.  Or?

HannesWell:
> The build looks like a bit of a mine field. I assume none of this is related to my changes. Or?

The Jenkins build is only unstable (unfortunately the GH UI cannot distinguish between unstable and failed) due to new compiler and Maven warnings:
grafik

But none of them seem to be related. For the Compiler warnings they even look incorrect and the IDE doesn't complain about unnecessary `@SuppressWarnings("restriction")`.

merks:
I'm going to assume that given the GHA build pass that there are no issues related to these changes involved.

PR #1720, opened 2025-01-13
Deprecate insufficient Display#setRescalingAtRuntime() method

HeikoKlare:
The method Display#setRescalingAtRuntime() allows to activate the monitor-specific scaling mechanism on Windows. This must, however, actually be activated before the Display is instantiated, like it is done during the Display constructor execution if the system property for global activation of monitor-specific scaling is specified. In consequence, calling #setRescalingAtRuntime() does not take the full intended effect.

This change deprecates the method so that the system property needs to be set for proper initialization of the monitor-specific scaling functionality. It also adapts the according tests to the deprecation of
the API and alternatives to be used.

⚠️ #1728 has been added as a preparation for this

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/DisplayWin32Test.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/DisplayWin32Test.java
@@ -8,36 +8,46 @@ import org.junit.jupiter.api.*;
 import org.junit.jupiter.api.extension.*;
 
 @ExtendWith(PlatformSpecificExecutionExtension.class)
+@ExtendWith(ResetMonitorSpecificScalingExtension.class)
 public class DisplayWin32Test {
 
-	private Display display;
-
-	@BeforeEach
-	public void createDisplay() {
-		display = new Display();
-	}
-
-	@AfterEach
-	public void destroyDisplay() {
-		display.dispose();
+	@Test
+	public void monitorSpecificScaling_activate() {
+		DPIUtil.setMonitorSpecificScaling(true);
+		Display display = Display.getDefault();
+		assertTrue(display.isRescalingAtRuntime());
+		assertEquals(OS.DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2, OS.GetThreadDpiAwarenessContext());
 	}
 
 	@Test
+	public void monitorSpecificScaling_deactivate() {
+		DPIUtil.setMonitorSpecificScaling(false);
+		Display display = Display.getDefault();
+		assertFalse(display.isRescalingAtRuntime());
+	}
+
+	@Test
+	@SuppressWarnings("removal")
 	public void setRescaleAtRuntime_activate() {
+		Display display = Display.getDefault();
 		display.setRescalingAtRuntime(true);
 		assertTrue(display.isRescalingAtRuntime());
 		assertEquals(OS.DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2, OS.GetThreadDpiAwarenessContext());
 	}
 
 	@Test
+	@SuppressWarnings("removal")
 	public void setRescaleAtRuntime_deactivate() {
+		Display display = Display.getDefault();
 		display.setRescalingAtRuntime(false);
 		assertFalse(display.isRescalingAtRuntime());
 		assertEquals(OS.DPI_AWARENESS_CONTEXT_SYSTEM_AWARE, OS.GetThreadDpiAwarenessContext());
 	}
 
 	@Test
+	@SuppressWarnings("removal")
 	public void setRescaleAtRuntime_toggling() {
+		Display display = Display.getDefault();
 		display.setRescalingAtRuntime(false);
 		assertFalse(display.isRescalingAtRuntime());
 		assertEquals(OS.DPI_AWARENESS_CONTEXT_SYSTEM_AWARE, OS.GetThreadDpiAwarenessContext());
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
@@ -6858,7 +6858,10 @@ public boolean isRescalingAtRuntime() {
  * @param activate whether rescaling shall be activated or deactivated
  * @return whether activating or deactivating the rescaling was successful
  * @since 3.127
+ * @deprecated this method should not be used as it needs to be called already
+ *             during instantiation to take proper effect
  */
+@Deprecated(since = "2025-03", forRemoval = true)
 public boolean setRescalingAtRuntime(boolean activate) {
 	// not implemented for Cocoa
 	return false;
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
@@ -949,7 +949,7 @@ protected void create (DeviceData data) {
 	checkSubclass ();
 	checkDisplay (thread = Thread.currentThread (), true);
 	if (DPIUtil.isMonitorSpecificScalingActive()) {
-		setRescalingAtRuntime(true);
+		setMonitorSpecificScaling(true);
 		DPIUtil.setAutoScaleForMonitorSpecificScaling();
 	}
 	createDisplay (data);
@@ -5263,6 +5263,7 @@ private class ThemeData {
 		return OS.OpenThemeData(hwndMessage, themeName, dpi);
 	}
 }
+
 /**
  * {@return whether rescaling of shells at runtime when the DPI scaling of a
  * shell's monitor changes is activated for this device}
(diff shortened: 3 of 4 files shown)


HeikoKlare:
@akoch-yatta

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
```diff
@@ -6858,7 +6858,10 @@ public boolean isRescalingAtRuntime() {
  * @param activate whether rescaling shall be activated or deactivated
  * @return whether activating or deactivating the rescaling was successful
  * @since 3.127
+ * @deprecated this method should not be used as it needs to be called already
+ *             during instantiation to take proper effect
  */
+@Deprecated(since = "2025-03", forRemoval = true)
```
akoch-yatta, line 6864: Is that the correct value for "since"? I personally would expect the version number "3.129" instead of the release name here

merks: I recall @HannesWell discussing this topic, but I don't remember where.  I think the point was that most users have no clue what the number version number means and most importantly don't know when the removal will happen from just the version number.  With a date-version like this they know (if they know the policy) that in 2027-03 it can end up being removed.

fedejeanne: FWIW I agree with Ed. Since the removal will be done "in X amount of time" then having a direct map from _version_ to _point in time_ is more useful.

Nice to know that this topic has been discussed and this form of documenting it has been accepted.

fedejeanne (approved):

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
akoch-yatta: I am of course fine with that, if there is in agreement about it. Still, it is at least in a conflict with javadoc
 of the since attribute, isn't it? We are using the version number in the "@since"
```
     * Returns the version in which the annotated element became deprecated.
     * The version string is in the same format and namespace as the value of
     * the {@code @since} javadoc tag. The default value is the empty
     * string.
```
I just had never seen this kind of usage anywhere, so I wondered and wanted to ask. But, as you said, there was an agreement for that, I can live with that 😄

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
```diff
@@ -5286,8 +5287,15 @@ public boolean isRescalingAtRuntime() {
...
  * @since 3.127
+ * @deprecated this method should not be used as it needs to be called already
+ *             during instantiation to take proper effect
  */
+@Deprecated(since = "2025-03", forRemoval = true)
 public boolean setRescalingAtRuntime(boolean activate) {
+	return setMonitorSpecificScaling(activate);
+}
+
+private boolean setMonitorSpecificScaling(boolean activate) {
```
fedejeanne, line 5298: Why the renaming (in this PR)?

fedejeanne, line 5298: Why the renaming in this PR?

fedejeanne (approved):
LGTM I only have 1 question.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
HeikoKlare: Because of introducting the new method `setMonitorSpecificScaling()` and to make the existing one consistent with that one as they relate to the same feature which is otherwise named differently.

fedejeanne: Ah, you mean this one right?

image

The intent wasn't clear in the description of this PR or in the commit message, hence my question.

HeikoKlare: Now I got the point. I lost several changes when rebasing the PR on master, which I now added again. The deprecation of API of course also means that usages have to be adapted (accepting deprecation or replacing it). I have now updated the PR with the according adaptations, which also includes that method "renaming", which just factors out the functionality into a new method with a proper name and visibility. "Proper" regaring the name means that it conform to the notion of "monitor-specific scaling" we use by now.

bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/DisplayWin32Test.java
```diff
@@ -8,48 +8,89 @@
...
+			display.dispose();
+		}
+	}
+
+	@Test
+	public void monitorSpecificScaling_deactivate() {
+		DPIUtil.setMonitorSpecificScaling(false);
+		Display display = new Display();
+		try {
+			assertFalse(display.isRescalingAtRuntime());
```
fedejeanne, line 45: Is it unnecessary/incorrect to assert that the scaling is set to `DPI_AWARENESS_CONTEXT_SYSTEM_AWARE` just like the "old" test `setRescaleAtRuntime_deactivate` does?

```suggestion
			assertFalse(display.isRescalingAtRuntime());
			assertEquals(OS.DPI_AWARENESS_CONTEXT_SYSTEM_AWARE, OS.GetThreadDpiAwarenessContext());
```

This asymmetry in the tests looks confusing to me.

bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/DisplayWin32Test.java
```diff
@@ -8,48 +8,89 @@
...
+		} finally {
+			display.dispose();
+		}
+	}
+
+	@Test
+	@SuppressWarnings("removal")
 	public void setRescaleAtRuntime_activate() {
```
fedejeanne, line 53: NIT: in these cases, when it is essential that the test (or any piece of code for that matter) runs "under special conditions", I personally like to extract a method that makes this "special condition" explicit _e.g._: 

```java
private void runInNewDisplay(Consumer<Display> consumer) {
	try {
		consumer.accept(display);
	} finally {
		display.dispose();
	}
}
```

This makes it clearer that the new implementation now force these tests to run in a new display

fedejeanne (commented):
Looks good, only minor comments

bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/DisplayWin32Test.java
HeikoKlare: Yes, it is incorrect.

fedejeanne: Care to elaborate? I'm missing some pieces of the puzzle here.

HeikoKlare: The application runs in "default" mode in this test case, so the actual DPI awareness depends on what the JVM is configured to use as process-wide DPI awareness. That's why assuming a specific DPI awareness is incorrect.

fedejeanne: Ah, the missing piece (in my mind) was the fact that `Display::setRescalingAtRuntime` actively overrides the system default both when passing `true` and `false` as a parameter while `DPIUtil::setMonitorSpecificScaling` only overrides the system value when passing `true` as a parameter. 
The problem I see here (unrelated to this PR) is that both these methods have the same name but they act differently, which is odd:
- `org.eclipse.swt.internal.DPIUtil.setMonitorSpecificScaling(boolean)`: sets a system property
- `org.eclipse.swt.widgets.Display.setMonitorSpecificScaling(boolean)`: sets **and overrides** the mode.

This currently produces (AFAIK) no bug because `Display.setMonitorSpecificScaling(boolean)` is only called if the system property has been set (_e.g._ if `DPIUtil.setMonitorSpecificScaling(true)` has been called) but I think the naming of the method in `Display` could be improved in order to make more explicit what it is supposed to do and how. 

How about renaming `Display.setMonitorSpecificScaling` to `Display.forceMonitorSpecificScaling` (in a separate PR) so that it is 100% clear that this method **overrides** the system value?

No further changes required for this PR. Approved on my side.

fedejeanne (approved):

HeikoKlare:
I see the point regarding method naming. In my opinion, the name `setMonitorSpecificScaling` is reasonable for the `Display` class. `forceMonitorSpecificScaling` would be misleading, as it does actually not "force" anything. I'd rather see the change to improve naming in the `DPIUtil` class, as it might make sense to keep the naming conforming to the system property using the termin "updateOnRuntime" referring to "autoScale", which in my opinion might be even better naming in that class. Since this PR does not touch names of method sin `DPIUtil`, I would not adapt it here.
And we need to keep in mind that of course methods with the same name may have a different meaning depending on their context (i.e., the class/interface they belong to), as the context also gives them semantics. But as said, in this case I see some chance for improvement in `DPIUtil`.

bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/DisplayWin32Test.java
HeikoKlare: In general, I agree that this pattern should be applied. In this case it was rather about proper cleanup than about running in a new display, but would still have been possible to realize it that way.
This, however, became obsolete by preparatory test method refactoring, which made this kind of construct completely obsolete, as displays are now automatically cleaned up by an according JUnit 5 extension.

Issue #1718, opened 2025-01-13
Slow org.eclipse.swt.widgets.Display.withCrLf(String)


jukzi:
reproduce: open help/"Installation Details"/Configuration (~200_000 lines)
=> UI freeze warning
regression from https://github.com/eclipse-platform/eclipse.platform.swt/pull/1708 @Christopher-Hermann 
A regexp is performance overkill.

```
eclipse.buildId=4.35.0.I20250112-1800
java.version=22.0.2
java.vendor=Eclipse Adoptium
BootLoader constants: OS=win32, ARCH=x86_64, WS=win32, NL=de_DE
Command-line arguments:  -os win32 -ws win32 -arch x86_64

org.eclipse.ui.monitoring
Warning
Mon Jan 13 15:43:12 CET 2025
UI freeze of 1,4s at 15:43:11.504


org.eclipse.ui.monitoring
Info
Mon Jan 13 15:43:12 CET 2025
Sample at 15:43:12.503 (+0,333s)
Thread 'main' tid=1 (RUNNABLE)

Stack Trace
	at java.base@22.0.2/java.util.regex.Pattern$BmpCharProperty.match(Pattern.java:4133)
	at java.base@22.0.2/java.util.regex.Pattern$Branch.match(Pattern.java:4914)
	at java.base@22.0.2/java.util.regex.Pattern$GroupHead.match(Pattern.java:4969)
	at java.base@22.0.2/java.util.regex.Pattern$Start.match(Pattern.java:3787)
	at java.base@22.0.2/java.util.regex.Matcher.search(Matcher.java:1765)
	at java.base@22.0.2/java.util.regex.Matcher.find(Matcher.java:785)
	at java.base@22.0.2/java.util.regex.Matcher.replaceAll(Matcher.java:1228)
	at java.base@22.0.2/java.lang.String.replaceAll(String.java:3111)
	at org.eclipse.swt.widgets.Display.withCrLf(Display.java:5159)
	at org.eclipse.swt.widgets.Text.setText(Text.java:2305)
	at org.eclipse.ui.internal.about.AboutSystemPage.lambda$2(AboutSystemPage.java:181)
	at org.eclipse.ui.internal.about.AboutSystemPage$$Lambda/0x000001bb65f30a88.run(Unknown Source)
	at org.eclipse.swt.widgets.RunnableLock.run(RunnableLock.java:40)
[...]

PR #1722, opened 2025-01-14
[FIX] Change to old withCrLf for better performance

Christopher-Hermann:
Changing back to the old withCrLf coding. Regex is causing performance issues.

Furthermore, handling of mixed \n and \r\n is improved.

Fixes: #1718

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
@@ -5145,16 +5145,65 @@ String wrapText (String text, long handle, int width) {
 }
 
 static String withCrLf (String string) {
-	if (string == null) {
-		return string;
+	/* Create a new string with the CR/LF line terminator. */
+	int i = 0;	
+	int length = string.length();
+	StringBuilder result = new StringBuilder (length);
+	while (i < length) {
+		int j = string.indexOf ('\n', i);
+		if (j > 0 && string.charAt(j - 1) == '\r') {
+			result.append(string.substring(i, j + 1));
+			i = j + 1;
+		} else {
+			if (j == -1) j = length;
+			result.append (string.substring (i, j));
+			if ((i = j) < length) {
+				result.append ("\r\n"); //$NON-NLS-1$
+				i++;
+			}
+		}
 	}
-	// Replace \r\n, \r, or \n with \r\n
-	return string.replaceAll("(\r\n|\r|\n)", "\r\n");
+	
+	/* Avoid creating a copy of the string if it has not changed */
+	if (string.length()== result.length()) return string;
+	return result.toString ();
 }
 
 static char [] withCrLf (char [] string) {
-	String withCrLf = withCrLf(new String(string));
-	return withCrLf.toCharArray();
+	/* If the string is empty, return the string. */
+	int length = string.length;
+	if (length == 0) return string;
+
+	/*
+	* Check for an LF or CR/LF and assume the rest of
+	* the string is formated that way.  This will not
+	* work if the string contains mixed delimiters.
+	* Also, compute the number of lines.
+	*/
+	int count = 0;
+	for (int i = 0; i < string.length; i++) {
+		if (string [i] == '\n') {
+			count++;
+			if (count == 1 && i > 0 && string [i - 1] == '\r') return string;
+		}
+	}
+	if (count == 0) return string;
+
+	/*
+	* The string is formatted with LF.
+	*/
+	count += length;
+
+	/* Create a new string with the CR/LF line terminator. */
+	char [] result = new char [count];
+	for (int i = 0, j = 0; i < length && j < count; i++) {
+		if (string [i] == '\n') {
+			result [j++] = '\r';
+		}
+		result [j++] = string [i];
+	}
+
+	return result;
 }
 
 static boolean isActivateShellOnForceFocus() {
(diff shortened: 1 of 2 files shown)


jukzi:
reduces time from 9 to 4 sec :-)

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
```diff
@@ -5152,16 +5152,90 @@ String wrapText (String text, long handle, int width) {
...
+	int i = string.indexOf ('\n', 0);
+	if (i == -1) return string;
+	if (i > 0 && string.charAt (i - 1) == '\r') {
 		return string;
 	}
-	// Replace \r\n, \r, or \n with \r\n
-	return string.replaceAll("(\r\n|\r|\n)", "\r\n");
+
+	/*
+	* The string is formatted with LF.  Compute the
```
jukzi, line 5172: this computation is needles

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
```diff
@@ -5152,16 +5152,90 @@ String wrapText (String text, long handle, int width) {
...
+	int count = 1;
+	while (i < length) {
+		if ((i = string.indexOf ('\n', i)) == -1) break;
+		count++; i++;
+	}
+	count += length;
+
+	/* Create a new string with the CR/LF line terminator. */
+	i = 0;
+	StringBuilder result = new StringBuilder (count);
```
jukzi, line 5186:
```suggestion
	StringBuilder result = new StringBuilder (string.length());
``` -> 3.4 s

File: bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java [hunk 3]
```diff
@@ -5152,16 +5152,90 @@ String wrapText (String text, long handle, int width) {
 }
 
 static String withCrLf (String string) {
-	if (string == null) {
+
+	/* If the string is empty, return the string. */
```
jukzi, line 5156: Also a premature optimization that could be left away

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
```diff
@@ -5152,16 +5152,90 @@ String wrapText (String text, long handle, int width) {
...
+		} else {
+			if (j == -1) j = length;
+			result.append (string.substring (i, j));
+			if ((i = j) < length) {
+				result.append ("\r\n"); //$NON-NLS-1$
+				i++;
+			}
+		}
+	}
+	return result.toString ();
```
jukzi, line 5201: should be save to avoid a copy with `if (string.length== result.length()) return string;`

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
```diff
@@ -5152,16 +5152,90 @@ String wrapText (String text, long handle, int width) {
...
 
 static String withCrLf (String string) {
-	if (string == null) {
+
+	/* If the string is empty, return the string. */
+	int length = string.length ();
+	if (length == 0) return string;
+
+	/*
+	* Check for an LF or CR/LF and assume the rest of
```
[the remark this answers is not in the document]
Christopher-Hermann: Not sure about that one. If there is a really long text with correct line breaks (\r\n) this will save some time. For example `withCrLf("\r\n".repeat(100_000_000));`. 

On the other hand, as you said, this is not working with partly mixed delimiters. So I'm not really sure how to proceed here.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
Christopher-Hermann: Done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
Christopher-Hermann: Done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
Christopher-Hermann: Done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
Christopher-Hermann: Done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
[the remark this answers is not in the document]
jukzi: > Not sure about that one. If there is a really long text with correct line breaks (\r\n) this will save some time.

did  you measure it? Anyhow Functional correctness beats performance requirements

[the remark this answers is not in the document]
Christopher-Hermann: Using `withCrLf("\r\n".repeat(100_000_000));` results in a noticeable performance difference: the operation takes 13ms with the check and 1382ms without it. Given that most strings passed to Display.withCrLf are likely already correctly formatted with proper line breaks, this optimization could significantly enhance the overall performance of Eclipse.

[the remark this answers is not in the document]
jukzi: > 13ms with the check and 1382ms without it.

that sounds like a big difference, but its only that fast because the check stops at the first occurrence. the following handling of the String however will still need to walk through the entire string so that in the broader context it may not be a hotspot but negligible

jukzi (approved):

Christopher-Hermann:
Test seems to be unrelated, but I'm not sure...

Issue #1546, opened 2024-10-21
Eclipse having problem with dark theme on newer Windows 11 24H2


provision95:
**Describe the bug**
Eclipse shows errors, stating that Not properly disposed SWT resource and having problems with css. I believe due to this dark theme does not work properly now on newest Windows.

**To Reproduce**
Update to Windows 11 24H2 (OSBuild 26100.2033)
Open Eclipse(latest).
Use Dark Mode.

**Expected behavior**
Expected a proper working dark theme, as it used to be.

**Screenshots**
![image](https://github.com/user-attachments/assets/37bbf877-9063-4d2a-b8dc-c90ba6550945)
![image](https://github.com/user-attachments/assets/5a22133c-d8aa-43f6-a1db-aa44d83a8d2a)
![image](https://github.com/user-attachments/assets/738cea6e-bf0f-476f-9860-22a6ef810f2f)

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 

2. Additional OS info (e.g. OS version, Linux Desktop, etc)
- Window 11 Pro 24H2 (OSBuild 26100.2033)
- Eclipse IDE for Enterprise Java and Web Developers (includes Incubating components)
Version: 2024-09 (4.33.0)
Build id: 20240905-0614

4. JRE/JDK version
org.eclipse.justj.openjdk.hotspot.jre.full.win32.x86_64_23.0.0.v20240919-1706\jre
 
**Version since**
 Eclipse or SWT version since when the behavior is seen [e.g. 4.23]
4.33.0

**Workaround (or) Additional context**
Any known workarounds for the problem?

Did not find any other than using default white theme.

BeckerWdf:
> Eclipse shows errors, stating that Not properly disposed SWT resource and having problems with css.

Can you pls. provided these errors (from the error log for example).

> Expected a proper working dark theme, as it used to be.
So what exactly is wrong on the screenshot? The light context menu?

Phillipus:
I can see light context menus, scroll bars and drop-downs.

Perhaps files like [this one](https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/bundles/org.eclipse.swt/Eclipse%20SWT%20PI/win32/library/os_custom.c) need to be updated?

HeikoKlare:
[...]

PR #1725, opened 2025-01-14
Backport Fix AllowDarkModeForWindowWithTelemetryId to 4.31 maintenance branch

elsazac:
Back port fix : https://github.com/eclipse-platform/eclipse.platform.swt/issues/1546

clipse shows errors, stating that Not properly disposed SWT resource and having problems with css. I believe due to this dark theme does not work properly now on newest Windows.

Fix : https://github.com/eclipse-platform/eclipse.platform.swt/pull/1547Win32
Update search for AllowDarkModeForWindowWithTelemetryId for 24H2


bundles/org.eclipse.swt/META-INF/MANIFEST.MF
MohananRahul, line 1: https://github.com/eclipse-platform/eclipse.platform.swt/blob/R4_31_maintenance/bundles/org.eclipse.swt/pom.xml also needs +1

Issue #1714, opened 2025-01-03
IllegalArgumentException in HTMLTextPresenter during quickfix


jukzi:
don't know how it happend, found in the logfile:

```
eclipse.buildId=4.35.0.I20250102-1800
java.version=22.0.2
java.vendor=Eclipse Adoptium
BootLoader constants: OS=win32, ARCH=x86_64, WS=win32, NL=de_DE
Command-line arguments:  -os win32 -ws win32 -arch x86_64 -data C:\Users\jkubitz\platform-2025-01-03_2\ws

org.eclipse.ui
Error
Fri Jan 03 11:11:00 CET 2025
Unhandled event loop exception

java.lang.IllegalArgumentException: Argument not valid
	at org.eclipse.swt.SWT.error(SWT.java:4913)
	at org.eclipse.swt.SWT.error(SWT.java:4847)
	at org.eclipse.swt.SWT.error(SWT.java:4818)
	at org.eclipse.swt.graphics.TextStyle.<init>(TextStyle.java:189)
	at org.eclipse.swt.custom.StyleRange.<init>(StyleRange.java:68)
	at org.eclipse.jface.internal.text.html.HTMLTextPresenter.updatePresentation(HTMLTextPresenter.java:170)
	at org.eclipse.jface.text.DefaultInformationControl.setInformation(DefaultInformationControl.java:341)
	at org.eclipse.jface.text.AbstractInformationControlManager.internalShowInformationControl(AbstractInformationControlManager.java:1153)
	at org.eclipse.jface.text.AbstractInformationControlManager.presentInformation(AbstractInformationControlManager.java:1120)
	at org.eclipse.jface.text.AbstractInformationControlManager.setInformation(AbstractInformationControlManager.java:431)
	at org.eclipse.jface.text.contentassist.AdditionalInfoController.computeInformation(AdditionalInfoController.java:567)
	at org.eclipse.jface.text.AbstractInformationControlManager.doShowInformation(AbstractInformationControlManager.java:1101)
	at org.eclipse.jface.text.AbstractInformationControlManager.showInformation(AbstractInformationControlManager.java:1091)
	at org.eclipse.jface.text.contentassist.AdditionalInfoController.showInformation(AdditionalInfoController.java:541)
	at org.eclipse.jface.text.contentassist.AdditionalInfoController$1.showInformation(AdditionalInfoController.java:478)
[...]

PR #1732, opened 2025-01-16
Revert "Corrected TextLayout to produce right number of lines when '\r\n' sequence comes."

jukzi:
Reverts eclipse-platform/eclipse.platform.swt#1320

IllegalArgumentException in HTMLTextPresenter during quickfix
https://github.com/eclipse-platform/eclipse.platform.swt/issues/1714


jukzi:
cc @deepika, ok to revert until you found a solution?

jukzi:
@eclipse-platform/eclipse-platform-project-leads please revert regression

HeikoKlare (commented):
@deepika-u will you be able to propose a fix for #1714 today?
Otherwise, reverting the change producing the regressions for M2 is, in my opinion, the right thing to do. Even though it unfortunately introduces the fixed issue #184 again (which should be reopened after merging this revert).

deepika-u:
Yeah please go ahead with a revert. I have a fix but need some more time to test and finalize on the fix before it is proposed as a pr.

PR #1734, opened 2025-01-16
Callback for dynamic drawing with a GC on images

akoch-yatta:
This PR adds a new interface to be used to provide dynamic drawing on GCs created with Images. It is mostly targeted at solving the following use cases:

```
Image image = new Image (display, width, height);
GC gc = new GC (image);
// draw on the gc
otherGc.drawImage(image, ...);
```
or in combination with a ImageDataProvider
```
return new Image(display, (ImageDataProvider) zoom -> {
     Image image = new Image (display, width, height);
     GC gc = new GC (image);
     // draw on the gc
     gc.dispose ();

     final ImageData imageData = image.getImageData (zoom);
     image.dispose();
     return imageData;
});
```

Both use cases have the problem, that they do not work properly in a multi-zoom setting. Not properly working means that there will be a destructive up- or downscaling on the created image data in the background. This PR moves the initialization of the GC into the implementation of each image and can therefor be handled properly for all platforms.

I saw those use cases with two different settings: with and without transparency necessary. To address the use case with transparency we added the TransparancyColorImageGcDrawer as internal class to be used for CTabFolder::createButtonImage.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/ImageGcDrawer.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/ImageGcDrawer.java
@@ -1,0 +1,48 @@
+/*******************************************************************************
+ * Copyright (c) 2025 Yatta and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Yatta - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.swt.graphics;
+
+/**
+ * Interface to provide a callback mechanism to draw on different GC instances
+ * depending on the zoom the image will be used for. A common use case is when
+ * the application is moved from a low DPI monitor to a high DPI monitor. This
+ * provides API which will be called by SWT during the image rendering.
+ *
+ * This interface needs to be implemented by client code to provide logic that
+ * draws on the empty GC on demand.
+ *
+ * @since 3.129
+ */
+public interface ImageGcDrawer {
+
+	/**
+	 * Draws an image on a GC for a requested zoom level.
+	 *
+	 * @param gc          The GC will draw on the underlying Image and is configured
+	 *                    for the targeted zoom
+	 * @param imageWidth  The width of the image in points to draw on
+	 * @param imageHeight The height of the image in points to draw on
+	 */
+	void drawOn(GC gc, int imageWidth, int imageHeight);
+
+	/**
+	 * Executes post processing on ImageData. This method will always be called
+	 * after <code>drawOn</code> and contain the resulting ImageData.
+	 *
+	 * @param imageData The resulting ImageData after <code>drawOn</code> was called
+	 */
+	default void postProcess(ImageData imageData) {
+	}
+
+}
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/TransparencyColorImageGcDrawer.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/TransparencyColorImageGcDrawer.java
@@ -1,0 +1,31 @@
+/*******************************************************************************
+ * Copyright (c) 2025 Yatta and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Yatta - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.swt.internal;
+
+import org.eclipse.swt.graphics.*;
+
+public abstract class TransparencyColorImageGcDrawer implements ImageGcDrawer {
+
+	private final Color transparencyColor;
+
+	public TransparencyColorImageGcDrawer(Color transparencyColor) {
+		this.transparencyColor = transparencyColor;
+	}
+
+	@Override
+	public void postProcess(ImageData imageData) {
+		imageData.transparentPixel = imageData.palette.getPixel(transparencyColor.getRGB());
+	}
+
+}
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Image.java
@@ -15,6 +15,7 @@ package org.eclipse.swt.graphics;
 
 
 import java.io.*;
+import java.util.*;
 
 import org.eclipse.swt.*;
 import org.eclipse.swt.internal.*;
@@ -152,6 +153,11 @@ public final class Image extends Resource implements Drawable {
 	private ImageDataProvider imageDataProvider;
 
 	/**
+	 * ImageGcDrawer to provide a callback to draw on a GC for various zoom levels
+	 */
+	private ImageGcDrawer imageGcDrawer;
+
+	/**
 	 * Style flag used to differentiate normal, gray-scale and disabled images based
 	 * on image data providers. Without this, a normal and a disabled image of the
 	 * same image data provider would be considered equal.
@@ -263,6 +269,7 @@ public Image(Device device, Image srcImage, int flag) {
 	this.type = srcImage.type;
 	this.imageDataProvider = srcImage.imageDataProvider;
 	this.imageFileNameProvider = srcImage.imageFileNameProvider;
+	this.imageGcDrawer = srcImage.imageGcDrawer;
 	this.styleFlag = srcImage.styleFlag | flag;
 	this.currentDeviceZoom = srcImage.currentDeviceZoom;
 
@@ -662,6 +669,36 @@ public Image(Device device, ImageDataProvider imageDataProvider) {
 }
 
 /**
+ * The provided ImageGcDrawer will be called on demand whenever a new variant of the
+ * Image for an additional zoom is required. Depending on the OS specific implementation
+ * these calls will be done during the instantiation or later when a new variant is
+ * requested.
+ *
+ * @param device the device on which to create the image
+ * @param imageGcDrawer the ImageGcDrawer object to be called when a new image variant
+ * for another zoom is required.
+ * @param width the width of the new image in points
+ * @param height the height of the new image in points
+ *
+ * @exception IllegalArgumentException <ul>
+ *    <li>ERROR_NULL_ARGUMENT - if device is null and there is no current device</li>
+ *    <li>ERROR_NULL_ARGUMENT - if the ImageGcDrawer is null</li>
+ * </ul>
+ * @since 3.129
+ */
+public Image(Device device, ImageGcDrawer imageGcDrawer, int width, int height) {
+	super(device);
+	if (imageGcDrawer == null) {
+		SWT.error(SWT.ERROR_NULL_ARGUMENT);
+	}
+	this.imageGcDrawer = imageGcDrawer;
+	currentDeviceZoom = DPIUtil.getDeviceZoom();
+	ImageData imageData = drawWithImageGcDrawer(width, height, currentDeviceZoom);
+	init (imageData);
+	init ();
+}
+
+/**
  * Refreshes the image for the current device scale factor.
  * <p>
  * <b>IMPORTANT:</b> This function is <em>not</em> part of the SWT
@@ -718,6 +755,17 @@ boolean refreshImageForZoom () {
 			destroy ();
 			ImageData resizedData = DPIUtil.autoScaleImageData (device, data.element(), data.zoom());
 			init(resizedData);
+			init();
+			refreshed = true;
+			currentDeviceZoom = deviceZoomLevel;
+		}
+	} else if (imageGcDrawer != null) {
+		int deviceZoomLevel = deviceZoom;
+		if (deviceZoomLevel != currentDeviceZoom) {
+			ImageData data = drawWithImageGcDrawer(width, height, deviceZoomLevel);
+			/* Release current native resources */
+			destroy ();
+			init(data);
 			init();
 			refreshed = true;
 			currentDeviceZoom = deviceZoomLevel;
@@ -904,6 +952,9 @@ public boolean equals (Object object) {
 		return (styleFlag == image.styleFlag) && imageDataProvider.equals (image.imageDataProvider);
 	} else if (imageFileNameProvider != null && image.imageFileNameProvider != null) {
 		return (styleFlag == image.styleFlag) && imageFileNameProvider.equals (image.imageFileNameProvider);
+	} else if (imageGcDrawer != null && image.imageGcDrawer != null) {
+		return styleFlag == image.styleFlag && imageGcDrawer.equals(image.imageGcDrawer) && width == image.width
+				&& height == image.height;
 	} else {
 		return surface == image.surface;
 	}
@@ -1110,8 +1161,24 @@ public ImageData getImageData (int zoom) {
 	} else if (imageFileNameProvider != null) {
 		ElementAtZoom<String> fileName = DPIUtil.validateAndGetImagePathAtZoom (imageFileNameProvider, zoom);
 		return DPIUtil.scaleImageData (device, new ImageData (fileName.element()), zoom, fileName.zoom());
+	} else if (imageGcDrawer != null) {
+		return drawWithImageGcDrawer(width, height, zoom);
 	} else {
 		return DPIUtil.scaleImageData (device, getImageDataAtCurrentZoom (), zoom, currentDeviceZoom);
+	}
+}
+
+private ImageData drawWithImageGcDrawer(int width, int height, int zoom) {
+	Image image = new Image(device, width, height);
+	GC gc = new GC(image);
+	try {
+		imageGcDrawer.drawOn(gc, width, height);
+		ImageData imageData = image.getImageData(zoom);
+		imageGcDrawer.postProcess(imageData);
+		return imageData;
+	} finally {
+		gc.dispose();
+		image.dispose();
 	}
 }
 
@@ -1179,6 +1246,8 @@ public int hashCode () {
 		return imageDataProvider.hashCode();
 	} else if (imageFileNameProvider != null) {
 		return imageFileNameProvider.hashCode();
+	} else if (imageGcDrawer != null) {
+		return Objects.hash(imageGcDrawer, width, height);
 	} else {
 		return (int)surface;
 	}
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
@@ -172,8 +172,13 @@ private Image (Device device, int nativeZoom) {
  * @see #dispose()
  */
 public Image(Device device, int width, int height) {
+	this(device, width, height, DPIUtil.getNativeDeviceZoom());
+}
+
+
+private Image(Device device, int width, int height, int nativeZoom) {
 	super(device);
-	initialNativeZoom = DPIUtil.getNativeDeviceZoom();
+	initialNativeZoom = nativeZoom;
 	final int zoom = getZoom();
 	width = DPIUtil.scaleUp (width, zoom);
 	height = DPIUtil.scaleUp (height, zoom);
@@ -600,6 +605,31 @@ public Image(Device device, ImageDataProvider imageDataProvider) {
 	init (resizedData, getZoom());
 	init();
 	this.device.registerResourceWithZoomSupport(this);
+}
+
+/**
+ * The provided ImageGcDrawer will be called on demand whenever a new variant of the
+ * Image for an additional zoom is required. Depending on the OS-specific implementation
+ * these calls will be done during the instantiation or later when a new variant is
+ * requested.
+ *
+ * @param device the device on which to create the image
+ * @param imageGcDrawer the ImageGcDrawer object to be called when a new image variant
+ * for another zoom is required.
+ * @param width the width of the new image in points
+ * @param height the height of the new image in points
+ *
+ * @exception IllegalArgumentException <ul>
+ *    <li>ERROR_NULL_ARGUMENT - if device is null and there is no current device</li>
+ *    <li>ERROR_NULL_ARGUMENT - if the ImageGcDrawer is null</li>
+ * </ul>
+ * @since 3.129
+ */
+public Image(Device device, ImageGcDrawer imageGcDrawer, int width, int height) {
+	super(device);
+	this.imageProvider = new ImageGcDrawerWrapper(imageGcDrawer, width, height);
+	initialNativeZoom = DPIUtil.getNativeDeviceZoom();
+	init();
 }
 
 private ImageData adaptImageDataIfDisabledOrGray(ImageData data) {
@@ -1140,6 +1170,9 @@ long [] createGdipImage(Integer zoom) {
 void destroy () {
 	device.deregisterResourceWithZoomSupport(this);
 	if (memGC != null) memGC.dispose();
+	if (this.imageProvider != null) {
+		this.imageProvider.destroy();
+	}
 	destroyHandle();
 	memGC = null;
 }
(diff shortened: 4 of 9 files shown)


bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/ImageGcDrawer.java
```diff
@@ -0,0 +1,50 @@
...
+ *******************************************************************************/
+package org.eclipse.swt.graphics;
+
+/**
+ * Interface to provide a callback mechanism to draw on different GC instances
+ * depending on the zoom the image will be used for. A common use case is when the
+ * application is moved from a low DPI monitor to a high DPI monitor.
+ * This provides API which will be called by SWT during the image rendering.
+ *
+ * This interface needs to be implemented by client code to private logic that draws
```
HeikoKlare, line 22:
```suggestion
 * This interface needs to be implemented by client code to provide logic that draws
```

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/ImageGcDrawer.java
```diff
@@ -0,0 +1,50 @@
...
+ * on the empty GC on demand.
+ *
+ * @since 3.129
+ */
+public interface ImageGcDrawer {
+
+
```
HeikoKlare, line 29:
```suggestion

```

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/ImageGcDrawer.java
```diff
@@ -0,0 +1,50 @@
...
+	/**
+	 * Provides a GC to draw on for a requested zoom level.
```
HeikoKlare, line 31: This does not describe the method, but what an implementor of the method has to do. The method itself receives a GC and (the implementation) draw on it to provide an image for the requested zoom.
```suggestion
	 * Draws an image on a GC for a requested zoom level.
```

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/ImageGcDrawer.java
```diff
@@ -0,0 +1,50 @@
...
+	 * <p>
```
HeikoKlare, line 32:
```suggestion
```

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/ImageGcDrawer.java
```diff
@@ -0,0 +1,50 @@
...
+	 * @param gc
+	 *            The GC will draw on the underlying Image and is configured for the targeted zoom
+	 * @since 3.129
+	 */
+	void drawOn(GC gc);
+
+	/**
+	 * Implement this method if any post processing of the ImageData created by the operations on the
+	 * GC in <code>drawOn</code> is necessary.
+	 * <p>
```
HeikoKlare, line 43:
```suggestion
```

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/ImageGcDrawer.java
```diff
@@ -0,0 +1,50 @@
...
+	 *
+	 * @param gc
+	 *            The GC will draw on the underlying Image and is configured for the targeted zoom
+	 * @since 3.129
+	 */
+	void drawOn(GC gc);
+
+	/**
+	 * Implement this method if any post processing of the ImageData created by the operations on the
+	 * GC in <code>drawOn</code> is necessary.
```
HeikoKlare, line 42: See above

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/TransparancyColorImageGcDrawer.java
```diff
@@ -0,0 +1,18 @@
+package org.eclipse.swt.internal;
```
HeikoKlare, line 1: Copyright header missing

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_Image.java
```diff
@@ -675,6 +694,22 @@ public void test_equalsLjava_lang_Object() {
...
+	try {
+		image = new Image(display, imageGcDrawer, 10, 10);
+		image1 = image;
+
+		assertFalse(":g:", image.equals(null));
+
+		assertTrue(":h:", image.equals(image1));
+
+		image1 = new Image(display, imageGcDrawer, 10, 10);
+		assertTrue(":i:", image.equals(image1));
```
HeikoKlare, line 708: Can we leave out the assertion messages? They don't seem to provide any benefit. They may have been used to identify the failing assertion before, but having line number in failures nowadays that should not be necessary. Here, they are also copied over from the try/finnyl before, so messages are not unique for the method.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_Image.java
```diff
@@ -675,6 +694,22 @@ public void test_equalsLjava_lang_Object() {
 		image.dispose();
 		image1.dispose();
 	}
+
+	// ImageDataProvider
```
HeikoKlare, line 698:
```suggestion
	// ImageGcDrawer
```

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_Image.java
```diff
@@ -760,6 +795,13 @@ public void test_getBoundsInPixels() {
 	bounds = image.getBounds();
 	image.dispose();
 	assertEquals(":d: Image.getBoundsInPixels method doesn't return bounds in Pixel values.", boundsInPixels, DPIUtil.autoScaleUp(bounds));
+
+	// create image with ImageGcDrawer
+	image = new Image(display, imageGcDrawer, bounds.width, bounds.height);
+	boundsInPixels = image.getBoundsInPixels();
+	bounds = image.getBounds();
+	image.dispose();
+	assertEquals(":d: Image.getBoundsInPixels method doesn't return bounds in Pixel values.", boundsInPixels, DPIUtil.autoScaleUp(bounds));
```
HeikoKlare, line 804:
```suggestion
	assertEquals("Image.getBoundsInPixels method doesn't return bounds in Pixel values for ImageGcDrawer.", boundsInPixels, DPIUtil.autoScaleUp(bounds));
```

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_Image.java
```diff
@@ -968,6 +1010,16 @@ public void test_hashCode() {
 		image.dispose();
 		image1.dispose();
 	}
+
+	// ImageGcDrawer
+	try {
+		image = new Image(display, imageGcDrawer, 10, 10);
+		image1 = new Image(display, imageGcDrawer, 10, 10);
+		assertEquals(":d:", image1.hashCode(), image.hashCode());
```
HeikoKlare, line 1018: See above

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -602,6 +607,32 @@ public Image(Device device, ImageDataProvider imageDataProvider) {
 	this.device.registerResourceWithZoomSupport(this);
 }
 
+/**
+ * The provided ImageGcDrawer will be called on demand whenever a new variant of the
+ * Image for an additional zoom is required. Depending on the OS specific implementation
+ * these calls will be done during the instantiation or later when a new variant is
+ * requested
+ * <p>
```
HeikoKlare, line 615:
```suggestion
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2076,6 +2121,13 @@ private class ImageFileNameProviderWrapper extends AbstractImageProviderWrapper
 		this.provider = provider;
 	}
 
+	@Override
+	protected Rectangle getBounds(int zoom) {
+		ImageHandle imageHandle = zoomLevelToImageHandle.values().iterator().next();
+		Rectangle rectangle = new Rectangle(0, 0, imageHandle.width, imageHandle.height);
+		return DPIUtil.scaleBounds(rectangle, zoom, imageHandle.zoom);
```
HeikoKlare, line 2128: This code seems to be replacated 3 times. Could it be put into a `getBounds()` method of `ImageHandle`?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2154,6 +2223,70 @@ ImageDataProviderWrapper createCopy(Image image) {
 	}
 }
 
+private class ImageGcDrawerWrapper extends AbstractImageProviderWrapper {
+	private ImageGcDrawer drawer;
+	private int width;
+	private int height;
+	private boolean isDestroyed = false;
```
HeikoKlare, line 2230: Wouldn't it make sense to store this directly in Image? It seems to be as if it just used as a flag whether `dipose()` was called for the image containing this wrapper.

HeikoKlare (commented):
The proposed API and the according changes look reasonable to me.
I have so far taken a look at the implementation excluding the Image adaptations for MacOS and Linux.
I will also have a look at the other implementations afterwards and of course need to test the change.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: I don't think it works without refactoring more. I would like to have a more cleaner Hierarchy Image -> AbstractImageProviderWrapper -> ImageHandle in the relation, that will lead to something like you propose, but for that we need to extract an AbstractImageProviderWrapper  for the non-dynamic case. So I would postpone this,

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/ImageGcDrawer.java
akoch-yatta: done

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/ImageGcDrawer.java
akoch-yatta: done

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/ImageGcDrawer.java
akoch-yatta: done

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/ImageGcDrawer.java
akoch-yatta: done

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/TransparancyColorImageGcDrawer.java
akoch-yatta: done

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/ImageGcDrawer.java
akoch-yatta: probably done

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_Image.java
akoch-yatta: sure, I remove them, where they do not contain relevant information

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_Image.java
akoch-yatta: done

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_Image.java
akoch-yatta: adapted

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: Hmm, correct. But what would you do then. Use this flag always? The reason why I introduced it for now is, that the handles for this new interface are always created on demand, so it would be detected as disposed from the beginning, which is obviously not good. I didn't find a better solution for now to address as to introduce this flag. But therefor I would like to have it only an affect in this subclass for now.

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet382.java
```diff
@@ -96,6 +108,10 @@ public void handleEvent(Event e)  {
 						drawImages(mainGC, gcData, "Normal",280, imageWithDataProvider);
 						drawImages(mainGC, gcData, "Disabled",320, disabledImageWithData);
 						drawImages(mainGC, gcData, "Greyed",360, greyImageWithData);
+
+						drawImages(mainGC, gcData, "Normal", 400, imageWithGcDrawer);
+						drawImages(mainGC, gcData, "Disabled", 440, disabledImageWithGcDrawer);
+						drawImages(mainGC, gcData, "Greyed", 480, greyImageWithGcDrawer);
```
HeikoKlare, line 114: Nit: these additional images increase the vertical size of the shell contents, so the current size (set to a fixed value at the end of the main method) is insufficient. Increasing height to 550 works fine for me.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2154,6 +2222,70 @@ ImageDataProviderWrapper createCopy(Image image) {
 	}
 }
 
+private class ImageGcDrawerWrapper extends AbstractImageProviderWrapper {
```
HeikoKlare, line 2225: Should we overwrite equals/hashcode here to also consider width/height? The superclass only bases those methods on the provider (i.e., the `ImageGcDrawer` in this case).

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -602,6 +607,31 @@ public Image(Device device, ImageDataProvider imageDataProvider) {
 	this.device.registerResourceWithZoomSupport(this);
 }
 
+/**
+ * The provided ImageGcDrawer will be called on demand whenever a new variant of the
+ * Image for an additional zoom is required. Depending on the OS specific implementation
```
HeikoKlare, line 612:
```suggestion
 * Image for an additional zoom is required. Depending on the OS-specific implementation
```

HeikoKlare (commented):
I've now tested the change on Windows and MacOS, also together with the consuming Platform UI PR https://github.com/eclipse-platform/eclipse.platform.ui/pull/2722.
Both platform implementations work fine. Initializing the UI with existing (non-monitor-specific) scaling capabilities at different zoom levels (100%, 200% and on Windows in between) shows behavior that is consistent to before the change.
Enabling monitor-specific scaling on Windows shows improved behavior. I particularly tested with higher resolution primary monitor and lower resolution secondary monitor, moving between these monitors and changing the primary monitor zoom at runtime.

I have reviewed again after recent changes and only have nitpicky comments. I still need to have a look at and test the Linux implementation.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: done

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet382.java
akoch-yatta: good point, done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2154,6 +2222,70 @@ ImageDataProviderWrapper createCopy(Image image) {
 	}
 }
 
+private class ImageGcDrawerWrapper extends AbstractImageProviderWrapper {
+	private ImageGcDrawer drawer;
+	private int width;
+	private int height;
+	private boolean isDestroyed = false;
```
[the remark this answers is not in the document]
akoch-yatta: done

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/TransparancyColorImageGcDrawer.java
```diff
@@ -0,0 +1,31 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Yatta - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.swt.internal;
+
+import org.eclipse.swt.graphics.*;
+
+public abstract class TransparancyColorImageGcDrawer implements ImageGcDrawer {
```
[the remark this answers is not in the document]
akoch-yatta: Ah, damn, done

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/TransparancyColorImageGcDrawer.java
```diff
@@ -0,0 +1,31 @@
...
+
+	private final Color transparancyColor;
```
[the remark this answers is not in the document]
akoch-yatta: done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: Hmm, thought about this as well, although it doesn't make much sense to use the same Drawer with different width height values. I am currently thinking about whether it makes more sense to use the handle for the equals/hashCode. Does the same drawer really should return in two images being equal?

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/TransparencyColorImageGcDrawer.java
```diff
@@ -0,0 +1,31 @@
...
+ *******************************************************************************/
+package org.eclipse.swt.internal;
+
+import org.eclipse.swt.graphics.*;
+
+public abstract class TransparencyColorImageGcDrawer implements ImageGcDrawer {
+
+	private final Color transparencyColor;
+
+	public TransparencyColorImageGcDrawer(Color transparancyColor) {
```
HeikoKlare, line 22: Still one leftover :-) 
```suggestion
	public TransparencyColorImageGcDrawer(Color transparencyColor) {
```

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Image.java
```diff
@@ -1110,11 +1160,33 @@ public ImageData getImageData (int zoom) {
...
+			imageGcDrawer.drawOn(gc);
+			ImageData imageData = image.getImageData(zoom);
+			imageGcDrawer.postProcess(imageData);
+			return imageData;
+		} finally {
+			gc.dispose();
+			image.dispose();
+		}
+	}
+	SWT.error(SWT.ERROR_INVALID_ARGUMENT, null, ": ImageGcDrawer [" + imageGcDrawer + "] is null.");
```
HeikoKlare, line 1186: I do not see any comparable way of throwing an error because of a null provider in the GTK implementation of Image and, if I don't miss anything, this line of code can never be reached anyway, as all callers properly ensure that the imageGcDrawer is null. So I would propose to complete remove the if block and have the drawer being null a precondition for caling that method.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
HeikoKlare: Good question. For the sake of consistency, I propose to keep it aligned with all the other providers that can be passed to the Image class, which do also not use the handle but simply the provider. That also makes sense to me, as initializing two images with the same provider should yield the same results for the same method call on the two of them (in particular, return the same image data for a given zoom).
Given that an ImageGcDrawer does not make sense for different width/height values, it means that it is actually tied to a specific width/height, isn't it? Shouldn't the drawer than maybe also provide its size (in points) via some `getSize()` method and use that in the Image (and maybe also in its own `drawOn()` method) instead of passing width and height additionally, but requiring it to conform to the actual implementation of `drawOn()`?

HeikoKlare (commented):
I've now also taken a look at the GTK code and only have minor comments. I have amended the PR commit with the proposed changes. Please see if they are find and otherwise feel free to revert. Except for that (and potentially hashCode/equals), the PR is now fine for me.

I have tested the behavior on Linux/GTK (Ubuntu 24.04.1) and found no regression with respect to existing behavior. I tested with two monitors at different zoom values, both 100% and 200%, as well as different combinations of fractional scaling and changes between them. I found that together with https://github.com/eclipse-platform/eclipse.platform.ui/pull/2722 we now get the LineNumberRuler scaling updated on DPI changes on GTK.
Changing from 100% to 150% yields the following results:

### Before the changes
image

### After the changes
image

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
HeikoKlare: Adapted `drawOn()` signature as discussed offline with recent commit. Width/height are now passed to that method to not rely on the drawer using the same bounds as provided to the image instance to which the drawer is passed. hashCode/equals have been adapted accordingly.

HeikoKlare (approved):

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/ImageGcDrawer.java
```diff
@@ -0,0 +1,48 @@
...
+	 */
+	void drawOn(GC gc, int imageWidth, int imageHeight);
+
+	/**
+	 * Executes post processing on ImageData. This method will always be called
+	 * after <code>drawOn</code> and contain the resulting ImageData.
+	 *
+	 * @param imageData The resulting ImageData after <code>drawOn</code> was called
+	 */
+	default void postProcess(ImageData imageData) {
```
fedejeanne, line 45: Adding `default` methods to a public interface may cause issues in consumers. See https://github.com/eclipse-platform/eclipse.platform/blob/master/docs/Evolving-Java-based-APIs-2.md#evolving-api-interfaces

I only mention this because the name of the method is pretty generic, which increases the chances of other interface maybe declaring the same method and actually causing pre-existing binaries to throw a `IncompatibleClassChangeError ` (https://docs.oracle.com/javase/specs/jls/se8/html/jls-13.html#jls-13.5.6).

HeikoKlare: Adding a default method to an interface is indeed an API change that might introduced problems and thus has to be assessed properly. However, in this case we do not add a method to an interface but create a completely new interface. There are no implementations of that interface so far that may face any incompatibility.

When taking a look at API evolution, in this case we only need to consider "Add API type to API package", which is a proper change.

Do you see any other potential issues here?

fedejeanne: you're right, if someone wants to implement this new interface then the problem should arise immediately at compile time.

PR #1736, opened 2025-01-17
Use AutoScaleMethod "smooth" by default for fractional scaling

HeikoKlare:
Currently, AutoScaleMethod "nearest" is used by default, leading to jagged image scaling results. The method "smooth" is only used on GTK with a fractional scaling value not being a multiple of 100. The reason for not using "smooth" scaling on all platform was a bug regarding icon transparency that has long been resolved.

With this change, AutoScaleMethod "smooth" is used by default in fractional scaling scenarios (i.e., with deviceZoom not being an integer multiple of 100) and when using monitor-specific scaling on Windows. To still use "nearest" in these scenarios, it can be activated via system property.

The change has also been tested for GIFs that are only rarely used nowadays. In the screenshots below, the "JobFactory" view uses a GIF as its icon.

## Why now?

A better interpolation method was already asked for in: https://bugs.eclipse.org/bugs/show_bug.cgi?id=493455
 
The issue that previously prevented us from using "smooth" scaling on Windows have meanwhile been resolved:
- https://bugs.eclipse.org/bugs/show_bug.cgi?id=97506

There are existing workarounds, e.g., for `CompositeImageDescriptor`, which should be adapted in a follow-up step as well. See:
- https://bugs.eclipse.org/bugs/show_bug.cgi?id=495782

## Is this safe to do?

This change only taken effect when using any `swt.scaleMode` that supports fractional scaling, such as the mode `quarter`, `exact` or an explicit zoom value. None of them used as default value. The default behavior (known as "integer"/"integer200") will not be affected, as the deviceZoom will be multiples of 100.
So this change only affects the monitor-specific scaling mode on Windows (which is experimental anyway) and applications where custom configurations of `swt.autoScale` are in place anyway. In those situation, if the new image scaling behavior is undesired, `swt.autoScaleMethod` may be custom configured to "nearest" as well.
I propose to add an according information in the news for this.

## How does it look?

The following screenshots are taken on a 150% monitor with `swt.autoScale=quarter`.

### Before

image
image

### After
image
image

image
image

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
@@ -79,10 +79,9 @@ public class DPIUtil {
 	 * <li>"nearest": nearest-neighbor interpolation, may look jagged</li>
 	 * <li>"smooth": smooth edges, may look blurry</li>
 	 * </ul>
-	 * The current default is to use "nearest", except on
-	 * GTK when the deviceZoom is not an integer multiple of 100%.
-	 * The smooth strategy currently doesn't work on Win32 and Cocoa, see
-	 * <a href="https://bugs.eclipse.org/493455">bug 493455</a>.
+	 * The current default is to use "smooth" on GTK when deviceZoom is an integer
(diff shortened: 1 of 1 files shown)


akoch-yatta (approved):
I had a look into the code. The logic is easy and looks good to me.

PR #1737, opened 2025-01-19
Adapt Snippet367 and Snippet382 to ImageGcDrawer

HeikoKlare:
Two snippets refer to an old signature of the ImageGcDrawer interface, leading to compilation issues. This change change adapts them accordingly.


HeikoKlare:
Snippets are not compiled in CI anyway.

PR #1745, opened 2025-01-21
Using a flag to check if a DPI handling is needed on WM_DPICHANGED event

amartya4256:


amartya4256:
We chose to go with #1746

PR #1748, opened 2025-01-21
[GTK4] Fix GtkWidget warning when calculating initial shell bounds

ptziegler:
It might be possible that no (custom) titlebar is set for the newly created shell. In such a case, gtk_widget_measure() must not be called with the NULL handle, in order to avoid the following error:

> gtk_widget_measure: assertion 'GTK_IS_WIDGET (widget)' failed

Furthermore, the header is now retrieved via gtk_window_get_titlebar(), rather than assuming that it's the second child of the shell handle.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT PI/gtk/library/gtk4.c
+++ b/bundles/org.eclipse.swt/Eclipse SWT PI/gtk/library/gtk4.c
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2024 IBM Corporation and others.
+ * Copyright (c) 2000, 2025 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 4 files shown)


ptziegler:
This was originally added via https://bugs.eclipse.org/bugs/show_bug.cgi?id=573701 which sadly, doesn't give any indication under which circumstances the following applies:

- Retrieved GtkHeaderBar height & add it to the height given to
set_default_size in order to accommodate for the GtkHeaderBar's height
being included in the client area of GtkWindow

I've tested the an unmodified shell under both GTK3 and GTK4 and in both cases, the shells have a size of 900x590 pixels. Most noticeably, they don't have a custom header, which is instead contributed by the window manager. So perhaps this differs between e.g. KDE and Gnome? 
ShellGtk4

Regardless, I've tried setting a custom header, which then resulted in an initial size of 900x609 pixels. So whatever use case this was implemented for not break as a result of this.
ShellWithHeaderBar

akurtakov (approved):
Looks good to me.

PR #1752, opened 2025-01-22
WebViewWrapperFuture handling on job termination timeout

amartya4256:
This Pull Request introduces following behaviour in Edge browser:

- Encapsulates the ICoreWebView*::Release for all the different webView* instances in the WebViewWrapper in a method WebViewWrapper::releaseWebViews
- The futures complete exceptionally if there occurs a timeout at processOSMessagesUntil
- If the webViewWrapperFuture finishes exceptionally (during the initialization of webViews), there's a streamlined rollback of all the resources.

contributes to #1466 and [#2734](https://github.com/eclipse-platform/eclipse.platform.ui/issues/2734)

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
@@ -294,6 +294,33 @@ class WebViewWrapper {
 	private ICoreWebView2_11 webView_11;
 	private ICoreWebView2_12 webView_12;
 	private ICoreWebView2_13 webView_13;
+
+	void releaseWebViews() {
+		if(webView != null) {
+			webView.Release();
+			webView = null;
+		}
+		if(webView_2 != null) {
+			webView_2.Release();
+			webView_2 = null;
+		}
+		if(webView_10 != null) {
+			webView_10.Release();
+			webView_10 = null;
+		}
+		if(webView_11 != null) {
+			webView_11.Release();
+			webView_11 = null;
+		}
+		if(webView_12 != null) {
+			webView_12.Release();
+			webView_12 = null;
+		}
+		if(webView_13 != null) {
+			webView_13.Release();
+			webView_13 = null;
+		}
+	}
 }
 
 class WebViewProvider {
@@ -318,6 +345,10 @@ class WebViewProvider {
 
 	private void abortInitialization() {
 		webViewWrapperFuture.cancel(true);
+	}
+
+	void releaseWebView() {
+		getWebViewWrapper().releaseWebViews();
 	}
 
 	private ICoreWebView2_2 initializeWebView_2(ICoreWebView2 webView) {
@@ -748,14 +779,9 @@ void setupBrowser(int hr, long pv) {
 void browserDispose(Event event) {
 	containingEnvironment.instances.remove(this);
 	webViewProvider.scheduleWebViewTask(() -> {
-		webViewProvider.getWebView(false).Release();
+		webViewProvider.releaseWebView();
 		if (environment2 != null) environment2.Release();
 		if (settings != null) settings.Release();
-		if (webViewProvider.isWebView_2Available()) webViewProvider.getWebView_2(false).Release();
-		if (webViewProvider.isWebView_10Available()) webViewProvider.getWebView_10(false).Release();
-		if (webViewProvider.isWebView_11Available()) webViewProvider.getWebView_11(false).Release();
-		if (webViewProvider.isWebView_12Available()) webViewProvider.getWebView_12(false).Release();
-		if (webViewProvider.isWebView_13Available()) webViewProvider.getWebView_13(false).Release();
 		if(controller != null) {
 			// Bug in WebView2. Closing the controller from an event handler results
 			// in a crash. The fix is to delay the closure with asyncExec.
--- a/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
@@ -264,7 +264,9 @@ static int callAndWait(long[] ppv, ToIntFunction<IUnknown> callable) {
 	phr[0] = callable.applyAsInt(completion);
 	// "completion" callback may be called asynchronously,
 	// so keep processing next OS message that may call it
-	processOSMessagesUntil(() -> phr[0] != COM.S_OK || ppv[0] != 0, Display.getCurrent());
+	processOSMessagesUntil(() -> phr[0] != COM.S_OK || ppv[0] != 0, exception -> {
+		throw exception;
+	}, Display.getCurrent());
 	completion.Release();
 	return phr[0];
 }
@@ -282,7 +284,9 @@ int callAndWait(String[] pstr, ToIntFunction<IUnknown> callable) {
 	phr[0] = callable.applyAsInt(completion);
 	// "completion" callback may be called asynchronously,
 	// so keep processing next OS message that may call it
-	processOSMessagesUntil(() -> phr[0] != COM.S_OK || pstr[0] != null, browser.getDisplay());
+	processOSMessagesUntil(() -> phr[0] != COM.S_OK || pstr[0] != null, exception -> {
+		throw exception;
+	}, browser.getDisplay());
 	completion.Release();
 	return phr[0];
 }
@@ -324,9 +328,8 @@ class WebViewWrapper {
 }
 
 class WebViewProvider {
-
 	private CompletableFuture<WebViewWrapper> webViewWrapperFuture = wakeDisplayAfterFuture(new CompletableFuture<>());
-	private CompletableFuture<Void> lastWebViewTask = webViewWrapperFuture.thenRun(() -> {});;
+	private CompletableFuture<Void> lastWebViewTask = webViewWrapperFuture.thenRun(() -> {});
 
 	ICoreWebView2 initializeWebView(ICoreWebView2Controller controller) {
 		long[] ppv = new long[1];
@@ -339,7 +342,12 @@ class WebViewProvider {
 		webViewWrapper.webView_11 = initializeWebView_11(webView);
 		webViewWrapper.webView_12 = initializeWebView_12(webView);
 		webViewWrapper.webView_13 = initializeWebView_13(webView);
-		webViewWrapperFuture.complete(webViewWrapper);
+		boolean success = webViewWrapperFuture.complete(webViewWrapper);
+		// Release the webViews if the webViewWrapperFuture has already timed out and completed exceptionally
+		if(!success && webViewWrapperFuture.isCompletedExceptionally()) {
+			webViewWrapper.releaseWebViews();
+			return null;
+		}
 		return webView;
 	}
 
@@ -398,13 +406,19 @@ class WebViewProvider {
 
 	private WebViewWrapper getWebViewWrapper(boolean waitForPendingWebviewTasksToFinish) {
 		if(waitForPendingWebviewTasksToFinish) {
-			processOSMessagesUntil(lastWebViewTask::isDone, browser.getDisplay());
+			processOSMessagesUntil(lastWebViewTask::isDone, exception -> {
+				lastWebViewTask.completeExceptionally(exception);
+				throw exception;
+			}, browser.getDisplay());
 		}
 		return webViewWrapperFuture.join();
 	}
 
 	private WebViewWrapper getWebViewWrapper() {
-		processOSMessagesUntil(webViewWrapperFuture::isDone, browser.getDisplay());
+		processOSMessagesUntil(webViewWrapperFuture::isDone, exception -> {
+			webViewWrapperFuture.completeExceptionally(exception);
+			throw exception;
+		}, browser.getDisplay());
 		return webViewWrapperFuture.join();
 	}
 
@@ -488,8 +502,9 @@ class WebViewProvider {
  * leads to a failure in browser initialization if processed in between the OS
  * events for initialization. Thus, this method does not implement an ordinary
  * readAndDispatch loop, but waits for an OS event to be processed.
+ * @throws Throwable
  */
-private static void processOSMessagesUntil(Supplier<Boolean> condition, Display display) {
+private static void processOSMessagesUntil(Supplier<Boolean> condition, Consumer<SWTException> timeoutHandler, Display display) {
 	MSG msg = new MSG();
 	AtomicBoolean timeoutOccurred = new AtomicBoolean();
 	// The timer call also wakes up the display to avoid being stuck in display.sleep()
@@ -502,8 +517,12 @@ private static void processOSMessagesUntil(Supplier<Boolean> condition, Display 
 		}
 	}
 	if (!condition.get()) {
-		SWT.error(SWT.ERROR_UNSPECIFIED, null, " Waiting for Edge operation to terminate timed out");
+		timeoutHandler.accept(createTimeOutException());
 	}
+}
+
+private static SWTException createTimeOutException() {
+	return new SWTException(SWT.ERROR_UNSPECIFIED, "Waiting for Edge operation to terminate timed out");
 }
 
 static ICoreWebView2CookieManager getCookieManager() {
@@ -622,16 +641,9 @@ private void createInstance(int previousAttempts) {
 }
 
 private IUnknown createControllerInitializationCallback(int previousAttempts) {
-	Runnable initializationRollback = () -> {
-		if (environment2 != null) {
-			environment2.Release();
-			environment2 = null;
-		}
-		containingEnvironment.instances().remove(this);
-	};
 	Runnable initializationAbortion = () -> {
 		webViewProvider.abortInitialization();
-		initializationRollback.run();
+		releaseEnvironment();
 	};
 	return newCallback((resultAsLong, pv) -> {
 		int result = (int) resultAsLong;
@@ -657,7 +669,7 @@ private IUnknown createControllerInitializationCallback(int previousAttempts) {
 			initializationAbortion.run();
 			break;
 		default:
-			initializationRollback.run();
+			releaseEnvironment();
 			if (previousAttempts < MAXIMUM_CREATION_RETRIES) {
 				System.err.println(String.format("Edge initialization failed, retrying (attempt %d / %d)", previousAttempts + 1, MAXIMUM_CREATION_RETRIES));
 				createInstance(previousAttempts + 1);
@@ -671,10 +683,23 @@ private IUnknown createControllerInitializationCallback(int previousAttempts) {
 	});
 }
 
+private void releaseEnvironment() {
+	if (environment2 != null) {
+		environment2.Release();
+		environment2 = null;
+	}
+	containingEnvironment.instances().remove(this);
+}
+
 void setupBrowser(int hr, long pv) {
 	long[] ppv = new long[] {pv};
 	controller = new ICoreWebView2Controller(ppv[0]);
 	final ICoreWebView2 webView = webViewProvider.initializeWebView(controller);
+	if(webView == null) {
+		controller.Release();
+		releaseEnvironment();
+		return;
+	}
 	webView.get_Settings(ppv);
 	settings = new ICoreWebView2Settings(ppv[0]);
 


fedejeanne:
I would additionally reduce the timeout introduced in https://github.com/eclipse-platform/eclipse.platform.swt/pull/1677 to 10 seconds (+5 second from your new timeout) just to make sure that your change here actually covers all failing tests and use cases. If it does then I would expect to see no timeouts hitting **10s**, only timeouts after **5s**.

WDYT?

amartya4256:
> I would additionally reduce the timeout introduced in #1677 to 10 seconds (+5 seconds from your new timeout) just to make sure that your change here actually covers all failing tests and use cases. If it does then I would expect to see no timeouts hitting **10s**, only timeouts after **5s**.
> 
> WDYT?

Alright, I'll add it here.

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -311,6 +311,17 @@ ICoreWebView2 initializeWebView(ICoreWebView2Controller controller) {
 		return webView;
 	}
 
+	private CompletableFuture<ICoreWebView2> initializeWebViewFutureWithTimeOut() {
+		CompletableFuture<ICoreWebView2> webViewFuture = new CompletableFuture<>();
+		webViewFuture.orTimeout(1, TimeUnit.MILLISECONDS).exceptionally(exception -> {
+			// Throw exception on the Display thread directly to prevent CompletableFuture
+			// to wrap the exception and throw it silently
+			browser.getDisplay().execute(() -> SWT.error(SWT.ERROR_UNSPECIFIED, exception, "Edge Browser initialization timed out"));
```
HeikoKlare, line 319: In case we run into a timeout, the initialization needs to be properly rolled back in terms of the state of `Edge` being cleaned up and potentially OS resources to be freed, like in the abortion logic of `createControllerInitializationCallback()`.

HeikoKlare, line 319: Instead of only throwing an error, should we maybe hide the (uninitialized) browser widget and add a label informing the user about the failed initialization instead? Calling something like this inside the execution on display might do the trick:
```java
	private void replaceWithErrorLabel() {
		browser.setVisible(false);
		Label errorLabel = new Label(browser.getParent(), SWT.WRAP);
		errorLabel.setForeground(browser.getDisplay().getSystemColor(SWT.COLOR_RED));
		errorLabel.setText("Edge browser initialization failed");
		errorLabel.setLocation(0, 0);
		errorLabel.setSize(browser.getSize());
	}
```

HeikoKlare (commented):
I think we need to have another thorough look at what may cause Edge initialization to block the UI thread. When trying to test the timeout logic, I came across several places where I can introduce blocking operations that will not be caught by this proposal.

When systematically going through the public methods of Edge, I see that the calls to `getWebView*()` and `isWebView*Availabile()` are the ones that block in case initialization or any of the subsequently scheduled futures do not finish timely.
Then, taking a look at which calls may potentially block inside those methods, I see these two:
- waitForFutureToFinish()
- future.join()

The latter will be captured for the initialization future by the timeout added in this PR. The former will only partly be resolved, as inside `waitForFutureToFinish()` there is a call to `processNextOSMessage()`, which spins a loop that may be blocking as well. Thus, I think we need to break that potentially blocking operation as well. Maybe we can simply spin `Display.readAndDisplay()` in the loop checking for future completion (inside `waitForFutureToFinish()`) instead of having the additional loop inside `processNextOSMessage()`?

Then, what about the other callers of `processNextOSMessage()`? In particular, `createEnvironment()` is called synchronously when creating the first Edge instance. May that one block as well? I also see further calls of that method of which I am not sure if they are correct. We need to check them as well.

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
amartya4256: I really like his idea. But how about we combine this with also throwing an error even though it doesn't pop up in a dialog. The error is atleast visible in the Error Log of the IDE.

amartya4256: Also if we don't use this execute method here, it will sill throw the error but the error will be wrapped in CompletionException and not SWTException since, the errors in the CompletableFuture are caught implicitly and wrapped with CompletionException. In case of SWT.error as well, I see the exception to be silent and there's no dialog coming up. What do you prefer? Is that okay to not wrap the exception with SWT.error and let it throw the CompletableFuture CompletionException and just replace the browser with a label?
image
image
The image above is for when I dont throw a wrapped SWT exception and let the future handle it

amartya4256: I have extracted the environment release method and added that here as now we have a webview wrapper which makes sure that we dont have multiple futures dealing with multiple instances of webView but one future initializing everything as a whole. And I have verified, the initialization of webview* happens very quickly after obtaining the webview instance which gives the initialization an atomic behaviour. What do you think?

amartya4256:
> When systematically going through the public methods of Edge, I see that the calls to `getWebView*()` and `isWebView*Availabile()` are the ones that block in case initialization or any of the subsequently scheduled futures do not finish timely. Then, taking a look at which calls may potentially block inside those methods, I see these two:
> 
> * waitForFutureToFinish()
> * future.join()


I evaluated your points and i have implemented everything into a webviewWrapper which is responsible for containing all the webviews and we need just one future (webViewWrapperFuture) now which provides us this wrapper from where all the webview* can be obtained, so now the dependency flow is like this: webViewWrapperFuture -> lastWebViewTask. for every webView*, it will be responsible - in case of a webViewWrapperFuture timeout, everything times out.

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
HeikoKlare: Generally sounds good, but I am sure whether it works like it is implemented now. You will run into the timeout handler and release the environment while the webview initialization is still running asynchronously at the OS. Is that the right thing to do? I would rather expect that to eventually be done when the initialization callback is executed (at some point in time after the callback) or, maybe, after a rather high amount of further waiting time.

HeikoKlare: I would still be in favor of logging some kind of error. Even though the user will primarily see the information in the UI, someone debugging issues may only have a look at the logs, which is why having such an error documented there is reasonable.

HeikoKlare:
> I evaluated your points and i have implemented everything into a webviewWrapper which is responsible for containing all the webviews and we need just one future (webViewWrapperFuture) now which provides us this wrapper from where all the webview* can be obtained, so now the dependency flow is like this: webViewWrapperFuture -> lastWebViewTask. for every webView*, it will be responsible - in case of a webViewWrapperFuture timeout, everything times out.

That sounds good. There are several good proposals in this PR now. Can we please separate into different smaller PRs to ease reviewing and testing? In particular, putting the web views inside a wrapper captured by one future is something that we might do as a simple "cleanup" and preparatory PR for this.

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
amartya4256: so replaceBrowser with label and log an error? the error is logged automatically by the future of exception. Do you mean that or should we throw an exception using, `browser.getDisplay().execute(() -> SWT.error(SWT.ERROR_UNSPECIFIED, exception, "Edge Browser initialization timed out"));`?

HeikoKlare:
> The question about if processNextOSMessage is a blocking anything: what kind of messages are processed in there and what control do we have? Looks like, we are just calling display.readAndDispatch(), which is just processing the messages on the display thread from the OS, if something unrelated to Edge was blocking it here, it could have been blocked anywhere in the platform. And We have all the handlers well defined to handling the Edge messages in the callbacks properly, my concern is about the block:
> 
> ```
> while (!OS.PeekMessage (msg, 0, 0, 0, OS.PM_NOREMOVE)) {
> 		display.sleep();
> 	}
> ```

This code will put the thread to sleep until some OS message is received. If for some reason no message arrives anymore, this will block the whole application. Not sure how likely it is, but I would be in favor of avoiding the risk. As mentioned above, it might be possible to simplify this code anway.

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
HeikoKlare: I have no strong opinion this, but I see two things to consider:
1. If possible, we should avoid throwing a future-specific exception to consumers of SWT/Edge and rather provide proper information inside an SWTException to them.
2. We should have a look at the where consumers may face that exception and how that conforms to the existing APIs they are using (usually those of `Browser`).

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -316,10 +316,23 @@ ICoreWebView2 initializeWebView(ICoreWebView2Controller controller) {
 		return webView;
 	}
 
+	private CompletableFuture<WebViewWrapper> initializeWebViewFutureWithTimeOut() {
+		CompletableFuture<WebViewWrapper> webViewWrapperFuture = new CompletableFuture<>();
+		webViewWrapperFuture.orTimeout(3, TimeUnit.SECONDS).exceptionally(exception -> {
```
laeubi, line 321:
```suggestion
		webViewWrapperFuture.orTimeout(3, TimeUnit.SECONDS).exceptionallyAsync(exception -> {...}, browser.getDisplay())
```

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
laeubi: if one is interested in the exception, `get()` is better than `join()`

laeubi:
### Just a general remark here:

The usage of `CompleteableFuture` here (even though very convenient and powerful) does not seem to be the proper way when looking at the overall design here.

Instead one should probably better use a Task-List design here, so that different task are processed one after the other, this could be for example a (single thread) `ExecutorService` where one `submit(...)` tasks, or a `Queue` that is processed by a special thread (e.g. if SWT thread is needed) unless it is empty.

If `ExecutorService` is used then one can also combine it with `CompleteableFuture`, but then it should be used in an event driven way without ever using a `join()`.

laeubi:
> When systematically going through the public methods of Edge

**I think the key is actually**: What **public API methods** really require blocking operations.

These methods should not use `join()` at all, they must use a busy loop that drives the event queue while waiting on this similar to what we do here:

https://github.com/eclipse-platform/eclipse.platform.swt/blob/5d535ce0d6bc288861042f9595f9a8a3a1bf9662/bundles/org.eclipse.swt/Eclipse%20SWT%20Custom%20Widgets/common/org/eclipse/swt/custom/BusyIndicator.java#L111-L121

https://github.com/eclipse-platform/eclipse.platform.swt/blob/5d535ce0d6bc288861042f9595f9a8a3a1bf9662/bundles/org.eclipse.swt/Eclipse%20SWT%20Custom%20Widgets/common/org/eclipse/swt/custom/BusyIndicator.java#L136-L140

(the wake is important to break out the sleep!)

HeikoKlare:
Thanks for the feedback! The proposals make sense and revising the design again might be a good thing. But we should move that discussion into a separate issue. This PR is part of efforts to mitigate the risk of freezes / UI thread blocks by using Edge for the upcoming release. For that reason, we need to improve the current design instead of revising the design, which we may defer to the next development cycle.

laeubi:
@HeikoKlare as a workaround replace all parts that do a `join()` on the event thread with this code:

```
public static void waitForFuture(CompletableFuture<?> future, Display display) {
	if (!future.isDone()) {
		future.handle((nil1, nil2) -> {
			if (!display.isDisposed()) {
				try {
					display.wake();
				} catch (SWTException e) {
					// ignore then, this can happen due to the async nature between our check for
					// disposed and the actual call to wake the display can be disposed
				}
			}
			return null;
		});
		while (!future.isDone() && !display.isDisposed()) {
		 		display.sleep();
		 	}
		 }
	}
}
```

Then you don't need special timeout handling and won't see UI freeze.

laeubi:
it looks odd, but I think

```
private <T> void waitForFutureToFinish(CompletableFuture<T> future) {
	while(!future.isDone()) {
		Display display = Display.getCurrent();
		future.handle((nil1, nil2) -> {
			if (!display.isDisposed()) {
				try {
					display.wake();
				} catch (SWTException e) {
					// ignore then, this can happen due to the async nature between our check for
					// disposed and the actual call to wake the display can be disposed
				}
			}
			return null;
		});
		MSG msg = new MSG();
		while (!OS.PeekMessage (msg, 0, 0, 0, OS.PM_NOREMOVE)) {
			if (!future.isDone()) {
				display.sleep();
			}
		}
		display.readAndDispatch();
	}
}
```

would probably prevent blocking (future itself does not use (a)sync exec).

fedejeanne:
This PR will probably be superseded by #1776 , right?

HeikoKlare:
> This PR will probably be superseded by #1776 , right?

Not necessarily. We will need to check that in detail. E.g., #1776 does currently not cancel any futures. Also things like the "replacement" of the browser widget (if we agree that we want to try that), is something #1776 does not cover.

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
amartya4256: I am going with `browser.getDisplay().execute(() -> SWT.error(SWT.ERROR_UNSPECIFIED, exception, "Edge Browser initialization timed out"));` since with this we make sure our SWT exception is not caught by CompletableFuture and silenced on timeout.

laeubi: This is a very strange way to print an exception... and it will require a synchronous call to the SWT thread... why not use `display.getErrorHandler () `directly?!?

amartya4256: You are right. It's better to let the other handlers handle the dispose
I have changed the flow of how things are initialized and disposed in case there's a time out. Its a little something bit of extra change but now I can handle exceptions a bit better and the flow also looks clearer to me. Can you have a look?

amartya4256: Yes I agree that looks better. But I think in this case runtimeExceptionHandler should be the right one, since its an instance of exception.

amartya4256: Yes I agree that looks better. But do you think its is good to check whether to use RunTimeHandler or ErrorHandler?

HeikoKlare:
Regarding the error label: I found that different consumers already have fallback logic when the browser instantiation failed, like the internal browser view showing an error message and the JDT control using a "simple" viewer with the note that the browser was not instantiated correctly:
image

So we should drop the idea of doing something similar in browser itself.

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -640,7 +649,10 @@ private IUnknown createControllerInitializationCallback(int previousAttempts) {
 		switch ((int) result) {
 		case COM.S_OK:
 			new IUnknown(pv).AddRef();
-			setupBrowser((int) result, pv);
+			boolean success = setupBrowser((int) result, pv);
+			if(!success) {
+				initializationRollback.run();
+			}
```
HeikoKlare, line 655: Couldn't we call the rollback right in the `setupBrowser` method to avoid that this caller need to rely on the called method terminating in a specific state that will be rolled back but instead make explicit in the aborting method what state needs to be rolled back when calling the rollback function?

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -496,10 +500,15 @@ private static void processOSMessagesUntil(Supplier<Boolean> condition, Display
 		}
 	}
 	if (!condition.get()) {
+		timeoutHandler.ifPresent(handler -> handler.run());
```
HeikoKlare, line 503:
```
timeoutHandler.ifPresent(Runnable:run);
```

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -331,8 +330,13 @@ ICoreWebView2 initializeWebView(ICoreWebView2Controller controller) {
 		webViewWrapper.webView_10 = initializeWebView_10(webView);
 		webViewWrapper.webView_11 = initializeWebView_11(webView);
 		webViewWrapper.webView_12 = initializeWebView_12(webView);
-		webViewWrapper.webView_13 = initializeWebView_13(webView);
-		webViewWrapperFuture.complete(webViewWrapper);
+		webViewWrapper.webView_13= initializeWebView_13(webView);
```
HeikoKlare, line 333: Please revert formatting change.

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -483,7 +487,7 @@ private <T> CompletableFuture<T> wakeDisplayAfterFuture(CompletableFuture<T> fut
  * events for initialization. Thus, this method does not implement an ordinary
  * readAndDispatch loop, but waits for an OS event to be processed.
  */
-private static void processOSMessagesUntil(Supplier<Boolean> condition, Display display) {
+private static void processOSMessagesUntil(Supplier<Boolean> condition, Optional<Runnable> timeoutHandler, Display display) {
```
HeikoKlare, line 490: Instead of using optional, can't we just accept a Runnable an pass `Function::identity` whenever there is nothing to do?

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -496,10 +500,15 @@ private static void processOSMessagesUntil(Supplier<Boolean> condition, Display
...
 	if (!condition.get()) {
+		timeoutHandler.ifPresent(handler -> handler.run());
 		SWT.error(SWT.ERROR_UNSPECIFIED, null, " Waiting for Edge operation to terminate timed out");
 	}
 }
 
+private static SWTException createTimeOutException() {
+	return new SWTException(SWT.ERROR_UNSPECIFIED, " Waiting for Edge operation to terminate timed out");
+}
+
```
HeikoKlare, line 511: This method should then also be called inside `processOSMessageUntil()` to procude the exception (i.e., use `throw createTimeOutException()` there).

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -319,6 +340,11 @@ private void abortInitialization() {
 		webViewWrapperFuture.cancel(true);
 	}
 
+	void releaseWebView() {
+		processOSMessagesUntil(webViewWrapperFuture::isDone, browser.getDisplay());
```
HeikoKlare: Why is that necessary? This method may only be called when the wrapper future is done (as it is called from a future scheduled _after_ the wrapper future).

HeikoKlare (commented):
In general, the changes look sound to me. I have rather comments on local design. I just do not understand the necessity for the added `processOSMessagesUntil()` call.

In addition, please revise the commit message to (1) be more descriptive and (2) refer to proper issues. They currently refer to some HiDPI issues, which are not related to Edge.

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
amartya4256: It was just done to make it safe to execute if used somewhere else later. Since webViewWrapperFuture is already done, it will not go in the loop and exit the processOSMessagesUntil method right away. But if in future, we use it somewhere where there is no check for future to have finished, we can be in problem. Anyways semantically, you cant release webViews without having them created in the first place; so I think it is a good check here.

- In our use case this loop processOSMessagesUntil  would be exited right away. 
What do you think? Should I keep it like this?

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
amartya4256: What do you mean by making it explicit in the aborting method? One thing I can thing of is passing the rollback runnable to setupBrowser as a rollback handler and execute it on rollback

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
amartya4256: Optional empty kinda tells that there's nothing passed for timeoutHandler. Don't you think it sounds more correct than passing an identity function, which basically means that it returns what is passed to it?

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
amartya4256: Done

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
HeikoKlare: You could also directly call the runnable from within the `setupBrowser()` method (by explicit reference, without passing it as a lambda) to make the behavior more explicit, can't you?

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
HeikoKlare: I mean that a handler can by itself already empty, so having an optional that allows to not have a handler instead of having a handler that does nothing sounds like unnecessary complexity to me. In particular, since you need to additionally check the optional for containing an actual value while a (potentially empty) runnable can simply be executed.

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
HeikoKlare: The idea sounds reasonable. But isn't then the whole method equivalent to just `getWebViewWrapper().releaseWebViews()`?

amartya4256: Yes, it is: I will adapt it like that. thanks

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
amartya4256: Alrighty then Function::identity it is

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
amartya4256: Currently the runnable is defined in the scope of createControllerInitializationCallback method. I would need to extract it outside for an explicit access in setupBrowser. Then it makes sense to convert it into a method and call in both createControllerInitializationCallback and setUpBrowser where intended.

amartya4256: Alright, I modified it accordingly

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -367,13 +396,13 @@ private ICoreWebView2_13 initializeWebView_13(ICoreWebView2 webView) {
 
 	private WebViewWrapper getWebViewWrapper(boolean waitForPendingWebviewTasksToFinish) {
 		if(waitForPendingWebviewTasksToFinish) {
-			processOSMessagesUntil(lastWebViewTask::isDone, browser.getDisplay());
+			processOSMessagesUntil(lastWebViewTask::isDone, Function::identity, browser.getDisplay());
```
HeikoKlare, line 399: Why doesn't the timeout handler complete exceptionally here?

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -471,10 +500,15 @@ private static void processOSMessagesUntil(Supplier<Boolean> condition, Display
 		}
 	}
 	if (!condition.get()) {
-		SWT.error(SWT.ERROR_UNSPECIFIED, null, " Waiting for Edge operation to terminate timed out");
+		timeoutHandler.run();
+		throw createTimeOutException();
```
HeikoKlare, line 504: Couldn't we make timeoutHandler a `Consumer<Exception>` which can then process the exception itself? Some will use the exception to exceptionally complete the future and others will just throw the exception.

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -471,10 +500,15 @@ private static void processOSMessagesUntil(Supplier<Boolean> condition, Display
...
 	}
 }
 
+private static SWTException createTimeOutException() {
+	return new SWTException(SWT.ERROR_UNSPECIFIED, " Waiting for Edge operation to terminate timed out");
```
HeikoKlare, line 509:
```suggestion
	return new SWTException(SWT.ERROR_UNSPECIFIED, "Waiting for Edge operation to terminate timed out");
```

HeikoKlare (commented):
I have one remaining question regarding correctness and two minor proposals for improvement.

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
amartya4256: We need to specify the future that we want to timeout. It will not always be the webViewWrapperFuture, but it can also be lastWebViewTaskFuture, if something fails in the middle because of a deadlock. I think in evaluate we have the same problem.

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
amartya4256: done

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
amartya4256: Or should we just go with a null check?

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
HeikoKlare: Okay, now I better see the issue there. What we would actually like to have is to exceptionally complete exactly that future that did not finish (in time). With the current implementation, that seems to be impossible, so we may only complete the initialization future or the last scheduled future exceptionally. Taking that as given, I would say that inside `getWebViewWrapper(boolean waitForPendingWebviewTasksToFinish)` we should then, as an additional first statement, call `getWebViewWrapper()` (including the waiting for the initialization to finish) first. This even seems to be a bug right now, as calling `getWebViewWrapper(false)` will directly run into `webViewWrapperFuture.join()`, which may never terminate.

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
HeikoKlare: What I actually meant is to combine the `timeoutHandler.run()` and `throw createTimeOutException()` statements by just having a timeoutHandler that also throws the exception if desired. So there would be no Function::identity or empty consumers but rather some `ex -> throw ex`. Just saw that this is of course not possible with a `Consumer`, but would be possible with a `ThrowingConsumer`.

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
amartya4256: getWebViewWrapper(false) was intended to be called from inside a scheduled future with the supplier i.e. when we use scheduleWebViewTask, because it will only execute once all the tasks before it are completed, which is, the webviewWrapperFuture will also be complete. ProcessOSMessagesUntil works fine for the initialization logic but my first suggestion was to use a CompletableFuture TimeOut which is not that straight forwardly possible after integrating the processOSMessageUntil logic. However, I don't believe that exceptionally completing lastWebViewTask is a problem, given it will only do that if the previous task has still not finished, which is legitimate but the wait time can vary depending on when the next task's call was made.

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
amartya4256: So we don't throw exception in the processOSMessagesUntil method directly but rather using the ThrowingConsumer.

amartya4256: Seems like it throws checked exception and asks me to change the signature of the methods

amartya4256: I think if we create the exception in processOSMessagesUntil then we should also explicitly throw it there. The futures just need the exception to complete exceptionally. I think its fine keep their functional scope to just completing silently. i mean they are going to throw an error to the consumer for the join() method anyway.

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
HeikoKlare: Okay, now I got it. Thank you for the explanation!

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
HeikoKlare: Since SWTException is unchecked, just having an ordinary `Consumer` should work work fine. I'd like to avoid that we throw exceptions all over the place and rather have a dedicated place where this kind of exception is generated and one responsibility for throwing it.

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -294,12 +294,32 @@ class WebViewWrapper {
 	private ICoreWebView2_11 webView_11;
 	private ICoreWebView2_12 webView_12;
 	private ICoreWebView2_13 webView_13;
+
+	void releaseWebViews() {
+		if(webView != null) {
+			webView.Release();
```
laeubi, line 300:
```suggestion
			webView.Release();
			webView = null;
```

same below...

PR #1753, opened 2025-01-22
Unify internal strict mode flag for win32 and gtk

akoch-yatta:
This contribution renames the internal strict mode flag in GTK from org.eclipse.swt.internal.gtk.enableStrictChecks to org.eclipse.swt.internal.enableStrictChecks and adds it to win32 to be used to check for correctly configured GCs for monitor-specific-scaling. The helps tracing invalid custom Drawables or GCs that are created with an invalid configuration.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Device.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Device.java
@@ -33,6 +33,8 @@ import org.eclipse.swt.widgets.*;
  */
 public abstract class Device implements Drawable {
 
+	static boolean strictChecks = System.getProperty("org.eclipse.swt.internal.enableStrictChecks") != null;
+
 	/* Debugging */
 	public static boolean DEBUG;
 	boolean debug = DEBUG;
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
@@ -3950,6 +3950,10 @@ void init(Drawable drawable, GCData data, long hDC) {
 }
 
 private static int extractZoom(long hDC) {
+	if (Device.strictChecks) {
+		System.err.println("***WARNING: GC is initialized with a missing zoom. This indicates an "
+				+ "incompatible custom Drawable implementation.");
+	}
 	long hwnd = OS.WindowFromDC(hDC);
 	long parentWindow = OS.GetAncestor(hwnd, OS.GA_ROOT);
 	long monitorParent = OS.MonitorFromWindow(parentWindow, OS.MONITOR_DEFAULTTONEAREST);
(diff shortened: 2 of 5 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -102,6 +102,20 @@ public final class GC extends Resource {
...
+	 * <li><b>true</b>: Enables additional checks and prints errors if an invalid configuration is detected.
+	 * This mode should only be used during development.</li>
+	 * <li><b>false</b>: No additional checks are executed.</li>
+	 * </ul>
+	 * <b>Important:</b> This flag is currently only used on Win32. Setting it to
+	 * true on GTK or cocoa wont have any effect.
+	 * <p>
+	 * The current default is "false".
+	 */
+	private static final String GC_STRICT_MODE = "gc.strict";
```
iloveeclipse, line 117: On GTK side we already have `-Dorg.eclipse.swt.internal.gtk.enableStrictChecks` flag. I believe it would make sense to use same value here.

akoch-yatta: Thx for the pointer. Would it be okay to just change the GTK one, so the old one would be gone?

iloveeclipse: "gc.strict" seem to be "gc" specific. I don't think we have too many checks in SWT that qualify for a dedicated GC checks, we just have few checks that could be enabled.

The old flag is not API, it is only for SWT developers / supporting SWT problem finding. I believe it shouldn't be a big deal to change it.

I just saw you introducing that in Windows port too for same reason so would be more consistent if we have cross platform flag for SWT low level checks.

akoch-yatta: Yes, I totally agree, I would unify that. Just wanted to make sure, we wouldn't break something with it. Then I will take your proposal unify it with -Dorg.eclipse.swt.internal.enableStrictChecks in this PR

iloveeclipse (approved):
Thanks

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Device.java
```diff
@@ -33,6 +33,8 @@
  */
 public abstract class Device implements Drawable {
 
+	static boolean strictChecks = Boolean.getBoolean("org.eclipse.swt.internal.enableStrictChecks");
```
HeikoKlare, line 36: Should this be `System.getProperty("org.eclipse.swt.internal.enableStrictChecks") != null` for the sake of consistency as well?

Otherwise, `-Dorg.eclipse.swt.internal.enableStrictChecks` will enable the checks on GTK but not on Windows. On Windows, you explicitly have to assign "true". This also means that the test configuration for GitHub Actions and Jenkins are inconsistent in this PR, as GitHub Actions will enable strict mode on Windows, while Jenkins will not.

HeikoKlare (approved):

PR #1754, opened 2025-01-22
Calculate chevron size in CTabFolderRenderer without Device#getDPI()

ShahzaibIbrahim:
getDPI always returns the points for primary monitor and can produce faulty behavior. Replacing the logic to calculate the chevron font size. Visibly 70% size of the font used for CTabItem is now used. Also adjusting the indent and adding the ZoomChange event to trigger redraw during DPI change.

**NOTE: Needs to be tested on linux and macOS**

### Steps to Reproduce

- Run the CustomControlExample with following arguments in 150% primary monitor
 > -Dswt.autoScale.updateOnRuntime=true
 > -Dswt.autoScale=quarter

- Change the tab to "**CTabFolder**"
- Change the Size to **100x100**
- Move the shell to other monitor with (100 or 200%) 
- The number along with chevron image should look fine.
- **Additional**: The tabs should be aligned correctly as well. 

### Before Fix

image

### After Fix
image

Scenarios Tested yet:
Case 	    Quarter   no args
150 -> 100 	OK	   OK
150 -> 200	OK	   OK

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
@@ -102,7 +102,8 @@ public class CTabFolderRenderer {
 	static final int FLAGS = SWT.DRAW_TRANSPARENT | SWT.DRAW_MNEMONIC | SWT.DRAW_DELIMITER;
 	static final String ELLIPSIS = "..."; //$NON-NLS-1$
 	private static final String CHEVRON_ELLIPSIS = "99+"; //$NON-NLS-1$
-	private static final int CHEVRON_FONT_HEIGHT = 10;
+	private static final float CHEVRON_FONT_SIZE_FACTOR = 0.7f;
+	private static final int CHEVRON_BOTTOM_INDENT = 4;
 
 	//Part constants
 	/**
@@ -926,7 +927,7 @@ public class CTabFolderRenderer {
 		Display display = parent.getDisplay();
 		Font font = getChevronFont(display);
 		int fontHeight = font.getFontData()[0].getHeight();
-		int indent = Math.max(2, (chevronRect.height - fontHeight - 4) /2);
+		int indent = Math.max(2, (chevronRect.height - fontHeight - CHEVRON_BOTTOM_INDENT) /2);
 		int x = chevronRect.x + 2;
 		int y = chevronRect.y + indent;
 		int count = parent.getChevronCount();
@@ -1696,9 +1697,8 @@ public class CTabFolderRenderer {
 
 	private Font getChevronFont(Display display) {
 		if (chevronFont == null) {
-			Point dpi = display.getDPI();
-			int fontHeight = 72 * CHEVRON_FONT_HEIGHT / dpi.y;
 			FontData fd = parent.getFont().getFontData()[0];
+			int fontHeight = (int) (fd.height * CHEVRON_FONT_SIZE_FACTOR);
 			fd.setHeight(fontHeight);
 			chevronFont = new Font(display, fd);
 		}
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
@@ -341,6 +341,7 @@ void init(int style) {
 			case SWT.Selection:        onSelection(event); break;
 			case SWT.Activate:         onActivate(event); break;
 			case SWT.Deactivate:       onDeactivate(event); break;
+			case SWT.ZoomChanged:	   onZoomChange(event); break;
 		}
 	};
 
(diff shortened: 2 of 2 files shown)


bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
```diff
@@ -926,7 +926,7 @@ void drawChevron(GC gc, Rectangle chevronRect, int chevronImageState) {
 		Display display = parent.getDisplay();
 		Font font = getChevronFont(display);
 		int fontHeight = font.getFontData()[0].getHeight();
-		int indent = Math.max(2, (chevronRect.height - fontHeight - 4) /2);
+		int indent = Math.max(2, (chevronRect.height - fontHeight - 8) /2);
```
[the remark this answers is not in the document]
ShahzaibIbrahim: The indent before looks bit weird (not so aligned with the chevron icon)

HeikoKlare (commented):
Looks good on Linux/GTK. I don't see any noticable difference to existing appearance.

In the following screenshot, the left is existing and the right is with this PR:
image

Screenshots are taken at 100% monitor scaling. Looks the same (blurry scaled) with other display scalings, both before and after the change.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
```diff
@@ -1696,9 +1696,8 @@ Color getFillColor() {
 
 	private Font getChevronFont(Display display) {
 		if (chevronFont == null) {
-			Point dpi = display.getDPI();
-			int fontHeight = 72 * CHEVRON_FONT_HEIGHT / dpi.y;
```
HeikoKlare, line 1700: The static variable `CHEVRON_FONT_HEIGHT` is now unused and thus needs to be removed.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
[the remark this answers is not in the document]
HeikoKlare: Can you explain this in more detail? The indent is used for vertical placement of the chevron, so what do mean is not aligned with the chevron icon due to this value?

For me, both looks pretty much the same (in particular because the Math.max calculation only leads to a difference of 1 pixel on Windows and 2 pixels on Linux with my configurations):
image

HeikoKlare (commented):
@BeckerWdf I just saw that you originally introduced the scaling logic for the chevrons several years ago. Do you still remember that code and have an opinion on these changes or should we just move on?

HeikoKlare (commented):
Also fine on MacOS:
image

BeckerWdf:
> @BeckerWdf I just saw that you originally introduced the scaling logic for the chevrons several years ago. Do you still remember that code and have an opinion on these changes or should we just move on?

You mean this change: https://github.com/eclipse-platform/eclipse.platform.swt/commit/f25be5b085812ca2605f7ed4b92f7c31fe9170d1?

BeckerWdf:
> > You mean this change: [f25be5b](https://github.com/eclipse-platform/eclipse.platform.swt/commit/f25be5b085812ca2605f7ed4b92f7c31fe9170d1)?
> 
> Yes, that change.

Sorry. I don't remember the details. Just also read the bugzilla.
If it looks good with this PR on windows, mac and linux I think you can continue with this.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
```diff
@@ -1696,9 +1696,8 @@ Color getFillColor() {
...
 			FontData fd = parent.getFont().getFontData()[0];
+			int fontHeight = (int) (fd.height * 0.7);
```
BeckerWdf, line 1700: This class has lots of magic numbers. The "0.7" is one of them. Can we create a constant with a telling name of it?

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
BeckerWdf, line 929: can we avoid the magic number "8" here?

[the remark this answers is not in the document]
ShahzaibIbrahim: I had a problem at some point where the vertical indent was so high that it just cut off the text from bottom. I can't seem to reproduce that issue right now. This adjustment was to address that issue. For now I will revert it as not reproducible.

ShahzaibIbrahim: I need to keep at least 4 here to avoid text being cut-off from bottom. I can put this in a variable though

HeikoKlare (approved):
Looks good now. Can you squash the commits please?

PR #1755, opened 2025-01-23
Use SWT constant and make the code more readable.

deepika-u:
This commit refactors the classes in swt by using swt provided constant and makes the code more readable. Instead, the SWT.SPACE static constant is now leveraged for representing a string containing single character which is space in these classes.

### DIFF
--- a/bundles/org.eclipse.swt.tools.spies/src/org/eclipse/swt/tools/views/SpyView.java
+++ b/bundles/org.eclipse.swt.tools.spies/src/org/eclipse/swt/tools/views/SpyView.java
@@ -139,7 +139,7 @@ public class SpyView extends ViewPart {
 		if (object == null) return "null";
 		String name = object.toString ();
 		if (fullyQualifiedAction.isChecked()) {
-			int index = name.indexOf(' ');
+			int index = name.indexOf(SWT.SPACE);
 			if (index >= 0 && name.length() >= 1) {
 				name = object.getClass ().getName () + name.substring(index);
 			}
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/StyledText.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/StyledText.java
@@ -5295,7 +5295,7 @@ int insertBlockSelectionText(String text, boolean fillWithSpaces) {
 				int numSpaces = maxLength - length;
 				StringBuilder buffer = new StringBuilder(length + numSpaces);
 				buffer.append(line);
-				for (int j = 0; j < numSpaces; j++) buffer.append(' ');
+				for (int j = 0; j < numSpaces; j++) buffer.append(SWT.SPACE);
 				lines[i] = buffer.toString();
 			}
 		}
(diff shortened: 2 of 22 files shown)


deepika-u:
@merks @HeikoKlare : can you review this please?

HeikoKlare (commented):
I am not sure about this change, as for me several questions arise:
- Is there a specific reason for doing this with `SWT.SPACE` and not every other character constant in SWT, like `SWT.TAB` and the like? 
- Why replace exactly these usages of `' '` and not the others within the SWT bundles?
- Why is that reasonable/necessary now? Is there a specific trigger or did you come across this by accident? I am asking because the constant has been there for more than 10 years now.

To me, it looks as if those SWT constants are rather used for expressing key codes and not as constants for general ASCII characters. If we really want to use them for that, I would expect that we do that for every appearance of `' '` or that we have a common notion of where we use it and where we don't.

merks:
I don't think this makes the code more readable.  It makes the code more verbose and hence even arguably less readable.

deepika-u:
> I am not sure about this change, as for me several questions arise:
> 
>     * Is there a specific reason for doing this with `SWT.SPACE` and not every other character constant in SWT, like `SWT.TAB` and the like?
> 
>     * Why replace exactly these usages of `' '` and not the others within the SWT bundles?
> 
>     * Why is that reasonable/necessary now? Is there a specific trigger or did you come across this by accident? I am asking because the constant has been there for more than 10 years now.
> 
> 
> To me, it looks as if those SWT constants are rather used for expressing key codes and not as constants for general ASCII characters. If we really want to use them for that, I would expect that we do that for every appearance of `' '` or that we have a common notion of where we use it and where we don't.

Actually i totally agree to your view point. While checking on another issue, i got to check on the contents of the SWT.java file indetail so landed into this pr actually. While updating the instances of SPACE occurences only i got see '\t', '\n' characters are also being used. By the way i didnt find any contant for an empty string like ZERO_LENGTH_STRING or EMPTY_STRING, do you think we should add one?

- At first place, i have did only for windows(x86_64) platform for which i missed 1 instance, i have added it now. But now i updated in other OS specific files as well wherever needed.
- I was not sure on the impact(in terms of performance) of replacing space character by this constant in os calls(Text.java, Menu.java, Combo.java kind of files) so i have left them untouched.
- Ofcourse i didnt update in commented(Text.java) sections of code.
- When individual characters are mentioned like occurrences in JSON.java, WebBrowser.java - i havent intentionally changed those so that it is more sensible that way itself.

Ofcourse this is a cosmetic change and nothing else should change except the readability for a developer post this pr.

laeubi (changes requested):
I don't think this is appropriate here.

If I look at the javadoc it says:

> ASCII character convenience constant for the space character

and comparing this with other like ESC and how it is used in the current codebase it becomes quite clear that this is for using in key events when the user has pressed the space-bar an not a general purpose constant to be used in programming code.

Taking this further, one soon need to replace the `:` or `;` as well and literally each and every single character. I also don't think it improves readability, if I see a space I can easily read it as a space (ignoring the fact that there are different space characters like non breakable space and alike for now), also there is no risk that a space changes to something else and we want it to be the case very where (like with a string constant).

If one really want to improve readability, I see some places where a loop is used and instead one maybe want to use `" ".repeat(n)` instead.

Issue #1756, opened 2025-01-24
[GTK] GC#copyArea() produces wrong results when source and target area are overlapping


HeikoKlare:
The `GC` provides an operation `copyArea()` that is supposed to copy an area inside the GC's surface, i.e., from a source area of the surface into a target area of the same surface. When the source and target area are overlapping, the result is not as expected.

## Expected Behavior Example

Take the following original surface data:
![Image](https://github.com/user-attachments/assets/4c433bcb-8d52-4263-87e8-a457361a2361)

Let's assume we want to copy the red-marked area down to the yellow-marked area:
![Image](https://github.com/user-attachments/assets/56800520-0d78-4f3f-a1f7-910bcc2c2fe9)

Then we would expect the following result:
![Image](https://github.com/user-attachments/assets/6dfc1063-3ed6-4b6b-a4f8-90a4ce67bfbd)

## Actual Behavior Example

The expected result is only achieved under very specific conditions:
- You are not using GTK4 or Wayland
- You do not create the GC on an Image instance or the Image you pass was created via the Image constructor just accepting width/height or a size

The reason for the latter is that for that exact case (creating Image via width/height or size) a very specific workaround has been implemented via
https://bugs.eclipse.org/bugs/show_bug.cgi?id=571166:
https://github.com/eclipse-platform/eclipse.platform.swt/blob/413426e72a98e560582c902a9e0eb1c9d4211b02/bundles/org.eclipse.swt/Eclipse%20SWT/gtk/org/eclipse/swt/graphics/Image.java#L1263-L1267

In all other cases (using GTK4, using Wayland, using an Image created in a different way) you will see this result:
![Image](https://github.com/user-attachments/assets/6b5c1abd-9bfa-4443-b6f6-f4fb7348d144)

Since this is a derivation from the existing behavior of GTK and the behavior on Windows and MacOS, consumers may experience inconsistent cross-platform behavior, such as in https://github.com/eclipse-platform/eclipse.platform.ui/issues/2740

## Cause

The cause for the mentioned issue is in different behavior of copy operations inside the same surface depending on the used API and environment on which you create a surface. The usage of `GDK.gdk_window_create_similar_surface()` on X11 seems to perform some sufficient buffering of the data to be copied inside the same surface. In all other cases, the operation already overwrites its own source region while still reading from in.

[...]

PR #1757, opened 2025-01-24
[GTK] Fix GC#copyArea() for overlapping source and target areas #1756

HeikoKlare:
The GC provides a #copyArea() method that copies an area inside the surface of the GC to another area in the same surface. When using a surface directly created via the Cairo API (or with the GDK API on Wayland), this operation is broken.
When the source and target area are overlapping and the source is above-left of the target area, the source area will appear replicated in the target area, as Cairo does not perform any buffering but does a linewise/chunkwise write from the source area to the target area.

This change fixes the behavior of GC#copyArea() by explicitly making Cairo first pipe the copied area into an intermediate surface to be painted into the actual target area afterwards. It also removes the existing workaround in images that avoids the instantiation of a surface via the Cairo API, as the reason for that workaround was the issue fixed by this change. The effect can be seen in the description of https://github.com/eclipse-platform/eclipse.platform.swt/issues/1756. The operation now produces the result shown as the expected one in that issue, also on GTK4 and when using Wayland. An according regression test is added.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/1756

Can be tested in the line number ruler, which has also proper behavior on GTK4 and when using Wayland with this fix. Can also be tested together with https://github.com/eclipse-platform/eclipse.platform.ui/pull/2754 resolving the workaround that was implemented to deal with the bug in SWT.

There might be better solutions than this (e.g., methods directly provided by Cairo that perform the buffering), but I did not find them yet and actually any solution will have to do something like the solution proposed here (buffering the area to be copied).

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/GC.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/GC.java
@@ -548,7 +548,13 @@ void copyAreaInPixels(int srcX, int srcY, int width, int height, int destX, int 
 		Cairo.cairo_set_source_surface(handle, data.image.surface, deltaX, deltaY);
 		Cairo.cairo_rectangle(handle, destX, destY, width, height);
 		Cairo.cairo_set_operator(handle, Cairo.CAIRO_OPERATOR_SOURCE);
+		// As source and target area may be overlapping, we need to draw on
+		// an intermediate surface to avoid that parts of the source area are
+		// overwritten before reading it to be copied
+		Cairo.cairo_push_group(handle);
 		Cairo.cairo_fill(handle);
+		Cairo.cairo_pop_group_to_source(handle);
+		Cairo.cairo_paint(handle);
 	} else if (drawable != 0) {
 		Cairo.cairo_save(handle);
 		Cairo.cairo_rectangle(handle, destX, destY, width, height);
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Image.java
@@ -1260,11 +1260,7 @@ void init(int width, int height) {
 	this.type = SWT.BITMAP;
 
 	/* Create the pixmap */
-	if (GTK.GTK4) {
-		surface = Cairo.cairo_image_surface_create(Cairo.CAIRO_FORMAT_RGB24, width, height);
-	} else {
-		surface = GDK.gdk_window_create_similar_surface(GDK.gdk_get_default_root_window(), Cairo.CAIRO_CONTENT_COLOR, width, height);
-	}
+	surface = Cairo.cairo_image_surface_create(Cairo.CAIRO_FORMAT_ARGB32, width, height);
 	if (surface == 0) SWT.error(SWT.ERROR_NO_HANDLES);
 	// When we create a blank image we need to set it to 100 in GTK3 as we draw using 100% scale.
 	// Cairo will take care of scaling for us when image needs to be scaled.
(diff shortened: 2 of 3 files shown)


HeikoKlare:
@iloveeclipse can you please have a look at this? You were involved in the development of the workaround for the copyArea() issue (https://github.com/eclipse-platform/eclipse.platform.ui/pull/2754), which this PR replaces with an actual solution of the underlying problem, as well as in the discussion on the regression we had because of that in the line number ruler this week. So would be great to have your opinion and/or testing of this (in the best case together with the removal of the Platform UI workaround https://github.com/eclipse-platform/eclipse.platform.ui/pull/2754).

iloveeclipse:
> I'm by no means a Cairo expert, so if I'm completely off here just say so and ignore my input :-)

Me too :-)

I've tested this patch alone and also with reverted the UI workaround for line number ruler, everything seem to work

HeikoKlare:
> Drawing into a temporary surface: isn't this what [`cairo_push_group`](https://gitlab.freedesktop.org/cairo/cairo/-/blob/master/src/cairo.c?ref_type=heads#L658) and `cairo_pop_group_to_source` are for?

Thank you for the pointer! I saw code using these methods in other cases of `copyArea()` as well, but did not further look into it. I will test whether that can be used for an even simpler solution.

> I've tested this patch alone and also with reverted the UI workaround for line number ruler, everything seem to work

Thank you for testing! If we agree that this change (or an improvement incorporating the above mentioned pointer) is fine and has low risk of regression, I would be in favor of merging as soon as possible to have some testing of this in daily use until M3/release.

iloveeclipse:
Last patch seem to work too. Please rebase on master.

tomaswolf:
Glad to see my idea wasn't misleading at all. And thanks for using it; the code is much cleaner now!

HeikoKlare:
Tested again after updating to I20250124-1800 on Ubuntu 24.04: With X11 everything is working as before. Using Wayland the previous glitches with the line number ruler are fixed.

ptziegler:
We seem to be facing a test failure in GEF, ever since the M3 went live (or to be more precise, starting with [I20250124-1800](https://download.eclipse.org/eclipse/updates/4.35-I-builds/I20250124-1800), so I'm fairly certain this commit is responsible for this behavior.

The test in question is checking whether anti-aliasing is working correctly, as shown below. The first image is how it currently looks like, the second image is what it's supposed to look like. Note that the second image is also what's shown on e.g. Windows.
image

Weirdly enough, this issue only happens on the Jenkins/GitHub builds which are using Xvnc, but not when executed locally.  Reason seems to be that Xvnc only supports up to [24](https://linux.die.net/man/1/xvnc) bit.
> Specify the pixel depth in bits of the desktop to be created. Default is 24, other possible values are 8, 15, and 16 - anything else is likely to cause strange behaviour by applications. 

So my question: @HeikoKlare is there a technical reason for going from 24 to 32 bit? Because if I understand it correctly, Windows creates at most a 24bit handle:

https://github.com/eclipse-platform/eclipse.platform.swt/blob/857ac0b2c4c7059e7957cd99e46a4501673cc67e/bundles/org.eclipse.swt/Eclipse%20SWT/win32/org/eclipse/swt/graphics/Image.java#L1480-L1487 

The same for MacOS:
https://github.com/eclipse-platform/eclipse.platform.swt/blob/857ac0b2c4c7059e7957cd99e46a4501673cc67e/bundles/org.eclipse.swt/Eclipse%20SWT/cocoa/org/eclipse/swt/graphics/Image.java#L1441

Would it make sense to go back to 24 bit for Linux as well, rather than being the odd one out? Especially if it risks causing visual glitches on systems that only support True Color?

HeikoKlare:
Thank you for the catch, @ptziegler!

> So my question: @HeikoKlare is there a technical reason for going from 24 to 32 bit? Because if I understand it correctly, Windows creates at most a 24bit handle:

I do not remember any technical reason for doing that. Actually, if I remember correctly, the intended change was to simply remove the `else` case (previously kept for GTK 3 compatibility). I have no clue why I changed the color type of the remaining case.

Since both previous cases (if/else, GTK4/3) used a 24 bit color type for the created surface, it should be safe to revert that change. Thus I proposed: 
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/1835

Maybe you want to check whether that PR fixes your test?

PR #1760, opened 2025-01-24
Refactor stream handling for Cocoa and GTK in ImageLoader class

Michael5601:
This PR refactors the stream handling by using try-with-resource for opening streams in the `ImageLoader` class for Cocoa and GTK.


Michael5601:
@HannesWell can you please review and merge this PR?

laeubi:
If closing fails we are probably all lost anyways...

Michael5601:
> We might update the Javadoc as well to clarify that an error will also be thrown when closing the stream fails.

I agree and updated the javadoc accordingly.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/ImageLoader.java
```diff
@@ -168,25 +168,17 @@ public ImageData[] load(InputStream stream) {
  *    <li>ERROR_NULL_ARGUMENT - if the file name is null</li>
  * </ul>
  * @exception SWTException <ul>
- *    <li>ERROR_IO - if an IO error occurs while reading from the file</li>
+ *    <li>ERROR_IO - if an IO error occurs while reading from the file or closing the stream fails</li>
```
laeubi, line 171:
```suggestion
 *    <li>ERROR_IO - if an IO error occurs</li>
```

laeubi:
I think we should not complicate this.... reading includes closing, and that a stream is used is and implementation detail, so simply say it do so when I/O error occurs when reading the file is enough.

HeikoKlare:
> I think we should not complicate this.... reading includes closing, and that a stream is used is and implementation detail, so simply say it do so when I/O error occurs when reading the file is enough.

Agreed 👍

Michael5601:
Now it should be ready.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/ImageLoader.java
```diff
@@ -168,25 +168,17 @@ public ImageData[] load(InputStream stream) {
  *    <li>ERROR_NULL_ARGUMENT - if the file name is null</li>
  * </ul>
  * @exception SWTException <ul>
- *    <li>ERROR_IO - if an IO error occurs while reading from the file</li>
+ *    <li>ERROR_IO - if an IO error occurs while reading the file</li>
```
HannesWell, line 171: This adjustment of the exception description is fine in my opinion, but I also think the previous text was ok.
However we have now an inconsistency with the `ImageLoader` class for Windows were the from is still present.
Plus the `load(InputStream)` method has a similar documentation for `ERROR_IO`, were we should stay consistent too.
So if you want to remove the `from`, please also remove it for Windows and the `load(InputStream)` method, or revert this this removal. As you prefer :)

HannesWell (changes requested):
> I think we should not complicate this.... reading includes closing, and that a stream is used is and implementation detail, so simply say it do so when I/O error occurs when reading the file is enough.

Fully agree, even though Heiko is technically correct, this is a potential change in behavior, I think this is neglectable. In other clean-ups (from myself and others) in other parts of Eclipse the same simplification was already applied multiple times. So I also think this is fine.

HannesWell (approved):

HannesWell:
The failure of browser-tests is unrelated and known, submitting.
Thanks for this code enhancement.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/ImageLoader.java
Michael5601: thank you, did not see it in time. :)

HannesWell: No problem and you are welcome. :)

PR #1762, opened 2025-01-24
[win32] Scale up bounds as rectangle in GC

akoch-yatta:
This commit adresses render artifacts visible when drawing on a GC with monitor-specific scaling enabled e.g. on 175%. Reason is that scaling up all attributes of bounds separately can lead to rounding errors e.g. between y and height (one being scaled up, the other one not). Scaling them together as a rectangle solves this limitation.

<s>This PR is not yet meant to be merged, it should be used to test different scenarios where the GC is involved on uneven zoom settings like 125% or 175%.</s>


HeikoKlare (approved):
Like posted for related PRs: This is part of completing a fix in 2016 (https://github.com/eclipse-platform/eclipse.platform.swt/commit/3b55f60471b210ca4631c9c97fbf4ec77086b0b3) improving the initial HiDPI support (https://github.com/eclipse-platform/eclipse.platform.swt/commit/e02d49aefe42ac4c77b81048299ab069ddb5c2ba) for fractional scale values, which was only applied to according calculations in DPIUtil but not to all consumers that contained the same inaccuracy.

PR #1768, opened 2025-01-27
[win32] Fix scaling inside Region::intersect

akoch-yatta:
This commit adresses possibly invalid result when using Region::intersects on a zoom not dividable by 100. Reason is that scaling up all attributes of the rectangle separately can lead to rounding errors. Scaling them together as a rectangle solves this limitation.


HeikoKlare (approved):
This is part of completing a fix in 2016 (https://github.com/eclipse-platform/eclipse.platform.swt/commit/3b55f60471b210ca4631c9c97fbf4ec77086b0b3) improving the initial HiDPI support (https://github.com/eclipse-platform/eclipse.platform.swt/commit/e02d49aefe42ac4c77b81048299ab069ddb5c2ba) for fractional scale values, which was only applied to according calculations in `DPIUtil` but not to all consumers that contained the same inaccuracy.

PR #1769, opened 2025-01-27
[win32] Unify scaling inside Control::setBounds

akoch-yatta:
This commit unifies scale up results in Control::setBounds. With zoom values not dividable by 100 different result depending on using setBounds(Rectangle) or setBounds(int, int, int, int) were possible. Scaling the bounds always as rectangle solves this limitation.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/ControlWin32Tests.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/ControlWin32Tests.java
@@ -55,6 +55,25 @@ class ControlWin32Tests {
 				fontComparison.originalFontHeight, fontComparison.currentFontHeight);
 	}
 
+	@Test
+	public void testCorrectScaleUpUsingDifferentSetBoundsMethod() {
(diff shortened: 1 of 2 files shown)


HeikoKlare (approved):
This is part of completing a fix in 2016 (https://github.com/eclipse-platform/eclipse.platform.swt/commit/3b55f60471b210ca4631c9c97fbf4ec77086b0b3) improving the initial HiDPI support (https://github.com/eclipse-platform/eclipse.platform.swt/commit/e02d49aefe42ac4c77b81048299ab069ddb5c2ba) for fractional scale values, which was only applied to according calculations in `DPIUtil` but not to all consumers that contained the same inaccuracy.

fedejeanne:
@akoch-yatta this PR caused a regression --> https://github.com/eclipse-platform/eclipse.platform.swt/issues/2003. What was the observable bug that was fixed by this PR? Is it safe to revert?

akoch-yatta:
@fedejeanne this PRs prevents rendering artifacts that would reappear when reverting. I dont't Out of my mind which ones are prevented by this change

HeikoKlare:
fwiw: the mentioned regression should have been fixed via this PR: 
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2047

Issue #669, opened 2023-05-12
Edge Browser freezes indefinitely when launching one Browser from another via a BrowserFunction 


daniemur:
### Summary
Edge `Browser` freezes indefinitely when launching one Browser ("child" Browser) from another ("parent" Browser) via a `BrowserFunction` using `Display#syncExec` to create the child Browser. The same behavior is not exhibited when the child Browser is created within a `Display#asyncExec` call.

Also want to note when SWT gets into this waiting state, closing the window does not kill the process. The process will run until killed manually.

### Code To Reproduce
```
package snippets;

import org.eclipse.swt.SWT;
import org.eclipse.swt.browser.Browser;
import org.eclipse.swt.browser.BrowserFunction;
import org.eclipse.swt.browser.ProgressAdapter;
import org.eclipse.swt.browser.ProgressEvent;
import org.eclipse.swt.layout.FillLayout;
import org.eclipse.swt.widgets.Display;
import org.eclipse.swt.widgets.Shell;

@SuppressWarnings({ "javadoc", "nls" })
public class EdgeSnippet
{
    private static Display display;
    private static Shell shell;

    public static void main(String[] args)
    {
        display = new Display();
        shell = new Shell(display);
        shell.setLayout(new FillLayout());

        Browser parentBrowser = new Browser(shell, SWT.EDGE);
        parentBrowser.setText("<!DOCTYPE html> \n"
                + "<html>\n"
                + "<head> \n"
                + "</head> \n"
                + "<body>\n"
                + "   <button onclick = \"openChildBrowser()\"> Open child browser </button> \n"
                + "   <p> \n"
                + "      <div id = \"result\"> </div> \n"
                + "   </p> \n"
                + "</body> \n"
                + "</html>");
        parentBrowser.addProgressListener(new ProgressAdapter()
        {
            @Override
            public void completed(ProgressEvent event)
            {
                new OpenChildBrowserBrowserFunction(parentBrowser, "openChildBrowser");
            }
        });

        shell.pack();
        shell.open();
        while (!shell.isDisposed())
        {
            if (!display.readAndDispatch())
            {
                display.sleep();
            }
        }
        display.dispose();
    }

    private static class OpenChildBrowserBrowserFunction extends BrowserFunction
    {
        public OpenChildBrowserBrowserFunction(Browser browser, String name)
        {
            super(browser, name);
        }

        @Override
        public Object function(Object[] arguments)
        {
[...]

PR #1770, opened 2025-01-27
[Win32] Summarize ICoreWebView2* instances in WebViewWrapper

amartya4256:
This PR contributes to refactoring of all the webView* instances into a wrapper class WebViewWrapper for better encapsulation and fault proof initialization using a single future.

Contributes to https://github.com/eclipse-platform/eclipse.platform.swt/issues/669


HeikoKlare (approved):
Definitely a reasonable cleanup, thanks! I have just removed the change of the timeout in browser tests, as it was unrelated to this PR (I guess it was accidentally taken over from https://github.com/eclipse-platform/eclipse.platform.swt/pull/1752).

Issue #1771, opened 2025-01-27
Edge: Freeze in Browser#evaluate() if evaluated script calls into SWT


sratz:
**Describe the bug**
With Edge, when using `browser.evaluate(script)` and `script` somehow causing a callback into SWT, the `evaluate()` freezes / hangs in 100% CPU.

**To Reproduce**
```java
@Test
public void test_EvaluateCallingIntoSwt() throws Exception {
	AtomicBoolean initialLoad = new AtomicBoolean();
	AtomicBoolean openWindowListenerCalled = new AtomicBoolean();
	browser.addProgressListener(ProgressListener.completedAdapter(e -> initialLoad.set(true)));
	browser.addOpenWindowListener(event -> {
		event.required = true; // block default
		openWindowListenerCalled.set(true);
	});
	browser.setText("""
			<button id="button" onClick="window.open('https://eclipse.org');">open eclipse.org</button>
			""");
	waitForPassCondition(initialLoad::get);

	browser.evaluate("""
				document.getElementById("button").click();
			""");

	waitForPassCondition(openWindowListenerCalled::get);
}
```

**Expected behavior**
Test succeeds with IE and Edge.

**Actual behavior**
Test hangs with Edge, succeeds with IE.
Stack trace shows:
```
Thread [main] (Suspended)	
	org.eclipse.swt.internal.win32.OS.PeekMessage(org.eclipse.swt.internal.win32.MSG, long, int, int, int) line: not available [native method]	
	org.eclipse.swt.browser.Edge.processNextOSMessage() line: 371	
	org.eclipse.swt.browser.Edge.callAndWait(java.lang.String[], java.util.function.ToIntFunction<org.eclipse.swt.internal.ole.win32.IUnknown>) line: 284	
	org.eclipse.swt.browser.Edge.evaluate(java.lang.String) line: 746	
	org.eclipse.swt.browser.Edge(org.eclipse.swt.browser.WebBrowser).evaluate(java.lang.String, boolean) line: 406	
	org.eclipse.swt.browser.Browser.evaluate(java.lang.String, boolean) line: 665	
	org.eclipse.swt.browser.Browser.evaluate(java.lang.String) line: 614	
	org.eclipse.swt.tests.junit.Test_org_eclipse_swt_browser_Browser.test_EvaluateCallingIntoSwt() line: 2700	
```

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
[...]

Issue #339, opened 2022-08-24
[Edge] No more handle exceptions from Edge browser


vogella:
In our RCP application we see no more handle exceptions during startup if we use Edge. This does not happen all the time, there seems to be a race condition or the error is reported incorrectly similar to https://bugs.eclipse.org/bugs/show_bug.cgi?id=576196.

If the user starts again the RCP application, the errors does not happen again, sometimes is takes 3-4 attempts to start the application. We are using latest version from https://developer.microsoft.com/de-de/microsoft-edge/webview2/#download-section, currently 104.0.1293.63 pointing to the download with  -Dorg.eclipse.swt.browser.DefaultType=edge -Dorg.eclipse.swt.browser.EdgeDir=c:\PROGRA~1\WebView2 in the runtime arguments.

This happens in the Edge.createmethod, relevant stack trace:
```
org.eclipse.swt.SWTError: No more handles [0x8007139f]
at org.eclipse.swt.SWT.error(SWT.java:4944)
at org.eclipse.swt.browser.Edge.error(Edge.java:190)
at org.eclipse.swt.browser.Edge.create(Edge.java:343)
at org.eclipse.swt.browser.Browser.<init>(Browser.java:99)
at de.example.rcp.base.views.NavigationPanelView.createPartControl(NavigationPanelView.java:242)
```

Here is the code which triggers the exception:
```
		try {
			if (hiddenBrowser == null || hiddenBrowser.isDisposed()) {
				hiddenBrowser = new Browser(bar, SWT.EDGE);
				hiddenBrowser.setBounds(0, 0, 0, 0);
				hiddenBrowser.setVisible(false);
			}
		} catch (Exception  | Error  e) {
			LOG.warn("Unable to create hidden browser", e);
		}
```

callAndWait seems to spins the event queue, maybe this causes the other thread to sync with main thread and to dispose the composite?

```
static int callAndWait(long[] ppv, ToIntFunction<IUnknown> callable) {
	int[] phr = {COM.S_OK};
	IUnknown completion = newCallback((result, pv) -> {
		phr[0] = (int)result;
		if ((int)result == COM.S_OK) {
			ppv[0] = pv;
			new IUnknown(pv).AddRef();
		}
		return COM.S_OK;
	});
	ppv[0] = 0;
	phr[0] = callable.applyAsInt(completion);
	completion.Release();
	Display display = Display.getCurrent();
[...]

PR #1773, opened 2025-01-27
[Win32] Spin event loop in Edge instead of only processing OS messages

HeikoKlare:
Before recent improvements of the Edge implementation, it was possible that a browser instance was disposed during WebView initialization, which made the initialization fail exceptionally. This has been worked around by not processing all kinds of asynchronously scheduled events (like a disposal) while waiting for WebView initialization but only those being processes by the OS event queue. This was still necessary to process the OS callbacks for WebView initialization and other operations.

In some cases, this may lead to an Edge browser instance blocking the UI thread, as some asynchronously scheduled tasks need to be processed but are not. In addition, enhancements of the Edge implementation made the initialization happen synchronously anyway, such that the browser instantiation itself cannot throw an error anymore, but the asynchronously happening initialization is now capable of processing asynchronous browser disposals.

This change simplifies the event processing inside Edge to not only process the next OS message but to just ordinarily spin the event loop. This reduces the risk of operations executed on an Edge browser to block execution.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
@@ -261,7 +261,7 @@ static int callAndWait(long[] ppv, ToIntFunction<IUnknown> callable) {
 	// "completion" callback may be called asynchronously,
 	// so keep processing next OS message that may call it
 	while (phr[0] == COM.S_OK && ppv[0] == 0) {
-		processNextOSMessage();
+		spinEventLoop();
 	}
 	completion.Release();
 	return phr[0];
@@ -281,7 +281,7 @@ static int callAndWait(String[] pstr, ToIntFunction<IUnknown> callable) {
 	// "completion" callback may be called asynchronously,
 	// so keep processing next OS message that may call it
 	while (phr[0] == COM.S_OK && pstr[0] == null) {
-		processNextOSMessage();
+		spinEventLoop();
 	}
 	completion.Release();
 	return phr[0];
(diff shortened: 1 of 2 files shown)


sratz (approved):
Looks good. Great to see that removing the old workarounds is possible now.

HeikoKlare:
Thank you, @sratz! I wanted to ping you on this anyway, so great that you already took a look! As you have probably seen, I have also incorporated the regression test you provided in #1771 in a separate commit with you as an author. I hope that's fine for you.

One additional information on this change for which feedback would be good: it introduces a slight change in behavior. Edge now processes _all kinds_ of events scheduled by OS or on the display during initialization and other operations requiring event queue processing. One change in behavior due to this can be seen in the test change. As simplified version is this:
```
Display.getCurrent().asyncExec(() -> {
	shell.dispose();
});
Browser b = new Browser(shell, SWT.None);
b.getUrl();
```
Before this change, that code initialized a browser and `getUrl()` succeeded, as the disposal was only processed when some code spinned the event queue later on. With this change, the disposal is already processed during browser initialization (waited for when calling `getUrl()`), such that `getUrl()` will fail as the browser was never fully initialized due to concurrent disposal. I tested the different use cases that failed to work two years ago (see https://github.com/eclipse-platform/eclipse.platform.swt/issues/339) and they still work fine. And it actually appears to be a bit strange to me to execute operations on the browser while the UI elements in which it is embedded are still being intialized. In other words: I would expect that between browser initialization and calls to blocking functionality (like setting/getting URL, evaluating scripts etc), the event loop will be spinned in productive use anyway. So I do not consider this a problem in productive use. Still, any feedback on this would be welcome.

sratz:
> I have also incorporated the regression test you provided in #1771 in a separate commit with you as an author. I hope that's fine for you.

Sure, was going to suggest to include that test, anyways :)



> it introduces a slight change in behavior

Yes, I have thought about the constructor behavior as well. I was wondering if we can at least make that one somehow synchronous. I think currently it always succeeds and all initialization errors go to nirvana.

I guess this is acceptable for now.

laeubi:
@HeikoKlare I think this is actually a bit dangerous as it can have some side effects (as you already noticed) and code must not assume that a control gets disposed in any way while being in the event thread, e.g. your example w ill not only dispose the browser but **all** controls and might trigger even more code to run.

Just assume the example you given but not *dispose* anything but open a new blocking dialog, then construction of the browser control will suddenly open the dialog even though that from program flow it is not to be expected in any way.

So I wonder what are these 

> some asynchronously scheduled tasks need to be processed but are not.

that are needed at init time of the browser? Can the possibly run once in the firs paint event or similar? Or do they need to run in the UI thread anyways?

HeikoKlare:
> Yes, I have thought about the constructor behavior as well. I was wondering if we can at least make that one somehow synchronous. I think currently it always succeeds and all initialization errors go to nirvana.

Actually it's not only about the constructor, but also about `evalute` calls, which will now spin the event loop and process any kind of events at a place, where it might not be expected. Without spinning the event loop, we will not be able to process situation like in the example of #1771. To me, it looks as if `evalute` would actually have to be processed asynchronously, but since that method returns the result of the evaluation, we cannot easily change that without a breaking API change. It reminds me of the deadlock detection functionality we had before making Edge initialization and some other operations asynchronous.

I agree with @laeubi that processing the whole event loop (and not only OS events) might be dangerous here. I will take a further look at the WebKit implementation, as there are similiar implementations in the `Webkit2AsyncToSync` class, also using timeouts to avoid UI blocks (which we should adapt anyway). And there must also be some mechanism to deal with the situation in the test, as the test succeeds on GTK.

HeikoKlare:
Converted this back to draft, as we agreed that the fix is questionable. Thus, #1771 remains an issue for now and potentially needs a different kind of fix.
At least, the test will now fail with a timeout after instead of non-terminating:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/1776

HeikoKlare:
Closing this as obsolete: we agreed that this is not the right thing to do and related issues (in particular https://github.com/eclipse-platform/eclipse.platform.swt/issues/1771), have now already been resolved by different means.

PR #1776, opened 2025-01-29
[Win32] Avoid blocking operations on Edge due to OS message processing

HeikoKlare:
The Edge browser implementation currently spins the event queue without any possibility to escape in case no OS message or not the expected OS message to proceed occurs and is processed.

In order to avoid that operations called on Edge deadlock, this change introduces timeouts for OS event processing inside Edge:
- The existing processNextOSMessage() method is replaced with processOSMessagesUntil(condition) that spins the event queue until the condition is met or a timeout occurred
- CompletableFutures used for the queuing events on the WebView instance are extended to wake the display in order to avoid that the OS message processing is unnecessary sleeping until a timeout wakes it up

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
@@ -21,6 +21,7 @@ import java.nio.file.Path;
 import java.time.*;
 import java.util.*;
 import java.util.concurrent.*;
+import java.util.concurrent.atomic.*;
 import java.util.function.*;
 
 import org.eclipse.swt.*;
@@ -42,6 +43,7 @@ class Edge extends WebBrowser {
 	static final String APPLOCAL_DIR_KEY = "org.eclipse.swt.internal.win32.appLocalDir";
 	static final String EDGE_USER_DATA_FOLDER = "org.eclipse.swt.internal.win32.Edge.userDataFolder";
 	static final String EDGE_USE_DARK_PREFERED_COLOR_SCHEME = "org.eclipse.swt.internal.win32.Edge.useDarkPreferedColorScheme"; //$NON-NLS-1$
+	static final String WEB_VIEW_OPERATION_TIMEOUT = "org.eclipse.swt.internal.win32.Edge.timeout"; //$NON-NLS-1$
 
 	// System.getProperty() keys
 	static final String BROWSER_DIR_PROP = "org.eclipse.swt.browser.EdgeDir";
@@ -58,6 +60,7 @@ class Edge extends WebBrowser {
 	private static final String ABOUT_BLANK = "about:blank";
 
 	private static final int MAXIMUM_CREATION_RETRIES = 5;
+	private static final Duration MAXIMUM_OPERATION_TIME = Duration.ofMillis(Integer.getInteger(WEB_VIEW_OPERATION_TIMEOUT, 5_000));
 
 	private record WebViewEnvironment(ICoreWebView2Environment environment, ArrayList<Edge> instances) {
 		public WebViewEnvironment(ICoreWebView2Environment environment) {
@@ -260,14 +263,12 @@ static int callAndWait(long[] ppv, ToIntFunction<IUnknown> callable) {
 	phr[0] = callable.applyAsInt(completion);
 	// "completion" callback may be called asynchronously,
 	// so keep processing next OS message that may call it
-	while (phr[0] == COM.S_OK && ppv[0] == 0) {
-		processNextOSMessage();
-	}
+	processOSMessagesUntil(() -> phr[0] != COM.S_OK || ppv[0] != 0, Display.getCurrent());
 	completion.Release();
 	return phr[0];
 }
 
-static int callAndWait(String[] pstr, ToIntFunction<IUnknown> callable) {
+int callAndWait(String[] pstr, ToIntFunction<IUnknown> callable) {
 	int[] phr = new int[1];
 	IUnknown completion = newCallback((result, pszJson) -> {
 		phr[0] = (int)result;
@@ -280,9 +281,7 @@ static int callAndWait(String[] pstr, ToIntFunction<IUnknown> callable) {
 	phr[0] = callable.applyAsInt(completion);
 	// "completion" callback may be called asynchronously,
 	// so keep processing next OS message that may call it
-	while (phr[0] == COM.S_OK && pstr[0] == null) {
-		processNextOSMessage();
-	}
+	processOSMessagesUntil(() -> phr[0] != COM.S_OK || pstr[0] != null, browser.getDisplay());
 	completion.Release();
 	return phr[0];
 }
@@ -298,7 +297,7 @@ class WebViewWrapper {
 
 class WebViewProvider {
 
-	private CompletableFuture<WebViewWrapper> webViewWrapperFuture = new CompletableFuture<>();
+	private CompletableFuture<WebViewWrapper> webViewWrapperFuture = wakeDisplayAfterFuture(new CompletableFuture<>());
 	private CompletableFuture<Void> lastWebViewTask = webViewWrapperFuture.thenRun(() -> {});;
 
 	ICoreWebView2 initializeWebView(ICoreWebView2Controller controller) {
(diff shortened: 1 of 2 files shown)


bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -365,95 +363,91 @@ private ICoreWebView2_13 initializeWebView_13(ICoreWebView2 webView) {
...
 	 */
 	void scheduleWebViewTask(Runnable action) {
-		lastWebViewTask = lastWebViewTask.thenRun(() -> {
-			action.run();
+		lastWebViewTask = wakeDisplayAfterFuture(lastWebViewTask.thenRun(action::run));
+	}
+
+	<T> CompletableFuture<T> wakeDisplayAfterFuture(CompletableFuture<T> future) {
+		return future.handle((nil1, nil2) -> {
+			Display display = Display.getDefault();
```
laeubi, line 432: At best use the display of the browser here as Windows can have multiple displays

HeikoKlare: Indeed, I adapted that.

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -58,6 +59,7 @@ class Edge extends WebBrowser {
 	private static final String ABOUT_BLANK = "about:blank";
 
 	private static final int MAXIMUM_CREATION_RETRIES = 5;
+	private static final Duration MAXIMUM_OPERATION_TIME = Duration.ofSeconds(5);
```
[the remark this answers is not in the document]
HeikoKlare: Good idea. I've extracted the system property name into a separate variable and adapted the name to be consistent with other system properties used by Edge.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
```diff
@@ -302,7 +302,9 @@ private int reportOpenedDescriptors() {
 }
 
 private Browser createBrowser(Shell s, int flags) {
+	// TODO Reduce to reasonable value
 	long maximumBrowserCreationMilliseconds = 90_000;
+	System.setProperty("org.eclipse.swt.internal.win32.Edge.timeout", Long.toString(maximumBrowserCreationMilliseconds));
```
laeubi, line 307:
```suggestion
	System.setProperty("org.eclipse.swt.internal.win32.Edge.timeout", Long.toString(TimeUnit.SECONDS.toMilllis(90));
```

HeikoKlare: I have moved this into properly typed, static constants and according initialization.

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -462,13 +457,21 @@ private <T> void waitForFutureToFinish(CompletableFuture<T> future) {
...
+	AtomicBoolean timeoutOccurred = new AtomicBoolean();
+	// The timer call also wakes up the display to avoid being stuck in display.sleep()
+	display.timerExec((int) MAXIMUM_OPERATION_TIME.toMillis(), () -> timeoutOccurred.set(true));
+	while (!display.isDisposed() && !condition.get() && !timeoutOccurred.get()) {
+		if (OS.PeekMessage(msg, 0, 0, 0, OS.PM_NOREMOVE)) {
+			display.readAndDispatch();
+		} else {
+			display.sleep();
+		}
+	}
```
fedejeanne, line 471: I see that the timer calls `timeoutOccurred.set(true)`, which is supposed to break the `while`, but if the `while` happens to call `display.sleep()` then it could hang indefinitely and therefore never evaluate the break-condition again... **unless** someone calls `Display::wake` (which is currently in `wakeDisplayAfterFuture(future)`).

I think I have the pieces of the puzzle, I just don't see how `wake` would come to the rescue and avoid this `while`-loop to hang indefinitely in the call to `display.sleep()`.

Would you mind explaining that? :-)

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -365,95 +364,91 @@ private ICoreWebView2_13 initializeWebView_13(ICoreWebView2 webView) {
...
 	 * Schedule a given runnable in a queue to execute when the webView is free and
 	 * has finished all the pending tasks queued before it.
 	 */
 	void scheduleWebViewTask(Runnable action) {
-		lastWebViewTask = lastWebViewTask.thenRun(() -> {
-			action.run();
+		lastWebViewTask = wakeDisplayAfterFuture(lastWebViewTask.thenRun(action::run));
+	}
+
+	<T> CompletableFuture<T> wakeDisplayAfterFuture(CompletableFuture<T> future) {
```
fedejeanne, line 431: NIT: it could be `private`

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
laeubi: Yes there is some very small timeframe where one can still run into an issue, but I have no good idea how to completely prevents this without having too much extra code here, so I think it at least an improvement here.

HeikoKlare: I have to admit that I don't get it. Setting `timeoutOccurred` to true happens during OS message processing (namely the timer message processing). This requires a `display.readAndDispatch()` to be executed, which in the mentioned code happens in line 467. I don't see how `display.sleep()` should be executed after the `timerExec` setting `timeoutOccurred` has been processed. Can you explain?

fedejeanne: > Setting timeoutOccurred to true happens during OS message processing (namely the timer message processing)
Ah, that was the part I was missing, thank you! I see now that this is done with `display.timerExec(...)` (emphasis on `display`) which "solves the puzzle" for me :-) .

@laeubi with ...

> Yes there is some very small timeframe where one can still run into an issue,

.. you mean while peeking at the `OS` queue (which might be empty)?

fedejeanne: I ask because this is the only part that I found problematic in the past (https://github.com/eclipse-platform/eclipse.platform.swt/pull/1752#issuecomment-2621521570) namely `OS.PeekMessage(...)` seems to block when the OS queue is empty. But I expect that `display.timerExec(...)` ends up firing an `OS.WM_TIMER` (event) from the OS and this ends up calling `org.eclipse.swt.widgets.Display.runTimer(long)` which in will put something in the queue and call `timeoutOccurred.set(true)`, breaking the deadlock.

HeikoKlare: No, there is nothing putting sometime actively in a queue. It's just that an `OS.WM_TIMER` event is fired which directly wakes up the display (i.e., relieves from `sleep()`). Then `PeekMessage` returns true as there is a message (the time event) in the (OS) queue, which is processed by `readAndDispatch()`, setting `timeoutOccurred` to true, which is why the loop breaks afterwards.
At least, that is the intended flow. Might be that we still have some flaw in there, but that's at least how it is supposed to work.

fedejeanne: 👍 
Sounds good! I'll mark this conversation as resolved since my question has been answered. Thank you both!

fedejeanne (approved):
This PR looks good to go. As soon as the tests run, we can merge.

I re-triggered the pipeline 5 minutes ago since they had failed because of an unreachable URL before (hopefully just a hiccup).

PR #1780, opened 2025-01-30
[GTK4] Implement correct moveAbove()/moveBelow() control behavior

ptziegler:
Regardless of which control are given as input parameters, the controls are currently always added as last child to their parents. Underlying issue seems to be that the swt_fixed_restack() method does not handle the GTK4 widgets correctly. As a solution, perform the movement using the proper API.

To understand the combination in which GTK methods need to be called, consider the expected behavior based on the GTK and SWT documentation:

[GTK4 Documentation]

- gtk_widget_insert_after(widget, parent, previous_sibling)

  It will be placed after previous_sibling, or at the beginning if previous_sibling is NULL.

- gtk_widget_insert_before(widget, parent, next_sibling)

  It will be placed before next_sibling, or at the end if next_sibling is NULL.

[SWT Documentation]

- moveAbove(control)

  Moves the receiver above the specified control in the drawing order. If the argument is null, then the receiver is moved to the top of the drawing order

This means that if the specified control is NULL,
gtk_widget_insert_after(...) needs to be called, otherwise gtk_widget_insert_before(...).

- moveBelow(control)

  Moves the receiver below the specified control in the drawing order. If the argument is null, then the receiver is moved to the bottom of the drawing order.

Here the inverse applies. If the specified control is NULL, gtk_widget_insert_before(...) needs to be called, otherwise gtk_widget_insert_after(...).

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT PI/gtk/library/gtk4.c
+++ b/bundles/org.eclipse.swt/Eclipse SWT PI/gtk/library/gtk4.c
@@ -2083,6 +2083,26 @@ JNIEXPORT jlong JNICALL GTK4_NATIVE(gtk_1widget_1get_1root)
 }
 #endif
 
+#ifndef NO_gtk_1widget_1insert_1after
+JNIEXPORT void JNICALL GTK4_NATIVE(gtk_1widget_1insert_1after)
+	(JNIEnv *env, jclass that, jlong arg0, jlong arg1, jlong arg2)
+{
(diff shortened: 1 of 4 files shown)


ptziegler:
Just a little demo to illustrate how it behaves using Snippet6.

[Recording.webm](https://github.com/user-attachments/assets/e21a12d4-37fa-4426-8846-c5c30a7e3131)

ptziegler:
Everything is green now. Unless there are any objections, I'll plan on merging this later today.

laeubi:
> Everything is green now. Unless there are any objections, I'll plan on merging this later today.

Just go ahead!

PR #1783, opened 2025-01-31
[win32] Unify up-scaling bounds using Rectangle

akoch-yatta:
This commit unifies scale up results in different places. With zoom values not dividable by 100 scaling up via rectangle or scaling up all values separately can give different result. Scaling always as rectangle solves this limitation.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Canvas.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Canvas.java
@@ -200,11 +200,8 @@ public void scroll (int destX, int destY, int x, int y, int width, int height, b
 	int zoom = getZoom();
 	destX = DPIUtil.scaleUp(destX, zoom);
 	destY = DPIUtil.scaleUp(destY, zoom);
-	x = DPIUtil.scaleUp(x, zoom);
-	y = DPIUtil.scaleUp(y, zoom);
-	width = DPIUtil.scaleUp(width, zoom);
-	height = DPIUtil.scaleUp(height, zoom);
-	scrollInPixels(destX, destY, x, y, width, height, all);
+	Rectangle rectangle = DPIUtil.scaleUp(new Rectangle(x, y, width, height), zoom);
+	scrollInPixels(destX, destY, rectangle.x, rectangle.y, rectangle.width, rectangle.height, all);
 }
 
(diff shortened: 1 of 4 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Canvas.java
```diff
@@ -200,11 +200,8 @@ public void scroll (int destX, int destY, int x, int y, int width, int height, b
 	int zoom = getZoom();
 	destX = DPIUtil.scaleUp(destX, zoom);
 	destY = DPIUtil.scaleUp(destY, zoom);
-	x = DPIUtil.scaleUp(x, zoom);
-	y = DPIUtil.scaleUp(y, zoom);
-	width = DPIUtil.scaleUp(width, zoom);
-	height = DPIUtil.scaleUp(height, zoom);
-	scrollInPixels(destX, destY, x, y, width, height, all);
+	Rectangle rect = DPIUtil.scaleUp(new Rectangle(x, y, width, height), zoom);
```
HeikoKlare, line 203: nit: at all other places the variable is called `rectangle`

HeikoKlare (commented):
Changes looks sound to me. I was only wondering about whether scaling destination coordinates independently is correct, but found that to be fine (in particular since source x/y are only used for "correcting" width/height).

Like posted for related PRs: This is part of completing a fix in 2016 (https://github.com/eclipse-platform/eclipse.platform.swt/commit/3b55f60471b210ca4631c9c97fbf4ec77086b0b3) improving the initial HiDPI support (https://github.com/eclipse-platform/eclipse.platform.swt/commit/e02d49aefe42ac4c77b81048299ab069ddb5c2ba) for fractional scale values, which was only applied to according calculations in DPIUtil but not to all consumers that contained the same inaccuracy.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Canvas.java
akoch-yatta: Adapted it 👍

akoch-yatta:
> Changes looks sound to me. I was only wondering about whether scaling destination coordinates independently is correct, but found that to be fine (in particular since source x/y are only used for "correcting" width/height).

Yes, from the usages I found this the correct/best solution

PR #1785, opened 2025-01-31
[Win32] Correct wrong formatting in os.c

HeikoKlare:
Whenever the os.c file is auto-updated (e.g., due to changes in the OS class), a change is generated that fixes an inconsistent formatting. This change updates the os.c file to incorporate that formatting fix.


HeikoKlare:
Only formatting correction. Test failures are known and unrelated.

PR #1786, opened 2025-01-31
[Win32] Ensure and document thread-safety of SWTFontProvider

HeikoKlare:
The SWTFontProvider is Win32-internal for accessing font registries that manage the fonts for a device at different zoom levels. Currently, this class does not consider it's accesses' thread-safety.

This change makes the SWTFontProvider thread-safe under current usage contexts:
- Make the internal map thread-safe as it may be modified from different threads (used for the different devices/displays) concurrently. Using a ConcurrentHashMap ensures that read access stays fast. Write access only happens once per device creation.
- Document non-thread-safety of the provider's operations and the necessity to always call it from the same thread for the same device. All current callers fulfill this requirement, thus no race conditions may occur.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/SWTFontProvider.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/SWTFontProvider.java
@@ -14,6 +14,7 @@
 package org.eclipse.swt.internal;
 
 import java.util.*;
+import java.util.concurrent.*;
 
 import org.eclipse.swt.graphics.*;
 import org.eclipse.swt.widgets.*;
@@ -26,20 +27,46 @@ import org.eclipse.swt.widgets.*;
  * take the provided values for the zoom into consideration and return scaled variant of a font if necessary.
  */
 public class SWTFontProvider {
-	private static Map<Device, SWTFontRegistry> fontRegistries = new HashMap<>();
+	private static final Map<Device, SWTFontRegistry> fontRegistries = new ConcurrentHashMap<>();
 
 	private static SWTFontRegistry getFontRegistry(Device device) {
 		return fontRegistries.computeIfAbsent(device, SWTFontProvider::newFontRegistry);
 	}
 
+	/**
+	 * Returns the system font for the given device at the specified zoom.
+	 *
+	 * <b>Note:</b> This operation is not thread-safe. It must thus always be called
+	 * from the same thread for the same device, such as the display's UI thread.
+	 *
+	 * @param device the device to retrieve the font for, must not be {@code null}
+	 * @param zoom   the zoom for which the font shall be scaled
+	 */
 	public static Font getSystemFont(Device device, int zoom) {
 		return getFontRegistry(device).getSystemFont(zoom);
 	}
 
+	/**
+	 * Returns the font with the given font data for the given device at the
+	 * specified zoom.
+	 *
+	 * <b>Note:</b> This operation is not thread-safe. It must thus always be called
+	 * from the same thread for the same device, such as the display's UI thread.
+	 *
+	 * @param device   the device to retrieve the font for, must not be {@code null}
+	 * @param fontData the data for the font to retrieve, must not be {@code null}
+	 * @param zoom     the zoom for which the font shall be scaled
+	 */
 	public static Font getFont(Device device, FontData fontData, int zoom) {
 		return getFontRegistry(device).getFont(fontData, zoom);
 	}
 
+	/**
+	 * Disposes the font registry for the given device, if one exists.
+	 *
+	 * @param device the device to dispose the font registry for, must not be
+	 *               {@code null}
+	 */
 	public static void disposeFontRegistry(Device device) {
 		SWTFontRegistry fontRegistry = fontRegistries.remove(device);
 		if (fontRegistry != null) {


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/SWTFontProvider.java
```diff
@@ -26,20 +27,45 @@
...
 
 	private static SWTFontRegistry getFontRegistry(Device device) {
 		return fontRegistries.computeIfAbsent(device, SWTFontProvider::newFontRegistry);
 	}
 
+	/**
+	 * Returns the system font for the given device at the specified zoom.
+	 *
+	 * <b>Note:</b> This operation is not thread-safe. It must thus always be called
+	 * from the same thread for the same device, such as the display's UI thread.
```
fedejeanne, line 40: It seems odd to say "it's not thread-safe" while making it thread-safe (implementation detail). How about something more relaxed, like not _guaranteeing_ thread-safety?:

```suggestion
	 * <b>Note:</b> This operation is not guaranteed to be thread-safe. It should thus always be called
	 * from the same thread for the same device, such as the display's UI thread.
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/SWTFontProvider.java
```diff
@@ -26,20 +27,45 @@
...
+	 * @param device   the device to retrieve the font for
+	 * @param fontData the data for the font to retrieve
+	 * @param zoom     the zoom for which the font shall be scaled
+	 */
 	public static Font getFont(Device device, FontData fontData, int zoom) {
 		return getFontRegistry(device).getFont(fontData, zoom);
 	}
 
+	/**
+	 * Disposes the font registry for the given device, if existing.
```
fedejeanne, line 65:
```suggestion
	 * Disposes the font registry for the given device, if it exists.
```

fedejeanne (approved):
Changes look good, just a minor comment regarding the wording in the JavaDoc.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/SWTFontProvider.java
HeikoKlare: I don't get it. Those methods are not thread-safe and will not be made thread-safe by this change, so what is wrong about saying that they are not thread-safe and instead relaxing it to "not guaranteed to be thread-safe"?

fedejeanne: The methods might not be synchronized but they will work just fine for multi-threaded apps calling them with the same device from different threads (because of the usage of `ConcurrentHashMap`).

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/SWTFontProvider.java
```diff
@@ -26,20 +27,45 @@
  * take the provided values for the zoom into consideration and return scaled variant of a font if necessary.
  */
 public class SWTFontProvider {
-	private static Map<Device, SWTFontRegistry> fontRegistries = new HashMap<>();
+	private static Map<Device, SWTFontRegistry> fontRegistries = new ConcurrentHashMap<>();
```
laeubi, line 30: According to the [javadoc](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/concurrent/ConcurrentHashMap.html)

> Like [Hashtable](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/Hashtable.html) but unlike [HashMap](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/HashMap.html), this class does not allow null to be used as a key or value.

So wen should add null checks at the getters for `display` below

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/SWTFontProvider.java
HeikoKlare: But that's not the meaning "thread-safe".

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/SWTFontProvider.java
HeikoKlare: Done with c774924

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/SWTFontProvider.java
merks: It seems slightly odd that it's not final.    I know it wasn't final before and that it doesn't really matter...

HeikoKlare: Indeed, just changed it to final.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/SWTFontProvider.java
fedejeanne: I go by [this one](https://en.wikipedia.org/wiki/Thread_safety)

> In multi-threaded computer programming, a function is thread-safe when it can be invoked or accessed concurrently by multiple threads without causing unexpected behavior, race conditions, or data corruption.

... which doesn't mention at which level the thread-safeness is guaranteed. 

Which definition do you use?

HeikoKlare: Exactly that definition is not fulfilled. What is a "thread-safeness level"?

laeubi: @fedejeanne the cache lookup is save, but the method being called from the cached object might not! Thats how far I understand the issue here...

HeikoKlare: Exactly. And we have already agreed that we can accept this as this class is internal and the methods are (currently) only called from one thread per device (like the display's UI thread). That's why even though the method is not thread-safe, currently no concurrency issues may occur. And that's also why we agreed to add this documentation, as further usages of these methods could of course introduce concurrency issues like race conditions if performed in other threads.

fedejeanne: > ... the cache lookup is save, but the method being called from the cached object might not!
@laeubi you're right, `getSystemFont(zoom)` is not thread-safe. Thank you!

PR #1787, opened 2025-01-31
[Win32] Thread-safe modification of Edge instances

HeikoKlare:
The WebView environment used for Edge manages a list of all currently used Edge instances inside an application. This data structure is modified at different places, which may be executed by different threads, such introducing a risk of race conditions on the unsynchronized data structure. In addition, when disposing a Display (such as at application shutdown), Edge instances are disposed in an according dispose listener. This can cause ConcurrentModificationExceptions as the listener iterates through the list of Edge instances and calls their disposal, which in turn removes the Edge instance from that list.

This change makes the data structure thread-safe and copy-on-write to avoid that ConcurrentModificationException occur when iterating over it while removing elements from it.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
@@ -20,6 +20,7 @@ import java.nio.file.*;
 import java.nio.file.Path;
 import java.time.*;
 import java.util.*;
+import java.util.List;
 import java.util.concurrent.*;
 import java.util.concurrent.atomic.*;
 import java.util.function.*;
@@ -62,9 +63,9 @@ class Edge extends WebBrowser {
 	private static final int MAXIMUM_CREATION_RETRIES = 5;
 	private static final Duration MAXIMUM_OPERATION_TIME = Duration.ofMillis(Integer.getInteger(WEB_VIEW_OPERATION_TIMEOUT, 5_000));
 
-	private record WebViewEnvironment(ICoreWebView2Environment environment, ArrayList<Edge> instances) {
+	private record WebViewEnvironment(ICoreWebView2Environment environment, List<Edge> instances) {
 		public WebViewEnvironment(ICoreWebView2Environment environment) {
-			this (environment, new ArrayList<>());
+			this (environment, new CopyOnWriteArrayList<>());
 		}
 	}
 


bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -62,9 +63,9 @@ class Edge extends WebBrowser {
 	private static final int MAXIMUM_CREATION_RETRIES = 5;
 	private static final Duration MAXIMUM_OPERATION_TIME = Duration.ofMillis(Integer.getInteger(WEB_VIEW_OPERATION_TIMEOUT, 5_000));
 
-	private record WebViewEnvironment(ICoreWebView2Environment environment, ArrayList<Edge> instances) {
+	private record WebViewEnvironment(ICoreWebView2Environment environment, List<Edge> instances) {
 		public WebViewEnvironment(ICoreWebView2Environment environment) {
-			this (environment, new ArrayList<>());
+			this (environment, Collections.synchronizedList(new ArrayList<>()));
```
laeubi, line 68:
```suggestion
			this (environment, new CopyOnWriteArrayList<>());
```

laeubi:
> This change makes the data structure thread-safe and avoids that ConcurrentModificationException occur by iterating through a copy of the list of Edge instances.

I don'tsee a copy in the PR... but if you use `CopyOnWriteArrayList` you can safely iterate the list while it is modified.

HeikoKlare:
Good point, thank you! I pushed an incomplete state where the copy operatoin was missing. But using an copy-on-write data structure makes more sense anyway, as it does not require a local avoidance of concurrent modification when accessing the data structure but just solves that by the type of data structure itself, also making it easier to understand (as it's not immediately clear why there is a need for making a copy of the data structure).

I updated the PR accordingly.

HeikoKlare:
Additional info: this is a stack trace of an exception that is to be avoided by this PR:
```
java.util.ConcurrentModificationException
	at java.base/java.util.ArrayList$Itr.checkForComodification(ArrayList.java:1095)
	at java.base/java.util.ArrayList$Itr.next(ArrayList.java:1049)
	at org.eclipse.swt.browser.Edge.lambda$12(Edge.java:554)
	at org.eclipse.swt.widgets.Display.release(Display.java:3779)
	at org.eclipse.swt.graphics.Device.dispose(Device.java:339)
	at org.eclipse.ui.internal.ide.application.IDEApplication.start(IDEApplication.java:194)
	at org.eclipse.equinox.internal.app.EclipseAppHandle.run(EclipseAppHandle.java:208)
	at org.eclipse.core.runtime.internal.adaptor.EclipseAppLauncher.runApplication(EclipseAppLauncher.java:143)
	at org.eclipse.core.runtime.internal.adaptor.EclipseAppLauncher.start(EclipseAppLauncher.java:109)
	at org.eclipse.core.runtime.adaptor.EclipseStarter.run(EclipseStarter.java:439)
	at org.eclipse.core.runtime.adaptor.EclipseStarter.run(EclipseStarter.java:271)
	at org.eclipse.equinox.launcher.Main.invokeFramework(Main.java:668)
	at org.eclipse.equinox.launcher.Main.basicRun(Main.java:605)
	at org.eclipse.equinox.launcher.Main.run(Main.java:1481)
	at org.eclipse.equinox.launcher.Main.main(Main.java:1454)
```

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -62,9 +63,9 @@ class Edge extends WebBrowser {
 	private static final int MAXIMUM_CREATION_RETRIES = 5;
 	private static final Duration MAXIMUM_OPERATION_TIME = Duration.ofMillis(Integer.getInteger(WEB_VIEW_OPERATION_TIMEOUT, 5_000));
 
-	private record WebViewEnvironment(ICoreWebView2Environment environment, ArrayList<Edge> instances) {
+	private record WebViewEnvironment(ICoreWebView2Environment environment, List<Edge> instances) {
```
laeubi, line 66: I wonder if a record is really useful here. can the instances probably be part of `ICoreWebView2Environment`?

HeikoKlare: `ICoreWebView2Environment` is a native interface (https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2environment?view=webview2-1.0.2957.106), so I guess nothing we could (or at least should) extend.

laeubi: okay than its not an option of course ...

HeikoKlare:
Jenkins failures are known and unrelated (in particular since the change only affects Edge, which is not executed in the Jenkins build).

PR #1792, opened 2025-02-03
Skip some Edge test when testing IE

fedejeanne:
2 tests for Edge were running for IE. This PR corrects that.

## Expected result
The GH Actions shouldn't show these tests running for IE. If you see something like this in the output then it's wrong:

```
Running Test_org_eclipse_swt_browser_Browser#test_Constructor_multipleInstantiationsInDifferentShells[browser flags: 524,288]
```

```
Running Test_org_eclipse_swt_browser_Browser#test_Constructor_multipleInstantiationsInDifferentThreads[browser flags: 524,288]
```

### DIFF
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
@@ -342,6 +342,7 @@ public void test_Constructor_asyncParentDisposal() {
 
 @Test
 public void test_Constructor_multipleInstantiationsInDifferentShells() {
+	assumeTrue("This test is intended for Edge only", isEdge);
 	final int numberOfBrowsers = 5;
 	for (int i = 0; i < numberOfBrowsers; i++) {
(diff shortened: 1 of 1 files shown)


HeikoKlare (commented):
Wouldn't it be better to replace SWT.Edge with swtBrowserSettings, so that the tests are simply executed for IE as well?

fedejeanne:
That would mean we have another 2 tests to execute and maintain in a browser which was discontinued in mid 2022.

fedejeanne:
Failing checks are unrelated

HeikoKlare (commented):
The browser discontinued in mid 2022 is still the default browser for every released Eclipse IDE until today. For quite a while, there will still be people using it. Edge will be introduced as default browser soon, but likely that still extensions need to be made to make it proper replacement for everyone with every usage scenario. So we won't get rid of IE that easily. In addition, you also disable one of the tests for _every_ browser but Edge, including the WebKit implementation for Cocoa and Gtk.

What concerns do you have regarding maintenance? The tests seem to run just fine and in case they start to fail for IE, we can still decide on proper action. Setting `SWT.Edge` as flag for a browser without guarding that use case by running on Windows seems to be improper anyway. For example, `test_Constructor_multipleInstantiationsInDifferentShells` is executed on every OS with flag `SWT.Edge`.

PR #1793, opened 2025-02-03
[GTK4] Guard new dialog API behind a 4.10 version check

ptziegler:
This restricts the GTK4-specific dialogs to only be executed when the system is using at least GTK 4.10. In all other cases, the (deprecated) implementation that was removed with the following commits has been restored:

- FontDialog: 10567bb0a45af7417885a3379c1949c1ad7fee8c
- FileDialog: 4179dc7ac6c38e56b20105f01a9a404a1c808f5a
- DirectoryDialog: 2e61b4be73c51afe0264028534306fff874e2ea9


akurtakov:
I have just tested this patch with ControlExample Directory/FileDialog are fine but FontDialog crashes with:
```
Exception in thread "main" java.lang.IllegalArgumentException: Argument cannot be null
	at org.eclipse.swt.SWT.error(SWT.java:4913)
	at org.eclipse.swt.SWT.error(SWT.java:4847)
	at org.eclipse.swt.SWT.error(SWT.java:4818)
	at org.eclipse.swt.graphics.Font.<init>(Font.java:78)
	at org.eclipse.swt.widgets.FontDialog.open(FontDialog.java:214)
	at org.eclipse.swt.examples.controlexample.DialogTab.createButtonSelected(DialogTab.java:221)
	at org.eclipse.swt.events.SelectionListener$1.widgetSelected(SelectionListener.java:83)
	at org.eclipse.swt.widgets.TypedListener.handleEvent(TypedListener.java:286)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:5857)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1652)
	at org.eclipse.swt.widgets.Display.runDeferredEvents(Display.java:5067)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:4519)
	at org.eclipse.swt.examples.controlexample.ControlExample.main(ControlExample.java:240)
```

ptziegler:
> I have just tested(on gtk4-4.16.5-2.fc41) this patch with ControlExample Directory/FileDialog are fine but FontDialog crashes with:

You're right. I only tested the implementation with Snipet133, so I didn't notice the case in which the dialog is called without an initial font. I've added an additional NULL check in case no font data has been set.

akurtakov:
Thanks for the fast changes! This one is good to go in.

Issue #1789, opened 2025-02-03
Edge: unexpected concurrent processing of OS messages


HeikoKlare:
The Edge implementation for using WebView2 in SWT employs a reduced event queue processing in `processOSMessagesUntil()` to process callback for initialization and other tasks that subsequent operations on an Edge instance have to wait for. The method is supposed to _only_ process messages coming from the OS and not to process any tasks asynchronously scheduled at the display. To this end, the method performs an `OS.PeekMessage()` before calling `display.readAndDisplay()` as that is supposed to only process a single OS message (on Windows):

https://github.com/eclipse-platform/eclipse.platform.swt/blob/4a201e25d53f7a61a52e6205714549919ff3c2f4/bundles/org.eclipse.swt/Eclipse%20SWT%20Browser/win32/org/eclipse/swt/browser/Edge.java#L467-L471

This assumption seems to be broken. I found the following call sequence in an exception during local browser test execution, which shows that when calling `display.readAndDisplay()`, messages in the synchronizer are processed instead of an OS message (which should be there because of the previous condition check and which should be processed instead of the synchornizer messages when it is there):
```
	at org.eclipse.swt.widgets.Synchronizer.runAsyncMessages(Synchronizer.java:132)
	at org.eclipse.swt.widgets.Display.runAsyncMessages(Display.java:4077)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:3693)
	at org.eclipse.swt.browser.Edge.processOSMessagesUntil(Edge.java:468)
```

By concept, this should _not_ happen. There are two potential reasons for this:
- The assumption that `display.readAndDispatch()` only processes an OS message in case one is present does not hold
- Some other thread processed the OS message concurrently

So we should investigate why the conceptual assumptions of the Edge implementation do not hold.

This is the full strack trace of the test for reference:
```
java.util.concurrent.CancellationException
	at java.base/java.util.concurrent.CompletableFuture.cancel(CompletableFuture.java:2478)
	at org.eclipse.swt.browser.Edge$WebViewProvider.abortInitialization(Edge.java:320)
	at org.eclipse.swt.browser.Edge.lambda$14(Edge.java:602)
	at org.eclipse.swt.browser.Edge.lambda$15(Edge.java:608)
	at org.eclipse.swt.browser.Edge.lambda$5(Edge.java:238)
	at org.eclipse.swt.internal.win32.OS.DestroyWindow(Native Method)
[...]

PR #1797, opened 2025-02-03
[Win32] Make Edge only process relevant OS messages #1789

HeikoKlare:
When waiting for initialization of the WebView2 component in the Edge implementation, OS messages are processed by peeking the queue and performing a display.readAndDispatch() in case a message is present. The call to OS.PeekMessage() does, however, already perform some processing of the messages. In fact, Edge is not supposed to process all kinds of OS messages at all (like paint or input events) but should rather only process the WebView2 initialization callback message, arriving as a posted OS message. This change thus limits the message peeking inside Edge to only process the message type PM_QS_POSTMESSAGE.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/1789

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
@@ -510,7 +510,7 @@ private static void processOSMessagesUntil(Supplier<Boolean> condition, Consumer
 	// The timer call also wakes up the display to avoid being stuck in display.sleep()
 	display.timerExec((int) MAXIMUM_OPERATION_TIME.toMillis(), () -> timeoutOccurred.set(true));
 	while (!display.isDisposed() && !condition.get() && !timeoutOccurred.get()) {
-		if (OS.PeekMessage(msg, 0, 0, 0, OS.PM_NOREMOVE)) {
+		if (OS.PeekMessage(msg, 0, 0, 0, OS.PM_NOREMOVE | OS.PM_QS_POSTMESSAGE)) {
 			display.readAndDispatch();
 		} else {
 			display.sleep();


fedejeanne:
@amartya4256 here's the reproducer so you can test locally: https://github.com/vi-eclipse/Eclipse-Platform/issues/223#issuecomment-2630858242

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -510,7 +510,7 @@ private static void processOSMessagesUntil(Supplier<Boolean> condition, Consumer
 	// The timer call also wakes up the display to avoid being stuck in display.sleep()
 	display.timerExec((int) MAXIMUM_OPERATION_TIME.toMillis(), () -> timeoutOccurred.set(true));
 	while (!display.isDisposed() && !condition.get() && !timeoutOccurred.get()) {
-		if (OS.PeekMessage(msg, 0, 0, 0, OS.PM_NOREMOVE)) {
+		if (OS.PeekMessage(msg, 0, 0, 0, OS.PM_NOREMOVE | OS.PM_QS_POSTMESSAGE)) {
```
amartya4256, line 513: Just wondering if we can make it more specific to the edge instance by providing a hwnd? But another concern is, does the filtering mean that the possibility of this falling into display.sleep() is higher?

akoch-yatta (approved):
Without having proper knowledge of the messages queues in windows, limiting peeking into the message queue as much as possible, makes sense to me.

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
HeikoKlare: Might be possible, but we need to be aware that this will not mean that this message is processed by `display.readAndDispatch()` then, as there messages are processed by priority. Posted messages have high priority and will be processed before, e.g., input events, so if there is a posted message, it will always be processed before an eventual input event also being in the queue. 
This change is only avoiding that we run into `display.readAndDispatch()` without a message being in the queue as that has already been processed via this `PeekMessage` call.

The possibility of falling into `display.sleep()` is of course higher, but that's also by intent. We limit processing to specific messages, so instead of processing other messages we will sleep and wait for the next message instead. But since we will wake from `display.sleep()` whenever a new message arrives, this is not an issue. Or do you have any further concerns here?

HeikoKlare: Well, if we don't receive messages (which also means that Edge initialization fails), we run into our timeout. That's what we have the `display.timerExec()` for, setting the timeout value and breaking the loop execution.

amartya4256: Alright, then its safe from the infiite sleeping problem here.

laeubi: By the way I always wondered why there is no `Display.sleep(timeout)` is it not supported on the OS level? How does a "real" native app handle this case?

HeikoKlare: Good question. If you are simply spinning the queue, waiting for messages and processing them, such a timeout should actually not be necessary, as without any messages arriving from the SO there is something completely broken anyway. In the case we have here, if we simply spinned the whole event loop instead of limiting to processing specific messages, we would not run into that situation either (see https://github.com/eclipse-platform/eclipse.platform.swt/pull/1773).

PR #1802, opened 2025-02-04
Calculate Scaling for Font without Device#getDPI() 

ShahzaibIbrahim:
getDPI always returns points for primary monitor and can produce faulty behavior. Printing the code is broken because Printer DPI (400) is different from the Monitor DPI (100). Now we compute points from the zoom level instead of currentFontDPI.

### How to Test

1.  Run the runtime Workspace
2.  Move the shell from 100% primary monitor to 200% secondary monitor
3. Press Ctrl + P 
4. Save the preview as pdf
5. See if the code is printed with correct formatting. 

### Before Fix

image

### After Fix

image

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/StyledText.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/StyledText.java
@@ -304,9 +304,11 @@ public class StyledText extends Canvas {
 				}
 			}
 		}
-		Point screenDPI = styledText.getDisplay().getDPI();
+
 		Point printerDPI = printer.getDPI();
 		resources = new HashMap<> ();
+		int scaleFactorX = printerDPI.x / 100;
+		int scaleFactorY = printerDPI.y / 100;
 		for (int i = 0; i < lineCount; i++) {
 			Color color = printerRenderer.getLineBackground(i, null);
 			if (color != null) {
@@ -323,7 +325,7 @@ public class StyledText extends Canvas {
 			}
 			int indent = printerRenderer.getLineIndent(i, 0);
 			if (indent != 0) {
-				printerRenderer.setLineIndent(i, 1, indent * printerDPI.x / screenDPI.x);
+				printerRenderer.setLineIndent(i, 1, indent * scaleFactorX);
 			}
 		}
 		StyleRange[] styles = printerRenderer.styles;
@@ -367,17 +369,17 @@ public class StyledText extends Canvas {
 			if (!printOptions.printTextFontStyle) {
 				style.fontStyle = SWT.NORMAL;
 			}
-			style.rise = style.rise * printerDPI.y / screenDPI.y;
+			style.rise = style.rise * scaleFactorY;
 			GlyphMetrics metrics = style.metrics;
 			if (metrics != null) {
-				metrics.ascent = metrics.ascent * printerDPI.y / screenDPI.y;
-				metrics.descent = metrics.descent * printerDPI.y / screenDPI.y;
-				metrics.width = metrics.width * printerDPI.x / screenDPI.x;
+				metrics.ascent = metrics.ascent * scaleFactorY;
+				metrics.descent = metrics.descent * scaleFactorY;
+				metrics.width = metrics.width * scaleFactorX;
 			}
 		}
-		lineSpacing = styledText.lineSpacing * printerDPI.y / screenDPI.y;
+		lineSpacing = styledText.lineSpacing * scaleFactorY;
 		if (printOptions.printLineNumbers) {
-			printMargin = 3 * printerDPI.x / screenDPI.x;
+			printMargin = 3 * scaleFactorX;
 		}
 	}
 	/**
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Device.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Device.java
@@ -268,12 +268,20 @@ int computePixels(float height) {
 }
 
 float computePoints(LOGFONT logFont, long hFont) {
-	return computePoints(logFont, hFont, -1);
+	return computePoints(logFont, hFont, SWT.DEFAULT);
 }
 
-float computePoints(LOGFONT logFont, long hFont, int currentFontDPI) {
+float computePoints(LOGFONT logFont, long hFont, int zoom) {
 	long hDC = internal_new_GC (null);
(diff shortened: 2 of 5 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Device.java
```diff
@@ -268,12 +268,20 @@ int computePixels(float height) {
...
-	return computePoints(logFont, hFont, -1);
+	return computePoints(logFont, hFont, SWT.DEFAULT);
 }
 
-float computePoints(LOGFONT logFont, long hFont, int currentFontDPI) {
+float computePoints(LOGFONT logFont, long hFont, int zoom) {
 	long hDC = internal_new_GC (null);
-	int logPixelsY = OS.GetDeviceCaps(hDC, OS.LOGPIXELSY);
+
+	float conversionFactor = 72f;
```
akoch-yatta, line 277: We needed to adapt this calculation because we broke the printing logic when adding this logic in the first place.
We differ between isAutoScalable() == true or false. If false then the original logic is used, if true the provided zoom is mapped to DPI. This does implicitly the same as before because the adjustedZoomFactor that was calculated resulted in the same conversion factor in the end. 
Instead of having this logic here, we could extract it more implicitly into Printer and Display, but that would require an API change to add an additional method to fetch/calculate this factor.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Device.java
```diff
@@ -268,12 +268,20 @@ int computePixels(float height) {
...
+	if (isAutoScalable() && zoom != SWT.DEFAULT) {
+		// For auto scalable devices we need to use a dynamic
+		// DPI value that is extracted from the zoom
+		conversionFactor /= DPIUtil.mapZoomToDPI(zoom);
+	} else {
+		conversionFactor /= OS.GetDeviceCaps(hDC, OS.LOGPIXELSY);
+	}
```
akoch-yatta, line 284: We needed to adapt this calculation because we broke the printing logic when adding this logic in the first place.
We differ between isAutoScalable() == true or false. If false then the original logic is used, if true the provided zoom is mapped to DPI. This does implicitly the same as before because the adjustedZoomFactor that was calculated resulted in the same conversion factor in the end. 
Instead of having this logic here, we could extract it more implicitly into Printer and Display, but that would require an API change to add an additional method to fetch/calculate this factor.

fedejeanne (commented):
FWIW I tested this in Mac and it doesn't produce any regressions. See https://github.com/vi-eclipse/Eclipse-Platform/issues/159#issuecomment-2637894703

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/StyledText.java
```diff
@@ -304,9 +304,11 @@ void cacheLineData(StyledText styledText) {
 				}
 			}
 		}
-		Point screenDPI = styledText.getDisplay().getDPI();
+
 		Point printerDPI = printer.getDPI();
 		resources = new HashMap<> ();
+		int scaleFactorX = printerDPI.x / 100;
```
akoch-yatta, line 310: Having the scale factor being based on the screen DPI leads to unexpected result, e.g. too big/small indentation. Having a screen dpi independent factor leads to consistent results

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/StyledText.java
```diff
@@ -304,9 +304,11 @@ void cacheLineData(StyledText styledText) {
...
+		int scaleFactorY = printerDPI.y / 100;
```
akoch-yatta, line 311: Having the scale factor being based on the screen DPI leads to unexpected result, e.g. too big/small indentation. Having a screen dpi independent factor leads to consistent results

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Device.java
fedejeanne: Do you plan to change something else here or was this conversation resolved?

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/StyledText.java
fedejeanne: I can't test this on my Mac at the moment since I'm not at home but maybe @BeckerWdf could quickly test it?

fedejeanne (commented):
The code looks fine and my previous test (on Mac) was green. I can't test on my Mac at the moment since I'm at the office, but once it's tested then this PR is good to go on my account.

fedejeanne:
FTR I'm talking about testing the very last change (https://github.com/eclipse-platform/eclipse.platform.swt/compare/e789f41c2192b54821d6315a47324aab475b0683..fbc1b3a140d1f78049f6b18c39a3ce0e32c2553a) on Mac

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/StyledText.java
BeckerWdf: I can try. Pls. tell me what I need to to do and what should I look at.

BeckerWdf: I just read: "Move the shell from 100% primary monitor to 200% secondary monitor". I only have my built-in MacBook display available for the next two week. So I cannot test this.

fedejeanne: I see. No problem, I can test this tonight. Thank you anyway!

fedejeanne: It works

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Device.java
fedejeanne: If you expect no more changes then let me know, this PR is ready to go

PR #1806, opened 2025-02-05
Make DPIZoomChangeHandlers concurrent R/W able #62

amartya4256:
Issue: #62 and #127 

This PR changes the type of `DPIZoomChangeRegistry::dpiZoomChangeHandlers` from `TreeMap` to `ConcurrentSkipListMap` as at times there can be classes which can register their handlers amidst a DPI change event leading to concurrent read and write which can lead to a `ConcurrentModificationException`

### Reasoning
1. A `ConcurrentModificationException` happens because while we are processing the DPI change and going over the loop:
```java
for (Entry<Class<? extends Widget>, DPIZoomChangeHandler> entry : dpiZoomChangeHandlers.entrySet()) {
	Class<? extends Widget> clazz = entry.getKey();
	DPIZoomChangeHandler handler = entry.getValue();
	if (clazz.isInstance(widget)) {
		handler.handleDPIChange(widget, newZoom, scalingFactor);
	}
}
``` 
Some Classes which are getting loaded concurrently are trying to register their handler in the `dpiZoomChangeHandlers`. (Verified this that not all the handlers are registered in the first DPI changed event by debugging and counting the number of handlers available at that time and comparing that to the number of handlers at a later stage)

2. Who are these classes which are trying to register handler? Amidst a DPI change event being processed for a shell, all the components (classes) whose instances are contained as children of the shell are already loaded by construct, which means none of the children classes are the ones which are registering at the same time. This rules out the possibility of missing out on any handler during a DPI change handling process. So, these classes which are registering are those classes' instances which are definitely not in the shell and currently being loaded.

3. Which brings us to our third reasoning: making the `dpiZoomChangeHandlers` able to process R/W operations concurrently would fix the issue. We decided to use `ConcurrentSkipListMap` and tested out the following things:
- The order in which the classes are indexed is the same as `TreeMap`, _i.e._ the hierarchy of the classes holds.
- I did a performance test in terms of time in ms for put operation per handler for both the Structures. Here's the result: 

Image

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/DPIZoomChangeRegistry.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/DPIZoomChangeRegistry.java
@@ -15,13 +15,14 @@ package org.eclipse.swt.internal;
 
 import java.util.*;
 import java.util.Map.*;
+import java.util.concurrent.*;
 
 import org.eclipse.swt.*;
 import org.eclipse.swt.widgets.*;
 
 public class DPIZoomChangeRegistry {
 
-	private static Map<Class<? extends Widget>, DPIZoomChangeHandler> dpiZoomChangeHandlers =  new TreeMap<>(
+	private static Map<Class<? extends Widget>, DPIZoomChangeHandler> dpiZoomChangeHandlers =  new ConcurrentSkipListMap<>(
 			(o1, o2) -> {
 	            if(o1.isAssignableFrom(o2)) {
 	            	return -1;


fedejeanne (commented):
This PR looks "harmless enough" and therefore _probably_ OK. It would be nice to be able to reproduce the error though, to see how this PR fixes it.

One pointer: the documentation says that
> Iterators and spliterators are [weakly consistent](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/concurrent/package-summary.html#Weakly).

Which means:
> they are guaranteed to traverse elements as they existed upon construction exactly once, and may (but are not guaranteed to) reflect any modifications subsequent to construction. 

I assume this uncertainty is not a problem in your use case?

amartya4256:
> This PR looks "harmless enough" and therefore _probably_ OK. It would be nice to be able to reproduce the error though, to see how this PR fixes it.

As mentioned in https://github.com/vi-eclipse/Eclipse-Platform/issues/183, it is very difficult to reproduce this exception, as I also wasn't able to reproduce this anyhow. But after looking at the stacktrace, it makes sense that its the concurrency problem as mentioned in the PR.

> > they are guaranteed to traverse elements as they existed upon construction exactly once, and may (but are not guaranteed to) reflect any modifications subsequent to construction.
> 
> I assume this uncertainty is not a problem in your use case?

Exactly Since when we are reading and executing the loop, all the classes which are relevant have already been loaded so we don't care about the newly loaded classes.

fedejeanne:
> Exactly Since when we are reading and executing the loop, all the classes which are relevant have already been loaded so we don't care about the newly loaded classes.

Sorry for insisting here but I want to make sure we are on the same page: "not caring" as in "it's OK if they are also processed (their method `handleDPIChange` is called)" or as in "they shouldn't be processed"?. If it's the latter then this PR may introduce unintended behavior.

fedejeanne:
I'm currently trying to understand the issue and the implications of this change. Can we please postpone merging this PR until tomorrow?

laeubi:
@fedejeanne I'm also a bit confused should not everything be guarde by the display thread? Is maybe a SWT thread check missing somewhere. And as Windows support multiple displays, the map must maybe be per display?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/DPIZoomChangeRegistry.java
```diff
@@ -15,13 +15,14 @@
...
 import java.util.Map.*;
+import java.util.concurrent.*;
 
 import org.eclipse.swt.*;
 import org.eclipse.swt.widgets.*;
 
 public class DPIZoomChangeRegistry {
 
-	private static Map<Class<? extends Widget>, DPIZoomChangeHandler> dpiZoomChangeHandlers =  new TreeMap<>(
+	private static Map<Class<? extends Widget>, DPIZoomChangeHandler> dpiZoomChangeHandlers =  new ConcurrentSkipListMap<>(
```
laeubi, line 25: I just wanted to mention that his will leak classloaders and lead to bundles that are updated/uninstalled retained forever.

laeubi:
May I ask why we need a global static registry at all? Can't we register individual listeners to widgets on demand like it is done for dispose? This currently has the potential to leak classloaders and does not look like a good design. If we need something global, it should be part of the `Display` and probably notify the listener with the actual Widget instance, then the Listener can decide itself if it is responsible or not.

fedejeanne:
> @fedejeanne I'm also a bit confused should not everything be guarde by the display thread? Is maybe a SWT thread check missing somewhere. 

I am currently checking that. AFAIK a `ConcurrentModificationException` can occur if:
a. The same thread modifies the structure it is currently iterating (I've seen this happening when doing recursion and the read-method ends up calling the write-method somewhere "hidden" in the stack)
b. Thread "A" writes while thread "B" is iterating.

I haven't spotted any scenario for case "a" so I was checking "b". No luck so far.

I did prove that the calls to `registerHandler()` may happen later in time _e.g._ if you open the _Declaration_ view and this happens to be the first time you need a Caret...
image
... then the handler for `Caret` is registered later than others (_e.g._ `Shell` or `Menu`).

The part I couldn't break yet is letting the read (`applyChange`) happen **while** someone (say `Caret`) tries to write (`registerHandler`).

w.r.t. your other questions...
> And as Windows support multiple displays, the map must maybe be per display?

and 

> May I ask why we need a global static registry at all? Can't we register individual listeners to widgets on demand like it is done for dispose? This currently has the potential to leak classloaders and does not look like a good design. If we need something global, it should be part of the Display and probably notify the listener with the actual Widget instance, then the Listener can decide itself if it is responsible or not.

IIRC we went with with "propagation (to children)" because using listeners would impact performance as there might be a lot of widgets that need to be registered. Maybe @akoch-yatta can explain it better?

laeubi:
> IIRC we went with with "propagation (to children)" because using listeners would impact performance as there might be a lot of widgets that need to be registered

But actually now it is also notify based on the widget, installing a listener itself should not be that impactful...

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/DPIZoomChangeRegistry.java
fedejeanne: Hm, interesting point, I hadn't seen it before. That would mean that some consumer registers a custom subclass of `Widget` by calling this method and then unloads the bundle that contained that custom subclass, right? If that's the case, some sort of `WeakHashMap` could help or simply offering a `deregister` method, right?

One thing in advance, this method is part of `internal` so I see this issue as having slightly lower priority. 

WDYT?

akoch-yatta: Every custom widget is based on SWT widget, so they inherit the awareness. If there is DPI relevant state in a custom widget, you can refresh that by reacting on a ZoomChanged event.

akoch-yatta:
> May I ask why we need a global static registry at all? Can't we register individual listeners to widgets on demand like it is done for dispose? This currently has the potential to leak classloaders and does not look like a good design. If we need something global, it should be part of the `Display` and probably notify the listener with the actual Widget instance, then the Listener can decide itself if it is responsible or not.

We tried multiple different approaches including using individual listener, but we had different issues with that. Some of our thoughts are outlined in https://github.com/eclipse-platform/eclipse.platform.swt/pull/1064. A lot obviously evolved over time, but most comments probably still hold. We could move this discussion over into an issue. As everything (except the ZoomChanged event) is only used internally, we are quite free to adapt, if we idea for improvements.

laeubi:
> As everything (except the ZoomChanged event) is only used internally, we are quite free to adapt, if we idea for improvements.

I have read through it but I'm not found if you have tried to use

`org.eclipse.swt.widgets.Display.addListener/addFilter(int, Listener)`

to watch on a global level, having a global static registry really does not look the right thing to do here (sorry for not notice that discussion earlier).

So if I really want to be notified globally one can do something like this (if I understand the code correctly):


```
Display display = ...;
display.addFilter(SWT.ZoomChanged, new Listener() {
	
	@Override
	public void handleEvent(Event event) {
		if (event.widget instanceof SomethingInteresting) {
			
		}
	}
});
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/DPIZoomChangeRegistry.java
laeubi: > IIRC one of the key points was: guaranteeing the order in which the event is processed is important.

I can only see that we need different handlers for different classes what is possible and as far as I know ordering is also some kind of guaranteed where I would like to see an example where it is actually important.

fedejeanne:
@laeubi could we move this discussion to https://github.com/vi-eclipse/Eclipse-Platform/issues/232 ?

Changing the current approach would be a major undertaking so I'd like to keep this PR focused on the current approach and the known issue (the concurrent modification).

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/DPIZoomChangeRegistry.java
fedejeanne: Can we move this discussion to https://github.com/vi-eclipse/Eclipse-Platform/issues/232 ?

laeubi: Theoretically yes, but why not at the SWT repo where the code lives and where it actually is implemented and seem to cause issues now?

akoch-yatta: > > IIRC one of the key points was: guaranteeing the order in which the event is processed is important.
> 
> I can only see that we need different handlers for different classes what is possible and as far as I know ordering is also some kind of guaranteed where I would like to see an example where it is actually important.

What does "some kind of guaranteed" mean? You mean by order of registration for the event?
Order is e.g. important when internal state is updated a subclass relies upon. The easiest example for this is updating the nativeZoom in the Widget callback. I don't remember more examples out of my head. There are probably more examples about resources or resizing, but I would need to have a deeper look. I see it like the order of constructor execution: you always call super first, so the super class can make sure it adapt its provided state accordingly first.

One thing we could rethink is externalizing re-layouting, that is triggered inside of the callbacks currently. This was inherited from the initial approach and there were issues in the beginning when we tried to move that, but perhaps those were related to side effects, that we solved in the mean time.

> And then the same question arise: If it really is important how are custom widgets supposed to do it?

Custom widgets are in my point of view quite different from the base widgets in SWT, e.g. they are extensions of existing widgets, a composition of existing widgets or just drawing directly on a GC. 
If there is DPI relevant state in a custom widget, you can refresh that by reacting on a ZoomChanged event, that will be propagated additionally. If you directly refer to the order, then you can probably create issues when you have a hierarchy of custom widgets. We did e.g. a test run with nebula widgets and as far as I remember we only found issues where DPI relevant state was used.

fedejeanne (commented):
-

laeubi:
> Changing the current approach would be a major undertaking so I'd like to keep this PR focused on the current approach and the known issue (the concurrent modification).

The problem is that I think the concurrent modification is actually showing a problem in the current implementation, so "fixing" the exception seems more hiding the problem than being a suitable fix here (just my opinion).

fedejeanne:
> The problem is that I think the concurrent modification is actually showing a problem in the current implementation, so "fixing" the exception seems more hiding the problem than being a suitable fix here (just my opinion).

Moved to https://github.com/vi-eclipse/Eclipse-Platform/issues/232#issuecomment-2639651600 👍

fedejeanne (approved):
I think the current PR is the best we can do without any major changes to the current approach and I propose to merge this PR as-is.

amartya4256:
Added one more commit for logging here: https://github.com/eclipse-platform/eclipse.platform.swt/pull/1806/commits/01317337c6675d2fe771399995d185954824403e

Added it as a separate commit so that we can discuss if this approach makes sense or if we even want this kind of logging at all.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/DPIZoomChangeRegistry.java
```diff
@@ -48,13 +49,17 @@ public static void applyChange(Widget widget, int newZoom, float scalingFactor)
 		if (widget == null) {
 			return;
 		}
+		int initialHandlerCount = dpiZoomChangeHandlers.entrySet().size();
```
fedejeanne, line 52: No need to risk instantiating a new set:
```suggestion
		int initialHandlerCount = dpiZoomChangeHandlers.size();
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/DPIZoomChangeRegistry.java
```diff
@@ -48,13 +49,17 @@ public static void applyChange(Widget widget, int newZoom, float scalingFactor)
...
 		for (Entry<Class<? extends Widget>, DPIZoomChangeHandler> entry : dpiZoomChangeHandlers.entrySet()) {
 			Class<? extends Widget> clazz = entry.getKey();
 			DPIZoomChangeHandler handler = entry.getValue();
 			if (clazz.isInstance(widget)) {
 				handler.handleDPIChange(widget, newZoom, scalingFactor);
 			}
 		}
+		if(initialHandlerCount != dpiZoomChangeHandlers.entrySet().size()) {
```
fedejeanne, line 60:
```suggestion
		if(initialHandlerCount != dpiZoomChangeHandlers.size()) {
```

fedejeanne (commented):
@amartya4256 would it be possible to also save the last registered handler and print it in case the concurrent R/W happens? Or is it too costly?

amartya4256:
> @amartya4256 would it be possible to also save the last registered handler and print it in case the concurrent R/W happens? Or is it too costly?

Hmm finding the added handler means comparing the previous and the new state. Which means we need to preserve the previous state as a copy. I think it might be expensive.

we can do something like this:

```
	private static List<Class<? extends Widget>> noobs = new ArrayList<>();

	public static void applyChange(Widget widget, int newZoom, float scalingFactor) {
		if (widget == null) {
			return;
		}
		noobs.clear();
		for (Entry<Class<? extends Widget>, DPIZoomChangeHandler> entry : dpiZoomChangeHandlers.entrySet()) {
			Class<? extends Widget> clazz = entry.getKey();
			DPIZoomChangeHandler handler = entry.getValue();
			if (clazz.isInstance(widget)) {
				handler.handleDPIChange(widget, newZoom, scalingFactor);
			}
		}
		if(!noobs.isEmpty()) {
			// log noobs
		}
		Event event = new Event();
		event.type = SWT.ZoomChanged;
		event.widget = widget;
		event.detail = newZoom;
		event.doit = true;
		widget.notifyListeners(SWT.ZoomChanged, event);
	}

	public static void registerHandler(DPIZoomChangeHandler zoomChangeVisitor, Class<? extends Widget> clazzToRegisterFor) {
		noobs.add(clazzToRegisterFor);
		dpiZoomChangeHandlers.put(clazzToRegisterFor, zoomChangeVisitor);
	}
```

but I don't know how important of an inforamtion it is. But with this, we avoid accessing the size of the map as that's an extra O(n) 
everytime we call applyChange which is a lot

fedejeanne:
I was thinking more like only the very last registered handler, something like:

```java
public static void registerHandler(DPIZoomChangeHandler zoomChangeVisitor, Class<? extends Widget> clazzToRegisterFor) {
	lastRegistered = clazzToRegisterFor;
	dpiZoomChangeHandlers.put(clazzToRegisterFor, zoomChangeVisitor);
}
```

and then print it when necessary:
```java
public static void applyChange(Widget widget, int newZoom, float scalingFactor) {
	...
	if(initialHandlerCount != dpiZoomChangeHandlers.entrySet().size()) {
		System.out.println(lastRegistered.getCanonicalName() + " registered while processing a DPI Change event.");
	}
	lastRegistered = null; // prevent leaking class loaders
	...
}
```

laeubi:
I really wont convoluted that thing more with loggings especially system out printings... It is either Ok then nothing is needed or it is an error, n wich case an exception should be thrown (like now).

If we think it is an error, than `applyChange` should set a boolean rejecting new registrations during the call throwing an exception at registerHandler... but as mentioned in the other discussion, I think the whole static nature  here is kind of flawed and better be reworked so we won't get into the situation in the first place!

fedejeanne:
Then let's leave the logging and further hacks out, accept the original PR (use `ConcurrentSkipListMap`) and discuss a possible improvement later. M3 is not a good moment to make ground-breaking changes.

Ok @amartya4256 ?

amartya4256:
I'll omit the last commit out.

fedejeanne (approved):

PR #1807, opened 2025-02-05
Set the auto scale method explicitly in ImagesWin32Tests

fedejeanne:
Rename the test class and let it test the "smooth" scale method only, which was the root cause of the problem all along.

Contributes to https://github.com/eclipse-platform/eclipse.platform.swt/pull/1790

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/graphics/ImageSmoothScalingWin32Tests.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/graphics/ImageSmoothScalingWin32Tests.java
@@ -22,8 +22,25 @@ import org.junit.jupiter.api.*;
 import org.junit.jupiter.api.extension.*;
 
 @ExtendWith(PlatformSpecificExecutionExtension.class)
-@ExtendWith(WithMonitorSpecificScalingExtension.class)
-class ImagesWin32Tests {
+class ImageSmoothScalingWin32Tests {
+
+	private static final String SWT_AUTOSCALE_METHOD = "swt.autoScale.method";
+	private static String originalAutoScaleMethod;
+
+	@BeforeAll
+	static void setUpAutoScaleMethodSmooth() {
+		originalAutoScaleMethod = System.getProperty(SWT_AUTOSCALE_METHOD);
+		System.setProperty(SWT_AUTOSCALE_METHOD, "smooth");
+	}
+
+	@AfterAll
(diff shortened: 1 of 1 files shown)


HeikoKlare (commented):
Some thoughts on this:
- The test is somehow specific to "smooth" scaling (at least the test was introduced as a regression test related to that scaling mode).
- The test class is annotated with the `WithMonitorSpecificScalingExtension` to enable monitor-specific scaling, which, in turn, enables "smooth" scaling.

So given that, I would expect one of the following solution:
- Make the test specific to "smooth" scaling, which would include renaming the test and removing the extension
- Make monitor-specific scaling properly enable smooth mode (for which we have a dedicated issue), so that the additional setup logic for "smooth" scaling is not required.

fedejeanne:
I went for option nr. 1: 

>     * Make the test specific to "smooth" scaling, which would include renaming the test and removing the extension

Why?
Removing `@ExtendWith(WithMonitorSpecificScalingExtension.class)` makes it work a proper regression test, which means that we were testing the smooth scaling behavior all along without knowing it. I double checked this assumption by switching the VM parameter `-Dswt.autoScale.method` between `nearest` and `smooth` and again: `nearest` lets the test pass always and `smooth` doesn't (_i.e._, it behaves like a regression test).

HeikoKlare (commented):
Changing the test to focus on smooth scaling is fine for me, but the current solution will not work. `SWT_AUTOSCALE_METHOD` is read only once in the static initialized of `DPIUtil`. There is no guarantee that static intializer is executed after the setup method of the test class. And even if you were able to ensure that, the current restoration logic will not apply as the system property that is reset will not be re-evaluated afterwards.

fedejeanne:
Hm, you're right: as it is, this test passes only by chance and it doesn't restore the previous auto scale method. I guess we will need to wait until the auto scale method can be properly set in runtime (if that's what we want) or we can force the auto scale method somehow, probably with a new method in `DPIUtil`.

I'll postpone this discussion (and this PR) until we discuss how to proceed with https://github.com/vi-eclipse/Eclipse-Platform/issues/227, which is related.

akurtakov:

fedejeanne:
## <TL; DR>
The current status of this test class on `master` is OK so there is no need to apply this PR

## Longer explanation
The test class is already testing the `smooth` method by activating the _monitor-specific UI scaling_ via the annotation `@ExtendWith(WithMonitorSpecificScalingExtension.class)`. 
The annotation also restores the status prior to the test when finishing the test class so all is good.

PR #1809, opened 2025-02-06
Allow smooth scaling for 100% monitors too

fedejeanne:
Smooth scaling is used if:
- (in Windows) Monitor-specific scaling is active
- (in Linux) if the zoom level of the monitor is not divisible by 100
- (in Mac) never

In Linux, once the mode changes to smooth, it stays in smooth.

# How to test
- One 100% monitor and another in a "fractional" zoom _e.g._ 150%
- Start the `ControlExample` with the following VM parameters:
```
```
- (Linux) Drag the application back and forth between monitors

# Expected result
## In Windows
The `autoScaleMethod` should **always** be set to `SMOOTH` (in `org.eclipse.swt.internal.DPIUtil.setDeviceZoom(int)`)

## In Linux
- If the application starts in the monitor with the "non fractional" zoom level (_e.g._ 100% or 200%) then the mode should be `NEAREST`
- The `autoScaleMethod` should be set to `SMOOTH` when starting at/switching to a monitor with a "fractional" zoom level (_e.g._ 150%). 
- Once in `SMOOTH` mode, it should never switch back to `NEAREST`

## In Mac
No changes: never `SMOOTH`.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
@@ -603,7 +603,10 @@ public static void setDeviceZoom (int nativeDeviceZoom) {
 
 	DPIUtil.deviceZoom = deviceZoom;
 	System.setProperty("org.eclipse.swt.internal.deviceZoom", Integer.toString(deviceZoom));
-	if (deviceZoom != 100 && autoScaleMethodSetting == AutoScaleMethod.AUTO) {
+
+	// in GTK, preserve the current method when switching to a 100% monitor
+	boolean preserveScalingMethod = SWT.getPlatform().equals("gtk") && deviceZoom == 100;
+	if (!preserveScalingMethod && autoScaleMethodSetting == AutoScaleMethod.AUTO) {
 		if (sholdUseSmoothScaling()) {
 			autoScaleMethod = AutoScaleMethod.SMOOTH;
 		} else {


fedejeanne:
@HeikoKlare so all in all you propose:
- If it starts in a "non-fractional" monitor (100%, 200%, 400% etc) then use `NEAREST`
- When switching to (or starting at) a "fractional" monitor, use `SMOOTH`
- Once in `SMOOTH`, never go back to `NEAREST`

All of this shall be done for Windows **and** Linux (never for Mac) and only when using the "auto" scale mode.

Is that correct?

fedejeanne:
And of course, in Windows this only applies when activating the monitor-specific scaling

HeikoKlare:
Exactly 👍 Just that I would apply the logic to Linux only. On Windows I would use "nearest" unless monitor-specific scaling is enabled (then use "smooth" but also when deviceZoom is 100).

fedejeanne:
Done @HeikoKlare

fedejeanne:
In a0aebf91fcddd44e956bac20796888dafae0dcff I made sure that the corner case of coming back to a 100% monitor in Linux remains unchanged by this PR _i.e._
- Starting with a 100% monitor: `NEAREST`
- 100% --> 125% --> 100%: `SMOOTH`
- 100% --> 125% --> 200%: `NEAREST`

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
```diff
@@ -603,7 +603,9 @@ public static void setDeviceZoom (int nativeDeviceZoom) {
 
 	DPIUtil.deviceZoom = deviceZoom;
 	System.setProperty("org.eclipse.swt.internal.deviceZoom", Integer.toString(deviceZoom));
-	if (deviceZoom != 100 && autoScaleMethodSetting == AutoScaleMethod.AUTO) {
+	boolean isGtk = SWT.getPlatform().equals("gtk");
+	if (!(isGtk && deviceZoom == 100) // in GTK, preserve the current mode when switching to a 100% monitor
+			&& autoScaleMethodSetting == AutoScaleMethod.AUTO) {
```
HeikoKlare, line 608: The functionality seems to be correct. I would only propose to clean up the code a bit to not have such a complex if statement with an additional comment in there. What about something more simple like:
```suggestion
	boolean preserveScalingMethod = SWT.getPlatform().equals("gtk") && deviceZoom == 100;
	if (!preserveScaling && autoScaleMethodSetting == AutoScaleMethod.AUTO) {
```

fedejeanne: Done in 8dace1dcbfe1ad4dec23bd3eee6a40fdaeac7645

PR #1813, opened 2025-02-07
[win32] Refresh thread dpi awareness on hooks

akoch-yatta:
This commit resets the DPI awareness for the WH_MSGFILTER hook, if called. For performance reasons that is only done, when runAsyncMessages will be triggered, as this can cause UI updates, which would be executed with the wrong DPI awareness context and cause UI glitches.

### How to reproduce

Have monitor specific scaling active and at least two monitors.

1. Open an IDE opened on a primary monitor with a low zoom (e.g. 100%)
2. Open a Java File in the Editor
3. Move the IDE to a secondary monitor with a higher zoom (e.g. 200%)
4. Use the vertical scrollbar to scroll up and down and you should see an effect, e.g. on the LineNumberRuler

https://github.com/user-attachments/assets/13a8e84a-5edf-4eaa-9b92-e9684b6e5a03

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
@@ -1455,7 +1455,7 @@ public Widget findWidget (Widget widget, long id) {
 
 long foregroundIdleProc (long code, long wParam, long lParam) {
 	if (code >= 0) {
-		if (!synchronizer.isMessagesEmpty()) {
+		Runnable processMessages = () -> {
 			sendPostExternalEventDispatchEvent ();
 			if (runMessagesInIdle) {
 				if (runMessagesInMessageProc) {
@@ -1481,6 +1481,15 @@ long foregroundIdleProc (long code, long wParam, long lParam) {
 			int flags = OS.PM_NOREMOVE | OS.PM_NOYIELD | OS.PM_QS_INPUT;
 			if (!OS.PeekMessage (msg, 0, 0, 0, flags)) wakeThread ();
 			sendPreExternalEventDispatchEvent ();
+		};
+		if (!synchronizer.isMessagesEmpty()) {
+			// Windows hooks will inherit the thread DPI awareness from
+			// the process. Whatever DPI awareness was set before on
+			// the thread will be overwritten before the hook is called.
+			// This requires to reset the thread DPi awareness to make
+			// sure, all UI updates caused by this will be executed
+			// with the correct DPI awareness
+			runWithProperDPIAwareness(processMessages);
 		}
 	}
 	return OS.CallNextHookEx (idleHook, (int)code, wParam, lParam);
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
```diff
@@ -3425,7 +3425,17 @@ long msgFilterProc (long code, long wParam, long lParam) {
...
+					// the process. Whatever DPI awareness was set before on
+					// the thread will be overwritten before the hook is called.
+					// This requires to reset the thread DPi awareness to make
+					// sure, all UI updates caused by this will be executed
+					// with the correct DPI awareness
+					long previousDPIAwareness = refreshDPIAwareness();
 					if (runAsyncMessages (false)) wakeThread ();
+					if (previousDPIAwareness > 0) {
+						OS.SetThreadDpiAwarenessContext(previousDPIAwareness);
+					}
```
HeikoKlare, line 3438: What about extracting the whole set/reset logic into a separate method, so that this call looks something like this:
```suggestion
					runWithProperDPIAwareness(() -> {
						if (runAsyncMessages(false)) wakeThread();
					});
```
and `refreshDPIAwareness()` is extended to `runWithProperDPIAwareness(Runnable operation)`

akoch-yatta: Adapted it accordingly

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
```diff
@@ -1456,31 +1456,39 @@ public Widget findWidget (Widget widget, long id) {
...
-			* events and posted messages (PM_QS_INPUT | PM_QS_POSTMESSAGE),
-			* it is possible to cause WaitMessage() to sleep until a new
-			* input event is generated causing sync runnables not to be
-			* executed.
-			*/
-			MSG msg = new MSG();
-			int flags = OS.PM_NOREMOVE | OS.PM_NOYIELD | OS.PM_QS_INPUT;
-			if (!OS.PeekMessage (msg, 0, 0, 0, flags)) wakeThread ();
-			sendPreExternalEventDispatchEvent ();
+				/*
```
HeikoKlare, line 1474: I know that preserving formatting should not be the primary goal when processing changes, but if so much code like here is affected where the history might be interesting for different reasons (e.g., to find the commit when this handling of the "bug in Windows" was introduced, I was wondering whether we could still achieve that here without reducing comprehensibility. For example, wrapping the whole block inside the `if (!synchronizer.isMessagesEmpty()) {` into a `processMessages` runnable or the like, which is then executed with proper DPI awareness inside that if block, i.e.:
```java
if (code >= 0) {
   Runnable processMessages = () -> {
      // all the existing code
   }
   if (!synchronizer.isMessagesEmpty()) {
      runWithProperDPIAwareness(processMessages);
   }
}
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
```diff
@@ -5382,4 +5398,18 @@ private boolean setDPIAwareness(int desiredDpiAwareness) {
 	return true;
 }
 
+private void runWithProperDPIAwareness(Runnable operation) {
+	if (isRescalingAtRuntime()) {
```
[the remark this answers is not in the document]
akoch-yatta: Oh yes, thx for catching that. That would be an annoying regression. I put it in an explicit if-else now. Have a look

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
akoch-yatta: Valid point. I adapted it

Issue #1794, opened 2025-02-03
[GTK4] UnsatisfiedLinkError in Display.flushExposes


ptziegler:
**Describe the bug**

_X11 only_
When using the GTK4 port, trying to create multiple instances of a shell may lead to an error due to calling GTK3-exclusive methods.

**To Reproduce**

Create two or more instances of Snippet133. This issue doesn't happen all the time but should happen at some point.

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [ ] Windows
- - [x] Linux
- - [ ] macOS 

**Stacktrace:**

```
Exception in thread "main" java.lang.UnsatisfiedLinkError: 'long org.eclipse.swt.internal.gtk.GDK.gdk_x11_window_lookup_for_display(long, long)'
	at org.eclipse.swt.internal.gtk.GDK.gdk_x11_window_lookup_for_display(Native Method)
	at org.eclipse.swt.widgets.Display.checkIfEventProc(Display.java:1026)
	at org.eclipse.swt.internal.gtk.OS.XCheckIfEvent(Native Method)
	at org.eclipse.swt.widgets.Display.flushExposes(Display.java:1735)
	at org.eclipse.swt.widgets.Control.update(Control.java:6791)
	at org.eclipse.swt.widgets.Shell.setVisible(Shell.java:3005)
	at org.eclipse.swt.widgets.Shell.open(Shell.java:2050)
	at org.eclipse.swt.snippets.Snippet133.open(Snippet133.java:95)
	at org.eclipse.swt.snippets.Snippet133.main(Snippet133.java:57)
```

ptziegler:
As shown in the stacktrace, the issue is caused by calling [gdk_x11_window_lookup_for_display](https://docs.gtk.org/gdk3-x11/class.X11Window.html) as opposed to the (with 4.18 deprecated) [gdk_x11_surface_lookup_for_display](https://docs.gtk.org/gdk4-x11/type_func.X11Surface.lookup_for_display.html).

This method is used to invalidate a specific area of the window in order to force a repaint operation. But to my knowledge, this is not possible in GTK4; You can only redraw entire widgets, but not sub-sections.

[...]

PR #1820, opened 2025-02-10
[GTK4] Stop handling expose events directly

ptziegler:
The handling of expose events has been internalized with GTK4 making it impossible to e.g. invalidate an area of a surface. Furthermore, the native method that are called no longer exist in GTK4, leading to an UnsatisfiedLinkError.

From the documentation:
Application and widget code should not handle expose events directly; invalidation should use the GtkWidget API, and drawing should only happen inside GtkWidget::draw implementations.

Closes https://github.com/eclipse-platform/eclipse.platform.swt/issues/1794

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
@@ -1015,6 +1015,9 @@ static void checkDisplay (Thread thread, boolean multiple) {
 }
 
 long checkIfEventProc (long display, long xEvent, long userData) {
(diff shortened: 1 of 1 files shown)


akurtakov:
Thanks, this looks like the right thing to do as GdkSurface (GdkWindow replacement in Gtk 4) doesn't provide such api.

akurtakov:
@ptziegler As you are committer now feel free to merge PRs yourself if/when I can't find time to review them in timely manner.

ptziegler:
> @ptziegler As you are committer now feel free to merge PRs yourself if/when I can't find time to review them in timely manner.

You are right, of course. I usually like to wait one or two days in case someone else wants to have a look at it, but I frequently end up forgetting about it because of other tasks popping up. 😅

PR #1822, opened 2025-02-11
[win32] fix to use zoom dpi for fonts via display

akoch-yatta:
With commit cd8bb13 we adapted the calculation how to scale up fonts to fix the calculation for fonts used by the Printer. This introduced an issue if monitor-specific scaling is active and the primary monitor zoom is changed, because two different calculation were mixed now. This commit fixes this inconsistency.

## How to reproduce

- Activate the experimental monitor-specific scaling flag
- Open a text editor
- change the primary zoom

Some fonts will be scaled by the wrong factor
image

@HeikoKlare Can you have a look


fedejeanne (commented):
I can't judge the code itself but I can confirm that the error is fixed by this PR 👍 

Thank you @akoch-yatta !

HeikoKlare (approved):

PR #1823, opened 2025-02-11
Use ImageGcDrawer for bgimages in CTabFolder

akoch-yatta:
This commit addresses rendering fragments that can occur when the background images of CTabFolder will be scaled with the monitor-specific scaling active in win32. Using the ImageGcDrawer to redraw the images on demand for the correct zoom removed the fragments.

### How to reproduce

Activate monitor-specific scaling
Open the IDE on a monitor with 125%
Move the IDE to a monitor with 150%

You will see some fragments in the CTabFolders
image

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
@@ -4007,10 +4007,7 @@ void updateBkImages(boolean colorChanged) {
 						if (colorChanged || !bounds.equals(bkImageBounds[i])) {
 							bkImageBounds[i] = bounds;
 							if (controlBkImages[i] != null) controlBkImages[i].dispose();
-							controlBkImages[i] = new Image(control.getDisplay(), bounds);
-							GC gc = new GC(controlBkImages[i]);
-							renderer.draw(CTabFolderRenderer.PART_BACKGROUND, 0, bounds, gc);
-							gc.dispose();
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
```diff
@@ -4007,10 +4007,7 @@ void updateBkImages(boolean colorChanged) {
 						if (colorChanged || !bounds.equals(bkImageBounds[i])) {
 							bkImageBounds[i] = bounds;
 							if (controlBkImages[i] != null) controlBkImages[i].dispose();
-							controlBkImages[i] = new Image(control.getDisplay(), bounds);
-							GC gc = new GC(controlBkImages[i]);
-							renderer.draw(CTabFolderRenderer.PART_BACKGROUND, 0, bounds, gc);
-							gc.dispose();
+							controlBkImages[i] = new Image(control.getDisplay(), (gc, imageWidth, imageHeight) -> renderer.draw(CTabFolderRenderer.PART_BACKGROUND, 0, bounds, gc), bounds.width, bounds.height);
```
HeikoKlare, line 4010: Shouldn't the call to `draw` pass some bounds depending on passed `imageWidth`/`imageHeight` instead of the bounds from outside?

akoch-yatta: Practically not. As long as a drawer is not used for different images with different size, there will be no difference. As we would need to use x and y from bounds anyway, I wouldn't mix that up and create a new rectangle like new Rectangle(bounds.x, bounds.y, imageWidth, imageHeight)

HeikoKlare (approved):
Using an ImageGcDrawer here makes sense and the change properly fixes the described issue.

PR #1824, opened 2025-02-11
Guard Image Type Icon from Smooth Scaling

ShahzaibIbrahim:
Enabling the smooth scaling for icon produce erroneous result. Resulting in icon losing its transparency.

### How to Test

Use the snippet to see the difference in behavior of the Image and Icon during the DPI change. 
```
shell.setText("Image and Icon Example");
shell.setLayout(new RowLayout());

Image icon = display.getSystemImage(SWT.ICON_WORKING);
InputStream is = ImageExample.class.getResourceAsStream("eclipse.png");
Image image = new Image(display, is);

new Label(shell, SWT.NONE).setImage(icon);
new Label(shell, SWT.NONE).setImage(image);
shell.pack();

shell.open();

while (!shell.isDisposed()) {
		display.sleep();
	}
}
display.dispose();
```

- Run the snippet with -Dswt.autoScale.updateOnRuntime=true
- Move the shell from primary (100%) monitor to secondary (200%) Or use any other combination of zooms
- See if the icons losing transparency. 

### Before Fix

image

### After Fix 

image

**Note: Image is only added in snippet to show smooth scaling still applicable for bitmap images.**

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
@@ -292,8 +292,8 @@ private static ImageData autoScaleImageData (Device device, final ImageData imag
 	int height = imageData.height;
 	int scaledWidth = Math.round (width * scaleFactor);
 	int scaledHeight = Math.round (height * scaleFactor);
-	return switch (autoScaleMethod) {
-	case SMOOTH -> {
+	boolean useSmoothScaling = autoScaleMethod == AutoScaleMethod.SMOOTH && imageData.getTransparencyType() != SWT.TRANSPARENCY_MASK;
+	if (useSmoothScaling) {
 		Image original = new Image (device, (ImageDataProvider) zoom -> imageData);
 		/* Create a 24 bit image data with alpha channel */
 		final ImageData resultData = new ImageData (scaledWidth, scaledHeight, 24, new PaletteData (0xFF, 0xFF00, 0xFF0000));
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
```diff
@@ -294,6 +294,10 @@ private static ImageData autoScaleImageData (Device device, final ImageData imag
 	int scaledHeight = Math.round (height * scaleFactor);
 	return switch (autoScaleMethod) {
 	case SMOOTH -> {
+        if (imageData.maskPad == SWT.TRANSPARENCY_MASK) {
```
akoch-yatta, line 297: Two remarks from my side:

- Indentation looks wrong
- isn't imageData.getTransparencyType() == SWT.TRANSPARENCY_MASK  the better check regarding API? I'm more for using API methods outside of the class

HeikoKlare: Agree with the transparency type check.

In addition, this duplicates the `scaledTo` call and introduces exceptional behavior inside the `smooth` case. I would be in favor of replacing the `switch(autoScaleMethod)` with something making the condition for using smooth scaling more explicit, like:
```java
boolean useSmoothScaling = autoScaleMethod == SMOOTH && imageData.getTransparencyType != SWT.TRANSPARENCY_MASK`;
if (useSmoothScaling) {
   ...
} else {
   imageData.scaledTo(scaledWidth, scaledHeight);
}
```

HeikoKlare (approved):

PR #1825, opened 2025-02-12
Removal for OS.LOGPIXELS* usage where possible #62

amartya4256:
This PR removes the usage of OS.LOGPIXELS* in a few places in the code wherever these values can be obtained using other means as the OS.LOGPIXELS provides the dpi values of the primary monitor at the app-startup time. It's usage have been either removed completely (wherever possible) or have been replaced by DPI-Aware OS.GetDpiForMonitor calls.

contributes to #62 and #127

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/ScalingSWTFontRegistry.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/ScalingSWTFontRegistry.java
@@ -77,7 +77,7 @@ final class ScalingSWTFontRegistry implements SWTFontRegistry {
 
 	@Override
 	public Font getSystemFont(int zoom) {
-		ScaledFontContainer container = getOrCreateBaseSystemFontContainer(device);
+		ScaledFontContainer container = getOrCreateBaseSystemFontContainer(device, zoom);
 
 		Font systemFont = container.getScaledFont(zoom);
 		if (systemFont != null) {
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/FontDialog.java
```diff
@@ -195,14 +195,15 @@ public FontData open () {
 	if (effectsVisible) {
 		lpcf.Flags |= OS.CF_EFFECTS;
 	}
+	int [] dpiX = new int[1];
+	int [] dpiY = new int[1];
+	OS.GetDpiForMonitor(getParent().getMonitor().handle, OS.MDT_EFFECTIVE_DPI, dpiX, dpiY);
```
HeikoKlare, line 200: Is it safe to do an unguarded call to `GetDpiForMonitor` here? The method has been introduced with Windows 8.1, so it will break for older Windows versions. I would usually be fine with that, but I do not want to respond to complains of people still using Windows 7 for whatever reason again if we can avoid it...

HeikoKlare: Can't we just use `getParent().getZoom()` and transform that to DPI via `DPIUtil.mapZoomToDPI()`?

HeikoKlare (commented):
Can we please split this up into two separate PRs addressing each of the usages independently (i.e., one for the font registry and one for the font dialog)?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/FontDialog.java
amartya4256: Yes that's something we can do. But what if there's a monitor where dpiX and dpiY values differ? Or we don't have such case? Because here we specifically use dpiY values.

HeikoKlare: That would affect all the other places where DPI-related values have been replaced with zoom as well. I am not sure whether there is any reasonable displaying device with different DPI on x and y axis (today).

amartya4256: alright then I'll use this here: `DPIUtil.mapZoomToDPI(getParent().getZoom())`

amartya4256:
Instead of splitting, is it okay if I put them in two commits in the same PR? Since, under an umbrella they all are basically the same thing: to replace the usage of logPixelsY, but the only differenc is that it has been removed in 2 different contexts.

What would you suggest @HeikoKlare ?

HeikoKlare:
I would be in favor of having separate PRs whereever possible (i.e., if changes do not depend on each other). It eases reviewing in testing. Once you are confident with one change but maybe not with the other, you can simply merge the former and keep working/testing on the latter.

PR #1828, opened 2025-02-13
Prepare the implementation of dynamically sized image file-formats

HannesWell:
Required for
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/1638

This is a purely internal change that also reduces the number of places where `IOExceptions` are handled (all int the same way) in `FileFormat.isFileFormat()` implementations.

@HeikoKlare do you want to review this? Should we aim to land this for this release?
CC @Michael5601

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/image/FileFormat.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/image/FileFormat.java
@@ -20,6 +20,7 @@ import java.util.function.*;
 
 import org.eclipse.swt.*;
 import org.eclipse.swt.graphics.*;
+import org.eclipse.swt.internal.DPIUtil.*;
 
 /**
  * Abstract factory class for loading/unloading images from files or streams
@@ -54,27 +55,60 @@ public abstract class FileFormat {
 		} catch (NoClassDefFoundError e) { } // ignore format
 	}
 
+	public static final int DEFAULT_ZOOM = 100;
+
+	private static Optional<FileFormat> determineFileFormat(LEDataInputStream stream) {
+		return FORMAT_FACTORIES.stream().skip(1).map(Supplier::get).filter(f -> {
+			try {
+				return f.isFileFormat(stream);
+			} catch (IOException e) {
+				return false;
+			}
+		}).findFirst();
+	}
+
+	private static final int MAX_SIGNATURE_BYTES = 18 + 2; // e.g. Win-BMP or OS2-BMP plus a safety-margin
+
+	public static boolean isDynamicallySizableFormat(InputStream is) {
+		Optional<FileFormat> format = determineFileFormat(new LEDataInputStream(is, MAX_SIGNATURE_BYTES));
+		return format.isPresent() && !(format.get() instanceof StaticImageFileFormat);
+	}
+
+	static abstract class StaticImageFileFormat extends FileFormat {
+
+		abstract ImageData[] loadFromByteStream();
+
+		@Override
+		List<ElementAtZoom<ImageData>> loadFromByteStream(int fileZoom, int targetZoom) {
+			return Arrays.stream(loadFromByteStream()).map(d -> new ElementAtZoom<>(d, fileZoom)).toList();
+		}
+	}
+
 	LEDataInputStream inputStream;
 	LEDataOutputStream outputStream;
 	ImageLoader loader;
 	int compression;
 
-/**
- * Return whether or not the specified input stream
- * represents a supported file format.
- */
-abstract boolean isFileFormat(LEDataInputStream stream);
+	/**
+	 * Return whether or not the specified input stream represents a supported file
+	 * format.
+	 */
+	abstract boolean isFileFormat(LEDataInputStream stream) throws IOException;
 
-abstract ImageData[] loadFromByteStream();
+	/**
+	 * Format that do not implement {@link StaticImageFileFormat} MUST return
+	 * {@link ImageData} with the specified {@code targetZoom}.
+	 */
+	abstract List<ElementAtZoom<ImageData>> loadFromByteStream(int fileZoom, int targetZoom);
 
 /**
  * Read the specified input stream, and return the
  * device independent image array represented by the stream.
  */
-public ImageData[] loadFromStream(LEDataInputStream stream) {
+public List<ElementAtZoom<ImageData>> loadFromStream(LEDataInputStream stream, int fileZoom, int targetZoom) {
 	try {
 		inputStream = stream;
-		return loadFromByteStream();
+		return loadFromByteStream(fileZoom, targetZoom);
 	} catch (Exception e) {
 		if (e instanceof IOException) {
 			SWT.error(SWT.ERROR_IO, e);
@@ -89,14 +123,14 @@ public ImageData[] loadFromStream(LEDataInputStream stream) {
  * Read the specified input stream using the specified loader, and
  * return the device independent image array represented by the stream.
  */
-public static ImageData[] load(InputStream is, ImageLoader loader) {
+public static List<ElementAtZoom<ImageData>> load(InputStream is, ImageLoader loader, int fileZoom, int targetZoom) {
 	LEDataInputStream stream = new LEDataInputStream(is);
-	FileFormat fileFormat = FORMAT_FACTORIES.stream().skip(1) //
-			.map(Supplier::get).filter(f -> f.isFileFormat(stream)) //
-			.findFirst().orElse(null);
- 	if (fileFormat == null) SWT.error(SWT.ERROR_UNSUPPORTED_FORMAT);
+	FileFormat fileFormat = determineFileFormat(stream).orElseGet(() -> {
+		SWT.error(SWT.ERROR_UNSUPPORTED_FORMAT);
+		return null;
+	});
 	fileFormat.loader = loader;
-	return fileFormat.loadFromStream(stream);
+	return fileFormat.loadFromStream(stream, fileZoom, targetZoom);
 }
 
 /**
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/InternalImageLoader.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/InternalImageLoader.java
@@ -14,13 +14,15 @@
 package org.eclipse.swt.graphics;
 
 import java.io.*;
+import java.util.*;
 
+import org.eclipse.swt.internal.DPIUtil.*;
 import org.eclipse.swt.internal.image.*;
 
 class InternalImageLoader {
 
-	static ImageData[] load(InputStream stream, ImageLoader imageLoader) {
-		return FileFormat.load(stream, imageLoader);
+	static List<ElementAtZoom<ImageData>> load(InputStream stream, ImageLoader imageLoader, int fileZoom, int targetZoom) {
+		return FileFormat.load(stream, imageLoader, fileZoom, targetZoom);
 	}
 
 	static void save(OutputStream stream, int format, ImageLoader imageLoader) {
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/ImageLoader.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/ImageLoader.java
@@ -18,6 +18,8 @@ import java.io.*;
 import java.util.*;
 
 import org.eclipse.swt.*;
+import org.eclipse.swt.internal.DPIUtil.*;
+import org.eclipse.swt.internal.image.*;
 
 /**
  * Instances of this class are used to load images from,
@@ -148,10 +150,16 @@ void reset() {
  * </ul>
  */
 public ImageData[] load(InputStream stream) {
+	load(stream, FileFormat.DEFAULT_ZOOM, FileFormat.DEFAULT_ZOOM);
+	return data;
+}
+
+List<ElementAtZoom<ImageData>> load(InputStream stream, int fileZoom, int targetZoom) {
 	if (stream == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
 	reset();
-	data = InternalImageLoader.load(stream, this);
-	return data;
+	List<ElementAtZoom<ImageData>> images = InternalImageLoader.load(stream, this, fileZoom, targetZoom);
+	data = images.stream().map(ElementAtZoom::element).toArray(ImageData[]::new);
+	return images;
 }
 
 /**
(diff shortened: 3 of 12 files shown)


bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/image/FileFormat.java
```diff
@@ -54,27 +54,47 @@ public abstract class FileFormat {
...
+		@Override
+		ImageData[] loadFromByteStream(int zoom) {
+			if (zoom == STATIC_ZOOM_FACTOR) {
+				return loadFromByteStream();
+			}
+			return Arrays.stream(loadFromByteStream()) //
+					.map(d -> {
+						int width = (int) Math.round(d.width * zoom / 100.);
+						int height = (int) Math.round(d.height * zoom / 100.);
+						return d.scaledTo(width, height);
```
HeikoKlare, line 72: `ImageData#scaledTo()` has no shortcut returning itself when desired width/height are equal to original width/height, so this will always execute a scaling algorithm that in most cases will return a result equivalent to the input. We could also improve `ImageData#scaledTo()` to return itself in these cases, but that would also change the (implicit) contract of that method as you are not guaranteed to receive a new ImageData object anymore. So I would propose to check for zoom == 100 here and then simply return `d` instead of `d.scaledTo(...)`.

HeikoKlare:
> Should we aim to land this for this release?

For me, it would be fine to bring this into the upcoming release. I do not see any risk in it (give the adaptation I have mentioned in the review) and it would make preparation for the next release cycle in terms of follow-up PRs regarding SVG support easier.

iloveeclipse:
I would postpone this to 4.36, given that we've just shipped M3, we are in RC phase and the change is not trivial.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/image/FileFormat.java
```diff
@@ -54,27 +54,47 @@ public abstract class FileFormat {
 		} catch (NoClassDefFoundError e) { } // ignore format
 	}
 
+	public static final int STATIC_ZOOM_FACTOR = 100;
+
+	static abstract class StaticFileFormat extends FileFormat {
```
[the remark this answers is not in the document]
HannesWell: > I'd propose to name this `StaticImageFileFormat` as the "static" refers to the image an not the file format.

That's a good point. Applied it.

> Or maybe even more precise call it `RasterizedImageFileFormat` to be later extended with `VectorizedImageFileFormat`?

That's IMO a bit too specific. The main point is that the image has a fixed zoom respectively cannot be 'really' re-renderer with another zoom. And for that StaticImageFileFormat seem to be a good, relatively general and simple name.
And there isn't another class for vectorized images necessary later as the FileFormat base class becomes 'dynamic' with this change.
But in the end this is all internal and therefore we could also change names later if necessary.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/image/FileFormat.java
HannesWell: > So I would propose to check for zoom == 100 here and then simply return d instead of d.scaledTo(...).

Isn't this already handled this way because the method calls and returns the result of `loadFromByteStream()` if `zoom == STATIC_ZOOM_FACTOR` at the top?

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/image/FileFormat.java
```diff
@@ -54,27 +54,47 @@ public abstract class FileFormat {
 		} catch (NoClassDefFoundError e) { } // ignore format
 	}
 
+	public static final int STATIC_ZOOM_FACTOR = 100;
```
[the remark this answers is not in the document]
HannesWell: > So in total, maybe just `DEFAULT_ZOOM`?

Sounds good as well. Changed it.

HannesWell:
> > Should we aim to land this for this release?
> 
> For me, it would be fine to bring this into the upcoming release. I do not see any risk in it (give the adaptation I have mentioned in the review) and it would make preparation for the next release cycle in terms of follow-up PRs regarding SVG support easier.

I second that.

> I would postpone this to 4.36, given that we've just shipped M3, we are in RC phase and the change is not trivial.

It's right that this is not trivial, but in the end it's just moving around some small parts of the code. And the only really new code-path, i.e. the path in `StaticImageFileFormat.loadFromByteStream()` if zoom != 100 is currently not called because that method is only called with a zoom of 100.
Therefore I would say that it's not a risky change.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/image/FileFormat.java
HeikoKlare: Indeed, sorry, I missed that one. You have already perfectly covered that case 🙂

HeikoKlare:
I also see no risk in the change since, as you explained, it is just some renaming and simple refactoring and new code paths are not executed at all. Still, I understand the concerns, in particular since we are beyond M3 and even though the change is unrisky, it does not look like that when only taking a quick look.

So I am not sure what to do. We should not push this through despite raised concerns. It's not worth to ask PMC for an advise/decision either (and you are part of that anyway 😉 ).

laeubi:
@HeikoKlare as far as I understand the only reason is "make PRs happen earlier", what I usually do in this situation is:

- Fetch the PR into my fork
- Create a new branch based on this and push it with my PR
- later simply rebase my PR after this is merged

This also will give some more confidence that what is done here is actually the right thing, and as `master` is effectively locked now there is also no risk of "missing" things while master evolve, so I also think its better to wait (and maybe improve).

Once master is open we can merge this right away and maybe even then more PRs.

HannesWell:
> So I am not sure what to do. We should not push this through despite raised concerns. It's not worth to ask PMC for an advise/decision either (and you are part of that anyway 😉 ).

While I agree with you that this change should be save, I think we should postpone it if others would be uncomfortable with having it, as it's not that important. As we had discussed already privately with Michael and as Christoph has described it the other PRs can include this PR's change temporarily.

HeikoKlare:
Sure, I am completely fine with everything said. Just one remark on:

> as far as I understand the only reason is "make PRs happen earlier"

For me, that's not what it is about for me. It is rather two other reasons:
1. Higher effort for creating and reviewing changes on top of another: I know how to do it (technically), we do that all the time during freeze, but still it produces additional effort.
2. Fine-grained feedback: the more changes we postpone to after the freeze period, the more is merged together in a short period of time making it more difficult to identify what has possibility cause regressions (or the other way round: which changes worked fine, e.g., in I-Builds) -- rapid feedback, small increments, ...

That does not outweigh accepting any risk, so don't take this as an argumentation for merging. It's just to explain why for me it's not only about getting things into a specific release but about efficiently using the time we have with (and without) freeze periods :-)

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
```diff
@@ -778,6 +778,7 @@ public Image(Device device, ImageFileNameProvider imageFileNameProvider) {
 	super(device);
 	if (imageFileNameProvider == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
 	this.imageFileNameProvider = imageFileNameProvider;
+	//TODO: implement fine-grained zoom handling here as well?
```
HannesWell, line 781: For Mac the implementation seems to be completely different than for the other platforms and it even looks to me that anything else than 100% and 200% is not considered?

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/ImageLoader.java
```diff
@@ -166,6 +167,14 @@ public ImageData[] load(InputStream stream) {
 	return imgDataArray;
 }
 
+private List<ElementAtZoom<ImageData>> load(InputStream stream, int fileZoom, int targetZoom) {
+	if (stream == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
+	reset();
+	ImageData[] imgDataArray = getImageDataArrayFromStream(stream);
+	data = imgDataArray;
```
HannesWell, line 174: We have to discuss when we use the native implementation, which is not capable (currently) to render SVGs for different zooms and when to use the software implementations of the file-format.

At the moment I would say that for SVGs the software solution should be used if `fileZoom!=targetZoom` and for all other formats the currently used native impl should be used?
Alternatively one could always use the software SVG-renderer for consistency or we could make it configurable by a flag for the case `fileZoom==targetZoom`?

Deciding in advance if the stream is an SVG or not is possible (the stream is read into a byte-array below anyways) but not trivial.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
HeikoKlare: From my understanding, MacOS/Cocoa is only capable of 100% and 200% displays and thus everything in the SWT implementation for MacOS is to 100/200 as well.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/ImageLoader.java
HeikoKlare: I don't have a strong opinion here. I would be in favor the following:
- Use the SVG software solution as default. Why? We then have consistent behavior across the operating systems and will not run into Linux-specific issues because of the native rendering, thus reducing maintenance efforts.
- Make the native implementation (either for fileZoom==targetZoom or even for every zoom) optional. Maybe having a system property for SVG rendering with arguments "native" and "nativeAtDefault" or the like would suffice to capture every desired use case. On the other hand, every valid argument for system properties need to be maintained and considered everywhere necessary, so introducing another maintenance overhead which maybe no one ever needs. So maybe when setting a property to use native SVG support (on Linux), only rendering at 100% is supported to exactly cover the case is supported right now (and other resolutions are done by auto-scaling), unless someone implements anything additionally?

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
HannesWell: I have now implemented something that seemed suitable just from looking at the code on a Windows computer (no way to try it out), but I'm absolutly not sure if that's reasonable. And I have the impression that code should also be enhanced, because currently it loads the image twice in some cases which might be unnecessary and could be avoided.

In any case it would be good to get support from somebody who has a Mac at hand to try this out and to get some details.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/ImageLoader.java
HannesWell: >     * Use the SVG software solution as default. Why? We then have consistent behavior across the operating systems and will not run into Linux-specific issues because of the native rendering, thus reducing maintenance efforts.
> 
Agree.

>     * Make the native implementation (either for fileZoom==targetZoom or even for every zoom) optional. Maybe having a system property for SVG rendering with arguments "native" and "nativeAtDefault" or the like would suffice to capture every desired use case. On the other hand, every valid argument for system properties need to be maintained and considered everywhere necessary, so introducing another maintenance overhead which maybe no one ever needs. So maybe when setting a property to use native SVG support (on Linux), only rendering at 100% is supported to exactly cover the case is supported right now (and other resolutions are done by auto-scaling), unless someone implements anything additionally?

As far as I know the native implementation is used for performance reasons. And since only SVG is (currently) able to natively scale the image dynamically I think we should continue using the native implementation for all other file formats as is. And only for SVGs we should always render in software, scaling 'natively'.

HannesWell:
After a session with Heiko last week I have now updated this PR to further prepare implementation of dynamically sized image file-formats like SVGs in https://github.com/eclipse-platform/eclipse.platform.swt/pull/1638 without the need to add any new API for that.

The basic idea is to apply the concept of the `ImageFileNameProvider` in a second step: After asking a `ImageFileNameProvider` for a file for a specific device-zoom, we ask the `FileFormat` to render the input for that `device-zoom` even if the file is for a lower zoom. Of course File-Formats are free to ignore that request and just render the image as usual, which also happens for all currently supported format. But the format for SVG can then later 'natively' scale up the image even if the file is for a zoom-factor of 100 (or any other as there is not really one zoom for an SVG).

Besides the `Image(Device device, ImageFileNameProvider imageFileNameProvider` constructor we have already implemented in that session I have now also implemented the same strategy for the other constructors that load from a 'file' source:
- Image(Device device, String filename)
- Image(Device device, InputStream stream)

For Windows this looks already good to me, but for Mac and Linux I think the implementation is not complete or needs some clarification.

HeikoKlare:
Info: with the latest push, I have rebased the PR on top of master, in particular including an update of the ImageLoader implementation as that has been refactored into ImageLoader and InternalImageLoader on master.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
HeikoKlare: Unfortunately, MacOS has no on-demand loading for image data at different zooms implemented yet. Thus, whenever possible, two representations of each image are created upon image instantiation. We need to implement this here as well, which is why the changes are generally fine.
I still have enhanced the changes with support for the constructors taking `InputStream` and `String` to also render multiple sizes of SVGs in case they refer to an SVG. In addition, I have revised the constructor accepting an `ImageFileNameProvider` to not unnecessary try to rescale the 200% image but just add a representation for it as is, like done in other constructors as well.

HeikoKlare:
Info: with the recent pushed I haved rebased the change onto preparation work in #1886 and restructured the commits to the following two:
1. Adapt FileFormat implementation to accept file and target zooms and be prepared for dynamic format types (SVGs), adapt InternalImageLoader accordingly and make ImageLoader fit to the new internal API
2. Adapt ImageLoader/ImageDataLoader to accept file and target zoom and make use of the enhanced API in the Image implementations.

The former would completely works without the latter.

HeikoKlare:
I have extracted the second commit out of this PR to keep this preparatory work separated from the adoption in the OS-specific `Image` classes and merge them one after another. Follow-up PR is:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/1889

PR #1829, opened 2025-02-14
possibility to suppress GTK interactive search (popup)

chrstg:


iloveeclipse (changes requested):
Please provide meaningful commit message.

See 
* https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md#what-does-a-valid-pr-look-like-checklist
* https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md#commit_message_recommendations

After that, please amend your commit and force push to same branch.

Please note, that we are at 4.35 RC phase so no API changes are allowed for this release.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/SWT.java
```diff
@@ -1564,6 +1564,19 @@ public class SWT {
...
+    /**
+     * Style constant for disabling GTK interactive search popup (value is 1&lt;&lt;10).
+     * <p><b>Used By:</b></p>
+     * <ul>
+     * <li><code>List</code></li>
+     * <li><code>Table</code></li>
+     * <li><code>Tree</code></li>
+     * </ul>
+     *
+     * @since 3.129
```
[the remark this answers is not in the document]
chrstg: thanks for the help!

chrstg:
> Please note, that we are at 4.35 RC phase so no API changes are allowed for this release.

So we'll have to wait for the next release.
Anything to do for me until then?

iloveeclipse:
> Anything to do for me until then?

Yes. Currently you have 3 commits including one merge commit.
Merge commits are not allowed, and for such a small change there is no need for multiple commits either.

Please squash your changes to **one commit**, rebase it on top of master branch and force push to same branch on github.

iloveeclipse:
> isn't it good practive to separate functional commit and version increase commit?

Correct, however I can't distinguish your commits in github UI because they have exact same commit message.
So updated instructions were: get rid of merge commit, provide proper commit messages, rebase on master & force push.

chrstg:
> > isn't it good practive to separate functional commit and version increase commit?
> 
> Correct, however I can't distinguish your commits in github UI because they have exact same commit message. So updated instructions were: get rid of merge commit, provide proper commit messages, rebase on master & force push.

force push is rejected with message
! [remote rejected]       master -> master (refusing to allow a Personal Access Token to create or update workflow `.github/workflows/maven.yml` without `workflow` scope)

Either way i didn't branch on my fork.
I guess it's better to start over from scratch with a new pull request?!

iloveeclipse:
> I guess it's better to start over from scratch with a new pull request?!

Yes. Id you work on master, you shouldn't do this in the future.
See also https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md#creating-a-pull-request

merks:
What I sometimes do is locally rename the branch I'm using for the PR, check out master again, pull, create a branch with the name previously used for the PR, commit the changes to that branch, and then force push that.  That updates the current PR with completely new clean content rather than close a PR and open a new PR...

chrstg:
the Jenkins build failes with 
04:51:13  [ERROR] [API ERROR] File unknown at line 0: The field org.eclipse.swt.SWT.NO_SEARCH has been added to a class (location: unknown)
04:51:13  [ERROR] [API ERROR] File MANIFEST.MF at line 1: The major version should be incremented in version 3.130.0, since API breakage occurred since version 3.128.0 (location: /home/jenkins/agent/workspace/eclipse.platform.swt_PR-1829/eclipse.platform.swt/binaries/org.eclipse.swt.gtk.linux.ppc64le/META-INF/MANIFEST.MF)

Did i increase the versions in a wrong way?

merks:
The API tools are very conservative about what is a breaking change.  Adding a public static field to a class is considered API-breaking and hence it wants a 4.0.0.  But that's not really acceptable  so the error needs to be suppressed.  Normally that's done via the a quick-fix in the IDE; was the error not already evident in the IDE?

chrstg:
> The API tools are very conservative about what is a breaking change. Adding a public static field to a class is considered API-breaking and hence it wants a 4.0.0. But that's not really acceptable so the error needs to be suppressed. Normally that's done via the a quick-fix in the IDE; was the error not already evident in the IDE?

In first place i didn't set up API tools / baseline correctly. Now it showed up and offered the quick fixes.

chrstg:
btw, the binary fragments require host bundle-version="[3.128.."  . couldn't it be increased to [3.129 ?

merks:
Yes, I think these much match.  Note that you will need to redo all this at the start of the next release cycle, so I'm not sure you want to bang your head on this now...

merks:
> In first place i didn't set up API tools / baseline correctly. Now it showed up and offered the quick fixes.

It's probably better to use the setup which does all the necessary and desired things:

https://github.com/eclipse-platform/eclipse.platform.swt?tab=readme-ov-file#contributing-to-swt

E.g., it will update the target platform and the API baseline when the time comes for that...

chrstg:
> Yes, I think these much match. Note that you will need to redo all this at the start of the next release cycle, so I'm not sure you want to bang your head on this now...

i was afraid of that :D  but next time it'll go in a minute

chrstg:
we're fine with next release.
Just for curiosity: Since it's no API change anymore could it be integrated in current release?

merks:
Yes, next time you'll know what to do.   😁 

If the builds are green, then in principle yes it could be merged.   Of course they are API changes just not a major version API change.  I approved a build run just now...

laeubi:
Technically it is an API change, but honestly nothing affecting anyone or of any risk... it adds a new field to a class that is just a collection of constants (and should actually be marked with `@NoExtends`), and the feature it self is disabled by default.

sratz:
Isn't the problem rather that these constant are only introduced for the `gtk` platform, but all SWT API is the same across all platforms? So the constant would also have to be added to `win32` and `macos`.

mickaelistria:
> Isn't the problem rather that these constant are only introduced for the gtk platform, but all SWT API is the same across all platforms? So the constant would also have to be added to win32 and macos.

Yes, it's also a concern for me. How much do we want such platform-specific settings to leak into API. Wouldn't a system property or a `control.getData("gtk.skipSearch")` be a better choice ?
@chrstg do you want to disable search for only 1 particular viewer, or for all of them?

sratz:
At least the Javadoc needs to be consistent on all platforms and say that it's only a hint for GTK.

chrstg:
> At least the Javadoc needs to be consistent on all platforms and say that it's only a hint for GTK.

I agree. As far as I know this interactive search is only on GTK but I don't know others than Windows and Linux.

chrstg:
> > Isn't the problem rather that these constant are only introduced for the gtk platform, but all SWT API is the same across all platforms? So the constant would also have to be added to win32 and macos.
> 
> Yes, it's also a concern for me. How much do we want such platform-specific settings to leak into API. Wouldn't a system property or a `control.getData("gtk.skipSearch")` be a better choice ? @chrstg do you want to disable search for only 1 particular viewer, or for all of them?

I thought about that solution either. At the moment we want to globally suppress it. But it's only our project and for the moment. The flag seems more flexible generally.

chrstg:
> Isn't the problem rather that these constant are only introduced for the `gtk` platform, but all SWT API is the same across all platforms? So the constant would also have to be added to `win32` and `macos`.
> 
> Edit: Nevermind, I was looking at this wrong.



> Technically it is an API change, but honestly nothing affecting anyone or of any risk... it adds a new field to a class that is just a collection of constants (and should actually be marked with `@NoExtends`), and the feature it self is disabled by default.

It's a little bit more complicated with these SWT flags since there values overlap (same value used by several flags). 
Therefore one has to be careful with these values. I picked the 1<<10 (1024) because the Widgets that use it (Table, Tree, List) don't use any flag with this value 1<<10.

So I guess it's good to still mark it as API change since new flags could potentially break things.

chrstg:
hmm. it closed itself while rebasing?!
..shall I reopen and continue on this pull-request or start over again from scratch?

chrstg:
Hi @iloveeclipse 
can we get this one through the gate.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/List.java
```diff
@@ -258,6 +258,9 @@ void createHandle (int index) {
 	if ((getShell ().style & SWT.ON_TOP) != 0) {
 		GTK.gtk_tree_view_set_search_column (handle, -1);
 	}
+    if ((style & SWT.NO_SEARCH) != 0) {
```
iloveeclipse, line 261: Can we combine this "if" with the previous one, in a style consistent with the Tree / Table code below and extract the condition into `searchEnabled()` method?

iloveeclipse:
> can we get this one through the gate.

See my comment in code above, beside this it looks good & ready to merge. Ich you change the code, please first rebase on master, then amend commit & force push.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/List.java
chrstg: sure, did it.
(even when imho i think that the implementation of the no_search() method is ugly, too many negations)

iloveeclipse:
> i assume it's not from my changes ?!

For sure not. It's Friday, both hard- and software is already in the weekend mode :-)
I've restarted the build.

chrst4:
> > i assume it's not from my changes ?!
> 
> For sure not. It's Friday, both hard- and software is already in the weekend mode :-) I've restarted the build.

something seems broken on the linux-risc build. All pull requests seem to fail there.

iloveeclipse:
> something seems broken on the linux-risc build. All pull requests seem to fail there.

I've just triggered https://github.com/eclipse-platform/eclipse.platform.swt/pull/1931 to have a proof.

iloveeclipse:
> I've just triggered #1931 to have a proof.

At least that one didn't wanted to rebuild binaries and failed for a different reason, but it was on top of master (yours is not). Let rebase this PR.

iloveeclipse:
>  Let rebase this PR.

... and finally we have a green state!

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/List.java
```diff
@@ -1299,6 +1292,20 @@ public void removeSelectionListener(SelectionListener listener) {
 	eventTable.unhook (SWT.DefaultSelection,listener);
 }
 
+private boolean search_enabled() {
```
iloveeclipse, line 1295: `search_enabled` should be better named `searchEnabled` to be consistent with existing code in Tree/Table.

iloveeclipse (approved):
I will handle `search_enabled` vs `searchEnabled` inconsistent naming in a follow up PR.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/List.java
iloveeclipse: => see https://github.com/eclipse-platform/eclipse.platform.swt/pull/1934

chrstg:
thanks a lot. Great!

PR #1831, opened 2025-02-14
Improve Program Icon Scaling

amartya4256:
This PR contributes to the improvement in the scaling of program icons. The implementation uses a windows API SHGetImageList to obtain imagelists at small, large and extra large sizes. Then the icon nearest to the required size is retrieved and its image data is scaled up returned.

Typically the small, large and extra large icons are supposed to have 16, 32 and 48 px of sizes respectively. But apparently, the API SHGetImageList turns out to be System-DPI aware. It means if the primary monitor zoom at app startup was 100% then those scaled values are valid but for instance if it was 200% then the sizes from small, large and extra large would be 32, 64 and 128 px. Hence, with this approach we do get a room to scale up with better image quality since, the retrieved images are usually, 1x, 2x and 3x the size of the primary monitor zoom at startup. However, if there exists a monitor with significantly lower zoom than the primary monitor zoom at startup, then the 1x zoom (small) icon will be used, which will end up in compressed images.

This approach still opens headspace in one direction (scale up), hence, the quality is indeed better while scaling up. However, while scaling down below the primary monitor zoom at app startup, it will compress the icon.

contributes to #62 and #127

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT PI/win32/org/eclipse/swt/internal/win32/OS.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT PI/win32/org/eclipse/swt/internal/win32/OS.java
@@ -2302,6 +2302,11 @@ public static final boolean SetWindowText (long hWnd, TCHAR lpString) {
 	return SetWindowText (hWnd, lpString1);
 }
 
+public static final int SHDefExtractIcon (TCHAR lpszFile, int iIndex, int uFlags, long [] phiconLarge, long [] phiconSmall, int nIconSize) {
+	char [] lpszFile1 = lpszFile == null ? null : lpszFile.chars;
+	return SHDefExtractIcon (lpszFile1, iIndex, uFlags, phiconLarge, phiconSmall, nIconSize);
+}
+
 public static final boolean UnregisterClass (TCHAR lpClassName, long hInstance) {
 	char [] lpClassName1 = lpClassName == null ? null : lpClassName.chars;
 	return UnregisterClass (lpClassName1, hInstance);
@@ -4455,6 +4460,12 @@ public static final native long SetWindowsHookEx (int idHook, long lpfn,  long h
  * @param lpXform cast=(XFORM *),flags=no_out
  */
 public static final native boolean SetWorldTransform(long hdc, float[] lpXform);
+/**
+ * @param lpszFile cast=(LPWSTR)
+ * @param phiconLarge cast=(HICON FAR *)
+ * @param phiconSmall cast=(HICON FAR *)
+ */
+public static final native int SHDefExtractIcon (char [] lpszFile, int iIndex, int uFlags, long [] phiconLarge, long [] phiconSmall, int nIconSize);
 /** @param pszPath cast=(LPCWSTR),flags=no_out */
 public static final native long SHGetFileInfo (char [] pszPath, int dwFileAttributes, SHFILEINFO psfi, int cbFileInfo, int uFlags);
 public static final native boolean ShellExecuteEx (SHELLEXECUTEINFO lpExecInfo);
--- a/bundles/org.eclipse.swt/Eclipse SWT Program/win32/org/eclipse/swt/program/Program.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Program/win32/org/eclipse/swt/program/Program.java
@@ -379,10 +379,45 @@ public ImageData getImageData () {
  * @since 3.125
  */
 public ImageData getImageData (int zoom) {
-	// OS.SHGetFileInfo is System DPI-aware, hence it retrieves the icon with zoom
-	// of primary monitor at the application startup
-	int initialNativeZoom = getPrimaryMonitorZoomAtStartup();
+	ImageData imageData = getImageDataUsingIconFilePath(zoom);
+	return (imageData != null) ? imageData : getImageDataUsingExtension(zoom);
+}
+
+private ImageData getImageDataUsingIconFilePath(int zoom) {
+	int nIconIndex = 0;
+	String fileName = iconName;
+	int index = iconName.indexOf (',');
+	if (index != -1) {
+		fileName = iconName.substring (0, index);
+		String iconIndex = iconName.substring (index + 1, iconName.length ()).trim ();
+		try {
+			nIconIndex = Integer.parseInt (iconIndex);
+		} catch (NumberFormatException e) {}
+	}
+	int length = fileName.length ();
+	if (length > 1 && fileName.charAt (0) == '\"') {
+		if (fileName.charAt (length - 1) == '\"') {
+			fileName = fileName.substring (1, length - 1);
+		}
+	}
+	TCHAR lpszFile = new TCHAR (0, fileName, true);
+	long [] hIcon = new long[1];
+	int size = OS.GetSystemMetricsForDpi(OS.SM_CXSMICON, DPIUtil.mapZoomToDPI(zoom));
+	OS.SHDefExtractIcon (lpszFile, nIconIndex, 0, hIcon, null, size);
+	if (hIcon [0] == 0) {
+		return null;
+	}
+	Image image = Image.win32_new (null, SWT.ICON, hIcon [0], zoom);
+	ImageData imageData = image.getImageData (zoom);
+	image.dispose ();
+	return imageData;
+}
+
+private ImageData getImageDataUsingExtension(int zoom) {
 	if (extension != null) {
+		// OS.SHGetFileInfo is System DPI-aware, hence it retrieves the icon with zoom
+		// of primary monitor at the application startup
+		int initialNativeZoom = getPrimaryMonitorZoomAtStartup();
 		SHFILEINFO shfi = new SHFILEINFO ();
 		int flags = OS.SHGFI_ICON | OS.SHGFI_USEFILEATTRIBUTES;
 		boolean useLargeIcon = 100 * zoom /  initialNativeZoom >= 200;
@@ -401,30 +436,7 @@ public ImageData getImageData (int zoom) {
 			return imageData;
 		}
 	}
-	int nIconIndex = 0;
-	String fileName = iconName;
-	int index = iconName.indexOf (',');
-	if (index != -1) {
-		fileName = iconName.substring (0, index);
-		String iconIndex = iconName.substring (index + 1, iconName.length ()).trim ();
-		try {
-			nIconIndex = Integer.parseInt (iconIndex);
-		} catch (NumberFormatException e) {}
-	}
-	int length = fileName.length ();
-	if (length > 1 && fileName.charAt (0) == '\"') {
-		if (fileName.charAt (length - 1) == '\"') {
-			fileName = fileName.substring (1, length - 1);
-		}
-	}
-	TCHAR lpszFile = new TCHAR (0, fileName, true);
-	long [] phiconSmall = new long[1], phiconLarge = null;
-	OS.ExtractIconEx (lpszFile, nIconIndex, phiconLarge, phiconSmall, 1);
-	if (phiconSmall [0] == 0) return null;
-	Image image = Image.win32_new (null, SWT.ICON, phiconSmall [0], initialNativeZoom);
-	ImageData imageData = image.getImageData (zoom);
-	image.dispose ();
-	return imageData;
+	return null;
 }
 
 private int getPrimaryMonitorZoomAtStartup() {
--- a/bundles/org.eclipse.swt/Eclipse SWT PI/win32/library/os.c
+++ b/bundles/org.eclipse.swt/Eclipse SWT PI/win32/library/os.c
@@ -7534,6 +7534,28 @@ JNIEXPORT jint JNICALL OS_NATIVE(SHDRAGIMAGE_1sizeof)
 }
 #endif
 
+#ifndef NO_SHDefExtractIcon
+JNIEXPORT jint JNICALL OS_NATIVE(SHDefExtractIcon)
+	(JNIEnv *env, jclass that, jcharArray arg0, jint arg1, jint arg2, jlongArray arg3, jlongArray arg4, jint arg5)
+{
+	jchar *lparg0=NULL;
+	jlong *lparg3=NULL;
+	jlong *lparg4=NULL;
+	jint rc = 0;
+	OS_NATIVE_ENTER(env, that, SHDefExtractIcon_FUNC);
+	if (arg0) if ((lparg0 = (*env)->GetCharArrayElements(env, arg0, NULL)) == NULL) goto fail;
+	if (arg3) if ((lparg3 = (*env)->GetLongArrayElements(env, arg3, NULL)) == NULL) goto fail;
+	if (arg4) if ((lparg4 = (*env)->GetLongArrayElements(env, arg4, NULL)) == NULL) goto fail;
+	rc = (jint)SHDefExtractIcon((LPWSTR)lparg0, arg1, arg2, (HICON FAR *)lparg3, (HICON FAR *)lparg4, arg5);
+fail:
+	if (arg4 && lparg4) (*env)->ReleaseLongArrayElements(env, arg4, lparg4, 0);
+	if (arg3 && lparg3) (*env)->ReleaseLongArrayElements(env, arg3, lparg3, 0);
+	if (arg0 && lparg0) (*env)->ReleaseCharArrayElements(env, arg0, lparg0, 0);
+	OS_NATIVE_EXIT(env, that, SHDefExtractIcon_FUNC);
+	return rc;
+}
+#endif
+
 #ifndef NO_SHELLEXECUTEINFO_1sizeof
 JNIEXPORT jint JNICALL OS_NATIVE(SHELLEXECUTEINFO_1sizeof)
 	(JNIEnv *env, jclass that)
--- a/bundles/org.eclipse.swt/Eclipse SWT PI/win32/library/os_stats.h
+++ b/bundles/org.eclipse.swt/Eclipse SWT PI/win32/library/os_stats.h
@@ -569,6 +569,7 @@ typedef enum {
 	SCROLLBARINFO_1sizeof_FUNC,
 	SCROLLINFO_1sizeof_FUNC,
 	SHDRAGIMAGE_1sizeof_FUNC,
+	SHDefExtractIcon_FUNC,
 	SHELLEXECUTEINFO_1sizeof_FUNC,
 	SHFILEINFO_1sizeof_FUNC,
 	SHGetFileInfo_FUNC,
--- a/bundles/org.eclipse.swt/Eclipse SWT PI/win32/library/os_structs.c
+++ b/bundles/org.eclipse.swt/Eclipse SWT PI/win32/library/os_structs.c
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2022 IBM Corporation and others.
+ * Copyright (c) 2000, 2025 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
--- a/bundles/org.eclipse.swt/Eclipse SWT PI/win32/library/os_structs.h
+++ b/bundles/org.eclipse.swt/Eclipse SWT PI/win32/library/os_structs.h
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2021 IBM Corporation and others.
+ * Copyright (c) 2000, 2025 IBM Corporation and others.
  *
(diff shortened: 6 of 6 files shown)


bundles/org.eclipse.swt/Eclipse SWT Program/win32/org/eclipse/swt/program/Program.java
```diff
@@ -434,6 +430,60 @@ private int getPrimaryMonitorZoomAtStartup() {
...
+	return Image.win32_new(Display.getCurrent(), SWT.ICON, hIcon, closestZoomAvailable);
+}
+
+private long getImageListForSize(int size) {
+	long [] ppv = new long [1];
+	COM.SHGetImageList(size, COM.IID_IImageList, ppv);
+	return ppv[0];
+}
+
+private Integer getClosestKey(NavigableMap<Integer, Long> map, int x) {
```
HeikoKlare, line 479:
```suggestion
private static int getClosestKey(NavigableMap<Integer, Long> map, int x) {
```

bundles/org.eclipse.swt/Eclipse SWT Program/win32/org/eclipse/swt/program/Program.java
```diff
@@ -434,6 +430,60 @@ private int getPrimaryMonitorZoomAtStartup() {
...
+    Integer floorKey = map.floorKey(x);
+    Integer ceilingKey = map.ceilingKey(x);
+    if (floorKey == null) return ceilingKey;
+    if (ceilingKey == null) return floorKey;
+    return (Math.abs(x - floorKey) <= Math.abs(x - ceilingKey)) ? floorKey : ceilingKey;
```
HeikoKlare, line 484: Shouldn't we rather take the icon with the next-higher than the required resolution and scale that down instead of taking the nearest one and scaling it up or down?

bundles/org.eclipse.swt/Eclipse SWT Program/win32/org/eclipse/swt/program/Program.java
```diff
@@ -434,6 +430,60 @@ private int getPrimaryMonitorZoomAtStartup() {
 	return DPIUtil.mapDPIToZoom(dpi);
 }
 
+private NavigableMap<Integer, Long> getImageListinAllAvailableSizesMappedWithZoom() {
+	TreeMap<Integer, Long> zoomToHImageList = new TreeMap<>();
+	int primaryMonitorZoomAtStartup = getPrimaryMonitorZoomAtStartup();
+	// SHIL_SYSSMALL is the size of icon at startup zoom of the primary monitor. The
+	// sizes can differ as specified by OS.GetSystemMetrics.
+	long hImageListAtPrimaryMonitorZoomAtStartup = getImageListForSize(OS.SHIL_SYSSMALL);
+	int sizeAtPrimaryMonitorZoomAtStartup = getIconSizeOfImageList(hImageListAtPrimaryMonitorZoomAtStartup);
```
HeikoKlare, line 439: This means the smallest retrievable icon still depends on the primary monitor zoom at startup. I.e., if I start an application with primary monitor zoom 200%, I will not be able to retrieve appropriate icons for, e.g., 100%. Isn't there any Windows API that allows us to simply retrieve icons in either "generic" scaling without any alignment to the primary monitor zoom (such as 16x16, 32x32, 64x64) or to request an icon at a specific zoom?

HeikoKlare (commented):
Before taking a deeper look into the changes, I have some questions regarding the overall change and in particular the used API, as the current proposal only seems to partially resolve the underlying issue and I am wondering if we cannot find a solution that completely resolves the issue with scaled program icons.

bundles/org.eclipse.swt/Eclipse SWT Program/win32/org/eclipse/swt/program/Program.java
amartya4256: Sounds good. I was just wondering in case we have images at 200, 400 and 600% for example and we want to get one for 225%, if we use 400, it would be a bit of an overkill compression on the image. Don't you think?

bundles/org.eclipse.swt/Eclipse SWT Program/win32/org/eclipse/swt/program/Program.java
amartya4256: At first I also thought that this API behaves the same way but appraently after implementing I found that the sizes are dependent on the System DPI value. I couldn't find any other alternative so far.

bundles/org.eclipse.swt/Eclipse SWT Program/win32/org/eclipse/swt/program/Program.java
HeikoKlare: It's probably best to try out what gives the best results. It also depends on which scaling mode you use, as smooth scaling will usually produce more acceptable results (also when downscaling) compared to nearest scaling.

I once did a small comparison of different scalings applied on program icons: https://github.com/eclipse-platform/eclipse.platform.swt/pull/1014#issuecomment-1926909752

And we also have quite a few results documented in this PR: https://github.com/eclipse-platform/eclipse.platform.swt/pull/1138

bundles/org.eclipse.swt/Eclipse SWT Program/win32/org/eclipse/swt/program/Program.java
HeikoKlare: What about something like `SHDefExtractIconA`? That one seems to be capable of extracting an icon in a specified size.
https://learn.microsoft.com/en-us/windows/win32/api/shlobj_core/nf-shlobj_core-shdefextracticona

amartya4256: Apparently, SHDefExtractIconA only works if the file is available. It doesn't work for a file path which is not available. SHGetFileInfo however doesn't need the file to exist as it doesn't take the file as input but rather it's path and parses the icon which is applicable for the file type. This can be controled by setting the flag to SHGFI_USEFILEATTRIBUTES. It then returns the icon index in system image list. And then we use the new method SHGetImageList to btain the image lists at different zoom level and retrieve the icon at the index with the right size.

SHDefExtractIconA  doesn't have this flexibility since it needs the file to exist to extract the icon using the file.

Here's more about what kind of input we are providing to SHGetFileInfo: https://learn.microsoft.com/en-us/windows/win32/api/shellapi/nf-shellapi-shgetfileinfoa

bundles/org.eclipse.swt/Eclipse SWT Program/win32/org/eclipse/swt/program/Program.java
amartya4256: Done

bundles/org.eclipse.swt/Eclipse SWT Program/win32/org/eclipse/swt/program/Program.java
```diff
@@ -381,26 +383,20 @@ public ImageData getImageData () {
...
+		if (shfi.iIcon >= 0) {
+			Image icon = getImageForZoom(zoom, shfi.iIcon);
+			ImageData imageData = icon.getImageData(zoom);
+			icon.dispose();
 			return imageData;
 		}
 	}
+
+	int initialNativeZoom = getPrimaryMonitorZoomAtStartup();
 	int nIconIndex = 0;
```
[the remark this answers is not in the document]
amartya4256: Yes I'll adapt this

bundles/org.eclipse.swt/Eclipse SWT Program/win32/org/eclipse/swt/program/Program.java
amartya4256: Okay I tested out both the approach. Since CompositeImageDescriptor::getImageData which is responsible for calling the Program::getImageData and it seems to not work well with quarter scaling because of an existing bug in supportsZoomLevel method (org.eclipse.jface.resource.CompositeImageDescriptor.supportsZoomLevel(int)), it seems to be always using using the integer200 soon (see Bug 97506: [HiDPI] ImageData.scaledTo() should use a better interpolation method), it seems to be working fine with always using the ceiling key and scaling it down. So, I would use ceiling key.

bundles/org.eclipse.swt/Eclipse SWT Program/win32/org/eclipse/swt/program/Program.java
amartya4256: Apparently, SHDefExtractIconA only works if the file is available. It doesn't work for a file path which is not available. SHGetFileInfo however doesn't need the file to exist as it doesn't take the file as input but rather it's path and parses the icon which is applicable for the file type. This can be controled by setting the flag to SHGFI_USEFILEATTRIBUTES. It then returns the icon index in system image list. And then we use the new method SHGetImageList to btain the image lists at different zoom level and retrieve the icon at the index with the right size.

SHDefExtractIconA  doesn't have this flexibility since it needs the file to exist to extract the icon using the file and after implementing it, I can see some bad behaviour, i.e. image not being retrieved sometimes when the file is not available.

Here's more about what kind of input we are providing to SHGetFileInfo: https://learn.microsoft.com/en-us/windows/win32/api/shellapi/nf-shellapi-shgetfileinfoa

bundles/org.eclipse.swt/Eclipse SWT Program/win32/org/eclipse/swt/program/Program.java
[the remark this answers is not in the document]
amartya4256: SHDefExtractIcon works better here. It acceots the same parameters as ExtractIconEx but also accepts the size of the icon. The only thing it needs is for the pszFile to be present there. But that part of the implementation is also applicable for DrawIconEx. In other words, we will be just adding the ability to retrieve the file of our choice of size here.

bundles/org.eclipse.swt/Eclipse SWT Program/win32/org/eclipse/swt/program/Program.java
amartya4256: Implemented using the suggested API and the previous implementation as a fallback if the pszFile is not valid (extracts using the extension in that case, if extension also not present, uses the default icon - also complies with previous behaviour)

bundles/org.eclipse.swt/Eclipse SWT Program/win32/org/eclipse/swt/program/Program.java
```diff
@@ -418,15 +396,42 @@ public ImageData getImageData (int zoom) {
...
+		return getImageDataUsingExtension(zoom);
+	}
+	Image image = Image.win32_new (null, SWT.ICON, hIcon [0], zoom);
 	ImageData imageData = image.getImageData (zoom);
 	image.dispose ();
 	return imageData;
 }
 
+private ImageData getImageDataUsingExtension(int zoom) {
+	if (extension != null) {
```
fedejeanne, line 412: NIT: invert the logic in order to reduce the complexity:
```java
if (extension == null) {
     return null;
}

...
```

HeikoKlare: Even better: perform this check in the caller to avoid returning `null` and requiring this if blokc in general.

bundles/org.eclipse.swt/Eclipse SWT Program/win32/org/eclipse/swt/program/Program.java
```diff
@@ -418,15 +396,42 @@ public ImageData getImageData (int zoom) {
...
+		int initialNativeZoom = getPrimaryMonitorZoomAtStartup();
```
HeikoKlare, line 413: Please keep the existing comment for why we use primary monitor zoom at startup. Otherwise it will be even harder to comprehend for other / in the future.

bundles/org.eclipse.swt/Eclipse SWT Program/win32/org/eclipse/swt/program/Program.java
```diff
@@ -418,15 +396,42 @@ public ImageData getImageData (int zoom) {
...
 	TCHAR lpszFile = new TCHAR (0, fileName, true);
-	long [] phiconSmall = new long[1], phiconLarge = null;
-	OS.ExtractIconEx (lpszFile, nIconIndex, phiconLarge, phiconSmall, 1);
-	if (phiconSmall [0] == 0) return null;
-	Image image = Image.win32_new (null, SWT.ICON, phiconSmall [0], initialNativeZoom);
+	long [] hIcon = new long[1];
+	int size = OS.GetSystemMetricsForDpi(OS.SM_CXSMICON, DPIUtil.mapZoomToDPI(zoom));
+	OS.SHDefExtractIcon (lpszFile, nIconIndex, 0, hIcon, null, size);
+	if (hIcon [0] == 0) {
+		return getImageDataUsingExtension(zoom);
```
HeikoKlare, line 403: Since you clean the code by extracting the from-extension-case into a separate method, what about doint the same for the from-icon-path-case? I.e., extract everything until `OS.SHDefExtractIcon` into an according method as well. Then the code becomes more easy to read as you can easily distinguish the two cases and when each of them applies without looking into the implementation of each of them.

bundles/org.eclipse.swt/Eclipse SWT PI/win32/org/eclipse/swt/internal/win32/OS.java
```diff
@@ -4445,6 +4445,16 @@ public static int HRESULT_FROM_WIN32(int x) {
...
+/**
+ * @param lpszFile cast=(LPWSTR)
+ * @param phiconLarge cast=(HICON FAR *)
+ * @param phiconSmall cast=(HICON FAR *)
+ */
+public static final native int SHDefExtractIcon (char [] lpszFile, int iIndex, int uFlags, long [] phiconLarge, long [] phiconSmall, int nIconSize);
+public static final int SHDefExtractIcon (TCHAR lpszFile, int iIndex, int uFlags, long [] phiconLarge, long [] phiconSmall, int nIconSize) {
+	char [] lpszFile1 = lpszFile == null ? null : lpszFile.chars;
+	return SHDefExtractIcon (lpszFile1, iIndex, uFlags, phiconLarge, phiconSmall, nIconSize);
+}
```
HeikoKlare, line 4457: Such utility methods are uncommon for the `OS` class. Why do we need this and cannot just pass `lpszFile.chars` directly like we do for `SHGetFileInfo`?

HeikoKlare (commented):
I've tested the code and it works fine for me. Once we get rid of the zoom limitations in CompositeImageDescriptor, we will even see properly scaled program icons in the package explorer.

I was just wondering whether preferring loading the icon from the icon path instead of the extension may have any drawback or can produce regression (like the ones we faced after completely removing that case in https://github.com/eclipse-platform/eclipse.platform.swt/pull/999), but I was not able to produce any problematic scenario so far.

So with the few requested changes, this is fine from my side and can be merged.

bundles/org.eclipse.swt/Eclipse SWT Program/win32/org/eclipse/swt/program/Program.java
fedejeanne: I lean towards my solution since it means less responsibility for the caller (easier to use) and since the name of the extracted method says `...UsingExtension`, which makes it OK to consider the extension inside of it.

But it's a matter of taste so it's your call @amartya4256 . Please consider adding a JavaDoc/comment to the method so the user knows what to expect.

HeikoKlare: Please prefer clean code practice over Javadoc for private methods that easily becomes outdated.

fedejeanne: you lost me. How is my proposal _not_ clean?

HeikoKlare: I didn't say that. I was just referring to the proposal of adding Javadoc to a private method, which is usually something you can avoid via clean design. But I am not sure what the Javadoc would be supposed to express here.

Since the method is almost independent from the state of the class instance anyway, one might even make a static method `getImageDataForExtension(String extension, int zoom)` out of it and assume the passed extension to be non-null.

fedejeanne: > I didn't say that. I was just referring to the proposal of adding Javadoc to a private method

👍 

> Since the method is almost independent from the state of the class instance anyway, one might even make a static method `getImageDataForExtension(String extension, int zoom)` out of it and assume the passed extension to be non-null.

I'm against assuming the passed extension to be non-null because that would mean that the caller needs to know this, which is an internal detail of `getImageDataForExtension` / `getImageDataUsingExtension`. More specifically, the caller needs to know what happens when calling `TCHAR pszPath = new TCHAR (0, extension, true);` with `extension == null`. The way it is written now (or using the early `return` I propose) are 2 ways of keeping the code self-explanatory without having to read the whole method. Leaving the `null`-check outside the method would also mean that in case there are more callers, they need to handle the `null` case too (duplicating code).

HeikoKlare: > I'm against assuming the passed extension to be non-null because that would mean that the caller needs to know this,

That's an invalid argument because that would require every method to check every argument for not being null. But it's common design practice to prefer non-null arguments over nullables ones for good reasons. And in this case, the caller does not need to know anything about the implementation, as the method can return null anyway. And that passing null as extension returns null is nothing unexpected, in particular when you call a method that states that it loads something from an extension and you pass null as the extension. Currently, it's actually exactly the opposite: when calling the method you need to know about the implementation that when the instance variable for the extension is null, the method will return null as well. 

> Leaving the null-check outside the method would also mean that in case there are more callers, they need to handle the null case too (duplicating code).

That's also not valid, as this is a private method which you can simply change if new callers are added. Assuming that beforehand would be an example of premature optimization.

Still, this is really not a crucial point but just a minor detail in the internal micro design of a class. I just wanted to make proposals to improve the design and made my point already, so I do not want to spend more time into this discussion. Feel free to use whatever design you prefer here as long as it's functionally correct.

bundles/org.eclipse.swt/Eclipse SWT PI/win32/org/eclipse/swt/internal/win32/OS.java
amartya4256: This method is very similar to OS::SHDefExtractIcon so I tried to stick to the way that had been implemented. However, in our case, at this one place where we call this method, we never have lpszFile set to null so I can get rid of it.

HeikoKlare: Thanks for the clarification. I wasn't aware of those methods. They are placed in a separate "section" of the OS class under `Ansi/Unicode wrappers`, so it would be fine to keep your implementation but it should then be moved to that section of the class.

bundles/org.eclipse.swt/Eclipse SWT Program/win32/org/eclipse/swt/program/Program.java
amartya4256: Yeah nice idea

bundles/org.eclipse.swt/Eclipse SWT Program/win32/org/eclipse/swt/program/Program.java
amartya4256: inside that if block we have one more if check for if we could retrieve the icon:

```
               if (shfi.hIcon != 0) {
			Image image = Image.win32_new (null, SWT.ICON, shfi.hIcon, initialNativeZoom);
			ImageData imageData = image.getImageData (zoom);
			image.dispose ();
			return imageData;
		}
```

So even if I make a check at the beginning of the method for extension == null, or outside the method, i will still have to return a null explicitly in the end if this clause is not satisfied. To me it seems natural to just keep it the way it is. Can we agree to keep it just like this?

bundles/org.eclipse.swt/Eclipse SWT PI/win32/org/eclipse/swt/internal/win32/OS.java
amartya4256: Alright. I will move it :)

HeikoKlare:
FTR: binaries successfully built and pushed with commit https://github.com/eclipse-platform/eclipse.platform.swt/commit/a92d9dbcf23706e13bed5b35a5b0add503cff08e

PR #1663, opened 2024-12-14
Add workflow to run with GTK4 enabled

laeubi:
Currently everything is run with GTK3 by default but we can't tell the current state of affairs with GTK4.

This now adds a new action that run the build (and tests) with GTK4 enabled.

PR #1832, opened 2025-02-14
[GTK4] Fix segmentation fault when creating multiple displays

ptziegler:
Subclasses are not created in GTK4, leading to the assignment of NULL pointers when disposing the display. If a second display is created afterwards, this leads to a segmentation fault.

Contributes to
https://github.com/eclipse-platform/eclipse.platform.swt/pull/1663

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
@@ -4817,26 +4817,28 @@ void releaseDisplay () {
 	keysChangedProc = 0;
 
 	/* Dispose subclass */
-	long pangoLayoutType = OS.PANGO_TYPE_LAYOUT ();
-	long pangoLayoutClass = OS.g_type_class_ref (pangoLayoutType);
-	OS.G_OBJECT_CLASS_SET_CONSTRUCTOR (pangoLayoutClass, pangoLayoutNewProc);
-	OS.g_type_class_unref (pangoLayoutClass);
-	pangoLayoutNewProc = 0;
-	long imContextType = GTK.GTK_TYPE_IM_MULTICONTEXT ();
-	long imContextClass = OS.g_type_class_ref (imContextType);
-	OS.G_OBJECT_CLASS_SET_CONSTRUCTOR (imContextClass, imContextNewProc);
-	OS.g_type_class_unref (imContextClass);
-	imContextNewProc = 0;
-	long pangoFontFamilyType = OS.PANGO_TYPE_FONT_FAMILY ();
-	long pangoFontFamilyClass = OS.g_type_class_ref (pangoFontFamilyType);
-	OS.G_OBJECT_CLASS_SET_CONSTRUCTOR (pangoFontFamilyClass, pangoFontFamilyNewProc);
-	OS.g_type_class_unref (pangoFontFamilyClass);
-	pangoFontFamilyNewProc = 0;
-	long pangoFontFaceType = OS.PANGO_TYPE_FONT_FACE ();
-	long pangoFontFaceClass = OS.g_type_class_ref (pangoFontFaceType);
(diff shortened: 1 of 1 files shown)


ptziegler:
See https://github.com/eclipse-platform/eclipse.platform.swt/blob/857ac0b2c4c7059e7957cd99e46a4501673cc67e/bundles/org.eclipse.swt/Eclipse%20SWT/gtk/org/eclipse/swt/widgets/Display.java#L3774-L3806

Two questions though:

1) On GTK3, the `printerOptionWidgetNewProc` is allocated but never freed, when disposed. I don't think this is deliberate, but rather a memory leak?

2) Would this still be fine for the RC1 or is it safer to postpone it until after the 2025-03 release?

iloveeclipse:
> * On GTK3, the `printerOptionWidgetNewProc` is allocated but never freed, when disposed. I don't think this is deliberate, but rather a resource leak?

Probably leak, but can wait for 4.36 as not new.

> * Would this still be fine for the RC1 or is it safer to postpone it until after the 2025-03 release?

This patch seem to be safe for GTK3, and only improves (anyway bad state of) GTK4, so would be OK for RC1. On the other side, there is no reason to rush for exact same reason, so I would propose to wait for 4.36 too. I know at least one company with patches for SWT Display class, and last minute changes are always disruptive.

ptziegler:
> This patch seem to be safe for GTK3, and only improves (anyway bad state of) GTK4, so would be OK for RC1. On the other side, there is no reason to rush for exact same reason, so I would propose to wait for 4.36 too. I know at least one company with patches for SWT Display class, and last minute changes are always disruptive.

That's a fair point. Then I'll wait for the start of the next release cycle.

laeubi:
> I know at least one company with patches for SWT Display class, and last minute changes are always disruptive.

Then this company should at best contribute the patches so it becomes less disruptive for them.

> This patch seem to be safe for GTK3, and only improves (anyway bad state of) GTK4, so would be OK for RC1. 

I agree and would vote for merging this ASAP, it is not an "improvement" if the process crash with a core dump but a server bug.

laeubi (approved):

laeubi:
Lets see if it already improves the situation, I added the gtk4 label so it should run the tests with gtk4 enabled.

laeubi:
The segmentation fault is gone, good job, now there are new error that popup but I think that's kind of expected.

ptziegler:
At least from my side, the main motivation was to fix the early crash of the test suite. So this is not something I need to have in the RC1. And if the only benefit or merging this is to disrupt someone elses workflow, then I'm also leaning towards postponing this for the next release.

> I agree and would vote for merging this ASAP, it is not an "improvement" if the process crash with a core dump but a server bug.

I don't think (or rather hope) there is anyone who uses the GTK4 productively. Especially since e.g. the UnsatisfiedLinkError when handling Expose events made the port effectively unusable, outside of the Snippets. So whether this is merged or not would hardly make a difference.

> The segmentation fault is gone, good job, now there are new error that popup but I think that's kind of expected.

That is true, though now it seems like the run gets stuck while executing the Browser tests. 😦 

image

laeubi:
>  And if the only benefit or merging this is to disrupt someone elses workflow

Every change breaks someones workflow :-P

grafik
https://xkcd.com/1172/

akurtakov:
@ptziegler What do you mean by "Subclasses are not created in GTK4" ?

ptziegler:
> @ptziegler What do you mean by "Subclasses are not created in GTK4" ?

This method is effectively skipped in GTK4. It's not the Java inheritance I'm talking about, but e.g. the Pango Layout subclasses that are set when the Display is created. Perhaps I can rephrase it to make it less confusing.

https://github.com/eclipse-platform/eclipse.platform.swt/blob/32032fbde5a6ab9bc13b83a059b93568c0bf3ae2/bundles/org.eclipse.swt/Eclipse%20SWT/gtk/org/eclipse/swt/widgets/Display.java#L3774-L3806

akurtakov (approved):

akurtakov:
Jenkins succeeds. Merging.

PR #1833, opened 2025-02-15
Run the Github verification with GTK4 enabled if it has GTK4 label

laeubi:
This is a different approach to

- https://github.com/eclipse-platform/eclipse.platform.swt/pull/1663

this should reuse the usual workflow, but set the `GTK4=1` when the PR is created with the gtk4 label ... as Jenkins is still running with gtk3 we still check both ways.


ptziegler:
Clean and simple. Thanks! 😄

Issue #207, opened 2022-06-10
Get access to ICoreWebView2-objects in Edge.java


jthier:
Problem:
We use the Edge as our browser in an RCP program and have multiple problems with the default behavior of the browser. Sadly the  Edge Object doesn't grant us any useful access to the ICoreWebView2 Objects, which would allow us to react to zoomfactor changes, manipulate the zoomfactor or change the shortcut behavior.

Solution:
Getter methods for the following objects would be necessary to use the Edge-Browser in a convenient way:
- ICoreWebView2Controller
- ICoreWebView2Settings
- 
Optional access to the other objects would also be useful:
- ICoreWebView2
- ICoreWebView2_2
- ICoreWebView2Environment2

Examples:
public ICoreWebView2Controller getICoreWebView2Controller() {
    return controller;
}
public ICoreWebView2Settings getICoreWebView2Settings() {
    return settings;
}

Possible alternatives:
It would also be an idea to enable setter-methods for a lot of variables, but this would be an unneccessary overhead.

PR #1834, opened 2025-02-15
[Win32] Add SSO support for SWT Browser using Edge backend

konczdev:
## Overview
This PR adds support for Single Sign-On (SSO) in SWT Browser (using Edge backend), 
allowing the browser to leverage the user's Windows credentials for authentication with 
Azure AD resources and Microsoft account-enabled websites.

## Features
- Adds two new system properties to control SSO functionality:
  - `org.eclipse.swt.browser.Edge.allowSingleSignOnUsingOSPrimaryAccount`
  - `org.eclipse.swt.browser.Edge.allowSSO` (shorthand alternative)
- Updates README documentation with new property descriptions
- Implements support for Single Sign-On through the Edge backend

## Implementation Details
- SSO is disabled by default (matches WebView2's default behavior)
- Either property can be set to `true` to enable the feature

## Documentation
- Added new section in README explaining the SSO properties and their effects

## Related Links
- [Microsoft WebView2 SSO Documentation](https://learn.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2environmentoptions.allowsinglesignonusingosprimaryaccount)

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
@@ -49,6 +49,8 @@ class Edge extends WebBrowser {
 	// System.getProperty() keys
 	static final String BROWSER_DIR_PROP = "org.eclipse.swt.browser.EdgeDir";
 	static final String BROWSER_ARGS_PROP = "org.eclipse.swt.browser.EdgeArgs";
+	static final String ALLOW_SINGLE_SIGN_ON_USING_OS_PRIMARY_ACCOUNT_PROP = "org.eclipse.swt.browser.Edge.allowSingleSignOnUsingOSPrimaryAccount";
+	static final String ALLOW_SINGLE_SIGN_ON_USING_OS_PRIMARY_ACCOUNT_PROP_2 = "org.eclipse.swt.browser.Edge.allowSSO";
 	static final String DATA_DIR_PROP = "org.eclipse.swt.browser.EdgeDataDir";
 	static final String LANGUAGE_PROP = "org.eclipse.swt.browser.EdgeLanguage";
 	static final String VERSIONT_PROP = "org.eclipse.swt.browser.EdgeVersion";
@@ -563,6 +565,7 @@ WebViewEnvironment createEnvironment() {
 	String browserDir = System.getProperty(BROWSER_DIR_PROP);
 	String browserArgs = System.getProperty(BROWSER_ARGS_PROP);
 	String language = System.getProperty(LANGUAGE_PROP);
+	boolean allowSSO = Boolean.getBoolean(ALLOW_SINGLE_SIGN_ON_USING_OS_PRIMARY_ACCOUNT_PROP) || Boolean.getBoolean(ALLOW_SINGLE_SIGN_ON_USING_OS_PRIMARY_ACCOUNT_PROP_2);
 	String dataDir = getDataDir(display);
 
 	// Initialize options
@@ -578,6 +581,11 @@ WebViewEnvironment createEnvironment() {
 	if (language != null) {
 		char[] pLanguage = stringToWstr(language);
 		options.put_Language(pLanguage);
+	}
+
+	if (allowSSO) {
+		int[] pAllowSSO = new int[]{1};
+		options.put_AllowSingleSignOnUsingOSPrimaryAccount(pAllowSSO);
 	}
 
 	// Create the environment
--- a/bundles/org.eclipse.swt/Readme.WebView2.md
+++ b/bundles/org.eclipse.swt/Readme.WebView2.md
@@ -67,6 +67,14 @@ language+country code that defines the browser UI language and preferred
 language for HTTP requests (`Accept-Languages` header).
 Example values: `en`, `ja`, `en-GB`, `de-AT`.
 
+The properties `org.eclipse.swt.browser.Edge.allowSingleSignOnUsingOSPrimaryAccount` or its shorthand
+`org.eclipse.swt.browser.Edge.allowSSO` enable Single Sign-On with Azure Active Directory (AAD)
+resources using the logged-in Windows account. This also enables SSO with websites using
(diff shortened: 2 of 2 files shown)


bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -49,6 +49,8 @@ class Edge extends WebBrowser {
 	// System.getProperty() keys
 	static final String BROWSER_DIR_PROP = "org.eclipse.swt.browser.EdgeDir";
 	static final String BROWSER_ARGS_PROP = "org.eclipse.swt.browser.EdgeArgs";
+	static final String ALLOW_SINGLE_SIGN_ON_USING_OS_PRIMARY_ACCOUNT_PROP = "org.eclipse.swt.browser.Edge.allowSingleSignOnUsingOSPrimaryAccount";
```
sratz, line 52: Please only use a single property. No need for redundant aliases that makes it more complicated than necessary.

sratz: I also think this should not be a system property but instead a `Control.setData(prop, value)`, so it can be set on individual browser instances.

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -580,6 +583,11 @@ WebViewEnvironment createEnvironment() {
 		options.put_Language(pLanguage);
 	}
 
+	if (allowSSO) {
```
sratz, line 586: I am wondering:
- Is there any harm in always setting the flag?
- I suppose WebKit on macOS / Linux have nothing comparable where we could copy from regarding default behavior / configurability?

sratz (changes requested):

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
konczdev: I tried to search the webkit repo for this or similar configurations but I found nothing.

> Is there any harm in always setting the flag?

Honestly, I don't know. I just discovered this issue accidentally: we have an Eclipse RCP application where the user can see an internal help page with an embedded browser. And this stopped working at all after the company introduced the Azure AD SSO and using the Edge browser the authentication is unsuccessful.

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
sratz: > Where would you put the `ALLOW_SINGLE_SIGN_ON_USING_OS_PRIMARY_ACCOUNT_PROP` property key constant with the `setData` implementation? Something like `Browser::EDGE_ALLOW_SINGLE_SIGN_ON_USING_OS_PRIMARY_ACCOUNT_PROPERTY`?

I was thinking along the lines of

```java
Browser browser = new Browser(parent, SWT.NONE);
browser.setData("org.eclipse.swt.browser.Edge.allowSingleSignOnUsingOSPrimaryAccount", Boolean.TRUE);
```
so that it can be controlled per instance and not only globally via system property.

The problem is, though, that we'd need to know this flag already when the constructor is called and the WebView2 instance is created.

> In long term #207 is maybe related.

True, exposing a whole proper platform-specific API is another idea.
Having a bunch of platform-specific flags to set on the `Widget#setData(String, Object)` is a more light-weight approach.

konczdev: Ok, I will update the pull request this way.

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -564,6 +565,11 @@ WebViewEnvironment createEnvironment() {
 	String browserDir = System.getProperty(BROWSER_DIR_PROP);
 	String browserArgs = System.getProperty(BROWSER_ARGS_PROP);
 	String language = System.getProperty(LANGUAGE_PROP);
+
+	Boolean instanceAllowSSO = (Boolean) browser.getData(ALLOW_SINGLE_SIGN_ON_USING_OS_PRIMARY_ACCOUNT_PROP);
```
sratz, line 569: I don't think the instance-specific approach will work with the `setData()` approach, since we need to already read the value while being inside the `new Browser(...)` constructor, so the `setData()` call after the instance is already created is too late.

I see three possible solutions:
1. We drop the instance-specific approach for now and only use the system property
2. We could call `setData()` on the parent, so we can check in the constructor. That would be a hack and would e.g. make it impossible to have two browsers next to each other in the same parent with different settings.
3. We use STYLE flags. That would make them public API for all platforms. We probably don't want to put so many platform-specific hints in there.

I think starting with 1) would be the simplest approach for now, like you initially suggested in this pull request.

There are many other interesting options in
`ICoreWebView2EnvironmentOptions` ... `ICoreWebView2EnvironmentOptions8` that we might want to be able to look at in the long run, so I think we will need are more holistic approach to control edge-specific settings -- potentially per instance.

konczdev: Thanks for the review, you're right, I overlooked this during implementation. I'll fix it as soon as I find a little time.

konczdev: Reverted to use the system property only.

sratz (approved):
Change looks good to me.

Any objections, @HeikoKlare?

HeikoKlare (approved):
This looks sound, so no objections from my side.

PR #1835, opened 2025-02-15
Revert change of Image surface color type from 24 to 32 bit

HeikoKlare:
In a previous change (5c2611d), the case distinction in creation of an image's surface depending on using GTK 4 or not has been removed. By accident, this change replaced the previously used 24 bit color format (for both cases) with a 32 bit color format.
This change revert that accidental modification.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Image.java
@@ -1260,7 +1260,7 @@ void init(int width, int height) {
 	this.type = SWT.BITMAP;
 
 	/* Create the pixmap */
-	surface = Cairo.cairo_image_surface_create(Cairo.CAIRO_FORMAT_ARGB32, width, height);
+	surface = Cairo.cairo_image_surface_create(Cairo.CAIRO_FORMAT_RGB24, width, height);
 	if (surface == 0) SWT.error(SWT.ERROR_NO_HANDLES);
 	// When we create a blank image we need to set it to 100 in GTK3 as we draw using 100% scale.
 	// Cairo will take care of scaling for us when image needs to be scaled.


ptziegler:
I've set up a test environment where I can run the problematic GEF test with an SWT version that contains this fix. But it seems like the test fails, regardless of whether a 32-bit or a 24-bit color depth is chosen.

Only when I switch back to the previous implementation via `gdk_window_create_similar_surface()` does the test succeed. This give me the impression that this method copies more attributes from the window than just color depth. The documentation even explicitly says so:

> Create a new surface that is as compatible as possible with the given window. For example the new surface will have the same fallback resolution and font options as `window`.

In the image I posted in the other PR, the text of the second line is slightly shorter than the text in the first line. So missing "font option" would explain those differences. Conversely, this then also means that the GTK4 implementation is incomplete, as there I would expect the same behavior...

What about reverting the changes to the Image class altogether? From what I can tell, the decision to move away from `cairo_image_surface_create` was done intentionally with https://github.com/eclipse-platform/eclipse.platform.swt/commit/5f4db6cac20e7848d08e2ea1bbe5bd403d3937b3 which to me sounds unrelated to the clipping issue you fixed in the other PR.

HeikoKlare:
> In the image I posted in the other PR, the text of the second line is slightly shorter than the text in the first line. So missing "font options" would explain those differences. Conversely, this then also means that the GTK4 implementation is incomplete, as there I would expect the same behavior...

That's interesting. I do not know what other properties are copied by the call to `gdk_window_create_similar_surface()` causing the difference to using the Cairo API. But we should be aware that this is the only place in `Image` where this method is still used, and this was only (re-)introduced as a workaround because the actual reason for the issue with the `copyArea()` method was not identified back then (https://bugs.eclipse.org/bugs/show_bug.cgi?id=571166) but is now fixed with https://github.com/eclipse-platform/eclipse.platform.swt/pull/1757. All other places in `Image` use `cairo_image_surface_create()`, so relying on some specify behavior of copying properties from the root surface when using the constructor calling this `init()` method with  `gdk_window_create_similar_surface()` seems to be wrong (or at least error prone) to me.

> What about reverting the changes to the Image class altogether? From what I can tell, the decision to move away from `cairo_image_surface_create` was done intentionally with [5f4db6c](https://github.com/eclipse-platform/eclipse.platform.swt/commit/5f4db6cac20e7848d08e2ea1bbe5bd403d3937b3) which to me sounds unrelated to the clipping issue you fixed in the other PR.

The call to `gdk_window_create_similar_surface()` has already been removed several years ago, together with all the others (https://github.com/eclipse-platform/eclipse.platform.swt/commit/b84f0060d2d31f1cd93acbbd184953522870c7d2). Only at this single place the change was reverted (https://github.com/eclipse-platform/eclipse.platform.swt/commit/af9f5cee1641970fa03ce813da7ba1dfc6e94141) because of a bug introduced to the line number ruler (which is the bug I addressed with the mentioned PR).
So even though I am not deep into the GTK implementation of SWT, it seems to me like getting rid of `gdk_window_create_similar_surface()` is the right thing to do (also regarding GTK 4 compatibility). Note that we have currently no single consumer of that method in the GTK SWT implementation.

Can you detail what exactly you (i.e., the GEF code consuming this method) expect from it? Maybe we can directly address that requirement rather than reverting to a workaround. But in case it's necessary to revert, I don't see any reasons against that (except that we need to find a proper solution to still get rid of that workaround as a follow-up task). We only need to make sure that the line number ruler still works when reverting the change to the `Image` class (see https://github.com/eclipse-platform/eclipse.platform.ui/pull/2754), but from a quick test I did locally it still seems to work fine.

ptziegler:
The issue stems from one of our [test cases](https://github.com/eclipse-gef/gef-classic/blob/cc892f0ad372fe399517326a2451a871ca97d59b/org.eclipse.draw2d.tests/src/org/eclipse/draw2d/test/AdvancedGraphicsTests.java#L140-L172) failing because of this change.

In short: In our Graphics object, we push the current state of the GC object on a stack, modify the text anti-aliasing and then afterwards restore the old state with a pop operation.

This no longer works and even though we call `setTextAntialiasing()` with the original value returned by `getTextAntialiasing()`, we can now see a different behavior when actually paining strings.

ptziegler:
Or maybe to rephrase the problem: When creating a GC object, `getTextAntialiasing()` returns `SWT.DEFAULT` but behaves as if `SWT.OFF` is set.

ptziegler:
> Can we somehow reproduce the failure of the added test locally? On my system it succeeds even on the master state, so I think it's up to the specifics of the GitHub Actions environment you mentioned in the other PR, isn't it?

That's the reason why this a real pain to figure out... The only way I see this problem is when an Xvnc server is running. I also don't see the issue locally but only when the tests are executed on either the Jenkins and as a GitHub action.

ptziegler:
Out of interest: What system are you using? 

I've just booted up my Ubuntu VM and there I see the same issue that is reported on the build nodes:

image

That said, I think I see what the problem is. Let's consider the following situation:

- On Gnome, the default anti-aliasing setting seems to be "Grayscale", which corresponds to `SWT.ON`.
- A GC object might be created without Cairo context (i.e. data.cairo == null) 

In this case, `getTextAntialias()` wrongfully returns `SWT.DEFAULT` because the method is unable to read the proper value and uses this value as fallback.

https://github.com/eclipse-platform/eclipse.platform.swt/blob/da081e9f0c53e2d6c835632380575fe744198e3d/bundles/org.eclipse.swt/Eclipse%20SWT/gtk/org/eclipse/swt/graphics/GC.java#L2532-L2547

PR #1844, opened 2025-02-18
Fix Sash Layout Shaking in win32 #62

amartya4256:
This PR fixes the check for enabling and disabling the redraw of controls in a sashlayout which leads to a shaky sashlayout on resizing it.

contributes to

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
@@ -3635,8 +3635,8 @@ public void setRedraw (boolean redraw) {
 	 *
 	 * https://github.com/eclipse-platform/eclipse.platform.swt/issues/1122
 	 */
-	boolean isShown = isVisible() && !isDisposed();
-	if (!redraw && isShown && embedsWin32Control()) {
+	if (!redraw && embedsWin32Control()) {
+		drawCount++;
 		return;
 	}
 
@@ -3677,6 +3677,9 @@ public void setRedraw (boolean redraw) {
 }
 
 private boolean embedsWin32Control () {
+	if (this.isDisposed() || !this.isVisible()) {
+		return false;
+	}
 	if (this instanceof Browser browser) {
 		// The Edge browser embeds webView2
 		return "edge".equals(browser.getBrowserType());


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
```diff
@@ -3635,8 +3635,8 @@ public void setRedraw (boolean redraw) {
 	 *
 	 * https://github.com/eclipse-platform/eclipse.platform.swt/issues/1122
 	 */
-	boolean isShown = isVisible() && !isDisposed();
-	if (!redraw && isShown && embedsWin32Control()) {
+	if (!redraw && embedsWin32Control()) {
+		drawCount++;
```
amartya4256, line 3639: This draw count was also missing since on setRedraw(false) we do a drawCount ++, if we don't do it then the value keeps getting subtracted and never increases when the browser is present. This leads to the control flow never entering the if else block and always shake.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
```diff
@@ -3679,7 +3679,7 @@ public void setRedraw (boolean redraw) {
 private boolean embedsWin32Control () {
 	if (this instanceof Browser browser) {
 		// The Edge browser embeds webView2
-		return "edge".equals(browser.getBrowserType());
+		return "edge".equals(browser.getBrowserType()) && browser.isVisible() && !browser.isDisposed();
```
HeikoKlare, line 3682: I think we need to check for `isDisposed()` first (before any of the other checks) as they make all fail with an error for a disposed widget.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
HeikoKlare: Good catch! That's definitely a bug that a composite that once contained a Win32 control can not recover from.

HeikoKlare (commented):
Except for the mentioned order of condition checks, the change is sound and works as expected. Would nice if you could also improve the commit message to explain the issue.

It particularly fixes a regression introduced during 2025-03 development:
- #1588

So this should be merged for 2025-03 to avoid that the regression lands in the release.
@HannesWell may I ask you for PMC approval on merging this (with the mentioned improvement) for 2025-03?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
amartya4256: Here's the new order then
isDisposed > isVisible > isEdge

amartya4256: I have also added the check in OLEClientSite

amartya4256: In fact the only bug in this implementation was the drawCount

fedejeanne (approved):

fedejeanne:
@akurtakov you may also approve this one :-)

akurtakov:
I can not test and verify this but as people working on SWT on win32 are in agreement that this is good PR here is my approval. 
PMC_APPROVED

PR #1847, opened 2025-02-19
Fix Caret Rescaling for win32 #62

amartya4256:
This PR fixes caret rescaling by making sure the the field height is updated for every caret and Caret::resize is called for those carets where applicable.

contributes to

## To reproduce
This is my configuration and how I reproduce the error in my current Eclipse IDE.
- Activate _Monitor-specific UI scaling_ either via the preferences (`General > Appearance`) or by using the flag `-Dswt.autoScale.updateOnRuntime=true`
- 125% monitor on the left (primary)
- 175% monitor on the right
- Open editor, click somewhere -> caret appears
- Move to the other monitor -> caret broken: it's in the wrong position and has the wrong size

In this [video](https://github.com/user-attachments/assets/95b20774-b2e4-4f45-95f7-25bf97a519c6) you can see how I moved from the monitor on the right to the one on the left and then back. The caret is broken after I move to the monitor on the left and stays broken when I come back to the monitor on the right.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Caret.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Caret.java
@@ -663,9 +663,10 @@ public void setVisible (boolean visible) {
  */
 public static void win32_setHeight(Caret caret, int height) {
 	caret.checkWidget();
-	if(caret.height == height && caret.isCurrentCaret()) return;
-	caret.height = height;
-	caret.resized = true;
+	if(caret.height != height) {
+		caret.height = height;
+		caret.resized = true;
+	}
 	if(caret.isVisible && caret.hasFocus()) caret.resize();
 }
 


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Caret.java
```diff
@@ -663,9 +663,10 @@ public void setVisible (boolean visible) {
  */
 public static void win32_setHeight(Caret caret, int height) {
 	caret.checkWidget();
-	if(caret.height == height && caret.isCurrentCaret()) return;
-	caret.height = height;
-	caret.resized = true;
+	if(caret.height != height) {
```
fedejeanne, line 666: Why do you need to remove `caret.isCurrentCaret()` from this check?

fedejeanne (commented):
It works as expected and fixes the issue, I just have one question

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Caret.java
amartya4256: we don't care if the caret is current or not. We want to update the height for all the carets otherwise some carets would be inconsistent.

fedejeanne: That makes sense, I'm just concerned with unwanted side effects. After all, someone decided to put that check in there in the first place. 

Do you have any example of a "non-current caret" so I can test for it? FTR I re-added the check locally and (in my simple test scenario) the problem was fixed too, that's why I ask.

fedejeanne (approved):
@akurtakov requesting pmc approval to merge this one

akurtakov:
I would like to have one more win32 developer to test and verify before giving approval. @HannesWell @HeikoKlare or someone else would you please test/review.

akurtakov:
@eclipse-platform/eclipse-platform-project-leads If some of you can review please step in as we are pretty close to RC1.

fedejeanne:
I added the information on how to reproduce the error.

HannesWell:
> I added the information on how to reproduce the error.

I'm sorry but I'm still not able to reproduce it, not even to trigger the code in question.
I followed the instruction you provided and tried it with a Java-Editor (as in your video).
What seems strange is that the next within the Editor doesn't change its size although e.g. the shell titel changes it's size when moving from the 125% scaled monitor to the 175% scaled one or the other way round.

HannesWell (approved):
> @HannesWell The issue description seems to miss the information that it is related to monitor-specific scaling only. The code in question is only executed when that feature is enabled. 

That was indeed the missing part. Now I was able to reproduce it and can confirm that this fixes the issue. Thanks for this.

> The above mentioned also means that the change will _only_ have an effect when the (experimental) monitor-specific scaling preference is enabled, which makes the change rather safe for being that late in the development cycle.

Agree. But I'd leave the final decision to the platform PLs. :)

HeikoKlare (commented):
@amartya4256 I agree that this change is reasonable. However, part of the issue addressed here seems to also be that `StyledText::updateAndRefreshCarets` calls the update functionality on the same caret multiple times. Isn't that somehow unexpected and unintended as well? I.e., shouldn't `StyledText::updateAndRefreshCarets` probably first collect all carets to update (removing duplicated) and then update them instead of calling the update on the same caret multiple times?

fedejeanne:
> But that could be done in a follow-up PR for cleanup. This PR seems to be fine as is to fix the underlying issue.

@HeikoKlare approved?

@eclipse-platform/eclipse-platform-project-leads can this still be merged (now for RC2)? It's been approved by 2 (soon maybe even 3) committers and it's been deemed safe to merge.

merks:
@iloveeclipse 

I think this should be merged but it would be nice if at least one project lead agrees.

iloveeclipse:
I personally would not merge non trivial SWT changes in RC2. This change is not a trivial one.

HeikoKlare:
> I personally would not merge non trivial SWT changes in RC2. This change is not a trivial one.

I agree that non-trivial changes should not be merged for RC2, but I would like to emphasize that this change has basically no risk (and may thus be considered rather "trivial"). The adapted code is only used for monitor-specific scaling, which is an experimental feature on Windows that is not enabled by default. An experimental feature might not work perfectly fine anway (which is why it's flagged "experimental"), but improving it as far as possible for a release (such as with this change) improves the chance that people keep testing it and report back issues so that once delivering that feature as non-experimental and enabled-by-default, it will be in a better state.

iloveeclipse:
> The adapted code is only used for monitor-specific scaling, which is an experimental feature on Windows that is not enabled by default.

I haven't reviewed the code in debugger but I trust your assessment. So no objections from my side then.

HeikoKlare:
Thank you!

FWIW, here is a screenshot showing the single call of that method (done by the DPI change handler only used for monitor-specific scaling on Windows):
image

So with no objections from platform PL side, I consider your comment as PMC+1, @merks?

HeikoKlare:
Thank you all!

PR #1849, opened 2025-02-20
Edge: Ignore gotFocus if we were the one who set focus to WebView2

### LINKED ISSUE #1848: Edge: Focus jumps back to browser control when trying to leave
**To Reproduce**
- Start with a simple eclipse installation (e.g. platform runtime binary)
- Close welcome page
- Open 'Internal Web Browser' view and move it to the editor area
- Create a dummy project with a dummyFile.txt
- Open that file
- Switch to the browser tab
- Press Ctrl-E -> switch to dummyFile.txt
- Editor tab jumps to dummyFile.txt
- Editor tab immediately jumps back to Browser

**Screenshots**

Image

**Environment:**
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 

The are much more situations where this occurs. The Ctrl-E is just a very easy one to reproduce.

The problem is the following:
- When the Ctrl-E popup closes, focus is returned to SWT-Edge and we receive an `SWT.FocusIn` event (`Edge.browserFocusIn(Event)`).
- We also need to inform WebView2 of this fact and call `controller.MoveFocus(COM.COREWEBVIEW2_MOVE_FOCUS_REASON_PROGRAMMATIC);`
- SWT switches to another editor
- -----async----
- Now, **with an async delay**, WebView2 informs us again in `handleGotFocus` that it has received focus.
- We move focus to SWT-Browser again, even though this event is now obsolete

sratz:
Fixes #1848.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
@@ -81,7 +81,8 @@ class Edge extends WebBrowser {
 	static boolean inCallback;
 	boolean inNewWindow;
 	HashMap<Long, LocationEvent> navigations = new HashMap<>();
-	private boolean ignoreFocus;
+	private boolean ignoreGotFocus;
+	private boolean ignoreFocusIn;
 	private String lastCustomText;
 
 	private static record CursorPosition(Point location, boolean isInsideBrowser) {};
@@ -829,8 +830,17 @@ void browserDispose(Event event) {
 }
 
 void browserFocusIn(Event event) {
-	if (ignoreFocus) return;
+	if (ignoreFocusIn) return;
 	// TODO: directional traversals
+
+	// https://github.com/eclipse-platform/eclipse.platform.swt/issues/1848
+	// When we call ICoreWebView2Controller.MoveFocus(int) here,
+	// WebView2 will call us back in handleGotFocus() asynchronously.
+	// We need to ignore that next event, as in the meantime the user might
+	// have moved focus to some other control and reacting on that event
+	// would bring us back to the Browser.
+	ignoreGotFocus = true;
+
 	controller.MoveFocus(COM.COREWEBVIEW2_MOVE_FOCUS_REASON_PROGRAMMATIC);
 }
 
@@ -1350,6 +1360,11 @@ int handleNewWindowRequested(long pView, long pArgs) {
 }
 
 int handleGotFocus(long pView, long pArg) {
+	if (ignoreGotFocus) {
+		ignoreGotFocus = false;
+		return COM.S_OK;
+	}
+	// https://github.com/eclipse-platform/eclipse.platform.swt/issues/1139
 	// browser.forceFocus() does not result in
 	// Shell#setActiveControl(Control)
 	// being called and therefore no SWT.FocusIn event being dispatched,
@@ -1357,9 +1372,9 @@ int handleGotFocus(long pView, long pArg) {
 	// The solution is to explicitly send a WM_SETFOCUS
 	// to the browser, and, while doing so, ignoring any recursive
 	// calls in #browserFocusIn(Event).
-	ignoreFocus = true;
+	ignoreFocusIn = true;
 	OS.SendMessage (browser.handle, OS.WM_SETFOCUS, 0, 0);
-	ignoreFocus = false;
+	ignoreFocusIn = false;
 	return COM.S_OK;
 }
 


bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -1350,16 +1360,20 @@ int handleNewWindowRequested(long pView, long pArgs) {
 }
 
 int handleGotFocus(long pView, long pArg) {
+	if (ignoreGotFocus) {
+		ignoreGotFocus = false;
+		return COM.S_OK;
+	}
 	// browser.forceFocus() does not result in
```
HeikoKlare, line 1367: Maybe we should add a link to https://github.com/eclipse-platform/eclipse.platform.swt/issues/1139 for this comment as well, as we now have several focus-related workaround where it might be(come) interesting to take a look at the different issue descriptions and discussions.

Just to be sure: the workaround for the other issue will not be affected by this change as in that case the starting point is `handleGotFocus` and not `browserFocusIn`, will it?

HeikoKlare (commented):
In general, the change looks fine to me.

I was just wondering what happens in case multiple focus-in events happen. Let's say the user for some reason rapidly switched focus between the browser and some other control, such that `browserFocusIn` is called multiple times. Would it be possible that the async calls of `handleGotFocus` are all processed afterwards, such that the second and further calls will erroneously be processed again? In order words: might it be necessary to count the number of `browserFocusIn` calls and decrement them in `handleGotFocus` or would that be wrong?

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
sratz: > Maybe we should add a link to #1139 for this comment as well, as we now have several focus-related workaround where it might be(come) interesting to take a look at the different issue descriptions and discussions.

Good idea to link more issues, yes, will add this.



> Just to be sure: the workaround for the other issue will not be affected by this change as in that case the starting point is `handleGotFocus` and not `browserFocusIn`, will it?

Correct
- in the other case the starting point is the WebView2 (`handleGotFocus`, the user has clicked directly in the browser control). We send out an SWT event that we need to ignore ourselves in `browserFocusIn`. This aspect is synchronous so we can directly do a `disable -> send event -> enable` idiom.
- this case / PR is about the other way round: SWT receives a focus event because the shell containing the browser was (re-)activated and SWT restores focus to the Browser instance. We need to notify WebView2 of this fact so it can re-set the cursor/focus. But then WebView2 also informs as back with a focus event in `handleGotFocus`. But this happens asynchronously, so we can only do it heuristically.

They should be mutually exclusive, otherwise we would have already infinite ping-pong loops.

sratz:
> I was just wondering what happens in case multiple focus-in events happen. Let's say the user for some reason rapidly switched focus between the browser and some other control, such that `browserFocusIn` is called multiple times. Would it be possible that the async calls of `handleGotFocus` are all processed afterwards, such that the second and further calls will erroneously be processed again? In order words: might it be necessary to count the number of `browserFocusIn` calls and decrement them in `handleGotFocus` or would that be wrong?

Thought about this, too. Counting up/down an integer could be an option, but I was not sure if that would on the other hand cause other weird issues where we would somehow not react to an event that we should have.

Ignoring one single event solved the issues in the scenarios that we tested so I think this is a good starting point.

sratz:
Do you think we should get this into RC2?

HeikoKlare (approved):
>> I was just wondering what happens in case multiple focus-in events happen. Let's say the user for some reason rapidly switched focus between the browser and some other control, such that browserFocusIn is called multiple times. Would it be possible that the async calls of handleGotFocus are all processed afterwards, such that the second and further calls will erroneously be processed again? In order words: might it be necessary to count the number of browserFocusIn calls and decrement them in handleGotFocus or would that be wrong?
> 
> Thought about this, too. Counting up/down an integer could be an option, but I was not sure if that would on the other hand cause other weird issues where we would somehow not react to an event that we should have.
> 
> Ignoring one single event solved the issues in the scenarios that we tested so I think this is a good starting point.

Agreed. Also thought about potential "side" effects of counting the number of events, so following your proposal of starting with this simple solution sounds fine (in particular since some broken focus handling, as the worse case of a remaining issue, is of course annoying but usually not a severe, blocking issue).

> Do you think we should get this into RC2?

I am not sure about this. I would be slightly in favor of holding it back unless you found the issue significantly annoying or problematic in your use cases. The CTRL+E scenario does not seem to be that crucial to me and since we will effectively have no implicit testing of it between merging the PR and the release, we should be really sure that it cannot make anything worse than better. I am rather confident about that, but still we know that you never know.

sratz:
> I am not sure about this. I would be slightly in favor of holding it back unless you found the issue significantly annoying or problematic in your use cases. The CTRL+E scenario does not seem to be that crucial to me and since we will effectively have no implicit testing of it between merging the PR and the release, we should be really sure that it cannot make anything worse than better. I am rather confident about that, but still we know that you never know.

Agreed, this came up rather later late with thorough systematic test tours. I guess it's not critical as it feels more like a fluke to users than an actual systematic issue.

So I am fine with deferring to 2025-06.

sratz:
> > I was just wondering what happens in case multiple focus-in events happen. Let's say the user for some reason rapidly switched focus between the browser and some other control, such that `browserFocusIn` is called multiple times. Would it be possible that the async calls of `handleGotFocus` are all processed afterwards, such that the second and further calls will erroneously be processed again? In order words: might it be necessary to count the number of `browserFocusIn` calls and decrement them in `handleGotFocus` or would that be wrong?
> 
> Thought about this, too. Counting up/down an integer could be an option, but I was not sure if that would on the other hand cause other weird issues where we would somehow not react to an event that we should have.
> 
> Ignoring one single event solved the issues in the scenarios that we tested so I think this is a good starting point.

We encountered the first scenario where the counting would be needed.

It's actually not necessarily that the user clicks rapidly, it can be a simple logic in the same event loop iteration:

```java
Browser.setFocus();
Browser.setFocus();
```

Then what happens is exactly that the boolean is not enough:

```
Browser.setFocus();
\-...
  \-WM_SET_FOCUS
    \-...
      \- Edge.browserFocusIn()
            ignoreGotFocus = true;
            sendAsyncEvent();

Browser.setFocus();
\-...
  \-WM_SET_FOCUS
    \-...
      \- Edge.browserFocusIn()
            ignoreGotFocus = true;
            sendAsyncEvent();

Edge.handleGotFocus():
// ignore + reset ignoreGotFocus;

Edge.handleGotFocus():
// flag is gone but the event should have been ignored as well.
```

Using counters instead of booleans solves this.

I'll do some more testing and will provide a PR.

PR #1850, opened 2025-02-21
Fix blurry shell icon #62

fedejeanne:
Windows only: wse the original width and height of the images set to the shell when searching for the best fit instead of adapting them to the zoom, otherwise the smallest image will always be picked and it will result in a blurry shell icon when scaling up for monitors with a higher zoom levels.

contributes to

## How to test
- Start the _Runtime Application_ with **Monitor-specific UI scaling** active
- Move it to a 200% monitor: the shell (window) icon should look sharper

**Before this PR**
image

**After this PR**
image

## Is this a regression?
~**Yes**~ Sort of (one needs to activate the experimental feature _Monitor-specific UI scaling_ in order to experience this behavior), introduced when we started using the `SMOOTH` scaling method (instead of `NEAREST`) on Windows (df0ae71aa5a5e6b1016cd4207c630196b6e7e394 / ffa10849c52645026eb96daca7a8194d48017abb). 

As demonstration, here's how the shell icon looks like **without** this PR and using the `NEAREST` scaling method:

image

And this is how it looks like in **4.34** (identical):
image

Since reverting back to the `NEAREST` scaling method has other implications (and overall the scaling method `SMOOTH` is better), I still propose to accept this PR as a regression fix even though it technically doesn't restore the previous behavior but it makes it better.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Decorations.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Decorations.java
@@ -899,7 +899,7 @@ void setImages (Image image, Image [] images) {
 				System.arraycopy (images, 0, bestImages, 0, images.length);
 				datas = new ImageData [images.length];
 				for (int i=0; i<datas.length; i++) {
-					datas [i] = images [i].getImageData (getZoom());
+					datas [i] = images [i].getImageData ();
 				}
 				images = bestImages;
 				sort (images, datas, getSystemMetrics (OS.SM_CXSMICON), getSystemMetrics (OS.SM_CYSMICON), depth);


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Decorations.java
```diff
@@ -899,7 +899,7 @@ void setImages (Image image, Image [] images) {
 				System.arraycopy (images, 0, bestImages, 0, images.length);
 				datas = new ImageData [images.length];
 				for (int i=0; i<datas.length; i++) {
-					datas [i] = images [i].getImageData (getZoom());
+					datas [i] = images [i].getImageData ();
```
amartya4256, line 902: Removing the getZoom call from here just calls getImageData(100), which doesn't change the ultimate effect of this. For example, even if you use getZoom() (e.g. 200) in this block and in the setImages you pass a small and a largeImage. If they have original size of 16 px and 32 px, it will still select smallImage of 16 px (32 px on 200) and largeImage of 32 px (64 px on 200). On debugging, I see it still allocates the image with the smallest size to smallIcon and the largest to largeIcon. I reckon the previous implementation is fine here.

I think the problem is rather in how you obtain the handle.
```
if (smallIcon != null) {
		switch (smallIcon.type) {
			case SWT.BITMAP:
				smallImage = Display.createIcon (smallIcon, getZoom());
				hSmallIcon = Image.win32_getHandle(smallImage, getZoom());
				break;
			case SWT.ICON:
				hSmallIcon = Image.win32_getHandle(smallIcon, getZoom());
				break;
		}
	}
	OS.SendMessage (handle, OS.WM_SETICON, OS.ICON_SMALL, hSmallIcon);
	if (largeIcon != null) {
		switch (largeIcon.type) {
			case SWT.BITMAP:
				largeImage = Display.createIcon (largeIcon, getZoom());
				hLargeIcon = Image.win32_getHandle(largeImage, getZoom());
				break;
			case SWT.ICON:
				hLargeIcon = Image.win32_getHandle(largeIcon, getZoom());
				break;
		}
	}
	OS.SendMessage (handle, OS.WM_SETICON, OS.ICON_BIG, hLargeIcon);
```

The call to Image::win32_getHandle will scale it down to the current zoom. Rather, we should use the correct zoom value for which the icon is supposed to be used. But the question is how the default image sizes which are passed in the method differ.

amartya4256: I do see the effect tho, but I dont understand why data has to do anything with it

fedejeanne: Hm, odd. I do see a difference. I am printing the results by adding these lines 

image

These are the outputs:
**Before this PR**
```
Zoom: 100
smallIcon bounds=Rectangle {0, 0, 16, 16}
largeIcon bounds=Rectangle {0, 0, 32, 32}
Zoom: 125
smallIcon bounds=Rectangle {0, 0, 16, 16}
largeIcon bounds=Rectangle {0, 0, 32, 32}
Zoom: 200
smallIcon bounds=Rectangle {0, 0, 16, 16}
largeIcon bounds=Rectangle {0, 0, 32, 32}
```
**After this PR**
```
Zoom: 100
smallIcon bounds=Rectangle {0, 0, 16, 16}
largeIcon bounds=Rectangle {0, 0, 32, 32}
Zoom: 125
smallIcon bounds=Rectangle {0, 0, 16, 16}
largeIcon bounds=Rectangle {0, 0, 48, 48}
Zoom: 200
smallIcon bounds=Rectangle {0, 0, 32, 32}
largeIcon bounds=Rectangle {0, 0, 48, 48}
```

As you can see, **without** this PR there is absolutely no difference in the output and **with** this PR, bigger icons are selected for higher zoom levels.

akurtakov:
@fedejeanne A procedural comment here - Please ask reviews from @eclipse-platform/eclipse-platform-project-leads and PMC . That way the former group will get notification thus better chances for review. 
About the PR itself - I'm not in a position to review/test it thus can not approve it.

fedejeanne:
cc PMC @HannesWell

HeikoKlare:
> **Yes**, this is a regression introduced when we started using the `SMOOTH` scaling method (instead of `NEAREST`) on Windows ([df0ae71](https://github.com/eclipse-platform/eclipse.platform.swt/commit/df0ae71aa5a5e6b1016cd4207c630196b6e7e394) / [ffa1084](https://github.com/eclipse-platform/eclipse.platform.swt/commit/ffa10849c52645026eb96daca7a8194d48017abb)).

We do not use `SMOOTH` scaling instead of `NEAREST` on Windows. The scaling mode is only changed automatically when monitor-specific scaling is used (which is still an experimental feature). Otherwise, one needs to set an according system property to enable `SMOOTH` scaling. So is this a regression limited to monitor-specific scaling or does it also affect default behavior of SWT / Eclipse RCP. And also the question arises whether this change affects default behavior as well, in particular, because this change does not fix the regression by reverting some change or the like but rather applies some improvement with behavior that has never been there like that before (so we may not know what effects this actually has, do we?). So wouldn't it be wise to do a proper fix for M1 instead of applying this quickfix in hope for improving an experimental feature but maybe also affecting default behavior in an unexpected way?

fedejeanne:
> So wouldn't it be wise to do a proper fix for M1 instead of applying this quickfix in hope for improving an experimental feature but maybe also affecting default behavior in an unexpected way?

To quote [you](https://github.com/eclipse-platform/eclipse.platform.swt/pull/1847#issuecomment-2674938362):

> An experimental feature might not work perfectly fine anway (which is why it's flagged "experimental"), but improving it as far as possible for a release (such as with this change) improves the chance that people keep testing it and report back issues so that once delivering that feature as non-experimental and enabled-by-default, it will be in a better state.

HeikoKlare:
How is that quote related to this change? It refers to a PR that changes code which is _only_ called when an experimental feature is active. This PR affects default behavior of SWT, as the code is executed in every SWT-based application.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Decorations.java
fedejeanne: > I do see the effect tho, but I dont understand why data has to do anything with it

My intention when using `getImageMetadata()` (without `zoom`) was to retrieve the "original" metadata _i.e._ the size for the 100% image. Doesn't the method retrieve that?

If I do this:
```java
for (int i=0; i<images.length; i++) {
	ImageData imageData = images[i].getImageData();
	System.out.println("width=" + imageData.width + ", height=" + imageData.height + ", depth=" + imageData.depth);
}
System.out.println("------------------");
for (int i=0; i<images.length; i++) {
	ImageData imageData = images[i].getImageData(getZoom());
	System.out.println("width=" + imageData.width + ", height=" + imageData.height + ", depth=" + imageData.depth);
}
```

I see this:
```
width=16, height=16, depth=24
width=32, height=32, depth=24
width=48, height=48, depth=24
------------------
width=24, height=24, depth=32
width=48, height=48, depth=32
width=72, height=72, depth=32
```
As you can see, the second set of sizes (and `depth`???) are scaled, which doesn't fit the requirements.

Am I seeing this wrong, @amartya4256 ?

HeikoKlare: Isn't an icon that is 150% the height and width exactly what is expected at 150% zoom?
And the depth is not scaled to 150% (neither mathematically nor technically). It rather seems as if auto-scaling changes the image data type from some RGB type (24 bit) to some ARGB type (32 bit).

fedejeanne: It is what I expect but not what I need: I need to find the image whose _original_ intended size (without rescaling) best matches the desired size in pixels. For example, if I have 2 images of sizes 16x16 and 32x32 and I want to create a shell icon of size 64x64 then I want to use the image that was created with a size of 32x32 (and scale it later).

amartya4256: Apart from the depth discussion, here's the flow of the whole thing:

1. setImages takes images [] array - usually 2 images in it - one small image, one large image
2. If you call getImageData (with or without zoom), you can see the size difference as shown in Federico's logs.
3. by default small icon has a size 16 px and large icon 32 px with better resolution - which means large icon looks more detailed.
4. when you call getImageData(getZoom()) for both the image - let's say at 200%, it becomes 32 px and 64 px each.
5. Compare method is called with the desired size information using getSystemMetrics (OS.SM_CXSMICON) and getSystemMetrics (OS.SM_CXICON), in a usual case it should be 16 and 32 px.
6. getSystemMetrics (OS.SM_CXSMICON) is used to evaluate the small icon and getSystemMetrics (OS.SM_CXICON) is used to evaluate the large icon.
7. If we pass the already scaled image data to sort method with getSystemMetrics (OS.SM_CXICON) to evaluate the large icon, it looks for the image with closest size to 32. Since small image's imaeg data is already 32, it doesn't choose the 64 (large) one and it ends up looking pixelated at a higher zoom level.

For now, this fix is a good hack but since we are working with per dpi awareness, we should use the DPI aware system metrics (GetSystemMetricsForDPI) and find out a way to still obtain the large image instance for largeIcon. Anyway, the whole implementation of sort and compare method which uses loops to find out which one is big and which one is small looks quite old school and we can make it better design wise.

HeikoKlare: I took a deeper look into #1878 to understand what really happens here and left some comments there. From my current understanding, it would make sense to use `getImageData(100)` here, as the values are compared against 100% icon sizes, so why not explicitly retrieve the 100% data (and really doing that by passing 100 to avoid that the reader needs to understand the contract of parameter-less `getImageData()`)?

PR #1851, opened 2025-02-21
[win32] Cache font height to prevent inconsistency

akoch-yatta:
When a font is initialized with FontData the height in points of it and the desired target zoom will be used to calculate the font height in pixels, e.g. font height 10.0 pt on zoom 175 results in pixel height of 23.333, rounded down to 23px.
When the FontData of an existing font is retrieved (Font::getFontData) the Logfont data from the OS together with the font zoom (Font.zoom) are used to recalculate the original font height, e.g. 23px results in 9.86 pt 
As the conversion to pixels involved rounding, information is lost and the reversed calculation will most likely result in a different value than was used on creation time

With monitor-specific scaling a font for a different zoom will be calculated on the basis of the FontData of an existing base font. This can lead to font heights in pixels that differ from the "correct" value. To solve this limitation each font handle and its original font height in points are cached in a static Map to recreate the original value when Font::getFontData is called.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Font.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Font.java
@@ -14,6 +14,8 @@
 package org.eclipse.swt.graphics;
 
 
+import java.util.*;
+
 import org.eclipse.swt.*;
 import org.eclipse.swt.internal.*;
 import org.eclipse.swt.internal.win32.*;
@@ -55,6 +57,9 @@ public final class Font extends Resource {
 	 * (Warning: This field is platform dependent)
 	 */
 	int zoom;
+
+	private static Map<Long, Float> fontHeightCache = new HashMap<>();
+
 /**
  * Prevents uninitialized instances from being created outside the package.
  */
@@ -176,6 +181,7 @@ public Font(Device device, String name, int height, int style) {
 @Override
 void destroy() {
 	OS.DeleteObject(handle);
+	fontHeightCache.remove(handle);
 	handle = 0;
 }
 
@@ -213,7 +219,13 @@ public FontData[] getFontData() {
 	if (isDisposed()) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
 	LOGFONT logFont = new LOGFONT ();
 	OS.GetObject(handle, LOGFONT.sizeof, logFont);
-	float heightInPoints = device.computePoints(logFont, handle, zoom);
+
+	float heightInPoints;
+	if (fontHeightCache.containsKey(handle)) {
+		heightInPoints = fontHeightCache.get(handle);
+	} else {
+		heightInPoints = device.computePoints(logFont, handle, zoom);
+	}
 	return new FontData[] {FontData.win32_new(logFont, heightInPoints)};
 }
 
@@ -240,6 +252,7 @@ void init (FontData fd) {
 	handle = OS.CreateFontIndirect(logFont);
 	logFont.lfHeight = lfHeight;
 	if (handle == 0) SWT.error(SWT.ERROR_NO_HANDLES);
+	fontHeightCache.put(handle, fd.height);
 }
 
 /**
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/ScalingSWTFontRegistry.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/ScalingSWTFontRegistry.java
@@ -53,8 +53,6 @@ final class ScalingSWTFontRegistry implements SWTFontRegistry {
 
 		private Font scaleFont(int zoom) {
 			FontData fontData = baseFont.getFontData()[0];
-			int baseZoom = computeZoom(fontData);
-			fontData.data.lfHeight = Math.round(1.0f * fontData.data.lfHeight * zoom / baseZoom);
 			Font scaledFont = Font.win32_new(device, fontData, zoom);
 			addScaledFont(zoom, scaledFont);
 			return scaledFont;
@@ -134,9 +132,8 @@ final class ScalingSWTFontRegistry implements SWTFontRegistry {
 		if (fontKeyMap.containsKey(fontData)) {
 			container = fontKeyMap.get(fontData);
 		} else {
-			int calculatedZoom = computeZoom(fontData);
-			Font newFont = Font.win32_new(device, fontData, calculatedZoom);
-			container = new ScaledFontContainer(newFont, calculatedZoom);
+			Font newFont = Font.win32_new(device, fontData, zoom);
+			container = new ScaledFontContainer(newFont, zoom);
 			fontHandleMap.put(newFont.handle, container);
 			fontKeyMap.put(fontData, container);
 		}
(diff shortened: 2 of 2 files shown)


HeikoKlare (commented):
First of all, this solved the described inconsistent scaling scenario for me.

I am just wondering whether it is really necessary to manage a cache. I simplified this to just store the height passed via the font data in the `init` method as an instance variable and it produces the same result in the scenario described in the PR message.
The only difference when using a cache is that also the fonts created via `win32_new(...)` passing a handle use the same height, but I do not even see that those methods are called by anything related to StyledText which is where we faced the issues with inconsistent scaling.

akoch-yatta:
> First of all, this solved the described inconsistent scaling scenario for me.
> 
> I am just wondering whether it is really necessary to manage a cache. I simplified this to just store the height passed via the font data in the `init` method as an instance variable and it produces the same result in the scenario described in the PR message. The only difference when using a cache is that also the fonts created via `win32_new(...)` passing a handle use the same height, but I do not even see that those methods are called by anything related to StyledText which is where we faced the issues with inconsistent scaling.

It solved the effect yes, but e.g. each call to Control::getFont would return a Font that returns "wrong" FontData with a possibly wrong height. So if you create a font using that font as basis you risk a wrong font height.

HeikoKlare:
> It solved the effect yes, but e.g. each call to Control::getFont would return a Font that returns "wrong" FontData with a possibly wrong height. So if you create a font using that font as basis you risk a wrong font height.

If a control does not have an explicit font set, it retrieves one from the OS, which has never been added to the cache anyway. If you replace the return in control with this, you can debug that the font data are not retrieved from the cache:
```java
        Font f = Font.win32_new (display, hFont, getShell().nativeZoom);
	f.getFontData();
	return f;
```

I think when we introduce such a cache, we should ensure that every handle is added to it, so that the "fallback logic" in `Font::getFontData` if no value is present in the cache is not necessary. This could be achieved by doing the calculation done in `getFontData` for those fonts not initialized with according FontData to store the height of but just with a handle right at initialization. But we should only do that if storing the height at each font itself is not sufficient.

akoch-yatta:
> > It solved the effect yes, but e.g. each call to Control::getFont would return a Font that returns "wrong" FontData with a possibly wrong height. So if you create a font using that font as basis you risk a wrong font height.
> 
> If a control does not have an explicit font set, it retrieves one from the OS, which has never been added to the cache anyway. If you replace the return in control with this, you can debug that the font data are not retrieved from the cache:
> 
> ```java
>         Font f = Font.win32_new (display, hFont, getShell().nativeZoom);
> 	f.getFontData();
> 	return f;
> ```
> 
> I think when we introduce such a cache, we should ensure that every handle is added to it, so that the "fallback logic" in `Font::getFontData` if no value is present in the cache is not necessary. This could be achieved by doing the calculation done in `getFontData` for those fonts not initialized with according FontData to store the height of but just with a handle right at initialization. But we should only do that if storing the height at each font itself is not sufficient.

Yes, fair point. I simplified the PR.

I proposed to define the new fontHeight attribute as final to make sure it is always initialized with a value, which lead to a questionable font height of 0 in the Font(Device) constructor. I'm open for a better idea.

HeikoKlare:
> I proposed to define the new fontHeight attribute as final to make sure it is always initialized with a value, which lead to a questionable font height of 0 in the Font(Device) constructor. I'm open for a better idea.

That constructor is only called from the two `win32_new` methods. Can't we change this constructor to accept a zoom and a handle (which are currently set by the `win32_new` methods) and then initialize the new `fontHeight` from the handle with the functionality currently implemented in `getFontData()`?

akoch-yatta:
> > I proposed to define the new fontHeight attribute as final to make sure it is always initialized with a value, which lead to a questionable font height of 0 in the Font(Device) constructor. I'm open for a better idea.
> 
> That constructor is only called from the two `win32_new` methods. Can't we change this constructor to accept a zoom and a handle (which are currently set by the `win32_new` methods) and then initialize the new `fontHeight` from the handle with the functionality currently implemented in `getFontData()`?

Yes, that should work. For me it didn't work properly, because the DLLs (either from the runtime or the github actions) still created a UnsatisfiedLink exception although they properly work on runtime. I didn't get it to work with that, but I noticed, they are way smaller than the ones in the repo (created via Jenkins?).

akoch-yatta:
> > I proposed to define the new fontHeight attribute as final to make sure it is always initialized with a value, which lead to a questionable font height of 0 in the Font(Device) constructor. I'm open for a better idea.
> 
> That constructor is only called from the two `win32_new` methods. Can't we change this constructor to accept a zoom and a handle (which are currently set by the `win32_new` methods) and then initialize the new `fontHeight` from the handle with the functionality currently implemented in `getFontData()`?

In the current state of my PR I introduced a new private constructor to do exactly that, so the mentioned constructor is not used anymore at all. 
I don't know what the use a this package protected constructor is (or was). It was only used internally anyway (event before my PR) so it could have been private. Thats why I didn't delete it

HeikoKlare:
> In the current state of my PR I introduced a new private constructor to do exactly that, so the mentioned constructor is not used anymore at all.
> I don't know what the use of this package protected constructor is (or was). It was only used internally anyway (even before my PR) so it could have been private. Thats why I didn't delete it

Ah, I didn't check the latest changes to the PR, that's of course fine. I also don't have any clue what these package protected constructors were used for. They exist for every Resource on every OS and some state that it "Prevents uninitialized instances from being created outside the package". But I don't think how that should be possible even without that constructor. Note that in the original SWT code (24 years ago) that was actually a package protected default constructors which at some point in time was already changed to accept a device as an argument. I cannot image what different it might make if we changed that constructor (except that reflective calls to it would fail...). Still, it's probably better to keep it as is (i.e., initialize Fonts via that constructor with a height of 0) and leave a potential cleanup of those constructors as a potential separate task.

HeikoKlare (commented):
The changes look fine now. I will also test them again, but can you please first have a look at the failing test, @akoch-yatta?

akoch-yatta:
> The changes look fine now. I will also test them again, but can you please first have a look at the failing test, @akoch-yatta?

One test actually failed because it tested something wrong. Font height in points of 10 is in pixels not double the size in the test setup (13 -> 26 before and wrong, 13 -> 27 now). With a point height of 9 it works again and I would argue it is fine, because it just tests that the scaled variant is different as expected.
The other test fails because of the baseFont being disposed. But with our recent changes, we could get rid of this baseFont overall, so I will create a different PR that does the refactoring and rebase this PR on top of that.

HeikoKlare:
> The other test fails because of the baseFont being disposed. But with our recent changes, we could get rid of this baseFont overall, so I will create a different PR that does the refactoring and rebase this PR on top of that.

Can you rebase this on top of #1868 to see if this one is done then?

akoch-yatta:
> > The other test fails because of the baseFont being disposed. But with our recent changes, we could get rid of this baseFont overall, so I will create a different PR that does the refactoring and rebase this PR on top of that.
> 
> Can you rebase this on top of #1868 to see if this one is done then?

Despite my assumption this is not necessary. The PRs are not related. Interestingly both PRs would make the same test fail because of different reasons, so this change is in both PRs, but this will solved itself after rebasing before the merge.

HeikoKlare:
> Despite my assumption this is not necessary. The PRs are not related. Interestingly both PRs would make the same test fail because of different reasons, so this change is in both PRs, but this will solved itself after rebasing before the merge.

I see, thanks for the clarification.

HeikoKlare (approved):

PR #1854, opened 2025-02-25
Add test to Combo

fedejeanne:
Checks existing behavior which seems to be a de-facto API _i.e._ that calling `getSelectionIndex()` during a modification event returns the up-to-date selection index.


akurtakov:
Version bump of the tests bundle is needed for the build to proceed.

Issue #1843, opened 2025-02-18
[TEST][windows] Seeing new Junit failure from 16th Feb build onwards


deepika-u:
**Describe the bug**
test_setTextContainingScript_applicationLayerProgressListenerMustSeeUpToDateDom test case is failing.

**Expected behavior**
This junit was passing in the previous builds, so did anything change recently regarding this?

**Detailed error**
```
Multiple Failures (2 failures) org.junit.ComparisonFailure: expected:<...ressListener: Found [1] h1 tag(s)> but was:<...ressListener: Found [0] h1 tag(s)> java.lang.RuntimeException: AwtScreenshot VM finished with exit code 1: Exception in thread "main" java.awt.AWTError: no screen devices at java.desktop/sun.awt.Win32GraphicsEnvironment.getDefaultScreenDevice(Win32GraphicsEnvironment.java:101) at java.desktop/java.awt.Robot.<init>(Robot.java:135) at org.eclipse.test.AwtScreenshot.main(AwtScreenshot.java:43)

org.opentest4j.MultipleFailuresError: Multiple Failures (2 failures)
org.junit.ComparisonFailure: expected:<...ressListener: Found [1] h1 tag(s)> but was:<...ressListener: Found [0] h1 tag(s)>
java.lang.RuntimeException: AwtScreenshot VM finished with exit code 1:
Exception in thread "main" java.awt.AWTError: no screen devices
at java.desktop/sun.awt.Win32GraphicsEnvironment.getDefaultScreenDevice(Win32GraphicsEnvironment.java:101)
at java.desktop/java.awt.Robot.<init>(Robot.java:135)
at org.eclipse.test.AwtScreenshot.main(AwtScreenshot.java:43)

at org.eclipse.test.TracingSuite$TracingRunNotifier.fireTestFinished(TracingSuite.java:176)
at org.eclipse.test.TracingSuite.runChild(TracingSuite.java:287)
Suppressed: org.junit.ComparisonFailure: expected:<...ressListener: Found [1] h1 tag(s)> but was:<...ressListener: Found [0] h1 tag(s)>
at org.junit.Assert.assertEquals(Assert.java:117)
at org.junit.Assert.assertEquals(Assert.java:146)
[...]

PR #1856, opened 2025-02-25
[win32] Refactor path to support multiple handles

akoch-yatta:
This PR refactors Path in the win32 implementation to better support multiple handles for different zoom settings of a Path when monitor-specific scaling is enables. The previous implementation only applied adaptions on the path to the initial handle and relied on the path not to be changed after usage. This doesn't cover all scenarios and can lead to incompletely rendered path when reusing Path object over different zoom settings.

### How to test
You can test this adaption with the new "Path Re-usage"-Tab in the GraphicsExample. You need to:

1. Set up at least two monitors with different zoom, e.g. 100% and 200%
2. Start the GraphicsExample with additonal VM arguments:
```
-Dswt.autoScale=quarter
```

3. Move to the "Path Re-usage"-Tab, play around with the buttons and switch monitors in between. Without the changes in Path you will loose operations or don't see some off them on the monitors the Path was not created on.


amartya4256:
Design-wise, looks good to me. I would approve it (I dont have the button to approve the PR). @HeikoKlare I would request you to test it out as well.

amartya4256 (approved):

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Path.java
```diff
@@ -259,10 +227,7 @@ public void addPath(Path path) {
 	if (isDisposed()) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
 	if (path == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
 	if (path.isDisposed()) SWT.error(SWT.ERROR_INVALID_ARGUMENT);
-	//TODO - expose connect?
-	Gdip.GraphicsPath_AddPath(getHandle(initialZoom), path.getHandle(initialZoom), false);
-	currentPoint.X = path.currentPoint.X;
-	currentPoint.Y = path.currentPoint.Y;
+	this.operations.addAll(path.operations);
```
HeikoKlare, line 230: Is it correct to remove the original code here?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Path.java
```diff
@@ -679,22 +506,425 @@ void moveToInPixels(float x, float y) {
...
+		* point of the subpath after calling CloseFigure().  The
+		* fix is to remember the subpath starting point and use
+		* it instead.
+		*/
+		currentPoint.X = startPoint.X;
+		currentPoint.Y = startPoint.Y;
+	}
+}
+
+private record FlattenOperation(float flatness) implements Operation {
```
HeikoKlare, line 789: This seems to be unused.

HeikoKlare (commented):
Most of the changes look fine to me (see my comments for exception). 
I am just wondering what the introduced operations are used for in `Path`: if I am not mistaken, the stored operations are not used anywhere. Each operation is directly applied to all present handles for all zooms and in case a new handle for a new zoom is created, the existing logic via passing the `PathData` is used.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Path.java
akoch-yatta: Ah, you're right. When splitting up the PRs I left that in the wrong PR

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Path.java
```diff
@@ -679,22 +506,425 @@ void moveToInPixels(float x, float y) {
...
+	public AddStringOperation(String string, float x, float y, Font font) {
+		this.string = string;
+		this.x = x;
+		this.y = y;
+		this.font = font;
+	}
+
+	@Override
+	public void apply(PathHandle pathHandle) {
+		if (font.isDisposed()) SWT.error(SWT.ERROR_INVALID_ARGUMENT);
```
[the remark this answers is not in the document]
akoch-yatta: Btw this change makes even more sense, because later only the FontData is used anyway

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Path.java
akoch-yatta: Almost. I just missed to apply the new operation to add to the existing handles

akoch-yatta:
> Most of the changes look fine to me (see my comments for exception). I am just wondering what the introduced operations are used for in `Path`: if I am not mistaken, the stored operations are not used anywhere. Each operation is directly applied to all present handles for all zooms and in case a new handle for a new zoom is created, the existing logic via passing the `PathData` is used.

They have a use now. The are used in Path::addPath. Without that, we would have needed them only in #1869, because there the handles are completly generated on demand using the operations

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Path.java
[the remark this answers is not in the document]
HeikoKlare: Great, even better.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Path.java
HeikoKlare: I am not sure whether we should do that since actually the new operation does exactly what you would expect it to do. I just wanted to mentioned that we _might_ oversee something that eventually leads to derivations of behavior, so that we are aware this change. But I don't see why this might be anything else than minor changes in result because of limited precision of float values and different ways of applying the operations.

But I just checked out the tab you added to the GraphicsExample, which shows that the old and new implementation of `addPath` behave different. So we may need to revise this.

akoch-yatta: Oh damn, you are right. There is a connecting path if you have done something before. That could probably be emulated be an additional move operation, but I would probably favor my approach above more

HeikoKlare: That's perfectly fine. It's just that we now have the problem of storing a Path that might be disposed when querying the handle for the encapsulating Path at another zoom level. So we probably need to clone the path to be stored?

HeikoKlare (changes requested):
See [comment ](https://github.com/eclipse-platform/eclipse.platform.swt/pull/1856#discussion_r1983685520):

> It's just that we now have the problem of storing a Path that might be disposed when querying the handle for the encapsulating Path at another zoom level. So we probably need to clone the path to be stored?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Path.java
akoch-yatta: No, in this PR the new Path is created on its path data and then scaled. The operations are always applied immediately. But I added a test to ensure it doesn't break later

akoch-yatta:
> See [comment ](https://github.com/eclipse-platform/eclipse.platform.swt/pull/1856#discussion_r1983685520):
> 
> > It's just that we now have the problem of storing a Path that might be disposed when querying the handle for the encapsulating Path at another zoom level. So we probably need to clone the path to be stored?

Just answered above at the same time. It is not an issue in this PR. I will need to address that in the next one

HeikoKlare (approved):
I see. I missed that the operations are not stored (anymore) in this PR. Thanks for the clarification!

HeikoKlare:

PR #1858, opened 2025-02-25
[win32] Use operations for transform manipulation

akoch-yatta:
This PR refactors Transform in the win32 implementation to better support multiple handles for different zoom settings of a Transform when monitor-specific scaling is enabled. The previous implementation only applied adaptions to the initial handle of the transform and relied on the transform not to be changed afterwards. This doesn't cover all scenarios and can lead to unexpected behavior when re-using Transform objects over different zoom settings.

### How to test
You can test this adaption with the new "Transform Re-usage"-Tab in the GraphicsExample. You need to:

1. Set up at least two monitors with different zoom, e.g. 100% and 200%
2. Start the GraphicsExample with additonal VM arguments:
```
-Dswt.autoScale=quarter
```

3. Move to the "Transform Re-usage"-Tab, play around with the buttons and switch monitors in between. Without the changes in Transform you will loose operations or don't see some off them on the monitors the Transform was not created on.


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Transform.java
```diff
@@ -388,8 +395,166 @@ public void transform(float[] pointArray) {
  */
 public void translate(float offsetX, float offsetY) {
 	if (isDisposed()) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
-	Drawable drawable = getDevice();
-	Gdip.Matrix_Translate(getHandle(initialZoom), DPIUtil.scaleUp(drawable, offsetX, initialZoom), DPIUtil.scaleUp(drawable, offsetY, initialZoom), Gdip.MatrixOrderPrepend);
+	storeAndApplyOperationForAllHandles(new TranslateOperation(offsetX, offsetY));
+}
+
+private class TransformHandle {
```
laeubi, line 401:
```suggestion
private final class TransformHandle {
```

or use a record

akoch-yatta: Good point, used a record here

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Transform.java
```diff
@@ -405,17 +552,21 @@ public String toString() {
...
+		TransformHandle transformHandle = new TransformHandle(handle, zoom);
+		zoomToHandle.put(zoom, transformHandle);
+		return transformHandle;
 	}
-	return zoomLevelToHandle.get(zoomLevel);
+	return zoomToHandle.get(zoom);
 }
 
+long getHandle(int zoomLevel) {
+	return getTransformHandle(zoomLevel).handle;
```
amartya4256, line 570: I think this can be removed and getTransformHandle(zoomLevel).handle can be used everywhere instead.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Transform.java
```diff
@@ -388,8 +398,145 @@ public void transform(float[] pointArray) {
...
+	public void apply(TransformHandle transformHandle) {
+		if (isDisposed()) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
+		Drawable drawable = getDevice();
+		long handle = transformHandle.handle;
+		int zoom = transformHandle.zoom;
+		Gdip.Matrix_Translate(handle, DPIUtil.scaleUp(drawable, offsetX, zoom), DPIUtil.scaleUp(drawable, offsetY, zoom), Gdip.MatrixOrderPrepend);
+	}
+}
+
+private interface Operation {
```
amartya4256, line 533: Just a thought. Since we will use Operation for all the resource, does it make sense to extract this as a FunctionalInterface in separate file and use it everywhere? I can see similar usage in Path and Region as well.

amartya4256: For the records like TransformHandle, RegionHandle, we can have a ResourceHandle instead.

akoch-yatta: Thought about the same. But I am hesitant to introduce some external class (even if package protected or internal package), because for now it is only used in win32.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Transform.java
akoch-yatta: Wouldn't this only be possible if we expose the TransformHandle class? 
Hmm, thinking about it I probably would not like to do that. TransformHandle is intended as an internal class to simplify the management of different OS handles. From external calls only the OS handle is relevant

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Transform.java
```diff
@@ -223,7 +230,7 @@ public void identity() {
  */
 public void invert() {
 	if (isDisposed()) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
```
HeikoKlare, line 232: Isn't this check obsolete as it's done by the operation (applies to all methods)?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Transform.java
```diff
@@ -388,8 +398,145 @@ public void transform(float[] pointArray) {
...
+	public void apply(TransformHandle transformHandle) {
+		if (isDisposed()) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
+		long handle = transformHandle.handle;
+		int zoom = transformHandle.zoom;
+		Gdip.Matrix_Multiply(handle, matrix.getHandle(zoom), Gdip.MatrixOrderPrepend);
+	}
+}
+
+private class RotateOperation implements Operation {
+	private float angle;
```
HeikoKlare, line 439: The values in all classes can be made final, can't they?

examples/org.eclipse.swt.examples/src/org/eclipse/swt/examples/graphics/TransformReuseTab.java
```diff
@@ -0,0 +1,119 @@
+/*******************************************************************************
+ * Copyright (c) 2025 Yatta.
```
HeikoKlare, line 2: Maybe align with other copyright notices?
```suggestion
 * Copyright (c) 2025 Yatta and others.
```

HeikoKlare (commented):
Change looks overall good, thanks! Only minor comments from my side.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Transform.java
akoch-yatta: yes, you are right, I can remove the checks in the operations

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Transform.java
akoch-yatta: yes, definitely

examples/org.eclipse.swt.examples/src/org/eclipse/swt/examples/graphics/TransformReuseTab.java
akoch-yatta: Sure

HeikoKlare (approved):
For me, the changes are fine now. I have tested them with the GraphicsExample. I have also executed GEF Classic's Draw2d tests against the changes as an additional means of validation: it's SWTGraphics also makes use of SWT's Transform.

PR #1859, opened 2025-02-25
[win32] Adapt Region to support creation of handles with operations with disposed resources

akoch-yatta:
This PR refactors Region in the win32 implementation to better support multiple handles for different zoom settings by creating all handles only on demand. It applies the same patterns as #1856 and #1858

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Region.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Region.java
@@ -15,6 +15,7 @@ package org.eclipse.swt.graphics;
 
 
 import java.util.*;
+import java.util.function.*;
 import java.util.stream.*;
 
 import org.eclipse.swt.*;
@@ -80,9 +81,8 @@ public Region () {
 public Region (Device device) {
 	super(device);
 	initialZoom = DPIUtil.getDeviceZoom();
-	long handle = OS.CreateRectRgn (0, 0, 0, 0);
+	long handle = newEmptyRegionHandle();
 	zoomToHandle.put(initialZoom, handle);
-	if (handle == 0) SWT.error(SWT.ERROR_NO_HANDLES);
 	init();
 	this.device.registerResourceWithZoomSupport(this);
 }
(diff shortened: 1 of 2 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Region.java
```diff
@@ -217,46 +219,27 @@ boolean containsInPixels (int x, int y) {
...
- * if they represent the <em>same</em> object using a class
- * specific comparison.
- *
- * @param object the object to compare with this object
- * @return <code>true</code> if the object is the same as this object and <code>false</code> otherwise
- *
- * @see #hashCode
- */
-@Override
-public boolean equals (Object object) {
```
HeikoKlare, line 255: API tools complain about hashcode/equals as public API methods being removed. I am not sure whether this is intended or a corner case that is not properly covered, still consumers may rely on proper implementations of the methods as they have existed before. On the other hand, the existing implementation of equals were effectively the same as the one from `Object`, as from my understanding the handle (for initial zoom) for two regions can never be the same. And `hashCode` used a unique memory pointer before and does it now as well (just taken from a different object).
So in my opinion the change is fine, but we should probably do it consistently across the three OS-specific implementations in a preparatory PR. It would probably be cleanest still remove those methods (and not replace them with just calls to the `super` implementation), but then we need to add API filters such that the API tooling does not complain.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Region.java
```diff
@@ -591,13 +565,40 @@ public void translate (Point pt) {
...
-
 private void storeAndApplyOperationForAllHandles(Operation operation) {
 	operations.add(operation);
-	zoomToHandle.forEach((zoom, handle) -> operation.apply(handle, zoom));
+	zoomToHandle.forEach((zoom, handle) -> operation.apply(handle));
+}
+
+private <T> T applyOnAnyHandle(Function<RegionHandle, T> function) {
+	if (zoomToHandle.isEmpty()) {
+		RegionHandle handle = newRegionHandle(DPIUtil.getDeviceZoom());
```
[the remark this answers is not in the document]
akoch-yatta: Adapted that. FYI, I additionally replaced DPIUtil.getDeviceZoom() with device.getDeviceZoom()

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Region.java
akoch-yatta: Adapted it as recommended and added the api filters

akoch-yatta:
@HeikoKlare I needed to adapt the PR a bit more, because tests failed because some operations that take a Region as input param failed because this region was disposed when a temporary handle was creation. As the OS.CombineRgn calls need a proper Region handle, I extracted a applyOnTemporaryHandle method to create and cleanup temporary handles on demand to theoretically encapsulate Region on Region operations as deep as you want.

HeikoKlare (approved):
Change looks good. I pushed a minor change to also use the introduced `newEmptyRegionHandle()` method in the constructor.

PR #1860, opened 2025-02-26
Stop crash of AccessibleObject on GTK 4

akurtakov:
GTK 4 port crashes on start up if worspace chooser has to be shown as AccessibleObject is totally disfunctional on GTK 4.x version but later on it tries to send events.
Disable it for now.


akurtakov:
Linux builds are fine and this PR touches only linux thus merging.

PR #1864, opened 2025-02-26
[win32] Dynamic handle creation for transform

akoch-yatta:
This adapts Transform in the win32 implementation to create handles only on demand. If a non-handle specific operation like isIdentity() is called, a temporary handle will be created and disposed afterwards if no handle exists already.

It contains the branch used for #1858 and is a second refactoring on transform, there it must be merged after #1858

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Transform.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Transform.java
@@ -14,6 +14,7 @@
 package org.eclipse.swt.graphics;
 
 import java.util.*;
+import java.util.function.*;
 
 import org.eclipse.swt.*;
 import org.eclipse.swt.internal.*;
@@ -39,11 +40,11 @@ import org.eclipse.swt.internal.gdip.*;
  */
 public class Transform extends Resource {
 
-	private int initialZoom;
-
 	private Map<Integer, TransformHandle> zoomToHandle = new HashMap<>();
 
 	private List<Operation> operations = new ArrayList<>();
+
+	private boolean isDestroyed;
 
 /**
  * Constructs a new identity Transform.
@@ -140,12 +141,8 @@ public Transform(Device device, float[] elements) {
  */
 public Transform (Device device, float m11, float m12, float m21, float m22, float dx, float dy) {
 	super(device);
-	initialZoom = DPIUtil.getDeviceZoom();
 	this.device.checkGDIP();
-	long handle = Gdip.Matrix_new(m11, m12, m21, m22,
-			DPIUtil.scaleUp(this.device, dx, initialZoom), DPIUtil.scaleUp(this.device, dy, initialZoom));
-	if (handle == 0) SWT.error(SWT.ERROR_NO_HANDLES);
-	zoomToHandle.put(initialZoom, new TransformHandle(handle, initialZoom));
+	storeAndApplyOperationForAllHandles(new SetElementsOperation(m11, m12, m21, m22, dx, dy));
 	init();
 	this.device.registerResourceWithZoomSupport(this);
 }
@@ -161,19 +158,14 @@ void destroy() {
 	device.deregisterResourceWithZoomSupport(this);
 	zoomToHandle.values().forEach(TransformHandle::destroy);
 	zoomToHandle.clear();
-	operations.clear();
+	this.isDestroyed = true;
 }
 
 @Override
 void destroyHandlesExcept(Set<Integer> zoomLevels) {
-	zoomToHandle.entrySet().removeIf(entry -> {
-		final Integer zoom = entry.getKey();
-		if (!zoomLevels.contains(zoom) && zoom != initialZoom) {
-			entry.getValue().destroy();
-			return true;
-		}
-		return false;
-	});
+	// As long as we keep the operations, we can cleanup all handles
+	zoomToHandle.values().forEach(TransformHandle::destroy);
+	zoomToHandle.clear();
 }
 
 /**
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Transform.java
```diff
@@ -539,6 +538,19 @@ private void storeAndApplyOperationForAllHandles(Operation operation) {
...
 }
 
+private <T> T applyOnAnyHandle(Function<TransformHandle, T> function) {
+	if (zoomToHandle.isEmpty()) {
+		TransformHandle handle = newTransformHandle(DPIUtil.getDeviceZoom());
+		try {
+			return function.apply(handle);
+		} finally {
+			handle.destroy();
+		}
```
HeikoKlare, line 548: Is this necessary to ensure that applying an operation behaves as if it was actually applied (including potential failures) even if no handles are currently available? Would be nice to have a short documentation about that because my first intent was to ask if this code isn't obsolete, so might be good to relieve other readers from that thought.

HeikoKlare: I see, thank you for the explanation. I missed the return.

Then I have a far more general question: This means we always take an arbitrary handle for an arbitrary zoom to generate the results for all methods from `isIdentity()` to even `getElements(...)`. Doesn't this mean that all the overhead for the zoom-specific calculations inside the transform object are not necessary, as it's only about scaling the translation values in the matrix (index 4 and 5) upon access to it? In other words, won't the matrix managed by the transform object have the exact same values for every zoom, except that index 4 and 5 are scaled? That would mean we could simply do all the calculations on a single handle and just create another handle for another zoom out of it by rescaling index 4 and 5 of the matrix on demand.

akoch-yatta: We could probably simplify newTransformHandle with this, yes. But the rest wouldn't be affected. If multiple handles for different zoom are requested you need to apply each operation on all handles from that point on.

HeikoKlare: I just wanted to provide a snippet but saw that we actually do exactly what I had in mind in the existing implementation of `getHandle()`. I see that the issue is that we have to cache the handle for the next retrieval of the same zoom (and to clean is up of the Transform object is destroyed), as we would otherwise loose the dispose management for the handles.
So forget about that, it seems like I just had to clarify that for myself ;-)

Then I think it's fine like it is. Just having a small comment on this code and maybe improving clarity by renaming the handle to `temporaryHandle` would be nice. But this is already quite nitpicky.

HeikoKlare (approved):

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Transform.java
```diff
@@ -364,15 +368,19 @@ public void shear(float shearX, float shearY) {
...
-	int length = pointArray.length;
 	Drawable drawable = getDevice();
-	for (int i = 0; i < length; i++) {
-		pointArray[i] = DPIUtil.scaleUp(drawable, pointArray[i], initialZoom);
-	}
-	Gdip.Matrix_TransformPoints(getHandle(initialZoom), pointArray, length / 2);
-	for (int i = 0; i < length; i++) {
-		pointArray[i] = DPIUtil.scaleDown(drawable, pointArray[i], initialZoom);
-	}
+	applyOnAnyHandle(transformHandle -> {
```
amartya4256, line 372: should it not be storeAndApplyOperationForAllHandles instead?

amartya4256: Is this operation used to transform the provided points using the transform resource?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Transform.java
```diff
@@ -388,8 +396,136 @@ public void transform(float[] pointArray) {
...
+private <T> T applyOnAnyHandle(Function<TransformHandle, T> function) {
+	if (zoomToHandle.isEmpty()) {
+		TransformHandle temporaryHandle = newTransformHandle(DPIUtil.getDeviceZoom());
+		try {
+			return function.apply(temporaryHandle);
+		} finally {
+			temporaryHandle.destroy();
+		}
+	} else {
+		return function.apply(zoomToHandle.values().iterator().next());
```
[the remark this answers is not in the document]
amartya4256: Anyway, I think the name is a bit misleading. It sounds liek we are applying some operation (as in to modify) on the handle. A better suggestion could be "applyUsingAnyHandle"

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Transform.java
amartya4256: Strange choice of name for the API. Then, it's fine.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Transform.java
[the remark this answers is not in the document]
amartya4256: I think its wise to adapt the same for all the resources then.

akoch-yatta:
Rebased on master after #1858  was merged and addressed the comment from @amartya4256

PR #1867, opened 2025-02-27
Simplify scaling of ImageData with DPIUtil

HannesWell:
Wrapping an `ImageData` object and its zoom into a `ElementAtZoom` object just to unwrap it seems to be unnecessary.

Or is this part of a greater plan @akoch-yatta or @HeikoKlare?


HeikoKlare (commented):
Thank you for this cleanup! There is not greater plan for that method and it is obviously obsolete in it's current state. Maybe it was different at some point in time.

HannesWell:
With the current state of https://github.com/eclipse-platform/eclipse.platform.swt/pull/1828, I think the method to be removed here gets a revival. So this should at least be delayed until that PR is completed and then we can close or submit this based on the resulting content.

HeikoKlare:
Hannes, I have created #1887 with a slightly more "correct" replacement (as we should replace autoScale* calls with scale* calls on Windows) and added you as co-author there. I Thus closing this one as superceded.

PR #1868, opened 2025-02-28
[win32] Refactor ScalingSWTFontRegistry

akoch-yatta:
This PR refactors ScalingSWTFontRegistry to simplify creation and caching of scaled variants of custom and system fonts. This is possible because the conversion of SWT font height in points to the OS font height in pixels no longer relies on the primary monitor zoom, but can directly calculated based on the target zoom a font will be used on now.


HeikoKlare (approved):
The change looks sound to me and works as expected. The code for the registry is much cleaner now. Great to see such simplification!

Does this supercede #1825?

akoch-yatta:
> The change looks sound to me and works as expected. The code for the registry is much cleaner now. Great to see such simplification!
> 
> Does this supercede #1825?

Good pointer, yes. This makes the change from #1825 in this class obsolete

PR #1869, opened 2025-02-28
[win32] Dynamic handle creation for Path

akoch-yatta:
This commit adapts Path in the win32 implementation to create handles only on demand. If a non-handle specific operation like getBounds() is called, a temporary handle will be created and disposed afterwards if no handle exists already.

It contains the branch used for https://github.com/eclipse-platform/eclipse.platform.swt/pull/1856 and is a second refactoring on transform, there it must be merged after https://github.com/eclipse-platform/eclipse.platform.swt/pull/1856

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Path.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Path.java
@@ -14,6 +14,7 @@
 package org.eclipse.swt.graphics;
 
 import java.util.*;
+import java.util.function.*;
 
 import org.eclipse.swt.*;
 import org.eclipse.swt.internal.*;
@@ -42,9 +43,11 @@ import org.eclipse.swt.internal.win32.*;
  * @since 3.1
  */
 public class Path extends Resource {
-	private int initialZoom;
-
 	private Map<Integer, PathHandle> zoomToHandle = new HashMap<>();
+
+	private List<Operation> operations = new ArrayList<>();
+
+	private boolean isDestroyed;
 
 /**
  * Constructs a new empty Path.
@@ -72,16 +75,8 @@ public class Path extends Resource {
  * @see #dispose()
  */
 public Path (Device device) {
-	this(device, DPIUtil.getDeviceZoom());
-}
-
-private Path(Device device, int zoom) {
 	super(device);
 	this.device.checkGDIP();
-	initialZoom = zoom;
-	long handle = Gdip.GraphicsPath_new(Gdip.FillModeAlternate);
-	if (handle == 0) SWT.error(SWT.ERROR_NO_HANDLES);
-	zoomToHandle.put(initialZoom, new PathHandle(handle, initialZoom));
 	init();
 	this.device.registerResourceWithZoomSupport(this);
 }
@@ -125,11 +120,10 @@ public Path (Device device, Path path, float flatness) {
 	if (path == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
 	if (path.isDisposed()) SWT.error(SWT.ERROR_INVALID_ARGUMENT);
 	flatness = Math.max(0, flatness);
-	long handle = Gdip.GraphicsPath_Clone(path.getHandle(path.initialZoom));
-	if (flatness != 0) Gdip.GraphicsPath_Flatten(handle, 0, flatness);
-	if (handle == 0) SWT.error(SWT.ERROR_NO_HANDLES);
-	initialZoom = path.initialZoom;
-	zoomToHandle.put(initialZoom, new PathHandle(handle, initialZoom));
+	path.operations.forEach(this::storeAndApplyOperationOnAllHandles);
+	if (flatness != 0) {
+		storeAndApplyOperationOnAllHandles(new FlattenOperation(flatness));
+	}
 	init();
 	this.device.registerResourceWithZoomSupport(this);
 }
@@ -163,15 +157,9 @@ public Path (Device device, Path path, float flatness) {
  * @since 3.4
  */
 public Path (Device device, PathData data) {
-	this(device, data, DPIUtil.getDeviceZoom());
-
-}
-
-private Path(Device device, PathData data, int zoom) {
-	this(device, zoom);
+	this(device);
 	if (data == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
 	init(data);
-	this.device.registerResourceWithZoomSupport(this);
 }
 
 /**
@@ -205,7 +193,7 @@ private Path(Device device, PathData data, int zoom) {
  */
 public void addArc (float x, float y, float width, float height, float startAngle, float arcAngle) {
 	if (width == 0 || height == 0 || arcAngle == 0) return;
-	applyOperationForAllHandles(new AddArcOperation(x, y, width, height, startAngle, arcAngle));
+	storeAndApplyOperationOnAllHandles(new AddArcOperation(x, y, width, height, startAngle, arcAngle));
 }
 
 /**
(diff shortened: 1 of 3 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Path.java
```diff
@@ -679,22 +494,438 @@ void moveToInPixels(float x, float y) {
...
+private <T> T applyOnAnyHandle(Function<PathHandle, T> function) {
+	if (zoomToHandle.isEmpty()) {
+		PathHandle temporaryHandle = newPathHandle(DPIUtil.getDeviceZoom());
+		try {
+			return function.apply(temporaryHandle);
+		} finally {
+			temporaryHandle.destroy();
+		}
+	} else {
+		return function.apply(zoomToHandle.values().iterator().next());
```
amartya4256, line 927: Same concern as here: https://github.com/eclipse-platform/eclipse.platform.swt/pull/1864#discussion_r1983312112

amartya4256: Addressed in the reply to https://github.com/eclipse-platform/eclipse.platform.swt/pull/1864#discussion_r1983312112

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Path.java
```diff
@@ -125,11 +120,10 @@ public Path (Device device, Path path, float flatness) {
...
 	if (path.isDisposed()) SWT.error(SWT.ERROR_INVALID_ARGUMENT);
 	flatness = Math.max(0, flatness);
-	long handle = Gdip.GraphicsPath_Clone(path.getHandle(path.initialZoom));
-	if (flatness != 0) Gdip.GraphicsPath_Flatten(handle, 0, flatness);
-	if (handle == 0) SWT.error(SWT.ERROR_NO_HANDLES);
-	initialZoom = path.initialZoom;
-	zoomToHandle.put(initialZoom, new PathHandle(handle, initialZoom));
+	operations.addAll(path.operations);
+	if (flatness != 0) {
+		operations.add(new FlattenOperation(flatness));
```
HeikoKlare, line 125: This comment depends on the other general comment on error handling: In order to not hide potential failures, should we do `storeAndApply...` here? E.g., could passing a negative flatness yield an error that may pop up on access instead of instantiation with this change?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Path.java
```diff
@@ -926,10 +932,31 @@ private interface Operation {
 	void apply(PathHandle pathHandle);
 }
 
-private void applyOperationForAllHandles(Operation operation) {
+private void storeAndApplyOperationOnAllHandles(Operation operation) {
+	operations.add(operation);
 	zoomToHandle.values().forEach(operation::apply);
```
HeikoKlare, line 937: I just want to raise a thought here: In case `zoomToHandle` is empty, don't we need to apply the current (and all previous operations) to a temporary handle such that potential errors are still thrown when applying the operation at consumer side rather than when dynamically instantiating the resource handle for an according zoom? Otherwise this change would lead to different behavior, as some errors would be thrown at different calls.
On the other hand, such errors should not occur and would be a problem in productive use anyway (so hopefully no one catches them) and they are even not part of the contract of the methods. And of course, creating a properly initialized handle for the application of every operation would somehow conflict with the on-demand-creation introduced with this PR.

So I for myself would conclude that the change is fine, still I would like to have a common understanding and agreement on this: @akoch-yatta @amartya4256

HeikoKlare (commented):
Some comments or points for discussion on the overall design. I will test the change after we have resolved them.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Path.java
akoch-yatta: Negative flatness is not possible because of line 122.
He is definitely no handle available, so storeAndApply would do nothing except storing it. But out of consistency it still makes sense to do that here, because if there would be handles, it should probably be applied here.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Path.java
```diff
@@ -125,11 +120,10 @@ public Path (Device device, Path path, float flatness) {
 	if (path == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
 	if (path.isDisposed()) SWT.error(SWT.ERROR_INVALID_ARGUMENT);
 	flatness = Math.max(0, flatness);
-	long handle = Gdip.GraphicsPath_Clone(path.getHandle(path.initialZoom));
-	if (flatness != 0) Gdip.GraphicsPath_Flatten(handle, 0, flatness);
-	if (handle == 0) SWT.error(SWT.ERROR_NO_HANDLES);
-	initialZoom = path.initialZoom;
-	zoomToHandle.put(initialZoom, new PathHandle(handle, initialZoom));
+	operations.addAll(path.operations);
```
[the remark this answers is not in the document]
akoch-yatta: Hmm, in my testing I didn't find any issues with that. But regarding risk minimization, it probably makes sense to keep the logic as close as possible to the previous state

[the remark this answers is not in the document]
akoch-yatta: Having a second look into it, that would still be possible but more complex then I thought. I would need to dynamize the place where new OS handles are created, because the output of the ClonePath is already the new OS handle.
Instead I now only added a test to use this constructor with a Path with some operation and compared the PathData. What do you think @HeikoKlare ?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Path.java
akoch-yatta: I adapted these places to use storeAndApply instead

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Path.java
akoch-yatta: Yes, I agree to the assumption. Of course some parts of the logic are now executed later, e.g. when the first handle is fetched. 
But I would as well say, I don't see a practical problem with that. If there are any issues, you should still catch those with testing your implementation. Only scenario when you would miss that is when the Path is not used at all.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Path.java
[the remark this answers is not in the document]
HeikoKlare: You would need to create another, temporary handle based on the operations and call the clone operation on it, right? Seems rather complex to achieve the goal that should be same as just executing the operations, still the failing test on MacOS seems to indicate that currently the behavior is not correct, is it?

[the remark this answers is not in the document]
HeikoKlare: When writing the comment, I forgot that the MacOS implementation is not affected by the change at all. I agree that that's really weird because it means that the behavior of the Path API is different across the operating systems. Instead of moving the test to the Windows-only tests, I would rather add an `assume` to the test that it's not run on MacOS and add a message stating that behavior seems to be broken on MacOS, since the test _should_ actually run on MacOS successfully.

HeikoKlare (approved):

HeikoKlare:

PR #1870, opened 2025-03-02
Unify ImageLoader implementations and extract differences

HannesWell:
Extract existing differences in the `ImageLoader` implementation for Linux/GTK into an `InternalImageFileLoader` class that also exists for Windows and MacOS but just delegates to FileFormat.

This reduces the maintenance effort for `ImageLoader` as only one file has to be updated and any existing differences are handled in the much smaller and internal `InternalImageFileLoader` .


HannesWell:
@HeikoKlare if you and the others think this is fine, I suggest to submit this before other changes in this area as it avoids triplicated work to keep API changes in sync.

HeikoKlare (approved):
Making the `ImageLoader` a common class with all the common contents across the three OS-specific implementations looks like a nice refactoring to me 👍

PR #1874, opened 2025-03-04
Extract LOGPIXELS* usage to FontDialog::getDPI

amartya4256:
This PR removes the usage of OS.LOGPIXELS* in FontDialog. These values can be obtained using other means as the OS.LOGPIXELS provides the dpi values of the primary monitor at the app-startup time. It's usage has been replaced by passing the zoom directly in FontDialog::open method.

contributes to

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/FontDialog.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/FontDialog.java
@@ -200,9 +200,7 @@ public FontData open () {
 	if (fontData != null && fontData.data != null) {
 		LOGFONT logFont = fontData.data;
 		int lfHeight = logFont.lfHeight;
-		long hDC = OS.GetDC (0);
-		int pixels = -(int)(0.5f + (fontData.height * OS.GetDeviceCaps(hDC, OS.LOGPIXELSY) / 72));
-		OS.ReleaseDC (0, hDC);
+		int pixels = -(int)(0.5f + (fontData.height * getDPI() / 72));
 		logFont.lfHeight = pixels;
 		lpcf.Flags |= OS.CF_INITTOLOGFONTSTRUCT;
 		OS.MoveMemory (lpLogFont, logFont, LOGFONT.sizeof);
@@ -256,9 +254,9 @@ public FontData open () {
 			 * This will not work on multiple screens or for printing. Should use DC for the
 			 * proper device.
 			 */
-			long hDC = OS.GetDC(0);
-			int logPixelsY = OS.GetDeviceCaps(hDC, OS.LOGPIXELSY);
+			int logPixelsY = getDPI();
 			int pixels = 0;
+			long hDC = OS.GetDC(0);
 			if (logFont.lfHeight > 0) {
 				/*
 				 * Feature in Windows. If the lfHeight of the LOGFONT structure is positive, the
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/FontDialog.java
```diff
@@ -195,14 +196,13 @@ public FontData open () {
 	if (effectsVisible) {
 		lpcf.Flags |= OS.CF_EFFECTS;
 	}
+	int dpi = DPIUtil.mapZoomToDPI(getParent().getZoom());
```
akoch-yatta, line 199: This one would create a regression with int200, e.g. on a 150% primary monitor LOGPIXELSY would have returned 144, but getZoom would return 100 which is converted to 96.
We need to use getParent().nativeZoom here

akoch-yatta (changes requested):
Apart from my change request this PR looks good

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/FontDialog.java
amartya4256: Yes thats right. Oops!

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/FontDialog.java
```diff
@@ -195,14 +196,13 @@ public FontData open () {
 	if (effectsVisible) {
 		lpcf.Flags |= OS.CF_EFFECTS;
 	}
+	int dpi = DPIUtil.mapZoomToDPI(getParent().nativeZoom);
```
HeikoKlare, line 199: If I am not mistaken, this will introduce a regression: `OS.GetDeviceCaps(hDC, OS.LOGPIXELSY)` has a consistent value throughout application lifecycle (depending on primary monitor zoom at startup). This value will be different depending on the monitor on which the parent shell is placed. Maybe the old implementation was not correct, but then we need to document that and have a description how to reproduce. Actually, I was not even able to call the affected code so far. You must initialize the FontDialog with specific FontData before starting it to execute at all and so far I found no consumer doing that. Can you please provide a description (any existing snippet or functinality in SDK) how to call the code and test that it still works as expected?

akoch-yatta: Are we talking about monitor-specific scaling active? I know how to run into the code, but I wanna have a look at it with the correct scenario you have in your mind.

HeikoKlare: Yes, with monitor-specific scaling active. With monitor-specific inactive, the values in both cases should be effectively (the same) constants throughout an application lifecycle.

HeikoKlare: Thank you the clarification! Instead of closing this PR, maybe we can take the chance to adapt it to make the understanding we have more explicit in the code to easy understanding it in the future?
More concretely for example something like:
- Extract the `OS.GetDeviceCaps()` call into a separate `getDPI()` method
- Leave a comment in that method stating that we need to use that value, which is static throughout application lifecycle, because of the specific DPI awareness mode we use here that always depends on the DPI at application startup.

amartya4256: Done

HeikoKlare (approved):

PR #1876, opened 2025-03-04
Version bump for 4.36 stream

HeikoKlare:
Also removes obsolete API filters for previous version

### DIFF
--- a/binaries/org.eclipse.swt.cocoa.macosx.aarch64/.settings/.api_filters
+++ b/binaries/org.eclipse.swt.cocoa.macosx.aarch64/.settings/.api_filters
@@ -467,8 +467,8 @@
     <resource path="META-INF/MANIFEST.MF">
         <filter id="926941240">
             <message_arguments>
+                <message_argument value="3.130.0"/>
                 <message_argument value="3.129.0"/>
-                <message_argument value="3.128.0"/>
             </message_arguments>
         </filter>
     </resource>
--- a/binaries/org.eclipse.swt.cocoa.macosx.aarch64/META-INF/MANIFEST.MF
+++ b/binaries/org.eclipse.swt.cocoa.macosx.aarch64/META-INF/MANIFEST.MF
@@ -3,7 +3,7 @@ Fragment-Host: org.eclipse.swt;bundle-version="[3.128.0,4.0.0)"
 Bundle-Name: %fragmentName
 Bundle-Vendor: %providerName
 Bundle-SymbolicName: org.eclipse.swt.cocoa.macosx.aarch64; singleton:=true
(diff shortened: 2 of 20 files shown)


HeikoKlare:
I think it makes sense to immediately bump the minor version (instead of the micro version) since there will be according changes requiring a minor version bump during the release cycle anyway (as we already have PRs in the pipeline requiring it).

akurtakov:
As far as I understand it now there is a cyclic problem unless the build succeeds new natives will not be pushed to git but until this happens I-build is unusable. What is the proposal to break this cycle?

HeikoKlare:
What is the cause for the cyclic dependency? I would have expected that version bump + build script update should enable a successful build of SWT (including natives build and deployment). But the build does not find platform dependencies anymore:
```
Error:  Cannot resolve project dependencies:
Error:    You requested to install 'osgi.bundle; org.eclipse.core.runtime 0.0.0, max=2147483647' but it could not be found
```

Should we just revert https://github.com/eclipse-platform/eclipse.platform.swt/pull/1873 so that SWT is in a consistent state again and then do everything in proper order (first bump version, then update build script)?

akurtakov:
The issue seems to come from parent pom using 4.36 I-builds repo. This used to be done only after there is successful build in given stream so there is content in Ibuild repo. @MohananRahul why was it changed earlier this cycle?

akurtakov:
So this was caused by https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2799 .

akurtakov:
I'll switch the Ibuilds repo to 4.35 . It would be best if creation of new stream repo is filled initially with latest from previous stream. That should make it less fragile and allow more automation to kick in as there will be working version initially.

HeikoKlare:
Sorry, I accidentally pushed again...

akurtakov:
Let's see what happens :)

akurtakov:
@HeikoKlare bumping minor made apitools complain 
```
Caused by: org.apache.maven.plugin.MojoFailureException: There are API errors:
16:07:37  META-INF/MANIFEST.MF:6 The minor version should be the same for version 3.130.0, since no new APIs have been added since version 3.129.0
```
Please either bump micro or add api filters.

HeikoKlare:
Jenkins build succeeds. Should we merge to have the natives built and puhsed, @akurtakov?

akurtakov:
Done.

HeikoKlare:
Thank you! Binaries are pushed: https://github.com/eclipse-platform/eclipse.platform.swt/commit/6f2bd2de36ec5f6ca6165699a349674cf7b6ccd1

PR #1877, opened 2025-03-05
Unify instantiation of image handles in ImageFileNameProvider

ShahzaibIbrahim:
Initializing image handle only using initNative method for both the cases where zoom is equal to the filezoom and when its not equal to the filezoom. The handle in case of not equal is created temporarily and destroyed later before calling init.

### Dependent on:
- [ ] https://github.com/eclipse-platform/eclipse.platform.swt/pull/1898

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
@@ -810,186 +810,6 @@ public static long win32_getHandle (Image image, int zoom) {
 	return image.getImageMetadata(zoom).handle;
 }
 
-ImageHandle initNative(String filename, int zoom) {
-	ImageHandle imageMetadata = null;
-	long handle = 0;
-	int width = -1;
-	int height = -1;
-	device.checkGDIP();
-	boolean gdip = true;
-	/*
-	* Bug in GDI+.  For some reason, Bitmap.LockBits() segment faults
-	* when loading GIF files in 64-bit Windows.  The fix is to not use
-	* GDI+ image loading in this case.
-	*/
-	if (gdip && C.PTR_SIZEOF == 8 && filename.toLowerCase().endsWith(".gif")) gdip = false;
-	/*
-	* Bug in GDI+. Bitmap.LockBits() fails to load GIF files in
-	* Windows 7 when the image has a position offset in the first frame.
-	* The fix is to not use GDI+ image loading in this case.
-	*/
-	if (filename.toLowerCase().endsWith(".gif")) gdip = false;
-	if (gdip) {
-		int length = filename.length();
-		char[] chars = new char[length+1];
-		filename.getChars(0, length, chars, 0);
-		long bitmap = Gdip.Bitmap_new(chars, false);
-		if (bitmap != 0) {
-			int error = SWT.ERROR_NO_HANDLES;
-			int status = Gdip.Image_GetLastStatus(bitmap);
-			if (status == 0) {
-				if (filename.toLowerCase().endsWith(".ico")) {
-					this.type = SWT.ICON;
-					long[] hicon = new long[1];
-					status = Gdip.Bitmap_GetHICON(bitmap, hicon);
-					handle = hicon[0];
-					imageMetadata = new ImageHandle(handle, zoom);
-				} else {
-					this.type = SWT.BITMAP;
-					width = Gdip.Image_GetWidth(bitmap);
-					height = Gdip.Image_GetHeight(bitmap);
-					int pixelFormat = Gdip.Image_GetPixelFormat(bitmap);
-					switch (pixelFormat) {
-						case Gdip.PixelFormat16bppRGB555:
-						case Gdip.PixelFormat16bppRGB565:
-							handle = createDIB(width, height, 16);
-							break;
-						case Gdip.PixelFormat24bppRGB:
-						case Gdip.PixelFormat32bppCMYK:
-							handle = createDIB(width, height, 24);
-							break;
-						case Gdip.PixelFormat32bppRGB:
-						// These will lose either precision or transparency
-						case Gdip.PixelFormat16bppGrayScale:
-						case Gdip.PixelFormat48bppRGB:
-						case Gdip.PixelFormat32bppPARGB:
-						case Gdip.PixelFormat64bppARGB:
-						case Gdip.PixelFormat64bppPARGB:
-							handle = createDIB(width, height, 32);
-							break;
-					}
-					if (handle != 0) {
-						/*
-						* This performs better than getting the bits with Bitmap.LockBits(),
-						* but it cannot be used when there is transparency.
-						*/
-						long hDC = device.internal_new_GC(null);
-						long srcHDC = OS.CreateCompatibleDC(hDC);
-						long oldSrcBitmap = OS.SelectObject(srcHDC, handle);
-						long graphics = Gdip.Graphics_new(srcHDC);
-						if (graphics != 0) {
-							Rect rect = new Rect();
-							rect.Width = width;
-							rect.Height = height;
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2170,6 +2002,200 @@ public int hashCode() {
 	ImageFileNameProviderWrapper createCopy(Image image) {
 		return image.new ImageFileNameProviderWrapper(provider);
 	}
+
+	ImageHandle initNative(String filename, int zoom) {
```
HeikoKlare, line 2006: Please do not move this method around to keep change as small as possible. Are there any changes in this method at all?

HeikoKlare (changes requested):
There is a failing test for image initialization on Windows, so the change does to not seem to be correct as is.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: But this method is only used in the ImageFileNameProviderWrapper now. Doesn't it make sense to encapsulate the method and move it there?

HeikoKlare: I see the point and even the (questionable) point of making access to history for the single code line easier to does not apply here and I just checked and there is no information that is that interesting in it (just "restore from accidental removal", "improve formatting" and HiDPI related changes).

I only see the Image constructors accepting a filename which could make use of it (as in many cases an image can probably be initialized via `initNative` from that file). On the other hand, the method is rarely used and may be deprecated anyway, so not sure whether it really makes sense to use it there.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1185,6 +1005,17 @@ private void destroyHandle () {
 	zoomLevelToImageHandle.clear();
 }
 
+private void destroyHandle(int zoom) {
```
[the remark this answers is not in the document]
ShahzaibIbrahim: As discussed, for now we will change the logic for destroyHandle to be more simpler and later we will try to figure out on how to deal with the temporary handles.

ShahzaibIbrahim:
Tests that are failing will be resolved once https://github.com/eclipse-platform/eclipse.platform.swt/pull/1898 is merged. So it makes this PR dependent.

HeikoKlare (changes requested):
This PR now basically reverts the commit made to Image recently pushed on master. Please adapt to current behavior of the class first.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2138,16 +1958,22 @@ ImageData getImageData(int zoom) {
 	@Override
 	ImageHandle getImageMetadata(int zoom) {
 		ElementAtZoom<String> fileForZoom = DPIUtil.validateAndGetImagePathAtZoom (provider, zoom);
-		ImageHandle nativeInitializedImage = null;
-		if (fileForZoom.zoom() == zoom) {
-			nativeInitializedImage = initNative(fileForZoom.element(), zoom);
-		}
-		if (nativeInitializedImage == null) {
+		ImageHandle nativeInitializedImage = initNative(fileForZoom.element(), zoom);
```
akoch-yatta, line 1961: that must be fileForZoom.zoom now instead of zoom

akoch-yatta (changes requested):

HeikoKlare (commented):
In general, the change looks good. I only have a remark regarding adaptation of gray/disabled images.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2130,27 +1956,39 @@ protected Rectangle getBounds(int zoom) {
...
-			if (imageDataAtZoom.zoom() != zoom) {
-				imageData = DPIUtil.scaleImageData(device, imageDataAtZoom, zoom);
-			}
+			imageData = scaleIfNecessary(imageDataAtZoom, zoom);
 			imageData = adaptImageDataIfDisabledOrGray(imageData);
-			init(imageData, zoom);
+		} else {
+			ElementAtZoom<ImageData> imageDataAtZoom = new ElementAtZoom<>(nativeInitializedImage.getImageData(), fileForZoom.zoom());
+			imageData = scaleIfNecessary(imageDataAtZoom, zoom);
+			destroyHandleForZoom(zoom);
```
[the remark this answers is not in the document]
ShahzaibIbrahim: No. In this case, it is called inside initNative method.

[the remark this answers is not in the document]
akoch-yatta: That is already done in initNative. We could take it out there and unify the caller. It is the only place where initNative is called anyhow.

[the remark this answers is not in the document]
HeikoKlare: But it's not done for every type of image, is it? Not sure whether that's relevant, but in the copy constructor the gray/disabled application is always done after `getImageData()`, thus even if the images was loaded via `initNative()`, so that seems a bit inconsistent. I would be in favor of just doing that at one single place (if that's possible), so that it does not matter whether `initNative()` or the `ImageLoader` was used. What do you think?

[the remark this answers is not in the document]
akoch-yatta: Yes, probably right. I adapted that place

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2213,27 +2039,38 @@ protected Rectangle getBounds(int zoom) {
...
-			ImageData imageData = imageDataAtZoom.element();
-			if (imageDataAtZoom.zoom() != zoom) {
-				imageData = DPIUtil.scaleImageData(device, imageDataAtZoom, zoom);
-			}
-			imageData = adaptImageDataIfDisabledOrGray(imageData);
-			init(imageData, zoom);
+			imageDataAtZoom = ImageDataLoader.load(fileForZoom.element(), fileForZoom.zoom(), zoom);
+		} else {
+			imageDataAtZoom = new ElementAtZoom<>(nativeInitializedImage.getImageData(), fileForZoom.zoom());
+			destroyHandleForZoom(zoom);
```
HeikoKlare, line 2055: Is it correct to destroy the handle if it was retireved from the `zoomLevelToImageHandle` map? I guess we should only do that if it was not in that map before (i.e., if it was temporarily created for retrieving the image data), shouldn't we?

akoch-yatta: That is confusing, yes, but logically okay.  zoom != fileForZoom.zoom(). It is again a place, where we will need to do a second round, when we refactored all the places to be able to better create temporary handles.

HeikoKlare: Ah, I see, it is because the calling method `Image#getImageData(int)` ensures that. I didn't see that. Yes, that's definitely confusing and we should clean up in a second round.

HeikoKlare (approved):

PR #1879, opened 2025-03-06
Rewrite the algorithm that sets smallIcon/largeIcon in Decorations class

fedejeanne:
The output is the same but the new algorithm doesn't sort and select the 1st one, it directly searches for the best one, reducing the complexity from **O(n^2)** (bubble sort) to **O(n)** (linear search)


fedejeanne:
Failing test is unrelated: #1843

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Decorations.java
```diff
@@ -949,6 +921,53 @@ void setImages (Image image, Image [] images) {
...
+	int diffWidth1 = Math.abs (data1.width - width);
+	int diffWidth2 = Math.abs (data2.width - width);
+
+	if (diffWidth1 == diffWidth2) {
+		int transparent1 = data1.getTransparencyType ();
+		int transparent2 = data2.getTransparencyType ();
+		if (transparent1 == transparent2) {
+			if (data1.depth == data2.depth) return false;
+			// TODO: analyze this, it seems broken.
+			return data1.depth > data2.depth && data1.depth <= depth;
```
fedejeanne, line 958: On a closer look this seems OK since it takes the highest pixel depth that does **not** surpass the wanted `depth`

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Decorations.java
```diff
@@ -949,6 +921,53 @@ void setImages (Image image, Image [] images) {
...
+			closestIndex = i;
+			closestData = imageData[i];
+		}
+	}
+
+	return closestIndex;
+}
+
+/**
+ * @return <code>true</code> if <code>data1</code> is closer to the desired width, height and depth than <code>data2</code>.
```
HeikoKlare, line 946: This contract does not seem to be correct: `height` is not used at all.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Decorations.java
```diff
@@ -949,6 +921,53 @@ void setImages (Image image, Image [] images) {
...
+ */
+private boolean isCloserThan (ImageData data1, ImageData data2, int width, int height, int depth) {
```
HeikoKlare, line 948: Would be easier to comprehend if the data parameters had more comprehensible names, such as `referenceData` and `dataToTest` so that the method name `isCloserThan` has a meaning regarding the parameters.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Decorations.java
```diff
@@ -949,6 +921,53 @@ void setImages (Image image, Image [] images) {
...
+		datas[i] = images[i].getImageData();
+	}
+	return datas;
+}
+
+private int findIndexOfClosest(ImageData[] imageData, int width, int height, int depth) {
+	int closestIndex = 0;
+	ImageData closestData = imageData[0];
+	for (int i=1; i<imageData.length;i++) {
+		if (isCloserThan(imageData[i],closestData,  width, height, depth)) {
```
HeikoKlare, line 936: From my understanding, the parameter order is wrong, isn't it? `closestData` should be preferred over `imageData[i]`, but currently it's the other way round.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Decorations.java
```diff
@@ -949,6 +921,53 @@ void setImages (Image image, Image [] images) {
...
 
+private ImageData[] getImageData(Image[] images) {
+	ImageData[] datas = new ImageData[images.length];
+	for (int i = 0; i < images.length; i++) {
+		datas[i] = images[i].getImageData();
+	}
+	return datas;
+}
+
+private int findIndexOfClosest(ImageData[] imageData, int width, int height, int depth) {
```
HeikoKlare, line 932: Can't this whole method be replaced by a usual `Arrays.sort` passing the `isCloserThan` implementation as a comparator? To avoid all the index calculations assuming that different arrays are sorted in the same way, you could also wrap an image with its image data into a record, so that you can execute the sort on the image data of the record while then retrieving the image stored in the first element after the sort operation.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Decorations.java
```diff
@@ -949,6 +921,53 @@ void setImages (Image image, Image [] images) {
...
+}
+
+/**
+ * @return <code>true</code> if <code>data1</code> is closer to the desired width, height and depth than <code>data2</code>.
+ */
+private boolean isCloserThan (ImageData data1, ImageData data2, int width, int height, int depth) {
+	int diffWidth1 = Math.abs (data1.width - width);
+	int diffWidth2 = Math.abs (data2.width - width);
+
+	if (diffWidth1 == diffWidth2) {
```
HeikoKlare, line 952: Replace with early return?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Decorations.java
```diff
@@ -893,22 +873,14 @@ void setImages (Image image, Image [] images) {
...
-			}
-			smallIcon = images [0];
-			if (images.length > 1) {
-				sort (images, datas, getSystemMetrics (OS.SM_CXICON), getSystemMetrics (OS.SM_CYICON), depth);
-			}
-			largeIcon = images [0];
+
+			ImageData[] imageData = getImageData(images);
+
+			int smallIconIndex = findIndexOfClosest(imageData, getSystemMetrics (OS.SM_CXSMICON),getSystemMetrics (OS.SM_CYSMICON), depth);
```
HeikoKlare, line 879: Assigning the `getSystemMetrics(...)` calls to some local variables with comprehensive names would ease to understand why comparing against 100% image data might be reasonable.
E.g., having `int smallIconWidthAt100 = getSystemMetrics(OS.SM_CXSMICON)` would indicate that using `image.getImageData(100)` makes sense to compare against. Currently, one would still need to gather all the understainding (meaning of the system metrics value, meaning of `getImageData()` without passing a zoom and matching that) again.

HeikoKlare (commented):
Simplifying this functionality and improve comprehensibility is a good thing.
I left some proposal for improving the code and making the logic comprehensible. Most essential for me would be to assign the compares values that are hard to understand (system metrics, image data at specific zooms) to comprehensive, self-explaining variables, as that's exactly the part that confused me all the time and made it difficult to understand what happens here.

Apart from that, formatting does currently seems to be broken at several places (in particular, missing whitespaces).

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Decorations.java
fedejeanne: `closestData` contains the image data that is _currently_ closer and if the `if` turns out to be true then `closestData` needs to be updated. After the `for`-loop is done, `closestData` may have been replaced a few times and it would contain the final data that is closest to the target.

If `isCloserThan(a, b, ...)` returns `true` then `a` is closer than `b` (_i.e._ it's a better choice). As a rule of thumb I code such methods in a way that one could read them as `a.isCloserThan(b,...)` (the order of the parameters is important).

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Decorations.java
fedejeanne: Done. I also changed the name of the private method `getImageData(...)` to `getImageDataAt100(...)` and I explicitly passed the parameter `100` when calling `images[i].getImageData(100)` to make it more explicit that I am interested in the data at 100% zoom

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Decorations.java
fedejeanne: Done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Decorations.java
fedejeanne: Thank you for the suggestion, I changed the names of all 4 remaining parameters (I removed `height`) and **removed the JavaDoc** too:

```java
private boolean isCloserThan (ImageData dataToTest, ImageData referenceData, int targetWidth, int targetDepth) {
```

I hope the method is now self-explanatory enough and that the JavaDoc is not necessary anymore.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Decorations.java
fedejeanne: Removed

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Decorations.java
fedejeanne: Removed `TODO`

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Decorations.java
HeikoKlare: I see. It's not up to wrong parameter order but seems to be an inconsistency inside the `isCloserThan()` implementation. See, e.g., thes lines: 
```java
if (transparent1 == SWT.TRANSPARENCY_ALPHA) return true;
if (transparent2 == SWT.TRANSPARENCY_ALPHA) return false;
```
It will return the tested data in case it has alpha transparency, even though the reference data may also have that, thus preferring tested over reference data even is is not "closer". So maybe some of the checks should be resorted.
It could become clearer when renaming the `...1` and `...2` variables inside `isCloserThan()` according to being reference or tested data, because then you can easily grasp that tested data is preferred over reference data during the transparency type checks.

fedejeanne: > It will return the tested data in case it has alpha transparency, even though the reference data may also have that...

If they both have the same transparency then the algorithm decides based on pixel depth, see the `if (transparencyToTest == referenceTransparency)` in 74f094a38822ee9c70b12e064ee4b05532639e34.

I just renamed the variables and added some comments in 74f094a38822ee9c70b12e064ee4b05532639e34 in the hope that the algorithm is a bit more understandable.

fedejeanne: Another possible version could be:
```java
private static boolean isCloserThan (ImageData dataToTest, ImageData referenceData, int targetWidth, int targetDepth) {
	int diffWidthToTest = Math.abs (dataToTest.width - targetWidth);
	int diffReferenceWidth = Math.abs (referenceData.width - targetWidth);

	// The closer the width the better
	if (diffWidthToTest != diffReferenceWidth) return diffWidthToTest < diffReferenceWidth;

	int transparencyToTest = dataToTest.getTransparencyType ();
	int referenceTransparency = referenceData.getTransparencyType ();

	// If they have different transparency, favor (in this order): the one with
	// an alpha channel, the one with a mask, the one with a transparency pixel
	if (transparencyToTest != referenceTransparency) {
		if (transparencyToTest == SWT.TRANSPARENCY_ALPHA) return true;
		if (referenceTransparency == SWT.TRANSPARENCY_ALPHA) return false;
		if (transparencyToTest == SWT.TRANSPARENCY_MASK) return true;
		if (referenceTransparency == SWT.TRANSPARENCY_MASK) return false;
		if (transparencyToTest == SWT.TRANSPARENCY_PIXEL) return true;
		if (referenceTransparency == SWT.TRANSPARENCY_PIXEL) return false;
		return false;
	}
	
	// If they have the same transparency then the bigger the pixel depth (without
	// surpassing the target depth) the better
	if (dataToTest.depth == referenceData.depth) return false;
	return dataToTest.depth > referenceData.depth && dataToTest.depth <= targetDepth;
}
```

This one makes it more explicit that the transparency is the 2nd criteria by using `if (transparencyToTest != referenceTransparency)`, which is more aligned with the other early return (https://github.com/eclipse-platform/eclipse.platform.swt/pull/1879#discussion_r1988721511).

HeikoKlare (commented):
Definitely much cleaner and easier to understand now 👍 
Once formatting is correct, I think this is good to go.

fedejeanne:
Perfect! 0892529fdb3ca1e04f698fa8683c05adab9b7787 only changes the formatting of the affected code. I used `Ctrl + Shift + F`.

PR #1880, opened 2025-03-06
Remove obsolete hashCode/equals for Region

akoch-yatta:
This PR removes the obsolete equals and hashCode methods for Region. For the upcoming changes for win32 the current implementation cannot be used anymore because all handles will be created on demand. As the current implementation is equal in their behavior to the base behavior for these methods inherited from Object they will be removed.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Region.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Region.java
@@ -436,9 +436,7 @@ void destroy() {
  */
 @Override
 public boolean equals(Object object) {
-	if (this == object) return true;
-	if (!(object instanceof Region region)) return false;
-	return handle == region.handle;
+	return super.equals(object);
 }
 
 /**
@@ -511,7 +509,7 @@ long regionToRects(long message, long rgn, long r, long path) {
  */
 @Override
 public int hashCode() {
-	return (int)handle;
+	return super.hashCode();
 }
 
 /**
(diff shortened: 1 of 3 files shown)


HeikoKlare:
@laeubi we have question regarding API tooling, which you may know best about 🙂 

This PR removes `hashCode`/`equals` implementations as the actual implementation are completely useless. When simply removing them, the build fails because of API errors:
```
Caused by: org.eclipse.tycho.core.exceptions.VersionBumpRequiredException: There are API errors:
[2196](https://github.com/eclipse-platform/eclipse.platform.swt/actions/runs/13703850988/job/38324210490?pr=1880#step:7:2197)
The method org.eclipse.swt.graphics.Region.equals(Object) has been removed
[2197](https://github.com/eclipse-platform/eclipse.platform.swt/actions/runs/13703850988/job/38324210490?pr=1880#step:7:2198)
The method org.eclipse.swt.graphics.Region.hashCode() has been removed
[2198](https://github.com/eclipse-platform/eclipse.platform.swt/actions/runs/13703850988/job/38324210490?pr=1880#step:7:2199)
META-INF/MANIFEST.MF:1 The major version should be incremented in version 3.130.0, since API breakage occurred since version 3.129.0
```

When adding API filters (since there is no breaking API change as the methods are provided via Object already), the API tools validation complains about unused filter:
```
NORMAL:
The API problem filter for: 'The method org.eclipse.swt.graphics.Region.hashCode() has been removed' is no longer used
```
The complain about the unused filter also shows up in the IDE.

I guess this is some cornercase as implementations of methods already provided via `Object` are removed, such that they are of course still part of public API but not present in any explicit class of the inheritance hierarchy.

Do you know something about this situation or what we could do about it (either here at consumer side or in API tools itself)?

laeubi:
@HeikoKlare I don't know why API tool exactly complains, but of course equals/hashcode can be part of the API contract (e.g. as in Set/List/Map/ ...).

I probably would here just keep the methods and call `super`...

HeikoKlare:
Thank you for the feedback! Yeah, then keeping the methods and just calling `super` is probably the best option (@akoch-yatta fyi)

PR #1883, opened 2025-03-06
[MacOS] Fix getBounds for Image based on ImageGcDrawer

HeikoKlare:
When running with a device zoom of 200, the bounds of images created with an ImageGcDrawer have double the expected size. The reason is that the image is initialized with image data at zoom 200, from which also the width and height are retrieved. The used initialization logic does, however, always expect image data at zoom 100 to be passed.

This change fixes the initialization of images based on ImageGcDrawer to create both the 100 and 200 image data version and correctly initialize the image metadata with the former. The according test case is extended to validate the bounds in pixels of images not only against the bounds if the image but also against the expected/initial bounds.

This API was introduced with the previous release and is yet only used for minimize/maximize icons of part stacks and the the line number ruler of editors. They do not use the bounds returned by these images, which is why the bug did not lead to any visible derivation of expected behavior yet. Once such images are used in a context whether the returned bounds matters, the issue will be visible by images having the double the expected size if used on a monitor with a zoom of 200.


akoch-yatta (approved):
Without being able to test, because I don't have a Mac at hand, I can only have a look at the code change. This looks fine to me.

PR #1886, opened 2025-03-07
[Win32] Combine duplicate initialization logic for image from providers

HeikoKlare:
Images based on ImageFileNameProvider and ImageDataProvider are initialized with data based on the default zoom and can later on be reinitialized with data for others zooms when `getImageData()` is called. At both places, the same functionality is implemented twice.

This change makes the initial creation of image data use the already existing implementation for reinitializing it for another zoom. It also unifies the previous two implementations and removes obsolete calls to init() which are just required once for starting resource tracking.

Required to keep changes in https://github.com/eclipse-platform/eclipse.platform.swt/pull/1828 simple.


akoch-yatta (approved):
Changes look solid to me. Unifying the call from the constructor with almost equal one for rescaling makes sense.

HeikoKlare:
Failing test is unrelated and already documented: https://github.com/eclipse-platform/eclipse.platform.swt/issues/1843

PR #1887, opened 2025-03-07
[Win32] Replace autoScaleUp of ImageData with proper scaleImageData

HeikoKlare:
In some image constructors, image data are currently auto-scaled. This works by accident, as the device zoom used for auto scaling conforms to the zoom of the image to be initialized.

However, to make this explicit, with this change the proper zoom is passed to scale method. It also makes the autoScaleUp method in DPIUtil obsolete and removes it.

This supcedes
- #1867

which migrated to a different but likely inaccurate autoScale operation.

Required to keep changes in https://github.com/eclipse-platform/eclipse.platform.swt/pull/1828 simple.


HeikoKlare:
@akoch-yatta can you please also have a look at this ones? Just a simple enhacement.

akoch-yatta (approved):
Changes look good to me

PR #1890, opened 2025-03-10
Invert parameters of all call to AssertEquals in Spinner tests

fedejeanne:
They were being used wrong. Also, add the index of the array in one of the tests for better debugging.


fedejeanne:
Failing API checks are unrelated since I didn't modify those classes

PR #1894, opened 2025-03-10
Fix spelling of GitHub brand name

jbampton:

PR #1895, opened 2025-03-11
Version bump for new stream

akurtakov:
I missed that in
https://github.com/eclipse-platform/eclipse.platform.swt/pull/1894

### DIFF
--- a/bundles/org.eclipse.swt.tools/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.swt.tools/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-Name: %pluginName
 Bundle-Vendor: %providerName
 Bundle-SymbolicName: org.eclipse.swt.tools; singleton:=true
-Bundle-Version: 3.110.600.qualifier
+Bundle-Version: 3.110.700.qualifier
 Bundle-ManifestVersion: 2
 Export-Package: org.eclipse.swt.tools.internal; x-internal:=true
 Bundle-ActivationPolicy: lazy
--- a/features/org.eclipse.swt.tools.feature/feature.xml
+++ b/features/org.eclipse.swt.tools.feature/feature.xml
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.swt.tools.feature"
       label="%featureName"
-      version="3.109.600.qualifier"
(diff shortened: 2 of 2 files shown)


akurtakov:
Warnings are coming from some previous change thus merging to move a step ahead at least.

HeikoKlare:
> Warnings are coming from some previous change

Just to understand this correctly: it's just because we have some flaky behavior of the analysis in SWT so that sometimes the used reference `master` build has 33 warnings and sometimes it has 43, don't we? Or was there actually a previous change that needs to be revised?

akurtakov:
@HeikoKlare Comparison is done against latest master build (https://github.com/eclipse-platform/eclipse.platform.swt/blob/632681faa3b100ee2872495bb0efe7559df0163c/Jenkinsfile#L349) so when a PR that increases issues is pushed and built in master the new "baseline" becomes with higher number of warnings and thus these "new" issues will no longer be reported as new.
 Overtime this leads to gradual increase of warnings and decrease in code quality, unfortunately we had to go that path as using a baseline that marks any increase of warnings as failing builds led to people ignoring results even more. Long story short - there seems to be no way to make people care about that :( .

HeikoKlare:
@akurtakov Yes, I know that latest `master` build is always taken as baseline, so if you merge something with new warnings, that will implicitly become the new baseline without further notice.

Still, I think we have a different "issue" here (related to the tooling) and there were no warnings introduced in the recent past: the reported warnings for SWT `master` builds are flaky. Sometimes there are 33, sometimes there are 43. See for example the latest master build (https://ci.eclipse.org/releng/job/eclipse.platform.swt/job/master/991/):
image
And the previous one (https://ci.eclipse.org/releng/job/eclipse.platform.swt/job/master/989/):
image

The latest build used for reference will always have 33 issues reported, as builds with more issues (like the most recent master build with 43 issues) will be marked as instable and not taken as reference. So if your PR is built unluckily built with 43 warnings (even though you did not introduce new ones and no recent PR did either), you will see the check failing.

akurtakov:
I have no explanation for such behavior. It would need separate tracking and investigation.

HeikoKlare:
The behavior has been there for quite a while, but it does not occur that often and did bother us that much that anyone cared. I justed wanted to make aware that such reported warnings may just be a CI glitch and do not necessarily point to an actual quality degradation.

PR #1897, opened 2025-03-11
docs: edited icons for workflows

braveocheretovych:


merks (approved):
With these changes it looks like this:

image

That seems correct to me.  Can someone else confirm?

laeubi (commented):
This is not correct. It now only shows that the job succeeds (what hopefully is always the case) but no longer shows the (failed) tests.

grafik

HeikoKlare:
I agree that the tests badge should be there. It's explicitly generated and deployed by the workflow publishing test results:
https://github.com/eclipse-platform/eclipse.platform.swt/blob/7825d59d1b70efafa0dfb4bcf4eb6213c1d99e6f/.github/workflows/junit.yml#L73-L79

braveocheretovych:
I think that would be correct.

braveocheretovych:
Hey. You accept it

HeikoKlare:
As mentioned before (https://github.com/eclipse-platform/eclipse.platform.swt/pull/1897#issuecomment-2715290103), in my opinion the tests badge should not be removed: (https://gist.githubusercontent.com/eclipse-releng-bot/78d110a601baa4ef777ccb472f584038/raw/71510599eb84e852f3e135aa7a3ddf33854ca716/badge.svg

braveocheretovych:
> As mentioned before ([#1897 (comment)](https://github.com/eclipse-platform/eclipse.platform.swt/pull/1897#issuecomment-2715290103)), in my opinion the tests badge should not be removed: (https://gist.githubusercontent.com/eclipse-releng-bot/78d110a601baa4ef777ccb472f584038/raw/71510599eb84e852f3e135aa7a3ddf33854ca716/badge.svg

I fixed everything, check it again.

HeikoKlare:
This is what it would currently look like:
image

I agree with @laeubi that the freeze period check does not make sense here. Reason is that the badges show results from latest master build and in freeze period we will usually not merge into master, so this badge will always show "passed" unless someone merged within freeze period. And then it would even report "failed" until the first PR is merged after freeze period.

So only change that would remain with this PR is the addition of the badge for publishing test results. I think that one makes sense to add, as it indicates whether the badge containing the actual tests results may be out-of-date (as that one will not be updated in case the test results publish failed).

If not one objects on adding that one, @braveocheretovych could you please remove the freeze period badge and squash your commits so that we may merge this?

braveocheretovych:
So I edited what you asked for, the only thing I don't know is how to compress it all into one commit. Maybe I should close this one and create a new one, what do you think?

HeikoKlare:
> I don't know is how to compress it all into one commit. Maybe I should close this one and create a new one, what do you think?

You don't need to create a new PR. In Git, you can squash commits via interactive rebase (or some GUI tools also have a dedicated "squash" functionality). You can find several tutorial on how to do it in the web, e.g., https://www.datacamp.com/tutorial/git-squash-commits
In your case, when using a command line tool, you would need to do a `git rebase -i HEAD~4` and mark the three latest commits as `squash` instead of `pick` so that they are squash into a single one.

braveocheretovych:
I'm sorry to admit it, but I can't do this because I'm making the change through the edit tab on Github, I don't use the terminal to work.

HeikoKlare:
Alright, I've squashed the changes for you.

braveocheretovych:
Thankyou @HeikoKlare

braveocheretovych:
Do I need to take any further action?

HeikoKlare:
No, you don't need to take any further action. I just wanted to keep this open for a short while to give the chance to raise objections (see https://github.com/eclipse-platform/eclipse.platform.swt/pull/1897#issuecomment-2764496686).

braveocheretovych:
I understand, thanks for the clarification.

jonahgraham:
> I can also add an old shield if that's important?

The readme.md (before this change, but preserved with this PR) has it hardcoded to a specific old (3 year old) version. That 3 year old badge was from a day the tests failed, so the readme has shown failing badge for 3 years. See PR #2778 for an update.

PR #1898, opened 2025-03-11
Create handles for ImageDataProvider and ImageFilenameProvider on demand

ShahzaibIbrahim:
Currently, when an image is instantiated using `ImageDataProvider `or `ImageFileNameProvider`, a handle is created immediately. We aim to optimize this by deferring handle creation until it is explicitly needed. To achieve this, `getBounds()` will return the bounds of a 100% zoomed image when called after instantiation, without requiring a handle to be created upfront. This change improves efficiency by reducing unnecessary resource allocation during image initialization.

Attaching the snippet to test the use cases after this change:

```java

import org.eclipse.swt.*;

public class SnippetTestImage {

	final static String IMAGES_ROOT = "bin/org/eclipse/swt/snippets/";

		imageFileNameProviderExample();
		imageDataProviderExample();
	}

	private static void imageDataProviderExample() {
		Image image;
		ImageDataProvider imageDataProvider = zoom -> {
			String fileName;
			switch (zoom) {
			case 100:
				fileName = "eclipse16.png";
				break;
			case 150:
				fileName = "eclipse24.png";
				break;
			case 200:
			case 250:
				fileName = "eclipse32.png";
				break;
			default:
				return null;
			}
			return new ImageData(IMAGES_ROOT + fileName);
		};
		try {
			image = new Image(display, imageDataProvider);
			System.out.println("Image Bounds initially = " + image.getBounds());
			ImageData imgData = image.getImageData(250);
			System.out.println("Image Bounds on Demand (250) = " + imgData.height + " width = " + imgData.width);
			image.dispose();
		} catch (SWTException e) {
			System.out.println(e);
		}
	}

	private static void imageFileNameProviderExample() {
		Image image;
		ImageFileNameProvider provider = zoom -> {
			String fileName;
			switch (zoom) {
			case 100:
				fileName = "eclipse16.png";
				break;
			case 150:
				fileName = "eclipse24.png";
				break;
			case 200:
			case 250:
				fileName = "eclipse32.png";
				break;
			default:
				return null;
			}
			return IMAGES_ROOT + fileName;
		};
		try {
			image = new Image(display, provider);
			System.out.println("Image Bounds initially = " + image.getBounds());
			ImageData imgData = image.getImageData(250);
			System.out.println("Image Bounds on Demand (250) = " + imgData.height + " width = " + imgData.width);
			image.dispose();
		} catch (SWTException e) {
			System.out.println(e);
		}
	}
}

```

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
@@ -91,6 +91,11 @@ public final class Image extends Resource implements Drawable {
 	public int type;
 
 	/**
+	 * this field make sure the image is initialized without any errors
+	 */
+	boolean isInitialized = false;
+
+	/**
 	 * specifies the transparent pixel
 	 */
 	int transparentPixel = -1, transparentColor = -1;
@@ -552,7 +557,10 @@ public Image(Device device, ImageFileNameProvider imageFileNameProvider) {
 	super(device);
 	this.imageProvider = new ImageFileNameProviderWrapper(imageFileNameProvider);
 	initialNativeZoom = DPIUtil.getNativeDeviceZoom();
-	imageProvider.getImageMetadata(getZoom());
+	if (imageProvider.getImageData(100) == null) {
+		SWT.error(SWT.ERROR_INVALID_ARGUMENT, null,
+				": ImageFileNameProvider [" + imageFileNameProvider + "] returns null ImageData at 100% zoom.");
+	}
 	init();
 	this.device.registerResourceWithZoomSupport(this);
 }
@@ -590,7 +598,10 @@ public Image(Device device, ImageDataProvider imageDataProvider) {
 	super(device);
 	this.imageProvider = new ImageDataProviderWrapper(imageDataProvider);
 	initialNativeZoom = DPIUtil.getNativeDeviceZoom();
-	imageProvider.getImageMetadata(getZoom());
+	if (imageDataProvider.getImageData(100) == null) {
+		SWT.error(SWT.ERROR_INVALID_ARGUMENT, null,
+				": ImageDataProvider [" + imageDataProvider + "] returns null ImageData at 100% zoom.");
+	}
 	init();
 	this.device.registerResourceWithZoomSupport(this);
 }
@@ -640,6 +651,12 @@ private ImageData adaptImageDataIfDisabledOrGray(ImageData data) {
 	}
 
 	return returnImageData;
+}
+
+@Override
+void init() {
+	super.init();
+	this.isInitialized = true;
 }
 
 private ImageData applyDisableImageData(ImageData data, int height, int width) {
@@ -1891,18 +1908,23 @@ public static Image win32_new(Device device, int type, long handle, int nativeZo
 }
 
 private abstract class AbstractImageProviderWrapper {
+	private boolean isDestroyed;
+
 	protected abstract Rectangle getBounds(int zoom);
 	abstract ImageData getImageData(int zoom);
 	abstract ImageHandle getImageMetadata(int zoom);
 	abstract AbstractImageProviderWrapper createCopy(Image image);
-	abstract boolean isDisposed();
+
+	protected boolean isDisposed() {
+		return !isInitialized || isDestroyed;
+	}
 
 	protected void destroy() {
+		this.isDestroyed = true;
 	}
 }
 
 private class PlainImageProviderWrapper extends AbstractImageProviderWrapper {
-	private boolean isDestroyed;
 	private final int width;
 	private final int height;
 
@@ -1984,16 +2006,6 @@ private class PlainImageProviderWrapper extends AbstractImageProviderWrapper {
 	AbstractImageProviderWrapper createCopy(Image image) {
 		return image.new PlainImageProviderWrapper(width, height);
 	}
-
-	@Override
-	protected void destroy() {
-		this.isDestroyed = true;
-	}
-
-	@Override
-	boolean isDisposed() {
-		return isDestroyed;
-	}
 }
 
 private abstract class DynamicImageProviderWrapper extends AbstractImageProviderWrapper {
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2170,6 +2180,11 @@ public int hashCode() {
 	ImageFileNameProviderWrapper createCopy(Image image) {
 		return image.new ImageFileNameProviderWrapper(provider);
 	}
+
+	@Override
+	boolean isDisposed() {
+		return false;
+	}
```
ShahzaibIbrahim, line 2187: Previously dispose condition was to check if there are no handles present. Since we won't create any handles on instantiation, that logic is obsolete. This is why returning just false unless someone has better idea what to do here. (Same for ImageDataProvider)

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2080,6 +2078,17 @@ public static Image win32_new(Device device, int type, long handle, int nativeZo
 	return image;
 }
 
+private void destroyHandle(int zoom) {
```
akoch-yatta, line 2081: I would like to prevent such logic heavy cleanup method. Our initial idea to add each ImageHandle to zoomLevelToImageHandle in its constructor doesn't really make sense anymore as we start creating temporary handles like in this PR. I would like a propose to create an additional PR first to extract this logic of adding it to the Map out of ImageHandle. Then you could just call `destroyHandle(handleAt100.handle);` below and you are done.

ShahzaibIbrahim: If I take out the logic to add the handle into the map. I will still need to add the handle into the map on the caller side. Would that not create an issue as we always depends on caller to add it to the map (or not)

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: I would like to agree. But the contract (at least implicitly) for `isDisposed()` is, that it must return true, if` dispose()` on that resource was called. I am actually surprised no tests fail because of that.
In other resources we introduced a flag named `isDestroyed` to track that, e.g. have a look into #1884 . For this PR I propose to introduce this flag in both of the affected wrappers. When adapted all places to create handles dynamically we can move that flag up into the Image class itself.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2115,17 +2120,24 @@ private class ImageFileNameProviderWrapper extends AbstractImageProviderWrapper
 	 * ImageFileNameProvider to provide file names at various Zoom levels
 	 */
 	private final ImageFileNameProvider provider;
+	private boolean isDestroyed;
+	private int width;
+	private int height;
```
fedejeanne, line 2125:
```suggestion
	private final int width;
	private final int height;
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2181,17 +2198,24 @@ private class ImageDataProviderWrapper extends AbstractImageProviderWrapper {
 	 * ImageDataProvider to provide ImageData at various Zoom levels
 	 */
 	private final ImageDataProvider provider;
+	private int width;
+	private int height;
```
fedejeanne, line 2202:
```suggestion
	private final int width;
	private final int height;
```

fedejeanne (commented):
Just a NIT: `width` and `height` could be `final` (also in `ImageGcDrawerWrapper`)

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
HeikoKlare, line 2125: Wouldn't it make sense to store these values in the AbstractImageProviderWrapper, now that all it's subclasses have such properties? You could also move the concrete implementations for `isDisposed()` and `destroy()` into the superclass then.

akoch-yatta: When testing the branch I noticed a weird behavior. When smooth scaling is involved we can run into exceptions with this changes and I am thinking, what the proper expectation here is. So, in smooth scaling we are doing this:
`Image original = new Image (device, (ImageDataProvider) zoom -> imageData);`
This returns for each zoom the same ImageData. As we are using the imageData at zoom 100 to define with and height, we have a problem now, because the bounds of the Image at 150% will no longer match the ImageData size at 150%, because getBounds scales the bounds up.
For this concrete case you could argue the ImageDataProvider implementation should be adapted, but in general I think we need to go the less efficient route and not store width and height, but always calculate them on demand with a proper handle.

HeikoKlare: I understand and agree that we probably have to calculate height/width for the according handle.
Even if we could argue and reasonably assume that image data is scaled according to the zoom, there might different ways of rounding the sizes which can then lead to off-by-one errors in the sizes. Thus the bounds probably need to be calculated from the actual image handle / image data to be safe here.

fedejeanne: So you mean:
- Keep `width` and `height` "empty" (with -1) in the constructor
- Initialize for `BASE_ZOOM` when first calling `getBounds(zoom)`, which creates and disposes a handle for the zoom
- Keep the values after that

?

akoch-yatta: No, remove width and height completely, because the width and height for each zoom must not necessarily correlate with the zoom, e.g. width at 100% mustn't be doubled with the width at 200%. Therefor get bounds must always calculate width and height on demand. We could optimize by caching width and height for each zoom as well. But I don't know if that is really necessary

fedejeanne: Ah, I see, thank you for the clarification. In that case, I think it's worth maintaining a cache to avoid constantly creating and disposing handles. These 2 methods alone have lots of usages in the code so I expect `getBounds(zoom)` to be called quite often:

image

We could even remove entries from the cache inside `destroyHandlesExcept` (probably need to rename that method too) since that is the moment that zooms become "invalid".

akoch-yatta (approved):
Looks good to me now

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2123,9 +2135,9 @@ private class ImageFileNameProviderWrapper extends AbstractImageProviderWrapper
 
 	@Override
 	protected Rectangle getBounds(int zoom) {
-		ImageHandle imageHandle = zoomLevelToImageHandle.values().iterator().next();
-		Rectangle rectangle = new Rectangle(0, 0, imageHandle.width, imageHandle.height);
-		return DPIUtil.scaleBounds(rectangle, zoom, imageHandle.zoom);
+		ImageData imageData = getImageData(zoom);
```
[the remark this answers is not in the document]
HeikoKlare: I agree that we should not store the handle.
To better understand the whole context: what would happen if we threw away the initNative operation and simply load and cache the image data upon first request of any data for the image at that zoom, such that the image data can then be used for the bounds, for creating the handle etc.?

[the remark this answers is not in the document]
akoch-yatta: You mean completly remove initNative? As initNative is used first, I'm not sure, what effects that would have. E.g. with the transparent tiff images we have the issue, that initNative works, but the tiff image loader doesn't work. I am sure, that there are many more cases like that

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2073,20 +2091,15 @@ ImageHandle getImageMetadata(int zoom) {
...
+	ImageFileNameProviderWrapper createCopy(Image image) {
+		return image.new ImageFileNameProviderWrapper(provider);
 	}
 
 	@Override
 	public int hashCode() {
 		return Objects.hash(provider, styleFlag, transparentPixel, getZoom());
 	}
-
+
```
fedejeanne, line 2102:
```suggestion
```

fedejeanne (approved):
LGTM

fedejeanne:
The code looks fine but some tests fail because of leaked resources. I am retriggering tests. @ShahzaibIbrahim stay tuned

ShahzaibIbrahim:
@fedejeanne resource leak is fixed.

fedejeanne (approved):

PR #1901, opened 2025-03-11
[win32] Explicit zoom handle management

akoch-yatta:
This commit adapt the win32 implementation of Image to manage the map contains all available OS handle more explicitly. As of now the ImageHandle constructor adds all new handles directly to the map. To better support temporary image handles, this implicit adding is replaced by an explicit adding at convenience method to create new ImageHandles.


akoch-yatta:
Probably not necessary anymore, therefor I close this PR

PR #1903, opened 2025-03-12
Make `MonitorAwarePoint` a windows-specific implementation detail

nedtwigg:
- #1711 made `Point` and `Rectangle` part of a sealed class hierarchy which added `MonitorAware` variants
- this new hierarchy causes two problems
  - users cannot instantiate `Point` or `Rectangle` from Kotlin, because Kotlin does not allow the base of a sealed hierarchy to be one of the elements
  - even for Java users, equality is now muddled. Two points can have the same `x, y` values, but not be equal. For example, take a point, map it to a control, map it back, you'll get the same coordinates you started with, but `.equals` will be false now because one of the points will be a `MonitorAware`. This behavior only happens on Windows.

I'm proposing two PRs. This, the first one, doesn't fix the muddled equality bug, but it makes `MonitorAware` a package-private implementation detail that only affects the win32 platform.

The second one (#1904) removes `MonitorAware` and replaces it with logic that is contained within the `MultiZoomCoordinateMapper`. As of SWT in the release 4.35, these new `MonitorAware` variants are created in exactly two places.

  - https://github.com/eclipse-platform/eclipse.platform.swt/blob/e1f11c985538793b9ae512ddedba1fa3a2108fff/bundles/org.eclipse.swt/Eclipse%20SWT/win32/org/eclipse/swt/widgets/MultiZoomCoordinateSystemMapper.java#L155
  - https://github.com/eclipse-platform/eclipse.platform.swt/blob/e1f11c985538793b9ae512ddedba1fa3a2108fff/bundles/org.eclipse.swt/Eclipse%20SWT/win32/org/eclipse/swt/widgets/MultiZoomCoordinateSystemMapper.java#L264

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Point.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Point.java
@@ -41,7 +41,7 @@ import java.io.*;
  * @see <a href="http://www.eclipse.org/swt/">Sample code and further information</a>
  */
 
-public sealed class Point implements Serializable permits MonitorAwarePoint {
+public class Point implements Serializable {
 
 	/**
 	 * the x coordinate of the point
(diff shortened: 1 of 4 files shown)


HeikoKlare (commented):
Unfortunately, this is not possible for multiple reasons:
1. The `MonitorAware...` classes are used from another package, so this change does not compile. We also wanted to make both the coordinate system mappers as well as the monitor-aware classes internal, but there is unfortunately no JLS-conforming way to do that, so we decided to make the monitor-aware classes public but marking them as `@noreference`.
2. `Point` and `Rectangle` were `final` and are not supposed to be subclassed in an unexpected way (i.e., outside SWT). That's why sealed types are a valid replacement but just removing the `final` modifier is not.

nedtwigg:
- closing in favor of #1905

PR #1906, opened 2025-03-14
[win32] refactor Image::init with ImageData

akoch-yatta:
This PR refactors Image#init for ImageData in the win32 implementation. Currently this method serves two purposes: 

1. creating handles for the given ImageData 
2. Doing 1. but additionally initializing an image with those handles

These two use cases are made more explicit refactoring this logic out into concrete methods. This is a necessary refactoring to be able to create all OS handles on demand and create temporary OS handles if neede.


HeikoKlare (approved):
This is a nice improvement as it disentagles multiple concerns implemented in one method and allows to explicitly call more specific methods at consumers. I traced every possible flow through the effected code to compare old and new implementation (and have to admit that it's rather difficult to follow all the flows) and can confirm that I did not find any difference before/after. I also tested an SDK with the change and found everything to work fine.

PR #1909, opened 2025-03-17
[win32] Create ImageData based image handles on demand

akoch-yatta:
This PR refactors the Image constructors using plain ImageData to create all handles on demand instead of creating a first handle in the constructor.

## Requires
- [x] #1910

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
@@ -371,9 +371,7 @@ public Image(Device device, ImageData data) {
 	super(device);
 	if (data == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
 	initialNativeZoom = DPIUtil.getNativeDeviceZoom();
-	int deviceZoom = getZoom();
-	data = scaleImageData(data, deviceZoom, 100);
-	init(data, deviceZoom);
+	this.imageProvider = new PlainImageDataProviderWrapper(data);
 	init();
 	this.device.registerResourceWithZoomSupport(this);
 }
@@ -416,10 +414,7 @@ public Image(Device device, ImageData source, ImageData mask) {
 		SWT.error(SWT.ERROR_INVALID_ARGUMENT);
 	}
 	initialNativeZoom = DPIUtil.getNativeDeviceZoom();
-	source = scaleImageData(source, getZoom(), 100);
-	mask = scaleImageData(mask, getZoom(), 100);
-	mask = ImageData.convertMask(mask);
-	initIconHandle(this.device, source, mask, getZoom());
+	this.imageProvider = new MaskedImageDataProviderWrapper(source, mask);
 	init();
 	this.device.registerResourceWithZoomSupport(this);
 }
@@ -479,11 +474,9 @@ public Image(Device device, ImageData source, ImageData mask) {
  */
 public Image (Device device, InputStream stream) {
 	super(device);
+	if (stream == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
 	initialNativeZoom = DPIUtil.getNativeDeviceZoom();
-	int deviceZoom = getZoom();
-	ElementAtZoom<ImageData> imageCandidate = ImageDataLoader.load(stream, FileFormat.DEFAULT_ZOOM, deviceZoom);
-	ImageData data = scaleImageData(imageCandidate.element(), deviceZoom, imageCandidate.zoom());
-	init(data, deviceZoom);
+	this.imageProvider = new ImageDataLoaderStreamProviderWrapper(stream);
 	init();
 	this.device.registerResourceWithZoomSupport(this);
 }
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1947,6 +1937,156 @@ protected void destroy() {
...
+	protected Rectangle getBounds(int zoom) {
+		Rectangle rectangle = new Rectangle(0, 0, srcAt100.width, srcAt100.height);
+		return DPIUtil.scaleUp(rectangle, zoom);
+	}
+
+	@Override
+	protected ElementAtZoom<ImageData> loadImageData(int zoom) {
+		ImageData scaledSource = DPIUtil.scaleImageData(device, srcAt100, zoom, 100);
+		ImageData mask = ImageData.convertMask(maskAt100);
+		ImageData scaledMask = DPIUtil.scaleImageData(device, mask, zoom, 100);
```
HeikoKlare, line 2021: Two questions on this:
1. The order of these operations has changed. Is that correct?
2. If yes, can't we sotre the converted mask instead of cloning and storing the original one already in the constructor?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1947,6 +1937,156 @@ protected void destroy() {
...
+		return new ElementAtZoom<>(imageDataAt100, 100);
+	}
+
+	@Override
+	AbstractImageProviderWrapper createCopy(Image image) {
+		return image.new PlainImageDataProviderWrapper(this.imageDataAt100);
+	}
+}
+
+private class MaskedImageDataProviderWrapper extends ImageFromImageDataProviderWrapper {
```
HeikoKlare, line 2001: Previously, for images with masks the method `initIconHandle` had been called, doing some additional icon-specific initialization. Why isn't that necessary anymore? Is that now already captured by `Image#getImageMetadata()`?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1947,6 +1937,156 @@ protected void destroy() {
...
+		}
+	}
+
+	private ImageDataLoaderStreamProviderWrapper(byte[] inputStreamData) {
+		this.inputStreamData = inputStreamData;
+	}
+
+	@Override
+	protected ElementAtZoom<ImageData> loadImageData(int zoom) {
+		return ImageDataLoader.load(new ByteArrayInputStream(inputStreamData), FileFormat.DEFAULT_ZOOM, zoom);
```
HeikoKlare, line 2050: This is a nice improvement as in case you have a stream for a dynamically-sized image (like an SVG) you can then create it at the appropriate zoom.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1947,6 +1937,156 @@ protected void destroy() {
...
+		return new Rectangle(0, 0, scaledImageData.width, scaledImageData.height);
+	}
+
+	@Override
+	AbstractImageProviderWrapper createCopy(Image image) {
+		return new ImageDataLoaderStreamProviderWrapper(inputStreamData);
+	}
+}
+
+private class ImageDataLoaderFileProviderWrapper extends ImageFromImageDataProviderWrapper {
```
HeikoKlare, line 2065: Is this one really necessary? Can't we creat an `ImageFileNameProvider`, which only returns a file name for zoom 100?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: No, ImageFromImageDataProviderWrapper#getImageMetadata is just using the more generic init method now, which will use `initIconHandle(long[] handles, int zoom)`. If you mean `initIconHandle(Device device, ImageData source, ImageData mask, Integer zoom)` this logic is covered as well. The mask is applied in MaskedImageDataProviderWrapper#loadImageData and the other two calls are then covered by the generic init method

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: We actually had SVG in mind here. Wouldn't that lead to a worse result, if a ImageFileNameProvider like that would be used with an SVG?

HeikoKlare: I don't think so. The ImageFileNameProvider implementation retrieves the most fitting file name for the desired zoom, which in case of SVG is always 100 (and for raster graphics it might also be 150 or 200). For that file name, the ImageLoader is used to generate an image in the desired size, if possible. For raster graphics this is limited to what is provided by the file, whereas for SVGs the rasterizer will take that into account.
Or in short: for SVGS the Image constructor taking a file name currently produces the exact same results than the one with an ImageFileNameProvider anyway.

akoch-yatta: Adapted it accordingly

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
HeikoKlare: Thanks for the clarification and correction!

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
HeikoKlare: I see. I was referring to the second one (`initIconHandle(Device device, ImageData source, ImageData mask, Integer zoom)`), which is now called via the generic `init` method at on- demand-creation of the handle. I just did not find the conformance between the previous and current implementation, but it's rather because of the different ways taken in the previous implementation while the new implementation is definitely more straight-forward always utilizing the generic init method.

HeikoKlare (approved):

PR #1910, opened 2025-03-17
[win32] Set background color on all image handles

akoch-yatta:
This PR adapts the win32 implementation of Image, so ensure that the background color of an image is always set to all existing of later created image handles. Currently all handles that are created after the background color was set, would not receive the background color.

## Prerequisite for
- #1909

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
@@ -118,6 +118,12 @@ public final class Image extends Resource implements Drawable {
 	private int styleFlag = SWT.IMAGE_COPY;
 
 	/**
+	 * Sets the color to which to map the transparent pixel.
+	 * For further info see {@link #setBackground(Color)}
+	 */
+	private RGB backgroundColor;
+
+	/**
 	 * Attribute to cache current native zoom level
 	 */
 	private int initialNativeZoom = 100;
@@ -1087,6 +1093,10 @@ public boolean equals (Object object) {
 public Color getBackground() {
 	if (isDisposed()) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
 	if (transparentPixel == -1) return null;
+	if (backgroundColor != null) {
+		// if a background color was set explicitly, we use the cached color directly
+		return Color.win32_new(device, (backgroundColor.blue << 16) | (backgroundColor.green << 8) | backgroundColor.red);
+	}
 
 	/* Get the HDC for the device */
 	long hDC = device.internal_new_GC(null);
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2421,9 +2409,37 @@ public ImageHandle(long handle, int zoom) {
 		this.handle = handle;
 		this.zoom = zoom;
 		updateBoundsInPixelsFromNative();
+		if (backgroundColor != null) {
+			setBackground(backgroundColor);
+		}
 		setImageMetadataForHandle(this, zoom);
 	}
 
+	private void setBackground(RGB color) {
```
fedejeanne, line 2418: NIT: move the method back to where it was so it's easier to see the changes made by this PR.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1821,33 +1831,11 @@ public void setBackground(Color color) {
 	if (isDisposed()) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
 	if (color == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
 	if (color.isDisposed()) SWT.error(SWT.ERROR_INVALID_ARGUMENT);
-	zoomLevelToImageHandle.values().forEach(imageHandle -> setBackground(color, imageHandle.handle));
-}
-
-private void setBackground(Color color, long handle) {
+	if (backgroundColor != null) SWT.error(SWT.ERROR_INVALID_ARGUMENT);
```
fedejeanne, line 1834: We need to enforce this on other OSes too, it's part of the contract. You can do it in another PR if you want.

fedejeanne (commented):
I would like to have a concrete way to test this though. 
Also: 
- what was the bug that's being fixed?
- Is it a bug or simply an inconsistency in the API contract?
- Is there a test to make sure that an exception is thrown if `setBackground(...)` is called twice? If not, can you add it?
- Is there a test that makes sure that images created afterward maintain the background? If not, can you add it?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: I would like to move it, because it is tightly related to the ImageHandle. But we could separate that and do the logic change first and a separate PR just to do the move, if you want.

akoch-yatta:
> I would like to have a concrete way to test this though. Also:
> 
> * what was the bug that's being fixed?
> * Is it a bug or simply an inconsistency in the API contract?
I didn't actively saw the issue somewhere as this functionality is not used very much, but we should be able to create an issue with a Snippet. I will have a look into that.
> * Is there a test that makes sure that images created afterward maintain the background? If not, can you add it?
Yes, I can add one in the win32 specific tests

> * Is there a test to make sure that an exception is thrown if `setBackground(...)` is called twice? If not, can you add it?
No, but I can add one

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
fedejeanne: No need to, it's a "NIT" comment so it's your call . Leave it as-is :-)

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: I removed this line from this PR. If we wan't to enforce this, we should enforce this on all OS. If we wanna do it, we should have one PR for all implementations

fedejeanne (commented):
The changes look OK. One final question: is `transparentColor` even needed anymore? I see that it's only used **outside** of `Image` (in `GC`) but it looks to me as if it can be replaced by the newly added `backgroundColor`. 

WDYT @akoch-yatta ?

akoch-yatta:
> The changes look OK. One final question: is `transparentColor` even needed anymore? I see that it's only used **outside** of `Image` (in `GC`) but it looks to me as if it can be replaced by the newly added `backgroundColor`.
> 
> WDYT @akoch-yatta ?

There is quite a similar logic involved there. I am just a bit hesitant to do something like that, because I don't fully understand the whole logic there.

PR #1913, opened 2025-03-18
Smooth Scaling Rounding error fix for win32 #62

amartya4256:
This PR contributes to fixing the implementation of Smooth scaling of the ImageData to get rid of the rounding errors because of multiple scale ups and downs with fractional scale factor. The commit replicates and modifies the DPIUtil::autoScaleImageData method implementation in Image class to adapt the same.

contributes to

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
@@ -361,7 +361,7 @@ public Image(Device device, ImageData data) {
 	if (data == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
 	initialNativeZoom = DPIUtil.getNativeDeviceZoom();
 	int deviceZoom = getZoom();
-	data = DPIUtil.scaleImageData(device, new ElementAtZoom<>(data, 100), deviceZoom);
+	data = scaleImageData(data, deviceZoom, 100);
 	init(data, deviceZoom);
 	init();
 	this.device.registerResourceWithZoomSupport(this);
@@ -405,8 +405,8 @@ public Image(Device device, ImageData source, ImageData mask) {
 		SWT.error(SWT.ERROR_INVALID_ARGUMENT);
 	}
 	initialNativeZoom = DPIUtil.getNativeDeviceZoom();
-	source = DPIUtil.autoScaleUp(device, source);
-	mask = DPIUtil.autoScaleUp(device, mask);
+	source = scaleImageData(source, getZoom(), 100);
+	mask = scaleImageData(mask, getZoom(), 100);
 	mask = ImageData.convertMask(mask);
 	initIconHandle(this.device, source, mask, getZoom());
 	init();
@@ -470,7 +470,8 @@ public Image (Device device, InputStream stream) {
 	super(device);
 	initialNativeZoom = DPIUtil.getNativeDeviceZoom();
 	int deviceZoom = getZoom();
-	ImageData data = DPIUtil.scaleImageData(device, ImageDataLoader.load(stream, FileFormat.DEFAULT_ZOOM, deviceZoom), deviceZoom);
+	ElementAtZoom<ImageData> imageCandidate = ImageDataLoader.load(stream, FileFormat.DEFAULT_ZOOM, deviceZoom);
+	ImageData data = scaleImageData(imageCandidate.element(), deviceZoom, imageCandidate.zoom());
 	init(data, deviceZoom);
 	init();
 	this.device.registerResourceWithZoomSupport(this);
@@ -513,7 +514,8 @@ public Image (Device device, String filename) {
 	if (filename == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
 	initialNativeZoom = DPIUtil.getNativeDeviceZoom();
 	int deviceZoom = getZoom();
-	ImageData data = DPIUtil.scaleImageData(device, ImageDataLoader.load(filename, FileFormat.DEFAULT_ZOOM, deviceZoom), deviceZoom);
+	ElementAtZoom<ImageData> imageCandidate = ImageDataLoader.load(filename, FileFormat.DEFAULT_ZOOM, deviceZoom);
+	ImageData data = scaleImageData(imageCandidate.element(), deviceZoom, imageCandidate.zoom());
 	init(data, deviceZoom);
 	init();
 	this.device.registerResourceWithZoomSupport(this);
@@ -1237,7 +1239,7 @@ private ImageData getScaledImageData (int zoom) {
 	}
 	TreeSet<Integer> availableZooms = new TreeSet<>(zoomLevelToImageHandle.keySet());
 	int closestZoom = Optional.ofNullable(availableZooms.higher(zoom)).orElse(availableZooms.lower(zoom));
-	return DPIUtil.scaleImageData (device, getImageMetadata(closestZoom).getImageData(), zoom, closestZoom);
+	return scaleImageData(getImageMetadata(closestZoom).getImageData(), zoom, closestZoom);
 }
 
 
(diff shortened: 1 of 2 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1850,6 +1853,41 @@ private void setBackground(Color color, long handle) {
 	device.internal_dispose_GC(hDC, null);
 }
 
+private ImageData scaleImageData(Device device, int currentZoom, int targetZoom, ImageData imageData) {
```
fedejeanne, line 1856: 1. Should be `static`
2. Why change the order of the parameters? It only adds more lines to review and also makes the new method a non-drop-in replacement.

```suggestion
private static ImageData scaleImageData(Device device, final ImageData imageData, int targetZoom, int currentZoom) {
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1850,6 +1853,41 @@ private void setBackground(Color color, long handle) {
...
+	int width = imageData.width;
+	int height = imageData.height;
+	int scaledWidth = Math.round (width * scaleFactor);
+	int scaledHeight = Math.round (height * scaleFactor);
+	boolean useSmoothScaling = DPIUtil.isSmoothScalingEnabled() && imageData.getTransparencyType() != SWT.TRANSPARENCY_MASK;
+	if (useSmoothScaling) {
+		return scaleToUsingSmoothScaling(scaledWidth, scaledHeight, imageData);
+	} else {
+		return imageData.scaledTo (scaledWidth, scaledHeight);
+	}
```
fedejeanne, line 1868:
```suggestion
	}
	return imageData.scaledTo (scaledWidth, scaledHeight);
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1850,6 +1853,41 @@ private void setBackground(Color color, long handle) {
...
+}
+
+private ImageData scaleToUsingSmoothScaling(int width, int height, ImageData imageData) {
+	Image original = new Image (Display.getCurrent(), (ImageDataProvider) zoom -> imageData);
```
fedejeanne, line 1872: Why `Display.getCurrent()` and not simply `device`, like it is in the original method in `DPIUtil`? Is there any use case where using `device` brings problems?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1850,6 +1853,41 @@ private void setBackground(Color color, long handle) {
...
+	/* Create a 24 bit image data with alpha channel */
+	final ImageData resultData = new ImageData (width, height, 24, new PaletteData (0xFF, 0xFF00, 0xFF0000));
+	resultData.alphaData = new byte [width * height];
+	Image resultImage = new Image (Display.getCurrent(), (ImageDataProvider) zoom -> resultData);
```
fedejeanne, line 1876: Same as a few lines above: why `Display.getCurrent()`... ?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1850,6 +1853,41 @@ private void setBackground(Color color, long handle) {
...
+	int scaledHeight = Math.round (height * scaleFactor);
+	boolean useSmoothScaling = DPIUtil.isSmoothScalingEnabled() && imageData.getTransparencyType() != SWT.TRANSPARENCY_MASK;
+	if (useSmoothScaling) {
+		return scaleToUsingSmoothScaling(scaledWidth, scaledHeight, imageData);
+	} else {
+		return imageData.scaledTo (scaledWidth, scaledHeight);
+	}
+}
+
+private ImageData scaleToUsingSmoothScaling(int width, int height, ImageData imageData) {
```
fedejeanne, line 1871:
```suggestion
private static ImageData scaleToUsingSmoothScaling(int width, int height, ImageData imageData) {
```

fedejeanne (commented):
We briefly discussed this in the call but I'd like to point it out here again: wouldn't it be better to fix the smooth scaling for all OSes instead of duplicating the logic in `Image` (which makes it Windows-specific)? 

Did you try fixing it for all OSes already? Did something break?

If it's just a matter of risk management, we can simply try and fix it in `DPIUtil` and test in Linux and Mac too.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
amartya4256: I forgot to change it back to device, my bad.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
amartya4256: Oops

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
amartya4256: Done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
amartya4256: But cannot make it static since we need some of the attributes of the image later in a method we call from this method.

HeikoKlare:
> We briefly discussed this in the call but I'd like to point it out here again: wouldn't it be better to fix the smooth scaling for all OSes instead of duplicating the logic in `Image` (which makes it Windows-specific)?

Can you please share why the solution was to duplicate functionality? We now have the (almost?) same implementation inside `Image` (on Windows) and inside `DPIUtil`. No one will know which one to use when, so I would either expect to have the fix done in `DPIUtil` or to move the implementation to the OS-specific `Image` classes if there needs to be a difference between them.

amartya4256:
We need to draw directly using pixel specific implementation and we cannot access that gc::drawImage usign pixels from DPIUtil. Moreover, DPIUtil uses intermediate scale ups ad downs so I moved it to the Image. Another idea will be to move the method to ImageData and call that from DPIUtil. But for that I need an API in ImageData for smooth scaling. (ImageData is in commons)

HeikoKlare:
But why is that specific to Windows? Shouldn't the functionality then be generally moved into the Image classes?

amartya4256:
It cannot be completely moved to Image because its not completely specific to Image but also Cursor and a few other places.

amartya4256:
Better option would be to move the smooth scaling block in the DPIUtil::autoScaleImageData to a new public method in ImageData class. There we can get rid of the rounding error.

HeikoKlare:
> Better option would be to move the smooth scaling block in the DPIUtil::autoScaleImageData to a new public method in ImageData class. There we can get rid of the rounding error.

Doing that in ImageData would be best anyway since there are several consumers using ImageData#scaledTo, which will produce bad results even if smooth scaling is enabled.
In any way, I think the duplication we have now is quite inacceptable. GC#drawImage contains a hack (explicitly marked as such via an according comment), which this change now works around via another hack. So we should at least factor out the common code into DPIUtil and pass the GC-drawing functionality to that common code as a lambda (or, of course, even better move the functionality into ImageData).

amartya4256:
Then we should revert the commit and I'll create another PR with a new impl.

Issue #1915, opened 2025-03-18
[GTK3] Reparenting a widget results in losing focus


trancexpress:
Seen on Linux, GTK 3. The resulting bug is: https://github.com/eclipse-platform/eclipse.platform.ui/issues/2839

I.e. switching perspectives when a view is in both perspectives results in reparenting (so that the view is under the right tab, w.r.t. part stacks in each perspective).

Adding focus out event listening to `Control`, I see this stack trace:

```
java.lang.Exception: Control focus out StyledText {} [layout=null]
	at org.eclipse.swt.widgets.Control.lambda$0(Control.java:796)
	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:91)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:5862)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1652)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1678)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1657)
	at org.eclipse.swt.widgets.Control.sendFocusEvent(Control.java:4889)
	at org.eclipse.swt.widgets.Control.gtk_event_after(Control.java:3753)
	at org.eclipse.swt.widgets.Widget.windowProc(Widget.java:2620)
	at org.eclipse.swt.widgets.Control.windowProc(Control.java:6846)
	at org.eclipse.swt.widgets.Display.windowProc(Display.java:6169)
	at org.eclipse.swt.internal.gtk3.GTK3.gtk_widget_reparent(Native Method)
	at org.eclipse.swt.widgets.Control.gtk_widget_reparent(Control.java:2857)
	at org.eclipse.swt.widgets.Control.setParent(Control.java:5874)
```

Note that reparenting is done by platform UI for parents of the `StyledText`, e.g. the tab folders (and other widgets).

It would be nice if focus is restored after the reparenting operation.

PR #1916, opened 2025-03-18
Restore focus that is lost during Control.setParent()

trancexpress:
When calling Control.setParent() during a perspective switch, focus is lost on GTK3. This change restores the focus to the last control that had it, after the reparenting operation.

Fixes: #1915


bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Control.java
```diff
@@ -5888,6 +5890,10 @@ public boolean setParent (Composite parent) {
 	this.parent = parent;
 	setZOrder (null, false, true);
 	reskin (SWT.ALL);
+	// restore focus to the last Control that had it, if focus is now gone
+	if (focusControlBeforeReparent != null && display.getFocusControl() == null) {
+		focusControlBeforeReparent.setFocus();
```
iloveeclipse, line 5895: Can `focusControlBeforeReparent` be disposed as a result of reparenting? Maybe better to check that.

Beside this, works as expected.

iloveeclipse (approved):

PR #1920, opened 2025-03-19
Edge: Fix URI_FOR_CUSTOM_TEXT_PAGE in case tmpdir contains unicode chars

### LINKED ISSUE #1912: Edge: instead of displaying HTML content in the browser control the browser is opened
**Describe the bug**
The first problem happes when I try to install Eclipse with the Eclipse installer. The installer uses a browser control to display information. But instead of displaying the content within the control for each HTML snippet a window with a temporary file name is opened in the operating system's standard browser.

Within Eclipse the same happens for content that should be displayed in an information windows ("tooltip"). The information window is empty and a new window is opened in the standard browser.
If the Web Browser preferences are set to "Use internal web browser" a new empty part is opened within Eclipse where an empty HTML file is displayed.

In this case the part shows an error message:

Image

The Javadoc view is completely empty.

The problem may be caused due to a configuration problem of my Windows user because it does not happen when I login with another user id.

**To Reproduce**
Start the Eclipse installer.

**Environment:**
-  [ ] All OS
-  [x] Windows
-  [ ] Linux
-  [ ] macOS 

Windows 11 Professional 24H2

3. JRE/JDK version
Tested with JDK 21 and 23 

**Version since**
 Eclipse or SWT version since when the behavior is seen [e.g. 4.35]

Use older IE control.

sratz:
We must ensure that we always use the ASCII-String (URL-encoded) everywhere.

Refactor code a bit to make it possible to add a testcase.

Fixes #1912.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
@@ -35,6 +35,7 @@ import org.eclipse.swt.widgets.*;
 class Edge extends WebBrowser {
 	static {
 		Library.loadLibrary("WebView2Loader");
+		setupLocationForCustomTextPage();
 	}
 
 	// WebView2Loader.dll compatible version. This is NOT the minimal required version.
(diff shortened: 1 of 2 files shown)


HeikoKlare (approved):
The change looks sound to me. ASCII conversion was obviously not done at all places as required before.
The added test is heavily tied to implementation details but I don't see a better to test the behavior either.

sratz:
> The added test is heavily tied to implementation details but I don't see a better to test the behavior either.

Yes, so much static stuff :o

But the new way of having platform dependent test cases inside the fragment close to the main sources is great and at least makes such things testable at all!

sratz:
Build failure in compiling gtk.linux.riscv64 binaries is unrelated.

PR #1921, opened 2025-03-19
Do not check the device in Display::getPrimaryMonitor

fedejeanne:
Fixes https://github.com/vi-eclipse/Eclipse-Platform/issues/256

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
@@ -2302,7 +2302,6 @@ long getMsgProc (long code, long wParam, long lParam) {
  * @since 3.0
  */
 public Monitor getPrimaryMonitor () {
-	checkDevice ();
 	long hmonitor = OS.MonitorFromWindow (0, OS.MONITOR_DEFAULTTOPRIMARY);
 	return getMonitor (hmonitor);
 }
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_Font.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_Font.java
@@ -19,6 +19,7 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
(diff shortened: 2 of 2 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
```diff
@@ -2302,7 +2302,6 @@ long getMsgProc (long code, long wParam, long lParam) {
  * @since 3.0
  */
 public Monitor getPrimaryMonitor () {
-	checkDevice ();
```
laeubi, line 2305: Can this really be called from *any thread* and even concurrently? If yes then this should be reflected in the Javadoc. How about other OS?

fedejeanne: Honestly, it's probably a bad idea but I wanted to give it a try (and also see if the tests showed some errors).

Also, `checkDevice()` does more than checking the thread: it also checks for disposed resources so it can not be simply left out. If anything, one should explicitly skip the check for the thread but perform the check for disposed.

The problem I'm trying to solve (https://github.com/vi-eclipse/Eclipse-Platform/issues/256) boils down to creating fonts from non-ui threads, which means that one needs to be able to extract the zoom of the primary monitor from a non-ui thread too.

I'm going to see if there is another way of getting the zoom by using `Display::(a)syncExec` without causing a deadlock,

laeubi: > The problem I'm trying to solve (https://github.com/vi-eclipse/Eclipse-Platform/issues/256) boils down to creating fonts from non-ui threads, which means that one needs to be able to extract the zoom of the primary monitor from a non-ui thread too.

I think this is really worse, you should never create a `Font` outside of the display thread (even if that might works) and instead the `FontData` and possibly create a font on demand later on.

laeubi:
If I read between the lines, it *might* be that the actual issue is on Windows there are multiple `Display`s but with a `null` device it always uses `Display.getDefault()` while it better first should check for `Display.getCurrent()` ... in any case one can avoid this by always pass a proper device and not rely on the defaults.

laeubi:
Yes not everything is optimal here... At best `Font`s where like `Color`s to become independent from a device.

In any case I think removing a device check is not the right approach, if called from the wrong thread I would expect that simply a zoom-factor of 100 is assumed.

HeikoKlare:
> In any case I think removing a device check is not the right approach, 

Fully agree.

> if called from the wrong thread I would expect that simply a zoom-factor of 100 is assumed.

In general, a zoom factor of 100 will be wrong. For that reason, it might serve as a fallback but in case a potentially wrong value is assigned, we should probably log some warning to make aware of wrong usage and potential issues arising from it.

As a completely different way of thought: couldn't we just store the primary monitor for a display (or at least the primary monitor zoom)? It can be updated via OS events and will relieve us from the necessity of repeated OS calls for retrieving the (usually always same) primary monitor.

HeikoKlare:
Obsolete via https://github.com/eclipse-platform/eclipse.platform.swt/pull/1943

PR #1924, opened 2025-03-20
Replace usages of new Image(device, width, height)

ShahzaibIbrahim:
Replacing it with imageGcDrawer

### DIFF
--- a/bundles/org.eclipse.swt.tools.spies/src/org/eclipse/swt/tools/internal/Sleak.java
+++ b/bundles/org.eclipse.swt.tools.spies/src/org/eclipse/swt/tools/internal/Sleak.java
@@ -359,14 +359,8 @@ private void saveToFile(boolean prompt) {
 			if (saveImages) {
 				String suffix = String.format("%05d.png", i++);
 				String pngName = String.format("%s_%s", fileName, suffix);
-				Image image = new Image(saveAs.getDisplay(), 100, 100);
+				Image image = new Image(saveAs.getDisplay(), (gc, width, height) -> draw(gc, object), 100, 100);
 				try {
-					GC gc = new GC(image);
-					try {
-						draw(gc, object);
-					} finally {
-						gc.dispose();
-					}
 					ImageLoader loader = new ImageLoader();
 					loader.data = new ImageData[] { image.getImageData() };
 					loader.save(pngName, SWT.IMAGE_PNG);
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/StyledText.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/StyledText.java
@@ -1593,15 +1593,15 @@ void createCaretBitmaps() {
 		leftCaretBitmap.dispose();
 	}
 	int lineHeight = renderer.getLineHeight();
-	leftCaretBitmap = new Image(display, caretWidth, lineHeight);
-	GC gc = new GC (leftCaretBitmap);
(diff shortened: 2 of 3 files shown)


HeikoKlare (commented):
Can you please share some information on how to test the change to `StyledText` (for regression)? Where can those affected carets be seen?

bundles/org.eclipse.swt.tools.spies/src/org/eclipse/swt/tools/internal/Sleak.java
```diff
@@ -359,14 +359,11 @@ private void saveToFile(boolean prompt) {
 			if (saveImages) {
 				String suffix = String.format("%05d.png", i++);
 				String pngName = String.format("%s_%s", fileName, suffix);
-				Image image = new Image(saveAs.getDisplay(), 100, 100);
+				final ImageGcDrawer imageGcDrawer = (gc, width, height) -> {
+					draw(gc, object);
+				};
+				Image image = new Image(saveAs.getDisplay(), imageGcDrawer, 100, 100);
```
HannesWell, line 365:
```suggestion
				Image image = new Image(saveAs.getDisplay(), (gc, w, h) -> draw(gc, object), 100, 100);
```

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/StyledText.java
```diff
@@ -1593,31 +1593,31 @@ void createCaretBitmaps() {
...
-	leftCaretBitmap = new Image(display, caretWidth, lineHeight);
-	GC gc = new GC (leftCaretBitmap);
-	gc.setBackground(display.getSystemColor(SWT.COLOR_BLACK));
-	gc.fillRectangle(0, 0, caretWidth, lineHeight);
-	gc.setForeground(display.getSystemColor(SWT.COLOR_WHITE));
-	gc.drawLine(0,0,0,lineHeight);
-	gc.drawLine(0,0,caretWidth-1,0);
-	gc.drawLine(0,1,1,1);
-	gc.dispose();
+	final ImageGcDrawer imageGcDrawer = (gc, width, height) -> {
```
HannesWell, line 1596: I would have inlined the lambda, which also avoids finding a variable name for it.

In hindsight and with this example in mind, I think it would the code would have be a bit nicer if the `ImageGcDrawer` argument would be the last of the new constructor. But unfortunately this ship has sailed 😞

ShahzaibIbrahim:
> Can you please share some information on how to test the change to `StyledText` (for regression)? Where can those affected carets be seen?

I have tested it with snippet332 replacing the styleText string with "English متن بالعربية עברית " and forcing isBidiCaret() method in StyledText class to return true (cos it checks from OS if it is Bidirectional Platform) and with my changes the caret looks same compared to previous versions. I am not sure I could programmatically write any test for it.

HeikoKlare (commented):
> I have tested it with snippet332 replacing the styleText string with "English متن بالعربية עברית " and forcing isBidiCaret() method in StyledText class to return true (cos it checks from OS if it is Bidirectional Platform)

Thank you. I have just tested with that as well and found the adapted code to be working fine. I actually expected the carets to become "better" because of proper scaling, but they are regenerated on a zoom change for the StyledText anyway, which makes the results look exactly the same, even in monitor-specific scaling scenarios.

Could you please inline at least the single-line lamda as proposed by Hannes? For the others I (personally) would be in favor of having it assigned to a variable to not make the code overly complex, but having a better name like `leftCaretDrawer` or the like would be good.

HeikoKlare (approved):

PR #1928, opened 2025-03-21
[Win32] MultiZoomCoordinateSystemMapper: zero-sized rectangle handling

HeikoKlare:
The MultiZoomCoordinateSystemMapper currently produces a devide-by-zero error when transforming a rectangle of zero size. This can, for example, happen when a (dummy) shell with width=height=0 is created and its bounds are passed to the mapper.

This change ensures that the problematic calculation is avoided in case width or height are zero. In that case, no reasonable monitor can be identified anyway.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MultiZoomCoordinateSystemMapper.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MultiZoomCoordinateSystemMapper.java
@@ -192,6 +192,9 @@ class MultiZoomCoordinateSystemMapper implements CoordinateSystemMapper {
 	}
 
 	private Monitor getContainingMonitorForPoints(int x, int y, int width, int height) {
+		if (width <= 0 || height <= 0) {
+			return getContainingMonitorForPoints(x, y);
+		}
 		Monitor[] monitors = monitorSupplier.get();
 		Monitor selectedMonitor = null;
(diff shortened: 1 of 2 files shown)


HeikoKlare:
@amartya4256 can you please review this?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MultiZoomCoordinateSystemMapper.java
```diff
@@ -192,6 +192,9 @@ private Monitor getContainingMonitorForPoints(int x, int y) {
 	}
 
 	private Monitor getContainingMonitorForPoints(int x, int y, int width, int height) {
+		if (width <= 0 || height <= 0) {
+			return null;
+		}
```
amartya4256, line 197: I think it's better to return the monitor of the containing point (x,y)in this case using `getContainingMonitorForPoints(int x, int y)`
Since this function's caller is `getValidMonitorIfApplicable(int x, int y, int width, int height, Monitor monitor)` and it assigns the rectangle the defaultMonitor in case null is returned. For correctness, we can keep the monitor consistent by calling `getContainingMonitorForPoints(int x, int y)`. What do you think?

amartya4256: Looks perfect now. I t

HeikoKlare:

PR #1932, opened 2025-03-24
Jenkinsfile: Use Tycho 4.0.13 way to display api tools warnings

sratz:
Also make sure the issues are published even in case of a failed build, as otherwise they are not visualized.

See https://github.com/eclipse-tycho/tycho/pull/4897

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -344,12 +344,18 @@ pipeline {
 			}
 			post {
 				always {
-					junit 'eclipse.platform.swt/tests/*.test*/target/surefire-reports/*.xml'
-					archiveArtifacts artifacts: '**/*.log,*/binaries/*/target/*.jar', excludes: '**/*-sources.jar'
+					junit allowEmptyResults: true, testResults: 'eclipse.platform.swt/tests/*.test*/target/surefire-reports/*.xml'
+					archiveArtifacts allowEmptyArchive: true, artifacts: '**/*.log,*/binaries/*/target/*.jar', excludes: '**/*-sources.jar'
 					discoverGitReferenceBuild referenceJob: 'eclipse.platform.swt/master'
 					// To accept unstable builds (test errors or new warnings introduced by third party changes) as reference using "ignoreQualityGate:true"
-					recordIssues publishAllIssues: true, ignoreQualityGate:true, tools: [eclipse(name: 'Compiler and API Tools', pattern: '**/target/compilelogs/*.xml'), javaDoc()], qualityGates: [[threshold: 1, type: 'DELTA', unstable: true]]
-					recordIssues publishAllIssues: true, ignoreQualityGate:true, tool: mavenConsole(), qualityGates: [[threshold: 1, type: 'DELTA_ERROR', unstable: true]]
+					recordIssues enabledForFailure: true, publishAllIssues: true, ignoreQualityGate:true,
+						tools: [
+							eclipse(name: 'Compiler', pattern: '**/target/compilelogs/*.xml'),
+							issues(name: 'API Tools', id: 'apitools', pattern: '**/target/apianalysis/*.xml'),
+							javaDoc()
+						],
+						qualityGates: [[threshold: 1, type: 'DELTA', unstable: true]]
+					recordIssues enabledForFailure: true, publishAllIssues: true, ignoreQualityGate:true, tool: mavenConsole(), qualityGates: [[threshold: 1, type: 'DELTA_ERROR', unstable: true]]
 				}
 			}
 		}


sratz:
See
https://ci.eclipse.org/releng/job/eclipse.platform.swt/job/PR-1932/4/ for a test result.
All the issues are now recorded even if the build failed.

Note: the link to the apitools is kind of broken (the link on the main page is broken as the results are put under to `/native` instead of `/apitools`). This is a bug in Jenkins warnings-ng (already fixed): https://github.com/jenkinsci/warnings-ng-plugin/pull/470

laeubi (approved):
Thanks for looking into this, I think this is good in general and ready to be merged, still there seem some issues in the output itself (that needs fixing at Tycho):

grafik

1. There is some strange `null:0` file assigned to a message
2. SWT.java seems not correctly linked
3. One message is duplicated for the Manifest

laeubi:
As said this is more something we probably can improve in Tycho, so nothing to block this PR!

sratz:
> As said this is more something we probably can improve in Tycho, so nothing to block this PR!

Yep, at least if line is 0 / file is null, we should not put those in the XML in Tycho.

> SWT.java seems not correctly linked

As for linked resources, I am not sure how to handle that. Maybe @HannesWell do you have an idea?

> One message is duplicated for the Manifest

It's actually not a duplicate, but two times the same message for two fragments:
```
[ERROR] [API ERROR] File unknown at line 0: The field org.eclipse.swt.SWT.MOZILLA has been removed (location: unknown)
[ERROR] [API ERROR] File MANIFEST.MF at line 1: The major version should be incremented in version 3.130.0, since API breakage occurred since version 3.129.0 (location: /home/jenkins/agent/workspace/eclipse.platform.swt_PR-1932/eclipse.platform.swt/binaries/org.eclipse.swt.gtk.linux.ppc64le/META-INF/MANIFEST.MF)
[ERROR] [API ERROR] File MANIFEST.MF at line 6: The major version should be incremented in version 3.130.0, since API breakage occurred since version 3.129.0 (location: /home/jenkins/agent/workspace/eclipse.platform.swt_PR-1932/eclipse.platform.swt/binaries/org.eclipse.swt.gtk.linux.x86_64/META-INF/MANIFEST.MF)
```

laeubi:
@sratz by the way feel free to merge as you like, we can improve this later on. Would be great to have the same for other platform repos as well.

laeubi:
> As for linked resources, I am not sure how to handle that. 

I would try to resolve the real system path somewhere in the analysis step:

https://github.com/eclipse-tycho/tycho/blob/e59ede590b6c666f8b0d3fd4c5b84aa9f9c12ea3/tycho-apitools-plugin/src/main/java/org/eclipse/tycho/apitools/ApiAnalysis.java#L437

we then would need to get this into the [ApiAnalysisResult](https://github.com/eclipse-tycho/tycho/blob/main/tycho-apitools-plugin/src/main/java/org/eclipse/tycho/apitools/ApiAnalysisResult.java)

Maybe even directly in the [ApiProblemDTO](https://github.com/eclipse-tycho/tycho/blob/e59ede590b6c666f8b0d3fd4c5b84aa9f9c12ea3/tycho-apitools-plugin/src/main/java/org/eclipse/tycho/apitools/ApiProblemDTO.java#L44)

HannesWell:
> > SWT.java seems not correctly linked
> 
> As for linked resources, I am not sure how to handle that. Maybe @HannesWell do you have an idea?

Sorry I haven't fully read through all linked issues/PR, but what exactly is the problem? That the SWT.java occurs twice or that it occurs at all?

sratz:
> Sorry I haven't fully read through all linked issues/PR, but what exactly is the problem? That the SWT.java occurs twice or that it occurs at all?

No worries, I think I found the problem.

The path returned for the finding is 
`Eclipse SWT/common/org/eclipse/swt/SWT.java`

which needs to be resolved to an absolute path in the context of the eclipse project's root directly:
`/jenkins/workspace/.../eclipse.platform.swt/binaries/org.eclipse.swt.win32.win32.x86_64`

But to do that it needs to consider linked resources defined in the fragment:
`
	<linkedResources>
		<link>
			<name>Eclipse SWT</name>
			<type>2</type>
			<locationURI>SWT_HOST_PLUGIN/Eclipse%20SWT</locationURI>
		</link>
		<link>
			<name>Eclipse SWT Tests</name>
			<type>2</type>
			
			<locationURI>PARENT-2-PROJECT_LOC/bundles/org.eclipse.swt/Eclipse%20SWT%20Tests</locationURI>
		</link>
	</linkedResources>
	<variableList>
		<variable>
			<name>SWT_HOST_PLUGIN</name>
			<value>$%7BPARENT-2-PROJECT_LOC%7D/bundles/org.eclipse.swt</value>
		</variable>
	</variableList>
`

For that we need basic support of user-defined variables: https://github.com/eclipse-tycho/tycho/issues/3820.

I've got a preliminary solution ready and will provide an improvement PR to Tycho which should resolve this.

PR #1936, opened 2025-03-24
Unify and improve image disablement calculation algorithm

HeikoKlare:
Each OS-specific implementation of the Image class has its own implementation of an algorithm to calculate a disabled representation of a given image. In addition, the algorithm used on Windows and MacOS produces bad results, which is why usually pre-generated disabled icons are used.

This change centralizes the implementations of those algorithms in ImageColorTransformers. The implementations used for the pre-generated disabled icons is added as default algorithm with an enhanced and the existing GTK-conforming algorithms being provided as options. It is supposed to render the usage of pre-generated disabled version of icons obsolete in many cases.

This is based on the snippet for testing different algorithms proposed by @killingerm and the according discussion in https://github.com/eclipse-platform/eclipse.platform.swt/discussions/1741

### How to test
Most icons Eclipse are provided in a pre-generated, disabled version. To completely disable this and enforce all icons to be generated on-the-fly via the disablement algorithm, you can change the value of `IWorkbenchRegistryConstants#ATT_DISABLEDICON` so that the according attribute defining the disabled icon path does not match.

The following screenshot shows how the three version of icons look like:
- Top: pre-generated icons and icons generated by proposed on-the-fly algorithm
- Middle: icons with the proposed `enhanced` on-the-fly algorithm
- Bottom: icons with the current on-the-fly algorithm (to be removed)

image

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
@@ -14,6 +14,8 @@
 package org.eclipse.swt.graphics;
 
 
+import static org.eclipse.swt.internal.image.ImageColorTransformer.DEFAULT_DISABLED_IMAGE_TRANSFORMER;
+
 import java.io.*;
 import java.util.*;
 import java.util.function.*;
@@ -435,12 +437,14 @@ private void createRepFromSourceAndApplyFlag(NSBitmapImageRep srcRep, int srcWid
 	long data = rep.bitmapData();
 	C.memmove(data, srcData, srcWidth * srcHeight * 4);
 	if (flag != SWT.IMAGE_COPY) {
-		final int redOffset, greenOffset, blueOffset;
+		final int redOffset, greenOffset, blueOffset, alphaOffset;
 		if (srcBpp == 32 && (srcBitmapFormat & OS.NSAlphaFirstBitmapFormat) == 0) {
 			redOffset = 0;
 			greenOffset = 1;
 			blueOffset = 2;
+			alphaOffset = 3;
 		} else {
+			alphaOffset = 0;
 			redOffset = 1;
 			greenOffset = 2;
 			blueOffset = 3;
@@ -448,16 +452,6 @@ private void createRepFromSourceAndApplyFlag(NSBitmapImageRep srcRep, int srcWid
 		/* Apply transformation */
 		switch (flag) {
 		case SWT.IMAGE_DISABLE: {
-			Color zeroColor = this.device.getSystemColor(SWT.COLOR_WIDGET_NORMAL_SHADOW);
-			RGB zeroRGB = zeroColor.getRGB();
-			byte zeroRed = (byte)zeroRGB.red;
-			byte zeroGreen = (byte)zeroRGB.green;
-			byte zeroBlue = (byte)zeroRGB.blue;
-			Color oneColor = this.device.getSystemColor(SWT.COLOR_WIDGET_BACKGROUND);
-			RGB oneRGB = oneColor.getRGB();
-			byte oneRed = (byte)oneRGB.red;
-			byte oneGreen = (byte)oneRGB.green;
-			byte oneBlue = (byte)oneRGB.blue;
 			byte[] line = new byte[(int)srcBpr];
 			for (int y=0; y<srcHeight; y++) {
 				C.memmove(line, data + (y * srcBpr), srcBpr);
@@ -466,16 +460,12 @@ private void createRepFromSourceAndApplyFlag(NSBitmapImageRep srcRep, int srcWid
 					int red = line[offset+redOffset] & 0xFF;
 					int green = line[offset+greenOffset] & 0xFF;
 					int blue = line[offset+blueOffset] & 0xFF;
-					int intensity = red * red + green * green + blue * blue;
-					if (intensity < 98304) {
-						line[offset+redOffset] = zeroRed;
-						line[offset+greenOffset] = zeroGreen;
-						line[offset+blueOffset] = zeroBlue;
-					} else {
-						line[offset+redOffset] = oneRed;
-						line[offset+greenOffset] = oneGreen;
-						line[offset+blueOffset] = oneBlue;
-					}
+					int alpha = line[offset+alphaOffset] & 0xFF;
+					RGBA result = DEFAULT_DISABLED_IMAGE_TRANSFORMER.adaptPixelValue(red, green, blue, alpha);
+					line[offset+redOffset] = (byte) result.rgb.red;
+					line[offset+greenOffset] = (byte) result.rgb.green;
+					line[offset+blueOffset] = (byte) result.rgb.blue;
+					line[offset+alphaOffset] = (byte) result.alpha;
 					offset += 4;
 				}
 				C.memmove(data + (y * srcBpr), line, srcBpr);
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/image/ImageColorTransformer.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/image/ImageColorTransformer.java
@@ -1,0 +1,73 @@
+/*******************************************************************************
+ * Copyright (c) 2025 Vector Informatik GmbH and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.swt.internal.image;
+
+import org.eclipse.swt.*;
+import org.eclipse.swt.graphics.*;
+
+public interface ImageColorTransformer {
+
+	public static final String IMAGE_DISABLEMENT_ALGORITHM_GRAYED = "grayed";
+	public static final String IMAGE_DISABLEMENT_ALGORITHM_GTK = "gtk";
+	public static final String IMAGE_DISABLEMENT_ALGORITHM_DESATURATED = "desaturated";
+	public static final String IMAGE_DISABLEMENT_ALGORITHM = System.getProperty("org.eclipse.swt.image.disablement",
+			IMAGE_DISABLEMENT_ALGORITHM_GRAYED).strip();
+
+	public static final ImageColorTransformer DEFAULT_DISABLED_IMAGE_TRANSFORMER = switch (IMAGE_DISABLEMENT_ALGORITHM) {
+	case IMAGE_DISABLEMENT_ALGORITHM_GTK -> ImageColorTransformer.forRGB(0.5f, 0.5f, 0.5f, 0.5f);
+	case IMAGE_DISABLEMENT_ALGORITHM_DESATURATED -> ImageColorTransformer.forHSB(1.0f, 0.2f, 0.9f, 0.5f);
+	default -> ImageColorTransformer.forGrayscaledContrastBrightness(0.2f, 2.9f);
+	};
+
+	RGBA adaptPixelValue(int red, int green, int blue, int alpha);
+
+	public static ImageColorTransformer forHSB(float hueFactor, float saturationFactor, float brightnessFactor,
+			float alphaFactor) {
+		return (red, green, blue, alpha) -> {
+			float[] hsba = new RGBA(red, green, blue, alpha).getHSBA();
+			float hue = Math.min(hueFactor * hsba[0], 1.0f);
+			float saturation = Math.min(saturationFactor * hsba[1], 1.0f);
+			float brightness = Math.min(brightnessFactor * hsba[2], 1.0f);
+			float alphaResult = Math.min(alphaFactor * hsba[3], 255.0f);
+			return new RGBA(hue, saturation, brightness, alphaResult);
+		};
+	}
+
+	public static ImageColorTransformer forRGB(float redFactor, float greenFactor, float blueFactor,
+			float alphaFactor) {
+		return (red, green, blue, alpha) -> {
+			int redResult = (int) Math.min(redFactor * red, 255.0f);
+			int greenResult = (int) Math.min(greenFactor * green, 255.0f);
+			int blueResult = (int) Math.min(blueFactor * blue, 255.0f);
+			int alphaResult = (int) Math.min(alphaFactor * alpha, 255.0f);
+			return new RGBA(redResult, greenResult, blueResult, alphaResult);
+		};
+	}
+
+	public static ImageColorTransformer forIntensityThreshold(Device device) {
+		RGBA lowIntensity = device.getSystemColor(SWT.COLOR_WIDGET_NORMAL_SHADOW).getRGBA();
+		RGBA highIntensity = device.getSystemColor(SWT.COLOR_WIDGET_BACKGROUND).getRGBA();
+		return (red, green, blue, alpha) -> {
+			int intensity = red * red + green * green + blue * blue;
+			RGBA usedGraytone = intensity < 98304 ? lowIntensity : highIntensity;
+			return new RGBA(usedGraytone.rgb.red, usedGraytone.rgb.green, usedGraytone.rgb.blue, alpha);
+		};
+	}
+
+	public static ImageColorTransformer forGrayscaledContrastBrightness(float contrast, float brightness) {
+		return (red, green, blue, alpha) -> {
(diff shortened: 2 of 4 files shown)


bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/image/HSBImageProcessingDescriptor.java
```diff
@@ -0,0 +1,39 @@
...
+	private final float saturationFactor;
+	private final float brightnessFactor;
+	private final float alphaFactor;
+
+	public HSBImageProcessingDescriptor(float hueFactor, float saturationFactor, float brightnessFactor, float alphaFactor) {
+		this.hueFactor = hueFactor;
+		this.saturationFactor = saturationFactor;
+		this.brightnessFactor = brightnessFactor;
+		this.alphaFactor = alphaFactor;
+	}
```
HannesWell, line 27: This and the other implementations of `ImageProcessingDescriptor` could probably be a record.

Or even more compact, you could add static factory methods to the ImageProcessingDescriptor interface, that just return lambda implementations, e.g.:
```
public interface ImageProcessingDescriptor {
   RGBA adaptPixelValue(RGBA original);
   
  static ImageProcessingDescriptor forHSB(float hueFactor, float saturationFactor, float brightnessFactor, float alphaFactor) {
    return RGBA original -> {
		float[] hsba = original.getHSBA();
		hsba[0] = Math.min(hueFactor * hsba[0], 1.0f);
		hsba[1] = Math.min(saturationFactor * hsba[1], 1.0f);
		hsba[2] = Math.min(brightnessFactor * hsba[2], 1.0f);
		hsba[3] = Math.min(alphaFactor * hsba[3], 255.0f);
		return new RGBA(hsba[0], hsba[1], hsba[2], hsba[3]);
    }
  }
 ...
}
```
or similar and alike for RGB and Pixel transformations.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/image/ImageProcessingDescriptor.java
```diff
@@ -0,0 +1,19 @@
...
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.swt.internal.image;
+
+import org.eclipse.swt.graphics.*;
+
+public interface ImageProcessingDescriptor {
```
HannesWell, line 15: Isn't this more something like a pixel-transformation or color-transformation?

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/image/ImageProcessingDescriptor.java
```diff
@@ -0,0 +1,19 @@
...
+
+	RGBA adaptPixelValue(RGBA original);
```
HannesWell, line 17: At least with the current callers it would save us the intermediate object creation if this method would accept the arguments `int red, int green, int blue, int alpha` directly.

HannesWell (commented):
Thanks for this PR, I really like the result of the proposed on-the-fly algorithm, i.e. the second version in your screenshots.

Below I have written some thoughts about the code.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/image/HSBImageProcessingDescriptor.java
HeikoKlare: Yeah, good idea. I reduced this PR from an original version also covering the SVG-preprocessing case (in which the interface had more methods), but since that would be something we add in the future and since this is not public API, let's keep it simple now and extend it on demand. By having it implemented under an interface, the algorithm can still be passed around as a strategy.

I adapted in my recent push accordingly.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/image/ImageProcessingDescriptor.java
HeikoKlare: That's true. I am not sure about this as the interface is intended to cover more (like SVG preprocessing), which would not be pixel transformation but rather a more general processing concept. Still, currently it is not, so the question is whether we should make the name more specific now and change it later on with a potential extension or directly indicate via the name that it is supposed to be extended for such a case.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/image/ImageProcessingDescriptor.java
HeikoKlare: Valid point. That reduces conversion efforts. I have adapted the PR accordingly.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/image/ImageProcessingDescriptor.java
HannesWell: Difficult question. But since it's internal and not widely used we could make it specific now for the current use-cases and change it later, when there is demand. Maybe things will go different directions than expected and then we would have unnecessarily generic code.

> I am not sure about this as the interface is intended to cover more (like SVG preprocessing), which would not be pixel transformation

Wouldn't this then require different methods with different signatures if more than just one pixel is transformed at the time? This could then be a second interface with a helper to wrap a pixel-wise transformator?

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/image/ImageProcessingDescriptor.java
```diff
@@ -0,0 +1,53 @@
...
+package org.eclipse.swt.internal.image;
+
+import org.eclipse.swt.*;
+import org.eclipse.swt.graphics.*;
+
+public interface ImageProcessingDescriptor {
+
+	RGBA adaptPixelValue(int red, int green, int blue, int alpha);
+
+	public static ImageProcessingDescriptor createHSBImageProcessingDescriptor(float hueFactor, float saturationFactor,
```
HannesWell, line 20: Just a nit-pick, but isn't mentioning the ImageProcessingDescriptor again in the method name somehow duplicated, if one assumes that this called with the interface as qualifier?
`ImageProcessingDescriptor.createHSBImageProcessingDescriptor`

If we make the method shorter it would also highlight the 'type' of transformation better, for example:
`ImageProcessingDescriptor.createHSB` `ImageProcessingDescriptor.forHSB` or `ImageProcessingDescriptor.ofHSB` or alike.

HeikoKlare: Also thought about that and then went for the non-qualified way, but I agree that using the qualified way with those short factory method names is better to avoid duplication within the qualifier. I adapted the PR accordingly.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/image/ImageProcessingDescriptor.java
```diff
@@ -0,0 +1,53 @@
+/*******************************************************************************
+ * Copyright (c) 2025 Vector Informatik GmbH and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
```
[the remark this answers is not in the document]
HeikoKlare: Since we have all information about authors in version control, we usually don't add these contributor information anymore. It's optional and I used to not add it for me :-)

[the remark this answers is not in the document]
HannesWell: OK, looks like I missed that.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
```diff
@@ -77,6 +77,8 @@
  */
 public final class Image extends Resource implements Drawable {
 
+	private static final boolean USE_LEGACY_IMAGE_DISABLEMENT = Boolean.getBoolean("org.eclipse.swt.image.useLegacyDisablementAlgorithm");
```
HannesWell, line 80: Should we make this property a bit shorter by just using the swt prefix (e.g. like for the `swt.autoScale` flag)? Or at least only `eclipse.swt...`?
```suggestion
	private static final boolean USE_LEGACY_IMAGE_DISABLEMENT = Boolean.getBoolean("swt.image.useLegacyDisablementAlgorithm");
```

HannesWell (commented):
As already written I really like the enhanced on-the-fly algorithm, i.e. the second version in your screenshots.
If I understand this PR correctly this is only applied for Windows and Mac. Wouldn't it make sense to use it on Linux as well for consistency?

Besides that this looks good generally. Thanks to everyone involved.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
HeikoKlare: My impression was that every "public" flag uses the `org.eclipse.swt` prefix (like `org.eclipse.swt.browser.DefaultType`, `org.eclipse.swt.browser.IEVersion`, `org.eclipse.swt.browser.EdgeDataDir`, ...), not mentioning the varios `org.eclipse.swt.internal.*` flags. So for the sake of consistency, I used the `org.eclipse.swt` prefix here as well. The autoscale properties in `DPIUtil` seem to be the only place where a shorter prefix is used.

Still, this is only the reasoning behind the current choice, but I have no strong opinion on this. If we agree, I will gladly change to a shorter prefix.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -667,12 +672,11 @@ void init() {
 }
 
 private ImageData applyDisableImageData(ImageData data, int height, int width) {
+	ImageProcessingDescriptor disabledImageProcessor = USE_LEGACY_IMAGE_DISABLEMENT
+			? forIntensityThreshold(device)
+			: forHSB(1.0f, 0.2f, 0.9f, 0.5f);
```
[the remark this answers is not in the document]
HeikoKlare: Thanks! That's also necessary for the sake of comprehensibility. I've update it accordingly.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
HannesWell: > My impression was that every "public" flag uses the org.eclipse.swt prefix

I think consistency is important so I'm fine with it, if that's the case. I just don't a full overview.

HeikoKlare:
> If I understand this PR correctly this is only applied for Windows and Mac. Wouldn't it make sense to use it on Linux as well for consistency?

We can of course do it for Linux as well. I didn't do this as Linux uses an adapted algorithm (fitting with what GTK does) anyway. But for the sake of being/becoming consistent again, we could unify the algorithm across all OSes, as we keep the previous algorithms in a legacy mode anyway.

HeikoKlare:
@HannesWell Can you please have another look at this?
I have now adapted the PR as we discussed: By default an image disablement algorithm will be used that conforms with the algorithm used for the existing pre-generated disabled icons. This will allow us to simply remove all the disabled icons and use the on-the-fly algorithm instead. Via a system property, the `enhanced` mode with the new proposed disabled icon styling can be activated and with `gtk` a mode conforming to the GTK-like disablement implemented for Linux will be used. These will make most sense when no using pre-generated disabled icons at all (i.e., if we introduced a mode in Platform UI that avoids the usage of explicit disabled icons), as otherwise there will be a mixture of differently styled disabled icons in the UI.

HannesWell (commented):
Thanks for the update. All what you have described makes sense to me and I'm generally fine with this change and look forward to it. I didn't exactly follow the algorithm changes in all three Image implementations, but I trust your change.

I just have a remark regarding the naming of the values.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/image/ImageColorTransformer.java
```diff
@@ -0,0 +1,72 @@
...
+ *******************************************************************************/
+package org.eclipse.swt.internal.image;
+
+import org.eclipse.swt.*;
+import org.eclipse.swt.graphics.*;
+
+public interface ImageColorTransformer {
+
+	public static final String IMAGE_DISABLEMENT_ALGORITHM = System.getProperty("org.eclipse.swt.image.disablement",
+			"default");
```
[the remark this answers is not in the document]
HeikoKlare: Done with b2ebdc5b19c9e04edd17279273bc945f8b3007e4

HeikoKlare:
All comments have been resolved. If no one objects on this, I plan to merge this by end of today.

HannesWell (approved):
Thank you, looks good to me.

HeikoKlare:
Thank you, @HannesWell!
I have just finally tested all configurations on all OSes and found that the existing algorithm for Linux is kind of wrong. More precisely, it ignores the fact that alpha values are also encoded in the RGB values. This led to wrong results, in particular making the icon background (which is supposed to be transparent) non-transparent. I have adapted the algorithm to properly reflect that fact. It then produces the expected results for all configurations also on Linux.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Image.java
```diff
@@ -307,15 +309,28 @@ public Image(Device device, Image srcImage, int flag) {
...
+						// The alpha value is embedded into the RGB values as well, so extract it out
+						// of those values for transformation and reapply it afterwards
+						if (hasAlpha && a != 0) {
+							r = r * 255 / a;
+							g = g * 255 / a;
+							b = b * 255 / a;
+						}
+						RGBA result = DEFAULT_DISABLED_IMAGE_TRANSFORMER.adaptPixelValue(r, g, b, a);
+						if (hasAlpha) {
+							result.rgb.red = result.rgb.red * result.alpha / 255;
```
HannesWell, line 326: Just a nit-pick: this and a few other places could be slightly more compact.
```suggestion
							result.rgb.red *= result.alpha / 255;
```

HannesWell (commented):
Sounds good. Thanks for finding it.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Image.java
HeikoKlare: Note that this is integer arithmetics and, if I am not mistaken, the proposal changes the execution order.
My understanding is that:
```
result.rgb.red = result.rgb.red * result.alpha / 255 === result.rgb.red = (result.rgb.red * result.alpha) / 255
```
And
```
result.rgb.red *= result.alpha / 255 === result.rgb.red = result.rgb.red * (result.alpha / 255)
```

The latter will always yield zero if alpha < 255.

@HannesWell Am I missing something?

HannesWell: > Note that this is integer arithmetics and, if I am not mistaken, the proposal changes the execution order.

Indeed, I think you are right. Please disregard my comment.

HeikoKlare:
There was a successful Jenkins build without any new isues for this PR yesterday: https://ci.eclipse.org/releng/job/eclipse.platform.swt/job/PR-1936/7/
And nothing but an added comment and rebases on master changed since back then.

Thus merging despite Jenkins build failing due to https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/5893

HeikoKlare:
Since Jenkins build failed for the PR: the subsequent master build after merging was successful has no new issues (https://ci.eclipse.org/releng/job/eclipse.platform.swt/job/master/1108/)

PR #1938, opened 2025-03-24
Suppress warnings for getImageDataAtCurrentZoom()

elsazac:
This PR adds suppression for deprecation warnings related to getImageDataAtCurrentZoom() in the ClipboardExample code.

### DIFF
--- a/examples/org.eclipse.swt.examples/src/org/eclipse/swt/examples/clipboard/ClipboardExample.java
+++ b/examples/org.eclipse.swt.examples/src/org/eclipse/swt/examples/clipboard/ClipboardExample.java
@@ -389,6 +389,7 @@ void createFileTransfer(Composite copyParent, Composite pasteParent){
 	}));
 }
 
+@SuppressWarnings("deprecation")
 void createImageTransfer(Composite copyParent, Composite pasteParent){
 	final Image[] copyImage = new Image[] {null};
 	Label l = new Label(copyParent, SWT.NONE);


elsazac:
I’d like to proceed with merging this if there are no objections

HeikoKlare (commented):
I am bit hestitant with respect to this change, as this is example code which is supposed to show how to use something, but using a deprecated method is obviously nothing one should do. On the other hand, I do currently not have a better solution (except passing `DPIUtil.getDeviceZoom()` to the ordinary `getImageData()`).

This seems to be one the last usage of the `getImageAtCurrentZoom()` method (except for a test method), so in my opinion we should rather try to get rid of it.
@akoch-yatta do you have an idea how to properly get rid of the `getImageDataAtCurrentZoom()` here?

sratz:
> I do currently not have a better solution (except passing `DPIUtil.getDeviceZoom()` to the ordinary `getImageData()`).

Isn't that the perfect solution for such an example?
- It avoids deprecated API
- It shows the complexity of high DPI handling
- It explains how to the the currents device's zoom using proper API

HeikoKlare:
There are two problems with that approach:
- `DPIUtil` is internal and not supposed to be used outside SWT. The whole auto-scaling concept is supposed to be SWT-internal.
- `DPIUtil.getDeviceZoom()` does not have a proper meaning once monitor-specific scaling is used (as now provided for Windows). You then have to consider the context-dependent zoom.

In this case, it's about retrieving image data in some zoom that did not apply any auto-scaling as via clipboard you cannot pass a proper image with according metadata but only the image data for some specific zoom, so they should have the best available quality. But you may have images that can provide image data in any zoom of best quality (like those rendered from SVGs). Since you don't know where data from a clipboard will be used, you can only use some data in a best-effort manner.

Maybe, in the case where you really want to paste some image data into the clipboard, the best would even be to avoid any auto-scaling and just use the data at 100%, so calling `getImageData(100)` or just `getImageData()`.

HeikoKlare:
This became obsolete as the usage of `getImageDataAtCurrentZoom()` has been replaced with:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2158

PR #1940, opened 2025-03-24
[Build] Rename library files if native sources of platform are unchanged

HannesWell:
Build only the SWT binaries whose native sources are really changed and just rename the binaries for all other platforms that are not changed.

Avoiding a rebuild of effectively unchanged binaries has multiple advantages, if only a sub-set of all supported platforms is build:
- reduced build times
- reduced occupation of precious native build-nodes in Jenkins
- increased stability of the build because if the native build-nodes for the skipped platforms are unavailable the build isn't blocked anymore.
- less occupation of the git large-file storage as renaming a binary file has basically zero-cost, while a modified binary (even if just slightly) will add the entire file again.

In order to ensure that native libraries can be renamed without recompilation, we first require:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/1939

This PR uses the same strategy as already used in Equinox:
https://github.com/eclipse-equinox/equinox/blob/master/Jenkinsfile

A visible difference in Jenkins is that, if one triggers a build explicitly, there are now three checkboxes to enforce the rebuild of natives for each supported OS (just like in Equinox).

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -42,8 +42,8 @@ def runOnNativeBuildAgent(String platform, Closure body) {
 	}
 }
 
-/** Returns the download URL of the JDK against whoose C headers (in the 'include/' folder) and native libraries the SWT natives are compiled.*/
-def getNativeJdkUrl(String os, String arch){ // To update the used JDK version update the URL template below
+/** Returns the download URL of the JDK against whose C headers (in the 'include/' folder) and native libraries the SWT natives are compiled.*/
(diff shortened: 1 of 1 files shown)


HannesWell:
All testing done was successful and the results are as expected. So this is ready for submission.

@eclipse-platform/eclipse-platform-committers please pay special attention to the build and not build (respectively just rename) of the native binaries in case you are changing the native sources of one or multiple platforms. Thanks in advance.

HeikoKlare:
Thank you, Hannes! Great improvement.

PR #1941, opened 2025-03-25
[GTK4] Fix gtk_widget_measure warning

akurtakov:
There is a flood of 
```
(SWT:64178): Gtk-WARNING **: 12:04:34.730: Trying to measure GtkHeaderBar 0x7fb13846ca30 for width of 0, but it needs at least 137
``` 
like messages when developing. The issue is that gtk_widget_measure is being called with for_size=0 param which happens to be below the minimum size for many widgets. The documentation clearly states "If no size is known, -1 can be passed." and so is done now.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Control.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Control.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2018 IBM Corporation and others.
+ * Copyright (c) 2000, 2025 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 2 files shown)


bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Control.java
```diff
@@ -855,14 +855,14 @@ Point computeNativeSize (long h, int wHint, int hHint, boolean changed) {
 		int [] natural_size = new int [1];
 		if (wHint == SWT.DEFAULT) {
 			if (GTK.GTK4) {
-				GTK4.gtk_widget_measure(h, GTK.GTK_ORIENTATION_HORIZONTAL, height, null, natural_size, null, null);
+				GTK4.gtk_widget_measure(h, GTK.GTK_ORIENTATION_HORIZONTAL, height==0?-1:height, null, natural_size, null, null);
```
laeubi, line 858:
```suggestion
				GTK4.gtk_widget_measure(h, GTK.GTK_ORIENTATION_HORIZONTAL, height>0?height:-1, null, natural_size, null, null);
```

same for `width`

PR #1943, opened 2025-03-25
[win32] Make Font#extractZoom device independent

akoch-yatta:
This PR ensures Font#extractZoom is device independent. The previous logic didn't provide any advantages anymore after the conversion between points and pixels was changed for Fonts. By using DPIUtil#getNativeZoom the most probable value for the zoom of a new font a chosen.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Font.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Font.java
@@ -67,7 +67,7 @@ public final class Font extends Resource {
  */
 Font(Device device) {
 	super(device);
-	this.zoom = extractZoom(this.device);
+	this.zoom = DPIUtil.getNativeDeviceZoom();
 	this.fontHeight = 0;
 }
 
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Font.java
```diff
@@ -296,9 +296,9 @@ public String toString () {
 
 private static int extractZoom(Device device) {
 	if (device == null) {
-		return DPIUtil.getNativeDeviceZoom();
+		System.out.println ("***WARNING: Initializing a font with a null device will use the default device and can lead to errors when this font is used with another device."); //$NON-NLS-1$
```
HeikoKlare, line 299: I think this will currently never be executed as the passed device is taken from `this.device`, which is always initialized by the fallback mechanism in `Resource`. So I think we need to change the callers of this method to use the device passed to the according method/constructors instead of using `this.device`, don't we?

akoch-yatta: Yes, that is of course correct. I had it done locally already, but seem to have forgotten to push that. Sorry for this one.

HeikoKlare (approved):
Reason for the change is sound and this avoids thread access errors when creating fonts with a wrong or null device (which was also gracefully accepted in the past).

Issue #1944, opened 2025-03-26
SWT tests not executed in I20250325-1800


iloveeclipse:
Most likely regression from https://github.com/eclipse-platform/eclipse.platform.swt/pull/1638

See https://download.eclipse.org/eclipse/downloads/drops4/I20250325-1800/testResults.php

The tests page says:

![Image](https://github.com/user-attachments/assets/3416474a-2637-4a5c-b8c7-1b1477ff7fb3)

The p2 director logs say

* https://download.eclipse.org/eclipse/downloads/drops4/I20250325-1800/testresults/ep436I-unit-linux-x86_64-java23_linux.gtk.x86_64_23/directorLogs/director-org.eclipse.swt.tests.log
* https://download.eclipse.org/eclipse/downloads/drops4/I20250325-1800/testresults/ep436I-unit-linux-x86_64-java21_linux.gtk.x86_64_21/directorLogs/director-org.eclipse.swt.tests.log

```
Installing org.eclipse.swt.tests 3.107.800.v20250325-0759.
Installing org.eclipse.test 3.6.200.v20241031-0735.
Installing org.eclipse.ant.optional.junit 3.3.400.v20230714-1309.
Installing org.eclipse.test.performance 3.21.100.v20250311-0829.
Installing org.eclipse.test.performance.win32 3.1.600.v20230714-1309.
Cannot complete the install because one or more required items could not be found.Installation failed.

!SESSION 2025-03-25 22:48:11.695 -----------------------------------------------
	Software being installed: Eclipse SWT Tests 3.107.800.v20250325-0759 (org.eclipse.swt.tests 3.107.800.v20250325-0759)eclipse.buildId=4.36.0.I20250325-1800
java.version=21.0.6
java.vendor=Eclipse Adoptium

BootLoader constants: OS=linux, ARCH=x86_64, WS=gtk, NL=en_US
	Missing requirement: Eclipse SWT Tests 3.107.800.v20250325-0759 (org.eclipse.swt.tests 3.107.800.v20250325-0759) requires 'org.eclipse.equinox.p2.iu; org.eclipse.swt.svg 0.0.0' but it could not be found
There were errors. See log file: /home/jenkins/agent/workspace/AutomatedTests/ep436I-unit-linux-x86_64-java21/workarea/I20250325-1800/eclipse-testing/test-eclipse/eclipse/configuration/1742942891803.log
[...]

PR #1946, opened 2025-03-26
Consume SVG rasterizer via capability in SWT tests bundle

HeikoKlare:
Currently, requiring an SVG rasterizer needs manual placement of an according fragment on the classpath. With this change, requiring SVG support can be defined explicitly via a required capability that is provided by the according SWT fragment.

Contributes to https://github.com/eclipse-platform/eclipse.platform.swt/issues/1944

### DIFF
--- a/bundles/org.eclipse.swt.svg/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.swt.svg/META-INF/MANIFEST.MF
@@ -11,3 +11,4 @@ Import-Package: com.github.weisj.jsvg;version="[1.7.0,2.0.0)",
  com.github.weisj.jsvg.geometry.size;version="[1.7.0,2.0.0)",
  com.github.weisj.jsvg.parser;version="[1.7.0,2.0.0)"
 Export-Package: org.eclipse.swt.svg
+Provide-Capability: eclipse.swt;image.format="svg";version:Version="1.0"
--- a/tests/org.eclipse.swt.tests/META-INF/MANIFEST.MF
+++ b/tests/org.eclipse.swt.tests/META-INF/MANIFEST.MF
@@ -20,3 +20,4 @@ Require-Bundle: org.junit;bundle-version="4.13.2",
 Eclipse-BundleShape: dir
 Bundle-RequiredExecutionEnvironment: JavaSE-17
 Automatic-Module-Name: org.eclipse.swt.tests
+Require-Capability: eclipse.swt;filter:="(image.format=svg)"
--- a/tests/org.eclipse.swt.tests/META-INF/p2.inf
+++ b/tests/org.eclipse.swt.tests/META-INF/p2.inf
@@ -1,4 +1,0 @@
-# pull in the applicable implementation fragment at build time (bug 461427)
-requires.0.namespace = org.eclipse.equinox.p2.iu
-requires.0.name = org.eclipse.swt.svg
-requires.0.range = 0.0.0


HeikoKlare:
@laeubi I am not sure whether this is the correct way of doing it, but at least it seems to work when locally executing the build.

I also tried with a "custom" capability (not using `osgi.extender`) but that did not work.

bundles/org.eclipse.swt.svg/META-INF/MANIFEST.MF
```diff
@@ -11,3 +11,4 @@ Import-Package: com.github.weisj.jsvg;version="[1.7.0,2.0.0)",
  com.github.weisj.jsvg.geometry.size;version="[1.7.0,2.0.0)",
  com.github.weisj.jsvg.parser;version="[1.7.0,2.0.0)"
 Export-Package: org.eclipse.swt.svg
+Provide-Capability: osgi.extender;osgi.extender="org.eclipse.swt.svgrasterizer"
```
laeubi, line 14:
```suggestion
Provide-Capability: swt.renderer;swt.renderer="svg"
```

tests/org.eclipse.swt.tests/META-INF/MANIFEST.MF
```diff
@@ -20,3 +20,4 @@ Require-Bundle: org.junit;bundle-version="4.13.2",
 Eclipse-BundleShape: dir
 Bundle-RequiredExecutionEnvironment: JavaSE-17
 Automatic-Module-Name: org.eclipse.swt.tests
+Require-Capability: osgi.extender;filter:="(osgi.extender=org.eclipse.swt.svgrasterizer)"
```
laeubi, line 23:
```suggestion
Require-Capability: swt.renderer;filter:="(swt.renderer=svg)"
```

laeubi (commented):
Im not sure if this is really an [extender](https://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.namespaces.html#service.namespaces-osgi.extender.namespace)

> An Extender is a bundle that uses the life cycle events from another bundle, the extendee, to extend that bundle's functionality when that bundle is active. It can use metadata (headers, or files inside the extendee) to control its functionality. 

I have added a suggestion how I would assume it should work to use a custom namespace, or maybe [osgi.implementation](https://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.namespaces.html#service.namespaces-osgi.implementation.namespace) would be more suitable

> The Implementation Namespace is intended to be used for:
>  Preventing a bundle from resolving if there is not at least one bundle that provides an implementation of the specified specification or contract.

then it might look similar to

```
Provide-Capability: osgi.implementation;
    osgi.implementation="swt.svg";
    version:Version="1.0";
    uses:="<packages that are used e.g. jsvg>"
```

HeikoKlare:
@laeubi are you fine with these changes now and would you expect them to fix https://github.com/eclipse-platform/eclipse.platform.swt/issues/1944?

In that case, I would propose to merge and maybe trigger a new I-Build to see if tests are running fine again.

bundles/org.eclipse.swt.svg/META-INF/MANIFEST.MF
```diff
@@ -11,3 +11,6 @@ Import-Package: com.github.weisj.jsvg;version="[1.7.0,2.0.0)",
  com.github.weisj.jsvg.geometry.size;version="[1.7.0,2.0.0)",
  com.github.weisj.jsvg.parser;version="[1.7.0,2.0.0)"
 Export-Package: org.eclipse.swt.svg
+Provide-Capability: osgi.implementation;
+    osgi.implementation="swt.svgrasterizer";
+    version:Version="1.0"
```
HannesWell, line 16: With the given reference I have the impression that the `osgi.implementation` namespace is inteded for implementations of OSGi specifications not arbitrary specs.
While it would probably not be harmful to use the osgi-namespace, I think a custom one would be more suitable, for example:
```suggestion
Provide-Capability: eclipse.swt.implementation;image.format="svg";version:Version="1.0"
```
Later we could then also use this namespace to introduce a requirement of o.e.swt to it's name fragments that could then provide `Provide-Capability: eclipse.swt.implementation;implementation="native"` or similar.

HannesWell (approved):
Thanks for the update, this is more compact and a bit more generic.

HeikoKlare:
Thank you all for your input! Hopefully this even fixes #1944.

PR #1947, opened 2025-03-26
Modernize Text JUnit tests

akurtakov:
Use assertThrows and remove dummy assertion messages as they give no valuable information.


HeikoKlare (approved):

akurtakov:
@HeikoKlare Do you see local failure in test_backspaceAndDelete on gtk3 ?

HeikoKlare:
I have executed that method in both the StyledText and Text test classes. They both succeed when excuted standalone and also when executed via the `AllNonBrowserTests` suite.
However, I currently don't have a native Linux system available and just executed it on WSL, so not that meaningful (though it's running GTK3 on Wayland).

PR #1948, opened 2025-03-26
Add getGcStyle() method to accommodate styling

ShahzaibIbrahim:
GC can be initialize with `GC(drawable)` or `GC(drawable, style)`. Currently `Image(Device, ImageGcDrawer, height, width)` doesn't consider `style` attribute while initializing `GC`. We need to create a new constructor for `Image` _e.g._ `Image(Device, ImageGcDrawer, height, width, style)` for which later `GC` inside `ImageGcDrawer` is initialized considering the style provided to it.

Now from implementation side, you can override `getGcStyle()` to provide styling in `GC` _e.g._ `LEFT_TO_RIGHT`, etc

### Consumer usage

Before this change, there was no way for consumer to provide styling using `ImageGcDrawer`. 

```java
final ImageGcDrawer imageGcDrawer = (gc, width, height) -> {
  gc.fillRectangle(0, 0, 16, 16);
};
 Image = new Image(Display.getDefault(), imageGcDrawer, 1, 1);
}
```
Now with this change. Consumer can define the styling with which `GC` will be initialized later in `ImageGCDrawerWrapper`. 

```java
final ImageGcDrawer imageGcDrawer = new ImageGcDrawer() {
  @Override
  public void drawOn(GC gc, int iWidth, int iHeight) {
    gc.fillRectangle(0, 0, iWidth, iHeight);
 }

  @Override
  public int getGcStyle() {
    return SWT.LEFT_TO_RIGHT;
  }
};
 Image = new Image(Display.getDefault(), imageGcDrawer, 1, 1);
}
```

## Required by
- https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2116

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/ImageGcDrawer.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/ImageGcDrawer.java
@@ -13,6 +13,8 @@
  *******************************************************************************/
 package org.eclipse.swt.graphics;
 
+import org.eclipse.swt.*;
+
 /**
  * Interface to provide a callback mechanism to draw on different GC instances
  * depending on the zoom the image will be used for. A common use case is when
@@ -45,4 +47,15 @@ public interface ImageGcDrawer {
 	default void postProcess(ImageData imageData) {
 	}
 
+	/**
+	 * Returns the GC style used when creating the GC instance. Default
+	 * implementation returns <code>SWT.NONE</code>.
+	 *
+	 * @return the GC style constant
+	 *
+	 * @since 3.130
+	 */
+	default int getGcStyle() {
+		return SWT.NONE;
+	}
 }
--- a/binaries/org.eclipse.swt.cocoa.macosx.aarch64/.settings/.api_filters
+++ b/binaries/org.eclipse.swt.cocoa.macosx.aarch64/.settings/.api_filters
@@ -232,6 +232,14 @@
             </message_arguments>
         </filter>
     </resource>
+	<resource path="Eclipse SWT/common/org/eclipse/swt/graphics/ImageGcDrawer.java" type="org.eclipse.swt.graphics.ImageGcDrawer">
+        <filter id="404000815">
+            <message_arguments>
+                <message_argument value="org.eclipse.swt.graphics.ImageGcDrawer"/>
+                <message_argument value="getGcStyle()"/>
+            </message_arguments>
+        </filter>
+    </resource>
     <resource path="Eclipse SWT Drag and Drop/common/org/eclipse/swt/dnd/DragSourceListener.java" type="org.eclipse.swt.dnd.DragSourceListener">
         <filter id="576720909">
             <message_arguments>
--- a/binaries/org.eclipse.swt.cocoa.macosx.x86_64/.settings/.api_filters
+++ b/binaries/org.eclipse.swt.cocoa.macosx.x86_64/.settings/.api_filters
@@ -232,6 +232,14 @@
             </message_arguments>
         </filter>
     </resource>
+	<resource path="Eclipse SWT/common/org/eclipse/swt/graphics/ImageGcDrawer.java" type="org.eclipse.swt.graphics.ImageGcDrawer">
+        <filter id="404000815">
+            <message_arguments>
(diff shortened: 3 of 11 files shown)


bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/ImageGcDrawer.java
```diff
@@ -45,4 +47,13 @@ public interface ImageGcDrawer {
 	default void postProcess(ImageData imageData) {
 	}
 
+	/**
+     * Returns the GC style used when creating the GC instance.
+     * Default implementation returns <code>SWT.NONE</code>.
+     *
+     * @return the GC style constant
+     */
+    default int getGcStyle() {
```
laeubi, line 56: Any reason to not call this simply 
```suggestion
    default int getStyle() {
```
like in other places in SWT as well?

ShahzaibIbrahim: The reason being, you can **only** set the style for gc with this method. This is the reason I thought to be descriptive with the name.

HeikoKlare (commented):
The change looks basically fine. Can you please:
- correct the formatting and
- add a consumer that demonstrates necessity for the extension?

akoch-yatta:
@ShahzaibIbrahim Is there a PR with an implementation that is overriding this method to utilize it?

fedejeanne:
> @ShahzaibIbrahim Is there a PR with an implementation that is overriding this method to utilize it?

https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2116/

HannesWell:
Sorry for commenting after the fact, but I think we could avoid the extra method, if we introduce a static factory that besides and ImageDrawer accepts the style and returns a special implementation. In the `Image` class we could then check for that special calls and get the style from it, for example:
```
public interface ImageGcDrawer {
...
	public static ImageGcDrawer withGCStyle(int style, ImageGcDrawer drawer) {
		return new ExtendedImageDrawer(drawer, style);
	}
}
```
This way one can still use a lambda if a style is to be set and we avoid an extra method on the interface itself (of course we get the factory for it).

If you think this would be better, I can create a follow-up PR for it.

fedejeanne:
@HannesWell thank you for the suggestion!
While I usually like factory methods and builders, I don't think that coding style matches the current style in SWT and I'm worried that it would result counterintuitive for consumers. I did a quick search for these methods names:
- `with*`
- `using*`
- `builder*`
... and came up (almost) empty

Is there any downside with the current approach or is it a matter of using newer coding practices?

_(Disclaimer: I would have even gone with a _builder_ pattern combined with lambdas so one could extend the returned instance by chaining calls together and providing new functionalities (in form of lambdas). But as I was writing this, the word "_overkill_" kept popping up in my head 😅)_

HeikoKlare:
> If you think this would be better, I can create a follow-up PR for it.

I generally like the idea. What I currently don't see is how we could realize the idea without type switches. The `ImageGcDrawer` is consumed by `Image`. There, a GC is created, for which the GC style is required. So `Image` either has to retrieve the style from the `ImageGcDrawer` (making it necessary to check for a specific subtype) or it has to defer the responsibility for creating the GC to the `ImageGcDrawer`, which can also only be provided in the specific extended subclass (making it necessary to check for a specific subtype as well). Do you see a solution for this, @HannesWell?

HannesWell:
> Is there any downside with the current approach or is it a matter of using newer coding practices?

It's just that this pattern remembers me about the existing typed Event interfaces, their adapter classes with empty implementation for easier sub-classing and the static `XYAdpter` factory methods added the the interfaces as last.

And with the current approach we somehow have the Adapter directly in the interface, while (at least from my understanding and experience) the `XYAdpter` factory methods are usually more convenient to use (and more modern). Therefore I was thinking about using the latter style here again.

As a picture (or code) speaks more than a thousands words I have now added these ideas in https://github.com/eclipse-platform/eclipse.platform.swt/pull/1985#issuecomment-2776675817.

> > If you think this would be better, I can create a follow-up PR for it.
> 
> I generally like the idea. What I currently don't see is how we could realize the idea without type switches.

If I understand you right, that's exactly how this is solved. Within the image class it's checked if the special (SWT internal) type is passed. But lets discuss that in https://github.com/eclipse-platform/eclipse.platform.swt/pull/1985 :)

PR #1949, opened 2025-03-26
[GTK3] Override gtk_im_module only if it was set to xim

akurtakov:
From https://github.com/ibus/ibus/issues/2331#issuecomment-927110685 "There are two modes to communicate between IBus and GTK applications."
- GTK_IM_MODULE=wayland (composer mode - GNOME Wayland default) and GTK_IM_MODULE=ibus (no composer mode).

Note that "wayland" im module is another incarnation of ibus. The only thing that was broken for sure is "xim" so override only it under Gnome on X11 to prevent making SWT(and as a result Eclipse) use non-default IM modules and thus exposing different behaviour than other apps.


akurtakov:
Windows problems are irrelevant for this change. Merging.

Issue #1865, opened 2025-02-27
Native compilation error on Fedora Rawhide (42)


deamn:
**Describe the bug**
I try to compile native part on Fedora Rawhide and I have the following error. The same on Fedora 41 works well.
[exec] os.c: In function ‘Java_org_eclipse_swt_internal_gtk_OS_Call__JJII’:
     [exec] os.c:10106:21: error: too many arguments to function ‘(jint (*)(void))arg0’; expected 0, have 3
     [exec] 10106 |         rc = (jint)((jint (*)())arg0)(arg1, arg2, arg3);
     [exec]       |                    ~^~~~~~~~~~~~~~~~~ ~~~~
     [exec] os.c: In function ‘Java_org_eclipse_swt_internal_gtk_OS_Call__JJJ’:
     [exec] os.c:10118:21: error: too many arguments to function ‘(jint (*)(void))arg0’; expected 0, have 2
     [exec] 10118 |         rc = (jint)((jint (*)())arg0)(arg1, arg2);
     [exec]       |                    ~^~~~~~~~~~~~~~~~~ ~~~~
     [exec] os.c: In function ‘Java_org_eclipse_swt_internal_gtk_OS_XCheckIfEvent’:
     [exec] os.c:10512:71: error: passing argument 3 of ‘XCheckIfEvent’ from incompatible pointer type [-Wincompatible-pointer-types]
     [exec] 10512 |         rc = (jboolean)XCheckIfEvent((Display *)arg0, (XEvent *)arg1, (Bool (*)())arg2, (XPointer)arg3);
     [exec]       |                                                                       ^~~~~~~~~~~~~~~~
     [exec]       |                                                                       |
     [exec]       |                                                                       int (*)(void)
     [exec] In file included from /usr/include/gtk-3.0/gdk/gdkx.h:30,
     [exec]                  from /usr/include/gtk-3.0/gtk/gtksocket.h:34,
     [exec]                  from /usr/include/gtk-3.0/gtk/gtkx.h:29,
     [exec]                  from os.h:60,
     [exec]                  from os_structs.h:19,
     [exec]                  from os.c:20:
     [exec] /usr/include/X11/Xlib.h:2105:5: note: expected ‘int (*)(Display *, XEvent *, char *)’ but argument is of type ‘int (*)(void)’
     [exec]  2105 |     Bool (*) (
     [exec]       |     ^
[...]

PR #1950, opened 2025-03-27
Compile as C17 code

akurtakov:
GCC 15 defaults to C23 standard but it's not feasible to use it as even Xlib.h is not compatible yet e.g.
```
/usr/include/X11/Xlib.h:2105:5: note: expected ‘int (*)(Display *,
XEvent *, char *)’ but argument is of type ‘int (*)(void)’
 2105 |     Bool (*) (
```
Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/1865 .


akurtakov:
Jenkins is fine. Windows problem is unrelated to Linux change. Merging.

akurtakov:
@HeikoKlare What is the plan for this long standing Windows test problem?

Issue #1710, opened 2025-01-07
Cannot set cookies in Browser component with libwebkit2gtk 4.1 in Ubuntu 24.04 due to libsoup 3 incompatibility


miscose:
**Describe the bug**
When running SWT on Ubuntu 24.04 LTS with libwebkit2gtk 4.1 (default), `Browser.setCookie` always fails (returns false). It works with  libwebkit2gtk 4.0 on earlier Ubuntu versions. This seems to be due to libwebkit2gtk 4.1 upgrading to libsoup 3 and SWT only supporting libsoup 2.

**To Reproduce**
1. Install "Eclipse IDE for RCP and RAP Developers" 2024-12 on Ubuntu 24.04 LTS
2. Run this SWT snippet in a Java project: (with swt.jar from swt-4.34-gtk-linux-x86_64.zip on the classpath)
```
package setcookietest;

import org.eclipse.swt.SWT;
import org.eclipse.swt.browser.Browser;
import org.eclipse.swt.browser.ProgressListener;
import org.eclipse.swt.layout.FillLayout;
import org.eclipse.swt.widgets.Display;
import org.eclipse.swt.widgets.Shell;

public class SetCookieTest {

	public static void main(String[] args) throws Exception {
		Display display = new Display();
		Shell shell = new Shell(display);
		shell.setLayout(new FillLayout());
		Browser browser = new Browser(shell, SWT.WEBKIT);
		browser.setUrl("http://www.eclipse.org/swt");
		browser.addProgressListener(ProgressListener.completedAdapter(_ -> {
			boolean success = Browser.setCookie("cookie1=value1", "http://www.eclipse.org/swt");
			System.out.println("setCookie: " + success);
			System.exit(success ? 0 : 1);
		}));
		
		shell.open();
		while (!shell.isDisposed()) {
			if (!display.readAndDispatch())
				display.sleep();
		}
		display.dispose();
	}

}
```
3. Note that you get this output:
```
SWT WebKit: SoupURI == 0 when setting cookie
setCookie: false
```
4. Try the same on Ubuntu 20.04 LTS and get this output:
```
setCookie: true
```

**Expected behavior**
The same behavior on Ubuntu 24.04 LTS as on Ubuntu 20.04 LTS.

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [ ] Windows
- - [x] Linux
- - [ ] macOS 

[...]

PR #1951, opened 2025-03-27
Fix Browser.setCookie on WebkitGtk/Soup3

akurtakov:
Soup 3.x switched away from its own SoupURI to Glib's GUri (see https://libsoup.gnome.org/libsoup-3.0/migrating-from-libsoup-2.html#uri-type-changed ). Thus g_uri_parse have to be used instead of soup_uri_new.


akurtakov:
Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/1710 .

PR #1952, opened 2025-03-27
Restore equality for all kinds of Point/Rectangle

HeikoKlare:
The MonitorAwarePoint and MonitorAwareRectangle classes have been introduced as specializations of the existing Point and Rectangle classes solely to support monitor-specific scaling and only used in the context of that feature. They are not supposed to be used outside of SWT but just an internal construct. However, instances of those classes reach consumers of SWT API, e.g., via the Shell class. This broke equality assumptions for consumers, as, e.g., a comparison between a Point and a MonitorAwarePoint (returned by methods of Shell) are never equal, even though the consumers is not supposed to even know that a MonitorAwarePoint can be returned here.

To restore equality, this change aligns equality of the monitor-aware implementations of point and rectangle with their original implementations. In consequence, a Point and a MonitorAwarePoint will be considered equal if their x/y coordinates match. The monitor stored in a MonitorAwarePoint will be ignored. This analogously applies to MonitorAwareRectangle. Since those classes are only used internally, no one may rely on any other kind of equality specification for the monitor-aware implementations of point and rectangle, but existing equality for Points and Rectangles in general is restored.

__Note:__ `MonitorAwarePoint`/`MonitorAwareRectangle` are not part of public API and are currently only used when monitor-specific scaling on Windows is active, which is marked as an experimental feature. This is why no one currently relies on specific equality handling of those added types, allowing to correct their equality relation.

See also https://github.com/eclipse-platform/eclipse.platform.swt/pull/1711#issuecomment-2722324061 and other comments in that PR.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/MonitorAwarePoint.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/MonitorAwarePoint.java
@@ -13,8 +13,6 @@
  *******************************************************************************/
 package org.eclipse.swt.graphics;
 
-import java.util.*;
-
 import org.eclipse.swt.widgets.*;
 
 /**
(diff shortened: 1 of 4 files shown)


HeikoKlare:
@amartya4256 please have a look when you have time since you originally introduced those classes

@nedtwigg for your information

akoch-yatta (approved):
Changes look good to me and the explanation makes sense

PR #1955, opened 2025-03-27
Improve signature of methods in InternalImageLoader

HeikoKlare:
The InternalImageLoader is currently placed in the graphics package and marked as package internal. This change moves it to the internal package improves its name to NativeImageLoader. It also adapts its method's signatures to properly combine element and zoom information if appropriate.


HannesWell (approved):
Looks good. Thank you.

PR #1956, opened 2025-03-27
Fix NPE #1929

### LINKED ISSUE #1929: NPE in DirectoryDialog
**Describe the bug**
```
NullPointerException: Cannot invoke "java.util.Optional.orElse(Object)" because the return value of "org.eclipse.swt.widgets.DirectoryDialog.openDialog()" is null
    at org.eclipse.swt.widgets.DirectoryDialog.open(DirectoryDialog.java:170)
```

**To Reproduce**
I received a P2 update prompt, agreed to download updates and immediately went to import project in a new PDE workspace. Eclipse went into a lock for approximately 10 minutes, meanwhile I've recovered a log

[eclipse.txt](https://github.com/user-attachments/files/19409465/eclipse.txt)

**Expected behavior**
No NPE

**Screenshots**

**Environment:**
- - [] All OS
- - [ ] Windows
- - [ ] Linux
- - [x] macOS 

[installation.txt](https://github.com/user-attachments/files/19409466/installation.txt)

3. JRE/JDK version
 java.vendor.version=Temurin-23.0.1+11

**Version since**
SWT version 3.129

Wait long enough

basilevs:
This is a fixup for ab1d11972aab1c75149f271dca4b58ccfd7710a7.
The contract of `openDialog()` is to throw on error, and return an `Optional` on user action. Null return value is not expected.

This does not fix the issue #1929, but fixes the error reporting per #1026

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/DirectoryDialog.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/DirectoryDialog.java
@@ -195,7 +195,7 @@ public Optional<String> openDialog () {
 	directoryPath = null;
 	panel = NSOpenPanel.openPanel();
 	if (panel == null) {
-		return null;
+		throw new SWTException(SWT.ERROR_INVALID_RETURN_VALUE);
 	}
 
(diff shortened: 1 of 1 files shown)


basilevs:
@HeikoKlare hello, could you please take a look?

HeikoKlare (approved):
Totally makes sense to me. Thanks for the catch and for fixing it!

PR #1968, opened 2025-03-31
Remove superfluous implementations of hashCode/equals

fedejeanne:
The classes are marked `final` and the methods only call `super`. No need for them.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/MonitorAwarePoint.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/MonitorAwarePoint.java
@@ -48,15 +48,4 @@ public final class MonitorAwarePoint extends Point {
 	public Monitor getMonitor() {
 		return monitor;
 	}
-
-	@Override
-	public boolean equals(Object object) {
-		return super.equals(object);
-	}
-
-	@Override
-	public int hashCode() {
-		return super.hashCode();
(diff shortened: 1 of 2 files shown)


fedejeanne:
I drafted this PR because the Git bot says it removes/skips some tests, which is odd.

image

Since the PR is trivial (just a "nit"), it can wait until the build behaves normally.

HeikoKlare:
This change would require adaptation of API tooling. We had the same issue with `Region` (or some other resource class): `equals`/`hashcode` are part of public API, but API tooling does not seem to correctly consider they are still available (implicitly) vi `Object`. That's why you cannot remove them from any class once added. You can add an API filter, but that will result in errors as well.

tomaswolf:
Usually one uses this pattern to document that the additional fields (`monitor` in this case) do not participate in equality tests.

HeikoKlare:
> Usually one uses this pattern to document that the additional fields (monitor in this case) do not participate in equality tests.

Didn't think about that yet, but that argument applies quite well here as explicitly not making the monitor field participate in equality calculation here was the motivation for the recent changing removing the implementation of equals and hashcode. Making this explicit with an "empty" implementation of those methods sounds reasonable.

fedejeanne:
> Usually one uses this pattern to document that the additional fields (`monitor` in this case) do not participate in equality tests.

So... "_it's not a bug, it's a feature (of the tooling)_"? 

I can live with that. Thank you both for the hints.

PR #1969, opened 2025-03-31
[win32] use fixed zoom for default system font

akoch-yatta:
This commit addresses a regression that is caused by DPIUtil#getDeviceZoom not properly initialized yet, when created the first system font. Explicitly passing the target zoom to the Font prevent the issue.

### How to reproduce
Start a runtime on a zoom > 100 with default configuration (if not visible try a fresh workspace). The CTabFolder fonts will be bigger as expected


HeikoKlare (approved):
Change is reasonable and it fixes the described issue.

tobiasmelcher:
Thanks a lot for the fix. Fix works on my side.

fedejeanne:
Ignoring unrelated check failure

PR #1970, opened 2025-04-01
docs: replaced the link in the documentation

braveocheretovych:
In my opinion, the new link fits here better.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ The best known SWT-based application is [Eclipse](https://www.eclipse.org).
 ## Getting Started
 
 SWT comes with platform-specific jar files.
-Download them from https://www.eclipse.org/swt and add the jar file to your classpath.
+Download them from https://eclipse.dev/eclipse/swt and add the jar file to your classpath.
 
 ### Example
 ![Example](example.png)
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ The best known SWT-based application is [Eclipse](https://www.eclipse.org).
 ## Getting Started
 
 SWT comes with platform-specific jar files.
-Download them from https://download.eclipse.org/eclipse/downloads/index.html and add the jar file to your classpath.
+Download them from https://www.eclipse.org/swt and add the jar file to your classpath.
 
 ### Example
 ![Example](example.png)
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ The best known SWT-based application is [Eclipse](https://www.eclipse.org).
 ## Getting Started
 
 SWT comes with platform-specific jar files.
-Download them from https://eclipse.dev/eclipse/swt and add the jar file to your classpath.
+Download them from https://download.eclipse.org/eclipse/downloads/drops4/S-4.36M1-202504031800/ and add the jar file to your classpath.
 
 ### Example
 ![Example](example.png)


merks:
Could we hold off on this until the dust settles on this issue?

https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/5805

We really need to the documentation to actually be updated when the content is modified here:

https://github.com/eclipse-platform/www.eclipse.org-swt

braveocheretovych:
I understand, tell me, should I close this request pool or keep it open?

braveocheretovych:
Yes, I understand you, no problem, I will wait for your answer.

merks:
@braveocheretovych

The final redirected location is the following:

https://eclipse.dev/eclipse/swt

merks:
Sorry, sorry, only now do I read the whole context.   😱 

These are instructions for where to get downloads not where to get documentation.  That existing page is the index.  Unfortunately the documentation is rather helpful because it doesn't explain that one needs to pick a specific version, e.g., 

https://download.eclipse.org/eclipse/downloads/drops4/S-4.36M1-202504031800/

And then get an SWT zip from the list:

image

braveocheretovych:
It's okay, I completely agree with you, please check again.

braveocheretovych:
Maybe for convenience add #SWT at the end of the link, what do you think?https://download.eclipse.org/eclipse/downloads/drops4/S-4.36M1-202504031800/#SWT

merks:
The problem is that that link will be out-of-date in a few weeks and will be a broken link within 3 months.  I really think the documentation must specify a location that remains valid long term. 

Of course the user then ends up needing to hunt. So better to describe what is on that root download page and describe what’s on the subsequent page.

braveocheretovych:
Sorry, I'm communicating with you through a translator, there are some translation difficulties. I made a draft, check it out, did I understand everything correctly?

[SWT](https://eclipse.dev/eclipse/swt) постачається з файлами баночок, характерними для платформи. Завантажте їх із https://download.eclipse.org/eclipse/downloads/drops4/S-4.36M1-202504031800/ і додайте файл баночки до свого класного шляху.

braveocheretovych:
Hey @merks. Please help me solve this question

merks:
The summary is that I don't think can fix this just by changing the URL.  You need to describe stuff in English and if that's a challenge this is probably not an appropriate task to take on...

braveocheretovych:
Getting Started
SWT comes with platform-specific jar files. Download the latest version of SWT from the [official Eclipse SWT page](https://www.eclipse.org/swt/) or directly from the [Eclipse downloads page](https://download.eclipse.org/eclipse/downloads/drops4/S-4.36M1-202504031800). Add the appropriate platform-specific jar file to your classpath.

merks:
@HannesWell 

I think we have a bit of a language barrier here and I'm not sure that it makes sense to contribute language-specific documentation in the face of such a language barrier...

braveocheretovych:
yes, I agree with you, I'd better close this pull request then?

braveocheretovych:
Last question for you, what do you think, is there a need to add this icon here?
https://img.shields.io/pypi/pyversions/eclipse?logo=pypi

merks:
I don’t know how those badges work. Maybe @HannesWell has more knowledge.

HannesWell:
> I think we have a bit of a language barrier here and I'm not sure that it makes sense to contribute language-specific documentation in the face of such a language barrier...

That makes it indeed very difficult.

But besides that @merks is right with his statement that this link will not last long. So we should either describe verbally how to navigate to the desired section or establish some kind of latest redirection link that points to the latest. But I think the former is easier to implement.
But the current state isn't suitable.

> yes, I agree with you, I'd better close this pull request then?

OK, then lets close this for now. If you are interested to implement the above suggestion, feel free to let us now so that we can reopen this or create a new PR.
Still, thank you for your effort trying to make SWT's documentation better.

braveocheretovych:
Thank you for the kind words, I'm sorry I couldn't solve this problem on my own.

Issue #1919, opened 2025-03-19
Edge: Operation timed out on evaluate()


fipro78:
**Describe the bug**
I started an application that contains a `RichTextEditor` and get an error on startup:
```
org.eclipse.swt.SWTException: Waiting for Edge operation to terminate timed out
	at org.eclipse.swt.browser.Edge.createTimeOutException(Edge.java:525)
	at org.eclipse.swt.browser.Edge.processOSMessagesUntil(Edge.java:520)
	at org.eclipse.swt.browser.Edge.callAndWait(Edge.java:287)
	at org.eclipse.swt.browser.Edge.evaluate(Edge.java:902)
	at org.eclipse.swt.browser.WebBrowser.evaluate(WebBrowser.java:405)
	at org.eclipse.swt.browser.Browser.evaluate(Browser.java:665)
	at org.eclipse.swt.browser.Browser.evaluate(Browser.java:614)
	at org.eclipse.nebula.widgets.richtext.RichTextEditor$3.completed(RichTextEditor.java:284)
	at org.eclipse.swt.browser.Edge.lambda$45(Edge.java:1109)
	at org.eclipse.swt.widgets.RunnableLock.run(RunnableLock.java:40)
	at org.eclipse.swt.widgets.Synchronizer.runAsyncMessages(Synchronizer.java:132)
	at org.eclipse.swt.widgets.Display.runAsyncMessages(Display.java:4094)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:3710)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine$5.run(PartRenderingEngine.java:1151)
	at org.eclipse.core.databinding.observable.Realm.runWithDefault(Realm.java:339)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine.run(PartRenderingEngine.java:1042)
	at org.eclipse.e4.ui.internal.workbench.E4Workbench.createAndRunUI(E4Workbench.java:153)
	at org.eclipse.e4.ui.internal.workbench.swt.E4Application.start(E4Application.java:165)
	at org.eclipse.equinox.internal.app.EclipseAppHandle.run(EclipseAppHandle.java:208)
	at org.eclipse.core.runtime.internal.adaptor.EclipseAppLauncher.runApplication(EclipseAppLauncher.java:149)
[...]

PR #1973, opened 2025-04-01
Edge:evaluate deadlock fix

amartya4256:
This PR cotributes to replicating the behaviour of Edge:evaluate as
it is in WebKit - it must not wait for the execution of script to obtain
the result, in case evaluate is called inside a WebView callback. This
PR also makes sure that OpenWindowListeners are execute
synchronously or asynchronously depending on if the
handleNewWindowRequested is called from the evaluate script. Moreover,
this enables all the tests which were failing because of Edge:evaluate
limitations.

contributes to
https://github.com/eclipse-platform/eclipse.platform.swt/issues/1771 and
https://github.com/eclipse-platform/eclipse.platform.swt/issues/1919

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
@@ -81,6 +81,7 @@ class Edge extends WebBrowser {
 
 	static boolean inCallback;
 	boolean inNewWindow;
+	private boolean inEvaluate;
 	HashMap<Long, LocationEvent> navigations = new HashMap<>();
 	private boolean ignoreGotFocus;
 	private boolean ignoreFocusIn;
@@ -906,11 +907,21 @@ public Object evaluate(String script) throws SWTException {
 	// Feature in WebView2. ExecuteScript works regardless of IsScriptEnabled setting.
 	// Disallow programmatic execution manually.
 	if (!jsEnabled) return null;
-
+	if(inCallback) {
+		// Execute script, but do not wait for async call to complete as otherwise it
+		// can cause a deadlock if execute inside a WebView callback.
+		execute(script);
+		return null;
+	}
 	String script2 = "(function() {try { " + script + " } catch (e) { return '" + ERROR_ID + "' + e.message; } })();\0";
 	String[] pJson = new String[1];
-	int hr = callAndWait(pJson, completion -> webViewProvider.getWebView(true).ExecuteScript(script2.toCharArray(), completion));
-	if (hr != COM.S_OK) error(SWT.ERROR_FAILED_EVALUATE, hr);
+	inEvaluate = true;
+	try {
+		int hr = callAndWait(pJson, completion -> webViewProvider.getWebView(true).ExecuteScript(script2.toCharArray(), completion));
+		if (hr != COM.S_OK) error(SWT.ERROR_FAILED_EVALUATE, hr);
+	} finally {
+		inEvaluate = false;
+	}
 
 	Object data = JSON.parse(pJson[0]);
 	if (data instanceof String && ((String) data).startsWith(ERROR_ID)) {
@@ -1319,7 +1330,7 @@ int handleNewWindowRequested(long pView, long pArgs) {
 	args.GetDeferral(ppv);
 	ICoreWebView2Deferral deferral = new ICoreWebView2Deferral(ppv[0]);
 	inNewWindow = true;
-	browser.getDisplay().asyncExec(() -> {
+	Runnable openWindowHandler = () -> {
 		try {
 			if (browser.isDisposed()) return;
 			WindowEvent openEvent = new WindowEvent(browser);
@@ -1355,7 +1366,21 @@ int handleNewWindowRequested(long pView, long pArgs) {
 			args.Release();
 			inNewWindow = false;
 		}
-	});
+	};
+
+	// Creating a new browser instance within the same environment from inside the OpenWindowListener of another browser
+	// can lead to a deadlock. To prevent this, handlers should typically run asynchronously.
+	// However, if a new window is opened using `evaluate(window.open)`, running the handler asynchronously in that context
+	// may also result in a deadlock.
+	// Therefore, whether the listener runs synchronously or asynchronously should depend on the `inEvaluate` condition.
+	// That said, combining both situations—opening a window via `evaluate` and launching a new browser inside the OpenWindowListener—
+	// should be avoided altogether, as it significantly increases the risk of deadlocks.
+	if (inEvaluate) {
+		openWindowHandler.run();
+	} else {
+		browser.getDisplay().asyncExec(openWindowHandler);
+	}
+
 	return COM.S_OK;
 }
 
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
@@ -745,7 +745,6 @@ public void test_LocationListener_ProgressListener_cancledLoad () {
 
 @Test
 public void test_LocationListener_LocationListener_ordered_changing () {
-	assumeFalse("Currently broken for Edge", isEdge);
 	List<String> locations = Collections.synchronizedList(new ArrayList<>());
 	browser.addLocationListener(changingAdapter(event -> {
 		locations.add(event.location);
@@ -1578,8 +1577,6 @@ public void test_setJavascriptEnabled_multipleInstances() {
 */
 @Test
 public void test_LocationListener_evaluateInCallback() {
-	assumeFalse("behavior is not (yet) supported by Edge browser", isEdge);
-
 	AtomicBoolean changingFinished = new AtomicBoolean(false);
 	AtomicBoolean changedFinished = new AtomicBoolean(false);
 	browser.addLocationListener(new LocationListener() {
@@ -1628,8 +1625,6 @@ public void test_LocationListener_evaluateInCallback() {
 /** Verify that evaluation works inside an OpenWindowListener */
 @Test
 public void test_OpenWindowListener_evaluateInCallback() {
-	assumeFalse("behavior is not (yet) supported by Edge browser", isEdge);
-
 	AtomicBoolean eventFired = new AtomicBoolean(false);
 	browser.addOpenWindowListener(event -> {
 		browser.evaluate("SWTopenListener = true");
@@ -2189,7 +2184,6 @@ ProgressListener callCustomFunctionUponLoad = completedAdapter(event ->	browser.
  */
 @Test
 public void test_BrowserFunction_callback () {
-	assumeFalse("Currently broken for Edge", isEdge);
 	AtomicBoolean javaCallbackExecuted = new AtomicBoolean(false);
 
 	class JavascriptCallback extends BrowserFunction { // Note: Local class defined inside method.


bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -1332,30 +1364,24 @@ int handleNewWindowRequested(long pView, long pArgs) {
...
-					updateWindowFeatures(args, showEvent);
-					for (VisibilityWindowListener showListener : other.visibilityWindowListeners) {
-						showListener.show(showEvent);
-						if (other.browser.isDisposed()) return;
-					}
+					updateWindowFeatures(windowFeatures, showEvent);
+						for (VisibilityWindowListener showListener : other.visibilityWindowListeners) {
+							showListener.show(showEvent);
+							if (other.browser.isDisposed()) return;
+						}
```
fedejeanne, line 1377: The indentation is wrong here

fedejeanne (commented):
I think this PR does 2 things, right?
1. It changes the strategy about _how_ to execute callbacks by introducing the method `executeAsynchronously(...)`
2. It executes some new blocks of code asynchronously by using the new method

Would it be possible to split it in 2 different PRs? That way it would be possible to see for example which part of these changes (1. or 2.) fixes the tests and which part makes it easier to work with the code.

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -81,6 +81,7 @@ public WebViewEnvironment(ICoreWebView2Environment environment) {
 
 	static boolean inCallback;
 	boolean inNewWindow;
+	private boolean inEvaluate = false;
```
fedejeanne, line 84:
```suggestion
	private boolean inEvaluate;
```

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -906,10 +907,19 @@ public Object evaluate(String script) throws SWTException {
 	// Feature in WebView2. ExecuteScript works regardless of IsScriptEnabled setting.
 	// Disallow programmatic execution manually.
 	if (!jsEnabled) return null;
-
+	if(inCallback) {
```
fedejeanne, line 910: Please add a comment explaining why you need this whole `if`, just like you did for `if (inEvaluate)` in the other method.

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -906,10 +907,19 @@ public Object evaluate(String script) throws SWTException {
...
+		execute(script);
+		return null;
+	}
+	inEvaluate = true;
```
fedejeanne, line 914: Move this line closer to its usage, right after line 916: `String[] pJson = new String[1];`

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -906,10 +907,19 @@ public Object evaluate(String script) throws SWTException {
...
 	String script2 = "(function() {try { " + script + " } catch (e) { return '" + ERROR_ID + "' + e.message; } })();\0";
 	String[] pJson = new String[1];
-	int hr = callAndWait(pJson, completion -> webViewProvider.getWebView(true).ExecuteScript(script2.toCharArray(), completion));
+	int hr;
+	try {
+		hr = callAndWait(pJson, completion -> webViewProvider.getWebView(true).ExecuteScript(script2.toCharArray(), completion));
+	} finally {
+		inEvaluate = false;
+	}
 	if (hr != COM.S_OK) error(SWT.ERROR_FAILED_EVALUATE, hr);
```
fedejeanne, line 923: Since you don't need `hr` outside, you can put everything inside the `try` block
```suggestion
	try {
		int hr = callAndWait(pJson, completion -> webViewProvider.getWebView(true).ExecuteScript(script2.toCharArray(), completion));
	if (hr != COM.S_OK) error(SWT.ERROR_FAILED_EVALUATE, hr);
	} finally {
		inEvaluate = false;
	}
```

fedejeanne (commented):
I made an initial review, just a few minor comments regarding coding practices. I will test it later today, after these comments are addressed.

In general, it looks good. Thank you @amartya4256 !

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
fedejeanne: Done in ce625749c181e36ff209703684cf01e9695a917d

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
fedejeanne: Done in ce625749c181e36ff209703684cf01e9695a917d

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
fedejeanne: Done in ce625749c181e36ff209703684cf01e9695a917d

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -1355,7 +1364,21 @@ int handleNewWindowRequested(long pView, long pArgs) {
...
+	// asynchronously. And in case a new browser is initialized in the same
+	// environment inside an OpenWindowListener of another browser, there can also
+	// be a deadlock. Hence, generally it should be run asynchronously. However, a
+	// combination of opening a new window from evaluate and starting a new instance
+	// of browser inside the OpenWindowListener should be avoided.
+	if (inEvaluate) {
+		openWindowHandler.run();
+	} else {
+		browser.getDisplay().asyncExec(openWindowHandler);
+	}
```
fedejeanne, line 1380: I would move the code to a new private method and make use of the JavaDoc instead of the comment (`//`). That way you can use `<pre>`-blocks in order to add example (javascript) code, which is easier to understand than just words.

Something like:

```suggestion
	doHandleOpenWindow(openWindowHandler);
```

And you add the new method like this:

```java
/**
 * Explain the situation here and ...
 *
 * <pre>
 * // add some JavaScript code in here
 * evaluate(window.open)
 * </pre>
 *
 */
private void doHandleOpenWindow(Runnable r) {
	if (inEvaluate) {
		r.run();
	} else {
		browser.getDisplay().asyncExec(r);
	}
}
```

fedejeanne:
Test failure is unrelated #1843

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
amartya4256: To me the method name doesn't justify what it does. It seems like it can accept any runnable. The reason why I didn't move that piece of code was as follows:

1. It has only one caller
2. It is very specific to the openWindow handler.

Hence, even if we separate the method, we need a better name. e.g. doHandleRunnableInEvaluate - however, this whole logic is very specific to openWindow event. So in my opinion we do not need to create a new method here until we are reusing this pattern in multiple places.

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
amartya4256: Done

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
fedejeanne: @amartya4256 please restore my changes from ce625749c181e36ff209703684cf01e9695a917d

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
fedejeanne: Ok, but at least add code snippets to your comment or rephrase it so it's a bit simpler to grasp the issue. Something like: _calling 'evaluate(window.open)' may cause deadlocks if it runs asynchronously and starting a new browser instance in the OpenWindowListener ..._ etc etc

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
fedejeanne: @amartya4256 please restore my changes from ce625749c181e36ff209703684cf01e9695a917d

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
fedejeanne: @amartya4256 please restore my changes from ce625749c181e36ff209703684cf01e9695a917d

fedejeanne (changes requested):
@amartya4256 please restore all my changes from ce625749c181e36ff209703684cf01e9695a917d and improve the text of the comment. No need for a new method.

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java

HeikoKlare:
This PR produced a regression regarding:
- https://github.com/eclipse-platform/eclipse.platform.swt/issues/1919

While the root cause of that described issue led to a timeout before this PR, it now leads to a complete UI block from which the application does not recover. 

Please have a look @amartya4256.

This is the according stack trace where it deadlocks:
```
Thread [main] (Suspended)	
	owns: RunnableLock  (id=43)	
	OS.PeekMessage(MSG, long, int, int, int) line: not available [native method]	
	Edge.processOSMessagesUntil(Supplier<Boolean>, Consumer<SWTException>, Display) line: 515	
	Edge.callAndWait(String[], ToIntFunction<IUnknown>) line: 288	
	Edge.evaluate(String) line: 920	
	Edge(WebBrowser).evaluate(String, boolean) line: 405	
	Browser.evaluate(String, boolean) line: 665	
	Browser.evaluate(String) line: 614	
	RichTextEditorConfiguration.customizeToolbar() line: 618	
	RichTextEditor$1.function(Object[]) line: 261	
	Edge.handleCallJava(int, long, long) line: 1014	
	0x000001d17c0cdb78.CallJava(int, long, long) line: not available	
	OS.DispatchMessage(MSG) line: not available [native method]	
	Display.readAndDispatch() line: 3725	
	Edge.processOSMessagesUntil(Supplier<Boolean>, Consumer<SWTException>, Display) line: 516	
	Edge.callAndWait(String[], ToIntFunction<IUnknown>) line: 288	
	Edge.evaluate(String) line: 920	
	Edge(WebBrowser).evaluate(String, boolean) line: 405	
	Browser.evaluate(String, boolean) line: 665	
	Browser.evaluate(String) line: 614	
	RichTextEditor$3.completed(ProgressEvent) line: 285	
	Edge.lambda$45() line: 1130	
	0x000001d17c0d4a28.run() line: not available	
	RunnableLock.run(Display) line: 40	
	Synchronizer.runAsyncMessages(boolean) line: 132	
	Display.runAsyncMessages(boolean) line: 4114	
	Display.readAndDispatch() line: 3730	
	RichTextEditorExample.main(String[]) line: 61	
```

PR #1979, opened 2025-04-02
Enable check for using GC on Image in unsupported use cases

ShahzaibIbrahim:
Logging a warning for GC initialized with unsupported use cases.
1. all dynamic images (retrieved via any of the existing providers).
2. all images for which handles in other zoom values have already been created.

### Examples 

Run this examples with following VM arguments
```
-Dswt.autoScale=quarter
```

```java
import org.eclipse.swt.*;

public class SWTGCImageExample {
	static final ImageFileNameProvider filenameProvider = zoom -> {
		String path = null;
		switch (zoom) {
		case 150:
			path = "bin/org/eclipse/swt/snippets/eclipse.png";
			break;
		case 200:
			path = "bin/org/eclipse/swt/snippets/eclipse.png";
			break;
		default:
			path = "bin/org/eclipse/swt/snippets/eclipse.png";
		}
		return path;
	};

         Image image = new Image(display, filenameProvider);
         GC gc = new GC(image);

         // Draw something on the image
         gc.setForeground(display.getSystemColor(SWT.COLOR_RED));
         gc.drawRectangle(50, 50, 100, 100);
         gc.drawText("Hello SWT!", 60, 90);

         // Dispose GC after drawing
         gc.dispose();

         // Dispose image
         image.dispose();

         shell.dispose();
         display.dispose();
    }
}
```

```java
import org.eclipse.swt.*;

public class SWTImageHandleExample {

        // Create an image using the display
        Image image = new Image(display, 200, 200);

        // Obtain the handle from the Image using win32_getHandle
        long handle = image.win32_getHandle(image, 100);
        System.out.println("Created image handle: " + handle);

        // Initialize GC with the image
        GC gc = new GC(image);

        // Modify the image using GC
        gc.setBackground(display.getSystemColor(SWT.COLOR_BLUE));
        gc.fillRectangle(50, 50, 100, 100);

        // Dispose GC and Image
        gc.dispose();
        image.dispose();

        display.dispose();
    }
}

```


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1743,6 +1743,14 @@ private ImageHandle init(ImageData i, int zoom) {
 @Override
 public long internal_new_GC (GCData data) {
 	if (isDisposed()) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
+
+	if(imageProvider != null && (imageProvider instanceof ImageDataProviderWrapper || imageProvider instanceof ImageFileNameProviderWrapper)) {
```
akoch-yatta, line 1747: instanceof already implicitly does a null check, so you can remove that one.
And change `if(` to `if (`

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1743,6 +1743,14 @@ private ImageHandle init(ImageData i, int zoom) {
...
+		System.err.println("***WARNING: Image initialized with ImageDataProvider or ImageFileNameProvider is not supposed to be modified");
+	}
+
+	if(zoomLevelToImageHandle != null && !zoomLevelToImageHandle.isEmpty()) {
```
akoch-yatta, line 1751: space after if
I think the check is not complete, we should exclude the ImageGCDrawerWrapper. This one would be fine no matter the amount of handles.
And shouldn't we check for zoomLevelToImageHandle.size() > 1 instead? Isn't there always already one handle at this point or will it be created later?
I would extend the message and add "should not be modified **with a GC. Changes will not be applied on existing handles**"

ShahzaibIbrahim: At this point there should be no handle. That's why I added the empty check. We don't create handle on initialization anymore.

akoch-yatta: Ah, yes you are right. It will implicitly create the handle, when it is first used inside a GC call 👍

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1743,6 +1743,14 @@ private ImageHandle init(ImageData i, int zoom) {
 @Override
 public long internal_new_GC (GCData data) {
 	if (isDisposed()) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
+
+	if(imageProvider != null && (imageProvider instanceof ImageDataProviderWrapper || imageProvider instanceof ImageFileNameProviderWrapper)) {
+		System.err.println("***WARNING: Image initialized with ImageDataProvider or ImageFileNameProvider is not supposed to be modified");
```
laeubi, line 1748: If this is really wrong, then an exception should be thrown.

laeubi (changes requested):
I don't think that printing to system out is appropriate here.

Either it is an error, then we should throw an exception or it should simply be ignored.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
HeikoKlare: Throwing an exception is kind of a breaking change. Yes, it's kind of wrong to do this, but making erroneous consumers (that just had incorrect images so far) aware of this by potentially breaking the complete application through an exception is probably nothing we should do. Throwing an exception would be something that needs to be guarded by an explicit strict mode (we have one already).

laeubi: I don't see how consumers will be "made aware" other than forbid bad calls. E.g. system outs will go completely unnoticed in regular Eclipse installs as they never show up in the error log.

laeubi: The PR title says "GC with Image in unsupported use cases", so it either *is* supported the nothing has to be done or it is not supported then an Exception must be thrown as otherwise people will simply use it.

As this is also only done for Windows, I wonder what developers / framework consumer should benefit from that, and it more feels like a temporary debugging facility as the message is completely unclear what it is an error / unsupported at all and what should be the alternative. Bot provided examples look perfectly valid to me as one might want to draw over an image and then write it to a file or whatever.

In any case we have `org.eclipse.swt.widgets.Display.getRuntimeExceptionHandler()` and `org.eclipse.swt.widgets.Display.getErrorHandler()` facility so application devlopers at least have control over how the issue is reported/handled.

laeubi: > as it will not give you the results that you expect, at latest when changing the zoom

I don't think that the primary use-case of images is changing the zoom dynamically. Images are used mostly in application code as images, and I don't see that the API forbid (or discourage) getting a "zoomed" image and paint on it. As in the first example I might then retrieve the `ImageData` and do whatever I want with that. Or I paint the image on a component (in wich case I might need to handle zoom changes anyways). All the examples even dispose the image right away and these are common examples so I can't see how this is unsupported or qualifies for a warning.

For the second example I'm completely lost... How is one supposed to create a (user drawn) image if not by creating one with a GC and paint on it? Would we expect that users are creating images pixel-by-pixel?

> How may these handlers be used here?

At least platform registers these handlers and perform certain actions on it (at least log them) so for me there are two ways:

1. We reuse these and e.g. handle it like a `RuntimeException` in a handler (e.g with passing an `AssertionException`), this might be arguable if this is really "handler code" but I don't think it is really that strict. e.g. `org.eclipse.swt.widgets.Synchronizer.syncExec(Runnable)` use it and this is obviously not always a "handler" so any code running in the UI thread can be seen as "handler code" anyways.
2. We add a new pair of `get/setAssertionHandler()` that explicitly accepts an `AssertionException` so we can pass any failed assertions there so that application code is able to handle it in whatever fashion seems suitable
 

With that approach there is a stack trace to trace down the actual call (this is what concerns me most at the moment that there is literally no way to find out where the call is coming from (except attaching a debugger) and there is a little chance people complain and things get fixed.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
ShahzaibIbrahim: zoomLevelToImageHandle if not empty -> error // or if one handle is present which is not equal to current zoom -> error

ShahzaibIbrahim:
@laeubi I have added a snippet to demonstrate on how to use Image(Display, ImageGcDrawer, int, int) to load an image to the system and draw watermark over it (supported for all zoom level images.

Also I have again changed the condition to only log an error when using the strict mode. This was done as per @HeikoKlare suggestion since Display#setRuntimeExceptionHandler handler contract states to only use it whenever an exception is thrown by a listener or external callback function and this is not the case here.

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet384.java
```diff
@@ -0,0 +1,84 @@
+package org.eclipse.swt.snippets;
+/*******************************************************************************
```
HeikoKlare, line 2: Copyright header should be above package declaration.

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet384.java
```diff
@@ -0,0 +1,84 @@
...
+ * Mainly motivated by custom drawing behavior in the win32 implementation
+ * <p>
+ * For a list of all SWT example snippets see
+ * http://www.eclipse.org/swt/snippets/
+ * </p>
+ */
+public class Snippet384 {
+
+    public static void main(String[] args) {
+		System.setProperty("swt.autoScale.updateOnRuntime", "true");
```
HeikoKlare, line 31: Please auto-format the added code.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1716,6 +1724,9 @@ public long internal_new_GC (GCData data) {
 		SWT.error(SWT.ERROR_INVALID_ARGUMENT);
 	}
 
+	assertCondition(imageProvider != null && (imageProvider instanceof ImageDataProviderWrapper || imageProvider instanceof ImageFileNameProviderWrapper), "Image initialized with ImageDataProvider or ImageFileNameProvider is not supposed to be modified.");
+	assertCondition(!zoomLevelToImageHandle.isEmpty() && (zoomLevelToImageHandle.size() != 1 || !zoomLevelToImageHandle.containsKey(getZoom())), "Images with handles created for a different zoom level should not be modified.");
+
```
HeikoKlare, line 1729: I would rather guard this whole block behind `strictChecks` to avoid repeated evaluation of all the conditions when `strictChecks` is usually just `false`.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1690,6 +1692,12 @@ private ImageHandle init(ImageData i, int zoom) {
 	}
 }
 
+private void assertCondition(boolean condition, String message) {
+    if (condition && strictChecks) {
+        System.err.print(message + "\nSee https://github.com/eclipse-platform/eclipse.platform.swt/tree/master/examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet384.java for supported variants.");
```
HeikoKlare, line 1697: I would also add more general statement than only referring to an example snippet here. Something like "Images with custom parts should be created with an ImageGcDrawer", as that's the actualy replacement for any kind of image whose creation involved a GC by now.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -80,6 +80,8 @@
  */
 public final class Image extends Resource implements Drawable {
 
+	static boolean strictChecks = System.getProperty("org.eclipse.swt.internal.enableStrictChecks") != null;
```
HeikoKlare, line 83: This value is already defined in `Device.strictChecks`, so please reuse that.

HeikoKlare (commented):
I have some comments but in general I am fine with the proposed change.

>Also I have again changed the condition to only log an error when using the strict mode. This was done as per @HeikoKlare suggestion since Display#setRuntimeExceptionHandler handler contract states to only use it whenever an exception is thrown by a listener or external callback function and this is not the case here.

To clarify this: it's not (only) about the specified usage scenario of the runtime exception handler but also about how this debug information may be used. I don't see that anyone will inspect the error log view of tested products, configure a different handler for runtime exceptions at the display to find such errors or the like, so the logging capability will likely remain unused. There is already some other logging guarded behind the `strictChecks` property, which one can just activate for a product under test to then inspect the logs of, e.g., Jenkins builds, test runs etc.
We may think of improving the way in which the collection of debug information for `strictChecks` is done in general, but since all other places currently use standard output, we should stick with it here to have a uniform way of logging that information and may do that as a separate step for all affected places.

laeubi:
> I don't see that anyone will inspect the error log view of tested products

Whenever something goes wrong I go to the error log and I also report incidents from there, others do as well, so I don't see this a general problem.

> configure a different handler for runtime exceptions at the display to find such errors or the like

Also SWT can be used standalone so its hard to tell how people actually use it. In any case if no one is looking at the errors, then my initial recommendation stays: simply don't log this. If it is important enough for people, they will do.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
laeubi: Do we already have a N&N for `ImageGcDrawer` (I still find the name quite confusing) we can reference here? Such N&N can then also reference snippets and whatever seems useful in that area.)

HeikoKlare: https://eclipse.dev/eclipse/news/news.html?file=4.35/platform_isv.html

HeikoKlare:
> Whenever something goes wrong I go to the error log and I also report incidents from there, others do as well, so I don't see this a general problem.

If we guard this behind strict checks (like we do for all comparable checks), it will only pop-up in the error log if you enable strict checks, and I guess that usually no one does not.

> Also SWT can be used standalone so its hard to tell how people actually use it. In any case if no one is looking at the errors, then my initial recommendation stays: simply don't log this. If it is important enough for people, they will do.

The strict checks are an ability for developers to run an application with, as the name says, with more strict checks. You have to explicitly enable them to validate your application (manually or via automated tests) regarding those checks. The information is not that relevant for a consumer of an application (who sees the error log and may use the information to report a bug if it annoys them) but just for developers who can simply activate those strict checks to have further validation of an application. This is the same for the other existing strict checks. So why shouldn't we provide the ability to check an application regarding the risk of imperfect images or why should we introduce a different means of identifying them than the other strict checks we have?

HeikoKlare (approved):
I rebased on master resolving conflicts due to a snippet with the same name meanwhile being added. I also slightly adapted the implementation.

Change is now fine for me. @laeubi do you have any remaining concerns or can we merge this for M3?

laeubi (approved):
Looks good now. I assume there will be a N&N for SWT already to make users generally aware of `ImageDrawer` so this seems suitable here.

HeikoKlare:
Thank you!
Yes, the N&N is from previous release: https://eclipse.dev/eclipse/news/news.html?file=4.35/platform_isv.html

PR #1983, opened 2025-04-02
[win32] Use consistent font zoom retrieval

akoch-yatta:
This PR unifies the extraction of the target zoom of a font to be
affected by the zoom context of the underlying Widget. Before there
were different strategies like getting it from the Shell.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Widget.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Widget.java
@@ -2690,6 +2690,10 @@ GC createNewGC(long hDC, GCData data) {
 	return GC.win32_new(hDC, data);
 }
 
+int getNativeZoom() {
+	return nativeZoom;
+}
+
 int getZoom() {
 	return DPIUtil.getZoomForAutoscaleProperty(nativeZoom);
 }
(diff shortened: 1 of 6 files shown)


amartya4256 (approved):
Looks rather cleaner to me. Also using a getter makes it more abstract for the consumers since they don't need to worry about which component's nativeZoom they should use. Approved.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Widget.java
```diff
@@ -2690,6 +2690,10 @@ GC createNewGC(long hDC, GCData data) {
 	return GC.win32_new(hDC, data);
 }
 
+int getFontZoom() {
```
amartya4256, line 2693: Just a suggestion, can we not call it `getNativeZoom`? Since, I am not sure if getFontZoom would make sense for every kind of widget.

akoch-yatta: Hmm, yes, makes sense. It is solely used for fonts right now, but that method would not be limited to this purpose. I will rename it.

HeikoKlare (approved):

amartya4256 (approved):

HeikoKlare:

PR #1984, opened 2025-04-02
[GTK4] Manage propagation for gesture events and fix popup menus

ptziegler:
For e.g. the text widget, a default pop-up menu if provided by GTK and
activated, unless a custom menu is set within SWT. To avoid both menus
being shown at the same time, the event must be consumed when showing
the SWT menu.

In GTK3, this is done by returning TRUE in the callback method, which is
no longer possible in GTK4. Instead, the event needs to be marked as
"claimed" via gtk_gesture_set_sequence_state(), to prevent other, native
listeners from processing it as well.

To do so, the gtk_gesture_press_event() and gtk_gesture_release_event()
have been adapted to now return an integer, which may be one of the
following values:

- GTK_EVENT_SEQUENCE_NONE
- GTK_EVENT_SEQUENCE_CLAIMED
- GTK_EVENT_SEQUENCE_DENIED

For example, the value should be GTK_EVENT_SEQUENCE_CLAIMED after the
menu is shown.

Note: It is not possible to always claim an event as "claimed", as it
would then prevent the default value from being shown at all, even if no
SWT menu is set.

Additionally, the check for whether a menu can be shown has been
inverted ((state & MENU) != 0), to match the GTK3 logic. Because this
bit is always set, this effectively stops all SWT menus from being
shown.


ptziegler:
Left is how it currently looks like, right is how it's supposed to look like. The current menu is the default one provided by GTK and not the one that's created in Snippet 122.

image image

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Widget.java
```diff
@@ -2540,6 +2541,8 @@ void gesturePressReleaseProc(long gesture, int n_press, double x, double y, long
 			gtk_gesture_release_event(gesture, n_press, x, y, event);
 			break;
 	}
+
+	GTK4.gtk_gesture_set_sequence_state(gesture, sequence, GTK4.GTK_EVENT_SEQUENCE_CLAIMED);
```
ptziegler, line 2545: The GTK3 callbacks return true or false, depending on whether the event has been consumed. This has never been considered for the GTK4 events. I wonder if it makes sense to add this as well, rather than to always claim it?

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Control.java
```diff
@@ -3416,7 +3416,7 @@ void gtk_gesture_press_event (long gesture, int n_press, double x, double y, lon
 	display.clickCount = n_press;
 	if (n_press == 1) {
 		sendMouseEvent(SWT.MouseDown, eventButton, n_press, 0, false, eventTime, x, y, false, eventState);
-		if ((state & MENU) == 0) {
+		if ((state & MENU) != 0) {
```
ptziegler, line 3419: See https://github.com/eclipse-platform/eclipse.platform.swt/blob/10cb37c38f41cf91be10a1efc91ec3cc7ea2a1ea/bundles/org.eclipse.swt/Eclipse%20SWT/gtk/org/eclipse/swt/widgets/Control.java#L3544-L3550

akurtakov:
I tried running this PR and it crashed with 
```
Exception in thread "main" java.lang.UnsatisfiedLinkError: 'long org.eclipse.swt.internal.gtk3.GTK3.gtk_clipboard_wait_for_contents(long, long)'
	at org.eclipse.swt.internal.gtk3.GTK3.gtk_clipboard_wait_for_contents(Native Method)
	at org.eclipse.swt.dnd.Clipboard.gtk_clipboard_wait_for_contents(Clipboard.java:701)
	at org.eclipse.swt.dnd.Clipboard.getAvailableClipboardTypes(Clipboard.java:679)
	at org.eclipse.swt.dnd.Clipboard.getAvailableTypes(Clipboard.java:576)
	at org.eclipse.swt.dnd.Clipboard.getAvailableTypes(Clipboard.java:544)
	at org.eclipse.swt.snippets.Snippet122.lambda$2(Snippet122.java:70)
	at org.eclipse.swt.events.MenuListener$2.menuShown(MenuListener.java:80)
	at org.eclipse.swt.widgets.TypedListener.handleEvent(TypedListener.java:293)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:5864)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1652)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1678)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1657)
	at org.eclipse.swt.widgets.Menu._setVisible(Menu.java:290)
	at org.eclipse.swt.widgets.Display.runPopups(Display.java:5116)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:4515)
	at org.eclipse.swt.snippets.Snippet122.main(Snippet122.java:83)
```
Do you experience that ?

ptziegler:
> I tried running this PR and it crashed with
> 
> ```
> Exception in thread "main" java.lang.UnsatisfiedLinkError: 'long org.eclipse.swt.internal.gtk3.GTK3.gtk_clipboard_wait_for_contents(long, long)'
> 	at org.eclipse.swt.internal.gtk3.GTK3.gtk_clipboard_wait_for_contents(Native Method)
> 	at org.eclipse.swt.dnd.Clipboard.gtk_clipboard_wait_for_contents(Clipboard.java:701)
> 	at org.eclipse.swt.dnd.Clipboard.getAvailableClipboardTypes(Clipboard.java:679)
> 	at org.eclipse.swt.dnd.Clipboard.getAvailableTypes(Clipboard.java:576)
> 	at org.eclipse.swt.dnd.Clipboard.getAvailableTypes(Clipboard.java:544)
> 	at org.eclipse.swt.snippets.Snippet122.lambda$2(Snippet122.java:70)
> 	at org.eclipse.swt.events.MenuListener$2.menuShown(MenuListener.java:80)
> 	at org.eclipse.swt.widgets.TypedListener.handleEvent(TypedListener.java:293)
> 	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:91)
> 	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:5864)
> 	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1652)
> 	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1678)
> 	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1657)
> 	at org.eclipse.swt.widgets.Menu._setVisible(Menu.java:290)
> 	at org.eclipse.swt.widgets.Display.runPopups(Display.java:5116)
> 	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:4515)
> 	at org.eclipse.swt.snippets.Snippet122.main(Snippet122.java:83)
> ```
> 
> Do you experience that ?

Yes. Locally I've adapted `Clipboard.gtk_clipboard_wait_for_contents()` to always return `0` when running with GTK4.

And just as a general update: I'll plan on creating a separate PR to properly implement the event cancellation, because the current approach of always consuming the event doesn't cover all cases.

In short: If no menu is set, the default menu should be used. At least that's how it currently works with GTK3 and something you can see in the snippet if you comment out the call to `setMenu(...)`. When the event is always consumed, this no longer happens.

Conversly, if the event is never consumed, you get both the native menu and the SWT menu on top of each other...

ptziegler:
@akurtakov I've tried to find another way to produce a "bubbling event", but so far without success. So in the end, I've combined everything in a single commit.

For testing, I've used Snippet 122 with and without a menu, together with https://github.com/eclipse-platform/eclipse.platform.swt/pull/1990.

akurtakov:
This is clear improvement to the current state thus pushing. Further improvements (if needed) should be handled separately.

PR #1985, opened 2025-04-02
Rework new Image constructors accepting a ImageGcDrawer

HannesWell:
When implementing the `ImageGCDrawer` as lambda, the resulting code is better to read if the `ImageGCDrawer` is passed as last argument:
```java
new Image(device, 16, 16, (gc, w, h) -> {
	gc.draw ...
})
```
For example at https://github.com/eclipse-platform/eclipse.platform.ui/pull/2869 or in the adapted parts within SWT one can see that the current order encourages to code in way that is more verbose.

Furthermore this way the argument order also better reflects the order in which the arguments are applied:
1. an image with the given size is created.
2. the GC draws on that image

I know that changing an API after it was released is not ideal, but since it's very new I assume there are not yet many users and existing users are very likely able to adapt especially as the change is trivial. Consequently we have a real chance to remove the first shot eventually. On the other hand if we stick with the existing API we might be annoyed by its imperfection forever.
If we have this, I can take care of adopting the existing callers in the SDK.

Follow-up on
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/1734

@HeikoKlare, @fedejeanne or anybody else, what do you think?

### DIFF
--- a/binaries/org.eclipse.swt.win32.win32.x86_64/.settings/.api_filters
+++ b/binaries/org.eclipse.swt.win32.win32.x86_64/.settings/.api_filters
@@ -582,10 +582,11 @@
         </filter>
     </resource>
     <resource path="Eclipse SWT/common/org/eclipse/swt/graphics/ImageGcDrawer.java" type="org.eclipse.swt.graphics.ImageGcDrawer">
-        <filter id="404000815">
+        <filter id="403984517">
             <message_arguments>
                 <message_argument value="org.eclipse.swt.graphics.ImageGcDrawer"/>
-                <message_argument value="getGcStyle()"/>
+                <message_argument value="org.eclipse.swt.graphics.ImageDrawer"/>
+                <message_argument value="withGCStyle(int, ImageDrawer)"/>
             </message_arguments>
         </filter>
     </resource>
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
@@ -734,7 +734,7 @@ Image createButtonImage(Display display, int button) {
 	final Rectangle trim = renderer.computeTrim(button, SWT.NONE, 0, 0, 0, 0);
 	final Point imageSize = new Point(size.x - trim.width, size.y - trim.height);
 	Color transColor = renderer.parent.getBackground();
-	final ImageGcDrawer imageGcDrawer = new TransparencyColorImageGcDrawer(transColor) {
+	return new Image(display, imageSize.x, imageSize.y, new TransparencyColorImageGcDrawer(transColor) {
 		@Override
 		public void drawOn(GC gc, int imageWidth, int imageHeight) {
 			Rectangle imageBounds = new Rectangle(0, 0, imageWidth, imageHeight);
@@ -742,8 +742,7 @@ Image createButtonImage(Display display, int button) {
 			gc.fillRectangle(imageBounds);
 			renderer.draw(button, SWT.NONE, imageBounds, gc);
 		}
-	};
-	return new Image(display, imageGcDrawer, imageSize.x, imageSize.y);
+	});
 }
 
 private void notifyItemCountChange() {
@@ -4007,7 +4006,8 @@ void updateBkImages(boolean colorChanged) {
 						if (colorChanged || !bounds.equals(bkImageBounds[i])) {
 							bkImageBounds[i] = bounds;
 							if (controlBkImages[i] != null) controlBkImages[i].dispose();
-							controlBkImages[i] = new Image(control.getDisplay(), (gc, imageWidth, imageHeight) -> renderer.draw(CTabFolderRenderer.PART_BACKGROUND, 0, bounds, gc), bounds.width, bounds.height);
+							controlBkImages[i] = new Image(control.getDisplay(), bounds.width, bounds.height,
+									(gc, w, h) -> renderer.draw(CTabFolderRenderer.PART_BACKGROUND, 0, bounds, gc));
 							control.setBackground(null);
 							control.setBackgroundImage(controlBkImages[i]);
 						}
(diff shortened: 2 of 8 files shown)


HeikoKlare:
I don't have a strong opinion on this. I agree that the proposed order would have been better from the beginning and I also agree on the "semantics" of the argument order (first specifying image creation and then image contents). I only disagree with the other argument of producing more verbose code because of the current API preventing easily inlining a multi-line lambda, as that conflicts with my understanding of clean coding practices: If there is a multi-line lambda, that should be assigned to a variable that condensely expresses its meaning. When I inline a multi-line lambda into the call of another method (or a constructor, like here), the code becomes incomprehensible anyway, no matter whether that lambda is the last argument of the call or not. That's why I would encourage to never do that anyway

So in summary, I am not sure whether the change is worth the API extension + deprecation, but at the same time I don't have a strong argument against it, in particular since the API has just been introduced and is not consumed that much yet.

fedejeanne (dismissed):
> @HeikoKlare, @fedejeanne or anybody else, what do you think?

Same as Heiko: I have no strong opinion about this.

Your proposal is reasonable and I see the slight benefits. If no one objects, you may go for it @HannesWell :-) ✔️

HannesWell:
> I only disagree with the other argument of producing more verbose code because of the current API preventing easily inlining a multi-line lambda, as that conflicts with my understanding of clean coding practices: If there is a multi-line lambda, that should be assigned to a variable that condensely expresses its meaning. When I inline a multi-line lambda into the call of another method (or a constructor, like here), the code becomes incomprehensible anyway, no matter whether that lambda is the last argument of the call or not. That's why I would encourage to never do that anyway

I have to admit that I have the exact opposite onion about it. If a lambda is only used once, most of the times I find it much cleaner to line it, because a lambda usually specifies the behavior of something and I find it more clear if it's then written in the context of the 'thing' whose behavior it defines then somewhere else. One prominent example for me is the `map()` operation of a Stream-pipeline. Usually I would always define the map-operation as inlined lambda, even if it has multiple lines. Then one can read and understand the code sequentially without the need to jump mentally between (code) locations.

But to stick with this case, with a inlined lambda I can define the drawer and therefore the content of the image, literally within the image-constructor and can say I create a new Image and draw this content XYZ :
```
Image rightCaretBitmap = new Image(display, caretWidth, lineHeight, (gc, width, height) -> {
	gc.setBackground(display.getSystemColor(SWT.COLOR_BLACK));
	gc.fillRectangle(0, 0, width, height);
	gc.setForeground(display.getSystemColor(SWT.COLOR_WHITE));
	gc.drawLine(width-1,0,width-1,height);
	gc.drawLine(0,0,width-1,0);
	gc.drawLine(width-1,1,1,1);
});
```
Previously it read like: I create an `ImageGcDrawer` that draws content XYZ and then I create a new Image and pass the just defined drawer:
```
ImageGcDrawer rightCaretDrawer = (gc, width, height) -> {
	gc.setBackground(display.getSystemColor(SWT.COLOR_BLACK));
	gc.fillRectangle(0, 0, width, height);
	gc.setForeground(display.getSystemColor(SWT.COLOR_WHITE));
	gc.drawLine(width-1,0,width-1,height);
	gc.drawLine(0,0,width-1,0);
	gc.drawLine(width-1,1,1,1);
};
Image rightCaretBitmap = new Image(display, rightCaretDrawer, caretWidth, lineHeight);
```

In the end it's of course not a game-changer, but I personally found the former more clear than the latter.

HannesWell:
In a second commit I have now also added the suggestion from https://github.com/eclipse-platform/eclipse.platform.swt/pull/1948#issuecomment-2773931682, about how a Image(GC)Drawer could be extended in an alternative way. Please consider this just as a starting point for a discussion. This doesn't have to be the final result.
The idea is that simple ImageDrawers are just created as lambdas and if one wants to add post-processing and defines the style of the GC one can create special static (factory) methods that wrap the given drawer together with the post-processing/style info and as the Image class then can know this (internal) class it can check for the type and extract the additionally specified information if possible.

E.g. to create a drawer with post-processing one could then use:
```
Image image = new Image(display, width, height, ImageDrawer.create((gc, w, h) -> {
	gc.fillRectangle(0, 0, w, h);
	...
}, imageData -> {
	// do some post-processing
});
```
to create an Image with a drawer with a certain style one could use:
```
Image image = new Image(display, width, height, ImageDrawer.withGCStyle(SWT.LEFT_TO_RIGHT, (gc, w, h) -> {
	gc.fillRectangle(0, 0, w, h);
	...
});
```
or to create a drawer with a style and post-processing one could use:
```
Image image = new Image(display, width, height, ImageDrawer.create(withGCStyle(SWT.LEFT_TO_RIGHT, (gc, w, h) -> {
	gc.fillRectangle(0, 0, w, h);
	...
}), imageData -> {
	// do some post-processing
});
```

And since multiple things are then already change in this area I also wanted to suggest to replace `ImageGcDrawer` by an new interface `ImageDrawer` because I assumed the GC part is implied because I don't know another way drawing on an image than with a GC. And without GC the name is a bit simpler. But we can also realize the factory methods within the existing  `ImageGcDrawer` interface. So that's also just an idea. :)

laeubi:
+1 for having lamda as last argument and
+1 for `ImageDrawer` name (while I more think it is an `ImageDecorator` or `ImageProducer`)

Beside that I personally more like [BuilderPatterns](https://en.wikipedia.org/wiki/Builder_pattern) over constructors with many optional and/or varying variants.

Something like 
```
 Image image = Image.from(device)
  .withSize(width, height)
  .drawing((gc, w, h)-> {
     //draw on me
   }).process(imageData -> {
	// do some post-processing
}.create();
```

that way its much easier to adjust/change without the hassle to find the "right" position for parameter X and its easier to evolve and better to read.

fedejeanne (commented):
- -1 on having to explicitly know about `ExtendedImageDrawer` (just as Heiko pointed out)
- +1 on the usage of builders if you want to hide such details
- +0 (neutral) on renaming `ImageGcDrawer` to `ImageDrawer` but please adapt the names of the parameters and variables accordingly
- +0 on changing the order of the parameters in the constructor

All in all: I like factories/builders but I don't like having to know the classes explicitly in order for them to work. Having checks saying `if (drawer instanceof ExtendedImageDrawer extendedDrawer)` is fine inside a factory method (or a builder) but only there and only if that factory method (or builder) is the only one that needs to know about (and instantiate) `ExtendedImageDrawer`.

HTH

HeikoKlare:
How should we follow up on this? In my opinion, it's too late to deprecate the constructors added to `Image` as after more than a year, they will have quite some adoption now. Having better construction patterns (using factories/builders) would be nice at many places (including the proposals made here for `Image`), but I wonder if it's worth to start at some random place with a better object construction pattern without having a holistic concept. If we want to work on this, I would be in favor of having one common concept for, e.g., all `Resource` implementations at once, associated with a proper goal or rather a specific benefit that consumers get when adopting those new construction patterns.

akurtakov:
What is the plan here? If there is no interest or it no longer makes sense it's best to close and stop poluting the PRs queue.

PR #1986, opened 2025-04-02
[Build] Stash jdk-resources with respect to WS to prevent overwrites

HannesWell:
For linux.x86_64 the jdk-resources for gtk(3) and gtk4 are downloaded, stashed and deleted into/from the same directory. This leads to concurrency issues and build failures.


HannesWell:
I cannot tell why it has worked until yesterday, but with this the gtk(3/4) build should be robust again. At least the latest build passed.

PR #1987, opened 2025-04-03
Allow to test for loading image at zoom and fix faulty Win32 image load

HeikoKlare:
The ImageLoader and FileFormat implementations currently combine the check whether an image can be provided at a specific scale value and the provision of an image in the scale itself. In case a consumer wants to test whether an image can be provided at a specific scale, it needs to request that image at that scale, even if it will not be used later one, e.g., because it cannot be provided in the required scale.

With this change, ImageLoader and FileFormat provide separate methods for validating whether an image can be retrieved at a specific zoom (such as from an SVG) and the retrieval of the image itself. This is employed by the Cocoa Image implementation to avoid unnecessary load operations and in the Win32 Image implementation to correct erroneous scaling from an already initialized handle instead of loading at proper zoom (such as for an SVG). In addition, this fixes a faulty reuse of the
same stream in the Cocoa Image implementation at the same place. See https://github.com/eclipse-platform/eclipse.platform/pull/1797#pullrequestreview-2740239609 and https://github.com/eclipse-platform/eclipse.platform.ui/pull/2643#pullrequestreview-2727866854.

This will specifically be useful for consumers that need to load images on their own (such as the URLImageFileNameProvider and URLImageDataProvider in JFace), as they would otherwise need to test for the ability to retrieve an image at a specific scale by actually retrieving the image without using the loaded image data later on.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/ImageLoader.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/ImageLoader.java
@@ -163,6 +163,11 @@ List<ElementAtZoom<ImageData>> load(InputStream stream, int fileZoom, int target
 	return images;
 }
 
+static boolean canLoadAtZoom(InputStream stream, int fileZoom, int targetZoom) {
+	if (stream == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
+	return FileFormat.canLoadAtZoom(new ElementAtZoom<>(stream, fileZoom), targetZoom);
+}
+
 /**
  * Loads an array of <code>ImageData</code> objects from the
  * file with the specified name. Throws an error if either
@@ -194,6 +199,16 @@ List<ElementAtZoom<ImageData>> load(String filename, int fileZoom, int targetZoo
 		SWT.error(SWT.ERROR_IO, e);
 	}
 	return null;
+}
+
+static boolean canLoadAtZoom(String filename, int fileZoom, int targetZoom) {
+	if (filename == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
+	try (InputStream stream = new FileInputStream(filename)) {
+		return canLoadAtZoom(stream, fileZoom, targetZoom);
+	} catch (IOException e) {
+		SWT.error(SWT.ERROR_IO, e);
+	}
+	return false;
 }
 
 /**
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
@@ -20,7 +20,6 @@ import java.util.function.*;
 
 import org.eclipse.swt.*;
 import org.eclipse.swt.internal.*;
-import org.eclipse.swt.internal.DPIUtil.*;
 import org.eclipse.swt.internal.cocoa.*;
 import org.eclipse.swt.internal.graphics.*;
 import org.eclipse.swt.internal.image.*;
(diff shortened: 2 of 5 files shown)


bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/ImageLoader.java
```diff
@@ -155,14 +155,19 @@ public ImageData[] load(InputStream stream) {
...
-	List<ElementAtZoom<ImageData>> images = NativeImageLoader.load(new ElementAtZoom<>(stream, fileZoom), this, targetZoom);
-	data = images.stream().map(ElementAtZoom::element).toArray(ImageData[]::new);
+	List<ImageData> images = NativeImageLoader.load(new ElementAtZoom<>(stream, fileZoom), this, targetZoom);
+	data = images.stream().toArray(ImageData[]::new);
 	return images;
 }
 
+boolean canLoadAtZoom(InputStream stream, int fileZoom, int targetZoom) {
+	if (stream == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
+	return NativeImageLoader.canLoadAtZoom(new ElementAtZoom<>(stream, fileZoom), targetZoom);
```
HannesWell, line 168: Calling `FileFormat.canLoadAtZoom()` here directly would save us the indirection via `NativeImageLoader` that is identical over all three impls.
```suggestion
	return FileFormat.canLoadAtZoom(new ElementAtZoom<>(stream, fileZoom), targetZoom);
```

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/ImageLoader.java
```diff
@@ -155,14 +155,19 @@ public ImageData[] load(InputStream stream) {
...
 	if (stream == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
 	reset();
-	List<ElementAtZoom<ImageData>> images = NativeImageLoader.load(new ElementAtZoom<>(stream, fileZoom), this, targetZoom);
-	data = images.stream().map(ElementAtZoom::element).toArray(ImageData[]::new);
+	List<ImageData> images = NativeImageLoader.load(new ElementAtZoom<>(stream, fileZoom), this, targetZoom);
+	data = images.stream().toArray(ImageData[]::new);
 	return images;
 }
 
+boolean canLoadAtZoom(InputStream stream, int fileZoom, int targetZoom) {
```
HannesWell, line 166: This method could be static, which would avoid the creation of otherwise unused `ImageLoader` instances.
```suggestion
static boolean canLoadAtZoom(InputStream stream, int fileZoom, int targetZoom) {
```

HannesWell (changes requested):
> In case a consumer wants to test whether an image can be provided at a specific scale, it needs to request that image at that scale, even if it will not be used later one, e.g., because it cannot be provided in the required scale.

This is definitively use-full, but the current implementation seems to have the problem that it re-uses the initially given `InputStream` without resetting it because the check if an image can be loaded at a specific zoom involves determining it's format, which reads a portion of the beginning of the stream. IIRC this was also the reason why the check for scaling and loading the image was combined in one method, because internally the stream was wrapped into a `LEDataStream` that was reset then.
I assume that wrapping has to be moved further up the call-chain and then, maybe it would help to retrieve the `FileFormat` instance for a given stream and reuse it. Then it doesn't has to determined multiple times. Of course one has to pay attention for Linux where the `FileFormat` is not to load static images.

Besides that I currently just have a few minor remarks how this could be a bit simpler.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/ImageLoader.java
HeikoKlare: That's true. I added the method there because we wanted to use methods in the native loader from Platform UI as a first implementation, but we can also directly access the FileFormat implementation. I adapted that in the latest push.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/ImageLoader.java
HeikoKlare: Makes sense, adapted accordingly.

HeikoKlare:
> but the current implementation seems to have the problem that it re-uses the initially given `InputStream` without resetting it because the check if an image can be loaded at a specific zoom involves determining it's format,

Indeed, I missed that point. Actually, there was a mistake in the Cocoa implementation reusing the same stream for loading an image multiple times as well, which I now fixed in this PR. To avoid overcomplicating this, I have reduced the PR to only add the new method. In many use cases, there will be no reuse necessity for the stream anyway, which is why I now simply kept the existing implementation of the load method and added a validation method, which is used at specific places. As said, in the Cocoa implementation a stream was erroneously reused anyway, which I solved in the same way the Win32 implementation already does it: store the stream contents and wrap then into a stream on demand.

HannesWell (approved):
> Indeed, I missed that point. Actually, there was a mistake in the Cocoa implementation reusing the same stream for loading an image multiple times as well, which I now fixed in this PR. To avoid overcomplicating this, I have reduced the PR to only add the new method.

Yes, indeed. Thank you for the update and the fix. This looks good to me now.

PR #1988, opened 2025-04-03
[win32] Handle all scaled font via SWTFontProvider

akoch-yatta:
This PR extends the SWTFontProvider and the SWTFontRegistry to receive scaled variants of a font where only the handle is available. Common use case for this is OS.SendMessage (hwnd, OS.WM_GETFONT, 0, 0) to receive the handle of the current SWT font of an OS handle. As at this point the correct zoom the handle was created is not known with 100%, it is better to retrieve and manage those fonts via the SWTFontRegistry, expecially when monitor specific scaling is used. When the passed font handle is not already managed via the SWTFontRegistry it will create the SWT Font as it was create previously.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Composite.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Composite.java
@@ -1525,7 +1525,7 @@ LRESULT WM_PAINT (long wParam, long lParam) {
 					Control control = findBackgroundControl ();
 					if (control == null) control = this;
 					data.background = control.getBackgroundPixel ();
-					data.font = Font.win32_new(display, OS.SendMessage (handle, OS.WM_GETFONT, 0, 0), nativeZoom);
+
 					data.uiState = (int)OS.SendMessage (handle, OS.WM_QUERYUISTATE, 0, 0);
 					if ((style & SWT.NO_BACKGROUND) != 0) {
 						/* This code is intentionally commented because it may be slow to copy bits from the screen */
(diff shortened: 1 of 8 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Widget.java
```diff
@@ -2690,6 +2690,10 @@ GC createNewGC(long hDC, GCData data) {
 	return GC.win32_new(hDC, data);
 }
 
+int getFontZoom() {
```
amartya4256, line 2693: See https://github.com/eclipse-platform/eclipse.platform.swt/pull/1983/files#r2055618075
Same remark here.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Composite.java
```diff
@@ -1536,6 +1536,8 @@ LRESULT WM_PAINT (long wParam, long lParam) {
 						drawBackground (phdc [0], rect);
 					}
 					GC gc = createNewGC(phdc [0], data);
+					data.font = SWTFontProvider.getFont(display, OS.SendMessage (handle, OS.WM_GETFONT, 0, 0), data.nativeZoom);
```
amartya4256, line 1539: Why move it down here?

akoch-yatta: data.nativeZoom is adapted in the createNewGC call, so I needed to move it down there

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Widget.java
akoch-yatta: Adapted it

HeikoKlare (approved):

amartya4256 (approved):

HeikoKlare:

PR #1989, opened 2025-04-03
[Win32] Ensure consistent image data returned for filename/data provider

HeikoKlare:
Due to the on-demand creation of image handles, there is not necessarily a handles anymore from which image data is retrieved when requesting is via the getImageData(...) methods. This results in potentially different kinds of image data (including different anti-aliasing results) depending on whether a handle has already been created for an image at the given zoom or not.

This change adapts the Image implementation for ImageDataProvider and ImageFileNameProvider to always use the image data retrieved from a native handle. To this end, it temporarily creates a handle if necessary. In order to avoid repeated loading and handle creation for the same source of image, a cache for the already retrieved image data is introduced. For both the consistent retrieval of image data and the avoidance of repeated loading of image data according test cases are added.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
@@ -16,6 +16,7 @@ package org.eclipse.swt.graphics;
 
 import java.io.*;
 import java.util.*;
+import java.util.function.*;
 
 import org.eclipse.swt.*;
 import org.eclipse.swt.internal.*;
@@ -2176,8 +2177,9 @@ private abstract class DynamicImageProviderWrapper extends AbstractImageProvider
 }
 
 private abstract class BaseImageProviderWrapper<T> extends DynamicImageProviderWrapper {
+	private final Map<Integer, ImageData> cachedImageData = new HashMap<>();
+
 	protected final T provider;
-	private final Map<Integer, Rectangle> zoomToBounds = new HashMap<>();
 
 	BaseImageProviderWrapper(T provider, Class<T> expectedClass) {
 		checkProvider(provider, expectedClass);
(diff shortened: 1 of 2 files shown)


akoch-yatta (approved):
Looks good to me. I tested around a bit and didn't find a regression caused by this.

HeikoKlare:
Test failure is known and documented: https://github.com/eclipse-platform/eclipse.platform.swt/issues/1843

sratz:
I noticed that this introduces behavior that might violate the contract of `Image.getImageData()`:

`... Modifications made to the returned {@code ImageData} will not affect this {@code Image}.`

But there is now the following behavior:

```
		ImageData initialImageData = new ImageData(1, 1, 24, new PaletteData(new RGB(0, 0, 0), new RGB(255, 255, 255)));

		Image image = new Image(new Display(), initialImageData);

		ImageData imageData1 = image.getImageData(100);

		imageData1.setPixel(0, 0, 1);

		ImageData imageData2 = image.getImageData(100);

		System.out.println("initial: " + initialImageData.hashCode() + ", value: " + initialImageData.getPixel(0, 0));
		System.out.println("1      : " + imageData1.hashCode() + ", value: " + imageData1.getPixel(0, 0));
		System.out.println("2      : " + imageData2.hashCode() + ", value: " + imageData2.getPixel(0, 0));
```
output:
```
initial: 1509514333, value: 0
1      : 991505714, value: 1
2      : 991505714, value: 1

```

So while modifying the `ImageData` retrieved from the first call to `getImageData()` did not affect the displayed `Image` itself, it still has an effect on a subsequent call to `getImageData()`.

Is this desired / expected?

HeikoKlare:
> Is this desired / expected?

That's definitely not expected but simply seems like a bug introduced with this PR. Thank you for the catch, @sratz!
I have created a fix together with a test based on your snippet:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2427

PR #1990, opened 2025-04-03
[GTK4] Implement Clipboard.getAvailableTypes()

ptziegler:
The list of available types is calculated using the GDK4 clipboard and
no longer falls back to the GTK3 bindings.

Note that UTF8_STRING, COMPOUND_TEXT and STRING_ID are not implemented
for GTK4, which is why those types must not be used in the TextTransfer,
regardless of whether X11 or Wayland is used.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Drag and Drop/gtk/org/eclipse/swt/dnd/Clipboard.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Drag and Drop/gtk/org/eclipse/swt/dnd/Clipboard.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2019 IBM Corporation and others.
+ * Copyright (c) 2000, 2025 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -570,7 +570,6 @@ public TransferData[] getAvailableTypes() {
 public TransferData[] getAvailableTypes(int clipboards) {
 	checkWidget();
 
-	//TODO: [GTK4] This currently will not work in GTK4
 	TransferData[] result = null;
 	if ((clipboards & DND.CLIPBOARD) != 0) {
 		int[] types = getAvailableClipboardTypes();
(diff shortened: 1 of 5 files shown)


ptziegler:
> Doesn't this PR practically disable clipboard functionality almost entirely?

Yes. But as mentioned, the clipboard is not implemented for GTK4. This method internally calls `getAvailableClipboardTypes()` and/or `getAvailablePrimaryTypes()`, which then calls `gtk_clipboard_wait_for_contents()`.

The last method uses GTK3-only methods which will always cause an UnsatisfiedLinkError.

https://github.com/eclipse-platform/eclipse.platform.swt/blob/0f21a6ef45900df1f564d52be3cbced83e1d399e/bundles/org.eclipse.swt/Eclipse%20SWT%20Drag%20and%20Drop/gtk/org/eclipse/swt/dnd/Clipboard.java#L696-L711

laeubi:
> But as mentioned, the clipboard is not implemented for GTK4.

I'm not sure if it is relevant but G**D**K4 seems to have a clipboard class.

https://docs.gtk.org/gdk4/class.Clipboard.html

and this seems to be the method to use for getting the formats:

https://docs.gtk.org/gdk4/method.Clipboard.get_formats.html

so can we sue this or is there a subtile difference between G**T**K and G**D**K that makes it impossible?

ptziegler:
The GDK clipboard is already used in some places but the implementation is not yet complete.

https://github.com/eclipse-platform/eclipse.platform.swt/blob/0f21a6ef45900df1f564d52be3cbced83e1d399e/bundles/org.eclipse.swt/Eclipse%20SWT%20Drag%20and%20Drop/gtk/org/eclipse/swt/dnd/Clipboard.java#L44-L51

> Would you look into that?

I can try to have a look what else is needed to complete this port, but I currently have to deal with allergy & stuff, so no promises...

ptziegler:
I've updated the PR to instead implement the GTK4 case instead.

[Recording.webm](https://github.com/user-attachments/assets/b4f9ed94-ec98-48e7-a745-cf4eeae79705)

ptziegler:
As a side note: I've tested this together with https://github.com/eclipse-platform/eclipse.platform.swt/pull/1984 and quite often the copy & paste actions in the popup menu don't work and the following error is logged.

> (SWT:140252): Gtk-WARNING **: 20:10:05.026: Broken accounting of active state for widget 0x72becc95af10(GtkPopoverMenu)

This seems to be a GTK4 bug: https://gitlab.gnome.org/GNOME/gtk/-/issues/4563

bundles/org.eclipse.swt/Eclipse SWT Drag and Drop/gtk/org/eclipse/swt/dnd/Clipboard.java
```diff
@@ -656,27 +655,36 @@ public String[] getAvailableTypeNames() {
...
+	int[] types = new int[gtypes_length];
+	for (int i = 0 ; i < gtypes_length ; ++i) {
+		long[] ptr = new long[1];
+		C.memmove(ptr, gtypes + i * C.PTR_SIZEOF, C.PTR_SIZEOF);
+		types[i] = (int) ptr[0];
+	}
+	return types;
+}
 
+int[] gtk3_getAvailableTypes(long clipboard) {
```
akurtakov, line 685: Is there a reason for this one and the gtk4 one to not be private?

ptziegler: No, that's merely an oversight. The distinction between GTK3 and GTK4 should be purely internal.
I've added the modifier to both methods.

akurtakov:
For some reason the Linux GHA seems to switched to on run GTK4. As it's not due to this change and it actually fixes a crash I am ignoring it and merging.

Issue #1991, opened 2025-04-04
[Test] [Windows] test_LocationListener_ProgressListener_cancledLoad is flaky


fedejeanne:
**Describe the bug**
The test `org.eclipse.swt.tests.junit.Test_org_eclipse_swt_browser_Browser.test_LocationListener_ProgressListener_cancledLoad()` fails intermittently in GH actions (probably also in CI builds)

```java
Unexpected event fired. 
LocationChanging (should be true): true
LocationChanged unexpectedly (should be false): true
ProgressChanged unexpectedly (should be false): true

	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.assertTrue(Assert.java:42)
	at org.eclipse.swt.tests.junit.Test_org_eclipse_swt_browser_Browser.test_LocationListener_ProgressListener_cancledLoad(Test_org_eclipse_swt_browser_Browser.java:727)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
```

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS

fedejeanne:
Might have the same root cause as https://github.com/eclipse-platform/eclipse.platform.swt/issues/1843

PR #1993, opened 2025-04-04
Consider about:blank as a legal "empty" location in test #1991

fedejeanne:
This approach needs to be discussed, **don't merge it yet!**

(Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/1991)


fedejeanne:
Since the issue is that now the asynchronous execution of some methods makes canceling calls impossible, this PR is wrong. If anything, one revert the decision of doing things asynchronously (which requires major changes) or accept it and remove tests accordingly.

PR #1996, opened 2025-04-04
[Win32] Ensure consistent image data returned for data-based images

HeikoKlare:
Due to the on-demand creation of image handles, there is not necessarily a handles anymore from which image data is retrieved when requesting is via the getImageData(...) methods. This results in potentially different kinds of image data (including different anti-aliasing results) depending on whether a handle has already been created for an image at the given zoom or not.

This change adapts the implementation of Image based on static ImageData and streams to always use the image data retrieved from a native handle. To this end, it temporarily creates a handle if necessary. In order to avoid repeated loading and handle creation for the same source of image, a cache for the already retrieved image data is introduced.

<s>⚠️ This is based on and should thus be merged after https://github.com/eclipse-platform/eclipse.platform.swt/pull/2036</s>

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2052

Supercedes and thus closes #2051

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
@@ -1878,14 +1878,18 @@ private abstract class AbstractImageProviderWrapper {
 
 	protected ImageHandle newImageHandle(int zoom) {
 		ImageData resizedData = getImageData(zoom);
-		if (type == SWT.ICON && resizedData.getTransparencyType() != SWT.TRANSPARENCY_MASK) {
+		return newImageHandle(resizedData, zoom);
+	}
+
+	protected final ImageHandle newImageHandle(ImageData data, int zoom) {
+		if (type == SWT.ICON && data.getTransparencyType() != SWT.TRANSPARENCY_MASK) {
 			// If the original type was an icon with transparency mask and re-scaling leads
 			// to image data without transparency mask, this will create invalid images
 			// so this fallback will "repair" the image data by explicitly passing
 			// the transparency mask created from the scaled image data
-			return initIconHandle(device, resizedData, resizedData.getTransparencyMask(), zoom);
+			return initIconHandle(device, data, data.getTransparencyMask(), zoom);
 		} else {
-			return init(resizedData, zoom);
+			return init(data, zoom);
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1948,39 +1944,43 @@ protected void destroy() {
...
 	}
 
 	@Override
-	ImageData getImageData(int zoom) {
-		if (zoomLevelToImageHandle.containsKey(zoom)) {
-			return zoomLevelToImageHandle.get(zoom).getImageData();
-		}
-		if (!zoomLevelToImageHandle.isEmpty()) {
-			return getScaledImageData(zoom);
-		}
```
HeikoKlare, line 1969: Removing this results in not calling the `memGC` check anymore. However, I wonder whether was correct/useful anyway: it ensures that when `memGC` is set (i.e., when a GC draws on the image), the image data at a zoom according to `initialNativeZoom` is returned. However, this `initialNativeZoom` is just set to the native device zoom at image creation, but the actual "initial" data must not be in that zoom. Actually, for all `ImageFromImageDataProviderWrappers`, the "original" zoom of all data rather seems to be 100. So I wonder whether we somehow need to adopt the `memGC` functionality here to return the 100% image data?

akoch-yatta: My plan would be to get rid of `initialNativeZoom` as it is only used for the GC anymore, if I don't forget anything. 
As we learned by now, that GC works best at 100%, having the default use 100% makes most sense in my opinion. Second usage is the ImageGCDrawer and there we can do the configuration in the Wrapper

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1251,11 +1251,7 @@ public ImageData getImageData (int zoom) {
 	if (zoomLevelToImageHandle.containsKey(zoom)) {
 		return zoomLevelToImageHandle.get(zoom).getImageData();
 	}
-	if (imageProvider != null) {
-		return imageProvider.getImageData(zoom);
-	}
-
-	return getScaledImageData(zoom);
+	return imageProvider.getImageData(zoom);
```
akoch-yatta, line 1254: Isn't imageProvider == null, when `public static Image win32_new(Device device, int type, long handle, int nativeZoom)` is used?

akoch-yatta (commented):
Code looks good to me except the one question/remark, I have. I still need to do some testing.

akoch-yatta:
@HeikoKlare As I already had the change drafted some time back, I wrapped that up and created https://github.com/eclipse-platform/eclipse.platform.swt/pull/2027 to make sure imageProvider is always != null, which makes my comment not relevant anymore. I would propose to process that one first

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1948,39 +1944,43 @@ protected void destroy() {
...
-		if (zoomLevelToImageHandle.containsKey(zoom)) {
-			return zoomLevelToImageHandle.get(zoom).getImageData();
-		}
-		if (!zoomLevelToImageHandle.isEmpty()) {
-			return getScaledImageData(zoom);
-		}
-		ElementAtZoom<ImageData> loadedImageData = loadImageData(zoom);
-		return DPIUtil.scaleImageData(device, loadedImageData, zoom);
+	final ImageData getImageData(int zoom) {
+		Function<Integer, ImageData> imageDataRetrival = zoomToRetrieve -> {
```
fedejeanne, line 1959: NIT: typo
```suggestion
		Function<Integer, ImageData> imageDataRetrieval = zoomToRetrieve -> {
```

fedejeanne (commented):
I tested this and it looks good. I would even change the commit-text and add the it contributes (or even fixes) https://github.com/eclipse-platform/eclipse.platform.swt/issues/2052. I'm not sure if one can call it a "fix" because I always compare these 2 constructors of `Image`:
1. `public Image(Device device, ImageData data) {...}`
2. `public Image(Device device, Image srcImage, int flag) {...}`

and the 2nd one produces sharper images. 

That being said, I also tested constructor nr. 1 in a `for`-loop, like this:

```java
Image image = new Image(cachedFolder.getDevice(), clonedImageData);
for (int i=0; i<15; i++) {
	// Make it blurry
	image = new Image(image.getDevice(), image.getImageData());
}

return image;
```
And it doesn't matter how many times I iterate, the returned image has always the same quality/sharpness.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
HeikoKlare: This became obsolete via https://github.com/eclipse-platform/eclipse.platform.swt/pull/2027

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
HeikoKlare: Thanks! Fixed with latest update of the PR.

HeikoKlare:
I have rebased this on the cleanup PR:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2027

This required minor adaptations as the adapted image provider wrapper needs to take into account the specific handling for ICON-type images to avoid errors during smooth scaling.

fedejeanne (approved):
I retested it and I saw no problems with this PR ✔️

PR #2004, opened 2025-04-08
Replace usages of new Image(device, width, height) for Composite

ShahzaibIbrahim:


fedejeanne (approved):

fedejeanne (approved):
@ShahzaibIbrahim thank you, the changes LGTM.

fyi I made one slight alteration and let `paintGC` be initialized inside of the `if` (just like it was before). The reason for this is that I wouldn't want to initialize it if the style-bit `SWT.TRANSPARENT` is not set, which was a pre-requisite before your PR.

fedejeanne:
Failing test is unrelated #1843

fedejeanne:

PR #2007, opened 2025-04-13
Fix localization filenames and preserve signature of swt.jar

HannesWell:
In the SWT fragment jars published in the Eclipse p2-repository the localization files had the wrong names, which are used at dev-time to avoid localization when launching SWT from a development environment.

Furthermore the 'swt.jar' is now not modified anymore before adding it to the assembly for the download-page, which ensures intact signatures.

Fixes https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2189

### DIFF
--- a/binaries/pom.xml
+++ b/binaries/pom.xml
@@ -76,6 +76,11 @@
 				<artifactId>tycho-packaging-plugin</artifactId>
 				<configuration>
 					<timestampProvider>fragment-host</timestampProvider>
+					<archive>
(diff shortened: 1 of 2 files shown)


HannesWell:
Results look good, although the jars are not signed in the verification build but the jar built for the p2-repo and the one included in the SWT.zip published at the download page are identical.

With tonight's I-build the published zips should have the expected content.

Issue #219, opened 2022-06-21
[GTK4] GtkEventControllerMotion Enter/Leave events are not handled 


joel-majano:
**Describe the bug**

As part of the migration to GTK4, Enter/Leave procs were created, but they haven't been hooked up to do what they're supposed to to replicate GTK3 behavior. This causes things like shell tooltips to not work as these were previously handled by enter_notify_event handlers.

**To Reproduce**

Snippet 1, with the line `shell.setToolTipText("any text");` added

**Expected behavior**

Shell tooltip is shown when mouse hovered on top of shell.

**Environment:**
1. Select the platform(s) on which the behavior is seen:

- Linux

2. Additional OS info (e.g. OS version, Linux Desktop, etc)

- Fedora 36, GTK 4.6.5

4. JRE/JDK version
 
- OpenJDK 17

Tracked in #228

joel-majano:
This continues work from https://git.eclipse.org/r/c/platform/eclipse.platform.swt/+/183457 and https://bugs.eclipse.org/bugs/show_bug.cgi?id=539524.

PR #2008, opened 2025-04-13
[GTK4] Cleanup GtkEventControllerMotion Enter/Leave events

ptziegler:
Support for mouse motion events was already implemented with b70cd16c42914b617e770ad76a8e951fa6ebf325. However, following edge cases were missed:

- Moving the cursor out of the shell and back in doesn't fire a " mouse enter" event.

Similarly to the way motion events are handled in GTK3, the current control needs to be cleared on a "leave" event. Otherwise the "enter" event is suppressed when the cursor is moved out of and back in a shell again.

- Moving from one widget to another fires two "mouse exit" events.

One event is fired by "gtk4_motion_event()", the other by "leaveProc()". Only the former should be fired, as it contains the proper "x" and "y" coordinates.

Latter is fired because "getCursorControl()" always returns "null". This has been fixed by first calculating the GtkWindow from the GdkSurface and then selecting the widget at the mouse coordinates. The whole implementation has also been restructured so that the X11-based heuristic via the XQueryPointer is only used for GTK3 (because it would simply return early for GTK4 anyway).

- The tooltip window isn't moved when hovering over different widget

This is because the tooltip is set for the shell, even though it should be set for the individual widgets. As a result, only the text is updated but the window remains where it was initially shown.

Closes https://github.com/eclipse-platform/eclipse.platform.swt/issues/219


ptziegler:
You can use the following snippet for testing:

```java
public class Snippet {

		shell.setSize(200, 200);
		shell.setText("Mouse Enter/Exit");
		shell.setToolTipText("Shell");
		shell.setLayout(new GridLayout(2, true));
		shell.addMouseTrackListener(new MouseTrackAdapter() {
			@Override
			public void mouseEnter(MouseEvent e) {
				System.out.println("mouseEnter: %s, x=%d, y=%d".formatted("Shell", e.x, e.y));
			}

			@Override
			public void mouseExit(MouseEvent e) {
				System.out.println("mouseExit: %s, x=%d, y=%d".formatted("Shell", e.x, e.y));
			}
		});

		createButton(shell, "Button 1");
		createButton(shell, "Button 2");
		createButton(shell, "Button 3");
		createButton(shell, "Button 4");

		shell.open ();
		}
		display.dispose ();
	}

	private static void createButton(Shell parent, String text) {
		GridData gd = new GridData();
		gd.horizontalIndent = 20;
		gd.verticalIndent = 20;
		gd.widthHint = 60;
		gd.heightHint = 60;

		Button b = new Button(parent, SWT.PUSH);
		b.setText(text);
		b.setToolTipText(text);
		b.setLayoutData(gd);
		b.addMouseTrackListener(new MouseTrackAdapter() {
			@Override
			public void mouseEnter(MouseEvent e) {
				System.out.println("mouseEnter: %s, x=%d, y=%d".formatted(text, e.x, e.y));
			}

			@Override
			public void mouseExit(MouseEvent e) {
				System.out.println("mouseExit: %s, x=%d, y=%d".formatted(text, e.x, e.y));
			}
		});
	}
}
```

Expectation:

When moving the mouse cursor from the shell to one of the buttons, exactly two events should be fired:
- mouseExit (Shell)
- mouseEnter (Button)

When moving the mouse cursor from one of the buttons back to the shell, exactly two events should be fired:
- mouseExit (Button)
- mouseEnter (Shell)

When moving the mouse cursor in and out of the shell, exactly two events should be fired:
- mouseExit (Shell)
- mouseEnter (Shell)

ptziegler:
> The tooltip window isn't moved when hovering over different widget

[Recording.webm](https://github.com/user-attachments/assets/368c4f8b-3e66-495d-9f57-7447411bb1b2)

akurtakov:
Very nice fix. Failure is due to https://github.com/eclipse-platform/eclipse.platform.swt/issues/2014 . Merging.

PR #2013, opened 2025-04-14
Edge: Tests: Add test-case for evaluate opening new window

sratz:
See #1771
See #1973


sratz:
Crash in test `Test_org_eclipse_swt_browser_Browser#test_OpenWindowListener_openHasValidEventDetails` on Ubuntu is unrelated.

Issue #2014, opened 2025-04-14
Test_org_eclipse_swt_browser_Browser#test_OpenWindowListener_openHasValidEventDetails failing on Ubuntu since 2025-04-10


sratz:
https://ci.eclipse.org/releng/job/eclipse.platform.swt/job/master/1086/consoleFull

1119bac8cd800531d58ef028932193f359384136

```
20:57:36  Running Test_org_eclipse_swt_browser_Browser#test_OpenWindowListener_openHasValidEventDetails[browser flags: 0]
20:57:36  
20:57:36  ### Descriptors opened BEFORE test_OpenWindowListener_openHasValidEventDetails[browser flags: 0]: 137
20:57:36  
20:57:36  ** (My Application Name:782): WARNING **: 18:57:36.532: WebKitWebView returned by WebKitWebView::create signal was not created with the related WebKitWebView
20:57:36  
20:57:36  ### Descriptors opened AFTER test_OpenWindowListener_openHasValidEventDetails[browser flags: 0]: 140
20:57:36  Delta to previous test: 3
20:57:36  
20:57:36  Delta to first test: -29
```

https://ci.eclipse.org/releng/job/eclipse.platform.swt/job/master/1089/consoleFull

4466c96fcf23d1c22770288e97f9f2b2e4b68676

```
08:34:51  Running Test_org_eclipse_swt_browser_Browser#test_OpenWindowListener_openHasValidEventDetails[browser flags: 0]
08:34:52  
08:34:52  ### Descriptors opened BEFORE test_OpenWindowListener_openHasValidEventDetails[browser flags: 0]: 137
08:34:52  /usr/include/c++/13/optional:486: constexpr const _Tp& std::_Optional_base_impl<_Tp, _Dp>::_M_get() const [with _Tp = WebCore::WindowFeatures; _Dp = std::_Optional_base<WebCore::WindowFeatures, false, false>]: Assertion 'this->_M_is_engaged()' failed.
[...]

PR #2015, opened 2025-04-15
Do not treat warnings as errors in gtk4 port

akurtakov:
The port is in early stages and there are still many deprecated usages that should be ported rather than silenced (as in the gtk3 port).


akurtakov:
Failure is due to https://github.com/eclipse-platform/eclipse.platform.swt/issues/2014 . Merging.

Issue #2017, opened 2025-04-15
Tooltips not shown reliably for CTabFolder tabs with gtk3-3.24.31


iloveeclipse:
Found after updating from RHEL 7.9.
To reproduce: in the IDE, open few editors and hover over the editor tabs.
* On RHEL7.9 the tab tooltip always appears and shows the editor name/path.
* On RHEL7.9 the tab tooltip mostly does not appear and it requires several attempts to get the tooltip show up. This works reliably only if first hover over Java editor breadcrumb area to show tooltip there and then quickly move the mouse to the tab area. In this case tooltip is shown.

This is not a recent regression, it seem to be there at least up to 4.29 (didn't investigated further).

RHEL 7.9
```
rpm -q gtk3
gtk3-3.22.30-8.el7_9.x86_64
```
RHEL 9.2.
```
rpm -q gtk3
gtk3-3.24.31-2.el9.x86_64
```
It **seems** that the problem is in CTabFolder code that nulls out the tooltip before showing the new one:

https://github.com/eclipse-platform/eclipse.platform.swt/blob/f734dc67d78c0bdc2d7d90aafffa23d6ec57ccb7/bundles/org.eclipse.swt/Eclipse%20SWT%20Custom%20Widgets/common/org/eclipse/swt/custom/CTabFolder.java#L1817

Changing this to the `_setToolTipText(event.x, event.y);` fixes the problem.

I will push a fix & also check that on RHEL 7.9 .

PR #2018, opened 2025-04-15
CTabFolder: don't unset tooltip value on mouse enter

iloveeclipse:
Setting tooltip to null on mouse enter results in an unpredictable GTK tooltip behavior which in most cases simply doesn't show at all.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2017

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
@@ -1813,10 +1813,6 @@ void onMouse(Event event) {
 	}
 	int x = event.x, y = event.y;
 	switch (event.type) {
-		case SWT.MouseEnter: {
-			_setToolTipText(event.x, event.y);
-			break;
-		}
 		case SWT.MouseExit: {
 			for (int i=0; i<items.length; i++) {
 				CTabItem item = items[i];
@@ -1910,6 +1906,10 @@ void onMouse(Event event) {
 				return;
 			}
 			break;
+		}
+		case SWT.MouseEnter: {
+			// fall through to the "move" case, see
+			// https://github.com/eclipse-platform/eclipse.platform.swt/issues/2017
 		}
 		case SWT.MouseMove: {
 			_setToolTipText(event.x, event.y);
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
@@ -1814,7 +1814,7 @@ void onMouse(Event event) {
 	int x = event.x, y = event.y;
 	switch (event.type) {
 		case SWT.MouseEnter: {
-			setToolTipText(null);
+			_setToolTipText(event.x, event.y);
 			break;
 		}
 		case SWT.MouseExit: {


bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
```diff
@@ -1814,7 +1814,7 @@ void onMouse(Event event) {
 	int x = event.x, y = event.y;
 	switch (event.type) {
 		case SWT.MouseEnter: {
-			setToolTipText(null);
+			_setToolTipText(event.x, event.y);
```
laeubi, line 1817: I wonder if this should simply be merged with the `SWT.MouseMove` case?

iloveeclipse: The event coordinates are different, depending on the coordinates different tooltip can be shown.

iloveeclipse:
@akurtakov : any idea if GTK build has known issues?

```
11:54:15  ### Descriptors opened BEFORE test_OpenWindowListener_openHasValidEventDetails[browser flags: 0]: 139
11:54:15  /usr/include/c++/13/optional:486: constexpr const _Tp& std::_Optional_base_impl<_Tp, _Dp>::_M_get() const [with _Tp = WebCore::WindowFeatures; _Dp = std::_Optional_base<WebCore::WindowFeatures, false, false>]: Assertion 'this->_M_is_engaged()' failed.
11:55:05  /home/jenkins/agent/workspace/eclipse.platform.swt_PR-2018/eclipse.platform.swt@tmp/durable-48730c15/script.sh.copy: line 2:   799 Aborted                 (core dumped) mvn clean verify --batch-mode --threads 1C -V -U -e -DforkCount=0 -Papi-check -Dcompare-version-with-baselines.skip=false -Dorg.eclipse.swt.tests.junit.disable.test_isLocal=true -Dmaven.test.failure.ignore=true -Dmaven.test.error.ignore=true
script returned exit code 134
```

akurtakov:
It's currently broken due to https://github.com/eclipse-platform/eclipse.platform.swt/issues/2014

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
iloveeclipse: > seems to fix more cases even though a move is likely occurs short after an enter.

Not sure which cases you mean?
Interestingly I see "enter" reported after "move" on RHEL 9... I'm still playing with that, trying to see what happens on RHEL 7.9 differently.

laeubi: > Not sure which cases you mean?

If enter causes issue, I suspect that handling it differently might lead to other issues (e.g. close icon not properly updated), so I think just handle enter like _move_ or dosen't handle _enter_ at all?

> Interestingly I see "enter" reported after "move" on RHEL 9... 

Yes sometimes events are "tricky" on some OS as the OS might skip/delay some events, maybe that's the real issue here, the tooltip is set because of _move_ but then cleared by a later _enter_ happening.

iloveeclipse: >  I see "enter" reported after "move" on RHEL 9

Correction: it is **additionally** reported after tooltip is shown!

So I guess it is not about GTK but about KDE update (or both).

Here is the events we see on RHEL 9.2:
```
Enter
Move
Move
Move
Hover
Down
Exit
Enter
```

The two last "Exit, Enter" entries are **not present** on RHEL 7.9.

I guess they are generated by the tooltip shell that **should** "show up" and on RHEL 9.2 GTK sends "exit" because of that and "enter" yet again probably to set focus back into original shell. Now with the old code that nulls out the tooltip text the GTK code that is supposed to draw tooltip probably detects that there is nothing to draw and doesn't show anything. This also explains why it is **sometimes** works - namely, if the tooltip shell is already opened, it is preserved on the screen and doesn't cause context switch and  "Exit, Enter" events, so the tooltip is displayed.

iloveeclipse: I've experimented with this patch and with merging I don't see visual differences. 

But I assume there is a chance that we get an "enter" not following by a "move" (for the same reason we do get "exit" after "move" directly followed by "enter") but still something need to be drawn. So I will commit that as a dedicated commit to this PR, just in case it could cause some regression and so could be easily reverted.

iloveeclipse:
Browser test failure on Windows is unrelated, merging.

PR #2019, opened 2025-04-15
Support cascaded java callbacks in Edge #1919

fedejeanne:
Add a test to it.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/1919


amartya4256 (approved):
Looks good to me :)

amartya4256:
However, the test fails for Linux.

fedejeanne:
> However, the test fails for Linux.

I noticed, let's put a pin on it: #2021

Issue #2001, opened 2025-04-07
Test_org_eclipse_swt_graphics_Image test fails since I20250406-1800


iloveeclipse:
See https://download.eclipse.org/eclipse/downloads/drops4/I20250406-1800/testResults.php

Regression from https://github.com/eclipse-platform/eclipse.platform.swt/pull/1999

Typical fail looks like:

```
test_ConstructorLorg_eclipse_swt_graphics_Device_ImageDataProvider	Error	Missing scheme

java.lang.IllegalArgumentException: Missing scheme
at java.base/java.nio.file.Path.of(Path.java:200)
at org.eclipse.swt.tests.junit.SwtTestUtil.getPath(SwtTestUtil.java:600)
at org.eclipse.swt.tests.junit.Test_org_eclipse_swt_graphics_Image.lambda$1(Test_org_eclipse_swt_graphics_Image.java:77)
at org.eclipse.swt.internal.DPIUtil.lambda$2(DPIUtil.java:546)
at org.eclipse.swt.internal.DPIUtil.getElementAtZoom(DPIUtil.java:575)
at org.eclipse.swt.internal.DPIUtil.validateAndGetImageDataAtZoom(DPIUtil.java:546)
at org.eclipse.swt.graphics.Image.initFromImageDataProvider(Image.java:788)
at org.eclipse.swt.graphics.Image.<init>(Image.java:653)
at org.eclipse.swt.tests.junit.Test_org_eclipse_swt_graphics_Image.test_ConstructorLorg_eclipse_swt_graphics_Device_ImageDataProvider(Test_org_eclipse_swt_graphics_Image.java:372)
at java.base/java.lang.reflect.Method.invoke(Method.java:580)
```

HannesWell:
Yes, I'll try to fix it asap. Sorry for the delay.

HannesWell:
After looking into this and considering the error message (`Error Missing scheme`) and the stack-trace, that's what I think is happening:
In I-build tests the system-property `PLUGIN_PATH` is set to the root of the project in
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/e68f158ea30b4eef72c79011bc099f3505d67109/eclipse.platform.releng.tychoeclipsebuilder/eclipse-junit-tests/src/main/scripts/library.xml#L275
or
[...]

PR #2024, opened 2025-04-15
Resolve resources URLs as file-URLs if running in an OSGi environment

HannesWell:
Don't consider the 'PLUGIN_PATH' property anymore, which is set in I-build tests and whose handling doesn't work anymore.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2001 (hopefully).

### DIFF
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/SwtTestUtil.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/SwtTestUtil.java
@@ -20,12 +20,10 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
-import static org.junit.jupiter.api.Assertions.assertTrue;
 
+import java.io.IOException;
+import java.io.InputStream;
 import java.io.PrintStream;
-import java.net.URI;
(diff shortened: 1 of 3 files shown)


HeikoKlare (commented):
Is it good to add an Equinox dependency to the SWT tests? I find that a bit strange since those tests are (outside of an I-Build) executed as plain JUnit tests without any OSGi environment.

In case this is reasonable, we should probably remove the `PLUGIN_PATH` property from the build scripts, as that one is only used for the SWT tests, isn't it?

HannesWell:
> Is it good to add an Equinox dependency to the SWT tests? I find that a bit strange since those tests are (outside of an I-Build) executed as plain JUnit tests without any OSGi environment.

That was indeed not ideal and I just used that since avoiding it seemed cumbersome and it worked in all cases (by chance) as described initially because the class-loader then already returned file URLs.
But I tried it again and turned out to not be that complicated eventually.

> 
> In case this is reasonable, we should probably remove the `PLUGIN_PATH` property from the build scripts, as that one is only used for the SWT tests, isn't it?

Good point. I checked all the SDK git repos and didn't found another use of that property. So I think we can remove it. I'll create a follow up in the `platform.releng.aggregator` repo once we have a confirmation that these tests work again.

HannesWell:
> In case this is reasonable, we should probably remove the `PLUGIN_PATH` property from the build scripts, as that one is only used for the SWT tests, isn't it?

For the record, see https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2989.

PR #2027, opened 2025-04-16
[win32] Wrapper for existing image handles

akoch-yatta:
This PR adds a wrapper for images created with an existing OS handle to ensure all existing image constructors are tied to a final wrapper instance.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
@@ -111,7 +111,7 @@ public final class Image extends Resource implements Drawable {
 	/**
 	 * AbstractImageProvider to avail right ImageProvider (ImageDataProvider or ImageFileNameProvider)
 	 */
-	private AbstractImageProviderWrapper imageProvider;
+	private final AbstractImageProviderWrapper imageProvider;
 
 	/**
 	 * Style flag used to differentiate normal, gray-scale and disabled images based
@@ -138,16 +138,13 @@ public final class Image extends Resource implements Drawable {
 
 	private Map<Integer, ImageHandle> zoomLevelToImageHandle = new HashMap<>();
 
-/**
- * Prevents uninitialized instances from being created outside the package.
- */
-Image (Device device) {
-	this(device, DPIUtil.getNativeDeviceZoom());
-}
-
-private Image (Device device, int nativeZoom) {
+private Image (Device device, int type, long handle, int nativeZoom) {
 	super(device);
 	initialNativeZoom = nativeZoom;
+	this.type = type;
+	this.imageProvider = new ExistingImageHandleProviderWrapper(handle, nativeZoom);
+	this.isInitialized = true;
+	this.device.registerResourceWithZoomSupport(this);
 }
 
 /**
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1947,6 +1938,56 @@ protected void destroy() {
...
+	}
+
+	@Override
+	ImageHandle getImageMetadata(int zoom) {
+		if (zoomLevelToImageHandle.containsKey(zoom)) {
+			return zoomLevelToImageHandle.get(zoom);
+		} else {
+			ImageData resizedData = getImageData(zoom);
+			ImageData newData = adaptImageDataIfDisabledOrGray(resizedData);
+			init(newData, zoom);
```
HeikoKlare, line 1980: In the `Image#getImageMetadata(int)` method, there is a special initialization handling for the case that the image is an icon. What if the image created with this new provider is such an icon-type image, won't the initialization potentially lead to wrong results?
Or even more: the `imageProvider == null` case in the `Image#getImageMetadata(int)` method should become completely obsolete by this change, so it should be removed and there and completely be moved into this method, shouldn't it?

HeikoKlare (commented):
In general, the change looks sound to me. I was wondering if we can avoid creating a new type of provider by just using the image data of the given handle, but if I am not mistaken that's not possible (or not a good idea at least), because the current implementation allows to share handles between Image instances, which is also why disposed resource tracking is not instantiated for an Image created via an existing handle. Is that correct?

This kind of change seems to allow for more cleanups. The follow methods have implementations for `imageProvider == null`, which should become obsolete by this change and can be removed, can't they?
- `equals` / `hashCode`
- `getBounds`
- `getImageData`
- `getImageMetaData` (see detailled comment)

akoch-yatta:
> In general, the change looks sound to me. I was wondering if we can avoid creating a new type of provider by just using the image data of the given handle, but if I am not mistaken that's not possible (or not a good idea at least), because the current implementation allows to share handles between Image instances, which is also why disposed resource tracking is not instantiated for an Image created via an existing handle. Is that correct?

Yes, we definitely need to use the passed OS handle, because it is created outside of the control of the Image. Therefor I did not see another approach as creating a separate wrapper for it

> 
> This kind of change seems to allow for more cleanups. The follow methods have implementations for `imageProvider == null`, which should become obsolete by this change and can be removed, can't they?
> 
> * `equals` / `hashCode`
> * `getBounds`
> * `getImageData`
> * `getImageMetaData` (see detailled comment)

Yes, but I wanted to do that cleanup separately.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: Hmm, regarding the first point. probably yes, but isn't this an issue with e.g. MaskedImageDataProviderWrapper as well?
Second point. Yes, question to me is, when and how do we do the cleanup.

HeikoKlare (approved):

HeikoKlare:
Merging this despite Jenkins failure caused by https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/5892

Since master builds seem to still work on Jenkins, the subsequent master build should hopefully work, thus I will check results of that build to retrospectively validate this PR.

Issue #2031, opened 2025-04-17
Progress bar is off in splash screen


fedejeanne:
**Describe the bug**
The progress bar is off when I start Eclipse with a zoom level of 175%

_Hint: Ignore the black splash screen, it has already been reported here: for https://github.com/eclipse-platform/eclipse.platform.ui/issues/2897._

![Image](https://github.com/user-attachments/assets/38a1dc85-c8f6-4678-b283-d8218a256dfc)

**To Reproduce**
- Set the zoom level in Windows to 175%
- Monitor-specific scaling should be **active**
- Restart Eclipse

**Expected behavior**
It should look like this:

_Hint: again, ignore the black splash screen_

![Image](https://github.com/user-attachments/assets/6dfac918-a2de-4e2c-818d-046f9e82256d)

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [x] Windows
- - [ ] Linux (Haven't tested it)
- - [ ] macOS (Haven't tested it)

**Version since**
Caused by https://github.com/eclipse-platform/eclipse.platform.swt/pull/2030 (@akurtakov )

HeikoKlare:
> Caused by [#2030](https://github.com/eclipse-platform/eclipse.platform.swt/pull/2030) ([@akurtakov](https://github.com/akurtakov) )

Is the issue caused by setting `useCairoAutoScale` to true in `DPIUtil` even on non-Linux systems? So isn't the solution simply to initialize `useCairoAutoScale = Platform.PLATFORM.equals("gtk")` (and in the best case make that value a constant (i.e. `final`)?

akoch-yatta:
> > Caused by [#2030](https://github.com/eclipse-platform/eclipse.platform.swt/pull/2030) ([@akurtakov](https://github.com/akurtakov) )
> 
> Is the issue caused by setting `useCairoAutoScale` to true in `DPIUtil` even on non-Linux systems? So isn't the solution simply to initialize `useCairoAutoScale = Platform.PLATFORM.equals("gtk")` (and in the best case make that value a constant (i.e. `final`)?

Probably yes.

HeikoKlare:
Yeah, #2030 broke auto-scaling completely, not only for the splash screen:

![Image](https://github.com/user-attachments/assets/68d4a192-e38d-4255-936e-85a6f5e0d3f2)

Should be easily resolved via:
[...]

PR #2032, opened 2025-04-17
Correct Cairo autoscaling setting in DPIUtil and remove setter #2031

HeikoKlare:
Cairo auto-scaling was enabled in DPIUtil for every OS in a recent commit. This corrects the setting to be only applied on Linux/GTK and thus also issues arising from it. In addition, it removes the unnecessary setter for the auto-scaling and disables the tests that previously required it as they are meaningless without productive use of disabled Cairo auto-scale on Linux anyway.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2031

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
@@ -47,7 +47,7 @@ public class DPIUtil {
 	private static AutoScaleMethod autoScaleMethod = AutoScaleMethod.NEAREST;
 
 	private static String autoScaleValue;
-	private static boolean useCairoAutoScale = true;
+	private static final boolean USE_CAIRO_AUTOSCALE = SWT.getPlatform().equals("gtk");
 
 	/**
 	 * System property that controls the autoScale functionality.
@@ -484,7 +484,7 @@ public static Rectangle scaleUp(Drawable drawable, Rectangle rect, int zoom) {
  * @return float scaling factor
  */
 private static float getScalingFactor(int zoom) {
-	if (useCairoAutoScale) {
+	if (USE_CAIRO_AUTOSCALE) {
 		return 1;
 	}
 	if (zoom <= 0) {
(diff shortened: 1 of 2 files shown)


akurtakov:
Jenkins failure is due to https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/5892

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
```diff
@@ -47,7 +47,7 @@ private static enum AutoScaleMethod { AUTO, NEAREST, SMOOTH }
 	private static AutoScaleMethod autoScaleMethod = AutoScaleMethod.NEAREST;
 
 	private static String autoScaleValue;
-	private static boolean useCairoAutoScale = true;
+	private static final boolean useCairoAutoScale = Platform.PLATFORM.equals("gtk");
```
fedejeanne, line 50: How about:
```suggestion
	private static final boolean IS_LINUX = SWT.getPlatform().equals("gtk");
```

My reasoning:
1. The field became a constant --> use uppercase
2. "Cairo" doesn't seem to bring any useful information anymore, it's just an internal detail about using Linux/Gtk --> exchange "cairo" for "linux" in the name
3. `SWT.getPlatform()` is used in other parts of the class but `Platform.PLATFORM` isn't. You can also choose to change the other parts to `Platform.PLATFORM` and use the same convention everywhere (and reduce 1 level of indirection)

HeikoKlare:
I would be confident with merging this change without a successful Jenkins run. GH actions succeeded at third attempt, Jenkins still fails. Would be nice to have this merged to avoid test failures in other PRs. @akurtakov would you agree with merging this?

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
HeikoKlare: Thanks for the pointers. I incorporated 1 and 3. I did not implement 2 as this flag still indicates whether Cairo's auto-scaling is used (which is identified by being on Linux as no other scaling mode is currently used there). Changing to `IS_LINUX` removes semantics, as you will only see that on Linux a scale factor of 1 is always returned without knowing about the reasons (which is that Cairo's auto-scaling is used).

akurtakov:
Let's merge once GHA confirm it.

PR #2036, opened 2025-04-17
[win32] Cleanup image wrappers

akoch-yatta:
This PR refactors Image in the win32 implementation as Image#imageProvider is final and always != null now. With this the handling and creation of new handles and ImageData was unified.

Relies on https://github.com/eclipse-platform/eclipse.platform.swt/pull/2027 being merged first and includes this commit because of that.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
@@ -780,23 +780,7 @@ private ImageHandle getImageMetadata(int zoom) {
 	if (zoomLevelToImageHandle.get(zoom) != null) {
 		return zoomLevelToImageHandle.get(zoom);
 	}
-	if (imageProvider != null) {
-		return imageProvider.getImageMetadata(zoom);
-	} else {
-		ImageData resizedData = getImageData(zoom);
-		ImageData newData = adaptImageDataIfDisabledOrGray(resizedData);
-		if (type == SWT.ICON && newData.getTransparencyType() != SWT.TRANSPARENCY_MASK) {
-			// If the original type was an icon with transparency mask and re-scaling leads
-			// to image data without transparency mask, this will create invalid images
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -780,23 +780,7 @@ private ImageHandle getImageMetadata(int zoom) {
 	if (zoomLevelToImageHandle.get(zoom) != null) {
 		return zoomLevelToImageHandle.get(zoom);
 	}
-	if (imageProvider != null) {
-		return imageProvider.getImageMetadata(zoom);
-	} else {
-		ImageData resizedData = getImageData(zoom);
-		ImageData newData = adaptImageDataIfDisabledOrGray(resizedData);
```
HeikoKlare, line 787: This was removed at the place where this code was moved to. Was that intended, i.e., it is sufficient to only do the adaptation at the remaining places?

HeikoKlare (commented):
The overall change looks sound and is great cleanup step.
I only have one question regarding the adaptation for disabled/gray images.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: I think, it was not completly correct. I want to cover all places, where scaled variants of an ImageData is created and those _could_ be not adapted. My latest commit removed an unnecessary call (calling place already does it) and adds one call.

HeikoKlare (approved):

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: Yes, I agree. We should try to better streamline ImageData creation/scaling in Image

HeikoKlare:

Issue #2000, opened 2025-04-06
[windows] WebView2 runtime not found


basilevs:
**Describe the bug**
When launching Eclipse IDE for Java Developers 2025-03 for the first time an error

> WebView2 runtime not found

happens.

```
!SESSION 2025-04-06 15:49:12.638 -----------------------------------------------
eclipse.buildId=4.35.0.20250306-0811
java.version=21.0.6
java.vendor=Eclipse Adoptium
BootLoader constants: OS=win32, ARCH=x86_64, WS=win32, NL=en_US
Framework arguments:  -product org.eclipse.epp.package.java.product
Command-line arguments:  -os win32 -ws win32 -arch x86_64 -product org.eclipse.epp.package.java.product -data C:\Users\Test\eclipse-workspace2

!ENTRY ch.qos.logback.classic 1 0 2025-04-06 15:49:15.775
!MESSAGE Activated before the state location was initialized. Retry after the state location is initialized.

!ENTRY ch.qos.logback.classic 1 0 2025-04-06 15:49:17.449
!MESSAGE Logback config file: C:\Users\Test\eclipse-workspace2\.metadata\.plugins\org.eclipse.m2e.logback\logback.2.7.0.20241001-1350.xml

!ENTRY org.eclipse.egit.ui 2 0 2025-04-06 15:49:24.622
!MESSAGE Warning: The environment variable HOME is not set. The following directory will be used to store the Git
user global configuration and to define the default location to store repositories: 'C:\Users\Test'. If this is
not correct please set the HOME environment variable and restart Eclipse. Otherwise Git for Windows and
EGit might behave differently since they see different configuration options.
This warning can be switched off on the Team > Git > Confirmations and Warnings preference page.

!ENTRY org.eclipse.e4.ui.workbench 4 0 2025-04-06 15:49:33.973
!MESSAGE Not implemented [WebView2 runtime not found]
!STACK 0
org.eclipse.swt.SWTError: Not implemented [WebView2 runtime not found]
	at org.eclipse.swt.SWT.error(SWT.java:4957)
	at org.eclipse.swt.browser.Edge.createEnvironment(Edge.java:591)
	at org.eclipse.swt.browser.Edge.createInstance(Edge.java:634)
	at org.eclipse.swt.browser.Edge.create(Edge.java:630)
	at org.eclipse.swt.browser.Browser.<init>(Browser.java:99)
[...]

PR #2038, opened 2025-04-17
[Edge] Add link to Edge FAQ with information on installing runtime #2000

HeikoKlare:
When the OS lacks a WebView2 runtime installation, an error is shown once a browser is initialized since Edge is used as the default browser on Windows. This change adds a link to the according FAQ page with information on how to install such a runtime to the error message.

Resolves https://github.com/eclipse-platform/eclipse.platform.swt/issues/2000

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
@@ -599,7 +599,7 @@ WebViewEnvironment createEnvironment() {
 					pBrowserDir, pDataDir, options.getAddress(), completion.getAddress()));
 	options.Release();
 	if (hr == OS.HRESULT_FROM_WIN32(OS.ERROR_FILE_NOT_FOUND)) {
-		SWT.error(SWT.ERROR_NOT_IMPLEMENTED, null, " [WebView2 runtime not found]");
+		SWT.error(SWT.ERROR_NOT_IMPLEMENTED, null, " [WebView2 runtime not found. For details, see https://github.com/eclipse-platform/eclipse.platform/tree/master/docs/FAQ/FAQ_How_do_I_use_Edge-IE_as_the_Browser's_underlying_renderer.md]");
 	}
 	if (hr != COM.S_OK) error(SWT.ERROR_NO_HANDLES, hr);
 	ICoreWebView2Environment environment = new ICoreWebView2Environment(ppv[0]);


sratz:
Is GitHub raw file access supposed to be the "official" location of the FAQ now?

Or should this be published to https://eclipse.dev/eclipse/swt/faq.html somehow?

vogella:
> Is GitHub raw file access supposed to be the "official" location of the FAQ now?

Yes. Most of the FAQs were part of the wiki and the wiki entries have been migrated to Markdown. I would suggest to move the relevant entries of the https://eclipse.dev/eclipse/swt/faq.html also to the FAQ md. 

> Or should this be published to https://eclipse.dev/eclipse/swt/faq.html somehow?

sratz:
> May be we could forward https://eclipse.dev/eclipse/swt/faq.html to the GitHub raw file to at least keep the existing link point to the most up-to-date FAQ?

Hmm, there are many more SWT-specific FAQ entries in that list than we have under

https://github.com/eclipse-platform/eclipse.platform/tree/master/docs/FAQ/

So we'd lose a lot of information.

I think we would need to migrate all SWT FAQ entries that are still relevant there to markdown.

vogella:
> I was not aware that there is still so much information to be migrated (I thought that most of it was already done).

IIRC the information on the wikis were migrated (as the wiki was going down), so far not the rest.

merks:
FYI, I'm currently trying to make sense of what's all there (documentation in the form of `*.md`)  in different repositories. I am also experimenting with client-side markdown rendering such that it's possible to embed the `*.md` content within the website.  E.g., here we have a page that uses the query parameter `file=` to read the markdown file (via api.github.com) and render it locally:

image

I'm also looking to generate a list of a folder's contents (so that we don't have to maintain an index):

image

As a result, I stumbled upon this 

https://github.com/eclipse-platform/eclipse.platform/blob/master/docs/FAQ.md

I  noticed that the very first link on this page is broken:

https://github.com/eclipse-platform/eclipse.platform/blob/master/docs/FAQ/The_Official_Eclipse_FAQs.md

@vogella Can docs/FAQ.md be deleted given we have docs/FAQ/The_Official_Eclipse_FAQs.md?

In any case, I hope to make it easier to find and navigate this valuable content much of which @vogella  helped migrate...

vogella:
> @vogella Can docs/FAQ.md be deleted given we have docs/FAQ/The_Official_Eclipse_FAQs.md?

It is not the same content in  https://github.com/eclipse-platform/eclipse.platform/blob/master/docs/FAQ.md 

In theory someone would have to go though it, check if the entries are still relevant and move them to the other FAQ.I did some clean-up in this document but ran (way) out of time.

merks:
I see.  So we have 3 different FAQ sources, none of which are known to be complete.  😱  Oh well, one step at a time.

vogella:
> * https://eclipse.dev/eclipse/swt/faq.html

First link here seems also outdated it still points to https://bugs.eclipse.org/bugs/enter_bug.cgi?product=Platform&component=SWT and the first entry explains in details what to put into the Bugzilla issue. 
Also scrolling throw the SWT FAQ shows tons of outdated entries, like: "What do I need to run the SWT Browser inside Eclipse on Solaris?" or "Why do I get strange drawing behaviour when running 32 bit Eclipse on a 64 bit GTK platform?".

Maybe a first step would be to delete the obvious obsolete entries in the SWT FAQ.  (I'm not planning to do this, but I assume that would be a relative quick thing to do for an active SWT developer).

PR #2039, opened 2025-04-17
Replace remaining usage of 'PLUGIN_PATH' system-property in tests

HannesWell:
and remove definition of therefore now unused `plugin-path` property.


HannesWell:
Single test failure is unrelated.

PR #2044, opened 2025-04-21
Add FreeBSD aarch64/x86_64 platforms' build.

chirontt:
- part of the overall issue eclipse-platform/eclipse.platform.releng.aggregator#2959
- add 2 SWT binary modules for FreeBSD: `org.eclipse.swt.gtk.freebsd.aarch64` and `org.eclipse.swt.gtk.freebsd.x86_64`
- changes made to the `build.sh` script and _make_linux.mak_ (renamed to `make_unix.mak`) to support compiling FreeBSD native binaries for SWT.

To support cross-compiling from Linux to FreeBSD, for the _same_ hardware architecture, a Docker container is made available in my fork with detailed instructions:
https://github.com/chirontt/eclipse.platform.releng.aggregator/tree/master/cje-production/dockerfiles/alpine/freebsd-cross 
which will be submitted to the official repo at later stage.

### DIFF
--- a/binaries/pom.xml
+++ b/binaries/pom.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!-- 
 ###############################################################################
-# Copyright (c) 2024, 2024 Hannes Wellmann and others.
+# Copyright (c) 2024, 2025 Hannes Wellmann and others.
 # All rights reserved. This program and the accompanying materials
 # are made available under the terms of the Eclipse Public License v1.0
 # which accompanies this distribution, and is available at
@@ -9,6 +9,7 @@
 #
 # Contributors:
 #     Hannes Wellmann - initial API and implementation
+#     Tue Ton - support for FreeBSD
 ###############################################################################
 -->
 <project xmlns="http://maven.apache.org/POM/4.0.0"
@@ -34,6 +35,8 @@
 	<modules>
 		<module>org.eclipse.swt.cocoa.macosx.x86_64</module>
 		<module>org.eclipse.swt.cocoa.macosx.aarch64</module>
+		<module>org.eclipse.swt.gtk.freebsd.aarch64</module>
+		<module>org.eclipse.swt.gtk.freebsd.x86_64</module>
 		<module>org.eclipse.swt.gtk.linux.aarch64</module>
 		<module>org.eclipse.swt.gtk.linux.ppc64le</module>
 		<module>org.eclipse.swt.gtk.linux.riscv64</module>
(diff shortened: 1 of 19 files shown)


bundles/org.eclipse.swt/META-INF/p2.inf
```diff
@@ -44,3 +44,13 @@ requires.9.namespace = org.eclipse.equinox.p2.iu
 requires.9.name = org.eclipse.swt.gtk.linux.riscv64
 requires.9.range = [$version$,$version$]
 requires.9.filter = (&(osgi.os=linux)(osgi.ws=gtk)(osgi.arch=riscv64)(!(org.eclipse.swt.buildtime=true)))
+
```
akurtakov, line 47: This part can not go in unless actual binaries are produced and pushed .

local-build/org.eclipse.swt.fragments.localbuild/META-INF/p2.inf
```diff
@@ -38,3 +38,13 @@ requires.8.namespace = org.eclipse.equinox.p2.iu
 requires.8.name = org.eclipse.swt.gtk.linux.riscv64
 requires.8.range = 0.0.0
 requires.8.filter = (&(osgi.os=linux)(osgi.ws=gtk)(osgi.arch=riscv64))
+
```
akurtakov, line 41: This part can not go in unless actual binaries are produced and pushed.

binaries/pom.xml
```diff
@@ -34,6 +35,8 @@
 	<modules>
 		<module>org.eclipse.swt.cocoa.macosx.x86_64</module>
 		<module>org.eclipse.swt.cocoa.macosx.aarch64</module>
+		<module>org.eclipse.swt.gtk.freebsd.aarch64</module>
```
akurtakov, line 38: As these will stay "shell" bundles without actual binaries in them we should not include them so such useless bundles don't get published in the p2 repo. I would be fine if you include them via Maven "freebsd" profile though.

bundles/org.eclipse.swt/META-INF/p2.inf
chirontt: Not only this specific file, but without the SWT binaries being officially compiled for FreeBSD, this whole PR is moot/pointless and shouldn't be merged at all.

And with the SWT being a critical component of Eclipse, failure to merge this PR is also a critical failure for the whole effort in eclipse-platform/eclipse.platform.releng.aggregator#2959

local-build/org.eclipse.swt.fragments.localbuild/META-INF/p2.inf
chirontt: Ditto, per my reply above

binaries/pom.xml
chirontt: Ditto, per my reply above

akurtakov:
Closing as there doesn't seem to be a viable path ahead for this one.

Issue #2003, opened 2025-04-06
Links in empty Package Explorer are cut off


iloveeclipse:
I'm using Windows 11 with 125% monitor scaling.
It looks like one of the recent changes broke the dimensions calculation for Link elements, because they are cut off now:

![Image](https://github.com/user-attachments/assets/7851a7a7-5304-464f-8de3-f5170cc75fe3)

Switching "Monitor specific UI scaling" **off** fixes the problem:

![Image](https://github.com/user-attachments/assets/75a3588b-c06e-4c81-a058-1f3fb6cf724a)

Note: I have a notebook with a single built-in monitor and no other monitors connected to it.

fedejeanne:
Caused by https://github.com/eclipse-platform/eclipse.platform.swt/issues/2003 (@akoch-yatta)

HeikoKlare:
Reopening as the solution attempt introduced a regression and had to be reverted: https://github.com/eclipse-platform/eclipse.platform.swt/pull/2095

We will propose an updated fix soon.

HeikoKlare:
That's strange. I cannot reproduce with the almost same configuration: latest I-Build, Windows 11 (22631), Temurin 21.0.6+7, 125% primary monitor zoom
![Image](https://github.com/user-attachments/assets/6fc0db95-9e7f-4280-aca1-8e15c591d0ba)

So you experience the behavior in your host Eclipse, not in a runtime instance?

HeikoKlare:
I can confirm that this also happens for me in the Git repositories view with the same configuration:
![Image](https://github.com/user-attachments/assets/401f7752-81f5-4d2e-9238-485133a8481e)

It does not occur for the package explorer though.

fedejeanne:
Fixed by https://github.com/eclipse-platform/eclipse.platform.ui/pull/2975

HeikoKlare:
Just for the sake of completeness: the additionally reported Git repositories view issue has been resolved in EGit: https://github.com/eclipse-egit/egit/commit/bd9874c94f98b0a7438f741cd3682c123140b2be

PR #2047, opened 2025-04-22
Preserved width/height on scaleUp/Down a Rectangle

akoch-yatta:
This commit extends the logic when a Rectangle is scaled up/down from points to pixels regarding width and height.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2003


HeikoKlare (approved):
The change is sound and properly solves https://github.com/eclipse-platform/eclipse.platform.swt/issues/2003

I couldn't find any unexpected side effects so far, so let's give this a try.

PR #2048, opened 2025-04-22
Refactor DPIUtil:autoScaleImageData for better design

amartya4256:
This PR contributes to refactoring the logic for scaling the ImageData using GC from DPIUtil to Image for each platform, where it belongs better and DPIUtil calls the platform specific code from the method DPIUtil:autoScaleImageData.

contributes to #62 and #127

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Image.java
@@ -1551,4 +1551,27 @@ public String toString () {
 	return "Image {" + surface + "}";
 }
 
+/**
+ * <b>IMPORTANT:</b> This method is not part of the public
+ * API for Image. It is marked public only so that it
+ * can be shared within the packages provided by SWT.
+ *
+ * Draws a scaled image using the GC by another image.
+ *
+ * @param gc the GC to draw on the resulting image
+ * @param original the image which is supposed to be scaled and drawn on the resulting image
+ * @param width the width of the original image
+ * @param height the height of the original image
+ * @param scaleFactor the factor with which the image is supposed to be scaled
+ *
+ * @noreference This method is not intended to be referenced by clients.
+ */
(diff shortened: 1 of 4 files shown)


bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Image.java
```diff
@@ -1551,4 +1551,28 @@ public String toString () {
 	return "Image {" + surface + "}";
 }
 
+/**
+ * <b>IMPORTANT:</b> This method is not part of the public
+ * API for Image. It is marked public only so that it
+ * can be shared within the packages provided by SWT. However, it is not
+ * available on all platforms.
```
akoch-yatta, line 1558: Overall it looks good to me and reverts the code duplication by providing the adapted implementation only for windows for now.
One sentence in the JavaDoc thats not correct is :"However, it is not available on all platforms." I think, we should remove that

HeikoKlare: Agree with this remark. Apart from that, the change also looks good to me 👍

amartya4256: Adapted

HeikoKlare (approved):

HeikoKlare:

HannesWell:
Sorry for joining this discussion (too) late, but I think adding more 'internal' methods to a public API class, especially such an important one like Image, even if they are marked as `@noreference`, really should be avoided as much as possible.

In this case I think it would be better to add it to a platform-specific util class.
My first idea would be renaming `NativeImageLoader` to something like `NativeImage` and add it there. But better ideas are welcome.

HeikoKlare:
> In this case I think it would be better to add it to a platform-specific util class.
> My first idea would be renaming `NativeImageLoader` to something like `NativeImage` and add it there. But better ideas are welcome.

Sounds good to me 👍 So it would basically become an internal utility class for everything related to OS-specific image?

PR #2051, opened 2025-04-23
Use auto-scale "nearest" by default in Windows #2910

fedejeanne:
Defaulting to "smooth" creates blurry images when first displaying the image in a monitor with a high zoom level (e.g. 200%).

This is a temporary workaround until https://github.com/eclipse-platform/eclipse.platform.swt/issues/2052 is properly fixed.

The auto-scale method can still be manually set to "smooth" by setting this VM argument: `-Dswt.autoScale.method=smooth`

## FIXME
This PR is just a workaround and I am only keeping it here in store in case there is no proper fix in time for M2. **I'll keep it drafted until noon but feel free to review now**.

## Screenshots
When starting the workbench in a monitor at 200% zoom level and with _monitor-specific UI scaling_ enabled. (original [screenshots.zip](https://github.com/user-attachments/files/19862188/screenshots.zip))

In green you can see which images look better using the **nearest** scale method (and in rot, the same images when using **smooth** scaling, for comparison).

Auto-scale method **nearest** (the new default when applying this PR)
image

Auto-scale method **smooth** (the previous default)
image


HeikoKlare:
This would basically render the SDK unusable on HiDPI monitors when using Windows 🙁 
So from my side, this is not an acceptable workaround.

Issue #2049, opened 2025-04-22
SWTException: Unsupported or unrecognized format [No SVG rasterizer found]


iloveeclipse:
We see following error reported in the log during our automated tests based on 4.36_I20250421-0320 SDK build.

The test simply opens `org.eclipse.debug.ui.ExpressionView`.

I've attached a small reproducer: 

[issue_2049.zip](https://github.com/user-attachments/files/19865020/issue_2049.zip)

This is regression compared with build 4.36_I20250410-1800 where the same test passes.

Test runs on Java 21.0.6 / RHEL 7.9.

```
!ENTRY org.eclipse.ui 4 0 2025-04-22 17:34:24.129
!MESSAGE Unhandled event loop exception
!STACK 0
org.eclipse.swt.SWTException: Unsupported or unrecognized format [No SVG rasterizer found]
	at org.eclipse.swt.SWT.error(SWT.java:4945)
	at org.eclipse.swt.internal.image.SVGFileFormat.loadFromByteStream(SVGFileFormat.java:48)
	at org.eclipse.swt.internal.image.FileFormat.loadFromStream(FileFormat.java:114)
	at org.eclipse.swt.internal.image.FileFormat.load(FileFormat.java:136)
	at org.eclipse.swt.internal.NativeImageLoader.load(NativeImageLoader.java:56)
	at org.eclipse.jface.resource.URLImageDescriptor.loadImageFromStream(URLImageDescriptor.java:186)
	at org.eclipse.jface.resource.URLImageDescriptor.loadImageData(URLImageDescriptor.java:171)
	at org.eclipse.jface.resource.URLImageDescriptor.getImageData(URLImageDescriptor.java:164)
	at org.eclipse.jface.resource.URLImageDescriptor.getImageData(URLImageDescriptor.java:142)
	at org.eclipse.jface.resource.URLImageDescriptor.getImageData(URLImageDescriptor.java:135)
	at org.eclipse.jface.resource.DeferredImageDescriptor.getImageData(DeferredImageDescriptor.java:76)
	at org.eclipse.jface.resource.ImageDescriptor.getImageData(ImageDescriptor.java:434)
	at org.eclipse.debug.internal.ui.viewers.model.TreeModelLabelProvider.getImage(TreeModelLabelProvider.java:152)
	at org.eclipse.debug.internal.ui.viewers.model.InternalTreeModelViewer.setElementData(InternalTreeModelViewer.java:1410)
[...]

Issue #1965, opened 2025-03-30
SVGRasterizer tests fail in I-Builds


HeikoKlare:
With latest I-Build, the tests for the SVGRasterizer (`org.eclipse.swt.tests.junit.Test_org_eclipse_swt_internal_SVGRasterizer`) added to SWT have been executed for the first time. While they succeed in pure Tycho builds, they fail during I-Builds, e.g.:
https://download.eclipse.org/eclipse/downloads/drops4/I20250330-0640/testresults/html/org.eclipse.swt.tests_ep436I-unit-linux-x86_64-java21_linux.gtk.x86_64_21.html

The error indicates that the required SVG fragment is not present on the classpath:
```
org.eclipse.swt.SWTException: Unsupported or unrecognized format [No SVG rasterizer found]
at org.eclipse.swt.SWT.error(SWT.java:4945)
at org.eclipse.swt.internal.image.SVGFileFormat.loadFromByteStream(SVGFileFormat.java:48)
at org.eclipse.swt.internal.image.FileFormat.loadFromStream(FileFormat.java:114)
at org.eclipse.swt.internal.image.FileFormat.load(FileFormat.java:136)
at org.eclipse.swt.internal.NativeImageLoader.load(NativeImageLoader.java:56)
at org.eclipse.swt.graphics.ImageLoader.load(ImageLoader.java:161)
at org.eclipse.swt.graphics.ImageLoader.load(ImageLoader.java:192)
at org.eclipse.swt.graphics.ImageLoader.load(ImageLoader.java:185)
at org.eclipse.swt.graphics.ImageDataLoader.load(ImageDataLoader.java:35)
at org.eclipse.swt.graphics.ImageData.<init>(ImageData.java:378)
at org.eclipse.swt.tests.junit.Test_org_eclipse_swt_internal_SVGRasterizer.lambda$3(Test_org_eclipse_swt_internal_SVGRasterizer.java:57)
at org.eclipse.swt.internal.DPIUtil.lambda$2(DPIUtil.java:546)
at org.eclipse.swt.internal.DPIUtil.getElementAtZoom(DPIUtil.java:575)
at org.eclipse.swt.internal.DPIUtil.validateAndGetImageDataAtZoom(DPIUtil.java:546)
at org.eclipse.swt.graphics.Image.initFromImageDataProvider(Image.java:788)
at org.eclipse.swt.graphics.Image.<init>(Image.java:653)
[...]

PR #2055, opened 2025-04-23
Use proper classloader for finding SVGRasterizer #1965 #2049

HeikoKlare:
When initializing the SVGFileFormat class for loading SVGs, the current context classloader is used by the ServiceLoader to find an SVGRasterizer implementation. This classloader may be incorrect in some cases, i.e., it may not be the plain system classloader or an according OSGi classloader but, e.g., some specific classloader for test execution.

This change makes the ServiceLoader use the same classloader for finding an SVGRasterizer implementation than the classloader of the SVGRasterizer class itself.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/1965

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2049

## Remarks

I am not completely sure if this will always yield a "correct" classloader or whether the classloader of the `SVGRasterizer` may also be "wrong".

An alternative might be to force classloading of the `SVGFileFormat` class to happen easly in application lifecycle instead of at first usage where the current context classloader may vary.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/image/SVGFileFormat.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/image/SVGFileFormat.java
@@ -31,7 +31,8 @@ import org.eclipse.swt.internal.DPIUtil.*;
 public class SVGFileFormat extends FileFormat {
 
 	/** The instance of the registered {@link SVGRasterizer}. */
-	private static final SVGRasterizer RASTERIZER = ServiceLoader.load(SVGRasterizer.class).findFirst().orElse(null);
+	private static final SVGRasterizer RASTERIZER = ServiceLoader
+			.load(SVGRasterizer.class, SVGFileFormat.class.getClassLoader()).findFirst().orElse(null);
 
 	@Override
 	boolean isFileFormat(LEDataInputStream stream) throws IOException {


iloveeclipse:
Tomorrow.

iloveeclipse:
But feel free to merge, looks promising.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/image/SVGFileFormat.java
```diff
@@ -31,7 +31,11 @@
 public class SVGFileFormat extends FileFormat {
 
 	/** The instance of the registered {@link SVGRasterizer}. */
-	private static final SVGRasterizer RASTERIZER = ServiceLoader.load(SVGRasterizer.class).findFirst().orElse(null);
+	// The service loader uses the classloader of the SVGRasterizer class to avoid
+	// that a potentially non-fitting context classloader is used for searching for
+	// a rasterizer implementation (such as during test execution).
+	private static final SVGRasterizer RASTERIZER = ServiceLoader
+			.load(SVGRasterizer.class, SVGRasterizer.class.getClassLoader()).findFirst().orElse(null);
```
laeubi, line 38:
```suggestion
			.load(SVGRasterizer.class, SVGFileFormat.class.getClassLoader()).findFirst().orElse(null);
```

Even though `SVGRasterizer` and `SVGFileFormat` are in the same bundle and have the same classloader it seems semantically more correct to use `SVGFileFormat` as this is the class that loads the `SVGRasterizer` (what usually could be an API class from somewhere else).

Using an explicit classloader for serviceloader is best-practice anyways so I'm not sure we need an explicit comment here, but it of course does not harm.

iloveeclipse: Independently which version of the patch is used, I can confirm the original test and the simplified one are working now on RHEL 9.2. Thanks.

iloveeclipse (approved):

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/image/SVGFileFormat.java
HeikoKlare: > Even though `SVGRasterizer` and `SVGFileFormat` are in the same bundle and have the same classloader it seems semantically more correct to use `SVGFileFormat` as this is the class that loads the `SVGRasterizer` (what usually could be an API class from somewhere else).

I was somehow concerned that it's a bad idea to use the class loader of the class that is currently being loaded (as we are in the static initialization. But since it works fine and I agree that it is semantically more correct, I adapted the PR accordingly.

> Using an explicit classloader for serviceloader is best-practice anyways so I'm not sure we need an explicit comment here, but it of course does not harm.

Thank you for the pointer. I wasn't aware of this. So I just removed the comment as it seems unnecessary to document best practice explicitly again.

> Independently which version of the patch is used, I can confirm the original test and the simplified one are working now on RHEL 9.2. Thanks.

Thank you for testing! Good to know that it works for you. I have also successfuly retested the modified change on Windows.

merks: It's good to see the team working so effectively together.  💞

HannesWell (approved):
Looks very good. Great catch!
When looking at the mentioned issues I didn't thought about this at all, but it makes totally sense.

HeikoKlare:

PR #2056, opened 2025-04-24
Support disabling the GTK SessionManagerDBus via environment variables

HannesWell:
For example this simplifies disabling the `SessionManagerDBus` in CI environments where no bus is available.

At our CI-server, which formally doesn't have an X-Server installed and uses xvnc I see the following warnings:
```
SWT SessionManagerDBus: Failed to connect to org.gnome.SessionManager: Failed to execute child process “dbus-launch” (No such file or directory)
SWT SessionManagerDBus: Failed to connect to org.xfce.SessionManager: Failed to execute child process “dbus-launch” (No such file or directory)
```

Setting the system-property `org.eclipse.swt.internal.SessionManagerDBus.disable` disables the `SessionManagerDBus` right from the start and prevents these warnings. But I have to set that system-property in the Maven/Tycho pom.xml as well as in all headless eclipse executions we run in our build.
It would be more convenient, if I can just set an environment property for the entire pipeline that would be used as proposed here.


akurtakov:
Failing Windows test is unrelated. Thanks for this one!

Issue #2066, opened 2025-04-25
[Win32] Wrong font in status line


HeikoKlare:
**Describe the bug**
Some fonts, at least the ones used in the status line, are wrong since:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/1988

**To Reproduce**
Start an Eclipse SDK and open a Java editor. The text in the status line uses wrong font.

**Expected behavior**
Fonts should look as before the mentioned change.

**Screenshots**
Before:
![Image](https://github.com/user-attachments/assets/91f8103b-1800-4f62-89f5-a6633cccb47f)

After: 
![Image](https://github.com/user-attachments/assets/d36e4d2c-d1ea-4ce5-a4be-392eb7a7a529)

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 

**Version since**
Current no I-Builds available, but can be tracked back to change:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/1988

PR #2067, opened 2025-04-25
[win32] Fix GC font for composite

akoch-yatta:
This commit reverts moving the initialization of GCData.font in Composite#WM_Paint in commit 033d733.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2066


HeikoKlare (commented):
Thank you for the super fast fix!
The change looks good and is safe to do as it just reverts a recent change before which everything worked as expected. I also tested the change and is properly solves the mentioned issue.

Just to be sure: the commit in which the issue was introduced is https://github.com/eclipse-platform/eclipse.platform.swt/commit/504c82b1b210d272060374e04af60096f636b9fb and not https://github.com/eclipse-platform/eclipse.platform.swt/commit/033d73320e6246f025d979c68c19697d9b64334a, is it?

akoch-yatta:
Yes, probably copied the wrong hash somehow

HeikoKlare (approved):

HeikoKlare:

Failing version increment check is because of infrastructure issues and no version bump is required here as SWT has already been bumped.

PR #2069, opened 2025-04-27
[win32] Move of unified code from Wrapper to Image

akoch-yatta:
This PR is a simple refactoring and just moves the code that is equal to all AbstractImageProviderWrapper implementations back to Image.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
@@ -99,6 +99,11 @@ public final class Image extends Resource implements Drawable {
 	private boolean isInitialized;
 
 	/**
+	 * this field is used to mark destroyed images
+	 */
+	private boolean isDestroyed;
+
+	/**
 	 * specifies the transparent pixel
 	 */
 	int transparentPixel = -1, transparentColor = -1;
@@ -999,7 +1004,7 @@ long [] createGdipImage(Integer zoom) {
 void destroy () {
 	device.deregisterResourceWithZoomSupport(this);
 	if (memGC != null) memGC.dispose();
-	this.imageProvider.destroy();
+	this.isDestroyed = true;
 	destroyHandle();
 	memGC = null;
 }
@@ -1225,7 +1230,10 @@ public ImageData getImageData() {
  */
 public ImageData getImageData (int zoom) {
 	if (isDisposed()) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
-	return imageProvider.getImageData(zoom);
+	if (zoomLevelToImageHandle.containsKey(zoom)) {
+		return zoomLevelToImageHandle.get(zoom).getImageData();
+	}
+	return this.imageProvider.newImageData(zoom);
 }
 
 
@@ -1762,7 +1770,7 @@ public void internal_dispose_GC (long hDC, GCData data) {
  */
 @Override
 public boolean isDisposed() {
-	return this.imageProvider.isDisposed();
+	return !isInitialized || isDestroyed;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -562,7 +567,7 @@ public Image(Device device, ImageFileNameProvider imageFileNameProvider) {
 	super(device);
 	this.imageProvider = new ImageFileNameProviderWrapper(imageFileNameProvider);
 	initialNativeZoom = DPIUtil.getNativeDeviceZoom();
-	if (imageProvider.getImageData(100) == null) {
+	if (imageProvider.newImageData(100) == null) {
```
HeikoKlare, line 570: I was wondering if we could avoid/improve this, as it's quite some potentially unnecessary effort to load image data at 100% if they are never used. Then I saw that the documentation says:
> ERROR_INVALID_ARGUMENT - if the fileName provided by ImageFileNameProvider is null at 100% zoom

So the current implementation is even wrong: it should not be tested if image data at 100% can be loaded but if the passed `imageFileNameProvider` returns a non-null value at 100%.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -521,7 +526,7 @@ public Image (Device device, String filename) {
 		}
 		return null;
 	});
-	if (imageProvider.getImageData(100) == null) {
+	if (imageProvider.newImageData(100) == null) {
```
HeikoKlare, line 529: I just saw that there is no documentation about this kind of check and it's also not present in the other OS implementations. So probably we should just remove it?

HeikoKlare (commented):
The change looks generally good to me. I Just found two issues not directly related to the refactoring. I propose to maybe fix them in a separate PR (to make clear that that's a behavioral fix compared to this refactoring) to be merged before this one?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: The problem are the other exceptions mentioned:

```*    <li>ERROR_IO - if an IO error occurs while reading from the file</li>
 *    <li>ERROR_INVALID_IMAGE - if the image file contains invalid data </li>
 *    <li>ERROR_UNSUPPORTED_DEPTH - if the image file describes an image with an unsupported depth</li>
 *    <li>ERROR_UNSUPPORTED_FORMAT - if the image file contains an unrecognized format</li>
 ```
 
 Those will only be detected if the ImageData is loaded once. If we remove that call here it would fail later, which would change the behavior

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: As above implicitly checking other exception mentioned in the javadoc, e.g.

` *    <li>ERROR_IO - if an IO error occurs while reading from the file</li>`

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
HeikoKlare: I see. Since we are now caching the already loaded image data, it might make sense to then not use zoom 100 here but instead retrieve some data that we will likely use?

Still, in my opinion we throw the mentioned exception in the wrong case: it should not be thrown when no image data can be generated but when the fileName provided by the file name provider at 100% is null. But maybe that just happens in the same situations.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
HeikoKlare: But like above: the exception we throw does not seem to be documented. So I guess it would be more correct to just call `imageProvider.newImageData()` without throwing a exception if the result is null?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: Ah, now I fully get, what you mean. I just took a short look into the history. Before we only used `getImageMetadata `that seemed to trigger the exceptions, we could probably do a newImageHandle + immediatly destroy if afterwards. I will try to have a look tomorrow.

HeikoKlare (approved):

HeikoKlare:
Everything related to the above discussion has been extracted into https://github.com/eclipse-platform/eclipse.platform.swt/pull/2070, so this is fine now.

PR #2070, opened 2025-04-28
[win32] Move some Image initialization checks

akoch-yatta:
This PR moves some initialization checks related to ImageFileNameProvider and ImageDataProvider into the matching Wrapper implementation.


HeikoKlare (approved):
Looks good. I have slightly adapted the commit message to reflect that with recent changes there is not only code moved but also the case in which an exception is thrown has been corrected.

Issue #2057, opened 2025-04-24
Custom cursor size is doubled with Eclipse 4.35


wolfgang-ch:
This is the cursor with Eclipse 4.34 which has the correct size

<img width="287" alt="Image" src="https://github.com/user-attachments/assets/ae080f0c-4284-484e-95a5-71d437c46b6a" />

This is the cursor with Eclipse 4.35 which size is about doubled

<img width="287" alt="Image" src="https://github.com/user-attachments/assets/4f7338d9-20d4-4664-a67c-c91d8d9c4c6d" />

This happens since Eclipse 4.35 on a 4k display by running the app with `-Dswt.autoScale=200`

The custom cursor is create with this code

<img width="561" alt="Image" src="https://github.com/user-attachments/assets/2865d1c0-9ac6-427a-8d31-624d5672cd9f" />

**Environment:**
Select the platform(s) on which the behavior is seen:
- - [?] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 

**JRE/JDK version**
17 or 21
 
**Version since**
 4.35

**Workaround (or) Additional context**
Any known workarounds for the problem? NO

HeikoKlare:
I guess it's caused by this change:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/1603/

It seems like cursors were changed to take the zoom of the context in which they are registered into account. Since you pass a zoom (auto-scale) of 200 to the application, the cursor is now probably also scaled to 200%.

@ShahzaibIbrahim @fedejeanne Can you please test with the snippet used for #1603 for the auto-scale settings given in the issue description again and see whether that PR introduced an issue?

fedejeanne:
> [@ShahzaibIbrahim](https://github.com/ShahzaibIbrahim) [@fedejeanne](https://github.com/fedejeanne) Can you please test with the snippet used for [#1603](https://github.com/eclipse-platform/eclipse.platform.swt/pull/1603) for the auto-scale settings given in the issue description again and see whether that PR introduced an issue?

Spot on, thank you!
I created a PR to fix it: #2071

wolfgang-ch:
With Eclipse 4.36 and Windows 10, I can still see, that the cursor is scaled.

This is the test cursor image which needs to be loaded in the snippet

![Image](https://github.com/user-attachments/assets/dee2e678-64b8-407e-86b3-1f3d6c6d512e)

[...]

PR #2071, opened 2025-04-28
Use the native zoom when setting the cursor in Control #2057

fedejeanne:
Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2057

# How to test
- Use 2 monitors with different zoom levels (none of them should match the zoom level passed to `-Dswt.autoScale` in the next step)
- Run `Snippet119` with the following JVM parameters: 
  - `-Dswt.autoScale=200` (or use another zoom level that doesn't match any of the zooms in your monitors)
  - `-Dswt.autoScale.updateOnRuntime=true`

Expected: the size of the cursor still matches the size of the current monitor and **not** the one you provided for `-Dswt.autoScale`
 
- Move the window of the snippet to the monitor 

Expected: the size of the cursor still matches the size of the current monitor and **not** the one you provided for `-Dswt.autoScale`


HeikoKlare:
Note that #2057 is _without_ `-Dswt.autoScale.updateOnRuntime=true`. Does this work for that as well? I would expect `getNativeZoom()` to return the value of `swt.autoScale` in that case.

fedejeanne:
> Note that #2057 is _without_ `-Dswt.autoScale.updateOnRuntime=true`. Does this work for that as well? I would expect `getNativeZoom()` to return the value of `swt.autoScale` in that case.

The current proposal (using the native zoom) means that the size of the cursor will **not** be affected by the parameter `-Dswt.autoscale` but will react to the zoom of the current monitor instead. Additionally, if one does not use the monitor-specific UI scaling (_e.g._ if `-Dswt.autoScale.updateOnRuntime=false`) then the size of the cursor will be the same in all monitors and it will be given by the zoom of the primary monitor.

fedejeanne:
I see the following error in the check:

```bash
Error:  Cannot resolve dependencies of project org.eclipse.platform:org.eclipse.swt.svg:eclipse-plugin:3.130.0-SNAPSHOT
Error:   with context {osgi.os=linux, org.eclipse.update.install.features=true, org.eclipse.swt.buildtime=true, osgi.arch=x86_64, org.eclipse.update.install.sources=true, osgi.ws=gtk, org.eclipse.jdt.buildtime=true}
Error:    Software being installed: org.eclipse.swt.svg 3.130.0.qualifier
Error:    Missing requirement: org.eclipse.swt.svg 3.130.0.qualifier requires 'org.eclipse.equinox.p2.iu; org.eclipse.swt.win32.win32.x86_64 0.0.0' but it could not be found: See log for details
```

And it seems unrelated to these changes.

@HeikoKlare all good from your side ✔️ ?

HeikoKlare:
What about all the other places where `getZoom()` has been introduced to the handle retrieval in `OS.SetCursor`? I only see one such place fixed, so I wonder why the others do not need to be changed.
See the according commit: https://github.com/eclipse-platform/eclipse.platform.swt/pull/1603/files

The version increment check is failing, but I guess version increment is not necessary here?

fedejeanne:
> What about all the other places where `getZoom()` has been introduced to the handle retrieval in `OS.SetCursor`? I only see one such place fixed, so I wonder why the others do not need to be changed.
> See the according commit: https://github.com/eclipse-platform/eclipse.platform.swt/pull/1603/files

I see the places and it could also make sense to apply the same change in all of them, but I'm having trouble testing those changes since I can't find any snippet or execution path in the workbench that runs through them. I was only able to run through 1 other place, namely the one in `org.eclipse.swt.widgets.Tree.setCursor()`, but I haven't seen any odd behavior in there, the cursor (which is a "busy cursor") has the correct size. Do you (@HeikoKlare / @ShahzaibIbrahim ) know how to reach the other places affected by #1603 where the cursor is set (the ones doing `Cursor.win32_getHandle(someCursor, getZoom()`) ?

> The version increment check is failing, but I guess version increment is not necessary here?

Correct.

fedejeanne:
> > The version increment check is failing, but I guess version increment is not necessary here?
> 
> Correct.

FTR [this](https://github.com/eclipse-platform/eclipse.platform.swt/pull/2071#issuecomment-2835091437) is the error ☝️ 

> I see the following error in the check:
> 
> ```shell
> Error:  Cannot resolve dependencies of project org.eclipse.platform:org.eclipse.swt.svg:eclipse-plugin:3.130.0-SNAPSHOT
> Error:   with context {osgi.os=linux, org.eclipse.update.install.features=true, org.eclipse.swt.buildtime=true, osgi.arch=x86_64, org.eclipse.update.install.sources=true, osgi.ws=gtk, org.eclipse.jdt.buildtime=true}
> Error:    Software being installed: org.eclipse.swt.svg 3.130.0.qualifier
> Error:    Missing requirement: org.eclipse.swt.svg 3.130.0.qualifier requires 'org.eclipse.equinox.p2.iu; org.eclipse.swt.win32.win32.x86_64 0.0.0' but it could not be found: See log for details
> ```

fedejeanne:
I ran all the UI tests in Platform and also all the tests in SWT and the other places are not covered, so I assume they are very rare use-cases. In any case, I adapted all invocations of `org.eclipse.swt.graphics.Cursor.win32_getHandle(Cursor)` in b750b5108741a32cc7c777ea08683cc89ec3b426 so they don't pass the zoom as parameter and use the zoom of the current monitor instead since the mouse should always match the zoom of the monitor and nothing else.

PR #2073, opened 2025-04-28
Extend ImageGcDrawer to support transparency

akoch-yatta:
This PR extends the usage of ImageGcDrawar to support SWT.Transparency as style that will initialize each GC with a proper transparent bitmap.

**Important:** 
Adaption for GTK and Cocoa will be added when we agree on the approach

I can extract to DPIUtil change into a separate PR

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
@@ -903,8 +903,16 @@ public Image(Device device, ImageGcDrawer imageGcDrawer, int width, int height) 
 }
 
 private ImageData drawWithImageGcDrawer(ImageGcDrawer imageGcDrawer, int width, int height, int zoom) {
-	Image image = new Image(device, width, height);
-	GC gc = new GC(image);
+	int gcStyle = imageGcDrawer.getGcStyle();
+	Image image;
+	if ((gcStyle & SWT.TRANSPARENT) != 0) {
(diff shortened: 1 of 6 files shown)


akoch-yatta:
@fedejeanne Can you test this one on MacOS? With Snippet367 you should see the transparent drawing working or not.

fedejeanne:
> @fedejeanne Can you test this one on MacOS? With Snippet367 you should see the transparent drawing working or not.

👍 I'll carve out some time later this week to test it on my Mac at home.

fedejeanne (approved):
I reviewed all implementations and also tested in Mac and Linux at home. I made some small modifications (reduced visibility of new methods) and also undid the change in line 701 of the `Image` class in **linux** in 64388ded9503bcd139e75e5820963c7189900e5c . The reason is that this change does not seem to belong in this PR and it may introduce undesired behavior. I also double-checked what we discussed in our call today and it is true that `DPIUtil.getDeviceZoom()` not always returns `100` on Linux, which is why I decided not to take any chances and I simply undid that change. If you still think that the change is necessary then please propose a follow-up PR and let us discuss the reason there.

fedejeanne:
> ... also undid the change in line 701 of the `Image` class in **linux** in [64388de](https://github.com/eclipse-platform/eclipse.platform.swt/commit/64388ded9503bcd139e75e5820963c7189900e5c) . 

@akoch-yatta doing that let 2 tests fail in Linux so I had to revert back to `currentDeviceZoom = 100` in line 701. Let us discuss the _why_ more in detail next week

fedejeanne:
The failing check about the service versions is due to infrastructure problems

fedejeanne (approved):

PR #2076, opened 2025-04-29
Artifacts due to rounding errors in StyledText

ShahzaibIbrahim:
While selecting the text in code editor the styled text have some artifacts/blank lines due to rounding error while translating points to pixel. Now we are passing the x and y in drawInPixels method in points instead of pixels and expecting it to be scaled up in the method.

### How To Test

- Run the `RuntimeWorkspace` with the following **VM Arguments** with primary monitor zoom of **175%**
```
-Dswt.autoScale=quarter
```
- Write some code/text in the text editor
- Select the code block with mouse
- See if the there are any artifact in between lines

### Screenshot
**Before:**
image

**After:**
image


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/TextLayout.java
```diff
@@ -729,7 +729,7 @@ public void draw (GC gc, int x, int y) {
  */
 public void draw (GC gc, int x, int y, int selectionStart, int selectionEnd, Color selectionForeground, Color selectionBackground) {
 	checkLayout();
-	drawInPixels(gc, DPIUtil.scaleUp(getDevice(), x, getZoom(gc)), DPIUtil.scaleUp(getDevice(), y, getZoom(gc)), selectionStart, selectionEnd, selectionForeground, selectionBackground);
+	drawInPixels(gc, x, y, selectionStart, selectionEnd, selectionForeground, selectionBackground);
 }
 
 void drawInPixels (GC gc, int x, int y, int selectionStart, int selectionEnd, Color selectionForeground, Color selectionBackground) {
```
akoch-yatta, line 735: x and y are here now as well xInPoints and yInPoints. Please rename them

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/TextLayout.java
```diff
@@ -787,15 +787,15 @@ void drawInPixels (GC gc, int x, int y) {
 	drawInPixels(gc, x, y, -1, -1, null, null);
```
akoch-yatta, line 787: x and y are here now as well xInPoints and yInPoints. Please rename them

akoch-yatta (changes requested):
Looks good to me. Didn't find any issues when testing this. As drawInPixels is a package protected method I think its fine to change the params, there are no usages besides the adapted ones. I don't see any other chance to fix this behavior besides scaling everything together

I only have the two remarks to rename the params in the other drawInPixelsMethods

akoch-yatta (approved):

fedejeanne (changes requested):
I'm testing this and I still see the lines:

image

HeikoKlare (commented):
Works as expected for me on 175% primary monitor (and also on 125% secondary):
image

fedejeanne:
Thank you all for testing it!

Since the problem seems to be on my machine, I accepted this PR because it improves the situation for most users. @ShahzaibIbrahim I would like a follow-up PR to see if we can fix the issue on my machine too, I'll ping you later so you can help me develop it. I assume that the deciding factor here is my Windows **10** since I know that you, Andreas and Heiko all use Windows **11** and I know that Heiko has the exact same monitor that I have (and he also tested with the same configuration that I tested with).

PR #2079, opened 2025-05-01
[Build] Activate javadoc generation and 'bree-libs' profile

HannesWell:
This will probably reveal many javadoc mistakes that have to be fixed first.
See also
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2999#issuecomment-2844874776


akurtakov:
Change looks good. Failing Windows test is unrelated. It's good to finally have javadoc checks here by default too.

PR #2080, opened 2025-05-02
Force qualifier updates

merks:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3011

### DIFF
--- a/binaries/org.eclipse.swt.cocoa.macosx.aarch64/forceQualifierUpdate.txt
+++ b/binaries/org.eclipse.swt.cocoa.macosx.aarch64/forceQualifierUpdate.txt
@@ -1,2 +1,1 @@
-https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2595
 https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3011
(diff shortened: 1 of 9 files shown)


merks:
@iloveeclipse  @akurtakov 

Is it okay to force all fragments and the host?  It seems to me these all always have the same qualifier in the repository:

image

merks:
@iloveeclipse  @akurtakov @MohananRahul 

The build failures are unrelated. Note that once I complete the Orbit milestone build and contribute it to the Platform's TP, I will merge this PR if there is no feedback in the meantime.

MohananRahul (approved):

akurtakov:
I was AFK . Forcing them all wouldn't hurt.

akurtakov:
Merging as the test failure is unstable windows test.

HannesWell:
> Is it okay to force all fragments and the host? It seems to me these all always have the same qualifier in the repository:

 Yes the native fragments always have the same qualifier as their host and it's sufficient to only touch the latter, due to:
https://github.com/eclipse-platform/eclipse.platform.swt/blob/5c5eae1dd2bacdc4c3a071dd40ad88014ae23674/binaries/pom.xml#L78

which was configured via
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/1096

merks:
I see.  So for future reference, no SWT native fragment ever needs to be qualifier forced, only the host might need to be qualifier forced.  Thanks.

PR #2087, opened 2025-05-02
Create Font handles on demand

ShahzaibIbrahim:
Creating a new static method to retrieve font handle. Handles will no longer be created during initialization. Also destroy condition is changed since handle being 0 doesn't mean that it was destroyed anymore.

### How to test

- Run the following snippet with following VM arguments:
```
-Dswt.autoScale=quarter
```
```java
import org.eclipse.swt.*;

public class SWTFontExample {
        shell.setText("SWT Font Example");
        shell.setSize(400, 200);

        Label label = new Label(shell, SWT.CENTER);
        label.setText("Hello, SWT with Custom Font!");

        FontData[] fontData = label.getFont().getFontData();
        for (FontData fd : fontData) {
            fd.setHeight(20); // Set font size to 20
            fd.setStyle(SWT.BOLD); // Make it bold
        }

        Font newFont = new Font(display, fontData);
        label.setFont(newFont);

        shell.open();
        while (!shell.isDisposed()) {
                display.sleep();
            }
        }
        newFont.dispose();
        display.dispose();
    }
}
```

- The correct formatting should be applied on the text same as before this change was made.
- This change should not produce any regression and the handle will only be created on demand not during initialization of Font object.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Font.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Font.java
@@ -56,6 +56,16 @@ public final class Font extends Resource {
 	int zoom;
 
 	/**
+	 * this field is used to mark destroyed fonts
+	 */
+	private boolean isDestroyed;
+
+	/**
+	 * this field is used to store fontData provided during initialization
+	 */
+	private final FontData fontData;
+
+	/**
 	 * Font height in points. As the conversion to pixel height involves rounding the fontHeight must
 	 * be cached.
 	 */
@@ -63,6 +73,7 @@ public final class Font extends Resource {
 
 private Font(Device device, long handle, int zoom) {
 	super(device);
+	this.fontData = null;
 	this.handle = handle;
 	this.zoom = zoom;
 	this.fontHeight = device.computePoints(fetchLogFontData(), handle, zoom);
@@ -90,16 +101,18 @@ private Font(Device device, long handle, int zoom) {
  */
 public Font(Device device, FontData fd) {
 	super(device);
+	if (fd == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
 	this.zoom = DPIUtil.getNativeDeviceZoom();
-	init(fd);
+	this.fontData = new FontData(fd.toString());
 	this.fontHeight = fd.height;
 	init();
 }
 
 private Font(Device device, FontData fd, int zoom) {
 	super(device);
+	if (fd == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
 	this.zoom = zoom;
-	init(fd);
+	this.fontData = new FontData(fd.toString());
 	this.fontHeight = fd.height;
 	init();
 }
@@ -138,7 +151,7 @@ public Font(Device device, FontData[] fds) {
 	}
 	this.zoom = DPIUtil.getNativeDeviceZoom();
 	FontData fd = fds[0];
-	init(fds[0]);
+	this.fontData = new FontData(fd.toString());
 	this.fontHeight = fd.height;
 	init();
 }
@@ -171,8 +184,7 @@ public Font(Device device, String name, int height, int style) {
 	super(device);
 	if (name == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
 	this.zoom = DPIUtil.getNativeDeviceZoom();
-	FontData fd = new FontData (name, height, style);
-	init(fd);
+	this.fontData = new FontData (name, height, style);
 	this.fontHeight = height;
 	init();
 }
@@ -181,8 +193,7 @@ public Font(Device device, String name, int height, int style) {
 	super(device);
 	if (name == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
 	this.zoom = DPIUtil.getNativeDeviceZoom();
-	FontData fd = new FontData (name, height, style);
-	init(fd);
+	this.fontData =  new FontData (name, height, style);
 	this.fontHeight = height;
 	init();
 }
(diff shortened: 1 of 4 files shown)


laeubi:
@ShahzaibIbrahim I once played around with making Fonts GC disposable here:

- https://github.com/eclipse-platform/eclipse.platform.swt/pull/1054

do you think that would be applicable here as well? So wrapping the (long) handle inside an `FontHandle` object would then also make the usage of different booleans not required, because you can have a

```
FontHandle handle;

if (handle == null) {
  //never initialized
}

if (handle != null && handle.pointer == 0) {
   //was initialized but disposed
}
```

in the code.

ShahzaibIbrahim:
> do you think that would be applicable here as well? So wrapping the (long) handle inside an `FontHandle` object would then also make the usage of different booleans not required, because you can have a
> 
> ```
> FontHandle handle;
> 
> if (handle == null) {
>   //never initialized
> }
> 
> if (handle != null && handle.pointer == 0) {
>    //was initialized but disposed
> }
> ```
> 
> in the code.

@laeubi I personally like the idea. I could implement it, but just to be consistent with other resources in win32. I would like it to be handled as I have done with isDestoyed field. 

Maybe we could do it for all the resources in a separate issue/ticket.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Font.java
```diff
@@ -294,6 +306,28 @@ public String toString () {
...
+ *
+ * @param font the font to get the handle of
+ *
+ * @return handle of the font
+ *
+ * @noreference This method is not intended to be referenced by clients.
+ */
+public static long win32_getHandle (Font font) {
+	if(font.handle == 0) {
+		font.init(font.fontData);
```
fedejeanne, line 326: This breaks some tests because `font.fontData` is `null`

fedejeanne (changes requested):
I marked the new field `fontData` as `final` (and explicitly initialized it to `null` in 2 constructors). The code looks OK but some tests fail because when `fontData == null`, the newly added static method does an illegal call.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Font.java
HeikoKlare: This indicates that there is some implicit contract and protocol assigned to handle and fontData inside Font. I guess it would be more correct to have a `Font` hold a `Provider<Long> handleProvider`, for which to realizations exist:
1. A "static" handle provider just capturing the handle the Font was initialized with
2. An "on-demand" handle provider, which generates the handle out of an initial FontData on demand.

Then in the first case you do not have the issue that some FontData field is null and you implicitly have to know what this means and how to deal with it.

fedejeanne: Thank you for the pointer! We were looking into it and came up with another approach in e62c8ab5af780ad24b17f089435338fd8d1d7940. I just pushed, let's see if everything else is in order.

fedejeanne (approved):

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Font.java
fedejeanne: Everything looks fine. I removed 2 small mistakes in 02b640d1172dbd754605aba37787d4a31e4ce255 and this PR is ready to go on my account.
@HeikoKlare can you re-review and approve/merge ?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Font.java
```diff
@@ -294,6 +308,28 @@ public String toString () {
...
 }
 
+/**
+ * <b>IMPORTANT:</b> This method is not part of the public
+ * API for Font. It is marked public only so that it
+ * can be shared within the packages provided by SWT. It is not
+ * available on all platforms, and should never be called from
+ * application code.
+ *
+ * Creates a new handle for the requested font or return the existing one
```
HeikoKlare, line 318:
```suggestion
 * Creates a new handle for the requested font or returns the existing one
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Font.java
```diff
@@ -100,16 +111,18 @@ private Font(Device device, long handle, int zoom) {
  */
 public Font(Device device, FontData fd) {
 	super(device);
+	if (fd == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
 	this.zoom = DPIUtil.getNativeDeviceZoom();
-	init(fd);
+	this.fontData = new FontData(fd.toString());
```
HeikoKlare, line 116: Seems like with this change we now really need a proper copy constructor / operation for `FontData` (at least as a follow-up change).

HeikoKlare (changes requested):
This PR is currently broken. The first commit is supposed to be an independent refactoring, but the code state does not compile.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Font.java
ShahzaibIbrahim: Follow up ticket created: https://github.com/vi-eclipse/Eclipse-Platform/issues/296

HeikoKlare (changes requested):
Now the commit contents are the wrong way round and do not fit their commit messages: the first, refactoring commit contains the actual change (creating handles on demand) and the second change supposed to create handles on demand encapsulates the field and adds some additional checks and corrects formatting.

PR #2088, opened 2025-05-02
Deprecate Image(Rectangle) constructor and replace usages

ShahzaibIbrahim:
Deprecate Image constructor that accepts Rectangle in favor of Image(display, width, height).

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
@@ -527,7 +527,10 @@ private void createRepFromSourceAndApplyFlag(NSBitmapImageRep srcRep, int srcWid
  * </ul>
  *
  * @see #dispose()
+ *
+ * @deprecated use {@link Image#Image(Device, int, int)} instead
  */
+@Deprecated(since = "2025-06", forRemoval = true)
 public Image(Device device, Rectangle bounds) {
 	super(device);
 	if (bounds == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_Image.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_Image.java
@@ -119,6 +119,7 @@ public void test_ConstructorLorg_eclipse_swt_graphics_DeviceII() {
 	image.dispose();
 }
 
+@SuppressWarnings("removal")
 @Test
 public void test_ConstructorLorg_eclipse_swt_graphics_DeviceLorg_eclipse_swt_graphics_Rectangle() {
 	Image image;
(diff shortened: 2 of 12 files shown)


HannesWell:
As far as I can tell, this constructor is currently often used to take screenshots from Displays or Controls.
If we introduce a corresponding API for that, as proposed in https://github.com/eclipse-platform/eclipse.platform.swt/pull/2104, the callers would be reduced a lot, when using the new API.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
```diff
@@ -527,7 +527,11 @@ private void createRepFromSourceAndApplyFlag(NSBitmapImageRep srcRep, int srcWid
  * </ul>
  *
  * @see #dispose()
+ *
+ * @deprecated use {@link Image#Image(Device, int, int)} instead
+ * @since 3.130
  */
+@Deprecated(since = "2025-05", forRemoval = true)
```
HeikoKlare, line 534: The release is 2025-06
```suggestion
@Deprecated(since = "2025-06", forRemoval = true)
```

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
```diff
@@ -527,7 +527,11 @@ private void createRepFromSourceAndApplyFlag(NSBitmapImageRep srcRep, int srcWid
  * </ul>
  *
  * @see #dispose()
+ *
+ * @deprecated use {@link Image#Image(Device, int, int)} instead
+ * @since 3.130
```
HeikoKlare, line 532: The since tag is wrong. This method was there before 3.130.
```suggestion
```

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_Image.java
```diff
@@ -119,6 +119,7 @@ public void test_ConstructorLorg_eclipse_swt_graphics_DeviceII() {
 	image.dispose();
 }
 
+@SuppressWarnings("deprecation")
```
HeikoKlare, line 122: Why do we need to add this?

HeikoKlare (commented):
> As far as I can tell, this constructor is currently often used to take screenshots from Displays or Controls.

That means it's a convenience constructor for the case where you have bounds (as rectangle) at hand, but the decision was actually questionable as such bounds may have x/y != 0, which has no proper meaning when using this constructor.
So having specific API for the screenshot use cae as proposed in https://github.com/eclipse-platform/eclipse.platform.swt/pull/2104 would be nice.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_Image.java
ShahzaibIbrahim: We needed @SuppressWarnings("removal") instead. The reason to add it is that this test specifically test the Image(Rectangle) constructor. Which is set to be removed at some point.

fedejeanne (approved):
LGTM. 

I replaced some usages of the deprecated constructor in GTK and added a new test for the preferred constructor

fedejeanne:
Test failure unrelated: https://github.com/eclipse-platform/eclipse.platform.swt/issues/2113

HeikoKlare:
Please process the PRs replacing the consumers as soon as possible (e.g., https://github.com/eclipse-platform/eclipse.platform.ui/pull/2953). In general, it would be best to replace all known consumers before deprecating the method.
Otherwise PR builds for repos with consumers may show errors because of introduced warnings, even though they were not introduced by the PR itself.

PR #2090, opened 2025-05-05
Do not use DPIUtil.getNativeDeviceZoom() to get the cursor handle #2057

fedejeanne:
Supersedes https://github.com/eclipse-platform/eclipse.platform.swt/pull/2071
 
Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2057

## Requires
- #2092 

# How to test
- Use 2 monitors with different zoom levels (none of them should match the zoom level passed to `-Dswt.autoScale` in the next step)
- Run `Snippet119` with the following JVM parameters: 
  - `-Dswt.autoScale=200` (or use another zoom level that doesn't match any of the zooms in your monitors)
  - `-Dswt.autoScale.updateOnRuntime=true`

Expected: the size of the cursor matches the size of the current monitor and **not** the one you provided for `-Dswt.autoScale`
 
- Move the window of the snippet to the monitor 

Expected: the size of the cursor still matches the size of the current monitor and **not** the one you provided for `-Dswt.autoScale`


fedejeanne:
Test failure is unrelated: #1843

Issue #2060, opened 2025-04-24
The new Image disabling algorithm sets every disabled images hue to red


killingerm:
[Line 36](https://github.com/eclipse-platform/eclipse.platform.swt/blame/28fdeb6dca72d94a6f8c1a8a47c285d0c089a03e/bundles/org.eclipse.swt/Eclipse%20SWT/common/org/eclipse/swt/internal/image/ImageColorTransformer.java#L36) in class ImageColorTransformer.java puts an upper bound of 1.0f to the hue value of a pixel, but hue is actually in the interval [0, 360) and represents the degree on a color circle. Limiting to 1.0f sets the hue to its corresponding color red.

`float hue = Math.min(hueFactor * hsba[0], 1.0f);`

Maybe a modulo operation instead of Math.min would be more applicable to represent a shift around the color circle:

```
float hue = (hueFactor * hsba[0]) % 360.0f;
```

HeikoKlare:
Thank you for the catch, @killingerm! The proposal to use a modulo operation sounds reasonable to me. Do you want to provide a PR with that fix?

killingerm:
@HeikoKlare Yes, I'll provide a PR

PR #2091, opened 2025-05-05
Remove hue from ImageColorTransformer#forHSB #2060

HeikoKlare:
The ImageColorTransformer introduced for disabled icon calculation has a default implementation taking hue, saturation, brightness and alpha factors for transformation. The hue factor was by accident limited to 1, even though it is a value between 0 and 360. In addition, a multiplication of a hue value with a factor is not reasonable at all.

This change removes the ability to change the hue of a color value, as the existing usage of that method does not change the hue anyway. It renames the method accordingly.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2060


HeikoKlare:
@killingerm I have created this PR with you as co author as the fix should be in M2 (in case people start using the new disablement algorithm already promoted via the new and noteworty) and today is last chance to do it. I have also adapted you proposal as applying a factor to a hue value does, in my opinion, not make any sense anyway and since that value is not used by the existing consumers, I have now simply removed it.

killingerm:
> @killingerm I have created this PR with you as co author as the fix should be in M2 (in case people start using the new disablement algorithm already promoted via the new and noteworty) and today is last chance to do it. I have also adapted you proposal as applying a factor to a hue value does, in my opinion, not make any sense anyway and since that value is not used by the existing consumers, I have now simply removed it.

@HeikoKlare Thank you for fixing it, my IDE broke after an update and I had to reinstall therefore din't manage it in time. Also just removing it was probably the best option.

HeikoKlare:
No worries, I just urged with this to have a proper version of the disablement algorithm provided with M2, so to not ship a broken one that may give a bad first impression.

PR #2093, opened 2025-05-05
Edge: Ensure #getText() works even when javascript is disabled

### LINKED ISSUE #2029: Edge browser getText() returns null when scripts are disabled
**Describe the bug**
Since implementation of the `getText()` method of Edge SWT browser uses javascript snippet to get the content (`return evaluate("return document.documentElement.outerHTML;")`), it will always return `null` if the browser has scripts execution disabled.

**To Reproduce**
Consequences of this bug can be seen in Javadoc views of JDT (see https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2165 if interested).I am not aware of any other and/or more direct way to see manifestation of this problem.

**Expected behavior**
Edge browser widget should return it's current content as a String even if javascript execution is disabled for the browser.

**Screenshots**
n/a

**Environment:**
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 

Any Windows with Edge (WebView2) component available. Mine is Windows 11 24H2.

3. JRE/JDK version
I assume it's irrelevant. Mine is Temurin-21.0.2+13.
 
**Version since**
Since Eclipse 4.35 - when Edge was made default browser on Windows.

Causing https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2165 which can (and most probably will) be fixed with a workaround that stops relying on `getText()`.

sratz:
Fixes #2029.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
@@ -916,6 +916,14 @@ public Object evaluate(String script) throws SWTException {
 	// Feature in WebView2. ExecuteScript works regardless of IsScriptEnabled setting.
 	// Disallow programmatic execution manually.
 	if (!jsEnabled) return null;
+	return evaluateInternal(script);
+}
+
+/**
+ * Unconditional script execution, bypassing {@link WebBrowser#jsEnabled} flag /
+ * {@link Browser#setJavascriptEnabled(boolean)}.
+ */
+private Object evaluateInternal(String script) throws SWTException {
 	if(inCallback > 0) {
 		// Execute script, but do not wait for async call to complete as otherwise it
 		// can cause a deadlock if execute inside a WebView callback.
@@ -965,7 +973,7 @@ String getJavaCallDeclaration() {
 
 @Override
 public String getText() {
-	return (String)evaluate("return document.documentElement.outerHTML;");
+	return (String)evaluateInternal("return document.documentElement.outerHTML;");
 }
 
 @Override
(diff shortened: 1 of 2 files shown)


bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -916,6 +916,14 @@ public Object evaluate(String script) throws SWTException {
 	// Feature in WebView2. ExecuteScript works regardless of IsScriptEnabled setting.
 	// Disallow programmatic execution manually.
 	if (!jsEnabled) return null;
+	return evaluateInternal(script);
+}
+
+/**
+ * Uncoditional script execution, bypassing {@link WebBrowser#jsEnabled} flag /
```
HeikoKlare, line 923: nit
```suggestion
 * Unconditional script execution, bypassing {@link WebBrowser#jsEnabled} flag /
```

HeikoKlare (approved):
Change and test look completely sound. Just found a typo, but since the documentation is internal it may also stay as is.

I thought about whether there are any security concerns regarding this change, such as misusing `getText()` to execute something while Javascript is disabled. But someone actively calling `getText()` can also simply call `setJavascriptEnabled()`, which made me conclude that there is no such issue.

sratz:
> Change and test look completely sound. Just found a typo, but since the documentation is internal it may also stay as is.

Fixed, thanks!

> I thought about whether there are any security concerns regarding this change, such as misusing `getText()` to execute something while Javascript is disabled. But someone actively calling `getText()` can also simply call `setJavascriptEnabled()`, which made me conclude that there is no such issue.

If you have access to the code or can influence the `setJavascriptEnable()` state, then there is no issue whatsoever, yes.

But if not, we need to make sure that the scripts passed to `executeInternal(String)` are not controllable by the end-users at runtime, neither explicitly not inadvertently.

For this use-case this is guaranteed as the string is a constant.

Should we get this merged Today to get it into M2?

Issue #2072, opened 2025-04-28
[Regression] Can't click on JavaDoc popup, resize or scroll


nlisker:
From: https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2187

Version: 2025-06 (4.36)
Build id: I20250423-0800
Windows 10

![Image](https://github.com/user-attachments/assets/5bd50ca4-309b-41a4-8b1e-5055395901a9)

Opening a popup JavaDoc as shown above does not allow to scroll it, making the continuation unreadable. It's also not resizable. In fact, clicking on it does nothing, but if you click on it and then switch to another non-eclipse window, the popups stay as an overlay.

This is a regression. In
Version: 2024-12 (4.34)
Build id: I20241120-1800

It is both resizable and scrollable, and the controls on the bottom appear. The windowing also behaves correctly after clicking on it.

![Image](https://github.com/user-attachments/assets/3692e69b-8d12-44d0-8a79-2976785e3f6e)

## Workaround
Use the `Tab` key to move from the forst popup to the JavaDoc popup

![Image](https://github.com/user-attachments/assets/e2755a79-2693-4537-947f-374534396bf7)

Phillipus:
Tabbing into the popup works, as does clicking on the area below the text (where it says "Press 'tab' from proposal table or click for focus"). Just not clicking into the main area.

fedejeanne:
Thank you for reporting it, @nlisker . I added a workaround for the issue. I'll be looking at it together with @amartya4256 .

fedejeanne:
Ah, I just realized that I had already reported this a while back: https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2061

fedejeanne:
Since this issue manifests in JDT but it is actually originated in SWT, I'm reopening this issue and closing https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2061 as the duplicate (it's easier to track it that way)

PR #2097, opened 2025-05-05
Edge: Implement mouse-related event listener support

### LINKED ISSUE #2164: Edge: addMouseWheelListener and addMouseListener have no effect
**Describe the bug**
When using an Edge browser, if the methods `addMouseWheelListener` or `addMouseListener` are used, the listener is registered but never executed.

**To Reproduce**
```java
    Browser browser = new Browser(browserContainer, SWT.NONE);

    browser.addMouseWheelListener(e -> {
      System.out.println(e.toString());
    });
    browser.addMouseListener(new MouseListener() {

      @Override
      public void mouseUp(final MouseEvent e) {
        System.out.println(e.toString());
      }

      @Override
      public void mouseDown(final MouseEvent e) {
        System.out.println(e.toString());
      }

      @Override
      public void mouseDoubleClick(final MouseEvent e) {
        System.out.println(e.toString());
      }
    });
```

**Expected behavior**
The listeners are called when the scroll wheel is used or the buttons of the mouse is are clicked inside the Edge window

**Environment:**
- - [] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 

Use the SWT.IE when creating the browser.
Use a javascript function to react to the scroll wheel / buttons and register a normal `BrowserFunction` as a callback.

sratz:
- MouseListener
  - up
  - down
  - doubleClick

- MouseMoveListener
  - move

- MouseTrackListener
  - exit
  - enter

- MouseWheelListener
  - scroll

- DragDetectListener
  - dragDetected

The implementation is JavaScript-based by attaching listeners to the DOM for all relevant events and forwarding them to the WebView via `window.chrome.webview.postMessage()`.

The event handling is analogous to IE's
`IE#handleDOMEvent(org.eclipse.swt.ole.win32.OleEvent)`.

Note: Since the implementation is JavaScript-based, this requires JavaScript to be enabled on the Browser instance.

When JavaScript is disabled, we keep using the timer-based fallback implementation introduced in #1551.
As JavaScript is only truly enabled *after* navigation has finished, we also keep using this workaround when the Browser is first instantiated and before the first page has loaded.

This change also fixes the jsEnabled flag lifecycle by updating it in `handleNavigationCompleted()`, same as in IE.

This resolves #2164 as long as JavaScript is enabled.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
@@ -24,6 +24,7 @@ import java.util.List;
 import java.util.concurrent.*;
 import java.util.concurrent.atomic.*;
 import java.util.function.*;
+import java.util.stream.*;
 
 import org.eclipse.swt.*;
 import org.eclipse.swt.graphics.*;
@@ -88,8 +89,33 @@ class Edge extends WebBrowser {
 	private boolean ignoreFocusIn;
 	private String lastCustomText;
 
+	private boolean atLeastOneTopNavigationCompleted;
 	private static record CursorPosition(Point location, boolean isInsideBrowser) {};
 	private CursorPosition previousCursorPosition = new CursorPosition(new Point(0, 0), false);
+
+	private static final String WEBMESSAGE_KIND_MOUSE_RELATED_DOM_EVENT = "mouse-related-dom-event";
+	private static final String EVENT_MOUSEDOWN = "mousedown"; //$NON-NLS-1$
+	private static final String EVENT_MOUSEUP = "mouseup"; //$NON-NLS-1$
+	private static final String EVENT_MOUSEMOVE = "mousemove"; //$NON-NLS-1$
+	private static final String EVENT_MOUSEOVER = "mouseover"; //$NON-NLS-1$
+	private static final String EVENT_MOUSEOUT = "mouseout"; //$NON-NLS-1$
+	private static final String EVENT_DRAGSTART = "dragstart"; //$NON-NLS-1$
+	private static final String EVENT_DRAGEND = "dragend"; //$NON-NLS-1$
+	private static final String EVENT_DOUBLECLICK = "dblclick"; //$NON-NLS-1$
+	private static final String EVENT_MOUSEWHEEL = "wheel"; //$NON-NLS-1$
+	private static final Set<String> MOUSE_RELATED_DOM_EVENTS = Set.of( //
+			EVENT_MOUSEDOWN, EVENT_MOUSEUP, //
+			EVENT_MOUSEMOVE, //
+			EVENT_MOUSEOVER, EVENT_MOUSEOUT, //
+			EVENT_DRAGSTART, EVENT_DRAGEND, //
+			EVENT_DOUBLECLICK, //
+			EVENT_MOUSEWHEEL//
+	);
+	private static record MouseRelatedDomEvent(String eventType, boolean altKey, boolean ctrlKey, boolean shiftKey,
+			long clientX, long clientY, long button, boolean fromElementSet, boolean toElementSet, long deltaY) {
+	}
+	private int lastMouseMoveX;
+	private int lastMouseMoveY;
 
 	static {
 		NativeClearSessions = () -> {
@@ -775,6 +801,9 @@ void setupBrowser(int hr, long pv) {
 	handler = newCallback(this::handleSourceChanged);
 	webView.add_SourceChanged(handler, token);
 	handler.Release();
+	handler = newCallback(this::handleWebMessageReceived);
+	webView.add_WebMessageReceived(handler, token);
+	handler.Release();
 	handler = newCallback(this::handleMoveFocusRequested);
 	controller.add_MoveFocusRequested(handler, token);
 	handler.Release();
@@ -814,7 +843,7 @@ void setupBrowser(int hr, long pv) {
 	browser.addListener(SWT.FocusIn, this::browserFocusIn);
 	browser.addListener(SWT.Resize, this::browserResize);
 	browser.addListener(SWT.Move, this::browserMove);
-	scheduleMouseMovementHandling();
+	scheduleFallbackMouseMovementHandlingIfNeeded();
 
 	// Sometimes when the shell of the browser is opened before the browser is
 	// initialized, nothing is drawn on the shell. We need browserResize to force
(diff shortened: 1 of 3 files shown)


HeikoKlare:
May this fix https://github.com/eclipse-platform/eclipse.platform.swt/issues/2072?

sratz:
> May this fix ...?
> 
> * [[Regression] Can't click on JavaDoc popup, resize or scroll #2072](https://github.com/eclipse-platform/eclipse.platform.swt/issues/2072)

Not sure.

I just remember that we've been down this road before.

The problem is that if we disable JavaScript via `ICoreWebView2Settings.put_IsScriptEnabled(boolean)`, this will also prevent those listeners from running :(

akurtakov:
Is this still being considered?

HeikoKlare:
> Is this still being considered?

Yes, we still have https://github.com/eclipse-platform/eclipse.platform.swt/issues/2164 and just had an offline talk about how we want to proceed with this a few weeks ago. I expect that we proceed with this or have something else supercede it in the next weeks.

sratz:
> > Is this still being considered?
> 
> Yes, we still have #2164 and just had an offline talk about how we want to proceed with this a few weeks ago. I expect that we proceed with this or have something else supercede it in the next weeks.

As discussed offline, I changed the implementation to keep the fallback in place
- before first page load (where we don't have any javascript yet)
- when javascript is disabled

sratz:
This should be ready to go now.

sratz:
@HeikoKlare Should we get this into M1?

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -88,9 +89,34 @@ public WebViewEnvironment(ICoreWebView2Environment environment) {
...
+	private static final String EVENT_MOUSEUP = "mouseup"; //$NON-NLS-1$
+	private static final String EVENT_MOUSEMOVE = "mousemove"; //$NON-NLS-1$
+	private static final String EVENT_MOUSEOVER = "mouseover"; //$NON-NLS-1$
+	private static final String EVENT_MOUSEOUT = "mouseout"; //$NON-NLS-1$
+	private static final String EVENT_DRAGSTART = "dragstart"; //$NON-NLS-1$
+	private static final String EVENT_DRAGEND = "dragend"; //$NON-NLS-1$
+	private static final String EVENT_DOUBLECLICK = "dblclick"; //$NON-NLS-1$
+	private static final String EVENT_MOUSEWHEEL = "wheel"; //$NON-NLS-1$
+	private static final Set<String> MOUSE_RELATED_DOM_EVENTS = Set.of( //
+			EVENT_MOUSEDOWN, EVENT_MOUSEUP, //
```
amartya4256, line 107: Can we get rid of the trailing //? what is the purpose of them here?

sratz: I wanted to group the different semantics / pairs per line

merks: Yes I often do that too to make details more readable.

amartya4256: Okay good to know :)

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -1529,6 +1599,136 @@ int handleMoveFocusRequested(long pView, long pArgs) {
...
+		System.err.println(e);
+	}
+	return COM.S_OK;
+}
+
+/**
+ * Insipired by the mouse-event related parts of
+ * {@link IE#handleDOMEvent(org.eclipse.swt.ole.win32.OleEvent)}
+ */
+private void handleMouseRelatedDomEvent(MouseRelatedDomEvent domEvent) {
```
amartya4256, line 1637: Since this method can take care of mouse related events, shall we get rid of `handleMouseMovement` and implement it here instead? Then we can also handle them right here.

sratz: Not sure I understand.

`handleMouseMovement()` is the fallback handling when javascript is disabled which produces artifical events.

`handleMouseRelatedDomEvent(MouseRelatedDomEvent)` is the logic that is being invoked from within javascript.

Getting rid of `handleMouseMovement` would mean we'd have to produce artificial DOM events inside the `scheduleMouseMovementHandlingIfNeeded()` lambda just to unpack them again in `handleMouseRelatedDomEvent`.

I don't think this makes sense.

But we should probably rename
- `scheduleMouseMovementHandlingIfNeeded()` -> `scheduleFallbackMouseMovementHandlingIfNeeded()`
- `handleMouseMovement()` -> `handleFallbackMouseMovement()`

for clarity.

amartya4256: Oh I completely misunderstood. The name changes are fine. I would like to propose that we change the field `needsMouseMovementFallback` to a private method. I'll comment where its intended.

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -1323,6 +1388,11 @@ int handleNavigationCompleted(long pView, long pArgs, boolean top) {
 		// ProgressListener.completed from here.
 		sendProgressCompleted();
 	}
+	if (top) {
+		jsEnabled = jsEnabledOnNextPage;
+		needsMouseMovementFallback = !jsEnabled;
```
amartya4256: How about we remove it and create a private method `needsMouseMovementFallback()` which always returns `!jsEnabled`. Then we do not rely on `jsEnabled` being set in `handleNavigationCompleted` since it is a package protected field.

sratz: The problem is that the initial value of `jsEnabled` is true, but the initial value of `needsMouseMovementFallback` also must be true, because we need the fallback at least once until the first website has successfully loaded and we then receive javascript events.

That's why I introduced this separate field.

sratz: I'll add a comment to the field explaining this subtlety

sratz: Alternatively, we could introduce a new field `atLeastOnceTopNavigationCompleted`, and then have something like


```
private boolean atLeastOnceTopNavigationCompleted;

boolean needsMouseMovementFallback() {
  return !atLeastOnceTopNavigationCompleted || !jsEnabled;
}
```

amartya4256: Ahhh, since they are counter-opposite of each other. But this still puts needsMouseMovementFallback at the risk of being inconsistent. What if jsEnabled is updated from outside of Edge class. I know its not ideal but still possible.

sratz: `jsEnabled` is never changed directly from the outsite.

`setJavaScriptEnabled(boolean)` updates `jsEnabledOnNextPage` and only in `handleNavigationCompleted()` are `jsEnabled` + `needsMouseMovementFallback` then updated.

amartya4256: Yes I agree its not set from outside but since it is accessible from outside, I meant from access modifier perspective.
We should use a getter instead. I like your proposal above more than the current state for safety.

sratz:
> I tested the code out. Functionality-wise the code uses the new handler when javascript is enabled to capture cursor movements.

Thanks for testing and reviewing!

> Although, I think the `needsMouseMovementFallback` being assigned in the method `handleNavigationCompleted` is a bit iffy. I am just being nitpicky here. I approve this PR.

I fully agree. I was also struggling and refactoring this part over several iterations - and it always felt off. Even when refactoring it into a helper method, the place where this condition effectively must change is still in `handleNavigationCompleted`, as there the jsEnabledFlag is toggled. So I thought it's at least consistent to have 2 flags and not 1 flag + 1 helper method (which uses the first flag as well).

sratz:
I've given it some more thought and I think the flag `atLeastOneTopNavigationCompleted` isn't that bad actually. It expresses better why we need the fallback handling.

amartya4256:
> I've given it some more thought and I think the flag `atLeastOneTopNavigationCompleted` isn't that bad actually. It expresses better why we need the fallback handling.

Yes, I think it is very descriptive. Let's go with it.

sratz:
Test failures on Windows are unrelated.

PR #2099, opened 2025-05-05
[Win32] Move OS-version-related functionality to OsVersion class

HeikoKlare:
With the separation of the OS version check to the separate OsVersion class/DLL, version concerns have been spread over the OS and the OsVersion classes. This change consistently moves every such logic to the OsVersion class. It also replaces the manual checks of the current OS version against a specified version by an easy-to-read access to according constants identifying whether the current system is at least a specific Windows build version.

Follow up to:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2054


bundles/org.eclipse.swt/Eclipse SWT PI/win32/org/eclipse/swt/internal/win32/version/OsVersion.java
```diff
@@ -31,16 +31,36 @@ public class OsVersion extends Platform {
...
+
 	/**
 	 * Values taken from https://en.wikipedia.org/wiki/List_of_Microsoft_Windows_versions
 	 */
-	public static final int WIN32_BUILD_WIN10_1607 = 14393; // "Windows 10 August 2016 Update"
+	public enum OsBuild {
+		WIN10_1607(14393), // "Windows 10 August 2016 Update"
+		WIN10_1809(17763), // "Windows 10 October 2018 Update"
+		WIN10_2004(19041), // "Windows 10 May 2020 Update"
+		WIN11_21H2(22000); // Initial Windows 11 release
```
HannesWell, line 43: Could this just be static final boolean constants (that are initialized in a static initializer)?:
- IS_WIN10_1607
- IS_WIN10_1809
- ...

Then one would avoid the isPresent() call.
that are ini

HeikoKlare: Good idea to further simplify this by defining this as constants instead of doing dynamic calculations. Thank you!
Adapted in https://github.com/eclipse-platform/eclipse.platform.swt/commit/77655b5fa3f32f12d84e067c4b61c7be2e4416d7.

HeikoKlare:

PR #2059, opened 2025-04-24
[GTK] Stop using DPIUtil (take 2)

akurtakov:
Handles remaining widgets code. Continuation of
https://github.com/eclipse-platform/eclipse.platform.swt/pull/2058 .
Part of https://github.com/eclipse-platform/eclipse.platform.swt/issues/1300

PR #2100, opened 2025-05-05
[GTK] Simplify Tree/Table

akurtakov:
This is continuation of
https://github.com/eclipse-platform/eclipse.platform.swt/pull/2059 . There is no point in having these extra methods (and double checkWidget method calls for so many things).


akurtakov:
Failing Windows tests are unrelated to pure GTK changes. Merging.

PR #2103, opened 2025-05-05
Introduce simple API to save ImageData to file and support '.jpg' extension

HannesWell:
Currently it's quite cumbersome to save ImageData into a file. screenshot from a `Control` or a `Display`. This proposes a convenience API to simplify this task.

This also complements the ImageData.load(Path) method introduced in
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/1767

It becomes also very simple to take and save a screenshot together with 
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2104, .

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/ImageData.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/ImageData.java
@@ -15,9 +15,11 @@ package org.eclipse.swt.graphics;
 
 
 import java.io.*;
+import java.nio.file.Path;
 
 import org.eclipse.swt.*;
 import org.eclipse.swt.internal.*;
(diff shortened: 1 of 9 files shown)


laeubi (commented):
I don't think we should clutter the `ImageData` class more can't we use ImageLoader(sic) for this that already is used internally anyways?

laeubi:
I don't think that adding methods to `ImageData` is a good idea, this should only be a data carrier and is not a general purpose place to put continence methods.

So why cant this static methods live in `ImageLoader` (or we can have a new `ImageWriter`)?

akurtakov:
Would you please split the *.jpg support in its own PR so it can be merged now?

akurtakov:
Is there still interest in this one or it should be closed?

akurtakov:
Looks like no interest. Closing.

Issue #673, opened 2023-05-16
Screenshots on Ubuntu 22 are always white [Snippet215]


DieterMaiSysgo:
**Describe the bug**
When creating a screenshot as described by Snippet215, the resulting image is all white. I would expect that the image instead contains a content reflecting the pixels displayed when the screenshot was taken.

**To Reproduce**
Easiest would be to execute Snippet215 on Ubuntu 22.
But essentially the following code seems not to work
```
	GC gc = new GC(display);
	Rectangle bounds = display.getBounds();
	final Image image = new Image(display, display.getBounds());
	gc.copyArea(image, 0, 0);
```

**Expected behavior**
I would expect to be able to create an screenshot with SWT. Preferably as demonstrated in Snippet215

**Screenshots**
<img width="330" alt="SWT screenshot issue" src="https://github.com/eclipse-platform/eclipse.platform.swt/assets/91322115/d18913f3-e16f-4be0-87c3-528472934aa8">

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [ ] Windows
- - [x] Linux 
- - [ ] macOS 

2. Additional OS info (e.g. OS version, Linux Desktop, etc)
I tested this on Windows 10, Ubuntu 20 and Ubuntu 22. Only Ubuntu 22 has this issue.

3. JRE/JDK version
Tested with JDK 17 

**Version since**
Tested with 4.16 and 4.17. Both had the same issue

**Workaround (or) Additional context**
I was not able to find a workaround.
I found this issue while investigating why our SWTBot test does not produce any screenshots. SWTBot uses SWT to create screenshot in a similar way as Snippet215

Kind regards
Dieter Mai

iloveeclipse:
- Which GTK3 / Eclipse version is used?
- Are you on Wayland? If yes, please try with X11.

DieterMaiSysgo:
> * Which GTK3  version is used?
```
$ dpkg -l libgtk-3-0
Desired=Unknown/Install/Remove/Purge/Hold
| Status=Not/Inst/Conf-files/Unpacked/halF-conf/Half-inst/trig-aWait/Trig-pend
|/ Err?=(none)/Reinst-required (Status,Err: uppercase=bad)
||/ Name             Version          Architecture Description
+++-================-================-============-====================================
ii  libgtk-3-0:amd64 3.24.33-1ubuntu2 amd64        GTK graphical user interface library
```
[...]

PR #2104, opened 2025-05-05
Introduce simple API to take a screenshot

HannesWell:
Currently it's quite cumbersome to take a screenshot from a `Control` or a `Display`. This proposes a convenience API to simplify this task.

When fully used within SWT it significantly reduces the calls to the to be deprecated Image constructor:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2088

Together with https://github.com/eclipse-platform/eclipse.platform.swt/pull/2103, it becomes very simple to take and save a screenshot.

I'm not yet sure where such a method is added best. Currently it's a new static method in the general `SWT` class. Alternativly it could be a static 'factory' like method in the `Image`/`ImageData` class or it could be in a new utility class or somewhere else?

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/SWT.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/SWT.java
@@ -15,6 +15,7 @@
 package org.eclipse.swt;
 
 
+import org.eclipse.swt.graphics.*;
 import org.eclipse.swt.internal.*;
 import org.eclipse.swt.widgets.*;
 
@@ -5010,4 +5011,39 @@ static {
 		MOD4 = 0;
 	}
 }
+
+/**
+ * Takes a screenshot of the given {@code display} and returns it as image.
+ *
+ * @param display the display to take a screenshot from
+ * @return the ImageData of the screenshot taken
+ * @since 3.130
+ */
+public static ImageData takeScreenShot(Display display) {
+	return takeScreenShot(display, display, display.getBounds());
+}
+
+/**
+ * Takes a screenshot of the given {@code control} and returns it as image.
+ *
+ * @param control the control to take a screenshot from
+ * @return the ImageData of the screenshot taken
+ * @since 3.130
+ */
+public static ImageData takeScreenShot(Control control) {
+	return takeScreenShot(control, control.getDisplay(), control.getBounds());
+}
+
+private static ImageData takeScreenShot(Drawable source, Display display, Rectangle bounds) {
+	Image image = new Image(display, bounds);
+	GC gc = new GC(source);
+	try {
+		gc.copyArea(image, 0, 0);
+		return image.getImageData();
+	} finally {
+		gc.dispose();
+		image.dispose();
+	}
+}
+
 }
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Control.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Control.java
@@ -60,6 +60,7 @@ public abstract class Control extends Widget implements Drawable {
 	long firstFixedHandle = 0;
 	long keyController;
 	long redrawWindow, enableWindow, provider;
+	//TODO: derive alpha from color?
 	int drawCount, backgroundAlpha = 255;
 	long dragGesture, zoomGesture, rotateGesture, panGesture;
 	Composite parent;
--- a/tests/org.eclipse.swt.tests.gtk/ManualTests/org/eclipse/swt/tests/gtk/snippets/Bug531667_CanvasPrint_does_not_work.java
+++ b/tests/org.eclipse.swt.tests.gtk/ManualTests/org/eclipse/swt/tests/gtk/snippets/Bug531667_CanvasPrint_does_not_work.java
@@ -19,7 +19,6 @@ import java.io.File;
 import org.eclipse.swt.SWT;
 import org.eclipse.swt.graphics.Color;
 import org.eclipse.swt.graphics.GC;
-import org.eclipse.swt.graphics.Image;
 import org.eclipse.swt.graphics.ImageData;
 import org.eclipse.swt.graphics.ImageLoader;
 import org.eclipse.swt.graphics.Rectangle;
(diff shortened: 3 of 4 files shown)


laeubi:
What about a class `Screenshot` and then have `Screenshot.of(Display )` (because you actually capture always from a Display, e.g. if controls overlap one would still not get a control) and  then return an`Screenshot` instance that has methods to get the `Image` / `ImageData` (just in case someone want to operate on the Image itself) and then have `Screenshot#take(Control)` as convenience to call parameter or a `Screenshot.take(Recangle)` or just call `Screenshot.take()` for the full display, that would be much more flexible than a possible growing set of static methods.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/SWT.java
```diff
@@ -5010,4 +5011,39 @@ public static void error (int code, Throwable throwable, String detail) {
...
+ */
+public static ImageData takeScreenShot(Control control) {
+	return takeScreenShot(control, control.getDisplay(), control.getBounds());
+}
+
+private static ImageData takeScreenShot(Drawable source, Display display, Rectangle bounds) {
+	Image image = new Image(display, bounds);
+	GC gc = new GC(source);
+	try {
+		gc.copyArea(image, 0, 0);
```
ptziegler, line 5041: If the source is invisible, this will capture the area underneath the widget.

ptziegler:
Wouldn't it make more sense to have this be a property of the Control, rather than moving this to the SWT class? Eventually, this functionality should probably also be supported for Widgets in general and not just Controls, which doesn't really work with a static method.

laeubi:
> Wouldn't it make more sense to have this be a property of the Control, rather than moving this to the SWT class? Eventually, this functionality should probably also be supported for Widgets in general and not just Controls, which doesn't really work with a static method.

As mentioned one can not take a screenshot of a Control, this is only used to make a screendump _of the area_ where the control is currently located. To capture a control one can create a GC and "draw" the control then there.

ptziegler:
> As mentioned one can not take a screenshot of a Control, this is only used to make a screendump of the area where the control is currently located. To capture a control one can create a GC and "draw" the control then there.

But in that case I don't really the benefit for the user. Keeping an internal `Screenshot` class is fine, but I don't think it's a good idea to make this public API. Especially when capturing the display is something that's prohibited on both GTK4 and Wayland anyway... (https://github.com/eclipse-platform/eclipse.platform.swt/issues/673)

HannesWell:
> What about a class `Screenshot` and then have `Screenshot.of(Display )` (because you actually capture always from a Display, e.g. if controls overlap one would still not get a control) and then return an`Screenshot` instance that has methods to get the `Image` / `ImageData` (just in case someone want to operate on the Image itself) and then have `Screenshot#take(Control)` as convenience to call parameter or a `Screenshot.take(Recangle)` or just call `Screenshot.take()` for the full display, that would be much more flexible than a possible growing set of static methods.

As written in my initial message, I appreciate any suggestion where this new functionality is discoverable and flexibly usable. So if it makes sense to have various configuration options a new `Screenshot` class sounds good.



> As mentioned one can not take a screenshot of a Control, this is only used to make a screendump of the area where the control is currently located. To capture a control one can create a GC and "draw" the control then there.
> 

I see that there is also a method `Control.print()`:
https://github.com/eclipse-platform/eclipse.platform.swt/blob/142dc7a00625d23f0641ec83bfb3614adac0b990/bundles/org.eclipse.swt/Eclipse%20SWT/win32/org/eclipse/swt/widgets/Control.java#L2190

Can you tell if that behaves differently and e.g. allows to record the surface of that Control even if it's not visible respectively not supported to capture the entire display?

> But in that case I don't really see the benefit for the user. Keeping an internal `Screenshot` class is fine, but I don't think it's a good idea to make this public API. Especially when capturing the display is something that's prohibited on both GTK4 and Wayland anyway... (#673)

That's kind of funny, Windows offers a 'feature' where it take screenshots automatically all the time (https://support.microsoft.com/de-de/windows/mit-recall-ihre-schritte-zur%C3%BCckverfolgen-aa03f8a0-a78b-4b3e-b0a1-2eb8ac48701c, article is in Germany, I didn't found the English version quickly).
But back to topic, is it supported on GTK4/Wayland to capture at least parts of the screen?

But even if it's not supported at all respectively impossible on some OS/WS combinations an API could still help users, if it's stated that it's an optional operation and if it fails cleanly with an exception if one is really in that situation. Then clients would at least not use some random snippets that don't work all the time and not fail cleanly, but just produce unexpected result.

As this topic doesn't seem to be that easy (thank you for all that input), I think it's even more important to help users and provide the 'best', most robust solutions possible.

ptziegler:
> But back to topic, is it supported on GTK4/Wayland to capture at least parts of the screen?

Screen capture is explicitly unsupported in Wayland for the sake of "security":

"By design, Wayland is a lot more secure than X11 and does not allow one application to capture the content of other applications' windows"

Or in short: This has been pushed down to the DE. So you can still take screenshots with Wayland, but now there's a Gnome way to do it, a KDE way to do it and so on... Not really something that can be integrated in SWT.

I think the "official" way is via the [xdg-desktop-portal](https://wiki.archlinux.org/title/XDG_Desktop_Portal) but I've never worked with it and you'd be yet again at the mercy of the DE to support this portal...

With GTK4, the GdkWindow has been made internal, so the required API is no longer available. See

https://github.com/eclipse-platform/eclipse.platform.swt/blob/142dc7a00625d23f0641ec83bfb3614adac0b990/bundles/org.eclipse.swt/Eclipse%20SWT/gtk/org/eclipse/swt/graphics/GC.java#L480-L482

You have the [GtkSnapshot](https://docs.gtk.org/gtk4/class.Snapshot.html) as replacement but which only works for widgets.

To summarize: Screen capture in Linux sucks.

ptziegler:
To add to my previous comment: It _might_ still be possible to capture the SWT widgets in GTK4/Wayland, which I believe is exactly what `print(...)` is doing.

mickaelistria:
Have you evaluated whether `org.eclipse.ui.workbench.texteditor.tests.ScreenshotTest.takeScreenshot(...)` can be reused?

bt90:
> I think the "official" way is via the [xdg-desktop-portal](https://wiki.archlinux.org/title/XDG_Desktop_Portal) but I've never worked with it and you'd be yet again at the mercy of the DE to support this portal...

That is the official way and it's supported across all mayor desktop environments 🙂

akurtakov:
@HannesWell Do you still have interest in this one? Or it should be closed?

akurtakov:
Looks like no interest in this one. Closing.

PR #2106, opened 2025-05-06
Fix surefire warnings in build

akurtakov:
Fixes following warnings in the build:
```
[WARNING] useSystemClassLoader setting has no effect when not forking
[WARNING] The parameter forkCount should likely not be 0. Forking a JVM
for tests improves test accuracy. Ensure to have a <forkCount> >= 1.
```

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -350,7 +350,7 @@ pipeline {
 					dir('eclipse.platform.swt') {
 						sh '''
 							mvn clean verify \
-								--batch-mode --threads 1C -V -U -e -DforkCount=0 \
+								--batch-mode --threads 1C -V -U -e \
 								-Pbree-libs -Papi-check -Pjavadoc \
 								-Dcompare-version-with-baselines.skip=false \
 								-Dorg.eclipse.swt.tests.junit.disable.test_isLocal=true \


Jenkinsfile
```diff
@@ -350,7 +350,7 @@ pipeline {
 					dir('eclipse.platform.swt') {
 						sh '''
 							mvn clean verify \
-								--batch-mode --threads 1C -V -U -e -DforkCount=0 \
+								--batch-mode --threads 1C -V -U -e -DforkCount=1 \
```
HannesWell, line 353: If I'm not mistake, the default is one. So just removing it would make it simpler.
```suggestion
								--batch-mode --threads 1C -V -U -e \
```

akurtakov: Trying it now.

akurtakov:
Windows test failure is in master for long time. @HeikoKlare @fedejeanne  What is the plan for this windows test failure? If there is none I'll disable it on Monday.

HeikoKlare:
> What is the plan for this windows test failure? If there is none I'll disable it on Monday.

It was a a sporadically failing test for a while (only failing in GHA). It just started to fail rather consistently on Jenkins. It was not reproducible locally. I was now able to reproduce it on a system with a recent Windows build, so might be that we are now able to fix it via local reproduction. I added information in https://github.com/eclipse-platform/eclipse.platform.swt/issues/1843.

PR #2107, opened 2025-05-06
[win32] Adapt/copy all handles on Image copy

akoch-yatta:
This PR adapts and copies all handles in the copy constructor in Image. As each image now manages possibly more than one OS handles at the same time in the windows implementation, it is important to apply the operation on all of them.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
@@ -244,41 +244,45 @@ public Image(Device device, Image srcImage, int flag) {
 	if (srcImage == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
 	if (srcImage.isDisposed()) SWT.error(SWT.ERROR_INVALID_ARGUMENT);
 	initialNativeZoom = srcImage.initialNativeZoom;
-	Rectangle rect = srcImage.getBounds(getZoom());
 	this.type = srcImage.type;
 	this.imageProvider = srcImage.imageProvider.createCopy(this);
 	this.styleFlag = srcImage.styleFlag | flag;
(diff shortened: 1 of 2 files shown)


HeikoKlare (commented):
The change looks sound. Just wondering if it is possible and would make sense to move the handle copy logic to the `ImageHandle` class? I.e., having an `ImageHandle::copy` operation or the like to extract the rather complex functionaltiy from the constructor?

akoch-yatta:
> The change looks sound. Just wondering if it is possible and would make sense to move the handle copy logic to the `ImageHandle` class? I.e., having an `ImageHandle::copy` operation or the like to extract the rather complex functionaltiy from the constructor?

Sure, we could. Only argument from my side against it would be that it is just not used anywhere else. But I would be fine with both.

HeikoKlare:
> Sure, we could. Only argument from my side against it would be that it is just not used anywhere else.

That's true. But wouldn't you say it's more a concern of an ImageHandle to implement how it has to be copied rather than the concern of a consumer (in this case the Image constructor requiring a copy of the handle)? I always found that Image constructor rather complex and now that it is refactored to obviously copy multiple handles it seems quite natural to me to have than be encapsulated in the ImageHandle.
But in the end, this is more about readability and encapsulation of concerns than about functionality.

fedejeanne (commented):
The changes look fine, I'm just missing some way to test it more "heavily". By playing around with the _Runtime Workspace_ Eclipse Application, I only seem to navigate the `case SWT.IMAGE_DISABLE` but not the other ones. @akoch-yatta do you have any Snippet or use case that tests the other `case`s in the `switch`?

I have no strong opinion regarding whether or not the code should be moved to `ImageHandle` so it's your call :-)

HeikoKlare (approved):
Since the case distinction for bitmap/icon would need to be replaced in the ImageHandle copy and in the Image constructor, let's keep it like it is for now. 

PR #2108, opened 2025-05-06
[win32] Utilize temporary image handles

akoch-yatta:
This PR adapts all places, where operation where executed with OS handles created for a specific zoom that could use any zoom. Therefor now any existing handle is used or, if none exists yet, a temporary handle is created and destroyed afterwards.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
@@ -1099,47 +1099,49 @@ public Color getBackground() {
 
 	/* Get the HDC for the device */
 	long hDC = device.internal_new_GC(null);
-	long handle = win32_getHandle(this, getZoom());
-
-	/* Compute the background color */
-	BITMAP bm = new BITMAP();
-	OS.GetObject(handle, BITMAP.sizeof, bm);
-	long hdcMem = OS.CreateCompatibleDC(hDC);
-	long hOldObject = OS.SelectObject(hdcMem, handle);
-	int red = 0, green = 0, blue = 0;
-	if (bm.bmBitsPixel <= 8)  {
-		byte[] color = new byte[4];
-		OS.GetDIBColorTable(hdcMem, transparentPixel, 1, color);
-		blue = color[0] & 0xFF;
-		green = color[1] & 0xFF;
-		red = color[2] & 0xFF;
-	} else {
-		switch (bm.bmBitsPixel) {
-			case 16:
-				blue = (transparentPixel & 0x1F) << 3;
-				green = (transparentPixel & 0x3E0) >> 2;
-				red = (transparentPixel & 0x7C00) >> 7;
-				break;
-			case 24:
-				blue = (transparentPixel & 0xFF0000) >> 16;
-				green = (transparentPixel & 0xFF00) >> 8;
-				red = transparentPixel & 0xFF;
-				break;
-			case 32:
-				blue = (transparentPixel & 0xFF000000) >>> 24;
-				green = (transparentPixel & 0xFF0000) >> 16;
-				red = (transparentPixel & 0xFF00) >> 8;
-				break;
-			default:
-				return null;
+	return applyUsingAnyHandle(imageHandle -> {
+		long handle = imageHandle.handle;
+		/* Compute the background color */
+		BITMAP bm = new BITMAP();
+		OS.GetObject(handle, BITMAP.sizeof, bm);
+		long hdcMem = OS.CreateCompatibleDC(hDC);
+		long hOldObject = OS.SelectObject(hdcMem, handle);
+		int red = 0, green = 0, blue = 0;
+		if (bm.bmBitsPixel <= 8)  {
+			byte[] color = new byte[4];
+			OS.GetDIBColorTable(hdcMem, transparentPixel, 1, color);
+			blue = color[0] & 0xFF;
+			green = color[1] & 0xFF;
+			red = color[2] & 0xFF;
+		} else {
+			switch (bm.bmBitsPixel) {
+				case 16:
+					blue = (transparentPixel & 0x1F) << 3;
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1832,6 +1838,19 @@ public String toString () {
...
+	if (zoomLevelToImageHandle.isEmpty()) {
+		ImageHandle temporaryHandle = this.imageProvider.newImageHandle(DPIUtil.getDeviceZoom());
+		try {
+			return function.apply(temporaryHandle);
+		} finally {
+			temporaryHandle.destroy();
+		}
+	} else {
+		return function.apply(zoomLevelToImageHandle.values().iterator().next());
+	}
```
fedejeanne, line 1851: Move outside of the `else`

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1258,7 +1262,9 @@ public ImageData getImageData (int zoom) {
  */
 @Deprecated
 public ImageData getImageDataAtCurrentZoom() {
-	return getImageMetadata(getZoom()).getImageData();
+	return applyUsingAnyHandle(imageHandle -> {
+		return imageHandle.getImageData();
+	});
```
fedejeanne, line 1267:
```suggestion
	return applyUsingAnyHandle(ImageHandle::getImageData);
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1176,7 +1178,9 @@ Rectangle getBounds(int zoom) {
  */
 @Deprecated
 public Rectangle getBoundsInPixels() {
-	return getBounds(getZoom());
+	return applyUsingAnyHandle(imageHandle -> {
+		return imageHandle.getBounds();
+	});
```
fedejeanne, line 1183:
```suggestion
	return applyUsingAnyHandle(ImageHandle::getBounds);
```

fedejeanne (approved):
The code looks OK and I saw no regressions when running the _Runtime Workspace_ Eclipse Application. I only have some minor comments regarding coding style.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1176,7 +1178,9 @@ Rectangle getBounds(int zoom) {
  */
 @Deprecated
 public Rectangle getBoundsInPixels() {
-	return getBounds(getZoom());
+	return applyUsingAnyHandle(imageHandle -> {
+		return imageHandle.getBounds();
```
HeikoKlare, line 1182: Currently, the method would return kind of an "arbitrary" value as the zoom at which the bounds are calculated is quite random. Would it make sense to return something more resonable, such as the bounds in pixels at 100& or, maybe better, at `DPIUtil.getDeviceZoom()` (as the latter might be meaningful in the case monitor-specific scaling is disabled)?

The same applies to `getImageDataAtCurrentZoom()`. Note that that method is used in `getScaledImageData()` of the provider wrapper when we create a GC on an image. So we probably need to return the image data at a reasonable zoom level there?

HeikoKlare (commented):
In general, the change looks fine. In particular the usage of `deviceZoom` for the temporary handle (if necessary) makes sense to me. I only have question regarding the implementation of the deprecated methods, most importantly the `getImageDataAtCurrentZoom()` one.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
HeikoKlare: > As both methods are deprecated I would argue to try to prevent to create additional handles for that. What I could think of is extending the apply method to pass a preferred zoom and adapt the logic like:

I didn't mean to create another or a different handle but just to scale the bounds to some meaningful value. But I just realized that in case monitor-specific scaling is disabled we would use the image at DPIUtil.getDeviceZoom() anyway (as that is either the one that was created or the one to be temporarily created). In since the return value is at least consistent with what `getImageDataAtCurrentZoom()` returns, it might be okay.

> I am a bit hesitant to use DPIUtil.getDeviceZoom() instead of 100% because we are aware of the issues with GC and zoom != 100%

Okay, I was just confused because there is code that uses `getImageDataAtCurrentZoom()` when a GC is created (i.e., `memGC != null`) and that may now return a handle at an arbitrary zoom. Is that intended?

akoch-yatta: > > I am a bit hesitant to use DPIUtil.getDeviceZoom() instead of 100% because we are aware of the issues with GC and zoom != 100%
> 
> Okay, I was just confused because there is code that uses `getImageDataAtCurrentZoom()` when a GC is created (i.e., `memGC != null`) and that may now return a handle at an arbitrary zoom. Is that intended?

Question is, what is the expected behavior here? Way before it always returned the existing handle, isn't this solution quite equal to that? Only difference is when there are multiple handles, it choses the first by hash, but how would you order them?

HeikoKlare: I was mistakenly thinking that the image data may be at a wrong size and that may lead to issues when drawing on them, but since the drawing happen in point coordinates it will always work. Only consequence might be that the underlying image is at a different zoom and thus might be scaled, but the worst that could happen is that the image becomes more/less/differently blurry.

HeikoKlare (approved):

HeikoKlare:

PR #2110, opened 2025-05-06
[win32] Improve GC init for ImageGcDrawer

akoch-yatta:
This PR wraps the GC initialization for the ImageGcDrawer to directly pass the zoom from the wrapper when the GC is initialized. In all other cases the GC will be initialized with 100 zoom.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
@@ -1719,6 +1719,10 @@ private ImageHandle init(ImageData i, int zoom) {
  */
 @Override
 public long internal_new_GC (GCData data) {
+	return configureGC(data, 100);
+}
+
+private long configureGC(GCData data, int zoom) {
 	if (isDisposed()) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
 	/*
 	* Create a new GC that can draw into the image.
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1707,6 +1707,10 @@ private ImageHandle init(ImageData i, int zoom) {
  */
 @Override
 public long internal_new_GC (GCData data) {
+	return configureGC(data, 100);
```
HeikoKlare, line 1710: Just for clarification: why do we use 100 instead of `DPIUtil.getNativeDeviceZoom()` here? Wouldn't the latter be more consistent with existing behavior (i.e., in case monitor-specific scaling is disabled an images are drawn this way)?

akoch-yatta: We know GC works better with 100% then with e.g. 125 or 175 that could be provided with `DPIUtil.getNativeDeviceZoom()` so I would like to prevent using `DPIUtil.getNativeDeviceZoom() ` as default

akoch-yatta: yes, it sure does. Before DPIUtil.getNativeZoom was used, but not for a specific reason, especially not for the GC, but just for creating the probably most suitable initial handle here So, for this default case I would argue, no value is really the correct value here. I chose 100 out of two reasons
-  drawing on the GC works best with 100% - e.g. having GEF in mind.
- all "easy" use cases involving GC and Images should be covered with ImageGcDrawer, leaving the complex ones (like. GEF) which brings me back to the upper point

HeikoKlare: Thank you for the clarification!
Just like in https://github.com/eclipse-platform/eclipse.platform.swt/pull/2108#discussion_r2075556883, I was mistakenly thinking that the image data may be at a wrong size and that may lead to issues when drawing on them, but since the drawing happens in point coordinates it will always work.
The decision makes sense to me.

HeikoKlare (approved):

PR #2114, opened 2025-05-07
[Win32] Do not initialize font when already disposed

HeikoKlare:
The on-demand creation of Font handles does not consider that a font may already have been disposed when first retrieving it's handle, which triggers the on-demand initialization. To avoid that this initialization after disposal fails, this change adds a check for the font being destroyed when performing the initialization.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/2960


amartya4256 (approved):
Seems reasonable to me.

fedejeanne (approved):
Makes sense

PR #2115, opened 2025-05-07
[win32] Fix for Image disablement

akoch-yatta:
This PR adjusts the order in the Image copy constructor. As the call to ImageProviderWrapper#createCopy creates cached variants for some implementation. These created variants could be missing disablement/greyed state. Therfore this flag is set before that.


HeikoKlare (approved):

HeikoKlare:
ftr: disablement did not work properly anymore after https://github.com/eclipse-platform/eclipse.platform.swt/pull/2107
image

After this PR it looks like this:
image

PR #2122, opened 2025-05-07
[win32] Fix double buffer paint with ImageGcDrawer

akoch-yatta:
This PR fixes a regression introduced with f10721a that did not properly retain the currently drawn zoom level when an ImageGcDrawer is used. That leads to issues with zoom != 100.


akoch-yatta:
@HeikoKlare please have a look

HeikoKlare (approved):
The change makes sense. We had this "zoom state" stored for a GC created on an image upon handle creation before as well, so it's sound to readd it.

I just tested at 125% monitor zoom to see the effect. The line number ruler does simply not update:
linenumberrule_no_change

That's properly fixed by this PR.

PR #2128, opened 2025-05-08
New FontData Constructor to create a deep copy

ShahzaibIbrahim:
This change will give us a chance to use proper channel to create a deep copy of FontData object instead of using FontData(fd.toString)

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/FontData.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/FontData.java
@@ -187,6 +187,31 @@ public FontData(String string) {
 }
 
 /**
+ * Constructs a deep copy of the given font data object.
+ *
+ * @param fontData the FontData object to copy
+ *
+ * @exception IllegalArgumentException
+ * <ul>
+ *    <li>ERROR_NULL_ARGUMENT - if the argument is null</li>
+ * </ul>
+ * @since 3.131
+ */
+public FontData(FontData fontData) {
+	if (fontData == null) {
+		SWT.error(SWT.ERROR_NULL_ARGUMENT);
+	}
+
+	this.name = fontData.name;
+	this.height = fontData.height;
+	this.style = fontData.style;
+	this.nsName = fontData.nsName;
+	this.lang = fontData.lang;
+	this.country = fontData.country;
+	this.variant = fontData.variant;
+}
+
+/**
  * Constructs a new font data given a font name,
  * the height of the desired font in points,
  * and a font style.
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/FontData.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/FontData.java
@@ -240,6 +240,59 @@ public FontData(String string) {
 	}
 }
 
+
+/**
+ * Constructs a deep copy of the given font data object.
+ *
+ * @param fontData the FontData object to copy
+ *
+ * @exception IllegalArgumentException
+ * <ul>
+ *    <li>ERROR_NULL_ARGUMENT - if the argument is null</li>
+ * </ul>
+ * @since 3.131
+ */
+public FontData(FontData fontData) {
+	if (fontData == null) {
+		SWT.error(SWT.ERROR_NULL_ARGUMENT);
+	}
+	this.height = fontData.height;
+	this.lang = fontData.lang;
+	this.country = fontData.country;
+	this.variant = fontData.variant;
+	this.data = cloneLogFont(fontData);
+}
+
+private static LOGFONT cloneLogFont(FontData fontData) {
+	if (fontData.data == null) {
+		return null;
+	}
+
+	LOGFONT existingData = fontData.data;
+	LOGFONT newData = new LOGFONT();
+	newData.lfHeight = existingData.lfHeight;
+	newData.lfWidth = existingData.lfWidth;
+	newData.lfEscapement = existingData.lfEscapement;
+	newData.lfOrientation = existingData.lfOrientation;
+	newData.lfWeight = existingData.lfWeight;
+	newData.lfItalic = existingData.lfItalic;
+	newData.lfUnderline = existingData.lfUnderline;
+	newData.lfStrikeOut = existingData.lfStrikeOut;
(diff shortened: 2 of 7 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/FontData.java
```diff
@@ -240,6 +240,52 @@ public FontData(String string) {
...
+		// Deep copy of the char array lfFaceName
+		if (fontData.data.lfFaceName != null) {
+			newData.lfFaceName = new char[fontData.data.lfFaceName.length];
+			System.arraycopy(fontData.data.lfFaceName, 0, newData.lfFaceName, 0, fontData.data.lfFaceName.length);
+		}
+
+		this.data = newData;
+	} else {
+		this.data = null;
+	}
```
HeikoKlare, line 285: This could be placed in a dedicated private static `cloneLogfont` method, in which you can also early return on that font being null.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/FontData.java
```diff
@@ -240,6 +240,52 @@ public FontData(String string) {
...
+ * </ul>
+ * @since 3.130
+ */
+public FontData(FontData fontData) {
+	if (fontData == null)
+		SWT.error(SWT.ERROR_NULL_ARGUMENT);
+
+	if (fontData.data != null) {
+		LOGFONT newData = new LOGFONT();
+		setHeight(fontData.getHeightF());
```
HeikoKlare, line 261: Instead of setting specific fields via setters, shouldn't we copy over all data in a proper copy constructor, i.e., also the height, lang, country and variant fields? This seems to be a copy-over from the String-based constructor, in which some original data is missing and needs to be re-initialized from the data being given using according setters.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/FontData.java
```diff
@@ -186,6 +186,26 @@ public FontData(String string) {
...
+ *
+ * @param fontData the FondData object for which copy is needed to be made
+ *
+ * @exception IllegalArgumentException
+ * <ul>
+ *    <li>ERROR_NULL_ARGUMENT - if the argument is null</li>
+ * </ul>
+ * @since 3.130
+ */
+public FontData(FontData fontData) {
```
HeikoKlare, line 200: Same as for the other implementations: why not simply copy the fields from the original font data instead of using setters with additional checks that may not be useful when already having a properly initialized font data hand? In addition, some data like lang, country, and variant seem to not be copied by the copy constructors.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/FontData.java
```diff
@@ -186,6 +186,26 @@ public FontData(String string) {
 	}
 }
 
+/**
+ * Constructs a new FontData copy
```
HeikoKlare, line 190:
```suggestion
 * Constructs a deep copy of the given font data object.
```

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/FontData.java
```diff
@@ -186,6 +186,26 @@ public FontData(String string) {
...
+ *
+ * @param fontData the FondData object for which copy is needed to be made
```
HeikoKlare, line 192:
```suggestion
 * @param fontData the FontData object to copy
```

HeikoKlare (changes requested):
Having a proper copy constructor for FontData sounds reasonable to me.
However, the implementations are no proper copy constructors. They seem to be copies of the String-based constructors which is only able to copy the data present in the string, while a proper copy constructor can simply deep-copy all the data from the existing FontData object.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/FontData.java
```diff
@@ -240,6 +240,56 @@ public FontData(String string) {
...
+	this.country = fontData.country;
+	this.variant = fontData.variant;
+	this.data = cloneLogFont(fontData);
+}
+
+private static LOGFONT cloneLogFont(FontData fontData) {
+	if(fontData.data == null) return null;
+
+	LOGFONT newData = new LOGFONT();
+	newData.lfHeight = fontData.data.lfHeight;
```
HeikoKlare: nit: you could store `fontData.data` in a local variable `existingData` or the like to shorten the access

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/FontData.java
```diff
@@ -240,6 +240,56 @@ public FontData(String string) {
...
+ *
+ * @exception IllegalArgumentException
+ * <ul>
+ *    <li>ERROR_NULL_ARGUMENT - if the argument is null</li>
+ * </ul>
+ * @since 3.130
+ */
+public FontData(FontData fontData) {
+	if (fontData == null)
+		SWT.error(SWT.ERROR_NULL_ARGUMENT);
```
HeikoKlare: Please wrap all such expressions in braces

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/FontData.java
```diff
@@ -240,6 +240,56 @@ public FontData(String string) {
...
+
+	this.height = fontData.height;
+	this.lang = fontData.lang;
+	this.country = fontData.country;
+	this.variant = fontData.variant;
+	this.data = cloneLogFont(fontData);
+}
+
+private static LOGFONT cloneLogFont(FontData fontData) {
+	if(fontData.data == null) return null;
```
HeikoKlare: This formatting seems to be wrong

HeikoKlare (commented):
In general, this now looks fine to me. I only have some minor comments regarding code style.

fedejeanne:
@ShahzaibIbrahim can you please bump the appropriate versions and link/document the failing tests?

PR #2137, opened 2025-05-09
Always manage custom tooltip in Tree in Win32 #62

amartya4256:
The PR contributes to the creation of custom tooltip handles in the Tree while creating items. It is a necessity to always have custom tooltips as it needs to be fit inside a single monitor since there can be freezing problems while using per-monitor awareness if the tooltip spans over multiple monitors.

contributes to

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
@@ -211,7 +211,7 @@ void _addListener (int eventType, Listener listener) {
 		case SWT.PaintItem: {
 			customDraw = true;
 			style |= SWT.DOUBLE_BUFFERED;
-			if (isCustomToolTip ()) createItemToolTips ();
+			createItemToolTips ();
 			OS.SendMessage (handle, OS.TVM_SETSCROLLTIME, 0, 0);
 			int bits = OS.GetWindowLong (handle, OS.GWL_STYLE);
 			if (eventType == SWT.MeasureItem) {
@@ -2200,7 +2200,7 @@ void createItem (TreeItem item, long hParent, long hInsertAfter, long hItem) {
 				}
 			}
 		}
-
+		createItemToolTips ();
 		/*
 		 Note: Don't update scrollbars when drawing is disabled.
 		 This gives significant improvement for bulk insert scenarios.
@@ -2799,20 +2799,16 @@ boolean findCell (int x, int y, TreeItem [] item, int [] index, RECT [] cellRect
 			} else {
 				cellRect [0].right = Math.min (cellRect [0].right, rect.right);
 				if (OS.PtInRect (cellRect [0], pt)) {
-					if (isCustomToolTip ()) {
-						int state = (int)OS.SendMessage (handle, OS.TVM_GETITEMSTATE, lpht.hItem, OS.TVIS_SELECTED);
-						int detail = (state & OS.TVIS_SELECTED) != 0 ? SWT.SELECTED : 0;
-						Event event = sendMeasureItemEvent (item [0], order [index [0]], hDC, detail);
-						if (isDisposed () || item [0].isDisposed ()) break;
-						Rectangle boundsInPixels = DPIUtil.scaleUp(event.getBounds(), getZoom());
-						itemRect [0] = new RECT ();
-						itemRect [0].left = boundsInPixels.x;
-						itemRect [0].right = boundsInPixels.x + boundsInPixels.width;
-						itemRect [0].top = boundsInPixels.y;
-						itemRect [0].bottom = boundsInPixels.y + boundsInPixels.height;
-					} else {
-						itemRect [0] = item [0].getBounds (order [index [0]], true, false, false, false, false, hDC);
-					}
+					int state = (int)OS.SendMessage (handle, OS.TVM_GETITEMSTATE, lpht.hItem, OS.TVIS_SELECTED);
+					int detail = (state & OS.TVIS_SELECTED) != 0 ? SWT.SELECTED : 0;
+					Event event = sendMeasureItemEvent (item [0], order [index [0]], hDC, detail);
+					if (isDisposed () || item [0].isDisposed ()) break;
+					Rectangle boundsInPixels = DPIUtil.scaleUp(event.getBounds(), getZoom());
+					itemRect [0] = new RECT ();
+					itemRect [0].left = boundsInPixels.x;
+					itemRect [0].right = boundsInPixels.x + boundsInPixels.width;
+					itemRect [0].top = boundsInPixels.y;
+					itemRect [0].bottom = boundsInPixels.y + boundsInPixels.height;
 					if (itemRect [0].right > cellRect [0].right) found = true;
 					quit = true;
 				}
@@ -3835,10 +3831,6 @@ public int indexOf (TreeItem item) {
 	if (item.isDisposed()) error(SWT.ERROR_INVALID_ARGUMENT);
 	long hItem = OS.SendMessage (handle, OS.TVM_GETNEXTITEM, OS.TVGN_ROOT, 0);
 	return hItem == 0 ? -1 : findIndex (hItem, item.handle);
-}
-
-boolean isCustomToolTip () {
-	return hooks (SWT.MeasureItem);
 }
 
 boolean isItemSelected (NMTVCUSTOMDRAW nmcd) {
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
```diff
@@ -3837,10 +3833,6 @@ public int indexOf (TreeItem item) {
 	return hItem == 0 ? -1 : findIndex (hItem, item.handle);
 }
 
-boolean isCustomToolTip () {
-	return hooks (SWT.MeasureItem);
-}
```
HeikoKlare, line 3842: Can you please explain why this is not relevant anymore (or why it was relevant before)? In my understanding, it validates whether a MeasureItem listener was registered, but I have to admit that I do not understand the reasons. What will be the effect in case there was a tree for which such a listener was not registered and thus tool tips were not custom? I set a breakpoint for the case the `!isCustomToolTip()` and triggered tool tips in several Trees, but it was never triggered.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
```diff
@@ -2200,7 +2200,7 @@ void createItem (TreeItem item, long hParent, long hInsertAfter, long hItem) {
 				}
 			}
 		}
-
+		createItemToolTips ();
```
HeikoKlare, line 2203: Why do we need this additional creation here?

This has some effect that I do no understand: without this addition, I do not face any calls to `isCustomToolTip()` that yields "false", even for the breakpoints view the original issue is about.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
```diff
@@ -5644,16 +5636,7 @@ String toolTipText (NMTTDISPINFO hdr) {
...
 		RECT [] cellRect = new RECT [1], itemRect = new RECT [1];
 		if (findCell (pt.x, pt.y, item, index, cellRect, itemRect)) {
-			String text = null;
-			if (index [0] == 0) {
-				text = item [0].text;
-			} else {
-				String[] strings = item [0].strings;
-				if (strings != null) text = strings [index [0]];
-			}
-			//TEMPORARY CODE
```
HeikoKlare, line 5654: Nice to get temporary removed from > 15 years ago removed :-)

HeikoKlare (commented):
Looks like a nice simplification of the tooltip handling. But I have to admit that I do not understand why it works. Without these changes, `isCustomToolTip()` never yields "false" for me, even for the problematic breakpoints view. Please see the detailed comments for that.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
amartya4256: It was never triggered because of the folowing reasons:
1. For breakpoints view, Eclipse doesn't maintain the tooltipHandle as createItem(TreeItem, long, long, long) doesn't have a call to createItemToolTipHandles(), but only createItem(TreeColumn, int) and _addListener calls for that. For example, in case of outline view, the items are created using createItem(TreeItem, long, long, long) and then _addListener is called to register a measure listener, which also triggers createItemToolTipHandles().
2. There are 2 checks to verify if custom tooltips exist: if itemToolTipHandle field is available and isCustomToolTip is true.
3. In case of breakpoint, since neither tooltip handle is created not a measure event is registered, the tooltips are directly handled by windows, i.e. the painting event of tooltip is indeed recieved by Eclipse but no operation is performed and Windows is responsible to paint and position it.

Hence, Breakpoints view doesn't have any custom tooltips and cannot be manipulated to fit in the monitor. The only way we can make sure all the tooltips are fir in the monitor is by managing all the tooltip handle hence keeping them all custom and then repositioning and resizing them, which is done in this PR.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
amartya4256: The isCustomToolTip call in wmNotifyToolTip is guarded by itemToolTipHandle handle check. If that doesn't exist (which is the case when BReakpoint lets windows handle the tooltips completely and doesn't keep track of the handle) the isCustomToolTip will never be triggered.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
```diff
@@ -211,7 +211,7 @@ void _addListener (int eventType, Listener listener) {
 		case SWT.PaintItem: {
 			customDraw = true;
 			style |= SWT.DOUBLE_BUFFERED;
-			if (isCustomToolTip ()) createItemToolTips ();
+			createItemToolTips ();
```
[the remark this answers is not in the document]
amartya4256: I tried but it seems to misbehave if I create tooltips for example in the constructor before there are any item in the tree.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
```diff
@@ -8225,13 +8207,10 @@ private Rectangle getContainingMonitorBoundsInMultiZoomCoordinateSystem(MonitorA
...
 	switch (nmcd.dwDrawStage) {
 		case OS.CDDS_PREPAINT: {
-			if (isCustomToolTip ()) {
-				//TEMPORARY CODE
-				//nmcd.uDrawFlags |= OS.DT_CALCRECT;
-				//OS.MoveMemory (lParam, nmcd, NMTTCUSTOMDRAW.sizeof);
-				return new LRESULT (OS.CDRF_NOTIFYPOSTPAINT | OS.CDRF_NEWFONT);
-			}
-			break;
+			//TEMPORARY CODE
```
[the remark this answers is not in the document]
amartya4256: would have a look.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
HeikoKlare: My question remains: why was the usage of custom tooltip guarded with the `MeasureItem` hook being present before? I am a bit concerned that in case the hook is missing but we still handle tool items in a custom way, something does not work as expected.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
HeikoKlare: So maybe you can add an explanation of the overall solution: Do we basically need to call `createItemTooltips()` for every item being created now or what's the way it is supposed to work now?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
amartya4256: The only place where MeasureItem is relevant in tooltip is to find the bounds of the cell, in the method sendMeaureItemEvent. The event there however already has the bounds of the current item passed to the method. It is used in findCell, which already has the right reference to the item. So, in case there's no measureItem handler registered, it uses the default size of the cell (Which can be as long as the string, just like in the outline view).


coming back to isCustomToolTip, from my understanding, the usage of isCustomToolTip revolves around how it should be painted. If it is true, OS.CDDS_POSTPAINT event is sent here:
```
LRESULT wmNotifyToolTip (NMTTCUSTOMDRAW nmcd, long lParam) {
	switch (nmcd.dwDrawStage) {
		case OS.CDDS_PREPAINT: {
			if (isCustomToolTip ()) {
				//TEMPORARY CODE
				//nmcd.uDrawFlags |= OS.DT_CALCRECT;
				//OS.MoveMemory (lParam, nmcd, NMTTCUSTOMDRAW.sizeof);
				return new LRESULT (OS.CDRF_NOTIFYPOSTPAINT | OS.CDRF_NEWFONT);
			}
			break;
		}
...
```
 which leads to painting the tooltip manually leading to the same thing painted manually on the tooltip as windows would. But with this we get the freedom to reposition it (which is the main goal to prevent an infinite cycle being triggered)

I did a before and after comparision of the tooltips in the tree in different places where they don't have any tooltTipHandle maintained manually originally. I'll attach the screenshots:
Before: 
image
After:
image
Before:
image
After:
image

I do see a difference in the color of the text though.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
amartya4256: although we can bypass the itemToolTipHandle check when TTM_SHOW message is recieved. It is only recieved for the tooltip of the tree and we shouldn't care if we maintain the handle or not. If we recieve the message, we should try to find the right position of the tooltip and resposition it. If we can do that then we do not need to change the isCustomToolTip logic making the PR minimalistic.

amartya4256:
See https://github.com/eclipse-platform/eclipse.platform.swt/pull/2142 for a better approach.

amartya4256:
closing as taken over by https://github.com/eclipse-platform/eclipse.platform.swt/pull/2142

Issue #2139, opened 2025-05-10
[Win32] Program icon missing in application switch view after primary monitor zoom change


HeikoKlare:
**Describe the bug**
When changing the primary monitor zoom on Windows, the program icon of running Eclipse-based products is not shown anymor in the application switch view (Alt+Tab / Windows+Tab) and in the Window preview (hovering over the task bar entry.

**To Reproduce**
- Start a recent Eclipse SDK at 100% primary monitor zoom (monitor-specific scaling enabled as by default)
- Change the primary monitor zoom to 125%
- Use Alt+Tab or Windows+Tab or hover over the task bar entry of the application to see the missing icon

**Expected behavior**
The icon to be shown also after changing primary monitor zoom.

**Screenshots**
![Image](https://github.com/user-attachments/assets/9c726b2c-cbea-4f62-ad90-3bf66523467c)

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 

2. Additional OS info (e.g. OS version, Linux Desktop, etc)
Windows 10.0.22631 Build 22631

3. JRE/JDK version
Temurin 21.0.6.7

fedejeanne:
Caused by 393309f9be178a4e6cb36048e729ca2a5c075441

PR #2140, opened 2025-05-12
Preserve handles of images when changing zooms #2139

fedejeanne:
Preserve the handles of images created out of existing image handles and also from those which are still needed (according to the method AbstractImageProviderWrapper::getPreservedZoomLevels)

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2139


akoch-yatta (approved):
Looks good to me as well. This handle got lost, when we removed initialNativeZoom - that prevented the deletion before.

PR #2141, opened 2025-05-12
 Removing getDeviceZoom from tests

arunjose696:
DPIUtil.get[Native]DeviceZoom() is no longer meaningful when monitor-specific scaling is used, so its usage has been removed from most tests. In some cases , these have been left unchanged for now.

Among the occurrences mentioned in [Eclipse Platform Issue #195](https://github.com/vi-eclipse/Eclipse-Platform/issues/195), most have been addressed. Still the following four cases remain unmodified:

```
1)org.eclipse.swt.tests.junit.Test_org_eclipse_swt_graphics_GC.executeWithNonDefaultDeviceZoom(Runnable)
2)org.eclipse.swt.tests.junit.DPIUtilTests.setup()
3)org.eclipse.swt.tests.junit.Test_org_eclipse_swt_graphics_Image.test_updateWidthHeightAfterDPIChange()
```
In these three tests, `getDeviceZoom()` is used to capture the zoom level before execution and restore it afterward. No suitable replacement has been identified yet.

```
4)org.eclipse.swt.snippets.Snippet373.main(String[])
```
The snippet description says it is used to to display/refresh Monitor DPI dynamically. However, it uses incorrect DPI access methods and presents a misleading example. It would be better to either remove or rewrite this snippet.

### DIFF
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/DPIUtilTests.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/DPIUtilTests.java
@@ -37,11 +37,12 @@ import org.junit.jupiter.api.condition.DisabledOnOs;
 public class DPIUtilTests {
 
 	private int deviceZoom;
+	private static final int ZOOM_200 = 200;
 
 	@BeforeEach
 	public void setup() {
 		deviceZoom = DPIUtil.getDeviceZoom();
-		DPIUtil.setDeviceZoom(200);
+		DPIUtil.setDeviceZoom(ZOOM_200);
 	}
 
 	@AfterEach
@@ -57,10 +58,10 @@ public class DPIUtilTests {
 
 		float[] scaledValue = DPIUtil.autoScaleDown(valueAt200);
 		assertArrayEquals(valueAt100, scaledValue, .001f, String.format("Scaling down float array from device zoom (%d) to 100 failed",
-				DPIUtil.getDeviceZoom()));
+				ZOOM_200));
 		scaledValue = DPIUtil.autoScaleDown((Device) null, valueAt200);
 		assertArrayEquals(valueAt100, scaledValue, .001f, String.format("Scaling down float array from device zoom (%d) to 100 with device failed",
-				DPIUtil.getDeviceZoom()));
+				ZOOM_200));
 
 		scaledValue = DPIUtil.scaleDown(valueAt200, 200);
 		assertArrayEquals(valueAt100, scaledValue, .001f, "Scaling down float array from 200 failed");
@@ -84,10 +85,10 @@ public class DPIUtilTests {
 
 		int scaledValue = DPIUtil.autoScaleDown(valueAt200);
 		assertEquals(
-				valueAt100, scaledValue, String.format("Scaling down integer from device zoom (%d) to 100 failed", DPIUtil.getDeviceZoom()));
+				valueAt100, scaledValue, String.format("Scaling down integer from device zoom (%d) to 100 failed", ZOOM_200));
 		scaledValue = DPIUtil.autoScaleDown((Device) null, valueAt200);
 		assertEquals(
-				 valueAt100, scaledValue, String.format("Scaling down integer from device zoom (%d) to 100 with device failed", DPIUtil.getDeviceZoom()));
+				 valueAt100, scaledValue, String.format("Scaling down integer from device zoom (%d) to 100 with device failed", ZOOM_200));
 
 		scaledValue = DPIUtil.scaleDown(valueAt200, 200);
 		assertEquals(valueAt100, scaledValue, "Scaling down integer from 200 failed");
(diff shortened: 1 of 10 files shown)


tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/DPIUtilTests.java
```diff
@@ -37,11 +37,12 @@
 public class DPIUtilTests {
 
 	private int deviceZoom;
+	private static final int  ZOOM_200 = 200;
```
HeikoKlare, line 40:
```suggestion
	private static final int ZOOM_200 = 200;
```

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_GC.java
```diff
@@ -154,14 +154,6 @@ public void test_copyAreaIIIIII() {
 	ImageData imageData = image.getImageData();
 	PaletteData palette = imageData.palette;
 
-	if (DPIUtil.getDeviceZoom() != 100) {
-		//TODO Fix non integer scaling factors.
-		if (SwtTestUtil.verbose) {
-			System.out.println("Excluded test_copyAreaIIIIII(org.eclipse.swt.tests.junit.Test_org_eclipse_swt_graphics_GC)");
-		}
-		return;
-	}
```
HeikoKlare, line 163: Are we sure that this can this safely be removed, i.e, will the test run properly on with deviceZoom != 100 on every OS? When I execute this test on Windows with 175% primary monitor zoom, it fails (probably because of fractional scaling rounding errors). So maybe we should rather replace this with an according JUnit assumption but in general keep the check.

HeikoKlare (commented):
In general, the changes look sound to me. The removal of the assumptions for deviceZoom being 100 in some of the tests seems not okay to me, as the tests fail on specific zooms, but maybe the assumptions themselves can be improved.

@akoch-yatta would be nice if you could still have another look to ensure that the changes are sound and that we do not degrade the tests

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_GC.java
arunjose696: I tested this on Windows 11 with various zoom levels (100%, 125%, 150%, 175%, 200%, and 225%) and observed that the tests passed in all cases, which is why I removed the check.
Could you share the exact error you're encountering at 175%—is it an assertion failure?
Also, are you using any specific arguments in your run configuration?

If we replace the check with a JUnit assumption, don’t we still need to call DPIUtil.getDeviceZoom()? Or is there another way to get the primary monitor’s zoom level that works across platforms?

HeikoKlare: I just found that I cannot reproduce the issue with your exact branch, but when rebasing it on top of the lastest master branch state. I guess we made some adaptations since you last rebased this branch on master that affect these tests.

When executing the tests at 175% primary monitor zoom on that state, I get this error:
```
java.lang.AssertionError: :b: expected:<RGB {0, 0, 255}> but was:<RGB {255, 255, 255}>
	at org.junit.Assert.failNotEquals(Assert.java:835)
	at org.junit.Assert.assertEquals(Assert.java:120)
	at org.eclipse.swt.tests.junit.Test_org_eclipse_swt_graphics_GC.test_copyAreaIIIIII(Test_org_eclipse_swt_graphics_GC.java:160)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
```

> If we replace the check with a JUnit assumption, don’t we still need to call DPIUtil.getDeviceZoom()? Or is there another way to get the primary monitor’s zoom level that works across platforms?

Yes, we would then still need to use `DPIUtil.getDeviceZoom()`. But our primary goal it to reevaluate all existing usages whether they can be replaced or whether they are fine and we may consider it fine to be used here as we are not in the situation that the returned value is useless because of potentially having multiple shells placed at different monitors with different zoom values on Windows (which is actually the use case in which the usage of `get[Native]DeviceZoom()` fails).

arunjose696: >  we may consider it fine to be used here as we are not in the situation that the returned value is useless because of potentially having multiple shells placed at different monitors with different zoom values on Windows (which is actually the use case in which the usage of `get[Native]DeviceZoom()` fails).

Yeah makes sense, I will modify this to junit assumptions

arunjose696: Replaced the checks with junit assumption

Issue #557, opened 2023-02-22
Win: Eclipse IDE freeze with Hover across screens with different UI scaling


jukzi:
**Describe the bug**
Eclipse IDE is hanging for ever if a long Hover Text is shown across multiple screens,
100% reproducible on some eclipse installations, but does not occur on other eclipse installations with same eclipse version on same computer - i could not find the difference.

**To Reproduce**
Use windows 10.
Have 2 Screens. Left configured to 100% and right configured to 125% UI Scaling in Windows setting.

Having a Class
```
class Long {
    public void veryLongMethodName1veryLongMethodName2veryLongMethodName3veryLongMethodName4() {
    }
}
```
And Show the Outline, Hover the method name with the mouse.

![image](https://user-images.githubusercontent.com/51790620/220602355-3131dfa9-7cb6-48b7-9369-73716bf479b0.png)
The freeze ends as soon as the second display is changed back to 100%.

Sampling shows, that the CPU is busy and debugging shows Windows OS sends Events again and again and SWT is doing some things, but the hover text is never shown.
![image](https://user-images.githubusercontent.com/51790620/220601154-1841bd5d-872d-4eaa-954a-629e3257bea8.png)

**Expected behavior**
No freeze.

 
**Version since**
First seen in Eclipse 4.26

**Workaround**
Use 100% UI Scaling on all Monitors.

Ideas welcome. It feels like Windows/SWT is not sure which scaling to use and is hung in an endless resizing loop.

jukzi:
Looks like a Microsoft Windows OS bug:
According to https://learn.microsoft.com/de-de/windows/win32/controls/ttn-show "SetWindowPos" should be called when TTN_SHOW is received. And that's what SWT does in org.eclipse.swt.widgets.Tree.wmNotifyToolTip(NMHDR, long, long). When i replace SetWindowPos  with "return null" while SWT is in that endless loop, then the quickinfo text is not shown and the endless "windowProc" events stop.

SyntevoAlex:
I would suspect that SWT calls some kind of `redraw()`, or maybe it sets different size with `SetWindowPos()`, which causes Tooltip to endlessly repaint.

jukzi:
SetWindowPos always gets the same coordinates. Where could i see some kind of redraw()?

SyntevoAlex:
[...]

PR #2142, opened 2025-05-12
Fit tooltip of the tree on a single monitor #62

amartya4256:
This PR contributes to handling wmNotify event for tooltip of a Tree regardless of if the Tree object maintains the handle of the tooltip in itemToolTipHandle. If there's a tooltip which is not handled by wmNotifyToolTip, it is completely maintained by windows and hence can span over multiple montiors. Hence, with this PR such tooltips are also positioned inside a single monitor to prevent any infinite loop.

Here's the explanation:

1. For a tree where the tooltip is not maintained in itemToolTipHandle (Since createToolTips was not called -> no createItem(TreeItem, ...) calls or _addListener() call with SWT.Paint or MeasureItem), wmNotifyToolTip cannot handle its drawing and hence it is completely drawn and positioned by windows, which can be risky -> can lead to inifinite loop if the tooltip spans over multiple monitors with different zooms.
2. ```
        else if (hdr.code == OS.TTN_SHOW) {
		return positionTooltip(hdr, wParam, lParam, false);
	}
	``` 
	This block in wmNotify makes sure for such tooltips that all the positioning of any tooltip within a tree is handled by SWT. Hence, it calls for the same positioning logic as of wmNotifyToolTip
3. According to the old implementation of tooltips, inside positionTooltip method, findCell is called. Inside findCell, those tooltips for which isCustomToolTip is true (they have been registered with a measureItem hook), the toolTip is believed to contain the image and the text while the other case by default doesn't contain the image. It means on the size calculation of a tooltip of a treeItem which has an image but doesnt have a measureItem, would not have the size enough to contain both image and text. Just to make this behaviour consistent, in both the cases we are considering the image size if it available within the treeItem. And hence, it can be used by those toolTips which aren't maintained by Tree in SWT as well. (Line 2814)

Contributes to

Replaces and thus closes #2142

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
@@ -2811,7 +2811,7 @@ boolean findCell (int x, int y, TreeItem [] item, int [] index, RECT [] cellRect
 						itemRect [0].top = boundsInPixels.y;
 						itemRect [0].bottom = boundsInPixels.y + boundsInPixels.height;
 					} else {
-						itemRect [0] = item [0].getBounds (order [index [0]], true, false, false, false, false, hDC);
+						itemRect [0] = item [0].getBounds (order [index [0]], true, true, false, false, false, hDC);
 					}
 					if (itemRect [0].right > cellRect [0].right) found = true;
 					quit = true;
@@ -7397,6 +7397,8 @@ LRESULT wmNotify (NMHDR hdr, long wParam, long lParam) {
 	if (hdr.hwndFrom == itemToolTipHandle && itemToolTipHandle != 0) {
 		LRESULT result = wmNotifyToolTip (hdr, wParam, lParam);
 		if (result != null) return result;
+	} else if (hdr.code == OS.TTN_SHOW) {
+		return positionTooltip(hdr, wParam, lParam, false);
 	}
 	if (hdr.hwndFrom == hwndHeader && hwndHeader != 0) {
 		LRESULT result = wmNotifyHeader (hdr, wParam, lParam);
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
```diff
@@ -8145,44 +8147,55 @@ LRESULT wmNotifyToolTip (NMHDR hdr, long wParam, long lParam) {
...
+		// triggers additional display messages to SWT, creating an infinite loop
+		// of positioning and re-scaling events.
+		// Refer: https://github.com/eclipse-platform/eclipse.platform.swt/issues/557
+		Point cursorLocation = display.getCursorLocation();
+		Rectangle monitorBounds = cursorLocation instanceof MonitorAwarePoint monitorAwarePoint
+				? getContainingMonitorBoundsInMultiZoomCoordinateSystem(monitorAwarePoint)
+				: getContainingMonitorBoundsInSingleZoomCoordinateSystem(cursorLocation);
+		if (monitorBounds != null) {
+			Rectangle adjustedTooltipBounds = fitTooltipBoundsIntoMonitor(toolRect, monitorBounds);
+			System.out.println(adjustedTooltipBounds.x + " " + toolRect.left);
```
fedejeanne, line 8184:
```suggestion
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
```diff
@@ -8145,44 +8147,55 @@ LRESULT wmNotifyToolTip (NMHDR hdr, long wParam, long lParam) {
...
+			OS.SetWindowPos (hdr.hwndFrom, 0, adjustedTooltipBounds.x, adjustedTooltipBounds.y, adjustedTooltipBounds.width, adjustedTooltipBounds.height, flags);
+			return LRESULT.ONE;
+		}
+	} else {
+		// If managedTooltip is false and the cursor is not over the valid part of the
+		// target cell, Windows may still try to display the default tooltip. Since we
+		// can't prevent it from showing at this point, we set its bounds to zero to
+		// effectively hide it.
+		OS.SetWindowPos (hdr.hwndFrom, 0, 0, 0, 0, 0, flags);
+		return LRESULT.ONE;
```
fedejeanne, line 8194:
```suggestion
		result = LRESULT.ONE;
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
```diff
@@ -8145,44 +8147,55 @@ LRESULT wmNotifyToolTip (NMHDR hdr, long wParam, long lParam) {
...
+		// Refer: https://github.com/eclipse-platform/eclipse.platform.swt/issues/557
+		Point cursorLocation = display.getCursorLocation();
+		Rectangle monitorBounds = cursorLocation instanceof MonitorAwarePoint monitorAwarePoint
+				? getContainingMonitorBoundsInMultiZoomCoordinateSystem(monitorAwarePoint)
+				: getContainingMonitorBoundsInSingleZoomCoordinateSystem(cursorLocation);
+		if (monitorBounds != null) {
+			Rectangle adjustedTooltipBounds = fitTooltipBoundsIntoMonitor(toolRect, monitorBounds);
+			System.out.println(adjustedTooltipBounds.x + " " + toolRect.left);
+			OS.SetWindowPos (hdr.hwndFrom, 0, adjustedTooltipBounds.x, adjustedTooltipBounds.y, adjustedTooltipBounds.width, adjustedTooltipBounds.height, flags);
+			return LRESULT.ONE;
```
fedejeanne, line 8186:
```suggestion
			result = LRESULT.ONE;
```

fedejeanne (commented):
Only small comments but it looks good. @HeikoKlare once the changes are done then I consider this PR approved ✔️ 

Thx @amartya4256 !

HeikoKlare (commented):
Thank you for the proposal and the nice explanation of how it works.

When testing, I made some observation:
1. When a different window has focus while I hover over an element having such a tooltip (like in the breakpoints view), it has a wrong size and shows some artifacts. However, I found that we have the same behavior already in the current master state:
image
2. To see the tooltip, you need to move the mouse from somewhere outside of the element directly onto the text. If you first hover over the checkbox and then to the text, the tooltip will not show up. That's a slight derivation from existing behavior. Is there a way to fix that?
tooltips_hover_path

Finally, the issues linked in the commit seem to be inappropriate. I think the fitting issue is this, isn't it?
- https://github.com/eclipse-platform/eclipse.platform.swt/issues/557

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
amartya4256: Done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
amartya4256: Done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
amartya4256: Done

amartya4256:
> Thank you for the proposal and the nice explanation of how it works.
> 
> When testing, I made some observation:
> 
> 1. When a different window has focus while I hover over an element having such a tooltip (like in the breakpoints view), it has a wrong size and shows some artifacts. However, I found that we have the same behavior already in the current master state:
>    image
> 2. To see the tooltip, you need to move the mouse from somewhere outside of the element directly onto the text. If you first hover over the checkbox and then to the text, the tooltip will not show up. That's a slight derivation from existing behavior. Is there a way to fix that?
>    tooltips_hover_path
>          
>            [
>              
>            
>                tooltips_hover_path
>              ](https://private-user-images.githubusercontent.com/755472/443238945-96c33098-64bd-4251-9da6-9334ac33d6bb.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NDcxNTM2NzIsIm5iZiI6MTc0NzE1MzM3MiwicGF0aCI6Ii83NTU0NzIvNDQzMjM4OTQ1LTk2YzMzMDk4LTY0YmQtNDI1MS05ZGE2LTkzMzRhYzMzZDZiYi5naWY_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjUwNTEzJTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI1MDUxM1QxNjIyNTJaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT1mZGQ3Y2ZiODYzYWVjM2I5MGNjZWFmMjhhNWM5YTI2YTk2NTFkMDEyNmU5NGIxNDdhMzgwN2QzMTk0YWQ1NzQ1JlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCJ9.CPqFZsKIZJwpkrbnMsgr7p3Tx6oEiVPsc-SneeQmw7k)
>            
>            
>              
>                
>                  
>                
>                
>                  
>                  
>                
>              
>              [
>                
>                  
>                
>              ](https://private-user-images.githubusercontent.com/755472/443238945-96c33098-64bd-4251-9da6-9334ac33d6bb.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NDcxNTM2NzIsIm5iZiI6MTc0NzE1MzM3MiwicGF0aCI6Ii83NTU0NzIvNDQzMjM4OTQ1LTk2YzMzMDk4LTY0YmQtNDI1MS05ZGE2LTkzMzRhYzMzZDZiYi5naWY_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjUwNTEzJTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI1MDUxM1QxNjIyNTJaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT1mZGQ3Y2ZiODYzYWVjM2I5MGNjZWFmMjhhNWM5YTI2YTk2NTFkMDEyNmU5NGIxNDdhMzgwN2QzMTk0YWQ1NzQ1JlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCJ9.CPqFZsKIZJwpkrbnMsgr7p3Tx6oEiVPsc-SneeQmw7k)
> 
> Finally, the issues linked in the commit seem to be inappropriate. I think the fitting issue is this, isn't it?
> 
> * [Win: Eclipse IDE freeze with Hover across screens with different UI scaling #557](https://github.com/eclipse-platform/eclipse.platform.swt/issues/557)

For the second point, it doesn't happen because when entering the toolItem with the mouse, we only get one callback from windows, and if it is at the checkmark (in case of unmanaged tooltip), we keep the tooltip hidden and hence when you move to the right side a bit, there's no other callback recieved until you enter the tooltip again. This can be probably fixed by mouse tracking for tree, just like we did in edge tooltips, but I'd propose to do that in a separate ticket. And then we can already merge this. The follow up would be an improvement.

For the first point, it is an existent behavior so I didnt have a look into it.

amartya4256:
I think there should be a check here so that we don't affect anything for custom tooltips. 
This else should only execute when `managedTooltip == false`. I'll make the change.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
```diff
@@ -8145,44 +8147,54 @@ LRESULT wmNotifyToolTip (NMHDR hdr, long wParam, long lParam) {
...
+		Rectangle monitorBounds = cursorLocation instanceof MonitorAwarePoint monitorAwarePoint
+				? getContainingMonitorBoundsInMultiZoomCoordinateSystem(monitorAwarePoint)
+				: getContainingMonitorBoundsInSingleZoomCoordinateSystem(cursorLocation);
+		if (monitorBounds != null) {
+			Rectangle adjustedTooltipBounds = fitTooltipBoundsIntoMonitor(toolRect, monitorBounds);
+			OS.SetWindowPos (hdr.hwndFrom, 0, adjustedTooltipBounds.x, adjustedTooltipBounds.y, adjustedTooltipBounds.width, adjustedTooltipBounds.height, flags);
+			result = LRESULT.ONE;
+		}
+	} else {
+		// If managedTooltip is false and the cursor is not over the valid part of the
```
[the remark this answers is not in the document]
amartya4256: I think there should be a check here so that we don't affect anything for custom tooltips. 
This else should only execute when `managedTooltip == false`. I'll make the change.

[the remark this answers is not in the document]
amartya4256: Although this has no effect on custom tooltips visually, it's better to have that check.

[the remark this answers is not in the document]
fedejeanne: I tested and I can confirm that the extra check [`if (!managedTooltip)`](https://github.com/eclipse-platform/eclipse.platform.swt/compare/4593ef63276d51a400e6c110419012b00425896b..08f99b76ff318bd8dcd4599ccd0aee901ce0a8c5) doesn't change the functionality. Furthermore, the whole `else` block has been added by this PR, which also adds the call to the method `positionTooltip(..., false)` meaning that this extra check further ensures that the previous use case (which is now the call to `positionTooltip(..., true);` in line **8150**) does not execute.

All in all, I think this change is OK.

fedejeanne (dismissed):
Thank you @amartya4256 !

This PR solves the critical UI freeze problem at a very low cost of not seeing the popup when first hovering on the checkbox and only later on the name of the method (See point Nr. 2 in https://github.com/eclipse-platform/eclipse.platform.swt/pull/2142#pullrequestreview-2836875078).

I accept that trade and propose to look at the introduced behavior in a follow-up PR.

@HeikoKlare if there are no objections on your side, I'll merge this one today.

merks:
FYI, there appear to be server problems:

https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/6135

iloveeclipse:
This PR seem to cause a regression, see https://github.com/eclipse-platform/eclipse.platform.swt/issues/3108.
Please check.

Issue #501, opened 2022-12-14
[Windows 11] MenuItem with style SWT.CHECK and image doesn't show state


Phillipus:
Windows 11
Eclipse 4.26

Run this Snippet on Windows 10 and Window 11 and compare the drop down menu after selecting it:

```java
import org.eclipse.swt.SWT;
import org.eclipse.swt.graphics.Image;
import org.eclipse.swt.widgets.Display;
import org.eclipse.swt.widgets.Menu;
import org.eclipse.swt.widgets.MenuItem;
import org.eclipse.swt.widgets.Shell;

public class MenuImageTest {
    
	public static void main(String[] args) {
		final Display display = new Display();
		Shell shell = new Shell(display);
		shell.setText("Menu Image Test");
		
		Menu appMenuBar = display.getMenuBar();
		if(appMenuBar == null) {
			appMenuBar = new Menu(shell, SWT.BAR);
			shell.setMenuBar(appMenuBar);
		}
		
		MenuItem file = new MenuItem(appMenuBar, SWT.CASCADE);
		file.setText("File");
		
		Menu dropdown = new Menu(appMenuBar);
		file.setMenu(dropdown);
		
		MenuItem testMenu = new MenuItem(dropdown, SWT.CHECK);
		testMenu.setImage(new Image(display, MenuImageTest.class.getResourceAsStream("image.png")));
		testMenu.setText("Test Menu");
		
		shell.open();
		
		while (!shell.isDisposed()) {
			if (!display.readAndDispatch())
				display.sleep();
		}
		display.dispose();
	}
}
```

Use an image for `image.png` such as:

![image](https://user-images.githubusercontent.com/600504/207722001-9d80d78b-f518-4e56-a5bf-3895e56a2379.png)

On Windows 10 selected state looks like this:

![menu](https://user-images.githubusercontent.com/600504/207722071-ed4baff9-5f2f-4f4d-976c-757d4f89f25d.png)

In Windows 11 selected state looks like this:

![menu 2](https://user-images.githubusercontent.com/600504/207722131-70543bba-e082-4b43-87e7-5f15b3e10a48.png)

SyntevoAlex:
![image](https://user-images.githubusercontent.com/45797751/209442325-df426db9-535c-4c9d-83bf-5eb48aa159d2.png)

SyntevoAlex:
The screenshot above shows a difference between
* `SWT.CHECK` menu with `.setImage()`
[...]

PR #2143, opened 2025-05-13
[win32] Add workaround for missing selection indicator for menu item with image on Win11 (#501)

wahlbrink:
For menu items of type CHECK and RADIO with an image, it create an second image with a checkmark/circle over the original image.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MenuItem.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MenuItem.java
@@ -19,6 +19,7 @@ import org.eclipse.swt.events.*;
 import org.eclipse.swt.graphics.*;
 import org.eclipse.swt.internal.*;
 import org.eclipse.swt.internal.win32.*;
+import org.eclipse.swt.internal.win32.version.*;
 
 /**
  * Instances of this class represent a selectable user interface object
@@ -42,6 +43,8 @@ import org.eclipse.swt.internal.win32.*;
 public class MenuItem extends Item {
 	Menu parent, menu;
 	long hBitmap;
+	Image imageSelected;
+	long hBitmapSelected;
 	int id, accelerator, userId;
 	ToolTip itemToolTip;
 	/* Image margin. */
@@ -53,6 +56,11 @@ public class MenuItem extends Item {
 	// value in wmMeasureChild is increased by a fixed value (in points) when wmDrawChild is called
 	// This static is used to mitigate this increase
 	private final static int WINDOWS_OVERHEAD = 6;
+	// Workaround for: selection indicator is missing for menu item with image on Win11 (#501)
+	// 0= off/system behavior; 1= no image if selected; 2= with overlay marker (default)
+	private final static int CUSTOM_SELECTION_IMAGE = (OsVersion.IS_WIN11_21H2) ?
+			Integer.getInteger("org.eclipse.swt.internal.win32.menu.customSelectionImage", 2) : 0;
+
 	static {
 		DPIZoomChangeRegistry.registerHandler(MenuItem::handleDPIChange, MenuItem.class);
 	}
@@ -543,6 +551,12 @@ void releaseWidget () {
 	super.releaseWidget ();
 	if (hBitmap != 0) OS.DeleteObject (hBitmap);
 	hBitmap = 0;
+	if (hBitmapSelected != 0) OS.DeleteObject (hBitmapSelected);
+	hBitmapSelected = 0;
+	if (imageSelected != null) {
+		imageSelected.dispose();
+		imageSelected = null;
+	}
 	if (accelerator != 0) {
 		parent.destroyAccelerators ();
 	}
@@ -774,6 +788,18 @@ public void setImage (Image image) {
 	if (this.image == image) return;
 	if ((style & SWT.SEPARATOR) != 0) return;
 	super.setImage (image);
+	if (imageSelected != null) {
+		imageSelected.dispose();
+		imageSelected = null;
+	}
+	if ((style & (SWT.CHECK | SWT.RADIO)) != 0 && CUSTOM_SELECTION_IMAGE > 1
+			&& image != null && getSelection()) {
+		initCustomSelectedImage();
+	}
+	updateImage();
+}
+
+private void updateImage () {
 	MENUITEMINFO info = new MENUITEMINFO ();
 	info.cbSize = MENUITEMINFO.sizeof;
 	info.fMask = OS.MIIM_BITMAP;
@@ -781,7 +807,15 @@ public void setImage (Image image) {
 		info.hbmpItem = OS.HBMMENU_CALLBACK;
 	} else {
 		if (OS.IsAppThemed ()) {
-			info.hbmpItem = hBitmap = getMenuItemIconBitmapHandle(image);
+			hBitmap = getMenuItemIconBitmapHandle(image);
+			if ((style & (SWT.CHECK | SWT.RADIO)) != 0 && CUSTOM_SELECTION_IMAGE > 0) {
+				info.fMask |= OS.MIIM_CHECKMARKS;
+				info.hbmpUnchecked = hBitmap;
+				info.hbmpChecked = getMenuItemIconSelectedBitmapHandle();
+			}
+			else {
+				info.hbmpItem = hBitmap;
+			}
 		} else {
 			info.hbmpItem = image != null ? OS.HBMMENU_CALLBACK : 0;
 		}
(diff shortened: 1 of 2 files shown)


wahlbrink:
ControlExample
IDE-noTheme
IDE-DarkTheme

merks:
@wahlbrink 

This problem is the only reason why I have not upgraded by system to Windows 11.

Can you show a screen capture of how it looks with and without your changes?  I'm not sure I can really test/see that on Windows 10.

@HeikoKlare 

This problem has festered for years and I don't think Windows 11 itself will ever make the problem go away.  It would be great for this to be reviewed to make any necessary changes or improvements.

Phillipus:
What _should_ happen is that the menu item image is highlighted by a surrounding square (blue in light theme). Wasn't this possible as POC in https://github.com/eclipse-platform/eclipse.platform.swt/issues/501?

merks:
> What _should_ happen is that the menu item image is highlighted by a surrounding square (blue in light theme). Wasn't this possible as POC in #501?

The _should_ happen is based on how it looked before, which was quite (very) nice.   In fact kind of nicer visually than this proposal and more familiar.  But, at this point 6 months later, I'm willing to say what _should_ happen is **something**, maybe even **anything** reasonable, because that is better than nothing at all which we have currently and have had for far too long.

I really hope to see movement here...

Phillipus:
@merks Agreed. I just thought that the highlighted square was kind of working as POC in https://github.com/eclipse-platform/eclipse.platform.swt/issues/501 ?

merks:
> @merks Agreed. I just thought that the highlighted square was kind of working as POC in #501 ?

Yes, I think that approach is prettier. 😍  I just don't know if it's harder to implement and maybe harder to match the native styling.  I don't know much, except that I'm willing to compromise at this point...

HeikoKlare:
Thank you for this proposal!
I agree that any solution to the issue is better than none. And given that this proposes a way to custom-draw the icon to be shown, it might be easy to just exchange _how_ the icon is drawn later on if desired (replacing the check with a highlighted square in the background or whatever we agree on).
In the long run, we should probably try to mimic the highlighted square. Even though the check mark is what the more modern UI frameworks in Windows use, the highlighted square fits to every other visualization of selection we have in Eclipse right now.

At first glance, the change looks quite straightforward and should hopefully be of low risk. Still the question is if we want to process this for the upcoming release given that we are short before M3 (and at least I will not have time to have a more thorough look into this until Thursday).

merks:
> What _should_ happen is that the menu item image is highlighted by a surrounding square (blue in light theme). Wasn't this possible as POC in #501?

I scanned that very long issue and I'm not sure exactly where the POC **code** actually exists.  So while I like visually what I see now on Windows 10, if no one can spend the effort to emulate that (which seems tricky with respect to theme adjustments) then I like that this PR provides a working solution that we can actually commit.   

In my opinion, we need to move forward one way or the other before M1.  So if someone has a better way and is willing to create a PR for it, that would be a good thing.

HeikoKlare:
> In my opinion, we need to move forward one way or the other before M1. So if someone has a better way and is willing to create a PR for it, that would be a good thing.

Fully agree. From UX perspective the solution looks quite fine and I don't expect that anyone proposes a better one soon.
I still have it on my list to take a look at this from code perspective for M1 and will do as soon as I find the time so that we have it early in M1.

Phillipus:
> > What _should_ happen is that the menu item image is highlighted by a surrounding square (blue in light theme). Wasn't this possible as POC in #501?
> 
> I scanned that very long issue and I'm not sure exactly where the POC **code** actually exists.

No code but @wahlbrink proposed possible solution "(A) Mimic Win 10 / Add blue box (as proposed)" [here](https://github.com/eclipse-platform/eclipse.platform.swt/issues/501#issuecomment-2408643301) - is that possible or a no go?

merks:
I think anything that changes the size needed to render the decoration is disruptive and will likely lead to visual inconsistencies.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MenuItem.java
```diff
@@ -791,16 +825,84 @@ public void setImage (Image image) {
...
+		}
+		@Override
+		public void drawOn(GC gc, int imageWidth, int imageHeight) {
+			gc.setAdvanced(true);
+			gc.drawImage(image, imageWidth - imageBounds.width, (imageHeight - imageBounds.height) / 2);
+			int x0 = imageWidth - 16;
+			int y0 = imageHeight / 2 - 8;
+			if (((style & SWT.CHECK) != 0)) {
+				int[] points = new int[] { x0 + 4, y0 + 10, x0 + 6, y0 + 12, x0 + 12, y0 + 6 };
+				gc.setAntialias(SWT.ON);
```
HeikoKlare: nit: alias is set in if and else clause, so it could be moved outside

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MenuItem.java
```diff
@@ -791,16 +825,84 @@ public void setImage (Image image) {
...
+		public int getGcStyle() {
+			return SWT.TRANSPARENT;
+		}
+		@Override
+		public void drawOn(GC gc, int imageWidth, int imageHeight) {
+			gc.setAdvanced(true);
+			gc.drawImage(image, imageWidth - imageBounds.width, (imageHeight - imageBounds.height) / 2);
+			int x0 = imageWidth - 16;
+			int y0 = imageHeight / 2 - 8;
+			if (((style & SWT.CHECK) != 0)) {
```
HeikoKlare: nit: to make the method a bit simpler and code a bit more comprehensive, you might extract the if and else blocks into methods `drawCheck` and `drawRadio` or the like

HeikoKlare (commented):
Thank you for this concrete proposal on addressing the long-standing issue with missing selection indicators in menu items!

The implementation seems to work fine. I have tested it also with different monitor zooms and it adapts as expected. Also the different configuration options work as designed. I did not find any functional issues within the code as well. So regarding all that, the PR is fine to me.

I am just a bit concerned that the implementation is scattered across the `MenuItem` implementation. This makes it difficult to understand even in the review and will become even more difficult when reading the code fragment without the context of a PR.

I don't want to bother with rewriting the adaptation, but in my opinion encapsulating all the custom selection image functionality at a single place might significantly improve the quality. It might be an (inner) `Selection` class (or the like) encapsulating:
- the toggle state for which kind of customization to use based on the system propery 
- the initialization logic for the custom selected image
- the custom selected image (currently the `imageSelected` field), which can be initialized on demand
- the selected bitmap handle (currently the `hBitmapSelected` field), which can also be initialized on demand
- the dispose functionality for all data related to the selection image (to be called when setting a new image on the menu item or releasing the menu item widget)
- the functionality to fill the menu item info (called in `updateImage()` and `setSelection()`)

This would serve two means:
1. Locality: the functionality would be encapsulated at a single place so it can be understood, maintained and, if necessary, removed at that single place.
2. Comprehensibility: as a side effect of putting all partial functionalities into methods of that encapsulating class, the purpose of the single code blocks becomes more clear (just like when extracting the code blocks into methods with self-explaining names).

But to point it out again: even without such a design refinement the change is fine and it will definitely improve UX compared to the given state by showing selected states again anyway.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MenuItem.java
merks: I force pushed a rebased version with this suggestion addressed.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MenuItem.java
merks: I force pushed a rebased version with this suggestion addressed.

merks:
@HeikoKlare 

Unless someone creates a PR for a better fix for the problem I think we should move forward and fix the problem somehow.  It can always be improved later if someone with the right skills find time and motivation...

merks:
With the changes in https://github.com/eclipse-platform/eclipse.platform.swt/pull/2314 it looks like I'll need to update this again.  😢

merks:
Rebased yet again...

merks:
@iloveeclipse 

I assume you have no concerns, but I thought I'd ask before merging this.

iloveeclipse:
> @iloveeclipse
> 
> I assume you have no concerns, but I thought I'd ask before merging this.

Thanks for asking. I know the problem, it is annoying one, but fortunately I only seldom work with Windows...

No concerns from my side, just a question: if I understood it right, the workaround works only for menu items with icons? If so, wouldn't be possible to extend the workaround to menu items without icons? May be in a next PR?

merks:
FYI, menus without icons work just fine.

image

This PR is effectively drawing those "decorations" on top of the image.

For Windows 10 it looks like this:

image

Which is nicer, but it's decorating an area outside the bounds of the image and that's (apparently) not achievable, hence the comprise to draw the decorations on top (and in a way that works in light or dark mode on light or dark images).

iloveeclipse:
> FYI, menus without icons work just fine.

Good, haven't used Eclipse on Windows for longer time...

BeckerWdf:
So should he have a New&Noteworthy entry for this change?

iloveeclipse:
> So should he have a New&Noteworthy entry for this change?

Yes please. 

And thanks everyone worked on that workaround!

HeikoKlare:
> > So should he have a New&Noteworthy entry for this change?
> 
> Yes please.

I just made a proposal:
- https://github.com/eclipse-platform/www.eclipse.org-eclipse/pull/387

Feel free to comment or propose an alternative news entry.

PR #2144, opened 2025-05-13
Edge: Fix URI_FOR_CUSTOM_TEXT_PAGE in case of case mismatch

### LINKED ISSUE #2061: [Edge] Eclipse muzz never hung on displaying Javadoc tooltip with WebView2 installed
**Describe the bug**
After installing WebView2 placing mouse over any tooltip-help-equipped term in (Java) editor causes IDE hung forever.

**To Reproduce**
On M$-Edgeless system with Firefox as system-wide default browser ensure default behavior of Eclipse Javadoc help - in tooltips('Press F2 to focus' in bottom bar).
Install M$ Edge WebView2 Runtime (https://developer.microsoft.com/en-us/microsoft-edge/webview2/)
Restart Eclipse
Check Help like Javadoc wasn't spontaneously switched to external browser. Return it back to internal tooltips, if so. Ensure General / Web browser / Use internal web browser is selected.
Place mouse pointer within editor over a Java term initiating Javadoc tooltip - Eclipse UI stops to respond (forever) and "(Not Responding)" suffix appears in window title after short period of time.

**Expected behavior**
The Javadoc tooltips work EXACTLY same way before WebView2 installation - just displayed with no hungs.

**Environment:**
Windows

Windows 11

3. JRE/JDK version
Oracle JDK 24.GA
 
Wipe out WebView2 - this brings the annoying #2000 popups back, but Javadoc tooltips became functional too

sratz:
Use Path.equals() instead of URI.equals() to compare the URI paths.

This will will handle upper/lower-case differences correctly which can arise, e.g. when symlinks are involved.

Fixes #2061.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
@@ -1534,8 +1534,10 @@ public void stop() {
 
 static boolean isLocationForCustomText(String location) {
 		try {
-			return URI_FOR_CUSTOM_TEXT_PAGE.equals(new URI(location));
-		} catch (URISyntaxException e) {
+			URI locationUri = new URI(location);
+			return "file".equals(locationUri.getScheme())
+					&& Path.of(URI_FOR_CUSTOM_TEXT_PAGE).equals(Path.of(locationUri));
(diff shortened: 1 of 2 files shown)


HeikoKlare (approved):
The change looks sound. Thank you also for providing the regression tests!

I guess this can be merged, @sratz? Would be nice to bring it in today for some testing I-Builds before M3.

PR #2145, opened 2025-05-13
[win32] Allow disablement of autoscale per Widget

akoch-yatta:
This PR introduces the data attribute AUTOSCALE_DISABLED for the windows implementation for Widgets. This will disable autoscaling for this Widget and all of its children. This is a necessary requirement for complex usages utilizing GC. Without disabling autoscaling rendering artifacts will be introduced when only parts of the GC are redrawn on monitor specific scaling with some zoom value, e.g. 175%. This commit serves as starting point to harden this feature before considering extracting an API out of it.


HeikoKlare (approved):
This is a purely internal change with no effects (except for setting some data on widgets) in case the auto-scale disablement is not used. Thus it's safe to do, but for the sake of some testing we integrate it now for M3.

The described purpose is to properly support non-auto-scaled widgets, which are supposed to be purely custom drawn on a pixel basis instead of the auto-scaled SWT points (with all their rounding errors). One prominent example are diagrams as implemented by GEF (https://github.com/eclipse-gef/gef-classic).
We will follow up with a snippet to try out the behavior independent from GEF as a follow up.

PR #2149, opened 2025-05-13
Fail to get text from the clipboard when text is Latin-1 encoded but transfer data type is UTF8_STRING

josediazfer:
It's not possible to paste text data with accented characters (CP-1252) from a Windows VNC connection to a Linux server. Pasting text into other desktop applications works just like gedit. In the code to be merged, each type is attempted (in the order UTF8_STRING, COMPOUND_TEXT, STRING) until the text can be retrieved from the clipboard.

In the gtkclipboard.c file (request_text_received_func function) of the GTK3 native library it does something similar:

```
static void
request_text_received_func (GtkClipboard     *clipboard,
                            GtkSelectionData *selection_data,
                            gpointer          data)
{
  RequestTextInfo *info = data;
  gchar *result = NULL;

  result = (gchar *) gtk_selection_data_get_text (selection_data);

  if (!result)
    {
      /* If we asked for UTF8 and didn't get it, try compound_text;
       * if we asked for compound_text and didn't get it, try string;
       * If we asked for anything else and didn't get it, give up.
       */
      GdkAtom target = gtk_selection_data_get_target (selection_data);

      if (target == gdk_atom_intern_static_string ("text/plain;charset=utf-8"))
        {
          gtk_clipboard_request_contents (clipboard,
                                          gdk_atom_intern_static_string ("UTF8_STRING"),
                                          request_text_received_func, info);
          return;
        }
      else if (target == gdk_atom_intern_static_string ("UTF8_STRING"))
        {
          gtk_clipboard_request_contents (clipboard,
                                          gdk_atom_intern_static_string ("COMPOUND_TEXT"),
                                          request_text_received_func, info);
          return;
        }
      else if (target == gdk_atom_intern_static_string ("COMPOUND_TEXT"))
        {
          gtk_clipboard_request_contents (clipboard,
                                          GDK_TARGET_STRING,
                                          request_text_received_func, info);
          return;
        }
    }

  info->callback (clipboard, result, info->user_data);
  g_free (info);
  g_free (result);
}

```

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Drag and Drop/gtk/org/eclipse/swt/dnd/Clipboard.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Drag and Drop/gtk/org/eclipse/swt/dnd/Clipboard.java
@@ -300,23 +300,29 @@ public Object getContents(Transfer transfer, int clipboards) {
 
 	long selection_data = 0;
 	int[] typeIds = transfer.getTypeIds();
+	boolean textTransfer = transfer.getTypeNames()[0].equals("UTF8_STRING");
+	Object result = null;
 	for (int i = 0; i < typeIds.length; i++) {
 		if ((clipboards & DND.CLIPBOARD) != 0) {
 			selection_data = gtk_clipboard_wait_for_contents(GTKCLIPBOARD, typeIds[i]);
 		}
-		if (selection_data != 0) break;
-		if ((clipboards & DND.SELECTION_CLIPBOARD) != 0) {
+		if (selection_data == 0 && (clipboards & DND.SELECTION_CLIPBOARD) != 0) {
 			selection_data = gtk_clipboard_wait_for_contents(GTKPRIMARYCLIPBOARD, typeIds[i]);
 		}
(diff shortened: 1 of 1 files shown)


akurtakov:
I would appreciate if you split the functional change and reorganization in different PRs in order to make it easier what actually have you changed.

josediazfer:
Sorry, I don't understand exactly what to do. The change about a file where a small piece of code has been added and encapsulated in a new function.

josediazfer:
@akurtakov You're right, I just changed it so that the commit only has the functional change

akurtakov:
I wonder how can I reproduce this in native Linux environment. Would you please give some hints if you have an idea?

josediazfer:
You can reproduce the problem by establishing a VNC connection with the Remmina client from a Linux machine against a VNC server (x11vnc) as shown below. A connection made from the Windows operating system with other VNC clients (RealVNC, TigerVNC) it also happens.

fail_accents

akurtakov:
The code looks good to me. I'm merging it as it shouldn't make a difference for the currently supported cases.

PR #2153, opened 2025-05-14
Do not show warnings for missing javadoc in examples

akurtakov:
There is one extra change about unused lambda warnings which was autoadded by JDT that's why I keep it in.


akurtakov:
This removes some ~70 warnings in the workspace for things like missing javadoc for `throws Exception` in tests or similar.

PR #2154, opened 2025-05-14
Reorganized monitor-bound utils in Tree

amartya4256:
This commit contributes to moving
Tree:getContainingMonitorBoundsInPixels and
Tree:fitRectangleBoundsIntoMonitor to CoordinateSystemMapper and Display,
respectively for better clarity and encapsulation.

contributes to
https://github.com/eclipse-platform/eclipse.platform.swt/issues/128

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
@@ -1724,6 +1724,28 @@ public Point getCursorLocation () {
 	return coordinateSystemMapper.getCursorLocation();
 }
 
+Rectangle fitRectangleBoundsIntoMonitorWithCursor(RECT rect) {
+	Rectangle monitorBounds = coordinateSystemMapper.getContainingMonitorBoundsInPixels(getCursorLocation());
+	if (monitorBounds == null) {
+		return null;
+	}
+	int rectWidth = rect.right - rect.left;
+	int rectHeight = rect.bottom - rect.top;
+	if (rect.left < monitorBounds.x) {
+		rect.left = monitorBounds.x;
+	}
+	int monitorBoundsRightEnd = monitorBounds.x + monitorBounds.width;
+	if (rect.right > monitorBoundsRightEnd) {
+		if (rectWidth <= monitorBounds.width) {
+			rect.left = monitorBoundsRightEnd - rectWidth;
+		} else {
+			rect.left = monitorBounds.x;
+		}
+		rectWidth = monitorBoundsRightEnd - rect.left;
+	}
+	return new Rectangle(rect.left, rect.top, rectWidth, rectHeight);
+}
+
 Point getCursorLocationInPixels () {
 	POINT pt = new POINT ();
 	OS.GetCursorPos (pt);
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MultiZoomCoordinateSystemMapper.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MultiZoomCoordinateSystemMapper.java
@@ -271,4 +271,11 @@ class MultiZoomCoordinateSystemMapper implements CoordinateSystemMapper {
 		return DPIUtil.getZoomForAutoscaleProperty(monitor.zoom);
 	}
 
+	@Override
+	public Rectangle getContainingMonitorBoundsInPixels(Point point) {
+		Monitor monitor = point instanceof MonitorAwarePoint monitorAwarePoint ? monitorAwarePoint.getMonitor()
+				: getContainingMonitorForPoints(point.x, point.y);
+		return getMonitorClientAreaInPixels(monitor);
+	}
+
 }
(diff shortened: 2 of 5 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MultiZoomCoordinateSystemMapper.java
```diff
@@ -271,4 +271,12 @@ private int getApplicableMonitorZoom(Monitor monitor) {
 		return DPIUtil.getZoomForAutoscaleProperty(monitor.zoom);
 	}
 
+	@Override
+	public Rectangle getContainingMonitorBoundsInPixels(Point point) {
+		Monitor monitor = point instanceof MonitorAwarePoint monitorAwarePoint
+				? monitorAwarePoint.getMonitor()
+				:getContainingMonitorForPoints(point.x, point.y);
+		return getMonitorClientAreaInPixels(monitor);
```
HeikoKlare: Please auto-format this code.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
```diff
@@ -1724,6 +1724,25 @@ public Point getCursorLocation () {
 	return coordinateSystemMapper.getCursorLocation();
 }
 
+Rectangle fitRectangleBoundsIntoMonitorWithCursor(RECT rect) {
+	Rectangle monitorBounds = coordinateSystemMapper.getContainingMonitorBoundsInPixels(getCursorLocation());
```
HeikoKlare: The result may be null, so there needs to be a check and early return in case the result is null to avoid NPEs in the subsequent code.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
```diff
@@ -8175,15 +8175,10 @@ private LRESULT positionTooltip(NMHDR hdr, long wParam, long lParam, boolean man
...
-		Rectangle monitorBounds = cursorLocation instanceof MonitorAwarePoint monitorAwarePoint
-				? getContainingMonitorBoundsInMultiZoomCoordinateSystem(monitorAwarePoint)
-				: getContainingMonitorBoundsInSingleZoomCoordinateSystem(cursorLocation);
-		if (monitorBounds != null) {
-			Rectangle adjustedTooltipBounds = fitTooltipBoundsIntoMonitor(toolRect, monitorBounds);
-			OS.SetWindowPos (hdr.hwndFrom, 0, adjustedTooltipBounds.x, adjustedTooltipBounds.y, adjustedTooltipBounds.width, adjustedTooltipBounds.height, flags);
-			result = LRESULT.ONE;
-		}
+
+		Rectangle adjustedTooltipBounds = getDisplay().fitRectangleBoundsIntoMonitorWithCursor(toolRect);
```
HeikoKlare, line 8179: Looks to me as if the previous null-check was erroneously removed. There are multiple places at which null may be returned, so I think it needs to be checked here as well to avoid NPEs.

HeikoKlare (commented):
The overall change looks fine. In my understanding, this is not a pure refactoring but adds some additional functionality in `MultiZoomCoordinateSystemMapper::getContainingMonitorBoundsInPixels(Point)` for the case that the given point is not monitor-aware, right? But in my understanding that is rather a fallback as that path will currently not be taken as only the cursor location is passed to the method which, in case a multi-zoom mapper is used, will always be a monitor-aware point.

In addition to that, there seem to be some missing / lost null checks.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MultiZoomCoordinateSystemMapper.java
amartya4256: Done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
amartya4256: Done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
```diff
@@ -8175,15 +8175,12 @@ private LRESULT positionTooltip(NMHDR hdr, long wParam, long lParam, boolean man
...
-				: getContainingMonitorBoundsInSingleZoomCoordinateSystem(cursorLocation);
-		if (monitorBounds != null) {
-			Rectangle adjustedTooltipBounds = fitTooltipBoundsIntoMonitor(toolRect, monitorBounds);
+
+		Rectangle adjustedTooltipBounds = getDisplay().fitRectangleBoundsIntoMonitorWithCursor(toolRect);
+		if(adjustedTooltipBounds != null) {
 			OS.SetWindowPos (hdr.hwndFrom, 0, adjustedTooltipBounds.x, adjustedTooltipBounds.y, adjustedTooltipBounds.width, adjustedTooltipBounds.height, flags);
-			result = LRESULT.ONE;
 		}
+		result = LRESULT.ONE;
```
HeikoKlare: Is it intended that this is moved outside of the if block?

amartya4256: That was a miss

HeikoKlare (approved):
Looks good now, thank you! I also tested the changes and did not find any regressions. In particular, the tooltip alignment of trees to fit into the monitor still works as expected.

PR #2158, opened 2025-05-16
Fix zoom issue during copy paste in clipboard example

arunjose696:
Copy and paste no longer tries to autoscale the image. The image is always copied and pasted at 100% zoom.

The purpose of this change is to:  
1) Fix an issue in the clipboard example where the pasted image had different zoom levels across monitors  
2) Remove all calls to `DPIUtil.get[Native]DeviceZoom()`

Previously, when copying an image in the clipboard example, the image at the current zoom level was stored to clipboard using `getImageDataAtCurrentZoom()`. However, in normal Windows behavior (e.g., Paint), images are always copied at their original size regardless of the current zoom. When pasting, the image is pasted at 100% zoom, and the application handles zoom internally when displaying the image.

To match this behavior, copy and paste now always handle images at 100% scale, ignoring the current zoom and leaving zoom handling to the image class when it is fetched for display.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Drag and Drop/win32/org/eclipse/swt/dnd/ImageTransfer.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Drag and Drop/win32/org/eclipse/swt/dnd/ImageTransfer.java
@@ -15,7 +15,6 @@ package org.eclipse.swt.dnd;
 
 import org.eclipse.swt.*;
 import org.eclipse.swt.graphics.*;
-import org.eclipse.swt.internal.*;
 import org.eclipse.swt.internal.ole.win32.*;
 import org.eclipse.swt.internal.win32.*;
 
@@ -184,8 +183,9 @@ public Object nativeToJava(TransferData transferData) {
 					pSourceBits -= scanline;
 				}
 			}
-			Image image = Image.win32_new(null, SWT.BITMAP, memDib, DPIUtil.getNativeDeviceZoom());
-			ImageData data = image.getImageData (DPIUtil.getDeviceZoom ());
+			final int DEFAULT_IMAGE_STORAGE_ZOOM = 100;
+			Image image = Image.win32_new(null, SWT.BITMAP, memDib, DEFAULT_IMAGE_STORAGE_ZOOM);
+			ImageData data = image.getImageData ();
 			OS.DeleteObject(memDib);
 			image.dispose();
 			return data;
--- a/examples/org.eclipse.swt.examples/META-INF/MANIFEST.MF
+++ b/examples/org.eclipse.swt.examples/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %plugin.SWTStandaloneExampleSet.name
 Bundle-SymbolicName: org.eclipse.swt.examples; singleton:=true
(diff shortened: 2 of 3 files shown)


HeikoKlare (approved):
The change is sound. Always using the 100% version of an image via copy/paste makes sense. I was wondering about situations where an image is provided in at different resolution for different zooms (such as we do for icons with the `@2x` version or as would be possible for SVGs), but that does not seem to be a reasonable use case for copy/paste / DnD and in any case you would not know which zoom to actually use.
In addition, this aligns the implementation with the (also HiDPI-aware) implementation on MacOS (and Linux).

I have tested the change on Windows and MacOS (as the ClipboardExample is OS-agnostic). On Windows, I tested different primary monitor zooms (including values that are mapped to both device zooms of 100% and 200% with default int200-autoscaling).
It seems to work as expected.

I have also tested with monitor-specific scaling enabled (i.e., running the snippet with `-Dswt.autoScale.updateOnRuntime=true`) on a two-monitor setup with 200% and 100%. Behavior was completely broken before when switching between the monitors and now behaves properly fine and just as expected.

Actually, I think this is the proper solution for basically the same issue discussed 9 years ago: https://bugs.eclipse.org/bugs/show_bug.cgi?id=493047
There also a proposal for always using the image data at 100% was made, but was ignored in favor of using the adaptations to the `ClipboardExample`.

So I am all for merging this for 4.37 M1. But since this affects integral copy/paste behavior, I would like to have someone else have a look at the change and its behavior as well (@fedejeanne @akoch-yatta).

akoch-yatta (approved):
I gave it a short test round to see if a find any issue, but behavior looks good to me. Additionally I think using the zoom of 100% aligns now with the behavior of copy&paste of images in windows overall, therefor a +1 from me

HeikoKlare:
Just one minor request @arunjose696: could you please improve the commit message to have a short, comprehensive header and more information in a separated explaining body? You may just restructure the existing commit message or also take parts of your PR message for a more expressive commit body. See also the [commit message commendations](https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md#commit-message-recommendations).

arunjose696:
> Just one minor request @arunjose696: could you please improve the commit message to have a short, comprehensive header and more information in a separated explaining body? You may just restructure the existing commit message or also take parts of your PR message for a more expressive commit body. See also the [commit message commendations](https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md#commit-message-recommendations).

Done, should I patch my branch as per the bots comment?

HeikoKlare:
> Done

Thank you, but please only link to issues in the official SWT repository, not in the one we use for our work planning. In this case, we don't have an issue in this repo to link to, so you can just leave out those links. Sorry if I confused here by linking to the commit message recommendations.

> should I patch my branch as per the bots comment?

In general, yes, but please wait until the master branch is open again as the patch is currently not correct because we are in the transition to the new development stream. Once that's done, the bot should creat a proper patch proposal that you can add as a separate commit to the PR. It will contain required version bumps for the new development cycle.

arunjose696:
> > Done
> 
> Thank you, but please only link to issues in the official SWT repository, not in the one we use for our work planning. In this case, we don't have an issue in this repo to link to, so you can just leave out those links. Sorry if I confused here by linking to the commit message recommendations.

Yeah, I misunderstood I thought those were the commit message guidelines for eclipse-platform. Thanks for clarifying! 😅 Looks like I’ll need to go back and edit all the commit messages for other PRs I just updated.

Issue #2162, opened 2025-05-19
NPE in GC: Cannot read field "nativeZoom" because "this.data" is null


ptziegler:
I haven't found a small reproducer yet, but the problem occurs when opening and closing one of our dialogs via the context menu.

```
Caused by: java.lang.NullPointerException: Cannot read field "nativeZoom" because "this.data" is null
	at org.eclipse.swt.graphics.GC.getZoom(GC.java:5139)
	at org.eclipse.swt.graphics.GC.drawImage(GC.java:934)
	at org.eclipse.swt.widgets.Composite.WM_PAINT(Composite.java:1656)
	at org.eclipse.swt.widgets.Control.windowProc(Control.java:4835)
	at org.eclipse.swt.widgets.Canvas.windowProc(Canvas.java:336)
	at org.eclipse.swt.widgets.Display.windowProc(Display.java:5106)
	at org.eclipse.swt.internal.win32.OS.RedrawWindow(Native Method)
	at org.eclipse.swt.widgets.Control.update(Control.java:4615)
	at org.eclipse.swt.widgets.Display.update(Display.java:5076)
	at org.eclipse.jface.window.Window.runEventLoop(Window.java:830)
	at org.eclipse.jface.window.Window.open(Window.java:799)
```

Tested with org.eclipse.swt.win32.win32.x86_64 3.130.0.v20250507-1813

ptziegler:
The paint method raises a lot of questions...

Here, the `paintGC` instance is assigned to `gc`
https://github.com/eclipse-platform/eclipse.platform.swt/blob/8b45dd8b63249c606e02cec496d429598442e1e6/bundles/org.eclipse.swt/Eclipse%20SWT/win32/org/eclipse/swt/widgets/Composite.java#L1581-L1583

Later, the `gc` instance (and therefore also the `paintGC` instance) is explicitly disposed. However, the `paintGC` instance is then immediately used for painting.

https://github.com/eclipse-platform/eclipse.platform.swt/blob/8b45dd8b63249c606e02cec496d429598442e1e6/bundles/org.eclipse.swt/Eclipse%20SWT/win32/org/eclipse/swt/widgets/Composite.java#L1649-L1660

This is what's causing the NullPointerException. I just don't see why this wasn't an issue before...

ptziegler:
[...]

PR #2163, opened 2025-05-19
Revert "Replace usages of new Image(device, width, height)..." #2162

fedejeanne:
This reverts commit 807b159fb65000762a179e6793cdd29e86a99cd4.

Add regression test.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2162

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Composite.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Composite.java
@@ -1554,6 +1554,8 @@ LRESULT WM_PAINT (long wParam, long lParam) {
 			data.ps = ps;
 			data.hwnd = handle;
 			GC gc = GC.win32_new (this, data);
+
+			/* Get the system region for the paint HDC */
 			long sysRgn = 0;
 			if ((style & (SWT.DOUBLE_BUFFERED | SWT.TRANSPARENT)) != 0 || (style & SWT.NO_MERGE_PAINTS) != 0) {
 				sysRgn = OS.CreateRectRgn (0, 0, 0, 0);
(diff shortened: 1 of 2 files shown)


fedejeanne:
Since a proper fix is not trivial, I will simply revert the commit and create a follow-up PR where 807b159fb65000762a179e6793cdd29e86a99cd4 is improved and reintroduced. The follow-up PR will be targeted for M1.

cc @ShahzaibIbrahim

HeikoKlare (commented):
The revert is fine, as the original change obivous introduced a bug which we need to revert for the upcoming release.

But the provided regression test needs to be revised:
- It does not test anything specifically regarding transparent style of the Composite. It only tests that one specific kind of regression does not occur in the sense of not leading to an exception in case a specific setup for shell and composite around is used, but that has nothing to do with testing transparent style for composite itself but some very specific regression to composite.
- It creates a new shell even though the test class already provides infrastructure including a shell (and even a composite). In case another shell is required (which does not seem to be the case, as I can simply remove the additional shell), that one would at least need to be safely disposed after the test execution.
- Why is setting the layout to the shell necessary here? If the test is about transparent style, it should have nothing to do with the layout of the containing shell. If it has some relevance, it needs to become clear by proper test naming, commenting or the like.
- Same as for the point above applies to the paint listener to the composite.

fedejeanne:
I adapted the name of the test and used the `shell` in the parent class.

The listener and the layout are necessary for the use case, otherwise the test doesn't run into the issue. Since this is a very complicated scenario and even getting a reproducer is not trivial, I simply took the reproducer from #2162 and stripped it down to the bare minimum.

fedejeanne:
Check failures (tests in Linux are unrelated:

```
Error:  Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:4.0.13-SNAPSHOT:compile (default-compile) on project org.eclipse.swt.cocoa.macosx.x86_64: can't call classpath contributors: Could not acquire lock(s) -> [Help 1]
```

I see the same for all 3 platforms. Very likely a follow-up of this failure:

```
Error:  Plugin org.eclipse.tycho:tycho-versions-plugin:4.0.13-SNAPSHOT or one of its dependencies could not be resolved:
Error:  	The following artifacts could not be resolved: org.eclipse.tycho:tycho-versions-plugin:pom:4.0.13-SNAPSHOT (absent): org.eclipse.tycho:tycho-versions-plugin:pom:4.0.13-SNAPSHOT failed to transfer from https://repo.eclipse.org/content/repositories/tycho-snapshots/ during a previous attempt. This failure was cached in the local repository and resolution is not reattempted until the update interval of tycho-snapshots has elapsed or updates are forced. Original error: Could not transfer artifact org.eclipse.tycho:tycho-versions-plugin:pom:4.0.13-SNAPSHOT from/to tycho-snapshots (https://repo.eclipse.org/content/repositories/tycho-snapshots/): status code: 403, reason phrase: Forbidden (403)
```

HeikoKlare (approved):
Thank you for the test improvement!

PR #2169, opened 2025-05-20
Simplify Spinner test

akurtakov:
* Use assertThrows
* Do not override parent test method so it actually gets executed


akurtakov:
Failing tests fails in master too. Merging.

PR #2172, opened 2025-05-21
[GTK] Fix Spinner [get|set]TextLimit on GTK 4

akurtakov:


akurtakov:
Changes only GTK4 behavior so should be safe change . I would appreciate an extra review in an effort to have this in June release.

PR #2175, opened 2025-05-21
Fix setTextContainingScript Test #1466

amartya4256:
This PR fixes the test
org.eclipse.swt.tests.junit.Test_org_eclipse_swt_browser_Browser.test_setTextContainingScript_applicationLayerProgressListenerMustSeeUpToDateDom() as Edge seems to misbehave sporadically when an invalid script source is added using Edge:setText. Hence, with this commit we replace the script tag with a valid statement in the test and enable it.

Contributes to https://github.com/eclipse-platform/eclipse.platform.swt/issues/1466


fedejeanne (approved):
The change in the test looks fine.

Didn't we say we wanted to improve the JavaDoc of `Browser::setText(String)` to say mention that "we expect well-formed HTML (without broken scripts)" or something similar?

PR #2177, opened 2025-05-22
Add a negative requirement on darkest dark 2024.4.0.202412101111

merks:
https://github.com/eclipse-packaging/packages/issues/310

### DIFF
--- a/bundles/org.eclipse.swt/META-INF/p2.inf
+++ b/bundles/org.eclipse.swt/META-INF/p2.inf
@@ -44,3 +44,11 @@ requires.9.namespace = org.eclipse.equinox.p2.iu
 requires.9.name = org.eclipse.swt.gtk.linux.riscv64
 requires.9.range = [$version$,$version$]
 requires.9.filter = (&(osgi.os=linux)(osgi.ws=gtk)(osgi.arch=riscv64)(!(org.eclipse.swt.buildtime=true)))
+
+# Ensure that 4.36 is not installed with the 0.0.1,2024.4.0.202412101111 version of darkest dark devstyle.
+# https://github.com/eclipse-packaging/packages/issues/307
+requires.100.namespace=org.eclipse.equinox.p2.iu
+requires.100.name=com.genuitec.eclipse.theming.core.feature.feature.group
+requires.100.range = [0.0.1,2024.4.0.202412101111]
+requires.100.min = 0
+requires.100.max = 0


merks:
FYI, I can see that this works because Oomph's repository explorer can explore the contents of the workspace projects published as IUs by p2's publisher:

image

bundles/org.eclipse.swt/META-INF/p2.inf
```diff
@@ -44,3 +44,11 @@ requires.9.namespace = org.eclipse.equinox.p2.iu
 requires.9.name = org.eclipse.swt.gtk.linux.riscv64
 requires.9.range = [$version$,$version$]
 requires.9.filter = (&(osgi.os=linux)(osgi.ws=gtk)(osgi.arch=riscv64)(!(org.eclipse.swt.buildtime=true)))
+
+# Ensure that 4.36 is not installed with the 0.0.1,2024.4.0.202412101111 version of darkest dark devstyle.
+# https://github.com/eclipse-packaging/packages/issues/307
+requires.100.namespace=org.eclipse.equinox.p2.iu
+requires.100.name=com.genuitec.eclipse.theming.core.feature.feature.group
+requires.100.range = [0.0.1,2024.4.0.202412101111]
```
akurtakov, line 52: Do we know that they don't auto push things with new qualifier? IMO it's best this to become [0.0.1, 2024.4.1).

laeubi:
@merks adding this to SWT bundle really seems odd. Isn't this a problem of jface?

jonahgraham:
The thing they are code weaving badly is org.eclipse.ui.internal.BrandingProperties - within their modifications they are referring to now deleted org.eclipse.jface.resource.FileImageDescriptor

So I think there are numerous bundles that this negative req can go on, but I agree that JFace makes sense. But SWT may work too, at least for EPP because we don't really support updating JFace without updating SWT too.

laeubi:
> at least for EPP

If only EPP is of a concern it could probably better added there... SWT really has nothing to do with that as the File was deleted from JFace.

I really find it ugly to add this to a bundle as we effectively should not maintain a list of bad  third party plugins, but if we do so it should at least be on the right place.

merks:
So you guys think the right place is the JFace bundle?  It's ugly no matter where we put it.  😢

merks:
Here instead:

https://github.com/eclipse-platform/eclipse.platform.ui/pull/3002

merks:
> If only EPP is of a concern it could probably better added there... SWT really has nothing to do with that as the File was deleted from JFace.
> 
> I really find it ugly to add this to a bundle as we effectively should not maintain a list of bad third party plugins, but if we do so it should at least be on the right place.

Nothing in EPP prevents the Platform from updating.  The updated Platform is the problem so this is also a problem for the Eclipse SDK with darkest dark installed.   So this is a problem that darkest dark does not work with the latest Platform...

bundles/org.eclipse.swt/META-INF/p2.inf
merks: I really think folks are missing the point and are now split hairs.  We are excluding the broken version and are assuming that any new version will not be broken because that's what folks have promised.

Issue #2098, opened 2025-05-05
Test_org_eclipse_swt_browser_Browser.test_LocationListener_then_ProgressListener fails randomly


fedejeanne:
`Test_org_eclipse_swt_browser_Browser.test_LocationListener_then_ProgressListener` fails randomly, _e.g._ in https://github.com/eclipse-platform/eclipse.platform.swt/pull/2096

Stack-trace:
```java
Error:  org.eclipse.swt.tests.junit.Test_org_eclipse_swt_browser_Browser.test_LocationListener_then_ProgressListener[browser flags: 0] -- Time elapsed: 17.07 s <<< FAILURE!
java.lang.AssertionError: 

Unexpected listener states. Expecting true for all, but have:
Location changed: false
ProgressChangedAfterLocationChanged: false
progressChanged: false
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.assertTrue(Assert.java:42)
	at org.eclipse.swt.tests.junit.Test_org_eclipse_swt_browser_Browser.test_LocationListener_then_ProgressListener(Test_org_eclipse_swt_browser_Browser.java:676)
...
```

akurtakov:
Closing as this tests hasn't failed lately.

PR #2180, opened 2025-05-23
[win32] Adapt GC zoom when using plain images

akoch-yatta:
This PR adapts the zoom passed to the GC when it is initialized for a plain Image created via Image#Image(Device, int, int) with a given width and height. It will fall back to DPIUtil#getDeviceZoom() again to revert to original bevahiour and prevent blurry rescaling of old usages.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
@@ -2108,6 +2108,11 @@ private class PlainImageProviderWrapper extends AbstractImageProviderWrapper {
 	}
 
 	@Override
+	protected long configureGCData(GCData data) {
+		return configureGC(data, DPIUtil.getDeviceZoom());
+	}
+
+	@Override
 	protected Rectangle getBounds(int zoom) {
 		Rectangle rectangle = new Rectangle(0, 0, width, height);
 		return DPIUtil.scaleUp(rectangle, zoom);


HeikoKlare:
The issue was introduced with this PR:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2064

It can, e.g., be seen in double-buffering scenarios. One such scenario can be reproduced with the [WindowBuilder](https://github.com/eclipse-windowbuilder/windowbuilder) project when going back to this commit: https://github.com/eclipse-windowbuilder/windowbuilder/commit/f69b7dd8d10ddf1fc84fcec4b2e4ffbffa3c4bcc

Starting the application on Windows without monitor-specific scaling at 175% (or above) and creating a sample projects leads to this wrong scaling:
image

With the proposed fix, it will look like this:
image

With monitor-specific scaling enabled, it will like the above instead of having blurry scaling as before:
image

HeikoKlare (approved):
This change should go in the upcoming release (and also in RC1 to have some remaining time of testing) to preserve the behavior of users of the `Image(device, width, height)` constructor on Windows.

I consider this change rather safe to merge as it consistutes a partial revert of the recently merged change:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2064

So the risk that consumers already adapted to the changed behavior is low. At the same time, the risk that consumers that did not adapt will face unintended behavior is probably not that low.

HeikoKlare:
The thanks are yours, Patrick! ❤️ You have pointed us to that potential issue, otherwise we would have probably missed it.
Do you have any objections on merging this for RC1, @ptziegler?

jonahgraham:
> Do you have any objections on merging this for RC1

IIUC the RC1 is already complete and about to be submitted to SimRel - see https://www.eclipse.org/lists/eclipse-dev/msg12407.html

iloveeclipse:
> > Do you have any objections on merging this for RC1
> 
> IIUC the RC1 is already complete and about to be submitted to SimRel - see https://www.eclipse.org/lists/eclipse-dev/msg12407.html

Not really, see https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3055

merks:
We've been spinning a bit...

We should contribute something today, and then we/I can contribute something newer tomorrow. 

I have to run now though...

HeikoKlare:
Thank you all! I will merge this now and then trigger another I-Build, which we may then contribute as (an update for) RC1.

The failing test is unrelated and already documented: https://github.com/eclipse-platform/eclipse.platform.swt/issues/2098

PR #2173, opened 2025-05-21
MenuItem (GTK): allow additional shortcut words #1852

tmssngr:
This allows to use more shortcuts to be displayed for a menu item, e.g. `Ctrl+Keypad +` or `Alt+Down`.

PR #2182, opened 2025-05-24
Control: remove dependency on Browser and OleClientSite

tmssngr:
This is better design and allows to compile SWT without the "bundles\org.eclipse.swt\Eclipse SWT Browser\" source path.


tmssngr:
Huh, my email still was fine for PR #2173 3 days ago.

HeikoKlare (commented):
Getting rid of the instanceof checks would of course be great. I do not remember why exactly we have implemented it this way. Probably it was related to the fact that browser is a common widget, which is also why the proposed change does not compile.
@tmssngr do you plan to fix this PR?

HeikoKlare:
@tmssngr do you plan to continue working on this?

tmssngr:
No, I close it.

PR #2183, opened 2025-05-26
misc: fix spelling

jbampton:


akurtakov:
Thanks for the fixes! It's merged via the referenced PR due to auto version bump failing for us in this case for yet unknown reasons.

PR #2186, opened 2025-05-27
Improve API for cloning/creating rectangles

amartya4256:
This PR contributes to providing better interfaces and APIs for handling MonitorAware Rectangles and Points with Abstraction using prototype pattern.

All the Monitor Aware classes implement MonitorAware. Also th coordinate classes liek Rectangle and Point implement an interface copy. This should be introduced as a convention to use copy or copyWith, when a consumer wants to create a new object from the existing one. If there's any offsetting needed, the method copyWith can be used, making sure we copy all the abstract fields of the underlying Child class (if applicable) to the new object.

Also this PR introduces a new method Rectangle.of(Point, int, int) to make it easy to create rectangles using points instead of calling the Rectangle Constructor.

contributes to

### DIFF
--- a/binaries/org.eclipse.swt.cocoa.macosx.aarch64/META-INF/MANIFEST.MF
+++ b/binaries/org.eclipse.swt.cocoa.macosx.aarch64/META-INF/MANIFEST.MF
@@ -3,7 +3,7 @@ Fragment-Host: org.eclipse.swt;bundle-version="[3.128.0,4.0.0)"
 Bundle-Name: %fragmentName
 Bundle-Vendor: %providerName
 Bundle-SymbolicName: org.eclipse.swt.cocoa.macosx.aarch64; singleton:=true
-Bundle-Version: 3.130.100.qualifier
+Bundle-Version: 3.131.0.qualifier
 Bundle-ManifestVersion: 2
 Bundle-Localization: fragment
 Export-Package: 
--- a/binaries/org.eclipse.swt.cocoa.macosx.x86_64/META-INF/MANIFEST.MF
+++ b/binaries/org.eclipse.swt.cocoa.macosx.x86_64/META-INF/MANIFEST.MF
@@ -3,7 +3,7 @@ Fragment-Host: org.eclipse.swt;bundle-version="[3.128.0,4.0.0)"
 Bundle-Name: %fragmentName
 Bundle-Vendor: %providerName
 Bundle-SymbolicName: org.eclipse.swt.cocoa.macosx.x86_64; singleton:=true
-Bundle-Version: 3.130.100.qualifier
+Bundle-Version: 3.131.0.qualifier
 Bundle-ManifestVersion: 2
 Bundle-Localization: fragment
 Export-Package: 
--- a/binaries/org.eclipse.swt.gtk.linux.aarch64/META-INF/MANIFEST.MF
+++ b/binaries/org.eclipse.swt.gtk.linux.aarch64/META-INF/MANIFEST.MF
@@ -3,7 +3,7 @@ Fragment-Host: org.eclipse.swt;bundle-version="[3.128.0,4.0.0)"
 Bundle-Name: %fragmentName
 Bundle-Vendor: %providerName
 Bundle-SymbolicName: org.eclipse.swt.gtk.linux.aarch64; singleton:=true
-Bundle-Version: 3.130.100.qualifier
+Bundle-Version: 3.131.0.qualifier
 Bundle-ManifestVersion: 2
 Bundle-Localization: fragment
 Export-Package: 
--- a/binaries/org.eclipse.swt.gtk.linux.loongarch64/META-INF/MANIFEST.MF
+++ b/binaries/org.eclipse.swt.gtk.linux.loongarch64/META-INF/MANIFEST.MF
@@ -3,7 +3,7 @@ Fragment-Host: org.eclipse.swt;bundle-version="[3.128.0,4.0.0)"
 Bundle-Name: %fragmentName
 Bundle-Vendor: %providerName
 Bundle-SymbolicName: org.eclipse.swt.gtk.linux.loongarch64; singleton:=true
-Bundle-Version: 3.130.0.qualifier
+Bundle-Version: 3.131.0.qualifier
 Bundle-ManifestVersion: 2
 Bundle-Localization: fragment
 Export-Package: 
--- a/binaries/org.eclipse.swt.gtk.linux.ppc64le/META-INF/MANIFEST.MF
+++ b/binaries/org.eclipse.swt.gtk.linux.ppc64le/META-INF/MANIFEST.MF
@@ -3,7 +3,7 @@ Fragment-Host: org.eclipse.swt;bundle-version="[3.128.0,4.0.0)"
 Bundle-Name: %fragmentName
 Bundle-Vendor: %providerName
 Bundle-SymbolicName: org.eclipse.swt.gtk.linux.ppc64le;singleton:=true
-Bundle-Version: 3.130.100.qualifier
+Bundle-Version: 3.131.0.qualifier
 Bundle-ManifestVersion: 2
 Bundle-Localization: fragment
 Export-Package: 
--- a/binaries/org.eclipse.swt.gtk.linux.riscv64/META-INF/MANIFEST.MF
+++ b/binaries/org.eclipse.swt.gtk.linux.riscv64/META-INF/MANIFEST.MF
@@ -3,7 +3,7 @@ Fragment-Host: org.eclipse.swt;bundle-version="[3.128.0,4.0.0)"
 Bundle-Name: %fragmentName
 Bundle-Vendor: %providerName
 Bundle-SymbolicName: org.eclipse.swt.gtk.linux.riscv64; singleton:=true
-Bundle-Version: 3.130.100.qualifier
+Bundle-Version: 3.131.0.qualifier
 Bundle-ManifestVersion: 2
 Bundle-Localization: fragment
 Export-Package: 
--- a/binaries/org.eclipse.swt.gtk.linux.x86_64/META-INF/MANIFEST.MF
+++ b/binaries/org.eclipse.swt.gtk.linux.x86_64/META-INF/MANIFEST.MF
@@ -3,7 +3,7 @@ Fragment-Host: org.eclipse.swt;bundle-version="[3.128.0,4.0.0)"
 Bundle-Name: %fragmentName
 Bundle-Vendor: %providerName
 Bundle-SymbolicName: org.eclipse.swt.gtk.linux.x86_64; singleton:=true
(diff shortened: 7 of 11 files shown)


bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Rectangle.java
```diff
@@ -356,4 +356,29 @@ public Rectangle union (Rectangle rect) {
...
+ */
+@Override
+public Rectangle copy() {
+	return new Rectangle(x, y, width, height);
+}
+
+/**
+ * @since 3.130
+ */
+public Rectangle copyWith(int dx, int dy, int dWidth, int dHeight) {
```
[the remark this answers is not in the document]
amartya4256: The upper approach is not clean and the user might wanna rather call the constructor instead and create a new instance with the values. That's why having such API makes sense. Also had a talk with @akoch-yatta about it.

amartya4256:
> * I see that `Copyable` is merely a "better named version of `Cloneable`" and it was probably added so that the methods `copy()` and `copyWith(...)` are associated with each other (?) . If that's the case, I think getting rid of `copyWith(...)` and replacing `Copyable` with `Clonable` is really a valid option.

Cloneable has clone() method as protected. If it is fine to override it to public without breaking the API conventions of SWT, I can use that for sure.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/MonitorAwareRectangle.java
```diff
@@ -61,4 +61,14 @@ public int hashCode() {
 		return super.hashCode();
 	}
 
+	@Override
+    public Rectangle clone() {
```
HeikoKlare, line 65: I have to admit that I don't understand why a `MonitorAwareRectangle` should not reveal that it's clone is a `MonitorAwareRectangle`. So I would be in favor of returning the actual/correct type here (also in the `cloneWith`) method.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Rectangle.java
```diff
@@ -356,4 +356,29 @@ public Rectangle union (Rectangle rect) {
...
+ */
+@Override
+public Rectangle clone() {
+	return new Rectangle(x, y, width, height);
+}
+
+/**
+ * @since 3.130
+ */
+public Rectangle cloneWith(int dx, int dy, int dWidth, int dHeight) {
```
HeikoKlare, line 380: To be more expressive, either the parameters should or the method name should be more self-explaning. E.g., call the parameters `delta*` instead of `d*` and/or call the method `cloneTranslatedAndResized` or the like.

HeikoKlare (commented):
Overall, the change looks reasonable to me. Having proper clone functionality for those data classes is fine. I am not sure whether we can have a more expressive name for `Rectangle.of`.

> Cloneable has clone() method as protected. If it is fine to override it to public without breaking the API conventions of SWT, I can use that for sure.

Note that Cloneable has not `clone()` method at all. `clone()` is defined by `Object` in Java and needs to be overwritten and made public in classes that implement the `Clonable` interface. I don't see that this can conflict with any conventions or contracts in SWT. `Rectangle` cannot be extended, so there is not chance that there is a conflicting overwrite of the method.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/MonitorAwareRectangle.java
amartya4256: I had it as `MonitorAwareRectangle` earlier but I changed it w.r.t. @fedejeanne request. refer https://github.com/eclipse-platform/eclipse.platform.swt/pull/2186#discussion_r2111077306

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Rectangle.java
amartya4256: I would go with a better name since it's more convenient for the consumers to make sense of it.

laeubi:
Just an idea: 

Do we really need an own class here? Can't we add a (package protected?) field to the regular `Rectangle` and mark it as not for public use like we do with the handles in `Control`?

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Rectangle.java
amartya4256: Removing it and would be followed in the next PR

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/MonitorAwareRectangle.java
amartya4256: Reverted

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Rectangle.java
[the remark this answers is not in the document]
amartya4256: Would be followed in the next PR

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Rectangle.java
```diff
@@ -356,4 +356,39 @@ public Rectangle union (Rectangle rect) {
...
+ * @param width the width of the rectangle
+ * @param height the height of the rectangle
+ * @return a new {@code Rectangle} instance appropriate for the given point and dimensions
+ * @since 3.131
+ */
+public static Rectangle of(Point topLeft, int width, int height) {
+    if (topLeft instanceof MonitorAwarePoint monitorAwareTopLeft) {
+        return new MonitorAwareRectangle(topLeft.x, topLeft.y, width, height, monitorAwareTopLeft.getMonitor());
+    }
+    return new Rectangle(topLeft.x, topLeft.y, width, height);
```
arunjose696:
```suggestion
public static Rectangle of(Point topLeft, int width, int height) {
    return new Rectangle(topLeft.x, topLeft.y, width, height);
}

public static MonitorAwareRectangle of(MonitorAwarePoint topLeft, int width, int height) {
    return new MonitorAwareRectangle(topLeft.x, topLeft.y, width, height, topLeft.getMonitor());
}
```

Maybe a better style?

HeikoKlare: Would be nice if that worked but in Java method calls are bound based on the static types of arguments. Thus if you have a `MonitorAwarePoint` at hand which is statically typed as `Point`, the "wrong" method would be called. 

Here, `rect` would be of dynamic type `Rectangle` even though it is supposed to be a `MonitorAwareRectangle`:
```java
Point point = new MonitorAwarePoint(...);
Rectange rect = Rectangle.of(point, 1, 1);
```

HeikoKlare:
> Just an idea:

> Do we really need an own class here? Can't we add a (package protected?) field to the regular Rectangle and mark it as not for public use like we do with the handles in Control?

The class already exists. This change just adds further functionality.

Back then, we also discussed adding a field to the existing `Rectangle` class, but there are at least two reasons against it:
- The `Rectangle` is a quite general type: it is used all over the place and adding another (usually useless) field unnecessary bloats up instances
- A pure `Rectangle` has nothing to with a monitor (as it is not semantically tied to a monitor), so adding such a field would be strange also regarding the class' semantics.

laeubi:
If that holds true then one should actually not extend Rectangle but e.g. use a record that combines a rectangle with a monitor.

HeikoKlare:
> If that holds true then one should actually not extend Rectangle but e.g. use a record that combines a rectangle with a monitor.

If it's not a subclass, you cannot pass it via existing APIs that accept or return instances of Rectangle.

laeubi:
> > If that holds true then one should actually not extend Rectangle but e.g. use a record that combines a rectangle with a monitor.
> 
> If it's not a subclass, you cannot pass it via existing APIs that accept or return instances of Rectangle.

The obviously the monitor is an important property in some cases and there are only others where it is irrelevant. This is similar to how we do it with `Event`s where some fields are filled and others are not depending on the context / usage. I also don't see that we create millions of Point/Rectangles that the object size actually matters.

So either we should have new API that accepts a Point/Rectangle + Monitor or we should account for the fact that the monitor is a property of a Rectangles or point. Having only internal used classes with marker interface and trying to restrict what we made public seems not very useful and complicates the matter more than it solves it.

HeikoKlare:
> Having only internal used classes with marker interface and trying to restrict what we made public seems not very useful and complicates the matter more than it solves it.

Why? And what's the marker interface here?
I have to admit that I don't see the problem that we need to address. Unless it's necessary and well designed to expose a propery, why should we do that?
The proposed API for cloning rectangles in this PR is even reasonable independent of whether there are monitor-aware specialization or not.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Rectangle.java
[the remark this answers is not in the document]
fedejeanne: I see your point about the 2nd argument (moving the rectangle and updating the monitor) but I think this method still combines 2 things that should be addressed separately (and properly):

1. Duplicating the rectangle while keeping the monitor-awareness
2. Moving the rectangle (and update the monitor information if appropriate)

One can always do them "wrong" and lose the information about the monitor _e.g._ by doing
```
Rectangle r2 = new Rectangle(r1.x, r1.y, r1.width, r1.height);
r2.x += 5;
```

This is unavoidable since it's too easy and comfortable for the consumer to do it. But the solution is not to provide 1 method that does 2 things (copy + move) but instead to provide 2 methods that do 2 things properly (preserve / update the information about the monitor) and to inform the user that _using the other way of copying / moving rectangles may lose contextual information that is relevant for some use cases_ (remember that a `Rectangle` is a very general term which is not only valid in the context of "monitors"). 

Also, imagine that the consumer does not want to move the new rectangle but simply copy it, he/she would have to call `r1.copyWith(0,0,0,0)` which seems odd.

Another indicator of a smell is the fact that the name of the method is not really clear as to what it actually does: _with_ is rather open for interpretation. Something like _copyAndMove_ would be more fitting and this one clearly indicates that the method does 2 things.

HeikoKlare (approved):
I've corrected wrong code formatting and added the necessary version bump commit, as a minor bump for SWT is required due to added API.

Issue #2188, opened 2025-05-28
SWTError: "Item not added" in Image.setImageMetadataForHandle(ImageHandle imageMetadata, Integer zoom)


estepper:
**Describe the bug**
Since 2025-06, when I try to display an image that was created via `new Image(display, sourceImage, SWT.IMAGE_DISABLE)` from a `sourceImage = new Image(display, new ByteArrayInputStream(bytes))`, I get the following SWTError:

```
Exception in thread "main" org.eclipse.swt.SWTError: Item not added
	at org.eclipse.swt.SWT.error(SWT.java:4970)
	at org.eclipse.swt.SWT.error(SWT.java:4860)
	at org.eclipse.swt.SWT.error(SWT.java:4831)
	at org.eclipse.swt.graphics.Image.setImageMetadataForHandle(Image.java:1553)
	at org.eclipse.swt.graphics.Image$ImageHandle.<init>(Image.java:2619)
	at org.eclipse.swt.graphics.Image.initBitmapHandle(Image.java:1581)
	at org.eclipse.swt.graphics.Image.init(Image.java:1672)
	at org.eclipse.swt.graphics.Image$AbstractImageProviderWrapper.newImageHandle(Image.java:1911)
	at org.eclipse.swt.graphics.Image$ImageFromImageDataProviderWrapper.initializeHandleFromSource(Image.java:1990)
	at org.eclipse.swt.graphics.Image$ImageFromImageDataProviderWrapper.newImageHandle(Image.java:1983)
	at org.eclipse.swt.graphics.Image.getImageMetadata(Image.java:785)
	at org.eclipse.swt.graphics.Image.win32_getHandle(Image.java:809)
	at org.eclipse.swt.graphics.GC.drawImage(GC.java:1113)
	at org.eclipse.swt.graphics.GC.drawImage(GC.java:1052)
	at org.eclipse.swt.graphics.GC.drawImageInPixels(GC.java:948)
	at org.eclipse.swt.graphics.GC.drawImage(GC.java:941)
	at org.eclipse.swt.widgets.Label.wmDrawChildImage(Label.java:570)
	at org.eclipse.swt.widgets.Label.wmDrawChild(Label.java:617)
	at org.eclipse.swt.widgets.Control.WM_DRAWITEM(Control.java:5006)
	at org.eclipse.swt.widgets.Control.windowProc(Control.java:4793)
[...]

PR #2190, opened 2025-05-28
[Win32] Initialize copied image provider wrapper on correct image #2188

HeikoKlare:
When copying an image based on an input stream, the copied provider wrapper was erroneously created on the original image. This change adapts the instantiation to create the warpper on the correct image.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2188

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
@@ -2084,7 +2084,7 @@ private class ImageDataLoaderStreamProviderWrapper extends ImageFromImageDataPro
 
 	@Override
 	AbstractImageProviderWrapper createCopy(Image image) {
-		return new ImageDataLoaderStreamProviderWrapper(inputStreamData);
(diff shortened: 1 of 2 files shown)


akoch-yatta (approved):
That was obviously a bug, as the copy must be based on the passed image, as it is done in the other wrappers. Tested it with the snippet from the issue this PR fixes and works as expected.

merks (approved):
A very simple fix to implement exactly what’s intended along with a test case. Very good. 

+1 from the PMC for merging

HeikoKlare:
For the record: this was obviously a sloppy mistake
- All other introduced image provider wrappers create their copy on the new image that is passed as intended, so this just aligns the one faulty provider with the others
- There is a reproducing snippet and test case demonstrating the regression and the appropriateness of the fix

To avoid that we face the regression in the upcoming release, this fix needs to go in master before the release (so for RC2).

Thank you @merks for already giving your +1!

HeikoKlare:
And many thanks again to @estepper for reporting the issue with a great reproducer early enough to be fixed for the upcoming release ❤️

estepper:
A pleasure! I'm happy to see the fix going into the release. Thanks for your prompt action ;-)

PR #2193, opened 2025-06-02
Configure SWT build scripts for 4.37

MohananRahul:
https://github.com/eclipse-platform/eclipse.platform.swt/issues/2192

PR #2198, opened 2025-06-04
Add version bumps for 4.37

HannesWell:


HannesWell:
This has finally reached the same problem as described in
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2193#issuecomment-2940447768

But I could work around it by pushing this change directly to the eclipse.platform.swt repo (not my fork)
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2199

Thus closing this.

PR #2199, opened 2025-06-04
Add version bumps for 4.37

HannesWell:
Provoke version increment for 4.37.

The same as https://github.com/eclipse-platform/eclipse.platform.swt/pull/2198, but pushed to the SWT repo directly in an attempt to avoid the push-rights problem.


HannesWell:
Fixing the new api-tools warnings can be done later.

Issue #2203, opened 2025-06-05
CTabFolder: add LEFT/RIGHT behaviour


pazi146:
As part of my degree, I'm going to implement left/right positions for CTabFolder

Plan:
- enhance `CTabFolder` and `CTabFolderRenderer`
- enhance `CTabFolderRendering` (in `eclipse.platform.ui`)
- cleanup code

PR #2204, opened 2025-06-05
CTabFolder LEFT/RIGHT 01 #2203

pazi146:
getters/setters:
SWT.LEFT and SWT.RIGHT are now valid positions
tab width can be set

no changes changes in behaviour (RIGHT and LEFT act like TOP)


akurtakov:
Is this still being worked on or we should close this PR?

akurtakov:
Closing due to months with no reply and the patch is not ready.

PR #2207, opened 2025-06-05
[Build] Use Eclipse Platform Bot to commit built SWT binaries

HannesWell:
Use the `Eclipse Platform Bot` to commit built SWT binaries with its correct email as recorded in the EF-API (https://api.eclipse.org/bots ) and that is also associated with it's Github account. Before the `Eclipse Releng Bot` was used for that, but with an incorrect email.

Not it's the same bot that is also used for version increments:
https://github.com/eclipse-platform/eclipse.platform.swt/blob/5115799730f502b7c3b2d958d3cf2ec37ba71d11/.github/workflows/pr-checks.yml#L20-L21

I think that is more suitable than using the Releng Bot (even with a fixed mail).

Fixes
- https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/6227

@akurtakov, @iloveeclipse, @HeikoKlare any objection?

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -116,32 +116,31 @@ pipeline {
 					script {
 						def authorMail = sh(script: 'git log -1 --pretty=format:"%ce" HEAD', returnStdout: true)
 						echo 'HEAD commit author: ' + authorMail
-						if ('eclipse-releng-bot@eclipse.org'.equals(authorMail) && !params.forceNativeBuilds) {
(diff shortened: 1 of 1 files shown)


akurtakov:
Looks good to me.

HeikoKlare:
Looks good to me as well.

HannesWell:
Great, thanks.

PR #2210, opened 2025-06-06
Remove unused apifilters

akurtakov:


HeikoKlare:
We have the same unused filters for Windows and MacOS. May I extend your PR? Or should I provide a follow-up one?

akurtakov:
Feel free to extend it.

HeikoKlare:
I've updated the PR to consistently remove the unused filters from all fragments. There seem to be even more unused filters (basically all the filters for the SWTEventListener changes), but I would keep them for a follow up.

PR #2211, opened 2025-06-06
Amartya4256/c tab folder fix float aware

amartya4256:
This PR is to show that properly utilizing the float aware residuals indeed fixes the border issue of CTabFolder. We need a cleaner approach for this.


amartya4256:
Closing as the issue is fixed by https://github.com/eclipse-platform/eclipse.platform.swt/pull/2295

PR #2212, opened 2025-06-06
Deprecate DefaultSWTFontRegistry and Completely switch to ScalingSWTFontRegistry

arunjose696:
**Summary**
This PR introduces the following changes:

1)Always use ScalingSWTFontRegistry
2)Rename DefaultSWTFontRegistry -> LegacySWTFontRegistry as this is not default anymore
3)Add a system property to use LegacySWTFontRegistry without monitor specific scaling active
4)Deprecate the LegacySWTFontRegistry(DefaultSWTFontRegistry)
5)additionally the change also makes sure all fonts are cached when they are created, this is necessary to make sure the systemfonts would also be cached when they are created (this change is to fix  Test_org_eclipse_swt_graphics_GC.test_setFontLorg_eclipse_swt_graphics_Font which was broken with ScalingSwtFontRegistry)
 

**Reproduction Steps**
To use the legacy font registry instead of the default scaling registry, launch the application with the following system properties:

```
-Dswt.autoScale.updateOnRuntime=false
-Dswt.autoScale=false
-Dswt.fontRegistry=legacy
```
With these flags, the [LegacySWTFontRegistry](https://github.com/eclipse-platform/eclipse.platform.swt/compare/master...vi-eclipse:eclipse.platform.swt:arunjose696/312?expand=1#diff-b9861edb165cd679e5582ba8abf6d666ad53941b3fa4a0ba8c90fca0a675b8ffR134) will be used.

If Dswt.fontRegistry is  not set or Monitorscaling is enabled(current default behaviour) , SWT will default to using [ScalingSWTFontRegistry](https://github.com/eclipse-platform/eclipse.platform.swt/compare/master...vi-eclipse:eclipse.platform.swt:arunjose696/312?expand=1#diff-b9861edb165cd679e5582ba8abf6d666ad53941b3fa4a0ba8c90fca0a675b8ffR136).

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/ControlWin32Tests.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/ControlWin32Tests.java
@@ -32,7 +32,7 @@ import org.junit.jupiter.api.extension.*;
 class ControlWin32Tests {
 
 	@Test
-	public void testScaleFontCorrectlyInAutoScaleSzenario() {
+	public void testScaleFontCorrectlyInAutoScaleScenario() {
 		DPIUtil.setMonitorSpecificScaling(true);
 		Display display = Display.getDefault();
 
@@ -55,15 +55,37 @@ class ControlWin32Tests {
 	}
 
 	@Test
-	public void testDoNotScaleFontCorrectlyInNoAutoScaleSzenario() {
+	public void testScaleFontCorrectlyInNoAutoScaleScenario() {
 		DPIUtil.setMonitorSpecificScaling(false);
 		Display display = Display.getDefault();
 
 		assertFalse("Autoscale property is not set to false", display.isRescalingAtRuntime());
 		int scalingFactor = 2;
 		FontComparison fontComparison = updateFont(scalingFactor);
-		assertEquals("Font height in pixels is different when setting the same font again",
-				fontComparison.originalFontHeight, fontComparison.currentFontHeight);
+		assertEquals("Font height in pixels is not adjusted according to the scale factor",
+				fontComparison.originalFontHeight * scalingFactor, fontComparison.currentFontHeight);
+	}
+
+	@Test
+	public void testDoNotScaleFontInNoAutoScaleScenarioWithLegacyFontRegistry() {
+		DPIUtil.setMonitorSpecificScaling(false);
+		String originalValue = System.getProperty("swt.fontRegistry");
+		System.setProperty("swt.fontRegistry", "legacy");
+		try {
+			Display display = Display.getDefault();
+
+			assertFalse("Autoscale property is not set to false", display.isRescalingAtRuntime());
+			int scalingFactor = 2;
+			FontComparison fontComparison = updateFont(scalingFactor);
+			assertEquals("Font height in pixels is different when setting the same font again",
+					fontComparison.originalFontHeight, fontComparison.currentFontHeight);
+		} finally {
+			if (originalValue != null) {
+				System.setProperty("swt.fontRegistry", originalValue);
+			} else {
+				System.clearProperty("swt.fontRegistry");
+			}
+		}
 	}
 
 	@Test
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/LegacySWTFontRegistry.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/LegacySWTFontRegistry.java
@@ -19,19 +19,25 @@ import org.eclipse.swt.graphics.*;
 import org.eclipse.swt.internal.win32.*;
 
 /**
- * This class is used in the win32 implementation only to support
- * unscaled fonts in multiple DPI zoom levels.
+ * <p>
+ * Formerly {@code DefaultSWTFontRegistry}, this class is deprecated. Use {@code ScalingSWTFontRegistry} instead.
+ * To temporarily fall back to legacy font behavior ({@code LegacySWTFontRegistry})
+ * (e.g., if issues arise in existing RCP products), set the system property: {@code
+ * -Dswt.fontRegistry=legacy
+ * }
+ * </p>
  *
  * As this class is only intended to be used internally via {@code SWTFontProvider},
  * it should neither be instantiated nor referenced in a client application.
  * The behavior can change any time in a future release.
  */
-final class DefaultSWTFontRegistry implements SWTFontRegistry {
+@Deprecated(forRemoval= true, since= "2025-09")
+final class LegacySWTFontRegistry implements SWTFontRegistry {
 	private static FontData KEY_SYSTEM_FONTS = new FontData();
 	private Map<FontData, Font> fontsMap = new HashMap<>();
 	private Device device;
 
-	DefaultSWTFontRegistry(Device device) {
+	LegacySWTFontRegistry(Device device) {
 		this.device = device;
 	}
 
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/SWTFontProvider.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/SWTFontProvider.java
@@ -105,10 +105,15 @@ public class SWTFontProvider {
 		}
 	}
 
+	private static final String SWT_FONT_REGISTRY = "swt.fontRegistry";
+
+	@SuppressWarnings("removal")
 	private static SWTFontRegistry newFontRegistry(Device device) {
-		if (device instanceof Display display && display.isRescalingAtRuntime()) {
-			return new ScalingSWTFontRegistry(device);
+		if ("legacy".equalsIgnoreCase(System.getProperty(SWT_FONT_REGISTRY)) && device instanceof Display display
+				&& !display.isRescalingAtRuntime()) {
+			return new LegacySWTFontRegistry(device);
 		}
-		return new DefaultSWTFontRegistry(device);
+		return new ScalingSWTFontRegistry(device);
+
 	}
 }
--- a/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/internal/LegacySWTFontRegistryTests.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/internal/LegacySWTFontRegistryTests.java
@@ -23,15 +23,16 @@ import org.junit.jupiter.api.*;
 import org.junit.jupiter.api.extension.*;
 
 @ExtendWith(PlatformSpecificExecutionExtension.class)
-class DefaultSWTFontRegistryTests {
+class LegacySWTFontRegistryTests {
 	private static String TEST_FONT = "Helvetica";
 	private Display display;
 	private SWTFontRegistry fontRegistry;
 
+	@SuppressWarnings("removal")
 	@BeforeEach
 	public void setUp() {
 		this.display = Display.getDefault();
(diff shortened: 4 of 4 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/ScalingSWTFontRegistry.java
```diff
@@ -44,7 +44,9 @@ private Font getScaledFont(int zoom) {
 
 		private Font createAndCacheFont(int zoom) {
 			Font newFont = createFont(zoom);
-			customFontHandlesKeyMap.put(Font.win32_getHandle(newFont), this);
+			FontData clonedFontData = new FontData(newFont.getFontData()[0].toString());
```
arunjose696: cloning FontData should be replaced with new constructors if  #2128  gets merged

bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/ControlWin32Tests.java
```diff
@@ -44,15 +44,41 @@ public void testScaleFontCorrectlyInAutoScaleSzenario() {
...
+		try {
+			Display display = Display.getDefault();
+
+			assertFalse("Autoscale property is not set to false", display.isRescalingAtRuntime());
+			int scalingFactor = 2;
+			FontComparison fontComparison = updateFont(scalingFactor);
+			assertEquals("Font height in pixels is different when setting the same font again",
+					fontComparison.originalFontHeight, fontComparison.currentFontHeight);
+
+
```
akoch-yatta: you could get rid of the empty lines

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/LegacySWTFontRegistry.java
```diff
@@ -19,19 +19,25 @@
 import org.eclipse.swt.internal.win32.*;
 
 /**
- * This class is used in the win32 implementation only to support
- * unscaled fonts in multiple DPI zoom levels.
+ * <p>
+ * Formerly {@code DefaultSWTFontRegistry}, this class is deprecated. Use {@code ScalingSWTFontRegistry} instead.
+ * To temporarily fall back to legacy font behavior ({@code LegacySWTFontRegistry})
+ * (e.g., if issues arise in existing RCP products), set the system property: {@code
+ * -Dswt.fontRegistry=LEGACY
```
akoch-yatta: I think most system property values a lower case, so I would propose "legacy" instead of "LEGACY"

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/SWTFontProvider.java
```diff
@@ -105,10 +105,32 @@ public static void disposeFontRegistry(Device device) {
...
-			return new ScalingSWTFontRegistry(device);
+		String registryProp = System.getProperty(SWT_FONT_REGISTRY);
+		FontRegistryType fontRegistryValue = FontRegistryType.fromString(registryProp);
+		if (fontRegistryValue == FontRegistryType.LEGACY && device instanceof Display display
+				&& !display.isRescalingAtRuntime()) {
+			return new LegacySWTFontRegistry(device);
 		}
-		return new DefaultSWTFontRegistry(device);
+		return new ScalingSWTFontRegistry(device);
+
```
akoch-yatta: Usually I'm all for enum, but here I think this is overkill, the only think you want to check is `"legacy".equals(System.getProperty(SWT_FONT_REGISTRY))` + the rescaling at runtime check. in all other scenarios you want to use the scaling registry. Therfor I would remove the enum here and simply do a string check

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/ScalingSWTFontRegistry.java
akoch-yatta: This part is actually a fix for an issue independent of the refactoring. I would extract that one out + a separate test the checks that

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/LegacySWTFontRegistry.java
arunjose696: Done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/SWTFontProvider.java
arunjose696: Done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/ScalingSWTFontRegistry.java
arunjose696: Raised a new PR for this https://github.com/eclipse-platform/eclipse.platform.swt/pull/2237 , Not removing the changes in this PR as the CI runs would fail for this

bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/ControlWin32Tests.java
```diff
@@ -55,15 +55,38 @@ public void testSetFontWithMonitorSpecificScalingEnabled() {
 	}
 
 	@Test
-	public void testDoNotScaleFontCorrectlyInNoAutoScaleSzenario() {
+	public void testScaleFontCorrectlyInNoAutoScaleSzenario() {
```
fedejeanne: German -> English
```suggestion
	public void testScaleFontCorrectlyInNoAutoScaleScenario() {
```

bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/ControlWin32Tests.java
```diff
@@ -55,15 +55,38 @@ public void testSetFontWithMonitorSpecificScalingEnabled() {
...
 		DPIUtil.setMonitorSpecificScaling(false);
 		Display display = Display.getDefault();
 
-		assertFalse("Autoscale property is not set to false", display.isRescalingAtRuntime());
+		assertFalse("Autoscale property is not set to true", display.isRescalingAtRuntime());
```
fedejeanne: Revert this change. The message should show what is wrong, which is what it did before.

bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/ControlWin32Tests.java
```diff
@@ -55,15 +55,38 @@ public void testSetFontWithMonitorSpecificScalingEnabled() {
...
 		int scalingFactor = 2;
 		FontComparison fontComparison = updateFont(scalingFactor);
-		assertEquals("Font height in pixels is different when setting the same font again",
-				fontComparison.originalFontHeight, fontComparison.currentFontHeight);
+		assertEquals("Font height in pixels is not adjusted according to the scale factor",
+				fontComparison.originalFontHeight * scalingFactor, fontComparison.currentFontHeight);
+	}
+
+	@Test
+	public void testDoNotScaleFontCorrectlyInNoAutoScaleSzenarioWithLegacyFontRegistry() {
```
fedejeanne: Shorter and in English, please.

```suggestion
	public void testDoNotScaleFontInNoAutoScaleScenarioWithLegacyFontRegistry() {
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/SWTFontProvider.java
```diff
@@ -105,10 +105,14 @@ public static void disposeFontRegistry(Device device) {
 		}
 	}
 
+	private static final String SWT_FONT_REGISTRY = "swt.fontRegistry";
+
 	private static SWTFontRegistry newFontRegistry(Device device) {
-		if (device instanceof Display display && display.isRescalingAtRuntime()) {
-			return new ScalingSWTFontRegistry(device);
+		if ("legacy".equals(System.getProperty(SWT_FONT_REGISTRY)) && device instanceof Display display
```
fedejeanne: Just in case, compare ignoring case.

```suggestion
		if ("legacy".equalsIgnoreCase(System.getProperty(SWT_FONT_REGISTRY)) && device instanceof Display display
```

bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/ControlWin32Tests.java
```diff
@@ -55,15 +55,38 @@ public void testSetFontWithMonitorSpecificScalingEnabled() {
...
+			Display display = Display.getDefault();
+
+			assertFalse("Autoscale property is not set to false", display.isRescalingAtRuntime());
+			int scalingFactor = 2;
+			FontComparison fontComparison = updateFont(scalingFactor);
+			assertEquals("Font height in pixels is different when setting the same font again",
+					fontComparison.originalFontHeight, fontComparison.currentFontHeight);
+		}
+		finally {
+			if (originalValue != null) {
```
fedejeanne, line 83: Indentation looks wrong
```suggestion
		if (originalValue != null) {
```

fedejeanne (commented):
Just some minor comments

bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/ControlWin32Tests.java
arunjose696: Indentation has been done with eclipse format `ctrl+shift+f`

fedejeanne (commented):
@arunjose696 please add the necessary `@SuppressWarnings("removal")` so that the checks do not fail

arunjose696:
> @arunjose696 please add the necessary `@SuppressWarnings("removal")` so that the checks do not fail

Have added SuppressWarnings, to usages of deprecated class(LegacySWTFontRegistry), still see the warnings, @fedejeanne is it required to add annotations elsewhere?

fedejeanne:
Yes, you missed one spot: `bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/internal/LegacySWTFontRegistryTests.java`

You can find the new warnings when clicking in the failing check:

image

Then look for the _new_ issues (they are at the top of the list):

image

akurtakov:
This has caused warnings in I-build e.g. https://download.eclipse.org/eclipse/downloads/drops4/I20250625-0320/compilelogs/plugins/org.eclipse.swt.win32.win32.aarch64_3.131.0.v20250624-0912/test.html

arunjose696:
> This has caused warnings in I-build e.g. https://download.eclipse.org/eclipse/downloads/drops4/I20250625-0320/compilelogs/plugins/org.eclipse.swt.win32.win32.aarch64_3.131.0.v20250624-0912/test.html

This seems to be a false positive, have added necessary @SuppressWarnings for this function, dont see any warnings appearing in IDE as well.

fedejeanne:
It just occurs to me: could it be that the check is complaining about the _deprecation **without** removal_?

The IDE is smart enough to know that suppressing the "removal" notices also suppresses the "deprecation" notices, but I don't know about the check.

PR #2221, opened 2025-06-10
Refactor Image(Display, int, int) in Tests (No Ops)

ShahzaibIbrahim:
Replacing Image(Display, int, int) with Image(Display, ImageGcDrawer, int, int) in Tests. Basic Case

- [x] requires -> https://github.com/eclipse-platform/eclipse.platform.swt/pull/2229

### DIFF
--- a/tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/graphics/ImageWin32Tests.java
+++ b/tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/graphics/ImageWin32Tests.java
@@ -39,7 +39,8 @@ public class ImageWin32Tests {
 
 	@Test
 	public void testImageDataForDifferentFractionalZoomsShouldBeDifferent() {
-		Image image = new Image(display, 10, 10);
+		ImageGcDrawer noOpGcDrawer = (gc, height, width) -> {};
+		Image image = new Image(display, noOpGcDrawer, 10, 10);
 		int zoom1 = 125;
 		int zoom2 = 150;
 		ImageData imageDataAtZoom1 = image.getImageData(zoom1);
@@ -52,9 +53,10 @@ public class ImageWin32Tests {
 
 	@Test
 	public void testImageShouldHaveDimesionAsPerZoomLevel() {
+		ImageGcDrawer noOpGcDrawer = (gc, height, width) -> {};
 		int zoom = DPIUtil.getDeviceZoom();
 		int scalingFactor = 2;
-		Image image = new Image(display, 10, 10);
+		Image image = new Image(display, noOpGcDrawer, 10, 10);
 		try {
 			ImageData baseImageData = image.getImageData(zoom);
 			assertEquals("Width should equal the initial width on the same zoom", 10, baseImageData.width);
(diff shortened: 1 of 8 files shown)


tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/graphics/ImageWin32Tests.java
```diff
@@ -39,7 +39,8 @@ public void setUp() {
 
 	@Test
 	public void testImageDataForDifferentFractionalZoomsShouldBeDifferent() {
-		Image image = new Image(display, 10, 10);
+		ImageGcDrawer noOpGcDrawer =  (gc, height, width) -> {};
```
HeikoKlare: This line is repeatedly formatted incorrectly.
```suggestion
		ImageGcDrawer noOpGcDrawer = (gc, height, width) -> {};
```

HeikoKlare (changes requested):
This PR overlaps with https://github.com/eclipse-platform/eclipse.platform.swt/pull/2229. Please consider the comments I made there and please revise the PRs so that they do not change the same lines of code in the semantically same but syntactically different ways.

tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/graphics/ImageWin32Tests.java
ShahzaibIbrahim: Since there are overlapping changes, this should be looked at after https://github.com/eclipse-platform/eclipse.platform.swt/pull/2229 is merged.

fedejeanne: #2229 is now merged. Is there something else to do here or is this conversation resolved?

fedejeanne:
@ShahzaibIbrahim is there anything else to do here from your side or have all comments been addressed? I see that some conversations are still open.

Also, the test `org.eclipse.swt.tests.junit.Test_org_eclipse_swt_SWT.test_isLocal failed` failed in 2 consecutive runs. Can you look into that? If it's a flaky test, please report it.

ShahzaibIbrahim:
> @ShahzaibIbrahim is there anything else to do here from your side or have all comments been addressed? I see that some conversations are still open.
> 
> Also, the test `org.eclipse.swt.tests.junit.Test_org_eclipse_swt_SWT.test_isLocal failed` failed in 2 consecutive runs. Can you look into that? If it's a flaky test, please report it.

All the comments were addressed but it needs to be re-reviewed.

ShahzaibIbrahim:
@akurtakov the test test_isLocal fails for this PR only in the CI. Do you have any idea what is it about?

fedejeanne:
> @akurtakov the test test_isLocal fails for this PR only in the CI. Do you have any idea what is it about?

My educated guess is that this check fails because SWT was **not** built for Linux in this PR and the check basically says "you're not testing your changes, you're testing `master` right now".

But this is irrelevant since this PR doesn't change any production code that needs to be tested, it changes only the tests. So technically, it is **OK** to test `master` (more precisely: the latest official drop).

Am I seeing this right, @akurtakov ?

akurtakov:
I am traveling so I can't look into things this week. Please remind me next week if there is still a problem to look into it.

fedejeanne:
Thank you for fixing it, @akurtakov ! 

May I ask what was the trick? If you merely rebased on `master` then you must have magic hands because I tried that too and it didn't work... 😅

akurtakov:
I have just rebased it :) . My gut feeling is that this error appears in a period when a commit makes the build produce different artifact compared to last I-build or smth similar and it starts working after the next build is in I-build repo. I haven't looked whether that's the case so I could be totally wrong too.

fedejeanne:
Ok, but at least I have something to try the next time I see this check fail. Thank you for the hint!

PR #2224, opened 2025-06-11
[win32] Always use GC#isDisposed() for check

akoch-yatta:
This PR replaces the handle == 0 checks in GC by using isDisposed instead. It was done via search & replace


akoch-yatta:
> A simple cleanup that is obviously fine. I just wondered: should be take the chance and extract the actual check and error generation into a method `checkNonDisposed` or the like?

Makes sense. I actually thought about that, but didn't do it yet for some reason. I adapted it now

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -487,7 +487,7 @@ public void copyArea (Image image, int x, int y) {
 }
 
 void copyAreaInPixels(Image image, int x, int y) {
-	if (handle == 0) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
+	checkNonDisposed();
```
laeubi, line 490: I'm not an expert regarding englisch language, but should it be `checkNotDisposed`? (like in "check it is not disposed"). I would probably simply name it `checkDisposed` like [here](https://github.com/eclipse-equinox/equinox/blob/d3cc8d72910d11eaff49592a42a9ce21156120d8/bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/SubMonitor.java#L683).

PR #2226, opened 2025-06-11
[GTK] Reduce button.toggle padding on GTK 4

akurtakov:
Add css file for gtk4 and start accomodating content in it.


akurtakov:
Before: 
Screenshot From 2025-06-11 19-34-04
After:
Screenshot From 2025-06-11 19-34-29

PR #2189, opened 2025-05-28
Refactor Image(Display, int, int) in Tests

ShahzaibIbrahim:
Replacing Image(Display, int, int) with Image(Display, ImageGcDrawer, int, int) in Tests

PR #2229, opened 2025-06-12
Refactor Image(Display, int, int) in Tests (Automated Tests)

ShahzaibIbrahim:
Replacing Image(Display, int, int) with Image(Display, ImageGcDrawer, int, int) in Tests. Automated Test Case.

### DIFF
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_Image.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_Image.java
@@ -209,16 +209,17 @@ public void test_ConstructorLorg_eclipse_swt_graphics_DeviceLorg_eclipse_swt_gra
 	data = new ImageData(10, 10, 8, new PaletteData(0x30, 0x0C, 0x03));
 	// set red pixel at x=9, y=9
 	data.setPixel(9, 9, 0x30);
-	image = new Image(display, data);
-	Image gcImage = new Image(display, 10, 10);
-	GC gc = new GC(gcImage);
-	gc.drawImage(image, 0, 0);
+	final Image imageFromImageData = new Image(display, data);
+	ImageGcDrawer gcDrawer = (gc, width, height) -> {
+		gc.drawImage(imageFromImageData, 0, 0);
+	};
+	Image gcImage = new Image(display, gcDrawer, 10, 10);
 	ImageData gcImageData = gcImage.getImageData();
 	int redPixel = gcImageData.getPixel(9, 9);
 	assertEquals(getRealRGB(display.getSystemColor(SWT.COLOR_RED)), gcImageData.palette.getRGB(redPixel));
-	gc.dispose();
 	gcImage.dispose();
 	image.dispose();
+	imageFromImageData.dispose();
 }
 
 @Test
@@ -252,21 +253,23 @@ public void test_ConstructorLorg_eclipse_swt_graphics_DeviceLorg_eclipse_swt_gra
 	data6.setPixel(9, 9, 0x30);
 	data7 = new ImageData(10, 10, 1, new PaletteData(new RGB(0, 0, 0), new RGB(255, 255, 255)));
 	data7.setPixel(9, 9, 1);
-	image = new Image(display, data6, data7);
-	Image gcImage = new Image(display, 10, 10);
-	GC gc = new GC(gcImage);
+	Image image2 = new Image(display, data6, data7);
 	Color backgroundColor = display.getSystemColor(SWT.COLOR_BLUE);
-	gc.setBackground(backgroundColor);
-	gc.fillRectangle(0, 0, 10, 10);
-	gc.drawImage(image, 0, 0);
+	final ImageGcDrawer gcDrawer = (gc, width, height) -> {
+		gc.setBackground(backgroundColor);
+		gc.fillRectangle(0, 0, 10, 10);
+		gc.drawImage(image2, 0, 0);
+	};
+	Image gcImage = new Image(display, gcDrawer, 10, 10);
+
 	ImageData gcImageData = gcImage.getImageData();
 	int redPixel = gcImageData.getPixel(9, 9);
 	assertEquals(getRealRGB(display.getSystemColor(SWT.COLOR_RED)), gcImageData.palette.getRGB(redPixel));
 	int bluePixel = gcImageData.getPixel(0, 0);
 	assertEquals(getRealRGB(backgroundColor), gcImageData.palette.getRGB(bluePixel));
-	gc.dispose();
 	gcImage.dispose();
 	image.dispose();
+	image2.dispose();
 }
 
 @Test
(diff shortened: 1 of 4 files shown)


tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_TextLayout.java
```diff
@@ -1007,21 +1003,17 @@ public void test_bug568740_multilineTextStyle() {
...
-
 		layout.draw(gc, 0, 0);
-		return image;
-	} finally {
-		if (gc != null)
-			gc.dispose();
-	}
+	};
+	Image image = new Image(display, gcDrawer, rect.width, rect.height);
+	image.getImageData(); //this will ensure that the gc calls inside gcDrawer are executed
```
HeikoKlare: In my understanding, it's not about the GC calls but about other side effects (in this case, the `layout.draw` call). So this is more about ensuring that the GC drawer in general is executed for lazy generated images.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_TextLayout.java
```diff
@@ -974,8 +966,14 @@ public void test_bug568740_multilineTextStyle() {
...
-		int offset = 10;
-		layout.draw(gc, offset, offset);
+		font = new Font(display, SwtTestUtil.testFontName, 16, SWT.NORMAL);
+		final ImageGcDrawer gcDrawer = (gc, width, height) -> {
+			gc.setBackground(display.getSystemColor(SWT.COLOR_WHITE));
+			gc.fillRectangle(0, 0, width, height);
+			gc.setAntialias(SWT.OFF); // aa can change colors and break the test in worst case
+			layout.draw(gc, offset, offset);
+		};
+		image = new Image(display, gcDrawer, 200, 100);
```
HeikoKlare, line 976: This seems to be affected by the same issue as reported here: https://github.com/eclipse-platform/eclipse.platform.swt/pull/2189#discussion_r2135367536
The drawer will not be executed, so the side effect on the layout may not be visible in subsequent code, will it?

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_Image.java
```diff
@@ -700,21 +703,21 @@ public void test_getImageData_200() {
...
-	Image image1 = new Image(display, bounds.width, bounds.height);
+	Image image1 = new Image(display, (gc, width, height) -> {}, bounds.width, bounds.height);
 	image1.dispose();
 	SWTException e = assertThrows(SWTException.class, () -> image1.getImageData(zoom));
 	assertSWTProblem("Incorrect exception thrown for disposed image", SWT.ERROR_GRAPHIC_DISPOSED, e);
 
 	Image image;
 	// creates bitmap image and compare size of imageData
-	image = new Image(display, bounds.width, bounds.height);
+	image = new Image(display, (gc, width, height) -> {}, bounds.width, bounds.height);
```
HeikoKlare: This should be implemented as an additional test scenario and not replace the existing test for the constructor accepting just an image size. Otherwise we will lose test coverage of the existing functionality. Also see https://github.com/eclipse-platform/eclipse.platform.swt/pull/2189#discussion_r2112267790.
So please check again in which cases we may need to add an additional test scenario instead of adapting the existing one.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_Image.java
```diff
@@ -208,16 +208,17 @@ public void test_ConstructorLorg_eclipse_swt_graphics_DeviceLorg_eclipse_swt_gra
 	data = new ImageData(10, 10, 8, new PaletteData(0x30, 0x0C, 0x03));
 	// set red pixel at x=9, y=9
 	data.setPixel(9, 9, 0x30);
-	image = new Image(display, data);
-	Image gcImage = new Image(display, 10, 10);
-	GC gc = new GC(gcImage);
-	gc.drawImage(image, 0, 0);
+	final Image image2 = new Image(display, data);
```
HeikoKlare: Would be nice to use a more descriptive name than `image2`.

HeikoKlare (changes requested):

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_TextLayout.java
ShahzaibIbrahim: The drawer will be executed as soon as getBounds is called in line:980. I tested it.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_TextLayout.java
ShahzaibIbrahim: That's why I have called image.getImageData(); right after the image is created. This will ensure the gcDrawer  is executed and so is layout.draw(). I have tested this.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_Image.java
ShahzaibIbrahim: I have added the additional test for Image created with GCImage and keeping the existing functionality tests as is.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_TextLayout.java
HeikoKlare: Yes, that's correct. My comment was made with respect to the comment on the `getImageData()` call, which appear misleading to me.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_TextLayout.java
HeikoKlare: Might be, but before that methods on layout are executes (such as `layout.getLineBounds(0)` in line 978). Assuming that that call is uneffected by the `layout.draw()` call is a risky assumption that may easily break in the future (if it's not already broken now).

ShahzaibIbrahim: You are right, It's already too late until layout.draw is executed. I will call the getImageData right after the image is created to ensure layout.draw is called before the layout.getLineBounds(0)

HeikoKlare:
The links in the review dismiss is wrong. I actually meant this comment:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2229#discussion_r2149518192

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_TextLayout.java
HeikoKlare: I have updated the comment.

PR #2230, opened 2025-06-12
Refactor Image(Display, int, int) in Tests (Manual Test)

ShahzaibIbrahim:
Replacing Image(Display, int, int) with Image(Display, ImageGcDrawer, int, int) in Tests. Manual Tests.

### DIFF
--- a/tests/org.eclipse.swt.tests.win32/ManualTests/org/eclipse/swt/tests/win32/snippets/Bug560546_GC_drawString_vs_GC_drawText.java
+++ b/tests/org.eclipse.swt.tests.win32/ManualTests/org/eclipse/swt/tests/win32/snippets/Bug560546_GC_drawString_vs_GC_drawText.java
@@ -13,9 +13,16 @@
  *******************************************************************************/
 package org.eclipse.swt.tests.win32.snippets;
 
-import org.eclipse.swt.*;
-import org.eclipse.swt.graphics.*;
-import org.eclipse.swt.widgets.*;
+import org.eclipse.swt.SWT;
+import org.eclipse.swt.graphics.Font;
+import org.eclipse.swt.graphics.GC;
+import org.eclipse.swt.graphics.Image;
+import org.eclipse.swt.graphics.ImageGcDrawer;
+import org.eclipse.swt.widgets.Button;
+import org.eclipse.swt.widgets.Canvas;
+import org.eclipse.swt.widgets.Display;
+import org.eclipse.swt.widgets.MessageBox;
+import org.eclipse.swt.widgets.Shell;
 
 public class Bug560546_GC_drawString_vs_GC_drawText {
 	static Font fontSegoeUI;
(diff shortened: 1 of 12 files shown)


tests/org.eclipse.swt.tests.win32/ManualTests/org/eclipse/swt/tests/win32/snippets/Bug563475_DarkDisabledTable.java
```diff
@@ -52,14 +52,13 @@ static void setColors(Control control, Color backColor, Color foreColor) {
...
-		gc.setBackground(display.getSystemColor(SWT.COLOR_BLUE));
-		gc.fillRectangle(6, 6, 4, 4);
-		gc.dispose();
+		final ImageGcDrawer imageGcDrawer = (gc, width, height) -> {
+			gc.setBackground(transparentColor);
+			gc.fillRectangle(0, 0, width, height);
+			gc.setBackground(display.getSystemColor(SWT.COLOR_BLUE));
+			gc.fillRectangle(6, 6, 4, 4);
+		};
+		Image image = new Image(display, imageGcDrawer, 16, 16);
```
HeikoKlare, line 61: This is the same change as in https://github.com/eclipse-platform/eclipse.platform.swt/pull/2189. So this is affected by the same issue as reported here:
https://github.com/eclipse-platform/eclipse.platform.swt/pull/2189#pullrequestreview-2909451863

HeikoKlare (changes requested):

tests/org.eclipse.swt.tests.win32/ManualTests/org/eclipse/swt/tests/win32/snippets/Bug563475_DarkDisabledTable.java
ShahzaibIbrahim: With the latest changes, it should work.

tests/org.eclipse.swt.tests.win32/ManualTests/org/eclipse/swt/tests/win32/snippets/Bug560546_GC_drawString_vs_GC_drawText.java
```diff
@@ -73,39 +80,37 @@ void function(GC gc, int x, int y, String string) {
...
+						}
 					}
 				}
 			}
-		}
-
-		gc.dispose();
+			MessageBox messageBox = new MessageBox(shell);
+			messageBox.setMessage(report);
+			messageBox.open();
```
HeikoKlare: I don't think opening a message box should happen inside the drawer (i.e., while instantiating the image data), but it should stay where it was before.

HeikoKlare (commented):
All adaptations are fine and the manual tests still work as before. There is only one change that needs to be revised.

HeikoKlare (approved):

PR #2233, opened 2025-06-13
DPI_CHANGED rough performance analysis

amartya4256:
Composite:handleDPIChange
Composite:setSize,:sendResize
layout()
sendBoundsInPixels()
Gridlayout.layout
Image:win32_getHandle() - maybe all resources? (20%)


HeikoKlare:
Several improvements have been implemented, in particular by making the update (more) asynchronous. If we find further things need to follow up on this, we will start with another analysis. So this is initial analysis code is not needed anymore (in particular not as a PR). Thus closing.

PR #2234, opened 2025-06-13
[GTK] Button images should be GtkImage not GtkPicture

akurtakov:
From Gtk documentation "GtkImage displays its image as an icon, with a size that is determined by the application. See GtkPicture if you want to show an image at is actual size."


akurtakov:
Before:
Screenshot From 2025-06-13 17-08-16
After:
Screenshot From 2025-06-13 17-08-36

PR #2236, opened 2025-06-15
[macOS] Initialize image data from ImageGcDrawer only at device zoom

HeikoKlare:
An Image based on an ImageGcDrawer currently initializes two image representations for 100% and 200% zoom, so that an appropriate representation for each of the two zoom levels can be retrieved. This, however, leads to issues when creating a GC on an Image created in that way, as only one of the two representations (the one fitting to the current device zoom) will be modified by the GC. If later the representation with the other zoom is used, it will not contain the changes performed by the GC.

One such scenario is starting an application at a monitor with one zoom (used as the device zoom throughout the application lifecycle then) and moving the shell to a monitor with a different zoom. An image will then be modified only for the zoom of the first monitor, thus if using it inside a control, it will yield the unmodified version on the second monitor.

To avoid this issue, this change ensures that an Image based on an ImageGcDrawer is only initialized for the device zoom.

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/3039

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
@@ -572,7 +572,7 @@ public Image(Device device, ImageData data) {
 	NSAutoreleasePool pool = null;
 	if (!NSThread.isMainThread()) pool = (NSAutoreleasePool) new NSAutoreleasePool().alloc().init();
 	try {
-		init(data);
+		init(data, 100);
 		init();
 	} finally {
 		if (pool != null) pool.release();
@@ -623,7 +623,7 @@ public Image(Device device, ImageData source, ImageData mask) {
 		ImageData image = new ImageData(source.width, source.height, source.depth, source.palette, source.scanlinePad, source.data);
 		image.maskPad = mask.scanlinePad;
 		image.maskData = mask.data;
-		init(image);
+		init(image, 100);
 	} finally {
 		if (pool != null) pool.release();
 	}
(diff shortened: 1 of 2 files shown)


HeikoKlare:
@akoch-yatta can you please review the fix?

@BeckerWdf may I ask you as a frequent Mac user (if I am not mistaken) to test if this solves https://github.com/eclipse-platform/eclipse.platform.ui/issues/3039?

BeckerWdf:
> @BeckerWdf may I ask you as a frequent Mac user (if I am not mistaken) to test if this solves [eclipse-platform/eclipse.platform.ui#3039](https://github.com/eclipse-platform/eclipse.platform.ui/issues/3039)?

Sure I can try do test this myself. So what are the exact step to reproduce?

HeikoKlare:
Thank you, @BeckerWdf!

> So what are the exact step to reproduce?

Prerequisite: 2 monitors with different zooms (e.g., MacBook with 200% (Retina) display and external monitor at 100%).

Then do the following:
- Start Eclipse
- Open an editor with any file (e.g. a Java file)
- Ensure that you can scroll in the file, i.e., that the file has more lines than can be displayed on one page. You can just add empty lines if the document is too short.
- Move the window to the other monitor.
- Scroll within the editor.

Without the fix you should see that the scrollbar is stuck (it always shows the same initial line numbers like in the video here: https://github.com/eclipse-platform/eclipse.platform.ui/issues/3039#issue-3140781459). With the fix you should see that the scrollbar works properly.

BeckerWdf:
Stupid question: How do I set the zoom to 100 / 200 on macOS?

HeikoKlare:
That's actually a very good question. In my case, I just know that I have 200% on the MacBook screen and 100% on external monitor (it's obvious from sharpness and I once found the according device zoom values in SWT when debugging), but I never tried to identify or change that at the system level.

I would expect that you have a Retina display (200%) on your MacBook anyway. Regarding the external monitor, it looks as if you can "implicitly" set the zoom via the resolution. This is how it looks for me:
image
When I switch to the resolution marked as "HiDPI", 200% zoom is used, otherwise its 100%. But in case you have a high-resolution external monitor, it might be that MacOS does not even provide you the option to use a non-HiDPI-resolution?

BeckerWdf:
Ok i got it.
My external monitor has a native resolution of 3840x2160. In the system preferences 1920x1080 is used (and marked as "default". So this is have the native resolution or 200% zoom.
If i set the used resolution to the native one this then seems to be the 100% resolution. And with this setting I can reproduce the issue. The line numbers scroll correctly on the monitor on which eclipse was started by do not scroll at all on the other monitor.

BeckerWdf:
And the good news is: Your PR fixes the issue.

BeckerWdf (approved):

HeikoKlare:
Thank you for testing it!

PR #2237, opened 2025-06-16
Cache all Fonts in ScalingSWTFontRegistry on creation

arunjose696:
Previously, system fonts were not cached during their creation in ScalingSWTFontRegistry. This change ensures that  system fonts, are now cached in fontsKeyMap

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/ScalingSWTFontRegistry.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/ScalingSWTFontRegistry.java
@@ -44,7 +44,9 @@ final class ScalingSWTFontRegistry implements SWTFontRegistry {
 
 		private Font createAndCacheFont(int zoom) {
 			Font newFont = createFont(zoom);
-			customFontHandlesKeyMap.put(Font.win32_getHandle(newFont), this);
+			FontData clonedFontData = new FontData(newFont.getFontData()[0]);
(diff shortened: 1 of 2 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/ScalingSWTFontRegistry.java
```diff
@@ -44,7 +44,9 @@ private Font getScaledFont(int zoom) {
 
 		private Font createAndCacheFont(int zoom) {
 			Font newFont = createFont(zoom);
-			customFontHandlesKeyMap.put(Font.win32_getHandle(newFont), this);
+			FontData clonedFontData = new FontData(newFont.getFontData()[0].toString());
```
HeikoKlare: You may adapt this to the new copy constructor since https://github.com/eclipse-platform/eclipse.platform.swt/pull/2128 is merged now.

HeikoKlare (commented):
Overall, the change looks fine to me. The FontData copy can now be improved with the just added copy constructor.

Still I would like to have @akoch-yatta have another at look at this change as it affects quite central, crucial code.

HeikoKlare:
Taking another look, I understood again that this is really only about adding the system font to the cache based on font data.
I tested around with the change, also at different zooms. Since it affects the system font, issues with it should be easily perceivable.

Thus I merge this to proceed with the work based on top of it. Still would be nice if you can double-check once you are back, @akoch-yatta.

PR #2238, opened 2025-06-16
Use the existing image to initialize the GC in ImageGcDrawerWrapper

ShahzaibIbrahim:
In ImageGCDrawerWrapper#newImageHandle, we create a new image with existing data to initialize the GC, instead we could use the existing image object and the base handle should be created as it is done in PlainImageDataProvider.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
@@ -380,14 +380,6 @@ public Image(Device device, ImageData data) {
 	this.device.registerResourceWithZoomSupport(this);
 }
 
-private Image(Device device, ImageData data, int zoom) {
-	super(device);
-	if (data == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
-	this.imageProvider = new PlainImageDataProviderWrapper(data, zoom);
-	init();
-	this.device.registerResourceWithZoomSupport(this);
-}
-
 /**
  * Constructs an instance of this class, whose type is
  * <code>SWT.ICON</code>, from the two given <code>ImageData</code>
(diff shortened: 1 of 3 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2541,28 +2545,25 @@ ImageData newImageData(int zoom) {
...
 			resultData.alphaData = new byte [scaledWidth * scaledHeight];
-			image = new Image(device, resultData, zoom);
+			init(resultData, zoom);
 		} else {
-			image = new Image(device, width, height);
-		}
-		GC gc = new GC(new DrawableWrapper(image, zoom), gcStyle);
+			createBaseHandle(zoom, width, height);
+		};
+		GC gc = new GC(new DrawableWrapper(Image.this, zoom), gcStyle);
```
HeikoKlare: Can't we then remove the `DrawableWrapper` and just call?
```suggestion
		GC gc = new GC(Image.this, gcStyle);
```

HeikoKlare (commented):
I've tested the behavior with the LineNumberRuler and that seems to be work fine with different zooms. But with this change we will see the following warnings when strict checks are enabled, which need to be resolved:
```
***WARNING: Images with handles created for multiple zooms should not be modified.  It should be created with an ImageGcDrawer (see SWT Snippet 384).
```

HeikoKlare:
Is this something we can only do for the Windows implementation or may we adapt all the MacOS and Linux implementations of `Image` as well?

ShahzaibIbrahim:
> ***WARNING: Images with handles created for multiple zooms should not be modified.  It should be created with an ImageGcDrawer (see SWT Snippet 384).

This error should go away once [this](https://github.com/eclipse-platform/eclipse.platform.swt/pull/2243) PR is merged.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
ShahzaibIbrahim: There's a scenario where it's needed when another GC can be called over this image from the outside. I am not 100% of the scenario but maybe @akoch-yatta can explain it bit better on why it was used in the first place.

ShahzaibIbrahim:
> Is this something we can only do for the Windows implementation or may we adapt all the MacOS and Linux implementations of `Image` as well?

Yes, it make sense to adapt the MacOS and Linux implementation of Image. But since I can't test the changes, I will wait until @fedejeanne is back and can test those changes with him.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2541,28 +2545,29 @@ ImageData newImageData(int zoom) {
...
 			drawer.drawOn(gc, width, height);
-			ImageData imageData = image.getImageMetadata(zoom).getImageData();
+			ImageData imageData = Image.this.getImageMetadata(zoom).getImageData();
 			drawer.postProcess(imageData);
-			ImageData newData = adaptImageDataIfDisabledOrGray(imageData);
-			return init(newData, zoom);
+			if(imageData.data != null) {
+				zoomLevelToImageHandle.get(zoom).destroy();
+				init(imageData, zoom);
+			}
```
ShahzaibIbrahim: Here I want initialize the imageData only if postProcess is called. That's why I have a default implementation that nullify the imageData in order to know here if it was not implemented. I know it's not a clean solution and I am open to hear suggestion. 

My suggestions:

1. Have some field to track if postProcess was called. 
2. Always call init regardless of postProcess calling. 

@HeikoKlare

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
HeikoKlare: Seems like it would make sense to add a regression test for that scenario.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
HeikoKlare: Do I understand correctly that you want to identify if the passed `ImageGcDrawer` has a custom implementation of `postProcess` because if that's not the case there is no need to reinitialize the image?
I have to admit that I have no really good solution here. We may choose any more or less hacky solution, such as using reflection for identifying if the actual drawer has overwritten the method, using some other flag such as a global variable or some field specifically for that case inside `ImageData`, but that's all more or less bad.
It would obviously have been good if the `postProcess()` method was introduced with a `boolean` return type, such that one could identify if anything was done, but that ship has sailed. And there might be the biggest chance we have: Hannes started to rework the `ImageGcDrawer` (https://github.com/eclipse-platform/eclipse.platform.swt/pull/1985), so if that work was finalized, we could adapt the `postProcess` method to return whether it actually did something as well. That would not only improve the design of the `ImageGcDrawer` but also improve the performance of its usage inside `Image`. But since that interface has now been there for multiple releases and adaptations of consumer code to use it already started, I am not sure if we can still reasonably do that change or if we have to accept that this ship has sailed as well. @HannesWell what do you think about the work you started on `ImageGcDrawer`. Does it make sense to finalize it?

ShahzaibIbrahim: We thought about adding some variable to know if `postProcess` has been called but again it means the introduction of api field which we wanted to avoid and yes, you understood it correctly. We don't want to reinitialize the image if there is no postProcessing

HeikoKlare: The question is where to place such a variable. Extending `ImageData` for this case does not really make sense and the `ImageGcDrawer` is an interface with no state at all. Since there is nothing else involved in the call, you could only set some global state variable.

ShahzaibIbrahim:
Closing this PR as removing temporary images does not reduce resource usage; handles still need to be created and disposed. A larger refactoring would be required to address this properly. More info here: https://github.com/vi-eclipse/Eclipse-Platform/issues/328#issuecomment-3328264916

PR #2241, opened 2025-06-16
Fit Table Tooltip in a single monitor #62

amartya4256:
This PR contributes to the fitting of the table tooltip inside a
single monitor if it spans over multiple monitor to avoid any
infinite loop because of rescaling triggered by any DPI_CHANGED events.

Adapted from https://github.com/eclipse-platform/eclipse.platform.swt/pull/2142

contributes to
https://github.com/eclipse-platform/eclipse.platform.swt/issues/128

### How to test
First we need to create a table with narrow columns.

* Go to this method: org.eclipse.swt.examples.controlexample.TableTab.packColumns()
* Change it with this:
```
void packColumns () {
		int columnCount = table1.getColumnCount();
		for (int i = 0; i < columnCount; i++) {
			TableColumn tableColumn = table1.getColumn(i);
			tableColumn.setWidth(8);
//			tableColumn.pack();
		}
	}
```

* Launch ControlExample and go to table tab and position the shell in such a way that the table columns overlap on multiple monitors. 
* Hover over the overlappting columns and the tooltip must be positioned in the monitor where the cursor is contained, as shown in the picture.

image


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Table.java
```diff
@@ -7158,82 +7158,87 @@ LRESULT wmNotifyToolTip (NMHDR hdr, long wParam, long lParam) {
 		case OS.TTN_SHOW: {
 			LRESULT result = super.wmNotify (hdr, wParam, lParam);
 			if (result != null) return result;
-			if (hdr.code != OS.TTN_SHOW) tipRequested = true;
-			long code = callWindowProc (handle, OS.WM_NOTIFY, wParam, lParam);
-			if (hdr.code != OS.TTN_SHOW) tipRequested = false;
```
HeikoKlare, line 7163: Why can this code be removed? It will also result in the `code` returned by this method to usually be null (in case we are not in the `customToolTip` case with `TTN_SHOW`.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Table.java
```diff
@@ -7158,82 +7158,87 @@ LRESULT wmNotifyToolTip (NMHDR hdr, long wParam, long lParam) {
...
-								OS.MoveMemory (lpnmtdi.lpszText, new char [1], 2);
-								OS.MoveMemory (lParam, lpnmtdi, NMTTDISPINFO.sizeof);
-							}
-							RECT cellRect = item.getBounds (pinfo.iItem, pinfo.iSubItem, true, true, true, true, hDC);
-							RECT clientRect = new RECT ();
-							OS.GetClientRect (handle, clientRect);
-							if (itemRect.right > cellRect.right || itemRect.right > clientRect.right) {
-								//TEMPORARY CODE
-								String string = " ";
+			if (isCustomToolTip () || hdr.code == OS.TTN_SHOW) {
```
HeikoKlare: And why is the check for `TTN_SHOW` added here?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Table.java
```diff
@@ -7158,82 +7158,87 @@ LRESULT wmNotifyToolTip (NMHDR hdr, long wParam, long lParam) {
...
-								if (string != null) {
-									Shell shell = getShell ();
-									char [] chars = new char [string.length () + 1];
-									string.getChars (0, string.length (), chars, 0);
-									shell.setToolTipText (lpnmtdi, chars);
-									OS.MoveMemory (lParam, lpnmtdi, NMTTDISPINFO.sizeof);
-								}
-							}
-						}
+					if (string != null) {
```
HeikoKlare: I think when the code is reformatted anyway, we should take the chance to remove this commented out code and the unnecessary if (it's been like this for more than 15 years, so no one will miss it).

HeikoKlare (commented):
The change seems to work fine. But I have to admit that I don't know what the actual changes are supposed to do. I see the expected coordinates adjustment for the tooltip, but there are some further changes around, which I do not understand at all (I have added comments to those lines). Can you please elaborate on why these changes are necessary and what they do?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Table.java
amartya4256: It was made corresponding to this bug: https://bugs.eclipse.org/bugs/show_bug.cgi?id=178044
Commit id: ec9593cc82c5a72b6479eac5f3bdaf617409cb43
This was made for rollovers in 2007 but seems to have been fixed for the newer versions of windows. With the block present, the code goes in an infinite loop at 7162 and gets infinite TTN_GETDISPINFO messages, when the tooltip is by default over 2 monitors.
Thus, this could be removed now.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Table.java
amartya4256: We need to make sure we always position the tooltip regardless of if it is custom or not
Can also test it by removing the if block here. I don't remember anymore but I check quickly.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Table.java
amartya4256: sure

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Table.java
amartya4256: Yes resolved

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Table.java
amartya4256: The current windows works fine with rollover tooltips withut this block as well, however this block makes it worse when the tooltip is in the middle of 2 screens. You can try running the code with this block present and try to position your tooltip in the middle of 2 monitors.

HeikoKlare: I have to admit that I do not fully understand the assessment. The code was not introduced for the mentioned bug and in the mentioned commit. The commit only moved that code to a different place, such that it is also executed for the `TTN_SHOW` event. The commit that originally introduced the code is from 2005: https://github.com/eclipse-platform/eclipse.platform.swt/commit/dd05f312473477c63077f519775dd1f377f4a002. So the question remains why the code is now _completely_ unnecessary (not only for `TTN_SHOW` but only for the `TTN_GETDISPINFO` event). 

Just wondering: is the reaction to the `TTN_GETDISPINFO` still necessary at all? I tested with the `ControlExample` and there it does not make any difference whether I keep the switch-case for `TTN_GETDISPINFO` or whether I remove it (even though those events are actually fired). But since I don't know what this is necessary for, I cannot assess if there are remaining cases (maybe with very specific configurations) where it is needed.

In general, I currently still lack understanding of what exactly happens here. I can confirm that the change works as expected in all cases I have tested, but likely I don't know about all the use cases this complex code has been written for. So I am bit concerned that we break something when just removing some code. On the other hand, the code obviously leads to an issue (namely an inifite execution loop), which we definitely need to resolve. So it might lead to the trade-off that we need to accept that we might miss some use case here that we do not understand but therefor resolve a known bug.

Thus, my proposal:
- Please check again what this code is used for (as, like I said above, the existing assessment is not comprehensible for me)
- Have the change reviewed and tested by @fedejeanne, such that he can hopefully merge it

amartya4256: There's no explanation in the commit https://github.com/eclipse-platform/eclipse.platform.swt/commit/dd05f312473477c63077f519775dd1f377f4a002, however it relates to the bug report mentioned I suppose. (https://bugs.eclipse.org/bugs/show_bug.cgi?id=82739)

There's a snippet in the report and I tried that out and found no issues. I believe the block of code: 

```
tipRequested = true;Add commentMore actions
int code = callWindowProc (handle, OS.WM_NOTIFY, wParam, lParam);
tipRequested = false;
```
Doesn't have any visual effect anymore. 

Anyway it's hard to assess what it was introduced for but I gathered that the tooltip works similar to how it is implemented for Tree. Without having much changes in the original implementation, I tried to bring it as close as possible to Tree. However, I assessed the bug report https://bugs.eclipse.org/bugs/show_bug.cgi?id=82739 and the other bug report which was its duplicate (https://bugs.eclipse.org/bugs/show_bug.cgi?id=83703) and it doesn't seem to hamper the table tooltips in externalize stings view, the custom tooltip on the table in the control which comes using (ctrl+3) and also the snippets. These fixes were done for XP and Vista as it seems but I have tested them with the current windows 11 and my PR doesn't change any behaviour at all for them.

fedejeanne: If I test this PR with the snippet provided in https://bugs.eclipse.org/bugs/show_bug.cgi?id=82739#c1, I can see that this PR breaks the so called "Rollovers" (a special kind of tooltip, see: https://bugs.eclipse.org/bugs/show_bug.cgi?id=178044#c12).

So this PR breaks a use case. Please fix.

fedejeanne (commented):
This PR breaks a use case: see my comment in

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Table.java
fedejeanne: Here are the screenshots.

On `master`, I see the rollover when I hover over the second column and it looks fine.
image


With this PR, I don't see the text in the rollover:
image

I'm on **Windows 10**

amartya4256: Oh that was because of my recent commit. GETDISPINFO block in positionTooltip is only supposed to be executed in case of isCustomToolTip() == true since it tends to clean the default text on the tooltip for custom draw.
I have pushed the changes and tested the snippet again. It should be fine now.

fedejeanne: Thank you, that fixed the issue.

I went through the mentioned commits and issues and I also couldn't find any other use case which required the "fix" you are talking about so I assume this code is not necessary in newer Windows versions. 

IMO this code is not necessary anymore so I'll go ahead and merge this PR. We can keep an eye open just in case the missing use case is revealed and this change actually breaks something.

fedejeanne (approved):
It looks good now. Thank you @amartya4256 !

fedejeanne:
Failing test is unrelated: https://github.com/eclipse-platform/eclipse.platform.swt/issues/2098

PR #2243, opened 2025-06-17
[win32] Apply Operation pattern on GC

akoch-yatta:
This PR provides a minimal implementation to apply the operation pattern on GC that is used on most other resources to re-create resources on zoom changes. A difference in this implementation is that there is always exactly one OS handle per GC. On zoom change the existing handle is cleaned up a created for the new zoom. There is no use case that requires managing of multiple handles per GC.

This PR does not change any existing behavior. It encapsulated all actions inside the GC into operations. A different PR will utilize this to trigger a re-creation of an existing GC for a different zoom level.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
@@ -480,16 +480,34 @@ void checkGC(int mask) {
  * </ul>
  */
 public void copyArea (Image image, int x, int y) {
-	int deviceZoom = getZoom();
-	x = DPIUtil.scaleUp(drawable, x, deviceZoom);
-	y = DPIUtil.scaleUp(drawable, y, deviceZoom);
-	copyAreaInPixels(image, x, y);
-}
-
-void copyAreaInPixels(Image image, int x, int y) {
 	checkNonDisposed();
 	if (image == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
 	if (image.type != SWT.BITMAP || image.isDisposed()) SWT.error(SWT.ERROR_INVALID_ARGUMENT);
+	applyOperationForExistingHandle(new CopyAreaToImageOperation(image, x, y));
+}
+
+private class CopyAreaToImageOperation extends Operation {
+	private final Image image;
+	private final int x;
+	private final int y;
+
+	CopyAreaToImageOperation(Image image, int x, int y) {
+		this.image = image;
+		this.x = x;
+		this.y = y;
+	}
+
+	@Override
+	void apply() {
+		int zoom = getZoom();
+		int scaledX = DPIUtil.scaleUp(drawable, this.x, zoom);
+		int scaledY = DPIUtil.scaleUp(drawable, this.y, zoom);
+		copyAreaInPixels(this.image, scaledX, scaledY);
+	}
+}
+
+private void copyAreaInPixels(Image image, int x, int y) {
+	if (image.isDisposed()) SWT.error(SWT.ERROR_INVALID_ARGUMENT);
 	/* Copy the bitmap area */
 	Rectangle rect = image.getBounds(getZoom());
 	long memHdc = OS.CreateCompatibleDC(handle);
@@ -515,6 +533,7 @@ void copyAreaInPixels(Image image, int x, int y) {
  * </ul>
  */
 public void copyArea (int srcX, int srcY, int width, int height, int destX, int destY) {
+	checkNonDisposed();
 	copyArea (srcX, srcY, width, height, destX, destY, true);
 }
 
@@ -537,13 +556,31 @@ public void copyArea (int srcX, int srcY, int width, int height, int destX, int 
  * @since 3.1
  */
 public void copyArea (int srcX, int srcY, int width, int height, int destX, int destY, boolean paint) {
-	int zoom = getZoom();
-	Rectangle sourceRect = DPIUtil.scaleUp(drawable, new Rectangle(srcX, srcY, width, height), zoom);
-	Rectangle destRect = DPIUtil.scaleUp(drawable, new Rectangle(destX, destY, width, height), zoom);
-	copyAreaInPixels(sourceRect.x, sourceRect.y, sourceRect.width, sourceRect.height, destRect.x, destRect.y, paint);
+	checkNonDisposed();
+	applyOperationForExistingHandle(new CopyAreaOperation(new Rectangle(srcX, srcY, width, height), new Rectangle(destX, destY, width, height), paint));
 }
 
-void copyAreaInPixels(int srcX, int srcY, int width, int height, int destX, int destY, boolean paint) {
+private class CopyAreaOperation extends Operation {
+	private final Rectangle source;
+	private final Rectangle destination;
+	private final boolean paint;
+
+	CopyAreaOperation(Rectangle source, Rectangle destination, boolean paint) {
+		this.source = source;
+		this.destination = destination;
+		this.paint = paint;
+	}
+
+	@Override
+	void apply() {
+		int zoom = getZoom();
+		Rectangle sourceRect = DPIUtil.scaleUp(drawable, source, zoom);
+		Rectangle destRect = DPIUtil.scaleUp(drawable, destination, zoom);
+		copyAreaInPixels(sourceRect.x, sourceRect.y, sourceRect.width, sourceRect.height, destRect.x, destRect.y, paint);
+	}
+}
+
+private void copyAreaInPixels(int srcX, int srcY, int width, int height, int destX, int destY, boolean paint) {
 	checkNonDisposed();
 	long hwnd = data.hwnd;
 	if (hwnd == 0) {
@@ -562,6 +599,7 @@ void copyAreaInPixels(int srcX, int srcY, int width, int height, int destX, int 
 		OS.ScrollWindowEx(hwnd, destX - srcX, destY - srcY, lprcScroll, lprcClip, 0, null, flags);
 	}
 }
+
 static long createGdipFont(long hDC, long hFont, long graphics, long fontCollection, long [] outFamily, long[] outFont) {
 	long font = Gdip.Font_new(hDC, hFont);
 	if (font == 0) SWT.error(SWT.ERROR_NO_HANDLES);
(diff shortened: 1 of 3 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -5187,8 +5839,17 @@ private static int sin(int angle, int length) {
 	return (int)(Math.sin(angle * (Math.PI/180)) * length);
 }
 
-private int getZoom() {
+int getZoom() {
 	return DPIUtil.getZoomForAutoscaleProperty(data.nativeZoom);
 }
 
+private void storeAndApplyOperationForExistingHandle(Operation operation) {
```
amartya4256: I think we could call it ForCurrentHandle - it sounds more proper in this case, as compared to Resources.
By the way, what handle is it? Handle of image or something else? It seems to be different for different methods
Sometimes the operation applies to the gdipGraphics, lineStyle, foreground, brush, state, etc
Maybe we should just call it currentGCDAta or currentData?

HeikoKlare (commented):
I have taken a rough look at this and also tested the proposed change. The general concept is fine, it aligns with the other `Resource` implementations we have.

I did not have time yet to have an in-depth look into this and will unfortunately not be able to do so for the next week (before M1) due to being on vacation. Still, this should go into M1 to have as much testing of it as possible. So consider the concept approved from my side, but it would be nice if @fedejeanne can make an in-depth review of the proposal and then merge the final version for M1.

fedejeanne:
I started looking into it and made some minor nit-picky changes in 4f1ff0fca6bb4a83332e84d0a3e6d34031722513. I am moving to an in-depth review now and also testing.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -81,6 +82,9 @@ public final class GC extends Resource {
 
 	Drawable drawable;
 	GCData data;
+	private final GCData originalData = new GCData();
```
fedejeanne: Right now this field merely holds data but it is never read so it _could_ be removed. But I assume it is going to be used in a follow-up PR/commit, correct?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -81,6 +82,9 @@ public final class GC extends Resource {
...
+
+	private List<Operation> operations = new ArrayList<>();
```
fedejeanne: Same here: holds data, not being read. Necessary for a follow-up PR/commit?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -4259,19 +4668,49 @@ public void setBackground (Color color) {
...
-		Gdip.TextureBrush_delete(data.gdipBgPatternBrushAlpha);
-		data.gdipBgPatternBrushAlpha = 0;
+	storeAndApplyOperationForExistingHandle(new SetBackgroundPatternOperation(pattern));
+}
+
+private class SetBackgroundPatternOperation extends Operation {
+	private final Pattern pattern;
+
+	SetBackgroundPatternOperation(Pattern pattern) {
+		this.pattern = pattern; // TODO copy me
```
fedejeanne: Complete/remove this `TODO`

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -4733,25 +5290,38 @@ public void setLineCap(int cap) {
...
-			newDashes[i] = DPIUtil.scaleUp(drawable, (float) dashes[i], deviceZoom);
-			if (!changed && lineDashes[i] != newDashes[i]) changed = true;
+	storeAndApplyOperationForExistingHandle(new SetLineDashOperation(dashes));
+}
+
+private class SetLineDashOperation extends Operation {
+	private final int[] dashes;
+
+	SetLineDashOperation(int[] dashes) {
+		this.dashes = dashes;
```
fedejeanne: I would clone/copy the array passed as parameter (its ***content***) so that the local copy remains unchanged even if consumers change the original one.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -1698,16 +1819,29 @@ void drawBitmapColor(long imageHandle, int srcX, int srcY, int srcWidth, int src
...
 
-void drawLineInPixels (int x1, int y1, int x2, int y2) {
-	checkNonDisposed();
+private class DrawLineOperation extends Operation {
+	private final Point start;
+	private final Point end;
+
+	DrawLineOperation(Point start, Point end) {
+		this.start = start;
+		this.end = end;
```
fedejeanne: Copy/clone the original points since they are **not** immutable.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -1825,14 +1987,25 @@ public void drawPath (Path path) {
...
+	storeAndApplyOperationForExistingHandle(new DrawPointOperation(new Point(x, y)));
 }
 
-void drawPointInPixels (int x, int y) {
-	checkNonDisposed();
+private class DrawPointOperation extends Operation {
+	private final Point location;
+
+	DrawPointOperation(Point location) {
+		this.location = location;
```
fedejeanne: Same here: the parameter is not immutable which means you need a copy of it.

akoch-yatta: Would be fine, because the Points are created in GC as well, but adapted it, so the primitives are passed and the point is created in the operation

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
akoch-yatta: Would be fine, because the Points are created in GC as well, but adapted it, so the primitives are passed and the point is created in the operation

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
akoch-yatta: Good point

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
akoch-yatta: Yes, I will move it to the follow-up PR

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
akoch-yatta: Yes, I will move it to the follow-up PR

akoch-yatta:
@fedejeanne  Addressed all the comments in [a984ff1](https://github.com/eclipse-platform/eclipse.platform.swt/commit/a984ff1ad01f3fc4d01ca7f9e14393aa65895902)

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
akoch-yatta: Good point

fedejeanne (commented):
Thank you for addressing my comments. I squashed the commits (I had added one yesterday) and rebased on `master`. 

The code looks fine and I saw no regressions when running the _Runtime Workspace_ application, but I also tested using the `org.eclipse.swt.examples.graphics.GraphicsExample` and I noticed that clicking in some of the examples on the left panel yields this exception:

```java
Exception in thread "main" java.lang.NullPointerException: Cannot invoke "org.eclipse.swt.graphics.Transform.getElements(float[])" because "transform" is null
	at org.eclipse.swt.graphics.GC$SetTransformOperation.<init>(GC.java:5597)
	at org.eclipse.swt.graphics.GC.setTransform(GC.java:5589)
	at org.eclipse.swt.examples.graphics.PathTab.paint(PathTab.java:161)
	at org.eclipse.swt.examples.graphics.GraphicsExample.lambda$2(GraphicsExample.java:228)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:4360)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1214)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1238)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1223)
	at org.eclipse.swt.widgets.Composite.WM_PAINT(Composite.java:1636)
	at org.eclipse.swt.widgets.Control.windowProc(Control.java:4836)
	at org.eclipse.swt.widgets.Canvas.windowProc(Canvas.java:336)
	at org.eclipse.swt.widgets.Display.windowProc(Display.java:5128)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:3744)
	at org.eclipse.swt.examples.graphics.GraphicsExample.main(GraphicsExample.java:639)
```

These 2 trigger it:

image

Could you please look into that?

akoch-yatta:
> Thank you for addressing my comments. I squashed the commits (I had added one yesterday) and rebased on `master`.
> 
> The code looks fine and I saw no regressions when running the _Runtime Workspace_ application, but I also tested using the `org.eclipse.swt.examples.graphics.GraphicsExample` and I noticed that clicking in some of the examples on the left panel yields this exception:
> 
> ```java
> Exception in thread "main" java.lang.NullPointerException: Cannot invoke "org.eclipse.swt.graphics.Transform.getElements(float[])" because "transform" is null
> 	at org.eclipse.swt.graphics.GC$SetTransformOperation.<init>(GC.java:5597)
> 	at org.eclipse.swt.graphics.GC.setTransform(GC.java:5589)
> 	at org.eclipse.swt.examples.graphics.PathTab.paint(PathTab.java:161)
> 	at org.eclipse.swt.examples.graphics.GraphicsExample.lambda$2(GraphicsExample.java:228)
> 	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:91)
> 	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:4360)
> 	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1214)
> 	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1238)
> 	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1223)
> 	at org.eclipse.swt.widgets.Composite.WM_PAINT(Composite.java:1636)
> 	at org.eclipse.swt.widgets.Control.windowProc(Control.java:4836)
> 	at org.eclipse.swt.widgets.Canvas.windowProc(Canvas.java:336)
> 	at org.eclipse.swt.widgets.Display.windowProc(Display.java:5128)
> 	at org.eclipse.swt.internal.win32.OS.DispatchMessage(Native Method)
> 	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:3744)
> 	at org.eclipse.swt.examples.graphics.GraphicsExample.main(GraphicsExample.java:639)
> ```
> 
> Could you please look into that?

Thanks for the pointer. This was an issue yes. When looking through the full code again, I found two more places, that had a null pointer risk as well. I hope, I found all now.

PR #1803, opened 2025-02-04
[win32] strict check to identify long running GC

akoch-yatta:
This commit adds a check to identify long running or misconfigured GCs that render in a different zoom as its creating drawable intended to. Only executes if strict checking is enabled.

PR #2249, opened 2025-06-18
A strict check 200% image data is double the size of 100% image data

arunjose696:
A strict check to be used only for debugging means, this ensures the imageData is scaled linearly for 100 and 200 zooms.
 
**Steps to reproduce**

Out of the blow snippets. Snippet1  which uses an image data provider that always returns the same image, when executed with the VM argument `-Dorg.eclipse.swt.internal.enableStrictChecks=True`, logs the error` "ImageData should be linearly scaled across zooms." to stdout. Whereas snippet 2 which scales ImageData linearly runs without this error log.
<details><summary>snippet 1</summary>

```java
public class TestImageProvider {
        ImageDataProvider provider = (zoom) -> new ImageData(
            16, 16, 32,
            new PaletteData((new RGB(255, 0, 0)))
        );
        Image image = new Image(display, provider);
        System.out.println("Done execution");
    }
}
```
</details>
<details><summary>snippet 2</summary>

```java
public class TestImageProvider {
        ImageDataProvider provider = (zoom) -> new ImageData(
            16*zoom/100, 16*zoom/100, 32,
            new PaletteData((new RGB(255, 0, 0)))
        );
        Image image = new Image(display, provider);  
        System.out.println("Done execution");
    }
}
```
</details>

The below examples demonstrate testing  this again with PNG images. In the first snippet, the ImageDataProvider returns the image data from the same file for both 100% and 200% zoom levels without scaling, which causes the error message "ImageData should be linearly scaled across zooms." to be printed to stdout. In the second snippet it provides properly scaled image data at 200% zoom (using a scaled image), so it executes without logging this error.
To run these snippets, download the following PNG files and place them in your working directory:
[ pause.png](https://github.com/user-attachments/assets/eee3aeb9-1f33-4362-9559-a1ebcfacbf89)

[ pause@2x.png](https://github.com/user-attachments/assets/88380f50-7dbd-4163-aed4-4c9d678117a2)

<details><summary>snippet 1</summary>

```java
public class TestImageProvider {
    	System.out.println("Working directory: " + System.getProperty("user.dir"));

        ImageDataProvider provider = zoom -> {
            if (zoom == 100) {
                return new ImageData("pause.png");
            }
            else if (zoom == 200) {
                return new ImageData("pause.png");
            }
            else {
                return null;
            }
        };

        Image image = new Image(display, provider);
        System.out.println("Done execution");
    }
}

```
</details>

<details><summary>snippet 2</summary>

```java
public class TestImageProvider {
    	System.out.println("Working directory: " + System.getProperty("user.dir"));

        ImageDataProvider provider = zoom -> {
            if (zoom == 100) {
                return new ImageData("pause.png");
            }
            else if (zoom == 200) {
                return new ImageData("pause@2x.png");
            }
            else {
                return null;
            }
        };

        Image image = new Image(display, provider);
        System.out.println("Done execution");
    }
}
```

</details>

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
@@ -150,9 +150,10 @@ public static ImageData autoScaleImageData (Device device, final ImageData image
 	int height = imageData.height;
 	int scaledWidth = Math.round (width * scaleFactor);
 	int scaledHeight = Math.round (height * scaleFactor);
+	int defaultZoomLevel = 100;
 	boolean useSmoothScaling = isSmoothScalingEnabled() && imageData.getTransparencyType() != SWT.TRANSPARENCY_MASK;
 	if (useSmoothScaling) {
-		Image original = new Image (device, (ImageDataProvider) zoom -> imageData);
+		Image original = new Image(device, (ImageDataProvider) zoom -> (zoom == defaultZoomLevel) ? imageData : null);
 		ImageGcDrawer drawer =  new ImageGcDrawer() {
 			@Override
 			public void drawOn(GC gc, int imageWidth, int imageHeight) {
@@ -166,7 +167,7 @@ public static ImageData autoScaleImageData (Device device, final ImageData image
 			}
 		};
 		Image resultImage = new Image (device, drawer, scaledWidth, scaledHeight);
-		ImageData result = resultImage.getImageData (100);
+		ImageData result = resultImage.getImageData (defaultZoomLevel);
 		original.dispose ();
 		resultImage.dispose ();
 		return result;
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
@@ -612,8 +612,31 @@ public Image(Device device, ImageDataProvider imageDataProvider) {
 		SWT.error(SWT.ERROR_INVALID_ARGUMENT, null,
 				": ImageDataProvider [" + imageDataProvider + "] returns null ImageData at 100% zoom.");
 	}
+	if (Device.strictChecks) {
+		validateLinearScaling(imageDataProvider);
+	}
 	init();
 	this.device.registerResourceWithZoomSupport(this);
+}
+
+private void validateLinearScaling(ImageDataProvider provider) {
+	final int baseZoom = 100;
+	final int scaledZoom = 200;
+	final int scaleFactor = scaledZoom / baseZoom;
+	ImageData baseImageData = provider.getImageData(baseZoom);
+	ImageData scaledImageData = provider.getImageData(scaledZoom);
+
+	if (scaledImageData == null) {
+		return;
+	}
+
+	if (scaledImageData.width != scaleFactor * baseImageData.width
+			|| scaledImageData.height != scaleFactor * baseImageData.height) {
+		System.err.println(String.format(
+				"***WARNING: ImageData should be linearly scaled across zooms but size is (%d, %d) at 100%% and (%d, %d) at 200%%.",
+				baseImageData.width, baseImageData.height, scaledImageData.width, scaledImageData.height));
+		new Error().printStackTrace(System.err);
+	}
 }
 
 /**
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_Image.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_Image.java
@@ -1043,13 +1043,15 @@ public void test_imageDataIsCached() {
 public void test_imageDataSameViaDifferentProviders() {
 	assumeFalse("Cocoa generates inconsistent image data", SwtTestUtil.isCocoa);
 	String imagePath = getPath("collapseall.png");
-	ImageFileNameProvider imageFileNameProvider = __ -> {
-		return imagePath;
+	ImageFileNameProvider imageFileNameProvider = zoom -> {
+		return (zoom == 100) ? imagePath : null;
 	};
-	ImageDataProvider dataProvider = __ -> {
-		try (InputStream imageStream = Files.newInputStream(Path.of(imagePath))) {
-			return new ImageData(imageStream);
-		} catch (IOException e) {
+	ImageDataProvider dataProvider = zoom -> {
+		if (zoom == 100) {
+			try (InputStream imageStream = Files.newInputStream(Path.of(imagePath))) {
+				return new ImageData(imageStream);
+			} catch (IOException e) {
+			}
 		}
 		return null;
 	};
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_internal_SVGRasterizer.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_internal_SVGRasterizer.java
@@ -54,17 +54,11 @@ public class Test_org_eclipse_swt_internal_SVGRasterizer {
 
 	@Test
 	public void test_ConstructorLorg_eclipse_swt_graphics_Device_ImageDataProvider() {
-		ImageDataProvider validImageDataProvider = zoom -> {
-			String fileName = "collapseall.svg";
-			return new ImageData(getPath(fileName));
-		};
+		ImageDataProvider validImageDataProvider = zoom -> (zoom == 100) ? new ImageData(getPath("collapseall.svg")) : null;
 		Image image = new Image(Display.getDefault(), validImageDataProvider);
 		image.dispose();
 
-		ImageDataProvider corruptImageDataProvider = zoom -> {
-			String fileName = "corrupt.svg";
-			return new ImageData(getPath(fileName));
-		};
+		ImageDataProvider corruptImageDataProvider = zoom -> (zoom == 100) ? new ImageData(getPath("corrupt.svg")) : null;
 		SWTException e = assertThrows(SWTException.class,
 				() -> new Image(Display.getDefault(), corruptImageDataProvider));
 		assertSWTProblem("Incorrect exception thrown for provider with corrupt images", SWT.ERROR_INVALID_IMAGE, e);


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -610,10 +610,26 @@ public Image(Device device, ImageDataProvider imageDataProvider) {
 		SWT.error(SWT.ERROR_INVALID_ARGUMENT, null,
 				": ImageDataProvider [" + imageDataProvider + "] returns null ImageData at 100% zoom.");
 	}
+	if(Device.strictChecks) {
```
akoch-yatta: if( -> if (

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
```diff
@@ -301,7 +301,12 @@ private static ImageData autoScaleImageData (Device device, final ImageData imag
 	int scaledHeight = Math.round (height * scaleFactor);
 	boolean useSmoothScaling = isSmoothScalingEnabled() && imageData.getTransparencyType() != SWT.TRANSPARENCY_MASK;
 	if (useSmoothScaling) {
-		Image original = new Image (device, (ImageDataProvider) zoom -> imageData);
+		Image original = new Image(device, (ImageDataProvider) zoom -> {
+			if (zoom == 100) {
+				return imageData;
+			}
+			return null;
+		});
```
akoch-yatta: is zoom == 100 always correct, is it? Will it always be fetched for 100 and then scaled?

arunjose696: ImageData is only requested for 100 zoom, so only 100 is required https://github.com/eclipse-platform/eclipse.platform.swt/blob/085c1c645b23f28b696bbbabb9c91689eda5b277/bundles/org.eclipse.swt/Eclipse%20SWT/common/org/eclipse/swt/internal/DPIUtil.java#L323

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -610,10 +610,26 @@ public Image(Device device, ImageDataProvider imageDataProvider) {
...
 				": ImageDataProvider [" + imageDataProvider + "] returns null ImageData at 100% zoom.");
 	}
+	if (Device.strictChecks) {
+		validateLinearScaling(imageDataProvider);
+	}
 	init();
 	this.device.registerResourceWithZoomSupport(this);
 }
 
+private void validateLinearScaling(ImageDataProvider provider) {
```
amartya4256, line 622: This method seems tightly coupled using a lot of magic numbers. Maybe adapt it a little?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -610,10 +610,26 @@ public Image(Device device, ImageDataProvider imageDataProvider) {
...
+	ImageData data100 = provider.getImageData(100);
+	ImageData data200 = provider.getImageData(200);
+
+	if (data200 == null) {
+		return;
+	}
+
+	if (data200.width != 2 * data100.width || data200.height != 2 * data100.height) {
+		SWT.error(SWT.ERROR_INVALID_ARGUMENT, null, "ImageData should be linearly scaled across zooms.");
+	}
```
amartya4256:
```suggestion
	final int scaledZoom = 200;
	ImageData baseImageData = provider.getImageData(100);
	ImageData scaledImageData = provider.getImageData(scaledZoom);

	if (baseImageData == null || scaledImageData  == null) {
		return;
	}
	
	int expectedWidth = DPIUtil.scaleUp(baseImage.width, scaledZoom);
	int expectedHeight = DPIUtil.scaleUp(baseImage.height, scaledZoom);

	if (scaledImage.width != expectedWidth || scaledImage.height != expectedHeight) {
		SWT.error(SWT.ERROR_INVALID_ARGUMENT, null, "ImageData should be linearly scaled across zooms.");
	}
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -610,10 +610,26 @@ public Image(Device device, ImageDataProvider imageDataProvider) {
...
 	init();
 	this.device.registerResourceWithZoomSupport(this);
 }
 
+private void validateLinearScaling(ImageDataProvider provider) {
+	ImageData data100 = provider.getImageData(100);
+	ImageData data200 = provider.getImageData(200);
+
+	if (data200 == null) {
```
amartya4256: why not check for data100 as well?

arunjose696: data200 may be null in  a valid ImageDataProvider implementation which should pass the strict check(in this check we just make sure we dont throw exceptions if data200 is null). But data100 is guaranteed to exist and will have already been [validated ](https://github.com/eclipse-platform/eclipse.platform.swt/pull/2249/files#diff-bdf38bc65dcd13a55f97900d358d6d64cceb4da8b147f692b5620ee4c1e94807L609) before this line is reached, so no additional checks are necessary here.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
arunjose696: I agree about avoiding magic numbers. However, I felt that using DPIUtil.scaleUp here would slightly overcomplicate things for readability. Instead, I introduced a scaleFactor variable to clarify the scaling logic.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
arunjose696: Using magic numbers for 100 and 200 seems unavoidable, have converted them to final variables, if they are any better

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
akoch-yatta: Ah okay, then please put the 100 in a common local variable to use in both places, so it is not changed independently

arunjose696:
In CI failing checks I have the below error, Dont think this has anything to do with my changes.

```
[2025-07-01T10:06:14.756Z] [ERROR] Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:5.0.0-SNAPSHOT:compare-version-with-baselines (compare-attached-artifacts-with-release) on project org.eclipse.swt.tools: Baseline and reactor have the same fully qualified version, but different content
[2025-07-01T10:06:14.756Z] [ERROR] different
[2025-07-01T10:06:14.756Z] [ERROR]    META-INF/ECLIPSE_.RSA: present in baseline only
[2025-07-01T10:06:14.756Z] [ERROR]    META-INF/ECLIPSE_.SF: present in baseline only
[2025-07-01T10:06:14.756Z] [ERROR] -> [Help 1]
[2025-07-01T10:06:14.756Z] org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:5.0.0-SNAPSHOT:compare-version-with-baselines (compare-attached-artifacts-with-release) on project org.eclipse.swt.tools: Baseline and reactor have the same fully qualified version, but different content
[2025-07-01T10:06:14.756Z] different
[2025-07-01T10:06:14.756Z]    META-INF/ECLIPSE_.RSA: present in baseline only
[2025-07-01T10:06:14.756Z]    META-INF/ECLIPSE_.SF: present in baseline only
```

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_Image.java
```diff
@@ -1031,13 +1031,18 @@ public void test_imageDataIsCached() {
 public void test_imageDataSameViaDifferentProviders() {
 	assumeFalse("Cocoa generates inconsistent image data", SwtTestUtil.isCocoa);
 	String imagePath = getPath("collapseall.png");
-	ImageFileNameProvider imageFileNameProvider = __ -> {
-		return imagePath;
+	ImageFileNameProvider imageFileNameProvider = zoom -> {
+		if (zoom == 100)
+			return imagePath;
+		else
+			return null;
```
fedejeanne:
```suggestion
		return (zoom == 100) ? imagePath : null;
```

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_ImageData.java
```diff
@@ -799,6 +802,25 @@ public void test_setAlphaIII() {
 	assertSWTProblem("Incorrect exception thrown for putWidth < 0", SWT.ERROR_INVALID_ARGUMENT, ex);
 }
 
+@Test
+public void test_drawImageFailsWithNonScalingImageDataProviderWhenstrictCheckEnabled() {
```
fedejeanne:
```suggestion
public void test_drawImageFailsWithNonScalingImageDataProviderWhenStrictCheckIsEnabled() {
```

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_internal_SVGRasterizer.java
```diff
@@ -55,15 +55,23 @@ public void test_ConstructorLorg_eclipse_swt_graphics_Device_ImageFileNameProvid
...
 	public void test_ConstructorLorg_eclipse_swt_graphics_Device_ImageDataProvider() {
 		ImageDataProvider validImageDataProvider = zoom -> {
-			String fileName = "collapseall.svg";
-			return new ImageData(getPath(fileName));
+			if (zoom == 100) {
+				String fileName = "collapseall.svg";
+				return new ImageData(getPath(fileName));
+			} else {
+				return null;
+			}
```
fedejeanne: Either move the `return null` outside of the `else` or, even better, transform the whole thing to: `return (zoom == 100) ? return new ImageData(getPath("collapseall.svg")) : null`. Your choice.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_internal_SVGRasterizer.java
```diff
@@ -55,15 +55,23 @@ public void test_ConstructorLorg_eclipse_swt_graphics_Device_ImageFileNameProvid
...
 
 		ImageDataProvider corruptImageDataProvider = zoom -> {
-			String fileName = "corrupt.svg";
-			return new ImageData(getPath(fileName));
+			if (zoom == 100) {
+				String fileName = "corrupt.svg";
+				return new ImageData(getPath(fileName));
+			} else {
+				return null;
+			}
```
fedejeanne: Same as above

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_ImageData.java
```diff
@@ -799,6 +802,25 @@ public void test_setAlphaIII() {
...
+	Display display = Display.getDefault();
+	GC gc = new GC(display);
+	try {
+		ImageDataProvider wrongDataProvider = (zoom) -> new ImageData(16, 16, 32, new PaletteData());
+		Image image = new Image(display, wrongDataProvider);
+		gc.drawImage(image, 0, 0, 16, 16, 0, 0, 16, 16);
+		image.dispose();
+		if (System.getProperty("org.eclipse.swt.internal.enableStrictChecks") != null) {
+			fail("Expected an exception due to non-linearly scaled image data provider");
+		}
```
fedejeanne: You need to make sure that this property is set at the beginning of the test and then reset it at the end (back to its **original value**), otherwise this test might not test anything.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -610,10 +610,29 @@ public Image(Device device, ImageDataProvider imageDataProvider) {
...
 	init();
 	this.device.registerResourceWithZoomSupport(this);
 }
 
+private void validateLinearScaling(ImageDataProvider provider) {
+	final int zoom100 = 100;
+	final int zoom200 = 200;
+	final int scaleFactor =  zoom200 / zoom100;
+	ImageData data100 = provider.getImageData(zoom100);
+	ImageData data200 = provider.getImageData(zoom200);
```
fedejeanne: I assume you went for the "constants" in order to make it clear that you are comparing the images in its original size (100%) with the ones at `@2x` size (200%), right? If that's the case then please consider using other names for the variables, like:

```java
final int baseZoom = 100;
final int largeZoom = 200;
```

Fields and variables that merely mirror their content are not really expressing what they _are_ or _represent_ but simply what they _contain_ (which I can see in the assignments).

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
```diff
@@ -299,9 +299,15 @@ private static ImageData autoScaleImageData (Device device, final ImageData imag
...
 	int scaledHeight = Math.round (height * scaleFactor);
+	int defaultZoomLevel = 100;
 	boolean useSmoothScaling = isSmoothScalingEnabled() && imageData.getTransparencyType() != SWT.TRANSPARENCY_MASK;
 	if (useSmoothScaling) {
-		Image original = new Image (device, (ImageDataProvider) zoom -> imageData);
+		Image original = new Image(device, (ImageDataProvider) zoom -> {
+			if (zoom == defaultZoomLevel) {
+				return imageData;
+			}
+			return null;
```
fedejeanne:
```suggestion
			return (zoom == defaultZoomLevel) imageData : null;
```

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_ImageData.java
arunjose696: The org.eclipse.swt.internal.enableStrictChecks property is set as a static variable when the Display class loads. That means changing the system property during the test won’t enable or disable strict checks.

Right now, our CI runs with strict checks enabled, so the test will fail if the image data provider doesn’t scale correctly, as expected.

My idea was to keep this test so it shows the failure when strict checks are enabled, but also to demonstrate that if strict checks were disabled (like in some other environments), it wouldn’t fail.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
arunjose696: renamed to baseZoom and scaledZoom. I think scaledZoom is a bit better than largeZoom.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_Image.java
arunjose696: Done

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_internal_SVGRasterizer.java
arunjose696: Done

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_internal_SVGRasterizer.java
arunjose696: Done

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_ImageData.java
fedejeanne: The way you describe it means that you need to test for 2 things:
1. If strict checks are enabled, an exception should be thrown
2. If strict checks are disabled, no exception should be thrown

Currently you are testing either 1. or 2. depending on what the current system property says. This is confusing and should be improved. I also suggest that you use **assumptions** to better express what the test should actually test _e.g._ when testing 1. (checks enabled) use `assumeNotNull(System.getProperty("org.eclipse.swt.internal.enableStrictChecks"))`

fedejeanne: Recap: 
- Create 2 methods to test for checks enabled and disabled
- Use assumptions in them

arunjose696: Done.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -610,10 +610,29 @@ public Image(Device device, ImageDataProvider imageDataProvider) {
...
+	final int scaleFactor =  scaledZoom / baseZoom;
+	ImageData baseImageData = provider.getImageData(baseZoom);
+	ImageData scaledImageData = provider.getImageData(scaledZoom);
+
+	if (scaledImageData == null) {
+		return;
+	}
+
+	if (scaledImageData.width != scaleFactor * baseImageData.width || scaledImageData.height != scaleFactor * baseImageData.height) {
+		SWT.error(SWT.ERROR_INVALID_ARGUMENT, null, "ImageData should be linearly scaled across zooms.");
```
fedejeanne: Why throw an exception here? I thought this check was only for debugging purposes like other checks we added before (https://github.com/eclipse-platform/eclipse.platform.swt/pull/1979/, https://github.com/eclipse-platform/eclipse.platform.swt/pull/1803/, https://github.com/eclipse-platform/eclipse.platform.swt/pull/1753/)

@akoch-yatta ?

arunjose696: Replaced exception with a `System.err.println()`

fedejeanne: In that case you'll have to remove the 2 new tests since there is no way to test that something was written to the console

fedejeanne: And you should adapt the text of this PR since nothing **fails**, one simply sees some output in the console. Please also add a snippet to test by providing 2 image files, one of them (the `_@2x`) having the wrong size.

arunjose696: I’ve updated the PR text . I also added the two snippets one with an ImageDataProvider that scales the image data correctly and one that doesn’t.

To keep the setup simple, I didn’t use actual image files (e.g. SVGs), since that would require uploading multiple files and slightly complicate the example. Let me know if including real images is still necessary

fedejeanne: It is necessary, that is a valid use case and it should be tested too.

arunjose696: In PR text I’ve now added tests using actual .png image files in addition to the earlier code snippets.

I’m still a bit unclear on why this was necessary, as I thought the original snippets already demonstrated the intended behavior with the case where ImageData is unscaled. That said, I’ve still added a snippet with real images as requested. Please let me know if this is what you had in mind, or if there’s another use case I should be covering.

fedejeanne: You're right, I had the wrong expectation about this PR. 

FTR what I wanted to test is in JFace and you could see it for example by adding a modified version of `validateLinearScaling` that takes `ImageDescriptor` as a parameter and calling it at the end of `org.eclipse.jface.tests.images.FileImageDescriptorTest.testAdaptToURL()`. If you do that and provide a second file that adds `@2x` to the end of the file but it does **not** have double the size then the check would fail.

Since I'm not on Windows at the moment, I can't test this PR- I'll get back to it later this week.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -610,10 +610,29 @@ public Image(Device device, ImageDataProvider imageDataProvider) {
...
+	final int scaleFactor =  scaledZoom / baseZoom;
+	ImageData baseImageData = provider.getImageData(baseZoom);
+	ImageData scaledImageData = provider.getImageData(scaledZoom);
+
+	if (scaledImageData == null) {
+		return;
+	}
+
+	if (scaledImageData.width != scaleFactor * baseImageData.width || scaledImageData.height != scaleFactor * baseImageData.height) {
+		System.err.println("ImageData should be linearly scaled across zooms.");
```
HeikoKlare: I think we should print the stack trace here, so that someone can more easily find the problematic ImageDataProvider. In addition, we prefix all other strict check prints with `***WARNING: `, so we should probably do the same here.

HeikoKlare (commented):
In general, the PR looks good to me. I have only one comment regarding the printed message.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java

ptziegler:
This caused a regression when working with `swt.autoScale`. See https://github.com/eclipse-platform/eclipse.platform.swt/issues/2353

PR #2250, opened 2025-06-18
Ensure DrawImage deals with scenario ImageData not linearly scaled

arunjose696:
This change makes sure DrawImage handles the scenario where ImageData may not be linearly scaled across zooms.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
@@ -990,7 +990,7 @@ public void drawImage (Image image, int srcX, int srcY, int srcWidth, int srcHei
 	if (image.isDisposed()) SWT.error(SWT.ERROR_INVALID_ARGUMENT);
 
 	int gcZoom = getZoom();
-	int srcImageZoom = calculateZoomForImage(gcZoom, srcWidth, srcHeight, destWidth, destHeight);
+	int srcImageZoom = calculateZoomForImage(image, gcZoom, srcWidth, srcHeight, destWidth, destHeight);
 	drawImage(image, srcX, srcY, srcWidth, srcHeight, destX, destY, destWidth, destHeight, gcZoom, srcImageZoom);
 }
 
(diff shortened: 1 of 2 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -1026,20 +1026,22 @@ private int calculateZoomForImage(int gcZoom, int srcWidth, int srcHeight, int d
 
 private void drawImage(Image image, int srcX, int srcY, int srcWidth, int srcHeight, int destX, int destY,
 		int destWidth, int destHeight, int imageZoom, int scaledImageZoom) {
-	Rectangle src = DPIUtil.scaleUp(drawable, new Rectangle(srcX, srcY, srcWidth, srcHeight), scaledImageZoom);
+	Rectangle scaledBounds = image.getBounds(scaledImageZoom);
+	Rectangle unScaledBound = image.getBounds();
+	float scalingFactor = (float) scaledBounds.height / unScaledBound.height;
+	Rectangle src = DPIUtil.scaleUp(drawable, new Rectangle(srcX, srcY, srcWidth, srcHeight),(int) (scalingFactor * 100));
```
akoch-yatta: I thought about this a bit. Isn't it possible to do the check regarding getBounds in the calling place and then pass the "correct" scaledImageZoom, so the implementation in here can stay as is and the change is in the calling place in GC?

arunjose696: The earlier approach was to use scaledImageZoom as before to obtain Image.win32_getHandle(), using the "correct" zoom level only for calculating the bounds. , but taking another look it would not make much difference if we get handles at either of these. I have moved this logic to the calling site when the scaledImageZoom is calculated

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -1003,25 +1003,29 @@ private Collection<Integer> getAllCurrentMonitorZooms() {
...
+		scaledImageZoom = 200;
+	} else {
+		scaledImageZoom = 100;
+	}
+	Rectangle scaledBounds = image.getBounds(scaledImageZoom);
+	Rectangle unScaledBound = image.getBounds();
+	// validates if the image bounds are scaled up correctly as per required zoom, if it is not zoom is returned as per actual scaling factor.
+	float scalingFactor = (float) scaledBounds.height / unScaledBound.height;
+	scaledImageZoom = (int) (scalingFactor * 100);
+	return scaledImageZoom;
```
HeikoKlare, line 1028: I like the overall approach and I think we could achieve to properly deal with the situation of non-linearly scaled image data. But I think the implementation would need to be a bit more sophisticated. With this approach, since the corrected `scaledImageZoom` may be an arbitrary value, we would potentially start loading images at arbitrary zooms and thus create more and more unwanted handles. In addition, there is no guarantee that the image data that the image yields for such a zoom have fitting width/height as it depends on which original image data (such as the 100%, 150% or 200% data for PNGs) are used and scaled.
I think what we might actually need to adapt is the `srcWidth` and `srcHeight` as they must not be treated as points with an ordinary point-to-pixel conversion based on the zoom if the zoom is not appropriate. However, if you have an image with non-linear image data, you do not even know to which the original point-based `srcWidth` and `srcHeight` should fit. We currently assume that it's the 100% version of the image, but it may also be the case that someone creats an `ImageDataProvider` that always yields image data at the same zoom, which  could, for example, also be the current device zoom. If that is not 100%, there might be unexpected results.

I have just tested again the snippet in https://github.com/vi-eclipse/Eclipse-Platform/issues/313 that led to an exception before and found that it now (silently) works. So I wonder whether we should focus on improving the API (adding a note to `ImageDataProvider` that implementations must return linearly scaled data) and rely on the just added strict check and adaptations of consumers (if necessary) instead of making the logic here more complex.

What do you think, @akoch-yatta @arunjose696?

arunjose696: >  I wonder whether we should focus on improving the API (adding a note to `ImageDataProvider` that implementations must return linearly scaled data) and rely on the just added strict check and adaptations of consumers (if necessary) instead of making the logic here more complex.
> 
I think the better approach would be to document clearly in ImageDataProvider that implementations must return linearly scaled data.  Adding complexity to drawImage to "fix" bad input risks hiding these issues rather than solving them at the actual ( a incorrect ImageDataProvider implementation).

HeikoKlare: Fully agree. Just for the sake of completeness: the risk is of course that implementations of `ImageDataProvider` usually have been there for years already, so the contract improvement may stay unnoticed and the implementations will stay somehow incompatible. Still, I think that's the best we can do.
So, concrete proposal: We drop this PR and instead improve the contract of `ImageDataProvider`. We write a short news as a means to make aware of the potential issue and the contract adaptatino (as actually the assumption that image data is linearly scaled was always there but violations were silently accepted in a more or less resilient way).
@akoch-yatta would then also be good to have your final opinion when you are back before proceeding with that.

akoch-yatta: I agree. Everything else seem like a workaround, that would only cover up the overall issue, that this contract was (at least implicitly) there from the introduction of HiDPI support, but not enforced.

arunjose696:
Closing this PR as discussed

PR #2251, opened 2025-06-18
[Build] Remove check for SWT native binary rebuild commit

HannesWell:
Currently the build is aborted if the branch tip is at a commit, which was committed with the email of the 'Eclipse Platform Bot'. This is done to prevent rebuilds after a commit for a SWT native binary rebuild was pushed by the bot. But because that bot is also used for other kind of commits, this test leads to false positives. This is a consequence of
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2207

This reworks the pipeline to check if the head commit has a SWT-native-binary build tag attached, which is a more precise indicator.

This also fixes the check if a native rebuild is enforced, when testing if the build can be skipped.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -113,28 +113,18 @@ pipeline {
 			steps {
 				dir('eclipse.platform.swt') {
 					checkout scm
-					script {
-						def authorMail = sh(script: 'git log -1 --pretty=format:"%ce" HEAD', returnStdout: true)
-						echo 'HEAD commit author: ' + authorMail
-						def buildBotMail = 'platform-bot@eclipse.org'
-						if (buildBotMail.equals(authorMail) && !params.forceNativeBuilds) {
-							// Prevent endless build-loops due to self triggering because of a previous automated build of natives and the associated updates.
-							currentBuild.result = 'ABORTED'
-							error('Abort build only triggered by automated SWT-natives update.')
-						}
-						sh """
+					sh '''
 						java -version
 						git version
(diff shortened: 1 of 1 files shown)


HannesWell:
Similar to https://github.com/eclipse-equinox/equinox/pull/1038, an in-depth analysis showed that the pipeline does in fact not get into an endless native rebuild loop in case the check for a SWT-native-binary build tag is removed. Only one new build is triggered, when a commit for a launcher binary rebuild is pushed (as usual for a push), which does not rebuild the native binaries again (because the latest SWT-native-binary tag points to the then HEAD commit).
I therefore think it's better to continue with that solution as it also allows to simplify the overall pipeline and an extra build after a native rebuild seems acceptable to me. Does anybody disagree?

HannesWell:
> Similar to [eclipse-equinox/equinox#1038](https://github.com/eclipse-equinox/equinox/pull/1038), an in-depth analysis showed that the pipeline does in fact not get into an endless native rebuild loop in case the check for a SWT-native-binary build tag is removed. Only one new build is triggered, when a commit for a launcher binary rebuild is pushed (as usual for a push), which does not rebuild the native binaries again (because the latest SWT-native-binary tag points to the then HEAD commit).

Finally I got a full run that confirms this, so I consider this ready for submission (as there were no objections).
This approach also has the advantage that after a native binaries rebuild the master build (should) complete successfully and is not aborted, which the latter is interpreted and displayed as build failure by the GH UI.

PR #2253, opened 2025-06-19
Stop generating feature source

akurtakov:
With https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3140 they will no longer be published thus it makes no sense to generate them anymore.


akurtakov:
Failing test is unrelated. Merging.

PR #2257, opened 2025-06-20
Convert FileDialog#setFilterExtensions to varargs

vogella:
Nicer to use for end user and API compliant. If this change is integrated we must update the usage of this method in platform to avoid compiler warnings.


akurtakov:
As it's semantically coupled with setFilterNames it would be good if they use the same style.

vogella:
> As it's semantically coupled with setFilterNames it would be good if they use the same style.

Will do, I just wanted to see what the general feedback was before doing to much changes.

akurtakov:
It's a fine change IMO.

akurtakov:
Would you please fix the 2 new compiler warnings?

akurtakov:
To be finished with github.com/eclipse-platform/eclipse.platform.swt/pull/2700

Issue #1758, opened 2025-01-24
Radio and check buttons text truncated under LUbuntu 24.04LTS (LXQt)


maccasoft:
**Describe the bug**
Radio and check buttons text is truncated when running under LUbuntu 24.04LTS (LXQt) or generally KDE (users reports).

**To Reproduce**
An example with Eclipse SDK itself, open the Find / Replace dialog, all radio / check button labels are truncated.

**Expected behavior**
Text should not be truncated.

**Screenshots**

![Image](https://github.com/user-attachments/assets/32360885-88ba-4aa6-a4c3-373fe88d9a8f)

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [ ] Windows
- - [x] Linux
- - [ ] macOS 

2. Additional OS info (e.g. OS version, Linux Desktop, etc)
 LUbuntu 24.04.1LTS
 LXQt 1.4.0
 Qt 5.15.13

3. JRE/JDK version
 Temurin 21.0.6+7-LTS

**Version since**
 Eclipse 4.34.0
 SWT 3.128.0

**Workaround (or) Additional context**
No workaround, as far as I know.

deepika-u:
Tried on the below version of ubuntu and i am seeing the buttons text properly.

For reference -> 
![Image](https://github.com/user-attachments/assets/90c9615e-b905-4a8d-81fb-d678814d112f)

Environment 
Description:	Ubuntu 24.04 LTS
XDG_SESSION_TYPE=wayland

Eclipse SDK
Version: 2025-03 (4.35)
Build id: I20250108-1800
OS: Linux, v.6.8.0-51-generic, x86_64 / gtk 3.24.41
Java vendor: Oracle Corporation
Java runtime version: 23+36-2368
Java version: 23

maccasoft:
Yes, the button text display properly on Ubuntu 24.04 (Gnome), it doesn't display properly on **LUbuntu**, that's a variant of Ubuntu running **LXQt** as desktop environment.
I guess is an issue with the DE or the theme or both that doesn't report the widget size correctly.

Reference https://lubuntu.me/

deepika-u:
Thanks for the details.  👍

ptziegler:
The same behavior can also be seen on KDE (Kubuntu 24.04)
![Image](https://github.com/user-attachments/assets/4e7cf892-37f9-4b13-acfe-4a3095183091)

[...]

PR #2258, opened 2025-06-20
Increase right margin for check- and combo-box

ptziegler:
Both the text and the arrow are truncated when using the Breeze theme in KDE/Fedora 42.

Closes https://github.com/eclipse-platform/eclipse.platform.swt/issues/1758


ptziegler:
Without:

Without

With (Breeze):
Breeze

With (Default):
Default

With (Emacs):
Emacs

ptziegler:
I don't really have a way to test through all GTK versions, so I'm not sure which css file is the best one to adapt.

PR #2260, opened 2025-06-24
Deprecating Display#getDPI

ShahzaibIbrahim:

When calling getDPI method as **Device** even with Printer instance, it will show you deprecation warning but should be fine when using the **Printer** object while calling getDPI. Here's the example:
image

### Deprecate `Display#getDPI` and Refactor DPI Handling Across Platforms

This PR deprecates the `Display#getDPI` method and replaces all its usages with platform-appropriate `dots_per_inch` values. The motivation behind this change is to make DPI handling more explicit and robust by directly referencing the correct DPI value for each supported OS.

#### Platform DPI Values

Through testing and verification across all supported operating systems, the following standard base DPI values have been applied:

- **Windows:** 96 DPI  
- **Linux:** 96 DPI  
- **Cocoa (macOS):** 72 DPI  

These values are now used consistently wherever DPI calculations are required, ensuring accuracy and consistency across the codebase.

#### Printing and Font Scaling

Special care has been taken to ensure that the change does not affect how fonts are rendered when printing. The DPI factor used for printers remains consistent, so that printed fonts do not appear differently when the base DPI changes between platforms.

#### Summary of Changes

- Deprecated the `Display#getDPI` method.
- Replaced all direct calls to `getDPI` with platform-specific `dots_per_inch` values.
- Established and documented standard DPI values for Windows (96), Linux (96), and macOS (72).
- Ensured that font rendering and DPI scaling for printing remain consistent and unaffected by these changes.

---

This refactor simplifies DPI management, improves cross-platform predictability, and reduces future maintenance by making DPI values explicit and centralized.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Device.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Device.java
@@ -519,7 +519,18 @@ public int getDepth () {
  * @exception SWTException <ul>
  *    <li>ERROR_DEVICE_DISPOSED - if the receiver has been disposed</li>
  * </ul>
+ *
+ * @deprecated <p>This method returns a single global DPI value
+ * that does not reflect per-monitor DPI settings on modern operating systems.
+ * In environments with different scaling factors across monitors, it may provide
+ * a misleading or meaningless result, as it does not correspond to the actual DPI
+ * of any specific monitor.</p>
+ *
+ * <p>Note: While deprecated for general {@code Device} instances like {@code Display},
+ * this method may still be validly used when called on a {@code Printer} instance,
+ * where a single global DPI value is meaningful and expected.</p>
  */
+@Deprecated
 public Point getDPI () {
 	checkDevice ();
 	long hDC = internal_new_GC (null);
--- a/examples/org.eclipse.swt.examples/src/org/eclipse/swt/examples/imageanalyzer/ImageAnalyzer.java
+++ b/examples/org.eclipse.swt.examples/src/org/eclipse/swt/examples/imageanalyzer/ImageAnalyzer.java
@@ -1166,6 +1166,7 @@ public class ImageAnalyzer {
 		if (image == null) return;
 
 		try {
+			final int DOTS_PER_INCH = 96;
 			// Ask the user to specify the printer.
 			PrintDialog dialog = new PrintDialog(shell, SWT.NONE);
 			if (printerData != null) dialog.setPrinterData(printerData);
@@ -1173,9 +1174,8 @@ public class ImageAnalyzer {
 			if (printerData == null) return;
 
 			Printer printer = new Printer(printerData);
-			Point screenDPI = display.getDPI();
 			Point printerDPI = printer.getDPI();
-			int scaleFactor = printerDPI.x / screenDPI.x;
+			int scaleFactor = printerDPI.x / DOTS_PER_INCH;
 			Rectangle trim = printer.computeTrim(0, 0, 0, 0);
 			if (printer.startJob(currentName)) {
 				if (printer.startPage()) {
--- a/bundles/org.eclipse.swt/Eclipse SWT Printing/cocoa/org/eclipse/swt/printing/Printer.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Printing/cocoa/org/eclipse/swt/printing/Printer.java
@@ -624,7 +624,7 @@ public Point getDPI() {
 }
 
 Point getIndependentDPI() {
-	return super.getDPI();
+	return new Point(72, 72);
 }
 
 /**
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Device.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Device.java
@@ -391,7 +391,18 @@ public int getDepth () {
  * @exception SWTException <ul>
  *    <li>ERROR_DEVICE_DISPOSED - if the receiver has been disposed</li>
  * </ul>
+ *
+ * @deprecated <p>This method returns a single global DPI value
+ * that does not reflect per-monitor DPI settings on modern operating systems.
+ * In environments with different scaling factors across monitors, it may provide
+ * a misleading or meaningless result, as it does not correspond to the actual DPI
+ * of any specific monitor.</p>
+ *
+ * <p>Note: While deprecated for general {@code Device} instances like {@code Display},
+ * this method may still be validly used when called on a {@code Printer} instance,
+ * where a single global DPI value is meaningful and expected.</p>
  */
+@Deprecated
 public Point getDPI () {
 	checkDevice ();
 	return getScreenDPI();
@@ -611,8 +622,9 @@ protected void init () {
 	/* Initialize the system font slot */
 	boolean smallFonts = System.getProperty("org.eclipse.swt.internal.carbon.smallFonts") != null;
 	double systemFontSize = smallFonts ? NSFont.smallSystemFontSize() : NSFont.systemFontSize();
-	Point dpi = this.dpi = getDPI(), screenDPI = getScreenDPI();
-	NSFont font = NSFont.systemFontOfSize(systemFontSize * dpi.y / screenDPI.y);
+	final int DOTS_PER_INCH = 72;
+	Point dpi = this.dpi = getDPI();
+	NSFont font = NSFont.systemFontOfSize(systemFontSize * dpi.y / DOTS_PER_INCH);
 	font.retain();
 	systemFont = Font.cocoa_new(this, font);
 }
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Font.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Font.java
@@ -244,6 +244,7 @@ public boolean equals(Object object) {
  */
 public FontData[] getFontData() {
 	if (isDisposed()) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
+	final int DOTS_PER_INCH = 72;
 	NSAutoreleasePool pool = null;
 	if (!NSThread.isMainThread()) pool = (NSAutoreleasePool) new NSAutoreleasePool().alloc().init();
 	try {
@@ -258,8 +259,8 @@ public FontData[] getFontData() {
 		if ((traits & OS.NSBoldFontMask) != 0) style |= SWT.BOLD;
 		if ((extraTraits & OS.NSItalicFontMask) != 0) style |= SWT.ITALIC;
 		if ((extraTraits & OS.NSBoldFontMask) != 0) style |= SWT.BOLD;
-		Point dpi = device.dpi, screenDPI = device.getScreenDPI();
-		FontData data = new FontData(name, (float)handle.pointSize() * screenDPI.y / dpi.y, style);
+		Point dpi = device.dpi;
+		FontData data = new FontData(name, (float)handle.pointSize() * DOTS_PER_INCH / dpi.y, style);
 		data.nsName = nsName;
 		return new FontData[]{data};
 	} finally {
@@ -312,8 +313,9 @@ public int hashCode() {
 void init(String name, float height, int style, String nsName) {
 	if (name == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
 	if (height < 0) SWT.error(SWT.ERROR_INVALID_ARGUMENT);
-	Point dpi = device.dpi, screenDPI = device.getScreenDPI();
-	float size = height * dpi.y / screenDPI.y;
+	final int DOTS_PER_INCH = 72;
+	Point dpi = device.dpi;
+	float size = height * dpi.y / DOTS_PER_INCH;
 	NSFont systemFont = NSFont.systemFontOfSize(size);
 	NSFont boldSystemFont = NSFont.boldSystemFontOfSize(size);
 	String systemFontName = systemFont.familyName().getString();
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Device.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Device.java
@@ -455,7 +455,18 @@ public int getDepth () {
  * @exception SWTException <ul>
  *    <li>ERROR_DEVICE_DISPOSED - if the receiver has been disposed</li>
  * </ul>
+ *
+ * @deprecated <p>This method returns a single global DPI value
+ * that does not reflect per-monitor DPI settings on modern operating systems.
+ * In environments with different scaling factors across monitors, it may provide
+ * a misleading or meaningless result, as it does not correspond to the actual DPI
+ * of any specific monitor.</p>
+ *
+ * <p>Note: While deprecated for general {@code Device} instances like {@code Display},
+ * this method may still be validly used when called on a {@code Printer} instance,
+ * where a single global DPI value is meaningful and expected.</p>
  */
+@Deprecated
 public Point getDPI () {
 	checkDevice ();
 	return getScreenDPI();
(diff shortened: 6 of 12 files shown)


examples/org.eclipse.swt.examples/src/org/eclipse/swt/examples/imageanalyzer/ImageAnalyzer.java
```diff
@@ -1172,9 +1172,8 @@ void menuPrint() {
 			if (printerData == null) return;
 
 			Printer printer = new Printer(printerData);
-			Point screenDPI = display.getDPI();
 			Point printerDPI = printer.getDPI();
-			int scaleFactor = printerDPI.x / screenDPI.x;
+			int scaleFactor = printerDPI.x / 100;
```
akoch-yatta: Wasn't Display:getDPI always returning 96?

ShahzaibIbrahim: No, it depends on the display. Sometimes 144 or 96 or depending on the monitor. We then decided to use 100 as the base consistent value to calculate the scaleFactor like we did here: https://github.com/eclipse-platform/eclipse.platform.swt/pull/1802/files

fedejeanne (commented):
In general the change looks harmless, but I'm missing some context here. Could you please add some `@deprecated` comment to the JavaDoc of `Display::getDPI` and explain **why** it is wrong to use it? Replacing the value with `100` seems a bit esoteric.

HeikoKlare (commented):
The current change does actually not fit to what it is supposed to be (as mentioned in the PR title): it should be about deprecating `Device.getDPI()`, not `Display.getDPI()`, and about overwriting that deprecation for `Printer.getDPI()` as for the `Printer` class the method is still useful while for `Display` it is not. It would also need to be applied to the implementations on every OS.

ShahzaibIbrahim:
@HeikoKlare PR title is updated now. It indeed deprecate the Display#getDPI and we keep the Printer#getDPI as is. 

@fedejeanne I have added javadoc stating why is it being deprecated. also I adapted the value from 100 to 96 as Default Dots Per Inch (DPI) similar to here: https://github.com/eclipse-gef/gef-classic/pull/750/files

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
```diff
@@ -2037,6 +2037,31 @@ public int getIconDepth () {
...
+ * that does not reflect per-monitor DPI settings on modern operating systems.
+ * In environments with different scaling factors across monitors, this method
+ * provides a misleading or meaningless result, as it does not correspond to
+ * the actual DPI of any specific monitor.
+ * </p>
+ * <p>
+ * In most cases, callers should not rely on this method. If a fallback is
+ * needed, a value of 96 DPI (the default baseline for 100% scaling) may be
+ * used.
+ * </p>
```
fedejeanne: All this should go into the text of `@deprecated`

HeikoKlare:
> PR title is updated now. It indeed deprecate the Display#getDPI and we keep the Printer#getDPI as is.

But that's not what we want to do: it does not really make sense to deprecate an overwritten method, as consumers may just operate on the abstraction (i.e., `Device#getDPI()`), which is _not_ deprecated. So we need to deprecate the most abstract method (`Device#getDPI()`) und "un-deprecate" the specialization that we want to keep (`Printer#getDPI()`).

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
```diff
@@ -2037,6 +2037,31 @@ public int getIconDepth () {
...
+ *
+ * @return the DPI of the display device
+ *
+ * @deprecated Use per-monitor DPI or scaling APIs instead. This method is unreliable in multi-monitor setups with varying DPI.
```
fedejeanne: Which APIs/Methods for example?

ShahzaibIbrahim: Updated javadoc

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Device.java
```diff
@@ -521,7 +521,18 @@ public int getDepth () {
...
+ * that does not reflect per-monitor DPI settings on modern operating systems.
+ * In environments with different scaling factors across monitors, it may provide
+ * a misleading or meaningless result, as it does not correspond to the actual DPI
+ * of any specific monitor.</p>
+ *
+ * <p>Note: While deprecated for general {@code Device} instances like {@code Display},
+ * this method may still be validly used when called on a {@code Printer} instance,
+ * where a single global DPI value is meaningful and expected.</p>
  */
+@Deprecated
```
HeikoKlare, line 533: You need to do the deprecation for every OS. Then the check complaining about the unnecessary deprecation will not occur anymore (as the checks are execute on Linux on which this method is currently not deprecated).

ShahzaibIbrahim: @HeikoKlare I did it, but still getting same error?

HeikoKlare: No, it now complains about the usage of deprecated `getDPI()` method at some places:

> NORMAL:
> The method getDPI() from the type Device is deprecated

fipro78:
@HeikoKlare @fedejeanne @ShahzaibIbrahim 

Hi all,

since you are now planning to deprecate `Display#getDPI()` I wonder how I should deal with this in a custom SWT widget implementation.

In NatTable I added code to render the table according to the DPI. I used `Display#getDPI()` to calculate the factor to use for painting on the GC accordingly.

As I didn't find information about how to migrate anywhere, I digged into the code and searched for information in tickets. I changed the usage of `Display#getDPI()` to the following:

```
(Display.getDefault().getPrimaryMonitor().getZoom() * 96) / 100)
```

If I start the SWT application with `-Dswt.autoScale.updateOnRuntime=false` the rendering of NatTable looks correct. Checking the values I see:

```
Display#getDPI().x = 144
(Display.getDefault().getPrimaryMonitor().getZoom() * 96) / 100) = 144
```

If I set `-Dswt.autoScale.updateOnRuntime=true` the rendering of NatTable is not correct anymore, because it automatically upscales. Actually it looks like it upscales twice, once by SWT itself and once by NatTable. Checking the values I see:

```
Display#getDPI().x = 96
(Display.getDefault().getPrimaryMonitor().getZoom() * 96) / 100) = 144
```

In `Display#getDPI()` I can see that `DPIUtil.scaleDown()` is called. So I assume with `-Dswt.autoScale.updateOnRuntime=true` the value is downscaled, with `-Dswt.autoScale.updateOnRuntime=false` it is not. At least that is the case with 4.36. 

Is there an additional change in the upcoming SWT implementation that takes care of this?
Are there public methods that would help to deal with that fact? `DPIUtil` is internal, therefore also all the system property keys that would need to be checked.

Of course I can manually check the value of the system property `swt.autoScale.updateOnRuntime` (hardcoded as the key is not available via constant) and if it is set to true I take a constant value of 96 and if false calculate the DPI based on the primary monitor as shown above.

Is this the intention when deprecating `Display#getDPI()`? Then it should be documented somehow. And probably there should be some public API that helps in dealing with that situation. If I missed that information how to deal with the new DPI situation in SWT, please point me to that information, so I am able to make the necessary adjustments, or at least recommendations for NatTable adopters, to handle it correctly. And probably also in some Nebula widgets where `Display#getDPI()` is used at several places as already mentioned in https://github.com/vi-eclipse/Eclipse-Platform/issues/205

BTW, I checked the __Plug-in Migration Guide__ in the [Eclipse Help](https://help.eclipse.org/latest/index.jsp) and do not find any information about this topic. And actually I think the help generation is broken somehow, because the __Incompatibility__  section of almost all entries is the same, which doesn't seem correct to me.

akoch-yatta:
Hi @fipro78,

thanks for reaching out. 

Generally, it should not be necessary to do any scaling at all on client side. This is of course only the case if some assumptions that SWT has are the same as yours. I'll try to explain, what I understand from your comment

> 
> since you are now planning to deprecate `Display#getDPI()` I wonder how I should deal with this in a custom SWT widget implementation.
> 
> In NatTable I added code to render the table according to the DPI. I used `Display#getDPI()` to calculate the factor to use for painting on the GC accordingly.
> 
> As I didn't find information about how to migrate anywhere, I digged into the code and searched for information in tickets. I changed the usage of `Display#getDPI()` to the following:
> 
> ```
> (Display.getDefault().getPrimaryMonitor().getZoom() * 96) / 100)
> ```
> 
> If I start the SWT application with `-Dswt.autoScale.updateOnRuntime=false` the rendering of NatTable looks correct. Checking the values I see:
> 
> ```
> Display#getDPI().x = 144
> (Display.getDefault().getPrimaryMonitor().getZoom() * 96) / 100) = 144
> ```
> 
> If I set `-Dswt.autoScale.updateOnRuntime=true` the rendering of NatTable is not correct anymore, because it automatically upscales. Actually it looks like it upscales twice, once by SWT itself and once by NatTable. Checking the values I see:
> 
> ```
> Display#getDPI().x = 96
> (Display.getDefault().getPrimaryMonitor().getZoom() * 96) / 100) = 144
> ```
> 
> In `Display#getDPI()` I can see that `DPIUtil.scaleDown()` is called. So I assume with `-Dswt.autoScale.updateOnRuntime=true` the value is downscaled, with `-Dswt.autoScale.updateOnRuntime=false` it is not. At least that is the case with 4.36.

Yes and no. `swt.autoScale.updateOnRuntime` causes this different values, but if I understand your setup correctly, the root cause for the effect you see comes from `swt.autoScale`, which has different default values for `-Dswt.autoScale.updateOnRuntime=true` (`-Dswt.autoScale=quarter`) and `-Dswt.autoScale.updateOnRuntime=false`  (`-Dswt.autoScale=integer200`). If you e.g. start your application with `-Dswt.autoScale.updateOnRuntime=false -Dswt.autoScale=quarter` you should see the same issues.

Before talking about a solution I would like to understand what you would like to achieve. Let's say, i am on a 150% monitor - do you always want to draw on a 1,5 scale independent of the value of swt.autoScale? So do you always want to link your scaling to the monitor zoom?

fipro78:
> If you e.g. start your application with -Dswt.autoScale.updateOnRuntime=false -Dswt.autoScale=quarter you should see the same issues.

Correct.

> Let's say, i am on a 150% monitor - do you always want to draw on a 1,5 scale independent of the value of swt.autoScale? So do you always want to link your scaling to the monitor zoom?

I want to scale the NatTable according to the rest of the SWT application. Fonts, grid lines, images etc. In the past I had to inspect the DPI setting and do the calculation manually, because on the `GC` you are drawing pixel based, and 10 pixels on a 150% monitor should be 15 pixels. That was not done automatically in the past, so I had to do this manually. Without this modification, the line heights for example are to small.

As NatTable also supports zooming in and out of the table, I need to set the initial value according to the setting of the application, so it looks like the rest of the application on startup.

akoch-yatta:
> > Let's say, i am on a 150% monitor - do you always want to draw on a 1,5 scale independent of the value of swt.autoScale? So do you always want to link your scaling to the monitor zoom?
> 
> I want to scale the NatTable according to the rest of the SWT application. Fonts, grid lines, images etc. In the past I had to inspect the DPI setting and do the calculation manually, because on the `GC` you are drawing pixel based, and 10 pixels on a 150% monitor should be 15 pixels. That was not done automatically in the past, so I had to do this manually. Without this modification, the line heights for example are to small.

Sounds like a yes to me. In SWT the semantics of zoom got a bit messy over the years, but its hard to clean up or streamline it fully now, because a lot of workarounds (like yours) were build because of this. For example swt.autoScale=200 means images will be scaled by 200 (independent of monitor zoom), but fonts are still relying on the monitor zoom and the value of Display#getDPI is hardly unusable in this scenario.

> 
> As NatTable also supports zooming in and out of the table, I need to set the initial value according to the setting of the application, so it looks like the rest of the application on startup.

I assume you are fully drawing on a GC, am I correct? This sounds like a similar scenario GEF is facing and I assume your current solution is mostly similar to the solutions in GEF. I think the solution for you is to incorporate the monitor zoom into your custom table zoom. Two questions:

- Do you render something, that is currently affected by your DPI calculations, but **not** by the custom table zoom? 
- How do you do the font scaling in your custom zoom? Do you scaling the font height with your scale factor and create a font out of that?

fipro78:
> I assume you are fully drawing on a GC, am I correct?

Yes

> Do you render something, that is currently affected by your DPI calculations, but not by the custom table zoom?

I hope not. Even the embedded controls like dropdowns are affected by the custom table zoom.

> How do you do the font scaling in your custom zoom? Do you scaling the font height with your scale factor and create a font out of that?

Yes, see 

https://github.com/eclipse-nattable/nattable/blob/2942c0f7db81e6588ff1d1d984302046d117618d/org.eclipse.nebula.widgets.nattable.core/src/org/eclipse/nebula/widgets/nattable/NatTable.java#L438-L444 

and 

https://github.com/eclipse-nattable/nattable/blob/2942c0f7db81e6588ff1d1d984302046d117618d/org.eclipse.nebula.widgets.nattable.core/src/org/eclipse/nebula/widgets/nattable/util/GUIHelper.java#L131-L141

akoch-yatta:
Follow-up question: How/where is the custom zoom applied to the GC, e.g. respective to coordinates/sizes etc. Are your using GC#setTransform similar to [SWTGrapics](https://github.com/eclipse-gef/gef-classic/blob/efb1ee1408b2998d3bf7bc2ee37cb0f2eb65a675/org.eclipse.draw2d/src/org/eclipse/draw2d/SWTGraphics.java#L1025) or scaling it up yourself similar to [ScaledGraphics](https://github.com/eclipse-gef/gef-classic/blob/efb1ee1408b2998d3bf7bc2ee37cb0f2eb65a675/org.eclipse.draw2d/src/org/eclipse/draw2d/ScaledGraphics.java) before calling the GC API?

fipro78:
I am scaling myself before calling the GC API.

akoch-yatta:
Okay, so, as I assumed, the situation looks quite similar to GEF. We are currently spending quite some time to evaluate what is the best solution for this scenario and started to implement some hidden features for that in SWT. 

**Short disclaimer first:** what I will describe is under development, not API, will probably change until it is done and is currently windows only and could break GTK and MacOS. Compatibility with GTK and MacOS are probably quite easy to achieve, but we are focusing on having a proper solution for windows first.

One big problem with drawing (or partially drawing complex elements) on a GC is when you use an autoScale mode like quarter and draw on uneven monitor zooms  like 125% or 175% -> This will most certainly result in rendering artifacts because the conversions from points to pixels and vice versa leads to rounding issues. 
In our opinion the only solution is to always draw on an unscaled GC and do the scaling yourself. Mid/long-term plan is to (hopefully) add this as a feature in SWT itself, but for now it is on your side. As you already have a custom zoom implemented, that should be quite easy:

1. First step is to disable the autoscaling for the Canvas you draw on with your GC. You can do that with `canvas.setData("AUTOSCALE_DISABLED", true)`.
2. Second step is to incorporate the native zoom of the canvas (== monitor zoom) into your custom zoom. You can have a look into our WIP state for GEF: [Zoom adaptions](https://github.com/vi-eclipse/gef-classic/blob/3_24_static_zoom/org.eclipse.gef/src/org/eclipse/gef/internal/MonitorAwareZoomManager.java). Important is the `zoomChangedListener` to react on zoom changes on the Canvas and `canvas.getData("NATIVE_ZOOM")` to get the current zoom of it. If you multiply this zoom with your own custom zoom the result show achieve, what you want.

If you have follow-up questions, I would propose to either open a discussion here in the repo and continue there or we switch over to a discussion/issue in the NatTable repo.

fipro78:
@akoch-yatta 
Thanks for the updates. I have played around with this a bit, and it seems to work quite well. There are some flaws, but these are mainly related to the usage of `Display#getDPI()` in static helper classes. I will need to change those also to better align with the changes in SWT. IIRC these places are for `Image`s which I need to work on also. At least if I got the information from the N&N right.

I opened a discussion in the NatTable forum: https://github.com/eclipse-nattable/nattable/discussions/164

Would be great if you could share any updates there or point to other discussions with informations, so I can update NatTable and probably also some Nebula widgets accordingly (e.g. the RichTextPainter is basically the same).

PR #2263, opened 2025-06-25
[win32] Refresh GC handle when related image 

akoch-yatta:
This PR  adapts the behavior of the commonly used pattern of a GC that draw onto a plain image. In the windows implementation the monitor specific scaling could lead to issues with this pattern, because during creation time of an GC/Image combo it was not always clear with which zoom the final result will be drawn upon. This commit will redraw the GC on an image handle if a new zoom is requested. The handle for the old zoom of the image will be kept, but if another operation of the GC is executed it will only be applied on the new handle.

### How to test
You can use this small windows specific snippet to force an image redraw with a different zoom
```java

import org.eclipse.swt.*;

public class SnippetGcRedraw {

		shell.setText("GC Redraw");
		int width = 150, height = 200;
		Image image = new Image (display, width, height);

		GCData data = new GCData();
		GC gc = GC.win32_new(image, data);
		gc.setBackground(display.getSystemColor(SWT.COLOR_RED));
		gc.fillRectangle (0, 0, width, height);
		gc.drawLine (0, 0, width, height);
		gc.drawLine (0, height, width, 0);
		gc.drawText ("Default Image", 10, 10);
		final Point origin = new Point (0, 0);
		final Canvas canvas = new Canvas (shell, SWT.NONE);
		canvas.addListener (SWT.Paint, e -> {
			GC gc2 = e.gc;
			gc2.getGCData().nativeZoom = 150;
			gc2.drawImage (image, origin.x, origin.y);
			gc2.getGCData().nativeZoom = 200;
			gc2.drawImage (image, origin.x + 300, origin.y);
		});
		shell.setSize (1000, 1000);
		shell.open ();
		}
		gc.dispose ();
		image.dispose();
		display.dispose ();
	}
}
```
#### Before
If the image was used with a GC the same ImageData were used for all zooms
Screenshot 2025-06-25 080153

#### After
Left image is drawn for 150%, right is redrawn for 200%
Screenshot 2025-06-25 080338

**This PR relies upon #2243 and must be merged after it. Only the latest commit is part of this PR**

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
@@ -1707,7 +1707,7 @@ private long configureGC(GCData data, int zoom) {
 		SWT.error(SWT.ERROR_INVALID_ARGUMENT);
 	}
 
-	if(Device.strictChecks) {
+	if (Device.strictChecks) {
 		checkImageTypeForValidCustomDrawing(zoom);
 	}
 	/* Create a compatible HDC for the device */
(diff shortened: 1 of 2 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2554,7 +2564,6 @@ protected ImageHandle newImageHandle(int zoom) {
 		}
 		GC gc = new GC(new DrawableWrapper(image, zoom), gcStyle);
 		try {
-			gc.data.nativeZoom = zoom;
```
fedejeanne, line 2557: Why is this change needed?

fedejeanne (commented):
I tested it and it looks good. 

I only have one question (https://github.com/eclipse-platform/eclipse.platform.swt/pull/2263#discussion_r2171347280) but even with that change, the code looks good.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: It's probably not needed, just unecessary anymore. But probably completly independent from this PR

Issue #2267, opened 2025-06-26
[WebKit] Allow disabling WebKit Browser search per Browser widget


trancexpress:
In https://github.com/eclipse-platform/eclipse.platform.swt/issues/2222 we added search capabilities to `Browser` widgets based on GTK+ and WebKit.

Not every `Browser` needs those though and so far we only have the option to disable the search globally, not per instance. We would like to be able to enable/disable it per instance. We didn't do this with the initial PR since its not trivial to do API-wise.

laeubi:
> We didn't do this with the initial PR since its not trivial to do API-wise.

Why not? Could it be a `SWT.SEARCH` style flag? If not it would be good to explain the difficulties here in the issue in case someone later comes here to pick it up and do not need to dig in history.

PR #2268, opened 2025-06-26
[WebKit] Allow disabling WebKit Browser search per Browser widget

trancexpress:
Fixes: #2267


trancexpress:
Usage:

```
Browser browser = new Browser(shell, SWT.NONE);
...
DisableWebKitBrowserSearch.disableSearch(browser);
```

In a snippet:

```
package test.browser;

import org.eclipse.swt.SWT;
import org.eclipse.swt.browser.Browser;
import org.eclipse.swt.browser.DisableWebKitBrowserSearch;

public class TestBrowser {

	public static void main(String[] args) throws InterruptedException {
		shell.setSize(600, 800);
		shell.setText("Test Browser");

		Browser b = new Browser(shell, SWT.NONE);
		b.setText(
				"""
				<html>
				 <head>
				 </head>
				 <body>
				   Hello World
				   <h1>Hello World</h1>
				   <p>Hello World</p>
				 </body>
				</html>
				""");

		DisableWebKitBrowserSearch.disableSearch(b);

		shell.open();

		while (!shell.isDisposed()) {
				display.sleep();
		}
		display.dispose();
	}

}
```

trancexpress:
@iloveeclipse how about disabling the search per default and then enabling it per instance?

That way we can selectively enable the search for e.g. the Eclipse Help cases where we need it.

Technically we can add callbacks too, now that we have an idea how to target only WebKit `Browser` widgets. So that the search can be re-implemented from outside (e.g. nicer than what we have).

laeubi:
> Usage:
> 
> ```
> Browser browser = new Browser(shell, SWT.NONE);
> ...
> DisableWebKitBrowserSearch.disableSearch(browser);
> ```

This is a bit odd style I think, better would be

> ```
> // enable search in the browser
> Browser browser = new Browser(shell, SWT.SEARCH);
> 
> ```

this would also be backward compatible and OPT-IN behavior.

trancexpress:
> This is a bit odd style I think, better would be
> 
> > ```
> > // enable search in the browser
> > Browser browser = new Browser(shell, SWT.SEARCH);
> > ```
> 
> this would also be backward compatible and OPT-IN behavior.

Unfortunately we are implementing the search **only** for WebKit browsers. Any API we add at `Browser` (including a style constant) would need to do _something_ for every type of browser?

Or adding a constant that does something only for WebKit and GTK+ is fine? From my POV I don't think it is fine, but maybe there are other such constants? I'll check that when I can.

laeubi:
> Unfortunately we are implementing the search only for WebKit browsers. Any API we add at Browser (including a style constant) would need to do something for every type of browser?

I don't think so we have many style constants that are only a _hint_, e.g

https://github.com/eclipse-platform/eclipse.platform.swt/blob/e2cbda9823b020a4e230fe0e5aca3216023cb7fc/bundles/org.eclipse.swt/Eclipse%20SWT/common/org/eclipse/swt/SWT.java#L1567-L1579

and there is even already a SEARCH constant (currently only used by text fields)

https://github.com/eclipse-platform/eclipse.platform.swt/blob/e2cbda9823b020a4e230fe0e5aca3216023cb7fc/bundles/org.eclipse.swt/Eclipse%20SWT/common/org/eclipse/swt/SWT.java#L1556-L1565

trancexpress:
> I don't think so we have many style constants that are only a _hint_, e.g

Alright, let me check what I can do with the `SEARCH` constant.

bundles/org.eclipse.swt/Eclipse SWT Browser/gtk/org/eclipse/swt/browser/BrowserFactory.java
```diff
@@ -13,13 +13,17 @@
...
+
 class BrowserFactory {
 
 WebBrowser createWebBrowser (int style) {
 	boolean webkitInstalled = WebKit.IsInstalled ();
 	if (!webkitInstalled) return null;
 
-	return new WebKit ();
+	WebKit webKit = new WebKit ();
+	webKit.enableSearch = (style & SWT.SEARCH) == SWT.SEARCH;
```
laeubi, line 25:
```suggestion
	webKit.enableSearch = (style & SWT.SEARCH) != 0;
```

bundles/org.eclipse.swt/Eclipse SWT WebKit/gtk/org/eclipse/swt/browser/WebKit.java
```diff
@@ -795,7 +797,7 @@ public void create (Composite parent, int style) {
 				break;
 			}
 			case SWT.KeyDown: {
-				if (!disableBrowserSearch && event.keyCode == 'f' && (event.stateMask & SWT.CTRL) == SWT.CTRL) {
+				if (!disableBrowserSearch && enableSearch && event.keyCode == 'f' && (event.stateMask & SWT.CTRL) == SWT.CTRL) {
```
laeubi: disable Browser search should not be required here if clients want to opt in they should provide some kind of own way to pass `SWT.SEARCH` or not to the browser instance created.

trancexpress: Makes me wonder if we should use the VM option to just enable search in every `Browser` (aside from hover pop-ups, i.e. in a `SWT.TOOL` `Shell`, where we do breaking things with the search `Shell` handling)... WDYT @iloveeclipse , @laeubi ?

laeubi: I think having the JVM option can be removed. The search will be disabled by default and if one wants it it can be enabled for that particular widget. This would also be backward compatible.

iloveeclipse: > if we should use the VM option to just enable search in every Browser (aside from hover pop-ups, i.e. in a SWT.TOOL Shell, where we do breaking things with the search Shell handling)...

I would not do that. Assuming someone has product based on browser views/editors and has there either shortcut support of just don't want search to be possible, we would make them unhappy. Current idea with the optional flag seems to be fine and would cover most of the use cases we know.

> I think having the JVM option can be removed. The search will be disabled by default and if one wants it it can be enabled for that particular widget. This would also be backward compatible.

This is not correct, because our goal is to enable the search flag in the places where it makes sense in SDK (embedded browser, help view etc). So the flag will be used by default and if webkit crashes (which happens time to time due version changes, missing some rpms etc), there should be a way to "fix" it.

@trancexpress : I assume this would be the platform patch we would need next to enable search in both help view and external help:
```
diff --git a/ua/org.eclipse.help.ui/src/org/eclipse/help/ui/internal/browser/embedded/EmbeddedBrowser.java b/ua/org.eclipse.help.ui/src/org/eclipse/help/ui/internal/browser/embedded/EmbeddedBrowser.java
index 0cac583..538d6ac 100644
--- a/ua/org.eclipse.help.ui/src/org/eclipse/help/ui/internal/browser/embedded/EmbeddedBrowser.java
+++ b/ua/org.eclipse.help.ui/src/org/eclipse/help/ui/internal/browser/embedded/EmbeddedBrowser.java
@@ -123,3 +123,3 @@
 
-		browser = new Browser(shell, SWT.NONE);
+		browser = new Browser(shell, SWT.SEARCH);
 		browser.setLayoutData(new GridData(SWT.FILL, SWT.FILL, true, true));
diff --git a/ua/org.eclipse.help.ui/src/org/eclipse/help/ui/internal/views/BrowserPart.java b/ua/org.eclipse.help.ui/src/org/eclipse/help/ui/internal/views/BrowserPart.java
index f8da110..47a108d 100644
--- a/ua/org.eclipse.help.ui/src/org/eclipse/help/ui/internal/views/BrowserPart.java
+++ b/ua/org.eclipse.help.ui/src/org/eclipse/help/ui/internal/views/BrowserPart.java
@@ -99,3 +99,3 @@
 			final IToolBarManager tbm, IMenuManager menuManager) {
-		browser = new Browser(parent, SWT.NULL);
+		browser = new Browser(parent, SWT.SEARCH);
 		browser.addLocationListener(new LocationListener() {
```
I'm correct or how do you plan to use this patch?

bundles/org.eclipse.swt/Eclipse SWT Browser/gtk/org/eclipse/swt/browser/BrowserFactory.java
iloveeclipse: I found the explicit check in the patch better to read.

bundles/org.eclipse.swt/Eclipse SWT WebKit/gtk/org/eclipse/swt/browser/WebKit.java
trancexpress: That is how we would use the patch in its current form, yes.

bundles/org.eclipse.swt/Eclipse SWT Browser/gtk/org/eclipse/swt/browser/BrowserFactory.java
laeubi: I have never seen this and it feels like `if (b == true)` usage here ...

bundles/org.eclipse.swt/Eclipse SWT WebKit/gtk/org/eclipse/swt/browser/WebKit.java
laeubi: > This is not correct, because our goal is to enable the search flag in the places where it makes sense in SDK (embedded browser, help view etc). So the flag will be used by default and if webkit crashes (which happens time to time due version changes, missing some rpms etc), there should be a way to "fix" it.

Then there should be some kind of flag in the code using SWT.SEARCH and not in SWT itself if we feel uncomfortable with it.

iloveeclipse: > Then there should be some kind of flag in the code using SWT.SEARCH and not in SWT itself if we feel uncomfortable with it.

Makes little sense for me. Assuming we have N places where client code would use the SWT.SEARCH flag for browser (and some of them are outside of SDK), do you really expect that one has to add extra checks there and not in the single one? Then it is irrelevant for the webkit who calls it it it is going to crash. As a product owner one would need to have access to all the code that is integrated in Eclipse based application to make sure webkit does not crash my application, which is simply impossible.

iloveeclipse: `disableBrowserSearch && enableSearch` is funny.
I would propose to rename `disableBrowserSearch` to `disableBrowserSearchGlobally` or something like that.

iloveeclipse (approved):
Looks OK for me, just minor issue with naming.

iloveeclipse:
@trancexpress : please provide PR for platform, so we can have it in the next SDK build tomorrow.

PR #2270, opened 2025-06-27
Add float precision to MonitorAwarePoint and MonitorAwareRectangle

fedejeanne:


fedejeanne:
Not necessary anymore

PR #2271, opened 2025-06-27
Revise the Cursor class to support scaling

arunjose696:
Previously, cursors were initialized with a single ImageData, which caused issues on systems with varying zoom levels, for example, cursors were not scaled at all on Linux, or were blurrily auto-scaled on Windows.

This commit introduces a new constructor for the Cursor class that accepts an Image object instead of ImageData. This allows the code instantiating the cursor to pass an image capable of providing appropriately scaled image data based on the current zoom level.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Cursor.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Cursor.java
@@ -382,6 +382,10 @@ void createNSCursor(int hotspotX, int hotspotY, byte[] buffer, int width, int he
  */
 public Cursor(Device device, ImageData source, int hotspotX, int hotspotY) {
 	super(device);
+	setupCursorFromImageData(source, hotspotX, hotspotY);
+}
+
+private void setupCursorFromImageData(ImageData source, int hotspotX, int hotspotY) {
 	if (source == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
 	if (hotspotX >= source.width || hotspotX < 0 ||
 		hotspotY >= source.height || hotspotY < 0) {
@@ -446,6 +450,41 @@ public Cursor(Device device, ImageData source, int hotspotX, int hotspotY) {
 	} finally {
 		if (pool != null) pool.release();
 	}
+}
+
+/**
+ * Constructs a new cursor given a device, image describing
+ * the desired cursor appearance, and the x and y coordinates of
+ * the <em>hotspot</em> (that is, the point within the area
+ * covered by the cursor which is considered to be where the
+ * on-screen pointer is "pointing").
+ * <p>
+ * You must dispose the cursor when it is no longer required.
+ * </p>
+ *
+ * @param device the device on which to allocate the cursor
+ * @param imageDataProvider the ImageDataProvider for the cursor
+ * @param hotspotX the x coordinate of the cursor's hotspot
+ * @param hotspotY the y coordinate of the cursor's hotspot
+ *
+ * @exception IllegalArgumentException <ul>
+ *    <li>ERROR_NULL_ARGUMENT - if device is null and there is no current device</li>
+ *    <li>ERROR_NULL_ARGUMENT - if the image is null</li>
+ *    <li>ERROR_INVALID_ARGUMENT - if the hotspot is outside the bounds of the
+ * 		 image</li>
+ * </ul>
+ * @exception SWTError <ul>
+ *    <li>ERROR_NO_HANDLES - if a handle could not be obtained for cursor creation</li>
+ * </ul>
+ *
+ * @see #dispose()
+ *
+ * @since 3.131
+ */
+public Cursor(Device device, ImageDataProvider imageDataProvider, int hotspotX, int hotspotY) {
+	super(device);
+	if (imageDataProvider == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
+	setupCursorFromImageData(imageDataProvider.getImageData(100), hotspotX, hotspotY);
 }
 
 @Override
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Cursor.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Cursor.java
@@ -272,6 +272,11 @@ public Cursor(Device device, ImageData source, ImageData mask, int hotspotX, int
  */
 public Cursor(Device device, ImageData source, int hotspotX, int hotspotY) {
 	super(device);
+	setupCursorFromImageData(source, hotspotX, hotspotY);
+
+}
+
+private void setupCursorFromImageData(ImageData source, int hotspotX, int hotspotY) {
 	if (source == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
 	if (hotspotX >= source.width || hotspotX < 0 ||
 		hotspotY >= source.height || hotspotY < 0) {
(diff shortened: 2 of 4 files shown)


ptziegler:
Out of curiosity: Is there a benefit to using the `Image`, rather than e.g. the `ImageDataProvider`?
This now creates an additional handle that needs to be disposed together with the Cursor. Which is fine, of course, just not very convenient.
Latter might also play a little nicer with the `ImageDescriptor`, because you could simply create the cursor via `new Cursor(device, imageDescriptor::getImageData, hotspotX, hotspotY)`.

arunjose696:
> Out of curiosity: Is there a benefit to using the `Image`, rather than e.g. the `ImageDataProvider`? This now creates an additional handle that needs to be disposed together with the Cursor. Which is fine, of course, just not very convenient. Latter might also play a little nicer with the `ImageDescriptor`, because you could simply create the cursor via `new Cursor(device, imageDescriptor::getImageData, hotspotX, hotspotY)`.

If cursors are initialized with a single ImageData and later requested at a different zoom level, that image data is auto-scaled using DPUtil. This just scales up the original ImageData, which makes the cursor appear blurry at different zoom levels (e.g., when moving the cursor from one monitor to another). If Image is used this scaling can be non blurry.

laeubi:
> > Out of curiosity: Is there a benefit to using the `Image`, rather than e.g. the `ImageDataProvider`? This now creates an additional handle that needs to be disposed together with the Cursor. Which is fine, of course, just not very convenient. Latter might also play a little nicer with the `ImageDescriptor`, because you could simply create the cursor via `new Cursor(device, imageDescriptor::getImageData, hotspotX, hotspotY)`.
> 
> If cursors are initialized with a single ImageData and later requested at a different zoom level, that image data is auto-scaled using DPUtil. This just scales up the original ImageData, which makes the cursor appear blurry at different zoom levels (e.g., when moving the cursor from one monitor to another). If Image is used this scaling can be non blurry.

I think this does not really answers the question why not using `ImageDataProvider` in the first place (that might or might not construct an internal image and dispose it appropriately later on).

akoch-yatta:
> > > Out of curiosity: Is there a benefit to using the `Image`, rather than e.g. the `ImageDataProvider`? This now creates an additional handle that needs to be disposed together with the Cursor. Which is fine, of course, just not very convenient. Latter might also play a little nicer with the `ImageDescriptor`, because you could simply create the cursor via `new Cursor(device, imageDescriptor::getImageData, hotspotX, hotspotY)`.

> > 
> > 
> > If cursors are initialized with a single ImageData and later requested at a different zoom level, that image data is auto-scaled using DPUtil. This just scales up the original ImageData, which makes the cursor appear blurry at different zoom levels (e.g., when moving the cursor from one monitor to another). If Image is used this scaling can be non blurry.
> 
> I think this does not really answers the question why not using `ImageDataProvider` in the first place (that might or might not construct an internal image and dispose it appropriately later on).

We discussed using `ImageDataProvider`, but thought using an Image directly would give the most freedom. Most usecases with cursors seem to be related to image files. Therefor, using an `ImageDataProvider` would make the implementation a bit more complex with pure SWT. Still, i see the argument with the additonal handle and the use case with `ImageDescriptor` would probably solve the added complexity we had in mind. So I'm actually liking the idea.

arunjose696:
I've updated the PR to make the constructor accept an `ImageDataProvider ` instead of an `Image`. This approach is preferable, as it avoids creating undisposed handles and works better with ImageDescriptor.

The only minor downside I can think is that cursors won't be scaled on GTK or macOS, since these platforms would set the cursor at 100% zoom and wouldnt update them later. However, I believe this is acceptable, as it matches the current behavior.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Cursor.java
```diff
@@ -382,6 +382,9 @@ void createNSCursor(int hotspotX, int hotspotY, byte[] buffer, int width, int he
  */
 public Cursor(Device device, ImageData source, int hotspotX, int hotspotY) {
 	super(device);
+	setupCursorFromImageData(source, hotspotX, hotspotY);
+}
+private void setupCursorFromImageData(ImageData source, int hotspotX, int hotspotY) {
```
akoch-yatta: I would add a blank line in between.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Cursor.java
```diff
@@ -448,6 +451,41 @@ public Cursor(Device device, ImageData source, int hotspotX, int hotspotY) {
...
+ * the desired cursor appearance, and the x and y coordinates of
+ * the <em>hotspot</em> (that is, the point within the area
+ * covered by the cursor which is considered to be where the
+ * on-screen pointer is "pointing").
+ * <p>
+ * You must dispose the cursor when it is no longer required.
+ * </p>
+ *
+ * @param device the device on which to allocate the cursor
+ * @param image the image for the cursor
```
akoch-yatta: The javadoc doesn't fit for this parameter anymore

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Cursor.java
```diff
@@ -355,6 +360,41 @@ public Cursor(Device device, ImageData source, int hotspotX, int hotspotY) {
...
+ * the desired cursor appearance, and the x and y coordinates of
+ * the <em>hotspot</em> (that is, the point within the area
+ * covered by the cursor which is considered to be where the
+ * on-screen pointer is "pointing").
+ * <p>
+ * You must dispose the cursor when it is no longer required.
+ * </p>
+ *
+ * @param device the device on which to allocate the cursor
+ * @param image the image for the cursor
```
akoch-yatta: update javadoc

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_Cursor.java
```diff
@@ -153,6 +154,15 @@ public void test_ConstructorLorg_eclipse_swt_graphics_DeviceLorg_eclipse_swt_gra
...
 
+@Test
+public void test_ConstructorLorg_eclipse_swt_graphics_DeviceLorg_eclipse_swt_graphics_ImageData() {
+	// Test new Cursor(Device device, ImageData source, ImageData mask, int
+	// hotspotX, int hotspotY)
+	Image sourceImage = new Image(display, 10, 10);
+	Cursor cursor = new Cursor(display, sourceImage, 0, 0);
+	cursor.dispose();
+}
+
```
akoch-yatta: This test does not compile anymore - probably why the builds are red (didn't have a look into the logs)
And I would propose to give new tests a proper name. We can still start with test... but I would propose something like testConstructorWithImageDataProvider

arunjose696: Fixed and renamed

arunjose696:
Have all CI checks passing.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_Cursor.java
```diff
@@ -153,6 +154,15 @@ public void test_ConstructorLorg_eclipse_swt_graphics_DeviceLorg_eclipse_swt_gra
 	}
 }
 
+@Test
+public void test_ConstructorWithImageDataProvider() {
+	// Test new Cursor(Device device, ImageData source, ImageData mask, int
+	// hotspotX, int hotspotY)
+	Image sourceImage = new Image(display, 10, 10);
+	Cursor cursor = new Cursor(display, sourceImage::getImageData, 0, 0);
+	cursor.dispose();
```
akoch-yatta, line 164: Call to sourceImage.dispose() is missing to clean up the image and there is no real assertion in this test. What is it supposed to test/verify?

arunjose696: I’ll add the missing sourceImage.dispose() call to properly clean up the resource.

This test’s purpose is simply to verify that the Cursor constructor with the ImageData supplier doesn’t throw an exception during creation. The other test_Constructor* tests cover similar checks for different Cursor constructors, so this one follows the same pattern.

akoch-yatta: Ah okay, but then we should test the error cases that are specific for this constructor-> imageDataProvider == null

arunjose696: Done.

fedejeanne:
Failing test is unrelated #2098

PR #2278, opened 2025-07-01
Refactor: Move Cocoa-specific DPI methods to CocoaDPIUtil

ShahzaibIbrahim:
The `autoScale` methods no longer have relevance in the Win32 implementation, so this commit extracts the macOS-specific DPI scaling logic into a dedicated `CocoaDPIUtil` class. This improves platform separation and maintains cleaner, OS-specific code boundaries.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
@@ -1822,11 +1822,28 @@ public String toString () {
  * @noreference This method is not intended to be referenced by clients.
  */
 public static void drawScaled(GC gc, Image original, int width, int height, float scaleFactor) {
-	gc.drawImage (original, 0, 0, DPIUtil.autoScaleDown (width), DPIUtil.autoScaleDown (height),
+	gc.drawImage (original, 0, 0, CocoaDPIUtil.autoScaleDown (width), CocoaDPIUtil.autoScaleDown (height),
 			/* E.g. destWidth here is effectively DPIUtil.autoScaleDown (scaledWidth), but avoiding rounding errors.
 			 * Nevertheless, we still have some rounding errors due to the point-based API GC#drawImage(..).
 			 */
-			0, 0, Math.round (DPIUtil.autoScaleDown (width * scaleFactor)), Math.round (DPIUtil.autoScaleDown (height * scaleFactor)));
+			0, 0, Math.round (CocoaDPIUtil.autoScaleDown (width * scaleFactor)), Math.round (CocoaDPIUtil.autoScaleDown (height * scaleFactor)));
+}
+
+private final class CocoaDPIUtil {
+
+	/**
+	 * Auto-scale down int dimensions.
+	 */
+	public static int autoScaleDown(int size) {
+		return DPIUtil.scaleDown(size, DPIUtil.getDeviceZoom());
+	}
+
+	/**
+	 * Auto-scale down float dimensions.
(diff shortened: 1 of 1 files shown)


BeckerWdf:
> LGTM but since this is for Mac, I'd like a second opinion. @BeckerWdf ?

What exactly should I comment on? Should I test the change? If yes: How can I test this?

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
```diff
@@ -1822,11 +1822,28 @@ public String toString () {
...
+			0, 0, Math.round (CocoaDPIUtil.autoScaleDown (width * scaleFactor)), Math.round (CocoaDPIUtil.autoScaleDown (height * scaleFactor)));
+}
+
+private final class CocoaDPIUtil {
+
+	/**
+	 * Auto-scale down int dimensions.
+	 */
+	public static int autoScaleDown(int size) {
+		return DPIUtil.scaleDown(size, DPIUtil.getDeviceZoom());
```
BeckerWdf, line 1838: In the before state we called
`DPIUtil.autoScaleDown (width)`
and not 
`DPIUtil.scaleDown(size, DPIUtil.getDeviceZoom()`

Is this change correct?

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
```diff
@@ -1822,11 +1822,28 @@ public String toString () {
...
+	}
+
+	/**
+	 * Auto-scale down float dimensions.
+	 */
+	public static float autoScaleDown(float size) {
+		return DPIUtil.scaleDown(size, DPIUtil.getDeviceZoom());
```
BeckerWdf, line 1845: same as above

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
ShahzaibIbrahim: Yes it is correct. Practically in autoScaleDown method it does the same thing (use the device zoom). This change is made to get rid of autoScaleDown method in DPIUtil (will be done in another PR).

fedejeanne:
> > LGTM but since this is for Mac, I'd like a second opinion. @BeckerWdf ?
> 
> What exactly should I comment on? Should I test the change? If yes: How can I test this?

This PR brings no functional changes with it, it is merely a refactoring. The test consists in running Eclipse performing any action (open a view, edit something, drag stuff around) that runs through the changed method and **check that it behaves as it did before**. 

Since this is done in the context of monitors and zooms, you should try it with different zoom levels. The best way is to have 2 monitors with different zooms and test the same things moving the Workbench/Views between the monitors but if you only have 1 monitor, you could simply try by changing the zoom level between operations.

Usually, this kind of testing can be done in no more than 10 minutes :-) 

Thank you!

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
BeckerWdf: then this looks good to me.

fedejeanne:
Thank you for reviewing!

The failing check is unrelated (a test failure on a previous run: https://github.com/eclipse-platform/eclipse.platform.swt/issues/2098)

PR #2280, opened 2025-07-01
Refactor: Move GTK-specific DPI logic to GtkDPIUtil

ShahzaibIbrahim:
This commit introduces a dedicated `GtkDPIUtil` class to isolate GTK-specific DPI logic, improving platform separation and maintainability.

Since the GTK scaling factor is always 1, the `autoScaleUp` and `autoScaleDown` methods effectively return the input unchanged. These methods and their usages have been removed from the GTK implementation to eliminate unnecessary code.

Also clarifies that `autoScale` no longer applies to the Win32 implementation and is being decoupled accordingly.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
@@ -284,7 +284,7 @@ public static ImageData scaleImageData (Device device, final ElementAtZoom<Image
 	return scaleImageData(device, elementAtZoom.element(), targetZoom, elementAtZoom.zoom());
 }
 
-private static ImageData autoScaleImageData (Device device, final ImageData imageData, float scaleFactor) {
+public static ImageData autoScaleImageData (Device device, final ImageData imageData, float scaleFactor) {
 	// Guards are already implemented in callers: if (deviceZoom == 100 || imageData == null || scaleFactor == 1.0f) return imageData;
 	int width = imageData.width;
 	int height = imageData.height;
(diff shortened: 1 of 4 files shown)


bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/internal/GtkDPIUtil.java
```diff
@@ -0,0 +1,53 @@
...
+	 * Auto-scale up ImageData to device zoom that is at 100%.
+	 */
+	public static ImageData autoScaleUp (Device device, final ImageData imageData) {
+		return autoScaleImageData(device, imageData, 100);
+	}
+
+	/**
+	 * Auto-scale ImageData to device zoom that are at given zoom factor.
+	 */
+	public static ImageData autoScaleImageData (Device device, final ImageData imageData, int imageDataZoomFactor) {
```
akoch-yatta: Where is this method used? I only found the upper one used

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/internal/GtkDPIUtil.java
```diff
@@ -0,0 +1,53 @@
...
+ * native widgets.
+ * </p>
+ * <p>
+ * The {@code autoScaleDown(..)} convert from high DPI pixels to API coordinates
+ * (in SWT points).
+ * </p>
+ *
+ * @since 3.105
+ */
+public class GtkDPIUtil {
```
akoch-yatta: If the only two remaining calls are from Image, I would not add a new internal API class, but internalize it into Image, and still call DPIUtil directly from there

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/internal/GtkDPIUtil.java
ShahzaibIbrahim: Its not used anywhere. actually. I could just add the logic inside of autoScaleUp method.

fedejeanne (approved):
LGTM but since this is GTK, I would like the opinion of @iloveeclipse and/or @akurtakov before merging it.

PR #2283, opened 2025-07-03
CCombo Border

DenisUngemach:


laeubi:
@DenisUngemach as you seem to open PRs [quite often](https://github.com/eclipse-platform/eclipse.platform.swt/pulls?q=is%3Apr+is%3Aclosed+author%3ADenisUngemach) against the "wrong" repository, I would recommend you do the following:

1. go to the repository that you actually want to provide PRs
2. fork that repository and choose a different name (e.g. eclipse.platform.swt-i31)
3. the checkout and push to that repository

Then github should suggest you to open PRs against that fork by default.

Issue #2287, opened 2025-07-08
[GTK4] Missing x,y values for MouseUp event, Snippet62


Theo-Dann-Muirhead-Renesas:
Issue was a part of #344  but has a different root cause so I'm separating it. 

Running Snippet 62 with GTK4 and clicking on it outputs Down/Up  events. The Down events have associated x and y values which are missing for the Up events. This behaviour does not happen with GTK3. 

Head of SWT, Ubuntu 24.04.2 LTS, GTK 4.14.2

akurtakov:
Close as PR was merged.

Issue #344, opened 2022-08-24
[GTK4] Double click registers as one DOWN and two UP events instead of two of each


joel-majano:
Run Snippet 62 in GTK4 then double click in the window. You will see a lot of move events, and then one single "DOWN" event followed by two "UP" events. There should be two of each in the "DOWN, UP, DOWN, UP" sequence which can be obtained when using this snippet in GTK3.

Environment: Fedora 36, GTK 4.6.5, Wayland, OpenJDK17

lathapatil:
> start the snipped with SWT_GTK4=1 env variable.

Thank you for the quick reply. Does this also apply to JUnit testing and SWT examples?

akurtakov:
> > start the snipped with SWT_GTK4=1 env variable.
> 
> Thank you for the quick reply. Does this also apply to JUnit testing and SWT examples?

Yes, if the env variable is set SWT will try load GTK 4 libs and fallback to Gtk 3.x if loading 4.x fails.

lathapatil:
I observed some test failures when SWT_GTK=1 was set for AllWidgetTests, eventually leading to a testcase that rendered Eclipse unresponsive and caused it to close.

Perhaps we should consider creating an issue to address these problems once somebody else confirm the behavior.

akurtakov:
> I observed some test failures when SWT_GTK=1 was set for AllWidgetTests, eventually leading to a testcase that rendered Eclipse unresponsive and caused it to close.
> 
> Perhaps we should consider creating an issue to address these problems once somebody else confirm the behavior.

I am not aware of anyone else looking into this so this could mean waiting indefinetely.

jonahgraham:
Confirmed in HEAD of SWT + Ubuntu 25.04 + GTK4 4.18.4+ds-1 

On GTK3 double click in the window gives this output:

```
DOWN: button: 1, stateMask=0x0, x=102, y=23
UP: button: 1, stateMask=0x80000, x=102, y=23 BUTTON1
DOWN: button: 1, stateMask=0x0, x=102, y=23
UP: button: 1, stateMask=0x80000, x=102, y=23 BUTTON1
```

And on GTK4:

```
DOWN: button: 1, stateMask=0x0, x=81, y=19
UP: button: 1, stateMask=0x80000, x=0, y=0 BUTTON1
UP: button: 1, stateMask=0x80000, x=0, y=0 BUTTON1
```

[...]

PR #2288, opened 2025-07-08
Add missing x, y coordinates to gtk4 mouse event.

Theo-Dann-Muirhead-Renesas:
Summary:

Originally an issue pointed out from: #344 but has a different cause so I split it to a separate ticket:  #2287 

On GTK4 "MouseUp" events were missing x and y values. This was dues to missing variable inputs to Control.SendMouseEvent().


akurtakov:
This looks like really good catch.

akurtakov:
Thanks @Theo-Dann-Muirhead-Renesas ! Keep'em coming.

PR #2289, opened 2025-07-08
Mark Deprecated getBoundsInPixel for removal

ShahzaibIbrahim:
The deprecated method getBoundsInPixel() in Image is marked for removal along with its tests.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Image.java
@@ -1014,7 +1014,7 @@ public Rectangle getBounds() {
  * @deprecated This API doesn't serve the purpose in an environment having
  *             multiple monitors with different DPIs, hence deprecated.
  */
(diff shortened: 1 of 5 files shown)


bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Image.java
```diff
@@ -996,7 +996,11 @@ public Color getBackground() {
  */
 public Rectangle getBounds() {
 	if (isDisposed()) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
-	return getBoundsInPixels();
+	if (width != -1 && height != -1) {
+		return new Rectangle(0, 0, width, height);
+	}
+	// There are no pixmaps on GTK3, so this will just return 0
+	return new Rectangle(0, 0, 0, 0);
```
HeikoKlare: Maybe we should keep the existing method call and inline the method upon removal of the deprecated method instead of duplicating code now.

HeikoKlare (approved):

PR #2291, opened 2025-07-09
Fix Double Click not Registering Snippet 62

Theo-Dann-Muirhead-Renesas:
Fix for issue 344. 

Running snippet 62 and double clicking used to register two up events but only one down event when there should also have been two. This was due to the use of SWT.MouseDoubleClick instead of SWT.MouseDown. I have modified it such that in the event SWT.MouseDoubleClick fails, SWT.MouseDown is tried instead. This keeps the double click as an option for the future if support is added.  

I also changed the behaviour so triple clicks and higher are now treated as double clicks rather than being ignored.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Control.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Control.java
@@ -3432,8 +3432,15 @@ int gtk_gesture_press_event (long gesture, int n_press, double x, double y, long
 				}
 			}
 		}
-	} else if (n_press == 2) {
+	} else if (n_press >= 2) {
 		boolean cancelled = sendMouseEvent(SWT.MouseDoubleClick, eventButton, n_press, 0, false, eventTime, x, y, false, eventState);
+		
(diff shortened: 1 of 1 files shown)


akurtakov:
Just as a note for the future - please squash all your commits into single one for such small patches. Merge commits are strongly discouraged in this repo.

Theo-Dann-Muirhead-Renesas:
How do I squash the commits in my fork? I can't see a way to do that

akurtakov:
https://www.datacamp.com/tutorial/git-squash-commits looks like good explanation.

akurtakov:
image
from Egit - open History view, select the commits you want to squash, right click, Modify, Squash

Theo-Dann-Muirhead-Renesas:
{D0E10337-6F9C-41DF-BD29-66ACAE0C56B3}
I do not have that option

akurtakov:
Most likely this is because you try to squash commits from different chains and also include merge commits. 
It might be easier to start a new branch from master and cherry-pick only the commit you want in it. 
Try to always rebase and not merge to reduce cases like that.

Theo-Dann-Muirhead-Renesas:
I made a new pull request with just the one commit

Issue #2290, opened 2025-07-09
[GTK] Background images behavior differs from Windows and Mac


N1k145:
**Describe the bug**
When setting a background image and a background color on an composite, the Color is shown
On Windows and Mac the background image is show
Relates to https://github.com/eclipse-platform/eclipse.platform.ui/issues/3079

**To Reproduce**
```java
import org.eclipse.jface.layout.FillLayoutFactory;
import org.eclipse.jface.resource.ImageDescriptor;
import org.eclipse.swt.SWT;
import org.eclipse.swt.custom.CLabel;
import org.eclipse.swt.graphics.Color;
import org.eclipse.swt.graphics.RGB;
import org.eclipse.swt.layout.GridLayout;
import org.eclipse.swt.widgets.Composite;
import org.eclipse.swt.widgets.Display;
import org.eclipse.swt.widgets.Shell;

public class Main {

	public static void main(String[] args) {
		Display display = new Display();

		Shell shell = new Shell(display);

		FillLayoutFactory.fillDefaults().margins(0, 0).applyTo(shell);

		final Composite compositeWithImage = new Composite(shell, SWT.NONE);
		final GridLayout gridLayout = new GridLayout(2, false);
		gridLayout.marginHeight = 0;
		gridLayout.marginWidth = 0;
		compositeWithImage.setLayout(gridLayout);

		compositeWithImage.setBackgroundImage(ImageDescriptor.getMissingImageDescriptor().createImage());
		compositeWithImage.setBackground(new Color(new RGB(0, 255, 0)));
		
		final CLabel test1 = new CLabel(compositeWithImage, SWT.NONE | SWT.WRAP);
		test1.setText("Lorem ipsum dolor sit amet, consetetur sadipscing elitr");
		final CLabel test2 = new CLabel(compositeWithImage, SWT.NONE | SWT.WRAP);
		test2.setText("Lorem ipsum dolor sit amet, consetetur sadipscing elitr");
		final CLabel test3 = new CLabel(compositeWithImage, SWT.NONE | SWT.WRAP);
		test3.setText("Lorem ipsum dolor sit amet, consetetur sadipscing elitr");
		final CLabel test4 = new CLabel(compositeWithImage, SWT.NONE | SWT.WRAP);
		test4.setText("Lorem ipsum dolor sit amet, consetetur sadipscing elitr");
		final CLabel test5 = new CLabel(compositeWithImage, SWT.NONE | SWT.WRAP);
[...]

PR #2293, opened 2025-07-09
[GTK] Background images behavior differs from Windows and Mac

N1k145:
Document the intended behavior

Contributes to:
https://github.com/eclipse-platform/eclipse.platform.swt/issues/2290

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
@@ -3072,6 +3072,9 @@ void setBackground () {
  * <p>
  * Note: This operation is a hint and may be overridden by the platform.
  * </p>
+ * <p>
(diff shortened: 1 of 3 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
```diff
@@ -3114,6 +3117,10 @@ private void _setBackground (Color color) {
  * Note: This operation is a hint and may be overridden by the platform.
  * For example, on Windows the background of a Button cannot be changed.
  * </p>
+ * <p>
+ * Note: Setting a background image overrides a set background color.
+ * </p>
+ *
```
akurtakov: Please remove this extra line that is only in win32 as the javadoc basher will complain about it at end of release.

N1k145: Done. I did not notice that in the first look...

PR #2297, opened 2025-07-09
Fix Double Click not Registering Snippet 62 

### LINKED ISSUE #344: [GTK4] Double click registers as one DOWN and two UP events instead of two of each

Theo-Dann-Muirhead-Renesas:
Fixes #344.

Running snippet 62 and double clicking used to register two up events but only one down event when there should also have been two. This was due to the use of SWT.MouseDoubleClick instead of SWT.MouseDown. I have modified it such that in the event SWT.MouseDoubleClick fails, SWT.MouseDown is tried instead. This keeps the double click as an option for the future if support is added.

I also changed the behaviour so triple clicks and higher are now treated as double clicks rather than being ignored.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Control.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Control.java
@@ -3428,8 +3428,15 @@ int gtk_gesture_press_event (long gesture, int n_press, double x, double y, long
 				}
 			}
 		}
-	} else if (n_press == 2) {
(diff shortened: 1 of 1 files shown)


akurtakov:
Thanks for the patch. Merging as it's step in the right direction even if there is a lot more to support.

PR #2304, opened 2025-07-10
Use font device when setting the font instead of gc device

ShahzaibIbrahim:
When a font is initialized with a Printer as its device, it should only be used with a GC that was also created from that same Printer. Using such a font with a GC tied to a different device, like a Display, leads to a device mismatch. Since the printer may have a much higher DPI (e.g., 600 DPI) compared to the display (typically 96 DPI), this mismatch causes the text to render extremely large.

To explain I am adding a snippet here:

```
public class FontDeviceMismatchExample {
		// Create the shell
		shell.setText("Font Device Mismatch Demo");
		shell.setSize(400, 200);
		shell.open();

		// Creating font using printer DPI = ~600
		Printer printer = new Printer();
		Font printerFont = new Font(printer, "Arial", 12, SWT.NORMAL);
               
		GC gc = new GC(shell); // creating gc with shell that uses display DPI = ~96

		gc.setFont(printerFont); // here device mismatch
		gc.drawText("This text may appear huge due to DPI mismatch", 20, 50);

		while (!shell.isDisposed()) {
				display.sleep();
			}
		}

		// Cleanup
		gc.dispose();
		printerFont.dispose();
		printer.dispose();
		display.dispose();
	}
}
```

### Expected behavior

The correct behavior of this snippet should be Oversized text in a shell due to device mismatch. But that doesn't happen without proposed change. **WHY?** because while setting the font we used the display of the **GC** instead of **Font** which is incorrect.


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -4943,7 +4943,7 @@ private class SetFontOperation extends Operation {
 
 	@Override
 	void apply() {
-		data.font = font != null ? SWTFontProvider.getFont(device, font.getFontData()[0], data.nativeZoom) : SWTFontProvider.getSystemFont(device, data.nativeZoom);
+		data.font = font != null ? SWTFontProvider.getFont(font.getDevice(), font.getFontData()[0], data.nativeZoom) : SWTFontProvider.getSystemFont(font.getDevice(), data.nativeZoom);
```
arunjose696: this will crash with a NullPointerException if font is null, because the else part still calls font.getDevice()

arunjose696:
I am slightly confused why we use data.nativeZoom in setFont() instead of GC.getZoom()

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
ShahzaibIbrahim: You're right. Also system fonts are always created with device of GC, so it make sense to use gc display.

PR #2306, opened 2025-07-10
Use font zoom in stringExtent and textExtent

### LINKED ISSUE #2311: Line number remains is too small when setting the zoom with -Dswt.autoScale
Let's make sure issue is not already fixed in latest builds first.

- [x] I verified I can reproduce this issue against [latest Integration Build of Eclipse SDK](https://download.eclipse.org/eclipse/downloads/)

## Steps to reproduce

On a 100% monitor, set these parameters:
```
-Dswt.enable.autoScale=true
-Dswt.autoScale=150
-Dswt.autoScale.method=nearest
```

**Screenshots**

**In 4.37**
Image

**In 4.33**

Image

**Expected Behavior**
The line numbers should fit in the ruler, either they should have the same size as other fonts (preferred outcome) or the ruler should be wider so that the line numbers fit in it.

## Tested under this environment:
* OS & version: Windows 10
* Eclipse IDE/Platform version (as shown in *Help > About*):  2025-09 (4.37). Build id: I20250625-2330

## Community

- [x] I understand reporting an issue to this OSS project does not mandate anyone to fix it. Other contributors may consider the issue, or not, at their own convenience. The most efficient way to get it fixed is that I fix it myself and contribute it back as a good quality patch to the project.

arunjose696:
Previously, stringExtent calculated the width of a string in pixels using GC.getZoom(), which could lead to inconsistencies if the provided font had a different zoom level. This change ensures that the pixels-to-points conversion uses the zoom of the specified font.

Fixes #2311 

Reproduction Steps
On a 100% monitor, set these parameters:

```
-Dswt.enable.autoScale=true
-Dswt.autoScale=150
-Dswt.autoScale.method=nearest
```

Screenshots

<table>
  <tr>
    <td align="center" valign="top">
      
      <br/>
      <sub>Current Behaviour</sub>
    </td>
    <td align="center" valign="top">
      
      <br/>
      <sub>With This Change</sub>
    </td>
  </tr>
</table>

Have tested the changes on zooms **100,150,200&225** with` -Dswt.autoScale`=**150 ,100 &200,**

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
@@ -5644,7 +5644,7 @@ private class SetTransformOperation extends Operation {
  */
 public Point stringExtent (String string) {
 	if (string == null) SWT.error (SWT.ERROR_NULL_ARGUMENT);
-	return DPIUtil.scaleDown(drawable, stringExtentInPixels(string), getZoom());
+	return DPIUtil.scaleDown(drawable, stringExtentInPixels(string), data.font.zoom);
 }
 
 Point stringExtentInPixels (String string) {
@@ -5724,7 +5724,7 @@ public Point textExtent (String string) {
  * </ul>
  */
 public Point textExtent (String string, int flags) {
-	return DPIUtil.scaleDown(drawable, textExtentInPixels(string, flags), getZoom());
+	return DPIUtil.scaleDown(drawable, textExtentInPixels(string, flags), data.font.zoom);
 }
 
 Point textExtentInPixels(String string, int flags) {


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -5724,7 +5724,7 @@ public Point textExtent (String string) {
  * </ul>
  */
 public Point textExtent (String string, int flags) {
-	return DPIUtil.scaleDown(drawable, textExtentInPixels(string, flags), getZoom());
+	return DPIUtil.scaleDown(drawable, textExtentInPixels(string, flags), data.font.zoom);
```
fedejeanne, line 5727: Is this one also necessary? It doesn't affect the line number ruler AFAICT so it is probably for another (broken?) use case. Could you tell me which one?

The good thing is that this change didn't break anything either.

fedejeanne (commented):
Thank you for the contribution, @arunjose696 !

In general it looks good and I couldn't find any regressions with this PR. I just have a question regarding the necessity of one change.

Also, have you by any chance tried to make the font in the line number ruler smaller so that it has the same size of the code? I assume this would mean that the height of lines in the code would also change, making the code look more compact (which I like) but this is just a thought.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
arunjose696: I made this change so that textExtent and stringExtent behave the same. To me, both methods share the same logic the only difference is that one uses flags and the other does not. So, it wouldn’t make sense to use different scaledown zooms

arunjose696:
> Also, have you by any chance tried to make the font in the line number ruler smaller so that it has the same size of the code? I assume this would mean that the height of lines in the code would also change, making the code look more compact (which I like) but this is just a thought.

I tried this, but it turned out to be a bit tricky. The GC used to draw the line number ruler is [created from a BufferedImage](https://github.com/eclipse-platform/eclipse.platform.ui/blob/master/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/source/LineNumberRulerColumn.java#L709)(which is a Image type), and it inherits the font zoom level from the last requested zoom for the image. This zoom is used by the GC for drawing. I haven’t found a straightforward way to reset the nativeZoom value of a GC that’s created from an image.

For experimentation, I was able to make this work on the first draw of the line number ruler by explicitly calling BufferedImage.getImageData(LOWER_ZOOM) just before the GC was created, which as a side effect sets the currentZoom of the image to LOWER_ZOOM. So when the GC is created from that image, it uses the desired zoom. However, later when switching monitors, SWT will request a new image using the zoom specified by swt.autoScale. 

I feel this route would be much more complicated compared to current approach.

fedejeanne:
I see. Thank you for the fix and the thorough response, @arunjose696 !

If the size of the font in the line number ruler becomes an issue then we can circle back to the analysis. But I have no strong opinion about it and there are other things in more dire need of attention right now. I accepted your fix.

PR #2314, opened 2025-07-14
[Refactor DPIUtil] rename scaleUp/Down

ShahzaibIbrahim:
Renaming scaleDown and ScaleUp methods to pixelToPoint and pointToPixel respectively. The implementation detail (using a scale factor in body) doesn't clearly convey the intent of the method from its name. When someone (me) reads scaleDown(x) or scaleUp(x), the natural question is:

**“Scaling relative to what? What coordinate system or unit is this about?”**

That’s ambiguous without diving into the implementation.
In contrast, renaming to pixelToPoint(x) and pointToPixel(x) makes it immediately clear. We’re converting between pixel units (physical/screen space) and point units (logical/UI space). The direction of conversion is unambiguous.

The changes are done with **Eclipse Refactoring tool**. There are no manual changes.


arunjose696:
I tested the refactorings locally, and the changes look good to me.

PR #2321, opened 2025-07-16
Fixes NPE in StyledText on DPI Resizing

### LINKED ISSUE #2320: NPE in Styled Text Caret resizing
**Describe the bug**
When a Styled Text, that has a "null" Caret set, switches to another monitor an NPE is thrown

**Expected behavior**
No NPE as this is a valid state according to the rest of the code

**Environment:**
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 

3. JRE/JDK version
 
**Version since**
 Eclipse or SWT version since when the behavior is seen 4.35

Do not set a null caret

N1k145:
fixes #2320


HeikoKlare (approved):
Thank you for the fix, @N1k145!
Change is completely sound.

Issue #2308, opened 2025-07-10
Cursors not correctly sized when using swt.autoScale on a HighDPI screen.


ptziegler:
I'm using a monitor with 150% zoom. Now consider the following snippet:

```java
package test;

import org.eclipse.swt.SWT;
import org.eclipse.swt.graphics.Color;
import org.eclipse.swt.graphics.Cursor;
import org.eclipse.swt.graphics.GC;
import org.eclipse.swt.graphics.Image;
import org.eclipse.swt.graphics.ImageData;
import org.eclipse.swt.widgets.Display;
import org.eclipse.swt.widgets.Shell;

public class Snippet {
        public static void main(String[] args) {
                System.setProperty("swt.autoScale", "200");
                Cursor cursor = new Cursor(null, getImageData(100), 0, 0);

                Shell shell = new Shell(SWT.NO_TRIM);
                shell.setSize(100, 100);
                shell.setCursor(cursor);
                shell.setBackground(new Color(0, 255, 0));
                shell.open();

                Display display = shell.getDisplay();
                while (!display.isDisposed()){
                        display.readAndDispatch();
                }
                cursor.dispose();
        }

        private static ImageData getImageData(int zoom) {
                Image image = new Image(null, 50, 50);
                try {
                        GC gc = new GC(image);
                        gc.setBackground(new Color(255, 0, 0));
                        gc.fillRectangle(0, 0, 50, 50);
                        return image.getImageData(zoom);
                } finally {
                        image.dispose();
                }
        }
}
```

Expectation: The cursor should always take up 25% of the shell.

SWT 3.128.0 (2024-12), exactly right:
<img width="200" height="200" alt="Image" src="https://github.com/user-attachments/assets/8d39da7a-c82f-4651-a181-813fc8b155b0" />

SWT 3.129.0 (2025-03), **way** too big:
<img width="205" height="204" alt="Image" src="https://github.com/user-attachments/assets/198340e8-3f40-4135-b22d-5f5509eae575" />

SWT 3.130.0 (2025-06), too big:
<img width="200" height="200" alt="Image" src="https://github.com/user-attachments/assets/8c414bc0-6ee2-4e38-afd2-57a00359e807" />

SWT 3.131.0 (2025-09), too small:
<img width="200" height="200" alt="Image" src="https://github.com/user-attachments/assets/69d8fbe9-5d84-4a9e-a95c-77ae6e5ec95b" />

ptziegler:
[...]

PR #2322, opened 2025-07-17
Fix incorrect cursor scaling by using autoscale zoom

ShahzaibIbrahim:
---

### 🐞 **Problem Description**

In SWT, custom cursors were being created with incorrect sizes on high-DPI displays. The root cause was this line:

```java
OS.SetCursor(Cursor.win32_getHandle(cursor, getNativeZoom()));
```

The method `Cursor.win32_getHandle(...)` used `getNativeZoom()` to determine the scale at which the cursor image should be rendered.

However, **`getNativeZoom()` only reflects the raw device DPI scaling** (e.g., 125%, 150%), and does **not** account for SWT’s autoscale policy (controlled via the `swt.autoScale` property).

This leads to inconsistencies:

* Shells, fonts, and other UI elements in SWT may scale according to a **rounded zoom factor** (e.g., 100%, 200%),
* But the cursor was scaled using the unrounded native zoom, causing it to appear **too small or too large** compared to the rest of the UI.

---

### ✅ **The Fix**

To align cursor scaling with the rest of the SWT UI, the cursor zoom should be computed using the same logic that SWT uses for autoscaling across the toolkit.

Replace:

```java
Cursor.win32_getHandle(cursor, getNativeZoom())
```

With:

```java
Cursor.win32_getHandle(cursor, DPIUtil.getZoomForAutoscaleProperty(getNativeZoom()))
```

This uses:

* `DPIUtil.getZoomForAutoscaleProperty(...)`, which applies the logic from `swt.autoScale` (e.g., rounding zoom levels up to the nearest 100%)
* Ensures cursor size matches the size expectations of other UI components (like Shells, Fonts, Images)

---

###  **Various Cases with Snippets**

```java
public class Snippet {
		//System.setProperty("swt.autoScale", "int200");
		Cursor cursor = new Cursor(display, getImageData(100), 0, 0);
		Shell shell = new Shell(display, SWT.NO_TRIM);
		shell.setSize(100, 100);
		shell.setCursor(cursor);
		shell.setBackground(new Color(0, 255, 0));
		shell.open();
		while (!display.isDisposed()) {
			display.readAndDispatch();
		}
		cursor.dispose();
	}

	private static ImageData getImageData(int zoom) {
		Image image = new Image(null, 50, 50);
		try {
			GC gc = new GC(image);
			gc.setBackground(new Color(255, 0, 0));
			gc.fillRectangle(0, 0, 50, 50);
			return image.getImageData(zoom);
		} finally {
			image.dispose();
		}
	}
}
```

**- autoScale=1000**

**- autoScale=quarter, zoom = 175%**

**- autoScale=int200, zoom = 175%**


fedejeanne:
> The code looks fine, I'm just confused about which issue does this PR solve (and how to test it). Is it #2308 ? 

I see that this is documented in our internal issue (https://github.com/vi-eclipse/Eclipse-Platform/issues/357). You just need to adapt the text of the commit and this PR's description to say that if fixes #2308.

> The screenshots you provided look similar to the ones in the issue but I don't see the "cursor" in them (the red square).

Scratch that, I see now that they are GIFs and they ran only once so I missed the red square. It's there though 👍

fedejeanne:
The API check failure is unrelated.

PR #2324, opened 2025-07-17
Ensure sendMeasureItemEvent is only called for OS.TTN_SHOW or custom tooltips

arunjose696:
Prior to PR #2241, `Table.sendMeasureItemEvent` was only invoked when the tooltip was not a CustomTooltip. However, after that change, `sendMeasureItemEvent()` is now called for all tooltips.

This results in an unintended side effect: `sendMeasureItemEvent() ` compares the item's width(obtained from `TableItem.getBounds()`) to the width of column  it is present in (obtained from `OS.LVM_GETCOLUMNWIDTH`) and sometimes concludes incorrectly that the item doesn't fit. This causes an unnecessary scrollbar to appear, even though the item is fully contained within the column.

While the root cause is possibly in how column item width is calculated (the calculation involves a[ strange formula](https://github.com/eclipse-platform/eclipse.platform.swt/blob/4e274c245190db794eb3576e2ffa861e14b9361c/bundles/org.eclipse.swt/Eclipse%20SWT/win32/org/eclipse/swt/widgets/TableItem.java#L390) `rect.right += textRect.right - textRect.left + Table.INSET * 3 + 2;` which is not very intutive).   While this does change not fix the root issue in getBounds(), it aligns the current behavior with previous one before #2241

This workaround aligns with the original behavior prior to PR #2241 as sendMeasureItemEvent would not be called for non customToolTip and restores expected UI behavior for now.

**Steps to reproduce**

See the steps in https://github.com/vi-eclipse/Eclipse-Platform/issues/360

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Table.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Table.java
@@ -7188,23 +7188,21 @@ private LRESULT positionTooltip(NMHDR hdr, long lParam) {
 		if (newFont != 0) oldFont = OS.SelectObject (hDC, newFont);
 		long hFont = item.fontHandle (pinfo.iSubItem);
 		if (hFont != -1) hFont = OS.SelectObject (hDC, hFont);
-		Event event = sendMeasureItemEvent (item, pinfo.iItem, pinfo.iSubItem, hDC);
-		if (!isDisposed () && !item.isDisposed ()) {
-			RECT itemRect = new RECT ();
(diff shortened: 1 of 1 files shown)


amartya4256:
To simplify this, SendMeasureItemEvent should only be called in 2 cases:
1. When isCustomToolTip() is true: This compliments the state of the Table tooltip implmentation before Positioning of tooltip on a single monitor was implemented.
2. When TTN_SHOW is called: As we need to re-position any kind of tooltip.

Hence, the condition in which it must not be called on a WM_NOTIFY message is when isCustomTooltip() is false and the message is a TTN_GETDISPINFO as calling this method can modify the tooltip in this case.

PR #2326, opened 2025-07-21
Revert "[win32] Refresh GC with Image based on zoom change"

fedejeanne:
This reverts commit 8a8c67d14986abab9a2d7049cca2476ec8fcb7b0.


iloveeclipse:
> This reverts commit [8a8c67d](https://github.com/eclipse-platform/eclipse.platform.swt/commit/8a8c67d14986abab9a2d7049cca2476ec8fcb7b0).

Because of **what**? Regression?

fedejeanne:
> > This reverts commit [8a8c67d](https://github.com/eclipse-platform/eclipse.platform.swt/commit/8a8c67d14986abab9a2d7049cca2476ec8fcb7b0).
> 
> Because of **what**? Regression?

https://github.com/vi-eclipse/Eclipse-Platform/issues/362

HeikoKlare:
I guess this was created by accident. We currently investigate a potential regression, but will first focus on resolving it. If we don't manage to achieve that until M2, we may consider reverting the causing commit.

fedejeanne:
It was not created by accident but as a fail-safe in case we can't find a solution until M2. 

It will remain drafted so we can work on a better solution.

HeikoKlare:
Obsolete due to:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2328
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2331

PR #2328, opened 2025-07-21
Track and dispose resources created by operations in GC

arunjose696:
**Description:**

Each Operation subclass now maintains its own list of disposable SWT resources. These are added to a `disposables `list within the Operation and are disposed when the GC is disposed. This ensures cleanup of resources.

**Steps to Reproduce:**

There is currently an issue with SWT resource tracking due to monitor-specific scaling. To reproduce the warning:

1)Temporarily comment out [`resourcesWithZoomSupport.add(resource);`](https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/bundles/org.eclipse.swt/Eclipse%20SWT/win32/org/eclipse/swt/graphics/Device.java#L967)  in the Device class (i.e., the implementation of registerResourceWithZoomSupport):

2)Start the Eclipse runtime workspace with the following VM argument:

`-Dorg.eclipse.swt.graphics.Resource.reportNonDisposed=true`
3)Wait a  minute for the garbage collector to run.

You will then see the following warning in the logs:

```
!ENTRY org.eclipse.ui.ide 4 4 2025-07-22 10:57:59.955
!MESSAGE Not properly disposed SWT resource
!STACK 0 
java.lang.Error: SWT Resource was not properly disposed
```

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
@@ -1958,14 +1958,20 @@ private class DrawPathOperation extends Operation {
 	@Override
 	void apply() {
 		Path path = new Path(device, pathData);
-		long pathHandle = path.getHandle(getZoom());
-		if (pathHandle == 0) SWT.error(SWT.ERROR_INVALID_ARGUMENT);
-		initGdip();
-		checkGC(DRAW);
-		long gdipGraphics = data.gdipGraphics;
-		Gdip.Graphics_TranslateTransform(gdipGraphics, data.gdipXOffset, data.gdipYOffset, Gdip.MatrixOrderPrepend);
-		Gdip.Graphics_DrawPath(gdipGraphics, data.gdipPen, pathHandle);
-		Gdip.Graphics_TranslateTransform(gdipGraphics, -data.gdipXOffset, -data.gdipYOffset, Gdip.MatrixOrderPrepend);
+		try {
+			long pathHandle = path.getHandle(getZoom());
+			if (pathHandle == 0)
+				SWT.error(SWT.ERROR_INVALID_ARGUMENT);
+			initGdip();
+			checkGC(DRAW);
+			long gdipGraphics = data.gdipGraphics;
+			Gdip.Graphics_TranslateTransform(gdipGraphics, data.gdipXOffset, data.gdipYOffset, Gdip.MatrixOrderPrepend);
+			Gdip.Graphics_DrawPath(gdipGraphics, data.gdipPen, pathHandle);
+			Gdip.Graphics_TranslateTransform(gdipGraphics, -data.gdipXOffset, -data.gdipYOffset,
+					Gdip.MatrixOrderPrepend);
+		} finally {
+			path.dispose();
+		}
 	}
 }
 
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -5876,8 +5899,36 @@ private void createGcHandle(Drawable drawable, GCData newData, int nativeZoom) {
...
+
+private void disposeOperations() {
+    for (Operation op : operations) {
+        op.disposeAll();
+    }
+    operations.clear();
+}
+
 private abstract class Operation {
+	private final List<Resource> disposables = new ArrayList<>();
```
HeikoKlare, line 5913: Couldn't we track the resources directly in the GC instead of doing it in every operation?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -5876,8 +5899,36 @@ private void createGcHandle(Drawable drawable, GCData newData, int nativeZoom) {
...
 	abstract void apply();
+
+	protected void track(Resource resource) {
```
HeikoKlare: Maybe something like `disposeOnGcDisposal(Resource)` or `registerForDisplay(resource)` is more comprehensive than `track`?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -497,7 +497,8 @@ private class CopyAreaToImageOperation extends Operation {
 	private final int y;
 
 	CopyAreaToImageOperation(Image image, int x, int y) {
-		this.image = image;
+		this.image = new Image(image.getDevice(), image, SWT.IMAGE_COPY);
+		track(this.image);
```
HeikoKlare: Let's leave the adaptations to the image usages out of this PR and do it in a follow-up. It has to be solved in a more sophisticated way, so let's make this PR focus on disposing temporary resources.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
arunjose696: I believe the disposable resources are more closely tied to the operations themselves, so managing them within each Operation helps keep responsibilities seperated. Moving them into GC would make GC to be aware of and manage internal details of the operations, which introduces tight coupling. 

That said, if you still prefer disposal in GC, I can move the  tracking to GC class

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
arunjose696: Done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
arunjose696: Renamed to `registerForDisposal()`

PR #2331, opened 2025-07-21
[Win32] Avoid disposed image usage in GC operations

HeikoKlare:
When an image is drawn on a GC, that image may be disposed right afterwards. Since the GC can live much longer and a handle for that GC at a different zoom may be requested later on, such that the operations in the GC are re-applied, the stored image may be disposed by then. This can currently lead to errors because of access to disposed resources.

To avoid that, this change ensures that in case a drawn image becomes disposed, a copy of it is created and stored for the lifetime of the GC.

⚠️ This depends on the GC operation disposal logic of and thus contains those changes and needs to be merged after https://github.com/eclipse-platform/eclipse.platform.swt/pull/2328


arunjose696:
I have gone through the code, and it looks good. I also tested additional disposal scenarios for example, using multiple Image operations that refer to the same source Image which is then disposed. The fix works correctly in all cases I tested.

amartya4256 (approved):
Looks good.

Tests positively for both the cases:
1. The copy of the image is set to the GC before disposing.
2. On disposal on GC, no new copy of image is registered.

Approved.

Issue #2329, opened 2025-07-21
Sash bounds grow indefinitely while clicking or dragging if scaling is greater than 100%


ShadelessFox:
**Describe the bug**
Sash bounds increase rapidly when dragging the rubber band and then releasing it.

**To Reproduce**
1. Ensure scaling is set to 125% (in my case)
2. Ensure JVM option `-Dswt.autoScale=exact` is present

Use [Snippet 54](https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet54.java) as the baseline and add a listener for logging bounds of the sash:
```java
sash.addMouseListener(MouseListener.mouseUpAdapter(e -> System.out.println(sash.getBounds())));
```

Start clicking on the rubber band or drag it; notice increasing numbers being printed in the console

**Expected behavior**
Sash bounds stay the same when dragging the rubber band

**Screenshots**

https://github.com/user-attachments/assets/46cbe63c-d0d8-4fe5-8834-c390276d6c84

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 

2. Additional OS info (e.g. OS version, Linux Desktop, etc)
Windows 11

3. JRE/JDK version
openjdk version "24.0.1" 2025-04-15
 
**Version since**
SWT 4.36

**Workaround (or) Additional context**
I am not aware of any workarounds. Adding a mouseUp listener that would force the sash to layout might solve the issue.

HeikoKlare:
I cannot reproduce the issue on my system. It's also Windows 11 at 125% monitor zoom with the according VM argument. I tested with both the current SWT master state and the 4.36 state. It works fine for me: 
```
Rectangle {140, 0, 32, 262}
Rectangle {146, 0, 32, 262}
Rectangle {158, 0, 32, 262}
Rectangle {204, 0, 32, 262}
Rectangle {182, 0, 32, 262}
Rectangle {114, 0, 32, 262}
Rectangle {144, 0, 32, 262}
Rectangle {176, 0, 32, 262}
Rectangle {118, 0, 32, 262}
Rectangle {175, 0, 32, 262}
```

[...]

PR #2333, opened 2025-07-22
[Win32] Fix wrong width and height in Sash with SWT.SMOOTH when dragging

HeikoKlare:
When dragging a Sash with the SWT.SMOOTH, the width and height of the sash are erroneously scaled up by the current zoom. It is caused by a recent change which replaced a setBoundsInPixels with a setBounds call still using the same pixel-based width and height values.

This change adapts the according call to consistently use pixel-based values passed to the setBoundsInPixels method again.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2329


HeikoKlare:
@ShadelessFox can you test whether this change fixes the issue that you reproted?

ShadelessFox:
I confirm that these changes fix the issue :^)

amartya4256 (approved):
Fair! Since, the width and the height obtained are in Pixels, we should call setBoundsInPixels for that scale. Hence, approved.

PR #2337, opened 2025-07-22
Ensure font is fetched from registry even when zoom levels match

arunjose696:
This change ensures  font is fetched from registry even when zoom levels match.

The font returned by Font.win32_new is stored in the setFont() method of widgets. Previously, if targetZoom == font.zoom, the passed font was returned directly. However, the passed font could have been disposed by user code. This change ensures that the font is always fetched from the font registry, even when the zoom levels match, avoiding potential use of disposed fonts.

`Steps to reproduce `

The below snippet fails on master.

```
import org.eclipse.swt.*;

public class CanvasCaretFocus {
        shell.setSize(300, 200);
        shell.setText("Canvas Caret Focus");
        Canvas canvas = new Canvas(shell, SWT.BORDER);

        shell.open();
        Caret caret = new Caret(canvas, SWT.NONE);
        Font font = new Font(display, "Default", 10, SWT.BOLD);
    	caret.setFont(font );
    	font.dispose();

    	canvas.setFocus();
        canvas.setCaret(caret);



    }
}
```


HeikoKlare (commented):
Since you already have a simple reproducing snippet, maybe you could use that to add a regression test, so that we might see such a bug more early in the future?

arunjose696:
Closing this as I have opened a different PR  which fixes this issue
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2365

Issue #2335, opened 2025-07-22
[Win32] Non-disposed resource tracker not working properly


HeikoKlare:
**Describe the bug**
On Windows, the non-disposed resource tracker is not working properly. The resource tracker employs the GC to detect that resource objects are cleaned up by the GC without being disposed before. On Windows, however, every resource it registered at its device in order to clean up handles that have been created for multiple zoom levels:
https://github.com/eclipse-platform/eclipse.platform.swt/blob/34a4d766cc792f93b6ef8b8e3395b9c28b90af14/bundles/org.eclipse.swt/Eclipse%20SWT/win32/org/eclipse/swt/graphics/Device.java#L966-L968

This registration at the device puts every resource into a set inside the device, which results in the GC never cleaning such a resource up, even if no one references it.

**To Reproduce**
You can run any code with a resource leak and the resource tracker enabled, i.e., with `-Dorg.eclipse.swt.graphics.Resource.reportNonDisposed=true`. As soon as the GC runs, one would expect an error to be printed. For example, take the following snippet, which explicitly calls the GC:

```java
public static void main (String [] args) {
	Display display = new Display ();
	new Region(display);
	System.gc();
	display.dispose ();
}
```

With the current SWT master state, nothing will be printed. If you comment out the implementation of this method inside Device, an error shows up:
https://github.com/eclipse-platform/eclipse.platform.swt/blob/34a4d766cc792f93b6ef8b8e3395b9c28b90af14/bundles/org.eclipse.swt/Eclipse%20SWT/win32/org/eclipse/swt/graphics/Device.java#L966-L968

```
java.lang.Error: SWT Resource was not properly disposed
	at org.eclipse.swt.graphics.Resource.initNonDisposeTracking(Resource.java:191)
	at org.eclipse.swt.graphics.Resource.<init>(Resource.java:124)
	at org.eclipse.swt.graphics.Region.<init>(Region.java:82)
	at org.eclipse.swt.snippets.Snippet1.main(Snippet1.java:29)
```

**Expected behavior**
The resource tracker shall work as expected on Windows, which requires resources to not be referenced in any central data structure.
In particular, the above snippet with the tracker enabled should log an error.

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 

2. Additional OS info (e.g. OS version, Linux Desktop, etc)

_Irrelevant_

[...]

PR #2338, opened 2025-07-23
Fix Resource leak tracking issue on Windows

ShahzaibIbrahim:
On Windows SWT, the Device implementation keeps a strong reference to Resource objects via this method:

```java
void registerResourceWithZoomSupport(Resource resource) {
	resourcesWithZoomSupport.add(resource);
}
```
That means:
Every Resource with zoom support is added to the resourcesWithZoomSupport set in the Device. This is a `Set<Resource>`, which holds strong references. **Because of this**, the GC will never collect these resources, even if the application code drops all references, the `Device` is still referencing them.

To tackle that, we can use the WeakReference ([more about it here](https://sonique6784.medium.com/is-weakreference-%EF%B8%8F-really-a-solution-to-memory-leaks-f1ff0e87cc47)). That will allows GC to collect the resource. 

The only downside I see could be, that every time a `Resource ` object becomes garbage collected, its corresponding `WeakReference` object still exists in the set, which needs to be cleared. For that, my idea is to clean up the dead references whenever win32_destroyUnusedHandles is called more specifically whenever `WM_DISPLAYCHANGE` or `WM_DPICHANGED` event happens. 

Fixes: https://github.com/eclipse-platform/eclipse.platform.swt/issues/2335


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Device.java
```diff
@@ -982,8 +983,41 @@ public static void win32_destroyUnusedHandles(Display display) {
...
+private static class ResourceReference extends WeakReference<Resource> {
+
+	public ResourceReference(Resource referent) {
+		super(referent);
+	}
+
+	@Override
+	public boolean equals(Object obj) {
+		if (this == obj) return true;
+		if (!(obj instanceof ResourceReference)) return false;
```
HeikoKlare: This way you can avoid the cast two lines later:
```suggestion
		if (!(obj instanceof ResourceReference passedResource)) return false;
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Device.java
```diff
@@ -982,8 +983,41 @@ public static void win32_destroyUnusedHandles(Display display) {
...
+
+		Resource thisResource = this.get();
+		Resource passedResource = ((ResourceReference) obj).get();
+
+		return thisResource != null && passedResource != null && thisResource.equals(passedResource);
```
HeikoKlare:
```suggestion
		return Objects.equals(thisResource, passedResource);;
```

HeikoKlare (commented):
The implementation looks fine and works as expected. I have only minor comments regarding code style.

Issue #2346, opened 2025-07-24
[WIn32] CTabFolder head is missing


HeikoKlare:
**Describe the bug**
The header of a CTabFolder may not be rendered.
This is a regression from https://github.com/eclipse-platform/eclipse.platform.swt/pull/2336

**To Reproduce**
Start the `CustomControlExample` and go to the `CTabFolder` tab or open the global search (CTRL+H) in an Eclipse application (see screenshots below).

**Expected behavior**
CTabHolder header should be rendered properly.

**Screenshots**
<img width="729" height="472" alt="Image" src="https://github.com/user-attachments/assets/dca192aa-ab0b-4798-bfe9-c6b45c6bb9f4" />
<img width="851" height="470" alt="Image" src="https://github.com/user-attachments/assets/f96b5116-8d0d-48a8-b0b2-831593c5b3fa" />

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 

2. Additional OS info (e.g. OS version, Linux Desktop, etc)
WIndows-only

3. JRE/JDK version
_ireelevant_
 
**Version since**
Since I20250723-1800 (as regression was introduced in day before, see above).

PR #2348, opened 2025-07-24
[Win32] Represent GC#getClipping(Region) operation in Region #2346

HeikoKlare:
The GC#getClipping(Region) call sets the passed region to the current clipping region of the GC. Currently, this is not represented as an operation inside the passed Region, such that when calculating a handle for a different zoom of that region or when copying that region the according application of GC#getClipping(Region) to that region is missing.

This change adds the missing operation to the Region. Since the clipping needs to be calculated at the of executing the GC#getClipping(Region) method and not at the time of applying the region operation when retrieving a handle, the GC#getClipping(Region) stores an operation inside GC that maintains handles for the according clipping region at the time of executing the GC#getClipping(Region) method for every zoom at which the GC is requested. This ensures that the proper clipping state is available to the Region.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/pull/2347


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Region.java
```diff
@@ -468,6 +468,16 @@ public boolean isEmpty () {
 	});
 }
 
+/**
+ * Specific method for {@link GC#getClipping(Region)} because the current GC
+ * clipping settings at that specific point in time of executing the getClipping
+ * method need to be stored.
+ */
+void set(Function<Integer, Long> handleForZoomSupplier) {
+	final Operation operation = new OperationWithRegionHandle(Operation::set, handleForZoomSupplier);
```
amartya4256, line 492: Why not have OperationWithRegionHandle::set as an argument and not define it in abstract class Operation? Should it not work?
Edit: `(op, zoom, handle) -> ((OperationWithRegionHandle)op).set(zoom, handle)`

Then we can get rid of the UnsupportedOperationException and so many unimplemented methods.

amartya4256:
I have tested the behaviour and it seems to fix the regression by storing getClipping operation in gc and calling the getClippingRegion on opertaion apply, and also registering as a set operation inside the region which references the gc getClipping operation object's zoomToHandleStorage to get the manipulated handle as per the operation. However, I have added a few comments here and there.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -3790,6 +3790,44 @@ public void getClipping (Region region) {
 	checkNonDisposed();
 	if (region == null) SWT.error (SWT.ERROR_NULL_ARGUMENT);
 	if (region.isDisposed()) SWT.error (SWT.ERROR_INVALID_ARGUMENT);
+	storeAndApplyOperationForExistingHandle(new GetClippingOperation(region));
+}
+
+private class GetClippingOperation extends Operation {
+	private final Map<Integer, Long> zoomToHandleStorage = new HashMap<>();
```
amartya4256: A bit of a confusing name
```suggestion
	private final Map<Integer, Long> zoomToRegionHandle = new HashMap<>();
```

HeikoKlare: Indeed. Updated with [01413be](https://github.com/eclipse-platform/eclipse.platform.swt/pull/2348/commits/01413be107dbe816b83ecc0a0ff170c0610c9eb5)

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Region.java
amartya4256: Since, we already have UnsuppportedOperationException for some other Operations, let's stick with the current impl. in the the PR for consistency.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -3790,6 +3790,44 @@ public void getClipping (Region region) {
...
+}
+
+private class GetClippingOperation extends Operation {
+	private final Map<Integer, Long> zoomToRegionHandle = new HashMap<>();
+
+	public GetClippingOperation(Region region) {
+		region.set(zoom -> {
+			if (!zoomToRegionHandle.containsKey(zoom)) {
+				System.err.println("No clipping handle for zoom " + zoom + " has been created on this GC");
+				return zoomToRegionHandle.values().iterator().next();
```
arunjose696, line 3803: I tried going through it just  have a question, could this lead to visual inconsistencies if the returned handle doesn't match the requested zoom? Since iterator().next() returns an arbitrary value for the region when GetClippingOperation has not been run for this zoom

HeikoKlare: Yes, it can. In such a situation we cannot create a proper handle (i.e., one at the correct zoom), which is why the error message is printed. This situation should actually not occur at all, as it would not make any sense to use the region in any other zoom context than in the context of the GC from which the clipping is applied.
However, I just found that this leads to issues when some operation is executed on a region that has no handle yet (e.g., for only retrieving its bounds). Then a temporary handle for an arbitrary zoom (more precisely the device zoom) is used. If that device zoom does not fit to the GC context, we run into the error case. I currently try to store a zoom hint inside region so that a temporary handle is created at a proper zoom if such a hint is given. I will update the PR soon.

arunjose696: > This situation should actually not occur at all, as it would not make any sense to use the region in any other zoom context than in the context of the GC from which the clipping is applied. However, I just found that this leads to issues when some operation is executed on a region that has no handle yet (e.g., for only retrieving its bounds). 

I get this error in my console when I move the search window across monitors.

HeikoKlare:
I have extended the PR so that a region stores a zoom hint for temporary handles, in order to ensure that when a temporary handle is necessary, its zoom fits to the GC context such that a potential clipping region can be retrieved.

This is the according change:
https://github.com/eclipse-platform/eclipse.platform.swt/compare/01413be107dbe816b83ecc0a0ff170c0610c9eb5..53ae3d3f9bceb1a8bf7dc0216ad05ac13ea2ac9b

The behavior can be tested as follows:
- Have two monitors with different zooms
- Start an Eclipse SDK and open two workbench windows, place one on each of the monitors
- Open something that uses the CTabFolderRenderer on both of them, for example the MANIFEST file of a project (the multi-page editor for Manifest, build.properties etc. uses that renderer)
- Move around the windows on both monitors

Before the recent change you will have seen `No clipping handle zoom ...` errors being printed, which is not the case with the recent extensions.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
HeikoKlare: > I get this error in my console when I move the search window across monitors.

Yes, exactly. You may test again with the recent change, which should resolve that issue.

arunjose696: Tested the latest changes the issue is resolved

arunjose696:
> Before the recent change you will have seen `No clipping handle zoom ...` errors being printed, which is not the case with the recent extensions.

I’ve tested the changes , and they seem to have resolved the regression and the issue where No clipping handle for zoom ... errors were previously being printed.
Additionally, another issue related to inconsistent behavior when maximizing and minimizing widgets using CTabFolderRenderer also appears to be fixed with this new commit.

I’ve briefly reviewed the code, but I’m not confident approving it as I’m not familiar with the Region implementation or the need for temporary handles.

HeikoKlare:
Thank you for testing the PR!

> Additionally, another issue related to inconsistent behavior when maximizing and minimizing widgets using CTabFolderRenderer also appears to be fixed with this new commit.

Can give explain which kind of issue that was? It that inconsistent behavior present in current master state or was it only in the "intermediate" state of this PR?

arunjose696:
> Thank you for testing the PR!
> 
> > Additionally, another issue related to inconsistent behavior when maximizing and minimizing widgets using CTabFolderRenderer also appears to be fixed with this new commit.
> 
> Can give explain which kind of issue that was? It that inconsistent behavior present in current master state or was it only in the "intermediate" state of this PR?

It was in intermediate state of the PR before the latest push.

HeikoKlare:
Alright, thank you!

Then I'll merge this to have the `CTabFolder` issue fixed in M2. We still have some time after M2 to check if there is anything left to do that we might have missed so far.

PR #2349, opened 2025-07-24
Effectively remove replaced MonitorAwareRectangle/MonitorAwarePoint

HeikoKlare:
The MonitorAwareRectangle/MonitorAwarePoint classes have been replaced by Rectangle.WithMonitor and Point.WithMonitor. They have been marked as deprecated but not removed because of issues with API tooling, however they were marked as non-API so they may be removed without any deprecation period.

This change effectively removes the classes, only keeping package-internal stubs of them as otherwise API tooling currently fails.

This is not the cleanest solution as it is driven by tooling limitations, but I don't think we should unnecessarily keep those classes and I am currently stuck in debugging the tooling issue:
- https://github.com/eclipse-tycho/tycho/issues/5148

Since API tooling treats this change as a breaking API change despite the `@noreference` Javadoc tag, according API filters are required. This also means that any dependent bundle that reexports SWT and uses API tooling (such as `org.eclipse.jface` and `org.eclipse.ui`) need to be extended with API filters as well if we merge this:
image

### DIFF
--- a/binaries/org.eclipse.swt.cocoa.macosx.aarch64/.settings/.api_filters
+++ b/binaries/org.eclipse.swt.cocoa.macosx.aarch64/.settings/.api_filters
@@ -8,4 +8,20 @@
             </message_arguments>
         </filter>
     </resource>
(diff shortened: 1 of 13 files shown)


fedejeanne (approved):
Looks good now.

We can remove the dummy classes later (no rush, they are not public)

PR #2354, opened 2025-07-27
[GTK] Remove duplicate device-zoom initialization in Image

ptziegler:
The device zoom is hard-coded to 100% when creating blank images. Setting it to the current device zoom in the constructor is both wrong and unnecessary, given that the value will be immediately overwritten.


ptziegler:
See https://github.com/eclipse-platform/eclipse.platform.swt/blob/ebcb6b9482fb9fb2c8fd6f03326783a5a4025487/bundles/org.eclipse.swt/Eclipse%20SWT/gtk/org/eclipse/swt/graphics/Image.java#L1276-L1278

akurtakov:
Good catch. Merging.

PR #2358, opened 2025-07-30
Fix stack overflow in DPIUtil.autoScaleImageData

### LINKED ISSUE #2353: StackOverflowError in Image when auto-scaling differs from device zoom
**To Reproduce**

Run Snippet367 with `-Dswt.autoScale=150`.

Stacktrace:
```
Exception in thread "main" java.lang.StackOverflowError
	at org.eclipse.swt.internal.DPIUtil.getElementAtZoom(DPIUtil.java:289)
	at org.eclipse.swt.internal.DPIUtil.validateAndGetImageDataAtZoom(DPIUtil.java:260)
	at org.eclipse.swt.graphics.Image.initFromImageDataProvider(Image.java:797)
	at org.eclipse.swt.graphics.Image.<init>(Image.java:675)
	at org.eclipse.swt.internal.DPIUtil.autoScaleImageData(DPIUtil.java:156)
	at org.eclipse.swt.internal.DPIUtil.scaleImageData(DPIUtil.java:139)
	at org.eclipse.swt.graphics.Image.initFromImageDataProvider(Image.java:798)
	at org.eclipse.swt.graphics.Image.<init>(Image.java:675)
	at org.eclipse.swt.internal.DPIUtil.autoScaleImageData(DPIUtil.java:156)
	at org.eclipse.swt.internal.DPIUtil.scaleImageData(DPIUtil.java:139)
	at org.eclipse.swt.graphics.Image.initFromImageDataProvider(Image.java:798)
	at org.eclipse.swt.graphics.Image.<init>(Image.java:675)
...
```

**Environment:**
- - [ ] All OS
- - [x] Windows
- - [x] Linux
- - [ ] macOS 

**Version since**

This is a regression introduced with 216f1adafc7380edb734654cb4b25c7ecbd1a082.

arunjose696:
Fixes #2353 

In PR #2249, a strict check was introduced  that `ImageData` should be linearly scaled for 
zoom levels. And imageDataProviders in SWT repo not following this strict checks were modified (eg to return only imageData at 100%).

However, this change broke `DPIUtil.autoScaleImageData`. That function creates an Image using the system zoom level, draws it on a GC at scaled dimensions, and then requests ImageData at 100% zoom from the image. Since the `ImageDataProvider` was now restricted to 100% zoom only, requesting imageData at other zoom levels during creation of  image at system zoom level triggered recursive calls,  causing a stack overflow error.

The current commit reverts the behavior of the `ImageDataProvider` used in` DPIUtil.autoScaleImageData` to return the same `ImageData` at all zoom levels.
We also Disable the strict zoom check in this specific case, because the scaling is done by `GC.drawImage()` and the `ImageData` used to create the image at system zoom level is expected to be same as the passed `imageData`.

**Steps to reproduce**

Execute Issue0445_HiDPISmoothScaling when your device zoom is something other than 200%. This would give stackoverflow error without this change

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/ControlWin32Tests.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/ControlWin32Tests.java
@@ -14,12 +14,15 @@
 package org.eclipse.swt.widgets;
 
 import static org.junit.Assert.*;
+import static org.junit.jupiter.api.Assertions.assertEquals;
 
 import org.eclipse.swt.*;
 import org.eclipse.swt.graphics.*;
 import org.eclipse.swt.internal.*;
 import org.junit.jupiter.api.*;
 import org.junit.jupiter.api.extension.*;
+import org.junit.jupiter.params.*;
+import org.junit.jupiter.params.provider.*;
 
 /**
  * Automated Tests for class org.eclipse.swt.widgets.Control for Windows
@@ -105,6 +108,28 @@ class ControlWin32Tests {
 		button.setBounds(0, 47, 200, 47);
 		assertEquals("Control::setBounds(int, int, int, int) doesn't scale up correctly",
 				new Rectangle(0, 82, 350, 83), button.getBoundsInPixels());
+	}
+
+	@ParameterizedTest
+	@CsvSource({ "0.5, 100, true", "1.0, 200, true", "2.0, 200, true", "2.0, quarter, true", "0.5, 100, false",
+			"1.0, 200, false", "2.0, 200, false", "2.0, quarter, false", })
+	public void testAutoScaleImageData(float scaleFactor, String autoScale, boolean monitorSpecificScaling) {
+		DPIUtil.setMonitorSpecificScaling(monitorSpecificScaling);
+		DPIUtil.runWithAutoScaleValue(autoScale, () -> {
+			Display display = new Display();
+			try {
+				ImageData imageData = new ImageData(100, 100, 1, new PaletteData(new RGB(0, 0, 0)));
+				int width = imageData.width;
+				int height = imageData.height;
+				int scaledWidth = Math.round(width * scaleFactor);
+				int scaledHeight = Math.round(height * scaleFactor);
+				ImageData scaledImageData = DPIUtil.autoScaleImageData(display, imageData, scaleFactor);
+				assertEquals(scaledWidth, scaledImageData.width);
+				assertEquals(scaledHeight, scaledImageData.height);
+			} finally {
+				display.dispose();
+			}
+		});
 	}
 
 	record FontComparison(int originalFontHeight, int currentFontHeight) {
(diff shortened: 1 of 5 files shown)


fedejeanne (commented):
I'm not able to judge the technical side of this PR but isn't it correct to say that it fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2353 ? If it does then the commit text needs to be adapted.

arunjose696:
> I'm not able to judge the technical side of this PR but isn't it correct to say that it fixes #2353 ? If it does then the commit text needs to be adapted.

Yes this issue fixes #2353 . Now I have mentioned breifly mentioned this issue in the commit text.

fedejeanne:
Thank you for adding the information. I'm not able to judge this PR at a technical level (maybe @akoch-yatta / @HeikoKlare can) but I can tell that at least this PR does get rid of the stack overflow, as it promises. ✔️

arunjose696:
The failing test seems to be unrelated to the change. I have raised an issue about the failures https://github.com/eclipse-platform/eclipse.platform.swt/issues/2363

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
```diff
@@ -394,6 +392,10 @@ public static void setMonitorSpecificScaling(boolean activate) {
 	System.setProperty(SWT_AUTOSCALE_UPDATE_ON_RUNTIME, Boolean.toString(activate));
 }
 
+public static void setAutoScaleValue(String autoScaleValue) {
```
HeikoKlare: We should add a JavaDoc that states that this method is only supposed to be called in tests and may lead to unintended effects if called in production.

bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/ControlWin32Tests.java
```diff
@@ -107,6 +110,27 @@ public void testCorrectScaleUpUsingDifferentSetBoundsMethod() {
 				new Rectangle(0, 82, 350, 83), button.getBoundsInPixels());
 	}
 
+	@ParameterizedTest
+	@CsvSource({ "0.5, 100, true", "1.0, 200, true", "2.0, 200, true", "2.0, quarter, true", "0.5, 100, false",
+			"1.0, 200, false", "2.0, 200, false", "2.0, quarter, false", })
+	public void autoScaleImageData0(float scaleFactor, String autoScale, boolean monitorSpecificScaling) {
+		DPIUtil.setMonitorSpecificScaling(monitorSpecificScaling);
+		DPIUtil.setAutoScaleValue(autoScale);
```
HeikoKlare: Since the value is only initialized on class loading (i.e., at JVM startup), changing this value in a test without resetting it will leave the JVM in an unexpected state for the subsequent tests. To avoid that, the original value should be stored and reapplied after executing the test. This has to be done in the finally block, as otherwise a test failure would still leave behind an unexpected system state.

HeikoKlare (commented):
The change is sound, thank you!

I only have two minor comments.

HeikoKlare (approved):
I just saw that we already have `DPIUtil#runWithAutoScaleValue()` which can be used to run some code with a specific autoScale value making the `setAutoScaleValue` method for test purposes obsolete. I have adapted the test implementation with a usage of that method and removed the obsolete new method from `DPIUtil`, which renders my previous comments obsolete.

@arunjose696 please check if the change is fine for you.

arunjose696:
> @arunjose696 please check if the change is fine for you.

The new changes look good to me, Additionally I have added image disposals for the new images created which I missed.

HeikoKlare:
> The new changes look good to me, Additionally I have added image disposals for the new images created which I missed.

Great, thanks!

Issue #2355, opened 2025-07-28
[Win32] Retrieving handle for cursor leads to non-disposed resource error


HeikoKlare:
**Describe the bug**
When retrieving a handle for a cursor that does not yet exist (because it is required for a different zoom), a non-disposed resource error is printed in case the non-disposed resource tracker is enabled.

This is caused by the `Cursor#win32_getHandle()` method creating a cursor instance just to extract the handle out of it. So it's actually intended that the cursor is not disposed, as that would also dispose the handle, but the implementation itself is strange as it should not be necessary to create such a dummy cursor at all:
https://github.com/eclipse-platform/eclipse.platform.swt/blob/bcc58ad49cc99affab042843e80306af4a65fd33/bundles/org.eclipse.swt/Eclipse%20SWT/win32/org/eclipse/swt/graphics/Cursor.java#L432-L433

**To Reproduce**
The error can be produced with the following simple snippet executed on a Windows machine with a primary monitor zoom <= 150%:

```java
	Cursor cursor = new Cursor(Display.getDefault(), new ImageData(1, 1, 24, new PaletteData(0xFF, 0xFF00, 0xFF0000)), 0, 0);
	Cursor.win32_getHandle(cursor, 200);
	System.gc();
	Thread.sleep(1000);
```

Note that the `System.gc()` call is just there to ensure that the dummy cursor is cleaned up by GC, which yields the error, and the `Thread.sleep()` ensures that the JVM has sufficient time to print the error (as otherwise it may stop before the printing is finished).

**Expected behavior**
No non-disposed resource should be created and tracked when retrieving cursor handles. Actually I would not expect a dummy cursor to be created just for the sake of getting a handle at all.

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 

2. Additional OS info (e.g. OS version, Linux Desktop, etc)
Any Windows version

3. JRE/JDK version
_irrelevant_ 

**Version since**
Eclipse 2024-12, precisely https://github.com/eclipse-platform/eclipse.platform.swt/pull/1603

**Workaround (or) Additional context**
The error can just be ignored, there is no negative effect. It's just an irrelevant error that indicates a design flaw.

ShahzaibIbrahim:
### Problem

[...]

PR #2362, opened 2025-07-31
[Win32] Lazy load Cursor handles

ShahzaibIbrahim:
### Problem

Currently, we have a logic to always initialize the first **handle** in a `handle` field. Instead of using it, we could switch to lazy loading via using `zoomLevelToHandle` map. The handle will only created when `win32_getHandle` is called. 

---

### Proposed Solution

To fix this, the `Cursor` class should be refactored to manage zoom-level-specific handles more robustly and safely. The following changes are needed:

1. **Remove `handle` Field**
   Eliminate the dedicated `handle` field from the class. All handle accesses should go through the `zoomLevelToHandle` map, which now acts as the single source of truth for all zoom levels.

2. **Update All Related Methods**
   Refactor all methods that previously relied on the `handle` field to instead use `zoomLevelToHandle`. This includes:

   * `hashCode()`
   * `equals(Object obj)`
   * `isDisposed()`
   * `destroy()`

---

### Benefits

* Enables safe, lazy creation of per-zoom handles when needed.
* Simplifies and centralizes handle management logic for better maintainability.

### Example

```java
import org.eclipse.swt.*;

public class CursorZoomTest {

        // 1. Constructor: Cursor(Device, int style)
        Cursor cursorFromStyle = new Cursor(display, SWT.CURSOR_HAND);

        // 2. Constructor: Cursor(Device, ImageData source, ImageData mask, int hotspotX, int hotspotY)
        ImageData source = new ImageData(16, 16, 1, new PaletteData(new RGB[]{new RGB(0, 0, 0)}));
        ImageData mask = new ImageData(16, 16, 1, new PaletteData(new RGB[]{new RGB(0, 0, 0)}));
        Cursor cursorFromImageAndMask = new Cursor(display, source, mask, 0, 0);

        // 3. Constructor: Cursor(Device, ImageData source, int hotspotX, int hotspotY)
        Cursor cursorFromImageOnly = new Cursor(display, source, 0, 0);

        // 4. Constructor: Cursor(Device, ImageDataProvider imageDataProvider, int hotspotX, int hotspotY)
        ImageDataProvider provider = zoom -> source;
        Cursor cursorFromProvider = new Cursor(display, provider, 0, 0);

        Cursor[] cursors = {
            cursorFromStyle,
            cursorFromImageAndMask,
            cursorFromImageOnly,
            cursorFromProvider
        };

        int[] zoomLevels = {100, 125, 150, 175, 200, 250};

        for (int i = 0; i < cursors.length; i++) {
            System.out.println("Cursor #" + (i + 1));
            for (int zoom : zoomLevels) {
                long handle = Cursor.win32_getHandle(cursors[i], zoom);
                System.out.println("  Zoom " + zoom + ": handle = " + handle);
            }
        }

        // Dispose resources
        for (Cursor cursor : cursors) {
            cursor.dispose();
        }

        shell.dispose();
        display.dispose();
    }
}
```

**Output**
```
Cursor #1
  Zoom 100: handle = 65567
  Zoom 125: handle = 65567
  Zoom 150: handle = 65567
  Zoom 175: handle = 65567
  Zoom 200: handle = 65567
  Zoom 250: handle = 65567
Cursor #2
  Zoom 100: handle = 75434631
  Zoom 125: handle = 53612289
  Zoom 150: handle = 43651639
  Zoom 175: handle = 61083667
  Zoom 200: handle = 30212277
  Zoom 250: handle = 141363963
Cursor #3
  Zoom 100: handle = 33559507
  Zoom 125: handle = 263197751
  Zoom 150: handle = 84413253
  Zoom 175: handle = 21237883
  Zoom 200: handle = 56361347
  Zoom 250: handle = 46928295
Cursor #4
  Zoom 100: handle = 20124285
  Zoom 125: handle = 63705063
  Zoom 150: handle = 46729749
  Zoom 175: handle = 68358127
  Zoom 200: handle = 79891305
  Zoom 250: handle = 14290407

```

**Note that for the Cursor 1, we always get the same handle from OS but it's scaled correctly for all zoom levels.**

Fixes: https://github.com/eclipse-platform/eclipse.platform.swt/issues/2355

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
@@ -50,18 +50,6 @@ import org.eclipse.swt.internal.win32.*;
 public final class Cursor extends Resource {
 
 	/**
-	 * the handle to the OS cursor resource
-	 * (Warning: This field is platform dependent)
-	 * <p>
-	 * <b>IMPORTANT:</b> This field is <em>not</em> part of the SWT
-	 * public API. It is marked public only so that it can be shared
-	 * within the packages provided by SWT. It is not available on all
-	 * platforms and should never be accessed from application code.
-	 * </p>
-	 *
-	 */
-	private long handle;
-	/**
 	 * Attribute to cache current native zoom level
 	 */
 	private static final int DEFAULT_ZOOM = 100;
@@ -69,6 +57,8 @@ public final class Cursor extends Resource {
 	private HashMap<Integer, CursorHandle> zoomLevelToHandle = new HashMap<>();
 
 	private final CursorHandleProvider cursorHandleProvider;
+
+	private boolean isDestroyed;
 
 /**
  * Constructs a new cursor given a device and a style
@@ -119,7 +109,6 @@ public final class Cursor extends Resource {
 public Cursor(Device device, int style) {
 	super(device);
 	this.cursorHandleProvider = new StyleCursorHandleProvider(style);
-	this.handle = this.cursorHandleProvider.createHandle(this.device, DEFAULT_ZOOM).getHandle();
 	init();
 	this.device.registerResourceWithZoomSupport(this);
 }
@@ -160,27 +149,13 @@ public Cursor(Device device, int style) {
 public Cursor(Device device, ImageData source, ImageData mask, int hotspotX, int hotspotY) {
 	super(device);
 	this.cursorHandleProvider = new ImageDataWithMaskCursorHandleProvider(source, mask, hotspotX, hotspotY);
-	this.handle = this.cursorHandleProvider.createHandle(this.device, DEFAULT_ZOOM).getHandle();
 	init();
 	this.device.registerResourceWithZoomSupport(this);
 }
 
 private static CursorHandle setupCursorFromImageData(ImageData source, ImageData mask, int hotspotX, int hotspotY) {
-	if (source == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
 	if (mask == null) {
-		if (source.getTransparencyType() != SWT.TRANSPARENCY_MASK) {
-			SWT.error(SWT.ERROR_NULL_ARGUMENT);
-		}
 		mask = source.getTransparencyMask();
-	}
-	/* Check the bounds. Mask must be the same size as source */
-	if (mask.width != source.width || mask.height != source.height) {
-		SWT.error(SWT.ERROR_INVALID_ARGUMENT);
-	}
-	/* Check the hotspots */
-	if (hotspotX >= source.width || hotspotX < 0 ||
-		hotspotY >= source.height || hotspotY < 0) {
-		SWT.error(SWT.ERROR_INVALID_ARGUMENT);
 	}
 	/* Convert depth to 1 */
 	mask = ImageData.convertMask(mask);
@@ -229,18 +204,12 @@ private static CursorHandle setupCursorFromImageData(ImageData source, ImageData
 public Cursor(Device device, ImageData source, int hotspotX, int hotspotY) {
 	super(device);
 	this.cursorHandleProvider = new ImageDataCursorHandleProvider(source, hotspotX, hotspotY);
-	this.handle = this.cursorHandleProvider.createHandle(this.device, DEFAULT_ZOOM).getHandle();
 	init();
 	this.device.registerResourceWithZoomSupport(this);
 }
 
 private static CursorHandle setupCursorFromImageData(Device device, ImageData source, int hotspotX, int hotspotY) {
 	if (source == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
-	/* Check the hotspots */
-	if (hotspotX >= source.width || hotspotX < 0 ||
-		hotspotY >= source.height || hotspotY < 0) {
-		SWT.error(SWT.ERROR_INVALID_ARGUMENT);
-	}
 	long hBitmap = 0;
 	long hMask = 0;
 	if (source.maskData == null && source.transparentPixel == -1 && (source.alpha != -1 || source.alphaData != null)) {
@@ -341,7 +310,6 @@ public Cursor(Device device, ImageDataProvider imageDataProvider, int hotspotX, 
 	super(device);
 	if (imageDataProvider == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
 	this.cursorHandleProvider = new ImageDataProviderCursorHandleProvider(imageDataProvider, hotspotX, hotspotY);
-	this.handle = this.cursorHandleProvider.createHandle(this.device, DEFAULT_ZOOM).getHandle();
 	init();
 	this.device.registerResourceWithZoomSupport(this);
 }
@@ -363,7 +331,7 @@ public Cursor(Device device, ImageDataProvider imageDataProvider, int hotspotX, 
  */
 public static Long win32_getHandle (Cursor cursor, int zoom) {
 	if (cursor.isDisposed()) {
-		return cursor.handle;
+		return 0L;
 	}
 	if (cursor.zoomLevelToHandle.get(zoom) != null) {
 		return cursor.zoomLevelToHandle.get(zoom).getHandle();
@@ -376,9 +344,6 @@ public static Long win32_getHandle (Cursor cursor, int zoom) {
 }
 
 private void setHandleForZoomLevel(CursorHandle handle, Integer zoom) {
-	if (this.handle == 0) {
-		this.handle = handle.getHandle();	// Set handle for default zoom level
-	}
 	if (zoom != null && !zoomLevelToHandle.containsKey(zoom)) {
 		zoomLevelToHandle.put(zoom, handle);
 	}
@@ -386,28 +351,12 @@ private void setHandleForZoomLevel(CursorHandle handle, Integer zoom) {
 
 @Override
 void destroy () {
-	/*
-	* It is an error in Windows to destroy the current
-	* cursor.  Check that the cursor that is about to
-	* be destroyed is the current cursor.  If so, set
-	* the current cursor to be IDC_ARROW.  Note that
-	* Windows shares predefined cursors so the call to
-	* LoadCursor() does not leak.
-	*/
-	// TEMPORARY CODE
-//	if (OS.GetCursor() == handle) {
-//		OS.SetCursor(OS.LoadCursor(0, OS.IDC_ARROW));
-//	}
 	device.deregisterResourceWithZoomSupport(this);
-	destroyHandle();
-}
-
-private void destroyHandle () {
 	for (CursorHandle handle : zoomLevelToHandle.values()) {
 		handle.destroy();
 	}
 	zoomLevelToHandle.clear();
-	handle = 0;
+	this.isDestroyed = true;
 }
 
 /**
@@ -425,7 +374,7 @@ public boolean equals (Object object) {
 	if (object == this) return true;
 	if (!(object instanceof Cursor)) return false;
 	Cursor cursor = (Cursor) object;
-	return device == cursor.device && handle == cursor.handle;
+	return device == cursor.device && win32_getHandle(this, DEFAULT_ZOOM) == win32_getHandle(cursor, DEFAULT_ZOOM);
 }
 
 /**
@@ -440,7 +389,7 @@ public boolean equals (Object object) {
  */
 @Override
 public int hashCode () {
-	return (int)handle;
+	return win32_getHandle(this, DEFAULT_ZOOM).intValue();
 }
 
 /**
@@ -455,7 +404,7 @@ public int hashCode () {
  */
 @Override
 public boolean isDisposed() {
-	return handle == 0;
+	return isDestroyed;
 }
 
 /**
@@ -467,7 +416,7 @@ public boolean isDisposed() {
 @Override
 public String toString () {
 	if (isDisposed()) return "Cursor {*DISPOSED*}";
-	return "Cursor {" + handle + "}";
+	return "Cursor {" + zoomLevelToHandle + "}";
 }
 
 @Override
@@ -531,19 +480,23 @@ private static interface CursorHandleProvider {
 }
 
 private static class StyleCursorHandleProvider implements CursorHandleProvider {
-	private final int style;
+	private final long lpCursorName;
 
 	public StyleCursorHandleProvider(int style) {
-		this.style = style;
+		this.lpCursorName = getOSCursorIdFromStyle(style);
 	}
 
 	@Override
 	public CursorHandle createHandle(Device device, int zoom) {
 		// zoom ignored, LoadCursor handles scaling internally
-		return setupCursorFromStyle(this.style);
+		long handle = OS.LoadCursor(0, lpCursorName);
+		if (handle == 0) {
+			SWT.error(SWT.ERROR_NO_HANDLES);
+		}
+		return new CustomCursorHandle(handle);
 	}
 
-	private static final CursorHandle setupCursorFromStyle(int style) {
+	private static final long getOSCursorIdFromStyle(int style) {
 		long lpCursorName = 0;
 		switch (style) {
 		case SWT.CURSOR_HAND:
@@ -615,11 +568,7 @@ private static class StyleCursorHandleProvider implements CursorHandleProvider {
 		default:
 			SWT.error(SWT.ERROR_INVALID_ARGUMENT);
 		}
-		long handle = OS.LoadCursor(0, lpCursorName);
-		if (handle == 0) {
-			SWT.error(SWT.ERROR_NO_HANDLES);
-		}
-		return new CustomCursorHandle(handle);
+		return lpCursorName;
 	}
 }
 
@@ -639,6 +588,14 @@ private static abstract class HotspotAwareCursorHandleProvider implements Cursor
 	protected final int getHotpotYInPixels(int zoom) {
 		return Win32DPIUtils.pointToPixel(hotspotY, zoom);
 	}
+
+	protected static final void validateHotspotInsideImage(ImageData source, int hotspotX, int hotspotY) {
+		/* Check the hotspots */
+		if (hotspotX >= source.width || hotspotX < 0 ||
+			hotspotY >= source.height || hotspotY < 0) {
+			SWT.error(SWT.ERROR_INVALID_ARGUMENT);
+		}
+	}
 }
 
 private static class ImageDataProviderCursorHandleProvider extends HotspotAwareCursorHandleProvider {
@@ -646,6 +603,9 @@ private static class ImageDataProviderCursorHandleProvider extends HotspotAwareC
 
 	public ImageDataProviderCursorHandleProvider(ImageDataProvider provider, int hotspotX, int hotspotY) {
 		super(hotspotX, hotspotY);
+		ImageData source = provider.getImageData(DEFAULT_ZOOM);
+		if (source == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
+		validateHotspotInsideImage(source, hotspotX, hotspotY);
 		this.provider = provider;
 	}
 
@@ -668,6 +628,8 @@ private static class ImageDataCursorHandleProvider extends HotspotAwareCursorHan
 
 	public ImageDataCursorHandleProvider(ImageData source, int hotspotX, int hotspotY) {
 		super(hotspotX, hotspotY);
+		if (source == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
+		validateHotspotInsideImage(source, hotspotX, hotspotY);
 		this.source = source;
 	}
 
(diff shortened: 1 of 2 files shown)


arunjose696:
CI fails as Added tests uses windows only API, they should be probably moved to `org.eclipse.swt.tests.win32`

HeikoKlare (commented):
Can we please split to up into multiple commits or even PRs to make it easier to review and (hopefully not necessary) to find regressions? I see three changes combined here that we could split up:
- Fixing the hotspot coordinate calculation
- Refactoring the constructors (moving initialization logic to other methods)
- Managing multiple handles for different zooms

ShahzaibIbrahim:
> Can we please split to up into multiple commits or even PRs to make it easier to review and (hopefully not necessary) to find regressions? I see three changes combined here that we could split up:
> 
> * Fixing the hotspot coordinate calculation
> * Refactoring the constructors (moving initialization logic to other methods)
> * Managing multiple handles for different zooms

As you proposed, I have 3 PRs for the following

- Coordinate Calculation: https://github.com/eclipse-platform/eclipse.platform.swt/pull/2366
- Move Initialization: https://github.com/eclipse-platform/eclipse.platform.swt/pull/2367
- Cursor handle management: this

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
```diff
@@ -424,41 +419,40 @@ public Cursor(Device device, ImageDataProvider imageDataProvider, int hotspotX,
  * @noreference This method is not intended to be referenced by clients.
  */
 public static Long win32_getHandle (Cursor cursor, int zoom) {
-	if (cursor.isDisposed()) {
-		return cursor.handle;
+	if(cursor.zoomLevelToHandle.size() == 0) {
```
HeikoKlare: Why not check for `isDisposed()` anymore?

ShahzaibIbrahim: Previously, isDispose() method meant handle == 0. Which could either mean 
- The handle was destroyed
- The handle was never initialized

With these changes, we are only checking is destroy() is called for the object. Maybe I should update the isDisposed method to mimic the previous behavior. Which will look like this

`return isDestroyed || cursor.zoomLevelToHandle.size() == 0;`

ShahzaibIbrahim:
> * Bug that caused the ImageDataProvider not to be used when requesting a handle for another zoom (as `cursor.source` is not null then, thus this will be used)

I am not sure what bug is it you're talking about. `cursor.source`  is always going to be null when created with `style `**attribute**.  From what I understand you mean the source will not be null when requesting handle for other zoom like this snippet?

```
ImageData source = new ImageData(16, 16, 1, new PaletteData(new RGB[]{new RGB(0, 0, 0)}));
		ImageDataProvider provider = zoom -> source;
Cursor cursorFromProvider = new Cursor(Display.getDefault(), provider, 0, 0);
Cursor.win32_getHandle(cursorFromProvider, 200);
Cursor.win32_getHandle(cursorFromProvider, 100);
```
@HeikoKlare

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
ShahzaibIbrahim: The reason I create a handle at object initialization is to be in compliance with `equals` method. 

Here I have two different cursors:

```
Cursor cursor = new Cursor(display, SWT.CURSOR_WAIT);
Cursor otherCursor = new Cursor(display, SWT.CURSOR_CROSS);
```

and without handle creation at initialization: `cursor  == otherCursor` and as per tests we have in Test_org_eclipse_swt_graphics_Cursor#test_equalsLjava_lang_Object, the two objects are not equal since they expect to have `hashCode == handle` (default handle in new case i.e. for 100% zoom)

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
```diff
@@ -74,6 +62,8 @@ public final class Cursor extends Resource {
 	private final ImageData mask;
 	private final int hotspotX;
 	private final int hotspotY;
+	private boolean isDestroyed;
+	private Integer style;
```
ShahzaibIbrahim: @HeikoKlare would you still like to mark this style field as final? if we do so then I will need to get rid of Cursor(Device) and initialize all other final values in Cursor(Device, int). Also I will need to initialize style attribute as null in all other constructor. 

Also note that once we switch to using [`CursorHandleProvider`](https://github.com/vi-eclipse/Eclipse-Platform/issues/390), we will get rid of all these final fields anyway.

HeikoKlare (changes requested):
Please note that the simple removal of eager handle creation from the constructors will violates the existing contracts. There are exceptions specified, which (except for the `ERROR_NO_HANDLES` ones) must still be thrown in the constructor and not on a lazy handle request.

ShahzaibIbrahim:
> Please note that the simple removal of eager handle creation from the constructors will violates the existing contracts. There are exceptions specified, which (except for the `ERROR_NO_HANDLES` ones) must still be thrown in the constructor and not on a lazy handle request.

Here's all the contracts tested on Cursor object creation

```
ImageData source = new ImageData(16, 16, 1, new PaletteData(new RGB[]{new RGB(0, 0, 0)}));
ImageData mask = new ImageData(16, 16, 1, new PaletteData(new RGB[]{new RGB(0, 0, 0)}));

// Should throw ERROR_INVALID_ARGUMENT
Cursor cursor = new Cursor(Display.getDefault(), -99);

// ERROR_NULL_ARGUMENT - Source is null
 Cursor cursorFromImageAndMask = new Cursor(Display.getDefault(), null, mask, 0, 0);

// ERROR_NULL_ARGUMENT - Mask is null and Source doesn't heve a mask
 Cursor cursorFromImageAndMask1 = new Cursor(Display.getDefault(), source, null, 0, 0);

// ERROR_INVALID_ARGUMENT - source and the mask are not the same size
ImageData source32 = new ImageData(32, 32, 1, new PaletteData(new RGB[]{new RGB(0, 0, 0)}));
ImageData mask16 = new ImageData(16, 16, 1, new PaletteData(new RGB[]{new RGB(0, 0, 0)}));
Cursor cursorFromImageAndMask2 = new Cursor(Display.getDefault(), source32, mask16, 0, 0);

// ERROR_INVALID_ARGUMENT - hotspot is outside the bounds of the image
Cursor cursorFromImageAndMask3 = new Cursor(Display.getDefault(), source, mask, 18, 18);

// ERROR_NULL_ARGUMENT - Source is null
ImageData nullImageData = null;
Cursor cursorFromSourceOnly= new Cursor(Display.getDefault(), nullImageData, 0, 0);

// ERROR_INVALID_ARGUMENT - hotspot is outside the bounds of the image
Cursor cursorFromSourceOnly1 = new Cursor(Display.getDefault(), source, 17, 17);

// ERROR_NULL_ARGUMENT - Source is null
ImageData nullSource = null;
ImageDataProvider provider = zoom -> nullSource;
Cursor cursorFromProvider = new Cursor(Display.getDefault(), provider, 0, 0);
```

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_Cursor.java
```diff
@@ -172,6 +174,74 @@ public void test_ConstructorWithImageDataProvider() {
 	}
 }
 
+@Test
+public void test_InvalidArgumentsForAllConstructors() {
+	ImageData source = new ImageData(16, 16, 1, new PaletteData(new RGB[]{new RGB(0, 0, 0)}));
+	ImageData mask = new ImageData(16, 16, 1, new PaletteData(new RGB[]{new RGB(0, 0, 0)}));
+
+	try {
```
akurtakov: Please use assertThrows instead of this fail.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_Cursor.java
```diff
@@ -172,6 +174,74 @@ public void test_ConstructorWithImageDataProvider() {
...
+		Cursor cursor = new Cursor(Display.getDefault(), -99);
+		cursor.dispose();
+		fail("No exception thrown when wrong style was provided");
+	} catch (IllegalArgumentException e) {
+	}
+
+	try {
+		Cursor cursorFromImageAndMask = new Cursor(Display.getDefault(), null, mask, 0, 0);
+		cursorFromImageAndMask.dispose();
+		fail("No exception thrown when source is null");
```
akurtakov: Please use assertThrows instead of this fail.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_Cursor.java
```diff
@@ -172,6 +174,74 @@ public void test_ConstructorWithImageDataProvider() {
...
+	} catch (IllegalArgumentException e) {
+	}
+
+	try {
+		 Cursor cursorFromImageAndMask1 = new Cursor(Display.getDefault(), source, null, 0, 0);
+		 cursorFromImageAndMask1.dispose();
+		fail("No exception thrown when mask is null and source doesn't heve a mask");
```
akurtakov: Please use assertThrows instead of this fail.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_Cursor.java
```diff
@@ -172,6 +174,74 @@ public void test_ConstructorWithImageDataProvider() {
...
+	} catch (IllegalArgumentException e) {
+	}
+
+	try {
+		ImageData source32 = new ImageData(32, 32, 1, new PaletteData(new RGB[]{new RGB(0, 0, 0)}));
+		ImageData mask16 = new ImageData(16, 16, 1, new PaletteData(new RGB[]{new RGB(0, 0, 0)}));
+		Cursor cursorFromImageAndMask2 = new Cursor(Display.getDefault(), source32, mask16, 0, 0);
+		cursorFromImageAndMask2.dispose();
+		fail("No exception thrown when source and the mask are not the same size");
```
akurtakov: Please use assertThrows instead of this fail.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_Cursor.java
```diff
@@ -172,6 +174,74 @@ public void test_ConstructorWithImageDataProvider() {
...
+	} catch (IllegalArgumentException e) {
+	}
+
+	try {
+		Cursor cursorFromImageAndMask3 = new Cursor(Display.getDefault(), source, mask, 18, 18);
+		cursorFromImageAndMask3.dispose();
+		fail("No exception thrown when hotspot is outside the bounds of the image");
```
akurtakov: Please use assertThrows instead of this fail.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_Cursor.java
```diff
@@ -172,6 +174,74 @@ public void test_ConstructorWithImageDataProvider() {
...
+	} catch (IllegalArgumentException e) {
+	}
+
+	try {
+		ImageData nullImageData = null;
+		Cursor cursorFromSourceOnly= new Cursor(Display.getDefault(), nullImageData, 0, 0);
+		cursorFromSourceOnly.dispose();
+		fail("No exception thrown when source image data is null");
```
akurtakov: Please use assertThrows instead of this fail.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_Cursor.java
```diff
@@ -172,6 +174,74 @@ public void test_ConstructorWithImageDataProvider() {
...
+	} catch (IllegalArgumentException e) {
+	}
+
+	try {
+		ImageDataProvider provider = null;
+		Cursor cursorFromProvider = new Cursor(Display.getDefault(), provider, 0, 0);
+		cursorFromProvider.dispose();
+		fail("No exception thrown when ImageDataProvider is null");
```
akurtakov: Please use assertThrows instead of this fail.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_Cursor.java
```diff
@@ -172,6 +174,74 @@ public void test_ConstructorWithImageDataProvider() {
...
+	} catch (IllegalArgumentException e) {
+	}
+
+	try {
+		ImageData nullSource = null;
+		ImageDataProvider provider = zoom -> nullSource;
+		Cursor cursorFromProvider = new Cursor(Display.getDefault(), provider, 0, 0);
+		cursorFromProvider.dispose();
+		fail("No exception thrown when source is null in ImageDataProvider");
```
akurtakov: Please use assertThrows instead of this fail.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
```diff
@@ -531,19 +514,23 @@ private static interface CursorHandleProvider {
...
-		return setupCursorFromStyle(this.style);
+		long handle = OS.LoadCursor(0, lpCursorName);
+		if (handle == 0) {
+			SWT.error(SWT.ERROR_NO_HANDLES);
+		}
+		return new CustomCursorHandle(handle);
 	}
 
-	private static final CursorHandle setupCursorFromStyle(int style) {
+	private static final long setupCursorFromStyle(int style) {
```
HeikoKlare: This does not do any "setup" anymore but just retrieves some OS identifier for a cursor style, so method name should be adapted.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
```diff
@@ -646,6 +629,13 @@ private static class ImageDataProviderCursorHandleProvider extends HotspotAwareC
 
 	public ImageDataProviderCursorHandleProvider(ImageDataProvider provider, int hotspotX, int hotspotY) {
 		super(hotspotX, hotspotY);
+		ImageData source = provider.getImageData(DEFAULT_ZOOM);
+		if (source == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
+		/* Check the hotspots */
+		if (hotspotX >= source.width || hotspotX < 0 ||
+			hotspotY >= source.height || hotspotY < 0) {
+			SWT.error(SWT.ERROR_INVALID_ARGUMENT);
+		}
```
HeikoKlare: This check (and others in other constructors) now makes the same one inside `setupCursorFromImageData` obsolete as they are already checked here in the provider constructors, so the replicated checks should be removed there.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
```diff
@@ -684,6 +680,16 @@ private static class ImageDataWithMaskCursorHandleProvider extends ImageDataCurs
 
 	public ImageDataWithMaskCursorHandleProvider(ImageData source, ImageData mask, int hotspotX, int hotspotY) {
 		super(source, hotspotX, hotspotY);
+		if (mask == null) {
+			if (source.getTransparencyType() != SWT.TRANSPARENCY_MASK) {
+				SWT.error(SWT.ERROR_NULL_ARGUMENT);
+			}
+			mask = source.getTransparencyMask();
```
HeikoKlare: This changes the current behavior: now the provider is initialized with a mask retrieved from the source instead of just storing the originally passed mask (which may be null). Maybe this is a cleaner way of doing it but (1) it should then be done independent from this change of making handle creation lazy and (2) we then need to ensure that a mask that is first created from source and then scaled is the same as first scaling the source and then extracting a mask out of it.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_Cursor.java
```diff
@@ -162,6 +164,66 @@ public void test_ConstructorWithImageDataProvider() {
 			IllegalArgumentException.class, () -> new Cursor(display, (ImageDataProvider) null, 0, 0));
 }
 
+@Test
+public void test_InvalidArgumentsForAllConstructors() {
+	ImageData source = new ImageData(16, 16, 1, new PaletteData(new RGB[] { new RGB(0, 0, 0) }));
+	ImageData mask = new ImageData(16, 16, 1, new PaletteData(new RGB[] { new RGB(0, 0, 0) }));
+
+	assertThrows("IllegalArgumentException Expected: when wrong style was provided", IllegalArgumentException.class,
```
HeikoKlare: Parts of the error messages are rather obsolete. In particular the expectation of throwing an IllegalArgumentException is already encoded in the assertions themselves.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
```diff
@@ -407,7 +390,7 @@ private void destroyHandle () {
 		handle.destroy();
 	}
 	zoomLevelToHandle.clear();
-	handle = 0;
+	this.isDestroyed = true;
```
HeikoKlare, line 359: This `destroyHandle()` method could be combined with the `destroy()` method and all the comment/commented-out code could be removed from that method.

HeikoKlare (changes requested):
In general, the change looks good. But there is at least one potential behavorial change beyond the lazy handle creation in there that we need to clarify or resolve.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
ShahzaibIbrahim: You are right. I will open up a new ticket after verifying your 2nd statement. I have revert to current behavior. 

> (2) we then need to ensure that a mask that is first created from source and then scaled is the same as first scaling the source and then extracting a mask out of it.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
```diff
@@ -685,6 +644,21 @@ private static class ImageDataWithMaskCursorHandleProvider extends ImageDataCurs
 	public ImageDataWithMaskCursorHandleProvider(ImageData source, ImageData mask, int hotspotX, int hotspotY) {
 		super(source, hotspotX, hotspotY);
 		this.mask = mask;
+		validateMask(source, mask);
+	}
+
+	private void validateMask(ImageData source, ImageData mask) {
+		ImageData testMask = mask == null? null : (ImageData) mask.clone();
```
```suggestion
		ImageData testMask = mask == null ? null : (ImageData) mask.clone();
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
```diff
@@ -646,6 +592,13 @@ private static class ImageDataProviderCursorHandleProvider extends HotspotAwareC
 
 	public ImageDataProviderCursorHandleProvider(ImageDataProvider provider, int hotspotX, int hotspotY) {
 		super(hotspotX, hotspotY);
+		ImageData source = provider.getImageData(DEFAULT_ZOOM);
+		if (source == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
+		/* Check the hotspots */
+		if (hotspotX >= source.width || hotspotX < 0 ||
+			hotspotY >= source.height || hotspotY < 0) {
+			SWT.error(SWT.ERROR_INVALID_ARGUMENT);
+		}
```
HeikoKlare: Since we have the same check two times, maybe we could extract it out into a `validateHotspotInsideImage` or the like?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
```diff
@@ -685,6 +644,21 @@ private static class ImageDataWithMaskCursorHandleProvider extends ImageDataCurs
...
+	}
+
+	private void validateMask(ImageData source, ImageData mask) {
+		ImageData testMask = mask == null? null : (ImageData) mask.clone();
+		if (testMask == null) {
+			if (source.getTransparencyType() != SWT.TRANSPARENCY_MASK) {
+				SWT.error(SWT.ERROR_NULL_ARGUMENT);
+			}
+			testMask = source.getTransparencyMask();
+		}
```
HeikoKlare, line 660: Now this code only seems to be used for validation but not for creating the actual mask. Thus `setupCursorFromImageData(ImageData, ImageData, int, int)` will now fail when passing a `null` mask.

HeikoKlare (changes requested):
The latest change introduced a regression that will make the cursor setup fail when passing a `null` mask.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
```diff
@@ -639,13 +588,24 @@ protected final int getHotpotXInPixels(int zoom) {
 	protected final int getHotpotYInPixels(int zoom) {
 		return Win32DPIUtils.pointToPixel(hotspotY, zoom);
 	}
+
+	protected static final void validateHotspotInsideImage(ImageData source, int hotspotX, int hotspotY) {
+		if (source == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
```
HeikoKlare: This check is unrelated to the hotspot validation, so it should be moved back to the callers.

PR #2364, opened 2025-07-31
Win32DPIUtils:pointToPixel shouldn't have residual

amartya4256:
This commit contributes to scaling points and rectangles from point to pixels while making sure they do not have any residuals since the OS doesn't support sub-pixel drawing.

contributes to #62

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/Win32DPIUtils.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/Win32DPIUtils.java
@@ -205,7 +205,8 @@ public class Win32DPIUtils {
 		float scaleFactor = DPIUtil.getScalingFactor(zoom);
 		float scaledX = fPoint.getX() * scaleFactor;
 		float scaledY = fPoint.getY() * scaleFactor;
-		return new Point.OfFloat(scaledX, scaledY);
+		Point scaledPoint = new Point.OfFloat(scaledX, scaledY);
+		return new Point.OfFloat(scaledPoint.x, scaledPoint.y);
 	}
 
 	public static Point pointToPixel(Drawable drawable, Point point, int zoom) {
@@ -228,7 +229,8 @@ public class Win32DPIUtils {
 	}
 
 	private static Rectangle pointToPixel(Rectangle.OfFloat rect, int zoom) {
-		return scaleBounds(rect, zoom, 100);
+		Rectangle scaledRectangle = scaleBounds(rect, zoom, 100);
+		return new Rectangle.OfFloat(scaledRectangle.x, scaledRectangle.y, scaledRectangle.width, scaledRectangle.height);
 	}
 
 	public static Rectangle pointToPixel(Drawable drawable, Rectangle rect, int zoom) {
--- a/tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/tests/win32/Win32DPIUtilTests.java
+++ b/tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/tests/win32/Win32DPIUtilTests.java
@@ -223,14 +223,26 @@ public class Win32DPIUtilTests {
 	@Test
 	public void scaleDownscaleUpRectangleInvertible() {
 		int[] zooms = new int[] {25, 50, 75, 100, 125, 150, 175, 200, 225, 250, 275, 300, 325, 350, 375, 400};
+		for (int zoom : zooms) {
+			for (int i = 1; i <= 10000; i++) {
+				Rectangle rect = new Rectangle.OfFloat(0, 0, i, i);
+				Rectangle scaleDown = Win32DPIUtils.pixelToPoint(rect, zoom);
+				Rectangle scaleUp = Win32DPIUtils.pointToPixel(scaleDown, zoom);
+				assertEquals(rect.width, scaleUp.width);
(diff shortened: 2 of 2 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/Win32DPIUtils.java
```diff
@@ -205,7 +205,8 @@ public static Point pointToPixel(Point point, int zoom) {
 		float scaleFactor = DPIUtil.getScalingFactor(zoom);
 		float scaledX = fPoint.getX() * scaleFactor;
 		float scaledY = fPoint.getY() * scaleFactor;
-		return new Point.OfFloat(scaledX, scaledY);
+		Point scaledPoint = new Point.OfFloat(scaledX, scaledY);
+		return new Point.OfFloat(scaledPoint.x, scaledPoint.y);
```
fedejeanne: No need to instantiate twice, do either:
```java
// Cast to int so that there are no residuals
return new Point.OfFloat((int) scaledX, (int) scaledY);
```

or even better, simply declare `scaledX` and `scaledY` as `int`s
```java
// Use ints so that there are no residuals
int scaledX = ...;
int scaledY = ...;
return new Point.OfFloat(scaledX, scaledY);
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/Win32DPIUtils.java
```diff
@@ -228,7 +229,8 @@ public static Rectangle pointToPixel(Rectangle rect, int zoom) {
 	}
 
 	private static Rectangle pointToPixel(Rectangle.OfFloat rect, int zoom) {
-		return scaleBounds(rect, zoom, 100);
+		Rectangle scaledRectangle = scaleBounds(rect, zoom, 100);
+		return new Rectangle.OfFloat(scaledRectangle.x, scaledRectangle.y, scaledRectangle.width, scaledRectangle.height);
```
fedejeanne: Why do you need to create yet another copy? `scaleBounds(...)` already returns a shallow copy. In fact, in doing so, you might loose information about the monitor since you're not using `clone`, you're using a constructor.

fedejeanne (changes requested):

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/Win32DPIUtils.java
amartya4256: (int) will not round it properly. We will need to use Math.round instead.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/Win32DPIUtils.java
amartya4256: Just thinking if we could somehow reset the residual values.

fedejeanne: Ah, I see. Tricky to do without exposing too much about the internal workings of `OfFloat`.

I would be OK with doing this:
```java
Rectangle scaledRectangle = scaleBounds(rect, zoom, 100);
scaledRectangle.setX(Math.round(scaledRectangle.getX());
scaledRectangle.setY(Math.round(scaledRectangle.getY());
```

It's not perfectly encapsulated (exposes the fact that you use `Math.round(...)` instead of just casting to `int`) but at least it preserves the `monitor` (if it's there).

If you don't want to expose that detail then you should add yet another method to `Rectangle.OfFloat` that does the rounding and looses the residuals. Something like:

```java
void round() {
   setX(Math.round(getX());
   setY(Math.round(getY());
}
```

Your call.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/Win32DPIUtils.java
fedejeanne: See https://github.com/eclipse-platform/eclipse.platform.swt/pull/2364/files/ca6ef3a35062a41e9e47b39d9e3b071fabbc3811#r2379074441 since you are probably trying to do the same here and you have to preserve the other information (`monitor`) here too.

fedejeanne:
@amartya4256 I rebased on `master`, please pull.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/Win32DPIUtils.java
HeikoKlare: Isn't the point of this code to ensure that we have a consistent way of rounding (by delegating the responsibility to `Point.OfFloat`), which would break with replicating the actual rounding implementation (via Math.round) here?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/Win32DPIUtils.java
fedejeanne: @akoch-yatta are you taking over this PR too?

If a better (and simple) solution can't be found then I suggest we merge it as-is. I don't think this is going to be a performance bottleneck anyway and the rest of the PR looks fine.

I'll withdraw my RFC.

PR #2366, opened 2025-08-01
Fix Cursor hotspot coordinate calculation

ShahzaibIbrahim:
While retrieving the handle for the cursor for specific zoom, currently we scale up the cursor source image but not the hotspot coordinates. This commit ensures we have per-zoom coordinates for the cursor.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
@@ -429,11 +429,11 @@ public static Long win32_getHandle (Cursor cursor, int zoom) {
 			source = DPIUtil.scaleImageData(cursor.device, cursor.source, zoom, DEFAULT_ZOOM);
 		}
 		if (cursor.isIcon) {
-			Cursor newCursor = new Cursor(cursor.device, source, cursor.hotspotX, cursor.hotspotY);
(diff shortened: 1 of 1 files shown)


HeikoKlare (approved):
The change is fine, thank you!

Testing this change shows that there actually was an issue with the hotspot values when using monitor zooms other than 100%, as can be seen in the screen captures below.

There are no actual usages of those hotspot-capable cursor constructors inside the platform projects. But since I know that GEF is making quite some use of custom cursors, I also checked that code and found that the `FlyoutPaletteComposite` [uses the hotspot functionality for the drag cursors](https://github.com/eclipse-gef/gef-classic/blob/50e27e3f2f36a1c1d161a219e3088e8ba6d4f1ed/org.eclipse.gef/src/org/eclipse/gef/ui/palette/FlyoutPaletteComposite.java#L1613). You can also see that the behavior was broken before and is fixed by this PR in the screen capture below. Other usages just pass hotspot values of 0, where the added conversion does not make a difference. @ptziegler fyi, just in case I missed some other usage in GEF that might have relied on the existing behavior in some way.

Here are some screen captures at 175% monitor zoom (with `swt.autoScale=int200`). I also successfully tested with `swt.autoScale=150` on a 100% monitor but did not take video of it.

**Before**
cursor_wrong_hotspot cursor_wrong_hotspot_gef

**After**
cursor_correct_hotspot_gef cursor_correct_cursor_gef

PR #2367, opened 2025-08-01
[Win32] Refactor the Cursor constructors 

ShahzaibIbrahim:
Extracting out the OS handle creation logic into separate methods. Intention is to make them reusable if someone want to create only a OS handle within a Cursor class.


HeikoKlare (approved):
The refactoring is fine. I just slightly adapted it to avoid method movements, such that the change remains small both for easier history lookup and for easier review.

PR #2372, opened 2025-08-04
[GTK] Implement Display.getHighContrast

akurtakov:
Gtk ships with "HighContrast" and "HighContrastInverse" and Mate desktop ships "ContrastHigh*" theme. If the gtk theme is any of these say it's "high contrast" theme.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
@@ -2519,6 +2519,10 @@ public Control getFocusControl () {
  */
 public boolean getHighContrast () {
 	checkDevice ();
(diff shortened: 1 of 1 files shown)


merks:
Maybe a dumb thought. But are there low contrast themes?  Maybe one could just check for contrast and assume high.

akurtakov:
If GTK  and theme developers claim only a few of these are "HighContrast" I would trust them rather than provide new meaning of the same terms. 
No one knows what weird names one might have come up with so "Contrast" is not enough for me.

akurtakov:
I have searched a bit and there seem to be efforts for "low contrast" themes https://www.gnome-look.org/p/1013697 . It's in the description but nothing guarantees that no one named a theme "LowerContrast"/"LessContrast" or anything similar.

akurtakov:
As per previous comments I'm merging this one. It's better to have 100% sure API than "guesses".

PR #2376, opened 2025-08-05
Move Windows only functionality to Win32DPIUtil

ShahzaibIbrahim:
Moving methods that are consumed only by Windows from DPIUtil to Win32DPIUtil (global state handling for monitor-specific scaling)

**Note: The methods are moved using eclipse refactor tool.**

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/internal/CocoaDPIUtil.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/internal/CocoaDPIUtil.java
@@ -1,0 +1,32 @@
+/*******************************************************************************
+ * Copyright (c) 2025 Yatta Solutions and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Yatta Solutions - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.swt.internal;
+
+public class CocoaDPIUtil {
+
+	/**
+	 * Auto-scale down int dimensions.
+	 */
+	public static int pixelToPoint(int size) {
+		return DPIUtil.pixelToPoint(size, DPIUtil.getDeviceZoom());
+	}
+
+	/**
+	 * Auto-scale down float dimensions.
+	 */
+	public static float pixelToPoint(float size) {
+		return DPIUtil.pixelToPoint(size, DPIUtil.getDeviceZoom());
+	}
+
+}
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
@@ -96,17 +96,6 @@ public class DPIUtil {
 	 */
 	private static final String SWT_AUTOSCALE_METHOD = "swt.autoScale.method";
 
-	/**
-	 * System property to enable to scale the application on runtime
-	 * when a DPI change is detected.
-	 * <ul>
-	 * <li>"true": the application is scaled on DPI changes</li>
-	 * <li>"false": the application will remain in its initial scaling</li>
-	 * </ul>
-	 * <b>Important:</b> This flag is only parsed and used on Win32. Setting it to
-	 * true on GTK or cocoa will be ignored.
-	 */
-	private static final String SWT_AUTOSCALE_UPDATE_ON_RUNTIME = "swt.autoScale.updateOnRuntime";
 	static {
 		autoScaleValue = System.getProperty (SWT_AUTOSCALE);
 
@@ -115,6 +104,13 @@ public class DPIUtil {
 		autoScaleMethod = AUTO_SCALE_METHOD_SETTING != AutoScaleMethod.AUTO ? AUTO_SCALE_METHOD_SETTING : AutoScaleMethod.NEAREST;
 	}
 
+static String getAutoScaleValue() {
+	return autoScaleValue;
+}
+
+static void setAutoScaleValue(String autoScaleValueArg) {
+	autoScaleValue = autoScaleValueArg;
+}
 
 public static int pixelToPoint(int size, int zoom) {
 	if (zoom == 100 || size == SWT.DEFAULT) return size;
@@ -327,7 +323,7 @@ public static void setDeviceZoom (int nativeDeviceZoom) {
 	// in GTK, preserve the current method when switching to a 100% monitor
 	boolean preserveScalingMethod = SWT.getPlatform().equals("gtk") && deviceZoom == 100;
 	if (!preserveScalingMethod && AUTO_SCALE_METHOD_SETTING == AutoScaleMethod.AUTO) {
-		if (sholdUseSmoothScaling()) {
+		if (useSmoothScalingByDefaultProvider.shouldUseSmoothScaling()) {
 			autoScaleMethod = AutoScaleMethod.SMOOTH;
 		} else {
 			autoScaleMethod = AutoScaleMethod.NEAREST;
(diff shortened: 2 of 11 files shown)


bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
```diff
@@ -227,6 +227,11 @@ public static int mapZoomToDPI (int zoom) {
 	return roundedDpi;
 }
 
+public static boolean isMonitorSpecificScalingActive() {
+	boolean updateOnRuntimeValue = Boolean.getBoolean (SWT_AUTOSCALE_UPDATE_ON_RUNTIME);
+	return updateOnRuntimeValue;
+}
```
HeikoKlare: This functionality is specific to Win32, so it should be moved to the `Win32DPIUtils`. Otherwise it would not make sense to move the setter there.
The reason for not moving this seems to be the `shouldUseSmoothScaling` method. I propose to invert the functionality of that method. Instead of retrieving OS-specific information in that method, we may store a `useSmoothScalingByDefaultProvider`, which initially returns `false` and is overwritten by the OS-specific implementation via some `setUseSmoothScalingByDefaultProvider`. The `GtkDPIUtils` can then set an according provider with the current implementation in `shouldUseSmoothScaling()` (e.g., in a static initializer of `GtkDPIUtils`), and the `Win32DPIUtils` can either do the same or even set/reset the provider inside `setMonitorSpecificScaling()`.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
```diff
@@ -106,7 +106,7 @@ public static Optional<AutoScaleMethod> forString(String s) {
 	 * <b>Important:</b> This flag is only parsed and used on Win32. Setting it to
 	 * true on GTK or cocoa will be ignored.
 	 */
-	private static final String SWT_AUTOSCALE_UPDATE_ON_RUNTIME = "swt.autoScale.updateOnRuntime";
+	public static final String SWT_AUTOSCALE_UPDATE_ON_RUNTIME = "swt.autoScale.updateOnRuntime";
```
HeikoKlare: This should go into the `Win32DPIUtils` as well.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
```diff
@@ -58,7 +58,7 @@ public static Optional<AutoScaleMethod> forString(String s) {
 	private static final AutoScaleMethod AUTO_SCALE_METHOD_SETTING;
 	private static AutoScaleMethod autoScaleMethod;
 
-	private static String autoScaleValue;
+	public static String autoScaleValue;
```
HeikoKlare: We should not make this public. It might become package protected (as only accessor is `Win32DPIUtil` which is placed in the same package), but even better it should only be accessed with a package-protected getter.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
```diff
@@ -335,12 +335,19 @@ public static void setDeviceZoom (int nativeDeviceZoom) {
...
 }
 
+@FunctionalInterface
+public interface UseSmoothScalingProvider {
+    boolean shouldUseSmoothScaling();
+}
+
+private static UseSmoothScalingProvider useSmoothScalingByDefaultProvider = () -> false;
+
+public static void setUseSmoothScalingByDefaultProvider(UseSmoothScalingProvider provider) {
```
HeikoKlare:
```suggestion
static void setUseSmoothScalingByDefaultProvider(UseSmoothScalingProvider provider) {
```

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
```diff
@@ -335,12 +335,19 @@ public static void setDeviceZoom (int nativeDeviceZoom) {
 	}
 }
 
+@FunctionalInterface
+public interface UseSmoothScalingProvider {
```
HeikoKlare:
```suggestion
interface UseSmoothScalingProvider {
```

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
```diff
@@ -335,12 +335,19 @@ public static void setDeviceZoom (int nativeDeviceZoom) {
...
+    useSmoothScalingByDefaultProvider = provider != null ? provider : () -> false;
+}
+
 private static boolean sholdUseSmoothScaling() {
-	return switch (SWT.getPlatform()) {
-	case "gtk" -> deviceZoom / 100 * 100 != deviceZoom;
-	case "win32" -> isMonitorSpecificScalingActive();
-	default -> false;
-	};
+	return useSmoothScalingByDefaultProvider.shouldUseSmoothScaling();
```
HeikoKlare: This method can be inlined.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
```diff
@@ -335,12 +335,19 @@ public static void setDeviceZoom (int nativeDeviceZoom) {
...
 
+@FunctionalInterface
+public interface UseSmoothScalingProvider {
+    boolean shouldUseSmoothScaling();
+}
+
+private static UseSmoothScalingProvider useSmoothScalingByDefaultProvider = () -> false;
+
+public static void setUseSmoothScalingByDefaultProvider(UseSmoothScalingProvider provider) {
+    useSmoothScalingByDefaultProvider = provider != null ? provider : () -> false;
```
HeikoKlare:
```suggestion
    useSmoothScalingByDefaultProvider = provider;
```

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
ShahzaibIbrahim: Can't make it package protected since GTKDPIUtil now is part of Image and thus not in a same package.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
ShahzaibIbrahim: Can't make it package protected since GTKDPIUtil now is part of Image and thus not in a same package.

HeikoKlare: Then we should move that class to the same package, as I would try to avoid making this method public if not really necessary.

ShahzaibIbrahim: Moving **GTKDPIUtil** and **CocoaDPIUtil** in org.eclipse.swt.internal package. Now we can make `setUseSmoothScalingByDefaultProvider ` and `UseSmoothScalingProvider  `package protected

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/internal/CocoaDPIUtil.java
```diff
@@ -0,0 +1,19 @@
+package org.eclipse.swt.internal;
```
HeikoKlare, line 14: This needs a copyright header.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/internal/GtkDPIUtil.java
```diff
@@ -0,0 +1,21 @@
+package org.eclipse.swt.internal;
```
HeikoKlare, line 14: This needs a copyright header

HeikoKlare (commented):
Except for missing copyright headers, the PR looks fine to me.

PR #2377, opened 2025-08-05
Make fonts use  autoscaled zoom when     swt.autoScale is fixed

### LINKED ISSUE #2361: [Win32] Wrong text extent calculation with swt.autoScale=int200
**Describe the bug**
When running an Eclipse product without monitor-specific scaling (e.g., by disabling the preference or with `-Dswt.autoScale.updateOnRuntime=false`) and default auto-scale settings (i.e., the implicit `swt.autoScale=quarter`), wrong font sizes are considered at some places. This particularly includes the header size calculation of `CTabFolders`, which may become too large/small.
This also happens when running any pure SWT application with default auto-scale settings, such as the `CustomControlExample`.

It is a regression from:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2306

**To Reproduce**
- Set primary monitor zoom to something different than 100% or 200% (e.g., 150%).
- Start the `CustomControlExample` with default settings and go to the `CTabFolder` tab or start an Eclipse SDK.
- You will see that tab headers are too small.

**Expected behavior**
`CTabFolder` headers shall be sized correct.

**Screenshots**
Image

Image

**Environment:**
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 

**Version since**
I20250714-1800

arunjose696:
This change ensures that images and fonts share the same zoom level when  a fixed autoscale value is used.

Previously, when different GCs were created for images and widgets, their native zoom(GC.data.nativeZoom) levels could differ, causing inconsistencies.

For example, in `LineNumberRuler`, one GC is created from an image (used for rendering), and another is created from a widget (used for text measurement). The differing native zoom levels between these GCs led to incorrect text width calculations. Specifically, one GC was used to measure the text extent, while the other used those measurements to calculate the width of the columns where the text would be drawn and this resulted in #2311

### Steps to reproduce
1)Checkout the first commit.  Run steps in https://github.com/eclipse-platform/eclipse.platform.swt/pull/2306
2)Checkout the second commit, this solves the issue with lineNumberRuler

First commit which reverts changes from #2306  Fixes #2361 
Second commit fixes the original issue in #2311

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
@@ -57,8 +57,7 @@ public class DPIUtil {
 	}
 	private static final AutoScaleMethod AUTO_SCALE_METHOD_SETTING;
 	private static AutoScaleMethod autoScaleMethod;
-
-	private static String autoScaleValue;
+	private static AutoScaleConfig autoScaleConfig;
 
 	/**
 	 * System property that controls the autoScale functionality.
@@ -97,20 +96,40 @@ public class DPIUtil {
 	private static final String SWT_AUTOSCALE_METHOD = "swt.autoScale.method";
 
 	static {
-		autoScaleValue = System.getProperty (SWT_AUTOSCALE);
-
+		setAutoScaleValue(System.getProperty (SWT_AUTOSCALE));
 		String value = System.getProperty (SWT_AUTOSCALE_METHOD);
 		AUTO_SCALE_METHOD_SETTING = AutoScaleMethod.forString(value).orElse(AutoScaleMethod.AUTO);
 		autoScaleMethod = AUTO_SCALE_METHOD_SETTING != AutoScaleMethod.AUTO ? AUTO_SCALE_METHOD_SETTING : AutoScaleMethod.NEAREST;
 	}
 
 static String getAutoScaleValue() {
-	return autoScaleValue;
+	return autoScaleConfig.value;
 }
 
 static void setAutoScaleValue(String autoScaleValueArg) {
-	autoScaleValue = autoScaleValueArg;
+	autoScaleConfig = new AutoScaleConfig(autoScaleValueArg);
 }
+
+	private static final class AutoScaleConfig {
+		private final String value;
+		private final boolean useAutoScaledFontZoom;
+
+		AutoScaleConfig(String value) {
+			this.value = value;
+			useAutoScaledFontZoom = isIntegerAutoScale(value);
+		}
+
+		private static boolean isIntegerAutoScale(String value) {
+			if (value == null)
+				return false;
+			try {
+				Integer.parseInt(value);
+				return true;
+			} catch (NumberFormatException e) {
+				return false;
+			}
+		}
+	}
 
 public static int pixelToPoint(int size, int zoom) {
 	if (zoom == 100 || size == SWT.DEFAULT) return size;
@@ -343,7 +362,15 @@ static void setUseSmoothScalingByDefaultProvider(UseSmoothScalingProvider provid
 }
 
 public static int getZoomForAutoscaleProperty (int nativeDeviceZoom) {
-	return getZoomForAutoscaleProperty(nativeDeviceZoom, autoScaleValue);
+	return getZoomForAutoscaleProperty(nativeDeviceZoom, autoScaleConfig.value);
+}
+
+public static int getFontZoomForAutoscaleProperty(int nativeDeviceZoom) {
+	if (autoScaleConfig.useAutoScaledFontZoom ) {
+		return getZoomForAutoscaleProperty(nativeDeviceZoom);
+	} else {
+		return nativeDeviceZoom;
+	}
 }
 
 private static int getZoomForAutoscaleProperty (int nativeDeviceZoom, String autoScaleValue) {
@@ -376,13 +403,13 @@ private static int getZoomForAutoscaleProperty (int nativeDeviceZoom, String aut
 }
 
 public static void runWithAutoScaleValue(String autoScaleValue, Runnable runnable) {
-	String initialAutoScaleValue = DPIUtil.autoScaleValue;
-	DPIUtil.autoScaleValue = autoScaleValue;
+	String initialAutoScaleValue = autoScaleConfig.value;
+	setAutoScaleValue(autoScaleValue);
 	DPIUtil.deviceZoom = getZoomForAutoscaleProperty(nativeDeviceZoom);
 	try {
 		runnable.run();
 	} finally {
-		DPIUtil.autoScaleValue = initialAutoScaleValue;
+		setAutoScaleValue(initialAutoScaleValue);
 		DPIUtil.deviceZoom = getZoomForAutoscaleProperty(nativeDeviceZoom);
 	}
 }
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
@@ -342,7 +342,7 @@ void checkGC(int mask) {
 			}
 		}
 		if ((state & FONT) != 0) {
-			long fontHandle = SWTFontProvider.getFontHandle(data.font, data.nativeZoom);
+			long fontHandle = SWTFontProvider.getFontHandle(data.font, getFontZoom());
 			OS.SelectObject(handle, fontHandle);
 			long[] hFont = new long[1];
 			long gdipFont = createGdipFont(handle, fontHandle, gdipGraphics, device.fontCollection, null, hFont);
@@ -463,7 +463,7 @@ void checkGC(int mask) {
 		OS.SetTextColor(handle, data.foreground);
 	}
 	if ((state & FONT) != 0) {
-		long fontHandle = SWTFontProvider.getFontHandle(data.font, data.nativeZoom);
+		long fontHandle = SWTFontProvider.getFontHandle(data.font, getFontZoom());
 		OS.SelectObject(handle, fontHandle);
 	}
 }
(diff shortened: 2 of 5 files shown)


bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
```diff
@@ -347,6 +347,18 @@ public static int getZoomForAutoscaleProperty (int nativeDeviceZoom) {
 	return getZoomForAutoscaleProperty(nativeDeviceZoom, autoScaleValue);
 }
 
+public static int getNativeZoomForAutoscaleProperty (int nativeDeviceZoom) {
+	int zoom = 0;
+	try {
+		// Native zoom is set to the autoscaled zoom if autoscale value uses an integer value 
+		Integer.parseInt (autoScaleValue);
+		zoom = getZoomForAutoscaleProperty(nativeDeviceZoom);
```
akoch-yatta: Can't your just use zoom = Integer.parseInt (autoScaleValue) here?

arunjose696: The getZoomForAutoScaleProperty function includes extra logic to ensure that the zoom level stays within the minimum and maximum limits of 25 and 1600.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/TextLayout.java
```diff
@@ -770,9 +770,9 @@ public void draw (GC gc, int x, int y, int selectionStart, int selectionEnd, Col
 
 private int getNativeZoom(GC gc) {
 	if (gc != null) {
-		return gc.data.nativeZoom;
+		return DPIUtil.getNativeZoomForAutoscaleProperty(gc.data.nativeZoom);
 	}
-	return nativeZoom;
+	return DPIUtil.getNativeZoomForAutoscaleProperty(nativeZoom);
```
HeikoKlare: I wonder if this is really the right place/change to `TextLayout`. We should clarify what the field `nativeZoom` in `TextLayout` is supposed to be: is it really a `nativeZoom` (such as the native zoom of a GC of a StyledText) or it is rather a `fontZoom`? If it was really a `nativeZoom`, I would find the change weird because `computeRuns` calls this method to set the `nativeZoom` field, which would make it effectively some `fontZoom`. When I take a look at the usages of `nativeZoom`, it seems to rather be a `fontZoom`. Then, however, calling `DPIUtil.getNativeZoomForAutoscaleProperty(nativeZoom)` would not really make sense here, as `nativeZoom` already is such an adapted value.
In my understanding `nativeZoom` should be renamed to `fontZoom` and then only the change to the first return statement should be applied.

this really the right place for the adaptation in `TextLayout()`? I see three callers of this method:
- `computeRuns()` calls it to set a new `nativeZoom`: shouldn't that native zoom

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
```diff
@@ -347,6 +347,18 @@ public static int getZoomForAutoscaleProperty (int nativeDeviceZoom) {
 	return getZoomForAutoscaleProperty(nativeDeviceZoom, autoScaleValue);
 }
 
+public static int getNativeZoomForAutoscaleProperty (int nativeDeviceZoom) {
```
HeikoKlare: I find this method name a bit confusing. From it's name, I would expect the method to rather return the argument, as passing a native zoom to get a native zoom should be kind of a no-op. Isn't this rather a `getFontZoomForAutoscaleProperty()`, as it is supposed to return the zoom to be used for fonts based on the given native zoom?

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
```diff
@@ -347,6 +347,18 @@ public static int getZoomForAutoscaleProperty (int nativeDeviceZoom) {
...
+	int zoom = 0;
+	try {
+		// Native zoom is set to the autoscaled zoom if autoscale value uses an integer value 
+		Integer.parseInt (autoScaleValue);
```
HeikoKlare: I am not sure if this is a good approach of dealing with the different cases of font zoom calculation:
1. The exception-driven logic is not that easy to understand.
2. The performance of exception-driven logic is bad and this will be called quite often with a thrown exception being the standard case

Wouldn't it make sense when changing the `autoScaleValue` to store a `fontZoomCalulator` that, depending on whether the `autoScaleValue` to be set is an integer, either gets the `getZoomForAutoscaleProperty` call assigned or just returns the given zoom. That would avoid that this logic to identify how to do the calculation is called everytime a zoom is to be convered.

HeikoKlare (commented):
The changes look sound to me. I have some proposals and implicit questions whether I understood everything correctly.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
arunjose696: I have renamed this

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
arunjose696: Have modified the code to set a flag `useAutoScaledFontZoom` when the autoScaleValue  is set.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/TextLayout.java
arunjose696: Since the native zoom is set based on a font when a font is applied to a TextLayout, it makes sense to rename nativeZoom to fontZoom. 

 I reviewed all usages of the variable. There is a minor ambiguity in the getZoom() method, where `nativeZoom`(now `fontZoom`)  used. This method is called for line and offset operations, which I am slightly not sure if it is directly tied to font . However, since line height and offsets are also influenced by the font zoom, I believe the renaming still makes sense.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/TextLayout.java
```diff
@@ -67,7 +67,7 @@ public final class TextLayout extends Resource {
 
 	private MetricsAdapter metricsAdapter = new MetricsAdapter();
 
-	int nativeZoom = DPIUtil.getNativeDeviceZoom();
+	int fontZoom = DPIUtil.getNativeDeviceZoom();
```
HeikoKlare: Shouldn't this be `DPIUtil.getFontZoomForAutoscaleProperty(DPIUtil.getNativeDeviceZoom())` for the sake of consistency?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -5939,7 +5939,11 @@ private static int sin(int angle, int length) {
 }
 
 int getZoom() {
-	return DPIUtil.getZoomForAutoscaleProperty(data.nativeZoom);
+	return DPIUtil.getZoomForAutoscaleProperty(getNativeZoom());
+}
+
+int getNativeZoom() {
```
HeikoKlare: Should this be `getFontZoom()` for the sake of consistency?

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
```diff
@@ -347,6 +348,29 @@ public static int getZoomForAutoscaleProperty (int nativeDeviceZoom) {
...
+	if (useAutoScaledFontZoom) {
+		return getZoomForAutoscaleProperty(nativeDeviceZoom);
+	} else {
+		return nativeDeviceZoom;
+	}
+}
+
+private static void setAutoScaleValue(String value) {
+	autoScaleValue = value;
+	useAutoScaledFontZoom = isIntegerAutoScale(value);
```
HeikoKlare: nit: since these values are tightly coupled (they have to be set together), one could also encapsulate them in an immutable `AutoScaleValue` object that initializes the values consistently and also encapsulates the required `isIntegerAutoScale` functionality. But feel free to ignore this, as it's fine as is and would just be a slight design improvement.

HeikoKlare (commented):
Thank you for the update, looks even cleaner now!

I have only minor comments/proposals left and one question regarding the initialization of the `fontZoom` (former `nativeZoom`) in `TextLayout`.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/TextLayout.java
arunjose696: Done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
arunjose696: Done

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
arunjose696: I have created a class AutoScaleConfig to encapsulate this

HeikoKlare:
The change now looks fine to me. It just has conflicts now because we meanwhile merged https://github.com/eclipse-platform/eclipse.platform.swt/pull/2376.
Could you adapt this PR, @arunjose696?

arunjose696:
> The change now looks fine to me. It just has conflicts now because we meanwhile merged #2376. Could you adapt this PR, @arunjose696?

done

HeikoKlare:
@arunjose696 since we found that this PR was created under a wrong assumption (fonts to be scaled according to autoscale zoom instead of native zoom when using a fixed autoscale value), I think unfortunately this PR cannot be merged/adapted in a reasonable way. Can we close it or do you see anything of value in it that we should extract into another PR?

arunjose696:
Closing this PR

PR #2378, opened 2025-08-05
Initialization handles without creating a Cursor instance

ShahzaibIbrahim:
In win32_getHandle, a new cursor instance is created to retrieve the OS handle. With this change, we are able to safely retrieve the handle without creating a cursor instance.

### How to Reproduce

To reproduce the issue, run the following snippet with resource tracking turned on:

`-Dorg.eclipse.swt.graphics.Resource.reportNonDisposed=true`

```java

public class Snippet {
	public static void main(String[] args) throws InterruptedException  {
		Cursor cursor = new Cursor(Display.getDefault(), new ImageData(1, 1, 24, new PaletteData(0xFF, 0xFF00, 0xFF0000)), 0, 0);
		Cursor.win32_getHandle(cursor, 200);
		System.gc();
		Thread.sleep(1000);
	}
}
```

You will see the following error:

```
java.lang.Error: SWT Resource was not properly disposed
	at org.eclipse.swt.graphics.Resource.initNonDisposeTracking(Resource.java:191)
	at org.eclipse.swt.graphics.Resource.<init>(Resource.java:124)
	at org.eclipse.swt.graphics.Cursor.<init>(Cursor.java:282)
	at org.eclipse.swt.graphics.Cursor.win32_getHandle(Cursor.java:447)
	at org.eclipse.swt.snippets.Snippet.main(Snippet.java:9)

```

### Expected Result

With this PR, there should be no `NonDisposeResource` error since we are now not creating a new Cursor instance to retrieve the handle.

Fixes: https://github.com/eclipse-platform/eclipse.platform.swt/issues/2355

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
@@ -444,12 +444,12 @@ public static Long win32_getHandle (Cursor cursor, int zoom) {
 			source = DPIUtil.scaleImageData(cursor.device, cursor.source, zoom, DEFAULT_ZOOM);
 		}
 		if (cursor.isIcon) {
-			Cursor newCursor = new Cursor(cursor.device, source, Win32DPIUtils.pointToPixel(cursor.hotspotX, zoom), Win32DPIUtils.pointToPixel(cursor.hotspotY, zoom));
-			cursor.setHandleForZoomLevel(newCursor.handle, zoom);
+			long handle = setupCursorFromImageData(cursor.getDevice(), source, Win32DPIUtils.pointToPixel(cursor.hotspotX, zoom), Win32DPIUtils.pointToPixel(cursor.hotspotY, zoom));
+			cursor.setHandleForZoomLevel(handle, zoom);
 		} else {
-			ImageData mask = DPIUtil.scaleImageData(cursor.device, cursor.mask, zoom, DEFAULT_ZOOM);
-			Cursor newCursor = new Cursor(cursor.device, source, mask, Win32DPIUtils.pointToPixel(cursor.hotspotX, zoom), Win32DPIUtils.pointToPixel(cursor.hotspotY, zoom));
-			cursor.setHandleForZoomLevel(newCursor.handle, zoom);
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
```diff
@@ -74,6 +74,7 @@ public final class Cursor extends Resource {
 	private final ImageData mask;
 	private final int hotspotX;
 	private final int hotspotY;
+	private Integer style;
```
```suggestion
	private final Integer style;
```

HeikoKlare (commented):
We now have three mutually exclusive fields (`source`, `imageDataProvider` and `style`). Only one of them is set, depending on how the cursor was initialized. And there are three `setup...` methods that can be executed depending on which of those three fields is set.

Can't we wrap that into something like a `CursorHandleProvider` interface, which has three implementations, each just accepting the necessary information (`ImageData`, `ImageDataProvider` or int/style) and then creating the handle based on that specific information. That would avoid that we always have two unused fields and that we need to switch logic depending on which of them is set.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
ShahzaibIbrahim: Marking it final means that I can no longer call the base constructor **this(device)** since I will have to instantiate the value of style there and once it's initialized it cannot be reassigned in style constructor. If we really want to make it final then maybe instead of calling the base constructor we initialize other final values with null/default in style constructor.

ShahzaibIbrahim:
> Can't we wrap that into something like a `CursorHandleProvider` interface, which has three implementations, each just accepting the necessary information (`ImageData`, `ImageDataProvider` or int/style) and then creating the handle based on that specific information. That would avoid that we always have two unused fields and that we need to switch logic depending on which of them is set.

Good Idea, we can do it perhaps after we are done fixing this issue, in a separate PR?

HeikoKlare:
> Good Idea, we can do it perhaps after we are done fixing this issue, in a separate PR?

Yes, we can do that. But then we should probably leave out the addition of the `style` field in this PR, as it does not appear to be necessary for the fix of not creating unnecessary Cursor instances.

HeikoKlare (approved):

PR #2379, opened 2025-08-05
[win32] Use deviceZoom in single zoom mapper

akoch-yatta:
This PR replaces the usage of the zoom attribute in methods in the `SingleZoomCoordinateMapper` that translate `Point` and `Rectangle` from and to display coordinates to use `DPIUtil.getDeviceZoom()` instead. As each widget inherits its zoom from  `DPIUtil.getDeviceZoom()` when the `SingleZoomCoordinateMapper` is involved, the behavior is not changed in most scenarios. but this PR fixes the scenarios, if a widget is scaled to a different zoom, e.g. by utilizing `widget.setData("AUTOSCALE_DISABLED", true)` the current implementation will break all conversions. One important factor in the CoordinateMapper implementation is to have a consistent translation from and to display coordinates. In the `MultiZoomCoordinateSystemMapper` this is achieved by extracting the Monitor (zoom) out of the coordinate to translate. In `SingleZoomCoordinateMapper`, the implementation was not consistent yet, as e.g. in `SingleZoomCoordinateMapper#getCursorLocation` or `SingleZoomCoordinateMapper#setCursorLocation` `DPIUtil.getDeviceZoom()` is already used.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/SingleZoomCoordinateSystemMapper.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/SingleZoomCoordinateSystemMapper.java
@@ -77,22 +77,22 @@ class SingleZoomCoordinateSystemMapper implements CoordinateSystemMapper {
 
 	@Override
 	public Point translateFromDisplayCoordinates(Point point, int zoom) {
-		return Win32DPIUtils.pixelToPoint(point, zoom);
+		return Win32DPIUtils.pixelToPoint(point, DPIUtil.getDeviceZoom());
 	}
 
 	@Override
 	public Point translateToDisplayCoordinates(Point point, int zoom) {
-		return Win32DPIUtils.pointToPixel(point, zoom);
+		return Win32DPIUtils.pointToPixel(point, DPIUtil.getDeviceZoom());
 	}
 
 	@Override
 	public Rectangle translateFromDisplayCoordinates(Rectangle rect, int zoom) {
-		return Win32DPIUtils.pixelToPoint(rect, zoom);
+		return Win32DPIUtils.pixelToPoint(rect, DPIUtil.getDeviceZoom());
 	}
 
 	@Override
 	public Rectangle translateToDisplayCoordinates(Rectangle rect, int zoom) {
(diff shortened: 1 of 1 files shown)


HeikoKlare (commented):
The change definitely makes sense. As a follow up, we should remove the obsolete `zoom` parameter of those methods.

I wonder if the `mapMonitorBounds()` method is not affected in the same way?
One consumer is the `Display#getMonitor()` method:
https://github.com/eclipse-platform/eclipse.platform.swt/blob/9baef5850786b1e49b4080bb75abd8390aaa99d6/bundles/org.eclipse.swt/Eclipse%20SWT/win32/org/eclipse/swt/widgets/Display.java#L2233
Shouldn't those bounds also be scaled according to the device zoom instead of the zoom of that specific monitor in case of a single zoom mapper? Otherwise with a 100% primary monitor and a 200% secondary monitor, I would expect the bounds of the secondary monitor, scaled with 200% instead of 100%, to be wrong.

akoch-yatta:
> The change definitely makes sense. As a follow up, we should remove the obsolete `zoom` parameter of those methods.
> 
> I wonder if the `mapMonitorBounds()` method is not affected in the same way? One consumer is the `Display#getMonitor()` method:
> 
> https://github.com/eclipse-platform/eclipse.platform.swt/blob/9baef5850786b1e49b4080bb75abd8390aaa99d6/bundles/org.eclipse.swt/Eclipse%20SWT/win32/org/eclipse/swt/widgets/Display.java#L2233
> 
> 
> Shouldn't those bounds also be scaled according to the device zoom instead of the zoom of that specific monitor in case of a single zoom mapper? Otherwise with a 100% primary monitor and a 200% secondary monitor, I would expect the bounds of the secondary monitor, scaled with 200% instead of 100%, to be wrong.

Yes, I thought about that, but wanted to come up with a test for that. But I think we will need to adapt those as well.

HeikoKlare:
> Yes, I thought about that, but wanted to come up with a test for that. But I think we will need to adapt those as well.

Alright, we can also keep that for a separate PR and merge this one as is.

akoch-yatta:
I would propose to do that separately and merge this PR as is. I will test the original behavior to add a proper test set for mapMonitorBounds to ensure the correct behavior.

Issue #2374, opened 2025-08-05
> [eclipse-platform/eclipse.platform.swt#2372](https://github.com/eclipse-platform/eclipse.platform.swt/pull/2372) - Gtk implementation


fedejeanne:
_Originally posted by @fedejeanne in https://github.com/eclipse-platform/eclipse.platform.ui/issues/3138#issuecomment-3153996368_

It would be nice to have a notification that tells the user that the _High Contrast_ mode has been activated and offers him to restart the application. This already happens on Windows:

<img width="500" height="136" alt="Image" src="https://github.com/user-attachments/assets/edaa16bd-5e97-45a1-8103-43790d48db33" />

Here's the code that shows the dialog:

https://github.com/eclipse-platform/eclipse.platform.ui/blob/4a2727c448db22f5ebf7fad65987b100007de72e/bundles/org.eclipse.ui.ide.application/src/org/eclipse/ui/internal/ide/application/IDEWorkbenchAdvisor.java#L318-L333

akurtakov:
For the record - Runtime theme change doesn't seem to "fire" settings event on Gtk. That would have to be fixed on SWT side.

fedejeanne:
Thank you @akurtakov! You're right, I moved it to SWT.

I'm guessing this is also needed in **Mac**. Maybe someone could quickly test it? @BeckerWdf / @opcoach ?

PR #2380, opened 2025-08-05
[GTK] Fire settings event on Gtk theme change

akurtakov:
Listen for notify::gtk-theme-name signal and fire SWT.Settings even in this case. Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2374


bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
```diff
@@ -1323,7 +1327,11 @@ void createDisplay (DeviceData data) {
 	} else {
 		keymap = GDK.gdk_keymap_get_for_display(display);
 		OS.g_signal_connect (keymap, OS.keys_changed, keysChangedProc, 0);
-	}
+		}
```
Mailaender, line 1330: wrong indention

Issue #2385, opened 2025-08-06
Font size of Line number ruler is different when using swt.autoScale=int200


arunjose696:
After the following PR to use `ScalingSWTFontRegistry`, a regression was introduced where the font sizes of the line number ruler (for `int200`) differ:

- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2212

Previously, in `LegacySWTFontRegistry` (`DefaultSWTFontRegistry`), fonts were always created using the constructor `Font(Device, FontData)`. However, in `ScalingSWTFontRegistry`, fonts are now created using the `Font(Device, FontData, zoom)` constructor.

When fonts are created using `Font(Device, FontData)`, they implicitly use `DPIUtil.getNativeDeviceZoom()` for scaling. On the other hand, fonts created via the new constructor use the zoom value explicitly passed to `SWTFontProvider.getFontHandle(font, zoom)`.

For example, with `LegacySWTFontRegistry`, the `LineNumberRuler` had the same font size as the text editor.  
However, with `ScalingSWTFontRegistry`, the sizes may now differ due to the zoom being explicitly passed on font creation.

|(LegacySWTFontRegistry) |  (ScalingSWTFontRegistry) |
|----------------------------------|-----------------------------------|
| <img src="https://github.com/user-attachments/assets/aedfe412-bec5-482d-a938-a14c526278d8" width="300"/> | <img src="https://github.com/user-attachments/assets/adf64ed3-b93a-4291-8feb-83c74eedb7ae" width="300"/> |

<img  alt="Image" src="https://github.com/user-attachments/assets/adf64ed3-b93a-4291-8feb-83c74eedb7ae" />

<img  alt="Image" src="https://github.com/user-attachments/assets/aedfe412-bec5-482d-a938-a14c526278d8" />

### Steps to reproduce

Start eclipse On monitor with 225% zoom with

```
-Dswt.autoScale.updateOnRuntime=false
-Dswt.autoScale=int200 
```

PR #2387, opened 2025-08-06
[win32] Fix inconsistency when an Image with GC is drawn on a GC

akoch-yatta:
This PR adapts the logic when a bitmap or icon is drawn with the GC in the windows implementation. As of now, the autoscale zoom was used to identify the image handle to use. If the image is itself drawn with a ImageGCDrawer, information about the original native zoom was lost, which led to inconsistencies in fonts depending on the auto scale mode. This PR forwards the native zoom in this case to enable the Image to initialize the GC accordingly. This was e.g. an issue with auto scale mode integer200 as described in #2385.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GCData.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GCData.java
@@ -48,6 +48,7 @@ public final class GCData {
 	public float lineMiterLimit = 10;
 	public int alpha = 0xFF;
 	public int nativeZoom;
+	int imageZoom;
 
 	public Image image;
 	public PAINTSTRUCT ps;
@@ -67,6 +68,7 @@ public final class GCData {
 		originalData.font = font;
 		originalData.nativeZoom = nativeZoom;
 		originalData.image = image;
+		originalData.imageZoom = imageZoom;
 		originalData.ps = ps;
 		originalData.layout = layout;
 		originalData.hwnd = hwnd;
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
@@ -809,10 +809,15 @@ private ImageHandle getImageMetadata(ZoomContext zoomContext) {
  * @noreference This method is not intended to be referenced by clients.
  */
 public static long win32_getHandle (Image image, int zoom) {
-	if (image.isDisposed()) {
+	return image.getHandle(zoom, zoom);
+}
+
+long getHandle (int targetZoom, int nativeZoom) {
+	if (isDisposed()) {
 		return 0L;
 	}
-	return image.getImageMetadata(new ZoomContext(zoom)).handle;
+	ZoomContext zoomContext = imageProvider.getFittingZoomContext(targetZoom, nativeZoom);
+	return getImageMetadata(zoomContext).handle;
 }
 
 /**
@@ -1752,6 +1757,7 @@ private long configureGC(GCData data, ZoomContext zoomContext) {
 		}
 		data.device = device;
 		data.nativeZoom = zoomContext.nativeZoom();
+		data.imageZoom = zoomContext.targetZoom();
 		data.image = this;
 		data.font = SWTFontProvider.getSystemFont(device, zoomContext.nativeZoom());
 	}
@@ -1912,6 +1918,10 @@ private record ZoomContext(int targetZoom, int nativeZoom) {
 private abstract class AbstractImageProviderWrapper {
 
 	protected abstract Rectangle getBounds(int zoom);
+
+	protected ZoomContext getFittingZoomContext(int targetZoom, int nativeZoom) {
+		return new ZoomContext(targetZoom);
+	}
 
 	protected long configureGCData(GCData data) {
 		return configureGC(data, new ZoomContext(100));
@@ -2139,13 +2149,21 @@ private class PlainImageProviderWrapper extends AbstractImageProviderWrapper {
 	}
 
 	@Override
+	protected ZoomContext getFittingZoomContext(int targetZoom, int nativeZoom) {
+		if (memGC != null) {
+			return new ZoomContext(targetZoom, nativeZoom);
+		}
+		return super.getFittingZoomContext(targetZoom, nativeZoom);
+	}
+
+	@Override
 	public Collection<Integer> getPreservedZoomLevels() {
 		return Collections.singleton(baseZoom);
 	}
 
 	@Override
 	protected long configureGCData(GCData data) {
-		return configureGC(data, new ZoomContext(DPIUtil.getDeviceZoom()));
+		return configureGC(data, new ZoomContext(DPIUtil.getDeviceZoom(), DPIUtil.getNativeDeviceZoom()));
 	}
 
 	@Override
@@ -2575,6 +2593,11 @@ private class ImageGcDrawerWrapper extends DynamicImageProviderWrapper {
 		this.drawer = imageGcDrawer;
 		this.width = width;
 		this.height = height;
+	}
+
+	@Override
+	protected ZoomContext getFittingZoomContext(int targetZoom, int nativeZoom) {
+		return new ZoomContext(targetZoom, nativeZoom);
 	}
 
 	@Override
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
@@ -1253,7 +1253,7 @@ private void drawImage(Image srcImage, int srcX, int srcY, int srcWidth, int src
 		}
 		return;
 	}
-	long imageHandle = Image.win32_getHandle(srcImage, imageZoom);
+	long imageHandle = srcImage.getHandle(imageZoom, data.nativeZoom);
 	switch (srcImage.type) {
 		case SWT.BITMAP:
 			drawBitmap(srcImage, imageHandle, srcX, srcY, srcWidth, srcHeight, destX, destY, destWidth, destHeight, simple);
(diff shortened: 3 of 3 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2000,32 +2022,32 @@ private abstract class ImageFromImageDataProviderWrapper extends AbstractImagePr
...
 		};
-		return cachedImageData.computeIfAbsent(zoom, imageDataRetrieval);
+		return cachedImageData.computeIfAbsent(zoomContext.targetZoom, imageDataRetrieval);
 	}
 
 	@Override
-	protected ImageHandle newImageHandle(int zoom) {
+	protected ImageHandle newImageHandle(ZoomContext zoom) {
 		ImageData cachedData = cachedImageData.remove(zoom);
 		if (cachedData != null) {
```
arunjose696: Shouldnt the remove be called with` cachedImageData.remove(zoom.targetZoom)`

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2000,32 +2022,32 @@ private abstract class ImageFromImageDataProviderWrapper extends AbstractImagePr
...
 			handle.destroy();
 			return data;
 		};
-		return cachedImageData.computeIfAbsent(zoom, imageDataRetrieval);
+		return cachedImageData.computeIfAbsent(zoomContext.targetZoom, imageDataRetrieval);
 	}
 
 	@Override
-	protected ImageHandle newImageHandle(int zoom) {
+	protected ImageHandle newImageHandle(ZoomContext zoom) {
```
arunjose696:
```suggestion
	protected ImageHandle newImageHandle(ZoomContext zoomContext) {
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2607,10 +2634,10 @@ protected ImageHandle newImageHandle(int zoom) {
 		} else {
 			image = new Image(device, width, height);
 		}
-		GC gc = new GC(new DrawableWrapper(image, zoom), gcStyle);
+		GC gc = new GC(new DrawableWrapper(image, nativeZoom), gcStyle);
 		try {
 			drawer.drawOn(gc, width, height);
-			ImageData imageData = image.getImageMetadata(zoom).getImageData();
+			ImageData imageData = image.getImageData(nativeZoom);
```
arunjose696: Why was this changed, and other calls of `image.getImageMetadata(zoom).getImageData() `kept as is

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: Yes, good catch

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: Shortcut to the same logic. Main reason is that I would need to replace the call with` image.getImageMetadata(new ZoomContext(nativeZoom)).getImageData()` which looks a bit clumsy, so I used the shorter variant

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1913,6 +1921,20 @@ public static Image win32_new(Device device, int type, long handle, int nativeZo
...
+	private ZoomContext(int targetZoom) {
+		this(targetZoom, targetZoom);
+	}
+
+	private ZoomContext(int targetZoom, int nativeZoom) {
+		this.targetZoom = targetZoom;
+		this.nativeZoom = nativeZoom;
+	}
+}
+
```
fedejeanne: You could turn this into a record and help the C2 compiler optimize it (no guarantees though)
```suggestion
private static record ZoomContext(int targetZoom, int nativeZoom) {
	ZoomContext(int zoom) {
		this(zoom, zoom);
	}
}
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
fedejeanne: I noticed that `getImageData(int)` checks if the image is disposed but `getImageMetadata(ZoomContext)` doesn't. Couldn't that be a source of issues?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2000,32 +2022,32 @@ private abstract class ImageFromImageDataProviderWrapper extends AbstractImagePr
...
+			return newImageHandle(cachedData, zoomContext);
 		}
-		return initializeHandleFromSource(zoom);
+		return initializeHandleFromSource(zoomContext);
 	}
 
-	private ImageHandle initializeHandleFromSource(int zoom) {
-		ElementAtZoom<ImageData> imageDataAtZoom = loadImageData(zoom);
-		ImageData imageData = DPIUtil.scaleImageData(device, imageDataAtZoom.element(), zoom, imageDataAtZoom.zoom());
+	private ImageHandle initializeHandleFromSource(ZoomContext zoom) {
```
fedejeanne: NIT: preserve the naming standard
```suggestion
	private ImageHandle initializeHandleFromSource(ZoomContext zoomContext) {
```

akoch-yatta: adapted it

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: adapted it

akoch-yatta:
> I noticed that getImageData(int) checks if the image is disposed but getImageMetadata(ZoomContext) doesn't. Couldn't that be a source of issues?

@fedejeanne The image is created few lines above and it will be disposed immediately after it anyway, but I don't see a change this can be disposed before.

fedejeanne:
In the call that I highlighted everything is fine because `image.getImageData(nativeZoom)` does check for dispose and, as you say, the image is disposed right away 👍 

I was thinking more on the lines of: would it be worth playing it safe and checking for disposal when calling `getImageMetadata(ZoomContext)`?

At the moment, all calls seem safe because the (direct and indirect) **callers** check for disposal:

image

But I was just thinking: is checking for disposal a responsibility of the caller or should the method itself do it?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
fedejeanne: It's been reverted.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GCData.java
```diff
@@ -48,6 +48,7 @@ public final class GCData {
 	public float lineMiterLimit = 10;
 	public int alpha = 0xFF;
 	public int nativeZoom;
+	public int imageZoom;
```
HeikoKlare: I am not sure why all the other fields are public, but couldn't we make at least this one package private to further reduce risk of potential faulty usages?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1913,6 +1924,10 @@ private abstract class AbstractImageProviderWrapper {
 
 	protected abstract Rectangle getBounds(int zoom);
 
+	protected boolean supportComplexZoom() {
```
HeikoKlare: Can we find a more self-explaining name than "complex" for that kind of zoom? I think that one would need to look into the code to what's meant by "complex". Maybe we could even defer the logic of providing a proper `ZoomContext` to the provider, i.e., instead of an `supportComplexZoom()` we may have somelike like a `ZoomContext getFittingZoomContext(target, native)`.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GCData.java
akoch-yatta: sure, done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: Yes, that makes sense. The logic was a bit different when I created the supportComplexZoom method, but with the final logic delegating it to the imageProvider makes sense. Adapted that one as well

HeikoKlare (approved):
I have tested the proposed change with several different setups including combinations of
- two monitors (175% and 100%), each of them once used as primary
- monitor-specific scaling enabled/disabled
- `swt.autoScale` set to a fixed value (150) vs. not custom specified

and particularly with StyledText (incl. LineNumberRuler) and GEF diagrams.

I did not find any regressions with respect to existing behavior (and with respect to 2025-06) but found the two mentioned issues to be successfully resolved:
- https://github.com/eclipse-platform/eclipse.platform.swt/issues/2311
- https://github.com/eclipse-platform/eclipse.platform.swt/issues/2385

HeikoKlare:
ftr: this also contributes to https://github.com/eclipse-platform/eclipse.platform.ui/issues/3061

PR #2390, opened 2025-08-07
Refactor cursor handle creation using CursorHandleProvider abstraction

ShahzaibIbrahim:
Replaces ad-hoc handle creation logic in `win32_getHandle()` with a new `CursorHandleProvider` abstract class. This class has three concrete implementations to cover the different input sources used to create a cursor:

- **StyleCursorHandleProvider**: uses predefined cursor style ints.
- **ImageDataProviderCursorHandleProvider**: uses ImageDataProvider for source.
- **ImageDataCursorHandleProvider**: uses ImageData and mask directly.

This eliminates the need for checking which cursor fields are set, and removes the problem of having multiple unused fields in Cursor.

The abstract class defines a final helper method to create handles from scaled image data, centralizing logic and reducing duplication.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
@@ -68,24 +68,8 @@ public final class Cursor extends Resource {
 
 	private HashMap<Integer, Long> zoomLevelToHandle = new HashMap<>();
 
+	private final CursorHandleProvider cursorHandleProvider;
 	boolean isIcon;
-	private final ImageDataProvider imageDataProvider;
-	private final ImageData source;
-	private final ImageData mask;
-	private final int hotspotX;
-	private final int hotspotY;
-/**
- * Prevents uninitialized instances from being created outside the package.
- */
-Cursor(Device device) {
-	super(device);
-	this.source = null;
-	this.imageDataProvider = null;
-	this.mask = null;
-	this.hotspotX = -1;
-	this.hotspotY = -1;
-	this.device.registerResourceWithZoomSupport(this);
-}
 
 /**
  * Constructs a new cursor given a device and a style
@@ -134,42 +118,11 @@ Cursor(Device device) {
  * @see #dispose()
  */
 public Cursor(Device device, int style) {
-	this(device);
-	this.handle = setupCursorFromStyle(style);
+	super(device);
+	this.cursorHandleProvider = new StyleCursorHandleProvider(style);
+	this.handle = this.cursorHandleProvider.createHandle(this, DEFAULT_ZOOM);
 	init();
-}
-
-private static long setupCursorFromStyle(int style) {
-	long lpCursorName = 0;
-	switch (style) {
-		case SWT.CURSOR_HAND: 		lpCursorName = OS.IDC_HAND; break;
-		case SWT.CURSOR_ARROW: 		lpCursorName = OS.IDC_ARROW; break;
-		case SWT.CURSOR_WAIT: 		lpCursorName = OS.IDC_WAIT; break;
-		case SWT.CURSOR_CROSS: 		lpCursorName = OS.IDC_CROSS; break;
-		case SWT.CURSOR_APPSTARTING: 	lpCursorName = OS.IDC_APPSTARTING; break;
-		case SWT.CURSOR_HELP: 		lpCursorName = OS.IDC_HELP; break;
-		case SWT.CURSOR_SIZEALL: 	lpCursorName = OS.IDC_SIZEALL; break;
-		case SWT.CURSOR_SIZENESW: 	lpCursorName = OS.IDC_SIZENESW; break;
-		case SWT.CURSOR_SIZENS: 	lpCursorName = OS.IDC_SIZENS; break;
-		case SWT.CURSOR_SIZENWSE: 	lpCursorName = OS.IDC_SIZENWSE; break;
-		case SWT.CURSOR_SIZEWE: 	lpCursorName = OS.IDC_SIZEWE; break;
-		case SWT.CURSOR_SIZEN: 		lpCursorName = OS.IDC_SIZENS; break;
-		case SWT.CURSOR_SIZES: 		lpCursorName = OS.IDC_SIZENS; break;
-		case SWT.CURSOR_SIZEE: 		lpCursorName = OS.IDC_SIZEWE; break;
-		case SWT.CURSOR_SIZEW: 		lpCursorName = OS.IDC_SIZEWE; break;
-		case SWT.CURSOR_SIZENE: 	lpCursorName = OS.IDC_SIZENESW; break;
-		case SWT.CURSOR_SIZESE: 	lpCursorName = OS.IDC_SIZENWSE; break;
-		case SWT.CURSOR_SIZESW: 	lpCursorName = OS.IDC_SIZENESW; break;
-		case SWT.CURSOR_SIZENW: 	lpCursorName = OS.IDC_SIZENWSE; break;
-		case SWT.CURSOR_UPARROW: 	lpCursorName = OS.IDC_UPARROW; break;
-		case SWT.CURSOR_IBEAM: 		lpCursorName = OS.IDC_IBEAM; break;
-		case SWT.CURSOR_NO: 		lpCursorName = OS.IDC_NO; break;
-		default:
-			SWT.error(SWT.ERROR_INVALID_ARGUMENT);
-	}
-	long handle = OS.LoadCursor(0, lpCursorName);
-	if (handle == 0) SWT.error(SWT.ERROR_NO_HANDLES);
-	return handle;
+	this.device.registerResourceWithZoomSupport(this);
 }
 
 /**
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
```diff
@@ -578,4 +395,272 @@ void destroyHandlesExcept(Set<Integer> zoomLevels) {
 	});
 }
 
+private abstract class CursorHandleProvider {
+	protected abstract long createHandle(Cursor cursor, int zoom);
+
+	protected final long createHandleFromImageData(Cursor cursor, int zoom, ImageData source, ImageData mask, int hotSpotX, int hotSpotY) {
```
HeikoKlare: All the methods in `CursorHandleProvider` don't access any state, so I would keep them as static private methods of the `Cursor` class. `CursorHandleProvider` may then be a proper interface instead of an abstract class from which you inherent effectively side-effect free, static functions.

```suggestion
	private static final long createHandleFromImageData(Cursor cursor, int zoom, ImageData source, ImageData mask, int hotSpotX, int hotSpotY) {
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
```diff
@@ -134,42 +118,11 @@ public final class Cursor extends Resource {
...
-		case SWT.CURSOR_UPARROW: 	lpCursorName = OS.IDC_UPARROW; break;
-		case SWT.CURSOR_IBEAM: 		lpCursorName = OS.IDC_IBEAM; break;
-		case SWT.CURSOR_NO: 		lpCursorName = OS.IDC_NO; break;
-		default:
-			SWT.error(SWT.ERROR_INVALID_ARGUMENT);
-	}
-	long handle = OS.LoadCursor(0, lpCursorName);
-	if (handle == 0) SWT.error(SWT.ERROR_NO_HANDLES);
-	return handle;
+	this.device.registerResourceWithZoomSupport(this);
```
HeikoKlare, line 125: This seems to be wrong. We did not register for style-based cursors before (as no new OS handle will be created for them to track).

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
```diff
@@ -578,4 +395,272 @@ void destroyHandlesExcept(Set<Integer> zoomLevels) {
...
+		this.hotspotX = hotspotX;
+		this.hotspotY = hotspotY;
+	}
+
+	@Override
+	public long createHandle(Cursor cursor, int zoom) {
+		Image tempImage = new Image(cursor.device, this.provider);
+		ImageData source = tempImage.getImageData(zoom);
+		tempImage.dispose();
+		long handle = createHandleFromImageData(cursor, zoom, source, null, this.hotspotX, this.hotspotY);
```
HeikoKlare: In the existing implementation, for zoom==100 we directly retrieved the image data from the provider instead of creating a temp image for it. Under specific circumstances, the results may differ. So to keep this a pure refactoring, we should stick to the existing logic and do the simplification (always creating a temp image) as a follow up for 4.38 M1 where have more time to investigate whether that leads to unexpected side effects.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
HeikoKlare: It's hard to understand that the passed `source` is already scaled while the `mask` is not. It would be clearer if either this method would scale the image data (being basically a no-op when the data is already scaled such as in case of the ImageDataProvider`) or, probably better, that this method is inlined to the two usages and reduces to exactly what is necessary there. The latter would probably be less confusing.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
ShahzaibIbrahim: We did before by calling this(device) and there it was registered.

HeikoKlare: Can you just create a simple separate PR that removes the obsolete constructor? It seem unrelated to the changes in this PR.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
```diff
@@ -578,4 +395,274 @@ void destroyHandlesExcept(Set<Integer> zoomLevels) {
...
+	if (handle == 0)
+		SWT.error(SWT.ERROR_NO_HANDLES);
+	return handle;
+}
+
+private interface CursorHandleProvider {
+	long createHandle(Cursor cursor, int zoom);
+}
+
+private class StyleCursorHandleProvider implements CursorHandleProvider {
```
HeikoKlare: Please make these classes static (see https://github.com/eclipse-platform/eclipse.platform.swt/pull/2390#discussion_r2260224846). Otherwise one may accidentally access the outer class instance which should not be necessary and thus should be prevented.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
```diff
@@ -578,4 +395,274 @@ void destroyHandlesExcept(Set<Integer> zoomLevels) {
 	});
 }
 
+private static final long createHandleFromImageData(Cursor cursor, int zoom, ImageData source, ImageData mask,
+		int hotspotX, int hotspotY) {
+	return cursor.isIcon ? setupCursorFromImageData(cursor.device, source, hotspotX, hotspotY)
+			: setupCursorFromImageData(source, mask, hotspotX, hotspotY);
+}
+
+private static final long setupCursorFromImageData(Device device, ImageData source, int hotspotX, int hotspotY) {
```
HeikoKlare: Is there a specific reason to move the static methods at this exact place? They are now placed in the file even before their usages. In case we want to move them, they should at least be at some reasonable place (such as right after their first usage for easy reading). But I would actually be in favor of keeping them at their original places for now and move them in a follow-up step (maybe in a separate commit of the same PR) to make the actual change (introducing the provider classes) more easy to understand (both for the review now and also for people taking a look at the commit at a later point in time).

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
```diff
@@ -578,4 +395,274 @@ void destroyHandlesExcept(Set<Integer> zoomLevels) {
...
+	if (source == null)
+		SWT.error(SWT.ERROR_NULL_ARGUMENT);
```
HeikoKlare: If the code is reformatted anyway, please put expression blocks into braces.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
```diff
@@ -578,4 +395,274 @@ void destroyHandlesExcept(Set<Integer> zoomLevels) {
...
+		this.source = source;
+		this.mask = mask;
+		this.hotspotX = hotspotX;
+		this.hotspotY = hotspotY;
+	}
+
+	@Override
+	public long createHandle(Cursor cursor, int zoom) {
+		ImageData scaledSource = DPIUtil.scaleImageData(cursor.device, this.source, zoom, DEFAULT_ZOOM);
+		ImageData scaledMask = this.mask!=null? DPIUtil.scaleImageData(cursor.device, mask, zoom, DEFAULT_ZOOM) : mask;
```
```suggestion
		ImageData scaledMask = this.mask!=null? DPIUtil.scaleImageData(cursor.device, mask, zoom, DEFAULT_ZOOM) : null;
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
```diff
@@ -578,4 +395,274 @@ void destroyHandlesExcept(Set<Integer> zoomLevels) {
...
+		int scaledHotspotX = Win32DPIUtils.pointToPixel(hotspotX, zoom);
```
```suggestion
		int hotspotXInPixels = Win32DPIUtils.pointToPixel(hotspotX, zoom);
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
ShahzaibIbrahim: The constructor becomes obsolete as a result of this change (removing all the field). I can create a separate PR but I think it belongs here.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
ShahzaibIbrahim: Moving this method back to original place and unchanged.

HeikoKlare (approved):
The change looks good and works as expected (i.e., does not change behavior) for all scenarios I tested (monitor-specific scaling on/off, autoScale set to a fixed value).

With the latest update, I just inlined the `createHandleFromImageData` method, which took an unnecessary parameter, requires some dummy arguments (e.g., `mask` being null when the cursor image is an icon).

I just found that this change seems to by accident fix a bug: when a Cursor was created with a null `mask`, this mask was tried to be scaled for handles of others zooms before, which should habe failed. Now there is a proper null check that makes the behavior for zoom != 100 conform to the behavior at zoom == 100.

These changes also expose that the `isIcon` field is semantically not tied to the cursor but to the created handles (it's just that all handles of the same cursor are either icons or not). This is a chance for a follow up cleanup: create a `CursorHandle` class that is capable of the disposal appropriate for the type of handle.

PR #2391, opened 2025-08-07
Clarify linear scaling in ImageDataProvider contract

arunjose696:
This PR extends ImageDataProvider contract  to clarify that implementations need to yield linearly scaled image data. 
<s>This commit also extends the strict checks for verifying linear scaling to GTK and Cocoa implementations.</s>

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/ImageDataProvider.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/ImageDataProvider.java
@@ -28,16 +28,31 @@ public interface ImageDataProvider {
 
 	/**
 	 * Returns the image data for the given zoom level.
-	 * <p>
-	 * If no image is available for a particular zoom level, this method should
-	 * return <code>null</code>. For <code>zoom == 100</code>, returning
-	 * <code>null</code> is not allowed, and SWT will throw an exception.
+	 *
+	 * <ul>
+     *   <li>
+     *     If no image is available for a particular zoom level, this method should
+     *     return <code>null</code>. For <code>zoom == 100</code>, returning
+     *     <code>null</code> is not allowed, and SWT will throw an exception.
+     *   </li>
+     *   <li>
+     *     Implementations are expected to return {@link ImageData} that is
+     *     <b>linearly scaled</b> with respect to the zoom level.
+     *     For example, if <code>getImageData(100)</code> returns an image of
+     *     width <code>w</code> and height <code>h</code>, then
+     *     <code>getImageData(200)</code> must return an {@link ImageData}
+     *     of width <code>2 * w</code> and height <code>2 * h</code>,
+     *     if a non-{@code null} result is returned.
+     *   </li>
+     * </ul>
+	 *
 	 *
 	 * @param zoom
 	 *            The zoom level in % of the standard resolution (which is 1
 	 *            physical monitor pixel == 1 SWT logical point). Typically 100,
 	 *            150, or 200.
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Device.java
```diff
@@ -33,6 +33,7 @@
  * @see <a href="http://www.eclipse.org/swt/">Sample code and further information</a>
  */
 public abstract class Device implements Drawable {
+	static boolean strictChecks = System.getProperty("org.eclipse.swt.internal.enableStrictChecks") != null;
```
HeikoKlare: Gtk already has this in the `Display` class, so we should either use or move the property from there.

HeikoKlare (commented):
The change looks good to me. I only have one comment regarding the system property usage on Linux/GTK.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Device.java
arunjose696: I have moved the strict checks to Display class and made them package protected. I think this is better for consistency

HeikoKlare: If I see correctly you moved the constant to the Device class and made it protected (not packaged protected), which effectively makes them part of the API. This is why it would require a `@since` tag. And the build currently also fails because of visibility issues. Since the constant obviously needs to be accessed from multiple packages and since it is not OS-specific anymore, it might make sense to move it to some OS-common class of `org.eclipse.swt.internal` (maybe just a new one if none of the existing fits).

arunjose696: I initially misunderstood the difference between protected and package-protected, as I thought package-private and package-protected were different, but it’s clear now.

By the way, I have moved the strict checks to a new class: `org.eclipse.swt.internal.SWTInternalConstants`.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/SWTInternalConstants.java
```diff
@@ -0,0 +1,20 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Yatta Solutions - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.swt.internal;
+
+public class SWTInternalConstants {
+
+	public static boolean strictChecks = System.getProperty("org.eclipse.swt.internal.enableStrictChecks") != null;
```
HeikoKlare: nit: instead of adding this constant that can be used from everywhere, we might encapsulate it (make it private) and provide methods like `runIfStrictChecksEnabled(Runnable)` and `runWithStrictChecksDisabled(Runnable)` that represent the two use cases we currently have in consumers code.

arunjose696:
> The change looks fine now. I just have one proposal: instead of adding this constant that can be used from everywhere, we might encapsulate it (make it private) and provide methods like `runIfStrictChecksEnabled(Runnable)` and `runWithStrictChecksDisabled(Runnable)` that represent the two use cases we currently have in consumers code. And maybe we could do that in a separate PR (preparatory for the contract improval), so that we have the two changes (improvement of strict checks and contract improvement) separated.

Have made the refactoring of strictchecks to another PR https://github.com/eclipse-platform/eclipse.platform.swt/pull/2410. This PR now only has the contract extension

Issue #611, opened 2023-04-04
Image.equals does not properly test if two images are the same so uses of that in SWT make bad assumptions


merks:
**Describe the bug**

See https://github.com/eclipse-platform/eclipse.platform.ui/pull/682 for background.

Several places in the code, including but probably not limited to these, assume that setting an image to an image that is .equals can be ignored:

- org.eclipse.swt.custom.CTabItem.setImage(Image)
- org.eclipse.swt.widgets.TreeItem.setImage(int, Image)
- TableItem.setImage(int, Image)

But it might well be the case that they are really two different images and the caller might assume that they can dispose the image they thought they replaced.  That is in fact happening in https://github.com/eclipse-platform/eclipse.platform.ui/pull/682

It's not clear if the bug should be fixed by not using .equals at all if if .equals should be fixed to properly test that it's the same image (underlying resource).

I think that if image1.isDisposed() == false and image1.equals(image2) then after calling image1.dispose() it should be the case that image2.isDisposed() == true.

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [x] All OS
- - [ ] Windows
- - [ ] Linux
- - [ ] macOS 

2. Additional OS info (e.g. OS version, Linux Desktop, etc)

3. JRE/JDK version
 
**Version since**

Not sure...

**Workaround (or) Additional context**

Avoid using Image.equals...

laeubi:
Here is a snippet that reproduce the issue with purse SWT (FYI @iloveeclipse )

1. Start the snippet, you see an image and can switch the tabs and then see an image on the active tab
2. Press the button, the image either vanishes instantly or at least if you select a different tab (depends a bit when a redraw is triggered)

```
/*******************************************************************************
 * Copyright (c) 2000, 2013 IBM Corporation and others.
 *
 * This program and the accompanying materials
 * are made available under the terms of the Eclipse Public License 2.0
 * which accompanies this distribution, and is available at
 * https://www.eclipse.org/legal/epl-2.0/
 *
 * SPDX-License-Identifier: EPL-2.0
 *
 * Contributors:
 *     IBM Corporation - initial API and implementation
 *******************************************************************************/
package org.eclipse.swt.snippets;

/*
 * Create a CTabFolder with min and max buttons, as well as close button and
 * image only on selected tab.
 *
 * For a list of all SWT example snippets see
 * http://www.eclipse.org/swt/snippets/
 *
 * @since 3.0
 */
[...]

PR #2392, opened 2025-08-08
 Implements a StyleProcessor class to allow access to a list of set styles for a widget in String format.

neilcabanilla:
This PR allows us to fix issues with the CSS Spy in Eclipse PDE, detailed in [Issue #611](https://github.com/eclipse-pde/eclipse.pde/issues/611) Completed with @WillPeltz

**Changes**
-StyleProcessor class
Parses styles that can be set based on context, checks relevant styles against integer bitmask to see which are actually set, and converts them into String list

-getStyles method in Text and Button classes
Returns a list of styles currently set for a given widget in String format, allowing for style inspection at runtime

-Static function using instance of StyleProcessor in Text and Button classes 
Clearly defines which styles can be set for a widget based on context, in a human-readable way.

**Motivation**

Currently, there is no built-in way to view a widget’s set styles in a human-readable format, leading to projects like Eclipse PDE implementing complex and limited parsers for UI elements like the CSS Spy tool. This PR makes this process much simpler and easier, without requiring much upkeep. It should be quite simple to add support for and update widget classes.

**Notes**
-Since this implementation only focuses on Text and Button classes, other widget classes should be revised incrementally to use the StyleProcessor class.

-Backwards compatible; existing behavior is unchanged. 
Having both getStyles and checkStyle is a bit redundant, so merging them at some point in the future might be a good idea if it can be done without breaking existing behavior too much in downstream projects.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Button.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Button.java
@@ -13,8 +13,6 @@
  *     Conrad Groth - Bug 23837 [FEEP] Button, do not respect foreground and background color on Windows
  *******************************************************************************/
 package org.eclipse.swt.widgets;
-
-
 import java.util.List;
 
 import org.eclipse.swt.*;
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Button.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Button.java
@@ -14,6 +14,7 @@
  *******************************************************************************/
 package org.eclipse.swt.widgets;
 import java.util.List;
+import java.util.*;
 
 import org.eclipse.swt.*;
 import org.eclipse.swt.events.*;
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Button.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Button.java
@@ -262,7 +262,7 @@ static {
 	StyleProcessor processor = new StyleProcessor();
 
 	if (COMMAND_LINK) {
-		processor.oneOf("PUSH, ARROW, CHECK, RADIO, TOGGLE, SWT.COMMAND");
+		processor.someOf("COMMAND");
 	} else {
 		processor.oneOf("PUSH, ARROW, CHECK, RADIO, TOGGLE");
 	}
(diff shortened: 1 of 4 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Button.java
```diff
@@ -278,6 +298,15 @@ static int checkStyle (int style) {
...
 	}
 	return style;
+
+}
+/**
+ * test
+ * @return ArrayList<String> styles
+ * @since 3.131
+ */
+public ArrayList<String> getStyles() {
```
merks: One would never want to surface the List implementation class in any API; elsewhere too.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/StyleProcessor.java
```diff
@@ -0,0 +1,141 @@
+package org.eclipse.swt.widgets;
+
+import java.util.*;
+
+import org.eclipse.swt.*;
+
+public class StyleProcessor {
```
merks: Will this really become public API or can it just be package protected?

merks (commented):
Given the implementation details I see, could this not all be done externally without introducing APIs in SWT?

laeubi:
The recommendation was to access this reflectively and not to make any public API here.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/StyleProcessor.java
neilcabanilla: Hello @merks,

You're absolutely right, we actually intended for StyleProcessor to be package private, so that it can only be accessed within the widgets package. My teammate and I forgot that part, thank you for clarifying.

Thank you so much!
Teammate - @WillPeltz

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Button.java
neilcabanilla: Hello @merks,

Answering to your question...
> Given the implementation details I see, could this not all be done externally without introducing APIs in SWT?

We initially explored other approaches such as implementing it in a utility class or using annotations, however based on the comments  and guidance that we got [Issue #611](https://github.com/eclipse-pde/eclipse.pde/issues/611). We were advised to directly change the SWT repository, specifically the widget classes instead. That being said, we are flexible and happy to revise the implementation.

> One would never want to surface the List implementation class in any API; elsewhere too.

You are right with what you've said. My teammate and I are planning to change this ArrayList<String> to List<String> instead and avoiding making the method public. We've only did it this way for testing purposes to see if it works and get feedback.

Thank you so much!
Teammate - @WillPeltz

WillPeltz:
Hi @laeubi @merks,

Just wanted to follow up on our PR. @neilcabanilla and I took your feedback and made some fixes, so we would really appreciate if you could take another look when you have the chance. Thank you for your time.

Issue #2394, opened 2025-08-08
[Win32] Drag and drop offset on secondary monitor without monitor-specific scaling


HeikoKlare:
**Describe the bug**
When dragging and dropping elements on a secondary monitor when monitor-specific scaling is disabled, there is an offset between the actual and calculated drop position.
This is a result of a previous fix for the same when monitor-specific scaling was enabled, which did not consider the case when monitor-specific scaling is disabled:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2261

**To Reproduce**
- Start an Eclipse SDK with monitor-specific scaling disabled.
- Move the window to a secondary monitor that has a different zoom than the primary monitor
- Open, e.g., the project or package explorer and ensure that it contains some elements
- Drag and move an element inside that explorer

**Expected behavior**
Actual and calculated drop position should be the same on secondary monitors.

**Screenshots**
This is how it looks on 125% secondary monitor (with 100% primary monitor zoom):
![Image](https://github.com/user-attachments/assets/c081ce5d-b5ff-4117-9b1a-5f8db2f9ad52)

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 

**Version since**
I20250707-1800

PR #2395, opened 2025-08-08
[Win32] Fix drag and drop on secondary monitor without monitor-specific scaling

HeikoKlare:
To fix drag and drop behavior on multiple monitors with different zooms when using monitor-specific scaling, a recent change adapted the OS callbacks for drop operations to run with the required DPI awareness (PerMonitorV2) in that case. This change did not consider that without monitor-specific scaling (and particularly when DPI awareness "System" is used), running with that modified DPI awareness is not correct.

This change adapts the drop callback to only adapt the DPI awareness if appropriate (i.e., if monitor-specific scaling is enabled).

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2394

For reproduction see the original issue. Here is how it looks on 125% secondary monitor (with 100% primary monitor):

### Before
dragdrop_nonmonitoraware_issue

### After
dragdrop_nonmonitoraware_fix


arunjose696:
Tested the changes, To reproduce the issue as discussed offline had to do the steps in https://eclipse.dev/eclipse/swt/faq.html#win10dpiawareness the issue is fixed with the changes

The change LGTM

PR #2398, opened 2025-08-08
Convert some tests to JUnit 5

akurtakov:


akurtakov:
Modernize and simplify the code .

PR #2399, opened 2025-08-08
Simplifying dealing with Colors in tests

akurtakov:
There is no need to go through RGB, cache and etc. complications.


akurtakov:
Unstable browser test on windows is unrelated to the change. Merging.

PR #2401, opened 2025-08-10
[GTK] Do not try to remove fixedHandle if there is none

akurtakov:
GtkWidgets can act as containers thus SwtFixed don't have to used always.


akurtakov:
Windows test failure is unrelated to GTK only change. Merging.

PR #2403, opened 2025-08-10
[Win32] Properly encapsulate cursor handle creation and disposal

HeikoKlare:
Currently, the `Cursor`implementation has a design flaw related to the `isIcon` field:
- It is used by handle provider implementations to distinguish ways in which a handle shall be created, which requires access of the provider to data of a cursor and requires some instantiations to pass dummy values (mask==null). 
- It defines how handles for that cursor have to be created and destroyed, without any guarantee that creation and destroyal are implemented in a conforming way.

This change gets rid of the `isIcon` field by two means:
- It introduces a `CursorHandle` class that encapsulates the destroy logic for a handle and assign it upon creation of the handle, such that destroy logic fits to how the handle was created.
- It also introduces an `ImageDataWithMaskHandleProvider` that properly handles the case that a Cursor was created with a mask which was yet only identified by the `isIcon` field of a cursor.

It is split up into to commits addressing each of the above concerns but with the common goal of getting rid of the `isIcon` field.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
@@ -120,7 +120,7 @@ public final class Cursor extends Resource {
 public Cursor(Device device, int style) {
 	super(device);
 	this.cursorHandleProvider = new StyleCursorHandleProvider(style);
-	this.handle = this.cursorHandleProvider.createHandle(this, DEFAULT_ZOOM);
+	this.handle = this.cursorHandleProvider.createHandle(device, DEFAULT_ZOOM);
 	init();
 	this.device.registerResourceWithZoomSupport(this);
 }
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
```diff
@@ -120,7 +119,7 @@ public final class Cursor extends Resource {
 public Cursor(Device device, int style) {
 	super(device);
 	this.cursorHandleProvider = new StyleCursorHandleProvider(style);
-	this.handle = this.cursorHandleProvider.createHandle(this, DEFAULT_ZOOM);
+	this.handle = this.cursorHandleProvider.createHandle(device, DEFAULT_ZOOM).getHandle();
```
arunjose696, line 122: just a Nit: The handle field still stores the raw long, while `zoomLevelToHandle` stores the `CursorHandle`. This  introduces a risk of the two becoming out-of-sync if the logic changes in the future. IMO It might be cleaner to store the `CursorHandle` directly in handle and call `this.handle.getHandle()`  (or rename the method to getNativeHandle() for clarity) wherever the long handle is needed @HeikoKlare

ShahzaibIbrahim:
I have tried with different zoom levels (100/150/175) and changes looks good to me

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
HeikoKlare: It's true that the two values may get out of sync but that's actually not different from the current state. `CursorHandle` is immutable, so it will be as consistent to the current `handle` value as it was before. We need to get rid of the `handle`, but that's something that will be done with the lazy generation of handles implemented by Shahzaib.

PR #2404, opened 2025-08-11
[win32] Introduce ZoomContext in Image

akoch-yatta:
This PR extends the internal zoom used in Image in the windows implementation to a complex object. This is preparatory work to provide a consistent behavior independent of the auto scale mode, e.g. when an Image is drawn with a GC.

The introduced ZoomContext is only used in one place with two different values for targetZoom and nativeZoom, in Image#applyUsingAnyHandle. This is the correct way to fill ZoomContext here, but the zoom does not matter for the usages of this method anyway.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
@@ -803,11 +803,12 @@ private ImageData applyGrayImageData(ImageData data, int pHeight, int pWidth) {
 	return newData;
 }
 
-private ImageHandle getImageMetadata(int zoom) {
-	if (zoomLevelToImageHandle.get(zoom) != null) {
-		return zoomLevelToImageHandle.get(zoom);
+private ImageHandle getImageMetadata(ZoomContext zoomContext) {
+	int targetZoom = zoomContext.targetZoom();
+	if (zoomLevelToImageHandle.get(targetZoom) != null) {
+		return zoomLevelToImageHandle.get(targetZoom);
 	}
-	return imageProvider.newImageHandle(zoom);
+	return imageProvider.newImageHandle(zoomContext);
 }
 
 
@@ -828,10 +829,10 @@ private ImageHandle getImageMetadata(int zoom) {
  * @noreference This method is not intended to be referenced by clients.
  */
 public static long win32_getHandle (Image image, int zoom) {
-	if(image.isDisposed()) {
+	if (image.isDisposed()) {
 		return 0L;
 	}
-	return image.getImageMetadata(zoom).handle;
+	return image.getImageMetadata(new ZoomContext(zoom)).handle;
 }
 
 /**
@@ -1265,7 +1266,7 @@ public ImageData getImageData (int zoom) {
 	if (zoomLevelToImageHandle.containsKey(zoom)) {
 		return zoomLevelToImageHandle.get(zoom).getImageData();
 	}
-	return this.imageProvider.newImageData(zoom);
+	return this.imageProvider.newImageData(new ZoomContext(zoom));
 }
 
 
@@ -1742,7 +1743,7 @@ public long internal_new_GC (GCData data) {
 	return this.imageProvider.configureGCData(data);
 }
 
-private long configureGC(GCData data, int zoom) {
+private long configureGC(GCData data, ZoomContext zoomContext) {
 	if (isDisposed()) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
 	/*
 	* Create a new GC that can draw into the image.
@@ -1753,7 +1754,7 @@ private long configureGC(GCData data, int zoom) {
 	}
 
 	if (Device.strictChecks) {
-		checkImageTypeForValidCustomDrawing(zoom);
+		checkImageTypeForValidCustomDrawing(zoomContext.targetZoom());
 	}
 	/* Create a compatible HDC for the device */
 	long hDC = device.internal_new_GC(null);
@@ -1770,9 +1771,9 @@ private long configureGC(GCData data, int zoom) {
 			data.style |= SWT.LEFT_TO_RIGHT;
 		}
 		data.device = device;
-		data.nativeZoom = zoom;
+		data.nativeZoom = zoomContext.nativeZoom();
 		data.image = this;
-		data.font = SWTFontProvider.getSystemFont(device, zoom);
+		data.font = SWTFontProvider.getSystemFont(device, zoomContext.nativeZoom());
 	}
 	return imageDC;
 }
@@ -1882,7 +1883,7 @@ public String toString () {
 
 <T> T applyUsingAnyHandle(Function<ImageHandle, T> function) {
 	if (zoomLevelToImageHandle.isEmpty()) {
-		ImageHandle temporaryHandle = this.imageProvider.newImageHandle(DPIUtil.getDeviceZoom());
+		ImageHandle temporaryHandle = this.imageProvider.newImageHandle(new ZoomContext(DPIUtil.getDeviceZoom(), DPIUtil.getNativeDeviceZoom()));
 		try {
 			return function.apply(temporaryHandle);
 		} finally {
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1913,42 +1914,49 @@ public static Image win32_new(Device device, int type, long handle, int nativeZo
 	return new Image(device, type, handle, nativeZoom);
 }
 
+private record ZoomContext(int targetZoom, int nativeZoom) {
```
HeikoKlare, line 1925: Even though this is private/internal, it might help to have a short comment that explains the intent of `targetZoom` and `nativeZoom` so that one can more easily understand why either is used.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2607,10 +2619,10 @@ protected ImageHandle newImageHandle(int zoom) {
 		} else {
 			image = new Image(device, width, height);
 		}
-		GC gc = new GC(new DrawableWrapper(image, zoom), gcStyle);
+		GC gc = new GC(new DrawableWrapper(image, zoomContext), gcStyle);
 		try {
 			drawer.drawOn(gc, width, height);
-			ImageData imageData = image.getImageMetadata(zoom).getImageData();
+			ImageData imageData = image.getImageData(zoom);
```
HeikoKlare: Is it correct to use the target zoom instead of retrieving the metadata for the ZoomContext here?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2588,13 +2599,14 @@ protected long configureGCData(GCData data) {
...
+	ImageData newImageData(ZoomContext zoomContext) {
+		return getImageMetadata(zoomContext).getImageData();
 	}
 
 	@Override
-	protected ImageHandle newImageHandle(int zoom) {
-		currentZoom = zoom;
+	protected ImageHandle newImageHandle(ZoomContext zoomContext) {
+		currentZoom = zoomContext;
+		int zoom = zoomContext.targetZoom;
```
HeikoKlare: Maybe we should rename these variables to `targetZoom`, so it's clear which of the two zooms involved it taken.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2272,19 +2282,20 @@ Object getProvider() {
...
 		};
-		return cachedImageData.computeIfAbsent(zoom, imageDataRetrival);
+		return cachedImageData.computeIfAbsent(zoomContext.targetZoom, imageDataRetrival);
 	}
 
 
 	@Override
-	protected ImageHandle newImageHandle(int zoom) {
+	protected ImageHandle newImageHandle(ZoomContext zoomContext) {
+		int zoom = zoomContext.targetZoom;
```
HeikoKlare: Maybe we should rename these variables to `targetZoom`, so it's clear which of the two zooms involved it taken.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2272,19 +2282,20 @@ Object getProvider() {
...
-	ImageData newImageData(int zoom) {
+	ImageData newImageData(ZoomContext zoomContext) {
 		Function<Integer, ImageData> imageDataRetrival = zoomToRetrieve -> {
 			ImageHandle handle = initializeHandleFromSource(zoomToRetrieve);
 			ImageData data = handle.getImageData();
 			handle.destroy();
 			return data;
 		};
-		return cachedImageData.computeIfAbsent(zoom, imageDataRetrival);
+		return cachedImageData.computeIfAbsent(zoomContext.targetZoom, imageDataRetrival);
```
HeikoKlare: Does't the map need to store the data for a `ZoomContext` instead of only the targetZoom?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2159,31 +2167,33 @@ protected Rectangle getBounds(int zoom) {
...
-			return newImageHandle(zoom).getImageData();
+			return newImageHandle(zoomContext).getImageData();
 		}
 		return getScaledImageData(zoom);
 	}
 
 	@Override
-	protected ImageHandle newImageHandle(int zoom) {
+	protected ImageHandle newImageHandle(ZoomContext zoomContext) {
+		int zoom = zoomContext.targetZoom;
```
HeikoKlare: Maybe we should rename these variables to `targetZoom`, so it's clear which of the two zooms involved it taken.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2159,31 +2167,33 @@ protected Rectangle getBounds(int zoom) {
...
 		if (zoomLevelToImageHandle.isEmpty()) {
 			return createBaseHandle(zoom);
 		}
 		if (memGC != null) {
 			GC currentGC = memGC;
 			memGC = null;
 			createHandle(zoom);
-			currentGC.refreshFor(new DrawableWrapper(Image.this, zoom), zoom);
+			currentGC.refreshFor(new DrawableWrapper(Image.this, zoomContext), zoom);
```
HeikoKlare: I wonder if this correct: the `zoom` passed to the GC is used as its `nativeZoom`, so shouldn't that one actually be the `nativeZoom` of the context? Or don't we even need to forward the context to `GC`? Or will that be done in the follow-up change?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2159,31 +2167,33 @@ protected Rectangle getBounds(int zoom) {
 	}
 
 	@Override
-	ImageData newImageData(int zoom) {
+	ImageData newImageData(ZoomContext zoomContext) {
+		int zoom = zoomContext.targetZoom;
```
HeikoKlare: Maybe we should rename these variables to `targetZoom`, so it's clear which of the two zooms involved it taken.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1977,8 +1985,8 @@ protected Rectangle getBounds(int zoom) {
 	}
 
 	@Override
-	ImageData newImageData(int zoom) {
-		return getScaledImageData(zoom);
+	ImageData newImageData(ZoomContext zoomContext) {
+		return getScaledImageData(zoomContext.targetZoom);
```
HeikoKlare: Sometimes the record fields are accessed directly and sometimes via their accessors. Would be good to unify that (in the best case using the accessors).

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: Probably yes, but the three wrappers that use that map will not consider zoomContext.nativeZoom in the upcoming  PR, so I would keep that as is for now

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
HeikoKlare: As we discussed offline: this is only for retrieving the image data at a target zoom and the actual data has been generated with a proper `ZoomContext` based on the `DrawableWrapper` passed to the GC few lines above. So this is correct.

HeikoKlare (approved):
Thank you for the latest updates of the PR!
The change is sound and adds the necessary capabilities for properly distinguishing native and target zoom to the Image class. This change alone (without the follow-up changes) does not have any user-perceivable effect. I tested with different auto-scale settings (monitor-specific on/off, fixed auto-scale) and did not find any changes in behavior.

PR #2407, opened 2025-08-11
[Win32] Remove obsolete zoom parameter for refreshing GC handle

HeikoKlare:
When refreshing a GC with another zoom, that zoom is unnecessarily duplicated in the DrawableWrapper of the Image and as a parameter of the refresh method of the GC. This refresh method than unnecessarily uses that zoom to initialize the GCData's native zoom with it, even though that's overwritten with the proper zoom by initializing the GC with the DrawableWrapper afterwards.

This change removes the zoom duplication by removing the parameter in the refresh method of the GC. It removes the unnecessary assignment of the GCData's native zoom and moves the check for whether a refresh is necessary because of a zoom change from the GC to the image. In addition, it ensures that the memGC of an image is not accidentally set to null in case the memGC's zoom does already fit to the requested zoom.


akoch-yatta (approved):
Getting rid of the probably confusing parameter and inlining the logic into image makes sense. tested it with fixed zoom, integer200 and quarter

PR #2408, opened 2025-08-11
Check if the browser is disposed before scheduling runnable in Edge

fedejeanne:
For the unlikely (but possible) case in which the callback is executed too late and the browser/display has already been disposed.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
@@ -840,7 +840,7 @@ void browserDispose(Event event) {
 			if (inCallback > 0) {
 				ICoreWebView2Controller controller1 = controller;
 				controller.put_IsVisible(false);
-				browser.getDisplay().asyncExec(() -> {
+				asyncExec(() -> {
 					controller1.Close();
 					controller1.Release();
 				});
@@ -1004,7 +1004,7 @@ private String getExposedUrl(String url) {
 }
 
 int handleCloseRequested(long pView, long pArgs) {
-	browser.getDisplay().asyncExec(() -> {
+	asyncExec(() -> {
 		if (browser.isDisposed()) return;
 		WindowEvent event = new WindowEvent(browser);
 		event.display = browser.getDisplay();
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -1169,6 +1169,8 @@ int handleSourceChanged(long pView, long pArgs) {
 }
 
 void sendProgressCompleted() {
+	if (browser.isDisposed()) return;
```
HeikoKlare: Should we do this for other similar callbacks as well, such as `handleDocumentTitleChanged` and others? Or is there a guarantee they will not be called when the browser has been disposed?

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -1187,7 +1189,8 @@ int handleDOMContentLoaded(long pView, long pArgs) {
 	args.get_NavigationId(pNavId);
 	LocationEvent startEvent = navigations.get(pNavId[0]);
 	if (startEvent != null && startEvent.top) {
-		if (lastCustomText != null && getUrl().equals(ABOUT_BLANK)) {
+		lastCustomText = "fede";
+		if (lastCustomText != null /*&& getUrl().equals(ABOUT_BLANK)*/) {
```
HeikoKlare: This is probably the cause for the failing tests 🙂

fedejeanne: 🤦‍♂️ _Urlaubsreif_

Thank you, I removed it.

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
fedejeanne: I funneled all calls to `browser.getDisplay().asyncExec(...)` through a private method.

There is one remaining call that _could_ be of similar nature and may have the same problem: 

https://github.com/eclipse-platform/eclipse.platform.swt/blob/51d6029d93746f34c917ea615c5c3d9898613a3f/bundles/org.eclipse.swt/Eclipse%20SWT%20Browser/win32/org/eclipse/swt/browser/Edge.java#L885

WDYT?

HeikoKlare: Looks good!
That method is only called by the lambda inside that same method, which checks for disposal before calling the method. So in my understanding we cannot run into the problem here.

HeikoKlare (approved):

PR #2410, opened 2025-08-11
Unify strict checks and extend ImageDataProvider checks to GTK and Cocoa

arunjose696:
- Move strict checks into a separate class.
- Extend strict check to verify if imageData is linearly scaled by the imageDataProvider in GTK and COCOA implementations.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/StrictChecks.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/StrictChecks.java
@@ -1,0 +1,40 @@
+/*******************************************************************************
+ * Copyright (c) 2025 Yatta Solutions
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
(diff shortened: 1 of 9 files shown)


bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/StrictChecks.java
```diff
@@ -0,0 +1,35 @@
...
+    public static void runIfStrictChecksEnabled(Runnable runnable) {
+        if (STRICT_CHECKS) {
+            runnable.run();
+        }
+    }
+
+    public static void runWithStrictChecksDisabled(Runnable runnable) {
+        if (!STRICT_CHECKS) {
+            runnable.run();
+        }
```
HeikoKlare: This needs to be something like:
```java
boolean initialStrictChecks = STRICT_CHECKS;
STRICT_CHECKS = false;
try {
   runnable.run();
} finally {
   STRICT_CHECKS = initialStrictChecks;
}
```
Or maybe with some additional temporary disablement flag being set, so that `STRICT_CHECKS` can remain a constant.

arunjose696: Added a temporary flag for disabling strict checks, and let STRICT_CHECKS  as constant

PR #2415, opened 2025-08-13
[win32] ImageList#getImageSize should always return value in points

akoch-yatta:
This PR clarifies the contract of ImageList#getImageSize in the windows implementation to always return a value in points. Historically it returned values in pixel. This worked until recent changes in ImageList to support and handle multiple OS image list handles to support monitor specific scaling. Because of the change the ImageList size in pixels depends on the target zoom the ImageList will be used with. Because of this the method is adapted to return a value in points. All callers of this method are adapted accordingly to scale the value to pixels matching the zoom of the Widget if necessary.

### How to reproduce

This is not testable with a Runtime Workspace and must be setup correctly, but can happen easily in real scenarios.

1. Configure monitor setup as: Primary 150%, secondary 100%
2. **Very Important:** Open the File Explorer and move it to the secondary monitor
3. **Start** the latest [I-Build](https://download.eclipse.org/eclipse/downloads/drops4/I20250812-1800/download.php?dropFile=eclipse-SDK-I20250812-1800-win32-x86_64.zip) **from the File Explorer**
4. Some Images will use the 100% image variant, e.g. the _Select Working Set_ dialog
image

The image is smaller than expected
image

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/ImageList.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/ImageList.java
@@ -359,10 +359,14 @@ private void addPlaceholderImageToImageList(long imageListHandle, int bitmapWidt
 	OS.ReleaseDC(0, hDC);
 }
 
+/**
+ *
+ * {@return size of Images in the ImageList in points}
+ */
 public Point getImageSize() {
 	int [] cx = new int [1], cy = new int [1];
 	OS.ImageList_GetIconSize (handle, cx, cy);
-	return new Point (cx [0], cy [0]);
+	return Win32DPIUtils.pixelToPoint(new Point (cx [0], cy [0]), zoom);
 }
 
 public int indexOf (Image image) {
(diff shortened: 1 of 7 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/ImageList.java
```diff
@@ -359,10 +359,14 @@ private void addPlaceholderImageToImageList(long imageListHandle, int bitmapWidt
 	OS.ReleaseDC(0, hDC);
 }
 
+/**
+ *
+ * @return size of the ImageList in points
```
HeikoKlare: It's not the size of the list but the size of images in that list, isn't it?
You could also use `{@return ...}` to show the statement as both the documentation and the return value.

HeikoKlare (commented):
Change is fully reasonable and it's interesting that this did not produce more problems before. Testing shows that the mentioned issue is fixed.

I only have a minor comment regarding the added Javadoc comment.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/ImageList.java
akoch-yatta: done

fedejeanne (approved):
No regressions observed

Issue #2414, opened 2025-08-13
NullPointerException in Cursor constructor if device is null.


ptziegler:
**Describe the bug**

Trying to create a new Cursor fails with a NullPointerException. Example:

```java
ImageDataProvider provider = ...;
new Cursor(null, provider, 0, 0);
```

**Excpectation**

`Device.getDevice()` should be used, if `device` is null.

This regression has been introduced with e8aadbb8fa15ab03cfe7b972a5567eb2b36394ee. 

https://github.com/eclipse-platform/eclipse.platform.swt/blob/15f6cc4dbf437e1cb6915b2cc79c7e48a4de11b8/bundles/org.eclipse.swt/Eclipse%20SWT/win32/org/eclipse/swt/graphics/Cursor.java#L344

This method should be called with `this.device` instead of the constructor argument.

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS

ptziegler:
Full stacktrace:

```
Caused by: java.lang.NullPointerException: Cannot invoke "org.eclipse.swt.graphics.Device.internal_new_GC(org.eclipse.swt.graphics.GCData)" because "device" is null
	at org.eclipse.swt.graphics.Image.init(Image.java:1550)
	at org.eclipse.swt.graphics.Image.initIcon(Image.java:1617)
	at org.eclipse.swt.graphics.Cursor.setupCursorFromImageData(Cursor.java:292)
	at org.eclipse.swt.graphics.Cursor$ImageDataProviderCursorHandleProvider.createHandle(Cursor.java:662)
	at org.eclipse.swt.graphics.Cursor.<init>(Cursor.java:344)
	at org.eclipse.gmf.runtime.gef.ui.internal.l10n.Cursors.createCursor(Cursors.java:50)
	at org.eclipse.gmf.runtime.gef.ui.internal.l10n.Cursors.<clinit>(Cursors.java:43)
	... 1 more
```

ptziegler:
PR will follow,

HeikoKlare:
Good catch! Thank you for creating a PR. I really don't like this magic "auto-correction" of null devices when creating `Resource`s, but it is like it is.

PR #2418, opened 2025-08-13
Fix NullPointerException when creating cursor with null device #2414

ptziegler:
This is a follow-up to e8aadbb8fa15ab03cfe7b972a5567eb2b36394ee which introduced a regression when working with null devices.

If null is passed as argument, the value of Device.getDevice() is used as a fallback when initializing this field. This field must be used for all other operations in the constructor, rather than the constructor argument.

Closes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2414


HeikoKlare (approved):
Thank you!

PR #2422, opened 2025-08-15
[GTK4] Fix GtkWidget warning when updating shell bounds

ptziegler:
This is a continuation of bd1cdd6ea39b2785a4b28905eedeac28caabcaf0 where a 'GTK_IS_WIDGET (widget)' warning is logged when measuring the height of a NULL titlebar. This issue appears when e.g. calling setBounds() on a Shell.

To reproduce this warning, run e.g. Snippet16.


akurtakov:
I have removed the gtk4 label and added it to the project instead. The gtk4 label causes linux verification gha to run with gtk4 but this still causes crash of the whole build.

akurtakov:
Warning is printed on X11 only, the assumption seems to have been always correct on wayland but x11 wasn't considered at all. Very good catch for the gtk_window_get_titlebar usage!

akurtakov (approved):
Thanks for this one. As it touches GTK 4 code paths I merge it now.

PR #2424, opened 2025-08-16
[GTK4] Fix scrollbar warnings when using Breeze theme

ptziegler:
Trying to open e.g. a shell (or any widget with a scrollbar) produces the following warning:
> GtkGizmo (...) (slider) reported min height -2, but sizes must be >= 0

This is a bug with the KDE theme:
https://bugs.kde.org/show_bug.cgi?id=486766

More specifically, the theme is reporting the wrong size for the scroll-bar (6px, as opposed to 8px). To fix this, adjust the CSS theme to define a minimum size for the slider.

To reproduce, run Snippet1 with GTK4 and the Breeze theme.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT PI/gtk/org/eclipse/swt/internal/gtk/swt_theming_fixes_gtk_4_0_0.css
+++ b/bundles/org.eclipse.swt/Eclipse SWT PI/gtk/org/eclipse/swt/internal/gtk/swt_theming_fixes_gtk_4_0_0.css
@@ -5,3 +5,8 @@ button {
 	min-height: 0px;
 	min-width: 0px;
 }
+
+scrollbar slider {
+	min-height: 8px;
+	min-width: 8px;
+}


akurtakov:
This PR is going totally in the wrong direction IMO. SWT does whole lot of things to allow components to be smaller (down to 0,0) via SwtFixed that setting min size bigger for the sake of one theme feels totally wrong, it might even have unpredictable effects in some places.
I can be convinced to accept a modified version where this brezze specific css is moved to a separate breeze css file and is applied only when running theme is breeze.

ptziegler:
@akurtakov I see your point. Having a dedicated CSS file for Breeze would be possible, but I'd like to avoid such a hard dependency if at all possible. How were theming issues handled in the past?

akurtakov:
Most often it was actual issue in SWT - not taking account of padding, border, etc. and the fix was to make SWT query one more Gtk property and use it in calculations. For actual issues in themes e.g. Oxygen was quite problematic one it ended up with "Bugs that are reproducible only with themes other than Adwaita will be given a lower priority (or may not be fixed at all), compared to bugs which are reproducible on the target environments listed above." in https://eclipse.dev/eclipse/development/plans/eclipse_project_plan_4_37.xml#target_environments or https://github.com/eclipse-equinox/equinox/blob/80c4668d7baeeee85369cb3b671f5b5876e372bd/features/org.eclipse.equinox.executable.feature/library/eclipse.c#L500 .
I don't remember a case where a workaround for broken theme was applied for all.

akurtakov:
Do you plan to do anything here or we should close it?

ptziegler:
> Do you plan to do anything here or we should close it?

I sadly don't have much free time right now to work more on the GTK4 port. I'm closing it for now.

PR #2426, opened 2025-08-18
Use DwmGetWindowAttribute for Shell#getLocation to avoid shadow margins

ShahzaibIbrahim:
`GetWindowRect` returns the bounding rectangle of the entire window including invisible shadows and glass margins. This leads to `Shell#getLocation` reporting coordinates that do not match the actual visible frame location. For example, when positioning a dialog close to the left edge of a monitor, the reported X coordinate can be shifted by ~9 pixels compared to the visible window frame. The same discrepancy appears when opening dialogs under a maximized parent shell.

This patch changes `Shell#getLocation` to use:

    DwmGetWindowAttribute(hwnd, DWMWA_EXTENDED_FRAME_BOUNDS, ...)

This API returns the extended frame bounds rectangle in screen space, which corresponds to the **true visible window frame** and excludes the transparent shadow margins. This produces values that align with what the user sees on screen and avoids subtle layout issues in multi-monitor or mixed-DPI setups.

If DWM is not available (e.g., on older Windows versions or when composition is disabled), the implementation falls back to `GetWindowRect`, preserving backward compatibility.

**Benefits:**
- Correct window coordinates near monitor edges.
- No more shadow-margin offset (~9px) in reported locations.
- Consistent behavior across normal, minimized, and maximized shells.

### Steps to Reproduce

- Run the runtime workspace with monitor specific scaling turned on.
- Move the window to secondary monitor on the right (could be of any zoom)
- Resize the window to smaller size
- Open the **"Open Type"** window using `Ctrl + Shift + T`
- Move the Open Type window towards very left of the screen. 
- Close Open Type window
- Maximize the size of main window. 
- Open the **Open Type** dialog again
- You will see the dialog appearing on the left monitor instead of where it was last closed (on the right monitor)

### Expected result
The sub-shell or dialog should open where it was last closed. There should be no shadow margins that leads to change of coordinates and resulting in window opening in the other monitor.

- [ ] requires https://github.com/eclipse-platform/eclipse.platform.swt/pull/2458

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/FontDialog.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/FontDialog.java
@@ -249,36 +249,7 @@ public FontData open () {
 		if (success) {
 			LOGFONT logFont = new LOGFONT();
 			OS.MoveMemory(logFont, lpLogFont, LOGFONT.sizeof);
-
-			/*
-			 * This will not work on multiple screens or for printing. Should use DC for the
-			 * proper device.
-			 */
-			int logPixelsY = getDPI();
-			int pixels = 0;
-			long hDC = OS.GetDC(0);
-			if (logFont.lfHeight > 0) {
-				/*
-				 * Feature in Windows. If the lfHeight of the LOGFONT structure is positive, the
-				 * lfHeight measures the height of the entire cell, including internal leading,
-				 * in logical units. Since the height of a font in points does not include the
-				 * internal leading, we must subtract the internal leading, which requires a
-				 * TEXTMETRIC, which in turn requires font creation.
-				 */
-				long hFont = OS.CreateFontIndirect(logFont);
-				long oldFont = OS.SelectObject(hDC, hFont);
-				TEXTMETRIC lptm = new TEXTMETRIC();
-				OS.GetTextMetrics(hDC, lptm);
-				OS.SelectObject(hDC, oldFont);
-				OS.DeleteObject(hFont);
-				pixels = logFont.lfHeight - lptm.tmInternalLeading;
-			} else {
-				pixels = -logFont.lfHeight;
-			}
-			OS.ReleaseDC(0, hDC);
-
-			float points = pixels * 72f / logPixelsY;
-			fontData = FontData.win32_new(logFont, points);
+			fontData = FontData.win32_new(logFont, lpcf.iPointSize / 10f);
 			if (effectsVisible) {
 				int red = lpcf.rgbColors & 0xFF;
 				int green = (lpcf.rgbColors >> 8) & 0xFF;


bundles/org.eclipse.swt/Eclipse SWT PI/win32/org/eclipse/swt/internal/win32/OS.java
```diff
@@ -1862,6 +1862,7 @@ public class OS extends C {
 	public static final int WM_XBUTTONDBLCLK = 0x020D;
 	public static final int XBUTTON1 = 0x1;
 	public static final int XBUTTON2 = 0x2;
+	public static final int DWMWA_EXTENDED_FRAME_BOUNDS = 9;
```
akoch-yatta: As the other statics are sorted, I would move that up

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Shell.java
```diff
@@ -1015,9 +1015,13 @@ public int getImeInputMode () {
 
 @Override Point getLocationInPixels () {
 	if (OS.IsIconic (handle)) return super.getLocationInPixels ();
-	RECT rect = new RECT ();
-	OS.GetWindowRect (handle, rect);
-	return new Point (rect.left, rect.top);
+	RECT rect = new RECT();
+	OS.GetWindowRect(handle, rect);
```
akoch-yatta: This call is unnecessary

bundles/org.eclipse.swt/Eclipse SWT PI/win32/org/eclipse/swt/internal/win32/OS.java
```diff
@@ -3022,6 +3023,10 @@ public static int HRESULT_FROM_WIN32(int x) {
 public static final native boolean GetWindowPlacement (long hWnd, WINDOWPLACEMENT lpwndpl);
 /** @param hWnd cast=(HWND) */
 public static final native boolean GetWindowRect (long hWnd, RECT lpRect);
+/**
+ * @param hwnd cast=(HWND)
+ * */
```
akoch-yatta: Nit: The javadocs with one param are usually one-liner here, so I would do it uniformly

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Shell.java
```diff
@@ -1015,9 +1015,12 @@ public int getImeInputMode () {
 
 @Override Point getLocationInPixels () {
 	if (OS.IsIconic (handle)) return super.getLocationInPixels ();
-	RECT rect = new RECT ();
-	OS.GetWindowRect (handle, rect);
-	return new Point (rect.left, rect.top);
+	RECT rect = new RECT();
+	int hr = OS.DwmGetWindowAttribute(handle, OS.DWMWA_EXTENDED_FRAME_BOUNDS, rect, RECT.sizeof);
+	if (hr != 0) {
```
HeikoKlare: To improve comprehensibility:
```suggestion
	if (hr != OS.S_OK) {
```

HeikoKlare (changes requested):
The change looks good when using monitor-specific UI scaling and significantly improves UX in that case: shells properly open on the correct monitor and not on the monitor to the left if the parent is maximized. The given reproducer is a good way to test it.

For testing I have added some logging of the new and old values to compare where one can see that the x values are off by some pixels:
<details>
old: 140, 127; new: 147, 127
old: 88, 67; new: 95, 67
old: -8, -8; new: 0, 0
old: -8, -8; new: 0, 0
old: 44, 52; new: 51, 52
old: 44, 52; new: 51, 52
old: 1501, 1292; new: 1501, 1292
old: 1501, 1292; new: 1501, 1292
old: 37, 52; new: 44, 52
old: 34, 52; new: 41, 52
old: 31, 52; new: 38, 52
old: 28, 52; new: 35, 52
old: 24, 52; new: 31, 52
old: 22, 52; new: 29, 52
old: 18, 53; new: 25, 53
old: 15, 54; new: 22, 54
old: 13, 54; new: 20, 54
old: 10, 54; new: 17, 54
old: 7, 54; new: 14, 54
old: 4, 55; new: 11, 55
old: 3, 55; new: 10, 55
old: 2, 55; new: 9, 55
old: 1, 55; new: 8, 55
old: -1, 55; new: 6, 55
old: -2, 55; new: 5, 55
old: -3, 55; new: 4, 55
old: -4, 55; new: 3, 55
old: -5, 55; new: 2, 55
old: -7, 55; new: 0, 55
old: -8, 54; new: -1, 54
</details>

However, the change is incompatible with monitor-specific scaling being disabled. In that case, the offset between the two values is incorporated when reopening a shell, leading it to slightly move to the right on every reopen. See this capture:
shell_reopen_offset

Some interesting information on the issue in general can also be found here: https://stackoverflow.com/questions/34139450/getwindowrect-returns-a-size-including-invisible-borders

ShahzaibIbrahim:
> However, the change is incompatible with monitor-specific scaling being disabled. In that case, the offset between the two values is incorporated when reopening a shell, leading it to slightly move to the right on every reopen. See this capture:

I tested this and figured out it happens even with monitor-specific scaling **enabled**. The stored coordinates when windows is closed and when it's open are identical. However the move event is occurred after opening the shell and when the getLocation is called again and from OS DwmGetWindowAttribute location is requested for the handle, it returns the +7px which I don't understand why since the position hasn't changed. I am unable to figure out this issue and currently I don't have any solution to prevent this right shifting of the shell when it's reopened. @HeikoKlare

HeikoKlare:
> I tested this and figured out it happens even with monitor-specific scaling **enabled**.

That's interesting. I thought I have tested it and found that to work properly (as the windows use a different logical positioning when running in different DPI awareness), but maybe I did not analyze properly.

What about looking for solutions others may have already found the problem? This seems to be a general "issue" with Win32 API, so many consumers will have already run into the problem (like in the linked Stack Overflow post), so there is a chance to find something in the web (either with or without AI assistance ;-) ).

ShahzaibIbrahim:
> What about looking for solutions others may have already found the problem? This seems to be a general "issue" with Win32 API, so many consumers will have already run into the problem (like in the linked Stack Overflow post), so there is a chance to find something in the web (either with or without AI assistance ;-) ).

I have found out the problem with DWM is that when we use DwmGetWindowAttribute in getLocationInPixels(), but when setting the location (setLocationInPixels()), it calls SetWindowPos directly, which expects the top-left of the window, not the extended frame. The mismatch between the extended frame and the actual window area causes Windows to move the window, resulting in the observed +7px shift and a WM_MOVE event.

I have came up with the solution to original problem and that was "When the child shell is too close to left monitor, the monitor bounds are saved with shadow resulting in shell opening in the left monitor". My solution is why not while saving the bounds for persistence for reopening, we store the monitor bounds of the shell as well and during opening of the shell when we read the persisted bounds we check if the bounds are withing the saved monitor coordinate. if not we update the coordinates as per monitor which should fix it instead of introducing DwmGetWindowAttribute which brings more problems with it. I have opened up the PR and you can see the fix there -> https://github.com/eclipse-platform/eclipse.platform.ui/pull/3297

PR #2427, opened 2025-08-18
[Win32] Ensure Image#getImageData() returns no internal data

HeikoKlare:
This fixes a regression from https://github.com/eclipse-platform/eclipse.platform.swt/pull/1989 as reported by @sratz: https://github.com/eclipse-platform/eclipse.platform.swt/pull/1989#issuecomment-3195890789

A caching mechanism for ImageData introduced to some provider implementations inside the Image class leads to Image#getImageData() returning an ImageData object, which if modified affects the underlying Image. This contradicts the contract of that method, which states that changes to that object will not affect the Image.

This change ensures that Image#getImagetData() only returns objects that are not used internal of the Image instance. To this end, clones of the ImageData are created where necessary.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
@@ -2017,7 +2017,7 @@ private abstract class ImageFromImageDataProviderWrapper extends AbstractImagePr
 			handle.destroy();
 			return data;
 		};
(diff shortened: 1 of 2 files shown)


HeikoKlare:
@ShahzaibIbrahim I adapted `test_imageDataIsCached` but I don't see that the other tests you mention are affected as well. Can you elaborate on them?

ShahzaibIbrahim:
@HeikoKlare I tried with your latest changes and they are not failing anymore. I cannot mark my review stale for some reason.

sratz:
Thanks @HeikoKlare! I have also verified the fix in our product.

PR #2429, opened 2025-08-19
Scale Tree.INSET by zoom level instead of using fixed pixels

ShahzaibIbrahim:
The Tree.INSET constant defines the spacing between an item's image and text. Previously, this was a fixed value of 3px, which looks correct at 100% zoom but causes the image and text to appear cramped on higher-DPI displays.

This change defines INSET in points and converts it to pixels based on the current zoom level (e.g. 6px at 200%), ensuring consistent spacing across different monitor scales.

### Before Fix
image

### After Fix
image

As we can see now the space between text and image in Tree is zoom relative.


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/TreeItem.java
```diff
@@ -918,9 +918,9 @@ public Rectangle getTextBounds (int index) {
 Rectangle getTextBoundsInPixels (int index) {
 	if (!parent.checkData (this, true)) error (SWT.ERROR_WIDGET_DISPOSED);
 	RECT rect = getBounds (index, true, false, true);
-	if (index == 0) rect.left += Tree.INSET - 1;
+	if (index == 0) rect.left += Win32DPIUtils.pointToPixel(Tree.INSET - 1, getZoom());
 	rect.left = Math.min (rect.left, rect.right);
-	rect.right = rect.right - Tree.INSET + 1; // Add 1 px margin to avoid truncation of text seen with "Segoe UI" font
+	rect.right = rect.right - Win32DPIUtils.pointToPixel(Tree.INSET + 1, getZoom()); // Add 1 px margin to avoid truncation of text seen with "Segoe UI" font
```
arunjose696: The comment mentions to add 1 pixel margin, with the change the comment no longer makes sense, or we should add the the pixel after the pointToPixel call

arunjose696:
I tested the changes and can observe that the inset adjusts as mentioned in the issue. However, at 350% monitor zoom, the last character sometimes gets cut off.

For example, when I right-click on the JRE System Library and choose Show in Project Explorer, the last closing bracket is not fully visible with the current changes.( 350% zoom)

image

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/TreeItem.java
ShahzaibIbrahim: This comment is valid to 1px only in non-monitor specific scaling scenario. From now on we always consider these "magic values" in points and scale it up as per zoom level. That's why it was added to the Tree Inset when it's being scaled.

ShahzaibIbrahim: The comment could be changed to `1 pt` instead of `1 px`.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/TreeItem.java
```diff
@@ -918,9 +918,8 @@ public Rectangle getTextBounds (int index) {
 Rectangle getTextBoundsInPixels (int index) {
 	if (!parent.checkData (this, true)) error (SWT.ERROR_WIDGET_DISPOSED);
 	RECT rect = getBounds (index, true, false, true);
-	if (index == 0) rect.left += Tree.INSET - 1;
```
arunjose696, line 921: why is this inset no longer required? I think see at 100% zoom there is a difference in the inset  for the first tree item with powertoys.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/TreeItem.java
```diff
@@ -918,9 +918,8 @@ public Rectangle getTextBounds (int index) {
...
 	rect.left = Math.min (rect.left, rect.right);
-	rect.right = rect.right - Tree.INSET + 1; // Add 1 px margin to avoid truncation of text seen with "Segoe UI" font
```
arunjose696, line 923: Previously the inset was subtracted and now it is added any reason why this was changed?

arunjose696:
The issue I mentioned is resolved, and testing shows everything is working fine. I just have two questions about the change for clarification.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/TreeItem.java
ShahzaibIbrahim: There's no detail on why it was added here in the first place. There are a lot of calculations that are done which doesn't make sense or would make sense previously (when we used to have only one 100% monitor). With current state you will see the inset being consistent as per zoom level and we add the inset for  all tree items not just the first one. 

My proposal is that we try to find any issue with the current proposal in any of the zoom level and if there are not, I would like to keep it like this.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/TreeItem.java
ShahzaibIbrahim: This is extension to converting this INSET value into points and because of that the calculations have changed. Again, I am not 100% sure why there was a need to subtract INSET here for rect.right and it might have worked when INSET was in pixels (fixed value) but now we adjust the value by converting it into pixel which will cut off the text if we still subtract it.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/TreeItem.java
arunjose696: On testing I dont see any issues with the current state at different zooms, so I think the change does no harm.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/TreeItem.java
arunjose696: I have also tested around with the implementation and can see there are no issues observable, so I feel the change seems okay to me

arunjose696:
I have tested the changes at my end at 100,200 and 350 zooms and everything seems to work fine, so approving this.

arunjose696 (approved):

PR #2430, opened 2025-08-19
Scale Tree.HEADER_MARGIN by zoom level instead of using fixed pixels

ShahzaibIbrahim:
The Tree.HEADER_MARGIN constant defines extra fixed-width spacing for tree headers. Previously, this was a fixed pixel value, which worked at 100% zoom but did not scale properly on high-DPI monitors.

This change redefines HEADER_MARGIN in points and converts it to pixels based on the current zoom level, ensuring consistent header spacing across different display scales.

### Before and After 

Header Width at 350% zoom

image


arunjose696:
The change looks good to me. However, at 350% zoom, my measured pixel values using PowerToys are 233 (with the change) and 195 (without the change).

Previously, the column width behaved inconsistently depending on how the header was created:

- If the column header was started directly at 350% zoom, the width would be 195 px.
- If the header was created at 100% zoom and then moved to a monitor with 350% zoom, the width would become 233 px.

With this change, the column width is now consistent across both scenarios.

arunjose696 (approved):

akoch-yatta (approved):
Code looks fine and testing short changes work as expected

PR #2431, opened 2025-08-19
[Win32] Remove unused zoom parameter from CoordinateSystemMapper methods

HeikoKlare:
The zoom parameter in some of the translation methods of the CoordinateSystemMappers and the according methods in Display is unused. This change simply removes the obsolete parameters.


akoch-yatta (approved):
Looks good to me. Just the removal of the unused parameter in the affected method signature and the callers

Issue #2432, opened 2025-08-20
UI broken after DPI change event


HeikoKlare:
**Describe the bug**
I ended in a completely broken UI of an Eclipse IDE after changing the scaling of my primary monitor. The DPI change handling failed with an exception:
```
org.eclipse.swt.SWTException: Graphic is disposed
	at org.eclipse.swt.SWT.error(SWT.java:4946)
	at org.eclipse.swt.SWT.error(SWT.java:4861)
	at org.eclipse.swt.SWT.error(SWT.java:4832)
	at org.eclipse.swt.graphics.Image.getImageData(Image.java:1250)
	at org.eclipse.swt.internal.ImageList.set(ImageList.java:417)
	at org.eclipse.swt.internal.ImageList.getHandle(ImageList.java:342)
	at org.eclipse.swt.widgets.Tree.imageIndex(Tree.java:3754)
	at org.eclipse.swt.widgets.TreeItem.setImage(TreeItem.java:1664)
	at org.eclipse.swt.widgets.TreeItem.setImage(TreeItem.java:1693)
	at org.eclipse.swt.internal.CommonWidgetsDPIChangeHandlers.handleItemDPIChange(CommonWidgetsDPIChangeHandlers.java:46)
	at org.eclipse.swt.internal.DPIZoomChangeRegistry.applyChange(DPIZoomChangeRegistry.java:56)
```

**To Reproduce**
I have no reproducer yet, it just happened for me once by now.

**Expected behavior**
Rescaling on a DPI change (such as when changing primary monitor zoom) should result in a proper UI and, particularly, should not throw an exception.

I see multiple things that need to be done here
- Find out why this exception occurs at all
- Check for the image being disposed when setting it on the TreeItem during DPI change handling
- Embed every DPI change handler into a try/catch to ensure that a failure of a single handler does not lead to a completely (but at most partially) broken UI

**Screenshots**
<img width="1583" height="828" alt="Image" src="https://github.com/user-attachments/assets/6a83d958-12d7-43c0-8976-7d47fe118f36" />

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 

2. Additional OS info (e.g. OS version, Linux Desktop, etc)
Windows 11

**Version since**
I20250815-1800

PR #2433, opened 2025-08-20
[Win32] Avoid broken UI when DPI change handler fails #2432

HeikoKlare:
In case a DPI change handler fails with an exception, the complete rescaling process may fail. As an example, it may happen that an image set to an Item has already been disposed when a DPI change event occurs, but the DPI change handler still tries to set that image on the item again, leading to an exception because of the image being disposed.

This change ensures that an exception happening when processing a DPI change handler does not result in a complete rescaling failure and instead reports the listener error to the Display.

Contributes to https://github.com/eclipse-platform/eclipse.platform.swt/issues/2432


HeikoKlare (commented):
This should go in RC1 as it comes without risk of regression but reduces the risk of ending up in a broken UI when some UI rescaling handler fails.

PR #2436, opened 2025-08-21
Scale Tree.SORT_WIDTH by zoom level instead of using fixed pixels

ShahzaibIbrahim:
The Tree.SORT_WIDTH constant defines the width reserved for the sort indicator  
in tree headers. Previously, this was a fixed pixel value, which appeared too  
small on high-DPI displays.  

This change redefines SORT_WIDTH in points and converts it to pixels according  
to the current zoom level, ensuring consistent sort indicator spacing across  
different display scales.

### Example 

- Run the following snippet with 250% zoom

```java
import org.eclipse.swt.*;

public class TreeColumnSortWidthExample {


public static void main(String [] args) {
	shell.setText("Custom gradient selection for Tree");
	final Tree tree = new Tree(shell, SWT.MULTI | SWT.FULL_SELECTION);
	tree.setHeaderVisible(true);
	tree.setLinesVisible(true);
	int columnCount = 4;
	for (int i=0; i<columnCount; i++) {
		TreeColumn column = new TreeColumn(tree, SWT.NONE);
		column.setText("Column " + i);
	}
	int itemCount = 3;
	for (int i=0; i<itemCount; i++) {
		TreeItem item1 = new TreeItem(tree, SWT.NONE);
		item1.setText("item "+i);
		for (int c=1; c < columnCount; c++) {
			item1.setText(c, "item ["+i+"-"+c+"]");
		}
		for (int j=0; j<itemCount; j++) {
			TreeItem item2 = new TreeItem(item1, SWT.NONE);
			item2.setText("item ["+i+" "+j+"]");
			for (int c=1; c<columnCount; c++) {
				item2.setText(c, "item ["+i+" "+j+"-"+c+"]");
			}
			for (int k=0; k<itemCount; k++) {
				TreeItem item3 = new TreeItem(item2, SWT.NONE);
				item3.setText("item ["+i+" "+j+" "+k+"]");
				for (int c=1; c<columnCount; c++) {
					item3.setText(c, "item ["+i+" "+j+" "+k+"-"+c+"]");
				}
			}
		}
	}
	tree.addListener(SWT.EraseItem, event -> {
		event.detail &= ~SWT.HOT;
		if ((event.detail & SWT.SELECTED) != 0) {
			GC gc = event.gc;
			Rectangle area = tree.getClientArea();
			/*
			 * If you wish to paint the selection beyond the end of
			 * last column, you must change the clipping region.
			 */
			int columnCount1 = tree.getColumnCount();
			if (event.index == columnCount1 - 1 || columnCount1 == 0) {
				int width = area.x + area.width - event.x;
				if (width > 0) {
					Region region = new Region();
					gc.getClipping(region);
					region.add(event.x, event.y, width, event.height);
					gc.setClipping(region);
					region.dispose();
				}
			}
			gc.setAdvanced(true);
			if (gc.getAdvanced()) gc.setAlpha(127);
			Rectangle rect = event.getBounds();
			Color foreground = gc.getForeground();
			Color background = gc.getBackground();
			gc.setForeground(display.getSystemColor(SWT.COLOR_RED));
			gc.setBackground(display.getSystemColor(SWT.COLOR_LIST_BACKGROUND));
			gc.fillGradientRectangle(0, rect.y, 500, rect.height, false);
			// restore colors for subsequent drawing
			gc.setForeground(foreground);
			gc.setBackground(background);
			event.detail &= ~SWT.SELECTED;
		}
	});
	tree.setSortColumn(tree.getColumn(0));
	tree.setSortDirection(SWT.UP);
	for (int i=0; i<columnCount; i++) {
		tree.getColumn(i).pack();
	}
	tree.setSelection(tree.getItem(0));
	shell.setSize(500, 200);
	shell.open();
	while (!shell.isDisposed()) {
	}
	display.dispose();
}
}
```
- Note that the first window is before change and second window is after this change. The first column (sort column) has extra pixels as the SORT_WIDTH value is stored in points and scaled up to pixels as per zoom level.

image

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/TreeColumn.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/TreeColumn.java
@@ -380,7 +380,7 @@ public void pack () {
 	if (image != null || parent.sortColumn == this) {
 		Image headerImage = null;
 		if (parent.sortColumn == this && parent.sortDirection != SWT.NONE) {
-			headerWidth += Tree.SORT_WIDTH;
+			headerWidth += Win32DPIUtils.pointToPixel(Tree.SORT_WIDTH, getZoom()) ;
 		} else {
 			headerImage = image;
 		}


akoch-yatta (approved):
In itself the change makes sense and does what it should do. So we can merge this from my side.

Only thing I want to mention, that the overall logic doesn't seem to make sense anymore. Looks like (latest) since Windows 7 the sort indicator is above the column, but with Windows XP the indicator was next to the line, where the logic probably originated from.
image

@HeikoKlare WDYT?

HeikoKlare:
If we are sure that at least starting with Windows 10 the sort marker is always at the top of the column header and not at the right side anymore, it would make sense to get rid of the code. I cannot imagine that there is a way to configure that, but I would expect that it was a general UI/UX change in some Windows version and that this cannot be configured, but I do not know either and from a quick search I did not find out anything about such a change to Windows (but if it was changed for Vista or 7, it's already quite long ago).
So if that code was not touched since then anyway, I would be in favor of removing it. @ShahzaibIbrahim @akoch-yatta can you create a ticket for us to remove that code?

ShahzaibIbrahim:
> If we are sure that at least starting with Windows 10 the sort marker is always at the top of the column header and not at the right side anymore, it would make sense to get rid of the code. I cannot imagine that there is a way to configure that, but I would expect that it was a general UI/UX change in some Windows version and that this cannot be configured, but I do not know either and from a quick search I did not find out anything about such a change to Windows (but if it was changed for Vista or 7, it's already quite long ago). So if that code was not touched since then anyway, I would be in favor of removing it. @ShahzaibIbrahim @akoch-yatta can you create a ticket for us to remove that code?

I can create a new ticket for this. 

Here, https://github.com/vi-eclipse/Eclipse-Platform/issues/429

PR #2440, opened 2025-08-21
Simplify JavadocBasher

akurtakov:
* Use Java NIO
* Less casting
* Use new collections api
* Stop caring about CVS dirs
* Use foreach


akurtakov:
Releng only change thus fine to merge during freeze.

PR #2441, opened 2025-08-21
Scale Tree.EXPLORER_EXTRA by zoom level instead of using fixed pixels

ShahzaibIbrahim:
The Tree.EXPLORER_EXTRA constant specifies the additional spacing applied to tree items when using the Explorer theme. Previously, this was defined as a fixed pixel value, which caused the extra spacing to appear too small on high-DPI monitors.

This change redefines EXPLORER_EXTRA in points and converts it to pixels based on the current zoom level, ensuring consistent spacing in Explorer-themed trees across different display scales.

### How to Test

- Run the Runtime workspace on primary monitor 250%
- See the project/package explorer.
- The items should look little bit bigger.

### Results 
**Before :**
image

**After :**
image

**Note:** the difference is very minor (from 2px to 4px) so you might not notice the difference but using some sort of scale, you can compare the size of items before and after this change.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
@@ -467,8 +467,8 @@ LRESULT CDDS_ITEMPOSTPAINT (NMTVCUSTOMDRAW nmcd, long wParam, long lParam) {
 					if (explorerTheme) {
 						if (hooks (SWT.EraseItem)) {
 							RECT itemRect = item.getBounds (index, true, true, false, false, true, hDC);
-							itemRect.left -= EXPLORER_EXTRA;
-							itemRect.right += EXPLORER_EXTRA + 1;
+							itemRect.left -= Win32DPIUtils.pointToPixel(EXPLORER_EXTRA, getZoom());
+							itemRect.right += Win32DPIUtils.pointToPixel(EXPLORER_EXTRA, getZoom()) + 1;
 							pClipRect.left = itemRect.left;
 							pClipRect.right = itemRect.right;
 							if (columnCount > 0 && hwndHeader != 0) {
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
```diff
@@ -467,8 +467,8 @@ LRESULT CDDS_ITEMPOSTPAINT (NMTVCUSTOMDRAW nmcd, long wParam, long lParam) {
 					if (explorerTheme) {
 						if (hooks (SWT.EraseItem)) {
 							RECT itemRect = item.getBounds (index, true, true, false, false, true, hDC);
-							itemRect.left -= EXPLORER_EXTRA;
-							itemRect.right += EXPLORER_EXTRA + 1;
+							itemRect.left -= Win32DPIUtils.pointToPixel(EXPLORER_EXTRA, getZoom());
+							itemRect.right += Win32DPIUtils.pointToPixel(EXPLORER_EXTRA, getZoom()) + 1;
```
akoch-yatta: There is already the local variable `int zoom` in the method that you can use instead of `getZoom()`

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
```diff
@@ -1135,11 +1135,11 @@ LRESULT CDDS_ITEMPREPAINT (NMTVCUSTOMDRAW nmcd, long wParam, long lParam) {
 						if (measureEvent != null) {
 							pRect.right = Math.min (pClipRect.right, boundsInPixels.x + boundsInPixels.width);
 						} else {
-							pRect.right += EXPLORER_EXTRA;
-							pClipRect.right += EXPLORER_EXTRA;
+							pRect.right += Win32DPIUtils.pointToPixel(EXPLORER_EXTRA, getZoom());
```
akoch-yatta: Can you replace your calls of `getZoom()` with an extracted `int zoom = getZoom();`?

PR #2443, opened 2025-08-22
Scale Tree.GRID_WIDTH by zoom level instead of using fixed pixels

ShahzaibIbrahim:
The Tree.GRID_WIDTH constant specifies the extra width added to a tree item when lines between columns are visible (`lineVisible = true`). This compensates for the space taken by the grid line. Previously, this was defined as a fixed pixel value, which appeared too small on high-DPI monitors.

This change redefines GRID_WIDTH in points and converts it to pixels based on the current zoom level, ensuring consistent column spacing across different display scales.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
@@ -2960,11 +2960,11 @@ TreeItem getFocusItem () {
  */
 public int getGridLineWidth () {
 	checkWidget ();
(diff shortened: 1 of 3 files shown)


akoch-yatta:
@ShahzaibIbrahim While reviewing, I noticed that [here](https://github.com/eclipse-platform/eclipse.platform.swt/blob/d5dc66178647fa40b14165a7b54bc3be8e013174/bundles/org.eclipse.swt/Eclipse%20SWT/win32/org/eclipse/swt/widgets/Tree.java#L8237) Table.GRID_WIDTH is used in Tree for some reason. Can you adapt this place in this PR as well, please

PR #2444, opened 2025-08-22
[GTK] Fix Combo copy/paste tests on Gtk 4.x

akurtakov:
Gtk 4 no longer provides API for doing that but rather a built-in action to activate. This requires to spin the event loop to process it.


laeubi:
I agree with @HeikoKlare that spinning the event queue is not really nice and wondering why it is required at all should this not happen anyways (sooner or later) and is the problem probably more that one expect the past happen instantly?

Should we probably just enhance the javadoc that the paste can happen asynchronously and any code that relies on the happen-before should use a display.asyncExec call before access the widget state after that?

iloveeclipse:
I assume this can be closed now, superceeded by https://github.com/eclipse-platform/eclipse.platform.swt/pull/2489

akurtakov:
This is considered too risky so a documentation is being added and tests event loop spinned instead.

PR #2446, opened 2025-08-22
Modify handleMonitorSpecificDpiChange to call setBounds before sending zoomChangedEvent

arunjose696:
This change ensures that setBounds is applied before the zoomChangedEvent is dispatched. By doing so, event listeners of ZoomChanged could modify the shell size additionally if required.

For example, this is useful in https://github.com/eclipse-platform/eclipse.platform.ui/pull/3212 , where the shell size needs to be set by a ZoomChanged listener.


ShahzaibIbrahim:
The stated issue seems to fix and sounds logical to me. but @arunjose696 suggested that he is still doing some work on it. So I will come back later again to review.

akoch-yatta:
This PR could affect the dark theme main menu bar. I did not try it yet with this PR, but while finalizing #2520 I noticed changing the order broke the main menu bar in the dark theme. With the merged state this could not be the case anymore, but be aware to test this scenario in case we want to proceed with this PR

arunjose696:
> This PR could affect the dark theme main menu bar. I did not try it yet with this PR, but while finalizing #2520 I noticed changing the order broke the main menu bar in the dark theme. With the merged state this could not be the case anymore, but be aware to test this scenario in case we want to proceed with this PR

This change remains blocked due to the mentioned issue , as changing the order setBounds is called would cause problems, and the original issue which requires this fix https://github.com/eclipse-platform/eclipse.platform.ui/pull/3212 is now unconsidered closing this pull request

Issue #221, opened 2022-06-22
Default external web browser do not work on Ubuntu 22.04


jakublabenski:
**How to reproduce.**

1. Start eclipse from terminal to see standard output
2. In 'Web Browser" preferences set "external web browser" and enable "Default system web browser"
3. Create empty file with html extension
4. Double-click html file

Browser will not open and there will be message in terminal:
```
env: '<html file path>'': Permission denied
```

**Environment**

Ubuntu 22.04 with Firefox from snap
Eclipse 4.24 (this is not limited to this version)

**Problem**

When Firefox is installed from Snap (which is default in Ubuntu 22.04) then Exec property in desktop file will have full command line not just executable path:

Snap version (/var/lib/snapd/desktop/applications/firefox_firefox.desktop):

```
Exec=env BAMF_DESKTOP_FILE_HINT=/var/lib/snapd/desktop/applications/firefox_firefox.desktop /snap/bin/firefox %u
```
Deb version (/usr/share/applications/firefox.desktop):
```
Exec=firefox %u
```

To get browser command line method Program.gio_getProgram is used.

From bundles/org.eclipse.swt/Eclipse SWT Program/gtk/org/eclipse/swt/program/Program.java:
```java
static Program gio_getProgram (long application) {
	Program program = new Program();
        // ... cut ...
	long applicationCommand = OS.g_app_info_get_executable (application);
	if (applicationCommand != 0) {
		length = C.strlen (applicationCommand);
		if (length > 0) {
			buffer = new byte [length];
			C.memmove (buffer, applicationCommand, length);
			program.command = new String (Converter.mbcsToWcs (buffer));
		}
	}
```
Native `OS.g_app_info_get_executable` method returns executable only, not full command line.
So for Snap Firefox it returns `env`, which is the used as executable to show html file.
This is consistent with error message in terminal above. For non-snap firefox it would return `firefox`.

Possible fix could use `g_app_info_get_commandline` function that return command line instead of just executable.

Simple C program that demonstrates difference:
```C
#include <gio/gio.h>

#include <stdio.h>

int main()
{
    GAppInfo* info = g_app_info_get_default_for_type("text/html", 0);
    if (info) {
        printf("Name: %s\n", g_app_info_get_name(info));
        printf("Exec: %s\n", g_app_info_get_executable(info));
        printf("Cmdl: %s\n", g_app_info_get_commandline(info));
    } else {
        printf("error\n");
    }
};
```
[...]

PR #2448, opened 2025-08-26
[GTK] Support launching Flatpak/Snap applications

akurtakov:
Make Program class explicitly check for "/usr/bin/flatpak"(flatpak) and for "env"(snap) commands and use the commandline in this case instead of the command only.
Command line is not used every time to not introduce breakage unintentionally in some other cases.


akurtakov:
Fixes https://github.com/eclipse-platform/eclipse.platform.swt/pull/221

PR #2452, opened 2025-08-27
Scale Table.EXPLORER_EXTRA by zoom level instead of using fixed pixels

ShahzaibIbrahim:
The Table.VISTA_EXTRA and Table.EXPLORER_EXTRA constants specify additional spacing applied to table items when using the Vista or Explorer themes. Previously, both were defined as fixed pixel values, which worked at 100% zoom but did not
scale properly on high-DPI monitors.

This change redefines VISTA_EXTRA and EXPLORER_EXTRA in points and converts them to pixels based on the current zoom level, ensuring consistent spacing for themed table items across different display scales.

### How to Test

- Run Snippet 7 from Snippet Explorer with Primary monitor of 250%
- You will notice the width is bit (2 px) wider than previous state since we take zoom into consideration.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Table.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Table.java
@@ -3545,6 +3545,7 @@ void sendEraseItemEvent (TableItem item, NMLVCUSTOMDRAW nmcd, long lParam, Event
 			boolean backgroundWanted = !ignoreDrawHot || drawDrophilited || (!ignoreDrawSelection && clrSelectionBk != -1);
 
 			if (backgroundWanted) {
+				int explorerExtraInPixels = Win32DPIUtils.pointToPixel(EXPLORER_EXTRA, getZoom());
 				RECT pClipRect = new RECT ();
 				OS.SetRect (pClipRect, nmcd.left, nmcd.top, nmcd.right, nmcd.bottom);
 				RECT rect = new RECT ();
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Table.java
```diff
@@ -3561,10 +3561,10 @@ void sendEraseItemEvent (TableItem item, NMLVCUSTOMDRAW nmcd, long lParam, Event
 					OS.MapWindowPoints (hwndHeader, handle, headerRect, 2);
 					rect.left = headerRect.left;
 					pClipRect.left = cellRect.left;
-					pClipRect.right += EXPLORER_EXTRA;
+					pClipRect.right += Win32DPIUtils.pointToPixel(EXPLORER_EXTRA, getZoom());
 				} else {
-					rect.right += EXPLORER_EXTRA;
-					pClipRect.right += EXPLORER_EXTRA;
+					rect.right += Win32DPIUtils.pointToPixel(EXPLORER_EXTRA, getZoom());
+					pClipRect.right += Win32DPIUtils.pointToPixel(EXPLORER_EXTRA, getZoom());
```
akoch-yatta: I would propose to extract that into a local variable, so you have this calculation not done twice

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Table.java
```diff
@@ -4880,7 +4880,7 @@ boolean setScrollWidth (TableItem item, boolean force) {
 			*/
 			newWidth++;
 		}
-		newWidth += INSET * 2 + VISTA_EXTRA;
+		newWidth += INSET * 2 + Win32DPIUtils.pointToPixel(VISTA_EXTRA, getZoom());
```
akoch-yatta, line 4884: This one sounds like a Windows Vista specific variable? But why is it then applied all the time? What is the visible difference, when this value is not added?

ShahzaibIbrahim: While I agree and it does look wrong, It was applied all the time for some reason. I can open a ticket to put this under condition and add only if windows is Vista.

akoch-yatta: If it is only for Vista, we could remove it. Vista is not supported at all anymore. But as there is not issue known with that it is probably okay, to just do this refactoring as is.

akoch-yatta:
@ShahzaibIbrahim Are you sure about your How To Test? With the SnippetExplorer I don't run into any of the usages of this constants.

ShahzaibIbrahim:
> @ShahzaibIbrahim Are you sure about your How To Test? With the SnippetExplorer I don't run into any of the usages of this constants.

I might have mentioned wrong snippet here. I will check and confirm it in a bit.

Edit: It was a snippet 7, I mistyped. @akoch-yatta

akoch-yatta (approved):
Changes look good and ensure a consistent behavior independent of the zoom the Table is rendered for

PR #2453, opened 2025-08-27
Scale Table.GRID_WIDTH by zoom level instead of using fixed pixels

ShahzaibIbrahim:
The Table.GRID_WIDTH constant specifies the extra width added to a table item when lines between columns are visible (`lineVisible = true`). This accounts for the space taken by the grid line. Previously, it was defined as a fixed
pixel value, which appeared too small on high-DPI monitors.

This change redefines GRID_WIDTH in points and scales it to pixels based on the current zoom level. At 100% zoom this results in 2px instead of 1px, which is visually negligible but ensures consistency by keeping all constants defined in points and scaled according to zoom level.

### How to Test

- Run the following snippet on a Primary monitor with 250% zoom
- You will notice extra pixel in the width compared to previous state. 

```

import org.eclipse.swt.*;

public class LinesVisibleTableExample {

        shell.setText("SWT Table Example");

        // Create table
        final Table table = new Table(shell, SWT.BORDER | SWT.FULL_SELECTION);
        table.setHeaderVisible(true);
        table.setLinesVisible(true);

        // Create columns
        String[] titles = { "Level 0", "Level 1", "Level 2", "Level 3" };
        for (String title : titles) {
            TableColumn column = new TableColumn(table, SWT.NONE);
            column.setText(title);
            column.setWidth(100);
        }

        // Fill table with hierarchical-like data
        for (int i = 0; i < 4; i++) {
            for (int j = 0; j < 4; j++) {
                for (int k = 0; k < 4; k++) {
                    for (int l = 0; l < 4; l++) {
                        TableItem item = new TableItem(table, SWT.NONE);
                        item.setText(new String[] {
                            "TableItem (0) -" + i,
                            "TableItem (1) -" + j,
                            "TableItem (2) -" + k,
                            "TableItem (3) -" + l
                        });
                    }
                }
            }
        }

        shell.setSize(450, 250);
        shell.open();
        while (!shell.isDisposed()) {
        }
        display.dispose();
    }
}
```

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Table.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Table.java
@@ -2334,11 +2334,11 @@ int getFocusIndex () {
  */
 public int getGridLineWidth () {
 	checkWidget ();
-	return DPIUtil.pixelToPoint(getGridLineWidthInPixels(), getZoom());
+	return GRID_WIDTH;
 }
 
 int getGridLineWidthInPixels () {
-	return GRID_WIDTH;
+	return Win32DPIUtils.pointToPixel(GRID_WIDTH, getZoom());
 }
 
 /**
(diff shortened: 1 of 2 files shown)


akoch-yatta (approved):
Changes are very similar to #2443. Changes look sound and I didn't find any problems when testing the changes.

PR #2454, opened 2025-08-27
Scale Table.HEADER_MARGIN by zoom level instead of using fixed pixels

ShahzaibIbrahim:
The Table.HEADER_MARGIN constant defines the extra fixed-width spacing for table headers, while the Table.HEADER_EXTRA constant specifies the additional spacing applied to the header area. Previously, both were defined as fixed pixel values,
which worked at 100% zoom but did not scale properly on high-DPI monitors.

This change redefines HEADER_MARGIN and HEADER_EXTRA in points and converts them to pixels based on the current zoom level, ensuring consistent header spacing across different display scales.

### How to Test

- Run Snippet 8 from Snippet Explorer with Primary monitor of 250%
- You will notice the width of the header wider than previous state since we take zoom into consideration.


ShahzaibIbrahim:
@akoch-yatta this PR was left behind as part of this ticket: https://github.com/vi-eclipse/Eclipse-Platform/issues/404. Could you reveiw it?

Issue #2457, opened 2025-08-27
[GTK] Resetting `MenuItem` and `ToolItem` image to `null` doesn't clear the image


ShadelessFox:
**Describe the bug**

Speaking in code, calling `setImage(new Image(...))` followed by `setImage(null)` on either `MenuItem` or `ToolItem` doesn't clear the image.

**To Reproduce**
```java
import org.eclipse.swt.SWT;
import org.eclipse.swt.graphics.Image;
import org.eclipse.swt.layout.GridLayout;
import org.eclipse.swt.widgets.*;

public class ItemIconBug {
    public static void main(String[] args) {
        Display display = new Display();
        Image image = new Image(display, "eclipse.png");

        Shell shell = new Shell(display);
        shell.setLayout(new GridLayout());
        shell.setSize(200, 100);

        Menu menu = new Menu(shell, SWT.POP_UP);
        shell.setMenu(menu);

        MenuItem menuItem = new MenuItem(menu, SWT.PUSH);
        menuItem.setImage(image);
        menuItem.setImage(null);

        ToolBar toolBar = new ToolBar(shell, SWT.NONE);
        ToolItem toolItem = new ToolItem(toolBar, SWT.PUSH);
        toolItem.setImage(image);
        toolItem.setImage(null);

        shell.open();

        while (!shell.isDisposed()) {
            if (!display.readAndDispatch()) {
                display.sleep();
            }
        }

        display.dispose();
    }
}
```

Image used in the snippet:
![Image](https://github.com/user-attachments/assets/c7940cbb-33c4-4588-83b5-334c263ea2a8)

**Expected behavior**
The image gets cleared when calling `setImage(null)`

**Screenshots**
Linux (faulty behavior):
<img width="207" height="136" alt="Image" src="https://github.com/user-attachments/assets/1731a6c6-1363-4527-b9eb-eebd0020c256" />

Windows (expected behavior):
<img width="186" height="93" alt="Image" src="https://github.com/user-attachments/assets/a04f83f3-ca3f-4acc-8e42-7bed1165dc87" />

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [ ] Windows
- - [x] Linux
- - [ ] macOS 

2. Additional OS info (e.g. OS version, Linux Desktop, etc)
Linux Manjaro 25.0 KDE

3. JRE/JDK version
openjdk version "24.0.1" 2025-04-15
 
**Version since**
[...]

PR #2459, opened 2025-08-27
[GTK] Fix MenuItem/ToolItem setImage(null)

akurtakov:
Fix for https://bugs.eclipse.org/bugs/show_bug.cgi?id=568771 had the wrong assumption that _setImage should be called only when non null image.
Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2457

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/MenuItem.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/MenuItem.java
@@ -986,8 +986,9 @@ private void _setEnabledOrDisabledImage() {
 			defaultDisableImage = new Image(getDisplay(), image, SWT.IMAGE_DISABLE);
 		}
 		_setImage(defaultDisableImage);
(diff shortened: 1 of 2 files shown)


akurtakov:
From my testing this fixes the issue and disabled images are not regressing. I'll merge once new stream opens unless someone finds something breaks by then.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/MenuItem.java
```diff
@@ -987,7 +987,7 @@ private void _setEnabledOrDisabledImage() {
 		}
 		_setImage(defaultDisableImage);
 	}
-	if (enabled && image != null) _setImage(image);
+	if (enabled) _setImage(image);
```
HeikoKlare: nit: this could now just be an `else`

akurtakov: Thanks!

Issue #2037, opened 2025-04-17
[Win32] Shell image broken


tmssngr:
**Describe the bug**
After having upgraded SWT from commit 1ced8a9c to 3670670e in our application it happens occasionally that the image for some shell is missing.

**To Reproduce**
Unfortunately, I have not yet found a way to reproduce it reliably.

**Expected behavior**
The set image should be used. It work without problem until at least commit 1ced8a9c.

**Screenshots**
<img width="271" alt="Image" src="https://github.com/user-attachments/assets/d865a99d-d77c-4cc6-87b0-84f71e57bf85" />

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 

2. Additional OS info (e.g. OS version, Linux Desktop, etc)
Windows 11

tmssngr:
It looks like this behavior is not fixed, but changes during runtime (an invalid image can turn into a correct one).

tmssngr:
As this screenshot indicates, the icon is set (see hover), but does not show correctly in the status bar
<img width="167" alt="Image" src="https://github.com/user-attachments/assets/407b8cda-20a2-444b-830c-a5223587eee9" />

tmssngr:
@HeikoKlare Could this bug be resolved with your commit 28fdeb6d?

HeikoKlare:
To be honest, I cannot say whether that commit might have fixed it. The mentioned commit fixes a slight change in behavior that was introduced in the recent months, but without knowing which commit extractly introduced the behavior you report and without being able to properly reproduce it, it's hard to say whether it is fixed now 😟

HeikoKlare:
@tmssngr Can you please reevaluate with latest master state? We are not able to reproduce the issue.
There was another issue with the program icon used for the shell (https://github.com/eclipse-platform/eclipse.platform.swt/issues/2139), but the issue was introduced after the problematic state you refer to. When doing the fix, we also did not experience your issue.

fedejeanne:
> There was another issue with the program icon used for the shell ([#2139](https://github.com/eclipse-platform/eclipse.platform.swt/issues/2139)), but the issue was introduced after the problematic state you refer to.

[...]

PR #2466, opened 2025-08-29
[Win32] Correct zoom level used for image of decorations/shell #2037

HeikoKlare:
When setting the images for a decorations instance (in particular for a shell) or when updating them upon a zoom change, the two images best fitting to sizes required by the OS in their 100% version are calculated. Then, however, a zoomed version of the image is used based on the auto-scaled zoom of the decorations object. This zoom is, however, completely unrelated to what is required, as the auto-scaled zoom was not taken into account for identifying the best fitting image before.

In the best case, a zoomed version of the image exactly fitting to the size required by the OS would be generated. This would be possible via auto-scaling the image (or in case of SVGs rendering it at the desired size), but this would also lead to bad results if a simple auto-scaling mode was used. Thus, the best possible solution is to use the 100% version of the image, as its size does best fit to the required sizes as conforming to the calculation done before.

Contributes to https://github.com/eclipse-platform/eclipse.platform.swt/issues/2037

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Decorations.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Decorations.java
@@ -888,11 +888,11 @@ private void setImages (Image image, Image [] images) {
 	if (smallIcon != null) {
 		switch (smallIcon.type) {
 			case SWT.BITMAP:
-				smallImage = Display.createIcon (smallIcon, getZoom());
-				hSmallIcon = Image.win32_getHandle(smallImage, getZoom());
+				smallImage = Display.createIcon (smallIcon, 100);
+				hSmallIcon = Image.win32_getHandle(smallImage, 100);
 				break;
 			case SWT.ICON:
-				hSmallIcon = Image.win32_getHandle(smallIcon, getZoom());
+				hSmallIcon = Image.win32_getHandle(smallIcon, 100);
 				break;
 		}
 	}
@@ -900,11 +900,11 @@ private void setImages (Image image, Image [] images) {
 	if (largeIcon != null) {
 		switch (largeIcon.type) {
 			case SWT.BITMAP:
-				largeImage = Display.createIcon (largeIcon, getZoom());
-				hLargeIcon = Image.win32_getHandle(largeImage, getZoom());
+				largeImage = Display.createIcon (largeIcon, 100);
+				hLargeIcon = Image.win32_getHandle(largeImage, 100);
 				break;
 			case SWT.ICON:
-				hLargeIcon = Image.win32_getHandle(largeIcon, getZoom());
+				hLargeIcon = Image.win32_getHandle(largeIcon, 100);
 				break;
 		}
 	}


HeikoKlare:
> Just one final question, for my understanding: I noticed you only adapted the usages in `Decorations` and not in other places (for example, `Display.createIcon(...)` is also called in `TrayItem` and `TaskItem`). Why aren't the same changes needed in those consumers too?

This change is not about the usage of `Display.createIcon(...)` but about how the zoom to create the icon at is calculated. As described in commit and PR message, in this case we calculate the icon best fitting for the OS metrics based on the 100% of the icons, but we then use a version that is zoomed with an arbitrary value (arbitrary in the sense that it is completely unrelated to what is requested).
@fedejeanne So what do you think needs to be changed at those places?

fedejeanne:
Thank you for the clarification. 

> @fedejeanne So what do you think needs to be changed at those places?

If I understand the issue correctly, it should also manifest in the tray area of the OS, which means that the usages in `TrayItem::setImage(Image)` would also require the change. Does the issue also happen in there?

I'm not sure about `TaskItem` though, I'm not familiar with the code.

HeikoKlare:
> If I understand the issue correctly, it should also manifest in the tray area of the OS, which means that the usages in TrayItem::setImage(Image) would also require the change. Does the issue also happen in there?

Again: why should that be the case? There is no calculation in `TrayItem` based on system metrics, matching a specific zoom of multiple images to such metrics etc. (actually there is only a single image in that case), like it's on `Decorations`. Why should any other zoom fit better than the one returned of `getZoom()` and how should that be calculated?

fedejeanne:
I just assumed that the code being so similar in both classes, the issue might manifest itself in the tray. But I read between the lines that this is not the case :-)

Thank you.

HeikoKlare:
fwiw: there is nothing in between the lines. I just don't understand where the code is similar in both cases, as it's actually completely different. Maybe I miss something.
As I said, in the `TrayItem` there is just one image retrieved at the appropriate size for the context zoom. In `Decorations` there are multiple images provided and the best zoomed version of the best fitting image is retrieved by comparing against the size that is required according to system metrics. That's also what the commit and PR description explains. So those calculations are completely different and that's why I wanted to understand why there may be a need for adaptation. You seem to have something specific in mind when you say that the code is similiar and may be updated equally.

fedejeanne:
> I just don't understand where the code is similar in both cases

I meant this pattern of checking for a type (a `switch`) and, based on it, either create an icon (`Display.createIcon(x, zoom)`) and get a handle (`Image.win32_getHandle(x, zoom)`) or just get the handle.

I saw the pattern in 3 places:

<details><summary>TrayItem</summary>
<p>

https://github.com/eclipse-platform/eclipse.platform.swt/blob/1094ebd97ea0a859041103befc65b529724081a3/bundles/org.eclipse.swt/Eclipse%20SWT/win32/org/eclipse/swt/widgets/TrayItem.java#L458-L466

</p>
</details> 

<details><summary>TaskItem</summary>
<p>

https://github.com/eclipse-platform/eclipse.platform.swt/blob/1094ebd97ea0a859041103befc65b529724081a3/bundles/org.eclipse.swt/Eclipse%20SWT/win32/org/eclipse/swt/widgets/TaskItem.java#L474-L482

</p>
</details> 

<details><summary>Decorations (adapted in this PR)</summary>
<p>

https://github.com/eclipse-platform/eclipse.platform.swt/blob/1094ebd97ea0a859041103befc65b529724081a3/bundles/org.eclipse.swt/Eclipse%20SWT/win32/org/eclipse/swt/widgets/Decorations.java#L889-L897

---

https://github.com/eclipse-platform/eclipse.platform.swt/blob/1094ebd97ea0a859041103befc65b529724081a3/bundles/org.eclipse.swt/Eclipse%20SWT/win32/org/eclipse/swt/widgets/Decorations.java#L901-L909



</p>
</details>

HeikoKlare:
So you just refer to code duplication and propose to extract that pattern to a common utilty method? But what's the relation to the change in this PR then?

fedejeanne:
Nope, I just wanted to raise the question to see if the problem arises in those other 2 places too.

HeikoKlare:
Sorry, I still don't know which problem you refer you. What's the change you may expect at those other places?

HeikoKlare:
I guess we can close this discussion as there does not seem to be a concrete issue description or fix proposal?

PR #2468, opened 2025-08-29
[Win32] Improve decorations/shell images size selection

HeikoKlare:
The algorithm to identify the best fitting image set to a decorations/shell instance for a specific size required by the OS currently only considers the difference between the target size and the actual size of the image. Smaller and larger images with the same difference are treated equally. However, usually one wants to prefer downscaling a higher-resolution image than upscaling a lower-resolution one to improve the quality.

This change ensures that larger images are preferred over smaller images if the difference is 1.5 times as high or less. One particular change is that on a 150% monitor now a 200% image will be downscaled instead of upscaling a 100% image.


fedejeanne:
Test failure was unrelated: https://github.com/eclipse-platform/eclipse.platform.swt/issues/2098

PR #2470, opened 2025-09-01
Caret scaling implementation improvement on StyledText DPI change

amartya4256:
This PR improves the way carets of StyledText is scaled on DPI change event by utilizing APIs instead of win32 specific static methods.

Earlier we used to provide a Consumer in CommonWidgetDPIChangeHandlers.handleStyledTextDPIChange to StyledText.updateAndRefreshCarets for performing scaling on DPI change.

To move forward with DPI change performance improvement, we need to get rid of CommonWidgetsDPIChangeHandlers. Which is only possible if we move these handlers to their respective classes and register these handlers on a ZoomChanged event. Which is the next step here: https://github.com/eclipse-platform/eclipse.platform.swt/pull/2483

However, the consumer mentioned above uses a platform specific piece of code: `Caret.win32_setHeight(caretToRefresh, styledText.getLineHeight());` Which cannot be moved to Common Widgets.

Hence, we need to improve the way we scale StyledText and its carets. So, this PR does the following changes:
1. The StyledText at first updates the carets' location and size.
2. Then the DPI change handlers in the carets in the styledText are called, where we call the caret.resize() which explicitly draws the caret with the newly scaled size.
3. Earlier caret.resize() was called using win32_setHeight. As using any other API like setSize, setBounds, etc. had a condition that if the size (in points) doesn't change, resize will not be called leading to no repaint. However, on a DPI change, the size changes in pixels and a repaint must be triggered. Hence, resize is not explicitly called (leading to repaint) in Caret.handleDPIChange which is triggered right after StyledText has adapted the size, location and visibility of the carets.

This change removes the need to use a platform specific (win32_) static method on DPI change and enables us to move forward with implementation of asynchronous DPI change.

contributes to https://github.com/eclipse-platform/eclipse.platform.swt/issues/62 and https://github.com/eclipse-platform/eclipse.platform.swt/issues/128

Depends on https://github.com/eclipse-platform/eclipse.platform.swt/pull/2469 and https://github.com/eclipse-platform/eclipse.platform.swt/pull/2471

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/StyledText.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/StyledText.java
@@ -10912,18 +10912,17 @@ void updateSelection(int startOffset, int replacedLength, int newLength) {
  * @noreference This method is not intended to be referenced by clients.
  */
 public static void updateAndRefreshCarets(StyledText styledText, Consumer<Caret> caretUpdater) {
-	Set<Caret> caretSet = new HashSet<>();
-	caretSet.add(styledText.getCaret());
+	styledText.updateCaretVisibility();
+	styledText.setCaretLocations();
+	Set<Caret> caretSet = new LinkedHashSet<>();
 	caretSet.add(styledText.defaultCaret);
+	caretSet.add(styledText.getCaret());
 	if (styledText.carets != null) {
 		for (Caret caret : styledText.carets) {
 			caretSet.add(caret);
 		}
 	}
 	caretSet.stream().filter(Objects::nonNull).forEach(caretUpdater);
-
-	styledText.updateCaretVisibility();
-	styledText.setCaretLocations();
 
 }
 
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/CommonWidgetsDPIChangeHandlers.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/CommonWidgetsDPIChangeHandlers.java
@@ -51,11 +51,8 @@ public class CommonWidgetsDPIChangeHandlers {
 		if (!(widget instanceof StyledText styledText)) {
 			return;
 		}
-
-
 		StyledText.updateAndRefreshCarets(styledText, caretToRefresh -> {
 			DPIZoomChangeRegistry.applyChange(caretToRefresh, newZoom, scalingFactor);
-			Caret.win32_setHeight(caretToRefresh, styledText.getLineHeight());
 		});
 	}
 	private static void handleCComboDPIChange(Widget widget, int newZoom, float scalingFactor) {
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Caret.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Caret.java
@@ -658,29 +658,6 @@ public void setVisible (boolean visible) {
 	}
 }
 
-/**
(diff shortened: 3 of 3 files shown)


HeikoKlare (commented):
Can you explain what changes when replacing the calls to `SystemParametersInfo` to `SystemParametersInfoForDpi`? I.e., what will be different both in terms of user experience and/or in terms of internal functionality when not doing the changes to the DPI change handling as well.

In addition, should be split this change up into replacing the existing calls and adapting the DPI change handling? I would also appreciate an explanation of why the method calls inside `updateAndRefreshCarets()` have to be moved to understand what happens there.

amartya4256:
> Can you explain what changes when replacing the calls to `SystemParametersInfo` to `SystemParametersInfoForDpi`? I.e., what will be different both in terms of user experience and/or in terms of internal functionality when not doing the changes to the DPI change handling as well.
> 
> In addition, should be split this change up into replacing the existing calls and adapting the DPI change handling? I would also appreciate an explanation of why the method calls inside `updateAndRefreshCarets()` have to be moved to understand what happens there.

It is already split. It depends on https://github.com/eclipse-platform/eclipse.platform.swt/pull/2471. The new API is used to retrieve the system's preferred width for the cursor. Earlier, it used the system DPI aware value, which would always return the value for the first monitor on which the App was started on, could lead to the cursors looking too think or too wide. This kind of behaviour is although not always visible as mostly, all the monitors have the same width. With Per-monitor DPI aware API, we make sure that the retrieved width is specific to the current monitor's preference and doesn't appear too think or too flat.

HeikoKlare (changes requested):
> It is already split. It depends on https://github.com/eclipse-platform/eclipse.platform.swt/pull/2471.

I see, thank you! Just commented on the other one. Putting a change request on this one as the one it depends on (and currently included in this) introduces a regression. Thus just wanting to avoid that it merged by accident.

amartya4256:
> I am still missing an explanation of what this movement of the method calls does to resolve an issue. The commit still refers to changes are not present anymore. So this is not ready to be merged.

I have extended the description with the explanation :)

amartya4256:
> * Why do we still need the `Caret#win32_setHeight` method if its only caller is replaced?

Yes, I'll remove it.

amartya4256:
@heiko, I cannot reproduce the behaviour. Can you provide some more details about how you started your runtime, What's your primary monitor zoom, etc?

HeikoKlare:
Just found out that by accident I used a module-info file for testing (which uses a different editor) which shows the issue with this PR:
image

Setup is 100% primary monitor and 150% secondary monitor (to the left).

amartya4256:
> Setup is 100% primary monitor and 150% secondary monitor (to the left).

Interesting. I still cannot reproduce it.

HeikoKlare:
> Interesting. I still cannot reproduce it.

With the Manifest-Info Editor?

amartya4256:
> With the Manifest-Info Editor?

Yes, I started it in a module-info editor just as you described above in the image. Are you sure you have the latest changes from this branch maybe?

HeikoKlare:
> Yes, I started it in a module-info editor just as you described above in the image. Are you sure you have the latest changes from this branch maybe?

I have the latest state. And I started with a clean workspace again. I see that I sometimes need to move back to the 100% monitor to break the cursor, but after moving at least two times, it breaks. The position replacement happens when calling `caret.resize()` in the DPI change handler of `Caret` where `getXInPixels(), getYInPixels()` returns `3,0` (i.e., the beginning of the document) instead of the actual cursor position.

HeikoKlare (approved):
Thank you for the latest update. The change is fine now and also works without the regression I found in the previous version.

The test failures are unrelated and caused by the GHA update to Windows Server 2025. I have documented them here:
- https://github.com/eclipse-platform/eclipse.platform.swt/issues/2516

The Jenkins failure is probably related to:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3329

PR #2471, opened 2025-09-01
Use SystemParametersInfoForDpi  in Caret

amartya4256:
This PR uses SystemParametersInfoForDpi in Caret instead of SystemParametersInfo for better monitor specific scaling.

Contributes to #62 and #128


HeikoKlare (changes requested):
This change currently introduces a regression:
I tested the change and printed out the old and new values for comparison. Turns out that the new method call always writes `0` to the passed buffer and, in particular, the method call always returns `false`, such that the size is not taken into account.
One example consequence is that if I set a high OS cursor size (like 10), the cursor is still 1 pixel wide. For example, in a console window it used to look like this (with OS cursor size 10):
image
And now it looks like this:
image

Note that the documentation of the used method states that it _cannot_ be called with `SPI_GETCARETWIDTH` as done here but will always return `0` in that case: https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-systemparametersinfofordpi#remarks

amartya4256:
@HeikoKlare Turns out SystemParametersInfo doesn't work with SPI_GETCARETWIDTH : https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-systemparametersinfofordpi 
So, the other way of having DPI specific value is by scaling it ourselves. Hence, I have readapted this PR. Please have a look. https://github.com/eclipse-platform/eclipse.platform.swt/pull/2470 has the final implementation with the refactored handler. Hence, this PR only acts as an intermediate state. The goal should be to make sure this doesn't cause any regression.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Caret.java
```diff
@@ -646,6 +652,13 @@ public void setVisible (boolean visible) {
 	}
 }
 
+private int getPrimaryMonitorZoomAtStartup() {
+	long hDC = OS.GetDC(0);
+	int dpi = OS.GetDeviceCaps(hDC, OS.LOGPIXELSX);
+	OS.ReleaseDC(0, hDC);
+	return DPIUtil.mapDPIToZoom(dpi);
+}
```
HeikoKlare: We already have that same method in `MenuItem`. Would it make sense to extract it to, e.g., a package-protected method inside `Display` (and probably even static as it does not require any access to state)?

amartya4256: There's also one in Program. What should we do about it? If we make it public static then I can get rid of that too.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/Win32DPIUtils.java
```diff
@@ -306,6 +306,15 @@ public static boolean isMonitorSpecificScalingActive() {
 		return updateOnRuntimeValue;
 	}
 
+
+
```
HeikoKlare: Please remove these unnecessary whitelines

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
```diff
@@ -2553,6 +2553,13 @@ Font getSystemFont (int zoom) {
 	return systemFont;
 }
 
+static int getPrimaryMonitorZoomAtStartup() {
+	long hDC = OS.GetDC(0);
+	int dpi = OS.GetDeviceCaps(hDC, OS.LOGPIXELSX);
+	OS.ReleaseDC(0, hDC);
+	return DPIUtil.mapDPIToZoom(dpi);
+}
```
HeikoKlare: Why add this method to Display in addition to the DPI utils?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Caret.java
```diff
@@ -132,7 +132,9 @@ Rectangle getBoundsInPixels () {
 	if (width == 0) {
 		int [] buffer = new int [1];
 		if (OS.SystemParametersInfo (OS.SPI_GETCARETWIDTH, 0, buffer, 0)) {
-			return new Rectangle (getXInPixels(), getYInPixels(), buffer [0], getHeightInPixels());
+			int width = DPIUtil.pixelToPoint(buffer [0], Win32DPIUtils.getPrimaryMonitorZoomAtStartup());
+			int widthInPixels = Win32DPIUtils.pointToPixel(width, getNativeZoom());
```
HeikoKlare: Maybe these two lines (or even the whole logic for retrieving the size with `SystemParameterInfo` could be extracted into a common utility method.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
amartya4256: Oops forgot to remove.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Caret.java
amartya4256: Extracted the whole thing to a new method.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/Win32DPIUtils.java
amartya4256: Done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Caret.java
```diff
@@ -130,14 +130,24 @@ Rectangle getBoundsInPixels () {
 		return new Rectangle (getXInPixels(), getYInPixels(), rect.width, rect.height);
 	}
 	if (width == 0) {
-		int [] buffer = new int [1];
-		if (OS.SystemParametersInfo (OS.SPI_GETCARETWIDTH, 0, buffer, 0)) {
-			return new Rectangle (getXInPixels(), getYInPixels(), buffer [0], getHeightInPixels());
+		int widthInPixels = getSystemCaretWidthInPixelsForCurrentMonitor();
+		if (widthInPixels != 0) {
```
HeikoKlare: Isn't this now a change in behavior? It also uses `getWidthInPixels()` in case the width value delivered by `SystemParameterInfo` is zero and not only if the return value of that method is zero.

amartya4256: Shall we use -1 then? Since -1 cannot be a width. I thoughtof using null first but then we need to have type Integer.

amartya4256: Used OptionalInt instead.

HeikoKlare (approved):
Looks and works good. Thank you!

PR #2472, opened 2025-09-01
[Gtk] Make Tree/Table cellDataProc safe for disposed items

akurtakov:
Table/Tree.remove method ends up firing Gtk signal(s) that cause cellDataProc to start. Make cellDataProc explicitly check whether item is disposed to prevent operating on disposed item. Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/1606

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Table.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Table.java
@@ -199,6 +199,9 @@ long cellDataProc (long tree_column, long cell, long tree_model, long iter, long
 	int [] index = new int [1];
 	C.memmove (index, GTK.gtk_tree_path_get_indices (path), 4);
 	TableItem item = _getItem (index[0]);
+	if (item.isDisposed()) {
+		return 0;
+	}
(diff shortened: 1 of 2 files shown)


akurtakov:
Merging as even if a "better" fix could be done, this kind of "safety" in signal handlers is nice to always have.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Table.java
```diff
@@ -199,6 +199,9 @@ long cellDataProc (long tree_column, long cell, long tree_model, long iter, long
 	int [] index = new int [1];
 	C.memmove (index, GTK.gtk_tree_path_get_indices (path), 4);
 	TableItem item = _getItem (index[0]);
+	if (item.isDisposed()) {
+		return 0;
```
laeubi, line 203: do we still need to call `GTK.gtk_tree_path_free (path);` here?

akurtakov: Good catch. https://github.com/eclipse-platform/eclipse.platform.swt/pull/2473 should fix it.

PR #2474, opened 2025-09-01
Only redraw once after scaling every widget #62

amartya4256:
This PR contributes to reducing the number of redraws and only draw once all the widgets have performed scaling using their Widget:handleDpiChange handlers.

contributes to

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Composite.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Composite.java
@@ -1984,6 +1984,5 @@ private static void handleDPIChange(Widget widget, int newZoom, float scalingFac
 	for (Control child : composite.getChildren()) {
 		DPIZoomChangeRegistry.applyChange(child, newZoom, scalingFactor);
 	}
-	composite.redrawInPixels (null, true);
 }
 }


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
```diff
@@ -4961,6 +4962,7 @@ void handleMonitorSpecificDpiChange(int newNativeZoom, Rectangle newBoundsInPixe
 	DPIUtil.setDeviceZoom (newNativeZoom);
 	DPIZoomChangeRegistry.applyChange(this, newNativeZoom, scalingFactor);
 	this.setBoundsInPixels(newBoundsInPixels.x, newBoundsInPixels.y, newBoundsInPixels.width, newBoundsInPixels.height);
+	getShell().WM_SIZE(0, 0);
```
HeikoKlare: Why do we need to use `WM_SIZE` here instead of, e.g., just using some `redraw()` method (like it was sufficient before)? This method is actually a callback function for OS events, so it's strange to call it from application code. I would expect that we have some proper method for doing what we want to do (e.g., requesting a redraw, requesting for size adaptation or the like) and that such a method is then may be called by both the `WM_SIZE` callback and this new place. But first of all, the quesiton is whether existing methods are not already sufficient.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
```diff
@@ -4764,6 +4764,7 @@ public boolean setParent (Composite parent) {
 		int newZoom = parent.nativeZoom;
 		float scalingFactor = 1f * newZoom / oldZoom;
 		DPIZoomChangeRegistry.applyChange(this, newZoom, scalingFactor);
+		getShell().WM_SIZE(0, 0);
```
HeikoKlare: Why do we need to submit a resize event to the shell if just a control inside is reparented? Shouldn't it be sufficient to redraw the control and its contents?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
amartya4256: The thing with WM_SIZE is that it is overriden by many classes for specific implementations. And the WM_SIZE also propagates to all the parents with super.WM_SIZE hence it works the best when it comes to propagating the message down a tree architecture. Hence, even if I extract a method out of it what's needed, I'll have to do it at all the level as they all call each other's WM_SIZE.

However, on a second look. I realized that we do not need WM_SIZE until we move to asynchronous scaling as  it is synchronous and the parents have access to the children size and they only scale and draw once the children have scaled and drawn themselves. Hence, no bad size. So we do not need WM_SIZE in this PR. I'll remove it.

HeikoKlare:
@amartya4256 that means the redraw that we triggered for every composite so far was just redundant? Can you shortly explain why the redraw does still happen at the end of processing the DPI change handlers?

I tested the change and at a first glance it works fine for me. In particular, it already improves performance and the visual experience when doing the rescaling as everything is just redrawn at once and not every composite/control independently, which looks much better.

amartya4256:
> @amartya4256 that means the redraw that we triggered for every composite so far was just redundant? Can you shortly explain why the redraw does still happen at the end of processing the DPI change handlers?
> 
> I tested the change and at a first glance it works fine for me. In particular, it already improves performance and the visual experience when doing the rescaling as everything is just redrawn at once and not every composite/control independently, which looks much better.

On a  second look, the widgets are implemented in such a way that they implicitly redraw upon being resized or moved. There are multiple reasons. Some call layout which triggers it, some send Resize or Redraw event and the handler does it for them. But in the end, most of the widgets seem to be dealing with repainting using layout, which already aligns them and paints them and then the loop control is passed to the parent, who then resize and repaint themselves. So, internally, it is handling repaints the moment we change their bounds, hence, we can also remove shell resize and shell layout from the dpi change handler for Shell and Control's handleMonitorSpecificDPIChange.

HeikoKlare:
Sounds good. Thank you for the explanation! Let's keep potential further improvements for follow-up PRs.

HeikoKlare (approved):

PR #2477, opened 2025-09-02
Remove the SORT_WIDTH spacing logic in Table column rendering.

ShahzaibIbrahim:
Since Windows 7, the sort indicator in tree columns is rendered above the column header text, whereas in Windows XP it was placed next to the text. The original SWT implementation added extra pixels (SORT_WIDTH) to the column width to make space for the indicator.


arunjose696:
Have tested the changes with a table with image the changes look good

Issue #2478, opened 2025-09-02
Various GTK Tree/Table test fails in I20250902-0330


iloveeclipse:
I assume all of the failures below are regression from https://github.com/eclipse-platform/eclipse.platform.swt/pull/2472

See https://download.eclipse.org/eclipse/downloads/drops4/I20250902-0330/testResults.php
 
* https://download.eclipse.org/eclipse/downloads/drops4/I20250902-0330/testresults/html/org.eclipse.jdt.ui.tests_ep438I-unit-linux-x86_64-java24_linux.gtk.x86_64_24.html
* https://download.eclipse.org/eclipse/downloads/drops4/I20250902-0330/testresults/html/org.eclipse.jface.tests_ep438I-unit-linux-x86_64-java24_linux.gtk.x86_64_24.html
* https://download.eclipse.org/eclipse/downloads/drops4/I20250902-0330/testresults/html/org.eclipse.search.tests_ep438I-unit-linux-x86_64-java24_linux.gtk.x86_64_24.html
* https://download.eclipse.org/eclipse/downloads/drops4/I20250902-0330/testresults/html/org.eclipse.ui.tests_ep438I-unit-linux-x86_64-java24_linux.gtk.x86_64_24.html

```
Cannot invoke "org.eclipse.swt.widgets.TableItem.isDisposed()" because "item" is null

java.lang.NullPointerException: Cannot invoke "org.eclipse.swt.widgets.TableItem.isDisposed()" because "item" is null
at org.eclipse.swt.widgets.Table.cellDataProc(Table.java:203)
at org.eclipse.swt.widgets.Display.cellDataProc(Display.java:997)
at org.eclipse.swt.internal.gtk.GTK.gtk_list_store_clear(Native Method)
at org.eclipse.swt.widgets.Table.removeAll(Table.java:2802)
at org.eclipse.swt.widgets.Table.remove(Table.java:2693)
at org.eclipse.jface.viewers.TableViewer.doRemove(TableViewer.java:285)
at org.eclipse.jface.viewers.AbstractTableViewer.internalRefreshAll(AbstractTableViewer.java:709)
at org.eclipse.jface.viewers.AbstractTableViewer.internalRefresh(AbstractTableViewer.java:616)
at org.eclipse.jface.viewers.AbstractTableViewer.internalRefresh(AbstractTableViewer.java:608)
[...]

PR #2480, opened 2025-09-02
[Gtk]Guard Tree/Table cellDataProc against null item

akurtakov:
Should help with
https://github.com/eclipse-platform/eclipse.platform.swt/issues/2478


laeubi (approved):
Given that the line below this it checks for null it seems required for some cases (but then the check in 282 can probably be removed.

akurtakov:
MacOS failure is unrelated to the change. I'm merging this one and starting a new I-build. If there are still issues I'll revert all and give up on the issue.

PR #2483, opened 2025-09-02
Design improvement for scaling win32 widgets #62

amartya4256:
This PR improves thes design implementation for widgets in win32 by moving away from DPIZoomChangeRegistry to an event-driven design using ZoomChanged event and inheritance.

contributes to


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/TreeItem.java
```diff
@@ -1816,23 +1812,21 @@ String getNameText () {
...
 		for (int index = 0; index < cellFonts.length; index++) {
 			Font cellFont = cellFonts[index];
-			cellFonts[index] = cellFont == null ? null : Font.win32_new(cellFont, treeItem.getNativeZoom());
+			cellFonts[index] = cellFont == null ? null : Font.win32_new(cellFont, getNativeZoom());
 		}
 	}
-	for (TreeItem item : treeItem.getItems()) {
-		DPIZoomChangeRegistry.applyChange(item, newZoom, scalingFactor);
+	for (TreeItem item : getItems()) {
+		item.notifyListeners(SWT.ZoomChanged, event);
```
HeikoKlare, line 1828: Shouldn't this be a directcall to `handleDPIChange` for now?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Widget.java
```diff
@@ -191,6 +190,14 @@ public Widget (Widget parent, int style) {
 	reskinWidget ();
 	notifyCreationTracker();
 	this.setData(DATA_NATIVE_ZOOM, this.nativeZoom);
+	registerDPIChangeListener();
```
HeikoKlare, line 193: With this PR, we stick to the synchronous processing of the event via direct methods calls, so no need to aready add this listener, right?

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CCombo.java
```diff
@@ -2026,19 +2029,24 @@ public boolean traverse(int event){
 	return super.traverse(event);
 }
 
+private void handleCComboDPIChange(Event event) {
+	updateAndRefreshChildren(childWidget -> {
+		childWidget.notifyListeners(SWT.ZoomChanged, event);
+	});
+}
```
HeikoKlare: This intermediate method can be removed by directly making the notifications in the update method.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CCombo.java
```diff
@@ -173,6 +173,9 @@ public CCombo (Composite parent, int style) {
 	}
 
 	initAccessible();
+	addListener(SWT.ZoomChanged, event -> {
+		handleCComboDPIChange(event);
+	});
```
HeikoKlare:
```suggestion
	addListener(SWT.ZoomChanged, this::handleCComboDPIChange);
```

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CCombo.java
```diff
@@ -2026,19 +2029,24 @@ public boolean traverse(int event){
 	return super.traverse(event);
 }
 
+private void handleCComboDPIChange(Event event) {
```
HeikoKlare:
```suggestion
private void handleDPIChange(Event event) {
```

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/StyledText.java
```diff
@@ -10901,30 +10903,20 @@ void updateSelection(int startOffset, int replacedLength, int newLength) {
...
 			caretSet.add(caret);
 		}
 	}
-	caretSet.stream().filter(Objects::nonNull).forEach(caretUpdater);
-
-	styledText.updateCaretVisibility();
-	styledText.setCaretLocations();
-
+	caretSet.stream().filter(Objects::nonNull).forEach(caretToRefresh -> {
+		((Caret) caretToRefresh).notifyListeners(SWT.ZoomChanged, event);
```
HeikoKlare: The set is already types as `Caret`, so why cast to `Caret` here again?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Composite.java
```diff
@@ -1977,13 +1973,12 @@ public String toString() {
...
-	for (Control child : composite.getChildren()) {
-		DPIZoomChangeRegistry.applyChange(child, newZoom, scalingFactor);
+@Override
+void handleDPIChange(Event event, float scalingFactor) {
+	super.handleDPIChange(event, scalingFactor);
+	for (Control child : getChildren()) {
+		child.notifyListeners(SWT.ZoomChanged, event);
 	}
-	composite.redrawInPixels (null, true);
+//	redrawInPixels (null, true);
```
HeikoKlare: Remove

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Widget.java
```diff
@@ -2717,9 +2724,10 @@ int getZoom() {
 	return DPIUtil.getZoomForAutoscaleProperty(nativeZoom);
 }
 
-private static void handleDPIChange(Widget widget, int newZoom, float scalingFactor) {
-	widget.nativeZoom = newZoom;
-	widget.setData(DATA_NATIVE_ZOOM, newZoom);
+void handleDPIChange(Event event, float scalingFactor) {
```
HeikoKlare, line 2727: Why change the signature here instead of sticking to the existing parameters?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/TabFolder.java
```diff
@@ -1128,18 +1127,17 @@ LRESULT wmNotifyChild (NMHDR hdr, long wParam, long lParam) {
...
+	super.handleDPIChange(event, scalingFactor);
+	Display display = getDisplay();
+	if (imageList != null) {
+		display.releaseImageList (imageList);
+		imageList = null;
 	}
-	for (int i = 0; i < tabFolder.getItemCount(); i++) {
-		DPIZoomChangeRegistry.applyChange(tabFolder.items[i], newZoom, scalingFactor);
+	for (int i = 0; i < getItemCount(); i++) {
+		items[i].notifyListeners(SWT.ZoomChanged, event);
```
HeikoKlare, line 1139: Shouldn't this be `items[i].handleDPIChange(...)` for now?

amartya4256:
@HeikoKlare I think we have mixed up things. With this PR we just wanted to move to Event driven architecture and not make it asynchronous. I understand that both the terms go hand in hand but we must note that the event model in SWT is synchronous and if an event is sent by a caller, at first the stub is executed and the caller waits for the event to be processed and then moves forward. Hence, this PR doesn't implement asynchrony. That's also the reason why I called notifyListeners instead of calling handleDpiChange because every widget registers to a listener then they must receive the DPI_CHANGE event also through listeners and not by some other widget calling handleDPIChange on them. The follow up PR would implement asynchrony while calling notifyListeners by the following:

```
void sendZoomChangedEvent(Event event, Shell shell) {
	AtomicInteger handleDPIChangedScheduledTasksCount = (AtomicInteger) event.data;
	handleDPIChangedScheduledTasksCount.incrementAndGet();
	getDisplay().asyncExec(() -> {
		try {
			if(!this.isDisposed()) {
				notifyListeners(SWT.ZoomChanged, event);
			}
		} finally {
			if (handleDPIChangedScheduledTasksCount.decrementAndGet() <= 0) {
				shell.WM_SIZE(0, 0);
			}
		}
	});
}
```

and notifyListeners will be replaced by this method.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Widget.java
amartya4256: We agreed about passing the same event down the same hierarchy and since we have made the method non static now, we do not need Widget anymore but rather the event.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Composite.java
amartya4256: It has been removed in a previous PR. My bad, I would rebase this PR now.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/TreeItem.java
amartya4256: Hmm alright let me think of an order.

amartya4256:
So from what you are proposing and what I think how we could do it, I devised the following order:
1. Remove DPIZoomChangeHandler and use direct handleDPIChange calls for internal classes and use ZoomChangedEvent for Common widget.
2. Move to Event driven model
3. Make the event driven model asynchronous.

(The only problem I see is in the first step. If we handle ZoomChangedEvent for Common widgets like StyledText, etc, they are not scaled in the hierarchical order. I need to test if it could cause any trouble.)

HeikoKlare:
Sounds good, I would just try to split the first step into the adaptation of common/custom widgets and the adaptation of the native part to keep it as small as possible.

> (The only problem I see is in the first step. If we handle ZoomChangedEvent for Common widgets like StyledText, etc, they are not scaled in the hierarchical order. I need to test if it could cause any trouble.)

But since everything stays synchronous, it should not matter whether you call a registry or whether you send an event.

amartya4256:
@HeikoKlare As per our discussion, Replacing CommonWidgetZoomChangeHandler with ZoomChanged Event handler is not possible since the Common widgets need to call for the handleDPIChange for their associated widgets which is win32 specific code and we cannot call notifyListeners on these widgets without making them register to ZoomChanged event as they still use DPIZoomChangeRegistry. Hence, the following is necessary to make it work all together:
1. Register Widget with ZoomChanged event.
2. Implement inheritance based `super.handleDPIChange` for widgets to mimic DPIZoomChangeRegistry's heirarchial behaviour.
3. For handling DPI Change for associated / children widgets, replace DPIZoomChangedRegistry.applyChange with widget.notifyListeners (This still executes synchronously).
4. Move the CommonWidgetZoomChangeHandler methods to the Common Widgets and register them on ZoomChanged events.
5. Remove CommonWidgetZoomChangeHandler  & DPIZoomChangeRegistry/Handler.
6. Adapt the DPI_CHANGE event trigger point in Control to use Shell.notifyListeners

For the next step, i.e. asynchrony, we will replace these notifyListeners with a `Widget:sendZoomChangedEvent` which will call widget.notifyListener asyncly and handle the painting upon DPI scaling completion.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Widget.java
HeikoKlare: I see. But why does this method need `Event` at all? Taking a look at the consumers, wouldn`t `newZoom` be sufficient?

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CCombo.java
```diff
@@ -2026,19 +2027,24 @@ public boolean traverse(int event){
 	return super.traverse(event);
 }
 
+private void handleDPIChange(Event event) {
+	updateAndRefreshChildren(childWidget -> {
+		childWidget.notifyListeners(SWT.ZoomChanged, event);
+	});
```
HeikoKlare: I find this split into two methods with an additional lambda quite hard to read. Why not just ... ?
```suggestion
	text.notifyListeners(SWT.ZoomChanged, event);
	list.notifyListeners(SWT.ZoomChanged, event);
	arrow.notifyListeners(SWT.ZoomChanged, event);
	popup.notifyListeners(SWT.ZoomChanged, event);
```

amartya4256: Yes I missed that one out. Adapting.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Widget.java
akoch-yatta: If the Event is really not needed anywhere, I would propose to keep it simple and leave it to the newZoom

akoch-yatta: But I think I saw some places, where the event is fowarded to other e.g. Items like in Tree, so probably we need that still

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Decorations.java
```diff
@@ -1710,26 +1705,29 @@ LRESULT WM_WINDOWPOSCHANGING (long wParam, long lParam) {
...
-	Image[] images = decorations.getImages();
+	Image[] images = getImages();
 	if (images != null && images.length > 0) {
-		decorations.setImages(images);
+		setImages(images);
 	}
 
-	DPIZoomChangeRegistry.applyChange(decorations.getMenuBar(), newZoom, scalingFactor);
+	Menu menuBar = getMenuBar();
+	if(menuBar != null) {
```
akoch-yatta: Add a space between if and (

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Decorations.java
```diff
@@ -1710,26 +1705,29 @@ LRESULT WM_WINDOWPOSCHANGING (long wParam, long lParam) {
...
+		menuBar.notifyListeners(SWT.ZoomChanged, event);
+	}
 
-	if (decorations.menus != null) {
-		for (Menu menu : decorations.menus) {
-			DPIZoomChangeRegistry.applyChange(menu, newZoom, scalingFactor);
+	if (menus != null) {
+		for (Menu menu : menus) {
+			if(menu != null) {
```
akoch-yatta: Add a space between if and (

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Widget.java
amartya4256: Let me analyze. We definitely need it for the final state (https://github.com/eclipse-platform/eclipse.platform.swt/pull/2520) but it could be possible that we do not need this in this PR.

amartya4256: Having event is important if we do not want to recreate the same event everytime we have to call childWidget.notifyListener to keep the event consistent. We pass the same event generated by shell (from root node), down to the leaf node. This keeps the event consistency and is later used in #2520 by the last leaf to extract some data out of the event (which was generated by the root). Otherwise, it would be the responsibility of the widget to recreate the zoomChanged event which I find unnecessary as the event was already created by the responsible control (Shell) as that's where the callback was received.

HeikoKlare: I see. Thank you for the explanation! Then it makes sense like it is.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
```diff
@@ -4956,12 +4952,22 @@ LRESULT WM_DESTROY (long wParam, long lParam) {
...
+	Event zoomChangedEvent = createZoomChangedEvent(newNativeZoom);
+	notifyListeners(SWT.ZoomChanged, zoomChangedEvent);
 	this.setBoundsInPixels(newBoundsInPixels.x, newBoundsInPixels.y, newBoundsInPixels.width, newBoundsInPixels.height);
 }
 
+private Event createZoomChangedEvent(int zoom) {
+	Event event = new Event();
+	event.type = SWT.ZoomChanged;
+	event.widget = this;
+	event.detail = DPIUtil.getZoomForAutoscaleProperty(zoom);
```
akoch-yatta: This one mustn't be the autoscale zoom. Isn't it stored as native zoom?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Button.java
```diff
@@ -1561,16 +1559,15 @@ LRESULT wmDrawChild (long wParam, long lParam) {
...
-private static void handleDPIChange(Widget widget, int newZoom, float scalingFactor) {
-	if (!(widget instanceof Button button)) {
-		return;
-	}
+@Override
+void handleDPIChange(Event event, float scalingFactor) {
+	super.handleDPIChange(event, scalingFactor);
 	//Refresh the CheckSize
-	button.refreshCheckSize(newZoom);
+	refreshCheckSize(event.detail);
```
HeikoKlare: nit: would be nice to assign this to a `newZoom` value so that one can understand the meaining of this `event.detail` here

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
```diff
@@ -4956,12 +4952,22 @@ LRESULT WM_DESTROY (long wParam, long lParam) {
...
 	this.setBoundsInPixels(newBoundsInPixels.x, newBoundsInPixels.y, newBoundsInPixels.width, newBoundsInPixels.height);
 }
 
+private Event createZoomChangedEvent(int zoom) {
+	Event event = new Event();
+	event.type = SWT.ZoomChanged;
+	event.widget = this;
+	event.detail = zoom;
+	event.doit = true;
+	event.data = new AtomicInteger(0);
```
HeikoKlare: This value seems unused, so please remove.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Widget.java
```diff
@@ -2717,9 +2724,10 @@ int getZoom() {
...
+	int newZoom = event.detail;
```
HeikoKlare, line 2728: I wonder if we need to clone the event at some place, because otherwise if any event consumer changes the `detail` value, it will break the complete rescaling mechanism, won't it?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Shell.java
```diff
@@ -303,6 +302,7 @@ public Shell (Display display, int style) {
 	reskinWidget();
 	createWidget ();
 	this.nativeZoom = DPIUtil.mapDPIToZoom(OS.GetDpiForWindow(this.handle));
+	registerDPIChangeListener();
```
HeikoKlare: Should we document here that this is the only constructor that does not call the `Widget` constructor which registers the listener, which is why it needs to be added here?
ftr: I verified that this is actually the only place.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CCombo.java
```diff
@@ -2026,19 +2027,10 @@ public boolean traverse(int event){
...
-public static void updateAndRefreshChildren(CCombo combo, Consumer<Widget> childUpdater) {
-	childUpdater.accept(combo.text);
-	childUpdater.accept(combo.list);
-	childUpdater.accept(combo.arrow);
-	childUpdater.accept(combo.popup);
+private void handleDPIChange(Event event) {
+	text.notifyListeners(SWT.ZoomChanged, event);
+	list.notifyListeners(SWT.ZoomChanged, event);
+	arrow.notifyListeners(SWT.ZoomChanged, event);
+	popup.notifyListeners(SWT.ZoomChanged, event);
```
HeikoKlare: Can we be sure that they are all != null?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Widget.java
amartya4256: That's why we do not clone event at all but pass the same event down the tree - I wish we could make it final so that no one could modify it. Or maybe use a getter instead for cleaner access.

HeikoKlare: What do you mean by ... ?

> That's why we do not clone event at all but pass the same event down the tree

So it is required that a consumer modifies the `detail` field?

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CCombo.java
amartya4256: There are all initialized in the constructor but they can be nullified so I'll add a null check. Good point.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Shell.java
amartya4256: Done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Widget.java
amartya4256: No, The consumer only consumes (reads) the detail field and then passes the same event instance to its children. Only producer of the event (Shell) sets the detail to the new zoom while creating the event.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
amartya4256: Done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Button.java
amartya4256: Done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Widget.java
HeikoKlare: > The consumer only consumes (reads) the detail field

How can you ensure that? Anyone can register a listener and modify the event. Take this snippet where a listener to a zoom change on a label modifies the scaling of another label (you can imagine the potential consequences in more complex UIs):
```java
		shell.setText("Snippet 1");
		Label label1 = new Label(shell, SWT.None);
		label1.setText("Hello World");
		label1.addListener(SWT.ZoomChanged, e -> {
			e.detail = 250;
		});
		Label label2 = new Label(shell, SWT.None);
		label2.setText("Hello World");

		shell.setSize(800, 200);
		shell.open();
		while (!shell.isDisposed()) {
				display.sleep();
			}
		}
		display.dispose();
	}
```
Started on 100% monitor you get:
image

And moved to 150% monitor you get:
image

amartya4256: Its again about differentiating between what's possible and the semantics. With Event of type ZoomChanged event.detail always represents the new zoom and I can't comprehend why someone would want to set it to a value which is faulty. It's analogous to the win32 public methods which are public yet not considered API. I think we should add a documentation about it to explictly mention that detail represents zoom when used with ZoomChanged.

Talking about Copying the event. It can still be modified and passed on to the other widgets by `someWidget.notifyListeners(SWT.ZoomChanged, event)`

akoch-yatta: I would say, its about risk minimization. Nobody should change that field ever, but we cannot prevent it. If we always copy it before sending it to the next Control, we would minimize the impact.

HeikoKlare: If we agree that we accept the risk of accidental changes to that value, it's of course fine for me. I was more wondering whether we can't be defensive here and copy the event. If someone sends an event to someone else and modifies it before that, this caller expects a changed behavior. But if a completely unrelated consumer of an event gets a modified value, that's not expected. So for me it would be interesting to know about the problems with just copying the event to defensively deal with this risk.

amartya4256: @akoch-yatta We still cannot guarantee that outside SWT because, according to the snippet above, the consumer can implement it the way they want. The question I have is what you actually mean by copying? Simply creating another event instance out of the pre-existing one. Well, inside SWT yes that's safe. But as a developer who registers a ZoomChanged listener, they have the freedom to modify it the way they want. Going forward with Event, the only way I see here is if we properly document this approach.

akoch-yatta: I would think just sending a new Event instance in every internal notifyListeners call should do it.

amartya4256: I am sorry I don't understand what risk we are trying to minimize here. From outside SWT, you can still break it with the similar snippet as @HeikoKlare 's:
```
		shell.setText("Snippet 1");
		Label label1 = new Label(shell, SWT.None);
		label1.setText("Hello World");
		shell.addListener(SWT.ZoomChanged, e -> {
			e.detail = 125;
			for (Control control : shell.getChildren()) {
				control.notifyListeners(SWT.ZoomChanged, e);
			}
		});
		Label label2 = new Label(shell, SWT.None);
		label2.setText("Hello World");

		shell.setSize(800, 200);
		shell.open();
		while (!shell.isDisposed()) {
				display.sleep();
			}
		}
		display.dispose();
	}
```
image

laeubi: I like to note that all events in SWT are modifiable (I think by design) e.g. look at even higher level events like `MouseEvent` where you can change x/y. Also one can change the display and so on.

I'm not aware that SWT copy things all the way what would of course create a lot of garbage objects. So yes one can easily mess up thing but that's not what happens usually.

amartya4256: I agree with @laeubi extending to that I think creating a copy every time we send the event down the widget tree wouldn't have any functional benefits given the design of SWT events.

HeikoKlare: As said before, we can agree that it's fine to not copy the event and accept the risk of modifications by the consumer, in particular given that other places do the same.

Still this conclusion seems to be based on inconsistent understanding of current design and data flows, so we should clarify that first:

> creating a copy every time we send the event down the widget tree wouldn't have any functional benefits given the design of SWT events.

I do not understand that point. If we copy the event, modifications done by consumers will not affect other consumers (except the ones the modified event is _explicitly_ forwarded to). If you take the snippet I provided, the shell would copy the event for each of the labels it passes it to. So as a result, the second label gets a copy of the original event rather than the modified one of the first label. That _is_ a functional benefit.
As said, we don't have to conclude out of that that we need to copy the event, but we need to clarify where the current inconsistent understanding is coming from.

amartya4256: That is right only if the consumers do not send zoomchanged events to other widgets. Consumers like CCombo and StyledText do that and if they were to modify the detail and send the event to other widgets, it will still be broken. The idea of having the same event sent to every widget was to keep the connotation correct. The event is produced at DPI Changed event by the shell and passed down to every widget. If we go with copying the events inside SWT and the consumers do not do it, it is still prone to errors and also it brings design inconsistency.

laeubi: I'm not completely sure what the concerns are.

1. We are talking about quite low level SWT Event and `detail` is marked as "for application use" so already quite generic
2. Actually each low level event should have a higher level "typed" Event that is used in user code
3. I just skimmed through both of them on most of them use public modifiable field and I find a few places where this modification is even used to communicate values to the caller or to even modify / "fix" values there so it is not unusual.
4. There are some exceptions where getters are used or protected fields / classes
5. performance if often a concern

So over all this seem to quite well align with what we have in SWT and we can assume that usually people do not do malicious things (and if they want there are other ways e.g. reflection anyways) and it well aligns with not adding any overhead.

If on the other hand we think we can not trust the eventhandler and really bad things happen, one might simply set critical values again afterwards (here the zoom level) might be something between the lines of do a full copy.

HeikoKlare: Just talked with @amartya4256 and turns out we actually have the same understanding: copying the event (or restoring the zoom value after returning from forwarding the event as just proposed) would prevent the risk of custom listeners accidentally (not necessarily maliciously) modifying the event and breaking follow-up consumers by that. I will of course not prevent this if the modified event is forwarded anywhere else by the custom listener.

As said, I am totally fine with accepting this risk as I do not think anyone will do that maliciously (but at most accidentally) and as you all highlighted we have other similar situations where somethink like that could happen. So let's keep everything as is based on the common understanding that we now have. Thank you all for your input.

HeikoKlare (approved):

PR #2484, opened 2025-09-03
Move SVG rasterizer service to source folder

akoch-yatta:
With the SVG rasterizer contributed as separate project to SWT, running pure Java Snippets did not work out of the box, when the SVG project was added to the workspace. This commit moves the service provider configuration into a separate source folder so it will be properly initialized in this use case as well.


laeubi:
> With the SVG rasterizer contributed as separate project to SWT, running pure Java Snippets did not work out of the box, when the SVG project was added to the workspace. 

Yes this is currently a limitation on how JDT/PDE works when it comes to different resource handling.

> This commit moves the service provider configuration into a separate source folder so it will be properly initialized in this use case as well.

Any reason we need an additional source folder instead of just placing it in the existing one?

HeikoKlare:
> Any reason we need an additional source folder instead of just placing it in the existing one?

I don't think there is any need to do it. It is just to not mess up the actual source code folder with additional resources. Just like done in a default Maven setup with separate "src/java" and "src/resources" source folders.

bundles/org.eclipse.swt.svg/.classpath
```diff
@@ -3,5 +3,6 @@
 	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-17"/>
 	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
 	<classpathentry kind="src" path="src"/>
+	<classpathentry kind="src" path="resources"/>
```
laeubi, line 6: M2e is using  `excluding="**"`  that seem to prevent java files from being processed in such resource only folders.

HeikoKlare: What's the concrete proposal/question here?

laeubi:
```suggestion
	<classpathentry kind="src" path="resources" excluding="**"/>
```

that should prevent any java files from being placed here so you get a real "resource only" folder.

HeikoKlare: Okay, I understand the idea, but (1) it's also a bit weird that if you place a Java file into an actual source folder (no matter whether it's called "resources") is not considered and (2) the pattern appears completely incomprehensible to me, so I am not sure if anywill will understand what this is supposed to do when taking a look at it or when dealing with issues potentially arising from it.

laeubi: You wanted a "resource folder like maven" and that is how maven does it (or more formally m2e). That it is a "source folder" is an implementation detail of JDT because no one has minded to actually implement it like in maven (where resources can also be filtered for example before copied)

HeikoKlare: Just tested and adding `excluding="**"` completely excludes the directory, thus rendering the change completely useless as the service file will not be placed in the `bin` folder anymore.

PR #2486, opened 2025-09-04
Add OfFloat.from methods in Point/Rectangle

amartya4256:
This PR adds Rectangle.OfFloat.from and Point.OfFloat.from methods and removes the Win32DPIUtils.FloatAwareGeometryFactory class to make these methods OS-independent.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Point.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Point.java
@@ -43,7 +43,7 @@ import org.eclipse.swt.widgets.*;
  * @see <a href="http://www.eclipse.org/swt/">Sample code and further information</a>
  */
 
-public sealed class Point implements Serializable permits Point.OfFloat {
+public sealed class Point implements Serializable, Cloneable permits Point.OfFloat {
 
 	/**
 	 * the x coordinate of the point
@@ -119,6 +119,15 @@ public String toString () {
 }
 
 /**
+ * Creates and returns a shallow copy of this {@code Point}.
+ * @since 3.132
+ */
+@Override
+public Point clone() {
+	return new Point(x, y);
+}
+
+/**
  * Instances of this class represent {@link org.eclipse.swt.graphics.Point}
  * objects with the fields capable of storing more precise value in float.
  *
@@ -158,6 +167,21 @@ public static sealed class OfFloat extends Point permits Point.WithMonitor {
 		this.y = Math.round(y);
 		this.residualY = y - this.y;
 	}
+
+	@Override
+	public Point.OfFloat clone() {
+		return new Point.OfFloat(getX(), getY());
+	}
+
+	/**
+	 * Creates a shallow copy of the provided point as a Point.OfFloat instance.
+	 */
+	public static Point.OfFloat from(Point point) {
+		if (point instanceof Point.OfFloat pointOfFloat) {
+			return pointOfFloat.clone();
+		}
+		return new Point.OfFloat(point.x, point.y);
+	}
 }
 
 /**
@@ -187,6 +211,11 @@ public static final class WithMonitor extends Point.OfFloat {
 		this.monitor = monitor;
 	}
 
+	private WithMonitor(float x, float y, Monitor monitor) {
+		super(x, y);
+		this.monitor = monitor;
+	}
+
 	/**
 	 * {@return the monitor with whose context the instance is created}
 	 */
@@ -194,6 +223,10 @@ public static final class WithMonitor extends Point.OfFloat {
 		return monitor;
 	}
 
+	@Override
+	public Point.WithMonitor clone() {
+		return new WithMonitor(getX(), getY(), monitor);
+	}
 }
 
 }
--- a/binaries/org.eclipse.swt.cocoa.macosx.aarch64/META-INF/MANIFEST.MF
+++ b/binaries/org.eclipse.swt.cocoa.macosx.aarch64/META-INF/MANIFEST.MF
@@ -3,7 +3,7 @@ Fragment-Host: org.eclipse.swt;bundle-version="[3.128.0,4.0.0)"
 Bundle-Name: %fragmentName
 Bundle-Vendor: %providerName
 Bundle-SymbolicName: org.eclipse.swt.cocoa.macosx.aarch64; singleton:=true
-Bundle-Version: 3.131.100.qualifier
+Bundle-Version: 3.132.0.qualifier
 Bundle-ManifestVersion: 2
 Bundle-Localization: fragment
 Export-Package: 
--- a/binaries/org.eclipse.swt.cocoa.macosx.x86_64/META-INF/MANIFEST.MF
+++ b/binaries/org.eclipse.swt.cocoa.macosx.x86_64/META-INF/MANIFEST.MF
@@ -3,7 +3,7 @@ Fragment-Host: org.eclipse.swt;bundle-version="[3.128.0,4.0.0)"
 Bundle-Name: %fragmentName
 Bundle-Vendor: %providerName
 Bundle-SymbolicName: org.eclipse.swt.cocoa.macosx.x86_64; singleton:=true
-Bundle-Version: 3.131.100.qualifier
+Bundle-Version: 3.132.0.qualifier
 Bundle-ManifestVersion: 2
 Bundle-Localization: fragment
 Export-Package: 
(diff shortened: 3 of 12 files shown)


bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Point.java
```diff
@@ -158,6 +158,13 @@ public void setY(float y) {
 		this.y = Math.round(y);
 		this.residualY = y - this.y;
 	}
+
+	public static Point.OfFloat from(Point point) {
+		if (point instanceof Point.OfFloat pointOfFloat) {
+			return new Point.OfFloat(pointOfFloat.getX(), pointOfFloat.getY());
```
fedejeanne: Before this PR one simply returned the same point (no instantiation).

Is there a reason for returning `new Point.OfFloat(...)` instead of just the existing `pointOfFloat`?

Same goes for the rectangles.

laeubi: I can only guess it is because of MonitorPoints that also Point.OfFloat  but carry the monitor.

In general I think all these workarounds show that we have a major design flaw with the current way of how measure of position/sizes work and its to complicated to get right I therefore opened:

- https://github.com/eclipse-platform/eclipse.platform.swt/issues/2488

amartya4256: Yes, the from method is used to create a new object from the previous object. The idea is to allow any consumer to make a new Point.OfFloat object from any Point instance. We do not want to share the same instance as the consumer can manipulate this new object and we do not want to affect the old object. The from method semantically creates new object from the provided ones.

laeubi: > We do not want to share the same instance as the consumer can manipulate this new object and we do not want to affect the old object

Should we probably (as far as possible) make the object unmodifable then instead? And instead make thing like "set" mor a copy of a new object?

amartya4256: Since the fields are public, we cannot make it unmodifiable completely, atleast not in a simple way. It will break all the consumers which reset these fields. 

The idea of this PR is to move these helper methods which were implemented in DPIUtils and Win32DPIUtils inside the classes which should actually own it. I did such modifications in the from method only to make it semantically correct.

fedejeanne: Ok so basically you fixed a brittle design by making a defensive copy of the parameter. Something that wasn't happening before but should have happened.

I'm fine with that.

laeubi: > I'm fine with that.

At a minimum the behavior must be documented (e.g. that you always get a fresh copy you are free to modify/use in any way).

Beside that only because on can modify x/y (int) it does not mean we should/must allow this for the `Point.OfFloat` returned here, so an API using Point.OfFloat might consequently simply never use the integer variants.

fedejeanne: > > I'm fine with that.
> 
> At a minimum the behavior must be documented (e.g. that you always get a fresh copy you are free to modify/use in any way).

@amartya4256 can you please mention it in the JavaDocs? Something like "Returns a new instance..."

amartya4256: sure.

amartya4256:
> * Are the methods supposed to be public API or shall they only be used internal for now? In the latter case, they need to be marked `@noreference`. Otherwise, why do they need to be static at all and cannot just be `toFloat()`, as they become public API anyway? But since the classes themselves are currently `@noreference`, this method should probably be as well. In any case, they need a documentation.

We do not need @noreference since the classes are already marked with it. Adding the documentation to the method.

> * Since the methods are supposed to create OfFloat versions out of any other, wouldn't it make sense to place them inside `Point.OfFloat`, so that you write `Point.OfFloat.From(something)` instead of `Point.from(something)` but giving you an OfFloat instead of a pure Point.

They are already inside Point.OfFloat.

> * Is it correct and expected that a `Point.WithMonitor` loses the monitor information when passed to that method?

Adapting the method to call clone internally.

HeikoKlare:
> We do not need @noreference since the classes are already marked with it. Adding the documentation to the method.

> They are already inside Point.OfFloat.

I see. I expected them to be in `Point` as obviously a minor version increment was necessary which I would not have expected for something effectively not being API.

You need to add an `@since` tag to the clone method implementation now.

fedejeanne (approved):

laeubi:
@HeikoKlare @fedejeanne if we move it here, would that not mean they should become public API now? And if not why not move them into the internal packages?

fedejeanne:
> @HeikoKlare @fedejeanne if we move it here, would that not mean they should become public API now? And if not why not move them into the internal packages?

The `OfFloat` classes are already internal API, they are marked with `@noreference`.

laeubi:
> The OfFloat classes are already internal API, they are marked with @noreference.

This is not the same. So if we now move them to a "more public place" and it was promised to make them public API last time, why not make them public API *now* (or consequently keep them internal as is).

`@noreference` is a plain PDE API tools thing that no other tooling likely understands nor is it a requirement for any consumers of SWT.

fedejeanne:
Good point @laeubi , thank you.

@amartya4256 can you draft this one too until you decide how to proceed?

amartya4256:
@laeubi so is your proposal to make both Point.OfFloat and Rectangle.OfFloat public API?

laeubi:
Last time it was added "windows only" and @HeikoKlare stated we should make it sometimes public API (but not now).

So if we now move it "up in the chain" to make it available for other OS, it might be it is "the time" to do so ... that's just my feelings about it before someone uses it by accident.

HeikoKlare:
Making it available for other OSes and making it public API are two different things. Unless it's necessary, we should avoid making things public API as it will make it more difficult to change it afterwards. If there is no specific need to make this public API right now, I would not be in favor of doing so.

laeubi:
>  If there is no specific need to make this public API right now, I would not be in favor of doing so.

The point is that once we put it into commons it seems _there actually is a specific need for it_ as it means the thing has escaped from the (internal) win32 impl it was needed for.

HeikoKlare:
I can only repeat myself:
> Making it available for other OSes and making it public API are two different things.

laeubi:
> I can only repeat myself:
> 
> > Making it available for other OSes and making it public API are two different things.

And I can only repeat myself that the demand was to use it in a `LayoutManger` what is public API that needs to be used by any SWT user so its likely other (outside SWT) will need it if we make it available for "others"...

So using `@norefrefence` for "I don't like to maintain its API" is not sufficient. This is usually because one needs to expose a type for technical reasons but it should be a last resort and not the norm (e.g. assume a library requires the type to be public). Also in P2 we have had the case already that these become defacto API because people simply used it (and either did not noticed or suppressed the warning or did not use API tools in their setups).

amartya4256:
I mean this PR was intended to move the methods to the right class since having a utility in Win32DPIUtils in an inner factory class doesn't make much sense and the classes (Point.OfFloat and Rectangle.OfFloat) themselves seem to be the right candidate for such methods. So from a refactoring POV it seems to be a right thing to do. But what I believe the the discussion here is about if it should be an API or a @noreference. Am I right?

I think since we intend for it to be not used outside by any consumer outside SWT, there's no need for it to be public. We also have a lot of static methods in many Common classes which have noreference methods. For instance we have StyledText#updateAndRefreshCarets, Drawable#internal_new_GC, ImageData#internal_new

The same mistakes can be made with these methods as well. We intend to not encourage usage of OfFloat outside SWT at least for now.

fedejeanne:
Since there are no more comments and later PRs can improve upon this one, merging.

iloveeclipse:
@amartya4256 , @fedejeanne : please next time you change SWT bundle versions, make sure that host bundle and all fragments have same version. If this doesn't happen, SDK build will fail as we saw it today: https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3347.

fedejeanne:
Oh, I'm sorry @iloveeclipse , I totally missed the necessary change in the host bundle. Thank you for fixing it!

For future reference @amartya4256 : we forgot both the `pom.xml` and the `MANIFEST.MF` (see https://github.com/eclipse-platform/eclipse.platform.swt/pull/2525)

Issue #2485, opened 2025-09-04
GTK4: Combo copy and paste test failure


raghucssit:
I have taken this failure from https://github.com/eclipse-platform/eclipse.platform.swt/pull/2124

Both copy and paste tests fail on RHEL-10 with GTK4.

paste:
```
org.junit.ComparisonFailure: expected:<1[23]456> but was:<1[]456>
	at org.junit.Assert.assertEquals(Assert.java:117)
	at org.junit.Assert.assertEquals(Assert.java:146)
	at org.eclipse.swt.tests.junit.Test_org_eclipse_swt_widgets_Combo.test_paste(Test_org_eclipse_swt_widgets_Combo.java:504)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	Suppressed: java.lang.Throwable: Screenshot written to /tmp/org.eclipse.swt.tests.junit.Test_org_eclipse_swt_widgets_Combo.test_paste.png
		at org.eclipse.test.Screenshots$ScreenshotOnFailure.failed(Screenshots.java:41)
```

copy:
```
org.junit.ComparisonFailure: expected:<[23]123456> but was:<[]123456>
	at org.junit.Assert.assertEquals(Assert.java:117)
	at org.junit.Assert.assertEquals(Assert.java:146)
	at org.eclipse.swt.tests.junit.Test_org_eclipse_swt_widgets_Combo.test_copy(Test_org_eclipse_swt_widgets_Combo.java:239)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	Suppressed: java.lang.Throwable: Screenshot written to /tmp/org.eclipse.swt.tests.junit.Test_org_eclipse_swt_widgets_Combo.test_copy.png
		at org.eclipse.test.Screenshots$ScreenshotOnFailure.failed(Screenshots.java:41)
```

raghucssit:
I will investigate this issue..

akurtakov:
There is https://github.com/eclipse-platform/eclipse.platform.swt/pull/2444 for this one but it needs improvements.

raghucssit:
Manually this cannot be reproduced As I have tried the steps as per test case.
```
@Test
public void test_copy() {
	if (SwtTestUtil.isCocoa) {
		// TODO Fix Cocoa failure.
		if (SwtTestUtil.verbose) {
			System.out
[...]

PR #2489, opened 2025-09-08
Spin the event loop to process the paste operation completely.

raghucssit:
Manually when copy and paste done on combo it works as expected. But while calling copy and paste API on combo there is an inconsistency. It is due to pending UI events before we check text on combo.

see https://github.com/eclipse-platform/eclipse.platform.swt/issues/2485


tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_widgets_Combo.java
```diff
@@ -236,6 +237,11 @@ public void test_copy() {
 	combo.copy();
 	combo.setSelection(new Point(0,0));
 	combo.paste();
+	// Spin the event loop to let GTK process the clipboard + entry update
+	Display display = Display.getDefault();
```
laeubi: use `combo.getDisplay()` here there is no guarantee that the default display is the display used by the combo.

laeubi:
@raghucssit should we enhance javadoc to mention that calling copy / paste will not _immediate_ update the state and one should possibly use asyncExec wif accessing the control is desired (that would probably also work for the test to use async exec but it of course complicates the test assertions maybe).

raghucssit:
> @raghucssit should we enhance javadoc to mention that calling copy / paste will not _immediate_ update the state and one should possibly use asyncExec wif accessing the control is desired (that would probably also work for the test to use async exec but it of course complicates the test assertions maybe).

I have updated copy/paste doc.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Combo.java
```diff
@@ -494,6 +494,12 @@ Point computeNativeSize (long h, int wHint, int hHint, boolean changed) {
  * <p>
  * The current selection is copied to the clipboard.
  * </p>
+ * <strong>Note:</strong> On GTK4, clipboard updates may be asynchronous. This
+ * means that the new clipboard content may not be immediately available right
+ * after calling this method. To ensure the update is visible, spin the event
+ * loop (e.g., using {@link Display#readAndDispatch()} or
+ * {@link Display#asyncExec(Runnable)} before accessing the clipboard.
+ * </p>
```
laeubi:
```suggestion
 * <strong>Note:</strong> Copy data to the Clipboard may be asynchronous. This
 * means that the new clipboard content may not be immediately available right
 * after calling this method. To ensure the update is visible, use
 * {@link Display#asyncExec(Runnable)} before accessing the clipboard data.
 * </p>
```

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Combo.java
```diff
@@ -494,6 +494,12 @@ Point computeNativeSize (long h, int wHint, int hHint, boolean changed) {
  * <p>
  * The current selection is copied to the clipboard.
  * </p>
+ * <strong>Note:</strong> On GTK4, clipboard updates may be asynchronous. This
```
iloveeclipse: Starting `<p>` tag is missing

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Combo.java
```diff
@@ -1955,6 +1961,13 @@ long paintSurface () {
  * The selected text is deleted from the widget
  * and new text inserted from the clipboard.
  * </p>
+ * <p>
+ * <strong>Note:</strong> On GTK4, pasting occurs asynchronously. The widget
+ * text may not reflect the updated value immediately after calling this method.
+ * The new text will appear once pending events are processed in the event loop.
+ * If you need to assert the new value right after calling {@code paste()}, use
+ * {@link Display#readAndDispatch()} or {@link Display#asyncExec(Runnable)}.
+ * </p>
```
laeubi:
```suggestion
 * <p>
 * <strong>Note:</strong> Pasting data to controls may occurs asynchronously. The widget
 * text may not reflect the updated value immediately after calling this method.
 * The new text will appear once pending events are processed in the event loop.
 * Use {@link Display#asyncExec(Runnable)} before accessing <code>getText()</code>.
 * </p>
```

raghucssit: Thanks for the suggestion. I have updated the doc..

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Combo.java
raghucssit: Fixed.

iloveeclipse:
Known GTK4 tests are failing, the fixed Combo tests are passing.

jonahgraham:
> > @raghucssit should we enhance javadoc to mention that calling copy / paste will not _immediate_ update the state and one should possibly use asyncExec wif accessing the control is desired (that would probably also work for the test to use async exec but it of course complicates the test assertions maybe).
> 
> I have updated copy/paste doc.

I believe this update to the Javadoc (and API contract) is incorrect. Having copy/paste/cut move from Sync to Async behaviour should probably be considered a breaking change in SWT.

I am creating a new issue for further discussions - please see and share your thoughts in https://github.com/eclipse-platform/eclipse.platform.swt/issues/2598

PR #2492, opened 2025-09-09
Add Snippet 386: Demonstrates Cursor constructors and DPI scaling in SWT

ShahzaibIbrahim:
This snippet demonstrates several ways to construct and use cursors in an SWT application, including system cursors, custom image cursors, and DPI-aware cursors using ImageDataProvider. The user interface allows interactive selection of the cursor construction method, and updates the shell’s cursor in real time based on the current choice. The snippet displays the current system zoom level (DPI) and the expected cursor size, and draws reference ticks to help visualize scaling effects. It also responds to system DPI changes, ensuring accurate display and behavior at different zoom levels. The code is structured for clarity and modularity, making it suitable for direct use with the SWT Snippet Explorer.

**Preview**

image


examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet386.java
```diff
@@ -0,0 +1,225 @@
...
+					case 200:
+						return createSolidColorImageData(32, green);
+					case 250:
+						return createSolidColorImageData(40, green);
+					case 300:
+						return createSolidColorImageData(48, green);
+					case 350:
+						return createSolidColorImageData(56, green);
+					default:
+						return createSolidColorImageData(64, green);
```
arunjose696: just a nit: to avoid switch case
```suggestion
int size = Math.round(IMAGE_SIZE_IN_POINTS * (zoom / 100f));
return createSolidColorImageData(size, green);
```

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet386.java
```diff
@@ -0,0 +1,225 @@
...
+			this.combo = combo;
+		}
+
+		void updateCursor() {
+			int selection = combo.getSelectionIndex();
+			Cursor oldCursor = shell.getCursor();
+			if (oldCursor != null && !oldCursor.isDisposed()) {
+				oldCursor.dispose();
+			}
+			Cursor cursor = createCursor(selection);
```
arunjose696, line 179: Though not a major issue with snippets it might be a good idea to dispose this cursor too on cleanup

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet386.java
ShahzaibIbrahim: I would like to keep the switch case as it is convenient for someone to see what size we are actually drawing. The purpose of snippet is to understand the context.

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet386.java
```diff
@@ -0,0 +1,232 @@
...
+						return createSolidColorImageData(32, green);
+					case 250:
+						return createSolidColorImageData(40, green);
+					case 300:
+						return createSolidColorImageData(48, green);
+					case 350:
+						return createSolidColorImageData(56, green);
+					default:
+						return createSolidColorImageData(64, green);
+					}
```
HeikoKlare: Why not something like...?
```suggestion
					if (zoom % 50 == 0) {
					   return createSolidColorImageData(16 * zoom / 100, green);
					}
					return null;
```

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet386.java
```diff
@@ -0,0 +1,232 @@
...
+				ImageData mask = new ImageData(IMAGE_SIZE_IN_POINTS, IMAGE_SIZE_IN_POINTS, 1,
+						new PaletteData(new RGB[] { new RGB(0, 255, 0) }));
+				return new Cursor(display, source, mask, 0, 0);
+			}
+
+			case 2:
+				return new Cursor(display, createSolidColorImageData(IMAGE_SIZE_IN_POINTS, new RGB(128, 0, 128)), 0, 0);
+
+			case 3: {
+				RGB green = new RGB(0, 255, 0);
```
HeikoKlare, line 216: Why not define and use `green` also at the other places in the method?

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet386.java
```diff
@@ -0,0 +1,232 @@
...
+
+		int pixel = palette.getPixel(color);
+		for (int y = 0; y < size; y++)
+			for (int x = 0; x < size; x++)
+				imageData.setPixel(x, y, pixel);
+
+		return imageData;
+	}
+
+	// --- Cursor management ---
```
HeikoKlare:
```suggestion
```

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet386.java
```diff
@@ -0,0 +1,232 @@
...
+	}
+
+	private static void eventLoop(Display display, Shell shell) {
+		while (!shell.isDisposed()) {
+			if (!display.readAndDispatch())
+				display.sleep();
+		}
+	}
+
+	// --- Utility: creates solid color square ImageData ---
```
HeikoKlare: Replace such comments with properly named classes/methods (if not already there).

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet386.java
```diff
@@ -0,0 +1,232 @@
...
+	public static ImageData createSolidColorImageData(int size, RGB color) {
+		PaletteData palette = new PaletteData(0xFF0000, 0x00FF00, 0x0000FF);
+		ImageData imageData = new ImageData(size, size, 24, palette);
+
+		int pixel = palette.getPixel(color);
+		for (int y = 0; y < size; y++)
```
HeikoKlare: Please enclose expressions into blocks/braces.

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet386.java
```diff
@@ -0,0 +1,232 @@
...
+ * For a list of all SWT example snippets see
+ * http://www.eclipse.org/swt/snippets/
+ * </p>
+ */
+public class Snippet386 {
+
+	private static final int IMAGE_SIZE_IN_POINTS = 16;
+
+	public static void main(String[] args) {
+		System.setProperty("swt.autoScale.updateOnRuntime", "true");
```
HeikoKlare: This should be something optional, as it would only work on Windows.

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet386.java
```diff
@@ -0,0 +1,232 @@
...
+
+	private static void addPaintTicks(Shell shell) {
+		shell.addPaintListener(event -> {
+			drawTicks(event.gc);
+		});
+	}
+
+	private static void drawTicks(GC gc) {
+		@SuppressWarnings("restriction")
+		int deviceZoom = DPIUtil.getDeviceZoom();
```
HeikoKlare: This looks rather flaky. Can't you do something like `shell.getMonitor().getZoom()` or the like? And even without that, is `getDeviceZoom()` really correct here or shouldn't it be `getNativeZoom()`?

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet386.java
```diff
@@ -0,0 +1,232 @@
...
+		Combo combo = createConstructorCombo(shell);
+
+		addZoomChangedListener(shell, zoomLabel);
+		addPaintTicks(shell);
+
+		CursorManager cursorManager = new CursorManager(display, shell, combo);
+		combo.addListener(SWT.Selection, e -> cursorManager.updateCursor());
+		cursorManager.updateCursor();
+
+		shell.setSize(300, 600);
```
HeikoKlare: Is there a specific reasons for this size? Of course, there is to "correct" size, but the shell is very narrow and high, not even allowing the shell title to be fully displayed. Wouldn't it be better to have more like a quadratic shell with the cursor pattern being centered in the shell instead of using magic values like 300 and 100 in the positioning?

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet386.java
```diff
@@ -0,0 +1,232 @@
...
+		zoomLabel.setLayoutData(new GridData(SWT.FILL, SWT.CENTER, true, false));
+		setZoomLabelText(zoomLabel);
+		return zoomLabel;
+	}
+
+	private static void setZoomLabelText(Label label) {
+		@SuppressWarnings("restriction")
+		int zoom = DPIUtil.getDeviceZoom();
+		int expectedCursorSize = Math.round(IMAGE_SIZE_IN_POINTS * (zoom / 100f));
+		label.setText("Current zoom: " + zoom + "% Expected Cursor Size = " + expectedCursorSize);
```
HeikoKlare: I would be in favor of using one line per information to make it easier to read.

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet386.java
HeikoKlare: Didn't see this comment when also doing a review: it's the value shown in the UI what is used to understand what we are actually drawing rather than taking a look into the code?

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet386.java
ShahzaibIbrahim: Only because I wanted others to be able to see easily what size is set for each zoom. I could do as you said but I wanted something easier to read.

HeikoKlare: Okay, but it's even not correct: if you run the snippet at 125% zoom you will get a 64px cursor (the default one) and the UI says it expected 20px. I don't think that's what we want to have.

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet386.java
```diff
@@ -0,0 +1,232 @@
...
+		private final Combo combo;
+
+		CursorManager(Display display, Shell shell, Combo combo) {
+			this.display = display;
+			this.shell = shell;
+			this.combo = combo;
+
+			// Dispose the final cursor when the Display is disposed
+	        display.addListener(SWT.Dispose, e -> {
+	            Cursor cursor = shell.getCursor();
```
HeikoKlare: This throws an exception as the shell is already disposed here. Do we need to dispose the cursor at all if the display is disposed?

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet386.java
ShahzaibIbrahim: I could add other quarter values with their expected sizes.

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet386.java
ShahzaibIbrahim: You mean for other cursors as well? I use different colors for each cursor so user get to see the visual change in cursor when selecting other constructor from combo.

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet386.java
HeikoKlare: Yes, that's possible. It will break in less cases then (namely if you use exact scale value or if you use a monitor zoom higher than 400). But honestly I don't get why we need all that complexity. There is a simple formula to calculate the size, so I don't see why we need a verbose lookup table for specific values of that formula. As I stated in https://github.com/eclipse-platform/eclipse.platform.swt/pull/2492#discussion_r2341289474, the user of the snippet can easily grasp the expected size in the UI instead of taking a look into the code.
But don't get me wrong: I don't have a strong opinion here, just a bit of confusion about the necessity :-)

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet386.java
ShahzaibIbrahim: No I just checked, we don't need to dispose if the display is disposed. I will keep my logic to dispose the old cursors when new ones are created.

ShahzaibIbrahim:
@HeikoKlare I have addressed all the comments, also made some UI improvements and it is now ready for review.

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet386.java
```diff
@@ -0,0 +1,214 @@
...
+ * http://www.eclipse.org/swt/snippets/
+ * </p>
+ */
+public class Snippet386 {
+
+	private static final int IMAGE_SIZE_IN_POINTS = 16;
+
+	public static void main(String[] args) {
+		Display display = new Display();
+		Shell shell = createShell(display);
```
HeikoKlare, line 58: Why does `createShell` create a label for constructor selection and a separate method creates the constructors combo that belongs to it? Shouldn't the label and the combo be created together?

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet386.java
```diff
@@ -0,0 +1,214 @@
...
+
+		addZoomChangedListener(shell, zoomLabel);
+		Group section = new Group(shell, SWT.NONE);
+		section.setText("Scale");
+		section.setLayoutData(new GridData(SWT.FILL, SWT.TOP, true, true));
+		section.setLayout(new GridLayout(2, false));
+
+		GridData gd = new GridData(SWT.FILL, SWT.TOP, true, false);
+		gd.heightHint = 100;   // fixed height in pixels
+		section.setLayoutData(gd);
```
HeikoKlare: Why not keep it simple with something like?
```suggestion
        section.setLayoutData(new GridData(SWT.FILL, SWT.FILL, true, true));
		section.setLayout(new FillLayout());
```

HeikoKlare (commented):
I think the for a snippet the code is okay now. Please auto-format the code as it does currently not conform to the auto-formatting settings. I have some minor proposals for improvement, which you could also apply with the formatting.

As the snippet perfectly shows the wrong cursor size when using the `(Device, ImageData, ImageData, int, int)` constructor: do we already have a ticket for analyzing/improving that?

ShahzaibIbrahim:
> As the snippet perfectly shows the wrong cursor size when using the `(Device, ImageData, ImageData, int, int)` constructor: do we already have a ticket for analyzing/improving that?

I already have a ticket for that. See, https://github.com/vi-eclipse/Eclipse-Platform/issues/441

HeikoKlare:
@ShahzaibIbrahim can we merge this or do you plan any further adaptations?

ShahzaibIbrahim:
> @ShahzaibIbrahim can we merge this or do you plan any further adaptations?

We can merge it. I have made the adaptations already.

ShahzaibIbrahim:

Issue #2461, opened 2025-08-27
[Win32] FontMetrics.getAverageCharacterWidth should return unrounded value


wahlbrink:
At moment FontMetrics.getAverageCharacterWidth always returns rounded integer values. This leads to poor estimates of text columns (when using scaling) with issues like here: https://github.com/eclipse-platform/eclipse.platform.ui/issues/2836

**Expected behavior**
FontMetrics.getAverageCharacterWidth should return unrounded values.

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 

Eclipse Version tested: 2025-06 (4.36), I20250528-1830

HeikoKlare:
I can confirm this issue and I can confirm that from first experiments that improving `FontMetrics.getAverageCharacterWidth` would resolve the issue with wrong print margins being displayed (and maybe more).

It's important to note that issues should only occur when a monitor zoom != 100% is used: the problem with the implementation is that the conversion from pixel to point usually leads to fractional results which are unnecessarily rounded here.

@wahlbrink thank you for reporting! Since you mentioned "issues like ...", do you know about further user-experienced issues caused by this?

wahlbrink:
> do you know about further user-experienced issues caused by this?

Similar issue (computation in the other direction):
We try to estimate the width (in chars) of the console
https://gitlab.eclipse.org/eclipse/statet/statet/-/blob/main/r/org.eclipse.statet.r.console.ui/src/org/eclipse/statet/internal/r/console/ui/actions/AdjustWidthHandler.java#L149

HeikoKlare:
Do you plan to propose a PR?

PR #2496, opened 2025-09-10
[Win32] Return unrounded value in FontMetrics.getAverageCharacterWidth

wahlbrink:
Fixes: https://github.com/eclipse-platform/eclipse.platform.swt/issues/2461


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/FontMetrics.java
```diff
@@ -108,7 +108,7 @@ public int getAscent() {
  * @since 3.107
  */
 public double getAverageCharacterWidth() {
-	return getAverageCharWidth();
+	return DPIUtil.pixelToPoint((float)handle.tmAveCharWidth, getZoom());
```
HeikoKlare: Shouldn't we add a method converting doubles instead of trimming to float and then converting back?

Issue #2504, opened 2025-09-12
On win32,  if GC drawable is not autoscalable GC.drawImage can cut of parts of the image


Tarmean:
**Describe the bug**

GC.drawImage scales the image data but not the image source coordinates if the gc is not autoScalable.

- Use a GC with `gc.drawable.isAutoScalable()==false`
- Call `gc.drawImage(image, 0, 0, 50, 50, 0, 0, 100, 100);`

For this, `DrawScalingImageToImageOperation.calculateZoomForImage` gives a zoom of 200.

Because the gc drawable is not autoScalable, the image source coordinates are not scaled by 200
https://github.com/eclipse-platform/eclipse.platform.swt/blob/1094ebd97ea0a859041103befc65b529724081a3/bundles/org.eclipse.swt/Eclipse%20SWT/win32/org/eclipse/swt/graphics/GC.java#L1158

Because the image is scaled independent of isAutoScalable, the coordinates get out of sync
https://github.com/eclipse-platform/eclipse.platform.swt/blob/1094ebd97ea0a859041103befc65b529724081a3/bundles/org.eclipse.swt/Eclipse%20SWT/win32/org/eclipse/swt/graphics/GC.java#L1201

**To Reproduce**
A standalone SWT snippet to reproduce the behavior (preferable)

```
package test.package;

import org.eclipse.swt.graphics.Image;
import org.eclipse.swt.graphics.Rectangle;
import org.eclipse.swt.layout.FormAttachment;
import org.eclipse.swt.layout.FormData;
import org.eclipse.swt.layout.FormLayout;
import org.eclipse.swt.widgets.Canvas;
import org.eclipse.swt.widgets.Display;
import org.eclipse.swt.widgets.Shell;

public class ReproDemo {
	static final float targetScale = 2f;
	public static void main(String[] args) {
		Display display = new Display();
		Shell shell = new Shell(display);

		Image image = new Image(display, ReproDemo.class.getResourceAsStream("/50x50.png"));
		Rectangle bounds = image.getBounds();

		shell.setLayout(new FormLayout());
		Canvas canvas = new Canvas(shell, 0) {
			@Override
			public boolean isAutoScalable() {
				return false;
			}
		};
		FormData formData = new FormData();
		formData.left = new FormAttachment(0,0);
		formData.top = new FormAttachment(0,0);
		formData.width = Math.round(bounds.width * targetScale);
		formData.height = Math.round(bounds.height * targetScale);
[...]

PR #2507, opened 2025-09-15
Set zoom for source image as GCzoom when the GC is not autoscalable

arunjose696:
The calculated image zoom for the source image should be the same as the GCzoom(100) when the drawable of GC is not scalable 

Steps to reproduce 
Run the snippet in https://github.com/eclipse-platform/eclipse.platform.swt/issues/2504
With source and destination at 1:2 the image would be cropped without this change

Fixes: #2504


fedejeanne:

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -1140,6 +1140,9 @@ private int calculateZoomForImage(int gcZoom, int srcWidth, int srcHeight, int d
 			// unscaled images can use the GC zoom
 			return gcZoom;
 		}
+		if(!drawable.isAutoScalable()){
```
HeikoKlare:
```suggestion
		if (!drawable.isAutoScalable()){
```

arunjose696:
> Would it be possible to add a test for this to ensure that we do not accidentally run into the same situation again? Something like a simplified version of the snippet in #2504 with a very simple image in which you can easily assert based on specific pixel values if the image was properly sized or not.

Added test

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_GC.java
```diff
@@ -246,6 +248,60 @@ public void test_dispose() {
...
+    Canvas canvas = new Canvas(shell, SWT.NONE) {
+        @Override
+        public boolean isAutoScalable() {
+            return false;
+        }
+    };
+    FormData formData = new FormData();
+    formData.width = Math.round(bounds.width * targetScale);
+    formData.height = Math.round(bounds.height * targetScale);
+    canvas.setLayoutData(formData);
```
HeikoKlare: Can we get rid of the FormData/FormLayout dependencies? The test seems to just work fine as well if we remove setting the layout and replace this (and subsequent usages of `formData.width/height`):
```suggestion
    int canvasWidth = Math.round(bounds.width * targetScale);
    int canvasHeight = Math.round(bounds.width * targetScale);
    canvas.setSize(canvasWidth, canvasHeight);
```

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_GC.java
```diff
@@ -246,6 +248,60 @@ public void test_dispose() {
...
+    canvas.addPaintListener(e -> {
+        e.gc.drawImage(image, 0, 0, bounds.width, bounds.height,
+                       0, 0, formData.width, formData.height);
+    });
+
+    shell.open();
+    while (display.readAndDispatch()) {
+    }
```
HeikoKlare: Why is this necessary? Can't we remove that?

HeikoKlare (commented):
The change looks fine and it properly fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2504
I have tested with the snippet provided in that issue.

Thank you also for adding a proper regression test. I only have minor comments regarding that test.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_GC.java
arunjose696: This was added to ensure that the PaintListener actually ran and the canvas was fully painted before you call copyArea. Without it, there’s a slight risk that the paint hasn’t completed yet

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_GC.java
arunjose696: Removed FormData/FormLayout dependencies

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_GC.java
HeikoKlare: @arunjose696 Can you explain where this risk come from? `shell.open()` will send a paint event and trigger the paint listener of the canvas, so I would expect the `copyArea()` operation to work. What will the `readAndDispatch()` do that leads to "fully" painting the canvas? Is there any way to produce the issue you describe?

arunjose696: The reasoning to add readAndDispatch() was to consider the case when e.gc.drawImage()'s  draw request to the OS, might not have completed by the time gcCopy.copyArea() was called. Using readAndDispatch() was to ensure all OS events are processed, so there’s no pending work left. I noticed this problem only when running the test in debug mode with breakpoints in drawImage and copyArea calls. When running the test normally multiple times even with large image sizes, the issue doesn’t occur. Thus I have now removed the readAndDispatch() call.

iloveeclipse:
Caused regression in https://github.com/eclipse-platform/eclipse.platform.swt/issues/2557

arunjose696:
> Caused regression in #2557

This regression is in the test added in the same commit 

I think the reason here would be e.gc.drawImage()'s  draw request to the OS, might not have completed by the time gcCopy.copyArea() was called, in the test. This was a [consideration](https://github.com/eclipse-platform/eclipse.platform.swt/pull/2507#discussion_r2378259773) which was thought of,  to add readAndDispatch() call but could see there were no failures in the CI for linux when this PR was made so I decided to skip the call.

PR #2509, opened 2025-09-15
Extend ImageLoader to be able to load images at a given height and width

arunjose696:
Extend ImageLoader to able to load images at a given target height and width in addition to the current behavior of only being able to load images at a target zoom. This is required to load images in svg format at custom height and width.

The extensions added in this change is used in https://github.com/eclipse-platform/eclipse.platform.swt/pull/2526 That PR also contains snippets through which these changes are tested.

This depends on https://github.com/eclipse-platform/eclipse.platform.swt/pull/2514, Only last commit is relevant for this PR


arunjose696:
> Overall, this change looks good to me. I am just wondering if the term "target" in all the methods/variables is appropriate and helpful. In the zoom case, we have the distinction between `fileZoom` and `targetZoom` as we need to know which zoom an image represents. But for the size case there is no such distinction, as you just need to tell the size you want to have. So I think we should maybe remove all the "target" and just call everything `width`, `height`, `loadBySize` etc.

Done

fedejeanne (approved):

Michael5601:
Thank you for finishing the prototype. It's very nice to see all the work getting merged now :). 
And thank you @HeikoKlare for remembering to mention my name on the changes.

PR #2510, opened 2025-09-15
Improve Display test assert message.

raghucssit:


raghucssit:
@laeubi Please add GTK4 label we can see what are the actual returned coordinates on build machine.

Issue #2512, opened 2025-09-15
Rendering issue if "Show code minings for problem annotations" is enabled


tobiasmelcher:
**Describe the bug**
- Enable the text editor feature "Show code minings for problem annotations" in the preferences

<img width="845" height="939" alt="Image" src="https://github.com/user-attachments/assets/068e2836-6564-4cdb-bc4a-dbdf335ea77d" />

- open a Java file with lots of errors. 
For example, create a Java file with a main method containing 150 lines of `System.out.println()` statements without the terminating semicolon (`;`).
- Use the mouse wheel to scroll down through the source code.
Eventually, you will observe a rendering issue: code minings displaying error messages are drawn on top of the source code, making it difficult to read.

Please refer to the following screencast for a visual demonstration:

https://github.com/user-attachments/assets/e73d8dd3-d787-4826-b37e-056a07c03fe3

**Expected behavior**
Code minings are drawn correctly in the source code editor

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [x] All OS
- - [ ] Windows
- - [ ] Linux
- - [ ] macOS 

2. Additional OS info (e.g. OS version, Linux Desktop, etc)
I could reproduce the rendering issue on Windows and MacOS

PR #2513, opened 2025-09-15
StyledText#setLineVerticalIndent: fix overlapping code minings in editor

tobiasmelcher:
Ensure problem annotation code minings are rendered below source lines to prevent visual overlap during vertical scrolling.

Fixes: eclipse-platform/eclipse.platform.swt#2512

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/StyledText.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/StyledText.java
@@ -4265,11 +4265,19 @@ public int getOrientation () {
  */
 int getPartialBottomIndex() {
 	if (isFixedLineHeight()) {
-		int lineHeight = renderer.getLineHeight();
-		int partialLineCount = Compatibility.ceil(clientAreaHeight, lineHeight);
-		return Math.max(0, Math.min(content.getLineCount(), topIndex + partialLineCount) - 1);
+		return getPartialBottomIndexFixedLineHeight();
 	}
 	return getLineIndex(clientAreaHeight - bottomMargin);
+}
+/**
+ * Returns the index of the last partially visible line assuming a fixed line height.
+ *
+ * @return index of the last partially visible line.
+ */
+private int getPartialBottomIndexFixedLineHeight() {
+	int lineHeight = renderer.getLineHeight();
+	int partialLineCount = Compatibility.ceil(clientAreaHeight, lineHeight);
+	return Math.max(0, Math.min(content.getLineCount(), topIndex + partialLineCount) - 1);
 }
 /**
  * Returns the index of the first partially visible line.
@@ -9315,7 +9323,8 @@ public void setLineVerticalIndent(int lineIndex, int verticalLineIndent) {
 	}
 	int initialTopPixel = getTopPixel();
 	int initialTopIndex = getPartialTopIndex();
-	int initialBottomIndex = getPartialBottomIndex();
+	// use getPartialBottomIndexFixedLineHeight to include more additional lines to fix scrolling issue eclipse-platform/eclipse.platform.swt#2512
+	int initialBottomIndex = getPartialBottomIndexFixedLineHeight();
 	int verticalIndentDiff = verticalLineIndent - previousVerticalIndent;
 	renderer.setLineVerticalIndent(lineIndex, verticalLineIndent);
 	this.hasVerticalIndent = verticalLineIndent != 0 || renderer.hasVerticalIndent();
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/StyledText.java
```diff
@@ -9330,27 +9330,33 @@ public void setLineVerticalIndent(int lineIndex, int verticalLineIndent) {
...
-			}
-		} else { // move as few lines as possible
-			if (lineIndex - getTopIndex() < getBottomIndex() - lineIndex) {
-				setTopPixel(initialTopPixel + verticalIndentDiff);
-			} else {
-				// redraw below
-				redrawLines(lineIndex, getPartialBottomIndex() - lineIndex + 1, true);
-			}
+	}
+	redrawLines(lineIndex, initialTopIndex, initialBottomIndex, initialTopPixel, verticalIndentDiff);
```
mickaelistria: IIRC, this block of code can be very expensive. I see it's now going to be executed more often than usual. Instead of fully removing the guard, isn't there a way to refine it?
However, I have not actually tried it on real big files. If you did so and are confident performance remains good, I'm fine with a merge, but code itself makes me unsure about it.

tobiasmelcher: I reduced it to "resetCache(lineIndex, 1);". Do you now have less concerns regarding performance?

mickaelistria: resetCache per se is the operation that triggers a lot of time recomputing sizes and so on. So yes, adding this can be a concern for performance.
However, as mentioned, if you've happily tested on a big enough file, I would trust my concerns are vain and would be fine seeing this patch merged.

tobiasmelcher: I created following performance test to compare the runtime of the `setLineVerticalndent` method. 

```
@Test
public void testPerformanceSetLineVerticalIndent() throws Exception{
	Shell shell = new Shell();
	StyledText text = new StyledText(shell, SWT.V_SCROLL | SWT.H_SCROLL);
	shell.setSize(500,500);
	shell.setVisible(true);
	StringBuilder str = new StringBuilder();
	for (int i=0;i<10_000;i++) {
		str.append("Line ").append(i).append("\n");
	}
	int lineHeight = text.getLineHeight();
	text.setText(str.toString());
	while(Display.getDefault().readAndDispatch()) {
	}

	long start = System.currentTimeMillis();
	for (int i=0;i<10_000;i++) {
		text.setLineVerticalIndent(i, lineHeight*2);
	}
	long diff = System.currentTimeMillis()-start;
	System.out.println(diff);
}
```
On my machine, I see following numbers. 
3609ms in the old code base
17318ms with the changes of this pull request where resetCache is called much more often. Factor 4.8 slower is quite high. But I would say that this is not a real world scenario. setLineVerticalIndent will never be called for 10.000 lines.

I don't really understand what method `resetCache` is doing and why it is not working properly in the scrolling scenario mentioned in the issue (https://github.com/eclipse-platform/eclipse.platform.swt/issues/2512). Can you maybe give some more insights and give some hints which parts of the code might make the problem? I have the feeling that StyledTextRenderer#calculateIdle is not thread-safe.

I found a more efficient solution which will not has any performance impact by increasing the `initialBottomIndex` variable by value 2 in `setLineVerticalIndent`. But don't ask me for the reason, I cannot explain. I only see that the `else if (lineIndex > initialBottomIndex) {` branch is not complete from my point of view and the lines are then not correctly rendered together with the code minings.
image
 
How should we proceed here?

mickaelistria: > Factor 4.8 slower is quite high.

I guess this factor is the same even for smaller files, or is it quadratic?

> I don't really understand what method resetCache 

resetCache() resets some info for the given lines (width/height and maybe even the position). Whenever it's invoked, it will cause cairo to recompute the particular line position) before next drawing completion.
I think maybe there is a way to make things smarter by shifting some line positions without clearing the cache.
However, I haven't touched that code for a while so I'm not sure I remember things correctly enough...

> I found a more efficient solution which will not has any performance impact by increasing the initialBottomIndex variable by value 2 in setLineVerticalIndent

That's interesting. Maybe caching a few more lines ahead of time and anticipating the scrolling is a way to prevent for recomputing positions too often...

tobiasmelcher: Ok, thanks a lot. Then I would propose to go for the +2 approach.

BeckerWdf:
With the newest state of this PR I don't see the error happening any more.

BeckerWdf:
@mickaelistria: Can you pls. review again?

mickaelistria (approved):
This looks like a very good patch, as I think it actually has identified the root cause of the issue: as we're not fully sure of what's the last line when computing the lines to redraw, it makes much sense to iterate until the last one assuming a fixed/lowest line height to ensure we include all necessary lines.

BeckerWdf:
failing browser test (in org.eclipse.swt.tests.junit.Test_org_eclipse_swt_browser_Browser) look unrelated.

BeckerWdf (approved):

mickaelistria:
Thanks a lot @tobiasmelcher !

jonahgraham:
@tobias-melcher I think an earlier version of this commit was accidentally pushed to SWT in this branch https://github.com/eclipse-platform/eclipse.platform.swt/tree/styled_text_set_line_vertical_indent

If there is nothing left to do with it, can you delete this branch please?

tobiasmelcher:
> @tobias-melcher I think an earlier version of this commit was accidentally pushed to SWT in this branch https://github.com/eclipse-platform/eclipse.platform.swt/tree/styled_text_set_line_vertical_indent
> 
> If there is nothing left to do with it, can you delete this branch please?

I've now deleted the branch. Apologies — I unintentionally pushed to the wrong remote.

Issue #2516, opened 2025-09-16
Some Test_org_eclipse_swt_events_KeyEvent tests are failing with Windows Server 2025 on GHA


HeikoKlare:
Since GHA started to use Windows Server 2025 instead of Windows Server 2022, some tests of `Test_org_eclipse_swt_events_KeyEvent` are constantly failing in GHA runs.

- `testEnglishInternational_deadKey_Shift6`
- `testGerman_deadKey_unmatched `
- `testEnglishUs_CtrlBreak`
- `testEnglishUs_numpad`
- `testFrench_digits`

For example, in PR https://github.com/eclipse-platform/eclipse.platform.swt/pull/2470, the [last run that still mentions Windows Server 2022](https://github.com/eclipse-platform/eclipse.platform.swt/actions/runs/17676941028/job/50241309151) in the logs succeeded while [from the first build after that Windows Server 2025](https://github.com/eclipse-platform/eclipse.platform.swt/actions/runs/17712094723/job/50332111251) is mentioned in the logs and those tests fail.

arunjose696:
In addition to the mentioned tests have `Test_org_eclipse_swt_events_KeyEvent::testEnglishUs_stdKeys` also failing in https://github.com/eclipse-platform/eclipse.platform.swt/pull/2530

BeckerWdf:
this also fails for https://github.com/eclipse-platform/eclipse.platform.swt/pull/2527

iloveeclipse:
If no one plans to investigate the tests, and they are failing on every PR, should we disable them?

HeikoKlare:
I'd started to investigate them:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2532

I tried multiple things but I have not clue how to fix it yet. What I found so far:
- The tests fail because another shell that is _not_ a top level shell of the SWT display receives focus
- All subsequent tests fail because that same shell keeps having the focus
- It seems to be quite impossible to force focus back to the SWT shell. Adding `forceActive()` / `forceFocus()` etc. to the shell created for a test method does still end up in that other shell having focus. When I locally execute the tests and manually click into another window while the tests are executed, only the single test method fails but every subsequent method suceeds again as the focus will be properly set. This does _not_ work in CI execution. So that other shell seems to somehow enforce keeping focus.

I agree that disabling the tests seems to be the only reasonable way forward if we do not find out how to fix them. But it would also not be good to have not validation of the behavior either.
[...]

PR #2517, opened 2025-09-16
Use OS.CreateIconIndirect for colored cursors with source and mask

ShahzaibIbrahim:
Use OS.CreateIconIndirect for colored cursors with source and mask

Previously, the Cursor constructor that accepted both source and mask used OS.CreateCursor, which only supports monochrome cursors. As a result, any color information in the source was ignored and the cursor always appeared black. 

This change updates the constructor to use OS.CreateIconIndirect, allowing full-color cursors while still respecting the mask for transparency. DPI scaling of the source and mask now works correctly with colored cursors.

### How to Test 

Run the following snippet: 
```
import org.eclipse.swt.*;

public class Snippet386 {

	private static final int IMAGE_SIZE_IN_POINTS = 16;

		Shell shell = createShell(display);

		Combo combo = createConstructorCombo(shell);
		Label zoomLabel = createZoomLabel(shell);

		addZoomChangedListener(shell, zoomLabel);
		Group section = new Group(shell, SWT.NONE);
		section.setText("Scale");
		section.setLayoutData(new GridData(SWT.FILL, SWT.FILL, true, true));
		section.setLayout(new FillLayout());
		addPaintTicks(section);

		CursorManager cursorManager = new CursorManager(display, shell, combo);
		combo.addListener(SWT.Selection, e -> cursorManager.updateCursor());
		cursorManager.updateCursor();

		shell.setSize(400, 400);
		shell.open();

		eventLoop(display, shell);
		display.dispose();
	}

	private static Shell createShell(Display display) {
		shell.setText("Snippet 386");
		return shell;
	}

	private static Combo createConstructorCombo(Composite parent) {
		Label label = new Label(parent, SWT.NONE);
		label.setText("Choose Cursor Constructor:");

		Combo combo = new Combo(parent, SWT.READ_ONLY);
		combo.setItems("Cursor(Device, int)", "Cursor(Device, ImageData, ImageData, int, int)",
				"Cursor(Device, ImageData, int, int)", "Cursor(Device, ImageDataProvider, int, int)");
		combo.select(0);
		return combo;
	}

	private static Label createZoomLabel(Shell parent) {
		Label zoomLabel = new Label(parent, SWT.NONE);
		zoomLabel.setLayoutData(new GridData(SWT.FILL, SWT.CENTER, true, false));
		setZoomLabelText(parent, zoomLabel);
		return zoomLabel;
	}

	private static void setZoomLabelText(Shell shell, Label label) {
		int zoom = shell.getMonitor().getZoom();
		int expectedCursorSize = Math.round(IMAGE_SIZE_IN_POINTS * (zoom / 100f));
		label.setText("Current zoom: " + zoom + "% \nExpected Cursor Size = " + expectedCursorSize);
	}

	private static void addZoomChangedListener(Shell shell, Label zoomLabel) {
		shell.addListener(SWT.Resize, event -> {
			setZoomLabelText(shell, zoomLabel);
			shell.layout();
		});
	}

	private static void addPaintTicks(Composite composite) {
		composite.addPaintListener(event -> {
			drawTicks(composite, event.gc);
		});
	}

	private static void drawTicks(Composite shell, GC gc) {
		int deviceZoom = shell.getMonitor().getZoom();
		float devScale = deviceZoom / 100f;
		Point client = shell.getSize();
		int xPos = (int) ((client.x / 2) - (6 * (IMAGE_SIZE_IN_POINTS / devScale)));
		int tickHeight = 10;
		int yPos = 20;

		for (int tickIndex = 0; tickIndex < 6; tickIndex++) {
			xPos += (IMAGE_SIZE_IN_POINTS / devScale);
			int yOffset = (tickIndex % 3 == 1) ? 20 : (tickIndex % 3 == 2) ? 40 : 0;

			gc.drawLine(xPos, yPos, xPos, yPos + tickHeight);
			gc.drawText(Integer.toString(tickIndex * IMAGE_SIZE_IN_POINTS), xPos - 5, yPos + 12 + yOffset);
		}
	}

	private static void eventLoop(Display display, Shell shell) {
		while (!shell.isDisposed()) {
				display.sleep();
		}
	}

	public static ImageData createSolidColorImageData(int size, RGB color) {
		PaletteData palette = new PaletteData(0xFF0000, 0x00FF00, 0x0000FF);
		ImageData imageData = new ImageData(size, size, 24, palette);

		int pixel = palette.getPixel(color);
		for (int y = 0; y < size; y++) {
			for (int x = 0; x < size; x++) {
				imageData.setPixel(x, y, pixel);
			}
		}
		return imageData;
	}

	private static class CursorManager {
		private final Display display;
		private final Shell shell;
		private final Combo combo;

		CursorManager(Display display, Shell shell, Combo combo) {
			this.display = display;
			this.shell = shell;
			this.combo = combo;
		}

		void updateCursor() {
			int selection = combo.getSelectionIndex();
			Cursor oldCursor = shell.getCursor();
			if (oldCursor != null && !oldCursor.isDisposed()) {
				oldCursor.dispose();
			}
			Cursor cursor = createCursor(selection);
			if (cursor != null) {
				shell.setCursor(cursor);
			}
		}

		private Cursor createCursor(int selection) {
			switch (selection) {
			case 0:
				return new Cursor(display, SWT.CURSOR_HAND);
			case 1: {
				PaletteData rgbPalette = new PaletteData(0xFF0000, 0x00FF00, 0x0000FF);
				int bluePixel = rgbPalette.getPixel(new RGB(0, 0, 255));
				ImageData source = new ImageData(IMAGE_SIZE_IN_POINTS, IMAGE_SIZE_IN_POINTS, 24, new PaletteData(0xFF0000, 0x00FF00, 0x0000FF));

				for (int x = 0; x < IMAGE_SIZE_IN_POINTS; x++) {
					for (int y = 0; y < IMAGE_SIZE_IN_POINTS; y++) {
						source.setPixel(x, y, bluePixel);
					}
				}

				ImageData mask = new ImageData(IMAGE_SIZE_IN_POINTS, IMAGE_SIZE_IN_POINTS, 1, new PaletteData(new RGB[] {new RGB(0,0,0), new RGB(255,255,255)}));
				for (int x = 0; x < IMAGE_SIZE_IN_POINTS; x++) {
					for (int y = 0; y < IMAGE_SIZE_IN_POINTS; y++) {
						mask.setPixel(x, y, x % 2);
					}
				}

				return new Cursor(display, source, mask, IMAGE_SIZE_IN_POINTS / 2, IMAGE_SIZE_IN_POINTS / 2);
			}
			case 2: {
				RGB red = new RGB(255, 0, 0);
				return new Cursor(display, createSolidColorImageData(IMAGE_SIZE_IN_POINTS, red), 0, 0);
			}
			case 3: {
				RGB green = new RGB(0, 255, 0);
				ImageDataProvider provider = zoom -> {
					return createSolidColorImageData(IMAGE_SIZE_IN_POINTS * zoom / 100, green);
				};
				return new Cursor(display, provider, 0, 0);
			}
			default:
				return null;
			}
		}
	}
}
```

- Select the `Cursor(Device, ImageData, ImageData, int, int)` value from drop down
- See if the cursor is blue and transparency is also applied.

### Result 

Before:

After:


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
```diff
@@ -349,6 +342,39 @@ private void setHandleForZoomLevel(CursorHandle handle, Integer zoom) {
...
+ */
+
+private static int getPointerSizeScaleFactor() {
+	final int defaultCursorSize = 32;
+	int scaleFactor = 1; // Default: standard size
+
+	if (OsVersion.IS_WIN10_1809) {
+		int[] cursorBaseSize = OS.readRegistryDwords(OS.HKEY_CURRENT_USER, "Control Panel\\Cursors", "CursorBaseSize");
+		if (cursorBaseSize != null && cursorBaseSize.length > 0 && cursorBaseSize[0] > 0) {
+			scaleFactor = cursorBaseSize[0] / defaultCursorSize;
```
fedejeanne: The value in the registry is not always divisible by 32 so you are loosing precision here:

image

Is this intentional?

ShahzaibIbrahim: No it's unintentional. I assumed that it always increase by 32 as we increase the slider. I will now return this value in float for precision.

ShahzaibIbrahim:

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
```diff
@@ -697,10 +688,10 @@ private void validateMask(ImageData source, ImageData mask) {
 
 	@Override
 	public CursorHandle createHandle(Device device, int zoom) {
-		ImageData scaledSource = DPIUtil.scaleImageData(device, this.source, zoom, DEFAULT_ZOOM);
-		ImageData scaledMask = this.mask != null ? DPIUtil.scaleImageData(device, mask, zoom, DEFAULT_ZOOM)
-				: null;
-		return setupCursorFromImageData(scaledSource, scaledMask, getHotpotXInPixels(zoom), getHotpotYInPixels(zoom));
+		float zoomFactor = getPointerSizeScaleFactor() * ((float) zoom) / 100f;
```
HeikoKlare: Why change the return type of `getPointerSizeScaleFactor()` to float and have all these integer conversions / roundings in this method?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
```diff
@@ -255,10 +236,21 @@ private static CursorHandle setupCursorFromImageData(Device device, ImageData so
 			}
 		}
 		OS.MoveMemory(dibBM.bmBits, srcData, srcData.length);
-		hMask = OS.CreateBitmap(source.width, source.height, 1, 1, new byte[(((source.width + 7) / 8) + 3) / 4 * 4 * source.height]);
+		if (maskData != null) {
+			long[] maskResult = Image.initIcon(device, maskData, maskData);
+			hMask = maskResult[1];
```
HeikoKlare: If I am not mistaken, this will apply the mask data as a mask to itself, then extract the mask data out of the combined image to finally just execute `OS.CreateBitmap(i.width, i.height, 1, 1, maskData)`. Why not directly do that without all the unnecessary complexity?

HeikoKlare: And if you pass mask data, why should the source data also contain alpha information? This was not supported by the existing API, so I don't think we should implement it here. It should be valid to assume that if a mask is passed the source does not contain mask data. I think we should even deprecate the source+mask constructor and indicate that better other constructors taking transparency-aware data or even an `ImageDataProvider` should be used.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
fedejeanne: See https://github.com/eclipse-platform/eclipse.platform.swt/pull/2517#discussion_r2367342308

fedejeanne (commented):
My comment was addressed 👍 
See Heiko's comments

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
HeikoKlare: Thanks for the pointer. But then it's an issue unrelated to this PR but an issue with the existing implementation, isn't it? So please extract an appropriate PR for a separate issue out of this.

fedejeanne: I think we could simply fix it here. The method was introduced 2 weeks ago in 6e4241da7ffbe82ca6e307a794bcb8b5be661a00. That's probably why nobody noticed so far.

HeikoKlare: The issue is unrelated to this PR, so please have separate PRs for unrelated concerns.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
ShahzaibIbrahim: > And if you pass mask data, why should the source data also contain alpha information? This was not supported by the existing API, so I don't think we should implement it here. It should be valid to assume that if a mask is passed the source does not contain mask data.

In this case I shouldn't consider passed maskData here if alpha information is present in source data. I have updated the changes accordingly.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
ShahzaibIbrahim: New ticket for this float issue. I will address it in different PR: https://github.com/vi-eclipse/Eclipse-Platform/issues/456

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
```diff
@@ -258,7 +239,12 @@ private static CursorHandle setupCursorFromImageData(Device device, ImageData so
 		hMask = OS.CreateBitmap(source.width, source.height, 1, 1, new byte[(((source.width + 7) / 8) + 3) / 4 * 4 * source.height]);
 		if (hMask == 0) SWT.error(SWT.ERROR_NO_HANDLES);
 	} else {
-		ImageData mask = source.getTransparencyMask();
+		ImageData mask;
+		if (maskData != null) {
+			mask = maskData;
```
HeikoKlare: The names "mask" and "maskData" for effectively the same kind of object are not very helpful. All other methods in the class seem to use the name `mask`, so this method should probably also accept a parameter called `mask` for consistency. Then at this place you could calculate something like an `effectiveMask` or just assign the transparency mask of source in case `mask` is null.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
```diff
@@ -697,10 +683,18 @@ private void validateMask(ImageData source, ImageData mask) {
...
-		return setupCursorFromImageData(scaledSource, scaledMask, getHotpotXInPixels(zoom), getHotpotYInPixels(zoom));
+		float zoomFactor = getPointerSizeScaleFactor() * zoom / 100f;
+		int scaledSourceWidth = Math.round(this.source.width * zoomFactor);
+		int scaledSourceHeight = Math.round(this.source.height * zoomFactor);
+		ImageData scaledSource = this.source.scaledTo(scaledSourceWidth, scaledSourceHeight);
+		ImageData scaledMask = null;
+		if (this.mask != null) {
+			int scaledMaskWidth = Math.round(this.mask.width * zoomFactor);
+			int scaledMaskHeight = Math.round(this.mask.height * zoomFactor);
+			scaledMask = this.mask.scaledTo(scaledMaskWidth, scaledMaskHeight);
```
HeikoKlare: This is a behavior change. Previously `DPIUtil.scaleImageData()` was used for the mask and now `ImageData::scaledTo` is used producing a different kind of result. This either needs to be explained or reverted.

ShahzaibIbrahim: We talked about that and in your words, "probably the issue is that when using "proper" scaling for the source it creates an image with alpha data, such that the mask is not properly taken into account. Using scaledTo will of course lead to worse scaling results, but maybe it's the best solution for the source+mask use case then"

HeikoKlare: What's the existing behavior here? Did the smooth-scaled source/mask work with the existing way of creating the cursor?

ShahzaibIbrahim: With existing implementation we don't get the color and also scaling is broken. I have reverted my changes to use "proper scaling" via autoScaleImageData but when scaled the mask pixels defined are not scaled as expected. 

look at the example here:

For 100%, I have one transparent pixel after every blue pixel. 


But when scaled with autoscale, it's not as I expected (i.e. for every two pixel blue we get two pixel transparent)



We have to make a decision here, to keep smooth scaling or we don't expect transparency as we defined first using mask. Generally, it's not very ideal to use this constructor anyway.

HeikoKlare: > With existing implementation we don't get the color and also scaling is broken.

Since scaling is also broken with the existing implementation, we should consider that a separate issue and not address two concerns in this single PR. In This PR, we should focus on unifying the way in which the cursor is created, which solves the issue of wrong cursor sizing.

A separate PR may then fix the broken transparency when scaling by adapting the used scale method.

Even a further PR may add a deprecation to the constructor accepting source and mask because, as you said, it's not ideal to use it, particularly when adding new code that instantiates a cursor. The deprecation should refer to using the constructor accepting an image data provide for best results.

@ShahzaibIbrahim can you please create according follow-ups?

ShahzaibIbrahim:
Follow ups created: https://github.com/vi-eclipse/Eclipse-Platform/issues/461 and https://github.com/vi-eclipse/Eclipse-Platform/issues/462/ @fedejeanne

PR #2518, opened 2025-09-17
Modernize examples with newer Java constructs

akurtakov:


akurtakov:
Broken windows tests are unrelated.

PR #2519, opened 2025-09-17
[win32] Calculate value for NATIVE_ZOOM on demand

akoch-yatta:
To remove the memory footprint of Widget this PR removes the value for NATIVE_ZOOM from the data attribute and replaces it by calculating the value in Control#getData on demand.
This does not change the value that is returned by NATIVE_ZOOM.


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
```diff
@@ -1258,6 +1260,15 @@ public Cursor getCursor () {
 	return cursor;
 }
 
+@Override
+public Object getData(String key) {
+	if (DATA_NATIVE_ZOOM.equals(key)) {
```
laeubi: why do we need `DATA_NATIVE_ZOOM` at all?

Cant all places that currently use that key simply perform the code itself?

akoch-yatta: Currently it is some kind of a hidden feature, that should be replaced in the long run by an API solution. We want to evolve the usage of this data point (e.g. in GEF as an example) with this before we introduce a unnecessary or not suitable API. But until there is a proper usable solution, we would like to at least reduce the memory footprint of that.

laeubi: Can you explain how/where it is used and why a "hidden key" is better than a (preliminary) API? One would suspect both to be internal and subject to change so the benefit is a bit unclear. Especially as it seems only ever a property of the shell itself.

akoch-yatta: In general this data is currently needed for an experimental feature for GEF to disable the SWT autoscaling for the diagram canvas and incorporate all scaling to GEF. There we still need the information about the zoom, that _would_ have been applied by SWT if it was not disabled.
So, I did some discussion with @HeikoKlare and we discussed assuming the solution (which is currently there for windows) can be applied in the same way to GTK and cocoa. But we are unsure about this, so we would like to prevent adding perhaps unnecessary API. The adaption to GTK and cocoa will most likely not happen in this release cycle.

HeikoKlare (approved):

PR #2520, opened 2025-09-17
Make DPI_CHANGED event handling async 

amartya4256:
With this PR, all the widgets scale themselves asynchronously independent of the order saving the wait time over their children to scale.
Earlier, the widgets would scale synchronously on a `DPI_CHANGED` event from the OS, i.e., the shell receives the callback from the OS and calls and waits for its children to execute their `handleDPIChange` (Over all the superclasses of course) and every widget on `handleDPIChange` method being triggered, would call `handDPIChange` for their children widgets or associated widgets and wait for them to finish and them execute he remaining scaling statements and pass the control to their parents. This whole flow would keep the thread busy for a significant amount of time making the UI freeze for sometime.

With this approach, the Shell receives the callback from the OS and executes its `handleDPIChange` (Over all the superclasses of course) and sends a notification asynchronously, i.e., queues the `notifyListeners` to be executed for its children when the thread is free. Once the first queued `notifyListeners` runs, it would invoke the `handleDPIChange` for that particular widget, which would then schedule DPI scaling calls for its children and associated widgets and move on with its `handleDPIChange` execution. This would definitely change the hierarchical order in which the widgets scale themselves. Considering that the parents need to resize them on the basis of the size of the children, there needs to be a recalibration of the bounds for the widgets once all the widgets have scaled themselves. Hence, we keep a note of when all the widgets have completed their scaling, the last widget to scale itself triggers a Resize event, leading every widget to recalibrate their bounds. This leads to a better UI freeze free scaling where all the widgets inside the shell are still operable/clickable amidst the scaling.

Only the last commit is relevant to this PR.
Depends on https://github.com/eclipse-platform/eclipse.platform.swt/pull/2483


HeikoKlare:
What's the relation of this to ... ?
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2465

Changes look quite similar.

amartya4256:
https://github.com/eclipse-platform/eclipse.platform.swt/pull/2465 was a draft or a POC with all the changes as a whole and then I divided bits from it into trailing PRs. We will not merge https://github.com/eclipse-platform/eclipse.platform.swt/pull/2465 and I shall close it. This is the last PR that relates to https://github.com/eclipse-platform/eclipse.platform.swt/pull/2465.

HeikoKlare:
Okay, then it would be good to close that other PR since it misses a title and description anyway.

bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/internal/DPITestUtil.java
```diff
@@ -21,14 +24,34 @@ public final class DPITestUtil {
 	private DPITestUtil() {
 	}
 
+	private static int timeoutInMillis = 2000;
```
fedejeanne: Either use a constant or simply remove it and use 2000 in the call to `waitForDPIChange`
```suggestion
	private static final int TIMEOUT_MILLIS = 2000;
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Widget.java
```diff
@@ -2730,6 +2731,25 @@ void handleDPIChange(Event event, float scalingFactor) {
 	this.setData(DATA_NATIVE_ZOOM, newZoom);
 }
 
+void sendZoomChangedEvent(Event event, Shell shell) {
```
fedejeanne: Passing the initial counter inside `event.data` is a bit obscure. Why don't simply use a `static` field and set this method to be `final` so that no one meddles with the counting?

fedejeanne: Rejected since every event should have its own counter

bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/internal/DPITestUtil.java
akoch-yatta: Adapted that one and increased the timeout to 10 seconds to reduce false positives here

fedejeanne (approved):

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
```diff
@@ -4953,7 +4954,7 @@ LRESULT WM_DESTROY (long wParam, long lParam) {
 void handleMonitorSpecificDpiChange(int newNativeZoom, Rectangle newBoundsInPixels) {
 	DPIUtil.setDeviceZoom (newNativeZoom);
 	Event zoomChangedEvent = createZoomChangedEvent(newNativeZoom);
-	notifyListeners(SWT.ZoomChanged, zoomChangedEvent);
+	sendZoomChangedEvent(zoomChangedEvent, getShell());
 	this.setBoundsInPixels(newBoundsInPixels.x, newBoundsInPixels.y, newBoundsInPixels.width, newBoundsInPixels.height);
```
HeikoKlare, line 4971: Since the zoom change events are now processed asynchronously, this `setBoundsInPixels()` call will now usuallynot be processed after the events but before them. In case that's fine, probably the order of statements should be changed here, such that the control flow better reflects the final execution flow.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
```diff
@@ -4950,11 +4954,15 @@ LRESULT WM_DESTROY (long wParam, long lParam) {
 	return null;
 }
 
-void handleMonitorSpecificDpiChange(int newNativeZoom, Rectangle newBoundsInPixels) {
+private void handleMonitorSpecificDpiChange(int newNativeZoom, Rectangle newBoundsInPixels) {
```
HeikoKlare, line 4963: `WM_DPICHANGED` does nothing if `nativeZoom == newNativeZoom`. Since all events are processed asynchronous (even handling the initial DPI change event for the shell) it can happen that a shell is moved back and forth and the second event is not processed because of the above condition, leading to the shell having a wrong scaling. As an example, imagine you move a shell from 100% to 125% monitor. `WM_DPICHANGED` forwards this event as `100 != 125` and the event is queued. Then the shell is immediately moved back from 125% to 100%. `WM_DPICHANGED` does _not_ forward the event because the shell's native zoom is still 100 (the event to change it is only queued but not processed), just like the new zoom. So only the first event will be processed, leaving behind a 125% shell on a 100% monitor.

A solution would be to remove the condition such that every call to `WM_DPICHANGED` will result in an event to be processed. As an alternative, the initial event on shell level could be processed synchronously to immediately set the shell's native zoom.

Another effect of the above issue is that the shell can shrink or enlarge when rapidly moving around. That will also be fixed with the solution proposals.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
```diff
@@ -5876,6 +5885,27 @@ LRESULT wmScrollChild (long wParam, long lParam) {
...
+		try {
+			if (!this.isDisposed() && event.doit) {
+				notifyListeners(SWT.ZoomChanged, event);
+			}
+		} finally {
+			if (shell.isDisposed()) {
+				return;
+			}
+			if (handleDPIChangedScheduledTasksCount.decrementAndGet() <= 0) {
+				currentDpiChangeEvent = null;
```
HeikoKlare: Shouldn't this only be done when `event.doit`? Otherwise the event currently being processed could have been cancelled at the very end of processing it (thus counter is 0), so the current event would be cleared even though it is still being processed (as it's a different event than the current). Another option would be to only clear `currentDpiChangeEvent` if `currentDpiChangeEvent == event`.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
akoch-yatta: Good point, I removed this condition. I don't see any harm in doing it, actually I just verified it by injecting the same multiple events for the same zoom change.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
akoch-yatta: Both conditions should be fine, as long as `currentDpiChangeEvent `is only changed when `event.doit` is falsified. I still took the second proposal as it seemed safer to me.

akoch-yatta:
@fedejeanne Could you have a second look on this PR. I noticed an issue when changing monitors quite fast so the second DPI change event was triggered before the first one was fully handled. This issue should not appear anymore, at least I could not reproduce it anyhow.

fedejeanne:
btw. I'm on Windows **11** since this morning.

Everything looks good on the `master` branch.

akoch-yatta:
@fedejeanne Good catch with the dark theme. As light and dark theme menu bar is rendered differently this was caused as the call to getBoundsInPixels triggered a re-render of the menu bar, but the menu bar was not yet updated - the async call was not yet processed. 
As I didn't find a way to re-trigger the resizing of the menu bar, the easiest solution was to still do the Shell DPI change handling synchronously and only process all children asynchronously.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
HeikoKlare: Since you now made the event processing at shell level synchronous, it would probably be possible to revert that change.

fedejeanne (changes requested):
@akoch-yatta I tested again and I still see some issues.

## My setup (same as before)
- 175% monitor on the left (primary)
- 100% monitor on the right

## Issue 1: Icons have the wrong size

image

**To reproduce**
- Start the workbench on the 100% monitor (if necessary, move it there and restart it).

## Issue Nr. 2: Editors are cut-off

image

**To reproduce (only happens sporadically)**
- Either move back and forth really fast between the 2 monitors and then leave the workbench in the 175% monitor or
- Restart the application in the 175% monitor

---

The last one is really hard to reproduce (I've only seen it once when moving back and forth and once when restarting).

akoch-yatta:
@fedejeanne the first issue was caused by ane interfering DPI change events triggered by moving the shell and changing the parent of ImageBasedFrame + the ZoomChange event listener in ImageBasedFrame. This listener currently relies on a Resize event from the Shell, which is not thrown in the scenario, when the parent is changed. 
With this I wasn't able to produce Issue 2 anymore on my laptop.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Shell.java
```diff
@@ -132,6 +132,7 @@ public class Shell extends Decorations {
 	long toolIcon, balloonIcon;
 	long windowProc;
 	Control lastActive;
+
```
fedejeanne:
```suggestion
```

fedejeanne (approved):
LGTM (only an unnecessary empty line)

I wasn't able to reproduce error Nr. 2 either, thank you for the fix @akoch-yatta !

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Shell.java
akoch-yatta: Will get rid of this one

Issue #2498, opened 2025-09-11
GTK4: org.eclipse.swt.tests.junit.Test_org_eclipse_swt_widgets_Shell test fails.


raghucssit:
```
2025-09-08T15:18:08.1580506Z [ERROR] org.eclipse.swt.tests.junit.Test_org_eclipse_swt_widgets_Shell.test_Issue450_NoShellActivateOnSetFocus -- Time elapsed: 7.709 s <<< FAILURE!
2025-09-08T15:18:08.1581519Z java.lang.AssertionError: 
2025-09-08T15:18:08.1581878Z Shell did not activate
2025-09-08T15:18:08.1582828Z Expected: is <Shell {shell2} [layout=GridLayout {marginWidth=5 marginHeight=5 horizontalSpacing=5 verticalSpacing=5}]>
2025-09-08T15:18:08.1583657Z      but: was null
2025-09-08T15:18:08.1584129Z 	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
2025-09-08T15:18:08.1584956Z 	at org.eclipse.swt.tests.junit.SwtTestUtil.waitShellActivate(SwtTestUtil.java:477)
2025-09-08T15:18:08.1586245Z 	at org.eclipse.swt.tests.junit.Test_org_eclipse_swt_widgets_Shell.test_Issue450_NoShellActivateOnSetFocus(Test_org_eclipse_swt_widgets_Shell.java:971)
2025-09-08T15:18:08.1587397Z 	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
2025-09-08T15:18:08.1588646Z 	Suppressed: java.lang.Throwable: Screenshot written to /tmp/org.eclipse.swt.tests.junit.Test_org_eclipse_swt_widgets_Shell.test_Issue450_NoShellActivateOnSetFocus.png
2025-09-08T15:18:08.1590017Z 		at org.eclipse.test.Screenshots$ScreenshotOnFailure.failed(Screenshots.java:41)
2025-09-08T15:18:08.1590560Z 
2025-09-08T15:18:08.1591146Z [ERROR] org.eclipse.swt.tests.junit.Test_org_eclipse_swt_widgets_Shell.test_getLocation -- Time elapsed: 1.238 s <<< FAILURE!
2025-09-08T15:18:08.1592109Z java.lang.AssertionError: :a: expected:<10> but was:<0>
[...]

PR #2523, opened 2025-09-18
GTK4: Fix GTK4 shell location test.

raghucssit:
We cannot set/determine the location of the shell using Window Handle on
GTK4. So fix the test not to test the set and get location on GTK4.

see https://github.com/eclipse-platform/eclipse.platform.swt/issues/2498

### DIFF
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_widgets_Shell.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_widgets_Shell.java
@@ -457,6 +457,10 @@ public void test_getImeInputMode() {
 @Override
 @Test
 public void test_getLocation() {
+	//Setting location for Windows is not supported in GTK4
+	if (isGTK4()) {
+		return;
+	}
 	shell.setLocation(10,15);
 	assertEquals(":a:", 10, shell.getLocation().x);
 	assertEquals(":b:", 15, shell.getLocation().y);
@@ -998,5 +1002,28 @@ public void test_Issue450_NoShellActivateOnSetFocus() {
 			System.getProperties().remove(key);
 		}
 	}
+}
+
+@Override
+public void test_setLocationLorg_eclipse_swt_graphics_Point() {
+	//Setting location for Windows is not supported in GTK4
+	if (isGTK4()) {
+		return;
+	}
+	super.test_setLocationLorg_eclipse_swt_graphics_Point();
+}
+
+@Override
+public void test_setLocationII() {
+	//Setting location for Windows is not supported in GTK4
+	if (isGTK4()) {
+		return;
+	}
+	super.test_setLocationII();
+}
+
+public static boolean isGTK4() {
+    String gtkVersion = System.getProperty("org.eclipse.swt.internal.gtk.version", "");
+    return SwtTestUtil.isGTK && gtkVersion.startsWith("4");
 }
 }


raghucssit:
@iloveeclipse please add GTK4 label.

akurtakov:
So if one sets location via API and user moves the shell returned location will be not where it was moved but rather what was initially set via the API?!? This looks like totally broken logic.

iloveeclipse:
If I remember right, GTK4 provides only relative coordinates, so can we record the diff to absolute coordinates set by user, and apply the diff on next get to the value reported by GTK4?

laeubi:
> So if one sets location via API and user moves the shell returned location will be not where it was moved but rather what was initially set via the API?!? This looks like totally broken logic.

One might want to attach a move listener to fix that, but one should also note that not all shells are movable and it is better than having no location reporting at all, so not optimal but a first step maybe...

akurtakov:
> If I remember right, GTK4 provides only relative coordinates, so can we record the diff to absolute coordinates set by user, and apply the diff on next get to the value reported by GTK4?

I should start with the fact that absolute coordinates set by user will have zero effect when running on gtk3+Wayland(https://docs.gtk.org/gdk3/method.Window.move.html) or gtk4 (https://github.com/eclipse-platform/eclipse.platform.swt/blob/2edab106b0bd3611823ad5cdbd8336c693585efe/bundles/org.eclipse.swt/Eclipse%20SWT/gtk/org/eclipse/swt/widgets/Control.java#L1092 ). With the above said such a thing have to be limited to gtk3+x11 where things already behave like that even without such a hack.

akurtakov:
https://discourse.gnome.org/t/how-to-center-gtkwindows-in-gtk4/3112/11 and the whole thread is worth reading on the topic from one of the main gtk developers.

akurtakov:
https://docs.gtk.org/gtk4/migrating-3to4.html#adapt-to-coordinate-api-changes

raghucssit:
After some tries and readings.. I conclude that it is not possible to know Location of the Top Level Window.
`Shell.gtk_configure_event(long, long)` is triggered only on GTK3 but not for GTK4.
So there is a listener already to receive the configure event but it's never called by GTK4 because it has removed `gtk_window_get_position()`.
So we can only fix the test and update API doc.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_widgets_Shell.java
```diff
@@ -457,6 +458,10 @@ public void test_getImeInputMode() {
 @Override
 @Test
 public void test_getLocation() {
+	//Setting location for Windows is not supported in GTK4
+	if (GTK.GTK4) {
```
akurtakov: If you run the test on Windows/Mac - class GTK will not be available and ClassNotFoundException will be thrown.

raghucssit: Oh, then how to disable these tests for GTK4 ?

raghucssit: Is there any test parameter or maven configuration ?

akurtakov: It's best to use the osgi.ws system property equals to gtk, combined with org.eclipse.swt.internal.gtk.version system property startsWith("4"). That way test for gtk 3.x will still execute and only gtk 4.x will be disabled.

raghucssit: > It's best to use the osgi.ws system property equals to gtk, combined with org.eclipse.swt.internal.gtk.version system property startsWith("4"). That way test for gtk 3.x will still execute and only gtk 4.x will be disabled.

Thank you.. It works well.. I have fixed it now..

iloveeclipse:
@raghucssit , @akurtakov : I think we should update Javadoc on affected SWT methods (get/set) to add a note about this GTK4 limitation?

Beside this: wdyt about adding some logging callback from IDE to SWT, so that it could be used to report errors in the IDE log if some unexpected (not critical) SWT error appears? So far SWT can only throw exceptions or print to sysout, but I would prefer also to log missing functionality like here to IDE error log without breaking callers. WDYT?

raghucssit:
> I think we should update Javadoc on affected SWT methods (get/set) to add a note about this GTK4 limitation?

I will do this with this PR.

> @raghucssit , @akurtakov : I think we should update Javadoc on affected SWT methods (get/set) to add a note about this GTK4 limitation?
> 
> Beside this: wdyt about adding some logging callback from IDE to SWT, so that it could be used to report errors in the IDE log if some unexpected (not critical) SWT error appears? So far SWT can only throw exceptions or print to sysout, but I would prefer also to log missing functionality like here to IDE error log without breaking callers. WDYT?

I don't know how to do this.. May be we can do it with different PR because improving logging mechanism altogether seems some enhancement.

iloveeclipse:
> I don't know how to do this.. May be we can do it with different PR because improving logging mechanism altogether seems some enhancement.

Sure, I was interesting to hear your opinion.

raghucssit:
> I will do this with this PR.

Already API doc has the needed warning for wayland windowing system. I think it's enough. Can we say it as may be GTK4 limitation ?
```
setLocation()
 * <b>Warning:</b> When executing this operation on a shell, it may not have the
 * intended effect on some platforms. For example, executing this operation on a
 * shell when the environment uses the Wayland protocol, nothing will happen.
```
```
getLocation()
 * <b>Warning:</b> When executing this operation on a shell, it may not yield a
 * value with the expected meaning on some platforms. For example, executing
 * this operation on a shell when the environment uses the Wayland protocol, the
 * result is <b>not</b> a coordinate relative to the display. It will not change
 * when moving the shell.
 *
```

raghucssit:
Shell test has only 1 fail now. So this PR has effect.. Rest everything is ok. Good to merge.
```
Z [ERROR] Tests run: 140, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 10.40 s <<< FAILURE! -- in org.eclipse.swt.tests.junit.Test_org_eclipse_swt_widgets_Shell
2025-09-19T12:44:51.0459068Z [ERROR] org.eclipse.swt.tests.junit.AllTests.test_Issue450_NoShellActivateOnSetFocus -- Time elapsed: 4.969 s <<< FAILURE!
```

akurtakov:
Technically it's Gtk 4.x design decision in order to fix security concern so it's not a limitation but rather a deliberate decision.

PR #2514, opened 2025-09-15
Add internal API in SVGrasterizer to rasterize images at given height and width

arunjose696:
It shall be possible to draw images at a custom zoom or size while maintaining the best achievable quality. To support this, images originating from an SVG source must be rasterized at the required size.

The internal API introduced in this PR enables loading SVGs at custom sizes, allowing images to be drawn at any requested zoom level or resolution, as discussed in in https://github.com/vi-eclipse/Eclipse-Platform/issues/326

> **Goal** It shall be possible to draw images at a custom zoom/size at the best achievable quality. This means either
> 
> * rescaling from the most fitting zoomed version of the image that exists
> * rasterizing the image at the according size from an SVG source

This PR specifically adds the internal API to handle rasterization for images coming from SVG sources.

PR #2526, opened 2025-09-19
GC#DrawImage(image, destX, destY, destWidth, destHeight) now tries to load images at the requested destination size before drawing

arunjose696:
The DrawImage API with just destination coordinates now tries to load images at the requested size.
If the requested size is not available, it falls back to the closest available size by calculating the appropriate zoom.
Images are loadable at the target size if the image is created from an SVG stream or filename, or The image is created with an ImageDataProvider that is ImageDataAtSizeProvider.

This commit also introduces the ImageDataAtSizeProvider interface to support dynamic sizing. 

With this changes we need to test the following
1) The svgs are rendered at custom sizes without blurry scaling (this can be compared with current commit vs master)
2)  Drawing images with targetHeight and targetWidth with ImageDataAtSizeProvider (comparison can be made against ImageDataProvider vs ImageDataAtSizeProvider both at this commit)
Steps to reproduce:

<details><summary>  snippet 1 to check drawing svgs at custom sizes  (click to show snippet) </summary>
**Step 1**: Copy the eclipse.svg to same path where the snippets are and run the below snippet.
**Step 2**: Run the snippet

```java

import org.eclipse.swt.*;
import org.eclipse.swt.custom.*;

public class Snippet386 {

        Image[] images = new Image[] {

            new Image(display, createImageFileNameProvider("src/org/eclipse/swt/snippets/eclipse.svg")),

        };

        String[] descriptions = new String[] {

            "ImageFileNameProvider with SVGs scaled by SVG rasterization"
        };

        Slice[] slices = {
        		 new Slice("Full", 0.0, 0.0, 1.0, 1.0),
                 new Slice("Top Half", 0.0, 0.0, 1.0, 0.5)
        };

        createShellWithImages(display, images, descriptions, slices, "Snippet 386 - Flipped Layout");
    }

    private static ImageFileNameProvider createImageFileNameProvider(String fileName) {
        return zoom -> {
        	if(zoom==100)
        	return fileName;
        	return null;};
    }

    static class Slice {
        String name;
        double xFrac, yFrac, wFrac, hFrac;
        Slice(String name, double x, double y, double w, double h) {
            this.name = name;
            this.xFrac = x;
            this.yFrac = y;
            this.wFrac = w;
            this.hFrac = h;
        }
    }

    private static void createShellWithImages(Display display, Image[] images, String[] descriptions, Slice[] slices, String title) {
        Shell shell = new Shell(display, SWT.SHELL_TRIM | SWT.MAX | SWT.RESIZE);
        shell.setText(title);

        ScrolledComposite scrolledComposite = new ScrolledComposite(shell, SWT.V_SCROLL | SWT.H_SCROLL | SWT.BORDER);
        Canvas canvas = new Canvas(scrolledComposite, SWT.DOUBLE_BUFFERED);
        scrolledComposite.setContent(canvas);
        scrolledComposite.setExpandHorizontal(true);
        scrolledComposite.setExpandVertical(true);

        int boxW = 1500, boxH = 1500, gap = 20;
        int titleHeight = 20, descHeight = 40, sliceHeaderHeight = 30;

        int rows = images.length;
        int cols = slices.length;

        int canvasWidth = (boxW + gap) * cols + gap;
        int canvasHeight = (boxH + titleHeight + gap) * rows + descHeight + sliceHeaderHeight;
        canvas.setSize(canvasWidth, canvasHeight);
        scrolledComposite.setMinSize(canvasWidth, canvasHeight);

        canvas.addListener(SWT.Paint, e -> {
            // Column headers
            for (int col = 0; col < cols; col++) {
                int x = col * (boxW + gap) + gap;
                Font font = new Font(display, "Arial", 18, SWT.NORMAL);
                e.gc.setFont(font);
                e.gc.drawText(slices[col].name, x, 0, true);
                font.dispose();

            }

            // Rows
            for (int row = 0; row < rows; row++) {
                Image image = images[row];
                Rectangle rect = image.getBounds();
                int y = row * (boxH + titleHeight + gap) + descHeight + sliceHeaderHeight;

                Font font = new Font(display, "Arial", 18, SWT.NORMAL);
                e.gc.setFont(font);
                e.gc.drawText(descriptions[row], 0, y - 10, true);
                font.dispose();

                for (int col = 0; col < cols; col++) {
                    Slice s = slices[col];
                    int x = col * (boxW + gap) + gap;

                    int boxTop = y + titleHeight;
                    e.gc.drawRectangle(x, boxTop, (int) Math.round(boxW / s.wFrac), (int) Math.round(boxH / s.hFrac));

                    e.gc.drawImage(image, x, boxTop, (int) Math.round(boxW / s.wFrac), (int) Math.round(boxH / s.hFrac));
                }
            }
        });

        shell.setMaximized(true);
        shell.open();

        while (!shell.isDisposed()) {
        }

        for (Image img : images) img.dispose();
        display.dispose();
    }
}



```

</details>

<table>
  <tr>
    <td align="center" valign="top" style="padding-bottom: 20px;">
      
      <br/>
      <sub>before</sub>
    </td>
  </tr>
  <tr>
    <td align="center" valign="top">
      
      <br/>
      <sub>after</sub>
    </td>
  </tr>
</table>

<details><summary>**snippet 2** Drawing images with ImageDataAtSizeProvider</summary>

```java

import org.eclipse.swt.*;
import org.eclipse.swt.custom.*;

public class Snippet386 {

        Image[] images = new Image[] {
            new Image(display, createImageDataProvider()),
            new Image(display, createImageDataAtSizeProvider())
        };

        String[] descriptions = new String[] {

            "ImageDataProvider with fixed font size for a given zoom",
            "ImageDataAtSizeProvider which scales font size based on target height and width"
        };

        Slice[] slices = {
            new Slice("Full", 0.0, 0.0, 1.0, 1.0),
   
        };

        createShellWithImages(display, images, descriptions, slices, "Snippet 386 - Flipped Layout");
    }



    private static ImageDataProvider createImageDataProvider() {
        return new ImageDataProvider() {
            @Override
            public ImageData getImageData(int zoomLevel) {
                int scaleFactor = zoomLevel / 100;
                return createScaledTextImageData(100 * scaleFactor, 100 * scaleFactor);
            }
        };
    }

    private static ImageDataProvider createImageDataAtSizeProvider() {
        return new ImageDataAtSizeProvider() {
            @Override
            public ImageData getImageData(int zoomLevel) {
                int scaleFactor = zoomLevel / 100;
                return createScaledTextImageData(100 * scaleFactor, 100 * scaleFactor);
            }

            @Override
            public ImageData getImageData(int width, int height) {
                return createScaledTextImageData(width, height);
            }
        };
    }

    private static ImageData createScaledTextImageData(int width, int height) {
        String text = "abcd";

        int fontSize = Math.max(1, height / 100);
        Font font = new Font(display, "Arial", fontSize, SWT.NORMAL);

        Image tmp = new Image(display, 1, 1);
        GC measureGC = new GC(tmp);
        measureGC.setFont(font);
        Point textExtent = measureGC.textExtent(text);
        measureGC.dispose();
        tmp.dispose();

        double scale = Math.min((double) width / textExtent.x, (double) height / textExtent.y);
        font.dispose();
        font = new Font(display, "Arial", Math.max(1, (int) (fontSize * scale)), SWT.NORMAL);

        Image image = new Image(display, width, height);
        GC gc = new GC(image);
        gc.setFont(font);
        gc.setForeground(display.getSystemColor(SWT.COLOR_BLACK));
        gc.setBackground(display.getSystemColor(SWT.COLOR_WHITE));
        gc.fillRectangle(image.getBounds());

        gc.setLineWidth(Math.max(1, width / 20));
        gc.drawLine(0, 0, width / 2, height);

        Point newTextExtent = gc.textExtent(text);
        gc.drawText(text, (width - newTextExtent.x) / 2, (height - newTextExtent.y) / 2, true);

        gc.dispose();
        ImageData data = image.getImageData();

        image.dispose();
        font.dispose();

        return data;
    }

    static class Slice {
        String name;
        double xFrac, yFrac, wFrac, hFrac;
        Slice(String name, double x, double y, double w, double h) {
            this.name = name;
            this.xFrac = x;
            this.yFrac = y;
            this.wFrac = w;
            this.hFrac = h;
        }
    }

    private static void createShellWithImages(Display display, Image[] images, String[] descriptions, Slice[] slices, String title) {
        Shell shell = new Shell(display, SWT.SHELL_TRIM | SWT.MAX | SWT.RESIZE);
        shell.setText(title);

        ScrolledComposite scrolledComposite = new ScrolledComposite(shell, SWT.V_SCROLL | SWT.H_SCROLL | SWT.BORDER);
        Canvas canvas = new Canvas(scrolledComposite, SWT.DOUBLE_BUFFERED);
        scrolledComposite.setContent(canvas);
        scrolledComposite.setExpandHorizontal(true);
        scrolledComposite.setExpandVertical(true);

        int boxW = 500, boxH = 500, gap = 20;
        int titleHeight = 20, descHeight = 40, sliceHeaderHeight = 30;

        int rows = images.length;
        int cols = slices.length;

        int canvasWidth = (boxW + gap) * cols + gap;
        int canvasHeight = (boxH + titleHeight + gap) * rows + descHeight + sliceHeaderHeight;
        canvas.setSize(canvasWidth, canvasHeight);
        scrolledComposite.setMinSize(canvasWidth, canvasHeight);

        canvas.addListener(SWT.Paint, e -> {
            // Column headers
            for (int col = 0; col < cols; col++) {
                int x = col * (boxW + gap) + gap;
                Font font = new Font(display, "Arial", 18, SWT.NORMAL);
                e.gc.setFont(font);
                e.gc.drawText(slices[col].name, x, 0, true);
                font.dispose();

            }

            // Rows
            for (int row = 0; row < rows; row++) {
                Image image = images[row];
                Rectangle rect = image.getBounds();
                int y = row * (boxH + titleHeight + gap) + descHeight + sliceHeaderHeight;

                Font font = new Font(display, "Arial", 18, SWT.NORMAL);
                e.gc.setFont(font);
                e.gc.drawText(descriptions[row], 0, y - 10, true);
                font.dispose();

                for (int col = 0; col < cols; col++) {
                    Slice s = slices[col];
                    int x = col * (boxW + gap) + gap;

                    int srcX = (int) (rect.width * s.xFrac);
                    int srcY = (int) (rect.height * s.yFrac);
                    int srcW = (int) (rect.width * s.wFrac);
                    int srcH = (int) (rect.height * s.hFrac);

                    int boxTop = y + titleHeight;
                    e.gc.drawRectangle(x, boxTop, boxW, boxH);


                    e.gc.drawImage(image, x, boxTop, boxW, boxH);

                }
            }
        });

        shell.setMaximized(true);
        shell.open();

        while (!shell.isDisposed()) {
        }

        for (Image img : images) img.dispose();
        display.dispose();
    }
}


```

</details>

 This depends on #2514 and #2509, Only last commit is relevant to this PR

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/ImageDataAtSizeProvider.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/ImageDataAtSizeProvider.java
@@ -1,0 +1,49 @@
+/*******************************************************************************
+ * Copyright (c) 2025 Vector Informatik GmbH and others.
+ *
+ * This program and the accompanying materials are made available under the terms of the Eclipse
+ * Public License 2.0 which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Michael Bangas (Vector Informatik GmbH) - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.swt.graphics;
+
+/**
+ * Provides an API that is invoked by SWT when an image needs to be drawn at a
+ * specified width and height.
+ * <p>
+ * Client code must implement this interface to supply {@link ImageData} on demand.
+ * </p>
+ *
+ * @since 3.132
+ * @noreference This class is still experimental API and might be subject to change.
+ */
+public interface ImageDataAtSizeProvider extends ImageDataProvider {
+
+	/**
+	 * Returns the {@link ImageData} for the given width and height.
+	 *
+	 * <p>
+	 * <b>Implementation notes:</b>
+	 * </p>
+	 * <ul>
+	 * <li>Returning <code>null</code> is not permitted. If <code>null</code> is
+	 * returned, SWT will throw an exception when loading the image.</li>
+	 * <li>The returned {@link ImageData} must match the requested width and height
+	 * exactly. Implementations should ensure proper resizing and scaling of the
+	 * image based on the height and width requested</li>
+	 * </ul>
+	 *
+	 * @param width  the desired width of the {@link ImageData} to be returned
+	 * @param height the desired height of the {@link ImageData} to be returned
+	 * @return the {@link ImageData} that exactly matches the requested width and
+	 *         height
+	 * @since 3.132
+	 */
+	ImageData getImageData(int width, int height);
+
+}
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Image.java
@@ -1579,22 +1579,21 @@ public String toString () {
  *
  * @param gc the GC to draw on the resulting image
  * @param imageData the imageData which is used to draw the scaled Image
- * @param width the width of the original image
- * @param height the height of the original image
- * @param scaleFactor the factor with which the image is supposed to be scaled
+ * @param width the width to which the image is supposed to be scaled
+ * @param height the height to which the image is supposed to be scaled
  *
  * @noreference This method is not intended to be referenced by clients.
  */
-public static void drawScaled(GC gc, ImageData imageData, int width, int height, float scaleFactor) {
+public static void drawAtSize(GC gc, ImageData imageData, int width, int height) {
 	StrictChecks.runWithStrictChecksDisabled(() -> {
 		Image imageToDraw = new Image(gc.device, (ImageDataProvider) zoom -> imageData);
-		gc.drawImage(imageToDraw, 0, 0, width, height,
+		gc.drawImage(imageToDraw, 0, 0, imageData.width, imageData.height,
 				/*
 				 * E.g. destWidth here is effectively DPIUtil.autoScaleDown (scaledWidth), but
 				 * avoiding rounding errors. Nevertheless, we still have some rounding errors
 				 * due to the point-based API GC#drawImage(..).
 				 */
-				0, 0, Math.round(width * scaleFactor), Math.round(height * scaleFactor));
+				0, 0, width, height);
 		imageToDraw.dispose();
 	});
 }
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
@@ -1061,7 +1061,8 @@ private class DrawImageOperation extends ImageOperation {
 
 	private void drawImageInPixels(Image image, Point location) {
 		if (image.isDisposed()) SWT.error(SWT.ERROR_INVALID_ARGUMENT);
-		drawImage(image, 0, 0, -1, -1, location.x, location.y, -1, -1, true, getZoom());
+		long handle = Image.win32_getHandle(image, getZoom());
+		drawImage(image, 0, 0, -1, -1, location.x, location.y, -1, -1, true, handle);
 	}
 }
 
@@ -1157,8 +1158,7 @@ public void drawImage(Image image, int destX, int destY, int destWidth, int dest
 		SWT.error(SWT.ERROR_INVALID_ARGUMENT);
 	}
 
-	storeAndApplyOperationForExistingHandle(new DrawScalingImageToImageOperation(image, new Rectangle(0, 0, 0, 0),
-			new Rectangle(destX, destY, destWidth, destHeight)));
+	storeAndApplyOperationForExistingHandle(new DrawScaledImageOperation(image, new Rectangle(destX, destY, destWidth, destHeight)));
 }
 
 void drawImage(Image srcImage, int srcX, int srcY, int srcWidth, int srcHeight, int destX, int destY, int destWidth, int destHeight, boolean simple) {
@@ -1216,6 +1216,29 @@ private class DrawScalingImageToImageOperation extends ImageOperation {
 	}
 }
 
+private class DrawScaledImageOperation extends ImageOperation {
+	private final Rectangle destination;
+
+	DrawScaledImageOperation(Image image, Rectangle destination) {
+		super(image);
+		this.destination = destination;
+	}
+
+	@Override
+	void apply() {
+		int gcZoom = getZoom();
+		drawImage(getImage(), destination.x, destination.y, destination.width, destination.height, gcZoom);
+	}
+}
+
+private void drawImage(Image image, int destX, int destY, int destWidth, int destHeight, int imageZoom) {
+	Rectangle destPixels = Win32DPIUtils.pointToPixel(drawable, new Rectangle(destX, destY, destWidth, destHeight), imageZoom);
+	image.executeOnImageHandleAtSize((tempHandle, handleSize) -> {
+		drawImage(image, 0, 0, handleSize.x, handleSize.y, destPixels.x, destPixels.y, destPixels.width,
+				destPixels.height, false, tempHandle);
+	}, destPixels.width, destPixels.height);
+}
+
 private void drawImage(Image image, int srcX, int srcY, int srcWidth, int srcHeight, int destX, int destY,
 		int destWidth, int destHeight, int imageZoom, int scaledImageZoom) {
 	Rectangle src = Win32DPIUtils.pointToPixel(drawable, new Rectangle(srcX, srcY, srcWidth, srcHeight), scaledImageZoom);
@@ -1237,7 +1260,7 @@ private void drawImage(Image image, int srcX, int srcY, int srcWidth, int srcHei
 			}
 		}
 	}
-	drawImage(image, src.x, src.y, src.width, src.height, dest.x, dest.y, dest.width, dest.height, false, scaledImageZoom);
+	drawImage(image, src.x, src.y, src.width, src.height, dest.x, dest.y, dest.width, dest.height, false, image.getHandle(scaledImageZoom, data.nativeZoom));
 }
 
 private class DrawImageToImageOperation extends ImageOperation {
@@ -1254,17 +1277,19 @@ private class DrawImageToImageOperation extends ImageOperation {
 
 	@Override
 	void apply() {
-		drawImage(getImage(), source.x, source.y, source.width, source.height, destination.x, destination.y, destination.width, destination.height, simple, getZoom());
+		long handle = Image.win32_getHandle(getImage(), getZoom());
+		drawImage(getImage(), source.x, source.y, source.width, source.height, destination.x, destination.y, destination.width, destination.height, simple, handle);
 	}
 }
 
-private void drawImage(Image srcImage, int srcX, int srcY, int srcWidth, int srcHeight, int destX, int destY, int destWidth, int destHeight, boolean simple, int imageZoom) {
+private void drawImage(Image srcImage, int srcX, int srcY, int srcWidth, int srcHeight, int destX, int destY, int destWidth, int destHeight, boolean simple, long tempImageHandle) {
 	if (data.gdipGraphics != 0) {
 		//TODO - cache bitmap
-		long [] gdipImage = srcImage.createGdipImage(imageZoom);
+		long [] gdipImage = srcImage.createGdipImageFromHandle(tempImageHandle);
 		long img = gdipImage[0];
 		int imgWidth = Gdip.Image_GetWidth(img);
 		int imgHeight = Gdip.Image_GetHeight(img);
+
 		if (srcWidth == 0 && srcHeight == 0) {
 			srcWidth = imgWidth;
 			srcHeight = imgHeight;
@@ -1319,13 +1344,13 @@ private void drawImage(Image srcImage, int srcX, int srcY, int srcWidth, int src
 		}
 		return;
 	}
-	long imageHandle = srcImage.getHandle(imageZoom, data.nativeZoom);
 	switch (srcImage.type) {
 		case SWT.BITMAP:
-			drawBitmap(srcImage, imageHandle, srcX, srcY, srcWidth, srcHeight, destX, destY, destWidth, destHeight, simple);
+			drawBitmap(srcImage, tempImageHandle, srcX, srcY, srcWidth, srcHeight, destX, destY, destWidth, destHeight,
+					simple);
 			break;
 		case SWT.ICON:
-			drawIcon(imageHandle, srcX, srcY, srcWidth, srcHeight, destX, destY, destWidth, destHeight, simple);
+			drawIcon(tempImageHandle, srcX, srcY, srcWidth, srcHeight, destX, destY, destWidth, destHeight, simple);
 			break;
 	}
 }
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
@@ -820,6 +820,27 @@ long getHandle (int targetZoom, int nativeZoom) {
 	return getImageMetadata(zoomContext).handle;
 }
 
+@FunctionalInterface
+interface HandleAtSizeConsumer {
+	void accept(long imageHandle, Point handleSize);
+}
+
+void executeOnImageHandleAtSize(HandleAtSizeConsumer handleAtSizeConsumer, int widthHint, int heightHint) {
+	ImageData imageData;
+	imageData = this.imageProvider.loadImageDataAtSize(widthHint, heightHint);
+	executeOnImageHandle(handleAtSizeConsumer, imageData);
+}
+
+private void executeOnImageHandle(HandleAtSizeConsumer handleAtSizeConsumer, ImageData imageData) {
+	ImageHandle handleContainer = init(imageData, -1);
+	long tempHandle = handleContainer.handle;
+	try {
+		handleAtSizeConsumer.accept(tempHandle, new Point(imageData.width, imageData.height));
+	} finally {
+		handleContainer.destroy();
+	}
+}
+
 /**
  * <b>IMPORTANT:</b> This method is not part of the public
  * API for Image. It is marked public only so that it
@@ -829,24 +850,29 @@ long getHandle (int targetZoom, int nativeZoom) {
  *
  * @param gc the GC to draw on the resulting image
  * @param imageData the imageData which is used to draw the scaled Image
- * @param width the width of the original image
- * @param height the height of the original image
- * @param scaleFactor the factor with which the image is supposed to be scaled
+ * @param width the width to which the image is supposed to be scaled
+ * @param height the height to which the image is supposed to be scaled
  *
  * @noreference This method is not intended to be referenced by clients.
  */
-public static void drawScaled(GC gc, ImageData imageData, int width, int height, float scaleFactor) {
+public static void drawAtSize(GC gc, ImageData imageData, int width, int height) {
 
 	StrictChecks.runWithStrictChecksDisabled(() -> {
 		Image imageToDraw = new Image(gc.device, (ImageDataProvider) zoom -> imageData);
-		gc.drawImage(imageToDraw, 0, 0, width, height, 0, 0, Math.round(width * scaleFactor),
-				Math.round(height * scaleFactor), false);
+		gc.drawImage(imageToDraw, 0, 0, imageData.width, imageData.height, 0, 0, width,
+				height, false);
 		imageToDraw.dispose();
 	});
 }
 
+
+
 long [] createGdipImage(Integer zoom) {
 	long handle = Image.win32_getHandle(this, zoom);
+	return createGdipImageFromHandle(handle);
+}
+
+long[] createGdipImageFromHandle(long handle) {
 	switch (type) {
 		case SWT.BITMAP: {
 			BITMAP bm = new BITMAP();
@@ -1956,8 +1982,28 @@ private abstract class AbstractImageProviderWrapper {
 		return new ElementAtZoom<>(getImageMetadata(new ZoomContext(closestZoom)).getImageData(), closestZoom);
 	}
 
+	public ImageData loadImageDataAtSize(int widthHint, int heightHint) {
+		Optional<ImageData> exact = loadImageDataAtExactSize(widthHint, heightHint);
+		if (exact.isPresent()) {
+			return adaptImageDataIfDisabledOrGray(exact.get());
+		}
+		return loadImageDataClosestTo(widthHint, heightHint);
+	}
+
+	private ImageData loadImageDataClosestTo(int targetWidth, int targetHeight) {
+		Rectangle bounds = getBounds(100);
+		int imageZoomForWidth = 100 * targetWidth / bounds.width;
+		int imageZoomForHeight = 100 * targetHeight / bounds.height;
+		int imageZoom = DPIUtil.getZoomForAutoscaleProperty(Math.max(imageZoomForWidth, imageZoomForHeight));
+		return loadImageData(imageZoom).element();
+	}
+
+	protected Optional<ImageData> loadImageDataAtExactSize(int width, int height) {
+		return Optional.empty(); // exact size not available
+	}
+
 	protected ImageHandle newImageHandle(ZoomContext zoomContext) {
-		ImageData resizedData = getImageData(zoomContext.targetZoom());
+		ImageData resizedData = getScaledImageData (zoomContext.targetZoom());
 		return newImageHandle(resizedData, zoomContext);
 	}
 
@@ -1999,7 +2045,7 @@ private class ExistingImageHandleProviderWrapper extends AbstractImageProviderWr
 
 	@Override
 	ImageData newImageData(ZoomContext zoomContext) {
-		return getScaledImageData(zoomContext.targetZoom());
+		return getImageData(zoomContext.targetZoom());
 	}
 
 	@Override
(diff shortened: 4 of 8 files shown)


bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/image/FileFormat.java
```diff
@@ -73,7 +73,16 @@ private static Optional<FileFormat> determineFileFormat(LEDataInputStream stream
 	private static final int MAX_SIGNATURE_BYTES = 18 + 2; // e.g. Win-BMP or OS2-BMP plus a safety-margin
 
 	public static boolean isDynamicallySizableFormat(InputStream is) {
-		Optional<FileFormat> format = determineFileFormat(new LEDataInputStream(is, MAX_SIGNATURE_BYTES));
+		if (!is.markSupported()) {
+	        is = new BufferedInputStream(is);
+	    }
+		is.mark(MAX_SIGNATURE_BYTES);
```
HeikoKlare: Is this change related to this PR? It looks like it tried to fix some independent issue.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2588,6 +2698,48 @@ protected ElementAtZoom<ImageData> loadImageData(int zoom) {
 	ImageDataProviderWrapper createCopy(Image image) {
 		return image.new ImageDataProviderWrapper(provider);
 	}
+
+}
+
+private class dynamicallySizableImageFileNameProviderWrapper extends ImageFileNameProviderWrapper
```
HeikoKlare: Here seem to be duplicate / unused classes?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -1164,15 +1213,40 @@ private int calculateZoomForImage(int gcZoom, int srcWidth, int srcHeight, int d
...
+	@Override
+	void apply() {
+		int gcZoom = getZoom();
+		drawImage(getImage(), destination.x, destination.y, destination.width, destination.height, gcZoom);
+	}
+}
+
+private void drawImage(Image image, int destX, int destY, int destWidth, int destHeight, int imageZoom) {
+	Rectangle destPixels = Win32DPIUtils.pointToPixel(drawable, new Rectangle(destX, destY, destWidth, destHeight), imageZoom);
+	image.executeOnImageHandleAtSizeWithZoomFallback((tempHandle, handleSize) -> {
```
HeikoKlare: It's had to understand why we need `executeOnImageHandleAtSizeWithZoomFallback`, which calculates a fallback zoom on its own and `executeOnImageHandleAtSizeOrZoom` which passes it. The latter one seems to rely on the calculation `calculateZoomForImage()` inside GC, which should probably be removed with this PR, as the responsibility to determine the best fitting image at zoom should be completely moved to image.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -820,6 +842,51 @@ long getHandle (int targetZoom, int nativeZoom) {
 	return getImageMetadata(zoomContext).handle;
 }
 
+void executeOnImageHandleAtSizeOrZoom(BiConsumer<Long, Point> drawFunction, int targetWidth, int targetHeight,
```
HeikoKlare: `drawFunction` should probably be something generic like `handleConsumer` as there is nothing effectively restricting it to drawing. Additionally, I would prefer a better type than a BiConsumer taking generic types like Long and Point, e.g., a properly named functional interface.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1908,6 +1982,11 @@ public static Image win32_new(Device device, int type, long handle, int nativeZo
 	return new Image(device, type, handle, nativeZoom);
 }
 
+@FunctionalInterface
+interface SizeProvider {
+	ImageData getForSize(int width, int height);
+}
```
HeikoKlare: Is this used at all?

HeikoKlare (commented):
I scanned the PR and commented on some obvious or essential points I came across.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -486,7 +486,13 @@ public Image(Device device, ImageData source, ImageData mask) {
...
 	super(device);
 	if (stream == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
-	this.imageProvider = new ImageDataLoaderStreamProviderWrapper(stream);
+	if(ImageDataLoader.isDynamicallySizable(stream))
+	{
+		this.imageProvider = new DynamicallySizableImageDataLoaderStreamProviderWrapper(stream);
+	}
+	else {
+		this.imageProvider = new ImageDataLoaderStreamProviderWrapper(stream);
+	}
```
HeikoKlare: When integrating a preliminary versin of this into our product, I found that this implementation can easily break: `ImageDataLoader.isDynamicallySizable(stream)` may already read the stream such that the created wrapper cannot read it anymore. That's why the wrapper initially reads the whole stream and stores it internally to then create a `ByteArrayInputStream` out of it on demand. We probably need to do something similar here. Maybe we could even use the same wrapper for both cases? Actually, `DynamicallySizableImageDataLoaderSTreamProviderWrapper` has an additional `intpuStream` field, which is also used to load an image at size, but that field is never initialized. How does that work at all?

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/image/FileFormat.java
arunjose696: This was to mark and reset the streams as the streams could not be consumed twice, but as per the latest changes we read the whole stream and stores it internally so this is no longer needed

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
arunjose696: The main purpose was to use the zoom calculated at GC for source cordinates for the case of cropping, currently the scope of this PR has reduced to just handle drawImage with the new API so have just kept in `executeOnImageHandleAtSizeWithZoomFallback` and removed the other function

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
arunjose696: Have made this a functional interface.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
arunjose696: Removed

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
arunjose696: Have moved the logic to read full stream in image constructor and use it  with  a ByteArrayInputStream out of it on demand.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -820,6 +848,44 @@ long getHandle (int targetZoom, int nativeZoom) {
 	return getImageMetadata(zoomContext).handle;
 }
 
+@FunctionalInterface
+interface ImageDrawFunction {
+    void draw(long imageHandle, Point targetSize);
+}
```
HeikoKlare: It still think that there is no reason for restricting this to "drawing"; isn't it generally some `HandleAtSizeConsumer` or the like?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -820,6 +848,44 @@ long getHandle (int targetZoom, int nativeZoom) {
...
+
+void executeOnImageHandleAtSizeWithZoomFallback(ImageDrawFunction drawFunction, int targetWidth,
```
HeikoKlare: Why does the function need to accept the target size while this method also accepts the target width / height as parameters? Doesn't the caller of this method then already need to know targetWith/targetHeight and can use that in the defined drawFunction instead of waiting for the very same arguments passed here to be received again?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -820,6 +848,44 @@ long getHandle (int targetZoom, int nativeZoom) {
...
+		int targetHeight) {
+	ImageData imageData;
+	if (this.imageProvider instanceof DynamicallySizableImageDataProvider provider) {
+		imageData = provider.loadImageData(targetWidth, targetHeight);
+	} else {
+		Rectangle bounds = getBounds(100);
+		int imageZoomForWidth = Math.round(100 * targetWidth / bounds.width);
+		int imageZoomForHeight = Math.round(100 * targetHeight / bounds.height);
+		int imageZoom = DPIUtil.getZoomForAutoscaleProperty(Math.max(imageZoomForWidth, imageZoomForHeight));
+		imageData = getImageDataUnscaled(imageZoom);
```
HeikoKlare: Why not have every wrapper implement a `loadImageData(width, height)` method and let them have this implementation in the most abstract class?
Then you can completely remove the additional dynamic wrappers (for input stream, file name provider and data provider) and instead overwrite the `loadImageData(width, height)` methods for the `ImageFileNameProviderWrapper`, `ImageDataProviderWrapper` and `ImageDataLoaderStreamProviderWrapper`, such that they use the dynamic creation logic if their input is capable of doing so. The changes should become much more local then.

arunjose696: pushing loadImageData(width, height) into the base abstraction would definitely make the design cleaner and remove the need for instanceof.

The only issue I see is that in the current implementation the “else” branch doesn’t really load an image at the exact requested width/height. Instead, it just picks the closest available zoom level and returns that, which means the resulting ImageData dimensions may not match the loaded size arguments.

So while we could expose a unified loadImageData(targetWidth, targetHeight) method, it might be misleading if callers assume those dimensions are guaranteed. But as the wrappers are private I can just put a comment in the most abstract class the arguments are hints.

HeikoKlare: I see. Seems to a bit similar to the existing loadImageData method for zoom, which returns an ElementAtZoom to indicate that it may not be able to yield the perfect fit for the desired zoom. One might also just call the method loadImageDataClosestTo(width, height) or the like to make obvious that it does not necessarily yield a perfect fit.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
arunjose696: the argument is the size of the handle, this can be  different from the passed target height and width, I have renamed the variable to `handleSize`

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
arunjose696: I’ve moved `loadImageData `into the base abstraction. Instead of calling it directly as `loadImageDataClosestTo`, I’ve renamed the method to `loadImageDataAtSize` and updated the argument names to `widthHint` and `heightHint` to make it clear these are hints, not guaranteed dimensions.

`loadImageDataAtSize` now internally calls `loadImageDataAtExactSize`, which returns the image if an exact match is available. If not, it falls back to `loadImageDataClosestTo` to provide the nearest available size.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
arunjose696: Makes sense changed to `HandleAtSizeConsumer`

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
arunjose696: No longer used due to the lates refactoring so resolving

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1939,18 +1976,44 @@ public Collection<Integer> getPreservedZoomLevels() {
...
+	ImageData getImageData(int zoom, boolean scaled) {
 		TreeSet<Integer> availableZooms = new TreeSet<>(zoomLevelToImageHandle.keySet());
 		int closestZoom = Optional.ofNullable(availableZooms.higher(zoom)).orElse(availableZooms.lower(zoom));
-		return DPIUtil.scaleImageData(device, getImageMetadata(new ZoomContext(closestZoom)).getImageData(), zoom, closestZoom);
+		ImageData unscaledImageData = getImageMetadata(new ZoomContext(closestZoom)).getImageData();
+		if (!scaled) {
+			return unscaledImageData;
+		} else {
+			return DPIUtil.scaleImageData(device, unscaledImageData, zoom, closestZoom);
+		}
```
akoch-yatta:
```suggestion
		if (scaled) {
			return DPIUtil.scaleImageData(device, unscaledImageData, zoom, closestZoom);
		} 
		return unscaledImageData;
```

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/ImageDataAtSizeProvider.java
```diff
@@ -0,0 +1,22 @@
...
+ *     Michael Bangas (Vector Informatik GmbH) - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.swt.graphics;
+
+/**
+ * @since 3.132
+ */
+public interface ImageDataAtSizeProvider extends ImageDataProvider {
+
+	ImageData getImageData(int targetWidth, int targetHeight);
```
HeikoKlare: This methods needs a Javadoc. And I propose to simplify the parameter names.
```suggestion
	ImageData getImageData(int width, int height);
```

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/ImageDataAtSizeProvider.java
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * Contributors:
+ *     Michael Bangas (Vector Informatik GmbH) - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.swt.graphics;
+
+/**
+ * @since 3.132
+ */
+public interface ImageDataAtSizeProvider extends ImageDataProvider {
```
HeikoKlare, line 25: This interface needs a Javadoc.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Image.java
```diff
@@ -1581,11 +1581,12 @@ public String toString () {
...
  * @param width the width of the original image
  * @param height the height of the original image
- * @param scaleFactor the factor with which the image is supposed to be scaled
+ * @param targetWidth the width to which the image is supposed to be scaled
+ * @param targetHeight the height to which the image is supposed to be scaled
  *
  * @noreference This method is not intended to be referenced by clients.
  */
-public static void drawScaled(GC gc, ImageData imageData, int width, int height, float scaleFactor) {
+public static void drawAtTargetSize(GC gc, ImageData imageData, int width, int height, int targetWidth, int targetHeight) {
```
HeikoKlare: Can't we just call this `drawAtSize` and also simplify the parameters accordingly? I don't see that we need the existing width/height parameters, as their values just seem to be imageData.width/height.
```suggestion
public static void drawAtSize(GC gc, ImageData imageData, int width, int height) {
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -820,6 +820,36 @@ long getHandle (int targetZoom, int nativeZoom) {
 	return getImageMetadata(zoomContext).handle;
 }
 
+@FunctionalInterface
+interface HandleAtSizeConsumer {
+	void accept(long imageHandle, Point handleSize);
+}
+
+void executeOnImageHandleAtSizeWithZoomFallback(HandleAtSizeConsumer handleAtSizeConsumer, int targetWidth,
```
HeikoKlare: Is the `WithZoomFallback` extension necessary/relevant? For the consumer, it shouldn't matter if the handle is created with based on zoom or anything else, should it?
```suggestion
void executeOnImageHandleAtSize(HandleAtSizeConsumer handleAtSizeConsumer, int targetWidth,
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2014,13 +2076,13 @@ private abstract class ImageFromImageDataProviderWrapper extends AbstractImagePr
 	void initImage() {
 		// As the init call configured some Image attributes (e.g. type)
 		// it must be called
-		newImageData(new ZoomContext(100));
+		newImageData(new ZoomContext(100), true);
 	}
 
 	@Override
-	ImageData newImageData(ZoomContext zoomContext) {
+	ImageData newImageData(ZoomContext zoomContext, boolean scaled) {
```
HeikoKlare: Passing `scaled == false` will lead to entries in `cachedImageData` that are incompatible with the current contract of that cache. The cache is supposed to contain the appropriate image data for a given zoom and not randomly some unscaled or different scaled image data. I would expect this to likely/eventually break.
At the same time I do not understand all the complexity with `getImageDataUnscaled`, the extension of `newImageData()` with the `scaled` parameter and transitive effects. All this is only used seems to be used in `loadImageDataClosestTo`. Why not just check if the handle map already contains an appropriate handle for that case and, if not, create one with the exact required code, which should effectly only be the simplified implementation of `initializeHandleFromSource()` (i.e., the one without scaling), shouldn' it? So something like:
```java
  private ImageData loadImageDataClosestTo(int targetWidth, int targetHeight) {
    ...
    initializeUnscaledHandle(zoomContext)
  }

  private ImageHandle initializeUnscaledHandle(ZoomContext zoomContext) {
    ElementAtZoom<ImageData> imageDataAtZoom = loadImageData(zoomContext.targetZoom());
    ImageData imageData = imageDataAtZoom.element();
    ImageHandle handle = newImageHandle(imageData, zoomContext);
    ImageData result = handle.getImageData();
    handle.destroy();
    return imageData
}
```

When doing that cleanup, it turns out that this flow creates a temporary handle just for the sake of creating the appropriate image data but to then later on to create another temporary handle to execute an operation on. So we create two handles even though we only need one. It seems reasonable to me to combine these to cases, i.e., in case we need to create a temporary handle based on zoom for executing the drawing operation on, to just create a handle as in the above proposed `initializeUnscaledHandle` method and directly use that one to execute the drawing operation.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
arunjose696: The name `executeOnImageHandleAtSize` can give the impression that execution always occurs at the exact requested size, which isn’t necessarily true. To make this clearer, I’ve kept the suggested method name but updated the parameters to `widthHint` and `heightHint`, which better reflect that these values are treated as preferred sizes rather than strict requirements.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
arunjose696: This totally makes sense. The loadImageData() method can now be used to load ImageData without scaling. This works seamlessly with the existing executeOnImageHandleAtSize consumer, which only requires an ImageData instance to be returned. By using loadImageData(), we also avoid unnecessarily creating new image handles.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Image.java
arunjose696: Done

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/ImageDataAtSizeProvider.java
arunjose696: Added  javadoc

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/ImageDataAtSizeProvider.java
arunjose696: Added javadoc.

HeikoKlare (approved):

PR #2530, opened 2025-09-22
Fix GDI handle leak when clearing MenuItem images

arunjose696:
When setting a null image on a MenuItem, the existing bitmap handle was overwritten with 0 without being released, causing a GDI handle leak if the MenuItem previously had an icon. This commit makes sure the handle is always deleted if it was set before.

Steps to reproduce 
1)Open a Runtime workspace 
2)Double click(or open) a file in the Package explorer 
3)Compare the GDI handles in the task manager before and after; it increases by 3 without this fix

This regression was introduced in [#1914](https://github.com/eclipse-platform/eclipse.platform.swt/pull/1914). The fix ensures that old handles are properly deleted before being replaced.


arunjose696:
Failed tests have been raised  as an issue already https://github.com/eclipse-platform/eclipse.platform.swt/issues/2516

PR #2531, opened 2025-09-22
GetLocation as Point.WithMonitor to ensure correct monitor association

ShahzaibIbrahim:
Previously, dialog positioning could be incorrect if the monitor was misidentified when restoring location. By returning the shell's location as a Point.WithMonitor (including monitor reference), we ensure dialogs and secondary shells open on the intended monitor, even in multi-monitor setups. This fixes issues where dialogs could appear on the wrong monitor after moving the parent shell.

### Steps to Reproduce

- Run the runtime workspace with monitor specific scaling turned on.
- Move the window to secondary monitor on the right (could be of any zoom)
- Resize the window to smaller size
- Open the **"Open Type"** window using `Ctrl + Shift + T`
- Move the Open Type window towards very left of the screen. 
- Close Open Type window
- Maximize the size of main window. 
- Open the **Open Type** dialog again
- You will see the dialog appearing on the left monitor instead of where it was last closed (on the right monitor)

**There are three cases which you could test and see if it preserves the previous state:**

**Case 1:** 
1 monitor, dialog should open where it was last closed as per parent shell. if child dialog is 10 px to the right to the parent shell, it should be 10 px from the parent shell doesnt matter where parent shell is or what size it is. 

**Case 2:** 
2 monitor, dialog moved to the left monitor, for e.g. -324 close the dialog. Reopen it should be -324 px from the parent shell. 

**Case 3:** 
2 monitor, parent shell is moved to the left monitor, dialog was closed to right monitor. Now it should follow parent shell.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/CoordinateSystemMapperTests.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/CoordinateSystemMapperTests.java
@@ -218,6 +218,31 @@ class CoordinateSystemMapperTests {
 		assertEquals(rectInPts, mapper.translateFromDisplayCoordinates(rectInPxs));
 	}
 
+	@ParameterizedTest
+	@MethodSource("provideCoordinateSystemMappers")
+	void translateBoundsToSecondMonitor(CoordinateSystemMapper mapper) {
+		final int OFFSET = 30;
+		final int SIZE = OFFSET * 3;
+		setupMonitors(mapper);
+		Rectangle currentPosition = new Rectangle.WithMonitor(monitors[1].getClientArea().x - OFFSET, 0, SIZE, SIZE,
+				monitors[1]);
+		Rectangle result = mapper.translateFromDisplayCoordinates(currentPosition);
+		assertTrue(monitors[1].getClientArea().intersects(result));
+	}
+
+	@ParameterizedTest
+	@MethodSource("provideCoordinateSystemMappers")
+	void translatePointsToSecondMonitor(CoordinateSystemMapper mapper) {
+		final int OFFSET = 30;
+		final int SIZE = OFFSET * 3;
+		setupMonitors(mapper);
+		Point currentPosition = new Point.WithMonitor(monitors[1].getClientArea().x - OFFSET, 0, monitors[1]);
+		Point size = new Point(SIZE, SIZE);
+		Point result = mapper.translateFromDisplayCoordinates(currentPosition);
+		Rectangle resultRect = new Rectangle(result.x, result.y, size.x, size.y);
+		assertTrue(monitors[1].getClientArea().intersects(resultRect));
+	}
+
 	private Point createExpectedPoint(CoordinateSystemMapper mapper, int x, int y, Monitor monitor) {
 		if (mapper instanceof SingleZoomCoordinateSystemMapper) {
 			return new Point(x, y);
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MultiZoomCoordinateSystemMapper.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MultiZoomCoordinateSystemMapper.java
@@ -93,7 +93,8 @@ class MultiZoomCoordinateSystemMapper implements CoordinateSystemMapper {
 
 	@Override
 	public Point translateFromDisplayCoordinates(Point point) {
-		return translateLocationInPixelsToPoints(point.x, point.y);
+		Monitor monitor = point instanceof Point.WithMonitor pointWithMonitor ? pointWithMonitor.getMonitor() : null;
+		return translateLocationInPixelsToPoints(point.x, point.y, monitor);
 	}
 
 	@Override
@@ -117,7 +118,7 @@ class MultiZoomCoordinateSystemMapper implements CoordinateSystemMapper {
 	@Override
 	public Point getCursorLocation() {
 		Point cursorLocationInPixels = display.getCursorLocationInPixels();
-		return translateLocationInPixelsToPoints(cursorLocationInPixels.x, cursorLocationInPixels.y);
+		return translateLocationInPixelsToPoints(cursorLocationInPixels.x, cursorLocationInPixels.y, null);
 	}
 
 	@Override
@@ -131,8 +132,10 @@ class MultiZoomCoordinateSystemMapper implements CoordinateSystemMapper {
 		return getPixelsFromPoint(monitor, x, y);
 	}
 
-	private Point translateLocationInPixelsToPoints(int x, int y) {
-		Monitor monitor = getContainingMonitorForPixels(x, y);
+	private Point translateLocationInPixelsToPoints(int x, int y, Monitor monitor) {
+		if (monitor == null) {
+			monitor = getContainingMonitorForPixels(x, y);
+		}
 		return getPointFromPixels(monitor, x, y);
 	}
 
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Shell.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Shell.java
@@ -924,7 +924,7 @@ public int getAlpha () {
 	OS.GetWindowRect (handle, rect);
 	int width = rect.right - rect.left;
 	int height = rect.bottom - rect.top;
-	return new Rectangle (rect.left, rect.top, width, height);
+	return new Rectangle.WithMonitor (rect.left, rect.top, width, height, getMonitor());
 }
 
 ToolTip getCurrentToolTip () {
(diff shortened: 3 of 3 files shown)


HeikoKlare (commented):
This change does not work for me yet. It seems incomplete, as `MultiZoomCoordinateSystemMapper#translateFromDisplayCoordinates()` does not take the monitor of the point into account. If I adapt the mapper method to consider the monitor (like in all other `translate...()`) methods, it works fine.
@ShahzaibIbrahim maybe you had some other changes that we tested still in place when doing your final testing? Otherwise I don't know how this change alone may have solved the issue.

I have tested with two monitors, both at 150%, primary to the left. Then having an IDE on the right monitor (with x >> 0) and a dialog places at the border to the left monitor.

ShahzaibIbrahim:
Yes, somehow I missed it in the testing. It's not using the monitor that's passed through the Points. I will fix it.

ShahzaibIbrahim:

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Shell.java
```diff
@@ -1017,7 +1017,7 @@ public int getImeInputMode () {
 	if (OS.IsIconic (handle)) return super.getLocationInPixels ();
 	RECT rect = new RECT ();
 	OS.GetWindowRect (handle, rect);
-	return new Point (rect.left, rect.top);
+	return new Point.WithMonitor (rect.left, rect.top, getMonitor());
```
HeikoKlare, line 1020: Shouldn't we probably do the exact same thing in `getBoundsInPixels()`? In `getBounds()`, we also pass that value to the coordinate system mapper which, if specified, alreday takes the given monitor into account (as the coordinate system mapper now does for the `getLocation()` case due to this change as well).

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MultiZoomCoordinateSystemMapper.java
```diff
@@ -131,8 +132,10 @@ private Point translateLocationInPointsToPixels(int x, int y, Monitor monitor) {
 		return getPixelsFromPoint(monitor, x, y);
 	}
 
-	private Point translateLocationInPixelsToPoints(int x, int y) {
-		Monitor monitor = getContainingMonitorForPixels(x, y);
+	private Point translateLocationInPixelsToPoints(int x, int y, Monitor monitor) {
+		if(monitor == null) {
```
HeikoKlare:
```suggestion
		if (monitor == null) {
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Shell.java
ShahzaibIbrahim: I agree, but just a question: Should we incorporate this change in this PR?

HeikoKlare: Yes, I would do the adaptation consistently in one PR rather than leaving behind another inconsistent state.

bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/CoordinateSystemMapperTests.java
```diff
@@ -218,6 +218,31 @@ void translateRectangleInPixelsForZeroSize(CoordinateSystemMapper mapper) {
...
+				return monitors;
+			}
+
+			@Override
+			public Monitor getPrimaryMonitor() {
+				return monitors[0];
+			}
+		};
+		Shell shell = new Shell(display);
+		shell.setBounds(3839, 0, 100, 100);
```
HeikoKlare: You should not need a shell to test the coordinate system mapper.

bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/CoordinateSystemMapperTests.java
```diff
@@ -218,6 +218,31 @@ void translateRectangleInPixelsForZeroSize(CoordinateSystemMapper mapper) {
...
+	void translateShellBoundsToSecondMonitor() {
+		MultiZoomCoordinateSystemMapper mapper = getMultiZoomCoordinateSystemMapper();
+		Rectangle boundsInPixelsForLeftMonitor = new Rectangle(0, 0, 2560, 1440);
+		Rectangle boundsInPixelsForRightMonitor = new Rectangle(3840, 0, 2560, 1440);
+		monitors = new Monitor[] { createMonitor(mapper, boundsInPixelsForLeftMonitor, 150),
+				createMonitor(mapper, boundsInPixelsForRightMonitor, 150) };
+		Display display = new Display() {
+			@Override
+			public Monitor[] getMonitors() {
+				return monitors;
```
HeikoKlare: This cannot work because `Display` has a `monitors` field, which is null. You would need to use something like `CoordinateSystemMapperTests.this.monitor` if you wanted to access the outer-scope monitors. But as indicated in another comment, it should not be necessary at all to instantiate a shell/display.

bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/CoordinateSystemMapperTests.java
```diff
@@ -218,6 +218,31 @@ void translateRectangleInPixelsForZeroSize(CoordinateSystemMapper mapper) {
 		assertEquals(rectInPts, mapper.translateFromDisplayCoordinates(rectInPxs));
 	}
 
+	@Test
+	void translateShellBoundsToSecondMonitor() {
```
HeikoKlare: Would be good to also have a test for the method changed by this PR (`translateFromDisplayCoordinates(Point)`).

bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/CoordinateSystemMapperTests.java
```diff
@@ -218,6 +218,32 @@ void translateRectangleInPixelsForZeroSize(CoordinateSystemMapper mapper) {
...
+		MultiZoomCoordinateSystemMapper mapper = getMultiZoomCoordinateSystemMapper();
+		Rectangle boundsInPixelsForLeftMonitor = new Rectangle(0, 0, 2560, 1440);
+		Rectangle boundsInPixelsForRightMonitor = new Rectangle(3840, 0, 2560, 1440);
+		monitors = new Monitor[] { createMonitor(mapper, boundsInPixelsForLeftMonitor, 150),
+				createMonitor(mapper, boundsInPixelsForRightMonitor, 150) };
+		Rectangle currentPosition = new Rectangle.WithMonitor(3839, 0, 100, 100, monitors[1]);
```
HeikoKlare: Why not just ... ?
```suggestion
		monitors = setupMonitors(mapper);
		Rectangle currentPosition = new Rectangle.WithMonitor(1980, 0, 100, 100, monitors[1]);
```
Same for the other test method.

bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/CoordinateSystemMapperTests.java
```diff
@@ -218,6 +218,32 @@ void translateRectangleInPixelsForZeroSize(CoordinateSystemMapper mapper) {
 		assertEquals(rectInPts, mapper.translateFromDisplayCoordinates(rectInPxs));
 	}
 
+	@Test
+	void translateShellBoundsToSecondMonitor() {
+		MultiZoomCoordinateSystemMapper mapper = getMultiZoomCoordinateSystemMapper();
```
HeikoKlare: Why create a specific mapper here and not parameterize the method with `provideCoordinateSystemMappers`? The test should also succeed with the single-zoom coordinate system mapper.

bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/CoordinateSystemMapperTests.java
ShahzaibIbrahim: I did it because I needed different setup from that, two monitors 150% zoom for both of it and some specific coordinates.

ShahzaibIbrahim: Nevermind I have adjusted the test expected result to use the existing setupMonitors.

HeikoKlare: Why do you need exactly 150% zoom and "specific coordinates"? If I understand correctly, the case to be tested is a gap in the coordinate system between a primary monitor (left) and a secondary monitor (right). `setupMonitors()` exactly produces such a situation.
Even more appropriate than what I proposed would be to use instead of `1980` something like `monitor[1].getClientArea().width - OFFSET` with e.g. `OFFSET == 20` and as size something like `3 * OFFSET`. Then you are flexible no matter what changes may be made to the setup.

ShahzaibIbrahim:
The failing test should be resolved with the merging of: https://github.com/eclipse-platform/eclipse.platform.swt/pull/2547

ShahzaibIbrahim:

PR #2536, opened 2025-09-23
Fix Javadoc of SVGRasterizer

### LINKED ISSUE #2515: Javadoc of `SVGRasterizer` is incomplete or wrong
I just noticed that the javadoc is probably incomplete  or wrong when I reviewed [another PR](https://github.com/eclipse-platform/eclipse.platform.swt/pull/2514) in that area:

https://github.com/eclipse-platform/eclipse.platform.swt/blob/508ad4a15bb54e280b1727db2f4dd5bd3fc4ba17/bundles/org.eclipse.swt/Eclipse%20SWT/common/org/eclipse/swt/internal/image/SVGRasterizer.java#L23-L35

I see the following issues:

1. `IOException` is thrown but not documented under what condition this happens, the `@return` claims it return `null` on invalid data but why not throw `IOException` always it seems odd to have `null` here (then maybe `Optional` would have been a better choice) and what invalid means.
3. In the head it talks about a "zoom" (also parameter is named) but then in the description it says it is a scaling factor in percent, where a factor usually has no unit and percent itself is [usually defined as 1% be a factor of 0.01](https://en.wikipedia.org/wiki/Percentage) so for a scaling factor that represents the double size one would expect to have 2.0 as a factor.

> For example, 45% (read as "forty-five percent") is equal to the fraction ⁠45/100⁠, or 0.45. 

So I think the javadoc should be streamlined here to avoid confusion even if it is an internal API and only has one implementation at the moment.

arunjose696:
This commits fixes Javadoc of SVGRasterizer and adds tests to verify the contract

Fixes #2515

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/image/SVGRasterizer.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/image/SVGRasterizer.java
@@ -14,6 +14,7 @@ package org.eclipse.swt.internal.image;
 
 import java.io.*;
 
+import org.eclipse.swt.*;
 import org.eclipse.swt.graphics.*;
 
 /**
@@ -26,10 +27,18 @@ public interface SVGRasterizer {
 	 * specified zoom.
 	 *
 	 * @param stream the SVG image as an {@link InputStream}.
-	 * @param zoom   the scaling factor (in percent) e.g. {@code 200} for doubled
-	 *               size. This value must be greater zero.
-	 * @return the {@link ImageData} for the rasterized image, or {@code null} if
-	 *         the input is not a valid SVG file or cannot be processed.
+	 * @param zoom   the scaling percentage (e.g., 100 = original size, 200 = double size).
+	 *  	   This value must be greater zero.
+	 * @return the {@link ImageData} for the rasterized image.
+	 *
+	 * @exception SWTException
+	 * <ul>
+	 *    <li>ERROR_INVALID_IMAGE - if the SVG cannot be loaded</li>
+	 * </ul>
+	 * @exception IllegalArgumentException
+	 * <ul>
+	 *    <li>ERROR_INVALID_ARGUMENT - if the zoom is less than zero</li>
+	 * </ul>
 	 */
-	public ImageData rasterizeSVG(InputStream stream, int zoom) throws IOException;
+	public ImageData rasterizeSVG(InputStream stream, int zoom);
 }
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/JSVGRasterizerTest.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/JSVGRasterizerTest.java
@@ -1,0 +1,71 @@
+/*******************************************************************************
+ * Copyright (c) 2025 Yatta Solutions and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Yatta Solutions - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.swt.tests.junit;
+
+import static org.eclipse.swt.tests.junit.SwtTestUtil.assertSWTProblem;
+import static org.junit.Assert.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.io.ByteArrayInputStream;
+import java.nio.charset.StandardCharsets;
+
+import org.eclipse.swt.SWT;
+import org.eclipse.swt.SWTException;
+import org.eclipse.swt.graphics.ImageData;
+import org.eclipse.swt.svg.JSVGRasterizer;
+import org.junit.jupiter.api.Test;
+
+class JSVGRasterizerTest {
+
+	private final JSVGRasterizer rasterizer = new JSVGRasterizer();
+	private String svgString = """
+			<svg viewBox="0 0 100 100" xmlns="http://www.w3.org/2000/svg">
+			    <rect width="100%" height="100%"/>
+			</svg>
+			""";
+
+	private ByteArrayInputStream svgStream(String svg) {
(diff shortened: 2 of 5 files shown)


bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/image/SVGRasterizer.java
```diff
@@ -26,10 +27,19 @@ public interface SVGRasterizer {
...
 	 * @return the {@link ImageData} for the rasterized image, or {@code null} if
 	 *         the input is not a valid SVG file or cannot be processed.
+	 *
+	 * @exception SWTException
+	 *<ul>
+	 *    <li>ERROR_INVALID_IMAGE - if the SVG cannot be rasterized</li>
+	 *</ul>
+	 * @exception SWTException
+	 *<ul>
+	 *    <li>ERROR_INVALID_IMAGE - if the zoom is less than zero</li>
```
HeikoKlare: Shouldn't this be `SWT.ERROR_INVALID_ARGUMENT` or `IllegalArgumentException`.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/image/SVGRasterizer.java
```diff
@@ -26,10 +27,19 @@ public interface SVGRasterizer {
...
+	 *  	   This value must be greater zero.
 	 * @return the {@link ImageData} for the rasterized image, or {@code null} if
 	 *         the input is not a valid SVG file or cannot be processed.
+	 *
+	 * @exception SWTException
+	 *<ul>
+	 *    <li>ERROR_INVALID_IMAGE - if the SVG cannot be rasterized</li>
+	 *</ul>
+	 * @exception SWTException
+	 *<ul>
```
HeikoKlare: No need to have two lists of codes for the same type of exception.
```suggestion
```

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/JSVGRasterizerTest.java
```diff
@@ -0,0 +1,66 @@
...
+	@Test
+	void testRasterizeWithZoom() {
+		ImageData data = rasterizer.rasterizeSVG(svgStream(svgString), 200);
+		assertEquals(200, data.width);
+		assertEquals(200, data.height);
+	}
+
+	@Test
+	void testRasterizeWithZoomNegative() {
+		assertThrows(IllegalArgumentException.class, () -> {
```
HeikoKlare: This does currently not conform to the contract (which says that an `SWTException` is thrown). That should be unified. An if I am not mistaken, this exception is implicitly thrown by the initialization of `BufferedImage`, which is something we should not rely on. Instead, `JSVGRaster#rasterizeSVG()` should explicitly check for the zoom being < 0 and then handle the case accordingly.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/image/SVGRasterizer.java
arunjose696: changed this to IllegalArgumentException

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/JSVGRasterizerTest.java
arunjose696: I updated the contract to specify IllegalArgumentException. Now, JSVGRaster#rasterizeSVG() triggers SWT.error(SWT.ERROR_INVALID_ARGUMENT), which throws an IllegalArgumentException internally.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/image/SVGRasterizer.java
arunjose696: Modified to IllegalArgumentException

arunjose696:
> Can you please also add a reference to the resolved issue in the commit message?

Done

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/JSVGRasterizerTest.java
HeikoKlare: I see, thanks for the explanation!

arunjose696:
>It should be added to `AllNonBrowserTests`. @arunjose696 can you please do that? Then this should be ready to go from my side.
Done

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/image/SVGRasterizer.java
```diff
@@ -26,10 +27,19 @@ public interface SVGRasterizer {
 	 * specified zoom.
 	 *
 	 * @param stream the SVG image as an {@link InputStream}.
-	 * @param zoom   the scaling factor (in percent) e.g. {@code 200} for doubled
-	 *               size. This value must be greater zero.
+	 * @param zoom   the scaling percentage (e.g., 100 = original size, 200 = double size).
+	 *  	   This value must be greater zero.
 	 * @return the {@link ImageData} for the rasterized image, or {@code null} if
 	 *         the input is not a valid SVG file or cannot be processed.
```
HeikoKlare: @arunjose696  sorry for another comment on this. I just checked again if https://github.com/eclipse-platform/eclipse.platform.swt/issues/2515 is properly resolved and found that the `null` return value was also complained. And actually, the contract still seems to be wrong here:
- It says that null can be returned when the input is not a valid SVG file. But in that case an exception with `ERROR_INVALID_IMAGE` should be thrown. This is probably even wrong in updated documentation which say that it happens if an SVG cannot be rasterized, but it's actually when it cannot be loaded. So the contract should be adapted there.
- It says that null can be returned when the file cannot be processed. But it's unclear what it means that the file cannot be "processed", and also I do not see that the code can currently lead to a null result (but will potentially rather fail because of any other runtime exception).

So maybe we can just remove the statement about the result being null, as it cannot happen anyway?

arunjose696: Sorry this was missed out, have removed the statement about null, also have updated the contract to say ERROR_INVALID_IMAGE will be thrown when image cant be loaded.

PR #2538, opened 2025-09-24
WIP - [GTK4] support Clipboard

jonahgraham:
This is my "its sort of working" Clipboard for GTK4.

This PR is a draft and contains a bunch of TODOs, some are questions for myself, some are questions for others. There are a number of items in here I plan to pull into separate PRs, so this PR may not be so interesting to review yet. But if someone want to, they should be able to run the new test class Test_org_eclipse_swt_dnd_Clipboard to see basic functionality working. This test seems reliable on my machine on GTK3, but not so much on GTK4 yet.

Once complete this will:

Fixes #2126

Issue #2126, opened 2025-05-08
[GTK4] Implement Full Clipboard support for GTK4


jonahgraham:
Using the ClipboardExample pressing Paste on StyledText doesn't do anything

jonahgraham:
For this to work we need a full Clipboard get contents implementation on GTK4. At the moment, only [gdk_clipboard_get_content](https://docs.gtk.org/gdk4/method.Clipboard.get_content.html) is used to get data from the clipboard, but gdk_clipboard_get_content is *only* for getting content from the [clips in the current process](https://docs.gtk.org/gdk4/method.Clipboard.get_content.html#:~:text=contents%20are%20not%20owned%20by%20the%20current%20process). Depending on the clipboard manager that is running, the contents of even stuff copied from our process may be transferred off our process leading to a null value when calling Clipboard.getContents.

The correct method to use is [gdk_clipboard_read_value_async()](https://docs.gtk.org/gdk4/method.Clipboard.read_value_async.html) (and perhaps the provided wrappers) which asynchronously obtains the requested data - this is roughly equivalent to how GTK3 did it in [gtk_clipboard_wait_for_contents](https://docs.gtk.org/gtk3/method.Clipboard.wait_for_contents.html) which is done here https://github.com/eclipse-platform/eclipse.platform.swt/blob/de23a39c68fd534d4e34a4d7961d689ae9971ad2/bundles/org.eclipse.swt/Eclipse%20SWT%20Drag%20and%20Drop/gtk/org/eclipse/swt/dnd/Clipboard.java#L710-L726

jonahgraham:
I think the Clipboard.setContents doesn't work properly either, but not sure why yet

jonahgraham:
> I think the Clipboard.setContents doesn't work properly either, but not sure why yet

While there is something going wrong with setContents - it seems to be made worse depending on which clipboard indicator I am using. With https://extensions.gnome.org/extension/779/clipboard-indicator/ once I try some SWT GTK4 setContents then the whole indicator stops picking up new content. https://extensions.gnome.org/extension/4839/clipboard-history/ on the other hand doesn't pick up the GTK4 set items, but it returns to normal operation on GTK3 setContents. I take this as a sign that SWT GTK4 setContents is not behaving as expected.

jonahgraham:
[...]

PR #2532, opened 2025-09-22
WiP: Analyze Test_org_eclipse_swt_events_KeyEvent for Windows Server 2025

HeikoKlare:
This is WiP to analyze the issue with the Test_org_eclipse_swt_events_KeyEvent tests on Windows Server 2025:
- https://github.com/eclipse-platform/eclipse.platform.swt/issues/2516

PR #2539, opened 2025-09-24
Improve support for Arrays of Strings in JNI Generator

jonahgraham:
A number of places in the GTK4 API there is the possibility of passing in arrays of Strings, i.e. Java String[] -> char ** as input argument to C side.

This change adds better support to the JNI generator for this use case by converting the jobjectArray (String[]) to char ** on the C side.

Part of #2126
Split out of #2538

### DIFF
--- a/bundles/org.eclipse.swt.tools/JNI Generation/org/eclipse/swt/tools/internal/NativesGenerator.java
+++ b/bundles/org.eclipse.swt.tools/JNI Generation/org/eclipse/swt/tools/internal/NativesGenerator.java
@@ -299,6 +299,14 @@ boolean generateGetParameter(JNIParameter param, boolean critical, int indent) {
 				output(iStr);
 				output(", NULL)");
 			}
+		} else if (componentType.isType("java.lang.String")) {
+			if (param.getFlag(FLAG_UNICODE)) {
+				throw new Error("not done");
+			}
+
+			output("swt_getArrayOfStringsUTF(env, arg");
+			output(iStr);
+			output(")");
 		} else {
 			throw new Error("not done");
 		}
@@ -381,6 +389,16 @@ void generateSetParameter(JNIParameter param, boolean critical) {
 				output("0");
 			}
 			output(");");
+		} else if (componentType.isType("java.lang.String")) {
+			if (param.getFlag(FLAG_UNICODE)) {
+				throw new Error("not done");
+			}
+
+			output("swt_releaseArrayOfStringsUTF(env, arg");
+			output(iStr);
+			output(", lparg");
+			output(iStr);
+			output(");");
 		} else {
 			throw new Error("not done");
 		}
@@ -452,6 +470,12 @@ boolean generateLocalVars(JNIParameter[] params, JNIType returnType) {
 				output(componentType.getTypeSignature2());
 				output(" *lparg" + i);
 				output("=NULL;");
+			} else if (componentType.isType("java.lang.String")) {
+				if (param.getFlag(FLAG_UNICODE)) {
+					output("jchar **lparg" + i + "=NULL");
+				} else {
+					output("char **lparg" + i+ "=NULL");
+				}
 			} else {
 				throw new Error("not done");
 			}
--- a/bundles/org.eclipse.swt.tools/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.swt.tools/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-Name: %pluginName
 Bundle-Vendor: %providerName
 Bundle-SymbolicName: org.eclipse.swt.tools; singleton:=true
-Bundle-Version: 3.110.800.qualifier
+Bundle-Version: 3.110.900.qualifier
 Bundle-ManifestVersion: 2
 Export-Package: org.eclipse.swt.tools.internal; x-internal:=true
 Bundle-ActivationPolicy: lazy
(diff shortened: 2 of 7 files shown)


bundles/org.eclipse.swt.tools/JNI Generation/org/eclipse/swt/tools/internal/NativesGenerator.java
```diff
@@ -77,9 +77,57 @@ public void generate(JNIClass clazz) {
...
+	generateStaticMethods();
 	generateExcludes(methods);
 	generate(methods);
 }
+private void generateStaticMethods() {
+	if (!getOutputName().equals("gtk4")) {
+		return;
+	}
+
+	outputln("""
```
jonahgraham: This is probably not where these methods should go - but I don't know/see any existing place for such C side utility methods.

Please suggest better location?

iloveeclipse:
Why are .so files changed? Shouldn't the binaries be rebuilt by the SWT build job?

jonahgraham:
The 5 test failures here all seem to be what is being investigated in https://github.com/eclipse-platform/eclipse.platform.swt/pull/2532 and not caused by this PR.

jonahgraham:
> Why are .so files changed? Shouldn't the binaries be rebuilt by the SWT build job?

I don't know - they are changed in my workspace so I included them in the commit. I can exclude these change from this (and similar PRs) but how do tests work? Does the PR do the SWT build job? Actually I don't know what the SWT build job is, but I assumed it was something on Jenkins that runs on main branch?

iloveeclipse:
@HannesWell could explain it better, I believe you don't need to checkin any binaries in a PR, if native code is changed, Jenkins job will build everything needed and test with the updated binaries, and if merged, a job will be scheduled to push updated libraries to. I believe you should be able to see this in the git history.

jonahgraham:
OK - I will look around and update this soon.

HannesWell:
> could explain it better, I believe you don't need to checkin any binaries in a PR, if native code is changed, Jenkins job will build everything needed and test with the updated binaries, and if merged, a job will be scheduled to push updated libraries to. I believe you should be able to see this in the git history.

Nope, cannot explain it better, that's exactly how it is. :)
I can only add that also the GH actions rebuild the binaries for the platform they are running on.
In general we try to avoid to have binaries contributed as they are very difficult to review.

jonahgraham:
Thanks @HannesWell - I have moved follow up questions on this to https://github.com/eclipse-platform/eclipse.platform.swt/discussions/2544

bundles/org.eclipse.swt.tools/JNI Generation/org/eclipse/swt/tools/internal/NativesGenerator.java
akurtakov: os_custom.c (
https://github.com/eclipse-platform/eclipse.platform.swt/blob/cbf7f7ac77cc847d10eb00b4523fee50a57ace35/bundles/org.eclipse.swt/Eclipse%20SWT%20PI/gtk/library/os_custom.c#L182 ) is the place where one puts such functions that are not autogenerated .

bundles/org.eclipse.swt.tools/META-INF/MANIFEST.MF
```diff
@@ -2,12 +2,12 @@ Manifest-Version: 1.0
...
 Bundle-Vendor: %providerName
 Bundle-SymbolicName: org.eclipse.swt.tools; singleton:=true
-Bundle-Version: 3.110.800.qualifier
+Bundle-Version: 3.110.900.qualifier
 Bundle-ManifestVersion: 2
 Export-Package: org.eclipse.swt.tools.internal; x-internal:=true
 Bundle-ActivationPolicy: lazy
 Bundle-Localization: plugin
-Bundle-RequiredExecutionEnvironment: JavaSE-17
+Bundle-RequiredExecutionEnvironment: JavaSE-21
```
akurtakov: If you bump the BREE one have to bump the minor version e.g. become 3.111.0, it would be nice to have in separate PR too as .classpath and .settings/ have to be modified too. PDE does it automatically when clicking 

Image

jonahgraham: I seem to have accidentally included this change in this PR, New commit pushed.

bundles/org.eclipse.swt.tools/JNI Generation/org/eclipse/swt/tools/internal/NativesGenerator.java
jonahgraham: Done.

jonahgraham:
The only failing tests are the Windows KeyEvent ones tracked in https://github.com/eclipse-platform/eclipse.platform.swt/pull/2156.

This is ready to merge. As a few other people have had a look in here I will wait a bit for last minute comments before merging myself

akurtakov:
Test failures are keyevent on windows and generally crashing tests on gtk4. 
@jonahgraham Please don't add gtk4 label for such changes as we want to have them tested with gtk3 .

jonahgraham:
> Please don't add gtk4 label for such changes as we want to have them tested with gtk3 .

OK - I didn't realize that gtk4 label changed the tests. I see it now in 

https://github.com/eclipse-platform/eclipse.platform.swt/blob/fc8e3fe9bd584387201831976a098a2991f60e08/.github/workflows/maven.yml#L17

bundles/org.eclipse.swt.tools/META-INF/MANIFEST.MF
HannesWell: Just to spread the knowledge: The `Modify..` button in this editor does all of that automatically:
https://eclipse.dev/eclipse/news/news.html?file=4.33/pde.html

PR #2540, opened 2025-09-24
Prepare TransferData for GTK4 support

jonahgraham:
The fields in TransferData are platform specific - but GTK3 and GTK4 share the same class. The fields needed for GTK4 are quite different than for GTK3. Therefore this change refactors TransferData's internals to make it ready for GTK4 better.

This is achieved by splitting out all existing fields in a new GTK3 specific subtype, and including a starting point for GTK4 fields.

Some type checking is done to make sure that the correct object is used on GTK3 vs GTK4 respectively and this is done via the newly introduced gtk3() and gtk4() methods.

For now the GTK4 has no fields, but it will eventually have the internal fields needed for GTK4 serializing and deserializing of objects from the native side.

Part of #2126
Split out of #2538

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Drag and Drop/gtk/org/eclipse/swt/dnd/TransferData.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Drag and Drop/gtk/org/eclipse/swt/dnd/TransferData.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2012 IBM Corporation and others.
+ * Copyright (c) 2000, 2025 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -13,6 +13,7 @@
  *******************************************************************************/
 package org.eclipse.swt.dnd;
 
+import org.eclipse.swt.internal.gtk.*;
 
 /**
  * The <code>TransferData</code> class is a platform specific data structure for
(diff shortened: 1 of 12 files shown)


bundles/org.eclipse.swt/Eclipse SWT Drag and Drop/common/org/eclipse/swt/dnd/DragSourceEvent.java
```diff
@@ -138,7 +138,7 @@ public String toString() {
 	String string = super.toString ();
 	return string.substring (0, string.length() - 1) // remove trailing '}'
 		+ " operation=" + detail
-		+ " type=" + (dataType != null ? dataType.type : 0)
+		+ " type=" + (dataType != null ? dataType.gtk3().type : 0)
```
jonahgraham: Here.

bundles/org.eclipse.swt/Eclipse SWT Drag and Drop/common/org/eclipse/swt/dnd/DropTargetEvent.java
```diff
@@ -139,11 +139,11 @@ public String toString() {
 	sb.append(" dataTypes={ ");
 	if (dataTypes != null) {
 		for (TransferData dataType : dataTypes) {
-			sb.append(dataType.type); sb.append(' ');
+			sb.append(dataType.gtk3().type); sb.append(' ');
```
jonahgraham: Here.

bundles/org.eclipse.swt/Eclipse SWT Drag and Drop/common/org/eclipse/swt/dnd/DropTargetEvent.java
```diff
@@ -139,11 +139,11 @@ public String toString() {
...
 		}
 	}
 	sb.append('}');
-	sb.append(" currentDataType="); sb.append(currentDataType != null ? currentDataType.type : '0');
+	sb.append(" currentDataType="); sb.append(currentDataType != null ? currentDataType.gtk3().type : '0');
```
jonahgraham: Here.

jonahgraham:
> Its just for debugging AFAICT, but since `type` is available on all platforms today perhaps I should just leave that outside of the GTK3/GTK4 structures.

This is what I did for now - but I think it would be nicer if TransferData had a toString that the DropTargetEvent.toString called instead of looking into the private data.

jonahgraham:
@akurtakov Please let me know if such a refactor is acceptable or if I need to find a different way for GTK4 and GTK3 to co-exist in TransferData

jonahgraham:
For "[Compiler](https://github.com/eclipse-platform/eclipse.platform.swt/pull/2540/checks?check_run_id=51249616220) — 2 new issues, 19 tota"l check I see this (I assume the top two are the new ones). But this seems incorrect as we are using `*` imports in SWT. What is going on here?

image

The [continuous-integration/jenkins/pr-head](https://ci.eclipse.org/releng/job/eclipse.platform.swt/job/PR-2540/3/display/redirect) and [Jenkins](https://github.com/eclipse-platform/eclipse.platform.swt/pull/2540/checks?check_run_id=51248430301) failed too, but AFAICT that is the same issue as above.

Any help gratefully received.

akurtakov:
@jonahgraham The new compiler warnings are because there are unused imports. You have TransferData in the same package so there is no need to import it, importing it's inner classes is not needed either as you don't refer to their types at all in the respective classes. Long story short - just delete the 2 unused imports (jdt even gives warnings for them) and warnings are gone.

bundles/org.eclipse.swt/Eclipse SWT Drag and Drop/gtk/org/eclipse/swt/dnd/TransferData.java
```diff
@@ -46,66 +47,115 @@ public class TransferData {
...
-	 * @see TransferData#format for the size of one unit
-	 *
-	 * @noreference This field is not intended to be referenced by clients.
+	 * @noreference This class is not intended to be referenced by clients.
+	 * @noinstantiate This class is not intended to be instantiated by clients.
+	 * @noextend This class is not intended to be subclassed by clients.
+	 * @since 3.132
 	 */
-	public int length;
+	public static class TransferDataGTK3 {
```
akurtakov, line 55: I would prefer if we don't introduce new classes/methods/fields  that are accessible by endusers and are only hidden by PDE apitools as not every SWT user/developer is developing with PDE apitools and this has created some churn in the past.

akurtakov (commented):
That's quite an unusual way to do it for SWT. 
What I am missing here is what will TransferData on Gtk 4 contain? Will it be that different ? 
The usual approach is to keep the fields that are there, add the newly needed ones and set them when needed. If needed the classes using it can have if (GTK4) setPropertyA else setPropertyB. That would keep it more in line with the rest of the codebase. 
Maybe the differences are really that big that make such approach not feasible but I am not convinced of that for now.

jonahgraham:
> @jonahgraham The new compiler warnings are because there are unused imports. You have TransferData in the same package so there is no need to import it, importing it's inner classes is not needed either as you don't refer to their types at all in the respective classes. Long story short - just delete the 2 unused imports (jdt even gives warnings for them) and warnings are gone.

Oops - I went to look again, I wasn't looking in the correct place. If I had it would have been obvious what I was doing wrong. 

The reason I looked at wrong thing is that the link in the output doesn't actually go to the correct place:

image

I suspect there is supposed to be some annotation in my PR that isn't being created properly, so the link didn't go to the correct place and then I was acting blind 🤦

bundles/org.eclipse.swt/Eclipse SWT Drag and Drop/gtk/org/eclipse/swt/dnd/TransferData.java
jonahgraham: OK - I will keep that in mind.

jonahgraham:
> That's quite an unusual way to do it for SWT. What I am missing here is what will TransferData on Gtk 4 contain? 

In GTK4 it will be quite different.

For GTK3 the TransferData contains pointers to a block of memory `pValue` and its size `length`/`format`. 

For GTK it will be a handle to [GdkContentSerializer](https://docs.gtk.org/gdk4/class.ContentSerializer.html) or  [GdkContentDeserializer](https://docs.gtk.org/gdk4/class.ContentDeserializer.html) depending on whether we are doing java to native or the reverse.

> The usual approach is to keep the fields that are there, add the newly needed ones and set them when needed.

OK. For now I am going to abandon doing it this way and we can look at applying this refactor when the Clipboard is done if we find the contents of TransferData too complicated.

PR #2541, opened 2025-09-24
New test for Clipboard

jonahgraham:
Testing the Clipboard can be difficult because there are a lot of system issues that can interfere, but of particular concern are:

1. System clipboard managers which may take ownership of the clipboard at unexpected times.
2. Limitations as to when processes can access clipboard, such as on Wayland where only active window can access clipboard.
3. Different behaviour when copying within a single process than between processes.

These tests aim to resolve these issues.

For the system clipboard manager there are a lot of extra sleep calls to allow clipboard manager to complete operations before continuing tests. In addition, we run all the tests multiple times by default to ensure stability.

For the process limitations, we carefully control when the shell is created because we often cannot get focus back when shell ends up in the background. See the openAndFocusShell and openAndFocusRemote methods.

For the different behaviours, we spin up a simple Swing app in a new process (the "remote" in openAndFocusRemote above). This app can be directed, over RMI, to access the clipboard. This allows our test to place data on the clipboard and ensure that the remote app can read the data successfully.

For now this test only covers basic text (and a little of RTF). Adding Image and other transfers is part of the future work as such functionality is added in GTK4 while working on #2126

For the changes to SwtTestUtil that we required:

1. isGTK4 moved from Test_org_eclipse_swt_widgets_Shell.java to the Utils
2. processEvents was limited to 20 readAndDispatch calls per second due to the ordering of the targetTimestamp check. This change allows full speed readAndDispatching.
3. getPath was refactored to allow better control of source and destination of files extracted. See extracting of class files for remote Swing app in startRemoteClipboardCommands method

Part of #2126
Split out of #2538

### DIFF
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_dnd_Clipboard.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_dnd_Clipboard.java
@@ -1,0 +1,358 @@
+/*******************************************************************************
+ * Copyright (c) 2025 Kichwa Coders Canada, Inc.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.swt.tests.junit;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.io.BufferedReader;
+import java.io.InputStreamReader;
+import java.lang.ProcessBuilder.Redirect;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.rmi.NotBoundException;
+import java.rmi.registry.LocateRegistry;
+import java.rmi.registry.Registry;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+import java.util.function.BooleanSupplier;
+
+import org.eclipse.swt.dnd.Clipboard;
+import org.eclipse.swt.dnd.RTFTransfer;
+import org.eclipse.swt.dnd.TextTransfer;
+import org.eclipse.swt.dnd.Transfer;
+import org.eclipse.swt.widgets.Display;
+import org.eclipse.swt.widgets.Shell;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.RepeatedTest;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+import clipboard.ClipboardCommands;
+
+/**
+ * Automated Test Suite for class org.eclipse.swt.dnd.Clipboard
+ *
+ * @see org.eclipse.swt.dnd.Clipboard
+ * @see Test_org_eclipse_swt_custom_StyledText StyledText tests as it also does
+ *      some clipboard tests
+ */
+public class Test_org_eclipse_swt_dnd_Clipboard {
+
+	@TempDir
+	static Path tempFolder;
+	static int uniqueId = 1;
+	private Display display;
+	private Shell shell;
+	private Clipboard clipboard;
+	private TextTransfer textTransfer;
+	private RTFTransfer rtfTransfer;
+	private ClipboardCommands remote;
+	private Process remoteClipboardProcess;
+
+	@BeforeEach
+	public void setUp() {
+		display = Display.getCurrent();
+		if (display == null) {
+			display = Display.getDefault();
+		}
+
+		clipboard = new Clipboard(display);
+		textTransfer = TextTransfer.getInstance();
+		rtfTransfer = RTFTransfer.getInstance();
+	}
+
+	private void sleep() throws InterruptedException {
+		if (SwtTestUtil.isGTK4) {
+			/**
+			 * TODO remove all uses of sleep and change them to processEvents with the
+			 * suitable conditional, or entirely remove them
+			 */
+			SwtTestUtil.processEvents(100, null);
+		} else {
+			SwtTestUtil.processEvents();
+		}
+	}
+
+	/**
+	 * Note: Wayland backend does not allow access to system clipboard from
+	 * non-focussed windows. So we have to create/open and focus a window here so
+	 * that clipboard operations work.
+	 */
+	private void openAndFocusShell() throws InterruptedException {
+		shell = new Shell(display);
+		shell.open();
+		shell.setFocus();
+		sleep();
+	}
+
+	/**
+	 * Note: Wayland backend does not allow access to system clipboard from
+	 * non-focussed windows. So we have to open and focus remote here so that
+	 * clipboard operations work.
+	 */
+	private void openAndFocusRemote() throws Exception {
+		startRemoteClipboardCommands();
+		remote.setFocus();
+		remote.waitUntilReady();
+		sleep();
+	}
+
+	@AfterEach
+	public void tearDown() throws Exception {
+		sleep();
+		try {
+			stopRemoteClipboardCommands();
+		} finally {
+			if (clipboard != null) {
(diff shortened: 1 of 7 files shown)


jonahgraham:
My new test didn't run - I had looked at the test file next to my new test to see if I needed to manually add it to any lists. That was Test_org_eclipse_swt_custom_VerifyKeyListener which is also not running. On further examination I found 10 test classes that are not run in the suite. I assume this is an oversight and I am providing a new PR that adds all of them in.

akurtakov:
Thanks @jonahgraham . All tests are supposed to run so it's definetely an oversight.

tests/org.eclipse.swt.tests/data/clipboard/ClipboardCommands.java
```diff
@@ -0,0 +1,32 @@
+/*******************************************************************************
+ * Copyright (c) 2000, 2025 IBM Corporation and others.
```
akurtakov: I don't believe this is IBM copyright. It woudl be nice to put yours here and in all other files.

akurtakov:
I have removed the gtk4 label to be sure that tests work fine on gtk3 for now.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_dnd_Clipboard.java
```diff
@@ -0,0 +1,322 @@
...
+			}
+			if (shell != null) {
+				shell.dispose();
+			}
+			SwtTestUtil.processEvents();
+		}
+	}
+
+	private void startRemoteClipboardCommands() throws Exception {
+		// TODO: Can I get the jar and run it without extracting it?
```
mickaelistria: I think it's a bit tricky to achieve it in a way that works with or without OSGi. Without OSGi, you can relatively easily get the jar location for a file using `getClass().getResource(getClass().getSimpleName() + ".class")` and reading/transforming the associated URL; if you're inside OSGi, you'll have to deal with Utils like FileLocator to get the URL.
I'm not aware of a mechanism that works fluently in both cases and that only involves standard Java API. But if it exists, I can't wait to read someone sharing the trick.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_dnd_Clipboard.java
```diff
@@ -0,0 +1,322 @@
...
+		// TODO: Can I get all the files without having to list them?
```
mickaelistria: It would indeed be relatively simple with a jar....
But I'm pretty sure that some code exists somewhere (not in SWT nor Eclipse, but more generally on the web), that is able to read the bytes of a Class object or .class file and list all the types that are referenced by it, that could then be filtered; eg `findReferencedTypeNames(ClipboardTest.class).stream().filter(name -> name.startsWith(ClipboardTest.class.getPackageName())`

However, it would be more code for a case that shouldn't become so tricky.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_dnd_Clipboard.java
laeubi: I don't really understand the relation regarding OSGi or not OSGi, also I'm not sure why files need to be copied, if we assume we can get the location of the jar (or the output folder e.g. `bin/` or `target/classes`) one can pass this as a `-cp` paramter and mentioning the main class to start java will not care about any additional files.

To get the source quite generic one can use `getClass().getProtectionDomain().getCodeSource().getLocation();`

See for example:
- [org.eclipse.core.runtime.adaptor.EclipseStarter.getSysPathFromCodeSource()](https://github.com/eclipse-equinox/equinox/blob/25234a41e7b925bc85dbe4b0c1db95e2c6aae894/bundles/org.eclipse.osgi/container/src/org/eclipse/core/runtime/adaptor/EclipseStarter.java#L999-L1025)
- [org.eclipse.osgi.internal.framework.EquinoxConfiguration.initializeProperties()](https://github.com/eclipse-equinox/equinox/blob/25234a41e7b925bc85dbe4b0c1db95e2c6aae894/bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/internal/framework/EquinoxConfiguration.java#L1014-L1033)

one usually want to print the actual values to system out to se how they are formatted, but from that its possible to construct a basepath.

Beside that using the classloader and copy the classes seems not that bad approach.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_dnd_Clipboard.java
laeubi: As mentioned above simply use the jar or output directory should work, why copy everything into a temp location?

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_dnd_Clipboard.java
mickaelistria: > I don't really understand the relation regarding OSGi or not OSGi,

I seem to remember that the value one gets from `Class.getResource()` can be very different with or without OSGi, without OSGi it is usually an URL including the jar location, in OSGi it can be some internal bundle:// url that only the container can then resolve back to a filesystem location.

jonahgraham: Thank you both for your input. I think I am going to leave the code as is since it seems like a testing a bunch of use cases that are not trivially straightforward. The current code is simple as I based the copying out of the class file on equivalent other uses in the tests I am confident that this copy mechanism will work in all the currently supported use cases. SwtTestUtil.getPath is what is used throughout the test suite to get images, html and other resources.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_dnd_Clipboard.java
jonahgraham: Similar to https://github.com/eclipse-platform/eclipse.platform.swt/pull/2541#discussion_r2378890628 I think maintaining the short list of class files is easier than writing a heavier weight mechanism, at least for now.

I may want more later if/when the clipboard tester becomes more complicated, especially if adding tests for more complicated types requires a bunch of extra work here.

tests/org.eclipse.swt.tests/data/clipboard/ClipboardCommands.java
jonahgraham: OK - I will do that in my next update. I'll leave this unresolved until I get to it so I don't forget.

jonahgraham:
The only failing tests are the Windows KeyEvent ones tracked in #2156.

I can see that the clipboard tests have run on all platforms, and I had to do [minor fixes](https://github.com/eclipse-platform/eclipse.platform.swt/compare/b92940e47d13f271663711a195dc3d432ac2781e..9ab2c51678e306ec96d23d7e96d08035a972beb5) to get them to pass on mac + windows.

- Linux results on jenkins: https://ci.eclipse.org/releng/job/eclipse.platform.swt/job/PR-2541/5/testReport/org.eclipse.swt.tests.junit/Test_org_eclipse_swt_dnd_Clipboard/
- Linux result github: https://github.com/eclipse-platform/eclipse.platform.swt/actions/runs/18010784764/job/51242588646?pr=2541#step:7:2682
- Windows results github: https://github.com/eclipse-platform/eclipse.platform.swt/actions/runs/18010784764/job/51242588627?pr=2541#step:7:2701
- mac x64 results github: https://github.com/eclipse-platform/eclipse.platform.swt/actions/runs/18010784764/job/51242588624?pr=2541#step:7:2499
- max arm results github: https://github.com/eclipse-platform/eclipse.platform.swt/actions/runs/18010784764/job/51242588630?pr=2541#step:7:2063

I am making a final change to only run the tests repeatedly on GTK4 since they seem stable on other platforms we don't need longer runtimes. If there is any sign of instability in these tests I will revisit. (Plus I need to rebase)

jonahgraham:
This is ready to merge. As a few other people have had a look in here I will wait a bit for last minute comments before merging myself

Issue #2542, opened 2025-09-24
Not all SWT tests are running in the suite


jonahgraham:
See https://github.com/eclipse-platform/eclipse.platform.swt/pull/2541#issuecomment-3330540961

It seems that some tests are not included in the test suites. I don't know if this is intentional, but I assume it is an accidental oversight.

I am working on a PR that adds all the tests in [org.eclipse.swt.tests.junit](https://github.com/eclipse-platform/eclipse.platform.swt/blob/b7269d920bb431a54b550eeae24f1d5fc56d06d9/tests/org.eclipse.swt.tests/JUnit%20Tests/org/eclipse/swt/tests/junit) (but not subpackages) to the test suite.

PR #2543, opened 2025-09-24
Add all missing tests to suite, mark abstract tests as such

jonahgraham:
AllWidgetTests was missing a bunch of tests, some of them seem not
very consequential (simple listener tests such as for example
[Test_org_eclipse_swt_custom_VerifyKeyListener](https://github.com/eclipse-platform/eclipse.platform.swt/blob/d5dc66178647fa40b14165a7b54bc3be8e013174/tests/org.eclipse.swt.tests/JUnit%20Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_custom_VerifyKeyListener.java)

But other ones seem to test important functionality and are rather
more substantial, such as [Test_org_eclipse_swt_widgets_TreeColumn](https://github.com/eclipse-platform/eclipse.platform.swt/blob/d5dc66178647fa40b14165a7b54bc3be8e013174/tests/org.eclipse.swt.tests/JUnit%20Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_widgets_TreeColumn.java)

There are a few test classes which are super classes of other test classes
and exist to match all the SWT classes. These have been marked as abstract
so that it is obvious why they are not included. If you had try to run them
they would have mostly failed. The secondary advantage of explicitly
marking them abstract is that JDT won't run them as a test from within
Eclipse IDE.

There are three test classes that don't pass, I have added them to the
list commented out for now, but if they are easy enough to resolve I
will enable them too.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2542

### DIFF
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_custom_StyledTextContent.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_custom_StyledTextContent.java
@@ -15,8 +15,8 @@
 package org.eclipse.swt.tests.junit;
 
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertTrue;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import org.eclipse.swt.SWT;
 import org.eclipse.swt.custom.StyledText;
@@ -24,8 +24,8 @@ import org.eclipse.swt.custom.StyledTextContent;
 import org.eclipse.swt.custom.TextChangeListener;
 import org.eclipse.swt.graphics.Point;
 import org.eclipse.swt.widgets.Shell;
-import org.junit.Before;
-import org.junit.Test;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
 
 /**
  * Automated Test Suite for class org.eclipse.swt.custom.StyledTextContent
@@ -85,7 +85,7 @@ public class Test_org_eclipse_swt_custom_StyledTextContent {
 	Shell shell;
 	StyledText styledText;
 
-@Before
+@BeforeEach
 public void setUp() {
 	shell = new Shell();
 	styledText = new StyledText(shell, SWT.NULL);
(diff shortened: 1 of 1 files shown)


jonahgraham:
Note that this PR has two commits - I added a formatting/sorting commit before adding the missing tests. The substantial part of this PR is https://github.com/eclipse-platform/eclipse.platform.swt/pull/2543/commits/40a7de03308251216290c956a0e97e428c0646dd

jonahgraham:
I fixed a couple of the bitrotten tests. But Test_org_eclipse_swt_widgets_CoolItem is more complicated (to me) because it is about layouts and computing sizes. I am unsure whether my platform is the best one to work on that.

jonahgraham:
Reading the test output it look like most of the test failures are the KeyEvent issue highlighted in https://github.com/eclipse-platform/eclipse.platform.swt/issues/2516

However the newly enabled test shows a problem on mac as `test_getWidth` (org.eclipse.swt.tests.junit.Test_org_eclipse_swt_widgets_TreeColumn) fails on both mac x86_64 and aarch64

Interestingly the same test on TableColumn seems to work fine. So perhaps some code has been fixed on macos for TableColumn but not TreeColumn which I would have thought would be very similar code.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_custom_BidiSegmentListener.java
```diff
@@ -122,7 +122,6 @@ public void test_lineGetSegmentsLorg_eclipse_swt_custom_BidiSegmentEvent() {
 
 	// should all cause an exception on a bidi platform
 	if (SwtTestUtil.isBidi()) {
-		testListener(new int[] {lineLength / 2}, true);
```
akurtakov, line 125: Why was this line removed?

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_custom_StyledTextContent.java
```diff
@@ -87,16 +87,14 @@ public void setText(String text) {
...
-	else XINSET = 0;
 	shell = new Shell();
 	styledText = new StyledText(shell, SWT.NULL);
 	styledText.setContent(content);
 }
 
 @Test
 public void test_getCharCount() {
-	assertEquals(":a:", 0, styledText.getCharCount());
+	assertEquals(":a:", INITIAL_TEXT.length(), styledText.getCharCount());
```
akurtakov: While at it, please remove these ":a:" messages. They are totally useless and what's worse cause confusion when converting tests to Java 5 where assert* methods moved message as last parameter.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_custom_BidiSegmentListener.java
jonahgraham: Please see the commit message on https://github.com/eclipse-platform/eclipse.platform.swt/pull/2543/commits/66f16c12a234613c16ffa3e313dd9947e1d2a3ae - replicated here:

> Fix Test_org_eclipse_swt_custom_BidiSegmentListener
> 
> Removed the part of the test that now fails. A specific combination of segments used to cause an exception, but between the pair of commits https://github.com/eclipse-platform/eclipse.platform.swt/commit/d3c0efa94b11c766b17f4670a7e8438e97a87bdc and https://github.com/eclipse-platform/eclipse.platform.swt/commit/b4dfa25ce1ddde6037308a5f9e2105a81151f900 this no longer causes an exception. Therefore I proceeded to remove the invalid test.

(BTW if you are seeing this twice, I accidentally put the above on the wrong comment thread the first time)

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_custom_StyledTextContent.java
jonahgraham: I added a separate commit 6595104852f466730457e2cf759699d511d9bae9 that converts test to JUnit5 API.

akurtakov:
Failing windows tests are tracked in https://github.com/eclipse-platform/eclipse.platform.swt/issues/2516

PR #2546, opened 2025-09-25
Fix specific WidgetWin32Tests to not use Layout

amartya4256:
Some of the tests in WidgetWin32Tests assert the size of the controls after DPI_CHANGE. The tests work fine since theres no updateLayout called but its deffered to be executed after the thread is free and hence the controls size are unaffected. However this is risky and layouting can make the values arbitrary and the tests can fail. Hence, we simply test the absolute values without using any Layout in this PR.


akurtakov:
I have missed when this folder was committed but shouldn't it be in https://github.com/eclipse-platform/eclipse.platform.swt/tree/master/tests/org.eclipse.swt.tests.win32 ?

HeikoKlare:
They were introduce 1,5 years ago or so. They are unit tests and no component tests (testing not at bundle API level), which is why they are not in the separate Win32 test bundle but in the Win32 fragment.

akurtakov:
I see. Thanks for the explanation.

PR #2547, opened 2025-09-25
Reset device zoom to 100 in ResetMonitorSpecificScalingExtension

ShahzaibIbrahim:
Disposing the display alone is insufficient, as some tests do not create a new display and continue to use the previously set device zoom when calling DPIUtil.getDeviceZoom. This change ensures that device zoom is explicitly reset to 100 after each test, preventing unintended side effects from lingering zoom values across tests.


fedejeanne (approved):

fedejeanne:

PR #2548, opened 2025-09-25
Add drawImage API that takes only a destination rectangle

arunjose696:
Introduces a new method that draws the full image into a specified destination rectangle without requiring source bounds. Consumers no longer need to call image.getBounds() to determine image dimensions, unlike the existing drawImage method (Image image, int srcX, int srcY, int srcWidth, int srcHeight, int destX, int destY, int destWidth, int destHeight), which required knowing the image size in advance.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/GC.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/GC.java
@@ -1189,11 +1189,67 @@ public void drawImage(Image image, int srcX, int srcY, int srcWidth, int srcHeig
 	drawImage(image, srcX, srcY, srcWidth, srcHeight, destX, destY, destWidth, destHeight, false);
 }
 
+/**
+ * Draws the full source image into a specified rectangular area in the
+ * receiver. The image will be stretched or shrunk as needed to exactly fit the
+ * destination rectangle.
+ *
+ * @param image      the source image
+ * @param destX      the x coordinate in the destination
+ * @param destY      the y coordinate in the destination
+ * @param destWidth  the width in points of the destination rectangle
+ * @param destHeight the height in points of the destination rectangle
+ *
+ * @exception IllegalArgumentException
+ *                                     <ul>
+ *                                     <li>ERROR_NULL_ARGUMENT - if the image is
+ *                                     null</li>
+ *                                     <li>ERROR_INVALID_ARGUMENT - if the image
+ *                                     has been disposed</li>
+ *                                     <li>ERROR_INVALID_ARGUMENT - if any of
+ *                                     the width or height arguments are
+ *                                     negative.
+ *                                     </ul>
+ * @exception SWTException
+ *                                     <ul>
+ *                                     <li>ERROR_GRAPHIC_DISPOSED - if the
+ *                                     receiver has been disposed</li>
+ *                                     </ul>
+ * @exception SWTError
+ *                                     <ul>
+ *                                     <li>ERROR_NO_HANDLES - if no handles are
+ *                                     available to perform the operation</li>
+ *                                     </ul>
+ * @since 3.132
+ */
+public void drawImage(Image image, int destX, int destY, int destWidth, int destHeight) {
+	if (handle == null) {
+		SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
+	}
+	if (destWidth == 0 || destHeight == 0) {
+		return;
+	}
+	if (destWidth < 0 || destHeight < 0) {
+		SWT.error(SWT.ERROR_INVALID_ARGUMENT);
+	}
+	if (image == null) {
+		SWT.error(SWT.ERROR_NULL_ARGUMENT);
+	}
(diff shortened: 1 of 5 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -1213,7 +1249,10 @@ private void drawImage(Image srcImage, int srcX, int srcY, int srcWidth, int src
 		long img = gdipImage[0];
 		int imgWidth = Gdip.Image_GetWidth(img);
 		int imgHeight = Gdip.Image_GetHeight(img);
-
+		if(srcWidth==0 && srcHeight==0) {
```
fedejeanne: Mind the formatting (spacing)

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -1109,6 +1109,42 @@ public void drawImage (Image image, int srcX, int srcY, int srcWidth, int srcHei
...
+ * @exception IllegalArgumentException <ul>
+ *    <li>ERROR_NULL_ARGUMENT - if the image is null</li>
+ *    <li>ERROR_INVALID_ARGUMENT - if the image has been disposed</li>
+ *    <li>ERROR_INVALID_ARGUMENT - if any of the width or height arguments are negative.
+ * </ul>
+ * @exception SWTException <ul>
+ *    <li>ERROR_GRAPHIC_DISPOSED - if the receiver has been disposed</li>
+ * </ul>
+ * @exception SWTError <ul>
+ *    <li>ERROR_NO_HANDLES - if no handles are available to perform the operation</li>
```
fedejeanne: Where is this one thrown? I don't see it in the immediate code, so it must be in a call to another method, right?

fedejeanne:
> Question: why add a new API method, who needs it?

Answer to myself: it's in our internal issue (https://github.com/vi-eclipse/Eclipse-Platform/issues/459)

> A new API in GC is required to draw images at specified size. This is required for https://github.com/vi-eclipse/Eclipse-Platform/issues/326 .

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
arunjose696: The error would be thrown from other methods, The exception is copied from` public void drawImage(Image image, int srcX, int srcY, int srcWidth, int srcHeight, int destX, int destY, int destWidth, int destHeight)`, because this method and the drawImage overload with source coordinates follow the same internal drawing path; the only difference is that here the source dimensions are calculated from the image.

fedejeanne: 👍

fedejeanne (commented):
Changes make sense. 

Since there are no usages of this API, I can't test it and therefore I can't really _approve_ it. I don't oppose it though 👍

arunjose696:
> Changes make sense.
> 
> Since there are no usages of this API, I can't test it and therefore I can't really _approve_ it. I don't oppose it though 👍

I have now added a usage in the CTabFolder renderer. To test it, you can run an Eclipse runtime workspace:

- In open files, the unselected ones are drawn using the updated drawImage call in drawUnselected(). This now uses the new API, and you can verify that it has no visual impact on the image.
- The drawBackground method was also updated to use the new API, which made it possible to remove the getBounds() call.

For adapting the remaining usages I would be creating an issue

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/GC.java
```diff
@@ -1189,11 +1189,63 @@ public void drawImage(Image image, int srcX, int srcY, int srcWidth, int srcHeig
...
+ * @exception SWTError
+ *                                     <ul>
+ *                                     <li>ERROR_NO_HANDLES - if no handles are
+ *                                     available to perform the operation</li>
+ *                                     </ul>
+ * @since 3.132
+ */
+public void drawImage(Image image, int destX, int destY, int destWidth, int destHeight) {
+	if (handle == null)
+		SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
```
HeikoKlare: Please put expression blocks into braces when adding new code (applies to all added if statements).

arunjose696: Done

arunjose696:
I have also added the tests for this new API similar to other drawImage() calls

akoch-yatta:
Main purpose for this additional API is to get rid of any Image#getBounds calls in advance as this could result in creating unnecessary handles in windows. Additionally, most calls to this API are the exact use case of the new API: draw an Image scaled - so the original method should mainly be used to draw a dropped & scaled Image.

PR #2549, opened 2025-09-25
Make test valid for backspace/delete

jonahgraham:
I noticed this test taking a very long time (10 seconds). The test as written is invalid because the loop has a timeout with hitting the exit condition.

The "a" is never inserted into the text box (at least on gtk3), so the sending of back space appears to work when it didn't

I guess this isn't for merging until the test can be properly fixed. I am creating PR so I can see if it reports failing on build system in the same way I see it failing locally.

### DIFF
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_custom_StyledText.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_custom_StyledText.java
@@ -5877,6 +5877,8 @@ public void test_backspaceAndDelete() {
 		}
 	}
 
+	assertEquals(1, text.getText().length());
+
 	// Simulate the backspace, ensuring that the caret is in the correct position
 	text.invokeAction(ST.DELETE_PREVIOUS);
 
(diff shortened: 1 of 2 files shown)


jonahgraham:
The test fails on my machine because `display.post()` fails because post cannot find a window with Display.findFocusedWindow() returning 0.

jonahgraham:
I think this test would have been disabled years ago if it hadn't false positive'd at the time that Display.post tests were all disabled in [Bug 553754](https://bugs.eclipse.org/bugs/show_bug.cgi?id=553754) commit b846e85a7f2791357f34bfcd3174b2a8e7cd6073

jonahgraham:
The tests failed locally for me, but they pass on the build machine for all the targets currently tested.

akurtakov:
Do you run on Wayland or X11? Note that XWayland inherits all security changes of Wayland so one has to use plain X11 for real test. Test machines are running on X11 only AFAIK.

jonahgraham:
Yes, I think that is probably the issue. I will set up an x11 only run to see what happens.

jonahgraham:
These tests will now show as a failure on gtk4 - which is fine because it was previously showing as a false pass (for me at least).

Thanks @akurtakov - you were right, setting up a real x11 backend shows the expected behaviour.

I have update the top comment with the new commit message as I cleaned up theses two tests, and learned a bunch of new stuff too.

jonahgraham:
> Yes, I think that is probably the issue. I will set up an x11 only run to see what happens.

I documented my setup in https://kichwacoders.com/2025/09/30/testing-and-developing-swt-on-gtk/

jonahgraham:
The updated tests [pass on GitHub actions](https://github.com/eclipse-platform/eclipse.platform.swt/actions/runs/18140049529/job/51628417423), all platforms. But failed on Windows I-build https://download.eclipse.org/eclipse/downloads/drops4/I20250930-1800/testResults.php - so more work (and more for me to learn about different platforms)

PR #2550, opened 2025-09-25
Increase timeout for RMI service being registered in Clipboard test

jonahgraham:
The test failed to get the RMI within 1 second on macOS on GitHub actions one time, so increase the timeout here.

Fixup for https://github.com/eclipse-platform/eclipse.platform.swt/pull/2541


jonahgraham:
All failed tests are covered by #2516 

Test_org_eclipse_swt_dnd_Clipboard.test_Remote (the one I saw fail on macOS) takes 1.8 seconds on Linux, 6.5s on macOS (aarch+x64 similar times) and 8.7s on Windows.

So the fact that the original 1 second timeout was too short was a good guess and I am more confident in stability with a 10 second test.

PR #2551, opened 2025-09-26
When computing the size of a widget always return values rounded up

amartya4256:
When converting pixels to point there can be-  depending on the zoom level - a fractional result. Currently, in all cases the result is converted into an integer using `Math.round()` that will make values `+/-0.5` resulting in small values to be rounded towards a smaller value.
While it is maybe valid for a _location_, when using points to express a _dimension_ this is not okay as it will result in the reported (integer) value to be to small leading to errors when the SWT API is then used after performing additional computations maybe.
See
- eclipse-platform#2381
- eclipse-platform#2166

This PR also resolves the issue https://github.com/eclipse-platform/eclipse.platform.swt/issues/2166

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Point.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Point.java
@@ -139,13 +139,20 @@ public static sealed class OfFloat extends Point permits Point.WithMonitor {
 	private static final long serialVersionUID = -1862062276431597053L;
 
 	public float residualX, residualY;
+	private final RoundingMode roundingMode;
 
 	public OfFloat(int x, int y) {
 		super(x, y);
+		this.roundingMode = null;
 	}
 
 	public OfFloat(float x, float y) {
-		super(Math.round(x), Math.round(y));
+		this(x, y, RoundingMode.ROUND);
+	}
+
+	public OfFloat(float x, float y, RoundingMode roundingMode) {
+		super(roundingMode.round(x), roundingMode.round(y));
+		this.roundingMode = roundingMode;
 		this.residualX = x - this.x;
 		this.residualY = y - this.y;
 	}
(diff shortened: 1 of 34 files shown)


ShahzaibIbrahim:
I cannot un draft it as this PR was not created by me. I also have tested the original issue after my changes: https://github.com/eclipse-platform/eclipse.platform.swt/issues/2166. Also tested runtime workspace if something is broken. 

The commit message is needed to adapted and also some API check failures need to be fixed, but feel free to review/test it.

@HeikoKlare @akoch-yatta

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Point.java
```diff
@@ -139,13 +139,19 @@ public static sealed class OfFloat extends Point permits Point.WithMonitor {
 	private static final long serialVersionUID = -1862062276431597053L;
 
 	public float residualX, residualY;
+	public RoundingMode roundingMode;
```
ShahzaibIbrahim: @akoch-yatta this field was added as public while it is only used in this class internally. Was there a reason for that? Why not simply have it private?

akoch-yatta: Probably because it was work in progress. If it can be private, make it private

akoch-yatta (commented):
Checking all changed calls in this PR, they look semantically correct, regarding the different rounding for sizes and locations

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Point.java
laeubi: it should at least be final ... on the other hand residualX/y are public modifiable as well... I think one best make both private.

laeubi (approved):
@amartya4256 thanks for taking over this and working on this, it looks all sane for me.

ShahzaibIbrahim:
Changing residualX, residualY to private will be done in another PR and test failures are unrelated: https://github.com/eclipse-platform/eclipse.platform.swt/issues/2516.

From my side this PR is ready to be merge if no other objections.

akoch-yatta:
Test are still failing after rebuild, but unrelated #2516

PR #2556, opened 2025-09-29
[win32] Fix pointer size scaling precision for custom cursors

ShahzaibIbrahim:
The value in the registry for CursorBaseSize is not always divisible by 32, so integer division was causing loss of precision in pointer scaling. This commit changes getPointerSizeScaleFactor() to return a float using floating-point division, and ensures all usages of this method in Cursor.java handle the scale as a float. The float is then cast to int only when calling DPIUtil.scaleImageData, which only accepts integer zoom values. This preserves as much precision as possible and avoids rounding errors in pointer scaling for accessibility settings.

**Description**
Recently I commited: https://github.com/eclipse-platform/eclipse.platform.swt/commit/6e4241da7ffbe82ca6e307a794bcb8b5be661a00, where we read the cursor size value from accessibility setting of windows. I assumed that each value is multiple of 32 which was wrong. 

Here for the cursor size 4 is 80
Image

Image

and my method will return 2 (i assumed it should return 4). Changing it to float will return us 2.4 which is more precise when trying to scale up the size.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
@@ -352,14 +352,14 @@ private void setHandleForZoomLevel(CursorHandle handle, Integer zoom) {
  *         size, etc.)
  */
 
-private static int getPointerSizeScaleFactor() {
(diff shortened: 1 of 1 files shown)


ShahzaibIbrahim:

fedejeanne (approved):

fedejeanne:
Test failures are unrelated

PR #2558, opened 2025-09-29
Adding readAndDispatch() in drawImage test to make  the test not flaky

arunjose696:
The reasoning to add readAndDispatch() in test_drawImage_nonAutoScalableGC_bug_2504 is to consider the case when e.gc.drawImage()'s draw request to the OS, might not have completed by the time gcCopy.copyArea() was called. Using readAndDispatch() is to ensure all OS events are processed, so there’s no pending work left.


iloveeclipse (approved):

iloveeclipse:
Windows test failures are known and unrelated, see https://github.com/eclipse-platform/eclipse.platform.swt/issues/2516

iloveeclipse:
Thanks!

PR #2560, opened 2025-09-30
Fix warning about using static directly in tests

akurtakov:


akurtakov:
Ignore unstable test on Windows that can't be affected by this PR.

PR #2563, opened 2025-09-30
Deprecation: Cursor(Device, ImageData, ImageData, int, int)

ShahzaibIbrahim:
The constructor Cursor(Device, ImageData source, ImageData mask, int hotspotX, int hotspotY) allows creating a cursor by providing a source image and a 1-bit mask to define transparency. This approach requires developers to manually define mask pixels and works reliably only in fixed DPI environments.

With the introduction of multi-monitor DPI support, this constructor produces inconsistent results.

- Masks do not scale properly when using DPIUtil.autoScaleImageData, leading to visual artifacts.
- Developers need to handle scaling and mask adjustments manually, which is error-prone.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Cursor.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Cursor.java
@@ -276,7 +276,10 @@ public Cursor(Device device, int style) {
  * </ul>
  *
  * @see #dispose()
+ *
+ * @deprecated Use {@link #Cursor(Device, ImageDataProvider, int, int)} instead.
  */
+@Deprecated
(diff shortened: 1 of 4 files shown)


ShahzaibIbrahim:
@fedejeanne those warnings are all from the test. What I can do now is to suppress the warning for deprecation on those test.

fedejeanne (approved):

akurtakov:
It is a good practice before deprecating some method to go through Eclipse TLP codebase and switch to new way. Otherwise we end up with pile of deprecation warnings that make working on the codebase a nightmare. E.g. https://github.com/eclipse-platform/eclipse.platform.ui/pull/3278/checks?check_run_id=51871277212 . Can I count on you doing this work?

HeikoKlare:
Within the SDK, I still see new warnings in `Snippet92` and `Snippet386` caused by this change. @ShahzaibIbrahim @fedejeanne do you plan to clean them up?

Issue #2557, opened 2025-09-29
Test_org_eclipse_swt_graphics_GC.test_drawImage_nonAutoScalableGC_bug_2504 fails since I20250926-0030 on Linux


iloveeclipse:
`Test_org_eclipse_swt_graphics_GC.test_drawImage_nonAutoScalableGC_bug_2504` fails since I20250926-0030 on Linux.

Regression from d5273ee004f362a6541a9553969074b978cd7d21 / https://github.com/eclipse-platform/eclipse.platform.swt/pull/2507
See  
- https://download.eclipse.org/eclipse/downloads/drops4/I20250926-0030/
- https://download.eclipse.org/eclipse/downloads/drops4/I20250926-0030/testresults/html/org.eclipse.swt.tests_ep438I-unit-linux-x86_64-java21_linux.gtk.x86_64_21.html

```
Bottom-right pixel is not black! when source GC is not autoScalable ==> expected: <RGB {0, 0, 0}> but was: <RGB {255, 255, 255}>

org.opentest4j.AssertionFailedError: Bottom-right pixel is not black! when source GC is not autoScalable ==> expected: <RGB {0, 0, 0}> but was: <RGB {255, 255, 255}>
at org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:151)
at org.junit.jupiter.api.AssertionFailureBuilder.buildAndThrow(AssertionFailureBuilder.java:132)
at org.junit.jupiter.api.AssertEquals.failNotEqual(AssertEquals.java:197)
at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:182)
at org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:1156)
at org.eclipse.swt.tests.junit.Test_org_eclipse_swt_graphics_GC.test_drawImage_nonAutoScalableGC_bug_2504(Test_org_eclipse_swt_graphics_GC.java:293)
at java.base/java.lang.reflect.Method.invoke(Method.java:580)
at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
```

iloveeclipse:
@HeikoKlare, @arunjose696 : please check.

arunjose696:
This regression is in the test added in the same commit

[...]

PR #2565, opened 2025-09-30
Adding barrier and timeout in failing test test_drawImage_nonAutoScalableGC_bug_2504 due to test failing on Linux

arunjose696:
Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2557 for the time being

### DIFF
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_GC.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_GC.java
@@ -30,6 +30,7 @@ import static org.junit.jupiter.api.Assumptions.assumeFalse;
 import static org.junit.jupiter.api.Assumptions.assumeTrue;
 
 import java.lang.ref.WeakReference;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 
 import org.eclipse.swt.SWT;
@@ -249,7 +250,7 @@ public void test_dispose() {
 
 
 @Test
-public void test_drawImage_nonAutoScalableGC_bug_2504() {
+public void test_drawImage_nonAutoScalableGC_bug_2504() throws InterruptedException {
 	Shell shell = new Shell(display);
     float targetScale = 2f;
     int srcSize = 50;
@@ -273,15 +274,18 @@ public void test_drawImage_nonAutoScalableGC_bug_2504() {
     int canvasWidth = Math.round(bounds.width * targetScale);
     int canvasHeight = Math.round(bounds.height * targetScale);
     canvas.setSize(canvasWidth, canvasHeight);
+    AtomicBoolean paintListenerCalled = new AtomicBoolean();
     canvas.addPaintListener(e -> {
         e.gc.drawImage(image, 0, 0, bounds.width, bounds.height,
                        0, 0, canvasWidth, canvasHeight);
+        paintListenerCalled.set(true);
     });
 
     shell.open();
-    while (display.readAndDispatch()) {
-    }
-    Image target = new Image(display, canvasWidth, canvasHeight);
+	int timeoutMillis = 2000;
+	SwtTestUtil.processEvents(timeoutMillis, () -> paintListenerCalled.get());
+	assertTrue(paintListenerCalled.get(), "paint listener was never called");
+	Image target = new Image(display, canvasWidth, canvasHeight);
     GC gcCopy = new GC(canvas);
     gcCopy.copyArea(target, 0, 0);
     gcCopy.dispose();


tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_GC.java
```diff
@@ -249,6 +250,7 @@ public void test_dispose() {
 
 
 @Test
+@Disabled("Disabled due to test failing on Linux https://github.com/eclipse-platform/eclipse.platform.swt/issues/2557")
```
iloveeclipse: you could disable this for Linux only, assuming the problem is Linux specific

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_GC.java
```diff
@@ -250,6 +250,8 @@ public void test_dispose() {
 
 @Test
 public void test_drawImage_nonAutoScalableGC_bug_2504() {
+	assumeTrue(!System.getProperty("os.name").toLowerCase().contains("linux"),
```
akurtakov: Please use SwtTestUtil.isGTK here.

HeikoKlare:
Since the issue only occurs in I-Build test runs, what would be the plan to proceed when disabling the test now? The test would need to be reenabled with a different solution concept afterwards anyway, so shouldn't we proceed with trying to find a proper solution? I don't think it really matters if the test fails/succeeds for M1 or not, does it? So we don't need to disable it for the M1 build just to reenable it afterwards again and proceed with testing different solution attempts.

arunjose696:
> what would be the plan to proceed when disabling the test now? The test would need to be reenabled with a different solution concept afterwards anyway, so shouldn't we proceed with trying to find a proper solution? 

As the test fails only a specific machine the i-build runs in we could create a low priority issue in the swt repo with a help-wanted tag so someone who could reproduce this locally on their machine could take a look.

> I don't think it really matters if the test fails/succeeds for M1 or not, does it?

I assumed the tests failing would block other tests on SDK builds, I see the remaining tests are getting executed so I think it does not matter. So could try experimenting for a fix if this wouldn't have an impact for M1.

iloveeclipse:
> I don't think it really matters if the test fails/succeeds for M1 or not, does it

In general, if there is no clue how to fix a test failure, it would be nice to create ticket (done) and disable the test.
This helps (at least me) to identify **new** issues while checking SDK test results every day.

HeikoKlare:
> In general, if there is no clue how to fix a test failure, it would be nice to create ticket (done) and disable the test.

I agree. But we have ideas how to fix the failure documented in https://github.com/eclipse-platform/eclipse.platform.swt/issues/2557.
In my understanding, the reason for disabling the test now was to have it silenced for M1 (https://github.com/eclipse-platform/eclipse.platform.swt/issues/2557#issuecomment-3355258874), but to proceed with testing the fix proposal we need to reenable it again anyway. And I don't see a need to consider M1 as anything special here and propose to directly try the other fix proposals. If all of them fail and we have no clue how to proceed, we should of course disable the test.

iloveeclipse:
> But we have ideas how to fix the failure documented in #2557.

Ok, that wasn't clear tome. And for sure no need to disable test only for M1.

arunjose696:
I have updated this to add a timeout, this would need merging to master to test if the fix works.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_GC.java
```diff
@@ -273,14 +273,21 @@ public boolean isAutoScalable() {
...
 
     shell.open();
-    while (display.readAndDispatch()) {
-    }
+    long startTime = System.currentTimeMillis();
+    int timeoutMillis = 2000;
+	while (!paintListenerCalled[0] || System.currentTimeMillis() - startTime < timeoutMillis) {
+		while (display.readAndDispatch()) {
+			display.sleep();
+		}
```
HeikoKlare: Without the condition inversion, this has the change to lock the application:
```suggestion
		if (!display.readAndDispatch()) {
			display.sleep();
		}
```

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_GC.java
```diff
@@ -273,14 +273,21 @@ public boolean isAutoScalable() {
...
+	}
```
HeikoKlare: There should be an assert after the loop to be able to distinguish between whether the listener was not called or the timeout occurred.

jonahgraham: Please have a look if this does what you want, saving re-writing this type of loop:


```suggestion
    SwtTestUtil.processEvents(10000, () ->  paintListenerCalled[0]);
```

jonahgraham: PS You probably want this line after too:

```java
assertTrue(paintListenerCalled[0])
```

in case `processEvents` has a timeout without the condition passing.

HeikoKlare: > `SwtTestUtil.processEvents(10000, () ->  paintListenerCalled[0]);`

Thank you for the pointer! I was looking for that kind of util in SWT but somehow only found them in Platform and Platform UI.

> PS You probably want this line after too:
> 
> ```java
> assertTrue(paintListenerCalled[0])
> ```

Indeed: https://github.com/eclipse-platform/eclipse.platform.swt/pull/2565#discussion_r2395589962 🙂

HeikoKlare:
I've adapted the PR according to the comments. In case the checks finally succeed, I plan to merge today to get further feedback from today's I-Build.

PR #2567, opened 2025-09-30
[GTK4] Fix Test*Shell.test_setSizeCustomResize test failure

akurtakov:
Translating coordinates from client area to GtkWindow for Gtk 4.x made the wrong assumption that it should reduce allocation.height with the offset (header/titlebar height) which makes no sense as it shrinks the window vertically.


jonahgraham:
I see the test failure is a Clipboard related test. I'll create a PR/issue for that soon.

akurtakov:
Merging as it's actually reducing the failures on GTK 4.

PR #2570, opened 2025-10-01
Adding loadImageData(zoom) to the AbstractImageProvider Wrapper

arunjose696:
This refactoring is required to load unscaled ImageData for all wrappers The loadImageData for the added functions would have similar logic to newImageData() except the imageData returned would not be scaled but would be an ElementAtZoom.

This is required for https://github.com/eclipse-platform/eclipse.platform.swt/pull/2526#discussion_r2392464120


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1939,14 +1939,20 @@ public Collection<Integer> getPreservedZoomLevels() {
 		return Collections.emptySet();
 	}
 
+	protected abstract ElementAtZoom<ImageData> loadImageData(int zoom);
+
 	abstract ImageData newImageData(ZoomContext zoomContext);
 
 	abstract AbstractImageProviderWrapper createCopy(Image image);
 
 	ImageData getScaledImageData (int zoom) {
```
HeikoKlare, line 1948: This calls `getClosestAvailableImageData()` twice. It does not seem to be a performance issue as at least the second call should only do lookups, but still it seems better to just make the call once and use the stored results twice.

arunjose696: Done

akoch-yatta (approved):
Changes look good to as preparatory work for #2526 and testing didn't show me any regressions

PR #2573, opened 2025-10-01
Control.flushQueueOnDnd: simplify commit

tmssngr:
Inlined `sendOrPost` for this special case making it more obvious that an (immediate) `sendEvent` is happening.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Control.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Control.java
@@ -4849,9 +4849,9 @@ void flushQueueOnDnd() {
 	// to determine DnD threshold.
 	// This is to preserve backwards Cocoa/Win32 compatibility.
 	Event mouseDownEvent = dragDetectionQueue.getFirst();
-	mouseDownEvent.data = Boolean.valueOf(true); // force send MouseDown to avoid subsequent MouseMove before MouseDown.
+	mouseDownEvent.data = null;
 	dragDetectionQueue = null;
-	sendOrPost(SWT.MouseDown, mouseDownEvent);
+	sendEvent(SWT.MouseDown, mouseDownEvent);
 }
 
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Control.java
```diff
@@ -4861,9 +4861,9 @@ void flushQueueOnDnd() {
 	// to determine DnD threshold.
 	// This is to preserve backwards Cocoa/Win32 compatibility.
 	Event mouseDownEvent = dragDetectionQueue.getFirst();
-	mouseDownEvent.data = Boolean.valueOf(true); // force send MouseDown to avoid subsequent MouseMove before MouseDown.
+	mouseDownEvent.data = null;
```
akurtakov, line 4852: Setting data to null is not needed too. Please remove it, it might have the positive side effect of not losing any other data that could have been there.

tmssngr: It is taken from `sendOrPost`.

merks: Yes, formerly it was setting true into data so that this method would do a send and this method does also set data to null:

https://github.com/eclipse-platform/eclipse.platform.swt/blob/0e04bc92028cdc32d8378e9aa4b63f3b50e62efe/bundles/org.eclipse.swt/Eclipse%20SWT/gtk/org/eclipse/swt/widgets/Control.java#L5128-L5140

But I don't think there is a compelling reason to set data to null to maybe replace some other value.  If there is another value in data (in the future), replacing it with null is like to be unhelpful.

tmssngr: You propose to change `event.data` while I want to keep it null as is. Maybe some application code relies on having `event.data` set null?

Issue #2119, opened 2025-05-07
[GTK] Reports wrong mouse coordinates when starting a drag


tmssngr:
**Describe the bug**
A drag is detected after moving the mouse a few pixels. But SWT reports the drag with the coordinates *after* the drag was detected, not the coordinates where the left mouse button was pressed. This causes wrong elements to be dragged in our application.

**To Reproduce**
```java
import org.eclipse.swt.*;
import org.eclipse.swt.dnd.*;
import org.eclipse.swt.graphics.*;
import org.eclipse.swt.widgets.*;

public class DragDropPositions {

	public static void main(String[] args) {
		final Display display = new Display();

		final Shell shell = new Shell(display);

		final Listener listener = event -> {
			switch (event.type) {
			case SWT.MouseMove -> System.out.println("move: " + new Point(event.x, event.y));
			case SWT.MouseDown -> System.out.println("down: " + new Point(event.x, event.y));
			case SWT.KeyDown -> System.out.println("key " + shell.toControl(display.getCursorLocation()));
			}
		};
		shell.addListener(SWT.MouseMove, listener);
		shell.addListener(SWT.MouseDown,listener);
		shell.addListener(SWT.KeyDown,listener);

		final DragSource dragSource = new DragSource(shell, DND.DROP_MOVE);
		dragSource.setTransfer(TextTransfer.getInstance());
		dragSource.addDragListener(new DragSourceAdapter() {
			@Override
			public void dragStart(DragSourceEvent event) {
				System.out.println("drag start " + shell.toControl(event.x, event.y));
			}

			@Override
			public void dragSetData(DragSourceEvent event) {
				event.data = new String[] {""};
			}
		});
		final DropTarget target = new DropTarget(shell, DND.DROP_MOVE);
		target.setTransfer(TextTransfer.getInstance());
		target.addDropListener(new DropTargetAdapter() {
			@Override
			public void dragEnter(DropTargetEvent event) {
				System.out.println("drop enter " + shell.toControl(event.x, event.y));
				event.detail = DND.DROP_NONE;
			}

			@Override
			public void drop(DropTargetEvent event) {
				event.detail = DND.DROP_NONE;
			}
		});

		shell.setSize(400, 300);
		shell.open();

		while (!shell.isDisposed()) {
			if (!display.readAndDispatch()) {
				display.sleep();
			}
		}

		display.dispose();
	}
}
```
- start this snippet
[...]

PR #2574, opened 2025-10-01
[GTK] Reports wrong mouse coordinates when starting a drag #2119

tmssngr:
Control: don't send MouseMove events with stateMask set for BUTTON1 before the MouseDown event was sent

These events are already added to the dragDetectionQueue, so there is no need to send/post them now.


tmssngr:
What does my PR need to get accepted?

akurtakov:
I have rebased it before christmas and it got lost in the myriad of ongoing things. It's a good patch. I've rebased it again and plan to merge it soon.
Sorry about the delay. There are less and less people spending time doing reviews.

tmssngr:
@akurtakov I appreciate your effort. Have a nice weekend.

Issue #2568, opened 2025-09-30
Unknown test (org.eclipse.swt.tests.junit.Test_org_eclipse_swt_dnd_Clipboard) with error


jonahgraham:
The new clipboard test introduced an occasional test failure on Windows which I believe is down to a race condition on shutting down the remote test program. Since on Windows you can't delete files when a process is still running this race condition causes a test failure.

Example failure from https://github.com/eclipse-platform/eclipse.platform.swt/runs/51615302452

    Failed to close extension context
    org.junit.platform.commons.JUnitException: Failed to close extension context
    	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
    Caused by: java.io.IOException: Failed to delete temp directory C:\Users\RUNNER~1\AppData\Local\Temp\junit-11184487295439766883. The following paths could not be deleted (see suppressed exceptions for details): <root>
    	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:184)
    	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
    	at java.base/java.util.stream.SortedOps$RefSortingSink.end(SortedOps.java:395)
    	at java.base/java.util.stream.Sink$ChainedReference.end(Sink.java:261)
    	at java.base/java.util.stream.Sink$ChainedReference.end(Sink.java:261)
    	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:510)
    	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
    	at java.base/java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:151)
    	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateSequential(ForEachOps.java:174)
    	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
    	at java.base/java.util.stream.ReferencePipeline.forEach(ReferencePipeline.java:596)
    	... 1 more
    	Suppressed: java.nio.file.FileSystemException: C:\Users\RUNNER~1\AppData\Local\Temp\junit-11184487295439766883: The process cannot access the file because it is being used by another process
    		at java.base/sun.nio.fs.WindowsException.translateToIOException(WindowsException.java:92)
[...]

PR #2575, opened 2025-10-01
Avoid failing test because temporary directory cannot be deleted

jonahgraham:
On Windows the temporary directory sometimes cannot be deleted because the files are still locked and you get an error like:

The process cannot access the file because it is being used by another process

This change allows the deletion of the temp directory to fail without failing the whole test. In case the temp dir can't be deleted a warning is printed on System.err

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2568


laeubi:
@jonahgraham maybe something to report to junit team so `@TempDir` might retry a bit longer or not fail completely?

jonahgraham:
Already discussions in JUnit about it - see link in the issue https://github.com/eclipse-platform/eclipse.platform.swt/issues/2568#issuecomment-3353005711

Issue #2571, opened 2025-10-01
test_backspaceAndDelete StyledText and Text fail on I-build machine


jonahgraham:
In https://github.com/eclipse-platform/eclipse.platform.swt/pull/2549 I updated tests and they [pass on GitHub actions](https://github.com/eclipse-platform/eclipse.platform.swt/actions/runs/18140049529/job/51628417423), all platforms. But failed on Windows I-build https://download.eclipse.org/eclipse/downloads/drops4/I20250930-1800/testResults.php - so more work (and more for me to learn about different platforms)

Failure looks like:
  
    Multiple Failures (2 failures) java.lang.AssertionError: <no message> java.lang.RuntimeException: AwtScreenshot VM finished with exit code 1: Exception in thread "main" java.awt.AWTError: no screen devices at java.desktop/sun.awt.Win32GraphicsEnvironment.getDefaultScreenDevice(Win32GraphicsEnvironment.java:101) at java.desktop/java.awt.Robot.<init>(Robot.java:135) at org.eclipse.test.AwtScreenshot.main(AwtScreenshot.java:43)
    
    org.opentest4j.MultipleFailuresError: Multiple Failures (2 failures)
    java.lang.AssertionError: <no message>
    java.lang.RuntimeException: AwtScreenshot VM finished with exit code 1:
    Exception in thread "main" java.awt.AWTError: no screen devices
    at java.desktop/sun.awt.Win32GraphicsEnvironment.getDefaultScreenDevice(Win32GraphicsEnvironment.java:101)
    at java.desktop/java.awt.Robot.<init>(Robot.java:135)
    at org.eclipse.test.AwtScreenshot.main(AwtScreenshot.java:43)
    
    Suppressed: java.lang.AssertionError
    at org.junit.Assert.fail(Assert.java:87)
    at org.junit.Assert.assertTrue(Assert.java:42)
    at org.junit.Assert.assertTrue(Assert.java:53)
    at org.eclipse.swt.tests.junit.Test_org_eclipse_swt_custom_StyledText.test_backspaceAndDelete(Test_org_eclipse_swt_custom_StyledText.java:5916)
    at java.base/java.lang.reflect.Method.invoke(Method.java:580)
    Suppressed: java.lang.RuntimeException: AwtScreenshot VM finished with exit code 1:
    Exception in thread "main" java.awt.AWTError: no screen devices
[...]

PR #2576, opened 2025-10-01
Re-enable Display.post tests

jonahgraham:
These tests have been disabled overtime because making sure they pass consistently on Jenkins is difficult (See
[407862](https://bugs.eclipse.org/bugs/show_bug.cgi?id=407862) and [553754](https://bugs.eclipse.org/bugs/show_bug.cgi?id=553754))

With GitHub actions these problems seem to be less of an issue, so we should enable these tests again, at least on GHA.

This first commit is a test to see if they pass on GHA consistently, if so adding a new check so that tests can be skipped on Jenkins releng infra can be added.

Part of https://github.com/eclipse-platform/eclipse.platform.swt/issues/2571


jonahgraham:
This may trigger the same issue as https://github.com/eclipse-platform/eclipse.platform.swt/issues/2571 across more tests on the I-build. If so, I will re disable the tests, but only on the Jenkins machines.

PR #2591, opened 2025-10-03
Avoid reading org.eclipse.swt.internal.gdk.backend too early

jonahgraham:
The value of `org.eclipse.swt.internal.gdk.backend` is not set until a display is created at least once since this value is assigned in the Display constructor. Normally tests are run in such an order that SwtTestUtil isn't accessed until after the first Display is created. Make sure to create a display before calling SwtTestUtil.isX11.getAsBoolean().

This is a fix for a regression caused by moving the setting of org.eclipse.swt.internal.gdk.backend from OS to Display in 5d67ce66d23165bd252339134ca86889b9594b63. That commit fixed an unrelated bug, but sometimes bad initialization of isX11 is an unintended side effect causing tests to be skipped or run in unexpected ways since isX11 was not correct.

### DIFF
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/SwtTestUtil.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/SwtTestUtil.java
@@ -103,7 +103,16 @@ public class SwtTestUtil {
 	/** Useful if you want some tests not to run on Jenkins with user "genie.platform" */
 	public final static boolean isRunningOnContinousIntegration = isGTK && ("genie.platform".equalsIgnoreCase(System.getProperty("user.name")));
 
-	public final static boolean isX11 = isGTK
+	/**
+	 * The value of org.eclipse.swt.internal.gdk.backend is not set until a display
+	 * is created at least once since this value is assigned in the Display
+	 * constructor. Normally tests are run in such an order that SwtTestUtil isn't
+	 * accessed until after the first Display is created.
+	 *
+	 * Rather than requiring all users of SwtTestUtil to be careful with their access,
+	 * instead defer checking x11 by using a supplier to defer getting the value.
+	 */
+	public final static BooleanSupplier isX11 = () -> isGTK
 			&& "x11".equals(System.getProperty("org.eclipse.swt.internal.gdk.backend"));
 	public final static boolean isGTK4 = isGTK
 			&& System.getProperty("org.eclipse.swt.internal.gtk.version", "").startsWith("4");
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_dnd_Clipboard.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_dnd_Clipboard.java
@@ -351,7 +351,7 @@ public class Test_org_eclipse_swt_dnd_Clipboard {
 			String result = runOperationInThread(remote::getStringContents);
 			assertEquals(helloWorld, result);
 		} catch (Exception | AssertionError e) {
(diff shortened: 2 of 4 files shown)


jonahgraham:
I think some of the tests that were supposed to have been guarded by isX11 may have bitrotten, for example test_activateEventSend fails on my machine now that I have fixed the isX11 test.

jonahgraham:
> for example test_activateEventSend fails on my machine now that I have fixed the isX11 test.

It also fails on the build machine. A fix for the unstable test is in https://github.com/eclipse-platform/eclipse.platform.swt/pull/2592

jonahgraham:
I'm deferring resolving this until #2593 is complete as this change makes the tests look unstable, but it really is just exposing some long non-running tests.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
```diff
@@ -1211,6 +1211,9 @@ void createDisplay (DeviceData data) {
 		xDisplay = GTK.GTK4 ? 0 : GDK.gdk_x11_get_default_xdisplay();
 		// set GDK backend if we are on X11
 		System.setProperty("org.eclipse.swt.internal.gdk.backend", "x11");
+	} else {
```
jonahgraham: @akurtakov please have a look at this part of the change - is it ok to start setting org.eclipse.swt.internal.gdk.backend like this? Nothing in eclipse platform, except SwtTestUtil seems to use this property, so this change seems safe.

Nothing [in all of github's open source code](https://github.com/search?q=org.eclipse.swt.internal.gdk.backend&type=code) uses it either. But I thought I should check with you in case you knew of something else.

akurtakov: It is not true that wayland is the only option. Even if we ignore quartz, win32 as swt/gtk is not used on mac/windows - it could be smth else e.g.  broadway (https://jayjaybillings.com/2016/01/02/launching-eclipse-in-a-browser-with-the-gtk-broadway-backend/). 
I don't mind setting it to wayland after a check it's actually wayland.

akurtakov: Actually, it would be nice to have the property reported in the about dialog like other gtk details are 

Image

jonahgraham: I added it, as isWayland is current !isX11 it will only ever be x11 or wayland in current implementation. I added `unknown` as value on currently unreachable else block.

akurtakov: I'll look into making isWayland a real check.

jonahgraham:
Rebased - especially to make sure full test suite behaves as expected with recently merged 06e5e83d976652dc1815f64af6b19f61a3b11e65 / #2605

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
```diff
@@ -1211,6 +1211,14 @@ void createDisplay (DeviceData data) {
 		xDisplay = GTK.GTK4 ? 0 : GDK.gdk_x11_get_default_xdisplay();
 		// set GDK backend if we are on X11
 		System.setProperty("org.eclipse.swt.internal.gdk.backend", "x11");
+	} else if (OS.isWayland()) {
+		// only other option for backend is Wayland at the moment
+		System.setProperty("org.eclipse.swt.internal.gdk.backend", "wayland");
+	} else {
+		// unreachable at the moment since isWayland is defined as !isX11
```
jonahgraham: This comment is now out of date.

PR #2592, opened 2025-10-03
Process events until parent shell is activated

jonahgraham:
It may take some event cycles before the parent shell that is getting focus/activation back gets the notification. Therefore wait until the Activate listener is called

### DIFF
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_widgets_Shell.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_widgets_Shell.java
@@ -22,10 +22,12 @@ import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertSame;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.Assertions.fail;
+import static org.junit.jupiter.api.Assumptions.assumeTrue;
 
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.eclipse.swt.SWT;
 import org.eclipse.swt.events.SelectionAdapter;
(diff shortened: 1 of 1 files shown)


jonahgraham:
This + #2591 shows that the test still isn't stable. I am deferring resolving this until #2593 is complete.

jonahgraham:
Test commit with `RepeatedTest` of 1000 to see if this is stable or not on the build machine with the added processEvents

jonahgraham:
The test I am modifying was a regression test for https://bugs.eclipse.org/bugs/show_bug.cgi?id=436841, but only tested 1 of the 4 events. I added to the test all the events. The test was also written before GTK4 (???) and was disabled on non-x11. The test is now enabled on x11.

I'll updated the commit message to include the above on my next push.

jonahgraham:
> Test commit with `RepeatedTest` of 1000 to see if this is stable or not on the build machine with the added processEvents

The tests have successfully run 1000x, so I think this is ready to go.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_widgets_Shell.java
```diff
@@ -653,21 +655,48 @@ public void test_setBoundsLorg_eclipse_swt_graphics_Rectangle() {
...
-		int[] styles = {SWT.ON_TOP, SWT.APPLICATION_MODAL, SWT.PRIMARY_MODAL, SWT.SYSTEM_MODAL, SWT.NO_TRIM, SWT.BORDER, SWT.SHELL_TRIM};
-		for (int style : styles) {
-			Shell childShell = new Shell(testShell, style);
-			listenerCalled = false;
-			childShell.open();
-			childShell.dispose();
-			assertTrue(listenerCalled);
-		}
+public void test_activateEventSend() throws InterruptedException {
+	assumeTrue((SwtTestUtil.isGTK && SwtTestUtil.isX11) || SwtTestUtil.isGTK4,
```
jonahgraham, line 659: Note that because of https://github.com/eclipse-platform/eclipse.platform.swt/pull/2591 this assumption fails, even on GTK3 + x11. 

When I had done the 1000x repeated test I had included the fix in this PR as well to make sure it ran.

Once https://github.com/eclipse-platform/eclipse.platform.swt/pull/2591 is done this test will actually run properly.

In hindsight, I think I made it too complicated by having this as two PRs. Two commits in the one PR would have been sufficient (and simpler for me)

PR #2593, opened 2025-10-04
Convert tests to JUnit5

jonahgraham:
Because most all the tests remaining tests are in huge class hierarchy under Test_org_eclipse_swt_widgets_Widget they all need changing at once. The tests that are not under Test_org_eclipse_swt_widgets_Widget are generally pretty small, so included here to fully remove JUnit4 dependency.

This means that junit-vintage-engine and org.junit removed from Require-Bundle of the manifest.

Messages have been removed from assertions that didn't add value. The order of parameters to assertEquals and others changed, and for most of the calls I removed the message based on previous guidance. I tried to leave in messages that were more useful for documenting or diagnosing errors. But I removed messages that more or less repeated as a string what the item being tested is.

A lot of things work quite differently in JUnit5. Here are some highlights that have affected or benefited this change:

There is a better way of integrating screenshots into failed tests as tests can be intercepted between the test completion and the `@AfterEach` methods. Therefore we can take a screenshot prior to disposal, and then let the tear down message cleanup in a less convoluted way. See Test_org_eclipse_swt_widgets_Widget.afterTestExecutionCallback and tearDown methods to see this in action.

The way to get a test name is quite different. Instead of a TestName `@Rule` a `TestInfo` can be obtained. The name returned is not exactly the same format, but where the name needed a specific format, I adjusted the code appropriately. See Test_org_eclipse_swt_browser_Browser.testInfo and Test_org_eclipse_swt_widgets_Widget.testInfo for a couple of use cases.

Whole class parameterized tests don't really exist in a simple way in JUnit5. As there was very little use of this, rather than having complicated setup, I used sub-classes to provide the parameterization. See for example Test_org_eclipse_swt_widgets_DateTime and its new subclasses.

### DIFF
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/graphics/ImageDataTestHelper.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/graphics/ImageDataTestHelper.java
@@ -14,7 +14,7 @@
 
 package org.eclipse.swt.tests.graphics;
 
-import static org.junit.Assert.assertArrayEquals;
+import static org.junit.jupiter.api.Assertions.assertArrayEquals;
 
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/AllWidgetTests.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/AllWidgetTests.java
@@ -48,7 +48,9 @@ import org.junit.platform.suite.api.Suite;
 		Test_org_eclipse_swt_widgets_Composite.class, //
 		Test_org_eclipse_swt_widgets_CoolBar.class, //
 		// Failing test: Test_org_eclipse_swt_widgets_CoolItem.class, //
-		Test_org_eclipse_swt_widgets_DateTime.class, //
+		Test_org_eclipse_swt_widgets_DateTime_Style_CALENDAR.class, //
+		Test_org_eclipse_swt_widgets_DateTime_Style_DATE.class, //
+		Test_org_eclipse_swt_widgets_DateTime_Style_TIME.class, //
 		Test_org_eclipse_swt_widgets_DirectoryDialog.class, //
 		Test_org_eclipse_swt_widgets_Event.class, //
 		Test_org_eclipse_swt_widgets_ExpandBar.class, //
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/ImageTestUtil.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/ImageTestUtil.java
@@ -13,10 +13,10 @@
  *******************************************************************************/
 package org.eclipse.swt.tests.junit;
 
-import static org.junit.Assert.assertEquals;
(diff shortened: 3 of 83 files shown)


jonahgraham:
There are a couple of independent changes pulled out into PR #2591 and #2592 that I identified while refactoring the tests.

jonahgraham:
In addition to the top comment, there are a few reasons I would like to get this integrated if there are no objections.

1. Consistency - I have found it annoying to go between different classes and have to use different annotations/etc. Its pretty easy to mess things up because of using a JUnit4 annotation in a JUnit5 test and vice versa. By removing JUnit4 entirely that problem goes away.
2. I want to be able to use [`@Tag`s](https://docs.junit.org/current/user-guide/#writing-tests-tagging-and-filtering), like what I described in https://github.com/eclipse-platform/eclipse.platform.swt/issues/2126#issuecomment-3367784743
3. JUnit6 just came out this week - so we will presumably start migrating to that at some point. I think it would be nice to have at most 2 versions of JUnit co-existing instead of 3

jonahgraham:
TL;DR

The overall test counts look correct to me, this PR has expected number tests on Jenkins runs compared to latest master

---

The discovered test count has gone up a lot because of how I migrated the parameterized browser test. This means all the IE tests are [marked as skipped](https://ci.eclipse.org/releng/job/eclipse.platform.swt/job/PR-2593/2/testReport/org.eclipse.swt.tests.junit/Test_org_eclipse_swt_browser_Browser_IE/) on non-Windows platform instead of simply not existing in the test report. This increases the test count by 194 (The number of browser tests), which can be seen as the difference in total tests from latest [master](https://ci.eclipse.org/releng/job/eclipse.platform.swt/job/master/1352/testReport/) and this [PR](https://ci.eclipse.org/releng/job/eclipse.platform.swt/job/PR-2593/2/testReport/). On those same links you can see the total number of tests that pass is the same (4196).


I had originally identified a coupe of tests that were not running when I did the migration. See new `@Test` annotation on `Test_org_eclipse_swt_widgets_Shell.test_setLocationLorg_eclipse_swt_graphics_Point()`, `Test_org_eclipse_swt_widgets_Shell.test_setLocationII()`, and `Test_org_eclipse_swt_browser_Browser.test_ConstructorLorg_eclipse_swt_widgets_CompositeI()` - without those new annotations those tests were not running in JUnit5 but were in JUnit4.

akurtakov:
Thanks Jonah! I have been converting tests one by one for quite some time with this being the end goal - thanks for bringing it to closure.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_widgets_Shell.java.orig
```diff
@@ -0,0 +1,1030 @@
+/*******************************************************************************
```
akurtakov: This file shouldn't have been part of the PR, right?

akurtakov:
I am not thrilled by the new classes that used to be parametrized tests but I don't have better idea for now. Other than the *.orig file included by mistake it looks good to me.

jonahgraham:
> Thanks Jonah! I have been converting tests one by one for quite some time with this being the end goal - thanks for bringing it to closure.

Thanks for leading the way on that (and looking through the history on doing the 3 -> 4 migration too!). I was only wanting to do StyledText, but things ballooned because of the huge hierarchy of tests.

jonahgraham:
Is this ok to merge now?

akurtakov:
Yes it i. Merging.

PR #2559, opened 2025-09-29
Disable Win32 key event test in GitHub Actions runs #2516

HeikoKlare:
Since the upgrade of the default GitHub Actions runner for Windows to Windows Server 2025, the key event tests started to fail sporadically. Until a solution is found, this temporarily disables the tests to avoid that it hides actual issues when investigating the build results.

For the used system variable to identify the GHA environment, see https://docs.github.com/en/actions/reference/workflows-and-actions/variables#default-environment-variables

Contributes to https://github.com/eclipse-platform/eclipse.platform.swt/issues/2516

PR #2595, opened 2025-10-04
[Win32] Add PrtScr hotkey check for KeyEvent tests #2516

HeikoKlare:
The Win32 KeyEvent tests started to fail on GitHub Actions when the environment was updated to Windows Server 2025. Using the print screen button opens the Windows snipping tools which, for yet unknown reason, now receives the focus and does not close automatically when processing further key events, like it did before and does on Windows 11 systems.

To keep the tests working properly, this change excludes the PtrScr key as a system hotkey when not combined with any modifier key. It reverts the complete disablement of the tests performed in https://github.com/eclipse-platform/eclipse.platform.swt/pull/2559.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2516

@jonahgraham this is an extended version of your experiment https://github.com/jonahgraham/eclipse.platform.swt/pull/2
It limits the exclusion of the PtrScr key to the case where no state modifier is used, as this seems to be sufficient and ensures that we preserve testing of the other key combinations. I've successfully run this configuration three time in https://github.com/HeikoKlare/eclipse.platform.swt/pull/2.


jonahgraham (approved):

jonahgraham:
Thanks @HeikoKlare for tracking down minimal modification needed to keep as many tests running as possible.

PR #2599, opened 2025-10-06
Replace nativeZoom of widgets with disabled autoscaling with 100

ShahzaibIbrahim:
Few Widgets do not adhere to AUTO_SCALE_DISABLED field when set to true in Widget#setData. The proposed changes are as follow: 

- In Widget when setting DATA_AUTOSCALE_DISABLED to true we also update the nativeZoom value to 100 since it is a public field and we cannot rely on everyone calling the getter for it where we previously had a logic to return 100 in case of autoScale disablement. 
- Tracker class in some cases was using same logic as getZoom() but missing one additional detail which is present in getZoom(), that is to check if autoScale is disabled then use 100%
- In Control class when we set the bounds then we use to check for autoScale disablement but not when setting the size or location, so that's fixed too. 

### Detailed Analysis

These are the widgets or widget utils that should be affected when auto scale is disabled. 
```
org.eclipse.swt.dnd.DragSource.drag(Event)
org.eclipse.swt.dnd.DropTarget.convertPixelToPoint(int, int)
org.eclipse.swt.dnd.TableDragSourceEffect.getDragSourceImage(DragSourceEvent)
org.eclipse.swt.dnd.TableDropTargetEffect.dragOver(DropTargetEvent)
org.eclipse.swt.dnd.TreeDragSourceEffect.getDragSourceImage(DragSourceEvent)
org.eclipse.swt.dnd.TreeDropTargetEffect.dragOver(DropTargetEvent)

org.eclipse.swt.widgets.Button.Button(Composite, int)
org.eclipse.swt.widgets.Button.getCheckboxTextOffset(long)
org.eclipse.swt.widgets.Button.wmDrawChild(long, long)

org.eclipse.swt.widgets.Composite.wmNCPaint(long, long, long)

org.eclipse.swt.widgets.Control.getData(String)
org.eclipse.swt.widgets.Control.setParent(Composite)
org.eclipse.swt.widgets.Control.WM_DPICHANGED(long, long)

org.eclipse.swt.widgets.ExpandBar.computeSizeInPixels(int, int, boolean)
org.eclipse.swt.widgets.ExpandBar.drawThemeBackground(long, long, RECT)
org.eclipse.swt.widgets.ExpandBar.drawWidget(GC, RECT)

org.eclipse.swt.widgets.Menu.fixMenus(Decorations)

org.eclipse.swt.widgets.MenuItem.fixMenus(Decorations)
org.eclipse.swt.widgets.MenuItem.getMenuItemIconBitmapHandle(Image)
org.eclipse.swt.widgets.MenuItem.getMenuItemIconSelectedBitmapHandle()

org.eclipse.swt.widgets.TabFolder.drawThemeBackground(long, long, RECT)

org.eclipse.swt.widgets.Table.setCheckboxImageList(int, int, boolean)
org.eclipse.swt.widgets.Table.wmNotifyHeader(NMHDR, long, long)

org.eclipse.swt.widgets.TaskBar.createShellLink(MenuItem)

org.eclipse.swt.widgets.Text.WM_DRAWITEM(long, long)

org.eclipse.swt.widgets.Tree.calculateAndApplyIndentSize()
org.eclipse.swt.widgets.Tree.setCheckboxImageList()
org.eclipse.swt.widgets.Tree.wmNotifyHeader(NMHDR, long, long)
```

**When should auto scale disabling function?**

Whenever widget#nativeZoom or widget#getNativeZoom is called from any of the widget, the native zoom must be 100%.

### Reproduction
The drop scenario sketched above can be easily reproduced with `Snippet257`:
- Adapt the snippet as follows:
  - Start it with monitor-specific scaling enabled, e.g., add the following for line to the `main` method: `System.setProperty("swt.autoScale.updateOnRuntime", "true");`
  - Disabled autoscaling for StyledText control, i.e., call `text.setData("AUTOSCALE_DISABLED", "true");` after instantiating `text`
- Start the snippet and move the shell to a monitor with zoom != 100% (it's also sufficient if it starts on primary monitor with zoom != 100)
- Drag a text from somewhere es and try to drop it into the text widget.

Here is how it looks on 200% monitor because of the wrong coordinate conversion:
Image

### Expected Behavior
`Widget#nativeZoom` shall be adapted to always contain the actual native zoom, i.e., 100 in case `autoscaleDisabled==true`. All accesses to `Widget#nativeZoom` shall be summarized in this issue (or in the according PR) and evaluated regarding whether it is correct to use 100 when autoscaling is disabled or not. In cases where it may not be correct, `getData(DATA_NATIVE_ZOOM)` could be used to access the original native zoom.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
@@ -78,11 +78,14 @@ public abstract class Control extends Widget implements Drawable {
 	Region region;
 	Font font;
 	int drawCount, foreground, background, backgroundAlpha = 255;
+	boolean autoScaleDisabled = false;
 
 	/** Cache for currently processed DPI change event to be able to cancel it if a new one is triggered */
 	Event currentDpiChangeEvent;
 
 	private static final String DATA_SHELL_ZOOM = "SHELL_ZOOM";
+
+	private static final String DATA_AUTOSCALE_DISABLED = "AUTOSCALE_DISABLED";
 /**
  * Prevents uninitialized instances from being created outside the package.
  */
@@ -122,6 +125,7 @@ Control () {
 public Control (Composite parent, int style) {
 	super (parent, style);
 	this.parent = parent;
+	this.autoScaleDisabled = parent.autoScaleDisabled;
 	createWidget ();
 }
 
@@ -1268,6 +1272,19 @@ public Object getData(String key) {
 		return shell == null ? null : shell.nativeZoom;
 	}
 	return super.getData(key);
+}
+
+@Override
+public void setData(String key, Object value) {
+	super.setData(key, value);
+	if (DATA_AUTOSCALE_DISABLED.equals(key)) {
+		autoScaleDisabled = Boolean.parseBoolean(value.toString());
+		if (autoScaleDisabled) {
+			this.nativeZoom = 100;
+		} else {
+			this.nativeZoom = (int) getData(DATA_SHELL_ZOOM);
+		}
+	}
 }
 
 /**
@@ -3534,7 +3551,7 @@ public void setLayoutData (Object layoutData) {
  */
 public void setLocation (int x, int y) {
 	checkWidget ();
-	int zoom = getZoom();
+	int zoom = autoScaleDisabled ? parent.getZoom() : getZoom();
 	x = DPIUtil.pointToPixel(x, zoom);
 	y = DPIUtil.pointToPixel(y, zoom);
 	setLocationInPixels(x, y);
@@ -3791,7 +3808,7 @@ public void setRegion (Region region) {
  */
 public void setSize (int width, int height) {
 	checkWidget ();
-	int zoom = getZoom();
+	int zoom = autoScaleDisabled ? parent.getZoom() : getZoom();
 	width = DPIUtil.pointToPixel(width, zoom);
 	height = DPIUtil.pointToPixel(height, zoom);
 	setSizeInPixels(width, height);
@@ -4783,6 +4800,23 @@ public boolean setParent (Composite parent) {
 	OS.SetWindowPos (topHandle, OS.HWND_BOTTOM, 0, 0, 0, 0, flags);
 	reskin (SWT.ALL);
 	return true;
+}
+
+@Override
+GC createNewGC(long hDC, GCData data) {
+	data.nativeZoom = getNativeZoom();
+	if (autoScaleDisabled && data.font != null) {
+		data.font = SWTFontProvider.getFont(display, data.font.getFontData()[0], 100);
+	}
+	return GC.win32_new(hDC, data);
+}
+
+@Override
+int getZoom() {
+	if (autoScaleDisabled) {
+		return 100;
+	}
+	return super.getZoom();
 }
 
 abstract TCHAR windowClass ();
@@ -5959,6 +5993,9 @@ void sendZoomChangedEvent(Event event, Shell shell) {
 @Override
 void handleDPIChange(Event event, float scalingFactor) {
 	super.handleDPIChange(event, scalingFactor);
+	if (this.autoScaleDisabled) {
+		this.nativeZoom = 100;
+	}
 	resizeFont(this, getNativeZoom());
 
 	Image image = backgroundImage;
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Widget.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Widget.java
@@ -65,7 +65,6 @@ public abstract class Widget {
 	 * @noreference This field is not intended to be referenced by clients.
 	 */
 	public int nativeZoom;
-	boolean autoScaleDisabled = false;
 	int style, state;
 	Display display;
 	EventTable eventTable;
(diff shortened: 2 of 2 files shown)


bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/SWT.java
```diff
@@ -2163,6 +2163,17 @@ public class SWT {
...
+	/**
+	 * Style constant for disabling auto scale for widgets (value is 1&lt;&lt;32).
+	 * <br>Note that this is a <em>HINT</em>.
+	 * <p><b>Used By:</b></p>
+	 * <ul>
+	 * <li><code>Widget</code> and subclasses</li>
+	 * </ul>
+	 * @since 3.1
+	 */
+	public static final int AUTOSCALE_DISABLED = 1 << 32;
```
ShahzaibIbrahim: I introduced this style field in SWT so we that native zoom can be set to 100 on widget creation. But since this is specific for windows, I don't think it makes sense to put it for all OS. 

Need a suggestion for where can I put such field? @HeikoKlare

HeikoKlare: Please keep the two tasks separated: this change should only be about replacing the native zoom value and adapts its consumptions in case autoscaling is disabled. _How_ autoscaling is disabled is a different task, which we still need to address and which should not be mixed up with this.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Widget.java
```diff
@@ -2725,7 +2723,7 @@ int getZoom() {
 }
 
 void handleDPIChange(Event event, float scalingFactor) {
-	this.nativeZoom = event.detail;
+	this.nativeZoom = this.autoScaleDisabled ? 100 : event.detail;;
```
fedejeanne:
```suggestion
	this.nativeZoom = this.autoScaleDisabled ? 100 : event.detail;
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Widget.java
```diff
@@ -1473,6 +1473,7 @@ public void setData (String key, Object value) {
 
 	if (DATA_AUTOSCALE_DISABLED.equals(key)) {
 		autoScaleDisabled = Boolean.parseBoolean(value.toString());
+		this.nativeZoom = 100;
```
HeikoKlare: This does not take into account that `autoScaleDisabled` might be set to `false`. In that case, I would expect `nativeZoom` to be reset to shell zoom, i.e., `nativeZoom = getData(DATA_SHELL_ZOOM)`. That's only available in `Control` but this functionality should be moved to `Control` anyway as it does not make sense for `Widget`, shouldn't it?

ShahzaibIbrahim: Shouldn't we just set the native zoom to 100 in case autoScaleDisabled is set to true and not change it for the else condition here? I think that would be sufficient no?

ShahzaibIbrahim:
> How does one test this?

I have added how to test section.

> Shouldn't line 186 (in the constructor) also be adapted?
> 
> ```java
> 	this.nativeZoom = parent != null ? parent.nativeZoom : DPIUtil.getNativeDeviceZoom();
> ```

No, you can only disable autoscaling after the widget is created till now. So we don't need to adapt here.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Widget.java
HeikoKlare: But then you can never re-enable auto-scaling for that control.

ShahzaibIbrahim: Oh, I never thought that someone would re-enable it after disabling. I would have expected user to only set it to true to disable scaling. In that case I will extend the implementation to control and set it to shell zoom.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Widget.java
```diff
@@ -2725,7 +2716,7 @@ int getZoom() {
 }
 
 void handleDPIChange(Event event, float scalingFactor) {
-	this.nativeZoom = event.detail;
+	this.nativeZoom = this.autoScaleDisabled ? 100 : event.detail;
```
HeikoKlare: Shouldn't this also be moved to `Control`? I.e., shouldn't the `autoScaleDisabled` field be part of `Control` and `Control` then override the `nativeZoom` value with 100 if autoscaling is disabled or the like?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
```diff
@@ -4785,6 +4802,23 @@ public boolean setParent (Composite parent) {
...
+	return GC.win32_new(hDC, data);
+}
+
+@Override
+int getZoom() {
+	if (autoScaleDisabled) {
+		return 100;
+	}
+	return super.getZoom();
+}
```
HeikoKlare, line 4820: Shouldn't we get rid if this? When autoscale is disabled, zoom should be 100 anyway.

ShahzaibIbrahim: What if someone sets the autoScale value to 200 or some other number? Then the method getZoomForAutoscaleProperty will return 200 in that case.

HeikoKlare: I see, makes sense. Thank you for the explanation!

HeikoKlare (approved):

PR #2601, opened 2025-10-07
Cache the last created temporary handles at a requested size

arunjose696:
When rendering images at a specific targetSize, the most recently created temporary handle is cached. If a handle of a different size is needed, the previous handle is disposed of and replaced with the new one. The last cached handle is also disposed of when the image itself is destroyed.

This change does not affect the visual appearance and is purely an optimization. It is especially useful when a user scrolls through a diagram, as the same-size image may be redrawn multiple times. Without handle caching, new image handles would need to be created repeatedly whenever the image is redrawn. More details can be found at  https://github.com/vi-eclipse/Eclipse-Platform/issues/326#issuecomment-3242679364

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
@@ -141,6 +141,34 @@ public final class Image extends Resource implements Drawable {
 
 	private List<Consumer<Image>> onDisposeListeners;
 
+	private record CachedHandle(ImageHandle handleContainer, int requestedWidth, int requestedHeight) {
+
+		public void destroy() {
+			if (handleContainer != null) {
+				handleContainer.destroy();
+			}
+		}
+
+		public boolean isReusable(int height, int width) {
+			if(handleContainer == null) {
+				return false;
+			}
+			return (requestedHeight == height && requestedWidth == width)
+					|| (handleContainer.height == height && handleContainer.width == width);
+
+		}
+
+		public long getHandle() {
+			if (handleContainer != null) {
+				return handleContainer.handle;
+			}
+			return -1;
+		}
+	};
+
+	// Initialize lastRequestedHandle with -1 for size-related fields to indicate uninitialized values
+	CachedHandle lastRequestedHandle = new CachedHandle(null, -1, -1);
+
 private Image (Device device, int type, long handle, int nativeZoom) {
 	super(device);
 	this.type = type;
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1090,6 +1112,7 @@ private void destroyHandles(Predicate<Integer> filter) {
 			imageHandle.destroy();
 		}
 	}
+	lastRequestedHandle.destroy();
```
fedejeanne: This doesn't belong here, the method accepts a filter. Put it in `destroyHandles()` and, if necessary, in `destroyHandlesExcept(Set<Integer>)` too.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -141,6 +141,31 @@ public final class Image extends Resource implements Drawable {
...
+
+		public long getHandle() {
+			if (handleContainer != null) {
+				return handleContainer.handle;
+			}
+			return -1;
+		}
+	};
+
+	CachedHandle lastRequestedHandle = new CachedHandle(null, SWT.DEFAULT, SWT.DEFAULT, SWT.DEFAULT, SWT.DEFAULT);
```
fedejeanne: Move it above the type-declaration (together with other fields), make it `private` and pass `-1` instead of `SWT.DEFAULT` as parameters. There is no "default" in this context, only "uninitialized".

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -828,17 +853,14 @@ interface HandleAtSizeConsumer {
...
-	} finally {
-		handleContainer.destroy();
+	if (!lastRequestedHandle.isReusable(heightHint, widthHint)) {
+		lastRequestedHandle.destroy();
+		ImageHandle handleContainer = init(imageData, -1);
+		lastRequestedHandle = new CachedHandle(handleContainer, widthHint, heightHint, imageData.width,
+				imageData.height);
 	}
+	handleAtSizeConsumer.accept(lastRequestedHandle.getHandle(),
+			new Point(lastRequestedHandle.handleWidth, lastRequestedHandle.handleHeight));
```
fedejeanne:
```suggestion
			new Point(lastRequestedHandle.handleWidth(), lastRequestedHandle.handleHeight()));
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
arunjose696: Moved to destroyHandles()

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java

fedejeanne: I moved it above the type-declaration in 4a83dbe9d80d106ae0029277ee68c2446f2f58cc

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -828,17 +854,14 @@ interface HandleAtSizeConsumer {
 void executeOnImageHandleAtSize(HandleAtSizeConsumer handleAtSizeConsumer, int widthHint, int heightHint) {
 	ImageData imageData;
 	imageData = this.imageProvider.loadImageDataAtSize(widthHint, heightHint);
```
HeikoKlare: This should only be done inside the `if`, as otherwise the image data is unnecessarily loaded even if reuse happens.

arunjose696: done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -141,6 +141,32 @@ public final class Image extends Resource implements Drawable {
 
 	private List<Consumer<Image>> onDisposeListeners;
 
+	private record CachedHandle(ImageHandle handleContainer, int requestedWidth, int requestedHeight, int handleWidth,
```
HeikoKlare: Isn't handleWidth and handleHeight obsolete as it's already contained in `ImageHandle handleContainer`?

arunjose696: Fair point, removed

PR #2605, opened 2025-10-07
[GTK] Make isWayland real check

akurtakov:
And not just !isX11(). There are other backends like broadway, windows, quartz.


jonahgraham (approved):
To the best of my understanding this looks ok.

---

The rest of this is notes, perhaps more to record my own understanding than as a useful feedback on this PR.

 There are a lot of assumptions that that !x11 is wayland (e.g. [1](https://github.com/eclipse-platform/eclipse.platform.swt/blob/65b962b4f10ed4f1783ab63b21b72ffd696fe4d7/bundles/org.eclipse.swt/Eclipse%20SWT/gtk/org/eclipse/swt/widgets/Control.java#L5042) [2](https://github.com/eclipse-platform/eclipse.platform.swt/blob/d1859ce5e5bb37d53c1f4a866a4daa725153fd2d/bundles/org.eclipse.swt/Eclipse%20SWT/gtk/org/eclipse/swt/widgets/List.java#L430) [3](https://github.com/eclipse-platform/eclipse.platform.swt/blob/51c3eb07472cd6d6033b6dbbfeee57f35e70ee4e/bundles/org.eclipse.swt/Eclipse%20SWT/gtk/org/eclipse/swt/widgets/Menu.java#L343)) - but since for now !x11 is effectively wayland for all active use cases I think having the explicit check is good here.

I can't get SWT to launch on Broadway with GTK3 or GTK4, but I haven't debugged why gtk_init_check fails (I can get the gtk4-demo/gtk3-demo to run in broadway)

To match what is documented in https://docs.gtk.org/gdk4/wayland.html I guess we could be harsher and error (or warn) if !isWayland && !isx11. But the reality is that there are not any active development in non-wayland/x11 area. Initiative 31 did some investigation on using GTK on Windows and macOS, but my guess is that would still require an awful lot of additional special casing before a really stable implementation would work. If that happened, change isX11/isWayland to an enum of known/supported backends may make sense.

With the potential complexity of GTK3 or GTK4 multiplied by the number of backends (x11, wayland and others such as those mentioned here) my understanding as documented in my recent [blog post](https://kichwacoders.com/2025/09/30/testing-and-developing-swt-on-gtk/) is still woefully under developed.

akurtakov:
One reason to be explicit is to be a bit more future proof . Gtk 5 plans to drop X11 entirely (https://discourse.gnome.org/t/deprecated-x11-backend/30510 )  so having a "real" Wayland check is good idea. Merging.

PR #2610, opened 2025-10-08
[DO NOT SUBMIT] Experiments with Mac compile target

HannesWell:
Done in the context of
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3315

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -26,6 +26,12 @@ def runOnNativeBuildAgent(String platform, Closure body) {
 		case 'gtk4.linux.x86_64':
 			dockerImage = 'eclipse/platformreleng-debian-swtnativebuild:12'
 			break
+		case 'cocoa.macosx.aarch64':
+			agentLabel = 'nc1ht-macos11-arm64'
+			break
+		case 'cocoa.macosx.x86_64':
+			agentLabel = 'b9h15-macos11-x86_64'
+			break
 	}
 	if (dockerImage != null) {
 		podTemplate(inheritFrom: 'basic' /* inherit general configuration */, containers: [
@@ -233,7 +239,16 @@ pipeline {
 													sh build.sh clean
 													sh build.sh -gtk4 checklibs install-pi-only
 												elif [[ ${PLATFORM} == cocoa.macosx.* ]]; then
+													xcode-select --print-path
+													xcode-select --version
+													softwareupdate --list
+													softwareupdate --history
+													xcodebuild -version
+													xcodebuild -showsdks
 													sh build.sh install
+													for f in libs/*; do
+														vtool -show ${f}
+													done
 												else
 													echo "Unexpected build platform ${PLATFORM}"
 													exit 1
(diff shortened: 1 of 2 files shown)


HannesWell:
@Phillipus Can you help here to assess if the `MACOSX_DEPLOYMENT_TARGET` environment property serves the intended purpose from
- https://ci.eclipse.org/releng/job/eclipse.platform.swt/job/PR-2610/2/pipeline-overview/?selected-node=329
- https://ci.eclipse.org/releng/job/eclipse.platform.swt/job/PR-2610/2/pipeline-overview/?selected-node=335

and
- https://ci.eclipse.org/releng/job/eclipse.platform.swt/job/PR-2610/3/pipeline-overview/?selected-node=324
- https://ci.eclipse.org/releng/job/eclipse.platform.swt/job/PR-2610/3/pipeline-overview/?selected-node=343

I have to admit that I'm not sure.
Is `minos 10.15` respectively `version 10.10` the target version we are looking for? Or is it the `sdk` value?
The latter seem not to change with the env variable mention above, but `minos`/`version` seem to consider it?

HeikoKlare:
> Is `minos 10.15` respectively `version 10.10` the target version we are looking for? Or is it the `sdk` value?
> The latter seem not to change with the env variable mention above, but `minos`/`version` seem to consider it?

I did some research and tested on my own locally on Mac and I am quite sure that `minos` is the minimum required MacOS version to deploy the binaries to and execute them on. As you have already pointed to in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3315, the environment variable that you have touched defines that minimum version, see also https://cmake.org/cmake/help/latest/envvar/MACOSX_DEPLOYMENT_TARGET.html

I've tested locally on a system that was just updated to MacOS 26.0. In consequence, the `sdk` value always shows `26.0`, as the SDK seems to have the version of the system you use for compilation. I changed the `MACOSX_DEPLOYMENT_TARGET` value in the build script to different values, and the resulting binaries always have that same value as `minos` after compilation. Only exception is that compiling against any 10.x always results in `minos==11.0` (probably because I am on aarch64, for which no 10.x versions exist at all).

Also see this for some pointers confirming the assumption that you made: https://stackoverflow.com/questions/17143373/determine-minimum-osx-version-a-binary-was-compiled-for

HannesWell:
> I did some research and tested on my own locally on Mac and I am quite sure that `minos` is the minimum required MacOS version to deploy the binaries to and execute them on.

OK, great. Thank you for your help on this one!

But besides the minos version, would it also be necessary to keep the SDK as low as possible? Would it work to run on mac 14 if the binaries are linked against sdk 15 (assuming sdk 15 is the default on mac 15)?
The issue linked from @Phillipus indicates something like this:
- https://github.com/eclipse-platform/eclipse.platform.swt/issues/1012

But as far as I can tell, it's possible to install older versions of xcode/the SDK. So if that's necessary we should be able to do it even on newer OS versions. I'll ask in the EF gitlab issue about that.

HannesWell:
Experiments were successful. Closing.

PR #2612, opened 2025-10-09
Suppress warnings for deprecated Cursor constructor

ShahzaibIbrahim:
Cursor(Device, ImageData, ImageData, int, int) was deprecated recently which is still used in some of the snippets testing that exact constructor. That's why the warnings are being suppressed.


akurtakov:
Would you please migrate the snippets to the new way ? This is the place where people using SWT should come to see how things are used so showing them these deprecated APIs is counterproductive.

ShahzaibIbrahim:
@akurtakov we recently implemented Snippet386 covering all kind of Cursors. Also Snippet92 was to show how this specific cursor with source and mask should work. We either remove it completely but I don't see the reason to migrate it to new way since we already have example in Snippet386 for others.

fedejeanne:
I'm with Alex on this one: why show the deprecated constructor in the snippets in the first place? It's not like we want new users to learn how to use a deprecated constructor.

Instead, I would:
- Remove the 1st option (usage of deprecated constructor) from `Snippet386`
- Delete Snippet92

One other detail: I just noticed that there is no explanation as to _why_ has the constructor been deprecated (no linked GH issue either). That would be a nice addition to the JavaDocs, don't you think?

fedejeanne:
> One other detail: I just noticed that there is no explanation as to _why_ has the constructor been deprecated (no linked GH issue either). That would be a nice addition to the JavaDocs, don't you think?

To answer my own question: there's the commit message

image

... so we're good, no more "nice additions" required :-)

HeikoKlare:
`Snippet92` may be migrated by integrating the mask data into the the source image data. The it would still reflect the source + mask use case.

For `Snippet386` it does not really make sense to migrate something as that snippet shows the behavior of all constructors in comparison. And in my opinion it's still valuable to the deprecated constructor in there in case someone wants to compare those behaviors, e.g., because they still use separate image and mask files and want to compare differences or the like.

ShahzaibIbrahim:
I would then like to remove the Snippet92 completely. Since we have enough (~3) snippets covering how the constructor with source data only works.

HeikoKlare:
> I would then like to remove the Snippet92 completely. Since we have enough (~3) snippets covering how the constructor with source data only works.

Okay, can you please point to the one that shows how to create cursor with source + mask based on the single-ImageData constructor?

ShahzaibIbrahim:
> Okay, can you please point to the one that shows how to create cursor with source + mask based on the single-ImageData constructor?

In Snippet119, reflecting source + mask based on the single-ImageData constructor. 

See, https://github.com/eclipse-platform/eclipse.platform.swt/blob/6fa4ee7ee2372daa68b30dbedc2e85ead2e70d00/examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet119.java#L80-L91

HeikoKlare:
Thank you! Then I would be fine with removing the snippet.

ShahzaibIbrahim:
@HeikoKlare I have deleted the snippet. This PR is ready for review.

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet386.java
```diff
@@ -205,8 +205,9 @@ private Cursor createCursor(int selection) {
 						mask.setPixel(x, y, x % 2);
 					}
 				}
-
-				return new Cursor(display, source, mask, IMAGE_SIZE_IN_POINTS / 2, IMAGE_SIZE_IN_POINTS / 2);
+				@SuppressWarnings("deprecation")
+				Cursor cursor = new Cursor(display, source, mask, IMAGE_SIZE_IN_POINTS / 2, IMAGE_SIZE_IN_POINTS / 2);
+				return cursor;
```
HeikoKlare, line 210: Is it necessary to split this line up into two?

akurtakov:
Please remove the snippet from https://github.com/eclipse-platform/eclipse.platform.swt/blob/79e74c73c6fed64c8c2811e2d1977d40a9777a63/examples/org.eclipse.swt.snippets/Snippets.md?plain=1#L124 too.

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet386.java
ShahzaibIbrahim: Yes, I tried applying deprecation but apparently i can apply annotation to the local variable declaration not to the return statements.

ShahzaibIbrahim: Or I could apply it to the whole method, but I only wanted to suppress the warning for this method.

Issue #2494, opened 2025-09-08
Eclipse 2025-06: UI rendering issues on Windows with monitors using different scale factors


WinmaH:
When running Eclipse IDE for Enterprise Java and Web Developers (2025-06 release) on Windows 11, UI rendering artifacts appear on secondary monitors configured with a different scale factor (e.g., 125%). Some images/icons display with black boxes or incorrect rendering.

[Run -> Run Configuration]

<img width="262" height="556" alt="Image" src="https://github.com/user-attachments/assets/e29e3ed5-8912-4004-bea7-00c473fa93ec" />

Environment:

Eclipse IDE 2025-06 (tested also with Eclipse 2025-09 RC2)
Windows 11
Multi-monitor setup with different scale factors (e.g., primary monitor 100%, secondary monitor 125%)

Steps to Reproduce:

Launch Eclipse 2025-06 on Windows 11.
Use a dual-monitor setup where one monitor is set to 100% scaling and another is set to 125%.
Open the IDE on the secondary monitor (with 125% scaling).
Observe UI elements and icons.

Expected Behavior:
Icons and images should render normally, without artifacts, regardless of monitor scale factor.

Actual Behavior:
Some images/icons display with black boxes or rendering glitches on the monitor with a non-100% scale factor.

Additional Notes:

Issue is reproduced only in Windows OS
Issue occurs even when “Monitor-specific UI Scaling” is enabled in Eclipse Preferences.
Tried Windows compatibility settings → Override high DPI scaling behavior → System (Enhanced), but the issue persists.
Issue is not reproduced when both monitors use 100% scaling.
Verified with latest Eclipse 2025-09 RC2; the issue still exists.

HeikoKlare:
I have tried to reproduce the behavior with the 2025-06 release of the package for Enterprise Java and Web Developers. I use primary monitor at 100% and secondary monitor at 125%. I have started the application from both primary and secondary monitor and opened the Run Configurations dialog on the 125% monitor or moved it there. In all cases the icon backgrounds look correct for me:
<img width="984" height="792" alt="Image" src="https://github.com/user-attachments/assets/ca18761b-d5c6-44ff-96e3-5355c03ec175" />

Can you give information about the JRE you are using to run Eclipse? In my case, it's Temurin 21.0.6.7. Maybe you can also try to start Eclipse with a different JRE build or version.

[...]

PR #2615, opened 2025-10-10
[Win32] Fix transparency when scaling images with transparentPixel

ShahzaibIbrahim:
### Description

**Move Transparent Pixel Handling from Image to ImageHandle**

Previously, the transparency pixel was managed at the Image level, which caused issues when an Image had multiple handles at different zoom levels or color models (e.g., indexed vs. direct/ARGB). For example, a GIF image at 100% zoom may use a transparency pixel, but when scaled, SWT converts it to a direct image with alpha data, making the transparency pixel concept invalid for the scaled handle.

This change moves the transparency pixel field from the Image class to the ImageHandle class. Now, each handle manages its own transparency information, allowing a single Image to have multiple handles, each with the correct transparency type (pixel or alpha) according to its color model. This ensures correct transparency handling for all handles, regardless of how they were created or scaled.

### How to test
1. Run the following snippet on 100% zoom monitor
```
import org.eclipse.swt.*;


public class ButtonWithGifImage {
        shell.setText("SWT Button with GIF Image");
        shell.setSize(300, 200);

        button.setBounds(80, 60, 140, 60);
        button.setText("Click Me");

        Image image = new Image(display,
            ButtonWithGifImage.class.getResourceAsStream("sample.gif"));
        button.setImage(image);

        shell.open();
        while (!shell.isDisposed()) {
                display.sleep();
        }

        image.dispose();
        display.dispose();
    }
}

```
2. Move the window to a 150% DPI monitor (causing scaling) — verify the image still renders correctly (no jagged/incorrect background).
3. Move back to 100% — verify the image still renders correctly and no gray background appears.

GIF Image: 
sample

### Result 
 
Here's the preview how it looked before and after this PR change. 100 -> 150 -> 100

**Before:**
image

**After**
image

- [x] Requires: https://github.com/eclipse-platform/eclipse.platform.swt/pull/2651/files 

Fixes: https://github.com/eclipse-platform/eclipse.platform.swt/issues/2494

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
@@ -105,11 +105,6 @@ public final class Image extends Resource implements Drawable {
 	private boolean isDestroyed;
 
 	/**
-	 * specifies the transparent pixel
-	 */
-	int transparentPixel = -1, transparentColor = -1;
-
-	/**
 	 * the GC which is drawing on the image
 	 */
 	GC memGC;
@@ -277,7 +272,7 @@ public Image(Device device, Image srcImage, int flag) {
 						long hOldSrc = OS.SelectObject(hdcSource, srcImageHandle);
 						BITMAP bm = new BITMAP();
 						OS.GetObject(srcImageHandle, BITMAP.sizeof, bm);
-						imageMetadata = new ImageHandle(OS.CreateCompatibleBitmap(hdcSource, rect.width, bm.bmBits != 0 ? -rect.height : rect.height), imageHandle.zoom);
+						imageMetadata = new ImageHandle(OS.CreateCompatibleBitmap(hdcSource, rect.width, bm.bmBits != 0 ? -rect.height : rect.height), imageHandle.zoom, imageHandle.transparentPixel);
 						if (imageMetadata.handle == 0) SWT.error(SWT.ERROR_NO_HANDLES);
 						long hOldDest = OS.SelectObject(hdcDest, imageMetadata.handle);
 						OS.BitBlt(hdcDest, 0, 0, rect.width, rect.height, hdcSource, 0, 0, OS.SRCCOPY);
@@ -289,12 +284,11 @@ public Image(Device device, Image srcImage, int flag) {
 						/* Release the HDC for the device */
 						device.internal_dispose_GC(hDC, null);
 					}
-					transparentPixel = srcImage.transparentPixel;
 					break;
 				case SWT.ICON:
 					for (ImageHandle imageHandle : srcImage.zoomLevelToImageHandle.values()) {
 						Rectangle rect = imageHandle.getBounds();
-						imageMetadata = new ImageHandle(OS.CopyImage(imageHandle.handle, OS.IMAGE_ICON, rect.width, rect.height, 0), imageHandle.zoom);
+						imageMetadata = new ImageHandle(OS.CopyImage(imageHandle.handle, OS.IMAGE_ICON, rect.width, rect.height, 0), imageHandle.zoom, imageHandle.transparentPixel);
 						if (imageMetadata.handle == 0) SWT.error(SWT.ERROR_NO_HANDLES);
 					}
 					break;
@@ -888,7 +882,8 @@ long [] createGdipImage(Integer zoom) {
 }
 
 long[] createGdipImageFromHandle(ImageHandle imageHandle) {
-	long handle = imageHandle.handle;
+	long handle = imageHandle.getHandle();
+	int transparentPixel = imageHandle.transparentPixel;
 	switch (type) {
 		case SWT.BITMAP: {
 			BITMAP bm = new BITMAP();
@@ -1124,7 +1119,7 @@ public boolean equals (Object object) {
 	if (object == this) return true;
 	if (!(object instanceof Image)) return false;
 	Image image = (Image) object;
-	if (device != image.device || transparentPixel != image.transparentPixel) return false;
+	if (device != image.device) return false;
 	return (styleFlag == image.styleFlag) && imageProvider.equals(image.imageProvider);
 }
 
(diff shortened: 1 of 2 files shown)


merks:
Yeah!  🥳

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1908,10 +1903,14 @@ public void setBackground(Color color) {
 	if (isDisposed()) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
 	if (color == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
 	if (color.isDisposed()) SWT.error(SWT.ERROR_INVALID_ARGUMENT);
-	if (transparentPixel == -1) return;
-	transparentColor = -1;
+	if (this.getImageData().transparentPixel == -1) return;
 	backgroundColor = color.getRGB();
```
HeikoKlare, line 1898: Is it correct that now `backgroundColor` is always set even if `transparentPixel` is not -1?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1165,7 +1159,9 @@ public Color getBackground() {
 	/* Get the HDC for the device */
 	long hDC = device.internal_new_GC(null);
 	return applyUsingAnyHandle(imageHandle -> {
```
HeikoKlare: I don't think we can use `anyHandle` here anymore, but need to use exactly the one based on the transparent pixel (probably usually the 100% one).

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1156,7 +1150,7 @@ public boolean equals (Object object) {
  */
 public Color getBackground() {
 	if (isDisposed()) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
-	if (transparentPixel == -1) return null;
+	if (this.getImageData().transparentPixel == -1) return null;
```
HeikoKlare, line 1146: Could be quite expensive to always extract the image data here.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -288,11 +280,11 @@ public Image(Device device, Image srcImage, int flag) {
 						OS.SelectObject(hdcDest, hOldDest);
 						OS.DeleteDC(hdcSource);
 						OS.DeleteDC(hdcDest);
+						imageMetadata.transparentPixel = imageHandle.transparentPixel;
```
HeikoKlare: Why not pass this to the constructor call to ImageHandle above?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2816,15 +2809,24 @@ public void internal_dispose_GC(long handle, GCData data) {
...
+	long handle;
 	private final int zoom;
 	private int height;
 	private int width;
+	/**
+	 * specifies the transparent pixel
+	 */
+	int transparentPixel = -1, transparentColor = -1;
 
 	public ImageHandle(long handle, int zoom) {
```
HeikoKlare: Can't we remove this constructor to put the responsibility for thinking about whether a transparentPixel needs to be specified or not to the consumer? Otherwise we risk that we or someone unexperienced just calls this constructor without thinking about the transparent pixel, but would actually need to take into account that it must be set.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1908,10 +1904,7 @@ public void setBackground(Color color) {
 	if (isDisposed()) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
 	if (color == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
 	if (color.isDisposed()) SWT.error(SWT.ERROR_INVALID_ARGUMENT);
-	if (transparentPixel == -1) return;
-	transparentColor = -1;
 	backgroundColor = color.getRGB();
-	zoomLevelToImageHandle.values().forEach(imageHandle -> imageHandle.setBackground(backgroundColor));
```
HeikoKlare, line 1906: Why do we completely remove this? In case there are already existing handles, shouldn't we ...
- apply the color if possible?
- print a warning that it cannot be applied if the existing handles are scaled versions (maybe encapsulated in a strict check to avoid performance impacts)?

HeikoKlare (commented):
I think `that we agree that the `transparentPixel` belongs to the handle and not to the image, so this is generally the right thing to do. @ShahzaibIbrahim can you please extract the preparatory refactoring to change the return type of `getHandle()` and use that one instead of `win32_getHandle()` into a separate preparatory PR, such that this one becomes easier to review?
I left some additional comments regarding the functional changes in this PR.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
ShahzaibIbrahim: I get the first part, we should apply color if possible. But why print a warning at all? If they are scaled version the setBackground method will do the early exit anyway after checking for transparentPixel != -1. Isn't that so?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
ShahzaibIbrahim: Sounds reasonable.

ShahzaibIbrahim:
> `@ShahzaibIbrahim can you please extract the preparatory refactoring to change the return type of`getHandle()`and use that one instead of`win32_getHandle()

https://github.com/eclipse-platform/eclipse.platform.swt/pull/2651

ShahzaibIbrahim:
> Just to not forget about it, here a comment from the extracted refactoring PR ([#2651 (review)](https://github.com/eclipse-platform/eclipse.platform.swt/pull/2651#pullrequestreview-3356933488)) that we should address in this PR:
> 
> > As a follow up in #2615, we can get rid of the Image parameters in some GC methods, as they were only passed for the transparentPixel/transparentColor fields, which are then properly part of the ImageHandle.

There was only one method that was passing Image solely for transparentpixel/color. I have refactored it.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1154,57 +1149,59 @@ public boolean equals (Object object) {
...
-					green = (transparentPixel & 0xFF0000) >> 16;
-					red = (transparentPixel & 0xFF00) >> 8;
-					break;
-				default:
-					return null;
-			}
+	ImageHandle imageHandle = this.getHandle(100, 100);
+	if (imageHandle.transparentPixel == -1) {
+		return null;
+	}
```
HeikoKlare, line 1154: With this change, hDC is not disposed anymore.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2819,10 +2813,15 @@ class ImageHandle {
 	private final int zoom;
 	private int height;
 	private int width;
+	/**
+	 * specifies the transparent pixel
+	 */
+	int transparentPixel = -1, transparentColor = -1;
```
HeikoKlare: Can `transparentPixel` be final? I don't think it is changed anywhere.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
HeikoKlare: > But why print a warning at all? If they are scaled version the setBackground method will do the early exit anyway after checking for transparentPixel != -1. Isn't that so?

Yes, and that's potentially unexpected behavior. My comment was bit imprecise. The scenario I refer to is the following: you have an image whose handle at 100% is created with a transparent pixel. So if you create that image, set the background and then request handles (also for scaled version), they will all have the background applied. If you create such an image, then request multiple handles (including scaled version) and then set the background, that will only affect the 100% handle but not the others (because they have no transparent pixel that). That's unexpected.

But maybe we even have another here, because the image at 100% would need to be drawn to apply all the login in `GC#drawBitmap()` before creating a handle for a scaled version?

HeikoKlare (dismissed):
> There was only one method that was passing Image solely for transparentpixel/color. I have refactored it.

The `drawBitmap` method also takes an Image in addition to an ImageHandle, now only for the sake of accessing its `memGC`. I don't even think the logic accessing the `memGC` is correctly placed in that method but should probably be moved to the caller. However, that's a follow-up task.

I have one comment on an introduced handle leak, which is why I place a change request.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
akoch-yatta: Fair point, but as we currently even get it to properly work on a correct 100% image we need to revise the Image#setBackground logic overall anyway, if we wanna do something about it. So I would consider this a different topic, so I discussed with @ShahzaibIbrahim to raise a separate issue about this and write all his findings down to be able to decide if and how to proceed with it.

ShahzaibIbrahim: I have added a follow-up ticket for the following issues, see: https://github.com/vi-eclipse/Eclipse-Platform/issues/495

ShahzaibIbrahim:
> I did a small adaption by adding a new constructor to ImageHandle with transparentPixel to better differentiate the use cases.

I did the same before I got this comment from heiko, https://github.com/eclipse-platform/eclipse.platform.swt/pull/2615#discussion_r2439100322

akoch-yatta:
> > I did a small adaption by adding a new constructor to ImageHandle with transparentPixel to better differentiate the use cases.
> 
> I did the same before I got this comment from heiko, [#2615 (comment)](https://github.com/eclipse-platform/eclipse.platform.swt/pull/2615#discussion_r2439100322)

Ah okay, I didn't notice that. We need to adapt this in a separate PR now

ShahzaibIbrahim:
> Ah okay, I didn't notice that. We need to adapt this in a separate PR now

Separate PR -> https://github.com/eclipse-platform/eclipse.platform.swt/pull/2663

PR #2618, opened 2025-10-13
Removing Image#CreateSurface from GTK

arunjose696:
Removing createSurface method from Image class in GTK and its callers as this method currently does not do anything. This PR is just a cleanup removing method calls which does not make any sense. 

@akurtakov could you take a look if this cleanup is acceptable or is there some reason to keep the method?


akurtakov:
Looks like a really good catch that has survived in the repo from before the unfortunate https://github.com/eclipse-platform/eclipse.platform.swt/commit/093c579a4ffd9551acb901bba9617e7aa776989d which broke any research prior to it.

PR #2619, opened 2025-10-13
[MacOS] Support loading and drawing images at desired size

HeikoKlare:
With the support of SVGs and the addition of the ImageDataAtSizeProvider, it is possible for many Image instances to provide a handle that represents the image at sharply rendered specific size in addition to the existing support of providing an image at different zooms of a base size.

This change adds according support to the MacOS implementations of GC and Image. The GC#drawImage() method accepting only destination position and size now makes use of enhanced Image capabilities to load the image at the desired size if possible instead of just using the best fitting zoomed handle.

Can be tested with the snippets provided for the Windows implementation PR:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2526

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
@@ -14,6 +14,7 @@
 package org.eclipse.swt.graphics;
 
 
+import static org.eclipse.swt.internal.DPIUtil.pointToPixel;
 import static org.eclipse.swt.internal.image.ImageColorTransformer.DEFAULT_DISABLED_IMAGE_TRANSFORMER;
 
 import java.io.*;
@@ -1177,6 +1178,7 @@ private NSBitmapImageRep createRepresentation(ImageData imageData, AlphaInfo alp
 
 @Override
 void destroy() {
+	cachedImageAtSize.destroy();
 	if (memGC != null) memGC.dispose();
 	handle.release();
 	handle = null;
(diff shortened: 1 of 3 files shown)


bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
```diff
@@ -1837,5 +1839,78 @@ public static void drawAtSize(GC gc, ImageData imageData, int width, int height)
 	});
 }
 
+@FunctionalInterface
```
arunjose696: the name does not make sense as it does not consume a handle , but an image

HeikoKlare: Indeed. I just removed the interface as it's only an image consumer.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
```diff
@@ -1837,5 +1839,73 @@ public static void drawAtSize(GC gc, ImageData imageData, int width, int height)
...
+		return image != null && image.height == height && image.width == width;
+	}
+
+	private Optional<Image> loadImageAtSize(int destWidth, int destHeight) {
+		Optional<ImageData> imageData = loadImageDataAtExactSize(destWidth, destHeight);
+		if (imageData.isEmpty()) {
+			return Optional.empty();
+		}
+		Image image = new Image(device, imageData.get());
+		NSBitmapImageRep representation = image.getRepresentation(100);
```
arunjose696: Shouldnt the image.getRepresentation be inside the if condition? also not very clear to me why do we get and remove representation at 100.

HeikoKlare: Yes, it can be inside the if block. I fixed that with the latest commit.
We create the image based on image data in the line before (1881), which effectively creates an image with a 100% representation (out of the given image data). For that reason, we retrieve the representation at that same zoom. We then use the `createRepFromSourceAndApplyFlag` to apply the style flag on that representation (which adds another representation with that styling to the image), so that we need to remove the original (unstyled) representation afterwards.

arunjose696 (approved):
The code changes LGTM

PR #2622, opened 2025-10-14
 [GTK] Support loading and drawing images at desired size

arunjose696:
This change introduces support for loading and drawing images at custom sizes in the GTK versions of GC and Image. The GC#drawImage() method, which  only accepts the destination position and size, now attempts to load image at the destination size in case if it is provided by ImageFilenameProvider returning svgs or if the image is created with a imageDataAtSizeProvider.

Changes can be tested with snippets in #2526 for svgs and the tests with image dataAtSize provider which are enabled for linux along with this PR

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Image.java
@@ -18,6 +18,7 @@ import static org.eclipse.swt.internal.image.ImageColorTransformer.DEFAULT_DISAB
 
 import java.io.*;
 import java.util.*;
+import java.util.function.*;
 
 import org.eclipse.swt.*;
 import org.eclipse.swt.internal.*;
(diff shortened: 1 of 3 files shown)


bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Image.java
```diff
@@ -927,6 +928,76 @@ void destroy() {
...
+		Optional<ImageData> imageData = loadImageDataAtExactSize(destWidth, destHeight);
+		if (imageData.isEmpty()) {
+			return Optional.empty();
+		}
+		Image image = new Image(device, imageData.get(),100);
+		if (styleFlag != SWT.IMAGE_COPY) {
+			image.currentDeviceZoom=DPIUtil.getDeviceZoom();
+			Image styledImage = new Image(device, image, styleFlag);
+			styledImage.currentDeviceZoom = 100;
+			image.dispose();
```
HeikoKlare: Why do we have to (re)set the zoom of the images here? Naively, I would expect that this would also work:
```suggestion
		Image image = new Image(device, imageData.get(), DPIUtil.getDeviceZoom());
		if (styleFlag != SWT.IMAGE_COPY) {
			Image styledImage = new Image(device, image, styleFlag);
			image.dispose();
```
But somehow it does not (at least not in the tests; in a product it seems to work fine). Can you explain this?

arunjose696: In GTK in Image(device, image, styleFlag) constructor if the `DPIUtil.getDeviceZoom() != currentDeviceZoom ` cairo tries to scale the image with a scaling factor of the device zoom as we dont need the image to be scaled we  set the zoom to 
DPIUtil.getDeviceZoom() before this constructor is called. PFB the link to the part that does the scaling in code.

https://github.com/vi-eclipse/eclipse.platform.swt/blob/aac508522c73c6d53b3476ce327a0fb261b32a02/bundles/org.eclipse.swt/Eclipse%20SWT/gtk/org/eclipse/swt/graphics/Image.java#L288C2-L291C3

HeikoKlare: Yes, that's why I asked for instantiating the original (unstyled) image as:
```java
Image image = new Image(device, imageData.get(), DPIUtil.getDeviceZoom());
```
With that, I would expect the image to be initialized with the given image at device zoom, such that the styled image will not scale (as the currentDeviceZoom, taken from the original image, fits to DPIUtil.getDeviceZoom()). But still some scaling must happen then at some other place. Interestingly, with that change the unstyled images seem to pass the test and only the styled ones fail.

arunjose696: Sorry, missed that part in the comment. 

When we create an image using` new Image(device, imageData.get(), DPIUtil.getDeviceZoom())`, the `init(imageData, zoom)` method is invoked with zoom set to DPIUtil.getDeviceZoom(). This method then scales the image based on the calculated scale factor (zoom / 100f). So we would need to use an image at 100% zoom to not do this scaling.

https://github.com/vi-eclipse/eclipse.platform.swt/blob/aac508522c73c6d53b3476ce327a0fb261b32a02/bundles/org.eclipse.swt/Eclipse%20SWT/gtk/org/eclipse/swt/graphics/Image.java#L1385

HeikoKlare: Thank you for the explanation. I would like to identify what we actually want to do here and see whether we cannot do it in a more proper way than tweaking zooms during the image lifecycle (which will likely break with future changes).
In my understanding, what we would conceptually need to do is to create an image based on given image data that are considered to already be device zoom while the existing public constructor accepting image data always considers them at 100% and scales them to device zoom. Now what I still do not understand: this scaling is already done inside the public Image constructor accepting the image data (via `GtkDPIUtil.pointToPixel()`). This is what we already work around by directly passing image data with their zoom to the private constructor that also accepts a zoom.
What happens with the scale factor in the `init()` method is not clear to me. I played around with it a bit but did not fully understanding yet what the proper solution is. To me it somehow seems as if something is already broken now, because the private image constructor taking image data and zoom should actually operate based on the assumption that the given image data fits to the given zoom and also initializes the image with that exact zoom (so exactly what we want).

HeikoKlare: When testing around, I see that the image copy operation seems to produce unexpected results. This can also be reproduced with an existing test:
- `Test_org_eclipse_swt_graphics_Image#test_ConstructorLorg_eclipse_swt_graphics_DeviceImageI()`

That test fails on 200% zoom, which means that the image data of an image and a copy are different at non-100% zoom. This seems like a bug to me.

HeikoKlare: When starting an SDK product at 200%, you can see that the disabled toolbar icons (which are created via that copy constructor) are blurry. So the functionality is definitely broken.

HeikoKlare: I think I have a fix for the blurry scaling, which also results in my above proposal working (i.e, just creating an image with the proper zoom instead of tweaking the zooms). I will post a PR soon.

HeikoKlare: See:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2653

arunjose696: I have rebased on the PR and can see the tests pass with your proposal I have also tested my snippets which work fine as well.

laeubi:
GTK already has `org.eclipse.swt.graphics.Image.drawScaled(GC, ImageData, int, int, float)` can't that be used? this also seems to use the imageprovider to get a scaled variant?

HeikoKlare:
> GTK already has `org.eclipse.swt.graphics.Image.drawScaled(GC, ImageData, int, int, float)` can't that be used? this also seems to use the imageprovider to get a scaled variant?

I guess you refer to `drawAtSize()`? That method is used to scale an image and not to render it at a specific size. It can only use what's given at different zooms to do a best-possible scaling from. But it will usually lead to blurry results whereas the new functionality is capable of rendering sharply at any size, given an SVG, an `ImageDataAtSizeProvider` or the like. But that's nothing specific for GTK, it is the same for every OS and GTK is the last to adopt this functionality (Windows and MacOS are already implemented).

akurtakov:
I have tested this PR using Snippet 2 from https://github.com/eclipse-platform/eclipse.platform.swt/pull/2526 .
While blur is gone, the "half" drawings are actually same as full.  I haven't looked into why this happens. This is the case both with and without this PR.

HeikoKlare:
> While blur is gone, the "half" drawings are actually same as full. I haven't looked into why this happens. This is the case both with and without this PR.

Seems like the snippet is incorrect. It's the same on Windows: every image looks the same. I always used Snippet 1 from that PR for testing. With that snippet, I see the expected results also on GTK with this PR.

arunjose696:
Snippet2 is broken as it was initially intended to support the case for drawing the images cropped also in that PR but later we decided to leave drawing the cropped images for later. I have updated the snippet to only draw the full image now.

HeikoKlare (approved):

PR #2624, opened 2025-10-14
Add a SWTTest Junit extension to allow tests executed in the UI thread

laeubi:
Currently we require SWT-UI based tests to be executed by special runners known as the 'UIHarness' to be run in the UI thread, what has several problems:

1) is requires usually to fire up a whole workbench (and a workspace) 2) the whole testsuite has to run in the UI thread 3) no way to customize this further
4) actually this is nothing that belongs into the test framework

This now adds a new SWTTest JUnit extension that allows to mark a test class with an extension and get your methods executed in the UI thread.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/test/SWTTest.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/test/SWTTest.java
@@ -1,0 +1,64 @@
+/*******************************************************************************
+ * Copyright (c) 2025 Christoph Läubrich and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Christoph Läubrich - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.swt.test;
+
+import java.lang.reflect.*;
+
+import org.eclipse.pde.api.tools.annotations.*;
+import org.eclipse.swt.widgets.*;
+import org.junit.jupiter.api.extension.*;
+
+/**
+ * The {@link SWTTest} is an extension that can be using in JUnit Platform
+ * Runners as an extension to execute the test code in the default display
+ * thread.
+ * <p>
+ * <b>Usage: </b>You would never use this in any client code but when writing a
+ * test for the JUnit Platform like in this example:
+ * </p>
+ * <p>
+ * <pre>
+ * &#64;ExtendWith(SWTTest.class)
+ * public class MyTestThaRequireUI {
+ *
+ * }
+ * </pre>
+ * </p>
+ */
+@NoInstantiate
+public final class SWTTest implements InvocationInterceptor {
+
+	@Override
+	public void interceptTestMethod(Invocation<Void> invocation, ReflectiveInvocationContext<Method> invocationContext,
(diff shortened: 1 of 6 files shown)


laeubi:
This currently is just a proof-of-concept on how we can make testing more self-contained. One likely more wants to use an `@UI` / `@NoUI` annotations that can be added to different levels (e.g. methods or classes) and then even switch between the threads in different methods of the same testclass.

Other annotation might be used to drive the event-queue automatically after each test.

so a "typical" SWT test then can look like this:

```
@UI
@DriveEvents
class MySwtTest {
   

  void testMyWidget() {
    //Will run in the UI and drive the event-queue before/after
  }
  
  @NoUI
  void testRunningOutSide() {
    //Test something outside the UI thread
  }
}
```

mickaelistria:
That reminds me of https://bugs.eclipse.org/bugs/show_bug.cgi?id=548970 ;)
I think such an annotation to mark code that is expected to run in UIThread would be great in general; and a JUnit runner/extension that is able to look at this annotation on a Test to decide whether to run it in UI Thread of not would be great!

laeubi:
> That reminds me of https://bugs.eclipse.org/bugs/show_bug.cgi?id=548970 

Any coincidence is only random :-P

> I think such an annotation to mark code that is expected to run in UIThread would be great in general; and a JUnit runner/extension that is able to look at this annotation on a Test to decide whether to run it in UI Thread of not would be great!

I really like to go in that direction because it shifts responsibility from Tycho (where it does not belong to), to the actual "library" (SWT/Platform) what is a much better fit.

I'm just noch sure where to best place such code:

1. Should be some at SWT and some at platform?
2. Should it be a separate module or not?

We already have `org.eclipse.test` bundle that has dependencies on junit + platform and some code in Tycho seems to be "inspired" by this. And it seems quite a lot of test bundles actually reference this I'm just not 100% sure about its general purpose and it seems not deployed to maven yet ( @merks ? ).

Of course one might also start with something "fresh" but where should it be located? Tycho? Platform? and own project?

merks:
I didn't think that org.eclipse.test was meant to be "API" used outside of the Platform's own builds.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/test/SWTTest.java
```diff
@@ -0,0 +1,64 @@
...
+ * </p>
+ */
+@NoInstantiate
+public final class SWTTest implements InvocationInterceptor {
+
+	@Override
+	public void interceptTestMethod(Invocation<Void> invocation, ReflectiveInvocationContext<Method> invocationContext,
+			ExtensionContext extensionContext) throws Throwable {
+
+		if (Display.getCurrent() == null) {
```
iloveeclipse, line 46: Should it be `Display.getCurrent() != null` ?

iloveeclipse (commented):
Why is this added on SWT production code, not in SWT tests? 
Eclipse usually does not mix production & test code in one bundle / source directory.

laeubi:
> Why is this added on SWT production code, not in SWT tests?

Because it is not test code but something you will need to use in production see:

> I'm just noch sure where to best place such code ... https://github.com/eclipse-platform/eclipse.platform.swt/pull/2624#issuecomment-3401482402

mickaelistria:
While those are generic about SWT, I suggest we add such code to org.eclipse.ui.tests.harness, which is often used as a test depedency and provides the widely used `DisplayHelper` which is IMO comparable in term of goal (providing API-ish to facilitate writing tests for SWT-based apps).

laeubi:
The place do not matter much to me unless it:

1. is deployed on maven central
2. can be used outside platform without pulling all platform in

akurtakov:
What is the status here?

Issue #2608, opened 2025-10-08
Dialogs open with wrong zoom level


ptziegler:
**Describe the bug**

When switch between different zoom levels, the Dialog seems to remember the previous/initial zoom level. When the dialog is openend, the contents of this dialog are scaled based on the previous zoom and then rescaled to match the monitor zoom.

**To Reproduce**

I'm not sure about the exact steps, as I only noticed this problem some time later. But I switched between 100%, 125% and 150% zoom on my second monitor for some tests and then reverted back to 100%.

**Expected behavior**

The dialog should already be at the correct zoom level when opened.

**Screenshots**

https://github.com/user-attachments/assets/7973578f-3cf9-4d79-930b-7161ecfad6a2

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 

2. Additional OS info (e.g. OS version, Linux Desktop, etc)

Windows 11

**Version since**
2025-12 M1

ptziegler:
The inverse also seems to be true. I've changed my monitor zoom to 125% and restarted the IDE. Now everything is rescaled from 100% -> 125% after the dialog was opened.

https://github.com/user-attachments/assets/f1fc2008-2d6b-49c1-aab7-644a8aee33a2

HeikoKlare:
Thank you for reporting this! I faced the same issue yesterday and was about to document it in an issue. This happens because we introduced async processing of DPI change events: 
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2520

Unfortunately, (child) shells seem to be created with primary monitor zoom (or whatever monitor they are initialized on) and then (asynchronously) process the DPI change event to the monitor they are placed on. Before this was asynchronous, you did not see the additional relayouting, but now you will notice it. We will work on that.

fedejeanne:
# How to reproduce
> [!WARNING]
> The issue is **not** reproducible in the _Runtime Workspace_ (run configuration) that can be started from the IDE but it **is** reproducible in the IDE and in the SDK (see [downloads](https://download.eclipse.org/eclipse/downloads/))

## Setup
- 2 monitors with different zooms

## Steps
- Open the SDK / IDE by double clicking in the `eclipse.exe` in the File Explorer
[...]

PR #2626, opened 2025-10-14
[win32] Always reset cached current DPIChangeEvent

akoch-yatta:
This PR ensures to always reset the currently cached DPIChangeEvent of a Control in any case. Otherwise spawning dpi change events for multiple Controls that are related via parent->child hierarchy could lead to a currentDpiChangeEvent of the parent being canceled by the creation of the event of the child.

So, as as example: there is Composite A and B and A is the parent of B. Now an asynchronous DPI change event for A is created because A is added to a new parent, now the same is done for B while the first event is already propagated the cached event for B could be already there and will be canceled when adding the second event which will cancel further processing. Therefor it is important wo always nullify `Control#currentDpiChangeEvent` after the event for a Control is processed and not only when it is the final `Control`

This fixes the effect in https://github.com/eclipse-platform/eclipse.platform.swt/issues/2608#issuecomment-3400547868.
It can be reproduced in the runtime for me, by moving it to the secondary monitor - primary and secondary must have different zooms of course.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
@@ -5940,13 +5940,13 @@ void sendZoomChangedEvent(Event event, Shell shell) {
 					notifyListeners(SWT.ZoomChanged, event);
 				}
 			} finally {
+				if (event == currentDpiChangeEvent) {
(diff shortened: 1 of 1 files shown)


HeikoKlare:
Closing as superceded by:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2637
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2639

PR #2629, opened 2025-10-14
Automatic refactoring: move Win32DPIUtils.pointToPixel(size,zoom) to DPIUtil

arunjose696:
The change is required for  the below PRs as they would need to do point to pixel conversions

- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2284
- #2622
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2619

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
@@ -24,15 +24,6 @@ import org.eclipse.swt.graphics.*;
 /**
  * This class hold common constants and utility functions w.r.t. to SWT high DPI
  * functionality.
- * <p>
- * The {@code autoScaleUp(..)} methods convert from API coordinates (in
- * SWT points) to internal high DPI coordinates (in pixels) that interface with
- * native widgets.
- * </p>
- * <p>
- * The {@code autoScaleDown(..)} convert from high DPI pixels to API coordinates
- * (in SWT points).
- * </p>
  *
  * @since 3.105
  */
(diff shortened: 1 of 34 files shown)


bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
```diff
@@ -257,6 +257,15 @@ public static void validateLinearScaling(ImageDataProvider provider) {
 	}
 }
 
+/**
+ * Auto-scale up int dimensions to match the given zoom level
+ */
```
HeikoKlare: Please remove this comment. We replaced the "scale up"/"scale down" terminology with "pointToPixel"/"pixelToPoint", so it's misleading now. And this is actually no "auto"-scaling (which used to use an automatically determined zoom).

tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/tests/win32/Win32DPIUtilTests.java
```diff
@@ -146,15 +147,15 @@ public void scaleUpInteger() {
 		int valueAt150 = 8;
 		int valueAt100 = 5;
 
-		int scaledValue = Win32DPIUtils.pointToPixel(valueAt100, 200);
```
HeikoKlare, line 149: Shouldn't the test be moved to `DPIUtilTests` instead of adapting the referenced class here?

HeikoKlare (commented):
Moving this to `DPIUtil` seems reasonable to me, as `pixelToPoint` is also part of that class. I just have minor comments for proper refactoring and cleanup.

tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/tests/win32/Win32DPIUtilTests.java
arunjose696: Moved these to to `DPIUtilTests`

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
arunjose696: Removed the comment, also could see in Javadoc of DPIUtils class there were some mentions about Auto-scaleup which has also now been removed in this change

HeikoKlare: Great, thank you for checking and fixing that!

PR #2634, opened 2025-10-15
[Win32] Reuse zoom handle for drawing image at size if appropriate

HeikoKlare:
When drawing an image at a specified size, currently a new image handle is always created based on on-demand loaded image data. In case the fallback to loading image data for the best fitting zoom is used, a handle for that zoom may already exist and could be reused but is currently not considered.

With this change, an existing handle is used if appropriate instead of unnecessarily creating a temporary handle.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
@@ -136,30 +136,66 @@ public final class Image extends Resource implements Drawable {
 
 	private List<Consumer<Image>> onDisposeListeners;
 
-	private record CachedHandle(ImageHandle handleContainer, int requestedWidth, int requestedHeight) {
+	private class HandleAtSize {
+		private ImageHandle handleContainer = null;
+		private int requestedWidth = -1;
+		private int requestedHeight = -1;
 
 		public void destroy() {
-			if (handleContainer != null) {
+			if (handleContainer != null && !zoomLevelToImageHandle.containsValue(handleContainer)) {
 				handleContainer.destroy();
 			}
+			handleContainer = null;
+			requestedWidth = -1;
+			requestedHeight = -1;
 		}
 
-		public boolean isReusable(int height, int width) {
-			if(handleContainer == null) {
+		public ImageHandle refresh(int width, int height) {
+			if (!isReusable(width, height)) {
+				destroy();
+				requestedWidth = width;
+				requestedHeight = height;
+				handleContainer = createHandleAtExactSize(width, height)
+						.orElseGet(() -> getOrCreateImageHandleAtClosestSize(width, height));
+			}
+			return handleContainer;
+		}
+
+		private boolean isReusable(int width, int height) {
+			if (handleContainer == null || handleContainer.isDisposed()) {
 				return false;
 			}
 			return (requestedHeight == height && requestedWidth == width)
 					|| (handleContainer.height == height && handleContainer.width == width);
-
 		}
 
-		public ImageHandle getHandle() {
-			return handleContainer;
+		private Optional<ImageHandle> createHandleAtExactSize(int width, int height) {
+			Optional<ImageData> imageData = imageProvider.loadImageDataAtExactSize(width, height);
+			if (imageData.isPresent()) {
+				ImageData adaptedData = adaptImageDataIfDisabledOrGray(imageData.get());
+				ImageHandle imageHandle = init(adaptedData, -1);
+				return Optional.of(imageHandle);
+			}
+			return Optional.empty();
 		}
-	};
 
-	// Initialize lastRequestedHandle with -1 for size-related fields to indicate uninitialized values
-	CachedHandle lastRequestedHandle = new CachedHandle(null, -1, -1);
+		private ImageHandle getOrCreateImageHandleAtClosestSize(int widthHint, int heightHint) {
+			Rectangle bounds = getBounds(100);
+			int imageZoomForWidth = 100 * widthHint / bounds.width;
+			int imageZoomForHeight = 100 * heightHint / bounds.height;
+			int imageZoom = DPIUtil.getZoomForAutoscaleProperty(Math.max(imageZoomForWidth, imageZoomForHeight));
+			ImageHandle bestFittingHandle = zoomLevelToImageHandle.get(imageZoom);
+			if (bestFittingHandle == null) {
+				ImageData bestFittingImageData = imageProvider.loadImageData(imageZoom).element();
+				ImageData adaptedData = adaptImageDataIfDisabledOrGray(bestFittingImageData);
+				bestFittingHandle = init(adaptedData, -1);
+			}
+			return bestFittingHandle;
+		}
+
+	}
+
+	private final HandleAtSize lastRequestedHandle = new HandleAtSize();
 
 private Image (Device device, int type, long handle, int nativeZoom) {
 	super(device);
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -141,33 +141,66 @@ public final class Image extends Resource implements Drawable {
...
-				return handleContainer.handle;
+		private ImageHandle getOrCreateImageHandleAtClosestSize(int widthHint, int heightHint) {
+			Rectangle bounds = getBounds(100);
+			int imageZoomForWidth = 100 * widthHint / bounds.width;
+			int imageZoomForHeight = 100 * heightHint / bounds.height;
+			int imageZoom = DPIUtil.getZoomForAutoscaleProperty(Math.max(imageZoomForWidth, imageZoomForHeight));
+			ImageHandle bestFittingHandle = zoomLevelToImageHandle.get(imageZoom);
+			if (bestFittingHandle == null) {
+				ImageData bestFittingImageData = imageProvider.loadImageData(imageZoom).element();
+				ImageData adaptedData = adaptImageDataIfDisabledOrGray(bestFittingImageData);
```
arunjose696, line 190: Doesnt the` loadImageData(zoom)` already do the  `adaptImageDataIfDisabledOrGray`,  so I think this is not needed.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -854,15 +887,8 @@ interface HandleAtSizeConsumer {
...
-		ImageData imageData;
-		imageData = this.imageProvider.loadImageDataAtSize(widthHint, heightHint);
-		lastRequestedHandle.destroy();
-		ImageHandle handleContainer = init(imageData, -1);
-		lastRequestedHandle = new CachedHandle(handleContainer, widthHint, heightHint);
-	}
-	handleAtSizeConsumer.accept(lastRequestedHandle.getHandle(),
-			new Point(lastRequestedHandle.handleContainer().width, lastRequestedHandle.handleContainer().height));
+	ImageHandle imageHandle = lastRequestedHandle.refresh(widthHint, heightHint);
+	handleAtSizeConsumer.accept(imageHandle.handle, new Point(imageHandle.width, imageHandle.height));
```
arunjose696: Since we are now passing the `imageHandle` directly, the `HandleAtSizeConsumer` functional interface could be replaced with a standard `Consumer<ImageHandle>`, as all other arguments can be derived from the `imageHandle` itself.

HeikoKlare: `ImageHandle` is private and I did not want to expose it just for this single case. If we merge https://github.com/eclipse-platform/eclipse.platform.swt/pull/2651 soon (which will increase visibility of the `ImageHandle` class), I will adapt this accordingly.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
HeikoKlare: If I am not mistaken, `adaptImageDataIfDisabledOrGray()` is only called at places where a handle is created, such as `initializeHandleFromSource()`. None of the `loadIamgeData()` implementations seems to call it. Can you please check if I miss something, @arunjose696?

arunjose696: I see the following call stack

 loadImageData(int) -> getImageMetadata(zoomContext) -> newImageHandle(zoomContext)-> initializeHandleFromSource(zoomContext) -> adaptImageDataIfDisabledOrGray()

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
arunjose696: Makes sense

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
HeikoKlare: I see. This seems to be an inconsistency that some `loadImageData()` implementations adapt the image data while others don't. Precisely, the `PlainImageProviderWrapper` and `ExistingImageHandlerProviderWrapper` seem to be the only ones that adapt them (via calling `getClosedAvailableImageData()`). We should try to streamline that (independent of this PR), as we unnecessarily adapt images twice (and maybe there are even some that we do not adapt at all).

arunjose696: > We should try to streamline that (independent of this PR), as we unnecessarily adapt images twice (and maybe there are even some that we do not adapt at all).

Yeah this could be streamlined in a seperate PR if running adaptImageDataIfDisabledOrGray is a no-op if applied for second time, I could see in addition to `PlainImageProviderWrapper` and `ExistingImageHandlerProviderWrapper` , the ImageGCDrawer wrapper also calls `adaptImageDataIfDisabledOrGray`

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
HeikoKlare: I have created:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2652

We should merge that one first, so I can adapt this on top of that refactoring.

HeikoKlare: I rebased on top of that PR so that the code is not simplified as proposed.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
HeikoKlare: Agreed. I have created an item in our backlog so that we can have a look at streamlining this. Thank you for the catch!

akoch-yatta:
Overall this PR looks good and seem to work properly due to my testing. 

But I was wondering, why don't we do a similar check in `HandleAtSize#createHandleAtExactSize` as well? So iterating over the handles and seeing of one size matches - each `ImageHandle` knows its size. With `CTabRendering` you immediately run into the scenario where a handle (for a SVG) is created unnecessarily. I am fine with doing that in a separate PR, but does anything speak against it?

HeikoKlare:
> But I was wondering, why don't we do a similar check in `HandleAtSize#createHandleAtExactSize` as well? So iterating over the handles and seeing of one size matches - each `ImageHandle` knows its size. With `CTabRendering` you immediately run into the scenario where a handle (for a SVG) is created unnecessarily. I am fine with doing that in a separate PR, but does anything speak against it?

That definitely makes sense. I added the simple required logic with 12d5fc5.
I am not completely sure anymore why I didn't do it before. As you pointed out for the already implemented case of searching for a fitting handle first, you need to be in a situation where a zoom-based handle has been created at all. In many cases, you may only use the draw-at-size functionality without ever requesting a handle at zoom. Maybe that's why I didn't implement it (along with thinking that the implementation would be more complicated). Anyway, now all potential reuse scenarios should be covered.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -136,30 +136,75 @@ public final class Image extends Resource implements Drawable {
...
+			if (handleContainer == null || handleContainer.isDisposed()) {
 				return false;
 			}
 			return (requestedHeight == height && requestedWidth == width)
 					|| (handleContainer.height == height && handleContainer.width == width);
+		}
 
+		private Optional<ImageHandle> getHandleAtExactSize(int width, int height) {
+			for (ImageHandle handle : zoomLevelToImageHandle.values()) {
+				if (handle.width == width && handle.height == height) {
```
arunjose696: I think this would cause a slight issue as if the image was requested at for eg 2X the height elsewhere, they would not be loaded with imageDataAtSizeProvider, we would then end up using this image while doing the drawing even though it is possible to load the image at the given target size.

HeikoKlare: Thank you! That was exactly the reason for not doing it. I just adapted the PR again. Maybe we can still integrate something like this as a follow up in case we maybe revise the `ImageDataAtSizeProvider`.

PR #2637, opened 2025-10-16
[Win32] Fix wrong layout when opening shell or reparenting control #2608

HeikoKlare:
In multiple situation, relayouting due to DPI change events does not work as expected:
- When opening child shells, they may be relayouted after being opened if their position changed to a different monitor between the OS handle was created and a new position was set
- When reparenting controls between different shells, the target shell may not be properly relayout at all

Since the user experience when performing these two cases asynchronously is not good anyway (e.g., reparenting leads to unexpected asynchronous relayouting operations), this change adapts the two use cases to process the DPI change synchronously.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2608

This is an addition to https://github.com/eclipse-platform/eclipse.platform.swt/pull/2636, which already fixes some of the issue (in the best possible way) while this resolved the remaining issues (such as broken perspective switching) on top.

You can test this by having two monitors with different zoom, starting an SDK product and switching perspectives on either of the monitors. Without this fix, you will see the broken layout as described here: https://github.com/eclipse-platform/eclipse.platform.swt/issues/2608#issuecomment-3400547868
In addition, you can open a child shell (e.g., an "Open Type" dialog) and move it to a different monitor, such that it reopens on that other monitor afterwards. Without this fix, when reopening the dialog you will see effects such as in https://github.com/eclipse-platform/eclipse.platform.swt/issues/2608#issuecomment-3381694689.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/DPITestUtil.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/DPITestUtil.java
@@ -28,7 +28,7 @@ public final class DPITestUtil {
 
 	public static void changeDPIZoom (Shell shell, int nativeZoom) {
 		DPIUtil.setDeviceZoom(nativeZoom);
-		Event event = shell.createZoomChangedEvent(nativeZoom);
+		Event event = shell.createZoomChangedEvent(nativeZoom, true);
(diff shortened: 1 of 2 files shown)


akoch-yatta (approved):
Changes sound reasonable for the scenarios it describes. I tested it with different monitor setting and scenarios like opening child shells, triggering parent changes or moving shells between monitors and it works as expected and I didn't find a regression

PR #2640, opened 2025-10-17
[Win32] Fix shell zoom when disabling autoscaling on shell

HeikoKlare:
Recent simplifications to how autoscaling can be disabled for controls (https://github.com/eclipse-platform/eclipse.platform.swt/pull/2599) leads to a wrong native zoom being reported via getData("SHELL_ZOOM") when autoscaling is disabled at the shell itself. Instead of the actual zoom it always reports 100 because the shell's nativeZoom is overwritten when autoscaling is disabled for the shell.

This change adapts the way in which the shell zoom is retrieved when calling getData("SHELL_ZOOM") to be independent of the stored nativeZoom value.


akoch-yatta (approved):
Change is necessary and works to me as expected

PR #2642, opened 2025-10-17
Use assertThrows in Test*BidiSegmentListener

akurtakov:


akurtakov:
MacOS failure is network problem.Ignoring.

Issue #2643, opened 2025-10-17
[Win32] Some controls are incorrectly placed


HeikoKlare:
**Describe the bug**
Some controls receive wrong placements as a result of:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2636
- 
Originally reported by Ed: https://github.com/eclipse-platform/eclipse.platform.swt/issues/2641#issuecomment-3414534481

**To Reproduce**
Open the preferences in an up-to-date Eclipse SDK.

**Expected behavior**
Controls should be placed properly.

**Screenshots**

<img width="640" height="468" alt="Image" src="https://github.com/user-attachments/assets/c8b219fb-2d3c-41f8-b1fa-36d5f5090fb6" />

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [x] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 

**Version since**
I20251017-0000

**Workaround (or) Additional context**
To me it looks as if the new control initialization logic should only be applied to shells and not to every control as this is obviously a regression of applying the new position initialization to controls _inside_ a shell.

merks:
It looks good now. Thank you.

PR #2644, opened 2025-10-17
[Win32] Fix initial control location #2643

HeikoKlare:
With a recent change to the initial location of newly created controls, the locations of some controls inside shells are broken. The intent of that previous change was to improve the initial location of shells.

This change adapts the previous modification to only apply an improved pattern for initial shell placement when the control is actually shell and to keep the previous behavior for all other controls.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2643


akoch-yatta (approved):
As this change was targeted on fixing the positioning on Shell anyway it makes sense to limit the affected places to Shell and move the custom code to there

fedejeanne (approved):
Makes sense to localize the fix for shells and leave out other controls.

Thank you, Heiko!

PR #2646, opened 2025-10-17
[Win32] Autoscale cursor even on widgets that have autoscaling disabled

HeikoKlare:
Cursors are initialized with the zoom of the control on which they are set. However, cursors are usually expected to scale according to the autoscaled UI around those controls for which autoscaling was disabled.

This change adapts the cursor initializations to use the autoscale zoom of the containing shell instead of the zoom of the control itself. In consequence, cursors will have the same size in the whole shell, no matter on which control they are set.

Contributes to https://github.com/eclipse-gef/gef-classic/pull/821

### How to test
Best possibility to productively test is with Draw2d/GEF and according examples.

For simple testing, the following snippet can be used (adapted from https://github.com/eclipse-platform/eclipse.platform.swt/issues/2308). The lines for different properties `AUTOSCALE_DISABLED`, `swt.autoScale` and `swt.autoScale.updateOnRuntime` can be changed to test different combinations. The combination that is activated by default shows the bug in the existing implementation. The others allow to test that we do not have possible regressions regarding previous issue reports, such as https://github.com/eclipse-platform/eclipse.platform.swt/issues/2308.

In any case, a primary monitor with zoom != 100% should be used.

```java
public class Snippet1 {

//		System.setProperty("swt.autoScale", "200");
//		System.setProperty("swt.autoScale", "false");
		Cursor cursor = new Cursor(null, getImageData(100), 0, 0);

		Shell shell = new Shell(SWT.NO_TRIM);
		shell.setSize(100, 100);
		Composite c = new Composite(shell, SWT.NONE);
		c.setBackground(new Color(0, 255, 0));
		c.setData("AUTOSCALE_DISABLED", true);
		c.setCursor(cursor);
		shell.open();

		while (!display.isDisposed()) {
			display.readAndDispatch();
		}
		cursor.dispose();
	}

	private static ImageData getImageData(int zoom) {
		Image image = new Image(null, 50, 50);
		try {
			GC gc = new GC(image);
			gc.setBackground(new Color(255, 0, 0));
			gc.fillRectangle(0, 0, 50, 50);
			return image.getImageData(zoom);
		} finally {
			image.dispose();
		}
	}
}
```

The red cursor is expected to always have half the width and height of the green shell.

With the default configuration is looks like this without the change:
cursor_old

And this is how it looks with the change:
cursor_new

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
@@ -1283,7 +1283,7 @@ public void setData(String key, Object value) {
 		if (autoScaleDisabled) {
 			this.nativeZoom = 100;
 		} else {
-			this.nativeZoom = (int) getData(DATA_SHELL_ZOOM);
+			this.nativeZoom = getShellZoom();
 		}
 	}
 }
@@ -4817,6 +4817,13 @@ int getZoom() {
 	return super.getZoom();
 }
 
+int getShellZoom() {
+	if (getData(DATA_SHELL_ZOOM) instanceof Integer shellZoom) {
+		return shellZoom;
+	}
+	return nativeZoom;
+}
+
 abstract TCHAR windowClass ();
 
 abstract long windowProc ();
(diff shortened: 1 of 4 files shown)


arunjose696:
One minor observation: when I start GEF with "swt.autoScale" set to "false" on a 200% monitor, some cursors (e.g., select, marquee) are scaled, while others (e.g., connection) are not. This creates a slight inconsistency, but I assume this has always been the behavior.

Apart from this I have tested the changes using the snippet and also on GEF, and the cursors are now scaled correctly in both cases compared to master. The code looks good as well.

HeikoKlare:
> One minor observation: when I start GEF with "swt.autoScale" set to "false" on a 200% monitor, some cursors (e.g., select, marquee) are scaled, while others (e.g., connection) are not. This creates a slight inconsistency, but I assume this has always been the behavior.

Yes, that's also my understanding. It should have always been like this (either intended or by accident), as disabling auto-scaling with `swt.autoScale=false` results in every control/image/cursor etc. not being scaled. For the cursors, this does only apply to custom cursors and not the ones coming from the OS, which makes the experience inconsistent. This is why, in my opinion, cursors should always be scaled according to the current monitor on which they are used (so that they conform to the OS cursors), which is exactly what happens now when monitor-specific scaling is enabled.

Issue #2537, opened 2025-09-24
[Win] Images in MenuItems are scaled incorrectly (regression in R4_37)


tmssngr:
**Describe the bug**
With the release of SWT R4_37 the Windows zoom level causes (bitmap) images to be upscaled though `swt.autoScale` had been set to `false`.

**To Reproduce**
- run this snippet on a Windows with a monitor scaling set to >100%:
```java
```
- right-click to open a context menu

=> Results:
1) the painted image is not scaled (correct)
2) the image of the SWT.PUSH MenuItem is scaled incorrectly
3) the image of the unchecked SWT.CHECK MenuItem is scaled incorrectly
4) the image of the selected SWT.CHECK MenuItem is not scaled
5) a checkmark is drawn onto the image of the SWT.CHECK MenuItem (incorrect)

2)-5) are regressions

**Expected behavior**
- if `swt.autoScale` is set to `true`, scaling must not happen at all
- no checkmark must be drawn onto an image (we solve it in our applications by using different images)

**Screenshots**
Windows 11 with 150% zoom.
<img width="248" height="209" alt="Image" src="https://github.com/user-attachments/assets/a4b6cc2b-ddd6-4faf-9dd7-1d5d95ddbc1b" />

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 

2. Additional OS info (e.g. OS version, Linux Desktop, etc)

3. JRE/JDK version
 
**Version since**
R4_37, still reproducible with the latest master build.

**Workaround (or) Additional context**
Workaround: Use SWT R4_36.

HeikoKlare:
Thank you for the report.

> * the image of the SWT.PUSH MenuItem is scaled incorrectly
> * the image of the unchecked SWT.CHECK MenuItem is scaled incorrectly
> * the image of the selected SWT.CHECK MenuItem is not scaled
  
Seems like this was introduced with:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2421

We will have a look.

> * a checkmark is drawn onto the image of the SWT.CHECK MenuItem (incorrect)

Why is that incorrect? https://eclipse.dev/eclipse/markdown/?f=news/4.37/platform.md#toggle-indicator-in-windows-11-menus

> **Workaround (or) Additional context**
> Workaround: Use SWT R4_36.

Note that another workaround is of course to use auto-scaling.

[...]

PR #2648, opened 2025-10-17
[Win32] Only scale menu item images on monitor-specific scaling #2537

HeikoKlare:
The current mechanism to adapt the size of menu item images based on Windows system metrics, primary monitor zoom at startup and the current monitor zoom ensures that menu item images always fit to the required size for that menu. Since menus and their sizes are managed by the OS, adapting the image size as currently implemented leads to the best fitting image sizes.
However, this behavior is incompatible with the assumptions one could previously make on the sizes of those images. In particular, depending on the used auto-scale mode different kinds of scaling for those images were in place.

To achieve the best solution for all cases, this change results in the following:
- When monitor-specific scaling is enabled, menu item images are properly scaled as required for the menu. This is independent of whether any other additional legacy settings regarding auto-scaling is made, which will be overwritten by this.
- When monitor-specific scaling is disabled, the previous behavior is restored such that images are just scaled depending on the auto-scale mode but do not take into account which size is actually needed for that menu.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2537

## How to test

The issue can be reproduced and the fix tested with the snippet given in the original issue:
- https://github.com/eclipse-platform/eclipse.platform.swt/issues/2537#issue-3448031683

It also allows to test different combinations of auto-scale properties together with menus.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MenuItem.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/MenuItem.java
@@ -907,6 +907,9 @@ private long getMenuItemIconSelectedBitmapHandle() {
 }
 
 private int adaptZoomForMenuItem(int currentZoom, Image image) {
+	if (!display.isRescalingAtRuntime()) {
+		return DPIUtil.getZoomForAutoscaleProperty(currentZoom);
+	}
 	int primaryMonitorZoomAtAppStartUp = Win32DPIUtils.getPrimaryMonitorZoomAtStartup();
 	/*
 	 * Windows has inconsistent behavior when setting the size of MenuItem image and


HeikoKlare:
@tmssngr this is hopefully the best tradeoff for all current use cases. It should use the new, proper menu scaling when monitor-specific scaling is enabled (without any additional auto-scaling settings), and it should use the unmodified preexisting scaling behavior when monitor-specific scaling is disabled (to allow backwards compatibilities).

fedejeanne (approved):
LGTM:

image

tmssngr:
Sorry for the late reply. I had vacation. I don't think that scaling up bitmaps is a good idea, especially for fractional scaling.

HeikoKlare:
> I don't think that scaling up bitmaps is a good idea, especially for fractional scaling.

Thank you for that opinion. I am not sure what it has to do with this change and I am also not sure why Eclipse should stick to the broken behavior of integrating unscaled images into a scaled UI while the rest of the world is fine with implementing proper UI scaling without a bunch of options that are difficult to maintain.

tmssngr:
Typical scaling UIs have (more or less) blurry images, maybe by drawing directly (e.g. using SVG) or by scaling larger bitmaps down. I have not seen any application that up-scales bitmap images on fractional scaling, because the result looks terrible.

HeikoKlare:
> Typical scaling UIs have (more or less) blurry images, maybe by drawing directly (e.g. using SVG) or by scaling larger bitmaps down. I have not seen any application that up-scales bitmap images on fractional scaling, because the result looks terrible.

And how is that different from what SWT does? We only scale up if we have no higher-resolution image available (or if the higher resolution one cannot properly be scaled down).

tmssngr:
In our case the ImageDataProvider can provide ImageData for 100% and 200%. I reckon scaling the 200% image down would have created better results than scaling the 100% image up. But I would leave the decision whether to scale bitmaps at all to the user of SWT. This decision can be different in different SWT based applications.

HeikoKlare:
Best is of course to just use SVGs and render them at appropriate size. For the common case of providing 100% and 200% image versions, scaling the 200% version down to 125% usually gives worse results than scaling the 100% version of, which is why the algorithm in SWT favors upscaling in that case.

Issue #2598, opened 2025-10-06
Can we change cut/copy/paste API from synchronous to asynchronous?


jonahgraham:
_Originally posted by @raghucssit in https://github.com/eclipse-platform/eclipse.platform.swt/issues/2489#issuecomment-3269356329_

>> @raghucssit should we enhance javadoc to mention that calling copy / paste will not _immediate_ update the state and one should possibly use asyncExec wif accessing the control is desired (that would probably also work for the test to use async exec but it of course complicates the test assertions maybe).

> I have updated copy/paste doc.

I think this is incorrect to change this API from sync to async. This is probably a breaking change in SWT, but I pose this as a question because perhaps it is an acceptable change, but it is one that may have to extend to many more places (see below).

For example https://github.com/eclipse-platform/eclipse.platform.swt/issues/2491#issuecomment-3279584420 still failing is a side effect of this change because the asynchronous method\* gtk_text_buffer_paste_clipboard is called just before calling applySegments, but applySegments needs the new value of the text buffer to operate properly.

Another example is this test that does effectively what `TextCellEditor.performPaste` does, this test passes on GTK3, but fails with current GTK4 implementation.

```java
@Test
public void test_paste_and_selection() {
	text.setText("01234567890");
	text.setSelection(2, 4);
	assertEquals("01234567890", text.getText());
	text.copy();
	text.setSelection(3, 4);
	assertEquals(1, text.getSelectionCount());
	assertEquals(new Point(3, 4), text.getSelection());

	// Simulate what code such as org.eclipse.jface.viewers.TextCellEditor.performPaste() does:
	text.paste();
	assertEquals(0, text.getSelectionCount());
	assertEquals(new Point(5, 5), text.getSelection());
	assertEquals("012234567890", text.getText());
}
```

\* GTK4 changed this (and similar) methods from sync to async.

laeubi:
> I think this is incorrect to change this API from sync to async.

The API was never changed. a change in API would have been for example from `public void paste ()` to `public CompletableFuture<?> paste ()` or `public void paste (Runnable completionHandler)`.

Apart from that the API doc only states

> Pastes text from clipboard.
> The selected text is deleted from the widget
> and new text inserted from the clipboard.

[...]

PR #2650, opened 2025-10-17
Add getContentsAsync methods to Clipboard

jonahgraham:
Obtaining data from the clipboard is fundamentally an asynchronous operation, the event loop, at least at the OS level, needs to run for the data to become available.

This PR proposes new API that provides asynchronous equivalents of getContents methods called getContentsAsync that return CompleteableFutures.

This new API will allow SWT API consumers to interact better with the clipboard when the OS (window system) API is actually implemented in an asynchronous manner.

GTK3 and cocoa provides API that spins the event loop as needed so that a synchronous looking API can be provided to the user of the API. For example, in SWT we use gtk_clipboard_wait_for_contents which spins the loop in the GTK library itself
[here](https://gitlab.gnome.org/GNOME/gtk/-/blob/716458e86a222f43e64f7a4feda37749f3469ee4/gtk/gtkclipboard.c#L1436)

GTK4 does not provide such an API and leaves it to the user of the API to spin the event loop.

Win32 is somewhat of a hybrid. The API appears synchronous, but SWT needs to spin in a couple of places, such as
[getData](https://github.com/eclipse-platform/eclipse.platform.swt/blob/63e8925dc77db3b93ef521dc6cf2bd6ded7bab64/bundles/org.eclipse.swt/Eclipse%20SWT%20Drag%20and%20Drop/win32/org/eclipse/swt/dnd/Transfer.java#L46)

Part of #2598 and #2126

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Drag and Drop/cocoa/org/eclipse/swt/dnd/Clipboard.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Drag and Drop/cocoa/org/eclipse/swt/dnd/Clipboard.java
@@ -13,6 +13,7 @@
  *******************************************************************************/
 package org.eclipse.swt.dnd;
 
+import java.util.concurrent.*;
 
 import org.eclipse.swt.*;
 import org.eclipse.swt.internal.cocoa.*;
@@ -199,6 +200,10 @@ public void dispose () {
  * clipboard.  Refer to the specific subclass of <code>Transfer</code> to
  * determine the type of object returned.
  *
+ * <p>This method may iterate the event loop to be able to complete. Use
+ * {@link #getContentsAsync(Transfer)} to have control over when the event
+ * loop runs.</p>
+ *
  * <p>The following snippet shows text and RTF text being retrieved from the
  * clipboard:</p>
  *
@@ -225,6 +230,7 @@ public void dispose () {
  * </ul>
  *
  * @see Transfer
+ * @see #getContentsAsync(Transfer)
  */
 public Object getContents(Transfer transfer) {
 	return getContents(transfer, DND.CLIPBOARD);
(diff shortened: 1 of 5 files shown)


jonahgraham:
One aspect I am unsure of is the naming here.

We can't simply overload `getContents` because the only difference in the new method is the return type. But in `java.util.concurrent` it seems generally the `Async` suffixed methods imply running in a separate (asynchronous) thread.

Am I overthinking this, is `getContentsAsync` a good name? @laeubi any thoughts?

HeikoKlare:
> Am I overthinking this, is `getContentsAsync` a good name?

In my opinion, thinking about that name is definitely valuable. It might be mistaken as "I just have to wait long enough for the contents to be available" (as it would be the case with the referenced concurrent Java API using an `Async` suffix). E.g., you could use `getContentsAsync().join()`, but that will never return if no one else spins the event loop. There is a slight analogy to `Display#asyncExec()` as that will also only be executed if you spin the event loop (though there is no `CompletableFuture` but you could, e.g.,  put a barrier in the passed runnable), just that `Display#asyncExec()` is more common and users usually know how it works.
On the other hand, I have no better name proposal and a sophisticated documentation of what "async" means in this case is given by the Javadoc, so unless someone has a better proposal I would be fine the currently proposed name.

jonahgraham:
> I would be fine the currently proposed name.

Great. Thanks @HeikoKlare for having a look and sharing your thoughts here. I'll leave this open a little bit longer in case anyone else has thoughts or concerns.

laeubi:
> > I would be fine the currently proposed name.
> 
> Great. Thanks @HeikoKlare for having a look and sharing your thoughts here. I'll leave this open a little bit longer in case anyone else has thoughts or concerns.

I'm also fine with that name, an alternative I often use in this case is simply omit the "get" and use the "shortform" what is the `CompletableFuture<?> contents();` but not everyone is happy with that and its more a personal preference.

In any case every java developer should be aware that joining a feature is always a bad thing and a source of code-smell.

jonahgraham:
> alternative I often use in this case is simply omit the "get" and use the "shortform" 

I am pretty fond of this too, especially when making new API these days. I like "ProcessBuilder" in Java in this way.

akurtakov:
I believe this is ready to go . Only request from my side is to add a reference in getContents javadoc refering to the new method and guiding users to choose it.

jonahgraham:
> I believe this is ready to go . Only request from my side is to add a reference in getContents javadoc refering to the new method and guiding users to choose it.

Done: See [diffs here](https://github.com/eclipse-platform/eclipse.platform.swt/compare/54fcebdc54f9bf4e26f935755ab4e8c91dc1686b..34207113f7321799f4f4474f5c73f994f4f331d1)

PR #2652, opened 2025-10-20
[Win32] Remove obsolete HandleAtSizeConsumer

HeikoKlare:
With latest refactorings, the HandleAtSizeConsumer inside Image became obsolete as the ImageHandle passed to it already contains the other information encapsulated in that consumer. Thus, this follow-up refactoring replaces the HandleAtSizeConsumer with a simple ImageHandle consumer. To do so, it properly encapsulates and exposes the width/height information in an ImageHandle. It also aligns the name of the method processing such a consumer to fit the actual behavior and the according MacOS implementation.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
@@ -845,12 +845,7 @@ ImageHandle getHandle (int targetZoom, int nativeZoom) {
 	return getImageMetadata(zoomContext);
 }
 
-@FunctionalInterface
-interface HandleAtSizeConsumer {
-	void accept(ImageHandle imageHandle, Point handleSize);
-}
-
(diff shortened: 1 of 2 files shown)


arunjose696:
Changes look good to me and it has made the `ImageHandle` and `drawImage` cleaner than before. I just have minor nit comment

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2834,7 +2830,15 @@ long getHandle() {
...
-	public Rectangle getBounds() {
+	int getWidth() {
+		return width;
+	}
+
+	int getHeight() {
+		return height;
+	}
+
+	Rectangle getBounds() {
```
arunjose696, line 2841: minor NIT : @HeikoKlare could see the access modifier of getBounds is changed from public to package protected this makes sense as the class is package protected but then wouldnt it also make sense to make the constructor `ImageHandle(long handle, int zoom)` package protected as well

HeikoKlare: Good catch! Adapted with latest commit.

PR #2653, opened 2025-10-21
[GTK] Fix scaling of image copies and stylings

HeikoKlare:
The image copy and styling operation via the constructor Image(Device, Image, int) does currently not properly take into account a potential scaling of the original image. The surface for the new image is always created with the size in points, i.e., according to a 100% scaling. When the image is used at 200% it will thus appear blurry.

This change adapts the image copy and styling constructor to properly take the scaling of the original image into account. It make the constructor create a surface with a proper size and scaling.

This will also resolve the issues discussed in:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2622

### Before
image

### After
image

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Image.java
@@ -278,16 +278,18 @@ public Image(Device device, Image srcImage, int flag) {
 	if (flag != SWT.IMAGE_DISABLE) transparentPixel = srcImage.transparentPixel;
 
 	long imageSurface = srcImage.surface;
-	int width = this.width = srcImage.width;
-	int height = this.height = srcImage.height;
+	this.width = srcImage.width;
+	this.height = srcImage.height;
 	int format = Cairo.cairo_surface_get_content(imageSurface) == Cairo.CAIRO_CONTENT_COLOR ? Cairo.CAIRO_FORMAT_RGB24 : Cairo.CAIRO_FORMAT_ARGB32;
 	boolean hasAlpha = format == Cairo.CAIRO_FORMAT_ARGB32;
-	surface = Cairo.cairo_image_surface_create(format, width, height);
+	int dataWidth = DPIUtil.pointToPixel(this.width, DPIUtil.getDeviceZoom());
+	int dataHeight= DPIUtil.pointToPixel(this.height, DPIUtil.getDeviceZoom());
+	surface = Cairo.cairo_image_surface_create(format, dataWidth, dataHeight);
 	if (surface == 0) SWT.error(SWT.ERROR_NO_HANDLES);
-	if (DPIUtil.getDeviceZoom() != currentDeviceZoom) {
-		double scaleFactor = DPIUtil.getDeviceZoom() / 100f;
-		Cairo.cairo_surface_set_device_scale(surface, scaleFactor, scaleFactor);
-	}
+	double[] scaleX = new double[1];
+	double[] scaleY = new double[1];
+	Cairo.cairo_surface_get_device_scale(imageSurface, scaleX, scaleY);
+	Cairo.cairo_surface_set_device_scale(surface, scaleX[0], scaleY[0]);
 	long cairo = Cairo.cairo_create(surface);
 	if (cairo == 0) SWT.error(SWT.ERROR_NO_HANDLES);
 	Cairo.cairo_set_operator(cairo, Cairo.CAIRO_OPERATOR_SOURCE);
@@ -306,9 +308,9 @@ public Image(Device device, Image srcImage, int flag) {
 		switch (flag) {
 			case SWT.IMAGE_DISABLE: {
 				byte[] line = new byte[stride];
-				for (int y=0; y<height; y++) {
+				for (int y=0; y<dataHeight; y++) {
 					C.memmove(line, data + (y * stride), stride);
-					for (int x = 0, offset = 0; x < width; x++, offset += 4) {
+					for (int x = 0, offset = 0; x < dataWidth; x++, offset += 4) {
 						int a = line[offset + oa] & 0xFF;
 						int r = line[offset + or] & 0xFF;
 						int g = line[offset + og] & 0xFF;
@@ -339,9 +341,9 @@ public Image(Device device, Image srcImage, int flag) {
 			}
 			case SWT.IMAGE_GRAY: {
 				byte[] line = new byte[stride];
-				for (int y=0; y<height; y++) {
+				for (int y=0; y<dataHeight; y++) {
 					C.memmove(line, data + (y * stride), stride);
-					for (int x=0, offset = 0; x<width; x++, offset += 4) {
+					for (int x=0, offset = 0; x<dataWidth; x++, offset += 4) {
 						int a = line[offset + oa] & 0xFF;
 						int r = line[offset + or] & 0xFF;
 						int g = line[offset + og] & 0xFF;


HeikoKlare:
@akurtakov since you are the GTK expert and also recently cleaned up autoscaling stuff that, if I am not mistaken, was necessary before Cairo autoscaling, maybe you want to have a look at this fix for image scaling?

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Image.java
```diff
@@ -278,16 +278,16 @@ public Image(Device device, Image srcImage, int flag) {
...
+	int dataWidth = DPIUtil.pointToPixel(this.width, DPIUtil.getDeviceZoom());
+	int dataHeight= DPIUtil.pointToPixel(this.height, DPIUtil.getDeviceZoom());
+	surface = Cairo.cairo_image_surface_create(format, dataWidth, dataHeight);
 	if (surface == 0) SWT.error(SWT.ERROR_NO_HANDLES);
-	if (DPIUtil.getDeviceZoom() != currentDeviceZoom) {
-		double scaleFactor = DPIUtil.getDeviceZoom() / 100f;
-		Cairo.cairo_surface_set_device_scale(surface, scaleFactor, scaleFactor);
-	}
+	double scaleFactor = DPIUtil.getDeviceZoom() / 100f;
+	Cairo.cairo_surface_set_device_scale(surface, scaleFactor, scaleFactor);
```
akoch-yatta: Is setting the scale still necessary? I don't get what the purpose is when the surface already has the correct size.

HeikoKlare: Yes, it is necessary. Otherwise it will look like this:
image

In my understanding, you need to do two things (which are the two places I adapted):
1. Give the correct size in pixels to the surface; by now we used the size in points, which led to have the expected size at 200% zoom
2. Tell Cairo at what scale that surface is used to ensure that coordinates are properly transformed; this is basically the same we have to do in SWT with point/pixel conversion. If you leave that out, you have a surface of double the size (when device zoom is 200) but since you treat all coordinates as 100%, it will paint at half scale to the top-left area of the surface.

arunjose696: I am still not able to understand a slight detail on how the images were drawn fully (although blurrily scaled) before, as the height and width of the surface was in points I would expect to the surface to have only data in half width and height or the image to be partially cut off, when the zoom is 200%

HeikoKlare: That's because the surface scale was usually set to 1 as `currentDeviceZoom` and `DPIUtil.getDeviceZoom()` were equal. I don't know why the check was there at all.

arunjose696:
I tried running with a snippet to draw a image and its copy. I can see a copied image would be cut off to only draw the top left quarter with the current state in master, which is also fixed by changes in this PR. Could also use this snippet to test the changes here. 

```java


import org.eclipse.swt.*;

public class ImageDrawExample {

        shell.setText("Image Enabled/Disabled Drawing");
        shell.setSize(600, 600);

        Image enabledImage = new Image(display, "src/org/eclipse/swt/snippets/eclipse.svg");
        Image copyImage = new Image(display, enabledImage, SWT.IMAGE_COPY);

        shell.addListener(SWT.Paint, e -> {
            e.gc.drawText("Enabled Image:", 10, 10);
            e.gc.drawImage(enabledImage, 10, 30);

            e.gc.drawText("Disabled Image:", 300, 10);
            e.gc.drawImage(copyImage, 300, 30);
        });

        shell.open();
        while (!shell.isDisposed()) {
        }

        enabledImage.dispose();
        copyImage.dispose();
        display.dispose();
    }
}
```


I have also tested the changes with the runtime workspace and could observe the original mentioned issue is resolved.

akurtakov:
I used the ImageDrawExample and was really surprised by the fact that on Gtk 4 the drawing only top left of image is a problem on Wayland but not on x11. With the patch applied it's correct on both wayland and x11.  All of that on HiDPI screen. 
On Gtk 3 if using GDK_SCALE=1 env variable the image is correct under X11. 
With all of the above I believe that issue is only on HiDPI screens and the current drawing is not adjusted for the fact that drawing on wayland uses "logical" units while on x11 its directly using pixels. I'll look into the PR in more details.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Image.java
```diff
@@ -278,16 +278,16 @@ public Image(Device device, Image srcImage, int flag) {
...
-	surface = Cairo.cairo_image_surface_create(format, width, height);
+	int dataWidth = DPIUtil.pointToPixel(this.width, DPIUtil.getDeviceZoom());
+	int dataHeight= DPIUtil.pointToPixel(this.height, DPIUtil.getDeviceZoom());
+	surface = Cairo.cairo_image_surface_create(format, dataWidth, dataHeight);
 	if (surface == 0) SWT.error(SWT.ERROR_NO_HANDLES);
-	if (DPIUtil.getDeviceZoom() != currentDeviceZoom) {
-		double scaleFactor = DPIUtil.getDeviceZoom() / 100f;
-		Cairo.cairo_surface_set_device_scale(surface, scaleFactor, scaleFactor);
-	}
+	double scaleFactor = DPIUtil.getDeviceZoom() / 100f;
```
akurtakov:
```suggestion
	double[] scaleX = new double[1];
	double[] scaleY = new double[1];
	Cairo.cairo_surface_get_device_scale(imageSurface, scaleX, scaleY);
```

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Image.java
akurtakov:
```suggestion
	Cairo.cairo_surface_set_device_scale(surface, scaleX[0], scaleY[0]);
```

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Image.java
akurtakov: As this is copy of the image it's best to fetch the scale (cairo gives separate x/y scale factors) of the image copied.

HeikoKlare: Good catch! Totally makes sense. Adapted with ece3113.

HeikoKlare:
> I used the ImageDrawExample and was really surprised by the fact that on Gtk 4 the drawing only top left of image is a problem on Wayland but not on x11. With the patch applied it's correct on both wayland and x11. All of that on HiDPI screen. On Gtk 3 if using GDK_SCALE=1 env variable the image is correct under X11. With all of the above I believe that issue is only on HiDPI screens and the current drawing is not adjusted for the fact that drawing on wayland uses "logical" units while on x11 its directly using pixels. I'll look into the PR in more details.

@akurtakov Thank you for the great analysis and for having a look into the PR. I was not aware about those different semantics of points under Wayland and X11. If I understand correctly, the behavior will now be fine on every of the environments you mentioned, right?

I implemented the proposed adaptations, so is this ready to go from your side?

akurtakov:
I believe it's good to go.

Issue #2641, opened 2025-10-17
[Win32] Parts of controls are cut off


HeikoKlare:
**Describe the bug**
When using non-100%-monitor zoom, the frames around part stacks are sometimes cut off. For example, at 150% zoom (same happens on 125%), when resizing the left part stack (including the Package explorer), the right frame border is sometimes cut off: 
<img width="786" height="369" alt="Image" src="https://github.com/user-attachments/assets/b1575b8e-3047-4f5f-b040-600aa5842d43" />

<img width="829" height="351" alt="Image" src="https://github.com/user-attachments/assets/ab801787-1ee4-4478-86e9-565313ca5970" />

**To Reproduce**
Open an SDK at 125% or 150% monitor zoom and horizontally resize the part stacks.

**Expected behavior**
Nothing (in particular not the frames) should be cut off.

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 

**Version since**
Must be presented since end of September since it's caused by:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2551

merks:
I'm not sure if it's the same regression, but if you launch a runtime workspace all the preference pages are empty:

<img width="640" height="468" alt="Image" src="https://github.com/user-attachments/assets/aaa2a050-9de2-4878-b835-26cac8118050" />

I think it starts with this commit:

<img width="1732" height="302" alt="Image" src="https://github.com/user-attachments/assets/a43dbf23-f256-47b1-b9ec-e8ec3f1af25c" />

HeikoKlare:
> I'm not sure if it's the same regression, but if you launch a runtime workspace all the preference pages are empty:

That's a different and more recent issue. Didn't see that yet. Thanks for reporting!
It's a regression from:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2636

PR #2654, opened 2025-10-21
[Win32] Fix cut-off controls #2641

HeikoKlare:
In HiDPI scenarios with fractional monitor scales (such as 125% and 175%), some controls appear slightly cut off. The reason for this is inconsistent rounding of control sizes. The rounding is necessary during pixel/point conversion and leads to added or lost single pixels. One controls which obviously shows the behavior is the border of CTabFolders.

This change adapts the calculation in the following way:
- It ensures that pixelToPoint and pointToPixel conversions for sizes round in opposite ways instead of rounding up in both directions as currently done. This ensures that sizes being set (via pointToPixel) are rounded up while sizes being retrieved (via pixelToPoint) are rounded down.
- It extracts the pixelToPoint conversion for sizes that are calculated based on pixel values and are supposed to represent a conservative size (i.e., better be to high than to low).
- It ensures that also the pixelToPoint conversion of rectangles takes the proper rounding method for sizes into account.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2641

### Before
cutoff_broken

### After 
cutoff_fixed

Note that the Manifest editor is shown because you can also see improper changes of the scaling when resizing that editor based on the Forms UI, in particular with cut off lines when resizing vertically.

### DIFF
--- a/tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/tests/win32/Win32DPIUtilTests.java
+++ b/tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/tests/win32/Win32DPIUtilTests.java
@@ -103,7 +103,7 @@ public class Win32DPIUtilTests {
 	@Test
 	public void scaleDownRectangle() {
 		Rectangle valueAt200 = new Rectangle(100, 150, 10, 14);
-		Rectangle valueAt150 = new Rectangle(75, 113, 7, 10);
+		Rectangle valueAt150 = new Rectangle(75, 113, 8, 10);
 		Rectangle valueAt100 = new Rectangle(50, 75, 5, 7);
 
 		Rectangle scaledValue = Win32DPIUtils.pixelToPoint(valueAt200, 200);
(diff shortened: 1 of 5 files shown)


tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/tests/win32/Win32DPIUtilTests.java
```diff
@@ -103,7 +103,7 @@ public void scaleDownPoint() {
 	@Test
 	public void scaleDownRectangle() {
 		Rectangle valueAt200 = new Rectangle(100, 150, 10, 14);
-		Rectangle valueAt150 = new Rectangle(75, 113, 7, 10);
+		Rectangle valueAt150 = new Rectangle(75, 113, 8, 10);
```
HeikoKlare, line 106: Note that this aligns the test values with the opposite `scaleUpRectangle()` method, which also makes the behavior appear more consistent again.

akoch-yatta:
With some further testing I noticed some issues regarding jface forms, but they appear in the same way on master. Therefor we need to investigate that separately and is not a blocker for this PR

PR #2656, opened 2025-10-22
Add support loading images with a size hint

laeubi:
Currently when loading a SVG image it always uses the size declared in the document but storing an SVG at the size to load the image is not really sufficient as it for example require to store the same SVG multiple times if one wants to load it at different sizes.

This now adds a new parameter to the NativeImageLoader that allows to pass a sizeHintSupplier that when we get a dynamic image will load it at the supplied size. This can then later be used in JFace to retrive the target size from the URL.

@HeikoKlare can you take a look if it makes sense to you? I'm not completely sure about `fileZoom` and `targetZoom` here...

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/image/FileFormat.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/image/FileFormat.java
@@ -116,6 +116,9 @@ public abstract class FileFormat {
  * device independent image array represented by the stream.
  */
 public List<ElementAtZoom<ImageData>> loadFromStream(LEDataInputStream stream, int fileZoom, int targetZoom) {
+	return loadFromStream(stream, fileZoom, targetZoom, null);
+}
+public List<ElementAtZoom<ImageData>> loadFromStream(LEDataInputStream stream, int fileZoom, int targetZoom, Supplier<Point> sizeHintSupplier) {
 	try {
 		inputStream = stream;
 		return loadFromByteStream(fileZoom, targetZoom);
@@ -148,13 +151,16 @@ public ImageData loadFromStreamBySize(LEDataInputStream stream, int width, int h
  * return the device independent image array represented by the stream.
  */
 public static List<ElementAtZoom<ImageData>> load(ElementAtZoom<InputStream> is, ImageLoader loader, int targetZoom) {
+	return load(is, loader, targetZoom, null);
+}
+public static List<ElementAtZoom<ImageData>> load(ElementAtZoom<InputStream> is, ImageLoader loader, int targetZoom, Supplier<Point> sizeHintSupplier) {
 	LEDataInputStream stream = new LEDataInputStream(is.element());
 	FileFormat fileFormat = determineFileFormat(stream).orElseGet(() -> {
 		SWT.error(SWT.ERROR_UNSUPPORTED_FORMAT);
 		return null;
 	});
 	fileFormat.loader = loader;
-	return fileFormat.loadFromStream(stream, is.zoom(), targetZoom);
+	return fileFormat.loadFromStream(stream, is.zoom(), targetZoom, sizeHintSupplier);
 }
 
 public static ImageData load(InputStream is, ImageLoader loader, int width, int height) {
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/image/SVGFileFormat.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/image/SVGFileFormat.java
@@ -15,6 +15,7 @@ package org.eclipse.swt.internal.image;
 import java.io.*;
 import java.nio.charset.*;
 import java.util.*;
+import java.util.function.*;
 
 import org.eclipse.swt.*;
 import org.eclipse.swt.graphics.*;
@@ -50,6 +51,29 @@ public class SVGFileFormat extends FileFormat {
 		}
 		if (targetZoom <= 0) {
 			SWT.error(SWT.ERROR_INVALID_ARGUMENT, null, " [Cannot rasterize SVG for zoom <= 0]");
+		}
+		ImageData rasterizedImageData = RASTERIZER.rasterizeSVG(inputStream, 100 * targetZoom / fileZoom);
+		return List.of(new ElementAtZoom<>(rasterizedImageData, targetZoom));
+	}
+
+	@Override
+	public List<ElementAtZoom<ImageData>> loadFromStream(LEDataInputStream stream, int fileZoom, int targetZoom,
+			Supplier<Point> sizeHintSupplier) {
+		if (RASTERIZER == null) {
+			SWT.error(SWT.ERROR_UNSUPPORTED_FORMAT, null, " [No SVG rasterizer found]");
+		}
+		if (targetZoom <= 0) {
+			SWT.error(SWT.ERROR_INVALID_ARGUMENT, null, " [Cannot rasterize SVG for zoom <= 0]");
+		}
+		if (sizeHintSupplier != null) {
+			Point point = sizeHintSupplier.get();
+			if (point != null) {
+				double factor = targetZoom / 100d;
+				int w = (int) (factor * point.x);
+				int h = (int) (factor * point.y);
+				ImageData imageData = RASTERIZER.rasterizeSVG(inputStream, w, h);
+				return List.of(new ElementAtZoom<>(imageData, targetZoom));
+			}
 		}
 		ImageData rasterizedImageData = RASTERIZER.rasterizeSVG(inputStream, 100 * targetZoom / fileZoom);
 		return List.of(new ElementAtZoom<>(rasterizedImageData, targetZoom));
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/internal/NativeImageLoader.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/internal/NativeImageLoader.java
@@ -18,6 +18,7 @@ package org.eclipse.swt.internal;
 import java.io.*;
 import java.util.*;
 import java.util.List;
+import java.util.function.*;
 
 import org.eclipse.swt.*;
 import org.eclipse.swt.graphics.*;
@@ -34,6 +35,10 @@ public class NativeImageLoader {
 	// --- loading ---
 
 	public static List<ElementAtZoom<ImageData>> load(ElementAtZoom<InputStream> streamAtZoom, ImageLoader imageLoader, int targetZoom) {
+		return load(streamAtZoom, imageLoader, targetZoom, null);
+	}
+
+	public static List<ElementAtZoom<ImageData>> load(ElementAtZoom<InputStream> streamAtZoom, ImageLoader imageLoader, int targetZoom, Supplier<Point> sizeHintSupplier) {
 		// 1) Load InputStream into byte array
 		byte[] data_buffer;
 		InputStream stream = streamAtZoom.element();
@@ -53,7 +58,7 @@ public class NativeImageLoader {
 			} catch (IOException e) {
 				SWT.error(SWT.ERROR_IO);
 			}
-			return FileFormat.load(new ElementAtZoom<>(stream2, streamAtZoom.zoom()), imageLoader, targetZoom);
+			return FileFormat.load(new ElementAtZoom<>(stream2, streamAtZoom.zoom()), imageLoader, targetZoom, sizeHintSupplier);
 		}
 		List<ImageData> imgDataList = new ArrayList<>();
 		long loader = GDK.gdk_pixbuf_loader_new();


HeikoKlare (commented):
Great to see further interest in adopting sophisticated image loading/SVG features.

But I have to admit that I do not understand the necessity for this change. In particular, the first sentence of the PR description is not true:
> Currently when loading a SVG image it always uses the size declared in the document but storing an SVG at the size to load the image is not really sufficient as it for example require to store the same SVG multiple times if one wants to load it at different sizes.

There is already `SVGFileFormat#loadFromByteStreamBySize()` and `NativeImageLodaer#load()` accepting a requested size, so it's not clear to me why we need yet other methods that combine zoom and size information. Currently, they are properly separated as I either want to load an image at it's original size (maybe in a zoomed version) or I want to load it at a specific size. In case fallbacks from one to the other are needed, this can (and even is already now) implemented in the `Image` class (for the case you cannot load/rasterize an image at a specified size).

In my opinion, this PR needs a demonstration of an adoption of the changes to understand which use cases shall be supported. That would allow to assess whether the current proposal fits to those demands or how a fitting solution would need to look like.

laeubi:
As said the caller can not know if this is a fixed or dynamic sized image, I have pushed the corresponding Jface changes now here:

- https://github.com/eclipse-platform/eclipse.platform.ui/pull/3431
the other important part is that almost all icons / images in a usual application are loaded through URLs and ImageDescriptor and at many places (e.g. plugin.xml and e4xmi) you can't even control how it is loaded at all.

HeikoKlare:
Thank you for sharing the JFace change. That helps to better understand the use case.

> As said the caller can not know if this is a fixed or dynamic sized image,

We already have `FileFormat#isDynamicallySizable()` just like the `FileFormat#canLoadAtZoom()` which is already used by the `URLImageDescriptor`. If we want to move the responsibility for identifying the proper way to load some data at the best possible size to the FileFormat/ImageLoader, we should do that consistently, also for the existing functionality. But I am not sure if that would be the right place to do it, as it would put pretty much responsibility for different ways of loading images to the FileFormat itself and on SWT side currently all such functionality is wrapped into ImageLoader and/or Image.

No matter which way we go here, is it really intended that this now combines zoom and size information and falls back to zoom if the image is a raster graphics? E.g., when you request an image at 64x64 at 200% zoom, you want it to be loaded at 128x128 pixel if the image is an SVG and you want it at 200% of whatever the original size is if it's a raster image? At least that's how I understand the current implementation.

But maybe I don't get the full picture yet. I am currently on mobile and will try to have another look next week when I am back in the office.

laeubi:
@HeikoKlare I mostly tried to get it working as I don't understand the full zoom handling here that somehow uses two zoom factors.

The  bottom line is: one somehow needs to "hint" the SVG loader that even the SVG document uses e.g. 1024x1024 pixel as its "native" size it should be rendered as a 16x16px image and by how we use images I can't use any (mostly internal) images and of course zooming should work as well. So yes it is intentional that it does not applies to raster graphics (what usually it is good to store them at the target size to safe space and have maybe a low resolution variant.

So given I have a hint of 16x16px and a current zoom of 200% the result should be an image of 32x32 pixel as if I would have been using two static images that are 16x16px and a `@2x` variant of that thing that has 32x32 pixels.

HeikoKlare:
I am still concerned that the proposal mixes up too many things, in particular in such a basic thing as FileFormat/ImageLoader. Here are some thoughts:
- In case I want to tell the image loading that I would like to load/rasterize data at a specific size, why to I need to additionally pass a zoom? I can directly incorporate the zoom into the requested size. I currently do not see why this cannot simply be done in JFace.
- Since raster graphics cannot be loaded at a specified size, you need to make a distinction based on whether you have a dynamically sizable image or not anyway. E.g., the request for `bundle:/example.id/icons/obj16/search.png?size=128x128` will usually fail while for SVGs you can pass whatever size you want. Currently, the size parameter would be ignored for raster images, but that will lead to unexpected results, For the above URl, I will probably receive a 16x16 pixels image, which is probably not what I as a consumer want and expect.
- In any case, mixing file zoom and target size does not make sense, as the file toom specifies which zoom the file represents, but that is kind of overwritten by the target size anyway. Lets say you use 200% file zoom, 150% target zoom and a target size of 48x48px, what would you expect? You could say that the file zoom does not matter, so you return 150% of 48x48px, i.e., 72x72px. Here the file zoom would be obsolete. So you could also say that you treat the file at 200% zoom, meaning you consider the 200% file as 48x48 px, which means you want to have 150% of the according 100% version (24x24px), so the result should be at 36x36px. But that would be completely confusion. So in my opinion file zoom and target size should never be used together.
- The concept of zoom is a bit weird anyway or at least becomes more weird when using SVGs: currently, images define the size in which they will be used. It may just be altered based on a zoom. However, it would in many cases conceptually make more sense if an application says "I want to use this image at X by Y pixels" (e.g., a toolbar icon at 16x16 for 100% screens and maybe scaled by some value that could depend on UI scaling or others) and not even encode a size in an image at all. That becomes possible with SVGs. It would make the application decide what image size it wants (in terms of size including scaling) and relieve this UI-specific concepts from image handling. But this implicit size encoding in images is so deeply embedded (e.g., that icons are mostly 16x16) that it would be difficult to get rid of it. Still, this is a good indicator why mixing zooms into size-based image retrieval does not really make sense to me.

Even if we agreed that the current proposal implements expected behavior, I am still convinced that this should not and does not need to be embedded into FileFormat/ImageLoader. For a request of an image with specified size (and zoom), you can check via FileFormat if the image dynamically sizable and in that case just request the image data at the size (altered by zoom). If the image is not dynamically sizable, you can still just request the image for the zoom (which, as said above, does not sound reasonable to me) or deal with the situation in a different way (such as throwin in error). In any case, this should be completely possible on consumer side (i.e., in JFace).

laeubi:
> In case I want to tell the image loading that I would like to load/rasterize data at a specific size, why to I need to additionally pass a zoom? I can directly incorporate the zoom into the requested size. I currently do not see why this cannot simply be done in JFace.

Because then JFace would need to know about zoom changes. I would ask this the other way round: Why does (currently) the ImageFormat at all must know the zoom at all if it all can be done by the caller?

In anyway I would be happy if it would be the case, but wanted to note what was written [here](https://github.com/eclipse-platform/eclipse.platform.swt/pull/2803#issuecomment-3569563695), that actually what one want is to override the _native size_ (that is unknown to the caller of the function!) if possible, so everything should just "work as before", so if before a zoom level would be required I don't see why it should now be obsolete because then it would be obsolete in the first place.

> Since raster graphics cannot be loaded at a specified size

That's why I explicitly made this as a *hint* to the provider. A classical raster image would always be loaded at its native size and I don't want to change this (by intention) but for example an ICO format or XPM both support different sizes in the same file as well so possibly could be using the hint as well even though not really broadly used in SWT anyways.

> in any case, mixing file zoom and target size does not make sense

As said above, I don't aim at a "target size" at all that why I don't think the feature to raster an image at a given size is suitable. I just want it to tell (a hint) about the _initial size_ at 100% so it can scale this on whatever demands for a zoomed image is given there as if the author of a SVG would have shrink the image at the first place.

> The concept of zoom is a bit weird anyway or at least becomes more weird when using SVGs

That's for sure, but it was designed that way before and I dindt aimed to change this in general

> For a request of an image with specified size (and zoom), you can check via FileFormat if the image dynamically sizable

The main problem is that even if I could I would completely need to redo what SWT does (and I think not everything is public API) because I literally can't know:

1. What image format has to be used
2. What size the image has

unless I try to load it, so the whole code of content-type detection and image file format selection has to be duplicated at JFace.

> In any case, this should be completely possible on consumer side (i.e., in JFace).

That's why I created the PRs so we can play around with that, I think the general idea (extract the size from the URL) is there so if you can came up with a JFace only solution that would be great and maybe help in further refine the proposal either in one way or the other.

HeikoKlare:
> The main problem is that even if I could I would completely need to redo what SWT does (and I think not everything is public API) because I literally can't know:
> 
> 1. What image format has to be used
> 2. What size the image has
> 
> unless I try to load it, so the whole code of content-type detection and image file format selection has to be duplicated at JFace.

Maybe I still do not fully understand the goal yet. Wouldn't something like this work as a solution to https://github.com/eclipse-platform/eclipse.platform.ui/pull/3431? (sure, that's not clean yet ...)
```java
	private static ImageData loadImageFromStream(InputStream stream, int fileZoom, int targetZoom,
			Supplier<Point> hintProvider) {
		Point hintSize = hintProvider.get();
		if (hintSize != null) {
			byte[] storedStream;
			try {
				storedStream = stream.readAllBytes();
			} catch (IOException e) {
				throw new RuntimeException(e);
			}
			if (FileFormat.isDynamicallySizableFormat(new ByteArrayInputStream(storedStream))) {
				return NativeImageLoader.load(new ByteArrayInputStream(storedStream), new ImageLoader(),
						hintSize.x * targetZoom / fileZoom, hintSize.y * targetZoom / fileZoom);
			}
		}
		return NativeImageLoader.load(new ElementAtZoom<>(stream, fileZoom), new ImageLoader(), targetZoom).get(0)
				.element();
	}
```

> Because then JFace would need to know about zoom changes. I would ask this the other way round: Why does (currently) the ImageFormat at all must know the zoom at all if it all can be done by the caller?

The JFace implementations for the image stuff knows about zooms already now and has reimplemented parts of what SWT already does.
The `FileFormat` knows about zooms to support SVGs, as for backward compatibility you need to incorporate the "native"/default size of the image (as usually the image and not the consumer specifies the default size) and apply a zoom on top of it. For SVGs you cannot do that at consumer side because the SVG is already rasterized then, so you need to pass the information to the rasterizer.
In the discussion on this PR, we are in a situation where we do not care about the "native" size of the SVG at all, but we want to tell the loader that we effectively want to have an image a specified target size (hint + zoom), which is something you can completely do at consumer side and do not need to provide further lower level API for it, at least as far as I understand it so far.

I know you say:
> As said above, I don't aim at a "target size" at all that why I don't think the feature to raster an image at a given size is suitable

But when you pass the requested default size and the zoom to an image to be rasterized, how is that different from specifying a target size (which by definition is `defaultsize * zoom`)?

> That's why I explicitly made this as a hint to the provider. A classical raster image would always be loaded at its native size and I don't want to change this (by intention) [...]

As said in my previous comment, that's something that raises concerns as a potential API user. You expect the user to know that loading `bundle:/example.id/icons/obj16/search.png?size=128x128` will usually give you a 16x16px image and `bundle:/example.id/icons/obj16/search.svg?size=128x128` will give you a 128x128px one (both at 100% zoom). I actually don't think that we can expect consumers to be aware of which exact file formats will consider that size hint and which will not. I, as a consumer, would expect that no matter what format the file has, when passing such a size "hint" means that I will get the image at that exact size (at 100% zoom).
Maybe there are good reasons for the proposed behavior which I currently do not understand yet. In that case, the above implementation might be a potential solution to achieve it.

laeubi:
> Maybe I still do not fully understand the goal yet. Wouldn't something like this work as a solution to https://github.com/eclipse-platform/eclipse.platform.ui/pull/3431? (sure, that's not clean yet ...)

You are right that JFace uses already internal API, I have not noticed this yet and this is actually worse enough, as JFace imports SWT with `org.eclipse.swt;bundle-version="[<some lower bound>,4.0.0)"` so any change here will break all prior jface version possible! Making more use of internal API does not really improve the situation.

Beside that this has some SWT implementation inherent problems:

- `org.eclipse.swt.internal.image.FileFormat.isDynamicallySizableFormat(InputStream)` iterates over the file formats and only returns a boolean, so we can not reuse that result.
- Next then we call `FileFormat.load(streamAtZoom, imageLoader, width, height);` effectively (one might ask why we have `org.eclipse.swt.internal.NativeImageLoader.load(InputStream, ImageLoader, int, int)` at all if its only a delegate...) what the *again* determines the file format
- `SVGFileFormat` then finally do what we want (as a side note - `StaticImageFileFormat` seems to just returns the image data as is (so actually ignoring the size) breaking the contract).

Given that quite convoluted process it seems to code against a very specific internal implementation details, what make the situation even worse. That also was the reason why I initially though we should hand it down to SWT, so the loader can make sure to take a good decision instead of trying to do the same check (again) multiple times on multiple levels.

> As said in my previous comment, that's something that raises concerns as a potential API user. You expect the user to know that loading bundle:/example.id/icons/obj16/search.png?size=128x128 will usually give you a 16x16px image and bundle:/example.id/icons/obj16/search.svg?size=128x128 will give you a 128x128px one (both at 100% zoom). 

I must confess I don't understand the example, as no one will use it that way. A more consistent example would be I have
`bundle:/example.id/icons/16x16/search.png` will usually give you a 16x16px image and now I replace it with an SVG (that is maybe designed at 128x128px or whatever size the UI Designer finds useful) so I would replace it with `bundle:/example.id/icons/16x16/search.svg`  and expect it to be an image with 16x16 pixels as before (or 32x32 with 200% zoom or on retina displays or...).

The query parameter is more for the case that I previously have `bundle:/example.id/icons/16x16/search.png` and `bundle:/example.id/icons/128x128/search.png` and I would now replace it with `bundle:/example.id/icons/search.svg?size=16x16` and `bundle:/example.id/icons/search.svg?size=128x128`

>  I actually don't think that we can expect consumers to be aware of which exact file formats will consider that size hint and which will not

We currently (and before) have expected that users understand they can save their files with `@2x` and `@1.5x` so I don't think this is much of a difference here.

HeikoKlare:
It's true that having JFace uses SWT (internals) is not a good solution. I think the intention back then was to finally provide appropriate API in SWT once the relevant functionality for supporting SVGs and newly arising use cases are clear. The version range issue is something that was probably not taken into account. The essential problem is that it would be best if SWT would/could properly encapsulate image loading capabilities and just provide an appropriate API for image access (via the `Image` class), but JFace replicates and/or extends quite some stuff, requiring access to internals (or things you would rather like to consider internal).

> The query parameter is more for the case that I previously have `bundle:/example.id/icons/16x16/search.png` and `bundle:/example.id/icons/128x128/search.png` and I would now replace it with `bundle:/example.id/icons/search.svg?size=16x16` and `bundle:/example.id/icons/search.svg?size=128x128`

I understand that use case, but still `bundle:/example.id/icons/search.png?size=128x128` will give you a differently sized image than  `bundle:/example.id/icons/search.svg?size=128x128`, requiring the consumer to explicitly know which kinds of file formats will take the size into account and which will not.
To that you say:

> I must confess I don't understand the example, as no one will use it that way. 

Why do you think so? Why should no one want to use `bundle:/example.id/icons/search.png?size=128x128` to retrieve an image at that specified size or at least expect that it works like that? That's exactly what I as a consumer would expect that query parameter to do.

> We currently (and before) have expected that users understand they can save their files with @2x and @1.5x so I don't think this is much of a difference here.

The difference is that the `@2x` and `@1.5x` extensions apply to _every_ kind of image file. You can even use them with SVGs. As a consumer, you just need to know that those conventions exist and can apply them without thinking about if you have a file format that supports it or not. That's different when you define that `?sizeXxY` will lead to different results depending on what kind of file you pass to it.

laeubi:
> I understand that use case, but still `bundle:/example.id/icons/search.png?size=128x128` will give you a differently sized image than `bundle:/example.id/icons/search.svg?size=128x128`, requiring the consumer to explicitly know which kinds of file formats will take the size into account and which will not.

If this ever will be not a (hypothetical) problem, then SWT could still return a scaled up version of that image. I just doubt it will ever be used that way.

> The difference is that the @2x and @1.5x extensions apply to every kind of image file. You can even use them with SVGs. 

But it in **no way guarantees** that I get the file at this zoom factor! I can use a 16x16px `search.png` and save a 1024x1024 png under the name `search@2x.png` is always is and was a convention and the users have to use it correctly, now trying to construct some kind of contract from that feels odd to me and don't really brings the topic forward. Especially as if I look at the implementation of `StaticImageFileFormat#load(InputStream, ImageLoader, int, int)` then we would need to enforce the same rules there (what seems not the case at all).

So for me this is a feature that builds a way for user make things _better_ then before by a convention without any risk. It does not prevent users from getting wrong results if they put in wrong data and I would exspect such thing to be noticed on the very first start of their application.

HeikoKlare:
> If this ever will be not a (hypothetical) problem, then SWT could still return a scaled up version of that image. I just doubt it will ever be used that way.

Not without fundamentally changing how it works: JFace bypasses the SWT `Image` implementation (which contains a bunch of useful stuff such as scaling capabilities). The `ImageLoader` and everything below is currently not capable of properly scaling images.

> But it in **no way guarantees** that I get the file at this zoom factor! I can use a 16x16px `search.png` and save a 1024x1024 png under the name `search@2x.png` is always is and was a convention and the users have to use it correctly, now trying to construct some kind of contract from that feels odd to me and don't really brings the topic forward. Especially as if I look at the implementation of `StaticImageFileFormat#load(InputStream, ImageLoader, int, int)` then we would need to enforce the same rules there (what seems not the case at all).

Yes, the user has to follow a contract (`@2x` must be twice the size as the default image) to make things work as expected. How is that different from any other contract? You loose me with the point of "constructing some kind of contract from that". Where do we contract a contract here?
Please note that `StaticImageFileFormat#load(InputStream, ImageLoader, int, int)` has a return type (`ElementAtZoom`) as well (by intent) to reflect the potentially missing capability of providing image data at requested zoom. I have already explained before why such a kind of capability in `FileFormat` is necessary to support SVGs. And please also note that this is internal API and not a general URL format that any consumer might use. So I really don't see why we need to discuss about such inappropriate comparisons here.

> It does not prevent users from getting wrong results if they put in wrong data and I would exspect such thing to be noticed on the very first start of their application.

Not sure what exactly you are referring to here, but at least the provision of correctly sized `@2x` images is something you should be able to (dynamically) evaluate with the `org.eclipse.swt.internal.enableStrictChecks` flag.

I have already given a proposal few comments above (https://github.com/eclipse-platform/eclipse.platform.swt/pull/2656#issuecomment-3626325763) how the desired functionality could probably be implemented at consumer side (without judging if that technically good or not). I don't see any response on whether that would be correct (to understand if we are talking about the same goals at least). Please also note that the code even reflects the proposed inconsistent handling of the `size` query parameter.

laeubi:
> Not without fundamentally changing how it works

That's why I designed this as "best effort", but you broaden the scope to something that uses might want to reuse that as a generic rescaling capability what was never the goal.

So yes given I would use it with something that is not scalable (be it be a restriction of JFace, SWT or the Image format) it would not work, and that's why it is a *hint* for the cases where I use it because it works.

> you loose me with the point of "constructing some kind of contract from that". Where do we contract a contract here?

You said as a user you would expect that non scalable images are scaled, so I assume you implied some kind of API/contract where no such thing exits or is intended.

> I have already given a proposal few comments above (https://github.com/eclipse-platform/eclipse.platform.swt/pull/2656#issuecomment-3626325763) how the desired functionality could probably be implemented at consumer side (without judging if that technically good or not). I don't see any response on whether that would be correct (to understand if we are talking about the same goals at least). 

I already wrote that this will probably work to the cost of doing the same checks two times. But I don't think it is godd because of the same reason you mentioned that is

> JFace bypasses the SWT Image implementation (which contains a bunch of useful stuff such as scaling capabilities)

So that's why I initially proposed to pass that information down to SWT so it can make use of any "useful stuff such as scaling capabilities" instead of bypass SWT even more, but the whole discussion now focus on the JFace part (that is how we get it from the URL), while initially you wanted the discussion to continue here (where it would not matter how the data is acquired at all).

So I'm not sure if you just don't like the feature at all, don't think its useful for SWT to leverage that information or have something completely different in mind.

HeikoKlare:
Let's take a step back before this gets more offensive and clarify some things.

> So I'm not sure if you just don't like the feature at all, don't think its useful for SWT to leverage that information or have something completely different in mind.

I don't have anything against that feature. Quite the contrary, it would of course be great to make more out of the capabilities coming in with SVGs. I even made concrete proposals for how to proceed, but the discussion seems to focus on something different. So to be clear: I do _not_ want to block this but would like to see this move forward instead.

But yes, I also raised concerns about the proposed API / query parameter and I agree that it took more room than it should here. So I would like to clarify that I still have those concerns but I would not block any progress because of them. I think what concerns me most about that is the attitude of effectively adding new API that only works for specific use cases that the user has to know about and not allowing to be the least bit concerned about that. We can still agree that it's fine to only support specific use cases (for now) and that we have to properly make users aware of that in some way, but why can't we properly discuss that with potential concrete proposals how to best deal with that and instead have to discuss inappropriate comparisons to argue that any concern is invalid? We had quite some discussions and revisions on the initial extensions for SVG support as well and threw away multiple proposals, and it was definitely worth it.

So to summarize my concrete concern on the API: When introducing a kind of public API that allows to specific a size for an image to load (via a URL query parameter), I am concerned about doing it in way in which, as a consumer, you need to know about the specific cases in which that will work properly, as those cases that are not supported are not obvious. In particular, a user cannot know that a query parameter called `size` does not request an image in the specified `size` but will rather apply that as a hint that will only have a specific effect on specific types of images (SVGs). You call vector graphics "scalable" (yes it's part of the abbreviation SVG) and raster graphics non-scalable, but I would not expect every user to think that a `size` query can thus only be applied to vector graphics as raster graphics are non scalable. Of couse the latter are also scalable (but in a different way / with worse quality).
That said, I have these concerns because I as a user was/would be confused by the behavior (which could also be just up to me, according to how as explained I think about such a `size` parameter). But it's nothing that makes me block any progress on the proposed feature.

The more relevant concern I have completely got out of focus and is about how the desired feature can be realized:

> That's why I designed this as "best effort", but you broaden the scope to something that uses might want to reuse that as a generic rescaling capability what was never the goal.

I actually made a concrete proposal to address the demand as "best effort" in https://github.com/eclipse-platform/eclipse.platform.swt/pull/2656#issuecomment-3626325763. And to be precise here as well: I did not propose to implement a generic rescaling capability, I just mentioned what users will expect when you allow them to request an image at a specific size.

I think you can even simply that proposal to:
```java
	private static ImageData loadImageFromStream(InputStream stream, int fileZoom, int targetZoom,
			Supplier<Point> hintProvider) {
		Point hintSize = hintProvider.get();
		if (hintSize != null) {
			return NativeImageLoader.load(stream, new ImageLoader(),
						hintSize.x * targetZoom / fileZoom, hintSize.y * targetZoom / fileZoom);
		}
		return NativeImageLoader.load(new ElementAtZoom<>(stream, fileZoom), new ImageLoader(), targetZoom).get(0)
				.element();
	}
```

To summarize some of my statements: the required functionality is nothing that, in my opinion, should be part of the `FileFormat`/`NativeImageLoader`. I have already explained in https://github.com/eclipse-platform/eclipse.platform.swt/pull/2656#issuecomment-3458134368 that I see two general use cases that should/need to be supported by such low-level API:
1. The image file itself defines the "default" size (common behavior since ever): on top of this, a zoom can be applied to make that image fit to a different UI zoom. This is something you would probably not realize in that way when designing a new API, but for the sake of compatibility we of course have to keep it. Applying the zoom on top is somehow difficult for SVGs (as it must be applied on loading the image and not as postprocessing such as for raster graphics), which makes the API a bit convoluted.
2. The application defines the desired image size: in that case, the application can incorporate the zoom as it knows about the size anyway.

Since the second scenario can completely be realized at consumer side, there should be no need to provide minimal low-level API for it. One might discuss what is considered "too low level" for it, i.e., at which level reusable/convenience API is appropriate (e.g., would it be fine to have that at `NativeImageLoader` level instead of `FileFormat`?). This also addresses the statement in the last comment about adding the functionality proposed by this PR to not bypass the according `Image` APIs, but actually the proposed functionality extends the bypass mechanisms, as it adds more internal API. So that statement is contradictory in itself.
The way how JFace accesses all the low-level functionality is currently not nice, so far we agree. But that is something we cannot resolve with introducing more internal API while just using the existing internal API would be sufficient. If one really wants to improve this, they would probably have to think about how to change the way in which the image descriptors work, such that they don't bypass the `Image` class and its capabilities. An option could be the use of an `Image` with `ImageDataAtSizeProvider`, but that's contrary to the current way in which image descriptors work and anyway obviously out of scope here.

> but the whole discussion now focus on the JFace part (that is how we get it from the URL), while initially you wanted the discussion to continue here (where it would not matter how the data is acquired at all).

The discussion is going on here because this PR is currently nothing to be used on its own but is basically just a prerequisites for the JFace PR. And having two discussions at different places on the same topic is not very helpful, so since we started here I wanted to avoid that we discuss the same things in the JFace PR again.

laeubi:
> Let's take a step back before this gets more offensive and clarify some things.

This was not meant in any way offensive, sorry if it feels like that.

> I think you can even simply that proposal to

I like that idea and have now [adjusted my PR](https://github.com/eclipse-platform/eclipse.platform.ui/pull/3431) to use that here, so I agree that as a first sight we do not need any new SWT (internal) API for that.

Sadly there is a bug in the Equinox URL parsing that currently will make some of the test fail, I'll look into this.

PR #2661, opened 2025-10-22
Add explicit activation (by tester) to Clipboard tests on Wayland

jonahgraham:
On Wayland apps are not allowed to take ownership of the system clipboard unless the Wayland compositer (e.g KWin, mutter or others) believes that the user has initiated the request for ownership.

This means that activating and focusing the Shell may not be sufficient.

Therefore we require the person running the tests to press a button in the shell so the compositer believes that the clipboard access is legitimate.

This commit adds such a button, and a short timeout to press it. If not pressed in time the tests requiring SWT to take ownership of the clipboard will be skipped.

The button must be pressed multiple times throughout the test, therefore we group the affected tests to run first with TestMethodOrder annotation.

On GitHub actions + Jenkins these tests are skipped by default.

Note: At the moment activating + focusing the shell is sufficient to read from the system clipboard. In addition, X11 (and therefore the Swing remote clipboard test app) run under Xwayland do not require such explicit activation, so the key presses are only on the couple of tests that require ownership of system clipboard to operate properly.

Part of #2126

### DIFF
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/RemoteClipboard.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/RemoteClipboard.java
@@ -205,4 +205,9 @@ public class RemoteClipboard implements ClipboardCommands {
 	public void waitUntilReady() throws RemoteException {
 		remote.waitUntilReady();
 	}
+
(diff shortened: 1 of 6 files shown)


jonahgraham:
Because AFAICT we don't test on Wayland yet, the impact of this is minimal so far. I plan to merge this soon if no objections/questions.

PR #2668, opened 2025-10-24
Spin event loop in widget tests that access Clipboard

jonahgraham:
This carries on earlier work, such as PR #2489

Part of #2598


jonahgraham:
The failing windows run is unrelated (documented in https://github.com/eclipse-platform/eclipse.platform.swt/issues/2670)

PR #2669, opened 2025-10-24
Document macOS URLTransfer uses out of date transfer type name

jonahgraham:
Part of #2126

Perhaps if a macOS maintainer wants to test this they can update to the new transfer type. I don't know if there is a negative effect of leaving this or moving it.

The full DND transfer tests that I am writing will make it easier to test and maintain that the transfer is implemented. Have a look at Test_org_eclipse_swt_dnd_URLTransfer (once that PR is merged)


jonahgraham:
The failing windows run is unrelated (documented in https://github.com/eclipse-platform/eclipse.platform.swt/issues/2670)

Issue #2666, opened 2025-10-24
[GTK] HTMLTransfer loses last byte if copied data is odd length


jonahgraham:
When Eclipse receives a copy of HTMLTransfer on GTK the last byte is lost if data is odd length.

To reproduce:

1. Launch ClipboardExample
2. Open webpage
3. Copy a string from the webpage
4. Paste it in the HTML area of ClipboardExample

Observe that last byte copied is missing. If it isn't missing, make the copy size one character longer and repeat 3 and 4.

Note that the length that must be odd for this to fail includes the html characters, so the selected text in the browser could be odd or even.

Here is a screenshot showing a selection of what I copied and the pasted result (`exte` copied, only `ext` pasted in SWT):

<img width="874" height="321" alt="Image" src="https://github.com/user-attachments/assets/4717ed73-037e-411e-a4a8-c87f68e98f20" />

Repeating by making selection one character longer (`exten` copied and properly pasted):

<img width="877" height="356" alt="Image" src="https://github.com/user-attachments/assets/79ec5048-3fb2-4843-a036-b63bf844d9fe" />

jonahgraham:
This bug is a side effect of a fix applied in Bug 376589 (c5308d73c1e46a1db4edf39db94dfada1b6e65dc) with follow-up in Bug 384381 (de6fbaaec46412d3fdef0c01810182c0b01cafbd)

Those bugs changed HTML copy/paste from UTF16 to UTF8, but left behind this bug. 

PR on the way.

PR #2678, opened 2025-10-28
[GTK] Fix HTMLTransfer loses last byte if copied data is odd length

jonahgraham:
As a follow up to Bugs 376589 & 384381 which changed the serialization deserializtion from UTF16 to UTF8, there was some remnant code that ensured input was handled in pairs of bytes that should have been moved into the UTF16 part of the new if statement.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2666

This is a GTK only fix that affects GTK3 and GTK4


akurtakov:
Very good catch. Thank you!

PR #2679, opened 2025-10-28
Modernize ByteArrayTransfer javadoc

jonahgraham:
Now that ByteArrayTransfer has a unit test, the unit test verifies that the code in this Javadoc operates correctly. This commit synchronizes the javadoc back to the contents of the test classes.

This includes

- Simplifying the code a little
- Changing the transfer to support a single MyType instead of an array of them (which simplifies the example quite a bit)
- Add `@Override` to the appropriate methods
- Format the code to Eclipse default settings
- Validating the objects

Part of https://github.com/eclipse-platform/eclipse.platform.swt/issues/2126


akurtakov:
Thanks Jonah!

PR #2682, opened 2025-10-28
Remove vintage engine from swt tests

akurtakov:
As all tests are JUnit 5 now let's not pollute with useless things.


akurtakov:
No change in executed tests in both GHAs and Jenkins.

PR #2685, opened 2025-10-28
Stop swallowing System.err in JUnit tests

jonahgraham:
test_printStackTrace overrides System.err with its own print stream to check functionality. But these tests did not restore System.err, so all System.err output for the rest of the test suite was lost into the last of these ByteArrayOutputStreams that was created.

The tests themselves have little intrinsic value as they really test nothing of SWT itself, but it does help achieve code coverage.

This was discovered when working on https://github.com/eclipse-platform/eclipse.platform.swt/pull/2684 as I could not see the output when running all the tests.

### DIFF
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/AllNonBrowserTests.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/AllNonBrowserTests.java
@@ -111,8 +111,7 @@ public class AllNonBrowserTests {
 		}
 
 		for (Error leak : leakedResources) {
-			// For some reason, printing to System.err in JUnit test has no effect
-			leak.printStackTrace (System.out);
+			leak.printStackTrace ();
 		}
 
 		if (0 != leakedResources.size ()) {
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/SwtTestUtil.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/SwtTestUtil.java
@@ -19,9 +19,11 @@ import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.Assertions.fail;
 
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.PrintStream;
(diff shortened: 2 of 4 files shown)


tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/AllNonBrowserTests.java
```diff
@@ -111,8 +111,7 @@ public static void afterClass() {
 		}
 
 		for (Error leak : leakedResources) {
-			// For some reason, printing to System.err in JUnit test has no effect
-			leak.printStackTrace (System.out);
+			leak.printStackTrace ();
```
jonahgraham, line 114: FYI @SyntevoAlex I found out where System.err was going.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/SwtTestUtil.java
```diff
@@ -764,4 +766,26 @@ static <T> T runOperationInThread(int timeoutMs, ExceptionalSupplier<T> supplier
...
+/**
+ * Capture any output on System.err
+ *
+ * This method does not capture output on stderr from C level, such as
+ * Gdk-CRITICAL messages.
+ *
+ * @param runnable to run while capturing output
+ * @return output on System.err
+ */
+public static String runWithCapturedStderr(Runnable runnable) {
```
mickaelistria, line 779: FWIW, I've recently relied on AutoCloseable a lot for testing; it allows some good way of wrapping state relying less on Runnable.

For this case, it could go that way
```java
class CaptureStderr implements AutoCloseable, Supplier<String> {
  ByteArrayOutputStream errContent
  CaptureStderr() {
     errContent = new ByteArrayOutputStream();
	 System.setErr(new PrintStream(errContent, true, StandardCharsets.UTF_8));
  }
  @Override
  public void close() {
    System.setErr(stderr);
    errContent.close();
  }
  @Override
  public String get() {
    return errContent.toString(StandardCharsets.UTF_8);
  }
}
```
and then tests can be written
```java
try (new captureStdErr = new CaptureStdErr()) {
   // do stuff
  assertTrue(captureStdErr.get().contains("blah");
}
```

jonahgraham: I have been meaning to come back to thank you for this advice - while I have not updated this now closed PR, I have used this methodology in my latest testing PR: #2825

PR #2686, opened 2025-10-28
Fix test leak detection in JUnit tests

jonahgraham:
While working on https://github.com/eclipse-platform/eclipse.platform.swt/pull/2685 and specifically trying to verify that change fixes the comment below \* I realized that a migration step was missed changing the suites from JUnit4 to JUnit5.

On JUnit5 the `@BeforeAll` and `@AfterAll` only apply to the tests in the class itself, not to `@SelectClasses` too.

Instead JUnit5 provides `@BeforeSuite` and `@AfterSuite`

\* this comment:

> For some reason, printing to System.err in JUnit test has no effect


jonahgraham:
:cry: Unfortunately with leak detection not running for sometime some resources leaks have snuck into the testsuite. Positive side is that at least some the leaks are in the tests I just wrote for the clipboard.

I created PR https://github.com/eclipse-platform/eclipse.platform.swt/pull/2687 to verify that leak detection worked, but clearly it does again now!

PR #2689, opened 2025-10-28
[Win32] Remove superceded GC operations from history

HeikoKlare:
The GC stores every operation changing the GC state to be able to repply it if a handle for a different zoom is requested. There are some operations which, if executing an operation of the same type subsequently, supercede each other. This, for example, applies to setting fonts, as when settings two fonts after each other, only the last one set is relevant.

This change flags several GC operations as replaceable and implements the removal of previous operations from the operation history if a new operation is executed that supercedes the previous one. This slightly reduces the memory footprint.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
@@ -4799,7 +4799,7 @@ public void setBackground (Color color) {
 	storeAndApplyOperationForExistingHandle(new SetBackgroundOperation(color));
 }
 
-private class SetBackgroundOperation extends Operation {
+private class SetBackgroundOperation extends ReplaceableOperation  {
 	private final Color color;
 
 	SetBackgroundOperation(Color color) {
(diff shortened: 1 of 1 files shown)


arunjose696:
@HeikoKlare Just to confirm: the current logic only removes a replaceable operation if it’s immediately preceding the new one.
So in a sequence like setFont > setLineWidth > setFont, the first setFont stays in the list even though it’s superseded. would it make sense to remove the older setfonts, too ?

HeikoKlare:
> Just to confirm: the current logic only removes a replaceable operation if it’s immediately preceding the new one.
> So in a sequence like setFont > setLineWidth > setFont, the first setFont stays in the list even though it’s superseded. would it make sense to remove the older setfonts, too ?

That would add another requirement to replaceable operations: they must be commutative. I don't know if that is currently the case and it's quite prone to errors as you have to thoroughly check that all such marker operations are completely independent (commutative). While we may be able to do that right now, I would be afraid that future changes could easily break that if developers are not properly aware of this requirement. Or we would need a whole bunch of tests to ensure it.

So to conclude: the idea is definitely good, but I am not sure if it's worth the effort and, in particular, the risk coming along with it.

PR #2691, opened 2025-10-29
Update README: Replace mailing list with GitHub issues/discussions (#4)

laeubi:
We still mention bugzilla / mailinglist in the readme... now updated to using github instead.


laeubi:
Doc change only...

Issue #2621, opened 2025-10-14
[macOS 26.1] Multiline selection highlighing is visually broken


danishmau:
**Describe the bug**
On Eclipse 2025-06 M3 (4.36.0 M3), the select all (command + A) is not working, even cursor selection also not working on any source code (java, docker, bash etc). The OS is MacOS Tahoe 26.1
**To Reproduce**
A standalone SWT snippet to reproduce the behavior (preferable)
(or) 
Steps to reproduce the behavior.
Open any file on Eclipse Java IDE, try to select all code using command + A or cursor select on MacOS Tahoe 26.1
**Expected behavior**
On using shortcut like command + A or Select All or Cursor selection, it should select code, currently on command + A, the cursor is going down, below is error we received on console log

-[TUINSCursorUIController activate:]: SWTCanvasView doesn't conform to NSTextInputClient protocol.

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [ ] Windows
- - [ ] Linux
- - [√ ] macOS 

2. Additional OS info (e.g. OS version, Linux Desktop, etc)
Tahoe 26.1
3. JRE/JDK version
 JDK 21.0
**Version since**
 Eclipse or SWT version since when the behavior is seen 
Version: 2025-06 M3 (4.36.0 M3)
Build id: 20250522-0641

**Workaround (or) Additional context**
No workaround as of now

HeikoKlare:
Does this issue still occur with latest I-Build? I cannot reproduce it with I20251008-1230 and I20251013-1800.

Are you referring to a MacOS 26.1 beta version? If I am not mistaken, the latest release version is 26.0.1, which is also the version on which I tested.

danishmau:
Hi @HeikoKlare ,
The MacOS version is Version 26.1 Beta (25B5062e), I have installed the developer beta. I have tested the same build in older MacOS (Sonoma) and it is working there.

HeikoKlare:
Thank you for the additional information. Then let's see if that issue will be fixed at MacOS side for the 26.1 release or whether the release version will have the same issue. Good to know in advance that there may be an issue we need to address.

Phillipus:
Might be worth filing an issue with Apple.

Phillipus:
macOS 26.1 beta 4 released today. Could you test this version?

danishmau:
Hi @Phillipus @HeikoKlare ,

[...]

PR #2694, opened 2025-10-29
[mac] SWT: fix for #2621

DenisUngemach:
From MacOS 26 - > MacOs 26.1 it seems the method:
NSBezierPath.appendBezierPathWithRect(NSRect rect) can't handle huge integer values anymore. So a smaller number is defined and used. Doesn't matter for the purposes here.

See: https://github.com/eclipse-platform/eclipse.platform.swt/issues/2621
(not a mistake, PR to the swt repo master branch.)

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/TextLayout.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/TextLayout.java
@@ -509,7 +509,7 @@ public void draw(GC gc, int x, int y, int selectionStart, int selectionEnd, Colo
 				NSRect bounds = lineBounds[lineBounds.length - 1];
 				rect.x = pt.x + bounds.x + bounds.width;
 				rect.y = y + bounds.y;
-				rect.width = (flags & SWT.FULL_SELECTION) != 0 ? 0x7fffffff : (bounds.height + spacing) / 3;
+				rect.width = (flags & SWT.FULL_SELECTION) != 0 ? OS.MAX_TEXT_CONTAINER_SIZE : (bounds.height + spacing) / 3;
 				rect.height = Math.max(bounds.height + spacing, ascent + descent);
 				path.appendBezierPathWithRect(rect);
 			}


Phillipus:
@DenisUngemach In your PR you replace the value `0x7fffffff` with `10000000`. I found that using `0xffffffff` also works. So maybe this has something to do with max value sizes in Apple's code?

DenisUngemach:
> @DenisUngemach In your PR you replace the value `0x7fffffff` with `10000000`. I found that using `0xffffffff` also works. So maybe this has something to do with max value sizes in Apple's code?

Yes there are some bigger numbers which also work, but this number just should be a number big enough in order to have no width limit. For that 10000000 is completely fine.

Phillipus:
Numbers that don't work: `0x3fffffff - 0xbfffffff` - must be something special about that.

Phillipus:
I did some searching and AI questioning and it seems that `0x7fffffff` is a "magic" number that denotes "infinite" width (`Integer.MAX_VALUE`). This value is also used here:

https://github.com/eclipse-platform/eclipse.platform.swt/blob/e9a8f5f9c57c67fe4c668573a8173202f7049ff1/bundles/org.eclipse.swt/Eclipse%20SWT/cocoa/org/eclipse/swt/graphics/TextLayout.java#L817-L819

And `0xffffffff` represents -1 a "magic negative value" so perhaps this is what Apple expects to see now?

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/TextLayout.java
```diff
@@ -38,6 +38,9 @@
  */
 public final class TextLayout extends Resource {
 
+    // Mac features might not handle big integer values (like 0x7fffffff = Integer.MAX_VALUE) since MacOS26.1.
+    	private final static int MAX_VALUE = 10000000;
```
sratz: Should be `0x...`?

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/TextLayout.java
```diff
@@ -38,6 +38,9 @@
  */
 public final class TextLayout extends Resource {
 
+    // Mac features might not handle big integer values (like 0x7fffffff = Integer.MAX_VALUE) since MacOS26.1.
```
sratz: Can you link to the github issue here?

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/TextLayout.java
sratz: `MAX_VALUE_<WHAT?>`.
Can we be more specific here? What we would like to express essentially is "unlimited", so something like

`VALUE_RECT_HUGE_EFFECTIVELY_UNLIMITED`

DenisUngemach:
On my system the values till 900.000.000 seem to work. 
So I can also raise the MAX_VALUE to another higher value. 

But it is very unexpected, that in eclipse width values higher than 10.000.000 will be used.

Phillipus:
I suspect that if `0x7fffffff` is a "magic" number that denotes "infinite" width (Integer.MAX_VALUE) then the new number should be based on that assumption, that's why I suggest `0xffffffff`

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/TextLayout.java
DenisUngemach: > Should be `0x...`?

Why?

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/TextLayout.java
```diff
@@ -509,7 +513,7 @@ public void draw(GC gc, int x, int y, int selectionStart, int selectionEnd, Colo
 				NSRect bounds = lineBounds[lineBounds.length - 1];
 				rect.x = pt.x + bounds.x + bounds.width;
 				rect.y = y + bounds.y;
-				rect.width = (flags & SWT.FULL_SELECTION) != 0 ? 0x7fffffff : (bounds.height + spacing) / 3;
+				rect.width = (flags & SWT.FULL_SELECTION) != 0 ? MAX_VALUE : (bounds.height + spacing) / 3;
```
sratz: Let's use `OS.MAX_TEXT_CONTAINER_SIZE`. Interesting JavaDoc on that constant as well :)

https://github.com/eclipse-platform/eclipse.platform.swt/blob/fe1ede80f5f5a19e4a47aeefeb181fae60a6a7d8/bundles/org.eclipse.swt/Eclipse%20SWT%20PI/cocoa/org/eclipse/swt/internal/cocoa/OS.java#L43-L54

Phillipus:
Has this been tested on earlier versions of macOS?

sratz:
> Has this been tested on earlier versions of macOS?

Yes, on 26.0.

Phillipus:
> Yes, on 26.0.

But not on macOS 13.x, 14.x and 15.x?

Phillipus:
> > Yes, on 26.0.
> 
> But not on macOS 13.x, 14.x and 15.x?

FWIW - I tested on 15.7.1 (24G231) and it seems to be OK.

BeckerWdf:
> But not on macOS 13.x, 14.x and 15.x?

I don't see the possibility to test with macOS 13.x and 14.x from SAP side because our IT department always forces us update the new OS-Versions (for good reasons). macOS 13 and 14 are long gone in here. Sorry.

Phillipus:
> I don't see the possibility to test with macOS 13.x and 14.x from SAP side because our IT department always forces us update the new OS-Versions (for good reasons). macOS 13 and 14 are long gone in here. Sorry.

I just tested on macOS 13.7.8 and it's working. I would guess 14.x will be OK too.

BeckerWdf:
> > I don't see the possibility to test with macOS 13.x and 14.x from SAP side because our IT department always forces us update the new OS-Versions (for good reasons). macOS 13 and 14 are long gone in here. Sorry.
> 
> I just tested on macOS 13.7.8 and it's working. I would guess 14.x will be OK too.

Great to hear. Thanks for your help.

BeckerWdf:
So should we merge this?

Phillipus:
> So should we merge this?

Well, it can't make the situation any worse than it is now. ;-)

+1

BeckerWdf:
Even though we usually don't provide don't provide "Dot 1" releases for already released Platform / SimRel versions: Should we anyhow cherry-pick this to the maintenance branch?

Phillipus:
> Even though we usually don't provide don't provide "Dot 1" releases for already released Platform / SimRel versions: Should we anyhow cherry-pick this to the maintenance branch?

Does anyone actually create a working Eclipse binary from a maintenance branch?

akurtakov:
Every commiter can backport a PR to maintenance branch . There is no build that will happen from this maintenance branch . AFAIK IBM do some such builds.

Phillipus:
Tested on Build id: I20251029-1800 and looking good. 🎉

vnuta9:
@Phillipus , @akurtakov, @IvanGorshechnikov
 
Dear Folks,
Essentially I am on Version: 2025-12 M2 (4.38.0 M2) Build id: 20251030-0646. It seems to have been messed up due to multiple trials. If I had to delete just the Eclipse app on my mac 26.1 Tahoe,  keeping the workspace as it is, **which version of latest Eclipse download has this fix built-in which I can access. URL...** ? It is a distraction and wasted few hours of productive time already, so want to get over with it PRONTO. Please advise !

PR #2695, opened 2025-10-29
Update gtk-dev-guide.md: Modernize references and add GTK4 documentaton

laeubi:
- Replace Bugzilla and wiki references with GitHub 
- Use local images instead of wiki.eclipse.org links
- mentioning GTK4 build


jonahgraham (commented):
Thanks @laeubi for these updates.

Issue #2667, opened 2025-10-24
[GTK4] Webkit crash on close


akurtakov:
Start ControlExample and close it after it loads.
It doesn't exit cleanly but with :
```
SWT call to Webkit timed out after 10000ms. No return value will be provided.
Possible reasons:
1) Problem: Your javascript needs more than 10000ms to execute.
   Solution: Don't run such javascript, it blocks Eclipse's UI. SWT currently allows such code to complete, but this error is thrown 
     and the return value of execute()/evalute() will be false/null.

2) However, if you believe that your application should execute as expected (in under10000 ms),
 then it might be a deadlock in SWT/Browser/webkit2 logic.
 I.e, it might be a bug in SWT (e.g this does not occur on Windows/Cocoa, but occurs on Linux). If you believe it to be a bug in SWT, then
Please report this issue *with steps to reproduce* via:
 https://github.com/eclipse-platform/eclipse.platform.swt/issues/new?assignees=&labels=&projects=&template=bug_report.md&title=
For bug report, please attach this stack trace:
java.lang.Throwable: 
	at org.eclipse.swt.browser.WebKit.getStackTrace(WebKit.java:305)
	at org.eclipse.swt.browser.WebKit.getInternalErrorMsg(WebKit.java:298)
	at org.eclipse.swt.browser.WebKit$Webkit2AsyncToSync.execAsyncAndWaitForReturn(WebKit.java:1321)
	at org.eclipse.swt.browser.WebKit$Webkit2AsyncToSync.runjavascript(WebKit.java:1063)
	at org.eclipse.swt.browser.WebKit$Webkit2AsyncToSync.evaluate(WebKit.java:1033)
	at org.eclipse.swt.browser.WebKit.evaluate(WebKit.java:1348)
	at org.eclipse.swt.browser.WebKit.close(WebKit.java:898)
	at org.eclipse.swt.browser.WebKit.onDispose(WebKit.java:1841)
	at org.eclipse.swt.browser.WebKit.lambda$5(WebKit.java:787)
	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:91)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:5889)
        ...
```

akurtakov:
[...]

PR #2706, opened 2025-10-30
[GTK4] Update to new API for running javascript

jonahgraham:
The API for running javascript and integrating with javascript objects and types has changed substantially between GTK3 and GTK4 (technically between webkit2gtk 4 and webkitgtk 6)

The top level methods to call to run javascript changed:

- webkit_web_view_run_javascript -> webkit_web_view_evaluate_javascript
- webkit_web_view_run_javascript_finish -> webkit_web_view_evaluate_javascript_finish
- JSValue methods -> JSCValue methods

With this Browser.execute works, so for example Snippet161 now works and the ControlExample shuts down cleanly.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2667


akurtakov:
Snippet161 doesn't work for me under Wayland
image but works under X11 so I assume this is another problem.
As ControlExample no longer crashes - it's worth merging now.

jonahgraham:
> but works under X11 so I assume this is another problem.

For me it is almost the opposite - it works on Wayland and x11 with Xwayland, but fails on x11 with Xtigervnc.

Can you try with https://github.com/eclipse-platform/eclipse.platform.swt/pull/2713 to see if it is something simple like some code path that hits old API?

akurtakov:
I will, I just have to finish some other things. When I say it works with X11 (it means XWayland) as Fedora 43 no longer has Gnome X11 session at all (https://fedoraproject.org/wiki/Changes/WaylandOnlyGNOME)

PR #2716, opened 2025-10-30
Set Shell Size and the end of DPIChange

amartya4256:
Currently, the shell is scaled synchronously regardless of the sync/async behaviour of scaling. With this PR the size of the shell is set at the end of a DPIChange execution to ensure the final size set to the shell consistent with the final zoom level. 

#### Why it is necessary:
In case of a DPIChange event or multiple events, the scaling of all the widgets can happen synchronously or asynchronously. In case of asynchronous execution, the shell's size might be set before all the widgets are scaled which can lead to a weird result as resizing a shell leads to the widgets to relayout. Calling a Shell:layout at the end has similar results. It is only consistent to also set the size of the shell at the end of the execution followed by a Shell::layout.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/DPITestUtil.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/DPITestUtil.java
@@ -28,7 +28,8 @@ public final class DPITestUtil {
 
 	public static void changeDPIZoom (Shell shell, int nativeZoom) {
 		DPIUtil.setDeviceZoom(nativeZoom);
-		Event event = shell.createZoomChangedEvent(nativeZoom, true);
+		DPIChangeExecution dpiChangeExecution = new DPIChangeExecution(true, null);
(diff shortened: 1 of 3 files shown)


amartya4256:
The underlying problem occurs because the setSize or setBounds is called asynchronously, i.e. after everything is scaled and the final position of the window at this instant is not guaranteed to comply with the bounds which was supposed to be set when the scaling had started. Meaning sometimes after setting the size of the Window after scaling it, the majority of the window can get scaled and span over the other monitor, triggering another DPI_CHANGE event. Hence, this triggers multiple DPI_CHANGE event while dragging.

HeikoKlare:
Thank you for the explanation. That totally makes sense and shows that we cannot set the size/bounds asynchronously.
I think the issue we have found in https://github.com/vi-eclipse/Eclipse-Platform/issues/530 is even involved here: since no event loop is spinned while dragging, only a mouse movement leads to a processing of asynchronous messages. That's probably why the window rescales when you start moving the mouse again, as only then the scheduled asynchronous DPI change events are processed.

So with those insights, I propose to close this PR and related issues. Any objections, @amartya4256?

amartya4256:
@HeikoKlare let's close this.

PR #2717, opened 2025-10-30
[Win32] Pass point values to size computation methods

HeikoKlare:
This increase precision of size computation by reducing the number of pixel/point conversions. In particular, the minimumSize() method unnecessarily calculated in pixels even though all relevant values are in points. Inside the computeSizeInPixels() methods, several conversions were done as most other called methods operate on point values.

I was not able to demonstrate benefits of the higher precision yet, but it anyway cleans up the code by removing unnecessary conversions.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Sash.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Sash.java
@@ -126,8 +126,11 @@ static int checkStyle (int style) {
 	return checkBits (style, SWT.HORIZONTAL, SWT.VERTICAL, 0, 0, 0, 0);
 }
 
-@Override Point computeSizeInPixels (int wHint, int hHint, boolean changed) {
+@Override
+Point computeSizeInPixels (Point hintInPoints, boolean changed) {
 	checkWidget ();
+	int zoom = getZoom();
+	Point hintInPixels = Win32DPIUtils.pointToPixelAsSize(hintInPoints, zoom);
 	int border = getBorderWidthInPixels ();
 	int width = border * 2, height = border * 2;
 	if ((style & SWT.HORIZONTAL) != 0) {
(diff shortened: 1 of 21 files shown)


HeikoKlare:
> Is it possible to add a unit test to test the precision?

I did not find a way to demonstrate the higher precision yet. Probably you can produce some scenario with specific zoom and size values and an according layout involved, but even if you found such a scenario, the test for it would be very specific. I would rather flag this as a simplification with the ability to improve precision without actually claiming the latter.

amartya4256:
> > Is it possible to add a unit test to test the precision?
> 
> I did not find a way to demonstrate the higher precision yet. Probably you can produce some scenario with specific zoom and size values and an according layout involved, but even if you found such a scenario, the test for it would be very specific. I would rather flag this as a simplification with the ability to improve precision without actually claiming the latter.

I was just wondering if there's a direct way to test that. But its fine without.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Sash.java
```diff
@@ -126,17 +126,20 @@ static int checkStyle (int style) {
...
 	int width = border * 2, height = border * 2;
 	if ((style & SWT.HORIZONTAL) != 0) {
 		width += DEFAULT_WIDTH;  height += 3;
 	} else {
 		width += 3; height += DEFAULT_HEIGHT;
 	}
-	if (wHint != SWT.DEFAULT) width = wHint + (border * 2);
-	if (hHint != SWT.DEFAULT) height = hHint + (border * 2);
+	if (hintInPoints.x != SWT.DEFAULT) width = hintInPixels.x + (border * 2);
+	if (hintInPoints.y!= SWT.DEFAULT) height = hintInPixels.y + (border * 2);
```
akoch-yatta: Can you add space before the !=

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Spinner.java
```diff
@@ -291,9 +295,9 @@ void addVerifyListener (VerifyListener listener) {
 	}
 	if (width == 0) width = DEFAULT_WIDTH;
 	if (height == 0) height = DEFAULT_HEIGHT;
-	if (wHint != SWT.DEFAULT) width = wHint;
-	if (hHint != SWT.DEFAULT)
-		height = hHint;
+	if (hintInPoints.x != SWT.DEFAULT) width = hintInPixels.x;
+	if (hintInPoints.y != SWT.DEFAULT)
+		height = hintInPixels.y;
 	else {
```
akoch-yatta: very nit, but if you touch that if here anyway, you could add brackets

akoch-yatta (approved):
Good looks good. Only have nitpicky comments

PR #2718, opened 2025-10-30
[Win32] Improve pixel-to-point conversion precision

HeikoKlare:
With recent changes to the rounding of pixel-to-point conversions the calculated sizes became too small in several use cases, leading to cut-off elements. This change adapts the rounding method to reduce the number of relevant cut-offs.

Follow-up to:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2654

### How to test
Issues caused by pixel/point conversion roundings can best be seen at 125% or 175% monitor zoom.

Before at 125%:
image

After at 125%:
image

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/Win32DPIUtils.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/Win32DPIUtils.java
@@ -124,7 +124,7 @@ public class Win32DPIUtils {
 	}
 
 	public static Point pixelToPointAsSize(Point point, int zoom) {
-		return pixelToPoint(point, zoom, RoundingMode.DOWN);
+		return pixelToPoint(point, zoom, RoundingMode.ROUND);
 	}
 
 	public static Point pixelToPointAsLocation(Point point, int zoom) {


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/Win32DPIUtils.java
```diff
@@ -124,7 +124,7 @@ public static Point pixelToPointAsLocation(Drawable drawable, Point point, int z
 	}
 
 	public static Point pixelToPointAsSize(Point point, int zoom) {
-		return pixelToPoint(point, zoom, RoundingMode.DOWN);
```
arunjose696, line 127: @HeikoKlare now as pixelToPointAsLocation and pixelToPointAsSize do the same thing, do we even need these separate methods?

arunjose696:
Tested this changes at 125 and 175% zoom this seems to resolve the mentioned issue, when testing around I can also see the issee https://github.com/vi-eclipse/Eclipse-Platform/issues/499  seems to be resolved for me at 125% zoom

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/Win32DPIUtils.java
HeikoKlare: We do not necessarily need them, but we already embedded the distinction at all callers and we would of course still like to find a better concept for rounding the values, so to work on that it may be easier to keep the distinction for now. Otherwise, for the same reasons, you could also get rid of all the `RoundingMode` and just have a special handling for the case of `computeSize()`.

arunjose696: sounds fair

PR #2720, opened 2025-10-30
Streamline rounding for rectangles and add rounding mode to them

HeikoKlare:
For the float-aware Point class, a rounding mode has been introduced to store in which way the Point needs to be rounded when necessary. For rectangles, this has not been implemented yet and is added by this change. It also correct the way in which rectangle values are rounded by considering the top-left and bottom-right corner instead of top-left corner and width/height, also done in the Win32DPIUtils. So it streamlines the rounding behavior in Win32DPIUtils and the Rectangle class to uniformly scale the upper left and bottom right corner of a rectangle.

The changes also makes proper use of the floating-precision coordinates in the point/pixel conversions for rectangles in the Win32DPIUtils.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Rectangle.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Rectangle.java
@@ -86,6 +86,9 @@ public Rectangle (int x, int y, int width, int height) {
 	this.height = height;
 }
 
+private Rectangle() {
+}
+
 /**
  * Destructively replaces the x, y, width and height values
  * in the receiver with ones which represent the union of the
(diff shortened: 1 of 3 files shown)


amartya4256:
@HeikoKlare I see that only RoundingMode.ROUND is used in the code for now for rectangles. Could you please explain what RoundingMode is intended to be used in which case?

HeikoKlare:
With the current changes, we will revert to `ROUND` everywhere (except for the conservative calculations in `computeSize()`), as everything else broke behavior at one place or the other. But properly introducing the rounding mode everywhere allows to better test changes to the rounding mode at the different places.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Rectangle.java
```diff
@@ -440,28 +452,38 @@ public float getHeight() {
...
 	public void setY(float y) {
-		this.y = Math.round(y);
+		this.y = locationRounding.round(y);
 		this.residualY = y - this.y;
+		setHeight(getHeight());
 	}
 
 	public void setWidth(float width) {
-		this.width = Math.round(width);
+		this.width = sizeRounding.round(width + getX()) - x;
```
amartya4256, line 467: I remember there were cases in the past where some places would have different results when you would scale the rectangle using scaleBounds or scaleUp because they both used different approach to scaling. One had height and width calculated using the difference between top left and bottom right and the other as just the width and height values in the object. Are we sure that this approach would not break that behavior for certain consumers? Visually I cannot see anything in the IDE but I just wanted to pick this out just to be careful. I am going to consider that since there's no test failing, this might not be the case anymore?

HeikoKlare: I think there is no way to be 100% sure that we to not break any behavior for existing consumers. With all the back-and-forth of the kinds of rounding we have implemented, this will probably also have effects that we currently do not see (either positive or negative). 
Regarding the different scaling methods that you mention: they are still in place. We still have `Win32DPIUtils#scaleBounds()` which implement the "other" way of scaling (used for things related to image). The modified way of scaling only applies to what is passed to `Win32DPIUtils#pointToPixel()/pixelToPoint()`. And there the only removed case distinction is that when a `Rectangle.OfFloat` was passed, no forwarding to `scaleBounds()` happens anymore.

HeikoKlare:
> Since we have a public constructor to initialize a rectangle with a specific rounding, are there any possibilities where we will be using any other rounding mode in future? If not can we hide it for the time being?

Sure, I've made it private for now. The class and all its methods are `@noreference` anyway (so effectively no API), but since the constructor is not used yet, it's definititely better to not make it public at all.

akoch-yatta (approved):
Code looks fine. Although there is no obvious visible improvement, I'll follow the argument with consistency and would merge this. I did not find a regression when testing around with different zoom values

PR #2723, opened 2025-10-31
Setup Repository for GitHub Copilot

laeubi:
Currently the Repository is not setup for copilot and could make it quite hard for the tool to give useful results because of the special setup.

This now is a first attempt to make it more valuable for that case let it be helping with analysis of core dumps or migrating GTK4 by improve the setup for at least the GTK case:

- creating general copilot instructions file
- creating a specialized GTK instructions file
- adding workflow setup that prepares the environment
- add a document to track current GTK4 deprecation warnings

FYI @akurtakov

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,0 +1,158 @@
+# GitHub Copilot Instructions for Eclipse Platform SWT
+
+## Project Overview
+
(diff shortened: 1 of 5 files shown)


laeubi:
Nothing affects the build.

PR #2726, opened 2025-10-31
Fix some GTK4 deprecation warning

laeubi:
- Replace gtk_widget_get_allocation with gtk_widget_compute_bounds (with GTK4 version check)
- Replace gtk_css_provider_load_from_data with gtk_css_provider_load_from_string
- Replace gtk_widget_translate_coordinates with gtk_widget_compute_point

These changes reduce GTK4 deprecation warnings from 159 to 156.

---------


bundles/org.eclipse.swt/Eclipse SWT PI/gtk/library/gtk4.c
jonahgraham, line 13: @laeubi these files are auto generated from GTK4.java and OS.java respectively, so the changes should be made at the Java level for this.

jonahgraham:
PS If the problem really needs to be solved in the C code, then it needs to be added to os_custom.c and in OS.java set:

```
	/** @method flags=no_gen
	 * @category custom
	 */
```

laeubi:
@jonahgraham thanks for the pointer I'll take a look to find out what would be the proper way here.

akurtakov:
Closing this one as it manually changes two files that should be auto generated and does so wrong e.g. binding for gtk_css_provider_load_from_data calling gtk_css_provider_load_from_string

akurtakov:
Tip: gtk_css_provider_load_from_string/bytes has a since 4.12 while SWT still supports 4.0 (or at least claims to do so).

PR #2727, opened 2025-10-31
Removed layout calls in handleDPIChange methods

amartya4256:
This PR removes layout calls in handleDPIChange methods of different widgets since Shell:layout called at the end of DPIChange event triggers layout for all the children in order hence these layouts are not needed.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/CoolBar.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/CoolBar.java
@@ -1242,6 +1242,5 @@ void handleDPIChange(Event event, float scalingFactor) {
 		}
 	}
 	setItemLayoutInPixels(itemOrder, indices, scaledSizes);
-	updateLayout(true);
 }
 }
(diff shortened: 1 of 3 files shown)


HeikoKlare (changes requested):
This does not work for me. After simply moving from 150% monitor to 100% monitor, the toolbar looks like this for me:
image

So seems like the layout calls for CoolBar/ToolBar are somehow necessary.

amartya4256:
> So seems like the layout calls for CoolBar/ToolBar are somehow necessary.

apparently, the layout call in toolBar in case of when used with ImageBasedFrame does nothing since there's no layout set. The only thing which causes the problem is the wHint and hHint not being reset. Hence, we need to clear the cached wHint and hHint.

This is how it works:
1. handleDpiChange resets the cached wHint and hHint.
2. Calculated again using computeSize by parent or pack.

ImageBasedFrame calls pack. In a normal case where widgets have layout, this is calculated again using computeSize in layout.layout.

laeubi:
>  In a normal case where widgets have layout, this is calculated again using computeSize in layout.layout.

As mentioned earlier having a layout is not mandatory (even though often the case) and some widgets even say setting a layout on them is not useful. So if no layout is set I would assume that SWT needs to update the location/size itself.

amartya4256:
> As mentioned earlier having a layout is not mandatory (even though often the case) and some widgets even say setting a layout on them is not useful. So if no layout is set I would assume that SWT needs to update the location/size itself.

Agreed. It is just an example for what happens in a few cases. I have added a fix.

HeikoKlare (approved):
Change is reasonable, including the cache clearing in `ToolBar`. I first wondered if it would be better to adapt the early return in `computeSizeInPixels()` (which is the consumer of the cached values) to check the cheched values against `SWT.DEFAULT` and in that case recompute, but when I take a look at the calculations is seems like for the same zoom the calculation will always yield the same result, so early return on `SWT.DEFAULT` is fine.

I tested with an SDK and the individual controls with the `ControlExample` and did not find any issues.

PR #2729, opened 2025-11-03
[Win32] Adapt further places to round up when computing control size

HeikoKlare:
Several places in the computations of preferred sizes of controls used commercial rounding instead of rounding up in pixel/point calculations. This change further adapts additional places to ensure that the calculated size will be large enough and and subsequent point-to-pixel calculation when setting the bounds according the computations does not lead to smaller results.

This is a follow up to:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2717

I did not find any visual differences yet, so consider this a unification/cleanup to ensure that every conversion involved in `computeSize()` calculations rounds up.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/CoolItem.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/CoolItem.java
@@ -186,16 +186,16 @@ protected void checkSubclass () {
 public Point computeSize (int wHint, int hHint) {
 	checkWidget ();
 	int zoom = getZoom();
-	wHint = (wHint != SWT.DEFAULT ? DPIUtil.pointToPixel(wHint, zoom) : wHint);
-	hHint = (hHint != SWT.DEFAULT ? DPIUtil.pointToPixel(hHint, zoom) : hHint);
-	return Win32DPIUtils.pixelToPointAsConservativeSize(computeSizeInPixels(wHint, hHint), zoom);
+	return Win32DPIUtils.pixelToPointAsSufficientlyLargeSize(computeSizeInPixels(new Point(wHint, hHint)), zoom);
 }
-Point computeSizeInPixels (int wHint, int hHint) {
+Point computeSizeInPixels (Point sizeHintInPoints) {
+	int zoom = getZoom();
+	Point sizeHintInPixels = Win32DPIUtils.pointToPixelAsSufficientlyLargeSize(sizeHintInPoints, zoom);
 	int index = parent.indexOf (this);
 	if (index == -1) return new Point (0, 0);
-	int width = wHint, height = hHint;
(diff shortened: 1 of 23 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/CoolItem.java
```diff
@@ -188,7 +188,7 @@ public Point computeSize (int wHint, int hHint) {
 	int zoom = getZoom();
 	wHint = (wHint != SWT.DEFAULT ? DPIUtil.pointToPixel(wHint, zoom) : wHint);
 	hHint = (hHint != SWT.DEFAULT ? DPIUtil.pointToPixel(hHint, zoom) : hHint);
```
akoch-yatta: mustn't wHint and hHint be upscaled as with Point with ...AsSufficientlyLargeSize as well?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -613,7 +613,7 @@ private class CopyAreaOperation extends Operation {
 	@Override
 	void apply() {
 		int zoom = getZoom();
-		Rectangle sourceRect = Win32DPIUtils.pointToPixel(drawable, source, zoom);
+		Rectangle sourceRect = Win32DPIUtils.pointToPixelWithSufficientlyLargeSize(source, zoom);
 		Rectangle destRect = Win32DPIUtils.pointToPixel(drawable, destination, zoom);
```
akoch-yatta: Can you shortly explain why source must be upscaled like this, buts dest not? couldn't that lead to unwanted different values, so source in pixels is bigger than dest in pixels?

HeikoKlare: iirc the idea was to ensure that the source area taken for copy is not too small. But I think this might be a risky change and should be considered unrelated to the rest of this PR. Thus I will revert that change.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/CoolItem.java
HeikoKlare: It definitely makes sense to do so. That was not done by the previous refactoring because `CoolItem` is no `Control` (for which I adapted all `computeSizeInPixels()` methods. I now adapted the method here accordingly as well.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
akoch-yatta: Thanks, I am in favor of having a look at this change separately 👍

Issue #2734, opened 2025-11-05
Support for webm image format


tmssngr:
**Is your feature request related to a problem? Please describe.**
Allow to load webm image format(s).

**Describe the solution you'd like**
Just like support for png, jpeg, gif, bmp.

akurtakov:
I would assume you want webp (https://en.wikipedia.org/wiki/WebP) as webm is container format for various audio/video codecs thus totally out of scope for SWT.

With the previous said - 
SWT on Gtk can load it just fine already 

<img width="3664" height="2232" alt="Image" src="https://github.com/user-attachments/assets/992c8e58-7046-4b71-b845-cd6c2c26e53b" />

due to the fact that it relies on the system codecs for loading images. I don't see anyone implementing webp codec in swt , like it was done for other formats back in the day and codecs are stuck there. So best path forward would be for Windows and MacOS ports to add image support based on native codecs.

laeubi:
I added some experiment here, but the format is not trivial:

- https://github.com/laeubi/eclipse.platform.swt/pull/15

so if one wants to go that way it could be used as a skeleton and then further investigate how to implement the different decoders.

So the main question for e would be the usecase, if you just want to display some of them using the embedded browser might give good enough results.

PR #2735, opened 2025-11-05
Add "All Files" filter to ImageAnalyzer

akurtakov:
It allows to select more file types that are not supported on all platforms (e.g. webp).


akurtakov:
This change is for the sake of https://github.com/eclipse-platform/eclipse.platform.swt/issues/2734#issuecomment-3490300513

Issue #2733, opened 2025-11-05
[win32] Widget constructor calls overrideable public method; breaks EGit's `FindToolbar`


tomaswolf:
**Describe the bug**
The `Widget` constructor

https://github.com/eclipse-platform/eclipse.platform.swt/blob/a9598b8a0da4a9ab68920ae687dfa3e365f77a58/bundles/org.eclipse.swt/Eclipse%20SWT/win32/org/eclipse/swt/widgets/Widget.java#L179-L197

calls via `registerDPIChangeListener` the public and overrideable method `addListener`. If that method is overridden in a subclass it will run before the subclass instance is initialized. This is bad practice, and in fact it breaks EGit's `FindToolbar`: see https://github.com/eclipse-egit/egit/issues/117 .

**Possible fix**

Replace the call to `this.addListener` at

https://github.com/eclipse-platform/eclipse.platform.swt/blob/a9598b8a0da4a9ab68920ae687dfa3e365f77a58/bundles/org.eclipse.swt/Eclipse%20SWT/win32/org/eclipse/swt/widgets/Widget.java#L192

by `this._addListener` (with underscore, package-private).

fedejeanne:
Makes sense. Would you like to provide a PR?

fedejeanne:
FTR I see other calls in the code that may run into the same issue, constructors being the the most prominent ones:

<img width="658" height="104" alt="Image" src="https://github.com/user-attachments/assets/e1d0378b-e155-4f0c-90a6-5611e64d77f7" />

---

<img width="826" height="334" alt="Image" src="https://github.com/user-attachments/assets/29dbd19f-959e-4bab-a5aa-997392c1c8a6" />

---

There might be many more, I didn't look extensively

tomaswolf:
> Would you like to provide a PR?

Sorry, no. I don't have a Windows development machine, and I'm not set up for SWT development.

fedejeanne:
> > Would you like to provide a PR?
> 
> Sorry, no. I don't have a Windows development machine, and I'm not set up for SWT development.

@tomaswolf I took care of it, with some extra PRs to prevent similar problems in the future. The last PR should be merged before the next release. ✅ 

If I may: _Quid pro quo_ --> https://review.gerrithub.io/c/eclipse-jgit/jgit/+/1225695, https://review.gerrithub.io/c/eclipse-egit/egit/+/1225216 ? :-)

tomaswolf:
[...]

PR #2736, opened 2025-11-05
Do not call addListener in Widget::registerDPIChangeListener #2733

fedejeanne:
The method is not final, which may cause issues if a subclass overrides it because it ends up being called upon instantiation, when the object is not fully initialized

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2733


merks:
@fedejeanne 

FYI, it looks like the Jenkins build is affected this issue:

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3461

It's probably safe to rely on the results of the GHA verification.

PR #2737, opened 2025-11-05
[Refactoring] Streamline calls to Image#adaptImageDataIfDisabledOrGray()

ShahzaibIbrahim:
Enabling a consistent pattern to apply disablement or graying of an Image. They are applied whenever a handle is being created from init method. Ensuring streamline usage and avoid double calls to this method.

### How to test
- Run the `Snippet382` with following VM Argument: `swt.autoScale.updateOnRuntime=true`
- Move the shell between different zoom levels (100,125,150,175,200,250)
- The snippet tests different kind of image providers to create an Image.
- Expected result would be that we don't lose disabled or grayed image attributes. 

image

---

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
@@ -172,8 +172,7 @@ public final class Image extends Resource implements Drawable {
 		private Optional<ImageHandle> createHandleAtExactSize(int width, int height) {
 			Optional<ImageData> imageData = imageProvider.loadImageDataAtExactSize(width, height);
 			if (imageData.isPresent()) {
-				ImageData adaptedData = adaptImageDataIfDisabledOrGray(imageData.get());
-				ImageHandle imageHandle = init(adaptedData, -1);
+				ImageHandle imageHandle = init(imageData.get(), -1);
 				return Optional.of(imageHandle);
 			}
 			return Optional.empty();
@@ -187,8 +186,7 @@ public final class Image extends Resource implements Drawable {
 			ImageHandle bestFittingHandle = zoomLevelToImageHandle.get(imageZoom);
 			if (bestFittingHandle == null) {
 				ImageData bestFittingImageData = imageProvider.loadImageData(imageZoom).element();
-				ImageData adaptedData = adaptImageDataIfDisabledOrGray(bestFittingImageData);
-				bestFittingHandle = init(adaptedData, -1);
+				bestFittingHandle = init(bestFittingImageData, -1);
 			}
 			return bestFittingHandle;
 		}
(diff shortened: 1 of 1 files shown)


ShahzaibIbrahim:
@copilot could you review it first.

akoch-yatta:
One remark: as the change seems to affect Cursor as well - Is there a Snippet/test case or could you provide one to test correctness for Cursor?

ShahzaibIbrahim:
> One remark: as the change seems to affect Cursor as well - Is there a Snippet/test case or could you provide one to test correctness for Cursor?

I have added a sanity test for **disabled cursor** in the description.

ShahzaibIbrahim:
@HeikoKlare there seems to be a confusion. For Icon it was working even before. Although during all these refactoring the signature for org.eclipse.swt.graphics.Image.init(Device, ImageData, int) has changed which accepts the style flag and org.eclipse.swt.graphics.Image.initIcon(Device, ImageData, ImageData, int) method which is called from cursor now need to pass the style flag (which could be null since the image for that now can be initialized directly with disabled or gray flag. 

The short answer is the flag was applied even before this change for the images of type Icon (can be tested through `Snippet382`)

I have made a small change which will reflect previous behavior as is for the icons after refactoring.

HeikoKlare:
> The short answer is the flag was applied even before this change for the images of type Icon (can be tested through `Snippet382`)

When running `Snippet382`, neither of the `initIconHandle` methods seems to be called. I've set breakpoints in both of them and none of them was triggered.

I went through the call sequences a bit to see if I miss something and found that `init()` (where the adaptation will be moved to by this PR) is executed quite often (obviously for every handle to be created). In many cases, a temporary handle will be created first (just for the sake of generating image data) and the extracted image data will later be used to create an actual handle to be used. So can't we improve the implementation to only do the adaptation when the initial image data / handle is created and no repeated applications of the adaptation logic to the same data / handle happen?

ShahzaibIbrahim:
> When running `Snippet382`, neither of the `initIconHandle` methods seems to be called. I've set breakpoints in both of them and none of them was triggered.

```
import org.eclipse.swt.*;

public class Snippet70 {

	Color red = display.getSystemColor (SWT.COLOR_RED);
	Color white = display.getSystemColor (SWT.COLOR_WHITE);
	Color black = display.getSystemColor (SWT.COLOR_BLACK);

	Image image = new Image (display, 20, 20);
	GC gc = new GC (image);
	gc.setBackground (red);
	gc.fillRectangle (5, 5, 10, 10);
	gc.dispose ();
	ImageData imageData = image.getImageData ();

	PaletteData palette = new PaletteData (new RGB (0, 0, 0),new RGB (0xFF, 0xFF, 0xFF));
	ImageData maskData = new ImageData (20, 20, 1, palette);
	Image mask = new Image (display, maskData);
	gc = new GC (mask);
	gc.setBackground (black);
	gc.fillRectangle (0, 0, 20, 20);
	gc.setBackground (white);
	gc.fillRectangle (5, 5, 10, 10);
	gc.dispose ();
	maskData = mask.getImageData ();

	Image icon = new Image (display, imageData, maskData);
	Image disabledIcon = new Image(display, icon, SWT.IMAGE_DISABLE);
	Image grayedIcon = new Image(display, icon, SWT.IMAGE_GRAY);
	shell.setText("Snippet 70");
	Button button = new Button (shell, SWT.PUSH);
	button.setImage (disabledIcon);
	button.setSize (60, 60);

	Button button2 = new Button (shell, SWT.PUSH);
	button2.setImage (grayedIcon);
	button2.setSize (60, 60);
	shell.open ();
	}
	icon.dispose ();
	disabledIcon.dispose ();
	grayedIcon.dispose();
	image.dispose ();
	mask.dispose ();
	display.dispose ();
}
}
```
This snippet can test the icon disablement and grayness. Initially I though we had an example in 382.

PR #2739, opened 2025-11-05
Do not call addListener in constructors of CCombo and StyledText #2733

fedejeanne:
The method is not final, which may cause issues if a subclass overrides it because it ends up being called upon instantiation, when the object is not fully initialized

Since these 2 classes are not in the same package as other subclasses of Widget, I had to resort to expanding the functionality in TypedListener.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2733

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CCombo.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CCombo.java
@@ -173,7 +173,18 @@ public CCombo (Composite parent, int style) {
 	}
 
 	initAccessible();
-	addListener(SWT.ZoomChanged, this::handleDPIChange);
+
+	// Add listener asynchronously in order to delay execution. This works because
+	// other DPI changes are also executed asynchronously but if we choose to revert
+	// back to synchronous execution of the listeners then this async call may be
+	// executed too late and zoom change events may happen too early and not get
+	// propagated.
+	// See https://github.com/eclipse-platform/eclipse.platform.swt/issues/2733
+	getDisplay().asyncExec(() -> {
+		if (!isDisposed()) {
+			addListener(SWT.ZoomChanged, this::handleDPIChange);
+		}
+	});
 }
 static int checkStyle (int style) {
 	int mask = SWT.BORDER | SWT.READ_ONLY | SWT.FLAT | SWT.LEFT_TO_RIGHT | SWT.RIGHT_TO_LEFT | SWT.LEAD | SWT.CENTER | SWT.TRAIL;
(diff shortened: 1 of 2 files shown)


fedejeanne:
Need to rethink this since the replacement is also official API and can therefore also be overridden

fedejeanne:
I've added 2 internal-API `final` methods to avoid calling the public `addListener(int, Listener)` and to not have to change the existing (package-friendly, non-final) `_addListener(int, Listener)` because that would be a breaking change. It would technically be a breaking change on a non-API method but I decided to go with the new method anyway.

FTR the calls that I am correcting in this PR have been introduced recently:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2483

HeikoKlare (commented):
Maybe an unconventional thought on avoiding such a new API for just a very specific use case: what about wrapping the `addListener` calls into a `display.asyncExec()`? I think it's not necessary that the listeners are registered synchronously, so that could move it out of the constructor to avoid the potential problems with overwritten methods as well.

fedejeanne:
In principle it _could_ work but I fear that it would mean that one needs to check for disposal when overriding the public `addListener(...)` method. It seems obscure, I would like to avoid imposing implementation burdens on consumers, if possible.

HeikoKlare:
> I fear that it would mean that one needs to check for disposal when overriding the public `addListener(...)` method.

Why should that be necessary? You would check for disposal before calling `addListener` as common in `asyncExec` calls.

fedejeanne:
hm. on a second thought: consumers **should** be checking for disposal already. I mean the original `addListener` does that already via `checkWidget()`.

https://github.com/eclipse-platform/eclipse.platform.swt/blob/136b37582ea88460c822cab9db372e396f4f747a/bundles/org.eclipse.swt/Eclipse%20SWT/win32/org/eclipse/swt/widgets/Widget.java#L231-L235

> You would check for disposal before calling `addListener` as common in `asyncExec` calls.

You're right, that would be easily solvable on my side by calling `checkWidget()`.

I'll adapt my PR with your suggestion 👍

fedejeanne:
Adapted

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CCombo.java
```diff
@@ -173,7 +173,13 @@ public CCombo (Composite parent, int style) {
 	}
 
 	initAccessible();
-	addListener(SWT.ZoomChanged, this::handleDPIChange);
+
+	// Add listener asynchronously in order to delay execution.
+	// See https://github.com/eclipse-platform/eclipse.platform.swt/issues/2733
+	getDisplay().asyncExec(() -> {
+		checkWidget();
```
HeikoKlare: `checkWidget()` is not correct here, as it will erroneously throw an exception instead of just skipping execution when the widget is disposed. I.e., this should just wrap the `addListener()` call into an `if (!isDisposed())`.

fedejeanne: Fixed in ad8479ce2f96237c4219bffb0c89143860bb2333, thank you.

I also changed the comment and added a bit more details about what may happen if we decide to switch back to synchronous execution of zoom change events.

akoch-yatta:
> Except for the widget check, the change looks sound to me. @akoch-yatta what do you think? Do we miss anything?

I think, it is fine

PR #2742, opened 2025-11-06
Remove zoom parameter from CoordinateSystemMapper#mapMonitorBounds()

ShahzaibIbrahim:
The CoordinateSystemMapper#mapMonitorBounds() method currently accepts a zoom value and that is wrong for SingleZoomCoordinateMapper since it should always use the global zoom (DPIUtil#getDeviceZoom()) instead of monitor zoom. This commit changes how zoom is passed through Rectangle with monitor for MultiZoomCoordinateMapper while that monitor will be ignored for SingleZoomCoordinateMapper.

### How to Test

- Run the following snippet with no VM argument (monitor-specific scaling off) with two monitors connected (250% and 150%)
```
import org.eclipse.swt.*;

public class OpenNewShellExample {

        Shell mainShell = new Shell(display);
        mainShell.setText("Main Shell");
        mainShell.setSize(300, 200);

        Button openButton = new Button(mainShell, SWT.PUSH);
        openButton.setText("Open New Shell");
        openButton.setBounds(80, 70, 140, 30);

        openButton.addListener(SWT.Selection, e -> {
            Rectangle mainShellRect = mainShell.getMonitor().getBounds();
            Shell newShell = new Shell(mainShell);
            newShell.setText("New Shell");
            newShell.setSize(mainShellRect.width, mainShellRect.height);

            Label infoLabel = new Label(newShell, SWT.NONE);
            infoLabel.setLocation(100, 100);
            infoLabel.setSize(300, 20);
            infoLabel.setText("New Shell Size: " + mainShellRect.width + "x" + mainShellRect.height);
            newShell.open();
        });

        mainShell.open();
        while (!mainShell.isDisposed()) {
                display.sleep();
        }
        display.dispose();
    }
}
```
- Keep the shell on primary monitor and click "Open New Shell" button
- A new shell will open with the same size of primary monitor (Global Zoom)
- Close the new shell
- Move the first shell to secondary monitor 
- Click "Open New Shell" button again.
- The new shell that open should be the same size as the original zoom since monitor-specific scaling is not turned on.

### Results

**Before**
At 250%
image

at 150%
image

Notice that zoom is adapted of monitor even though monitor-specific scaling is turned off.

**After**
At 250% (Primary)
image

At 150% (Secondary) 
image

Therefore the size remain same (which is correct behaviour)

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/CoordinateSystemMapper.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/CoordinateSystemMapper.java
@@ -25,7 +25,7 @@ interface CoordinateSystemMapper {
 
 	Point map(Control from, Control to, int x, int y);
 
-	Rectangle mapMonitorBounds(Rectangle rectangle, int zoom);
+	Rectangle mapMonitorBounds(Rectangle.WithMonitor rectangle);
 
 	Point translateFromDisplayCoordinates(Point point);
 
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
@@ -2225,16 +2225,10 @@ Monitor getMonitor (long hmonitor) {
 	OS.GetMonitorInfo (hmonitor, lpmi);
 	Monitor monitor = new Monitor ();
 	monitor.handle = hmonitor;
-	Rectangle boundsInPixels = new Rectangle(lpmi.rcMonitor_left, lpmi.rcMonitor_top, lpmi.rcMonitor_right - lpmi.rcMonitor_left,lpmi.rcMonitor_bottom - lpmi.rcMonitor_top);
-	Rectangle clientAreaInPixels = new Rectangle(lpmi.rcWork_left, lpmi.rcWork_top, lpmi.rcWork_right - lpmi.rcWork_left, lpmi.rcWork_bottom - lpmi.rcWork_top);
 	int [] dpiX = new int[1];
 	int [] dpiY = new int[1];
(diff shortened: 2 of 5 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Display.java
```diff
@@ -2225,25 +2225,25 @@ Monitor getMonitor (long hmonitor) {
...
 	result = (result == OS.S_OK) ? DPIUtil.mapDPIToZoom (dpiX[0]) : 100;
-
-	int autoscaleZoom = DPIUtil.getZoomForAutoscaleProperty(result);
-	monitor.setBounds(coordinateSystemMapper.mapMonitorBounds(boundsInPixels, autoscaleZoom));
-	monitor.setClientArea(coordinateSystemMapper.mapMonitorBounds(clientAreaInPixels, autoscaleZoom));
 	if (result == 0) {
 		System.err.println("***WARNING: GetDpiForMonitor: SWT could not get valid monitor scaling factor.");
 		result = 100;
 	}
+	monitor.zoom = result;
```
HeikoKlare, line 2240: Comment of this code line was not moved

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/CoordinateSystemMapper.java
```diff
@@ -25,7 +25,7 @@ interface CoordinateSystemMapper {
 
 	Point map(Control from, Control to, int x, int y);
 
-	Rectangle mapMonitorBounds(Rectangle rectangle, int zoom);
+	Rectangle mapMonitorBounds(Rectangle rectangle);
```
HeikoKlare: Why not as follows to avoid unnecessary type checks?
```suggestion
	Rectangle mapMonitorBounds(Rectangle.WithMonitor rectangle);
```

ShahzaibIbrahim:
> In general, the change looks good to me. Could we add some tests to `CoordinateSystemMapperTests` that document and ensure the expected behavior for the two mappers? The snippets seems to be a good starting point for that.

I have added the test. Need to be reviewed.

PR #2703, opened 2025-10-30
[GTK4] Add missing Callback dispose in Display

jonahgraham:

PR #2744, opened 2025-11-07
[GTK4] Update callback.c debugging for GDK4

jonahgraham:
This was useful for diagnosing https://github.com/eclipse-platform/eclipse.platform.swt/pull/2703

To test this change you need to uncomment `#define DEBUG_CALL_PRINTS`


laeubi:
@jonahgraham maybe we should have generally a way to enable tracing with ENV varibales in SWT so it can be used without recompile the code?

jonahgraham:
> @jonahgraham maybe we should have generally a way to enable tracing with ENV varibales in SWT so it can be used without recompile the code?

Yes - it is a good idea, perhaps next time I need to look at it I'll add this option.

PR #2748, opened 2025-11-10
new flag for propagating auto scale disablement

arunjose696:
Previously, if a control had AutoScaleDisabled, all its children would inherit this setting by default. This change introduces a new flag PROPAGATE_AUTOSCALE_DISABLED (default: true). Setting it to false via ontrol.setData("PROPAGATE_AUTOSCALE_DISABLED", false) prevents children from inheriting AutoScaleDisabled. 

The fix is based on @amartya4256 branch 
- https://github.com/vi-eclipse/eclipse.platform.swt/tree/amartya4256/autoscale_disabled

This change fixes issues with cell editors in GEF-classic.

**Steps to reproduce**
 Follow steps in https://github.com/eclipse-gef/gef-classic/pull/850

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
@@ -82,6 +82,8 @@ public abstract class Control extends Widget implements Drawable {
 	private static final String DATA_SHELL_ZOOM = "SHELL_ZOOM";
 
 	private static final String DATA_AUTOSCALE_DISABLED = "AUTOSCALE_DISABLED";
+
+	private static final String PROPOGATE_AUTOSCALE_DISABLED = "PROPOGATE_AUTOSCALE_DISABLED";
 /**
  * Prevents uninitialized instances from being created outside the package.
  */
@@ -121,7 +123,13 @@ Control () {
 public Control (Composite parent, int style) {
 	super (parent, style);
 	this.parent = parent;
-	this.autoScaleDisabled = parent.autoScaleDisabled;
+	Boolean parentPropagateAutoscaleDisabled = (Boolean) parent.getData(PROPOGATE_AUTOSCALE_DISABLED);
+	if (parentPropagateAutoscaleDisabled == null || parentPropagateAutoscaleDisabled) {
+		this.autoScaleDisabled = parent.autoScaleDisabled;
+	}
+	if (!autoScaleDisabled) {
+		this.nativeZoom = getShellZoom();
+	}
 	createWidget ();
 }
 
@@ -618,7 +626,7 @@ public Point computeSize (int wHint, int hHint) {
  */
 public Point computeSize (int wHint, int hHint, boolean changed){
 	checkWidget ();
-	int zoom = getZoom();
+	int zoom = computeBoundsZoom();
 	//We should never return a size that is to small, RoundingMode.UP ensures we at worst case report
 	//a size that is a bit too large by half a point
 	return Win32DPIUtils.pixelToPointAsSufficientlyLargeSize(computeSizeInPixels(new Point(wHint, hHint), zoom, changed), zoom);
@@ -1206,7 +1214,7 @@ int getBorderWidthInPixels () {
  */
 public Rectangle getBounds (){
 	checkWidget ();
-	return Win32DPIUtils.pixelToPoint(getBoundsInPixels (), getZoom());
+	return Win32DPIUtils.pixelToPoint(getBoundsInPixels (), computeBoundsZoom());
 }
 
 Rectangle getBoundsInPixels () {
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
```diff
@@ -4824,6 +4835,13 @@ int getShellZoom() {
 	return nativeZoom;
 }
 
+int getScalingZoom() {
+	if (parent != null) {
+		return parent.getZoom();
+	}
+	return getZoom();
+}
```
akoch-yatta: I think it makes sense to extract this into a separate method, but two remarks:
- can be private
- I think about a better naming for this, e.g. `computeBoundsZoom` because we only use it in the context of changes to the bounds of the control

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
```diff
@@ -121,7 +124,12 @@ public abstract class Control extends Widget implements Drawable {
 public Control (Composite parent, int style) {
 	super (parent, style);
 	this.parent = parent;
-	this.autoScaleDisabled = parent.autoScaleDisabled;
+	this.propagateAutoScaleDisabled = parent != null ? parent.propagateAutoScaleDisabled : true;
```
akoch-yatta: I am not sure, if propagating it is the correct call. I would propose to only disable it via setData for the affected control

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
```diff
@@ -121,7 +124,12 @@ public abstract class Control extends Widget implements Drawable {
...
+	if (parent != null && parent.propagateAutoScaleDisabled) {
+		this.autoScaleDisabled = parent.autoScaleDisabled;
+	} else {
+		this.nativeZoom = getShellZoom();
+	}
```
akoch-yatta: I don't fully understand this if-else:
The if part makes sense to me, but the else part looks a bit misleading to me. I think it works as it is, but currently it says: if the parent does not propagate autoScale disablement you should use shell zoom, but that only works as long as the default value of `autoScaleDisabled` is false, if we would change that to true, this code break, long story short, I would propose to move that into a separate if like:
```java
if (!autoScaleDisabled) {
   this.nativeZoom = getShellZoom();
}
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
```diff
@@ -78,10 +78,13 @@ public abstract class Control extends Widget implements Drawable {
 	Font font;
 	int drawCount, foreground, background, backgroundAlpha = 255;
 	boolean autoScaleDisabled = false;
+	boolean propagateAutoScaleDisabled = true;
```
HeikoKlare: Could we just retrieve the value via `getData()` where necessary instead of storing another value for every control? It seems to happen so seldom that we may accept the overhead that `getData()` may have.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
```diff
@@ -1286,6 +1294,9 @@ public void setData(String key, Object value) {
 			this.nativeZoom = getShellZoom();
 		}
 	}
+	if (PROPOGATE_AUTOSCALE_DISABLED.equals(key)) {
+		this.propagateAutoScaleDisabled = Boolean.parseBoolean(value.toString());
```
HeikoKlare: Do we need to propagate the news setting to childs that may already exist? With the current implementation, childs will only be properly affected if `setData("PROPAGATE_AUTOSCALE_DISABLED")` is called before those children are created, isn't it? Or is that even intended, as when the first childs were created that flag was not set yet?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
arunjose696: DOne

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
arunjose696: The only currently planned usage is in GEF, where the data would be set directly in the constructor. This ensures that all children will see the flag set. Given our current use cases, the state appears fine. In my opinion, it is acceptable not to modify existing children created before the flag was introduced, in case the flag is used elsewhere.

HeikoKlare (approved):
The change makes sense to improve auto-scale disablement at control level and is in itself sound.
I tested in particular with GEF Classic (in combination with https://github.com/eclipse-gef/gef-classic/pull/850), both with GEF example diagrams and with our RCP products (which has some very specific CellEditor and locator implementations) and it works fine in every scenario I tested.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
HeikoKlare: Agreed, thank you.

PR #2751, opened 2025-11-11
Add AGENTS.md documentation for AI tools

vogella:
This adds comprehensive documentation for AI agents and tools working with the Eclipse SWT codebase. The new AGENTS.md file includes:

- Project overview and architecture details
- Complete build instructions (Maven/Tycho and native libraries)
- Repository structure and development workflows
- Essential development rules (OSGi, resource disposal, UI threading)
- Testing best practices and troubleshooting guidance
- AI agent-specific guidelines

CLAUDE.md is updated to reference AGENTS.md, following the pattern used in eclipse.platform.ui repository.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,1 @@
+@AGENTS.md


vogella:
> Please rename [existing copilot instructions](https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/.github/copilot-instructions.md) (any maybe adjust).

I assume you mean delete? Copilot can use Agends.md directly

> As AGENTS.MD is a standard I wonder why Claude is needing an own file?

I don't know why Claude does not follow the (new) standard. My assumption is that as the leading engine in code generation they play on their marketshare for code support and do not yet care about standards. It seems that standards are usually not pushed by the market leader. Just a personal opinion btw as I do not have insights in the Claude dev team.

laeubi:
> I assume you mean delete? Copilot can use Agends.md directly

No why delete and replace with something else new? So rename it and maybe ask claude to update it if you think it is not sufficient enough. At least I have tested it already with this instructions and made sure they are working. So why would one want to start from scratch?

> I don't know why Claude does not follow the (new) standard. My assumption is that as the leading engine in code generation they play on their marketshare for code support and do not yet care about standards. 

For me a tool not following common practice can hardly be leading but anyways it sounds odd... so is there a issue/ticket for claude to support this in the future?

vogella:
The new Agends.md file was created by on the copilot instruction so this request is already covered. 

> For me a tool not following common practice can hardly be leading 

Yes, this never happened before, like  IntelliJ not promoting LSP usage or MS Window ignoring standards ;-)

laeubi:
> The new Agends.md file was created by on the copilot instruction so this request is already covered.

Why is the there no rename in this commit? I really don't want to go through and make a diff on my own.

So this should contain:

- rename A > B
- then changes on top of it to get a diff
- add additional claude.md

otherwise its impossible to follow the history.

vogella:
Ok, will update the PR once I find time.

laeubi:
> Ok, will update the PR once I find time.

Thanks, I think unify this would be good think in general.

laeubi (approved):
Looks reasonable to me

akurtakov:
Is there a reason to not merge this one now?

vogella:
> Is there a reason to not merge this one now?

I typically never merge without a green build, but in this case, I think merging is fine.

vogella:
Because it is only  a documentation change for AI

PR #2752, opened 2025-11-11
Add deprecation annotations to members of deprecated classes

HeikoKlare:
Adapt according to: https://eclipse.dev/eclipse/markdown/?f=news/4.38/jdt.md#changes-concerning-deprecation-warnings


HeikoKlare:
I did not find any (invalid) consumers in the SDK that would need to be adapted because of the added deprecation. And even if they existed, they would have been considered as accesses of deprecated methods/fields until the recent ECJ change.

akoch-yatta (approved):
Looks good to me 👍

PR #2753, opened 2025-11-11
Adjust selection filling the curve for CTabFolder

ShahzaibIbrahim:
When CTabFolder is selected there is a visible gap on top as well as on the top left where there is a curve (in RCP when theming is disabled). Also the image was too close to selection highlight which is adjusted here as well.

### Results

Before Changes at 200%:

image

After Changes at 200%:

image

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
@@ -14,6 +14,8 @@
  *******************************************************************************/
 package org.eclipse.swt.custom;
 
+import java.util.*;
+
 import org.eclipse.swt.*;
 import org.eclipse.swt.graphics.*;
 import org.eclipse.swt.widgets.*;
@@ -93,8 +95,8 @@ public class CTabFolderRenderer {
 	static final int BUTTON_FILL = SWT.COLOR_LIST_BACKGROUND;
 	static final int BORDER1_COLOR = SWT.COLOR_WIDGET_NORMAL_SHADOW;
 
-	static final int ITEM_TOP_MARGIN = 2;
-	static final int ITEM_BOTTOM_MARGIN = 2;
+	static final int ITEM_TOP_MARGIN = 3;
+	static final int ITEM_BOTTOM_MARGIN = 3;
 	static final int ITEM_LEFT_MARGIN = 4;
 	static final int ITEM_RIGHT_MARGIN = 4;
 	static final int INTERNAL_SPACING = 4;
@@ -1382,7 +1384,15 @@ public class CTabFolderRenderer {
 				if (parent.selectionHighlightBarThickness > 0 && parent.simple) {
 					Color previousColor = gc.getBackground();
 					gc.setBackground(item.getDisplay().getSystemColor(parent.shouldHighlight() ? SWT.COLOR_LIST_SELECTION : SWT.COLOR_WIDGET_DISABLED_FOREGROUND));
-					gc.fillRectangle(x + 1 /* outline */, parent.onBottom ? y + height - 1 - parent.selectionHighlightBarThickness : y + 1, width - 2 /*outline*/, parent.selectionHighlightBarThickness);
+
+					int[] highlightShape = Arrays.copyOf(shape, shape.length);
+					// Update Y coordinates in shape to apply highlight thickness
+					int thickness = parent.selectionHighlightBarThickness;
+					boolean onBottom = parent.onBottom;
+					int bottomY = y + height - 1;
+					int highlightY = onBottom ? bottomY - thickness : thickness + 1;
+					highlightShape[1] = highlightShape[3] = highlightShape[highlightShape.length - 1] = highlightShape[highlightShape.length - 3] = highlightY;
+					gc.fillPolygon(highlightShape);
 					gc.setBackground(previousColor);
 				}
 


bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
```diff
@@ -1402,6 +1395,35 @@ void drawSelected(int itemIndex, GC gc, Rectangle bounds, int state ) {
...
+
+					int thickness = parent.selectionHighlightBarThickness;
+
+					if (parent.onBottom) {
+						gc.fillRectangle(x + 1, y + height - 1 - thickness, width - 2, thickness);
+					} else {
+						int[] selectedHighlightShape = new int[] {
+								x, y + 2, // left corner curve start
+								x + 1, y + 1,
+								x + 2, y, // top-left curve peak
```
HeikoKlare: In my understanding, this will just draw a isosceles trapezoid and not the round corner that is used by the border. You can also see that when zooming in, such as here at 225%:
image

Can't we just reuse the shape that is already calculated for the background and border to draw the highlight analogously?

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
```diff
@@ -1419,7 +1441,7 @@ void drawSelected(int itemIndex, GC gc, Rectangle bounds, int state ) {
 				if (!parent.single && item.closeRect.width > 0) maxImageWidth -= item.closeRect.width + INTERNAL_SPACING;
 				if (imageBounds.width < maxImageWidth) {
 					int imageX = xDraw;
-					int imageY = y + (height - imageBounds.height) / 2;
+					int imageY = parent.selectionHighlightBarThickness + y + (height - imageBounds.height) / 2;
```
HeikoKlare: This will lead to the icons vertically jumping around:
jumping_icons

But the icons should be visually stable. To avoid that the icons touch the highlight line, can't we make the highlight line smaller?

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
```diff
@@ -1402,6 +1395,35 @@ void drawSelected(int itemIndex, GC gc, Rectangle bounds, int state ) {
 				gc.setForeground(borderColor);
 				gc.drawPolyline(shape);
 
+				// draw highlight marker of selected tab
+				if (parent.selectionHighlightBarThickness > 0 && parent.simple) {
```
HeikoKlare, line 1384: Shouldn't we still draw the highlight line before the border/outline? Otherwise the highlight will be drawn on top of it, thus hiding the outline.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
ShahzaibIbrahim: I tried using the already calculated shape but that is also not enough. I needed to make the changes on that as well and that's how i came to this calculation later. 

This is how it looks now when selected and when not selected. 
image

ShahzaibIbrahim:
I have made some more changes (with the help of co-pilot), resolving some comments. 

1. Images stays where they were.
2. Thinner highlight line.
3. Outline is drawn after (previous behavior)

This is the closest result I could achieve with all this calculations.

image

ShahzaibIbrahim:
I tried using the shape at first but using shape fills the whole tab because it defines the entire tab’s outline like this:
image

while computed polygon only defines the thin highlight region that we see on the top only. There is a lot of operation performed on shape as well which I don't understand and I don't know how can I extract out that specific curve from it. That's my reason to just draw a highlight on top of it.

HeikoKlare:
> while computed polygon only defines the thin highlight region that we see on the top only. There is a lot of operation performed on shape as well which I don't understand and I don't know how can I extract out that specific curve from it. That's my reason to just draw a highlight on top of it.

But the polygon is not different from the shape (or if it is, it's wrong) except that the area it fills is larger (as it fills the whole header instead of only the topmost part of it). Why should the operation performed on the polygon be more simple than the ones on the existing shape? For people that have to maintain the code, it will become even more difficult when there are two implementations (that somehow need to be consistent) that they need to understand rather than one. If the current code is too hard to comprehend, we should first refactor it to improve comprehensibility. But replicating an implementation because we do not understand the existing one it not a good idea. And without understanding the existing shape implementation, how could we be sure that the added polygon one fits to it at all in every scenario?

HeikoKlare:
I quickly hacked this demonstration (just adopted what is done for the background, removing all unnecessary stuff and adapted the bottom left and bottom right coordinates of the shape):
```java
// draw highlight marker of selected tab
if (parent.selectionHighlightBarThickness > 0 && parent.simple) {
	int[] shape2 = Arrays.copyOf(shape, shape.length);
	shape2[1] = 3;
	shape2[3] = 3;
	shape2[shape2.length - 1] = 3;
	shape2[shape2.length - 3] = 3;
	Color defaultBackground = item.getDisplay().getSystemColor(parent.shouldHighlight() ? SWT.COLOR_LIST_SELECTION : SWT.COLOR_WIDGET_DISABLED_FOREGROUND);
	xx = x;
	yy = parent.onBottom ? y -1 : y + 1;
	ww = width;
	hh = height;
	if (!parent.single && !parent.simple) ww += curveWidth - curveIndent;
	drawBackground(gc, shape2, xx, yy, ww, hh, defaultBackground, null, null, null, false);
}
```

Which would already give you:
image

This of course needs to be improved to deal with all the different cases (such as highlight at bottom/bottom) and has to be simplified, but the PoC makes me think even more that it should be quite easy to adapt and resue the existing shape to draw a properly fitting highlight.
Unfortunately, the line still touched the icons on higher monitor zooms, so the values probably need some tweaking, but they were just random numbers for testing anyway.

ShahzaibIbrahim:
> I quickly hacked this demonstration (just adopted what is done for the background, removing all unnecessary stuff and adapted the bottom left and bottom right coordinates of the shape):

I used your snippet as a template to draw the highlight. Which in my opinion now looks better in both cases (top and bottom). Attaching the screenshot in order to compare

image

ShahzaibIbrahim:
> Reusing the existing shape is definitely a step forward. But with the current state the highlight line now becomes much thinner than it was before. At 100% is looks like this:

These are the results after my latest changings. I think now thickness for the selection is ideal and in no zoom that I tested, the highlight line doesn't touches icon. 

image

HeikoKlare:
Please note that you have to use tabs with icons that start at the topmost pixel to have them touch the highlight line, such as the mentioned Manifest editor icon. With Java editor icons the behavior was always fine. This is how the current state looks at 150 with the Manifest editor:
image

The overall appearance of the highlight line is of course still better, but it still touches the icon.

ShahzaibIbrahim:
After some investigation with @akoch-yatta, we found out there wasn't enough space in ctab to accommodate highlighting (3pts) and hence images were touching the highlight when zoom > 100. We adjusted the size of Ctab which looks fine now with all kind of images and enough space for highlight. Here are the SS on relevant zoom levels. 

image

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
```diff
@@ -93,8 +95,8 @@ public class CTabFolderRenderer {
 	static final int BUTTON_FILL = SWT.COLOR_LIST_BACKGROUND;
 	static final int BORDER1_COLOR = SWT.COLOR_WIDGET_NORMAL_SHADOW;
 
-	static final int ITEM_TOP_MARGIN = 2;
-	static final int ITEM_BOTTOM_MARGIN = 2;
+	static final int ITEM_TOP_MARGIN = 3;
+	static final int ITEM_BOTTOM_MARGIN = 3;
```
HeikoKlare, line 99: Would it be sufficient to only have a higher top margin but keep the bottom one as is? Or do we want to increase the padding consistently or even also have issues there?

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
```diff
@@ -1380,10 +1382,23 @@ void drawSelected(int itemIndex, GC gc, Rectangle bounds, int state ) {
...
+
+					Color oldBackground = gc.getBackground();
+					Color highlightColor = item.getDisplay().getSystemColor(
+							highlight ? SWT.COLOR_LIST_SELECTION : SWT.COLOR_WIDGET_DISABLED_FOREGROUND);
+					gc.setBackground(highlightColor);
+
+					int bottomY = y + height - 1;
+					int highlightY = onBottom ? bottomY - thickness : thickness + 1;
+
+					int[] shape2 = Arrays.copyOf(shape, shape.length);
```
HeikoKlare: Please us more self-explaining names such as `highlightShape` instead of `shape2`.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
```diff
@@ -1380,10 +1382,23 @@ void drawSelected(int itemIndex, GC gc, Rectangle bounds, int state ) {
...
-					gc.fillRectangle(x + 1 /* outline */, parent.onBottom ? y + height - 1 - parent.selectionHighlightBarThickness : y + 1, width - 2 /*outline*/, parent.selectionHighlightBarThickness);
-					gc.setBackground(previousColor);
+					int thickness = parent.selectionHighlightBarThickness;
+					boolean onBottom = parent.onBottom;
+					boolean highlight = parent.shouldHighlight();
+
+					Color oldBackground = gc.getBackground();
+					Color highlightColor = item.getDisplay().getSystemColor(
+							highlight ? SWT.COLOR_LIST_SELECTION : SWT.COLOR_WIDGET_DISABLED_FOREGROUND);
+					gc.setBackground(highlightColor);
```
HeikoKlare: nit: couldn't we just keep this part of the code as is?

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
```diff
@@ -1380,10 +1382,23 @@ void drawSelected(int itemIndex, GC gc, Rectangle bounds, int state ) {
 
 				// draw highlight marker of selected tab
 				if (parent.selectionHighlightBarThickness > 0 && parent.simple) {
-					Color previousColor = gc.getBackground();
-					gc.setBackground(item.getDisplay().getSystemColor(parent.shouldHighlight() ? SWT.COLOR_LIST_SELECTION : SWT.COLOR_WIDGET_DISABLED_FOREGROUND));
-					gc.fillRectangle(x + 1 /* outline */, parent.onBottom ? y + height - 1 - parent.selectionHighlightBarThickness : y + 1, width - 2 /*outline*/, parent.selectionHighlightBarThickness);
-					gc.setBackground(previousColor);
+					int thickness = parent.selectionHighlightBarThickness;
+					boolean onBottom = parent.onBottom;
```
HeikoKlare: It would be easier to read if the variables are corehently defines where they are used. Currently, you define some variable for the layout here in the beginning, then some coloring stuff is done, then again some layouting variables are calculated.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
ShahzaibIbrahim: I wanted to increase the overall size height of the tab. Making top margin 4 would have worked. But I wanted to increase the padding from both side consistently.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
HeikoKlare: @ShahzaibIbrahim why was this comment marked as resolved? the code is still modified

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
HeikoKlare: Okay, let's see if this will have any unexpected side effects. I guess you meant "making top margin 3 would have worked", as an increase of one to the top margin should be sufficient in any case (ignoring the idea of consistent margings at top and bottom).

ShahzaibIbrahim: Yes, let's wait and see if there are any side effects.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
ShahzaibIbrahim: I missed the half part of it and missed the gc.setBackgound part. Now it should be okay.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
HeikoKlare: @ShahzaibIbrahim this comment still seems unaddressed. Why not calculate the position `bottomY`/`highlightY` exactly where its required for changing the shape coordinates?

Issue #2714, opened 2025-10-30
[GTK4] Get tests running on GitHub actions


jonahgraham:
We've had GitHub actions for GTK4 for a while now, but we've never had a complete test run. There are numerous reasons I have already identified:

- [x] callback leaks (such as https://github.com/eclipse-platform/eclipse.platform.swt/pull/2703)
- [x] memory leaks (such as https://github.com/eclipse-platform/eclipse.platform.swt/pull/2704)
- [x] known failing tests obscure regressions
- [x] missing dependencies (such as no webkit for GTK4)
- [x] permission problems with new isolation around apparmour + webkit
- [x] run tests on wayland (currently uses x11)

This is the umbrella bug for me to attach all the PRs I plan to do in this area to. I have started work on each of the areas in the above list in https://github.com/eclipse-platform/eclipse.platform.swt/pull/2690 while trying to fix https://github.com/eclipse-platform/eclipse.platform.swt/issues/2667 but decided that I needed to split this out to a separate task now. PRs incoming over the coming days on this topic.

jonahgraham:
My goal for this issue is that GTK4 tests are on for all PRs alongside GTK3 and other tests so that we can ensure other changes that are not explicitly GTK4 don't regress progress in GTK4.

akurtakov:
I have https://github.com/eclipse-platform/eclipse.platform.swt/pull/2600 as a test for this. Feel free to use/rebase it for testing purposes.

jonahgraham:
Progress update, nearly there on goal of:

> My goal for this issue is that GTK4 tests are on for all PRs alongside GTK3 and other tests so that we can ensure other changes that are not explicitly GTK4 don't regress progress in GTK4.

The original checklist items (completed ones checked in original comment now)

> - callback leaks 

- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2703

> - memory leaks

- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2745
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2704
- related: https://github.com/eclipse-platform/eclipse.platform.swt/pull/2686
 
I ran some other work locally and have attempted to automate the testing of memory leak detection, but as of yet don't have it stable enough to be generally useful and sure that
 
> known failing tests obscure regressions

[...]

PR #2757, opened 2025-11-11
Add todo JUnit5 tags to GTK4 and wayland tests that need work still

jonahgraham:
Issue #2714 is to get tests running on GTK4 and wayland. One challenge is that some tests are known failures because work has not been completed in some narrow areas.

Rather than add `assumeFalse(GTK4)` which would disable the tests all the time, this PR marks the tests that are still to be done with `@Tag` so that we can update the CI to exclude these tests until someone has had a chance to review them more fully.

A subsequent commit will exclude these tagged tests on GitHub actions, unless a suitable label is set on the PR.


jonahgraham:
Jenkins is stuck, but GHA passed fine so I am proceeding.

PR #2758, opened 2025-11-11
Improve enablement when running under GitHub actions/Jenkins

jonahgraham:
Previous commits aimed to disable tests when running on Jenkins and/or GitHub actions. However the conditions were too aggressive in some cases, causing tests to be skipped even when run locally by devs. This change aims to improve the enablements so that tests are only skipped on CI.

`isRunningOnContinousIntegration` was unused, hence removed as it is related to this PR.


jonahgraham:
Test "1 out of 4 runs failed: test_setUrl_remote_with_post (org.eclipse.swt.tests.junit.Test_org_eclipse_swt_browser_Browser_IE)" failed because bugzilla was down AFAICT. I am rebasing change so that we see better results on next try.

PR #2759, opened 2025-11-11
Disable some clipboard tests when running on wayland with no x11 available

jonahgraham:
The remote clipboard test requires X11 because Java Swing does not have Wayland backend until Project Wakefield is implemented, therefore these clipboard related tests are skipped.


jonahgraham:
macOS build failed with:

`
[ERROR] error: Error fetching URL: https://docs.osgi.org/javadoc/osgi.core/8.0.0/ (java.net.ConnectException: Operation 
timed out)
`

which is unrelated to this change, so I am proceeding.

PR #2762, opened 2025-11-11
Split the SWT workflow into a reusable part

jonahgraham:
On its own this PR does not really change anything, but it is a building block to future PRs that will optimize the workflow.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,92 @@
+# This workflow will build a Java project with Maven, and cache/restore any dependencies to improve the workflow execution time
+# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-maven
+
+name: "SWT Build"
+
+on:
+  workflow_call:
+    inputs:
+      runner:
+        description: "build.runs-on value"
+        type: string
+        required: true
+      java:
+        description: "Java version"
+        type: string
+        required: true
+      native:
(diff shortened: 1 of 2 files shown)


jonahgraham:
I believe this is uncontroversial so I am proceeding with it.

laeubi:
@jonahgraham I think we now need to have one workflow call for the tests comment so one gets one comment / OS otherwhise the test report will become completely unusable. e.g assume the following

1. PR 1 changes Windows
2. PR 2 changes Linux
3. The last master has changed Mac

Now the test results on master (baseline) will be the mac tests and it gets compared to the state of Linux+Windows on the other PRs ... son one of them gets merged and you get a totally different baseline now...

laeubi:
btw one can maybe get some inspiration form Equinox where we have some extra tests published form the TCK run:

https://github.com/eclipse-equinox/equinox/blob/master/.github/workflows/unit-tests.yml

But maybe for SWT consequently one should just have three workflows and three publisher workflows...

jonahgraham:
> [...] one workflow call [...]

As of this PR there is still only one workflow call - I assume you want these comments to apply to https://github.com/eclipse-platform/eclipse.platform.swt/pull/2763 where we call the workflow many times?

If you think the comment does apply here, I will revert this PR and we can start the review from scratch.

See also https://github.com/eclipse-platform/eclipse.platform.swt/pull/2764#issuecomment-3523577988

laeubi:
> If you think the comment does apply here, I will revert this PR and we can start the review from scratch.

Now worries, I just wanted to note it here but haven't seen your other PR

PR #2763, opened 2025-11-11
Use simpler build matrix, split by OS

jonahgraham:
Rather than a single matrix that covers all OSes, split them up so that enumerating different ones is a little easier.
    
This has three benefits:
    
1. `matrix.exclude` is easier to write in this format
2. Tools like `act` work with simple string matrix dimensions (see new comment at top of maven.yml)
3. conditional execution on file name pattens are easier to implement (see subsequent PR on this topic)
    

jonahgraham:
No further comments received on this step of the refactoring for #2714 so I am going to merge so I can proceed with getting GTK4 tests running.

Issue #2741, opened 2025-11-05
[GTK4] Memory (and threads) leak in Shell


jonahgraham:
**Describe the bug**

SWT on GTK4 does not dispose the native resources allocated by the Shell. 

**To Reproduce**

Run the below example with a process monitor running

```java
package org.eclipse.swt.snippets;

import org.eclipse.swt.internal.gtk.*;
import org.eclipse.swt.widgets.*;

public class Snippet1 {

	public static void main(String[] args) {
		System.out.println("GTK4: " + GTK.GTK4);
		int i = 0;
		while (i++ < 2500000) {
			System.out.println(i);
			final Display display = new Display();
			Shell shell = new Shell(display);
			shell.open(); // only needed on wayland
			while (display.readAndDispatch()) {
			}
			display.dispose();
			System.gc();
		}
	}
}
```

**Expected behavior**

No memory or thread growth after first few iterations - the GTK3 version achieves this.

**Screenshots**

(apologies for the picture of my screen, while tests were running I couldn't get screenshots working)

Image shows, the following java processes. The snippet above was running for about 3 minutes, or about 600 iterations.

1. My IDE
2. GTK4 x11 of above snippet
3. GTK3 x11 of above snippet
4. GTK4 wayland of above snippet

![Image](https://github.com/user-attachments/assets/68c80c43-806d-4153-855d-3ccf5c7f635a)

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [ ] Windows
- - [x] Linux
- - [ ] macOS 

2. Additional OS info (e.g. OS version, Linux Desktop, etc)

Ubuntu 25.10

---

In addition to the obvious problem of a memory leak, this is one of the blockers for getting https://github.com/eclipse-platform/eclipse.platform.swt/issues/2714 complete because the test java process keeps running out of memory and the GitHub runner gets killed for trying to allocate too much memory.

jonahgraham:
The error is that there is no call to `gtk_window_destroy` for Shells (at least in the code path of this snippet). The call to `gtk_window_new` in `Shell.createHandle` is never freed. 

For x11 the majority of memory/threads is only allocated when `gtk_widget_realize` is called on the shell at the end of `createHandle`. Therefore, if you run valgrind most of the used memory will be recorded in the call to `gtk_widget_realize`

For wayland the memory is only allocated when the shell is opened and visible. That is why the shell.open line is needed in snippet when running on wayland.


[...]

PR #2764, opened 2025-11-11
Skip build and test on OSes that have no changes

jonahgraham:
Many of the PRs on SWT are OS specific, this change only runs the build on OSes if there are changes that can affect that OS.
    
This is going to be more important with #2714 because potentially a number of new jobs will be run per PR and avoiding some of the extra time waiting will be a benefit to all.
    
The junit.yml has been updated to record tests per platform in addition to all tests. When not all platforms are run in a PR, only the respective platform results are reported as a comment. If all platforms are run, the cumulative results are reported as a comment.
    
### DIFF
--- a/.github/workflows/junit.yml
+++ b/.github/workflows/junit.yml
@@ -1,3 +1,21 @@
+# This is the workflow for SWT that collects and reports junit results
+
+# This script runs from the master branch all the time. This means if you update it in a
+# Pull Request, the update won't take effect until after you merge the PR, the PR itself
+# will *not* use the edits.
+# Therefore to really test a change to this script, push it to your fork's master branch
+# and then create a PR against your own master branch.
+
+# You can run this locally to test changes with the act command https://nektosact.com/
+# Steps:
+# 1. Get a PAT (https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens)
+#    Use fine grained one, needs Read access to artifact metadata, and metadata
+# 2. Get the URL of some artifacts to test against from a previous run of the build
+#    script (such as your last run on your fork's master branch). It looks like:
+#    https://api.github.com/repos/jonahgraham/eclipse.platform.swt/actions/runs/19980765555/artifacts
+# 3. Run the act command:
+#    act -W '.github/workflows/junit.yml' -s GITHUB_TOKEN=your_token --input artifacts_url=your_url
+
 name: Publish Unit Test Results
 
 on:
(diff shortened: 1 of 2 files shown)


jonahgraham:
> The minor negative of skipping tests conditionally is that the **Test Results** summary comment will report less runs that base commit.

I put together an example of what this looks like on my fork. I pushed this commit as my master HEAD, and then made three PRs:

- test results on master: https://github.com/jonahgraham/eclipse.platform.swt/actions/runs/19306877246/job/55218442382
- cocoa only change https://github.com/jonahgraham/eclipse.platform.swt/pull/12
- win32 only change https://github.com/jonahgraham/eclipse.platform.swt/pull/11
- other change, so all run https://github.com/jonahgraham/eclipse.platform.swt/pull/13

You end up with, for example, on win32 this output:

image

It shows number of tests/runs down compared to master because so many tests were skipped, but it still shows 0 failures as expected. Further down you can see the non-win32 runs are skipped

image

HeikoKlare:
I totally like the idea of skipping tests for OSes that are not affected by a change. I am just a bit concerned about the condensed test results posted by the bot being rendered less useful. Maybe we could check if we can adapt that the results to be separated by OS, such that you can easily and properly compare for the OS(es) of interest for the current PR.

Edit: just saw that Christoph already made a similar comment: https://github.com/eclipse-platform/eclipse.platform.swt/pull/2762#issuecomment-3522950874

jonahgraham:
I am going to split this PR out from the rest of them for #2741

laeubi:
> It shows number of tests/runs down compared to master because so many tests were skipped, but it still shows 0 failures as expected. Further down you can see the non-win32 runs are skipped

As mentioned on the other PR (sorry for confusion!) I think we then better should have three check runs for all three oss to prevent the problem you see with the test drops and then have one comment for each os e.g.

- Test Results GTK
- Test Results Win32
- Test Results Cocoa

beside that I'm wondering if it really pay off much to skip the tests, the Jenkins if often the slowest one and Github has completed most often much earlier anyways.

Also I'm wondering what happens with changes to the common parts of SWT that can change behavior on other OS of course as well.

HeikoKlare:
> Also I'm wondering what happens with changes to the common parts of SWT that can change behavior on other OS of course as well.

From my understanding, there is a classification for those changes that enables all OS builds via `needs.changes.outputs.other == 'true'`

> beside that I'm wondering if it really pay off much to skip the tests, the Jenkins if often the slowest one and Github has completed most often much earlier anyways.

Even though this is often the case, in my experience "often" rather means 60% of the cases than 90%, so avoiding unnecessary GHA runs will my in opinion still increase the average roundtrip time. And I see at least two further reasons to do this:
- Flakyness: we repeatedly run into the situation that some builds start to get flaky for different reasons; reducing the number of (unnecessary) builds that can prevent you from making progress (both in moving actual changes forward as well as in investigating the reason for builds not working properly) is beneficial.
- Resources: why should we utilize resources when we definitely know that the results are useless? It consumers material and energy without any benefit. You can argue that SWT builds are comparably small such that this does not pay off, but it's at least an additional point in combination with the other arguments.

jonahgraham:
@laeubi / @HeikoKlare thanks for your input - I agree and will see if I have time to bring this to a conclusion.

>> Also I'm wondering what happens with changes to the common parts [...]

> From my understanding, [...]

Yes, that is correct, anything that isn't in the platform specific changes (e.g `**/win32/**`) will mean everything gets run. What this means is any PR that includes/updates a test, snippet, or example change will cause all platforms to run.

jonahgraham:
I have now implemented test result reporting based on what has run. You can see what it will look like once this change is merged in the 4 examples I provide in my fork:

### Single platform only

* https://github.com/jonahgraham/eclipse.platform.swt/pull/26

Only the platform that was run generates a comment.

image


### Common code (or any change that triggers all the platforms)

* https://github.com/jonahgraham/eclipse.platform.swt/pull/27

Nothing is skipped in this case. Since linux + macos are matrix jobs, the total number of matrix jobs run is 6 (will be more once #2768 is merged). The comment added to the PR is the summary of all the jobs. 

image

If you click through the results you can also see the results of the individual platforms. This is especially useful when reviewing skipped jobs and a few of the other annotations.

image


### 2 platforms change

* https://github.com/jonahgraham/eclipse.platform.swt/pull/28

In this case the 3rd platform SWT Matrix job is skipped. This means that two comments are created, for the two platforms that did run.

image


### Only markdown changed (or anything in a category with no associated jobs)

* https://github.com/jonahgraham/eclipse.platform.swt/pull/29

In this case all the SWT Matrix jobs are skipped and no test results are reported.

image

jonahgraham:
> 4 examples I provide in my fork:

Actually 5 - I realized I missed an important case, what failures look like. 

### How errors report

In this example PR I have a ridiculous failures that throws an exception in the Display constructor meaning almost everything fails. This is what that looks like.

* https://github.com/jonahgraham/eclipse.platform.swt/pull/30

image

### Test report diff to base commit

When I made the screenshots in https://github.com/eclipse-platform/eclipse.platform.swt/pull/2764#issuecomment-3619575441 the base commit had not finished running yet, so they were not showing diffs to base commit. I updated all the linked example PRs, but I am only redoing this screenshot:

image

jonahgraham:
This is now ready for review.

jonahgraham:
I plan to merge this soon if there are no objections.

jonahgraham:
This is now merged. Thanks @HeikoKlare for the review - and hopefully it will just work so you never need to dig in deeper than you have already.

Once the master builds and existing PRs rebase to include this change then we should see the fruits of this labour. FWIW I did check to make sure there was likely a benefit, and there were plenty of commits to SWT that only changed one platform.

PR #2765, opened 2025-11-11
Exclude TODO tests by default, include them with runtodotests label

jonahgraham:
In https://github.com/eclipse-platform/eclipse.platform.swt/pull/2757 the idea of using `@Tag("gtk4-todo")` was introduced to tag tests that are known TODO as such. This commit adds todos to the exclude groups list when running such tests.
    
There are not actually any GTK3 x11 todos yet, so this commit on its own may not make sense until GTK4 + wayland tests are enabled in a subsequent commit
    
Apply the label runtodotests to the PR and the todo tests will not be excluded.
    
In draft until dependent PR https://github.com/eclipse-platform/eclipse.platform.swt/pull/2764 is merged.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,6 +23,11 @@ on:
         type: boolean
         required: false
         default: false
+      runtodotests:
+        description: "Run TODO tagged tests too (one of true, false)"
+        type: boolean
+        required: false
+        default: false
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


jonahgraham:
The runtodotests label is https://github.com/eclipse-platform/eclipse.platform.swt/issues?q=label%3Aruntodotests

PR #2766, opened 2025-11-11
[GTK] Fix ImageList.createPixbuf to not do wrong scaling

akurtakov:
Difference comes from Gtk 4 working in logical points while Cairo works in pixels. For proper size of the pixbuf scale to the original image size.
Change Label to use GtkPicture instead of GtkImage to prevent downscaling (revert of
https://github.com/eclipse-platform/eclipse.platform.swt/commit/4c13551c541f8dbffd219d62b1bf78135d79cd3b ) as SWT shows images in labels at their real size and not as "icon" size.


akurtakov:
About Before:
Screenshot From 2025-11-11 20-25-43
About After:
Screenshot From 2025-11-11 20-27-37

Onboarding image is also the correct size with this change.

PR #2768, opened 2025-11-11
Add Wayland to build matrix on GitHub actions

jonahgraham:
Add Wayland to build matrix on GitHub actions

Add GDK_BACKEND environment variable setting and corresponding entry in the Linux matrix entry for wayland and x11.

There are a total of 4 combinations possible of GTK version and backend:

- GTK3 x11 (uses Xvfb)
- GTK3 wayland (uses mutter headless)
- GTK4 x11 (uses Xvfb)
- GTK4 wayland (uses mutter headless)

Prior to this PR only the x11 variants were tested. With this PR GTK4 on Wayland is also tested. GTK3 on wayland is too unreliable under tests at the moment and is excluded.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -84,6 +84,9 @@ jobs:
         java: ['21']
         gtk: [gtk3, gtk4]
         gdk_backend: [x11, wayland]
+        exclude:
+          - gtk: gtk3
+            gdk_backend: wayland
     uses: ./.github/workflows/build.yml
     with:
       runner: ubuntu-latest


akurtakov:
On most recent Ubuntu/Fedora/etc. distros users get Wayland session by default thus I would say that even on Gtk3 Wayland has become the most common setup.

jonahgraham:
I am open to whatever collection of backends/gtk versions. For now this PR tests as before plus GTK4 on wayland. We could enable all 6 - I don't want to overwhelm the number of builds that run if it isn't valuable enough.

jonahgraham:
@akurtakov I split out the question of which combinations to exclude to a future commit - this PR tests all of the valid combinations. Comment 0 and commit message updated.

jonahgraham:
The JVM crashes are a corner case that I will workaround because IMHO they are bugs in GTK not in SWT: https://github.com/eclipse-platform/eclipse.platform.swt/issues/2853

jonahgraham:
I have given up - for now at least - getting tests reliable on gtk3+wayland. As it turns out GTK3 itself is only tested on x11. GTK4 is tested on x11, wayland, and broadway.

I have therefore excluded the gtk3/wayland combination and I open this up for review now.

jonahgraham:
I plan to merge this soon if there are no objections.

jonahgraham:
Only thing left is to update commit message to reflect final state. I will do that when I rebase it on top of https://github.com/eclipse-platform/eclipse.platform.swt/pull/2764

jonahgraham:
I updated commit messages and the original PR comment https://github.com/eclipse-platform/eclipse.platform.swt/pull/2768#issue-3613631502

PR #2709, opened 2025-10-30
[Refactor] Moving methods from Win32DPIUtil to DPIUtil for autoscaling

ShahzaibIbrahim:
Workbench needs few methods from DPIUtil that were only present for Win32. Keeping the functionality for Win32DPIUtil as is, just making them available for DPIUtil.

PR #2771, opened 2025-11-12
Limit autoscale mode to quarter and exact only

ShahzaibIbrahim:
swt applications will be limited to use autoscale modes quarter and exact only otherwise user will see an error message stating the incompatibility. Also the tests testing other autoscale values have been removed.

This PR will further affect: https://github.com/eclipse-platform/eclipse.platform.ui/pull/3475

- [x] Requires: https://github.com/eclipse-platform/eclipse.platform.swt/pull/2709

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
@@ -51,11 +51,15 @@ public class DPIUtil {
 
 	private static String autoScaleValue;
 
-	private static final Set<String> ALLOWED_AUTOSCALE_VALUES_FOR_UPDATE_ON_RUNTIME = Set.of("quarter", "exact", "false");
+	private static final Set<String> ALLOWED_AUTOSCALE_VALUES_FOR_UPDATE_ON_RUNTIME = Set.of("quarter", "exact");
+
 	/**
-	 * System property to enable to scale the application on runtime
-	 * when a DPI change is detected.
+	 * System property to enable to scale the application on runtime when a DPI
+	 * change is detected.
 	 * <ul>
+	 * <li>"force": the application is scaled on DPI changes even if an unsupported
+	 * value for swt.autoScale are defined. See allowed values
+	 * {@link #ALLOWED_AUTOSCALE_VALUES_FOR_UPDATE_ON_RUNTIME}.</li>
 	 * <li>"true": the application is scaled on DPI changes</li>
 	 * <li>"false": the application will remain in its initial scaling</li>
 	 * </ul>
@@ -140,25 +144,19 @@ static void setAutoScaleValue(String autoScaleValueArg) {
  * scaling.
  */
 public static boolean isSetupCompatibleToMonitorSpecificScaling() {
-	if (DPIUtil.getAutoScaleValue() == null) {
+	if (!"win32".equals(SWT.getPlatform())) {
 		return false;
 	}
-
-	if (ALLOWED_AUTOSCALE_VALUES_FOR_UPDATE_ON_RUNTIME.contains(DPIUtil.getAutoScaleValue().toLowerCase())) {
+	if (autoScaleValue == null || "force".equals(System.getProperty(SWT_AUTOSCALE_UPDATE_ON_RUNTIME))) {
 		return true;
 	}
-	try {
-		Integer.parseInt(DPIUtil.getAutoScaleValue());
-		return true;
-	} catch (NumberFormatException e) {
-		// unsupported value, use default
-	}
-	return false;
+	String value = autoScaleValue.toLowerCase(Locale.ROOT);
+	return ALLOWED_AUTOSCALE_VALUES_FOR_UPDATE_ON_RUNTIME.contains(value);
 }
 
 public static boolean isMonitorSpecificScalingActive() {
-	boolean updateOnRuntimeValue = Boolean.getBoolean (DPIUtil.SWT_AUTOSCALE_UPDATE_ON_RUNTIME);
-	return updateOnRuntimeValue;
+	String updateOnRuntimeValue = System.getProperty(SWT_AUTOSCALE_UPDATE_ON_RUNTIME);
+	return "force".equalsIgnoreCase(updateOnRuntimeValue) || "true".equalsIgnoreCase(updateOnRuntimeValue);
 }
 
 public static int pixelToPoint(int size, int zoom) {
--- a/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/ControlWin32Tests.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/ControlWin32Tests.java
@@ -111,7 +111,7 @@ class ControlWin32Tests {
 	}
 
 	@ParameterizedTest
-	@CsvSource({ "0.5, 100, true", "1.0, 200, true", "2.0, 200, true", "2.0, quarter, true", "0.5, 100, false",
+	@CsvSource({ "2.0, quarter, true", "0.5, 100, false",
 			"1.0, 200, false", "2.0, 200, false", "2.0, quarter, false", })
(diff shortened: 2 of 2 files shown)


bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
```diff
@@ -140,20 +147,20 @@ static void setAutoScaleValue(String autoScaleValueArg) {
...
 public static boolean isSetupCompatibleToMonitorSpecificScaling() {
-	if (DPIUtil.getAutoScaleValue() == null) {
+	// Per-monitor DPI supported only on Windows
+	if (!"win32".equals(SWT.getPlatform())) {
 		return false;
 	}
 
-	if (ALLOWED_AUTOSCALE_VALUES_FOR_UPDATE_ON_RUNTIME.contains(DPIUtil.getAutoScaleValue().toLowerCase())) {
+	// Default means: treat as "quarter" (compatible)
+	if (autoScaleValue == null || "true".equalsIgnoreCase(System.getProperty(SWT_AUTOSCALE_DISABLE_COMPATIBILITY_CHECK))) {
```
amartya4256: instead of comparing string, I think it makes it cleaner to directly parse the property as boolean, using `Boolean.parseBoolean(System.getProperty(SWT_AUTOSCALE_DISABLE_COMPATIBILITY_CHECK))`

HeikoKlare: Indeed. It usually makes sense to use `System.getProperty()` when you want to allow that no value for the property needs to be specified to enable it, e.g., `-Dswt.autoScale.force` without `=true`. Then you can do something like 
```java
Object disableCompatibilityCheck = System.getProperty(SWT_AUTOSCALE_DISABLE_COMPATIBILITY_CHECK);
if (disableCompatibilityCheck  != null && !"false".equalsIgnoreCase(disableCompatibilityCheck) { 
... 
}
```

ShahzaibIbrahim: I will use @amartya4256 suggestion in this regard as I want user to really -Dswt.autoScale.force=true if one wants to force the incompatible auto scale value.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
```diff
@@ -140,20 +147,20 @@ static void setAutoScaleValue(String autoScaleValueArg) {
...
  */
 public static boolean isSetupCompatibleToMonitorSpecificScaling() {
-	if (DPIUtil.getAutoScaleValue() == null) {
+	// Per-monitor DPI supported only on Windows
+	if (!"win32".equals(SWT.getPlatform())) {
 		return false;
 	}
 
-	if (ALLOWED_AUTOSCALE_VALUES_FOR_UPDATE_ON_RUNTIME.contains(DPIUtil.getAutoScaleValue().toLowerCase())) {
+	// Default means: treat as "quarter" (compatible)
```
HeikoKlare: What does this comment refer to?

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
```diff
@@ -51,7 +51,14 @@ public static Optional<AutoScaleMethod> forString(String s) {
...
 	private static String autoScaleValue;
 
-	private static final Set<String> ALLOWED_AUTOSCALE_VALUES_FOR_UPDATE_ON_RUNTIME = Set.of("quarter", "exact", "false");
+	private static final Set<String> ALLOWED_AUTOSCALE_VALUES_FOR_UPDATE_ON_RUNTIME = Set.of("quarter", "exact");
+
+	/**
+	 * System property that enforces to use autoScale value despite incompatibility
+	 * For e.g. Monitor-specific scaling with int200 autoscale value
+	 */
+	private static final String SWT_AUTOSCALE_DISABLE_COMPATIBILITY_CHECK = "swt.autoScale.force";
```
HeikoKlare: What if extend the existing property `swt.autoScale.updateOnRuntime` to support the value `force` in addition? The current name is in my opinion not meaningful, as it only has an effect when using monitor-specific scaling, so incorporating it into the according property could be reasonable.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
```diff
@@ -140,20 +147,20 @@ static void setAutoScaleValue(String autoScaleValueArg) {
  * scaling.
  */
 public static boolean isSetupCompatibleToMonitorSpecificScaling() {
-	if (DPIUtil.getAutoScaleValue() == null) {
+	// Per-monitor DPI supported only on Windows
```
HeikoKlare: I think this code is comprehensive even without the comment. If it should be preserved, please use common terminology such as "monitor-specific scaling" instead of adding "per-monitor DPI" as another term for the same thing (unfortunately we already have the "updateOnRuntime" for the property).

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
```diff
@@ -140,20 +147,20 @@ static void setAutoScaleValue(String autoScaleValueArg) {
...
-		Integer.parseInt(DPIUtil.getAutoScaleValue());
-		return true;
-	} catch (NumberFormatException e) {
-		// unsupported value, use default
-	}
-	return false;
+
+	String value = autoScaleValue.toLowerCase(Locale.ROOT);
+
+	// Compatible only if one of the known values
```
HeikoKlare:
```suggestion
```

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
ShahzaibIbrahim: Meant to show why autoScaleValue == null is compatible.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
ShahzaibIbrahim: We can do it but there will a separate effort how this property is set. Currently we pass the boolean value to this method `org.eclipse.swt.internal.Win32DPIUtils.setMonitorSpecificScaling(boolean)` to enable it. To also set the value "force", we need to either create another util method to set "force" and also update the implementation in `org.eclipse.swt.internal.DPIUtil.isMonitorSpecificScalingActive()` (when the value is set to force, does that mean monitor specific scaling is true?)

HeikoKlare: The method `Win32DPIUtils.setMonitorSpecificScaling(boolean)` is only used for testing purposes, isn't it? So no need to add any further functionality in that regard.
`DPIUtil.isMonitorSpecificScalingActive()` would also just return `true` when the system property is set to `force`, as that's obviously supposed to activate it.

In my understanding, we have three use cases that could be reasonably realized as follows:
- Disable monitor-specific scaling: system property is not set (or set to "false"), `isMonitorSpecificScalingActive()` returns true)
- Enable monitor-specific scaling: system property is set to "true", `isMonitorSpecificScalingActive()` returns true and startup will fail if `!isSetupCompatibleToMonitorSpecificScaling()`
- Enforce monitor-specific scaling: system property is set to "force", `isMonitorSpecificScalingActive()` returns true and result of `isSetupCompatibleToMonitorSpecificScaling()` is ignored

ShahzaibIbrahim: Thanks for clarification. I will adapt as mentioned.

PR #2778, opened 2025-11-14
Make "SWT Matrix Tests" badge the current one

jonahgraham:
As part of working on https://github.com/eclipse-platform/eclipse.platform.swt/pull/2764 I was considering how to display badges going forward to align with test results per platform. As it turns out the "SWT Matrix Tests" has been stuck on an arbitrary old badge for many years now.

This PR updates it, but as it has been wrong for a long time, perhaps removing it entirely makes more sense, but for now I update it. There was push back in https://github.com/eclipse-platform/eclipse.platform.swt/pull/1897 against removing it.


jonahgraham:
This hasn't really solved the problem - see https://github.com/eclipse-platform/eclipse.platform.swt/issues/2783

PR #2780, opened 2025-11-14
[GTK] Simplify _setBackground() and _setForeground() in Control widget

ptziegler:
The local variable "set" is always set to true in _setBackground() and is therefore redundant. This is because some widgets might have a default color, so it must always be set.

For _setForeground(), the initial value of the local variable is immediately overridden and can therefore be skipped as well.


ptziegler:
I'm merging this now, as it doesn't change the execution flow.

PR #2781, opened 2025-11-14
Improve ImageDataAtSizeProvider for zoom-based use case

HeikoKlare:
The ImageDataAtSizeProvider so far required a custom implementation of the existing getImageData(zoom) method, even though in many use cases the image data are only retrieved at specified size and not a specified zoom. This change extends the implementation such a default size can be provided and the size-based retrieval is automatically called with that default size when calling the zoom-based retrieval method. This relieves interface implementations from providing a custom implementation of the zoom-based retrieval method.

In addition, a default implementation for the zoom-based data retrieval and the default size specification is added, such that for those use cases where image data will not retrieved at zoom at all, dummy data are returned to reduce potential overheads of creating unnecessary data. This allows to treat the interface as a functional interface that can be implemented as a lambda, as only a single method for which an implementation is required remains.

Finally, the contract of the interface is extended to explain the intended use cases for the interface.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/ImageDataAtSizeProvider.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/ImageDataAtSizeProvider.java
@@ -12,17 +12,65 @@
  *******************************************************************************/
 package org.eclipse.swt.graphics;
 
+import org.eclipse.swt.internal.*;
+
 /**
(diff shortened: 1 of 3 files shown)


arunjose696:
The changes look good to me. It might be worth adding an additional test for ImageDataAtSizeProvider that overrides getDefaultSize()

HeikoKlare:
> It might be worth adding an additional test for ImageDataAtSizeProvider that overrides getDefaultSize()

Good point! I added an according test to those for the `Image` class.

PR #2785, opened 2025-11-14
[GTK4] Fix Control.redraw

akurtakov:
Control.redraw was not implemented due to no ability to invalidate a rectangle. This change marks the whole widget for redraw as this seems to be the only API that Gtk still provides.
Fixes SWTChart scrolling.


jonahgraham (approved):
+1 for RC1

Issue #2702, opened 2025-10-29
[GTK] Double click behaviour changed between 2025-03 and 2025-09


stephan-herrmann:
**Describe the bug**

As reported by @haubi in https://github.com/eclipse-egit/egit/issues/110 a double click into the EGit Staging View may fail to select the element under the cursor, and will instead operate on the previously selected tree element.

Since EGit doesn't seem to have any recent related changes fingers are pointing to SWT for changed behavior.

**To Reproduce**

* Have a git staging view with several staged files
* Select the first file
* Select any editor or other view
* Perform a double click **on the second file** in the staging view

**Expected behavior**
A compare editor should open for the second file.

**Actual behavior**
Compare editor will open for the **first file**.

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [ ] Windows
- - [x] Linux
- - [ ] macOS 

EGit committer tried to reproduce on macOS and failed.

2. Additional OS info (e.g. OS version, Linux Desktop, etc)

I'm seeing it on Ubuntu 24.04.3 with
```
org.eclipse.swt.internal.gtk.theme=Breeze:dark
org.eclipse.swt.internal.gtk.version=3.24.41
org.eclipse.swt.internal.webkitgtk.version=2.48.7
osgi.ws=gtk
```
and
```
GTK2_RC_FILES=/etc/gtk-2.0/gtkrc:/home/stephan/.gtkrc-2.0:/home/stephan/.config/gtkrc-2.0
GTK_OVERLAY_SCROLLING=0
GTK_RC_FILES=/etc/gtk/gtkrc:/home/stephan/.gtkrc:/home/stephan/.config/gtkrc
SWT_GTK3=1
SWT_GTK4=0
```

3. JRE/JDK version

 java.version=24.0.1

**Version since**
Reported in https://github.com/eclipse-egit/egit/issues/110:
* working: SWT 3.129.0.v20250221-1734
* broken: SWT 3.131.0.v20250820-1556

**Workaround (or) Additional context**

One additional click is required to first set focus to the staging view.

See https://github.com/eclipse-egit/egit/issues/110

ptziegler:
Whatever is causing this has to be very likely be on the EGit side, not the SWT side. Specifically something introduced with 7.4.

This is the 2025-06 with EGit 7.3, where the staging view behaves as expected:

[...]

PR #2786, opened 2025-11-14
[GTK] Don't invalidate style context when setting background color #2702

ptziegler:
When setting the background color while inside an SWT.Activate event, pending events are not processed.

The gtk_style_context_invalidate() method is deprecated in GTK3 and doesn't have a GTK4 counterpart, as GTK invalidates the style context automatically.

With b7dee8a4521e8dd010d6bbb84d7533febf71b7db, this method call was already removed, but later reverted due to test failures in the Eclipse Platform. Rather than removing all references, this change only does so for the call to setBackground().

To reproduce, execute the Snippet388. When clicking on a tree/table item, the item should be selected. This is the current behavior on Windows and MacOS, but not on Linux.

Closes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2702

### DIFF
--- a/tests/org.eclipse.swt.tests.gtk/ManualTests/org/eclipse/swt/tests/gtk/snippets/Issue2702_DoubleClickBehavior.java
+++ b/tests/org.eclipse.swt.tests.gtk/ManualTests/org/eclipse/swt/tests/gtk/snippets/Issue2702_DoubleClickBehavior.java
@@ -1,0 +1,65 @@
+/*******************************************************************************
+ * Copyright (c) 2025 Patrick Ziegler and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Patrick Ziegler - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.swt.tests.gtk.snippets;
+
+import org.eclipse.swt.*;
+import org.eclipse.swt.graphics.*;
+import org.eclipse.swt.layout.*;
+import org.eclipse.swt.widgets.*;
+
+/**
+ * This snippet demonstrates a missed selection event when updating the
+ * background color within an SWT.Activate event. To reproduce, alternative
+ * between the table and the tree. On each click, the clicked tree/table item
+ * should be selected.
+ *
+ * @see <a href="https://github.com/eclipse-platform/eclipse.platform.swt/issues/2702">Issue 2702</a>
+ */
+public class Issue2702_DoubleClickBehavior {
+	private static Color ACTIVE = new Color(255, 255, 255);
+	private static Color INACTIVE = new Color(248, 248, 248);
+
+	public static void main(String[] args) {
+		Shell shell = new Shell();
+		shell.setLayout(new GridLayout(2, true));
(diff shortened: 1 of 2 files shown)


ptziegler:
This is definitely a risky change, but perhaps it inspires someone.

laeubi:
@ptziegler can you explain why you think it is risky and what the invalidate do? Will it simply "repaint" the control somehow? From what you describe it seems one only want to process such invalidate after all events are processed, so maybe one should check if the event queue is empty and if not schedule it (again) async?

ptziegler:
> can you explain why you think it is risky 

By invalidating the style context asynchronously, I'm putting a lot of faith into the assumption that the context pointer is still valid at that point. The check whether the widget is disposed is just a workaround. Also as pointed out in https://github.com/eclipse-platform/eclipse.platform.swt/issues/2702#issuecomment-3536171536, invalidating the context asynchronously doesn't make much sense when it is automatically invalidated by GTK. So it will likely cause the same issues that were caused when the line was simply removed.

> and what the invalidate do?

I'm really not an expert on GTK, but I think it simply "applies" the CSS style. Meaning updating the colors and bounds of the widget. In some of the tests that failed because of this, you got assertion errors like:

```
Warning: Button {Always r&un in background}
	Actual Width -> 166
	Recommended Width -> 168
```

Which sounds like the style was not updated in time.

> From what you describe it seems one only want to process such invalidate after all events are processed, so maybe one should check if the event queue is empty

At least as simple `display.readAndDispatch()` doesn't work. From my experience, GTK is also very finicky once you're inside an event. There more funny stuff you try to do (like processing a selection-event within an activate-event), the more I expect other things to break. I think https://github.com/eclipse-platform/eclipse.platform.swt/issues/697 is such an example where you can get a crash when restyling during a `gtk_size_allocate`.

Perhaps styling should be done asynchronously after activation? But that probably introduces a completely different set of problems...

akurtakov:
I am more and more convinced that dropping invalidate (no **immediate** styling effect) in next cycle and see what(if anything) breaks. Gtk 4.x no longer even have the invalidate so there must be a way to work without it, and newer Gtk versions (since the time invalidate call was introduced) should be less fragile to styling.

tests/org.eclipse.swt.tests.gtk/ManualTests/org/eclipse/swt/tests/gtk/snippets/Issue2702_DoubleClickBehavior.java
```diff
@@ -0,0 +1,65 @@
...
+ *******************************************************************************/
+package org.eclipse.swt.snippets;
+
+import org.eclipse.swt.*;
+import org.eclipse.swt.graphics.*;
+import org.eclipse.swt.layout.*;
+import org.eclipse.swt.widgets.*;
+
+/**
+ * This snippet demonstrates a missed selection event when updating the
```
akurtakov, line 22: Snippets are something adopters to use when they familiarize with SWT or need examples how to achieve smth. Reproducers for bugs are not smth we should make widely visible esp. considering this is GTK backend specific. Please move it to https://github.com/eclipse-platform/eclipse.platform.swt/tree/master/tests/org.eclipse.swt.tests.gtk/ManualTests/org/eclipse/swt/tests/gtk/snippets

ptziegler: Done

akurtakov:
Merging so it has maximum amount of test period.

PR #2790, opened 2025-11-17
Show all system colors in ControlExample

akurtakov:


akurtakov:
IE test failure on Windows is unrelated to the change. It's examples change. Merging.

PR #2795, opened 2025-11-19
[Refactoring] Streamline calls to Image#adaptImageDataIfDisabledOrGray()

ShahzaibIbrahim:
Enabling a consistent pattern to apply disablement or graying of an Image. They are applied whenever a handle is being created from init method. Ensuring streamline usage and avoid double calls to this method.

---

### How to test
- Run the `Snippet382` with following VM Argument: `swt.autoScale.updateOnRuntime=true`
- Move the shell between different zoom levels (100,125,150,175,200,250)
- The snippet tests different kind of image providers to create an Image.
- Expected result would be that we don't lose disabled or grayed image attributes. 

image

---

This adaption is made on the basis of this comment: https://github.com/eclipse-platform/eclipse.platform.swt/pull/2737#issuecomment-3517881120

Instead of calling the adaptImageDataIfDisabledOrGray method from init method (which could be executed even when temporary handle is created) we will now call it only when data is loaded for the first time. So instead of calling loadImageData, all providers must call loadImageDataWithGrayOrDisablement and also provide loadImageData implementation of its own. In this way we can achieve one common place to call adaptImageDataIfDisabledOrGray from.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
@@ -172,8 +172,7 @@ public final class Image extends Resource implements Drawable {
 		private Optional<ImageHandle> createHandleAtExactSize(int width, int height) {
 			Optional<ImageData> imageData = imageProvider.loadImageDataAtExactSize(width, height);
 			if (imageData.isPresent()) {
-				ImageData adaptedData = adaptImageDataIfDisabledOrGray(imageData.get());
-				ImageHandle imageHandle = init(adaptedData, -1);
+				ImageHandle imageHandle = init(imageData.get(), -1);
 				return Optional.of(imageHandle);
 			}
 			return Optional.empty();
@@ -186,9 +185,8 @@ public final class Image extends Resource implements Drawable {
 			int imageZoom = DPIUtil.getZoomForAutoscaleProperty(Math.max(imageZoomForWidth, imageZoomForHeight));
 			ImageHandle bestFittingHandle = zoomLevelToImageHandle.get(imageZoom);
 			if (bestFittingHandle == null) {
-				ImageData bestFittingImageData = imageProvider.loadImageData(imageZoom).element();
-				ImageData adaptedData = adaptImageDataIfDisabledOrGray(bestFittingImageData);
-				bestFittingHandle = init(adaptedData, -1);
+				ImageData bestFittingImageData = imageProvider.loadImageDataWithGrayOrDisablement(imageZoom).element();
+				bestFittingHandle = init(bestFittingImageData, -1);
 			}
 			return bestFittingHandle;
 		}
@@ -333,21 +331,12 @@ public Image(Device device, Image srcImage, int flag) {
 			}
 			break;
 		}
-		case SWT.IMAGE_DISABLE: {
-			for (ImageHandle imageHandle : srcImage.zoomLevelToImageHandle.values()) {
-				Rectangle rect = imageHandle.getBounds();
-				ImageData data = srcImage.getImageData(imageHandle.zoom);
-				ImageData newData = applyDisableImageData(data, rect.height, rect.width);
-				init (newData, imageHandle.zoom);
-			}
-			break;
-		}
+		case SWT.IMAGE_DISABLE:
 		case SWT.IMAGE_GRAY: {
 			for (ImageHandle imageHandle : srcImage.zoomLevelToImageHandle.values()) {
-				Rectangle rect = imageHandle.getBounds();
-				ImageData data = srcImage.getImageData(imageHandle.zoom);
-				ImageData newData = applyGrayImageData(data, rect.height, rect.width);
-				init (newData, imageHandle.zoom);
+				ImageData imageData = imageHandle.getImageData();
+				imageData = adaptImageDataIfDisabledOrGray(imageData ,this.styleFlag);
+				init(imageData, imageHandle.zoom);
 			}
 			break;
 		}
@@ -698,9 +687,9 @@ public Image(Device device, ImageGcDrawer imageGcDrawer, int width, int height) 
 	init();
 }
 
-private ImageData adaptImageDataIfDisabledOrGray(ImageData data) {
+private static ImageData adaptImageDataIfDisabledOrGray(ImageData data, int styleFlag) {
 	ImageData returnImageData = null;
-	switch (this.styleFlag) {
+	switch (styleFlag) {
 		case SWT.IMAGE_DISABLE: {
 			ImageData newData = applyDisableImageData(data, data.height, data.width);
 			returnImageData = newData;
@@ -726,7 +715,7 @@ void init() {
 	this.isInitialized = true;
 }
 
-private ImageData applyDisableImageData(ImageData data, int height, int width) {
+private static ImageData applyDisableImageData(ImageData data, int height, int width) {
 	PaletteData palette = data.palette;
 	ImageData newData = new ImageData(width, height, 32, new PaletteData(0xFF, 0xFF00, 0xFF0000));
 	newData.alpha = data.alpha;
@@ -775,7 +764,7 @@ private ImageData applyDisableImageData(ImageData data, int height, int width) {
 	return newData;
 }
 
-private ImageData applyGrayImageData(ImageData data, int pHeight, int pWidth) {
+private static ImageData applyGrayImageData(ImageData data, int pHeight, int pWidth) {
 	PaletteData palette = data.palette;
 	ImageData newData = data;
 	if (!palette.isDirect) {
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -333,21 +331,10 @@ public Image(Device device, Image srcImage, int flag) {
...
-		}
+		case SWT.IMAGE_DISABLE:
 		case SWT.IMAGE_GRAY: {
 			for (ImageHandle imageHandle : srcImage.zoomLevelToImageHandle.values()) {
-				Rectangle rect = imageHandle.getBounds();
-				ImageData data = srcImage.getImageData(imageHandle.zoom);
-				ImageData newData = applyGrayImageData(data, rect.height, rect.width);
-				init (newData, imageHandle.zoom);
+				srcImage.getImageData(imageHandle.zoom);
 			}
```
ShahzaibIbrahim: Now we only need to reload the data for other handle if exist in order to apply gray or disable image data.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2765,7 +2755,7 @@ protected ImageHandle newImageHandle(ZoomContext zoomContext) {
 			drawer.drawOn(gc, width, height);
 			ImageData imageData = image.getImageData(targetZoom);
 			drawer.postProcess(imageData);
-			ImageData newData = adaptImageDataIfDisabledOrGray(imageData);
+			ImageData newData = adaptImageDataIfDisabledOrGray(imageData, styleFlag);
```
ShahzaibIbrahim, line 2760: I still have to keep it because a new image is created here with plainImageDataProvider which loads the data directly from handle. I am open to suggestions here.

ShahzaibIbrahim:
@akoch-yatta @HeikoKlare  Please review the PR and comments, It might not be in final shape but here I could use some suggestions.

ShahzaibIbrahim:
@akoch-yatta can you please have a look/review if you got time?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1456,7 +1443,7 @@ private static ImageData directToDirect(ImageData src, int newDepth, PaletteData
 
 private record HandleForImageDataContainer(int type, ImageData imageData, long[] handles) {}
 
-private static HandleForImageDataContainer init(Device device, ImageData i) {
+private static HandleForImageDataContainer init(Device device, ImageData i, int styleFlag) {
```
HeikoKlare, line 1448: Why is this necessary? The new parameter does not seems to be used.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2736,7 +2725,7 @@ protected long configureGCData(GCData data) {
 
 	@Override
 	ImageData newImageData(ZoomContext zoomContext) {
-		return loadImageData(zoomContext.targetZoom).element();
+		return loadImageDataWithGrayOrDisablement(zoomContext.targetZoom).element();
```
HeikoKlare: Why is this necessary? The `loadImageData` finally goes to `newImageHandle` which already applied the gray/disablement, doesn't it?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -172,8 +172,7 @@ private boolean isReusable(int width, int height) {
 		private Optional<ImageHandle> createHandleAtExactSize(int width, int height) {
 			Optional<ImageData> imageData = imageProvider.loadImageDataAtExactSize(width, height);
 			if (imageData.isPresent()) {
-				ImageData adaptedData = adaptImageDataIfDisabledOrGray(imageData.get());
-				ImageHandle imageHandle = init(adaptedData, -1);
+				ImageHandle imageHandle = init(imageData.get(), -1);
```
HeikoKlare, line 175: Why don't we apply gray/disablement here anymore?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -333,21 +331,10 @@ public Image(Device device, Image srcImage, int flag) {
...
-			break;
-		}
+		case SWT.IMAGE_DISABLE:
 		case SWT.IMAGE_GRAY: {
 			for (ImageHandle imageHandle : srcImage.zoomLevelToImageHandle.values()) {
-				Rectangle rect = imageHandle.getBounds();
-				ImageData data = srcImage.getImageData(imageHandle.zoom);
-				ImageData newData = applyGrayImageData(data, rect.height, rect.width);
-				init (newData, imageHandle.zoom);
+				srcImage.getImageData(imageHandle.zoom);
```
HeikoKlare: Wny do we remove the gray/disabled adaptation and, especially, the `init` call here? Won't this lead to the new image not being initalized with appropriate handles anymore?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2271,21 +2262,20 @@ protected Rectangle getBounds(int zoom) {
...
-			return newImageHandle(zoomContext).getImageData();
+			return new ElementAtZoom<> (newImageHandle(new ZoomContext(targetZoom)).getImageData(), targetZoom);
 		}
-		return getScaledImageData(targetZoom);
-	}
-
-	@Override
-	protected ElementAtZoom<ImageData> loadImageData(int zoom) {
-		return getClosestAvailableImageData(zoom);
+		return new ElementAtZoom<> (getScaledImageData(targetZoom), targetZoom);
```
HeikoKlare: Why is this semantics change of the methods necessary? In case of a `PlainImageProviderWrapper` there is no need to apply any gray/disabled if we just create an empty handle. And if we rescale an existing handle, that one should already have gray/disablement applied.

HeikoKlare (commented):
Can you please explain the underlying pattern where the application of gray/disablement is supposed to be done now? The PR description states:
> In this way we can achieve one common place to call adaptImageDataIfDisabledOrGray from.

But I see calls of that method from at least:
- initializeHandleFromSource
- newImageData
- getOrCreateImageHandleAtClosestSize

To assess if this is a good concept/systematics where to apply gray/disablement, it would be helpful to have a description of that concept.

In addition to that, the PR seems to contain unnecessary changes (e.g., addition of unused parameters, duplicate application of adaptImageDataIfDisabledOrGray) and also seems to remove crucial code (handle initialization in copy constructor).

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
ShahzaibIbrahim: Because data will already be loaded with disablement/gray properties.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
ShahzaibIbrahim: Since the logic to apply gray/disablement is moved to whenever image data is loaded, we don't need to apply adaptions here (also the init method).

HeikoKlare: This is an image constructor, which copied/adapts the handles of the original image. Not doing that will leave the new image empty. How should that work, e.g., in case you have an image based on an existing handle? The new Image instance with GRAY/DISABLED flag would simply be empty.

ShahzaibIbrahim:
> Can you please explain the underlying pattern where the application of gray/disablement is supposed to be done now? The PR description states:
> 
> > In this way we can achieve one common place to call adaptImageDataIfDisabledOrGray from.
> 
> But I see calls of that method from at least:
> 
> * initializeHandleFromSource
> * newImageData
> * getOrCreateImageHandleAtClosestSize
> 
> To assess if this is a good concept/systematics where to apply gray/disablement, it would be helpful to have a description of that concept.
> 
> In addition to that, the PR seems to contain unnecessary changes (e.g., addition of unused parameters, duplicate application of adaptImageDataIfDisabledOrGray) and also seems to remove crucial code (handle initialization in copy constructor).

I tried running **Snippet382** with the updated examples, and it now fails for **ImageDataProvider** when an `imageDataAtSizeProvider` is passed. In this case we must call `loadImageDataAtExactSize` instead of `loadImageData`, which means the disablement logic is never applied. The only fix would be to apply disablement manually at this point—bringing us back to the original issue: how to **streamline** disablement handling.

The original idea was to trigger disablement for all types of `ImageWrapper` when the underlying data is loaded for the first time. However, different wrappers load their initial data through different code paths. The conclusion is that there is **no single common entry point** where we can reliably apply disablement for every case.

At the moment, we have two options:

1. **Keep the current behavior** as it is.
2. **Apply disablement during handle initialization**, as shown here: [PR 2737](https://github.com/eclipse-platform/eclipse.platform.swt/pull/2737), though this approach introduces its own issues, described in this comment: https://github.com/eclipse-platform/eclipse.platform.swt/pull/2737#issuecomment-3517881120

@HeikoKlare

Issue #2794, opened 2025-11-19
[GTK] Transparent combo popups


akurtakov:
Combo boxes popup is transparent when PARENT_BACKGROUND is set as Control's implementation relies on transparency to implement it on GTK.
One instance of the problem has been https://github.com/eclipse-pde/eclipse.pde/pull/2094 .
Other projects put workaround for it like https://github.com/eclipse-swtchart/swtchart/issues/347 and https://github.com/eclipse-chemclipse/chemclipse/issues/1372

laeubi:
@akurtakov the same happen in the E4 Model editor, very annoying!

vogella:
Also visible in the Combo of Eclipse Copilot plug-in

PR #2797, opened 2025-11-20
[Gtk] Ignore transparency when setting Combo popup color

akurtakov:
Ensures no transparency is set for Combo's popup (which generally makes UI harder to read).

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2794

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Combo.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Combo.java
@@ -2695,6 +2695,17 @@ void updateCss() {
 	// Deal with background
 	if (background != null) {
 		final String colorString = display.gtk_rgba_to_css_string(background);
+		GdkRGBA menuBackground = new GdkRGBA();
+		menuBackground.red = background.red;
+		menuBackground.green = background.green;
+		menuBackground.blue = background.blue;
+		menuBackground.alpha = 1.0;
+		/* Ensures that the popup menu is not transparent and as a result unreadable
+		 * This way effects like "light" transparency (alpha > 0.85) can not be achieved.
+		 * Having any transparency of the popup is generally unwanted as it hurts visibility so
+		 * if such a feature is usable to anyone it would need new dedicated API.
+		 */
+		final String menuColorString = display.gtk_rgba_to_css_string(menuBackground);
 
 		/*
 		 * Use 'background:' instead of 'background-color:' to also override
@@ -2702,7 +2713,7 @@ void updateCss() {
 		 * 'background-image:' for 'GtkToggleButton' used in READ_ONLY combo.
 		 */
 		css.append("* {background: " + colorString + ";}\n");
-		css.append("menu {background: " + colorString + ";}\n");
+		css.append("menu {background: " + menuColorString + ";}\n");
 
 		/*
 		 * Setting background color for '*' also affects selection background,


akurtakov:
The fix should be quite safe so if it gets needed approvals it could be shipped in RC1 even.

vogella (approved):
Looks save and the current behavior is super annoying

vogella:
FYI @jdneo  this should also fix this issue in Copilot

jdneo:
@vogella Thank you! :)

iloveeclipse:
@akurtakov : is there some SWT snippet where one can check the fix?

laeubi:
copilot says:

Issues Found:

- Line 2708 - Spacing: Missing space after the assignment operator in final String menuColorString =display.... Should be = display (with space) to match the formatting on line 2697 and maintain consistent code style throughout the file.

-  Line 2704 - Comment clarity: The comment states "alpha > 0.85%" which is ambiguous. Alpha values in GdkRGBA range from 0.0 to 1.0 (or 0% to 100% when expressed as percentages). The notation "0.85%" would represent 0.0085 as a decimal value, which doesn't align with the intent. This should be either "85%" or "0.85" (without the percent symbol) to accurately represent high transparency values.

**Note on Test Coverage**: While there are existing tests for setBackground with alpha values in Test_org_eclipse_swt_widgets_Combo.java (e.g., test_setBackgroundAlphaDropDownCombo), these tests verify that the background color getter returns the correct value but don't specifically test the CSS rendering of the popup menu. The new behavior that forces alpha=1.0 for the menu could benefit from a test, though this would likely require inspecting the generated CSS or visual validation which may be challenging to automate.

akurtakov:
@iloveeclipse I tested with:
```java
import org.eclipse.swt.*;

public class TransparentComboExample {

        shell.setLayout(new FillLayout(SWT.VERTICAL));

        Composite parent = new Composite(shell, SWT.NONE);
        parent.setLayout(new RowLayout(SWT.VERTICAL));


        Combo combo = new Combo(parent, SWT.READ_ONLY);
        for (int i = 1; i <= 5; i++) {
            combo.add("Item " + i);
        }
        combo.select(0);
        combo.setBackground(new Color(240, 240, 240, 0));

        Label l = new Label(parent, SWT.NONE);
        l.setText("Some test long ");
        l = new Label(parent, SWT.NONE);
        l.setText("Some test long ");
        l = new Label(parent, SWT.NONE);
        l.setText("Some test long ");

        shell.pack();
        shell.setSize(500, 500);
        shell.open();

        while (!shell.isDisposed()) {
                display.sleep();
            }
        }
        display.dispose();
    }
}
```
but real testing comes with platform css styling as it really plays with colors.

akurtakov:
@laeubi review comments fixed

iloveeclipse:
> but real testing comes with platform css styling as it really plays with colors.

OK, I don't see any differences (and no transparency problems) with/without patch if starting your snippet. I'm on RHEL 9.4 / X11 / gtk3-3.24.31-5.el9.x86_64. 
Any hint where I can see this in the SDK?

akurtakov:
@iloveeclipse if you revert https://github.com/eclipse-pde/eclipse.pde/pull/2094 you should see https://github.com/eclipse-pde/eclipse.pde/pull/2094#issuecomment-3487098682
Or in 
image

akurtakov:
FWIW, on Fedora 43 there is no pure X11 session anymore(thus can't test) but the problem is visible with XWayland.

iloveeclipse:
> FWIW, on Fedora 43 there is no pure X11 session anymore(thus can't test) but the problem is visible with XWayland.

That explains I can't reproduce it (I've tried with your PDE example & reverted workaround). So it seem Wayland issue only.

akurtakov:
As there have been two +1 I'm merging this one and starting new I-build.

akurtakov:
Verified that E4 tools problem is gone with in https://download.eclipse.org/eclipse/downloads/drops4/I20251120-0400/

PR #2799, opened 2025-11-20
[Win32] Properly convert event coordinates in Tracker

HeikoKlare:
The coordinates for mouse move events in Tracker are currently just scaled via pixel/point conversion methods. However, those are display coordinates which need to take the coordinate system specifics into account and thus have to be converted via the display methods for translating coordinates.

This change adapts the implementation accordingly.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tracker.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tracker.java
@@ -846,7 +846,11 @@ public void setRectangles (Rectangle [] rectangles) {
 	if (rectangles == null) error (SWT.ERROR_NULL_ARGUMENT);
 	Rectangle [] rectanglesInPixels = new Rectangle [rectangles.length];
 	for (int i = 0; i < rectangles.length; i++) {
-		rectanglesInPixels [i] = Win32DPIUtils.pointToPixel (rectangles [i], getZoom());
+		if (parent != null) {
+			rectanglesInPixels [i] = Win32DPIUtils.pointToPixel (rectangles [i], getZoom());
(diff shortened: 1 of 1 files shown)


akoch-yatta:
@HeikoKlare Are there any other test scenarios except the GEF Flyout Palette you are aware of that are affected by this change (positively)?

HeikoKlare:
@akoch-yatta you should be able to see the improvement with `Snippet31`:
- Start the snippet with `System.setProperty("swt.autoScale.updateOnRuntime", "true")`
- Move it to a secondary monitor with a zoom different from primary monitor zoom
- Drag out of the shell client area (so that the tracker is instantiated)

Without this fix you should see an offset rectangle whereas with this change it should be placed properly.

akoch-yatta:
I got this when I tested on a 175% primary monitor with autoscaling disabled in GEF - with autoscaling enabled it works fine:

https://github.com/user-attachments/assets/2caf242c-05af-43ad-a097-c72807806169

That is an issue in the Display:translateXXX methods, so it is nothing that should be part of this PR, correct?

HeikoKlare:
> I got this when I tested on a 175% primary monitor with autoscaling disabled in GEF - with autoscaling enabled it works fine:

You mean the wrong sizes of the rectangles? In my understanding that's rooted in how GEF calculates the sizes based on what sizes are provided with a mixture of SWT-auto-scale enabled and SWT-auto-scale disabled controls. If I remember correctly, it's something that we can fix (in a backward-compatible way) in GEF and I think this was the change for it but I did not propose it as a PR yet: https://github.com/vi-eclipse/gef-classic/commit/033744da24aab5e8092aa26d4493058cd422f3b9
This is the branch containing the change: https://github.com/vi-eclipse/gef-classic/commits/flyout-tracker-size/

PR #2801, opened 2025-11-20
Handle Checkbox ImageList scaling in Table in win32

amartya4256:
This PR handles scaling of ImageList in a table on DPI_CHANGE event by destroying the current imagelist and recalculating the height of the table item and then creating a new imagelist using the new size.

### How to test:
- Start a table with checkboxes at 250% zoom
- Move to a different zoom (more and less than 250)
- The spacing between the table items should be consistent.

**Before:**
image

**After:**
image

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Table.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Table.java
@@ -7336,6 +7336,20 @@ LRESULT wmNotifyToolTip (NMTTCUSTOMDRAW nmcd, long lParam) {
 	return null;
 }
 
+void destroyImageList() {
+	// Bug in windows: Cannot set the imageList handle to 0 directly, it doesn't
+	// seems to cache the previous imageList and show flaky behavior. Instead set
+	// the size of the image to the minimum (1,1) and then set the imageList to 0.
+	long hImageList = OS.ImageList_Create (1, 1, 0, 0, 0);
+	long oldStateImageList = OS.SendMessage (handle, OS.LVM_SETIMAGELIST, OS.LVSIL_STATE, hImageList);
+	long oldhImageList = OS.SendMessage (handle, OS.LVM_SETIMAGELIST, OS.LVSIL_SMALL, hImageList);
+	OS.ImageList_Destroy(oldStateImageList);
+	OS.ImageList_Destroy(oldhImageList);
+	OS.SendMessage (handle, OS.LVM_SETIMAGELIST, OS.LVSIL_STATE, 0);
+	OS.SendMessage (handle, OS.LVM_SETIMAGELIST, OS.LVSIL_SMALL, 0);
+	OS.ImageList_Destroy(hImageList);
+}
+
 @Override
 void handleDPIChange(Event event, float scalingFactor) {
 	super.handleDPIChange(event, scalingFactor);
@@ -7358,6 +7372,12 @@ void handleDPIChange(Event event, float scalingFactor) {
 		imageList = null;
 	}
 
+	if ((style & SWT.CHECK) != 0 ) {
+		destroyImageList();
+		int size = getItemHeightInPixels();
+		setCheckboxImageList(size, size, true);
+	}
+
 	// if the item height was set at least once programmatically with CDDS_SUBITEMPREPAINT,
 	// the item height of the table is not managed by the OS anymore e.g. when the zoom
 	// on the monitor is changed, the height of the item will stay at the fixed size.
@@ -7375,7 +7395,6 @@ void handleDPIChange(Event event, float scalingFactor) {
 		// Update scrollbar width if no columns are available
 		 setScrollWidth(scrollWidth);
 	}
-	 fixCheckboxImageListColor (true);
-	 settingItemHeight = false;
+	settingItemHeight = false;
 }
 }


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Table.java
```diff
@@ -7336,6 +7336,19 @@ LRESULT wmNotifyToolTip (NMTTCUSTOMDRAW nmcd, long lParam) {
 	return null;
 }
 
+void destroyImageList() {
+	long oldStateImageList = OS.SendMessage(handle, OS.LVM_GETIMAGELIST, OS.LVSIL_STATE, 0);
+	long oldhImageList = OS.SendMessage(handle, OS.LVM_GETIMAGELIST, OS.LVSIL_SMALL, 0);
+	long hImageList = OS.ImageList_Create (1, 1, 0, 0, 0);
+	OS.SendMessage (handle, OS.LVM_SETIMAGELIST, OS.LVSIL_STATE, hImageList);
+	OS.SendMessage (handle, OS.LVM_SETIMAGELIST, OS.LVSIL_SMALL, hImageList);
```
arunjose696: cant we get the old image lists from the LVM_SETIMAGELIST instead of LVM_GETIMAGELIST?

```suggestion

	long hImageList = OS.ImageList_Create (1, 1, 0, 0, 0);
	long oldStateImageList = OS.SendMessage (handle, OS.LVM_SETIMAGELIST, OS.LVSIL_STATE, hImageList);
	long oldhImageList =OS.SendMessage (handle, OS.LVM_SETIMAGELIST, OS.LVSIL_SMALL, hImageList);
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Table.java
```diff
@@ -7358,6 +7371,12 @@ void handleDPIChange(Event event, float scalingFactor) {
 		imageList = null;
 	}
 
+	if((style & SWT.CHECK) != 0 ) {
+		destroyImageList();
+		int size = getItemHeightInPixels();
+		setCheckboxImageList(size, size, false);
```
arunjose696: @amartya4256 previously `setCheckboxImageList` was called with `height `and `width `of the icon and `fixScroll` true, why do we consider only the height now, is it because the checkbox is always square I see no visual regressions but just slightly confused why the height and width can be passed when setting checklbox `ImageList`, also why is `fixscroll `now false?

amartya4256:
```
int width = OS.HIWORD(oneItem) - OS.HIWORD(empty), height = width;
			setCheckboxImageList(width, height, false);
```
Yes the height and the width are the same. Look at the block above from `createHandle` method.

amartya4256: `fixscroll` was a mistake. Fixed it.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Table.java
amartya4256: Adapted

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Table.java
```diff
@@ -7358,6 +7372,12 @@ void handleDPIChange(Event event, float scalingFactor) {
 		imageList = null;
 	}
 
+	if((style & SWT.CHECK) != 0 ) {
```
akoch-yatta: spacing between if and bracket 😄

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Table.java
```diff
@@ -7375,7 +7395,6 @@ void handleDPIChange(Event event, float scalingFactor) {
 		// Update scrollbar width if no columns are available
 		 setScrollWidth(scrollWidth);
 	}
-	 fixCheckboxImageListColor (true);
```
akoch-yatta, line 7378: Why is that call removed? Not necessary because of the other changes?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Table.java
```diff
@@ -7336,6 +7336,20 @@ LRESULT wmNotifyToolTip (NMTTCUSTOMDRAW nmcd, long lParam) {
...
+	// Bug in windows: Cannot set the imageList handle to 0 directly, it doesn't
+	// seems to cache the previous imageList and show flaky behavior. Instead set
+	// the size of the image to the minimum (1,1) and then set the imageList to 0.
+	long hImageList = OS.ImageList_Create (1, 1, 0, 0, 0);
+	long oldStateImageList = OS.SendMessage (handle, OS.LVM_SETIMAGELIST, OS.LVSIL_STATE, hImageList);
+	long oldhImageList = OS.SendMessage (handle, OS.LVM_SETIMAGELIST, OS.LVSIL_SMALL, hImageList);
+	OS.ImageList_Destroy(oldStateImageList);
+	OS.ImageList_Destroy(oldhImageList);
+	OS.SendMessage (handle, OS.LVM_SETIMAGELIST, OS.LVSIL_STATE, 0);
+	OS.SendMessage (handle, OS.LVM_SETIMAGELIST, OS.LVSIL_SMALL, 0);
```
akoch-yatta, line 7349: You should add a comment that this double call is necessary. Otherwise it looks like a mistake

amartya4256: There are 2 different calls. One for LVSIL_STATE and one for LVSIL_SMALL. Do you mean these 2 calls at 7348 and 7349?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Table.java
amartya4256: SetCheckboxImageList is it's replacement in the if block. They both do the similar things. With the new call we can pass the right size calculated using getHeight call.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Table.java
akoch-yatta: Oh, your're right. Oversaw the difference

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Table.java
amartya4256: Adapted

arunjose696 (approved):
Tested the latest changes. It fixes the issue  and code changes look alright hence approving

PR #2802, opened 2025-11-20
Adapt usages in SWT to new GC#drawImage API

arunjose696:
The previous drawImage method required specifying the full source and destination rectangles, which meant the caller needed to know the image size up front. This change replaces the old API calls with the newer, simplified drawImage overload that takes fewer parameters and no longer requires the caller to provide the image dimensions.


akoch-yatta:
@arunjose696 Could you split this PR up a bit. I would be in favor of one PR for each of:
- CLabel
- ImageData
- Table&Tree
- Snippets
- Tests
- Other examples (You could split up more here, e.g. browser, imageanalyzer, accessibility, but I am fine with one for all)

This makes it way easier to test and review each adaption separatly

arunjose696:
> @arunjose696 Could you split this PR up a bit. I would be in favor of one PR for each of:

Have split this to multiple PRs so closing this one

PR #2803, opened 2025-11-20
Simplify newImageData  pass integer zoom instead of zoomContext

arunjose696:
All callers currently use a zoom context with targetZoom == nativeZoom. This change passes a single integer zoom to AbstractImageProviderWrapper::newImageData, removing the unnecessary zoomContext parameter.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
@@ -1328,7 +1328,7 @@ public ImageData getImageData (int zoom) {
 	if (zoomLevelToImageHandle.containsKey(zoom)) {
 		return zoomLevelToImageHandle.get(zoom).getImageData();
 	}
-	return this.imageProvider.newImageData(new ZoomContext(zoom));
+	return this.imageProvider.newImageData(zoom);
 }
 
 
@@ -2007,7 +2007,7 @@ private abstract class AbstractImageProviderWrapper {
 
 	protected abstract ElementAtZoom<ImageData> loadImageData(int zoom);
 
-	abstract ImageData newImageData(ZoomContext zoomContext);
+	abstract ImageData newImageData(int zoom);
 
 	abstract AbstractImageProviderWrapper createCopy(Image image);
 
@@ -2068,8 +2068,8 @@ private class ExistingImageHandleProviderWrapper extends AbstractImageProviderWr
 	}
 
 	@Override
-	ImageData newImageData(ZoomContext zoomContext) {
-		return getScaledImageData(zoomContext.targetZoom());
+	ImageData newImageData(int zoom) {
+		return getScaledImageData(zoom);
 	}
 
 	@Override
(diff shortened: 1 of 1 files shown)


laeubi:
@arunjose696 @akoch-yatta I'd like to make you aware of

- https://github.com/eclipse-platform/eclipse.platform.ui/pull/3431

where I found the whole concept of passing zoom values (only) both hard to use reliable (e.g. zoom is only integer value while one requires sometimes fractional) and unflexible (one has to compute a suitable zoom level in the "upper layer" to get a matching image while one usually knows the target size quite well.

I therefore like to suggest to rethink the concept of zoom levels really match well for image providers (I think it does not). and not unify this to instead always use a target size what can trivially be derived from a given initial size and zoom level.

Such a `TargetSize` could then be a record of `(double zoom, int width, int height)` with the contract that if width + height have values > 0 this is assumed to be the "native size at zoom = 1 -> 100%". That way one can use that to override the "native size" and gets a scaled variant with the given zoom level (e.g. it might later be zoomed to 150% = zoom 1.5), runding would of course still happen but it will be as much exact as possible.

FYI @HeikoKlare

arunjose696:
@laeubi I’m not sure if [PR 3431](https://github.com/eclipse-platform/eclipse.platform.ui/pull/3431)
 has any effect on this change this one is just a minor refactoring.

I’m also slightly confused by the approach in your proposal: having default values for height and width in the TargetSize record feels a bit counterintuitive. I would expect this to be seperated,  or wouldnt the callers requesting image at a target size need to know about the default size of the image?

That said, I think these two PRs are unrelated and can be treated separately and if we need to have a target size along with zoom that would be a different proposal. I have also breifly reviewed the open PR

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2270,15 +2270,15 @@ protected Rectangle getBounds(int zoom) {
 	}
 
 	@Override
-	ImageData newImageData(ZoomContext zoomContext) {
-		int targetZoom = zoomContext.targetZoom();
+	ImageData newImageData(int zoom) {
+		int targetZoom = zoom;
```
HeikoKlare: `targetZoom` could be completely removed here

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2095,18 +2095,18 @@ private abstract class ImageFromImageDataProviderWrapper extends AbstractImagePr
...
-		newImageData(new ZoomContext(100));
+		newImageData(100);
 	}
 
 	@Override
-	ImageData newImageData(ZoomContext zoomContext) {
+	ImageData newImageData(int zoom) {
 		Function<Integer, ImageData> imageDataRetrieval = zoomToRetrieve -> {
-			ImageHandle handle = initializeHandleFromSource(zoomContext);
+			ImageHandle handle = initializeHandleFromSource(new ZoomContext(zoom));
```
HeikoKlare: For the sake of consistency, shouldn't this anyway be ... ?
```suggestion
			ImageHandle handle = initializeHandleFromSource(new ZoomContext(zoomToRetrieve));
```

HeikoKlare (approved):
Thank you for the latest update. Looks good to me now.

Issue #2806, opened 2025-11-22
Edge integration leaves unused Timers around


jonahgraham:
Consider this sample:

```java
import org.eclipse.swt.*;
import org.eclipse.swt.browser.*;
import org.eclipse.swt.layout.*;
import org.eclipse.swt.widgets.*;

public class LotsOfUSERObjects {

	public static void main(String[] args) {
		Display display = new Display();
		Shell shell = new Shell(display);
		shell.setLayout(new FillLayout());
		Browser browser = new Browser(shell, SWT.NONE);
		shell.open();
		for (int i = 0; i < 5000; i++) {
			browser.setText("Iteration " + i);
			new BrowserFunction(browser, "name");
		}
		while (!shell.isDisposed()) {
			if (!display.readAndDispatch())
				display.sleep();
		}
		display.dispose();
	}
}
```

If you observe this with the Windows task manager (or Process Explorer) you will see there are 10,000 User Objects created, but they don't do anything as they clean up eventually.

The above is a minimal reproducer from a huge Eclipse application. Obviously the back to back `setText` calls are pointless, but across the huge application it was less obvious that they were redundant, and you can conceive of a design that isn't redundant too. As this sample goes to exactly (or would go past the Windows limit) if any other controls or windows are created at that point a `SWTError: No more handles` error is produced at that place.

PRs that fix this on the way.

PR #2807, opened 2025-11-22
[Edge] Cancel the timer from running when we are done with it

jonahgraham:
Each use of a timer creates a User Object in Windows, and once we have stopped waiting for our desired event we don't need the cancellation tracking anymore.

If there are lots of calls to processOSMessagesUntil, such as repeated browser.setText within the MAXIMUM_OPERATION_TIME (5 second) window, we can end up with a huge growth in the number of User Objects, leading to potential `SWTError: No more handles` errors elsewhere in the client code.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2806


jonahgraham:
@HeikoKlare I don't know how to automate a regression test for this, I assume the existing test cases will cover it sufficiently.

jonahgraham:
> I.e., something like:

Done.

HeikoKlare (approved):
Great, thank you!

PR #2808, opened 2025-11-22
throw no more handles if a timer could not be created

jonahgraham:
At the moment the code silently fails on calling timerExec if there are no handles left or it fails for other reasons.

Relates to https://github.com/eclipse-platform/eclipse.platform.swt/issues/2806 where we can see that we are running out of handles, but the timerExec does not report a failure.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
@@ -5343,6 +5343,7 @@ public <T, E extends Exception> T syncCall(SwtCallable<T, E> callable) throws E 
  * @exception SWTException <ul>
  *    <li>ERROR_THREAD_INVALID_ACCESS - if not called from the thread that created the receiver</li>
  *    <li>ERROR_DEVICE_DISPOSED - if the receiver has been disposed</li>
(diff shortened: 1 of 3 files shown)


jonahgraham:
This is a draft for now because it needs consideration for other platforms and whether after 25 years there is a reason that this case was allowed to fail silently on Windows.

On macOS there is a pointless null check in the same place because an NPE would have already been raised. On GTK I need to investigate more because the functions are not documented as being able to fail.

macOS:

https://github.com/eclipse-platform/eclipse.platform.swt/blob/bb954ac3236fcdfa89e6ed6dd1dfcce5cf3bb317/bundles/org.eclipse.swt/Eclipse%20SWT/cocoa/org/eclipse/swt/widgets/Display.java#L5381-L5390

GTK:

https://github.com/eclipse-platform/eclipse.platform.swt/blob/286e614a1d9a3ebc85a4f5c4c194be5b7fd92288/bundles/org.eclipse.swt/Eclipse%20SWT/gtk/org/eclipse/swt/widgets/Display.java#L5750-L5759

jonahgraham:
> This is a draft for now because it needs consideration for other platforms and whether after 25 years there is a reason that this case was allowed to fail silently on Windows.
> 
> On macOS there is a pointless null check in the same place because an NPE would have already been raised. On GTK I need to investigate more because the functions are not documented as being able to fail.

My latest commit adds throwing no more handles on macos and gtk if the timer creation routines return 0/null so they also don't silently fail anymore.

I plan to submit this soon if there are no objections.

Issue #2804, opened 2025-11-21
Test_org_eclipse_swt_browser_Browser.test_evalute_Cookies() failing on GTK x86_64


elsazac:
**Describe the bug**

The `test_evaluate_Cookies()` in `org.eclipse.swt.tests.junit.Test_org_eclipse_swt_browser_Browser` test is failing on GTK x86_64 for both Java 21 and Java 25 in the test results for I-build [I20251120-1800](https://download.eclipse.org/eclipse/downloads/drops4/I20251120-1800).

```
expected: <false> but was: <true>

org.opentest4j.AssertionFailedError: expected: <false> but was: <true>
at org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:151)
at org.junit.jupiter.api.AssertionFailureBuilder.buildAndThrow(AssertionFailureBuilder.java:132)
at org.junit.jupiter.api.AssertFalse.failNotFalse(AssertFalse.java:63)
at org.junit.jupiter.api.AssertFalse.assertFalse(AssertFalse.java:36)
at org.junit.jupiter.api.AssertFalse.assertFalse(AssertFalse.java:31)
at org.junit.jupiter.api.Assertions.assertFalse(Assertions.java:231)
at org.eclipse.swt.tests.junit.Test_org_eclipse_swt_browser_Browser.test_evalute_Cookies(Test_org_eclipse_swt_browser_Browser.java:419)
at java.base/java.lang.reflect.Method.invoke(Method.java:580)
at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
Suppressed: java.lang.Throwable: Screenshot written to /home/jenkins/agent/workspace/AutomatedTests/ep438I-unit-linux-x86_64-java21/workarea/I20251120-1800/eclipse-testing/results/ep438I-unit-linux-x86_64-java21_linux.gtk.x86_64_21/org.eclipse.swt.tests.junit.Test_org_eclipse_swt_browser_Browser.test_evalute_Cookies().png
at org.eclipse.swt.tests.junit.Test_org_eclipse_swt_widgets_Widget.lambda$1(Test_org_eclipse_swt_widgets_Widget.java:79)
at java.base/java.util.Optional.ifPresent(Optional.java:178)
[...]

PR #2810, opened 2025-11-24
An attempt to fix test_evalute_Cookies

iloveeclipse:
Just a guess the failure could be somehow related to redirect from https://www.eclipse.org/swt to https://eclipse.dev/eclipse/swt/.

See https://github.com/eclipse-platform/eclipse.platform.swt/issues/2804

### DIFF
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
@@ -404,19 +404,26 @@ public void test_evalute_Cookies () {
 	browser.addProgressListener(ProgressListener.completedAdapter(event -> loaded.set(true)));
 
 	// Using JavaScript Cookie API on local (file) URL gives DOM Exception 18
-	browser.setUrl("https://www.eclipse.org/swt");
+	String url = "https://eclipse.dev/eclipse/swt/";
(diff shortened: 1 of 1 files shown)


akurtakov:
As a start you might want to print the content or change the assertFalse(res.isEmpty()) to assertEquals("", res) so the actual value is seen. 
Changing the URL doesn't seem to make a deifferent so my feeling is that this is related to some internal webkit change.

iloveeclipse:
> so my feeling is that this is related to some internal webkit change.

Are you able to run this test on latest webkit? Our RHEL 9.6 version is surely very old `webkit2gtk3-2.48.1-1.el9_5.x86_64`.

akurtakov:
I can't reproduce the problem locally on webkitgtk-2.50.2-1.fc43

akurtakov:
One thing worth considering here is that version 2.50 no longer links against libsoup but rather libsoup3 and it might actually be a change in libsoup . https://webkitgtk.org/2025/10/07/webkitgtk-soup2-deprecation.html indicates that for now only some distros dropped libsoup (2.x) support but starting with current indevelopment version of webkit (2.52) only libsoup3 is supported. This library is what handles the cookie stuff IIRC.

akurtakov:
Putting this upstream change here for later analysis (if needed) https://github.com/WebKit/WebKit/pull/51819/files#diff-38776530271615386c97930733aae45d563980de6a3ae553cfeb49db1458e63f

akurtakov:
Closing in favor of https://github.com/eclipse-platform/eclipse.platform.swt/pull/2825

PR #2493, opened 2025-09-09
SWT: Respect Windows accessibility cursor size setting

ShahzaibIbrahim:
On Windows, the mouse pointer size can be increased in Accessibility settings. Previously SWT always created cursors at their logical bitmap size (e.g. 16x16, 32x32), ignoring the accessibility scale.

This change reads `CursorBaseSize` from
`HKCU\Control Panel\Cursors` and uses it as a scale factor when creating SWT cursors. For example, with scale = 5, a 16px cursor bitmap is scaled to 80px before being displayed, matching the user’s configured pointer size.

This aligns SWT custom cursors with the system accessibility setting and improves usability for users with enlarged cursors.

### How to Test

- Run following snippet.

<details><summary>Snippet386</summary>
<p>

```java
public class Snippet386 {

	private static final int IMAGE_SIZE_IN_POINTS = 16;

	public static void main(String[] args) {
		System.setProperty("swt.autoScale.updateOnRuntime", "true");
		Display display = new Display();
		Shell shell = createShell(display);

		Label zoomLabel = createZoomLabel(shell);
		Combo combo = createConstructorCombo(shell);

		addZoomChangedListener(shell, zoomLabel);
		addPaintTicks(shell);

		CursorManager cursorManager = new CursorManager(display, shell, combo);
		combo.addListener(SWT.Selection, e -> cursorManager.updateCursor());
		cursorManager.updateCursor();

		shell.setSize(300, 600);
		shell.open();

		eventLoop(display, shell);
		display.dispose();
	}

	private static Shell createShell(Display display) {
		Shell shell = new Shell(display);
		shell.setText("Cursor Constructors Example");
		shell.setLayout(new GridLayout(1, false));

		Label label = new Label(shell, SWT.NONE);
		label.setText("Choose Cursor Constructor:");
		return shell;
	}

	private static Combo createConstructorCombo(Composite parent) {
		Combo combo = new Combo(parent, SWT.READ_ONLY);
		combo.setItems("Cursor(Device, int)", "Cursor(Device, ImageData, ImageData, int, int)",
				"Cursor(Device, ImageData, int, int)", "Cursor(Device, ImageDataProvider, int, int)");
		combo.select(0);
		return combo;
	}

	private static Label createZoomLabel(Composite parent) {
		Label zoomLabel = new Label(parent, SWT.NONE);
		zoomLabel.setLayoutData(new GridData(SWT.FILL, SWT.CENTER, true, false));
		setZoomLabelText(zoomLabel);
[...]

PR #2811, opened 2025-11-25
[Win32] Apply accessibility setting to ImageDataProvider-based cursor

HeikoKlare:
Cursors are scaled according to the accessibility settings of Windows. However, this setting is only applied to Cursor instances based on image data but not on those that are instantiated via an ImageDataProvider. This change also applies the scale factor in the latter case.

See:
- https://github.com/eclipse-gef/gef-classic/issues/872#issuecomment-3575191088

I consider this a safe change as it is simply an oversight in the existing implementation. The change has no effect in case the accessibility setting of Windows is not used. In case it is used, the quality of custom cursors based on an ImageDataProvider (as an example and in particuar for GEF) will have a better user experience.
This is why I would be in favor of still bringing this into the upcoming releae / RC2, even though we are already very late.

### How to test
Change the accessibility settings for the mouse cursor size in the Windows settings and check with any custom cursor that uses an `ImageDataProvider`, such as the shared cursors of GEF.
image

### Before
https://github.com/user-attachments/assets/a62f6d5b-be74-4022-801c-5e6932727ee7

### After
https://github.com/user-attachments/assets/ccd88a96-9a7f-450f-9abd-b96d467b3eda

This is basically a follow up to the following PR which applied the accessibility setting to the image-data-based construction of cursors:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2493

This depends on and thus has to be merged after:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2830

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
@@ -628,7 +628,8 @@ private static class ImageDataProviderCursorHandleProvider extends HotspotAwareC
 	@Override
 	public CursorHandle createHandle(Device device, int zoom) {
 		Image tempImage = new Image(device, this.provider);
-		ImageData source = tempImage.getImageData(zoom);
+		int scaledZoom = (int) (zoom * getPointerSizeScaleFactor());
+		ImageData source = tempImage.getImageData(scaledZoom);
 		tempImage.dispose();
 		return setupCursorFromImageData(device, source, null, getHotpotXInPixels(zoom), getHotpotYInPixels(zoom));
 	}


ptziegler (approved):
I think this is fine for the sake of consistency. There just seems to be one minor issue (which also happens with the other providers) that the image data is not updated when the scale factor is changed. Though I think that's best handled separately.

If I understand everything correctly, the cache stores the handle by monitor zoom level, even though the monitor zoom * scale factor would probably make more sense.

HeikoKlare:
Failures are due to `test_evalute_Cookies` test in both the GHA and the Jenkins run. See:
- https://github.com/eclipse-platform/eclipse.platform.swt/issues/2804

> There just seems to be one minor issue (which also happens with the other providers) that the image data is not updated when the scale factor is changed. Though I think that's best handled separately.
> 
> If I understand everything correctly, the cache stores the handle by monitor zoom level, even though the monitor zoom * scale factor would probably make more sense.

That's indeed a separate issue. As you say, it would probably make sense to include the accessibility scale factor into the handle map key and it should work then.

HeikoKlare:
We propose to merge this for RC2. According to the comments and arguments, we consider the change of low risk while it will improve the user experience for use cases of custom cursors. As mentioned in https://github.com/eclipse-platform/eclipse.platform.swt/pull/2811#issuecomment-3575841543, the failing checks are due to the preexisting `test_evalute_Cookies` test fails.

@iloveeclipse @akurtakov @merks what do think from PL/PMC perspective? If you agree to bring this into the release, it would be great to have your PL/PMC+1.

merks:
PMC+1

I trust that you folks have all the necessary expertise to make the best technical decisions for the good of the project.

iloveeclipse:
I must confess the change looks non trivial for me, but I have no idea about Windows / SWT specific zoom handling. 
So I abstain from voting.

HeikoKlare:
> I must confess the change looks non trivial for me, but I have no idea about Windows / SWT specific zoom handling.
> So I abstain from voting.

Thank you for the feedback!
I understand that it's difficult to assess without knowing the code and the specifics around it (Windows, zoom handling etc.).
Even though we cannot complete resolve this, I would like to share some help for understanding what happens and why we consider it safe (though maybe not completely trivial):
- Cursors (like any resources) are created according to the zoom for which they are requested
- On Windows, there is an OS setting for scaling cursors (accessibility) in addition to the monitor zoom
- This accessibility factor is already applied for one way of creating a Cursor (via an `ImageData` object) while it was missing for the creation way that we adapt here (via an `ImageDataProvider`), which is why this is no completely new behavior but just extends it to all cases where it should have already been applied before. This is the existing case: https://github.com/eclipse-platform/eclipse.platform.swt/blob/a0d6ffcfa2fe09e1962677bc78703adbeea6dff6/bundles/org.eclipse.swt/Eclipse%20SWT/win32/org/eclipse/swt/graphics/Cursor.java#L654-L658
- The effect of this change is as follows:
  - The cursor size will be multiplied by the accessibility scaling factor
  - That scaling factor is 1 by default and will then have no effect at all
  - In case the value cannot be retrieved for whatever reason (it is queried from the Windows registry) it also defaults to 1
  - Only if the user has specified a non-default accessibility setting and if that value can be retrived, the behavior can change at all
- Also note that this way of creating a cursor (via `ImageDataProvider`) is not a long-standing behavior that could break long-standing consumers but was just introduced with the previous Eclipse/SWT release

HeikoKlare:
I investigated the change again and found that the check against `DEFAULT_ZOOM` is a bit unclear to me. I adapted it with the latest commit to what I think it should be, but without properly knowing why it is necessary/useful, I am not in favor of proceeding with the change now anymore but would like to better understand that condition first.

HeikoKlare:
> I investigated the change again and found that the check against `DEFAULT_ZOOM` is a bit unclear to me. I adapted it with the latest commit to what I think it should be, but without properly knowing why it is necessary/useful, I am not in favor of proceeding with the change now anymore but would like to better understand that condition first.

I found that we just added/kept that code because we when doing a larger adaptation of the Cursor class we were already late in the development cycle and did not want to risk issue by changing this behavior as well: https://github.com/eclipse-platform/eclipse.platform.swt/pull/2390#discussion_r2260255140
Still we wanted to follow up in the next release cycle on that but obviously missed to do that. So we should proceed with this for at the beginning of the next release cycle first and then make this change on top of it.

I have created https://github.com/eclipse-platform/eclipse.platform.swt/pull/2830 and rebased this on top of it. So it should be merged after that other PR.

akoch-yatta:
Had a look into the latest change, and the approval from my side is still valid.

Issue #2812, opened 2025-11-25
[GTK4] Caret not being drawn in StyledText


jonahgraham:
The Caret is not being drawn on GTK4 StyledText (or it is being drawn, but it is not visible to the user).

This can be observed with Snippet163, which is one of the simplest snippets with StyledText. Snippet74 is *just* the caret, and the caret works there.

jonahgraham:
In my first pass of evaluations this is a difference I see between GTK3 and GTK4 caret drawing in StyledText.

The caret needs to be drawn (on or off) every `Caret.blinkRate`, which causes a bunch of drawing to happen every blinkRate (I'm calling it an iteration below)

In GTK3, each iteration calls `Display.windowProc` 4 times:

1. Shell {Snippet 163}
2. StyledText {}
3. StyledText {} (a second time)
4. Shell {Snippet 163} (a second time)

In (2) above the windowProc calls gtk_draw, which calls StyledText.handlePaint. Which means in each iteration the StyledText is being painted once.

In GTK4, each iteration calls `Display.snapshotDrawProc` 3 times:

1. Shell {Snippet 163}
2. StyledText {}
3. StyledText {} (a second time)

In 2 and 3 for GTK4 the snapshotDrawProc eventually calls gtk4_draw then StyledText.handlePaint. Which means in each iteration the StyledText is being painted twice.

The first time that snapshotDrawProc calls gtk4_draw, the Canvas.blink == true, so drawCaret is also called. But the second time gtk4_draw is called, since Canvas.blink == false, the caret doesn't get drawn, leaving the caret invisible as StyledText.handlePaint has redrawn over where the caret was.

If you "skip" the second paint of styledtext the caret is visible. For example, apply this hack (that may only work with Snippet163):

```diff
diff --git a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
index 132f293b132..3cc5a29e2ac 100644
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java      
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java      
@@ -27,6 +27,7 @@ import java.util.regex.*;
 import java.util.regex.Pattern;
 
[...]

PR #2813, opened 2025-11-25
[GTK4] Set source color when doing difference on pixels in Caret

jonahgraham:
With no source color explicitly set here, the difference is applied using whatever happens to have been the last set source color, or 0,0,0 if none have been set.

See https://github.com/eclipse-platform/eclipse.platform.swt/issues/2812#issuecomment-3577428337 for a longer explanation.


akurtakov:
Is this one still needed?

akurtakov:
I believe https://github.com/eclipse-platform/eclipse.platform.swt/pull/3356 is a more complete fix than this PR.

jonahgraham:
Using CAIRO_OPERATOR_OVER as done in #3356 makes sense, so closing this in favour of #3356.

PR #2814, opened 2025-11-25
[GTK4] Skip drawing of fixedHandle

jonahgraham:
The fixedHandle does not have any drawable components, and as written today, what gtk4_draw does on fixedHandle is draw the whole Widget, then Display.snapshotDrawProc iterates through all the children, of which the normal handle will be one (at least in StyledText case) and gtk4_draw will be called again on the same widget.

This fixes caret drawing, because on the first time through gtk4_draw the caret is drawn, but then on the second time the gtk4_draw draws over the freshly drawn caret.

This is option 1 in https://github.com/eclipse-platform/eclipse.platform.swt/issues/2812#issuecomment-3577503275

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2812

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Control.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Control.java
@@ -3834,7 +3834,17 @@ void cairoClipRegion (long cairo) {
 	eventRegion = actualRegion;
 }
 
+@Override
+void snapshotToDraw(long handle, long snapshot) {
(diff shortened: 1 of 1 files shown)


jonahgraham:
> This is option 1 in https://github.com/eclipse-platform/eclipse.platform.swt/issues/2812#issuecomment-3577503275

Option 2 doesn't work, so this is the only option I provide.

jonahgraham:
This fix by itself breaks other things. I am writing a test to make sure that Paint event is sent exactly once on a redraw, but that shows some other Widgets are not getting their paint event now with this change.

jonahgraham:
Closing - the underlying issue is resolved differently now.

PR #2818, opened 2025-11-26
[GTK4] Clamp height of combo drop-down arrow to client area height

### LINKED ISSUE #2817: [GTK4] CCombo draws arrow button too high
Consider this simplified version of Snippet39:

```java
public class Snippet39 {
		shell.setText("Snippet 39");

		CCombo combo = new CCombo(shell, SWT.NONE);
		combo.setText("text");

		shell.pack();
		shell.open();
		while (!shell.isDisposed()) {
		}
		display.dispose();
	}
}
```

The arrow button on the right is two pixels too high and loses the bottom of the button:

Image

In the unmodified snippet, the button is too high too, but it may be harder to see because the grid layout makes the shell bigger to make it visible. But you can see the button goes two pixels lower than the text:

Image

This error happens because `Button.computeSize` has a small mistake around adding border when the button is default:

https://github.com/eclipse-platform/eclipse.platform.swt/blob/d884e6b44dcaf11aec87088d1a5467f238b4c6ff/bundles/org.eclipse.swt/Eclipse%20SWT/gtk/org/eclipse/swt/widgets/Button.java#L284-L291

You can see it clearly by applying this diff and seeing that the arrow computeSize returns a height higher than the client area:

```diff
$ git diff 'bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CCombo.java'
diff --git a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CCombo.java b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CCombo.java
index c700fd1632..af1efe06ab 100644
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CCombo.java      
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CCombo.java      
@@ -1187,6 +1187,7 @@ void internalLayout (boolean changed) {
        int width = rect.width;
        int height = rect.height;
        Point arrowSize = arrow.computeSize (SWT.DEFAULT, height, changed);
+       System.out.println("ClientArea Height: " + rect.height + " Arrow Computed Height: " + arrowSize.y);
        text.setBounds (0, 0, width - arrowSize.x, height);
        arrow.setBounds (width - arrowSize.x, 0, arrowSize.x, arrowSize.y);
 }
```

I will provide a PR that clamps the height of the arrow to the client area, but it doesn't feel like quite the right fix.

jonahgraham:
Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2817

Note that this fix is needed for GTK4, but affects all platforms as it is in swt custom's CCombo. I am leaving this as draft for the moment because the preferred fix is probably fixing computeSizeInPixels of Button.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CCombo.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CCombo.java
@@ -1187,6 +1187,9 @@ void internalLayout (boolean changed) {
 	int width = rect.width;
 	int height = rect.height;
 	Point arrowSize = arrow.computeSize (SWT.DEFAULT, height, changed);
+	// clamp the height of the arrow to the maximum of the client area
+	// https://github.com/eclipse-platform/eclipse.platform.swt/issues/2817
+	arrowSize.y = Math.min(height, arrowSize.y);
 	text.setBounds (0, 0, width - arrowSize.x, height);
 	arrow.setBounds (width - arrowSize.x, 0, arrowSize.x, arrowSize.y);
 }


jonahgraham:
I am going to close this because it is almost for sure this isn't the correct solution.

Issue #2819, opened 2025-11-26
[GTK4] custom Paint is under instead of over control


jonahgraham:
Consider the following snippet:

```java
public class SnippetDrawOverButton {

public static void main(String [] args) {
	var display = new Display();
	var shell = new Shell(display);
	shell.setLayout(new FillLayout());
	Button button = new Button(shell, SWT.CHECK);
	button.setText("SnippetDrawOverButton");
	Font initialFont = button.getFont();
	FontData[] fontData = initialFont.getFontData();
	for (FontData element : fontData) {
		element.setHeight(24);
	}
	Font newFont = new Font(display, fontData);
	button.setFont(newFont);
	button.addPaintListener(event -> {
		Point rect = button.getSize();
		event.gc.setForeground(display.getSystemColor(SWT.COLOR_WHITE));
		event.gc.setLineWidth(5);
		event.gc.drawOval(1, 1, rect.x - 2, rect.y - 2);
	});
	shell.pack();
	shell.open ();
	while (!shell.isDisposed()) {
		if (!display.readAndDispatch()) display.sleep();
	}
	display.dispose();
}
}
```

If you run it on GTK3 you get the oval drawn over the control: (win32 looks similar to GTK3)

<img width="404" height="79" alt="Image" src="https://github.com/user-attachments/assets/2114ff7e-2835-47bc-b48c-f320d6d74dbc" />

But on GTK4 it draws under the control:

<img width="397" height="80" alt="Image" src="https://github.com/user-attachments/assets/b5d905aa-3a85-47cf-829d-9e63a06795d8" />

jonahgraham:
This is tightly related to https://github.com/eclipse-platform/eclipse.platform.swt/issues/2812 - both of them are about the order and how many gtk4_draws happen when GTK4 calls Display. But also how SWT and GTK drawing interact.

The Control.paintHandle implementation has changed many times since GTK4 support was first written. For example, I can hook paint signals (`Control.hookPaintSignals`) into either `fixedHandle` or `handle` for a `SWT.PUSH` style `Button`.

With the hook into `fixedHandle` (the current implementation I see this:

<img width="387" height="91" alt="Image" src="https://github.com/user-attachments/assets/997863d7-6252-4920-9b71-d50ad8d0649e" />

With the hook into `handle` I see:

[...]

PR #2822, opened 2025-11-27
[GTK4] Fix custom drawing in GTK4

jonahgraham:
This involves changing the mechanism with which we do custom drawing. Previously we tried to hook into GtkWidget's snapshot, but this required messing around with the vtable of the widget to call our handler instead.

This commit changes to using GtkOverlays + GtkDrawingAreas to allow custom drawing on top of the GtkWidgets.


jonahgraham:
This is draft because it is only the first step of https://github.com/eclipse-platform/eclipse.platform.swt/issues/2819#issuecomment-3583634720 - new commits/pushes as I get the other steps done.

akurtakov:
https://github.com/eclipse-platform/eclipse.platform.swt/pull/3069 fixes the issue so I propose we close this one for now. Using GtkOverlay will require far bigger changes than we can afford doing now.

jonahgraham:
I am closing this for now - the caret drawing has been fixed a different way. I still have concerns about rewriting snapshot by editing vtable, but to reduce noise I close this for now.

PR #2825, opened 2025-11-28
Run browser tests against a local HttpServer

jonahgraham:
For a while we have suffered with patchy tests due to limited connectivity to external websites when running Browser tests.

This commit brings in a new Http Server based on the now available Java 21 simple web server available in the JDK. This is sufficient to test the various paths currently used in the browser tests, and make them reliable (or at least more resilient against external sites going dark)

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2804

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2561

This has as a parent commit PR #2824

### DIFF
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/EchoHttpServer.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/EchoHttpServer.java
@@ -1,0 +1,136 @@
+/*******************************************************************************
+ * Copyright (c) 2025 Kichwa Coders Canada, Inc.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.swt.tests.junit;
+
+import java.io.BufferedReader;
+import java.io.Closeable;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.io.OutputStream;
+import java.net.InetAddress;
+import java.net.InetSocketAddress;
+import java.net.URLDecoder;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
+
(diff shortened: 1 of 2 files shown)


jonahgraham:
The Windows tests I modified are currently failing - I tried to make these new test more robust. For example in the POST test we were just checking that the returned data was "big enough" without checking content. I have changed that to make sure the results are exactly as expected, which works on Linux. But Edge + IE both re-interpret/re-format the HTML leading to comparison errors.

For example: expected: `<!DOCTYPE html><html><head>` but was `<HTML><HEAD><TITLE>`. 

I also think newlines change too. I'll provide an update that loosens the new requirements I just added.

---

This new server uses http://localhost:PORT where PORT is assigned per test (standard way using port 0 to get free port).

---

A while ago all the tests were changed from http: -> https: and this goes back to http:. There is a comment in the code that said:

> "please use https instead, http do not work on mac out of the box and your test will hang there!"

But the tests all seem to pass on macOS. I will spin this up on my mac machine to make sure they work locally too.

I can add https to the webserver, but there is a few hoops to jump through to make keys. However I don't think the functionality being tested needs https. In reality we should have some https tests - for example to make sure that failed certs are somewhat displayed to users, but AFAICT there are no such tests today.

jonahgraham:
> This pull request changes some projects for the first time in this development cycle.

This is really a parent commit - PR #2824 problem.

jonahgraham:
The last iteration of Jenkins all browswer tests passed: https://ci.eclipse.org/releng/job/eclipse.platform.swt/job/PR-2825/2/testReport/org.eclipse.swt.tests.junit/Test_org_eclipse_swt_browser_Browser/

I would like to merge this now as I have a number of other commits related to #2824 that sit on top of this.

jonahgraham:
The Jenkins build ran against https://github.com/eclipse-platform/eclipse.platform.swt/pull/2838 and shows my whole collection of commits to remove warnings has been successful. As I don't think there is anything controversial or risky about this change I am merging now.

jonahgraham:
Whoops - too many open PRs. https://github.com/eclipse-platform/eclipse.platform.swt/pull/2825#issuecomment-3603201033 was a comment meant for #2837 

Anyways, this is now merged. If there is any concern I will address it asap.

Issue #554, opened 2023-02-17
[Windows] Scroll bars are wider at 200% scaling


Phillipus:
**Describe the bug**
Using Eclipse on Windows 11 at 100% scaling, scroll bars are quite thin (expected behaviour). Setting the native display to 200% (for example, with 4k display) and the scroll bars are very wide.

**To Reproduce**
Start Eclipse with native display set to 200%.

**Expected behavior**
Scrollbars should be the same width at 100% and at 200% scaling.

**Screenshots**

At 100%:

![win100](https://user-images.githubusercontent.com/600504/219634989-9deccdee-da46-4d15-93ae-8b836e5e3429.png)

At 200%:

![win200](https://user-images.githubusercontent.com/600504/219634996-9ef72f54-15d9-4f99-93c2-e8343b8d3ef4.png)

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 

2. Additional OS info (e.g. OS version, Linux Desktop, etc)
Windows 11 Pro.
4k monitor at 200% scaling.

4. JRE/JDK version
JDK 17

**Workaround (or) Additional context**
The scrollbar width seems to be set at Eclipse startup.
I can set the native display to 100%, start Eclipse, then set the native display to 200% and the scrollbars remain thin.

deepika-u:
@Phillipus 
I have checked on the scaling of 100% and 200% on windows 11 and please find my observations below
1. I have checked the scroll bars on word. Seeing as below screen shot(top circled part corresponds to 100% and bottom circled part corresponds to 200% scaling). -> 
![Scaling_word_100Left_200Right](https://user-images.githubusercontent.com/46807064/220836721-7a0d771b-8861-42f9-87f4-38bb228c3791.jpg)

2. I have checked the scroll bars on notepad. Seeing as below screen shot. To be specific when it is hovered the horizontal scroll bar widens up(as shown in the image) where as vertical scroll bar is as-is at different scaling levels. From screen shot top 2 circles corresponds to 100% and bottom 2 circles corresponds to 200% scaling.
![Scaling_notepad_100top_200bottom](https://user-images.githubusercontent.com/46807064/220836838-3fefe17f-c3e8-4966-8423-82f502e25984.jpg)

[...]

PR #2827, opened 2025-11-28
Replacement of Image(device, int, int) constructor in examples/snippets

al-noori:
This PR takes over a part to change all occurences of the image(device, int, int) constructor with the image(device, gc, int, int) constructor.

Within examples/snippets, almost all usages of the stated constructor are now replaced by an ImageGcDrawer and the Image(device, gc int, int) constructor afterwards. This replacement has/could not be made for the snippets {387, 215, 292, 95, 139} and partially not for 141. The details for them are commented here: https://github.com/vi-eclipse/Eclipse-Platform/issues/310#issuecomment-3472301957.

### DIFF
--- a/examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet10.java
+++ b/examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet10.java
@@ -32,12 +32,12 @@ public class Snippet10 {
 		shell.setText("Advanced Graphics");
 		FontData fd = shell.getFont().getFontData()[0];
 		final Font font = new Font(display, fd.getName(), 60, SWT.BOLD | SWT.ITALIC);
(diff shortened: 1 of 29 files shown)


akoch-yatta:
I got the following difference when opening Snippet10 on a 175% monitor with master (left) and this PR (right):
image

Didn't check the other Snippets yet.

HeikoKlare:
> I got the following difference when opening Snippet10 on a 175% monitor with master (left) and this PR (right):

That looks like an issue with `GC#drawImage` and `ImageGcDrawer`. Should we create a separate ticket for that, @akoch-yatta?

akoch-yatta:
Yes, we had a look into it yesterday, but I didn't create a ticket yet. Just created [#554](https://github.com/vi-eclipse/Eclipse-Platform/issues/554)

PR #2828, opened 2025-11-28
Replacement of Image(device, int, int) constructor in examples/graphics

al-noori:
All usages of the stated constructor with an additional GC initialization are now replaced by an ImageGcDrawer and the Image(device, gc int, int) constructor afterwards in examples/graphics.

### DIFF
--- a/examples/org.eclipse.swt.examples/src/org/eclipse/swt/examples/graphics/GraphicsExample.java
+++ b/examples/org.eclipse.swt.examples/src/org/eclipse/swt/examples/graphics/GraphicsExample.java
@@ -28,10 +28,12 @@ import org.eclipse.swt.graphics.Color;
 import org.eclipse.swt.graphics.Device;
 import org.eclipse.swt.graphics.GC;
 import org.eclipse.swt.graphics.Image;
+import org.eclipse.swt.graphics.ImageGcDrawer;
 import org.eclipse.swt.graphics.Path;
 import org.eclipse.swt.graphics.Pattern;
 import org.eclipse.swt.graphics.Point;
 import org.eclipse.swt.graphics.Rectangle;
+import org.eclipse.swt.internal.TransparencyColorImageGcDrawer;
 import org.eclipse.swt.layout.FormAttachment;
 import org.eclipse.swt.layout.FormData;
 import org.eclipse.swt.layout.FormLayout;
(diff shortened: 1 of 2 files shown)


akoch-yatta:
If I try this PR on a setup 100% primary, 175% secondary I get the following issue:
image

To reproduce:
1. Open the ControlExample (with vm args -Dswt.autoScale=quarter -Dswt.autoScale.updateOnRuntime=true) (on primary)
2. Open Tree Tab
3. Change background color in "Colors and Fonts"
4. Move to secondary monitor

I get the same effect on master, when I change the color, while being on the secondary monitor, but it is happening now on DPI change as well (because the handle is now recreated in that scenario as well)

HeikoKlare:
> I get the same effect on master, when I change the color, while being on the secondary monitor, but it is happening now on DPI change as well (because the handle is now recreated in that scenario as well)

@akoch-yatta If I understand correctly, you refer to the mismatch between box and it filling, don't you? Should we create a separate ticket for it?

akoch-yatta:
@HeikoKlare I removed those and rebased the branch on master

examples/org.eclipse.swt.examples/src/org/eclipse/swt/examples/graphics/GraphicsExample.java
```diff
@@ -368,16 +377,26 @@ static Image createImage(Device device, Color color1, Color color2, int width, i
...
+			if (color.equals(device.getSystemColor(SWT.COLOR_BLACK))) {
+				gc.setForeground(device.getSystemColor(SWT.COLOR_WHITE));
+			}
+			gc.drawRectangle(0, 0, imageWidth - 1, imageHeight - 1);
+		}
+
+		@Override
+		public void postProcess(ImageData imageData) {
+			imageData.transparentPixel = imageData.palette.getPixel(getTransparencyColor(device, color).getRGB());
+		}
```
HeikoKlare: This should be obsolete

examples/org.eclipse.swt.examples/src/org/eclipse/swt/examples/graphics/GraphicsExample.java
```diff
@@ -347,19 +349,26 @@ static Image createThumbnail(Device device, String name) {
...
+		gc.drawRectangle(0, 0, iwidth - 1, iheight - 1);
+		pattern.dispose();
+	};
+	Image image = new Image(device, igc, width, height);
 	return image;
 }
 
+private static Color getTransparencyColor(Device device, Color color) {
+	Color transparencyColor = device.getSystemColor(SWT.COLOR_CYAN);
+	if (color == transparencyColor) {
```
HeikoKlare:
```suggestion
	if (transparencyColor.equals(color)) {
```

examples/org.eclipse.swt.examples/src/org/eclipse/swt/examples/graphics/GraphicsExample.java
al-noori: Removed it

examples/org.eclipse.swt.examples/src/org/eclipse/swt/examples/graphics/GraphicsExample.java
al-noori: Should be changed by now

PR #2835, opened 2025-12-01
Replace fail() with assertThrows() in TreeItem test

ptziegler:
This replaces the following pattern:

```
try {
	...
	fail();
} catch (IllegalArgumentException e) {
}
```

with

```
assertThrows(IllegalArgumentException.class, () -> ...);

### DIFF
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_widgets_TreeItem.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_widgets_TreeItem.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2018 IBM Corporation and others.
+ * Copyright (c) 2000, 2025 IBM Corporation and others.
  *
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_widgets_TreeItem.java [hunk 1]
```diff
@@ -812,20 +774,15 @@ public void test_setFontLorg_eclipse_swt_graphics_Font() {
...
+	treeItem.setFont(font2);
+	assertEquals(treeItem.getFont(), font2);
 
 	treeItem.setFont(null);
 	assertEquals(treeItem.getFont(), tree.getFont());
 
-	font.dispose();
-	try {
-		treeItem.setFont(font);
-		treeItem.setFont(null);
```
ptziegler, line 825: This call is effectively dead code. If `treeItem.setFont(font)` throws the expected `IllegalArgumentException`, this line is never called. And if it doesn't, the test fails and a new `treeItem` instance is created at the start of the next test.

ptziegler:
Merging, as this is only a cleanup task and doesn't change the scope covered by the tests.

Issue #2834, opened 2025-12-01
Inconsistency in TreeItem.getExpanded() between Linux and Windows


ptziegler:
**Describe the bug**
In certain configurations, a call to `getExpanded()` returns `false` on Linux, but `true` on Windows. The result was seen in https://github.com/eclipse-platform/eclipse.platform.ui/pull/3527, where "dummy" tree items are created on Linux but not on Windows, because JFace believes a tree item to be collapsed when it is not.

**To Reproduce**
Consider the following snippet:

```java
public static void main(String[] args) {
	Shell shell = new Shell();
	Tree tree = new Tree(shell, SWT.FULL_SELECTION);
	TreeItem item1 = new TreeItem(tree, SWT.NONE);
	TreeItem item2 = new TreeItem(item1, SWT.NONE);
	item1.setExpanded(true);
	System.out.println(item1.getExpanded());
	item2.dispose();
	System.out.println(item1.getExpanded());
	item2 = new TreeItem(item1, SWT.NONE);
	System.out.println(item1.getExpanded());
}
```

On Linux, the second call to `getExpanded()` returns `false`, because the tree doesn't contain any non-disposed children. However, the item is still expanded, as otherwise the third call would also return `false`.
On Windows, all calls return `true`.

In addition, the Linux-implementation of this widget has an `isExpanded` field that keeps track of the expanded state. This field is `true`, even though `getExpanded()` returns `false`.

**Expected behavior**
The `TreeItem` widget should behave consistently across multiple platforms.

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [x] Windows
- - [x] Linux
- - [ ] macOS

ptziegler:
This seems to be very similar to [Bug 455167](https://bugs.eclipse.org/bugs/show_bug.cgi?id=455167) - "Collapse All" creates invisible empty element at the collapsed subtree.

iloveeclipse:
For the history: the `isExpanded` field on Linux was added via 75d5803b14cf89e469ebb012f27eb8919a8482dc

I don't know why it was not used in `getExpanded()`, but this patch seem to fix the problem from https://github.com/eclipse-platform/eclipse.platform.ui/issues/3525 (together with https://github.com/eclipse-platform/eclipse.platform.ui/pull/3527)
```
[...]

PR #2836, opened 2025-12-01
Make TreeItem.getExpanded consistent across platforms #2834

ptziegler:
When the children of a previously expanded TreeItem are removed, the call to getExpanded() should continue to return "true". On both Linux and MacOS, this property is not persisted and therefore stored in a local variable. But on Linux, a call to getExpanded() still returns the result from a call to the GTK API.

To harmonize the behavior between the different operating systems, following changes are done:

1) The call to getExpanded() now always returns the local variable, similar to how it's done for MacOS.

2) The call to setExpanded() doesn't modify the tree item if it is already expanded or a leaf node.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/TreeItem.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/TreeItem.java
@@ -535,10 +535,7 @@ public boolean getChecked () {
  */
 public boolean getExpanded () {
 	checkWidget();
-	long path = GTK.gtk_tree_model_get_path (parent.modelHandle, handle);
-	boolean answer = GTK.gtk_tree_view_row_expanded (parent.handle, path);
-	GTK.gtk_tree_path_free (path);
-	return answer;
+	return isExpanded;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


ptziegler:
I've tested this with https://github.com/eclipse-platform/eclipse.platform.ui/pull/3527 and now the issue also disappears on Linux. From my side this is done and any feedback is greatly appreciated.

vogella:
Here is the feedback from Gemini CLI:

The Pull Request #2836: Make TreeItem.getExpanded consistent across platforms aims to harmonize the behavior of TreeItem
  between Linux (GTK) and macOS.

  Here is an analysis of the changes:

  Summary
  The core issue addressed is that on Linux, TreeItem.getExpanded() was directly querying the underlying GTK widget. If a
  TreeItem's children were removed, GTK would treat it as a leaf (non-expandable), causing getExpanded() to return false even
  if it was previously expanded. On macOS, this state is persisted in a local variable, returning true even after children
  are removed. This PR adopts the macOS approach for Linux.

  Key Changes
   1. `TreeItem.java` (GTK implementation):
       * `getExpanded()`: Now returns a local isExpanded field instead of calling GTK.gtk_tree_view_row_expanded. This
         ensures the state is remembered purely within SWT.
       * `setExpanded(boolean)`:
           * Added a check GTK.gtk_tree_model_iter_n_children(...) != 0.
           * The method now does nothing if the item is a leaf (has 0 children). It neither calls GTK functions nor updates
             the isExpanded field.
           * This effectively "locks" the expansion state when an item becomes a leaf.

   2. Tests (`Test_org_eclipse_swt_widgets_TreeItem.java`):
       * Updated test_getExpanded to verify:
           * setExpanded(true) on a fresh leaf does not change getExpanded() (stays false).
           * Removing children from an expanded item preserves getExpanded() == true.
           * Calling setExpanded(false) on that now-leaf item does not change getExpanded() (stays true).

  Assessment
  The changes accurately implement the stated goal.
   * Consistency: It ensures Linux behavior matches macOS regarding state persistence.
   * Logic: The updated setExpanded logic ensures that the "virtual" expansion state isn't lost or inadvertently flipped when
     the item temporarily lacks children.
   * Verification: The added tests cover the edge cases (leaf interaction and state persistence after child removal).

  Overall, the change looks correct and safe, improving cross-platform consistency for TreeItem state management.

iloveeclipse (approved):

iloveeclipse:
@ptziegler : many thanks for fix & testing. Let merge this ASAP, to get as much coverage as possible.

iloveeclipse:
Verified on my RHEL 9.6 & platform UI patch, everything works as expected.

iloveeclipse:
> Verified on my RHEL 9.6 & platform UI patch, everything works as expected.

We've found one regression, see https://github.com/eclipse-platform/eclipse.platform.swt/issues/2887. Fix is https://github.com/eclipse-platform/eclipse.platform.swt/pull/2888.

PR #2838, opened 2025-12-02
Warnings and webserver for test and other cleanups

jonahgraham:
This is the follow-up for https://github.com/eclipse-platform/eclipse.platform.swt/pull/2824 as one big PR with bug fixes, warning removals, and other cleanups all mixed together.

I will split this PR into bite-size pieces later, but I want to make sure that this PR does indeed remove the extra warnings and issues that #2824 has introduced.

This isn't marked as a draft so it runs on Jenkins too - its not really ready for review.

PR #2842, opened 2025-12-02
Fixup warnings due to deprecated URL constructor (in Edge/IE/WebKit)

jonahgraham:
The windows tests hit all the cases in Edge/IE. A bunch of code was unreachable for WebKit because WebKit has a bunch of custom code that is not needed on Edge/IE so this PR includes new tests to cover these cases, but they are only run on gtk at the moment.

Needed to resolve new warnings due to update to Java21 https://github.com/eclipse-platform/eclipse.platform.swt/pull/2824


jonahgraham:
The Jenkins build ran against https://github.com/eclipse-platform/eclipse.platform.swt/pull/2838 which was my whole collection of commits.

As we have clean GitHub actions I am merging this now.

Issue #2833, opened 2025-12-01
[Win32] Cursor size not updated when accessibility setting changes


HeikoKlare:
**Describe the bug**
When changing the cursor size in the Windows accessibility settings, the sizes of already created cursors (i.e., those that have already been instantiated for a specific monitor zoom) will not be updated for that zoom anymore and stick to the existing accessibility setting instead. This is caused by the cursor handle cache only taking the monitor zoom but not the accessibility zoom into account. See https://github.com/eclipse-platform/eclipse.platform.swt/pull/2811#pullrequestreview-3505205704 from @ptziegler.

**To Reproduce**
- Open Snippet386 and choose one of the constructors using `ImageData` (or with https://github.com/eclipse-platform/eclipse.platform.swt/pull/2811 you can also use the one using `ImageDataProvider`)
- Hover over the "Scale" area once such that the cursor appears
- Change the accessibility settings of your system to use a different cursor size
- Hover over the "Scale" area again

You will see that the cursor still has the "old" size (i.e., without the accessibility setting change) when hovering over that area.

**Expected behavior**
Cursor sizes should dynamically adapt to changed accessibility settings.

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 

**Workaround (or) Additional context**
Restarting the application or resetting the context in which the cursor was created helps. But since, e.g., in GEF Classic such cusors are static/shared, you have to restart the application.

What would probably solve the issue is to simply use `monitorZoom * accessibilityZoom` as key for the Cursor handle cache.

PR #2850, opened 2025-12-03
[Win32] Update Cursor size when accessibility setting changes

ShahzaibIbrahim:
currently cursor handle is created per zoom. So when cursor is created at 100% zoom it will be reused even if you change the cursor size from accessibility setting. With this change we create a new cursor handle with the key combination of (zoom x accessibilityFactor).

Fixes: https://github.com/eclipse-platform/eclipse.platform.swt/issues/2833

### To Reproduce

- Open Snippet386 and choose one of the constructors using ImageData
- Hover over the "Scale" area once such that the cursor appears
- Change the accessibility settings of your system to use a different cursor size
- Hover over the "Scale" area again

**Previous Behavior:** 

You will see that the cursor still has the "old" size (i.e., without the accessibility setting change) when hovering over that area.

**Expected behavior:**
Cursor sizes should dynamically adapt to changed accessibility settings.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
@@ -320,14 +320,15 @@ public static Long win32_getHandle (Cursor cursor, int zoom) {
 	if (cursor.isDisposed()) {
 		return 0L;
 	}
-	if (cursor.zoomLevelToHandle.get(zoom) != null) {
-		return cursor.zoomLevelToHandle.get(zoom).getHandle();
+	int zoomWithPointerSizeScaleFactor = (int) (zoom * getPointerSizeScaleFactor());
+	if (cursor.zoomLevelToHandle.get(zoomWithPointerSizeScaleFactor) != null) {
+		return cursor.zoomLevelToHandle.get(zoomWithPointerSizeScaleFactor).getHandle();
 	}
 
-	CursorHandle handle = cursor.cursorHandleProvider.createHandle(cursor.device, zoom);
-	cursor.setHandleForZoomLevel(handle, zoom);
+	CursorHandle handle = cursor.cursorHandleProvider.createHandle(cursor.device, zoomWithPointerSizeScaleFactor);
+	cursor.setHandleForZoomLevel(handle, zoomWithPointerSizeScaleFactor);
 
-	return cursor.zoomLevelToHandle.get(zoom).getHandle();
+	return cursor.zoomLevelToHandle.get(zoomWithPointerSizeScaleFactor).getHandle();
 }
 
 private void setHandleForZoomLevel(CursorHandle handle, Integer zoom) {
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java
```diff
@@ -320,14 +320,15 @@ public static Long win32_getHandle (Cursor cursor, int zoom) {
...
 		return 0L;
 	}
-	if (cursor.zoomLevelToHandle.get(zoom) != null) {
-		return cursor.zoomLevelToHandle.get(zoom).getHandle();
+	int zoomWithPointerSizeScaleFactor = (int) (zoom * getPointerSizeScaleFactor());
+	if (cursor.zoomLevelToHandle.get(zoomWithPointerSizeScaleFactor) != null) {
+		return cursor.zoomLevelToHandle.get(zoomWithPointerSizeScaleFactor).getHandle();
 	}
 
 	CursorHandle handle = cursor.cursorHandleProvider.createHandle(cursor.device, zoom);
```
HeikoKlare: Couldn't we pass the adapted zoom to the `createHandle()` method as well and remove all the `getPointerSizeScaleFactor()` from the different `createHandle()` implementations?

ShahzaibIbrahim: I thought about that but I then saw that in some of the method the normal zoom is also used. Like here in the getHotpotXInPixels and getHotpotYInPixels methods. Or should I pass both the zooms?

See,

https://github.com/vi-eclipse/eclipse.platform.swt/blob/2e63ce8fbd89f5986855512c5f22feb67edddee7/bundles/org.eclipse.swt/Eclipse%20SWT/win32/org/eclipse/swt/graphics/Cursor.java#L650-L655

HeikoKlare: That sound plain wrong to me. Why should the hotspot calculation use a different zoom than the cursor image itself? I guess in the snippets we should see broken hotspot behavior already now.

ShahzaibIbrahim: I will update the code to pass the adapted zoom to createHandle methods anyway (the behavior is also fine in this case).

HeikoKlare (approved):
The change works perfectly fine for me. The considered zoom now consists of monitor zoom and accessibility scale factor and adapts to changes to either of them accordingly.

I have extracted the fix for the hotspot calculation issue I reported in https://github.com/eclipse-platform/eclipse.platform.swt/pull/2850#discussion_r2585350097 to another PR so that we have separate commits/PRs tracking those independent changes: https://github.com/eclipse-platform/eclipse.platform.swt/pull/2851

I have successfully tested both those changes with this PR already and will rebase and merge this one right after #2851

PR #2854, opened 2025-12-03
[Build] Remove workarounds to fetch JDK for win32.aarch64/linux.riscv64

HannesWell:
Because SWT is now built against Java-21 and JustJ provides Java-21 JDKs for all supported platforms, the special handling to get JDKs and the headers for win32.aarch64 and linux.riscv64 are not necessary anymore.

Follow-up on
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2824


HannesWell:
> this LGTM and we don't need to wait on the aarch64 windows machine to come to life before submitting this.

Thanks and agree. The pipeline got past the point where the changed code is used. So yes, everything should be good.

Still I reported the offline state in
- https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/4726#note_5419510

PR #2856, opened 2025-12-06
[GTK] Remove always false local variable

jonahgraham:
This is a follow-up to 7460be01cd1d48f9eb991a9b41d5cdea86e3edaa which removed the setting of xFocus = to true but missed deleting variable itself.


jonahgraham:
> Test Results (linux)

Woohoo! faster runtime with just one OS needing to run.

PR #2860, opened 2025-12-08
Edge Browser Key Event Async Handling

amartya4256:
This PR does the following:

1. Makes Widget:filters public: This method is used by many clients which are primarily a widget - child of Widget class. Arguably, there can be Child classes of Widget outside the package, such as, Browser which also heavily use Event based norms. Hence, this method should be made public. Additionally, it is a getter and doesn't do any harm in any way upon exposing.
2. Edge Browser KeyEvent fix: This is a direct reference to https://github.com/eclipse-platform/eclipse.platform.ui/issues/3104. The freezing of edge browser happens because the KeyEvent Listener (which is a display filter) calls for another Edge Browser instance which won't complete as it is called in a callback, causing deadlocks. Hence, the KeyEvent listeners should be executed asynchronously in the methodEdge:handleAcceleratorKeyPressed. However, `args.set_handled` in the method must set the value to true only if the shortcut is handled by SWT, i.e., if there's a listener and handles the key combo. Since we cannot check that and set the value while executing asyncly, we needed to find another way to handle it.

Solution: set the Handled to true when there's a listener for key event.

Drawbacks: Of course that means that it will set Handled to true for every Key Event regardless of whatever combination of keys pressed.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
@@ -1573,9 +1573,12 @@ int handleAcceleratorKeyPressed(long pView, long pArgs) {
 			return COM.S_OK;
 		}
 
-		if (!sendKeyEvent(keyEvent)) {
-			args.put_Handled(true);
-		}
+		asyncExec(() -> {
+			sendKeyEvent(keyEvent);
+		});
+
+		boolean hasListener = browser.getListeners(keyEvent.type).length != 0 || browser.getDisplay().filters(keyEvent.type);
+		args.put_Handled(hasListener);
 	} else {
 		keyEvent.type = SWT.KeyUp;
 		browser.notifyListeners (keyEvent.type, keyEvent);
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Widget.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Widget.java
@@ -870,7 +870,7 @@ NSRect expansionFrameWithFrame_inView(long id, long sel, NSRect cellRect, long v
 	return result;
 }
 
-public boolean filters (int eventType) {
+boolean filters (int eventType) {
 	return display.filters (eventType);
 }
 
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
@@ -1224,7 +1224,33 @@ boolean filterEvent (Event event) {
 	return false;
 }
 
-boolean filters (int eventType) {
+/**
+ * Checks for if there exists a listener in the collection of listeners who will
+ * be notified when an event of the given type occurs anywhere in a widget. The event type is one of the event constants
+ * defined in class <code>SWT</code>. When the event does occur,
+ * the listener is notified by sending it the <code>handleEvent()</code>
+ * message.
+ *
+ * <p>
+ * Filters are listeners which are registered at display level by calling {@link Display#addFilter(int, Listener)}. These listeners are triggered when there happens an event of the type they are registered with, anywhere in an SWT application.
+ * </p>
+ *
+ * <p>
+ * Needs to be public as the consumers of display can register a filter from any package using public method <code>addFilter</code> and might need to check if there exists a filter for an event.
+ * </p>
+ *
+ * @param eventType the type of an event to check if there exists a listener registered using <code>addFilter</code>
+ * @return true if there exists a listener which reacts to the eventType otherwise false
+ *
+ * @see Listener
+ * @see SWT
+ * @see #addFilter
+ * @see #removeFilter
+ * @see #removeListener
+ *
+ * @since 3.133
+ */
+public boolean filters (int eventType) {
 	if (filterTable == null) return false;
 	return filterTable.hooks (eventType);
 }
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
@@ -1987,7 +1987,33 @@ boolean filterEvent (Event event) {
 	return false;
 }
 
-boolean filters (int eventType) {
+/**
+ * Checks for if there exists a listener in the collection of listeners who will
+ * be notified when an event of the given type occurs anywhere in a widget. The event type is one of the event constants
+ * defined in class <code>SWT</code>. When the event does occur,
+ * the listener is notified by sending it the <code>handleEvent()</code>
+ * message.
+ *
+ * <p>
+ * Filters are listeners which are registered at display level by calling {@link Display#addFilter(int, Listener)}. These listeners are triggered when there happens an event of the type they are registered with, anywhere in an SWT application.
+ * </p>
+ *
+ * <p>
+ * Needs to be public as the consumers of display can register a filter from any package using public method <code>addFilter</code> and might need to check if there exists a filter for an event.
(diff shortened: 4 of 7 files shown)


bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Widget.java
```diff
@@ -870,7 +870,7 @@ NSRect expansionFrameWithFrame_inView(long id, long sel, NSRect cellRect, long v
 	return result;
 }
 
-boolean filters (int eventType) {
+public boolean filters (int eventType) {
```
akurtakov: I am not convinced that this should become public at all. It would help if you add javadoc (mandatory if its to become public!)  that explains why/what/how/when this should be used.

HeikoKlare: Is it even necessary to do this? 
1. Couldn't you also just create a package-protected method in `Browser` that is accessed by `Edge` to get the information, or ...
2. couldn't you even just use `browser.getDisplay().filters...` here?

Note that I have not tried to understand the overall intent of the PR yet, so I cannot judge whether the proposal as a whole makes sense. The comment is just about making this method public.

amartya4256: Display:filters is also package protected. At first I thought we could make this public but then I saw all the widget children classes which use filters they use `Widget#filters` so for a better convenience, I made that public instead.

iloveeclipse: > I made that public instead.

Note, **if** this is supposed to be public API, it should provide also a proper javadoc clearly explaining the API contract (beside the usual rules that since tag is needed, bundle version bump etc).

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -1573,9 +1573,13 @@ int handleAcceleratorKeyPressed(long pView, long pArgs) {
...
 
-		if (!sendKeyEvent(keyEvent)) {
-			args.put_Handled(true);
-		}
+		asyncExec(() -> {
+			sendKeyEvent(keyEvent);
+		});
+
+		boolean hasListener = browser.getListeners(keyEvent.type).length != 0 || browser.filters(keyEvent.type);
+		System.out.println(hasListener);
```
iloveeclipse: this is not a draft PR, so please remove all debug code.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Widget.java
akurtakov: Just so it's 100% clear - no version bump should be needed as minor version is already bumped for this release via https://github.com/eclipse-platform/eclipse.platform.swt/pull/2824

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
amartya4256: Ahh thanks, missed that one out.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Widget.java
amartya4256: I added the javadoc. Does that look alright or should it be different?

iloveeclipse: > Does that look alright or should it be different?

Just read the method name and try to imagine you would as a user need to use it.
Here expected reaction looking at the API proposed:

> "filters"... What is that?
> Filter **what**?
> **What** kind of "eventType" should be given as argument?
> Why **filters** says anything about listeners: "if there exists a **listener**"?
> Where is **there**? On the current widget? All widgets in the hierarchy? Display?

If you would try to answer to the questions above, you would realize that this method helps to check whether there are **filters** registered **on the display** for given event type. 

So I believe **if** it should be API, it should be named "hasFilters(eventType)" and located at Display, next to addFilter() and "removeFilter()". And of course javadoc should match the javadoc provided by two other methods.

amartya4256: Naming-wise, I think its fine. `filters` here is used as a verb, as in, if it filters the event. I think, that was the intention of naming it like that here. So convenience wise the methods mean, add a filter to display, remove a filter from display and if display filters an event. hasFilter is also correct but so is filter. But if we name it `hasFilter`, either we need to provide a new method which is a copy of `filters` or we need to rename `filters` to `hasFilter` for which we need to adapt all its consumers. I think we can go with the name `filter` and provide more in-depth explanation in the javadoc just the way it is done is `addFilter`.
I'll re-adapt the javadoc and push.

arunjose696:
I have tested the changes, it fixes the problem mentioned in the issue. 

@amartya4256  However as this makes some key combinations like ctrl+shift+r or ctr+r not work on the browser I would prefer to have this documented in the contract of` handleAcceleratorKeyPressed() `  adding to the other limitations mentioned in the javadoc

amartya4256:
> I have tested the changes, it fixes the problem mentioned in the issue.
> 
> @amartya4256 However as this makes some key combinations like ctrl+shift+r or ctr+r not work on the browser I would prefer to have this documented in the contract of`handleAcceleratorKeyPressed()` adding to the other limitations mentioned in the javadoc

You can use `CTRL + SHIFT + R` to open the Open resource dialog. If you mean you cannot use other key combinations which are native to browser, yes you are right since with this implementation, if there's a key listener registered either by the display or the browser, when the browser is in focus, the native key combinations of browser won't work. Did you mean this?

arunjose696:
> You can use `CTRL + SHIFT + R` to open the Open resource dialog. If you mean you cannot use other key combinations which are native to browser, yes you are right since with this implementation, if there's a key listener registered either by the display or the browser, when the browser is in focus, the native key combinations of browser won't work. Did you mean this?

Yes I meant the shortcuts of the browser not working, I think this would be good to be documented as a limitation

HeikoKlare:
> If you mean you cannot use other key combinations which are native to browser, yes you are right since with this implementation, if there's a key listener registered either by the display or the browser, when the browser is in focus, the native key combinations of browser won't work.

That sounds like a severe limitation and contradicts what I would expect from the integrated browser. That even means you cannot easily use search functionality (CTRL + F) within the browser anymore.

amartya4256:
> That sounds like a severe limitation and contradicts what I would expect from the integrated browser. That even means you cannot easily use search functionality (CTRL + F) within the browser anymore.

Yes that is right. to find out which key combinations are registered with eclipse, we need to go inside the execution. We can somehow still change the behaviour but for that we need to modify the implementation in `KeyBindingDispatcher` where `KeyBindingDispatcher#generatePossibleKeyStrokes` is used to obtain the actual key combination and is compared somewhere with the Key Combination of the listener. Maybe we can add a method that checks for if this key combination is registered with us only then execute this asyncly otherwise tell the browser to execute its native action?

HeikoKlare:
> Maybe we can add a method that checks for if this key combination is registered with us only then execute this asyncly otherwise tell the browser to execute its native action?

How would you do that? Would you retrieve the listeners registered for the event from the browser and then try to extract from them if the key combination is processed by them? This information is currently not exposed by the listeners, so you would need to extend the interface of such listeners, wouldn't you? I am not sure if that is feasible, but maybe I do not fully understand the idea yet. Also note that `KeyBindingDispatcher` is Platform functionality, so nothing you can directly access from SWT.

Maybe we should simply go with the workaround proposed in https://github.com/eclipse-platform/eclipse.platform.ui/issues/3104#issuecomment-3061931734?

amartya4256:
Closing as taken over by https://github.com/eclipse-platform/eclipse.platform.ui/pull/3614

PR #2861, opened 2025-12-08
Use shorter strings in the macOS matrix configuration

jonahgraham:
By having the full names of runner and native it meant in many places GitHub displayed ... instead of the full matrix entry.

Therefore, use a simpler arch matrix and derive the runner/native values from it.


jonahgraham:
This PR changes the UI on GHA from:

image

to the not-cut-off this:

image

PR #2863, opened 2025-12-09
Add setAutoscaleDisabled API to provide a better means to disabling autoscaling at widget level

amartya4256:
This PR provides a new API `setAutoscaleDisabled` to provide a better means of disabling autoscaling at widget level across all the platforms. The current implementation is intended to return a false when this method is called in cocoa and GTK and only perform an action win32 and return since we do not support widget level autoscaling control on GTK and cocoa yet.

The convention of the method is as follows:
1. Takes a boolean parameter for disabling or enabling autoscale
2. returns true if successful (in case of win32) otherwise false (in case of GTK and cocoa)

Things to be done:
- Usage of different autoscaling modes as enabled, disabled and disabled_cascaded, etc
- Testing with GEF to see how it impacts.

depends on https://github.com/eclipse-platform/eclipse.platform.swt/pull/2930

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Control.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Control.java
@@ -3527,6 +3527,26 @@ NSTouch findTouchWithId(NSArray touches, NSObject identity) {
 	return null;
 }
 
+/**
+ * Sets the autoscaling mode for this widget. The capability is not supported on
+ * every platform, such that calling this method may not have an effect on
+ * unsupported platforms. The return value indicates if the autoscale mode was
+ * set properly. With {@link #isAutoScalable()}, the autoscale enablement can
+ * also be evaluated at any later point in time.
+ * <p>
+ * Currently, this is only supported on Windows.
+ * </p>
+ *
+ * @param autoscalingMode the autoscaling mode to set
+ *
+ * @return {@code false} if the operation was called on an unsupported platform
+ *
+ * @since 3.133
+ */
+public boolean setAutoscalingMode(AutoscalingMode autoscalingMode) {
+	return false;
+}
+
 void setBackground () {
 	if (!drawsBackground()) return;
 	Control control = findBackgroundControl ();
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/AutoscalingMode.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/AutoscalingMode.java
@@ -1,0 +1,28 @@
+package org.eclipse.swt.graphics;
+
+/**
+ * Defines the autoscaling behavior used when rendering or computing layout for
+ * controls that support DPI‑aware scaling.
+ * <p>
+ * This mode determines whether SWT takes care of the scaling of widgets with
+ * respect to the zoom level of the monitor.
+ * </p>
+ *
+ * <ul>
+ * <li><b>{@link #ENABLED}</b> – Autoscaling is explicitly enabled. Values (such
+ * as coordinates, dimensions, or layout metrics) will be adjusted according to
+ * the effective scaling factor.</li>
+ *
+ * <li><b>{@link #DISABLED}</b> – Autoscaling is explicitly disabled. Values
+ * will be used as‑is without applying any scaling transformations.</li>
+ *
+ * <li><b>{@link #DISABLED_INHERITED}</b> – Autoscaling is disabled for this
+ * control and this is inherited by its children. Child controls will also be
+ * initialized with DISABLED_INHERITED AutoscalingMode.</li>
+ * </ul>
+ *
+ * @since 3.133
+ */
+public enum AutoscalingMode {
+	ENABLED, DISABLED, DISABLED_INHERITED
+}
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
@@ -78,12 +78,24 @@ public abstract class Control extends Widget implements Drawable {
 	Region region;
 	Font font;
 	int drawCount, foreground, background, backgroundAlpha = 255;
-	boolean autoScaleDisabled = false;
+	AutoscalingMode autoscalingMode = AutoscalingMode.ENABLED;
 
+	/**
+	 * @deprecated use {@link Shell#getZoom(AutoscalingMode)} instead to obtain the zoom of the shell.
+	 */
+	@Deprecated(since = "2026-03", forRemoval = true)
 	private static final String DATA_SHELL_ZOOM = "SHELL_ZOOM";
 
+	/**
+	 * @deprecated use {@link Control#setAutoscalingMode(AutoscalingMode)} instead to set the autoscaling mode directly.
+	 */
+	@Deprecated(since = "2026-03", forRemoval = true)
 	private static final String DATA_AUTOSCALE_DISABLED = "AUTOSCALE_DISABLED";
 
+	/**
+	 * @deprecated use {@link Control#setAutoscalingMode(AutoscalingMode)} instead to set the autoscaling mode directly.
+	 */
+	@Deprecated(since = "2026-03", forRemoval = true)
 	private static final String PROPOGATE_AUTOSCALE_DISABLED = "PROPOGATE_AUTOSCALE_DISABLED";
 /**
  * Prevents uninitialized instances from being created outside the package.
@@ -124,11 +136,11 @@ Control () {
 public Control (Composite parent, int style) {
 	super (parent, style);
 	this.parent = parent;
-	Boolean parentPropagateAutoscaleDisabled = (Boolean) parent.getData(PROPOGATE_AUTOSCALE_DISABLED);
-	if (parentPropagateAutoscaleDisabled == null || parentPropagateAutoscaleDisabled) {
-		this.autoScaleDisabled = parent.autoScaleDisabled;
+	AutoscalingMode parentAutoscaleMode = parent.autoscalingMode;
+	if (parentAutoscaleMode == AutoscalingMode.DISABLED_INHERITED) {
+		this.autoscalingMode = parent.autoscalingMode;
 	}
-	if (!autoScaleDisabled) {
+	if (!isAutoscalingDisabled()) {
 		this.nativeZoom = getShellZoom();
 	}
 	createWidget ();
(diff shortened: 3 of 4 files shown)


HeikoKlare:
@amartya4256 can you please add a meaningful title and description to the PR, even if it's still a draft?

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Control.java
```diff
@@ -3527,6 +3527,26 @@ NSTouch findTouchWithId(NSArray touches, NSObject identity) {
 	return null;
 }
 
+/**
+ * Sets the autoscaling mode for this widget.
+ * <p>
+ * On Cocoa, autoscaling cannot be configured at the widget level.
+ * This method is therefore not supported on these platforms and has no effect.
```
HeikoKlare: On Cocoa and GTK? I see that the contracts differ across the platforms, but since consumers are usually OS-agnostic, there must be a uniform Javadoc for the same class on every OS,

I would phrase this the other way round and first point out the necessity to take into account that this functionality may be platform-specific and how to deal with it before actually mentioning which platform are supported (as usually you want to be platform-agnostic anyway, such that you do not care which of them are actually supported):
```suggestion
The capability is not supported on every platform, such that calling this method may not have an effect on unsupported platforms. The return value indicates if the autoscale mode was set properly. With {@link #isAutoScalable()}, the autoscale enablement can also be evaluated at any later pointin time.
<p>
Currently, this is only supported on Windows.
```

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Control.java
```diff
@@ -3527,6 +3527,26 @@ NSTouch findTouchWithId(NSArray touches, NSObject identity) {
...
+ * </p>
+ *
+ * @param autoscalingMode
+ *            the autoscaling mode to request; this argument is accepted but
+ *            ignored on this platform
+ *
+ * @return {@code false} to indicate that autoscaling mode cannot be changed
+ *         on GTK or Cocoa
```
HeikoKlare: I would not refer to specific platforms here but just say that it returns false if the operation was called on an unsupported platform.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Shell.java
```diff
@@ -1992,6 +1993,17 @@ public void setModified (boolean modified) {
...
 }
 
+/**
+ * Returns the native zoom of the device which is used globally by SWT.
+ *
+ * @return the native zoom for the current device
+ *
+ * @since 3.133
+ */
+public int getNativeZoom() {
```
HeikoKlare: In my opinion, we should call this `getZoom()`, as consumers should not bother with native/non-native zoms. And the effective zoom of a shell to deal with is always the native zoom, isn't it?

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/AutoscalingMode.java
```diff
@@ -0,0 +1,36 @@
...
+ *       adjusted according to the effective scaling factor.</li>
+ *
+ *   <li><b>{@link #DISABLED}</b> – Autoscaling is explicitly disabled.
+ *       Values will be used as-is without applying any scaling
+ *       transformations.</li>
+ *
+ *   <li><b>{@link #DISABLED_INHERITED}</b> – Autoscaling is disabled for this
+ *       component, but the decision originates from a parent or owner context.
+ *       This is typically used to indicate that the disabled state was not set
+ *       explicitly at this level but inherited from upstream configuration.</li>
```
HeikoKlare: Don't we need a mode to set on a control that the autoscale mode will (not) be inherited to childred of the control rather than to indicate that a control has inherited the mode? Since it's only used in a setter, how should the current semantics be reasonably applied anyway? And isn't that even how it is currently implemented?

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/AutoscalingMode.java
```diff
@@ -0,0 +1,36 @@
...
+ * </ul>
+ *
+ * @since 3.133
+ */
+
```
HeikoKlare:
```suggestion
```

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/AutoscalingMode.java
```diff
@@ -0,0 +1,36 @@
+package org.eclipse.swt.graphics;
+
+
+/**
+ * Defines the autoscaling behavior used when rendering or computing layout
+ * for widgets or graphics that support DPI‑aware scaling.
```
HeikoKlare: What is meant by "widgets or graphics"? This is only supported on `Control` isn't it?

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/AutoscalingMode.java
```diff
@@ -0,0 +1,36 @@
...
+ * <p>
+ * This mode determines whether autoscaling logic is applied, disabled, or
+ * inherited from a parent component. It allows callers to explicitly control
+ * how scaling should be handled in contexts where DPI, zoom factors, or
+ * user accessibility settings may influence size computation.
```
HeikoKlare: I find this a bit misleading as you cannot explicitly control how scaling is handled but what this mode is actually supposed to do it to define if SWT takes care of auto-scaling the control or not (and whether it should be inherited or not).

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Shell.java
amartya4256: That is not always true. When autoscaling is disabled getZoom returns autoscaled zoom depending on the scaling mode i.e. quarter, half, etc. Hence to do so, we need to rename the internal method getZoom to getAutoscaledZoom and then call this API `Shell#getZoom`. I'll create a separate PR.

amartya4256: https://github.com/eclipse-platform/eclipse.platform.swt/pull/2930

HeikoKlare:
@amartya4256 since the dependent PR is merged, this can/should be rebased on master now

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Control.java
```diff
@@ -3527,6 +3527,25 @@ NSTouch findTouchWithId(NSArray touches, NSObject identity) {
 	return null;
 }
 
+/**
+ * Sets the autoscaling mode for this widget.
+ * The capability is not supported on every platform, such that calling this method may not have an effect on unsupported platforms. The return value indicates if the autoscale mode was set properly. With {@link #isAutoScalable()}, the autoscale enablement can also be evaluated at any later pointin time.
```
HeikoKlare: "pointin" -> "point in"

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Control.java
```diff
@@ -3527,6 +3527,25 @@ NSTouch findTouchWithId(NSArray touches, NSObject identity) {
...
+ * <p>
+ * Currently, this is only supported on Windows.
+ * </p>
+ *
+ * @param autoscalingMode
+ *            the autoscaling mode to request; this argument is accepted but
+ *            ignored on this platform
```
HeikoKlare: "request" -> "set"?

What does "this argument is accepted but ignored on this platform" mean?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
```diff
@@ -1287,12 +1287,20 @@ public Object getData(String key) {
 public void setData(String key, Object value) {
 	super.setData(key, value);
 	if (DATA_AUTOSCALE_DISABLED.equals(key)) {
-		autoScaleDisabled = Boolean.parseBoolean(value.toString());
+		boolean autoScaleDisabled = Boolean.parseBoolean(value.toString());
 		if (autoScaleDisabled) {
+			Object autoscaleDisablementValue = getData(PROPOGATE_AUTOSCALE_DISABLED);
+			boolean propagateAutoscaling = autoscaleDisablementValue != null && Boolean.parseBoolean(autoscaleDisablementValue.toString());
+			autoscalingMode = propagateAutoscaling ? AutoscalingMode.DISABLED_INHERITED : AutoscalingMode.DISABLED;
 			this.nativeZoom = 100;
```
HeikoKlare: This code is there twice and could be reused via some `updateAutoScalingModeFromData()` method.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
```diff
@@ -1287,12 +1287,20 @@ public Object getData(String key) {
 public void setData(String key, Object value) {
 	super.setData(key, value);
 	if (DATA_AUTOSCALE_DISABLED.equals(key)) {
```
HeikoKlare: I propose to move all the autoscaling-data-specific code to a separate method and mark that one as deprecated for removal with the information that this hidden data objects will be removed in a future release and one has to migrate to the new API for that purpose. Same deprecation information should be added to constants "DATA_SHELL_ZOOM", "DATA_AUTOSCALE_DISABLED" and "PROPAGATE_AUTOSCALE_DISABLED".

HeikoKlare (commented):
I have not tested yet, but the code looks fine at a first glance. For now I have some comments regarding code style and documentation.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
```diff
@@ -78,12 +78,24 @@ public abstract class Control extends Widget implements Drawable {
...
+	/**
+	 * @deprecated use {@link Control#setAutoscalingMode(AutoscalingMode)} instead to set the autoscaling mode directly.
+	 */
+	@Deprecated(since = "2026-01", forRemoval = true)
 	private static final String DATA_AUTOSCALE_DISABLED = "AUTOSCALE_DISABLED";
 
+	/**
+	 * @deprecated use {@link Control#setAutoscalingMode(AutoscalingMode)} instead to set the autoscaling mode directly.
+	 */
+	@Deprecated(since = "2026-01", forRemoval = true)
```
akoch-yatta: The since should always be set according to the release cycle. This one would be included in "2026-03"

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
```diff
@@ -1286,13 +1298,27 @@ public Object getData(String key) {
...
+		updateAutoScalingModeFromData();
+	}
+}
+
+/**
+ * @deprecated use {@link Control#setAutoscalingMode(AutoscalingMode)} instead.
+ * These hidden data objects will be removed in a future release and one has to
+ * migrate to the new API for that purpose.
+ */
+@Deprecated(since = "2026-01", forRemoval = true)
```
akoch-yatta: Change since to "2026-03" as well

akoch-yatta (approved):
LGTM, I tested it with the current implementation in GEF that uses the existing DATA_ fields and adapted it with the new API methods to ensure it still works as expected

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
```diff
@@ -3353,6 +3385,31 @@ private void fitInParentBounds(Rectangle boundsInPixels, int zoom) {
...
+/**
+ * Sets the autoscaling mode for this widget.
+ * The capability is not supported on every platform, such that calling this method may not have an effect on unsupported platforms. The return value indicates if the autoscale mode was set properly. With {@link #isAutoScalable()}, the autoscale enablement can also be evaluated at any later point in time.
+ * <p>
+ * Currently, this is only supported on Windows.
+ * </p>
+ *
+ * @param autoscalingMode
+ *            the autoscaling mode to set; this argument is accepted but
+ *            ignored on this platform
```
akoch-yatta: can you adapt and unify this param. it was probably a leftover when the javadocs were different, but "this argument is accepted but ignored on this platform" is not really true

HeikoKlare (approved):

PR #2867, opened 2025-12-09
Throw ERROR_UNSUPPORTED_FORMAT when SVGRasterizer could not be loaded

ptziegler:
Trying to load an SVG while the JSVG rasterizer is not in the classpath throws a ServiceConfigurationError. This error should be caught and transformed into an SWTError, as specified by the contract of the ImageLoader.

By catching the ServiceConfigurationError, the RASTERIZER field is initialized with NULL. In this state, calling any method of the SVGFileFormat class automatically throws the expected SWTError.

To reproduce, load an SVG while `org.eclipse.swt.svg` is in the classpath, but `com.github.weisj.jsvg` is not. This call should fail with a:

> Provider org.eclipse.swt.svg.JSVGRasterizer could not be instantiated

Due to a NoClassDefFoundError when trying to load the JSVGRasterizer class, caused by the unresolvable dependency.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/image/SVGFileFormat.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/image/SVGFileFormat.java
@@ -31,8 +31,18 @@ import org.eclipse.swt.internal.DPIUtil.*;
 public class SVGFileFormat extends FileFormat {
 
 	/** The instance of the registered {@link SVGRasterizer}. */
-	private static final SVGRasterizer RASTERIZER = ServiceLoader
-			.load(SVGRasterizer.class, SVGFileFormat.class.getClassLoader()).findFirst().orElse(null);
+	private static final SVGRasterizer RASTERIZER;
+
+	static {
+		SVGRasterizer rasterizer = null;
+		try {
+			rasterizer = ServiceLoader
+					.load(SVGRasterizer.class, SVGFileFormat.class.getClassLoader()).findFirst().orElse(null);
+		} catch (ServiceConfigurationError e) {
+			// rasterizer not in classpath or could not be instantiated
+		}
+		RASTERIZER = rasterizer;
+	}
 
 	@Override
 	boolean isFileFormat(LEDataInputStream stream) throws IOException {


ptziegler:
We use this piece of code in GEF to determine whether SVGs are supported by SWT. Which fails for our snippets because of this.

```java
public static boolean isSvgSupported() {
    if (isSvgSupported != null) {
        return isSvgSupported;
    }
    String svg = """
            <svg width="1" height="1" version="1.1" viewBox="0 0 0 0" xmlns="http://www.w3.org/2000/svg"></svg>
            """; //$NON-NLS-1$
    try (InputStream is = new ByteArrayInputStream(svg.getBytes(StandardCharsets.UTF_8))) {
        new ImageLoader().load(is);
        isSvgSupported = true;
    } catch (IOException ignore) {
        // Should never happen
        isSvgSupported = false;
    } catch (SWTException e) {
        // SVGs unsupported
        isSvgSupported = false;
    }
    return isSvgSupported;
}
```

As you can see, `org.eclipse.swt.svg` is in the classpath of our example project, but `com.github.weisj.jsvg` is not. When the snippets are executed as normal Java applications, this error is then thrown.

image

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/image/SVGFileFormat.java
```diff
@@ -31,8 +31,18 @@
...
-			.load(SVGRasterizer.class, SVGFileFormat.class.getClassLoader()).findFirst().orElse(null);
+	private static final SVGRasterizer RASTERIZER;
+
+	static {
+		SVGRasterizer rasterizer = null;
+		try {
+			rasterizer = ServiceLoader
+					.load(SVGRasterizer.class, SVGFileFormat.class.getClassLoader()).findFirst().orElse(null);
+		} catch (ServiceConfigurationError e) {
+			// jSVG not in classpath
```
HeikoKlare: nit: the current rasterizer depending on JSVG is just one potential implementatin, but there could be others. So when this happens it is not necessarily JSVG but could be the dependency of any rasterizer implementation, So maybe the comment could be library-agnostic and just say "dependency of rasterizer is not in classpath" or the like?

HeikoKlare (commented):
Good catch! This works as expected and makes sense. I only have a nitpicky comment.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/image/SVGFileFormat.java
ptziegler: Good point. I've updated the comment.

HeikoKlare: great, thank you

PR #2868, opened 2025-12-10
Replacing usages of Display#getDPI in snippets

ShahzaibIbrahim:
Having the scale factor being based on the screen DPI leads to unexpected result e.g. Image too big/small. Having a screen dpi independent factor leads to consistent results


akoch-yatta (approved):
LGTM, small adjustments to get rid of Display#getDPI calls that do not provide proper values on windows with monitor specific scaling enabled

PR #2869, opened 2025-12-10
Removing usages of Display#getDPI from test

ShahzaibIbrahim:
This commit is made to prepare for Display#getDPI deprecation. removing/replacing all unnecessary usages.


tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_widgets_Display.java
```diff
@@ -1553,6 +1553,7 @@ public void test_wake() {
 /* custom */
 boolean disposeExecRan;
 
+@SuppressWarnings("deprecation")
```
HeikoKlare: This change does not deprecate any method, so why add a warning suppression here?

HeikoKlare (changes requested):
> Having the scale factor being based on the screen DPI leads to unexpected result e.g. Image too big/small. Having a screen dpi independent factor leads to consistent results

I do not understand how the commit/PR message is related to the actual change. There is no change in what DPI value is used at all. The change simply adds a faulty warning suppression and removes some text.

ShahzaibIbrahim:
Updated the message in PR and commit. SuppressWarning was added to test as Display#getDPI is set to be deprecated but it does make sense to do it when it is actually deprecated.

PR #2870, opened 2025-12-10
[Cocoa] Replacing usages of  Display#getDPI from cocoa

ShahzaibIbrahim:


ShahzaibIbrahim:
Just like in GTK, we have now only two remaining usages for getDPI in cocoa which we cannot be logically replaced. 
image

For later: we can deprecate the getDPI method now, but later we can keep the usage to internal by making getDPI package protected.

Currently this PR doesn't add anything to the deprecation of getDPI(). Hence closing it.

PR #2871, opened 2025-12-10
[GTK] Replacing usages of Display#getDPI from gtk

ShahzaibIbrahim:
This is a preparatory commit to remove all **possible** usages of getDPI from gtk. Since getDPI is set to be deprecated.

PR for Deprecation: https://github.com/eclipse-platform/eclipse.platform.swt/pull/2872

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Device.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Device.java
@@ -723,8 +723,8 @@ protected void init () {
 		}
 	}
 	defaultFont = OS.pango_font_description_copy (defaultFont);
-	Point dpi = getDPI(), screenDPI = getScreenDPI();
-	if (dpi.y != screenDPI.y) {
+	Point screenDPI = getScreenDPI();
+	if (this.dpi.y != screenDPI.y) {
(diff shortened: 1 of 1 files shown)


jonahgraham:
The GTK3 tests hung after 360m - I don't think that is because of your change and I have raised https://github.com/eclipse-platform/eclipse.platform.swt/issues/2876 to track that

ShahzaibIbrahim:
> > Having the scale factor being based on the screen DPI leads to unexpected result e.g. Image too big/small. Having a screen dpi independent factor leads to consistent results.
> 
> I do not understand how the commit/PR message relates to the actual change. What unexpected results to you refer to? Can you share an example?
> 
> And can you explain why it is correct to use a fixed DPI value of `96` in `Display`, even though the current `getDPI()` implementation has more sophisticated logic to determine the actual DPI? Should the implementation of `getDPI()` be changed then, in case the value is wrong?

After our discussion. I have changed the commit message and its content. Now we only get rid of getDPI usage only where its possible. For `Display` getDPI could be used for both `Device` and `Printer` both and because of this some places the value currently cannot be replaced.

For later: we can deprecate the getDPI method now, but later we can keep the usage to internal by making getDPI package protected.

HeikoKlare (approved):
Reusing the already stored DPI value instead of re-retrieving it is fine to avoid unnecessary multiple calls to `getDPI()`.

PR #2872, opened 2025-12-10
Deprecating Display#getDPI

ShahzaibIbrahim:

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Device.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Device.java
@@ -455,7 +455,18 @@ public int getDepth () {
  * @exception SWTException <ul>
  *    <li>ERROR_DEVICE_DISPOSED - if the receiver has been disposed</li>
  * </ul>
+ *
+ * @deprecated <p>This method returns a single global DPI value
+ * that does not reflect per-monitor DPI settings on modern operating systems.
+ * In environments with different scaling factors across monitors, it may provide
+ * a misleading or meaningless result, as it does not correspond to the actual DPI
+ * of any specific monitor.</p>
+ *
+ * <p>Note: While deprecated for general {@code Device} instances like {@code Display},
+ * this method may still be validly used when called on a {@code Printer} instance,
+ * where a single global DPI value is meaningful and expected.</p>
  */
+@Deprecated
 public Point getDPI () {
 	checkDevice ();
 	return getScreenDPI();
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Device.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Device.java
@@ -519,7 +519,18 @@ public int getDepth () {
  * @exception SWTException <ul>
  *    <li>ERROR_DEVICE_DISPOSED - if the receiver has been disposed</li>
  * </ul>
+ *
+ * @deprecated <p>This method returns a single global DPI value
+ * that does not reflect per-monitor DPI settings on modern operating systems.
+ * In environments with different scaling factors across monitors, it may provide
+ * a misleading or meaningless result, as it does not correspond to the actual DPI
+ * of any specific monitor.</p>
+ *
+ * <p>Note: While deprecated for general {@code Device} instances like {@code Display},
+ * this method may still be validly used when called on a {@code Printer} instance,
+ * where a single global DPI value is meaningful and expected.</p>
  */
+@Deprecated
 public Point getDPI () {
 	checkDevice ();
 	long hDC = internal_new_GC (null);
--- a/bundles/org.eclipse.swt/Eclipse SWT Printing/cocoa/org/eclipse/swt/printing/Printer.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Printing/cocoa/org/eclipse/swt/printing/Printer.java
@@ -623,6 +623,7 @@ public Point getDPI() {
 	}
 }
 
+@SuppressWarnings("deprecation")
 Point getIndependentDPI() {
 	return super.getDPI();
 }
--- a/bundles/org.eclipse.swt/Eclipse SWT Printing/win32/org/eclipse/swt/printing/PDFDocument.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Printing/win32/org/eclipse/swt/printing/PDFDocument.java
@@ -215,6 +215,7 @@ public class PDFDocument implements Drawable {
 		int screenDpiX = 96;
 		int screenDpiY = 96;
 		if (this.device != null) {
+			@SuppressWarnings("deprecation")
 			Point dpi = this.device.getDPI();
 			screenDpiX = dpi.x;
 			screenDpiY = dpi.y;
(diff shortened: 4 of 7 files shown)


bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Device.java
```diff
@@ -723,10 +735,9 @@ protected void init () {
 		}
 	}
 	defaultFont = OS.pango_font_description_copy (defaultFont);
-	Point dpi = getDPI(), screenDPI = getScreenDPI();
-	if (dpi.y != screenDPI.y) {
+	if (this.dpi.y != DOTS_PER_INCH) {
```
HeikoKlare: This removes the initialization of the `dpi` and `screenDPI` fields in `Device` while at least `dpi` will still be accessed afterwards. In Can you explain how this can be correct? In case those fields become obsolete, they may be removed, but simply not initializing them anymore seems wrong.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Device.java
```diff
@@ -519,7 +519,18 @@ public int getDepth () {
...
  * </ul>
+ *
+ * @deprecated <p>This method returns a single global DPI value
+ * that does not reflect per-monitor DPI settings on modern operating systems.
+ * In environments with different scaling factors across monitors, it may provide
+ * a misleading or meaningless result, as it does not correspond to the actual DPI
+ * of any specific monitor.</p>
+ *
+ * <p>Note: While deprecated for general {@code Device} instances like {@code Display},
+ * this method may still be validly used when called on a {@code Printer} instance,
```
HeikoKlare, line 530: Don't we need to deprecate the method in `Display` as well?

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Device.java
ShahzaibIbrahim: In this same method, on line 706. this.dpi is initialized with getDPI(). The semantics for this method hasn't changed. These are just two local variables being removed.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Device.java
ShahzaibIbrahim: Since getDPI is not implemented currently, that should be sufficient.

Image

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Device.java
HeikoKlare: I see, but then we effectively still have accesses to the deprecated `getDPI()` method, namely the initialization of the `dpi` field and all accesses to that field.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Device.java
HeikoKlare: I see, thank you. The method is not overwritten in `Display`, then it's of course fine.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Device.java
ShahzaibIbrahim: As discussed, we cannot remove few of the usages in Device for getDPI. Also moved the replacement changes in a  preparatory PR here: https://github.com/eclipse-platform/eclipse.platform.swt/pull/2871. 

This PR now only serves purpose to deprecate the method.

HeikoKlare (commented):
For the remaining usages of `Device#getDPI()`, this PR needs to adds according warnings suppressions.
On Windows, I currently see these three remaining usages of the method, of which I am not sure if the first two are expected:
image

akoch-yatta (approved):
LGTM, some required (internal) usages are guarded with SupressWarnings - I did not find any missed places.
The usages in PDFDocument must be removed and are not working properly in all combinations of zooms and autoscale modes, but simply removing it does not solve all issues. Therefor, it makes sense to have a holistic view on it in a second step to properly solve the existing issues.

Phillipus:
We access `getDPI`or scaling fonts depending on OS:

If this is deprecated can you suggest what to use instead?

Edit - I opened https://github.com/eclipse-platform/eclipse.platform.swt/issues/2925 with more details.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Device.java
```diff
@@ -455,7 +455,18 @@ public int getDepth () {
...
+ * that does not reflect per-monitor DPI settings on modern operating systems.
+ * In environments with different scaling factors across monitors, it may provide
+ * a misleading or meaningless result, as it does not correspond to the actual DPI
+ * of any specific monitor.</p>
+ *
+ * <p>Note: While deprecated for general {@code Device} instances like {@code Display},
+ * this method may still be validly used when called on a {@code Printer} instance,
+ * where a single global DPI value is meaningful and expected.</p>
  */
+@Deprecated
```
laeubi, line 469: The deprecation on Device is then wrong here in my opinion and Display should override the method and deprecate it individually. One might even define this for a display to return the smallest DPI setting of all monitors, or return the one of the _primary_ monitor.

laeubi:
Sorry for being late to this discussion—I somehow missed the PR initially and its broader implications. After reviewing the changes, I have serious concerns about the scope of this deprecation.

**The Problem with Deprecating `Device.getDPI()`**

While I understand the rationale regarding multi-monitor scenarios for `Display`, deprecating `getDPI()` at the `Device` level is overly broad and problematic. This method remains **critically important** for other `Device` subclasses:

1. **`Printer`** - DPI is essential for print operations (e.g., rasterizing images at the correct printer resolution)
2. **`PDFDocument`** - DPI matters for proper PDF rendering and scaling
3. **Platform-specific font scaling** - As @Phillipus demonstrated in [#2925](https://github.com/eclipse-platform/eclipse.platform.swt/issues/2925), applications rely on this for cross-platform font calculations

**Suggested Alternative Approach**

Instead of deprecating at the `Device` level, I propose:

1. **Keep `Device.getDPI()` non-deprecated** since it's valid and necessary for `Printer` and `PDFDocument` use cases

2. **Override and deprecate specifically in `Display`** with clear documentation: 
   ```java
   /**
    * @deprecated In multi-monitor environments, this method returns the DPI 
    * of the primary monitor only, which may not reflect the actual DPI of 
    * the monitor where your window is displayed.  
    * Use {@link #getDPI(Monitor)} to get monitor-specific DPI values.
    */
   @Override
   @Deprecated
   public Point getDPI() { ...  }
   ```

3. **Add a new method** to `Display`:
   ```java
   /**
    * Returns the DPI for the specified monitor.
    * @param monitor the monitor to query
    * @return the dots per inch of the specified monitor
    * @since 3.XXX
    */
   public Point getDPI(Monitor monitor) { ... }
   ```

4. **Document consistent behavior**:  Clarify that `Display.getDPI()` returns the DPI of the **primary monitor** to maintain backward compatibility while adding the new API for multi-monitor awareness.

**Migration Path**

This approach provides:
- ✅ Clear migration path for `Display` users to monitor-specific APIs
- ✅ No false deprecation warnings for legitimate `Printer`/`PDFDocument` use cases
- ✅ Backward compatibility with well-defined semantics
- ✅ Future-proof API for multi-monitor scenarios

Given that this PR has already been merged, I'd recommend considering a follow-up PR to refine the deprecation strategy along these lines.

Issue #2862, opened 2025-12-08
[Win32] Shell.getSize returns wrongly scaled values when dialog is opened on second monitor (monitor-specific scaling)


wahlbrink:
When opening a Dialog on another monitor, Shell.getSize() returns wrongly scaled values, if the monitors have different zoom factors. The size is computed by using the zoom factor of the monitor of the main app instead of the monitor on which the dialog is located.

**To Reproduce**
I noticed it in the IDE (requires two monitors with different zoom):

* Open the Eclipse IDE on the first screen.
* Open the preference dialog and move it to the second screen (size is correctly adapted)
* Close the preference dialog (wrong size is saved)
* Reopen the preference dialog (dialog has wrong size)

If monitor 1= 200% and monitor 2= 100%, the dialog shrinks to half its size each time.

**Short Analysis**
Shell.getSize() is implemented in Control. It uses `computeGetBoundsZoom()` which return the zoom of the parent control. For instances of Shell, the use of the parent seems to be wrong?

So
 * I guess it affects only dialogs with parentShell set.
 * Maybe also other methods which uses `computeGetBoundsZoom` or `computeBoundsZoom` work incorrectly.

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 

2. Additional OS info (e.g. OS version, Linux Desktop, etc)
Windows 11

3. JRE/JDK version
JDK 21 (Adoptium)

**Version since**
Eclipse 4.38 Build id: I20251201-0920

It was working correctly in 2025-09/4.37.

HeikoKlare:
Thank you for that report. That definitely seems to be a bug and an oversight in the implementations of the `computeGetBoundsZoom()` and `computeBoundsZoom()` methods. At a first glance, just overwriting them in `Shell` to return the shell's zoom should fix the issue.

vogella:
I think I saw this also recently at a client with his splash screen. Thanks for reporting and fixing this.

HeikoKlare:
@vogella can you give more details on that? I am not sure if/how this can affect the splash screen as well (as I don't think any child shells are involved there), so maybe that's a different issue?

vogella:
The client RCP used a splash handler and received the shell from the splash (get splash().get Shell()). The widgets were initially correctly sized but the first update of the progress bar changed their sizing incorrectly.

[...]

PR #2873, opened 2025-12-10
[Win32] Correct bounds/size value of child shells

HeikoKlare:
In point/pixel conversions for bounds/sizes of controls, the parent control's zoom is used to adhere to the most appropriate context when having auto-scale disabled controls. This, however, leads to child shells using the zoom of the parent shells, which can be placed on a different monitor with a different zoom, such that the bounds/sizes of such child shells are wrongly scaled.

With this change, shells always use their own zoom for bounds/sizes calculations instead of the one of a potential parent shell.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2862

### How to test
As described in https://github.com/eclipse-platform/eclipse.platform.swt/issues/2862:
> I noticed it in the IDE (requires two monitors with different zoom):
> 1. Open the Eclipse IDE on the first screen.
> 2. Open the preference dialog and move it to the second screen (size is correctly adapted)
> 3. Close the preference dialog (wrong size is saved)
> 4. Reopen the preference dialog (dialog has wrong size)
>
> If monitor 1= 200% and monitor 2= 100%, the dialog shrinks to half its size each time.

With primary monitor 100% and secondary 175% (and the preferences opening on secondary), it looks like this:

#### Before
childshell_wrongsize

#### After
childshell_correctsize

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/ControlWin32Tests.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/ControlWin32Tests.java
@@ -356,4 +356,16 @@ class ControlWin32Tests {
 		assertTrue(clientAreaInPixels.height <= childBoundsInPixels.y + childBoundsInPixels.height);
 	}
 
+	@Test
+	void testChildShellGetSize() {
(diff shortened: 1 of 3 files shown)


akoch-yatta (approved):
LGTM, I tested with the runtime and the issue is fixed as described.

vogella:
LGTM

PR #2879, opened 2025-12-12
Adjusted computation of horizontal toolbar size.

ShahzaibIbrahim:
The first part ensures the toolbar becomes tall enough to fit controls like Combo or Text, whose height is often larger than the native toolbar button height returned by TB_GETITEMRECT. Without this check, the toolbar height is based only on Windows’ button metrics, which are smaller, causing embedded controls to be clipped. By taking the max of the toolbar height and each control’s actual height, the final toolbar size always accommodates the tallest embedded control.

In the second part, the original code centered the control vertically, but when the control (like a combo box) was taller than the toolbar item height, the centering calculation produced a negative offset, shifting the control upward and causing clipping. Your fix uses Math.max(0, …) to prevent negative offsets, ensuring the control is never positioned above the item’s top. As a result, taller controls are aligned safely without being cut off.

### To Reproduce

- Open ControlExample
- Open Toolbar Tab
- Enable ComboChild

**Expected behavior**
Combobox is displayed correctly

### Result
---
**Before Fix**
image
---
**After Fix**
image

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/935

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/ToolBar.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/ToolBar.java
@@ -255,6 +255,12 @@ Point computeSizeInPixels (Point hintInPoints, int zoom, boolean changed) {
 			OS.SendMessage (handle, OS.TB_GETITEMRECT, count - 1, rect);
 			width = Math.max (width, rect.right);
 			height = Math.max (height, rect.bottom);
+			// Adjust height for embedded controls (Combo, Text, etc.)
+			for (ToolItem item : items) {
+			    if (item != null && item.getControl() != null) {
+			        height = Math.max(height, item.getControl().getBoundsInPixels().height);
+			    }
+			}
 		}
 		OS.SetWindowPos (handle, 0, 0, 0, oldWidth, oldHeight, flags);
 		if (redraw) OS.ValidateRect (handle, null);
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/ToolItem.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/ToolItem.java
@@ -583,7 +583,7 @@ void resizeControl () {
 		control.setSize (itemRect.width, itemRect.height);
 		Rectangle rect = control.getBounds ();
 		rect.x = itemRect.x + (itemRect.width - rect.width) / 2;
-		rect.y = itemRect.y + (itemRect.height - rect.height) / 2;
+		rect.y = itemRect.y + Math.max(0, itemRect.height - rect.height) / 2;
 		control.setLocation (rect.x, rect.y);
 	}
 }


tomaswolf:
If such a toolbar is used now in a view stack, do we get https://github.com/eclipse-platform/eclipse.platform.swt/pull/936#issuecomment-1854073938 again?

(If you need a view with such a toolbar, try EGit's staging view or history view. When the window is wide enough so that there is enough horizontal space to the right of the view stack's tabs to accomodate the toolbar's width, the toolbar should be on the same line as the tabs.)

If that layout issue re-appears: maybe the CTabFolder needs layout changes.

HeikoKlare:
> If you need a view with such a toolbar, try EGit's staging view or history view.

I also went through the EGit views when trying to find a toolbar that we might easily use for testing, but none of those seems to contain a combo, do they? We only notices this issue with combos so far, but they seem to be rarely used in toolbars. Only usage I know about is the zoom field of GEF.

tomaswolf:
> none of those seems to contain a combo, do they?

That's right. They do contain text input fields, though. If the issue occurs only with combos, that won't help then. IIRC the text inputs are also cut off at the bottom on Linux (but this change here is Windows only).

laeubi:
In general Toolbar with controls are often work quite poor and the best "workaround" is to add a large enough image (what the would make the toolbar higher). A similar issue (and maybe easier to reproduce?) occurs when one adds buttons with images of different size, e.g.  add a tool item with a 16x16 image and one with a 32x32 image then both are shrink to 16x16px.

By the way according to the image, the vertical (!) size is to small or do I misunderstand the issue?

arunjose696:
I have tested this change and could see this fixes the issue in GEF zoom field. I have a question about the changes in the `toolitem` and the below statement in the description


> In the second part, the original code centered the control vertically, but when the control (like a combo box) was taller than the toolbar item height,

I dont understand how this can happen as in the current code the height of tool item is the max height of all the controls, so why is the second part required in toolitem to adjust toolitems y?

ShahzaibIbrahim:
> I have tested this change and could see this fixes the issue in GEF zoom field. I have a question about the changes in the `toolitem` and the below statement in the description
> 
> > In the second part, the original code centered the control vertically, but when the control (like a combo box) was taller than the toolbar item height,
> 
> I dont understand how this can happen as in the current code the height of tool item is the max height of all the controls, so why is the second part required in toolitem to adjust toolitems y?

@arunjose696  In the case of a combo box, its height can exceed the height of the toolbar. The existing logic vertically centers the control by computing an offset based on the height difference. When the combo box is taller than the toolbar, this calculation produces a negative Y offset, causing the control to be positioned partially outside the toolbar and resulting in the top being clipped. By using Math.max(0, itemRect.height - rect.height) / 2, we clamp the offset to a non-negative value, ensuring the control is never positioned at a negative Y coordinate and preventing the top portion from being cut off.

akoch-yatta (commented):
I tested this PR in multiple scenarios and it improves the behavior, but I still have a wrongly sized toolbar after the initial creation in my setup:

Image

The toolbar is from GEF e.g. visible in the Logic Editor. If I move the shell to another monitor and back to the original it looks fine, if the combo was in the toolbar at this time:

Image

If not, it looks the same. What happens here is probably, that the ToolItem was added after the toolbar was created and sized. Question is, if this can be fixed in a reasonable effort.

HeikoKlare:
As this includes the same change as https://github.com/eclipse-platform/eclipse.platform.swt/pull/936, it produces the same rendering error as reported there: https://github.com/eclipse-platform/eclipse.platform.swt/pull/936#issuecomment-1854073938

See this toolbar which would fit into the header row but is still unexpectedly displayed below (due to the combo box being sized differently):
image

ShahzaibIbrahim:
> As this includes the same change as #936, it produces the same rendering error as reported there: [#936 (comment)](https://github.com/eclipse-platform/eclipse.platform.swt/pull/936#issuecomment-1854073938)
> 
> See this toolbar which would fit into the header row but is still unexpectedly displayed below (due to the combo box being sized differently): image

Since I have no reproducer of this case and no better solution for this layout issue, I will close this PR for now.

PR #2889, opened 2025-12-17
  Adapt Clabel to use new GC#drawImage API

arunjose696:
This change replaces the old API call in CLabel with the newer, simplified drawImage overload that takes fewer parameters and no longer requires the caller to provide the image dimensions.

### Steps to reproduce
1)Run CustomControlExample 
2)The below images are drawn with the drawImage API they remain the same as before

image


akoch-yatta (approved):
LGTM, it simplifies the call and I tested against master.

PR #2890, opened 2025-12-17
 Adapt  fillGradientRectangle to new GC#drawImage API

arunjose696:
This change replaces the old drawImage API call in fillGradientRectangle with the newer, simplified drawImage overload that takes fewer parameters and replaces the imageDataProvider with imageDataAtSizeProvider.

Steps to reproduce can be found in the below PR
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2284


akoch-yatta (approved):
LGTM, tested against master as mentioned in the PR

PR #2894, opened 2025-12-18
 Adapt  tests to use new GC#drawImage API wherever a full image is drawn

arunjose696:
This change replaces the old drawImage API call in tests with the newer, simplified drawImage overload that takes fewer parameters


akoch-yatta (approved):
LGTM, changes simplify the calling places

PR #2895, opened 2025-12-18
[Gtk4] Fix Shell.computeSize to take header bar into account

akurtakov:
HeaderBar is part of the regular window size  in Gtk 4 on Wayland BUT gtk_window_get_headerbar returns 0 if it is not set by the application but drawn internally by Gtk itself and thus it's size can not be queried. Add a "default" GtkHeaderBar explicitly for that and take its size into account when computing Shell size.


akurtakov:
This fixes initial size of Eclipse main window and dialogs, resizing is still broken. Fixes github.com/eclipse-platform/eclipse.platform.swt/issues/2826

akurtakov:
Merging to be able to continue with more noticeable problems.

Issue #2876, opened 2025-12-11
[GTK] Lots of GTK3 test runs hanging at potentially the same point


jonahgraham:
I see regular runs that hang on GitHub actions and the last line of output is:

```
[INFO] Running org.eclipse.swt.tests.junit.Test_org_eclipse_swt_custom_BusyIndicator
```

Such as https://github.com/eclipse-platform/eclipse.platform.swt/actions/runs/20100393887/job/57669450861?pr=2871

If we see this enough I will look into it more. For now this issue is used to track so I know how significant this problem really is.

PR #2903, opened 2025-12-20
Add timeout to Test_org_eclipse_swt_custom_BusyIndicator

laeubi:
It seem recently Test_org_eclipse_swt_custom_BusyIndicator is hang in the build forever sometimes.

This now adds a timeout so we can see if it is really hanging there or somewhere else.


jonahgraham:
I think this fixes (or at least worksaround) https://github.com/eclipse-platform/eclipse.platform.swt/issues/2876

laeubi:
> I think this fixes (or at least worksaround) #2876

Yes its a workaround for a complete hang... so not really a fix, but then hopefully will only fail this single test.

akurtakov:
https://download.eclipse.org/eclipse/downloads/drops4/I20251221-1800/testresults/html/org.eclipse.swt.tests_ep439I-unit-linux-x86_64-java21_linux.gtk.x86_64_21.html looks as a result of this one.

PR #2905, opened 2025-12-22
Implemented Double-Checked Locking in ImagehandleManager

amartya4256:
This PR handles the creation of handles / cached ImageData per zoom for Image objects in a thread-safe way by the means of double-checked locking. The implementation extends the ImageHandleManager in the following way:
- ImageHandleManager#getHandleOrCreate now uses thread-safe compute methods
- `ImageHandleManager#executeOnHandle` also uses `ConcurrentHashMap#compute` for thread-safety. This is utilized while calling `getImageData` to ensure thread safety.

Depends on https://github.com/eclipse-platform/eclipse.platform.swt/pull/2964
This fixes https://github.com/vi-eclipse/Eclipse-Platform/issues/562

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
@@ -18,7 +18,8 @@ import static org.eclipse.swt.internal.image.ImageColorTransformer.DEFAULT_DISAB
 
 import java.io.*;
 import java.util.*;
-import java.util.Map.*;
+import java.util.concurrent.*;
+import java.util.concurrent.atomic.*;
 import java.util.function.*;
 import java.util.stream.*;
 
@@ -138,7 +139,7 @@ public final class Image extends Resource implements Drawable {
 	private final ImageHandleManager imageHandleManager = new ImageHandleManager();
 
 	private class ImageHandleManager {
-		private Map<Integer, DestroyableImageHandle> zoomLevelToImageHandle = new HashMap<>();
+		private Map<Integer, DestroyableImageHandle> zoomLevelToImageHandle = new ConcurrentHashMap<>();
 
 		InternalImageHandle get(int zoom) {
 			final DestroyableImageHandle imageHandle = zoomLevelToImageHandle.get(zoom);
@@ -153,15 +154,7 @@ public final class Image extends Resource implements Drawable {
 			if (zoom == -1) {
 				return null;
 			}
-
-			DestroyableImageHandle imageHandle = (DestroyableImageHandle) get(zoom);
-			if (imageHandle != null) {
-				return imageHandle;
-			}
-
-			imageHandle = creator.get();
-			zoomLevelToImageHandle.put(zoom, imageHandle);
-			return imageHandle;
+			return zoomLevelToImageHandle.computeIfAbsent(zoom, __ -> creator.get());
 		}
 
 		boolean contains(int zoom) {
@@ -181,16 +174,23 @@ public final class Image extends Resource implements Drawable {
 		}
 
 		void destroyHandles(Predicate<Integer> filter) {
-			Iterator<Entry<Integer, DestroyableImageHandle>> it = zoomLevelToImageHandle.entrySet().iterator();
-			while (it.hasNext()) {
-				Entry<Integer, DestroyableImageHandle> zoomToHandle = it.next();
-				if (filter.test(zoomToHandle.getKey())) {
-					DestroyableImageHandle imageHandle = zoomToHandle.getValue();
-					it.remove();
-					zoomLevelToImageHandle.remove(imageHandle.zoom(), imageHandle);
-					imageHandle.destroy();
-				}
-			}
+		    zoomLevelToImageHandle.entrySet().removeIf(entry -> {
+		    	if (filter.test(entry.getKey())) {
+		    		entry.getValue().destroy();
+		    		return true;
+		    	}
+		    	return false;
+		    });
+		}
+
+		<R> R executeOnHandle(int zoom, Function<AtomicReference<DestroyableImageHandle>,R> executable) {
+			AtomicReference<R> computedValue = new AtomicReference<>();
+			zoomLevelToImageHandle.compute(zoom, (key, value) -> {
+				AtomicReference<DestroyableImageHandle> handleReference = new AtomicReference<>(value);
+				computedValue.set(executable.apply(handleReference));
+				return handleReference.get();
+			});
+			return computedValue.get();
 		}
 
 		@Override
@@ -1304,12 +1304,13 @@ public Rectangle getBounds() {
 
 Rectangle getBounds(int zoom) {
 	if (isDisposed()) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
-	if (imageHandleManager.contains(zoom)) {
-		InternalImageHandle imageMetadata = imageHandleManager.get(zoom);
-		Rectangle rectangle = new Rectangle(0, 0, imageMetadata.width(), imageMetadata.height());
-		return Win32DPIUtils.scaleBounds(rectangle, zoom, imageMetadata.zoom());
-	}
-	return this.imageProvider.getBounds(zoom);
+	return imageHandleManager.executeOnHandle(zoom, imageHandle -> {
+		if (imageHandle.get() != null) {
+			Rectangle rectangle = new Rectangle(0, 0, imageHandle.get().width(), imageHandle.get().height());
+			return Win32DPIUtils.scaleBounds(rectangle, zoom, imageHandle.get().zoom());
+		}
+		return this.imageProvider.getBounds(zoom);
+	});
 }
 
 /**
@@ -1383,15 +1384,16 @@ public ImageData getImageData() {
  */
 public ImageData getImageData (int zoom) {
 	if (isDisposed()) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
-	InternalImageHandle imageHandle = imageHandleManager.get(zoom);
-	if (imageHandle != null) {
-		return imageHandle.getImageData();
-	}
-
-	if (this.imageProvider.isPersistentImageHandleRequriedForImageData()) {
-		return imageHandleManager.getOrCreate(zoom, () -> imageProvider.newImageHandle(new ZoomContext(zoom))).getImageData();
-	}
-	return this.imageProvider.newImageData(zoom);
+	return imageHandleManager.executeOnHandle(zoom, imageHandle -> {
+		if (imageHandle.get() != null) {
+			return imageHandle.get().getImageData();
+		}
+		if (this.imageProvider.isPersistentImageHandleRequriedForImageData()) {
+			imageHandle.set(imageProvider.newImageHandle(new ZoomContext(zoom)));
+			return imageHandle.get().getImageData();
+		}
+		return this.imageProvider.newImageData(zoom);
+	});
 }
 
 /**


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1938,19 +1998,19 @@ public void setBackground(Color color) {
...
 
 <T> T applyUsingAnyHandle(Function<ImageHandle, T> function) {
-	if (zoomLevelToImageHandle.isEmpty()) {
+	if (imageHandleManager.hasNoBaseHandle()) {
 		ImageHandle temporaryHandle = this.imageProvider.newImageHandle(new ZoomContext(DPIUtil.getDeviceZoom(), DPIUtil.getNativeDeviceZoom()));
 		try {
 			return function.apply(temporaryHandle);
 		} finally {
-			temporaryHandle.destroy();
+			imageHandleManager.destroy(temporaryHandle);
```
HeikoKlare: Why do you need to destroy the handle via the handle manager even though it was not created via the manager (so it will not be present in its handle map anyway)?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2103,7 +2163,7 @@ ImageData newImageData(int zoom) {
 		Function<Integer, ImageData> imageDataRetrieval = zoomToRetrieve -> {
 			ImageHandle handle = initializeHandleFromSource(new ZoomContext(zoomToRetrieve));
 			ImageData data = handle.getImageData();
-			handle.destroy();
+			imageHandleManager.destroy(handle);
```
HeikoKlare: Same as above: why do you need to destroy the handle via the handle manager even though it was not created via the manager (so it will not be present in its handle map anyway)?

This applies to all `newImageData` methods.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2290,17 +2350,18 @@ protected ElementAtZoom<ImageData> loadImageData(int zoom) {
...
 				GC currentGC = memGC;
 				memGC = null;
-				createHandle(targetZoom);
-				currentGC.refreshFor(new DrawableWrapper(Image.this, zoomContext));
+				ImageHandle imageHandle = createHandle(targetZoom);
+				currentGC.refreshFor(new DrawableWrapper(Image.this, zoomContext), imageHandle);
+				return imageHandle;
 			}
-			return zoomLevelToImageHandle.get(targetZoom);
+			return imageHandleManager.getImageHandle(targetZoom);
```
HeikoKlare: This code looks a bit weird (but also did before) because we call a _new_ImageHandle method and then return an existing one. From my understanding, this line of code should never be reached anyway. In case there is a `memGC`, there should also be an image handle for it, so that `newImageHandle` will never be called unless the zoom of that handle is not fitting (and then the if block above is executed).

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -1325,8 +1393,9 @@ public ImageData getImageData() {
  */
 public ImageData getImageData (int zoom) {
 	if (isDisposed()) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
-	if (zoomLevelToImageHandle.containsKey(zoom)) {
-		return zoomLevelToImageHandle.get(zoom).getImageData();
+	ImageHandle imageHandle = imageHandleManager.getImageHandle(zoom);
+	if (imageHandle != null) {
+		return imageHandle.getImageData();
```
HeikoKlare: This code is still prone to race conditions: there might not be a handle yet, then another thread comes and creates that handle, but since we have not seen that handle here, we call `imageProvider.newImage`, which will expensively create another handle, extract image data out of it and then cache that image data (although the cached image data will never be used as there is a handle already, so that part of memory is effectively dead). Is that intended?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -6059,10 +6068,10 @@ private void removePreviousOperationIfSupercededBy(Operation operation) {
 	}
 }
 
-private void createGcHandle(Drawable drawable, GCData newData) {
+private void createGcHandle(Drawable drawable, GCData newData, ImageHandle imageHandle) {
 	long newHandle = drawable.internal_new_GC(newData);
 	if (newHandle == 0) SWT.error(SWT.ERROR_NO_HANDLES);
-	init(drawable, newData, newHandle);
+	initWithImageHandle(drawable, newData, newHandle, imageHandle);
```
HeikoKlare: Why is this change to `GC` necessary? Is it a bug in the existing implementation?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -132,18 +133,100 @@ public final class Image extends Resource implements Drawable {
...
+	private class ImageHandleManager {
+		private Map<Integer, ImageHandle> zoomLevelToImageHandle = new ConcurrentHashMap<>();
+		private Set<ImageHandle> imageHandles = new HashSet<>();
+
+		ImageHandle getImageHandle(int zoom) {
+			final ImageHandle imageHandleFuture = zoomLevelToImageHandle.get(zoom);
+			if (imageHandleFuture == null) {
+				return null;
+			}
+			return imageHandleFuture;
```
HeikoKlare: There is no future involved here and the value of `zoomLevelToHandle.get(zoom)` could directly be returned, couldn't it?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -132,18 +133,100 @@ public final class Image extends Resource implements Drawable {
...
 
-	private Map<Integer, ImageHandle> zoomLevelToImageHandle = new HashMap<>();
-
 	private List<Consumer<Image>> onDisposeListeners;
 
+	private final ImageHandleManager imageHandleManager = new ImageHandleManager();
+
+	private class ImageHandleManager {
+		private Map<Integer, ImageHandle> zoomLevelToImageHandle = new ConcurrentHashMap<>();
+		private Set<ImageHandle> imageHandles = new HashSet<>();
```
HeikoKlare: Why do we need two separate data structure managing image handles? They are managed in different ways and will mostly (but not competely) contains the same handles.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -132,18 +133,100 @@ public final class Image extends Resource implements Drawable {
...
+
+		boolean hasImageHandle(ImageHandle imageHandle) {
+		    return zoomLevelToImageHandle.values().contains(imageHandle);
+		}
+
+		boolean hasZoom(int zoom) {
+			return zoomLevelToImageHandle.containsKey(zoom);
+		}
+
+		boolean hasNoBaseHandle() {
```
HeikoKlare: Should this be `hasNoHandle()`? It's unclear what a "base" handle is supposed to be.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -132,18 +133,100 @@ public final class Image extends Resource implements Drawable {
...
+				}
+			}
+		}
+
+		void destroy(ImageHandle imageHandle) {
+			zoomLevelToImageHandle.remove(imageHandle.zoom, imageHandle);
+			imageHandle.destroy();
+			imageHandles.remove(imageHandle);
+			cleanupDanglingImageHandles();
+		}
```
HeikoKlare: This method feels wrong to me: either the `ImageHandleManager` manages the handles and does not auto destroy (with the other method accepting a Predicate) or a handle is managed separately, but then it should not be artificially be passed over this manager.

If I am not mistaken, at least the `imageHandles` set, the `register` method and the `destroy(ImageHandle)` should be removed.

HeikoKlare (commented):
I really like the idea to wrap the handle-for-zoom mapping into a separate class. That clearly assigns all related management functionality to that class. I would be in favor of extracting the pure refactoring of the management functionality in a dedicated class into a separate PR. That PR will just improve code quality and comprehensibility and not improve thread safety.

You find several questions and proposals in the detailed comments. The probably most essential point is that to my understanding the implementation is not fully thread safe.

In addition, I am not sure if the concept of using a synchronized data structure is fully sufficient for all cases here (e.g., if we also include the rest of the API such as `getBounds()` and `getImageData()`). At least the complexity could increase such that ensuring correctness and maintainability becomes complex. Maybe we better see that when we adapt those methods as well, but I could imagine that simply synchronizing write access might be sufficient (as we don't expect multiple consumers to write for different zooms) and provides reduced compexity and error proneness.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
amartya4256: It will. All the ImageHandles are registered in the manager. It is in the constructor of ImageHandle. Maybe we should only allow creation of ImageHandle using the manager then. This will simplify it. We want to register all the ImageHandles as we actively perform cleanup of dangling ImageHandles.

amartya4256: The dangling ImageHandles can be those which were created temporarily for obtaining the ImageData or for some other purpose. We want to track all the ImageHandles in the Manager.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
amartya4256: Yes this method is created from the earlier implementation of the PlainImageDataProvider as in case it uses GC to scale, it uses this specific method. In the init method of GC, it uses the Image.win32_getHandle, which would want to create a handle and since we already are in the middle of creation of the handle it will fail. Since we also had a specific method in the past for this case, the idea is to use the ImageHandle we just obtained and pass that directly instead of calling `win32_getHandle`.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
amartya4256: My bad, forgot to rename it. Went with approach of Futures earlier.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
amartya4256: The Set contains all the distinct imageHandles created for the image temporary and permanent ones. Any dangling imageHandle can be tracked here and disposed upon cleanups.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
amartya4256: Fair point.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
amartya4256: As discussed, fixed it using a synchronized block with before and after check

laeubi:
I have not looked into the full details here but using `syncronized` in SWT code looks wrong. The whole SWT code assumes it is always executed single threaded on the display thread, so where are different Threads are coming from here? If it is because of multiple `Display`s I think the cache should more be per display and avoid any complicate synchronization here.

HeikoKlare:
The point is that resources used to be accessible from any thread (even though it is not recommended, but it worked and consumers used it) and this behavior broke when introducing the multi-handle support. This change is just restoring that capability.

laeubi:
> The point is that resources used to be accessible from any thread (even though it is not recommended, but it worked and consumers used it) and this behavior broke when introducing the multi-handle support

Looking at the `Resource` (and its subclasses) I can't find any sign neither in code nor API that it is safe to access these from different threads.

If we still want to support this I think synchronized primitives are not the right way and instead concurrent data-structures must be used instead and we are prone to deadlocks here.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
laeubi:
```suggestion
		private Map<Integer, DestroyableImageHandle> zoomLevelToImageHandle = new ConurrentHashMap<>();
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -159,9 +159,20 @@ InternalImageHandle getOrCreate(int zoom, Supplier<DestroyableImageHandle> creat
...
-			return imageHandle;
+			synchronized (zoomLevelToImageHandle) {
+				imageHandle = (DestroyableImageHandle) get(zoom);
+				if (imageHandle == null) {
+					imageHandle = creator.get();
+					zoomLevelToImageHandle.put(zoom, imageHandle);
+				}
+				return imageHandle;
+			}
+		}
```
laeubi: Use `zoomLevelToImageHandle.computeIfAbsent(...)` or `compute(...)` on the concurrent map instead of own synchronization primitives

amartya4256: https://github.com/eclipse-platform/eclipse.platform.swt/commit/93108dc7d4
That's how I pretty much started off earlier but since we also need to implement this for `getImageData` which either retrieves it from `ImageHandle#getImageData` or creates an `ImageData` and caches it; we needed different mechanism and overall it made sense to use a consistent locking mechanism rather than using 2 different approaches in 2 different places. That's why we decided to go with double-checked locking using `synchronized`.

laeubi: I must confess I don't understand the befit of this approach and why it should not work for `getImageData`.

Assume `Thread 1` is entering here into the map.computeIfAbsent.

Now `Thread 2` is entering  `getImageData` and will be blocked on the get call until `Thread 1` completes, what has the same effect as `Thread 1` entering the synchronized and `Thread 2` is hitting the `null` and then entering the synchronization and needs to wait until Thread 1 finishes.

The main difference here is, that if we have more threads then with your approach ALL threads will be blocked regardless of zoomlevel (both in `getImageData` and `getOrCreate`) even if they aim for _different_ zoom levels!

... not talking even about the case that we have a race condition between two threads one that is removing it an one that is getting it (what will make a possibly destroyed image handle visible to the getting thread.

So actually I think one even needs to use `Map.compute` here and do two checks: If current value is `null` or current value is _disposed_ we need to create a new handle...

HeikoKlare: > why it should not work for `getImageData`.

`getImageData()` is not doing an atomic call on the `ImageHandleManager`, which would be necessary to make a synchronization via the data structure. Maybe there is some possibility to refactor the code such that it is wrapped into an atomic data structure access.

> ALL threads will be blocked regardless of zoomlevel (both in `getImageData` and `getOrCreate`) even if they aim for _different_ zoom levels

That's correct but not a problem at all. This synchronization is only necessary to avoid issues due to the extremely rare case of a race condition, so there is no necessity to ensure that access to the data structure with different zooms concurrently works in an efficient way.

HeikoKlare:
> Looking at the `Resource` (and its subclasses) I can't find any sign neither in code nor API that it is safe to access these from different threads.

That's true, there is nothing explicitly implemented to ensure thread-safety. However, most of the functionality was effectively thread-safe as after object initialization most of the state was effectively final (in particular the handles themselves), such that without any additional effort all access scenarios that did not modify state worked find in a concurrent access scenario.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -181,15 +186,24 @@ Set<Integer> getAllZooms() {
...
 		void destroyHandles(Predicate<Integer> filter) {
-			Iterator<Entry<Integer, DestroyableImageHandle>> it = zoomLevelToImageHandle.entrySet().iterator();
-			while (it.hasNext()) {
-				Entry<Integer, DestroyableImageHandle> zoomToHandle = it.next();
-				if (filter.test(zoomToHandle.getKey())) {
-					DestroyableImageHandle imageHandle = zoomToHandle.getValue();
-					it.remove();
-					zoomLevelToImageHandle.remove(imageHandle.zoom, imageHandle);
-					imageHandle.destroy();
+			executeOnHandle(0, __ -> {
```
HeikoKlare: This sounds semantically incorrect. You do not execute anything on a handle for zoom 0, but you want somehow miuse that method to synchronize on the handler. You would either have to iterate over the keys on perform an `executeOnHandle` for each each to process, or you should just synchronize the call here.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -181,15 +186,24 @@ Set<Integer> getAllZooms() {
...
+						imageHandle.destroy();
+					}
 				}
+				return null;
+			});
+		}
+
+		<R> R executeOnHandle(int zoom, Function<Optional<InternalImageHandle>,R> execution) {
+			synchronized (this) {
+					return execution.apply(Optional.ofNullable(get(100)));
```
HeikoKlare:
```suggestion
					return execution.apply(Optional.ofNullable(get(zoom)));
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
amartya4256: Yes I thought of using the same method for this as well. It is indeed a hack. So, I'll just use a synchronized block here instead, just like execute on handle.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
laeubi: I must confess I don't understand the "not doing an atomic call" it is not doing an "atomic call" either now and why double checked locking should work/help here it all.

I have already outlined why I think my approach works and that it is semantically working the same (except the problematic locking), so maybe one can explain what is the real case to solve here and what threads are involved and why a synchronized works better than a plain get.

If it is "extremely rare" the only issue is that maybe two image data are created (what are not cached anyways) so I would argue this is acceptable in this situation as it only might hurt performance a very little bit in "extreme rare cases".

HeikoKlare: > I must confess I don't understand the "not doing an atomic call" it is not doing an "atomic call" either now and why double checked locking should work/help here it all.

With the used locking/synchronization, we ensure that the effect of the synchronized code block is atomically visible to other threads that synchronize on the manager as well. They will not operate on any intermediate state. When you first retrieve the handle from the manager and, if it is not present, execute some other code based on that decision, this may lead to a race condition with another thread accessing the manager. What precisely will happen is that another thread creates the handle, the thread trying to access the image data will create image data based on a temporary handle (so far it's just inefficient) but also caches the created image data without them ever being consumed again (this is the erroneous state that will be reached), as the image data is only removed from the cache when an actual handle is created out of it, which will not happen as the other thread concurrently created that handle already. That's also why this is not acceptable contrary to:
> If it is "extremely rare" the only issue is that maybe two image data are created (what are not cached anyways) so I would argue this is acceptable in this situation as it only might hurt performance a very little bit in "extreme rare cases".

If there is a concrete proposal as an alternative the current one, we might agree on that or directly show the race condition on it.

laeubi: > precisely will happen is that another thread creates the handle, the thread trying to access the image data will create image data based on a temporary handle

This it what does not makes sense to me: `ImageData` has no handle (only image has) and the created `ImageData` (in case we have no cache hit) is never stored anywhere!

So can we probably have a a _concrete_ call sequence that describes the issue to be happens as I tried with https://github.com/eclipse-platform/eclipse.platform.swt/pull/2905#discussion_r2672916518 like

- We have Thread A + B
- Thread A calls ...
- Then a context switch happens and Thread B calls ...
- What leads to object X is put into map ...
- ...

Its really hard otherwise to catch up. From what I can see we have a Map that caches `ImageData` and handles (the map) for a zoom level and one method that wants to reuse the `ImageData` cached (if any). If nothing is cached we will get at worst an `ImageData` created twice (one because Thread A has created it while Thread B got a temporary one). And I don't see that with the current code structure this can be avoided.

Furthermore we have an (unsyncronized) map that performs non atomic get-and-set operation what is problematic. This would be fixed by using a `ConcurrentHashMap` and `computeIfAbsent` what makes the get-and-set an atomic operation and also ensures that while one thread is currently computing an `ImageData` another thread arriving after that will wait until this computation is complete and use the cached data.

HeikoKlare: > This it what does not makes sense to me: `ImageData` has no handle (only image has) and the created `ImageData` (in case we have no cache hit) is never stored anywhere!

The image data created by `getImageData()` _is_ cached in case there is no handle for that zoom in this map:
https://github.com/eclipse-platform/eclipse.platform.swt/blob/367a0a613ea507a525bd7648a47cb9c7b9f5d374/bundles/org.eclipse.swt/Eclipse%20SWT/win32/org/eclipse/swt/graphics/Image.java#L2142

I tried to sketch a sequence already, let's try more concretely:
- We have an image based on an `ImageDataProvider` and threads A + B
- A calls `getImageData(zoom)` and finds the manager not to contain a handle for that zoom
- Context switch happens to B which calls `getHandle(zoom, zoom)` and creates a handle for that zoom in the manager (stored in the zoom to handle map of the manager)
- Context switch happens to A which proceeds with `imageProvider.newImageData(zoom)`; this creates a temporary handle for that zoom, extracts the image data out of it and stores it in the provider's `cachedImageData`

In consequence, `cachedImageData` contains image data for that zoom which will never be accessed or removed as the current contract (for single-threaded access) ensures that this cache only contains image data for zooms for which the manager does not contain a handle yet and, in case such a handle gets created, the image data is removed from that cache. So the sketched sequence leads to a kind of memory leak.

> Furthermore we have an (unsyncronized) map that performs non atomic get-and-set operation what is problematic. This would be fixed by using a ConcurrentHashMap and computeIfAbsent what makes the get-and-set an atomic operation and also ensures that while one thread is currently computing an ImageData another thread arriving after that will wait until this computation is complete and use the cached data.

We could (and maybe even should) synchronize the map, yes. For most use cases it's not necessary as in case a thread sees an inconsistent state (especially missing a handle that has already been created), it will synchronize afterwards, which produces a memory barrier and results in properly updated data (that's why the double-checked locking works). What might go wrong is that you retrieve and access a handle that has already been disposed, as for that case there is no memory barrier involved.

As an additional note: I would be fully in favor of simply using a synchronized collection (which we probably need to use additionally anyway), I currently just don't see how to synchronize the image data creation just based on a synchronized zoom-to-handle collection. That would require some `compute` method on the collection that doesn ot yield a value of the map (in our case a handle) but something else (in our case an image data object). What might be possible is (mis) use a `compute` method on the map-to-handle collection and inside that method calculate the image data which is returned via writing a separate `AtomicReference` (that is local to the calling method). That more feels like a hack but could actually work and relieve us from all the `synchronized` stuff. @amartya4256 maybe you want to give this a try?

laeubi: > the current contract (for single-threaded access) ensures that this cache only contains image data for zooms for which the manager does not contain a handle yet and, in case such a handle gets created, the image data is removed from that cache.

Thanks, I think I now better understand what one want to  handle here. I'll try to look a bit deeper into this the next days but what currently makes it quite hard is that this class is bloated with a lot of  nested classes/interfaces that have implicit references to the image object.

Is there any compelling reason not to extract all these into (package private) classes and passing any required pointer to an Image instance explicitly in the constructor? I think this will make not only the code easier to read/understand but also make the responsibilities more visible.

My current impression is, that the distributed caches are actually a problem and should better be handled in a way where we have one Map (indexed by the zoom level) that returns a `ImageDataState`. This manger would then have (possibly synchronized) methods that acquire image datas and manage the handle and the can easily make sure things are cleared out (either handle or image data) as it only has to cope with a single field of each item instead of these distributed values all over the place. This would then also be more memory efficient (currently we create multiple maps wich each has a capacity of  16 items while likely one won't have that much zoom levels ever stored.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -150,18 +151,15 @@ InternalImageHandle get(int zoom) {
...
+				if (zoom == -1) {
+					return null;
+				}
 
-			imageHandle = creator.get();
-			zoomLevelToImageHandle.put(zoom, imageHandle);
-			return imageHandle;
+				DestroyableImageHandle imageHandle = creator.get();
+				return imageHandle;
+			});
```
HeikoKlare: Can't this simply be:
```java
if (zoom == -1) {
	return null;
}
return return zoomLevelToImageHandle.computeIfAbsent(zoom, __ -> creator.get());

File: bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java [hunk 14]
```diff
@@ -1304,8 +1312,8 @@ public Rectangle getBounds() {
 
 Rectangle getBounds(int zoom) {
 	if (isDisposed()) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
-	if (imageHandleManager.contains(zoom)) {
-		InternalImageHandle imageMetadata = imageHandleManager.get(zoom);
+	InternalImageHandle imageMetadata = imageHandleManager.get(zoom);
```
HeikoKlare: Wouldn't it make sense to use `executeOnHandle` as well?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -181,16 +179,26 @@ Set<Integer> getAllZooms() {
...
+		            continue;
+		        }
+		        zoomLevelToImageHandle.compute(zoom, (key, handle) -> {
+		            if (handle != null) {
+		                handle.destroy();
+		            }
+		            return null;
+		        });
+		    }
+		}
```
HeikoKlare: Shouldn't this be somelike the following for simplicity and thread-safety?
```suggestion
			zoomLevelToImageHandle.entrySet().removeIf(entry -> {
				if (filter.test(entry.getKey())) {
					entry.getValue().destroy();
					return true;
				}
				return false;
			});
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -181,16 +174,25 @@ Set<Integer> getAllZooms() {
...
+		    zoomLevelToImageHandle.entrySet().removeIf(entry -> {
+		    	if (filter.test(entry.getKey())) {
+		    		entry.getValue().destroy();
+		    		return true;
+		    	}
+		    	return false;
+		    });
+		}
+
+		<R> R executeOnHandle(int zoom, Supplier<DestroyableImageHandle> handleCreator, Function<Optional<InternalImageHandle>,R> executable) {
```
HeikoKlare: Seeing the implementation (especially in `getImageData()`), I wonder if it would not be easier to understand if there was only was lambda that allows to create the ImageHandle as well. Precisely, this method could be something like:
```java
<R> R executeOnHandle(int zoom, Function<AtomicReference<DestroyableImageHandle>,R> executable) {
	AtomicReference<R> computedValue = new AtomicReference<>();
	zoomLevelToImageHandle.compute(zoom, (key, value) -> {
		AtomicReference<DestroyableImageHandle> handleReference = new AtomicReference<>(value);
		computedValue.set(executable.apply(handleReference));
		return handleReference.get();
	});
	return computedValue.get();
}
```
And then the callers could set the handle reference if they need to create it.

PR #2906, opened 2025-12-22
Make DPIChangeExecution execute with a higher priority

amartya4256:
This PR introduces the usage of a separate thread to start processing the DPI Change events by executing the handlers using Display#syncExec to provide higher priority to the DPIChangeExecuting tasks.

It doesn't fix the issue in https://github.com/vi-eclipse/Eclipse-Platform/issues/530 but executes the DPI Change tasks with a higher priority making sure the execution happens quicker.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
@@ -16,6 +16,7 @@ package org.eclipse.swt.widgets;
 
 
 import java.util.*;
+import java.util.concurrent.*;
 import java.util.concurrent.atomic.*;
 import java.util.stream.*;
 
@@ -5979,9 +5980,34 @@ LRESULT wmScrollChild (long wParam, long lParam) {
 	return null;
 }
 
+private class DPIChangeOperation {
+	Runnable operation;
+	Callback callback;
+
+	DPIChangeOperation(Runnable operation) {
+		this.operation = operation;
+	}
+
+	void asyncExec() {
+		callback = new Callback(this, "runAsyncOperation", 0);
+		OS.SendMessageCallback(Control.this.handle, OS.WM_NULL, 0, null, callback.getAddress(), 0);
+	}
+
+	void syncExec() {
+		operation.run();
+	}
+
+	long runAsyncOperation() {
+		operation.run();
+		callback.dispose();
+		return 0;
+	}
+}
+
 static class DPIChangeExecution {
 	AtomicInteger taskCount = new AtomicInteger();
 	private boolean asyncExec = true;
+	ExecutorService executor = Executors.newSingleThreadExecutor();
 
 	private void process(Control control, Runnable operation) {
 		boolean currentAsyncExec = asyncExec;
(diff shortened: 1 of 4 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
```diff
@@ -5992,7 +5994,9 @@ private void process(Control control, Runnable operation) {
 			asyncExec &= (comp.layout != null);
 		}
 		if (asyncExec) {
-			control.getDisplay().asyncExec(operation::run);
+			executor.submit(() -> {
```
HeikoKlare: Just an idea: couldn't we even simplify this by just having a single async task that sequentially runs through the controls and performs a `syncExec` with the operation on them? Maybe could even get rid of the counter then as there would be a single reponsingle (the executor) for tracking the process.
Maybe that's not possible or too complicated to adapt the implementation, but could be an interesting potential for simplification.

amartya4256: There is currently one executor per DPI_CHANGE event yes. I could try getting rid of the counter and see how it affects the overall event.

amartya4256: After having a look, I think it is not possible to track the end of execution without using a counter because all the controls are notified from within handleDPIChange event and that happens during the execution of each control.

HeikoKlare:
@amartya4256 do you have some snippet or the like to share for testing this? Something like an action to trigger async executions that will block the DPI change processing in an Eclipse application or even a plain SWT snippet that spawns according async executions?

HeikoKlare:
Nevermind, I just used this simple snippet for test:
```java
	shell.setSize(500, 100);
	Label label = new Label(shell, SWT.NONE);
	label.setText("Hello");
	for (int i = 0; i < 2000; i++) {
		display.asyncExec(() ->  {
			try {
				Thread.sleep(100);
			} catch (InterruptedException e) {
			}
		});
	}
	shell.open ();
	}
	display.dispose ();
}
```
This mimics a bunch of async tasks being spawned, preventing DPI change handling from being processed. That's also what actually happens as you can see the label not being rescaled when moving the shell to another monitor.

However, this change does not fix it. Turns out that `syncExec` calls are scheduled after `asyncExec` calls, just that they will block the caller. You can see that in the `Synchronizer` implementation:
https://github.com/eclipse-platform/eclipse.platform.swt/blob/d4f11be9183890aab9d628de8e220f98ca8b3a82/bundles/org.eclipse.swt/Eclipse%20SWT/common/org/eclipse/swt/widgets/Synchronizer.java#L176

So I think to achieve that the DPI change processing is preferred over application-side tasks scheduled at the `Display`, you would not need to provoke the execution of OS callbacks, which are preferred in `readAndDispatch()`. Or you would need to implement some special handling for DPI change processing in `Display`/`Synchronizer` (such as a special queue for DPI change processing that is executed in `Display#readAndDispatch()` before the async messages), but I am not sure if it's a good idea to mess that up. Still it could be worth a try.

HeikoKlare:
Closing as superceded by:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2946

PR #2907, opened 2025-12-22
Add Rectangle.of(Point,Point) supporting OfFloat

ShahzaibIbrahim:
When the size point is Point.OfFloat, return Rectangle.OfFloat to retain sub-pixel precision; otherwise return the classic Rectangle.

This is introduced as needed to be used in https://github.com/eclipse-platform/eclipse.platform.ui/pull/3597

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Rectangle.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Rectangle.java
@@ -377,10 +377,39 @@ public Rectangle union (Rectangle rect) {
  * @since 3.131
  */
 public static Rectangle of(Point topLeft, int width, int height) {
-	if (topLeft instanceof Point.WithMonitor monitorAwareTopLeft) {
-		return new Rectangle.WithMonitor(monitorAwareTopLeft.getX(), monitorAwareTopLeft.getY(), width, height, monitorAwareTopLeft.getMonitor());
+	return Rectangle.of(topLeft, new Point(width, height));
+}
+
+/**
+ * Creates a new {@code Rectangle} using the specified top-left point and the
+ * dimensions point defining the width and height.
+ * <p>
+ * If the provided {@code Point} instances carry additional contextual
+ * information, an extended {@code Rectangle} type may be returned to preserve
+ * that context. Otherwise, a standard {@code Rectangle} is returned.
+ * </p>
+ *
+ * @param topLeft   the top-left corner of the rectangle
+ * @param dimension the point defining the width and height of the rectangle
+ * @return a new {@code Rectangle} instance appropriate for the given point and
+ *         dimensions
+ * @since 3.133
+ */
+public static Rectangle of(Point topLeft, Point dimension) {
+	final float x = (topLeft instanceof Point.OfFloat p) ? p.getX() : topLeft.x;
+	final float y = (topLeft instanceof Point.OfFloat p) ? p.getY() : topLeft.y;
+	final float w = (dimension instanceof Point.OfFloat p) ? p.getX() : dimension.x;
+	final float h = (dimension instanceof Point.OfFloat p) ? p.getY() : dimension.y;
+
+	if (topLeft instanceof Point.WithMonitor pm) {
+		return new Rectangle.WithMonitor(x, y, w, h, pm.getMonitor());
 	}
-	return new Rectangle(topLeft.x, topLeft.y, width, height);
+
+	if (topLeft instanceof Point.OfFloat || dimension instanceof Point.OfFloat) {
+		return new Rectangle.OfFloat(x, y, w, h);
+	}
+
+	return new Rectangle(topLeft.x, topLeft.y, dimension.x, dimension.y);
 }
 
 /**
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_Rectangle.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_Rectangle.java
@@ -298,4 +298,66 @@ public void test_unionLorg_eclipse_swt_graphics_Rectangle() {
 	assertSWTProblem("Incorrect exception thrown for rectangle == null", SWT.ERROR_NULL_ARGUMENT, e);
 }
 
+@Test
+void test_bothPointsAreOfFloat() {
+    Point.OfFloat topLeft = new Point.OfFloat(10.5f, 20.5f);
+    Point.OfFloat dimension = new Point.OfFloat(30.5f, 40.5f);
+
+    Rectangle rect = Rectangle.of(topLeft, dimension);
+
+    assertTrue(rect instanceof Rectangle.OfFloat);
+    Rectangle.OfFloat f = (Rectangle.OfFloat) rect;
+
+    assertEquals(10.5f, f.getX());
+    assertEquals(20.5f, f.getY());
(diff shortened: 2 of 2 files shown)


bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Rectangle.java
```diff
@@ -383,6 +383,28 @@ public static Rectangle of(Point topLeft, int width, int height) {
...
+ * </p>
+ *
+ * @param topLeft the top-left corner of the rectangle
+ * @param dimension the Point(width, height) of the rectangle
+ * @return a new {@code Rectangle} instance appropriate for the given point and
+ *         dimensions
+ * @since 3.133
+ */
+public static Rectangle of(Point topLeft, Point dimension) {
+	if(dimension instanceof Point.OfFloat ofFloatDimension) {
```
akoch-yatta: I think we need to extend this PR a bit:
- spacing between if
- we need to handle the case when top is a Point.OfFloat as well for consistency
- And we should add some tests for the four different possible cases, so (Point,  Point), (Point.OfFloat, Point), (Point, Point.OfFloat) and (Point.OfFloat, Point.OfFloat)

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Rectangle.java
```diff
@@ -383,6 +383,33 @@ public static Rectangle of(Point topLeft, int width, int height) {
...
+	if (topLeft instanceof Point.OfFloat ofFloatTopLeft && dimension instanceof Point.OfFloat ofFloatDimension) {
+		return new Rectangle.OfFloat(ofFloatTopLeft.getX(), ofFloatTopLeft.getY(), ofFloatDimension.getX(),
+				ofFloatDimension.getY());
+	} else if (topLeft instanceof Point.OfFloat ofFloatTopLeft) {
+		return new Rectangle.OfFloat(ofFloatTopLeft.getX(), ofFloatTopLeft.getY(), dimension.x, dimension.y);
+	} else if (dimension instanceof Point.OfFloat ofFloatDimension) {
+		return new Rectangle.OfFloat(topLeft.x, topLeft.y, ofFloatDimension.getX(), ofFloatDimension.getY());
+	}
+	return new Rectangle(topLeft.x, topLeft.y, dimension.x, dimension.y);
+}
```
amartya4256, line 413: So many if-elses look very fuzzy. Maybe a suggestion for something like this?

```

public static Rectangle of(Point topLeft, Point dimension) {
    final boolean isFloat=
            (topLeft instanceof Point.OfFloat) || (dimension instanceof Point.OfFloat);

    final float x = (topLeft instanceof Point.OfFloat p) ? p.getX() : topLeft.x;
    final float y = (topLeft instanceof Point.OfFloat p) ? p.getY() : topLeft.y;
    final float w = (dimension instanceof Point.OfFloat p) ? p.getX() : dimension.x;
    final float h = (dimension instanceof Point.OfFloat p) ? p.getY() : dimension.y;

    return isFloat
            ? new Rectangle.OfFloat(x, y, w, h)
            : new Rectangle((int) x, (int) y, (int) w, (int) h);
}


```

ShahzaibIbrahim: Sound reasonable. Adapted!

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Rectangle.java
```diff
@@ -383,6 +383,32 @@ public static Rectangle of(Point topLeft, int width, int height) {
...
+ * that context. Otherwise, a standard {@code Rectangle} is returned.
+ * </p>
+ *
+ * @param topLeft   the top-left corner of the rectangle
+ * @param dimension the Point(width, height) of the rectangle
+ * @return a new {@code Rectangle} instance appropriate for the given point and
+ *         dimensions
+ * @since 3.133
+ */
+public static Rectangle of(Point topLeft, Point dimension) {
```
HeikoKlare, line 398: There is already `of(Point, int, int)`. Since that and this new method are basically equal (just that the new one can carry additional context information in the dimension point), I would expect them to behave the same whenever possible. This particularly means:
- The existing method should also properly handle the passed topleft point being OfFloat
- This new method should also deal with topLeft being WithMonitor

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Rectangle.java
```diff
@@ -383,6 +383,32 @@ public static Rectangle of(Point topLeft, int width, int height) {
...
 }
 
+/**
+ * Creates a new {@code Rectangle} using the specified Point instances for the
+ * dimensions.
+ * <p>
+ * If the provided {@code Point} instance carries additional OfFloat
+ * information, an extended {@code Rectangle} type may be returned to preserve
+ * that context. Otherwise, a standard {@code Rectangle} is returned.
+ * </p>
```
HeikoKlare: "OfFloat" is nothing known at public API. The contract of the method should be accordingly abstract. I would propose to adopt the formulation of the existing `of(Point, int, int)` method.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Rectangle.java
ShahzaibIbrahim: Changes are as follow for both the methods now: Extract coordinates once, return WithMonitor if present, otherwise choose OfFloat when needed or fall back to a normal Rectangle.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Rectangle.java
```diff
@@ -377,12 +377,49 @@ public Rectangle union (Rectangle rect) {
...
+
+	if (topLeft instanceof Point.WithMonitor pm) {
+		return new Rectangle.WithMonitor(x, y, w, h, pm.getMonitor());
+	}
+
+	if (topLeft instanceof Point.OfFloat || dimension instanceof Point.OfFloat) {
+		return new Rectangle.OfFloat(x, y, w, h);
+	}
+
+	return new Rectangle((int) x, (int) y, (int) w, (int) h);
```
HeikoKlare: I would propose to just use the original values to avoid casts here.
```suggestion
	return new Rectangle(topLeft.x, topLeft.y, dimension.x, dimension.y);
```

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/Rectangle.java
```diff
@@ -377,12 +377,49 @@ public Rectangle union (Rectangle rect) {
...
-	if (topLeft instanceof Point.WithMonitor monitorAwareTopLeft) {
-		return new Rectangle.WithMonitor(monitorAwareTopLeft.getX(), monitorAwareTopLeft.getY(), width, height, monitorAwareTopLeft.getMonitor());
+	if (topLeft instanceof Point.WithMonitor pm) {
+		return new Rectangle.WithMonitor(pm.getX(), pm.getY(), width, height, pm.getMonitor());
 	}
+
+	if (topLeft instanceof Point.OfFloat p) {
+		return new Rectangle.OfFloat(p.getX(), p.getY(), width, height);
+	}
+
```
HeikoKlare: Can't you just replace this with `Rectangle.of(topLeft, new Point(width, height))` instead of duplicating the logic?

PR #2910, opened 2025-12-24
Update Snippet382 to demonstrate PlainImageDataProvider disablement

ShahzaibIbrahim:
Added two additional examples showing how to create images using PlainImageDataProvider—both with a basic provider and with a provider that already has an allocated handle. These examples illustrate how greyed and disabled images are generated in each scenario.


akoch-yatta (approved):
LGTM, extending the Snippet to show this scenario makes sense here as on windows this will create the OS handles in a different way

PR #2912, opened 2025-12-26
Snippet for demonstrating Cropping and Scaling of GC#drawImage

al-noori:
Previously, GC#drawImage was only used in Snippet 10 to demonstrate scaling. Since the method also supports cropping parts of a source image, and scaling itself can be achieved by passing only the destination coordinates/width/height, this PR adds a more expressive snippet.


examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet388.java
```diff
@@ -1,240 +1,176 @@
 /*******************************************************************************
- * Copyright (c) 2025 Eclipse Platform Contributors and others.
- *
- * This program and the accompanying materials
- * are made available under the terms of the Eclipse Public License 2.0
- * which accompanies this distribution, and is available at
- * https://www.eclipse.org/legal/epl-2.0/
+ * Copyright (c) 2000, 2016 IBM Corporation and others.
```
merks: I assume this was (mostly) copied, but even then the date does not reflect date of current content.

al-noori: Good catch, the header was copied from an existing snippet.
I’ve updated the copyright year to reflect the current content.

al-noori:
A preview screenshot and an entry in the snippets.md file are now included.

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet389.java
```diff
@@ -0,0 +1,177 @@
+/*******************************************************************************
+ * Copyright (c) 2025 IBM Corporation and others.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
```
HeikoKlare, line 10:
```suggestion
/*******************************************************************************
 * Copyright (c) 2026 Yatta Solutions and others.
 *
 * This program and the accompanying materials
 * are made available under the terms of the Eclipse Public License 2.0
 * which accompanies this distribution, and is available at
 * https://www.eclipse.org/legal/epl-2.0/
 *
 * SPDX-License-Identifier: EPL-2.0
 *******************************************************************************/
```

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet389.java
```diff
@@ -0,0 +1,177 @@
...
+
+package org.eclipse.swt.snippets;
+
+import org.eclipse.swt.*;
+import org.eclipse.swt.graphics.*;
+import org.eclipse.swt.layout.*;
+import org.eclipse.swt.widgets.*;
+
+public class Snippet389 {
```
HeikoKlare, line 33: Please provide a short documentation like for other snippets. Here is one of the most recent (but the description does not need to be that long): https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet387.java

Most important, the first sentence of the description should be good summary of the snippet to be automatically shown in the SnippetExplorer:
image

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet389.java
al-noori: Thank you for the correct header, i included it now.

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet389.java
al-noori: Thanks for clarification, i made a small description in front of the snippet. I have also changed the name of the folder in the resources to Snippet389.

HeikoKlare (approved):
The snippet looks fine. I would like to merge it soon as it currently kind of blocks https://github.com/eclipse-platform/eclipse.platform.swt/pull/2926, which introduces subsequent Snippet390. @akoch-yatta do you have any concerns? In case we find some need to improve the snippet afterwards, we can still adapt.

With the latest changes, I just fixed the following:
- Formatting of the Java file
- Wrong path to images (still contained "Snippet388", so snippet did not start)
- Replaced the preview image with one taken at 100% monitor zoom as the previous one was too large for the SnippetExplorer
- Removed faulty issue reference from commit message

akoch-yatta:
I'm fine with merging it, my original remarks were solved

HeikoKlare:
Adding the snippet to the SWT website via:
- https://github.com/eclipse-platform/www.eclipse.org-eclipse/pull/469

PR #2913, opened 2025-12-26
Fix GC#drawImage with ImageGcDrawer for Cropping and Scaling

al-noori:
The GC#drawImage method takes
(image, srcX, srcY, srcWidth, srcHeight, destX, destY, destWidth, destHeight)
as arguments and crops and scales from the source region to the destination
region.

Passing an image drawn via ImageGCDrawer led to the following issue:

The image handle from the subcall is resolved using the monitor zoom
(data.nativeZoom) and the calculated scaledImageZoom (gcZoom * scaleFactor).
This handle corresponds to an ImageData initialized at scaledImageZoom,
whereas the drawings of the second GC are performed using the monitor zoom,
subject to the auto-scale property.

This mismatch results in unaligned sizing of drawings. For example, a 200%
monitor zoom combined with a scale factor of 0.5 produces a scaledImageZoom of
100%. As a result, the ImageData is initialized at 100%, while drawing occurs
at 200%. 

Furthermore, the calculation of scaledImageZoom uses fallback logic that only
allows 100% and 200% as possible outcomes, which is clearly unintended in this
context.

The fix delegates resolving the correct handle to the Image class by passing
the width/height of the full image scaled by the scaledImageZoom. This
is a space on where scaled `src` coordinates/width/height lie. A callback
then creates a new handle for the height/width and respects the auto-scale property.

If the returned handle matches the full image scaled to the requested
scaledImageZoom in width and height, the source region
coordinates/width/height are passed directly in pixels at that zoom. Otherwise,
the internal zoom factor is derived from the returned handle’s width relative to
the full image, and the source region coordinates are converted to pixel values
using this internal zoom.

Note: Snippet 10 used to demonstrate that method. Although the sizing issue is fixed, the method is replaced by GC#drawImage(destX, destY, destWidth, destHeight) as the use case here is only scaling and not cropping.  Currently, a new snippet is in progress which will be developed to address the cropping + scaling scenario (see #2912 as a first version without drawings via the ImageGcDrawer).

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
@@ -15,16 +15,13 @@ package org.eclipse.swt.graphics;
 
 
 import java.util.*;
-import java.util.List;
 import java.util.function.*;
-import java.util.stream.*;
 
 import org.eclipse.swt.*;
 import org.eclipse.swt.graphics.Image.*;
 import org.eclipse.swt.internal.*;
 import org.eclipse.swt.internal.gdip.*;
 import org.eclipse.swt.internal.win32.*;
-import org.eclipse.swt.widgets.*;
 
 /**
  * Class <code>GC</code> is where all of the drawing capabilities that are
@@ -1178,18 +1175,52 @@ private class DrawScalingImageToImageOperation extends ImageOperation {
 
 	@Override
 	void apply() {
-		int gcZoom = getZoom();
-		int srcImageZoom = calculateZoomForImage(gcZoom, source.width, source.height, destination.width, destination.height);
-		drawImage(getImage(), source.x, source.y, source.width, source.height, destination.x, destination.y, destination.width, destination.height, gcZoom, srcImageZoom);
+		draw(getImage(), source.x, source.y, source.width, source.height, destination.x, destination.y, destination.width, destination.height);
 	}
 
-	private Collection<Integer> getAllCurrentMonitorZooms() {
-		if (device instanceof Display display) {
-			return Arrays.stream(display.getMonitors())
-				.map(Monitor::getZoom)
-				.collect(Collectors.toSet());
+	private void draw(Image image, int srcX, int srcY, int srcWidth, int srcHeight, int destX, int destY, int destWidth, int destHeight) {
+		int gcZoom = getZoom();
+		int requestedImageZoom = calculateZoomForImage(gcZoom, source.width, source.height, destination.width, destination.height);
+
+		Rectangle src =  new Rectangle(srcX, srcY, srcWidth, srcHeight);
+		Rectangle destPixels = Win32DPIUtils.pointToPixel(drawable, new Rectangle(destX, destY, destWidth, destHeight), gcZoom);
+		Rectangle fullImageBounds = image.getBounds();
+		Rectangle requestedFullImageBoundsPixels = Win32DPIUtils.pointToPixel(drawable, fullImageBounds, requestedImageZoom);
+
+		image.executeOnImageHandleAtBestFittingSize((tempHandle) -> {
+			Rectangle srcPixels = computeSourceRectangle(tempHandle, fullImageBounds, src);
+			drawImage(image, srcPixels.x, srcPixels.y, srcPixels.width, srcPixels.height, destPixels.x, destPixels.y, destPixels.width,
+					destPixels.height, false, tempHandle);
+		}, requestedFullImageBoundsPixels.width, requestedFullImageBoundsPixels.height);
+	}
+
+	private Rectangle computeSourceRectangle(ImageHandle imageHandle, Rectangle fullImageBounds, Rectangle src) {
+		/*
+		 * The point values (x, y, width, height) of the source "part" in points will be
+		 * computed to pixels depending on the factor of the full image bounds to the
+		 * actual OS handle size that will be used.
+		 */
+		float scaleFactor = Math.min(1f * imageHandle.width() / fullImageBounds.width, 1f * imageHandle.height() / fullImageBounds.height);
+		int closestZoomOfHandle = Math.round(scaleFactor * 100);
+		Rectangle srcPixels = Win32DPIUtils.pointToPixel(drawable, src, closestZoomOfHandle);
+
+		if (closestZoomOfHandle != 100) {
+			/*
+			 * This is a HACK! Due to rounding errors at fractional scale factors,
+			 * the coordinates may be slightly off. The workaround is to restrict
+			 * coordinates to the allowed bounds.
+			 */
+			int errX = srcPixels.x + srcPixels.width - imageHandle.width();
+			int errY = srcPixels.y + srcPixels.height - imageHandle.height();
+			if (errX != 0 || errY != 0) {
+				if (errX <= closestZoomOfHandle / 100 && errY <= closestZoomOfHandle / 100) {
+					srcPixels.intersect(new Rectangle(0, 0, imageHandle.width(), imageHandle.height()));
+				} else {
+					SWT.error (SWT.ERROR_INVALID_ARGUMENT);
+				}
+			}
 		}
-		return Collections.emptySet();
+		return srcPixels;
 	}
 
 	private int calculateZoomForImage(int gcZoom, int srcWidth, int srcHeight, int destWidth, int destHeight) {
@@ -1205,15 +1236,9 @@ private class DrawScalingImageToImageOperation extends ImageOperation {
 			return gcZoom;
 		}
 
-		float imageScaleFactor = 1f * destWidth / srcWidth;
+		float imageScaleFactor = Math.max(1f * destWidth / srcWidth, 1f * destHeight / srcHeight);
 		int imageZoom = Math.round(gcZoom * imageScaleFactor);
-		if (getAllCurrentMonitorZooms().contains(imageZoom)) {
-			return imageZoom;
-		}
-		if (imageZoom > 150) {
-			return 200;
-		}
-		return 100;
+		return imageZoom;
 	}
 }
 
@@ -1239,30 +1264,6 @@ private void drawImage(Image image, int destX, int destY, int destWidth, int des
 		drawImage(image, 0, 0, tempHandle.width(), tempHandle.height(), destPixels.x, destPixels.y,
 				destPixels.width, destPixels.height, false, tempHandle);
 	}, destPixels.width, destPixels.height);
-}
-
-private void drawImage(Image image, int srcX, int srcY, int srcWidth, int srcHeight, int destX, int destY,
-		int destWidth, int destHeight, int imageZoom, int scaledImageZoom) {
-	Rectangle src = Win32DPIUtils.pointToPixel(drawable, new Rectangle(srcX, srcY, srcWidth, srcHeight), scaledImageZoom);
-	Rectangle dest = Win32DPIUtils.pointToPixel(drawable, new Rectangle(destX, destY, destWidth, destHeight), imageZoom);
-	if (scaledImageZoom != 100) {
-		/*
-		 * This is a HACK! Due to rounding errors at fractional scale factors,
-		 * the coordinates may be slightly off. The workaround is to restrict
-		 * coordinates to the allowed bounds.
-		 */
-		Rectangle b = image.getBounds(scaledImageZoom);
-		int errX = src.x + src.width - b.width;
-		int errY = src.y + src.height - b.height;
-		if (errX != 0 || errY != 0) {
-			if (errX <= scaledImageZoom / 100 && errY <= scaledImageZoom / 100) {
-				src.intersect(b);
-			} else {
-				SWT.error (SWT.ERROR_INVALID_ARGUMENT);
-			}
-		}
-	}
-	drawImage(image, src.x, src.y, src.width, src.height, dest.x, dest.y, dest.width, dest.height, false, image.getHandle(scaledImageZoom, data.nativeZoom));
 }
 
 private class DrawImageToImageOperation extends ImageOperation {


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -880,6 +891,15 @@ void executeOnImageHandleAtBestFittingSize(Consumer<ImageHandle> handleAtSizeCon
 	handleAtSizeConsumer.accept(imageHandle);
 }
 
+void executeOnImageHandleAtBestFittingSizeAtZoom(Consumer<ImageHandle> handleAtSizeConsumer, int scaledZoom) {
+	ImageHandle imageHandle = lastRequestedHandle.getOrCreateImageHandleAtClosestSizeAtZoom(scaledZoom);
+	handleAtSizeConsumer.accept(imageHandle);
+}
+void executeOnImageHandleAtSizeOrZoom(BiConsumer<ImageHandle, Point> handleAtSizeConsumer, int targetWidth, int targetHeight, int zoom) {
```
arunjose696: is this used anywhere?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -179,6 +179,17 @@ private Optional<ImageHandle> createHandleAtExactSize(int width, int height) {
 			return Optional.empty();
 		}
 
+		public ImageHandle getOrCreateImageHandleAtClosestSizeAtZoom(int scaledZoom) {
```
arunjose696: shouldnt this be private

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -1243,6 +1228,7 @@ private void drawImage(Image image, int destX, int destY, int destWidth, int des
 
 private void drawImage(Image image, int srcX, int srcY, int srcWidth, int srcHeight, int destX, int destY,
 		int destWidth, int destHeight, int imageZoom, int scaledImageZoom) {
+
```
arunjose696:
```suggestion
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -1262,7 +1248,31 @@ private void drawImage(Image image, int srcX, int srcY, int srcWidth, int srcHei
 			}
 		}
 	}
-	drawImage(image, src.x, src.y, src.width, src.height, dest.x, dest.y, dest.width, dest.height, false, image.getHandle(scaledImageZoom, data.nativeZoom));
+	Rectangle fullImageBounds = image.getBounds();
+	Rectangle targetSrc = Win32DPIUtils.pointToPixel(drawable, fullImageBounds, scaledImageZoom);
+	Rectangle startSrc =  new Rectangle(srcX, srcY, srcWidth, srcHeight);
+	image.executeOnImageHandleAtBestFittingSizeAtZoom((tempHandle) -> {
```
arunjose696: IMO it is better to have an additional override for `Image#executeOnImageHandleAtBestFittingSize()` which takes source dimensions in addition to destination dimensions so the GC#calculateZoomForImage() can be removed and the calculation of zoom becomes the responsibility of the Image class.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -1262,7 +1248,31 @@ private void drawImage(Image image, int srcX, int srcY, int srcWidth, int srcHei
 			}
 		}
 	}
-	drawImage(image, src.x, src.y, src.width, src.height, dest.x, dest.y, dest.width, dest.height, false, image.getHandle(scaledImageZoom, data.nativeZoom));
+	Rectangle fullImageBounds = image.getBounds();
```
arunjose696: Can see `image.getBounds `is called above in case the `zoom!=100` can you see if the hack still makes sense, It would make sense to consider if we can stream line these calls

al-noori:
@arunjose696 i edited the PR such that the `drawImage(srcX,srcY,...)` method uses the same callback `image.executeOnImageHandleAtBestFittingSize` via width and height of the scaled full image as the other drawImage method. 
That is why i would hesitate to write an additional overwrite. I thereby intend to capture a quick adoption mentioned in: https://github.com/vi-eclipse/Eclipse-Platform/issues/468#issue-3476415417. This also made any changes to the image class redundant and i could leave out my additional methods. What is your opinion on this?

Regarding that hack: it seems to me that i does not make sense. The hack is motivated by potential rounding errors, but without further investigation, i assume that passing correct source coordinates/width/height which lie in the space of the full image, the statement `src.intersect(fullImageBoundsScaled)` will always return `src` with equal attributes as before regardless of errors.  
I streamlined the getBounds call as you mentioned and would leave the hack as it is for this PR. Investigating the hack and why it was introduced in the first place might be useful in the context of a different PR.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -5810,7 +5814,7 @@ void apply() {
  * Returns the extent of the given string. No tab
  * expansion or carriage return processing will be performed.
  * <p>
- * The <em>extent</em> of a string is the width and height of
+ * The <em>extent</em> ofbut  a string is the width and height of
```
arunjose696:
```suggestion
 * The <em>extent</em> of a string is the width and height of
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -1245,24 +1227,46 @@ private void drawImage(Image image, int srcX, int srcY, int srcWidth, int srcHei
...
 	if (scaledImageZoom != 100) {
 		/*
 		 * This is a HACK! Due to rounding errors at fractional scale factors,
 		 * the coordinates may be slightly off. The workaround is to restrict
 		 * coordinates to the allowed bounds.
 		 */
-		Rectangle b = image.getBounds(scaledImageZoom);
-		int errX = src.x + src.width - b.width;
-		int errY = src.y + src.height - b.height;
+		int errX = src.x + src.width - fullImageBoundsScaled.width;
```
arunjose696: Why do we use the `fullImageBoundsScaled` here ?, this looks like a pixel value previously the bounds in points was used in the hack

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -1245,24 +1227,46 @@ private void drawImage(Image image, int srcX, int srcY, int srcWidth, int srcHei
...
+	if (new Rectangle(0, 0, imageHandle.getWidth(), imageHandle.getHeight()).equals(fullImageBoundsScaled)) {
+		return src;
+	} else {
+		/*
+		 *  the achieved handle with its drawings has not the required size, thus we calculate the zoom of the handle
+		 *
+		 * with respect to the full 100% image. The point values (x,y,width,height) of the source "part" of the full image will
+		 * be computed to pixels by this zoom.
+		 */
+		float scaleFactor = 1f * imageHandle.getWidth() / fullImageBounds.width;
```
arunjose696: WHy do we only consider width for the scale factor, as there could theoretically be the case where width and height are differently scaled, in such cases shouldnt we scale the y cordinates and height with the scaling factor for height?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -1245,24 +1227,46 @@ private void drawImage(Image image, int srcX, int srcY, int srcWidth, int srcHei
 		int destWidth, int destHeight, int imageZoom, int scaledImageZoom) {
 	Rectangle src = Win32DPIUtils.pointToPixel(drawable, new Rectangle(srcX, srcY, srcWidth, srcHeight), scaledImageZoom);
 	Rectangle dest = Win32DPIUtils.pointToPixel(drawable, new Rectangle(destX, destY, destWidth, destHeight), imageZoom);
+	Rectangle fullImageBounds = image.getBounds();
+	Rectangle fullImageBoundsScaled = Win32DPIUtils.pointToPixel(drawable, fullImageBounds, scaledImageZoom);
```
arunjose696: a minor nit from my side is to use fullImageBoundsPx or fullImageBoundsPixels for the pixel value and something similar for the point value for better readability.

arunjose696:
> @arunjose696 i edited the PR such that the `drawImage(srcX,srcY,...)` method uses the same callback `image.executeOnImageHandleAtBestFittingSize` via width and height of the scaled full image as the other drawImage method. That is why i would hesitate to write an additional overwrite. I thereby intend to capture a quick adoption mentioned in: [vi-eclipse/Eclipse-Platform#468 (comment)](https://github.com/vi-eclipse/Eclipse-Platform/issues/468#issue-3476415417). This also made any changes to the image class redundant and i could leave out my additional methods. What is your opinion on this?

Using the same overload makes sense and has simplified the changes IMO. 

> Regarding that hack: it seems to me that i does not make sense. The hack is motivated by potential rounding errors, but without further investigation, i assume that passing correct source coordinates/width/height which lie in the space of the full image, the statement `src.intersect(fullImageBoundsScaled)` will always return `src` with equal attributes as before regardless of errors. I streamlined the getBounds call as you mentioned and would leave the hack as it is for this PR. Investigating the hack and why it was introduced in the first place might be useful in the context of a different PR.

Thanks for streamlining it , i have a some concerns in the changes though, and few more comments

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
al-noori: Actually, the size of the rectangle `b` is interpreted as pixels: take a monitor zoom of 200% and a scaling factor of 0.5, then the conversion in px is point * 200 * 0.5 = point * scaledImageZoom.  As `b` and `fullImageBoundsScaled` are the same, i replaced it to streamline the getBounds call.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
al-noori: Because the ratio of the source rectangle should be still the same in pixels otherwise the passed values in points to drawImage would not be respected. But i see that, to minimize native resampling, it would be better to acquire the maximum of the scale factor in width and the scale factor in height. This would align well with how the `imageZoom` is computed in `ImageHandle#getOrCreateImageHandle `, being the maximum of the scaling factor in width and in height * 100 of the full image satisfying the autoscale property.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
al-noori: Thank you for that, i re-edited it

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
al-noori: This is a good point, i appended the suffix "Pixels" to all names of rectangles with size in pixels, and left it out if the size is in points.

ShahzaibIbrahim:
@arunjose696 have you reviewed this already?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -1243,26 +1225,48 @@ private void drawImage(Image image, int destX, int destY, int destWidth, int des
...
+private Rectangle computeSourceRectangle(ImageHandle imageHandle, Rectangle fullImageBounds, Rectangle fullImageBoundsPixels, Rectangle src, Rectangle srcPixels) {
+	if (new Rectangle(0, 0, imageHandle.width(), imageHandle.height()).equals(fullImageBoundsPixels)) {
+		return srcPixels;
+	} else {
+		/*
+		 *  the achieved handle with its drawings has not the required size, thus we calculate the zoom of the handle
+		 *
+		 * with respect to the full 100% image. The point values (x,y,width,height) of the source "part" of the full image will
+		 * be computed to pixels by this zoom.
+		 */
```
arunjose696: @al-noori The existing comment restates the implementation. Encoding intent in names and keeping a short  comment makes this easier to maintain.

```suggestion
// Handle may be rendered at a different DPI than the original image
```

al-noori: Thank you for the comment. I updated it now and it should be more readable.

arunjose696: just noticed this is somehow reverted.

akoch-yatta (approved):
LGTM, I did a small adaption by moving up and re-using the src Rectangle to create one less Rectangle. I tested mostly with Snippet389 and the IDE and everything worked as expected.
Although there are now more calculations involved, I did not see an easier way to achieve the same result to use the best quality image handle available in a cropping+scaling scenario

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -1205,15 +1193,9 @@ private int calculateZoomForImage(int gcZoom, int srcWidth, int srcHeight, int d
 			return gcZoom;
 		}
 
-		float imageScaleFactor = 1f * destWidth / srcWidth;
+		float imageScaleFactor = Math.max(1f * destWidth / srcWidth, 1f * destHeight / srcHeight);
```
HeikoKlare, line 1239: Shouldn't we just calculate the scale factor in both axes? Some images (like SVGs or based on ImageDataAtSizeProvider) are able to scale both axes differently.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -1243,26 +1225,49 @@ private void drawImage(Image image, int destX, int destY, int destWidth, int des
 
 private void drawImage(Image image, int srcX, int srcY, int srcWidth, int srcHeight, int destX, int destY,
 		int destWidth, int destHeight, int imageZoom, int scaledImageZoom) {
```
HeikoKlare: With the adaptations in this PR, I think it would clearer if the `scaledImageZoom` was calculated in this method based on the `imageZoom` instead of passing it this method and calculate it out of the `gcZoom` (which is hard to comprehend).

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -1243,26 +1225,49 @@ private void drawImage(Image image, int destX, int destY, int destWidth, int des
...
 		/*
 		 * This is a HACK! Due to rounding errors at fractional scale factors,
 		 * the coordinates may be slightly off. The workaround is to restrict
 		 * coordinates to the allowed bounds.
 		 */
-		Rectangle b = image.getBounds(scaledImageZoom);
-		int errX = src.x + src.width - b.width;
-		int errY = src.y + src.height - b.height;
+		int errX = srcPixels.x + srcPixels.width - fullImageBoundsPixels.width;
+		int errY = srcPixels.y + srcPixels.height - fullImageBoundsPixels.height;
```
HeikoKlare: Why do we change this calculation? Previously, `src` was used instead of `srcPixels` and compared against `b`, which seems to have been the bounds in pixels. That does not sound correct. Is that the reason for changing it? But since the `srcPixels` might be different later (depending on what handle we get), this new calculation is not correct either, is it?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -1243,26 +1225,49 @@ private void drawImage(Image image, int destX, int destY, int destWidth, int des
...
+		Rectangle newSrcPixels = computeSourceRectangle(tempHandle, fullImageBounds, fullImageBoundsPixels, src, srcPixels);
+		drawImage(image, newSrcPixels.x, newSrcPixels.y, newSrcPixels.width, newSrcPixels.height, destPixels.x, destPixels.y, destPixels.width,
+				destPixels.height, false, tempHandle);
+	}, fullImageBoundsPixels.width, fullImageBoundsPixels.height);
+}
+
+private Rectangle computeSourceRectangle(ImageHandle imageHandle, Rectangle fullImageBounds, Rectangle fullImageBoundsPixels, Rectangle src, Rectangle srcPixels) {
+	if (new Rectangle(0, 0, imageHandle.width(), imageHandle.height()).equals(fullImageBoundsPixels)) {
+		return srcPixels;
+	} else {
```
HeikoKlare: Can't we just drop this if block, as the else block will yield the same result (just with a slight calculation overhead that will not be relevant in the overall computation time)? That would simplify the method signature and the intellectual load of understanding the behavior. It might also relieved us from calculating the `srcPixels` in the calling method at all, as in my understanding the usage of that value is not correct anyway (see my other comment on that).

Since the actual image zoom and thus the src coordinates in pixels always depend on the handle that the image will give us for the requested width/height, it would, in my opinion, be conceptually the right think to only calculate the srcPixels based on the handle once we have it.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
akoch-yatta: `src` was in pixels before, the naming has changed here to differ between both of them.
But this correction should be placed somewhere later, when the actual image handle size is known.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
akoch-yatta: Would be possible, but complicates the logic a lot more, as multiple upscaling would need to be done separately for different factors. Could still be done, but I would move it to a separate PR then

akoch-yatta:
@HeikoKlare I updated the PR and addressed your comments as discussed. I tested it again mostly with the Snippet on different zooms and it still works as expected. As I originally approved this PR, I would ask you to approve it as well, if the changes are fine for you.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -1178,18 +1175,52 @@ private class DrawScalingImageToImageOperation extends ImageOperation {
...
 	@Override
 	void apply() {
-		int gcZoom = getZoom();
-		int srcImageZoom = calculateZoomForImage(gcZoom, source.width, source.height, destination.width, destination.height);
-		drawImage(getImage(), source.x, source.y, source.width, source.height, destination.x, destination.y, destination.width, destination.height, gcZoom, srcImageZoom);
+		draw(getImage(), source.x, source.y, source.width, source.height, destination.x, destination.y, destination.width, destination.height);
+	}
+
+	private void draw(Image image, int srcX, int srcY, int srcWidth, int srcHeight, int destX, int destY, int destWidth, int destHeight) {
+		int imageZoom = getZoom();
```
HeikoKlare: This is not the image zoom but the GC zoom, isn't it?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -1178,18 +1175,52 @@ private class DrawScalingImageToImageOperation extends ImageOperation {
...
+		int requestedImageZoom = calculateZoomForImage(imageZoom, source.width, source.height, destination.width, destination.height);
+
+		Rectangle src =  new Rectangle(srcX, srcY, srcWidth, srcHeight);
+		Rectangle destPixels = Win32DPIUtils.pointToPixel(drawable, new Rectangle(destX, destY, destWidth, destHeight), imageZoom);
+		Rectangle fullImageBounds = image.getBounds();
+		Rectangle requestedFullImageBoundsPixels = Win32DPIUtils.pointToPixel(drawable, fullImageBounds, requestedImageZoom);
+
+		image.executeOnImageHandleAtBestFittingSize((tempHandle) -> {
+			Rectangle newSrcPixels = computeSourceRectangle(tempHandle, fullImageBounds, src);
```
HeikoKlare: nit: why are these "new" src pixels? Those are the effective src pixels for the given handle, arent't they?

PR #2915, opened 2025-12-29
Drop reference to archived repo

akurtakov:
All needed dependencies are installable from the same repository so no need for such special and wrong comment.


akurtakov:
Docs change only. Merging without waiting.

PR #2922, opened 2025-12-30
Mark ImageDataAtSizeProvider as referenceable

arunjose696:


binaries/org.eclipse.swt.win32.win32.x86_64/META-INF/MANIFEST.MF
```diff
@@ -3,7 +3,7 @@ Fragment-Host: org.eclipse.swt;bundle-version="[3.128.0,4.0.0)"
 Bundle-Name: %fragmentName
 Bundle-Vendor: %providerName
 Bundle-SymbolicName: org.eclipse.swt.win32.win32.x86_64; singleton:=true
-Bundle-Version: 3.133.0.qualifier
+Bundle-Version: 4.0.0.qualifier
```
akurtakov: Fragments and host in SWT should match version. We are not moving to 4.0 for this one. Please add API tools commented API filter.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/ImageDataAtSizeProvider.java
```diff
@@ -38,9 +38,7 @@
  * version of the image data is not of interest.
  * </ol>
  *
- * @since 3.132
- * @noreference This class is still experimental API and might be subject to
- *              change.
+ * @since 4.0
```
akurtakov: Since flag should not change. Just the noreference to be dropped.

akurtakov (changes requested):

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/ImageDataAtSizeProvider.java
arunjose696: my IDE complains when this is done , Let me see if the CI works
image

binaries/org.eclipse.swt.win32.win32.x86_64/META-INF/MANIFEST.MF
arunjose696: Adding the commented API filter. from the eclipse quickfix does not seem to work, Can you give an example how to add the API filter, making a class referenceable

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/graphics/ImageDataAtSizeProvider.java
HeikoKlare: Do you have the right baseline set in your IDE? It looks a bit odd that the generated filter references `3.131.0`:
```
                <message_argument value="3.133.0"/>
                <message_argument value="3.131.0"/>
```
Maybe you have an outdated setup? You should do a "Perform Setup Tasks" and check you API baseline for if it contains the right SWT reference version (should be 3.132.0):
image

akoch-yatta:
@akurtakov Can you have a second look here? I took that PR over and added api_filters that look okay to me.

akurtakov (approved):
Ideally, it should be a commented api filter with link to this issue. But it's not a blocker for me.

akoch-yatta:
Thank you @akurtakov There is no issue for that, but I added a comment to each filter. I assume your approval as still valid and will merge it when the build is green

PR #2927, opened 2026-01-05
Using ImageHandleManager for managing multiple Image handles in Image class

amartya4256:
This PR introduces ImageHandleManager, an inner class to Image responsible for managing the storage, creation and manipulation of ImageHandles. This class provides methods which can be used to dead with ImageHandle instances inside an image for managing it with convenience while hiding the complexity of how the ImageHandle lifecycle is managed inside itself.

All the places where zoomToImageHandle map is accessed for any purposes is replaced by the wrapper methods provided by ImageHandleManager. Each image initializes a final field imageHandleManager with the instance of ImageHandleManager. Every instance of ImageHandle created within the Image object which is meant to be stored per zoom is registered in the manager itself and the manager is responsible for cleaning the dangling resources when are not supposed to be used anymore or when a cleanup is needed.

This contributes to https://github.com/vi-eclipse/Eclipse-Platform/issues/562

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
@@ -4462,6 +4462,15 @@ private long identity() {
 }
 
 private void init(Drawable drawable, GCData data, long hDC) {
+	ImageHandle imageHandle = null;
+	Image image = data.image;
+	if (image != null) {
+		imageHandle = image.getHandle(data.imageZoom, data.nativeZoom);
+	}
+	init(drawable, data, hDC, imageHandle);
+}
+
+private void init(Drawable drawable, GCData data, long hDC, ImageHandle imageHandle) {
 	int foreground = data.foreground;
 	if (foreground != -1) {
 		data.state &= ~(FOREGROUND | FOREGROUND_TEXT | PEN);
@@ -4485,8 +4494,8 @@ private void init(Drawable drawable, GCData data, long hDC) {
 		data.font = SWTFontProvider.getFont(device, OS.GetCurrentObject(hDC, OS.OBJ_FONT), data.nativeZoom);
 	}
 	Image image = data.image;
-	if (image != null) {
-		data.hNullBitmap = OS.SelectObject(hDC, image.getHandle(data.imageZoom, data.nativeZoom).getHandle());
+	if (imageHandle != null) {
+		data.hNullBitmap = OS.SelectObject(hDC, imageHandle.getHandle());
 		image.memGC = this;
 	}
 	int layout = data.layout;
@@ -5935,12 +5944,12 @@ Point textExtentInPixels(String string, int flags) {
 	return new Point(rect.right, rect.bottom);
 }
 
-void refreshFor(Drawable drawable) {
+void refreshFor(Drawable drawable, ImageHandle imageHandle) {
 	if (drawable == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
 	destroy();
 	GCData newData = new GCData();
 	originalData.copyTo(newData);
-	createGcHandle(drawable, newData);
+	createGcHandle(drawable, newData, imageHandle);
 }
 
 /**
@@ -6059,10 +6068,10 @@ private void removePreviousOperationIfSupercededBy(Operation operation) {
 	}
 }
 
-private void createGcHandle(Drawable drawable, GCData newData) {
-	long newHandle = drawable.internal_new_GC(newData);
-	if (newHandle == 0) SWT.error(SWT.ERROR_NO_HANDLES);
-	init(drawable, newData, newHandle);
+private void createGcHandle(Drawable drawable, GCData newData, ImageHandle imageHandle) {
+	long gcHandle = drawable.internal_new_GC(newData);
+	if (gcHandle == 0) SWT.error(SWT.ERROR_NO_HANDLES);
+	init(drawable, newData, gcHandle, imageHandle);
 	for (Operation operation : operations) {
 		operation.apply();
 	}
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
@@ -20,6 +20,7 @@ import java.io.*;
 import java.util.*;
 import java.util.Map.*;
 import java.util.function.*;
+import java.util.stream.*;
 
 import org.eclipse.swt.*;
 import org.eclipse.swt.internal.*;
@@ -132,18 +133,83 @@ public final class Image extends Resource implements Drawable {
 	 */
 	static final int DEFAULT_SCANLINE_PAD = 4;
 
-	private Map<Integer, DestroyableImageHandle> zoomLevelToImageHandle = new HashMap<>();
-
 	private List<Consumer<Image>> onDisposeListeners;
 
+	private final ImageHandleManager imageHandleManager = new ImageHandleManager();
+
+	private class ImageHandleManager {
+		private Map<Integer, DestroyableImageHandle> zoomLevelToImageHandle = new HashMap<>();
+
+		InternalImageHandle get(int zoom) {
+			final DestroyableImageHandle imageHandle = zoomLevelToImageHandle.get(zoom);
+			if (imageHandle == null) {
+				return null;
+			}
+			return imageHandle;
+		}
+
+
+		InternalImageHandle getOrCreate(int zoom, Supplier<DestroyableImageHandle> creator) {
+			if (zoom == -1) {
+				return null;
+			}
+
+			DestroyableImageHandle imageHandle = (DestroyableImageHandle) get(zoom);
+			if (imageHandle != null) {
+				return imageHandle;
+			}
+
+			imageHandle = creator.get();
+			zoomLevelToImageHandle.put(zoom, imageHandle);
+			return imageHandle;
+		}
+
+		boolean contains(int zoom) {
+			return zoomLevelToImageHandle.containsKey(zoom);
+		}
+
+		boolean isEmpty() {
+		    return zoomLevelToImageHandle.isEmpty();
+		}
+
+		List<InternalImageHandle> getAllImageHandles() {
+			return zoomLevelToImageHandle.values().stream().collect(Collectors.toList());
+		}
+
+		Set<Integer> getAllZooms() {
+			return zoomLevelToImageHandle.keySet();
+		}
+
+		void destroyHandles(Predicate<Integer> filter) {
+			Iterator<Entry<Integer, DestroyableImageHandle>> it = zoomLevelToImageHandle.entrySet().iterator();
+			while (it.hasNext()) {
+				Entry<Integer, DestroyableImageHandle> zoomToHandle = it.next();
+				if (filter.test(zoomToHandle.getKey())) {
+					DestroyableImageHandle imageHandle = zoomToHandle.getValue();
+					it.remove();
+					zoomLevelToImageHandle.remove(imageHandle.zoom, imageHandle);
+					imageHandle.destroy();
+				}
+			}
+		}
+
+		@Override
+		public String toString() {
+			return zoomLevelToImageHandle.toString();
+
+		}
+	}
+
 	private class HandleAtSize {
-		private DestroyableImageHandle handleContainer = null;
+		private ImageHandle handleContainer = null;
+		private DestroyableImageHandle temporaryHandleContainer = null;
 		private int requestedWidth = -1;
 		private int requestedHeight = -1;
 
 		public void destroy() {
-			if (handleContainer != null && !zoomLevelToImageHandle.containsValue(handleContainer)) {
-				handleContainer.destroy();
+			if (temporaryHandleContainer != null) {
+				temporaryHandleContainer.destroy();
+				temporaryHandleContainer = null;
 			}
 			handleContainer = null;
 			requestedWidth = -1;
@@ -169,26 +235,26 @@ public final class Image extends Resource implements Drawable {
 					|| (handleContainer.getHeight() == height && handleContainer.getHeight() == width);
 		}
 
-		private Optional<DestroyableImageHandle> createHandleAtExactSize(int width, int height) {
+		private Optional<ImageHandle> createHandleAtExactSize(int width, int height) {
 			Optional<ImageData> imageData = imageProvider.loadImageDataAtExactSize(width, height);
 			if (imageData.isPresent()) {
 				ImageData adaptedData = adaptImageDataIfDisabledOrGray(imageData.get());
-				DestroyableImageHandle imageHandle = init(adaptedData, -1);
-				return Optional.of(imageHandle);
+				temporaryHandleContainer = init(adaptedData, -1);
+				return Optional.of(temporaryHandleContainer);
 			}
 			return Optional.empty();
 		}
 
-		private DestroyableImageHandle getOrCreateImageHandleAtClosestSize(int widthHint, int heightHint) {
+		private ImageHandle getOrCreateImageHandleAtClosestSize(int widthHint, int heightHint) {
 			Rectangle bounds = getBounds(100);
 			int imageZoomForWidth = 100 * widthHint / bounds.width;
 			int imageZoomForHeight = 100 * heightHint / bounds.height;
 			int imageZoom = DPIUtil.getZoomForAutoscaleProperty(Math.max(imageZoomForWidth, imageZoomForHeight));
-			DestroyableImageHandle bestFittingHandle = zoomLevelToImageHandle.get(imageZoom);
+			ImageHandle bestFittingHandle = imageHandleManager.get(imageZoom);
 			if (bestFittingHandle == null) {
 				ImageData bestFittingImageData = imageProvider.loadImageData(imageZoom).element();
 				ImageData adaptedData = adaptImageDataIfDisabledOrGray(bestFittingImageData);
-				bestFittingHandle = init(adaptedData, -1);
+				bestFittingHandle = temporaryHandleContainer = init(adaptedData, -1);
 			}
 			return bestFittingHandle;
 		}
@@ -296,7 +362,7 @@ public Image(Device device, Image srcImage, int flag) {
 		case SWT.IMAGE_COPY: {
 			switch (type) {
 				case SWT.BITMAP:
-					for (ImageHandle imageHandle : srcImage.zoomLevelToImageHandle.values()) {
+					for (ImageHandle imageHandle : srcImage.imageHandleManager.getAllImageHandles()) {
 						Rectangle rect = imageHandle.getBounds();
 						long srcImageHandle = imageHandle.handle;
 						/* Get the HDC for the device */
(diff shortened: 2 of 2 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -132,17 +132,84 @@ public final class Image extends Resource implements Drawable {
...
-	private Map<Integer, ImageHandle> zoomLevelToImageHandle = new HashMap<>();
-
 	private List<Consumer<Image>> onDisposeListeners;
 
+	private final ImageHandleManager imageHandleManager = new ImageHandleManager();
+
+	private class ImageHandleManager {
+		private Map<Integer, ImageHandle> zoomLevelToImageHandle = new HashMap<>();
+
+		ImageHandle getImageHandle(int zoom) {
```
HeikoKlare: Since this is to a certain degree a wrapper for a data structure map, wouldn't it maybe be sufficient to use according short method names rather than always repeating "handle" and "imageHandle"? I.e.:
- getImageHandle(zoom) -> get(zoom)
- getHandleOrCreate(zoom, creator) -> getOrCreate(zoom, creator)
- hasImageHandle(handle) -> contains(handle)
- hasZoom(zoom) -> contains(zoom) / containsHandleFor(zoom)
- hasNoHandle() -> isEmpty()

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2602,8 +2655,7 @@ ImageHandle initNative(String filename, int zoom) {
 								img.transparentPixel = transparentPixel;
 								img.alphaData = alphaData;
 
-								init(img, zoom);
-								imageMetadata = zoomLevelToImageHandle.get(zoom);
+								imageMetadata = init(img, zoom);
```
HeikoKlare, line 2655: Since init(data, zoom) does not register the handle at the map anymore, is it correct that we don't do it at this consumer?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2456,7 +2507,9 @@ protected ElementAtZoom<ImageData> loadImageData(int zoom) {
 			imageDataAtZoom = ImageDataLoader.loadByZoom(fileForZoom.element(), fileForZoom.zoom(), zoom);
 		} else {
 			imageDataAtZoom = new ElementAtZoom<>(nativeInitializedImage.getImageData(), fileForZoom.zoom());
-			nativeInitializedImage.destroy();
+			if (!imageHandleManager.getAllImageHandles().contains(nativeInitializedImage)) {
+				nativeInitializedImage.destroy();
+			}
```
HeikoKlare: There is already `imageHandleManager.hasImageHandle()` for this. But I think it would be more clear in general if we explicitly store locally that the image needs to be destroyed. E.g., we could simply have a boolean `destroyNativeImage` that is set to true when we use `initNative` and only destroy the handle then.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -132,17 +132,84 @@ public final class Image extends Resource implements Drawable {
...
+	}
+
 	private class HandleAtSize {
 		private ImageHandle handleContainer = null;
 		private int requestedWidth = -1;
 		private int requestedHeight = -1;
 
 		public void destroy() {
-			if (handleContainer != null && !zoomLevelToImageHandle.containsValue(handleContainer)) {
+			if (handleContainer != null && !imageHandleManager.hasImageHandle(handleContainer)) {
```
HeikoKlare: The `hasImageHandle` method is only used here and rather indicates a missing local state handling, I would propose to remove that method and manage the state locally instead.
The method is used to find out, if this `HandleAtSize` has created the handle on its own or whether it was retrieved from the `ImageHandleManager`. So the `HandleAtSize` could just store that it created the handle on its own (inside `refresh()`) and then just use that information here to determine if the handleContainer needs to be destroyed or not instead of indirectly retrieving that information again by querying the ImageHandleManager for whether it contains that handle.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -4461,6 +4461,15 @@ private long identity() {
...
 }
 
+private void initWithImageHandle(Drawable drawable, GCData data, long hDC, ImageHandle imageHandle) {
+	Image image = data.image;
+	data.image = null;
+	init(drawable, data, hDC);
+	data.hNullBitmap = OS.SelectObject(hDC, imageHandle.getHandle());
+	image.memGC = this;
+	data.image = image;
+}
```
HeikoKlare: Why are all these changes to `GC` necessary?

HeikoKlare (commented):
The changes look like a good refactoring to better encapsulate the management of handles for zooms and in general are fine fore me. I still have some proposals for improvements.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
amartya4256: Done.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
amartya4256: `initNative` is called by ImageFileNameProvider#loadImageData to create a temporary `ImageHandle`. Hence not needed. Earlier, it was created and then stored in the map even though the handle was not requested but just the `ImageData`.
Some consumers can call `init` for temporary handles as well. e.g. BaseImageProviderWrapper#newImageData. Hence, in case of only obtaining the ImageData, following the convention, we do not store the `ImageHandle` in the map.

A lot of previous consumers will call new ImageHandle which will automatically register it in the map and then they will call destroy on it which will then clear it from the map which was not a clean approach. This implementation gets rid of that and only adds it where it is needed in the ImageHandleManager.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
amartya4256: You are right. As I can see, `HandleAtSize` doesn't register any `ImageHandle` explicitly and `init` is called with -1, which from the previous convention also means that it wasn't event intended to be added to the map. And as I can see, it is never added in the `ImageHandleManager` with the current implementation so I would simply remove this check as it is always true. And I will get rid of the method.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
amartya4256: image

- PlainImageProviderWrapper#newImageHandle needs to perform `GC#init` amidst the creation of `ImageHandle` using `ImageHandleManager#getOrCreate` which is a synchronous call.
- `GC#init` calls `Image.win32_getHandle` which tries to create the handle since the current ImageHandle is not added to the map yet. As getOrCreate at first waits for the operation to complete and return a handle and then adds it to the map. We could provide a method like `register(ImageHandle)` which adds the ImageHandle amidst the creation of the ImageHandle using this method but then we put the manager at risk that some other consumer inside Image can misuse it. And what if newImageHandle (where the ImageHandle creation takes place) is called by some other consumer? Then the guarantee of synchronous ImageHandle creation will be lost.
- Hence, this takes a ImageHandle instead of the Image and perform the operation using the handle and not Image object.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
amartya4256: Makes sense

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
HeikoKlare: This is not correct. The `handleContainer` may be a handle from the `ImageHandleManager`, as `refresh()` may call `getOrCreateImageHandleAtClosestSize(width, height)` which can retrieve a handle from there. So it is not guaranteed that this handle is locally managed. The latest change may lead to a handle inside the `ImageHandleManager` being disposed.

HeikoKlare: Maybe we should add a check somewhere to better find accidental changes to the implementation that destroys handles from the manager outside of it? E.g., `ImageHandle#destroy()` may check for the handle not being contained in the `ImageHandleManager`.

Or may we can even better do that in an OO-safe way: remove the `destroy` method from `ImageHandle` and create a `DestroyableImageHandle` or the like. Then `ImageHandleManager` could only expose `ImageHandle` instances while internally managing destroyable ones. And the places where handles are created locally would maintain destroyable instances as well. Since we pass `ImageHandles` over to `GC` as well that would even increase safety as those other code places not not accidentally destroy handles anymore as well.

amartya4256: Although with this approach here we still need to check if it is in the manager because either we obtain it from the manager (Exposed Imagehandle) or create a temporary (DestroyableImageHandle)

amartya4256: Also the method which creates ImageHandle is consumed by both temporary user and permanent user, so we can utilize factory pattern here to create the instance of a specific class depending on who the consumer is.

Then we can have the following:
1. ImageHandle (no destroy, created by imageHandleManager, destroyed by a specific method inside the manager)
2. TemporaryHandle extends ImageHandle, has a destroy method.

- we can check if the instance is a temporaryhandle then destroy, in the case above. If type already know, destroying is straight forward.

HeikoKlare: It seems my comment was a bit imprecise. I did not want to propose a distinction between manager-handled `ImageHandle` and temporary `DestroyableImageHandle` but a distinction between "public" API (`ImageHandle`) and "internal" API (`DestroyableImageHandle`), such that only the manager of the handle's lifecycle (either the ImageHandleManager or whoever creates the handle temporarily) have a `DestroyableImageHandle` at hand and only an pure `ImageHandle` is passed outside of those contexts.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
HeikoKlare: > `GC#init` calls `Image.win32_getHandle`

I don't see a direct call of `Image.win32_getHandle()` in `GC#init()`. Are you referring to `image.getHandle()`?
Why not adapt the existing init method to accept the image handle at set it at the proper place and provide a new init method with the existing signature that just delegates to that new method with `image.getHandle()` passed as the additionally required image handle?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -6059,10 +6068,10 @@ private void removePreviousOperationIfSupercededBy(Operation operation) {
 	}
 }
 
-private void createGcHandle(Drawable drawable, GCData newData) {
+private void createGcHandle(Drawable drawable, GCData newData, ImageHandle imageHandle) {
 	long newHandle = drawable.internal_new_GC(newData);
 	if (newHandle == 0) SWT.error(SWT.ERROR_NO_HANDLES);
-	init(drawable, newData, newHandle);
+	initWithImageHandle(drawable, newData, newHandle, imageHandle);
```
HeikoKlare: It might be good to rename `newHandle` to `gcHandle` then to be able to easily distinguish between the image and the gc handle involved here.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
HeikoKlare: See:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2938

Maybe we can take that as a preparatory step and then do this one on top.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -3191,16 +3243,14 @@ private class DestroyableImageHandle extends InternalImageHandle {
...
 	}
 
 	@Override
 	boolean isDisposed() {
 		return isDisposed;
 	}
 
-	void destroy() {
-		zoomLevelToImageHandle.remove(zoom, this);
+	private void destroy() {
```
HeikoKlare: Can stay package protected as it's only usable inside this class anyway.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2443,18 +2493,22 @@ protected ElementAtZoom<ImageData> loadImageData(int zoom) {
...
+			canDestroy = true;
 		}
 		ElementAtZoom<ImageData> imageDataAtZoom;
 		if (nativeInitializedImage == null) {
 			imageDataAtZoom = ImageDataLoader.loadByZoom(fileForZoom.element(), fileForZoom.zoom(), zoom);
 		} else {
 			imageDataAtZoom = new ElementAtZoom<>(nativeInitializedImage.getImageData(), fileForZoom.zoom());
-			nativeInitializedImage.destroy();
+			if (canDestroy) {
+				((DestroyableImageHandle) nativeInitializedImage).destroy();
```
HeikoKlare: Instead of a boolean and doing an optimistic cast here, why not store the handle in an additional `DestroyableImageHandle`-typed variable and destroy it if present?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2290,14 +2341,14 @@ protected ElementAtZoom<ImageData> loadImageData(int zoom) {
...
-		if (zoomLevelToImageHandle.isEmpty()) {
+		if (imageHandleManager.isEmpty()) {
 			return createBaseHandle(targetZoom);
 		}
 		if (memGC != null) {
 			GC currentGC = memGC;
 			memGC = null;
-			ImageHandle imageHandle = createHandle(targetZoom);
-			currentGC.refreshFor(new DrawableWrapper(Image.this, zoomContext));
+			DestroyableImageHandle imageHandle = createHandle(targetZoom);
```
HeikoKlare, line 2349: This should remain `ImageHandle`

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
amartya4256: Yes sorry I meant Image#getHandle

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -132,18 +133,82 @@ public final class Image extends Resource implements Drawable {
...
+		private ImageHandle handleContainer = null;
 		private int requestedWidth = -1;
 		private int requestedHeight = -1;
+		private DestroyableImageHandle destroyableHandleContainer = null;
 
 		public void destroy() {
-			if (handleContainer != null && !zoomLevelToImageHandle.containsValue(handleContainer)) {
-				handleContainer.destroy();
+			if (destroyableHandleContainer != null) {
+				destroyableHandleContainer.destroy();
```
HeikoKlare: you need to (or at least should) null `destroyableHandleContainer` here.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -169,26 +234,26 @@ private boolean isReusable(int width, int height) {
...
 		}
 
-		private Optional<DestroyableImageHandle> createHandleAtExactSize(int width, int height) {
+		private Optional<ImageHandle> createHandleAtExactSize(int width, int height) {
 			Optional<ImageData> imageData = imageProvider.loadImageDataAtExactSize(width, height);
 			if (imageData.isPresent()) {
 				ImageData adaptedData = adaptImageDataIfDisabledOrGray(imageData.get());
-				DestroyableImageHandle imageHandle = init(adaptedData, -1);
+				DestroyableImageHandle imageHandle = destroyableHandleContainer = init(adaptedData, -1);
 				return Optional.of(imageHandle);
```
HeikoKlare:
```suggestion
				destroyableHandleContainer = init(adaptedData, -1);
				return Optional.of(destroyableHandleContainer);
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2443,18 +2491,21 @@ protected ElementAtZoom<ImageData> loadImageData(int zoom) {
 		}
 
 		// Load at file zoom (native or via loader) and rescale
-		DestroyableImageHandle nativeInitializedImage;
-		if (zoomLevelToImageHandle.containsKey(fileForZoom.zoom())) {
-			nativeInitializedImage = zoomLevelToImageHandle.get(fileForZoom.zoom());
+		InternalImageHandle nativeInitializedImage;
+		DestroyableImageHandle destroyableNativeInitializedImage = null;
```
HeikoKlare: I would rather call it `temporaryImageHandle` than `destroyableNatibeInitializedImage`, as it better reflects the temporary purpose of it.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -132,18 +133,82 @@ public final class Image extends Resource implements Drawable {
...
+
+		}
+	}
+
 	private class HandleAtSize {
-		private DestroyableImageHandle handleContainer = null;
+		private ImageHandle handleContainer = null;
 		private int requestedWidth = -1;
 		private int requestedHeight = -1;
+		private DestroyableImageHandle destroyableHandleContainer = null;
```
HeikoKlare: Maybe rather call this `localHandleContainer` or the like, as the object's role is about being temporary/local and not about being destroyable.
And nit: maybe move it up over the "requested" fields, so that both image handle fields are placed next to each other.

HeikoKlare (commented):
I have only minor/nitpicky comments left.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
amartya4256: How about temporary since we are using that in a few other places?

HeikoKlare: sure, sounds good

HeikoKlare (approved):

PR #2929, opened 2026-01-06
Extend Snippet388 to generate PDF directly drawing on GC

akoch-yatta:
Currently Snippet388 showcases the PDF export by using Shell#print. This PR extends it to additionally provide the possibility to export a direct rendering on the GC.


laeubi:
> This PR extends it to additionally provide the possibility to export a direct rendering on the GC

The snippet already prints an additional text when exporting to PDF so I'm not completely sure what is missing?

Also as we are painting to a canvas everything is already implicitly rendered through the GC anyways isn't it?

akoch-yatta:
> > This PR extends it to additionally provide the possibility to export a direct rendering on the GC
> 
> The snippet already prints an additional text when exporting to PDF so I'm not completely sure what is missing?

Yes, the text is there, but I would propose a set of more complex operations to be applied to showcase the usage

> Also as we are painting to a canvas everything is already implicitly rendered through the GC anyways isn't it?

Still, there are differences, e.g. if I replace shell::print with canvas::print in the Snippet (to make it as similar as possible by directly applying the GC operations) the export looks different: background color + SWT.BORDER effect of the Canvas is shown when using canvas::print.

I mostly adapted to snippet to test the behavior with Windows, but I thought it makes sense to have a Snippet that utilizes both uses cases in a complex scenario.

laeubi:
> Still, there are differences, e.g. if I replace shell::print with canvas::print in the Snippet (to make it as similar as possible by directly applying the GC operations) the export looks different: background color + SWT.BORDER effect of the Canvas is shown when using canvas::print.

Yes print will maybe print more of the controls parts. It was a bit clumsy as shell#print was not implemented in windows at all so there might me slight differences as well in when printing the canvas directly and when printing the whole shell.

akoch-yatta:
> > Still, there are differences, e.g. if I replace shell::print with canvas::print in the Snippet (to make it as similar as possible by directly applying the GC operations) the export looks different: background color + SWT.BORDER effect of the Canvas is shown when using canvas::print.
> 
> Yes print will maybe print more of the controls parts. It was a bit clumsy as shell#print was not implemented in windows at all so there might me slight differences as well in when printing the canvas directly and when printing the whole shell.

Which is fine in my opinion. I just wanted to give the option to check the export with both approaches.

At least in windows they provide very different results depending on zoom, autoscale mode and advanced mode setting of the GC. But that is a problem in Control#print most likely.

PR #2930, opened 2026-01-06
Add Shell#getZoom()

amartya4256:
This adds a `Shell#getZoom()` method to make the zoom of a shell as considered by the OS available to consumers. Currently, there is a (undocumented) data field on controls in the Windows implementation for this (`SHELL_ZOOM`). And for other operating system (which only use a static zoom value for everything), the value is provided internally via `DPIUtil#getNativeDeviceZoom()` and may also be calculated from `Device#getDPI()`.
It also serves as a preparation for allowing explicit disablement of auto-scaling for specific controls as proposed with https://github.com/eclipse-platform/eclipse.platform.swt/pull/2863, which requires access to the shell's zoom to custom-zoom auto-scale-disabled controls.

This PR also renames the method Widget#getZoom to getAutoscalingZoom since this methods return the value of the zoom based on the autoscaling technique, i.e. quarter, half, etc and whether autoscaling is enabled or disabled, and would conflict with the newly added `Shell#getZoom()` method.
Note: This change has been made using Eclipse refactoring tool.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Shell.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Shell.java
@@ -18,6 +18,7 @@ import java.util.*;
 import org.eclipse.swt.*;
 import org.eclipse.swt.events.*;
 import org.eclipse.swt.graphics.*;
+import org.eclipse.swt.internal.*;
 import org.eclipse.swt.internal.cocoa.*;
 
 /**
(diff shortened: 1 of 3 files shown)


HeikoKlare (commented):
What's the purpose of this renaming?

amartya4256:
> What's the purpose of this renaming?

Since https://github.com/eclipse-platform/eclipse.platform.swt/pull/2863#discussion_r2661342249 in this comment youhad requested to expose the API with a better name that makes sense, i.e. `getZoom()`, it makes sense to rename the internal method `getZoom()` to something specific to its behaviour, hence the refactoring.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Shell.java
```diff
@@ -1992,6 +1992,21 @@ public void setModified (boolean modified) {
...
 }
 
+/**
+ * Returns the zoom of the shell.
+ * <p>
+ * Hint: The returned value is the zoom of the shell as originally considered by
+ * the OS and not an adjusted zoom value as considered by SWT autoscaling capabilities.
+ * </p>
+ *
+ * @return the zoom for the current shell
```
HeikoKlare:
```suggestion
 * @return the zoom for this shell
```

PR #2934, opened 2026-01-06
Fix GC#drawImage methods with additional Transformation

al-noori:
If a transform is applied via the GC, then both drawImage APIs consider now the best fitting handle by requesting it through the width/height of the full image scaled by the scaleFactor (relating destination width/height to source width/height) times the width/height scaling induced by the transformation.

Note that only the second commit is relevant for the PR. The first one references to the changes made in #2913. 

To easily see and test, use the following snippet:

````java

public class Snippet389 {

	shell.setText("Snippet 389");
	ImageFileNameProvider fileNameProvider = zoom -> {
		if (zoom == 200) {
			return "src/org/eclipse/swt/snippets/eclipse32.png";
		} else if (zoom == 100) {
			return "src/org/eclipse/swt/snippets/eclipse16.png";
		}
		return null;
	};
	Image image = new Image(display, fileNameProvider);
	shell.addPaintListener(e -> {
		e.gc.drawImage(image, 0, 0, 32, 32);
		e.gc.drawImage(image, 0, 0, 16, 16, 0, 50, 32, 32);
	});

	shell.addPaintListener(e -> {
		e.gc.setTransform(new Transform(display, 2, 0, 0, 2, 0, 0));
		e.gc.drawImage(image, 25, 0, 16, 16);
		e.gc.drawImage(image, 0, 0, 16, 16, 25, 25, 16, 16);
		e.gc.setTransform(null);
	});

	shell.open ();

	}
	display.dispose ();
}
}
````

Below are screenshots of the snippet on a 100% zoom.

**Before:**
Before

**After:**
After

Contributes to https://github.com/eclipse-platform/eclipse.platform.swt/issues/2919

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
@@ -521,6 +521,20 @@ private abstract class ImageOperation extends Operation {
 		image.removeOnDisposeListener(disposeCallback);
 		super.disposeAll();
 	}
+
+	protected float calculateTransformationScale() {
+		Transform current = new Transform(device);
+		getTransform(current);
+		float[] m = new float[6];
+		current.getElements(m);
+		// this calculates the effective length in x and y
+		// direction without being affected by the rotation
+		// of the transformation
+		float scaleWidth = (float) Math.hypot(m[0], m[2]);
+		float scaleHeight = (float) Math.hypot(m[1], m[3]);
+		current.dispose();
+		return Math.max(scaleWidth, scaleHeight);
+	}
 }
 
 private class CopyAreaToImageOperation extends ImageOperation {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java [hunk 1]
```diff
@@ -1214,14 +1214,28 @@ void apply() {
 	}
 }
 
+private float calculateTransformationScale() {
+	Transform current = new Transform(device);
```
arunjose696: Transform is a resource so if it is created it should be disposed at the end of this function

arunjose696:
Have tested the changes myself and can see it fixes the said issue.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
al-noori: Thanks for the info, i edited the method such that the transform is disposed in the end.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -1252,18 +1254,33 @@ private class DrawScaledImageOperation extends ImageOperation {
...
+	}
+}
+
+private float calculateTransformationScale() {
+	Transform current = new Transform(device);
+	getTransform(current);
+	float[] m = new float[6];
+	current.getElements(m);
+	float scaleWidth = (float) Math.hypot(m[0], m[2]);
+	float scaleHeight = (float) Math.hypot(m[1], m[3]);
```
HeikoKlare: I think it would help to add a comment with an explanation of a link to some explaining source why this calculates the right scale values out of the matrix.
It's not that easy to understand as the matrix of course also contains a rotations, so I have to admit I do not easily understand if/why this calculation is the right one.

akoch-yatta: I added a short comment to it

akoch-yatta (approved):
LGTM, I only rebased, moved a bit of code around and added a comment with my commits. Still I would ask for a second approval (@HeikoKlare)

I tested especially with Snippet389 and a slightly adjusted version to test with the eclipse16.svg to ensure is is really creating the perfect solution when putting in high scale factors

Issue #2917, opened 2025-12-29
Non-blurred SVG image scaling with GC.drawImage(Image image, int destX, int destY, int destWidth, int destHeight) only works on Windows unless an ImageFileNameProvider is used


Phillipus:
- Eclipse 4.38

According to [this](https://eclipse.dev/eclipse/markdown/?f=news/4.38/platform_isv.md#new-gc-api-for-drawing-scaled-images):

> When using the new drawImage method, images that can provide size-specific data (such as SVGs) are loaded at the destination size, avoiding blurry scaling.

However that's not the case for Mac or Linux unless an `ImageFileNameProvider` is used.

Run this Snippet on Windows, Mac and Linux and compare the results:

```java
import java.nio.file.Path;
import org.eclipse.swt.graphics.Color;
import org.eclipse.swt.graphics.Image;
import org.eclipse.swt.graphics.ImageFileNameProvider;
import org.eclipse.swt.layout.FillLayout;
import org.eclipse.swt.widgets.Display;
import org.eclipse.swt.widgets.Shell;

public class SVGImageTests {
    
    public static void main(String[] args) throws Exception {
        final Display display = new Display();

        final Shell shell = new Shell(display);
        shell.setText("SVG Image Test");
        shell.setLayout(new FillLayout());
        shell.setBounds(250, 50, 1000, 400);
        shell.setBackground(new Color(255, 255, 255));
        
        String imgPath = "image.svg";
        Path fullPath = Path.of(SVGImageTests.class.getResource(imgPath).toURI());
        
        // Image loaded in stream
        Image image1 = new Image(display, SVGImageTests.class.getResourceAsStream(imgPath));

        // Image loaded in file path
        Image image2 = new Image(display, fullPath.toString());
        
        // Image loaded in ImageFileNameProvider
        Image image3 = new Image(display, (ImageFileNameProvider) zoom -> {
            return zoom == 100 ? fullPath.toString() : null;
        });
        
        shell.addPaintListener(e -> {
            e.gc.drawImage(image1, 20, 10, 300, 300);
            e.gc.drawText("Image(Display, InputStream)", 80, 300);
            
            e.gc.drawImage(image2, 340, 10, 300, 300);
            e.gc.drawText("Image(Display, String)", 420, 300);
            
            e.gc.drawImage(image3, 660, 10, 300, 300);
            e.gc.drawText("Image(Display, ImageFileNameProvider)", 700, 300);
        });
        
[...]

PR #2926, opened 2026-01-03
Add snippet to compare SVG image loading using three Image constructors

Phillipus:
Added this as requested [here](https://github.com/eclipse-platform/eclipse.platform.swt/pull/2924#pullrequestreview-3624326441).

Not sure about how these things are named so called it the next in line `Snippet 389` (Edit - now Snippet 390)

Please let me know if any changes required.

PR #2924, opened 2026-01-02
Delegating Image(Display, String) to ImageFileNameProvider to init Image

ShahzaibIbrahim:
In **GTK** and **Cocoa**, an image based on an SVG passed as filename to Image(Device, String) will be drawn now sharply with this change.

Contributes to https://github.com/eclipse-platform/eclipse.platform.swt/issues/2917

### How to Test

Run this Snippet on Windows, Mac and Linux and compare the results:
```
import java.nio.file.Path;
import org.eclipse.swt.graphics.Color;
import org.eclipse.swt.graphics.Image;
import org.eclipse.swt.graphics.ImageFileNameProvider;
import org.eclipse.swt.layout.FillLayout;
import org.eclipse.swt.widgets.Display;
import org.eclipse.swt.widgets.Shell;

public class SVGImageTests {
    
    public static void main(String[] args) throws Exception {
        final Display display = new Display();

        final Shell shell = new Shell(display);
        shell.setText("SVG Image Test");
        shell.setLayout(new FillLayout());
        shell.setBounds(250, 50, 1000, 400);
        shell.setBackground(new Color(255, 255, 255));
        
        String imgPath = "image.svg";
        Path fullPath = Path.of(SVGImageTests.class.getResource(imgPath).toURI());

        // Image loaded in file path
        Image image2 = new Image(display, fullPath.toString());
        
        // Image loaded in ImageFileNameProvider
        Image image3 = new Image(display, (ImageFileNameProvider) zoom -> {
            return zoom == 100 ? fullPath.toString() : null;
        });
        
        shell.addPaintListener(e -> {            
            e.gc.drawImage(image2, 340, 10, 300, 300);
            e.gc.drawText("Image(Display, String)", 420, 300);
            
            e.gc.drawImage(image3, 660, 10, 300, 300);
            e.gc.drawText("Image(Display, ImageFileNameProvider)", 700, 300);
        });
        
        shell.open();

        while(!shell.isDisposed()) {
            if(!display.readAndDispatch()) {
                display.sleep();
            }
        }

        display.dispose();
    }
}
```
Use this SVG image for the Snippet:

image

Here's the outputs...

On Windows all Image methods result in sharp output:

image

On Mac only using ImageFileNameProvider is sharp:

image

On Linux only using ImageFileNameProvider is sharp:

image

### After Fix

Image use for testing: 
eclipse16

GTK: 

image

Cocoa:

image

PR #2935, opened 2026-01-07
Sharp rendering of SVGs passed as InputStream

ShahzaibIbrahim:
Wrapping the input stream handling in an ImageDataProvider. In case stream data belong to a dynamically sizable image we create a provider with ImageDataAtSizeProvider else we use ImageDataProvider.

- The ImageDataAtSizeProvider will store the stream data and return accordingly loaded image data in its getImageData(width, height) and getImageData(zoom) methods.
- The ImageDataProvider will only return according image data at 100% and none for the other zoom, so it does not need to store the stream data.

Contributes to: https://github.com/eclipse-platform/eclipse.platform.swt/issues/2917

### How to Test

- Run the Snippet390 on Windows, Mac and Linux and compare the results
- Currently, In GTK and Cocoa, an image based on an SVG passed as inputStream to Image(Device, InputStream) was blurry. With the current change it should look as good as ImageFileNameProvider as shown in snippet.

---
### Results

**Without this PR:**
image

---
**After Changes:** 

GTK:

image

Cocoa:

image

- [x] Requires: https://github.com/eclipse-platform/eclipse.platform.swt/pull/2954

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
@@ -691,9 +691,8 @@ public Image(Device device, InputStream stream) {
 	NSAutoreleasePool pool = null;
 	if (!NSThread.isMainThread()) pool = (NSAutoreleasePool) new NSAutoreleasePool().alloc().init();
 	try {
-		byte[] input = stream.readAllBytes();
-		initWithSupplier(zoom -> ImageDataLoader.canLoadAtZoom(new ByteArrayInputStream(input), FileFormat.DEFAULT_ZOOM, zoom),
-				zoom -> ImageDataLoader.loadByZoom(new ByteArrayInputStream(input), FileFormat.DEFAULT_ZOOM, zoom).element());
+		ImageDataProvider imageDataProvider = createImageDataProvider(stream);
+		initUsingImageDataProvider(imageDataProvider);
 		init();
 	} catch (IOException e) {
 		SWT.error(SWT.ERROR_IO, e);
@@ -847,24 +846,14 @@ private void initUsingFileNameProvider(ImageFileNameProvider imageFileNameProvid
 public Image(Device device, ImageDataProvider imageDataProvider) {
 	super(device);
 	if (imageDataProvider == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
-	this.imageDataProvider = imageDataProvider;
-	ImageData data = imageDataProvider.getImageData (100);
-	if (data == null) SWT.error(SWT.ERROR_INVALID_ARGUMENT);
 	NSAutoreleasePool pool = null;
 	if (!NSThread.isMainThread()) pool = (NSAutoreleasePool) new NSAutoreleasePool().alloc().init();
 	try {
-		init (data, 100);
+		initUsingImageDataProvider(imageDataProvider);
 		init ();
 		StrictChecks.runIfStrictChecksEnabled(() -> {
 			DPIUtil.validateLinearScaling(imageDataProvider);
 		});
-		ImageData data2x = imageDataProvider.getImageData (200);
-		if (data2x != null) {
-			alphaInfo_200 = new AlphaInfo();
-			NSBitmapImageRep rep = createRepresentation (data2x, alphaInfo_200);
-			handle.addRepresentation(rep);
-			rep.release();
-		}
 	} finally {
 		if (pool != null) pool.release();
 	}
@@ -1511,11 +1500,15 @@ void init(ImageData image, int imageZoom) {
 	handle.setCacheMode(OS.NSImageCacheNever);
 }
 
-private void initWithSupplier(Function<Integer, Boolean> canLoadAtZoom, Function<Integer, ImageData> zoomToImageData) {
-	ImageData imageData = zoomToImageData.apply(100);
+private void initUsingImageDataProvider(ImageDataProvider imageDataProvider) {
+	this.imageDataProvider = imageDataProvider;
+	ImageData imageData = imageDataProvider.getImageData(100);
+	if (imageData == null) {
+		SWT.error(SWT.ERROR_INVALID_ARGUMENT);
+	}
 	init(imageData, 100);
-	if (canLoadAtZoom.apply(200)) {
-		ImageData imageData2x = zoomToImageData.apply(200);
+	ImageData imageData2x = imageDataProvider.getImageData(200);
+	if (imageData2x != null) {
 		alphaInfo_200 = new AlphaInfo();
 		NSBitmapImageRep rep = createRepresentation (imageData2x, alphaInfo_200);
 		handle.addRepresentation(rep);
@@ -1523,6 +1516,19 @@ private void initWithSupplier(Function<Integer, Boolean> canLoadAtZoom, Function
 	}
 }
 
+private static ImageDataProvider createImageDataProvider(InputStream stream) throws IOException {
+	byte[] streamData = stream.readAllBytes();
+	if (ImageDataLoader.isDynamicallySizable(new ByteArrayInputStream(streamData))) {
+		ImageDataAtSizeProvider imageDataAtSizeProvider = (width, height) -> ImageDataLoader
+				.loadBySize(new ByteArrayInputStream(streamData), width, height);
+		return imageDataAtSizeProvider;
+	}
+
+	ImageData imageData = ImageDataLoader
+			.loadByZoom(new ByteArrayInputStream(streamData), FileFormat.DEFAULT_ZOOM, 100)
+			.element();
+	return zoom -> zoom == 100 ? imageData : null;
+}
 
 void initAlpha_200(NSBitmapImageRep nativeRep) {
 	NSAutoreleasePool pool = null;
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Image.java
@@ -554,11 +554,17 @@ public Image(Device device, ImageData source, ImageData mask) {
  */
 public Image(Device device, InputStream stream) {
 	super(device);
-	currentDeviceZoom = DPIUtil.getDeviceZoom();
-	ElementAtZoom<ImageData> image = ImageDataLoader.loadByZoom(stream, FileFormat.DEFAULT_ZOOM, currentDeviceZoom);
-	ImageData data = DPIUtil.scaleImageData(device, image, currentDeviceZoom);
-	init(data);
-	init();
+	if (stream == null) {
+		SWT.error(SWT.ERROR_NULL_ARGUMENT);
+	}
+	try {
+		currentDeviceZoom = DPIUtil.getDeviceZoom();
+		this.imageDataProvider = createImageDataProvider(stream);
+		initFromImageDataProvider(currentDeviceZoom);
+		init();
+	} catch (IOException e) {
+		SWT.error(SWT.ERROR_IO, e);
+	}
 }
 
 /**
@@ -800,6 +806,20 @@ private void initFromImageDataProvider(int zoom) {
 	ElementAtZoom<ImageData> data = DPIUtil.validateAndGetImageDataAtZoom (imageDataProvider, zoom);
 	ImageData resizedData = DPIUtil.scaleImageData (device, data.element(), zoom, data.zoom());
 	init(resizedData);
+}
+
+private static ImageDataProvider createImageDataProvider(InputStream stream) throws IOException {
+	byte[] streamData = stream.readAllBytes();
+	if (ImageDataLoader.isDynamicallySizable(new ByteArrayInputStream(streamData))) {
+		ImageDataAtSizeProvider imageDataAtSizeProvider = (width, height) -> ImageDataLoader
+				.loadBySize(new ByteArrayInputStream(streamData), width, height);
+		return imageDataAtSizeProvider;
+	}
+
+	ImageData imageData = ImageDataLoader
+			.loadByZoom(new ByteArrayInputStream(streamData), FileFormat.DEFAULT_ZOOM, 100)
+			.element();
+	return zoom -> zoom == 100 ? imageData : null;
 }
 
 void createFromPixbuf(int type, long pixbuf) {


laeubi:
@ShahzaibIbrahim please provide a consistent PR description and at best show before/after screenshot.
Also commit descriptions seem not sufficient.

ShahzaibIbrahim:
> @ShahzaibIbrahim please provide a consistent PR description and at best show before/after screenshot. Also commit descriptions seem not sufficient.

Hi @laeubi, This PR is not ready and hence it's currently drafted.

laeubi:
> > @ShahzaibIbrahim please provide a consistent PR description and at best show before/after screenshot.  Also commit descriptions seem not sufficient. 
> 
> Hi @laeubi, This PR is not ready and hence it's currently drafted.

I appreciate that it's marked as a draft, but even draft PRs should contain a clear title and meaningful description explaining what you're working on and what the expected outcome is. This helps maintainers understand the context and provide better guidance early in the process.

The SWT repository should have draft PRs that clearly communicate:
- What issue or improvement you're addressing
- What approach you're taking
- What the current state is and what still needs to be done

If you need to run CI checks while developing, you can do that on your fork as well. Please update the PR description before continuing, so we can provide better feedback.  Thanks!

ShahzaibIbrahim:
> I appreciate that it's marked as a draft, but even draft PRs should contain a clear title and meaningful description explaining what you're working on and what the expected outcome is. This helps maintainers understand the context and provide better guidance early in the process.
> 
> The SWT repository should have draft PRs that clearly communicate:
> 
> * What issue or improvement you're addressing
> * What approach you're taking
> * What the current state is and what still needs to be done

I have updated the PR description with all the sufficient information I currently have.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
```diff
@@ -146,6 +146,11 @@ public final class Image extends Resource implements Drawable {
 	 */
 	private ImageGcDrawer imageGcDrawer;
 
+	/**
+	 * Byte array to store input stream data to draw on various Zoom levels
+	 */
+	private byte[] inputStreamData;
```
laeubi: Would this not better be part of a supplier instead of part of the image?

Phillipus:
Some early feedback on Mac for this one...

Here's a screenshot running this [Snippet](https://github.com/eclipse-platform/eclipse.platform.swt/pull/2926) of the changes in this PR and https://github.com/eclipse-platform/eclipse.platform.swt/pull/2924

Screenshot 2026-01-07 at 14 14 49

@ShahzaibIbrahim Thank-you for your work on this and https://github.com/eclipse-platform/eclipse.platform.swt/pull/2924

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
ShahzaibIbrahim: I need o make it part of Image class because I have to use it in private class `CachedImageAtSize#loadImageDataAtExactSize` not just in `initWithSupplier`. Or do you mean something else?

ShahzaibIbrahim:
@HeikoKlare I think tests are fixed. Can you please test my snippet on multiple zoom level on mac to mark this PR ready for review?

Phillipus:
Mac with latest changes:

Screenshot 2026-01-09 at 17 08 25

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Image.java
```diff
@@ -554,11 +559,17 @@ public Image(Device device, ImageData source, ImageData mask) {
...
-	ElementAtZoom<ImageData> image = ImageDataLoader.loadByZoom(stream, FileFormat.DEFAULT_ZOOM, currentDeviceZoom);
-	ImageData data = DPIUtil.scaleImageData(device, image, currentDeviceZoom);
-	init(data);
-	init();
+	try {
+		this.inputStreamData = stream.readAllBytes();
+		initDataFromInputStream(currentDeviceZoom);
+		init();
+	} catch (IOException e) {
+		SWT.error(SWT.ERROR_INVALID_ARGUMENT, e);
```
HeikoKlare: This is not the correct error w.r.t. the contract. It should be `SWT.ERROR_IO`.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Image.java
```diff
@@ -763,6 +774,15 @@ boolean refreshImageForZoom () {
 			refreshed = true;
 			currentDeviceZoom = deviceZoomLevel;
 		}
+	} else if (inputStreamData != null) {
+		int deviceZoomLevel = deviceZoom;
+		if (deviceZoomLevel != currentDeviceZoom) {
```
HeikoKlare: One could think of checking for `isDynamicallySizable` here, as otherwise you will unnecessary destroy and recreate the exact same handle. However, it's not done for the other providers as well and with my general remark, this will not be easily possible either.

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/SVGImageTests.java
```diff
@@ -0,0 +1,48 @@
+package org.eclipse.swt.snippets;
+import java.nio.file.Path;
+
+import org.eclipse.swt.graphics.*;
+import org.eclipse.swt.layout.*;
+import org.eclipse.swt.widgets.*;
+
+public class SVGImageTests {
```
HeikoKlare: There is already `Snippet390`, so this one is obsolete.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Image.java
```diff
@@ -161,6 +161,11 @@ public final class Image extends Resource implements Drawable {
 	 */
 	private ImageGcDrawer imageGcDrawer;
 
+	/**
+	 * Byte array to store input stream data to draw on various Zoom levels
```
HeikoKlare: There is no drawing happening with input streams (same for the Cocoa implementation).
```suggestion
	 * Byte array to store input stream data to convert into image data at various Zoom levels
```

examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/SVGImageTests.java
ShahzaibIbrahim: Yes, I meant to remove it in my final version. I added it for testing on my linux machine.

ShahzaibIbrahim:
> Following up on the [comment](https://github.com/eclipse-platform/eclipse.platform.swt/pull/2935/files#r2668332803) of @laeubi, wouldn't it be cleaner to wrap the input stream handling in an `ImageDataProvider` to not further mess up the `Image` implementation with mutual exclusive providers? It might work as follows (starting from the `InputStream` constructor):
> 
> * Read stream data from stream
> * Check if stream data belong to a dynamically sizable image, in that case create an `ImageDataAtSizeProvider` otherwise create an `ImageDataProvider`.
>   
>   * The `ImageDataAtSizeProvider` will store the stream data and return accordingly loaded image data in its `getImageData(width, height)` and `getImageData(zoom)` methods
>   * The `ImageDataProvider` will only return according image data at 100% and none for the other zoom, so it does not need to store the stream data
> 
> Nothing needs to be changed for the `loadImageDataAtExactSize` then, as it reuses the `ImageDataAtSizeProvider` functionality.

I have adapted the changes as per your recommendation. It looks good in GTK but some tests are failing in cocoa but I can't figure why. Can you please test for me in mac? @HeikoKlare

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
```diff
@@ -688,15 +688,29 @@ public Image(Device device, InputStream stream) {
...
+	}
+	ImageDataProvider imageDataProvider;
+	if (ImageDataLoader.isDynamicallySizable(new ByteArrayInputStream(streamData))) {
+	    imageDataProvider =  (ImageDataAtSizeProvider) (width, height) -> ImageDataLoader.loadBySize(
+		    new ByteArrayInputStream(streamData),
+		    width,
+		    height
+		);
+	} else {
+		imageDataProvider = zoom -> zoom == 100 ? ImageDataLoader.loadByZoom(new ByteArrayInputStream(streamData), FileFormat.DEFAULT_ZOOM, zoom).element() : null;
```
HeikoKlare: I would still say that for the plain image data provider we should not store the stream but rather the data (see https://github.com/eclipse-platform/eclipse.platform.swt/pull/2935#pullrequestreview-3646579488). We can only produce image data at one zoom out the stream anyway, so there is not need to do the processing again when requesting image data from the provider again.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
```diff
@@ -846,30 +860,34 @@ private void initUsingFileNameProvider(ImageFileNameProvider imageFileNameProvid
...
-			NSBitmapImageRep rep = createRepresentation (data2x, alphaInfo_200);
-			handle.addRepresentation(rep);
-			rep.release();
-		}
 	} finally {
 		if (pool != null) pool.release();
 	}
 }
 
+private void initUsingImageDataProvider(ImageDataProvider imageDataProvider) {
```
HeikoKlare: Would be good to keep this method at the same position as the previous `initFromSupplier` such that the change is smaller.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
```diff
@@ -688,15 +688,29 @@ public Image(Device device, InputStream stream) {
 	if (stream == null) {
 		SWT.error(SWT.ERROR_NULL_ARGUMENT);
 	}
+	byte[] streamData;
+	try {
+		streamData = stream.readAllBytes();
+	} catch (IOException e) {
+		SWT.error(SWT.ERROR_IO, e);
+		return;
```
HeikoKlare: No need to return as `SWT.error` will throw an error anyway. And I would rather create a separate method that maps the stream to an ImageDataProvider, such that you can simply replace the existing `initWithSupplier` call with an `createImageDataProvider()` and `initUsingImageDataProvider` call, without any ned to add further try/catch and the like.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
```diff
@@ -685,18 +684,14 @@ public Image(Device device, ImageData source, ImageData mask) {
  */
 public Image(Device device, InputStream stream) {
 	super(device);
-	if (stream == null) {
-		SWT.error(SWT.ERROR_NULL_ARGUMENT);
-	}
```
HeikoKlare, line 690: Why not keep the check here?

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
```diff
@@ -685,18 +684,14 @@ public Image(Device device, ImageData source, ImageData mask) {
...
 	try {
-		byte[] input = stream.readAllBytes();
-		initWithSupplier(zoom -> ImageDataLoader.canLoadAtZoom(new ByteArrayInputStream(input), FileFormat.DEFAULT_ZOOM, zoom),
-				zoom -> ImageDataLoader.loadByZoom(new ByteArrayInputStream(input), FileFormat.DEFAULT_ZOOM, zoom).element());
+		ImageDataProvider imageDataProvider = createImageDataProvider(stream);
+		initUsingImageDataProvider(imageDataProvider);
 		init();
 	} catch (IOException e) {
-		SWT.error(SWT.ERROR_INVALID_ARGUMENT, e);
+		SWT.error(SWT.ERROR_IO, e);
```
HeikoKlare, line 698: Does this mean the exception type was wrong? Then we should fix that in a separate PR to keep one PR/commit per concern.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
```diff
@@ -846,25 +841,11 @@ private void initUsingFileNameProvider(ImageFileNameProvider imageFileNameProvid
  */
 public Image(Device device, ImageDataProvider imageDataProvider) {
 	super(device);
-	if (imageDataProvider == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
```
HeikoKlare, line 849: Why not keep the check here?

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
```diff
@@ -1495,18 +1476,42 @@ void init(ImageData image, int imageZoom) {
...
-private void initWithSupplier(Function<Integer, Boolean> canLoadAtZoom, Function<Integer, ImageData> zoomToImageData) {
-	ImageData imageData = zoomToImageData.apply(100);
-	init(imageData, 100);
-	if (canLoadAtZoom.apply(200)) {
-		ImageData imageData2x = zoomToImageData.apply(200);
+private void initUsingImageDataProvider(ImageDataProvider imageDataProvider) {
+	if (imageDataProvider == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
+	this.imageDataProvider = imageDataProvider;
+	ImageData data = imageDataProvider.getImageData (100);
+	if (data == null) SWT.error(SWT.ERROR_INVALID_ARGUMENT);
```
HeikoKlare: This check could be kept in the existing caller, as the new caller by construction will never pass a provider where the data at 100% is null, will it?

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
```diff
@@ -1495,18 +1476,42 @@ void init(ImageData image, int imageZoom) {
...
+	init (data, 100);
+	StrictChecks.runIfStrictChecksEnabled(() -> {
+		DPIUtil.validateLinearScaling(imageDataProvider);
+	});
```
HeikoKlare: This can be kept in the existing caller, as the the new caller by construction fulfills the linear scaling constraint.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
```diff
@@ -1495,18 +1476,42 @@ void init(ImageData image, int imageZoom) {
...
 }
 
+private static ImageDataProvider createImageDataProvider(InputStream stream) throws IOException {
+	if (stream == null) {
+		SWT.error(SWT.ERROR_NULL_ARGUMENT);
+	}
+
+	byte[] streamData = stream.readAllBytes();
+	if (ImageDataLoader.isDynamicallySizable(new ByteArrayInputStream(streamData))) {
+		return (ImageDataAtSizeProvider) (width, height) ->
```
HeikoKlare: The cast should be obsolete.
```suggestion
		return (width, height) ->
```

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
```diff
@@ -1495,18 +1476,42 @@ void init(ImageData image, int imageZoom) {
...
+			ImageDataLoader.loadBySize(
+				new ByteArrayInputStream(streamData),
+				width,
+				height
+			);
+	}
+
+	ImageData imageData = new ImageData(new ByteArrayInputStream(streamData));
```
HeikoKlare: Why not use `ImagerLoader.loadByZoom()` anymore?

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
ShahzaibIbrahim: As per Javadocs contract yes, I can create a separate PR for that.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
ShahzaibIbrahim: True. New caller won't pass a provider with null data but in both cases we will need to extract image data for 100% for init call. If I don't do it here then either 1) I will have to pass the data as an argument or 2) I will have to call imageDataProvider.getImageData (100) first in the constructor to do the null check and again here for init. I just found this cleaner.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
ShahzaibIbrahim: The compiler doesn’t know which functional interface I meant. It will look for `ImageDataProvider` for which this width and height argument function doesn't exist.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
ShahzaibIbrahim: To be consistent with gtk implementation. There I used new ImageData(). Would this cause any problem?

HeikoKlare: You added that to the GTK implementation, didn't you? `ImageLoader` is basically the factory for `ImageData`. The `ImageData` constructors are basically copy constructors encapsulating a factory call, which does not really make sense to use.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
HeikoKlare: I see, then it makes sense like this.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
ShahzaibIbrahim: Separate PR: https://github.com/vi-eclipse/Eclipse-Platform/issues/584

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
HeikoKlare: I see. In my opinion it would be cleaner to assign this to a properly variable then, as I find this notation with a "casted" lambda hardly readable. Also please format this code properly with the auto-formatter.

bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
```diff
@@ -685,15 +684,14 @@ public Image(Device device, ImageData source, ImageData mask) {
  */
 public Image(Device device, InputStream stream) {
 	super(device);
+	NSAutoreleasePool pool = null;
```
HeikoKlare: Please don't unnecessary change the statement order here.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Image.java
```diff
@@ -802,6 +804,25 @@ private void initFromImageDataProvider(int zoom) {
 	init(resizedData);
 }
 
+private static ImageDataProvider createImageDataProvider(InputStream stream) throws IOException {
+	if (stream == null) {
+		SWT.error(SWT.ERROR_NULL_ARGUMENT);
+	}
```
HeikoKlare: For the sake of consistency and readability, this check should be done in caller, just like done in the Cocoa implementation.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Image.java
```diff
@@ -802,6 +808,20 @@ private void initFromImageDataProvider(int zoom) {
 	init(resizedData);
 }
 
+private static ImageDataProvider createImageDataProvider(InputStream stream) throws IOException {
+	byte[] streamData = stream.readAllBytes();
+	if (ImageDataLoader.isDynamicallySizable(new ByteArrayInputStream(streamData))) {
+		ImageDataAtSizeProvider imageDataAtSizeProvider = (width, height) -> ImageDataLoader
```
ptziegler, line 814: This doesn't work when loading the `ImageData` of an `Image` by zoom. Because the `getDefaultSize()` method is not overridden, a 1x1 `ImageData` is  returned at 100% zoom, otherwise `null`.

https://github.com/eclipse-platform/eclipse.platform.swt/blob/c6ec34ef45d989f3ed41cac15e1cc109376f919c/bundles/org.eclipse.swt/Eclipse%20SWT/common/org/eclipse/swt/graphics/ImageDataAtSizeProvider.java#L46-L55

HeikoKlare: Damn, I even though about that case when reviewing the PR and somehow didn't follow-up on it. Too bad that there was not even a test for it. Thank you for fixing that with #3012!

PR #2938, opened 2026-01-07
Split internal ImageHandle class into exposed API and internal methods

HeikoKlare:
The ImageHandle class currently contains some methods that are supposed to be exposed to other consumers (like GC) while others are only supposed to be used inside Image. To this end, different method visibilities are used. However, depending on where the ImageHandle is used inside Image, only specific functionality should be accessible as well. In particular, only the current lifecycle responsible for an ImageHandle should be able to destroy that handle.

To better ensure this, this change splits the ImageHandle class into multiple classes, such that only the intended functionality can be accessed at different consumers.

See https://github.com/eclipse-platform/eclipse.platform.swt/pull/2927#discussion_r2662299006


amartya4256 (approved):
The overall design looks much better with proper abstractions. As discussed, we could further improve it by making use of Interfaces but since there are many PRs urgently dependent on it, we should move forward with this and we can implement the usage of interfaces afterwards.

PR #2940, opened 2026-01-07
[GTK4] Fix CTabFolder chevron click GTK-Critical

akurtakov:
Error is like:
```
(SWT:339274): Gtk-CRITICAL **: 17:08:42.533: gtk_widget_compute_point:
assertion 'GTK_IS_WIDGET (widget)' failed
```

The problem comes from Control.getSurfaceOrigin method relying on fixedHandle to translate coordinated but not every widget has such, use eventHandle() instead.

Tested with Snippet165.


akurtakov:
This is kind of prereq for https://github.com/eclipse-platform/eclipse.platform.swt/pull/2895 as it fails for chevron button positioning.

PR #2941, opened 2026-01-07
[Win32] Encapsulate fields of ImageHandle

HeikoKlare:
The inner class `ImageHandle` currently exposes several fields, which are accessed and some even modified at different places. This change properly encapsulates those values and replaces the access with according accessors. It prepares for a further refactoring of the ImageHandle and its subclasses to a structure based on interface rather than class extensions.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
@@ -1236,7 +1236,7 @@ private void drawImage(Image image, int destX, int destY, int destWidth, int des
 	Rectangle destPixels = Win32DPIUtils.pointToPixel(drawable, new Rectangle(destX, destY, destWidth, destHeight),
 			imageZoom);
 	image.executeOnImageHandleAtBestFittingSize(tempHandle -> {
-		drawImage(image, 0, 0, tempHandle.getWidth(), tempHandle.getHeight(), destPixels.x, destPixels.y,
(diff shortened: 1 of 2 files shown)


laeubi:
@HeikoKlare a while back I made an experiment to make [font GC disposable](https://github.com/eclipse-platform/eclipse.platform.swt/pull/1054). Looking at the `ImageHandle` class it looks like it could be made GC disposable as well.

This will (as a last resort) allow the GC to reclaim unrefernced images automatically if not disposed directly, maybe an idea for future improvements. As far as I know SWT alread uses a cleaner for the warning when there are undisposed elements so that can be reused here.

HeikoKlare:
@laeubi thank you for the pointer. That should be possible, yes. You could for every resource implement an auto-disposal via GC, just like you proposed for the Font class. It's basically comparable to the existing `ResourceTracker`, which just prints a warning if the handle was not released, but you could of course also release the handle, just as you do in your PR.
On Windows, we now have multiple handles per resource (one for every required zoom). The `ImageHandle` class is a wrapper for one of such zoom-specific handles (as there is more handle-specific information for an image than for other resoures). Adding GC-auto-dispose functionality to that class would be possible, but is at the same time not reasonable as those handles are not referenced by consumers but are managed by the `Image` class itself. So in case such an auto-dispose would kick in, this would be an indicator for an implementation error in the `Image` class rather than show that a handle that is not required by consumers anymore.

amartya4256 (approved):
The changes look good. I have tested and confirm that I don't see any adverse effect as it should be since it is a straight-forward refactoring.

PR #2942, opened 2026-01-08
Add documentation for HiDPI support for Windows

akoch-yatta:
This PR adds a documentation for the HiDPI support for the win32 implementation. It is focusing on the perspective of a SWT developer not a SWT user. This is of course not covering everything as of now and perhaps not understandable for everyone, but as the complexity of this topic grew over time and a lot of knowledge is only in the heads of people we need to start somewhere. 

I would like to evolve that document over the next weeks. Feel free to put any questions remarks here, but I would like to not block merging it because of missing part and add those with further PRs

### DIFF
--- a/docs/hidpi-support-for-windows-dev-guide.md
+++ b/docs/hidpi-support-for-windows-dev-guide.md
@@ -1,0 +1,383 @@
+# SWT HiDPI Support in SWT for Windows
+
+## Table of Contents
+
+1. [Introduction](#introduction)
+2. [Design Principles](#design-principles)
+3. [Architecture Overview](#architecture-overview)
+4. [DPI Change Propagation](#dpi-change-propagation)
+5. [Resource Management](#resource-management)
+6. [Configuration of Autoscaling behavior](#configuration-of-autoscaling-behavior)
+7. [Additional Information and Challenges](#additional-information-and-challenges)
+
+---
+
+## Introduction
+
+HiDPI (High Dots Per Inch) support in SWT for Windows enables applications to function as expected on monitors with high pixel density.
+This documentation describes the behavior and implementation of HiDPI functionality in the Windows implementation of SWT.
+
+### What is HiDPI?
+
+HiDPI displays have a higher pixel density than conventional monitors.
+Windows supports various scaling levels (e.g., 125%, 150%, 175%, 200%) that allow users to make content readable on high-resolution displays.
+This UI scaling is of course not limited to HiDPI displays, but can be applied to conventional monitors as well.
+
+### History of HiDPI support
+
+The HiDPI support in SWT for Windows evolved in two phases:
+
+The first phase can be considered as *static* HiDPI support and was targeted on rendering a sharp UI for primary monitor zoom at startup.
+Changing the zoom at runtime was not properly support.
+Technically it focused on the SWT autoscaling mode *integer200* and a *System* DPI awareness for the OS.
+*integer200* results in the native zoom of amonitor being converted internally either to 100% scaling on native zoom < 175% and 200% otherwise.
+This means Widget will behave e.g. on a 150% the same as on a 100%, with the expections of Fonts.
+*System* DPI awareness means the native monitor zoom is only considered for the primary monitor on startup.
+If the application is moved to another monitor with another zoom, the OS will scale the application itself - usually resulting in a blurry UI.
+
+The second phase can be considered as *dynamic* HiDPI support and was targeted on rendering a sharp UI for every monitor zoom  while adapting automatically, e.g. if a Shell is moved to a monitor with different zoom.
+The existing implementation was extended and focused on the autoscaling mode *quarter* and a *PerMonitorV2* DPI awareness for the OS.
+*quarter* results in all elements using a value rounded down to a multiple of 25% internally in SWT, e.g. 125% will be used as 125%, a custom zoom of 130% would be rounded down to 125%.
+This decision was made as the proposed zooms of Windows itself are proposed as *quarter*.
+*PerMonitorV2* DPI awareness means the native monitor zoom should be considered for each Window (=Shell) separately.
+If a Shell is moved to another monitor with another zoom, the Shell and its contents rescale themselves to the new context.
+This is particularly important in multi-monitor environments where different monitors can have different DPI settings.
+
+---
+
+## Design Principles
+
+The introduction and extension of HiDPI support in SWT for Windows required structural changes designed to remain fully backward compatible.
+All adaptations follow these principles:
+
+### Backwards Compatibility
+
+Existing behavior should not worsen.
+For example, the different autoscale modes ("integer", "quarter", "false" or with fixed value) should still work as previously.
+One example for this is that font sizes were always based on the monitor zoom of the Shell, not on the autoscale mode.
+This includes of course the implementations for Linux and MacOS of SWT.
+Although proper HiDPI behavior in Windows requires more effort and complexity on SWT side than on Linux and MacOS, it must not have a negative side effect on the other implementation.
+
+
+### Per-Monitor DPI Awareness
+
+Each `Shell` maintains its own zoom.
+Widgets inherit this zoom on creation and adjust when the Shell's DPI changes.
+
+### Event-Driven Zoom Propagation
+
+DPI changes are propagated through the widget hierarchy via the standard SWT event system (`SWT.ZoomChanged`) to make it available for consumers as well.
+
+### Consistent Resource Scaling
+
+Images, cursors, and geometric resources maintain multiple OS handles internally for different zooms.
+Fonts are managed in a similar way via the `SWTFontProvider`.
+
+---
+
+## Architecture Overview
+
+### Per-Shell Zoom State
+
+The *static* HiDPI support for SWT operated with a single system-wide zoom factor, that was accessible via `DPIUtil.getNativeDeviceZoom()` for the native zoom and `DPIUtil.getDeviceZoom()` for the autoscaled zoom.
+While both methods can be used as fallback in certain cases, it is usually not recommended to use those utility function, as they might to fit the zoom of the current SWT `Shell`.
+
+The *dynamic* HiDPI support introduced a per-Shell `nativeZoom`, which:
+
+- is initialized when the Shell is created (from the current monitor),
+- is inherited by all child widgets,
+- updates when a DPI change event is received from the OS.
+
+Each widget stores its own copy of the zoom to avoid repeated Shell lookups.
+
+#### Widget Initialization
+
+The `nativeZoom` field is defined in the base `Widget` class as a public (but non-API) field:
+
+```java
+public int nativeZoom;
+```
+
+It should only be accessed from win32 specific components of SWT itself.
+
+**Standard Widgets** (non-Shell widgets):
+
+During widget construction all necessary configuration is done:
+
+```java
+public Widget(Widget parent, int style) {
+    this.nativeZoom = parent != null ? parent.nativeZoom : DPIUtil.getNativeDeviceZoom();
+    // ...
+    registerDPIChangeListener();
+}
+```
+
+- If a parent exists, the widget inherits the parent's `nativeZoom`.
+- For top-level widgets (no parent), `nativeZoom` is initialized from the current device zoom via `DPIUtil.getNativeDeviceZoom()`.
+This is considered a fallback and should never happen in practice, as top-level widgets are `Shells` and are covered in the next section.
+- All widgets automatically register a DPI change listener during construction.
+This listener will update the state of each `Widget` to fit the new zoom when a DPI change occurs.
+
+**Shell Initialization**:
+
+Shell widgets as top level widgets use a different initialization sequence:
+
+```java
+Shell(Display display, Shell parent, int style, long handle, boolean embedded) {
+    super();  // Calls parameterless Widget constructor
+    // ...
+initialization code ...
+    createWidget();  // Creates the native window handle
+    this.nativeZoom = DPIUtil.mapDPIToZoom(OS.GetDpiForWindow(this.handle));
+    registerDPIChangeListener();  // Explicitly called since Widget(parent, style) wasn't called
+}
+```
+
+Key differences for Shell:
+
+- Shell calls `super()` (parameterless constructor) instead of `super(parent, style)`.
+- `createWidget()` must be called first to create the native window handle (`this.handle`).
+- `nativeZoom` is initialized **after** `createWidget()` using `DPIUtil.mapDPIToZoom(OS.GetDpiForWindow(this.handle))` to convert the DPI value from the window handle to a zoom percentage.
+This approach ensures independence from internal SWT configuration and uses the actual native zoom retrieved directly from the OS for the monitor where the window is created.
+- `registerDPIChangeListener()` is explicitly called because the Widget constructor that normally registers it wasn't invoked.
+
+This ensures that Shell's `nativeZoom` reflects the actual DPI of the monitor where the window is created, which is essential for Per-Monitor DPI Awareness.
+
+### Event-Driven DPI Propagation (`SWT.ZoomChanged`)
+
+SWT propagates zoom changes through the standard event-based model:
+
+1. Windows emits a `WM_DPICHANGED` message.
+2. The Shell computes:
+   - `newZoom` (percentage),
+   - `scalingFactor = newZoom / oldZoom`.
+3. SWT sends a `SWT.ZoomChanged` event to the affected Shell.
+4. The event is propagated through the widget tree according to layout and dispatching rules.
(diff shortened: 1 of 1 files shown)


BeckerWdf:
Could this be something that should go into the org.eclipse.platform.doc.isv documentation?

For example into something like this: https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/eclipse.platform.common/bundles/org.eclipse.platform.doc.isv/guide/swt.htm

docs/HiDPI Support for Windows.md
```diff
@@ -0,0 +1,326 @@
...
+HiDPI (High Dots Per Inch) support in SWT for Windows enables applications to function as expected on monitors with high pixel density. This documentation describes the behavior and implementation of HiDPI functionality in the Windows implementation of SWT.
+
+### What is HiDPI?
+
+HiDPI displays have a higher pixel density than conventional monitors. Windows supports various scaling levels (e.g., 125%, 150%, 175%, 200%) that allow users to make content readable on high-resolution displays.
+
+### History of HiDPI support
+
+The HiDPI support in SWT for Windows evolved in two phases. 
+In the first phase, the support was added with the focus on the autoscaling mode *integer200* and a *System* DPI awareness for the OS. *integer200* results in all elements being treated as 100% scaling on OS scaling < 175% and 200% otherwise internally in SWT. *System* DPI awareness means the native monitor zoom is only considered for the primary monitor on startup. If the application is moved to another monitor with another zoom, the OS will scale the application itself - usually resulting in a blurry UI.
```
HeikoKlare: I do not understand the sentence "*integer200* results in all elements being treated as 100% scaling on OS scaling < 175% and 200% otherwise internally in SWT."
Can you rephrase/correct that?

docs/HiDPI Support for Windows.md
```diff
@@ -0,0 +1,326 @@
...
+The second phase extended the existing implementation by putting the focus on the autoscaling mode *quarter* and a *PerMonitorV2* DPI awareness for the OS. *quarter* results in all elements using a value rounded down to a multiple of 25% internally in SWT, e.g. 125% will be used as 125%, a custom zoom of 130% would be rounded down to 125%. This decision was made as the proposed zooms of Windows itself are proposed as *quarter*. *PerMonitorV2* DPI awareness means the native monitor zoom should be considered for each Window (=Shell) separately. If a Shell is moved to another monitor with another zoom, the Shell and its contents rescale themselves to the new context. This is particularly important in multi-monitor environments where different monitors can have different DPI settings.
+
+---
+
+## Design Principles
+
+The introduction of dynamic HiDPI support in SWT for Windows required structural changes designed to remain fully backward compatible. All adaptations follow these principles:
```
HeikoKlare: The term "dynamic" is used for the first time here and probably not comprehensive for everyone. Maybe you could adapt the above explained phases to rather point "static" HiDPI support (sharp for primary monitor zoom at startup) and "dynamic" HiDPI support (sharp for every monitor a shell is moved to).

docs/HiDPI Support for Windows.md
```diff
@@ -0,0 +1,326 @@
...
+
+### Backwards Compatibility
+
+Existing behavior should not worsen. For example, the different autoscale modes ("integer200", "quarter", "false" or with fixed value) should still work as previously. One example for this is that font sizes were always based on the monitor zoom of the Shell, not on the autoscale mode.
```
HeikoKlare: We could also mention compatibility to other OSes here, as it should be clear that adapting to Windows monitor-specific scaling does not mean you have to drop/degrade support for Linux/MacOS but it rather means that you may have to done few things on top that will not break anything on Linux/MacOS.

docs/HiDPI Support for Windows.md
```diff
@@ -0,0 +1,326 @@
...
+
+Images, cursors, and geometric resources maintain multiple OS handles internally for different zoom levels. Fonts are managed in a similar way via the `SWTFontProvider`.
+
+---
+
+## Architecture Overview
+
+### Per-Shell Zoom State
+
+After the first HiDPI adaption for SWT, it operated with a single system-wide zoom factor. Now, with the further adaptions HiDPI support introduces a per-Shell `nativeZoom`, which:
```
HeikoKlare: "Time-based" statement such as "now" will easily outdated. I propose to refer the different development state of HiDPI support here, as proposed above the initial "static" concept with a single zoom factor and the "dynamic" concept with a per-shell zoom.

docs/hidpi-support-for-windows-dev-guide.md
```diff
@@ -0,0 +1,326 @@
...
+
+### Per-Shell Zoom State
+
+After the first HiDPI adaption for SWT, it operated with a single system-wide zoom factor. Now, with the further adaptions HiDPI support introduces a per-Shell `nativeZoom`, which:
+
+- is initialized when the Shell is created (from the current monitor),
+- is inherited by all child widgets,
+- updates when a DPI change event is received from the OS.
+
+Each widget stores its own copy of the zoom to avoid repeated Shell lookups.
```
HeikoKlare, line 91: Maybe this could be a good point to mention that the static zooms provided as `DPIUtil.getDeviceZoom()` and `DPIUtil.getNativeDeviceZoom()` became inappropriate with the introduction of the dynamic mode?

docs/HiDPI Support for Windows.md
```diff
@@ -0,0 +1,326 @@
...
+
+```java
+@Override
+void handleDPIChange(Event event, float scalingFactor) {
+    super.handleDPIChange(event, scalingFactor);
+    // Widget-specific DPI change handling
+}
+```
+
+Each class contributes only the logic relevant to the state it introduces. Because listeners are registered per class, inheritance naturally results in hierarchical, layered updates.
```
HeikoKlare: I think it would be good to mention that the `handleDPIChange` method with its overwrites is used throughout internal widgets to avoid that too many handlers (one for each class) are registered. One could also register one listener per sublass and that's also what you need to do when implementing custom widgets. Otherwise, one might wonder how to overwrite that `handleDPIChange` method as it's package protected.

docs/HiDPI Support for Windows.md
```diff
@@ -0,0 +1,326 @@
...
+4. The event is propagated through the widget tree according to layout and dispatching rules.
+5. Each widget updates its internal DPI-dependent state via its registered callback.
+
+The `scalingFactor` is only used internally and passed to the event handling methods `handleDPIChange(Event event, float scalingFactor)` in each widget. It represents the relative change between the old and new zoom and is used to rescale pixel values (for example the column width of a `Table`) that cannot simply be recomputed from the absolute zoom alone.
+
+More details about the event propagation in [DPI Change Propagation](#dpi-change-propagation).
+
+### Class-Level Callback Registration
+
+Every widget class with DPI-dependent state registers a **private** listener for the `SWT.ZoomChanged` event. All widgets automatically register this listener during construction via `registerDPIChangeListener()`.
```
HeikoKlare: Do you mean widget "object" instead of widget "class"? This sounds as if every subclass would register such as listener, but that is not the case.

docs/HiDPI Support for Windows.md
```diff
@@ -0,0 +1,326 @@
...
+
+### Autoscale system properties
+
+#### Autoscale Modes
+
+The `swt.autoScale` system property controls how native DPI is converted to effective zoom:
+
+- **`false`**: Always returns 100% zoom.
+- **`integer`**: Rounds to integer multiples of 100% (e.g., 150% → 100%, 175% → 200%).
+- **`integer200`**: Like `integer`, but caps at 200%.
```
HeikoKlare: If I am not mistaken, `integer200` is effectively not in place anymore for quite a while now. And actually the `integer` mode is the default hat is used as a fallback (e.g., if you specify an undefined mode such as `true`).

docs/HiDPI Support for Windows.md
```diff
@@ -0,0 +1,326 @@
...
+- **`half`**: Rounds to multiples of 50% (e.g., 125% → 100% or 150%).
+- **`quarter`**: Rounds to multiples of 25% (e.g., 125%, 150%, 175%).
+- **`exact`**: Uses native zoom with 1% precision.
+- **`<value>`**: Fixed zoom percentage (e.g., `"150"` = 150%).
+
+#### Runtime Rescaling
```
HeikoKlare: "Runtime Resaling" is just another new name for what used to be called "second phase" and "dynamic". I propose to stick to a single uniform name. Unfortunately, the system property name is inconsistent to that, but we cannot change that anymore.

docs/HiDPI Support for Windows.md
```diff
@@ -0,0 +1,326 @@
...
+
+The `swt.autoScale.updateOnRuntime` system property controls dynamic DPI changes:
+
+- **`true`**: Application rescales when DPI changes (e.g., window moved to different monitor).
+- **`false`**: Application remains at initial scaling.
+- **`force`**: Forces runtime rescaling even with incompatible autoScale modes.
+
+**Note**: Runtime rescaling requires compatible autoScale modes (`quarter`, `exact`). Modes like `integer`/`integer200` or fixed values are not recommended with monitor-specific scaling.
```
HeikoKlare: It's not only that the modes are not recommended, but they are actually incompatible and will make Display initialization fail (unless you use `force` mode). Without that explanation the `force` would probably be incomprehensible.

docs/HiDPI Support for Windows.md
```diff
@@ -0,0 +1,326 @@
...
+
+
+---
+
+## Known Limitations and Challenges
```
HeikoKlare: I am not sure if we should call this "known limitations". I was expecting in particular the first subsections to contain actual limitations and what then wondering how far event odering, performance etc. are limited. Maybe we could rename this to "Additional Information and Challenges" or the like?

docs/HiDPI Support for Windows.md
```diff
@@ -0,0 +1,326 @@
...
+
+---
+
+## Introduction
+
+HiDPI (High Dots Per Inch) support in SWT for Windows enables applications to function as expected on monitors with high pixel density. This documentation describes the behavior and implementation of HiDPI functionality in the Windows implementation of SWT.
+
+### What is HiDPI?
+
+HiDPI displays have a higher pixel density than conventional monitors. Windows supports various scaling levels (e.g., 125%, 150%, 175%, 200%) that allow users to make content readable on high-resolution displays.
```
HeikoKlare: I would also mention that you can even use different zoom values when using a convential monitor to just increase UI zoom. And even though we often talk about "HiDPI support", it's actually just about scaling support, no matter if the display may actually be considered HiDPI or not. Maybe we can add a statement to clarify that here?

docs/HiDPI Support for Windows.md
akoch-yatta: I'll rephrase that

docs/HiDPI Support for Windows.md
akoch-yatta: Adapted the upper paragraph to introduce the terms static and dynamic

merks:
Please start each sentence on a new line.

docs/HiDPI Support for Windows.md
akoch-yatta: Oh yes, that is correct, that was a mix-up

akoch-yatta:
@HeikoKlare I tried to address all your remarks and rephrased/changed those. You can see the changes here: https://github.com/eclipse-platform/eclipse.platform.swt/compare/7ee146823957b49df3b9cf91a2dd0f17edd8f28f..3882e9197f9a8e4dd633a825d4c32f77155761a3

merks:
BTW, I didn't even know SWT had a docs folder.  It seems new from just a few months ago.   I'm not sure how we expect folks to find those docs.  From this page:

https://eclipse.dev/eclipse/

Should we add it here?

image

Documentation that hard to find is not so useful...

akoch-yatta:
> Please start each sentence on a new line.

I will do that now. To be honest, I was not aware of this Markdown behavior and my Markdown Preview editor as well as it did render a line break for a line break. So my intended line breaks are not even real lines break, so I will go through the document again.

merks:
The rendering inconsistencies is actually kind of annoying.  You can see that this page:

https://eclipse.dev/simrel/?file=eclipse-packaging/packages/master/RELEASING.md

and this page

https://github.com/eclipse-packaging/packages/blob/master/RELEASING.md

render the same way. 

See this for the raw view:

https://github.com/eclipse-packaging/packages/blob/master/RELEASING.md?plain=1

But, if I copy and paste that text into an issue, it renders line breaks differently  😱 

---

One reason I ask about the website is that the oomph setup configures a launcher for "local host":

image

And that gives you a live rendered view of your local contents just like it will look when you commit the changes:

image

That's especially helpful if you have a set of documents and not just a single document.  I have yet to find an editor that renders consistently nicely...

https://eclipse.dev/eclipse/markdown/?f=news/instructions.md#markdown-preview

akoch-yatta:
> BTW, I didn't even know SWT had a docs folder. It seems new from just a few months ago. I'm not sure how we expect folks to find those docs. From this page:
> 
> https://eclipse.dev/eclipse/
> 
> Should we add it here?
> 
> image
> Documentation that hard to find is not so useful...

I don't mind where to put it, although I personally would first have a look into the repo itself for documentation. But moving the document itself to another place would be fine for me, especially with the argument of having one place to collect everything there. I which repo would I need to put it?

merks:
> I which repo would I need to put it?

It's all really a bit of a messy dumping ground unfortunately.  For the "docs" folders in some of the repositories there isn't really a good index so I've really just done the bare minimum (e.g., a folder listing) to make the content somehow reachable from the website.  

I actually don't have any great advice to offer.  One could spend (many) months improving the website(s), or the markdown documents in the github repositories, but then then one has to wonder, do people even read anything today?  Maybe just the first paragraph of an AI summary? 

In the end, I don't think anyone really feels responsible for the state of the website.  Certainly while working on it there was minimal reaction and minimal feedback...

HeikoKlare:
> BTW, I didn't even know SWT had a docs folder. It seems new from just a few months ago. I'm not sure how we expect folks to find those docs. From this page:
> 
> https://eclipse.dev/eclipse/
> 
> Should we add it here?
> 
> image
> Documentation that hard to find is not so useful...

We should definitely do that. I have created a PR to make a proper project out of the docs folder, just like it is for Platform and Platform UI, so that you have it in your workspace and can easily modify it:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2948

We then need to embed the markdown in the website. And I think we should even migrate the SWT stuff from the website repo to the SWT repo: https://github.com/eclipse-platform/www.eclipse.org-eclipse/tree/master/swt
It does not really make sense that there is a differently maintained SWT page on the Eclipse website, with the contents being maintained at an uncommon place (the Website repo instead of the code repo like for other docs) and with a different semantics than the other links (Equinox and Platform link to the Eclipse projects while SWT links to the subpage):
image
Migrating the SWT website contents to the SWT repo will also relieve us from adding snippets/snippet references in two different repos (SWT and Website).

BeckerWdf:
> I think we agree that the documentation should be properly visible, which is what we achieve by having it added to the website next to other devloper information. Since this documentation is rather relevant for SWT developers than SWT consumers, I think that the better place than the ` org.eclipse.platform.doc.isv`. If you disagree, let me know.

That's ok with me.

vogella:
I suggest to avoid spaces in file names as this makes linking harder. Use - or _ instead.

docs/hidpi-support-for-windows-dev-guide.md
```diff
@@ -0,0 +1,379 @@
...
+
+The `swt.autoScale.updateOnRuntime` system property controls dynamic DPI changes:
+
+- **`true`**: Application rescales when DPI changes (e.g., window moved to different monitor).
+- **`false`**: Application remains at initial scaling.
+- **`force`**: Forces runtime rescaling even with incompatible autoscale modes.
+
+**Note**: Runtime rescaling requires compatible autoScale modes (`quarter`, `exact`).
+Modes like `integer` or fixed values are incompatible with monitor-specific scaling and will only be accepted when `force` is used.
+
```
vogella, line 332: Add example usage:

vogella:
@HeikoKlare thanks for adding this, that is super useful information.

merks:
> I suggest to avoid spaces in file names as this makes linking harder. Use - or _ instead.

+10000

akoch-yatta:
> I suggest to avoid spaces in file names as this makes linking harder. Use - or _ instead.

Very good point, I renamed the file

docs/hidpi-support-for-windows-dev-guide.md
akoch-yatta: I added a short example usage to both system properties and marked the default values

HeikoKlare:
@merks you still have pending a change request, can we resolve that? 🙂

PR #2943, opened 2026-01-08
[Win32] Replace ImageHandle class hierarchy with interfaces

HeikoKlare:
To ensure proper restricted access to specific functionality of Image's internal ImageHandle class, a hierarchy of classes with different visibilities is currently used. To better express that there is just a single type to instantiate and that access restriction is supposed to be done by only providing specific methods to specific consumers, this change replaces the existing class hierarchy with a single implementation class and a conversion of the existing other classes with interfaces. It also further restricts access to several methods of ImageHandle (zoom(), bounds(), isDisposed()) that are not required to be public.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
@@ -201,7 +201,7 @@ public final class Image extends Resource implements Drawable {
 	}
 
 	private class HandleAtSize {
-		private ImageHandle handleContainer = null;
+		private InternalImageHandle handleContainer = null;
 		private DestroyableImageHandle temporaryHandleContainer = null;
 		private int requestedWidth = -1;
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2863,57 +2863,84 @@ public void internal_dispose_GC(long handle, GCData data) {
...
+interface ImageHandle {
+	long handle();
+	int width();
+	int height();
+	int transparentPixel();
+	int transparentColor();
+	void setTransparentColor(int transparentColor);
+}
+
+interface InternalImageHandle extends ImageHandle{
```
amartya4256: we can make it private.

HeikoKlare: Indeed, changed with the latest commit.

PR #2945, opened 2026-01-09
Remove integer200 from DPIUtil javadoc

akoch-yatta:
This PR removes all mentions of integer200 from the javadoc and comments in DPIUtil. The described behavior of integer200 was removed by commit e59aa02 in 2022. Using integer200 will do not cause any specific behavior but will fall back to the default mode integer, which is provided the same behavior expect not capping at 200%.


HeikoKlare (approved):
The documentation is obviously inconsistent with the implementation and lacks an update, which is properly done with this change. Equinox consistency due to the mentioned SWT change from several years ago is to be fixed via:
- https://github.com/eclipse-equinox/equinox/pull/1213

PR #2946, opened 2026-01-09
[Win32] Process DPI change events via OS instead of Display#asyncExec

HeikoKlare:
Asynchronous processing of DPI change events is currently done by processing the event for each control via Display#asyncExec. This has several drawbacks:
- the events are processed rather late (even after earlier scheduled, lower-priority business events
- the events may not be processed while dragging a shell to another monitor in case a browser currently has focus
- the events are not processed at all if a Synchronizer is used that does not process async events (such as for an Eclipse workspace during startup)

<s>This change replaces the usage of Display#asyncExec for asynchronous processing of DPI events with a timer message sent via the OS and a queue maintained in every control for the events to process.</s>
This change replaces the usage of Display#asyncExec for asynchronous processing of DPI events with the a timer message beings sent with a callback to the processed operation via the OS.
WM_TIMER events are used as they are processed with low priority by the OS but still with higher priority by SWT than any business-logic-side tasks scheduled at a Display instance.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
@@ -15,6 +15,7 @@
 package org.eclipse.swt.widgets;
 
 
+import java.lang.reflect.*;
 import java.util.*;
 import java.util.concurrent.atomic.*;
 import java.util.stream.*;
@@ -5992,7 +5993,7 @@ static class DPIChangeExecution {
 			asyncExec &= (comp.layout != null);
 		}
 		if (asyncExec) {
-			control.getDisplay().asyncExec(operation::run);
+			DPIChangeProcessingCallback.schedule(control, operation);
 		} else {
 			operation.run();
 		}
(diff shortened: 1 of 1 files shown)


HeikoKlare:
> creates a timer with a callback pointing to the DPIExecution runnable which is then checked when WM_TIMER is received

@amartya4256 it does not create a timer with a callback but just with the hash code of the acutal operation stored in a list as a parameter that is then used to extract the actual operation from the list when the `WM_TIMER` event is processed.
Having an actual callback would be even better (and might be possible via [a timerproc function](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nc-winuser-timerproc)), but we have not implemented anything comparable so far. We might try to do so. That would definitely be cleaner and more failsafe as the hash code could eventually collide.

> A small question about if the order of execution is always guaranteed as with asnycExec, it was guaranteed. And if it not then throwing the exception still introduces regression.

I expect that the OS processes messages in the exact order in which they were posted, otherwise other places would probably break as well. The exception is rather there to find out that/if the OS does not behave like that.

HeikoKlare:
> That would definitely be cleaner and more failsafe as the hash code could eventually collide.

To avoid the risk of collisions, I introduced a base ID for the timer (higher than the integer range) and add the hashcode of the operation (an integer value) on top. This range of IDs is not used by any timer, so collisions with other timers are avoided.

HeikoKlare:
With the latest change, I only added a missing `KillTimer` for the callbacks. I expected the time to automatically be killed after being triggered, but turned out when testing the change with our product and a large number of controls shown the application ran out of user interface objects. With the `KillTimer` added, the number of user interface objects stays constants. I tested with heavily moving around shells with many complex controls and did not find any issues there either. The responsiveness is much better than with the existing implementation based on `Display#asyncExec()`.

PR #2949, opened 2026-01-12
[win32] Pass zoom to Scrollable#computeTrimInPixels

akoch-yatta:
The PR unifies the behavior of Scrollable#computeTrimInPixels to computeBoundsInPixels. As computeBoundsInPixels was extended to always require a zoom to calculate for, the same should apply to computeTrimInPixels as it is usually called from computeBoundsInPixels and must be calculate with the same zoom as context.


akoch-yatta:
This PR is no longer needed with https://github.com/eclipse-platform/eclipse.platform.swt/pull/2950 and https://github.com/eclipse-platform/eclipse.platform.swt/pull/2951

PR #2950, opened 2026-01-12
[win32] Use correct zoom in Control#getSystemMetrics

akoch-yatta:
This PR overrides Widget#getSystemMetrics for Control to utilize Shell#getZoom if possible to define the target zoom. This is necessary if the control has its autoscaling disabled.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/ScrollBar.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/ScrollBar.java
@@ -999,4 +999,12 @@ LRESULT wmScrollChild (long wParam, long lParam) {
 	return null;
 }
 
+@Override
+int getSystemMetrics(int nIndex) {
+	// Control#getSystemMetrics should be used if possible,
+	// as it considers if autoscaling of a Control is
+	// disabled which would affect the ScrollBar as well,
+	// therefore the value is fetched via the parent
+	return parent.getSystemMetrics(nIndex);
+}
 }
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
@@ -6097,6 +6097,13 @@ void handleDPIChange(Event event, float scalingFactor) {
 	}
 }
 
+@Override
+int getSystemMetrics(int nIndex) {
+	Shell shell = getShell();
+	int zoom = shell != null ? shell.getZoom() : nativeZoom;
+	return OS.GetSystemMetricsForDpi(nIndex, DPIUtil.mapZoomToDPI(zoom));
+}
+
 boolean adjustWindowRectEx(RECT lpRect, int dwStyle, boolean bMenu, int dwExStyle) {
 	Shell shell = getShell();
 	int zoom = shell != null ? shell.getZoom() : nativeZoom;
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/TreeColumn.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/TreeColumn.java
@@ -762,4 +762,13 @@ void handleDPIChange(Event event, float scalingFactor) {
 		setImage(image);
 	}
 }
+
+@Override
+int getSystemMetrics(int nIndex) {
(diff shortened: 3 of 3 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Widget.java
```diff
@@ -2708,7 +2708,11 @@ void handleDPIChange(Event event, float scalingFactor) {
 }
 
 int getSystemMetrics(int nIndex) {
-	return OS.GetSystemMetricsForDpi(nIndex, DPIUtil.mapZoomToDPI(nativeZoom));
+	return getSystemMetrics(nIndex, nativeZoom);
+}
+
+int getSystemMetrics(int nIndex, int zoom) {
+	return OS.GetSystemMetricsForDpi(nIndex, DPIUtil.mapZoomToDPI(zoom));
```
HeikoKlare: Is this change in `Widget` necessary at all? Can't that class stay as it is?

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/ScrollBar.java
```diff
@@ -199,10 +199,10 @@ Rectangle getBounds () {
 	if ((style & SWT.HORIZONTAL) != 0) {
 		y = rect.bottom - rect.top;
 		width = rect.right - rect.left;
-		height = getSystemMetrics (OS.SM_CYHSCROLL);
+		height = parent.getSystemMetrics (OS.SM_CYHSCROLL);
```
HeikoKlare: Maybe it would be easier to understand if `ScrollBar` and `TreeColumn` would overwrite `getSystemMetrics()` with a delegation to `parent.getSystemMetrics()`, so that one can also place a short comment in that method explaining why we use the parent's method and not the inherited one.

HeikoKlare (commented):
This looks like the right thing to do. I just have minor comments regarding potential improvements. And of course the PR title and description needs to be adapted to latest changes.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Widget.java
akoch-yatta: I removed that and used the OS call in Control#getSystemMetrics directly

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/ScrollBar.java
akoch-yatta: I adapted it acordingly

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/ScrollBar.java
```diff
@@ -999,4 +999,12 @@ LRESULT wmScrollChild (long wParam, long lParam) {
 	return null;
 }
 
+@Override
+int getSystemMetrics(int nIndex) {
+	// Control#getSystemMetrics should be used if possible,
+	// as it works better if autoscaling of a Control is
+	// enabled which would affect the ScrollBar as well,
```
HeikoKlare:
```suggestion
	// as it considers if autoscaling of a Control is
	// disabled which would affect the ScrollBar as well,
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/ScrollBar.java
```diff
@@ -999,4 +999,12 @@ LRESULT wmScrollChild (long wParam, long lParam) {
...
+	// therefore the value if fetched via the parent
```
HeikoKlare:
```suggestion
	// therefore the value is fetched via the parent
```

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/ScrollBar.java
akoch-yatta: adapted it in TreeColumn as well

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/ScrollBar.java
akoch-yatta: adapted it in TreeColumn as well

PR #2954, opened 2026-01-12
[MacOS] Correct faulty representation creation on image copy

HeikoKlare:
When copying an image based on some provider (file name, data, gcdrawer), a 200% representation is supposed to be copied if it exists. The current implementation, however, unintentionally creates such a representation if it does not exist yet.

This change corrects the implementation to not create a representation when copying an image.


akoch-yatta (approved):
LGTM, but I cannot test it on Mac. The split up of the code looks correct and all calls except the one that was mentioned was replaced with the getOrCreateRepresentation call

PR #2955, opened 2026-01-12
[Win32] Enable monitor-specific scaling by default

HeikoKlare:
By now, SWT had monitor-specific scaling on Windows disabled by default and used the existing, single-zoom HiDPI support. There are good reasons to now change the default of SWT to have monitor-specific scaling enabled:
- Eclipse applications default to have monitor-specific scaling enabled for several months now, so SWT and Eclipse default behaviors are different
- With default JVM settings, pure SWT application have no proper scaling as the existing HiDPI support requires DPI awareness "System" to function properly whereas the JVM default is "PerMonitorV2" and requires the application to scale on its own. So by default, SWT applications currently have no scaling at all
- The feature has matured throughout the last year with it being the default for Eclipse application for several months now.

This change adapts the SWT default to enable monitor-specific scaling on Windows.

### How to test

Just run any SWT snippet and move it between monitors of different zooms. It will adapt it's scale while it did not do prior to this PR.

### Feedback welcome

I would appreciate feedback on whether we agree on changing this default. I propose to merge this before M2 so that people get aware it early enough. When merging, I will also create an according N&N.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/DPIUtil.java
@@ -125,6 +125,10 @@ public static boolean isCustomAutoScale() {
 	return System.getProperty (SWT_AUTOSCALE) != null;
 }
 
+public static String getEffectiveAutoScaleValue() {
+	return autoScaleValue.toString();
+}
+
 /**
  * Returns {@code true} only if the current setup is compatible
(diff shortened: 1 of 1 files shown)


merks:
I'll be happy with a new default because the installer won't need a -D option then.  I was testing that the other day and it's much nicer with the new behavior.

vogella:
Of course +1, defaults should be using the best functionality and it's unlikely that a user "only wants his menu (done automatically by the os)" scaled.

vogella:
Thanks! Awesome!

PR #2960, opened 2026-01-13
Wrong SWT Error in Image(Device, InputStream) for IOException

ShahzaibIbrahim:
As per contract of org.eclipse.swt.graphics.Image.Image(Device, InputStream), SWT Error - ERROR_IO should be thrown if IOException is caught while reading the inputStream.


HeikoKlare (approved):
This obviously adapts the implementation to the existing contract. Great to also see a test for the behavior.
It seems like the behavior used to be correct but was accidentally modified for Cocoa with: 
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/1987

This change corrects that.

PR #2964, opened 2026-01-14
Fix wrong usage of getImageData and Permanent ImageHandle creation

amartya4256:
There are many places in several ImageProviders where we want to create a new ImageHandle or ImageData and use getImageData or create permanent ImageHandle to obtain just ImageData respectively. This PR replaces such usage with directly calling newImageData instead of getImageData and creating a temporary ImageHandle when calling newImageData.

Needed for https://github.com/eclipse-platform/eclipse.platform.swt/pull/2905

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
@@ -1388,6 +1388,9 @@ public ImageData getImageData (int zoom) {
 		return imageHandle.getImageData();
 	}
 
+	if (this.imageProvider.isPersistentImageHandleRequriedForImageData()) {
+		return imageHandleManager.getOrCreate(zoom, () -> imageProvider.newImageHandle(new ZoomContext(zoom))).getImageData();
+	}
 	return this.imageProvider.newImageData(zoom);
 }
 
@@ -2055,6 +2058,10 @@ private abstract class AbstractImageProviderWrapper {
 		return Collections.emptySet();
 	}
 
+	protected boolean isPersistentImageHandleRequriedForImageData() {
+		return false;
+	}
+
 	protected abstract ElementAtZoom<ImageData> loadImageData(int zoom);
 
 	abstract ImageData newImageData(int zoom);
@@ -2076,10 +2083,7 @@ private abstract class AbstractImageProviderWrapper {
 		return Optional.empty(); // exact size not available
 	}
 
-	protected DestroyableImageHandle newImageHandle(ZoomContext zoomContext) {
-		ImageData resizedData = getImageData (zoomContext.targetZoom());
-		return newImageHandle(resizedData, zoomContext);
-	}
+	abstract DestroyableImageHandle newImageHandle(ZoomContext zoomContext);
 
 	protected final DestroyableImageHandle newImageHandle(ImageData data, ZoomContext zoomContext) {
 		if (type == SWT.ICON && data.getTransparencyType() != SWT.TRANSPARENCY_MASK) {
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2077,7 +2077,7 @@ protected Optional<ImageData> loadImageDataAtExactSize(int width, int height) {
 	}
 
 	protected DestroyableImageHandle newImageHandle(ZoomContext zoomContext) {
-		ImageData resizedData = getImageData (zoomContext.targetZoom());
+		ImageData resizedData = newImageData (zoomContext.targetZoom());
 		return newImageHandle(resizedData, zoomContext);
 	}
```
HeikoKlare: This method is implemented by all but one subclass anyway, so I would propose to make the method abstract and always have it implemented in the subclass to easy understanding of what's going on there.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2322,12 +2322,18 @@ protected Rectangle getBounds(int zoom) {
 	@Override
 	ImageData newImageData(int zoom) {
 		if (imageHandleManager.isEmpty()) {
-			return imageHandleManager.getOrCreate(zoom, () -> createBaseHandle(zoom)).getImageData();
+			DestroyableImageHandle imageHandle = createBaseHandle(zoom);
+			ImageData imageData = imageHandle.getImageData();
+			imageHandle.destroy();
+			return imageData;
```
HeikoKlare: This change is not correct (at least semantically), as `createBaseHandle` sets some state that requires a base handle to be present afterwards. It might be okay to just create an empty ImageData object here and return it. But I am not sure if it might have any side effects and that actually creating a handle when requesting image data in the PlainImageProviderWrapper is necessary. The same applies below, where we do not rescale the image but create a new handle as well and retrieve image data from it, which will always be empty.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2319,16 +2329,13 @@ protected Rectangle getBounds(int zoom) {
 		return Win32DPIUtils.pointToPixel(rectangle, zoom);
 	}
 
+	@Override
+	protected boolean isImageHandleRequired() {
```
HeikoKlare: This name is really expressive as it's unclear what what a handle might be required. This method is about something like `isPersistentImageHandleRequriedForImageData()`.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2319,16 +2329,13 @@ protected Rectangle getBounds(int zoom) {
...
+		return imageHandleManager.isEmpty() || memGC != null;
+	}
+
 	@Override
 	ImageData newImageData(int zoom) {
-		if (imageHandleManager.isEmpty()) {
-			return imageHandleManager.getOrCreate(zoom, () -> createBaseHandle(zoom)).getImageData();
-		}
-		// if a GC is initialized with an Image (memGC != null), the image data must not be resized, because it would
-		// be a destructive operation. Therefor, a new handle is created for the requested zoom
```
HeikoKlare, line 2328: The information is important, so it must be preserved. Probably, an adapted version is appropriate in the (renamed) `isImageHandleRequired()`  method.

PR #2965, opened 2026-01-14
[Win32] Use proper zoom for computing size of CoolBar on DPI change

HeikoKlare:
CoolBar's handleDPIChange() implementation calls computeSizeInPixels(), which is supposed be called with the zoom calculated by computeBoundsZoom() in order to properly reflect the zoom in an autoscale-disabled context. The same is done for the call in computeSize().
Even though the usage of a CoolBar inside an autoscale-disabled composite is unlikely, this unifies the used zoom for the sake of correctness and consistency.


akoch-yatta (approved):
LGTM, makes sense for consistency

PR #2967, opened 2026-01-15
Execute tests with non-default autoscaling settings in I-Build

HeikoKlare:
The SWT tests are currently executed with the default autoscaling settings during I-Builds. For Tycho test execution (e.g., in GitHub Actions runs), the configuration is modified to run in "quarter" autoscaling mode. Since multiple different autoscaling modes need to be supported (especially on Windows), the execution with just one configuration leads to a high risk of undetected regressions with other configurations.

This change extends the I-Build test execution to run the test suite with OS default settings and with one relevant non-default set of settings. For Windows, this will specifically cover the monitor-specific scaling with "quarter" autoscaling and non-monitor-specific scaling with "integer" autoscaling. For Linux and MacOS, which do not support monitor-specific scaling, this reduces to testing "integer" and "quarter" modes. This is achieved by an additional, accordingly configured test suite, which is added to the test.xml to be executed in I-Builds.
To keep the Tycho test execution times low (also used in GitHub Actions runs), that test suite is not added to Tycho test execution. The configuration is adapted to use the OS default settings for the Tycho runs (instead of using "quarter" mode, which is now covered by I-Build tests instead).

This becomes particularly useful when changing the Windows default for autoscaling in SWT to monitor-specific scaling, as it will ensure regression testing with the pre-existing default autoscaling settings:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2955

### DIFF
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/AllNonBrowserTests.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/AllNonBrowserTests.java
@@ -14,10 +14,15 @@
  *******************************************************************************/
 package org.eclipse.swt.tests.junit;
 
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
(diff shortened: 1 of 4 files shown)


HeikoKlare:
Let's take today's I-Build to see if this works as expected.

Issue #2968, opened 2026-01-15
GTK: GC.fillPolygon produces blurry results if line operation was performed before


tmssngr:
**Describe the bug**
If a line operation is performed before a `gc.fillPolygon`, the resulting polygon outline gets blurry.

**To Reproduce**
```
import org.eclipse.swt.SWT;
import org.eclipse.swt.graphics.Color;
import org.eclipse.swt.widgets.Display;
import org.eclipse.swt.widgets.Shell;

public class GcTest {

    public static void main(String[] args) {
        Display display = new Display();

        Shell shell = new Shell(display);
        shell.addListener(SWT.Paint, event -> {
            event.gc.setBackground(new Color(255, 0, 0));
            event.gc.fillRectangle(1, 1, 10, 10);

            event.gc.setAntialias(SWT.ON);
            event.gc.setLineWidth(3);
            event.gc.drawLine(50, 0, 50, 10);

            event.gc.fillPolygon(new int[] {
                    1, 20,
                    11, 20,
                    15, 25,
                    11, 30,
                    1, 30
            });
        });

        shell.setSize(400, 300);
        shell.open();

        while (!shell.isDisposed()) {
            if (!display.readAndDispatch()) {
                display.sleep();
            }
        }

        display.dispose();
    }
}
```
Uncommenting the line `event.gc.drawLine(50, 0, 50, 10);` will cause a clear result.

**Expected behavior**
`fillPolygon` must not be influenced by line widths.

**Screenshots**
With line-operation before:
<img width="826" height="991" alt="Image" src="https://github.com/user-attachments/assets/74a6610e-5423-4aa6-a612-cc0a44f9aa4b" />

Without line-operation before:
<img width="826" height="976" alt="Image" src="https://github.com/user-attachments/assets/8516a3d4-468f-4cbd-8c32-d4b20420d512" />

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [ ] Windows
- - [x] Linux
- - [ ] macOS 

2. Additional OS info (e.g. OS version, Linux Desktop, etc)
Tried with Linux Mint.

akurtakov:
Fixed with https://github.com/eclipse-platform/eclipse.platform.swt/pull/2969

PR #2969, opened 2026-01-15
GTK: GC.fillPolygon produces blurry results if line operation was performed before #2968

tmssngr:
The cairoXoffset/cairoYoffset must not be used for the fill operation, because no line width needs to be compensated.


akurtakov:
Very good catch. Thanks for it.

PR #2975, opened 2026-01-16
Extract cluttered autoscaling mode handling into dedicated classes

HeikoKlare:
The handling of different autoscaling modes is currently spread across the DPIUtil class, based on String matching and system property retrieval.
This change extracts and AutoScale class that encapsulates functionality and information related to a specific autoscaling mode (string representation, conversion, compatibility to monitor-specific scaling) into a dedicated class.

This is particularly a preparation for changing the default mode on Windows to quarter/monitor-specific scaling, as adapting the String matching logic for that would become incomprehensive.


akoch-yatta (approved):
LGTM, the autoscale handling in DPIUtil got more complex over the last months and years, so encapsulating this better makes sense

HeikoKlare:
The reported removed tests seem to be based on an missing update of test results (the last update of the comment is older than the latest CI run). The logs actually show that those tests were executed and all the jobs (including the tests) succeeded accordingly.

image

PR #2986, opened 2026-01-20
Ignore undefined all-zero alpha data for 32-bit icons without mask

ShahzaibIbrahim:
I have a png image (icon), which has no maskData and no alpha pixels (that means it is fully opaque):
image

On Windows, GetDIBits() may return undefined alpha bytes for 32-bit BI_RGB bitmaps, commonly resulting in all-zero alpha even for fully opaque images.

When icon mask data is missing, this caused opaque icons to be incorrectly treated as fully transparent.

Reject all-zero alpha data and only replace empty mask data when the alpha channel contains meaningful (non-opaque, non-zero) values.

### How to test

- Run the following snippet on 100% (only reproducible on 100%, otherwise image will be scaled, image semantics will change and we will never run into this issue):

**Note:** Add the attached image on the same folder as snippet

```
import org.eclipse.swt.*;

public class ImageTransparencyFix {
        shell.setText("Image Transparency Toolbar Demo");
        shell.setSize(300, 200);

        Image original = new Image(display, ImageTransparencyFix.class.getResourceAsStream("png-image.png"));

        ImageData data = original.getImageData();
        Image transparentImage = new Image(display, data, data.getTransparencyMask());

        ToolBar imageToolBar = new ToolBar(shell, SWT.FLAT);
        for (int i = 0; i < 3; i++) {
            ToolItem item = new ToolItem(imageToolBar, SWT.PUSH);
            item.setImage(transparentImage);
            item.setToolTipText("Image ToolItem ToolTip " + i);
        }

        // 4) Open UI
        shell.open();
        while (!shell.isDisposed()) {
        }

        // 5) Dispose resources
        transparentImage.dispose();
        original.dispose();
        display.dispose();
    }
}
```

- You will see the images on toolbar will be completely transparent.
image

### After fix 
image

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
@@ -3142,12 +3142,14 @@ private class DestroyableImageHandle implements InternalImageHandle {
 					if (!hasMaskData && depth == 32) {
 						alphaData = new byte[width * height];
 						boolean hasAlphaData = false;
+						boolean isAllZeroAlpha = true;
 						for (int pixelIndex = 0; pixelIndex < alphaData.length; pixelIndex++) {
 							alphaData[pixelIndex] = data[pixelIndex * 4 + 3];
 							hasAlphaData |= alphaData[pixelIndex] != -1;
+							isAllZeroAlpha &= alphaData[pixelIndex] == 0;
 						}
 						// In case there is alpha data, replace the empty mask data with proper alpha data
-						if (hasAlphaData) {
+						if (hasAlphaData && !isAllZeroAlpha) {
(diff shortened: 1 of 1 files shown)


ShahzaibIbrahim:
Drafted because I need suggestion here in the case where, what if the image was made completely transparent by choice? Meaning user define the image to be completely transparent (all 0 alpha pixels). Is this a viable/possible case here? Or that this fix is enough? as we can't know for sure if GetDIBits() is returning 0 as un-defined alpha-byte or was it really 0. 

Documentation for GetDIBits -> https://learn.microsoft.com/en-us/windows/win32/api/wingdi/nf-wingdi-getdibits#return-value

arunjose696:
Changes look good to me. I have tested the snippet, and it fixes the problem mentioned in the issue. However, there are some CI failures, which I don’t think are related. @ShahzaibIbrahim, could you please rerun the CI?

PR #2990, opened 2026-01-21
Extend Snippet389 to support scaling and rotating

akoch-yatta:
This commit extends Snippet389 to support scaling and rotating the (cropped) image.


akoch-yatta:
> Nice improvement of the snippet to also test the effects of transformations on the image. Maybe you could also slightly extend the snippet documentation to the reflect the new capabilities? Except for that, the change is fine.

Good point, I added the scaling and rotation.

akoch-yatta:
Additionally I check if the selected source region is the full image and then I will explicitly use GC#drawImage(Image, int, int, int, int), GC#drawImage(Image, int, int, int, int, int, int, int, int) is only used if we have a cropping scenario. This way both APIs can be tested with it.

Issue #2919, opened 2025-12-29
GC#drawImage() does not consider applied Transform when selecting best-fitting image source


HeikoKlare:
**Describe the bug**
When a `Transform` is applied to a `GC` and any of the `drawImage` methods is used, the selection of the best fitting scaled version of the image does not take into account that applied transformation. E.g., drawing a 16x16 image that exists at 100% and 200% (or even at any zoom if provided as SVG) with a `Transform` scaling by 2 in each direction, the 100% image is used an scaled up even though a 200% image as perfect fit exists.

**To Reproduce**
The following snippet shows the behavior. It can be placed in the SWT snippets project to have the according images available.
When starting the snippet, the images at the left are drawn with `GC#drawImage()` methods using double the destination size than the source size (where the best fitting scaled version is taken). The images at the right are drawn with `GC#drawImage()` methods using the same source and destination size but using a `Transform` applied to the GC that scaled by factor 2:
```java
import org.eclipse.swt.widgets.*;

public class Snippet389 {

public static void main (String [] args) {
	Display display = new Display ();
	Shell shell = new Shell(display);
	shell.setText("Snippet 389");
	ImageFileNameProvider fileNameProvider = zoom -> {
		if (zoom == 200) {
			return "src/org/eclipse/swt/snippets/eclipse32.png";
		} else if (zoom == 100) {
			return "src/org/eclipse/swt/snippets/eclipse16.png";
		}
		return null;
	};
	Image image = new Image(display, fileNameProvider);
	shell.addPaintListener(e -> {
		e.gc.drawImage(image, 0, 0, 32, 32);
		e.gc.drawImage(image, 0, 0, 16, 16, 0, 50, 32, 32);
	});

	shell.addPaintListener(e -> {
		e.gc.setTransform(new Transform(display, 2, 0, 0, 2, 0, 0));
		e.gc.drawImage(image, 25, 0, 16, 16);
		e.gc.drawImage(image, 0, 0, 16, 16, 25, 25, 16, 16);
		e.gc.setTransform(null);
	});

	shell.open ();

	while (!shell.isDisposed ()) {
		if (!display.readAndDispatch ()) display.sleep ();
	}
	display.dispose ();
}
}
```

**Expected behavior**
The `GC#drawImage()` methods shall take into account a `Transform` applied to the `GC`, such that the best fitting image handle is used. Especially in case of an SVG, it shall be rendered as a perfect fit.

[...]

PR #2994, opened 2026-01-21
[Win32] Consider transformation scale in GC.drawImage(image, x, y)

HeikoKlare:
The GC.drawImage(image, x, y) does not consider a potential transformation applied to the GC when retrieving the best-fitting image handle. The other GC.drawImage() methods are already updated accordingly.

The GC.drawImage(image, x, y) method has a special use case of drawing an image that is currently being rendered by another GC (like used in double-buffering scenarios for rulers), which also requires potentially different native and autoscale zooms to be used be applied when retrieving a handle.

For that reason, this change only delegates a call for GC.drawImage(image, x, y) to the existing scaling GC.drawImage() method in case the transformation currently applied to the GC is not the identity, as that's not the use case for rendering an image currently generated by another GC.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2919

This also fixes:
- https://github.com/eclipse-gef/gef-classic/issues/922

I have tested the change with the snippet in the following PR which addresses the above mentioned issue at consumer (GEF) side:
- https://github.com/eclipse-gef/gef-classic/pull/925

@ptziegler fyi

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
@@ -522,19 +522,6 @@ private abstract class ImageOperation extends Operation {
 		super.disposeAll();
 	}
 
-	protected float calculateTransformationScale() {
-		Transform current = new Transform(device);
-		getTransform(current);
-		float[] m = new float[6];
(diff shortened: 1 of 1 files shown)


akoch-yatta (approved):
LGTM, tested it with the mentioned Snippet in the GEF PR and an extended version of Snippet389

PR #2997, opened 2026-01-22
[Gtk] Prevent mapping to display coordinates on Gtk 4

akurtakov:
Gtk 4 and wayland do not support global coordinates thus trying to use it always end up giving the top-left corner of the window which is totally not the wanted one.


akurtakov:
Testable with Snippet165 but will afffect every place where Display.map with either from or to parameter being null.
Before:
Screenshot From 2026-01-22 17-11-23
After:
Screenshot From 2026-01-22 17-11-54

Issue #3003, opened 2026-01-23
Huge icons in Open Resources dialog


trancexpress:
To reproduce:

1. Start Eclipse in a new workspace.
2. Create a project.
3. Create a `text.txt` file in the root of the new project.
4. Create a `text.moc` file.
5. Ctrl+Shift+R to open the "Open Resources" dialog.
6. Type "test".
7. Observe huge icons.

See screen recording:

https://github.com/user-attachments/assets/b36c5e25-cf86-47c3-87e9-732dfdf7f737

I can reproduce this on Linux, GTK3, with:

```
Eclipse SDK
Version: 2026-03 (4.39)
Build id: I20260113-1800
```

Seems like the icons of external editors are not scaled properly.

SWT snippet:

```
package test;

import org.eclipse.swt.SWT;
import org.eclipse.swt.events.PaintEvent;
import org.eclipse.swt.events.PaintListener;
import org.eclipse.swt.graphics.Image;
import org.eclipse.swt.graphics.ImageData;
import org.eclipse.swt.layout.FillLayout;
import org.eclipse.swt.program.Program;
import org.eclipse.swt.widgets.Canvas;
import org.eclipse.swt.widgets.Display;
import org.eclipse.swt.widgets.Shell;

public class TestProgram {
	public static void main(String[] args) {
		Display display = new Display();
		Program program = Program.findProgram("moc");
		ImageData imageData = program.getImageData();
		Image image = new Image(display, imageData);
		Shell shell = new Shell(display);
		shell.setSize(100, 100);
		shell.setLayout(new FillLayout());
		shell.setText("Snippet 36");
		Canvas c = new Canvas(shell, SWT.BORDER);
		c.addPaintListener(new PaintListener() {
			@Override
			public void paintControl(PaintEvent e) {
				e.gc.drawImage(image, 0, 0);
			}
		});
		shell.open();
		while (!shell.isDisposed()) {
			if (!display.readAndDispatch())
				display.sleep();
		}
		display.dispose();
	}
}
```

And GTK3 snippet showing the height/width:

```
#include <gtk/gtk.h>

// gcc -g icon.c  `pkg-config --cflags --libs gtk+-3.0`  -o IconExample && ./IconExample


int
main (int argc, char **argv)
{
	GtkWidget *window, *scrolled_window;
	GtkListStore *liststore;
	GtkTreeIter iter;

	gtk_init(&argc, &argv);

[...]

PR #3004, opened 2026-01-23
[GTK] Scale external program icons to 16x16

trancexpress:
GTK3 API gtk_icon_theme_lookup_by_gicon() returns 48x48 icons, despite specifying size 16x16 in the arguments.

This change scales the retrieved icon to 16x16,
to avoid unexpected program icon sizes.

Fixes: #3003

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Program/gtk/org/eclipse/swt/program/Program.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Program/gtk/org/eclipse/swt/program/Program.java
@@ -168,7 +168,7 @@ public ImageData getImageData(int zoom) {
 			OS.g_object_unref(paintable);
 		} else {
 			long icon_theme = GTK3.gtk_icon_theme_get_default();
-			long gicon_info = GTK3.gtk_icon_theme_lookup_by_gicon(icon_theme, gicon, 16/*size*/, 0);
+			long gicon_info = GTK3.gtk_icon_theme_lookup_by_gicon_for_scale(icon_theme, gicon, 16/*size*/, 1/*scale*/, 0);
 			if (gicon_info != 0) {
 				pixbuf = GTK3.gtk_icon_info_load_icon(gicon_info, null);
 				OS.g_object_unref(gicon_info);
@@ -204,6 +204,10 @@ public ImageData getImageData(int zoom) {
 					}
 				}
 				data.alphaData = alphaData;
+				if (height > 16 && width > 16) {
+					ImageData scaled = data.scaledTo(16, 16);
+					data = scaled;
+				}
 			} else {
 				PaletteData palette = new PaletteData(0xFF0000, 0xFF00, 0xFF);
 				data = new ImageData(width, height, 24, palette, 4, srcData);
--- a/bundles/org.eclipse.swt/Eclipse SWT PI/gtk/library/gtk3.c
+++ b/bundles/org.eclipse.swt/Eclipse SWT PI/gtk/library/gtk3.c
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2025 IBM Corporation and others.
+ * Copyright (c) 2000, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -1093,6 +1093,18 @@ JNIEXPORT jlong JNICALL GTK3_NATIVE(gtk_1icon_1theme_1lookup_1by_1gicon)
 	GTK3_NATIVE_ENTER(env, that, gtk_1icon_1theme_1lookup_1by_1gicon_FUNC);
 	rc = (jlong)gtk_icon_theme_lookup_by_gicon((GtkIconTheme *)arg0, (GIcon *)arg1, (gint)arg2, (GtkIconLookupFlags)arg3);
 	GTK3_NATIVE_EXIT(env, that, gtk_1icon_1theme_1lookup_1by_1gicon_FUNC);
+	return rc;
+}
+#endif
+
+#ifndef NO_gtk_1icon_1theme_1lookup_1by_1gicon_1for_1scale
+JNIEXPORT jlong JNICALL GTK3_NATIVE(gtk_1icon_1theme_1lookup_1by_1gicon_1for_1scale)
+	(JNIEnv *env, jclass that, jlong arg0, jlong arg1, jint arg2, jint arg3, jint arg4)
+{
+	jlong rc = 0;
+	GTK3_NATIVE_ENTER(env, that, gtk_1icon_1theme_1lookup_1by_1gicon_1for_1scale_FUNC);
+	rc = (jlong)gtk_icon_theme_lookup_by_gicon_for_scale((GtkIconTheme *)arg0, (GIcon *)arg1, (gint)arg2, (gint)arg3, (GtkIconLookupFlags)arg4);
+	GTK3_NATIVE_EXIT(env, that, gtk_1icon_1theme_1lookup_1by_1gicon_1for_1scale_FUNC);
 	return rc;
 }
 #endif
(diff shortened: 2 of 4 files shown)


trancexpress:
@akurtakov any better ideas here?

I'll try to get a 16x16 icon with GTK3 APIs, somehow. I do have such sizes available... I'm not sure why 48x48 is always returned.

bundles/org.eclipse.swt/Eclipse SWT Program/gtk/org/eclipse/swt/program/Program.java
```diff
@@ -204,6 +204,8 @@ public ImageData getImageData(int zoom) {
 					}
 				}
 				data.alphaData = alphaData;
+				ImageData scaled = data.scaledTo(16, 16);
```
trancexpress: I guess we can check if the height or width are higher than 16 and only then scale...

iloveeclipse: Makes sense. 

Did you found the commit that changed this icons size?

akurtakov:
Is there 16x16 icon available at all (/usr/share/icons/themename/... or smth similar)? If not what are the available sizes? I believe Gtk returns the closest bigger size if the requested size is not available. If this is on HiDPI system it becomes even more complicated as 16x16 with scale=2 will try to return 32x32 which in case of not available might give you the next one 48x48.
On a HiDPI screen I see too small icon - both with and without the patch 
image

akurtakov:
I don't believe there would be such a "simple" fix.

bundles/org.eclipse.swt/Eclipse SWT Program/gtk/org/eclipse/swt/program/Program.java
trancexpress: Probably its this commit in platform UI:

https://github.com/eclipse-platform/eclipse.platform.ui/commit/f0b1071e25e6f03cb49660e647199117c79580b3

For too large icons, we used to show a default. After this commit we don't.

trancexpress:
> Is there 16x16 icon available at all (/usr/share/icons/themename/... or smth similar)? If not what are the available sizes? I believe Gtk returns the closest bigger size if the requested size is not available.

I think there is:

```
#include <gtk/gtk.h>

// gcc -g icon.c  `pkg-config --cflags --libs gtk+-3.0`  -o IconExample && ./IconExample


int
main (int argc, char **argv)
{
	GtkWidget *window, *scrolled_window;
	GtkListStore *liststore;
	GtkTreeIter iter;

	gtk_init(&argc, &argv);

	window = gtk_window_new(GTK_WINDOW_TOPLEVEL);
	g_signal_connect(window, "delete_event", gtk_main_quit, NULL);
	gtk_window_set_default_size(GTK_WINDOW(window), 200, 100);

	scrolled_window = gtk_scrolled_window_new(NULL, NULL);

	GIcon *gicon = g_icon_new_for_string("kate", NULL);
	GtkIconTheme *icon_theme = gtk_icon_theme_get_default();
	GtkIconInfo *gicon_info = gtk_icon_theme_lookup_by_gicon(icon_theme, gicon, 16/*size*/, 0);
	GdkPixbuf *pixbuf = gtk_icon_info_load_icon(gicon_info, NULL);

	int height = gdk_pixbuf_get_height(pixbuf);
	int width = gdk_pixbuf_get_width(pixbuf);
	g_print("height=%d, width=%d\n", height, width);

	gtk_container_add (GTK_CONTAINER (window), scrolled_window);

	gtk_widget_show_all(window);

	gtk_main();

	return 0;
}
```

```
$ gcc -g icon.c  `pkg-config --cflags --libs gtk+-3.0`  -o IconExample && ./IconExample
height=48, width=48
```

```
$ find /usr/share/icons/ -name "kate.png"  | sort -V
/usr/share/icons/hicolor/16x16/apps/kate.png
/usr/share/icons/hicolor/22x22/apps/kate.png
/usr/share/icons/hicolor/32x32/apps/kate.png
/usr/share/icons/hicolor/44x44/apps/kate.png
/usr/share/icons/hicolor/48x48/apps/kate.png
/usr/share/icons/hicolor/64x64/apps/kate.png
/usr/share/icons/hicolor/128x128/apps/kate.png
/usr/share/icons/hicolor/150x150/apps/kate.png
/usr/share/icons/hicolor/256x256/apps/kate.png
/usr/share/icons/hicolor/310x310/apps/kate.png
/usr/share/icons/hicolor/512x512/apps/kate.png
```

> If this is on HiDPI system it becomes even more complicated as 16x16 with scale=2 will try to return 32x32 which in case of not available might give you the next one 48x48. On a HiDPI screen I see too small icon - both with and without the patch

What is the difference for HiDPI? Will the zoom level argument be different?

The GTK3 documentation does say this: https://docs.gtk.org/gtk3/method.IconTheme.lookup_by_gicon.html

> When rendering on displays with high pixel densities you should not use a size multiplied by the scaling factor returned by functions like gdk_window_get_scale_factor(). Instead, you should use gtk_icon_theme_lookup_by_gicon_for_scale(), as the assets loaded for a given scaling factor may be different.

akurtakov:
Yes, looks like gtk_icon_theme_lookup_by_gicon_for_scale https://docs.gtk.org/gtk3/method.IconTheme.lookup_by_gicon_for_scale.html has to be used.

trancexpress:
Slightly modifying the prints:

```
g_print("height=%d, width=%d, %s\n", height, width, gtk_icon_info_get_filename(gicon_info));
```

Then I see:

```
$ gcc -g icon.c  `pkg-config --cflags --libs gtk+-3.0`  -o IconExample && ./IconExample
height=48, width=48, /usr/share/icons/breeze-dark/apps/48/kate.svg
```

```
$ find  /usr/share/icons/breeze-dark/apps/ -name "*kate*"
/usr/share/icons/breeze-dark/apps/22/kate-symbolic.svg
/usr/share/icons/breeze-dark/apps/16/kate-symbolic.svg
/usr/share/icons/breeze-dark/apps/48/kate.svg
/usr/share/icons/breeze-dark/apps/24/kate-symbolic.svg
/usr/share/icons/breeze-dark/apps/32/kate-symbolic.svg
```

In the Advantest environment, `emacs` is used to open the external file... There I see:

```
$ find  /usr/share/icons/breeze/apps/ -name "*emacs*"
/usr/share/icons/breeze/apps/48/emacs.svg
```

So we must be missing icons... at least for `breeze`...

trancexpress:
> Yes, looks like gtk_icon_theme_lookup_by_gicon_for_scale https://docs.gtk.org/gtk3/method.IconTheme.lookup_by_gicon_for_scale.html has to be used.

OK, I'll try to in the next few days.

I assume for the scale parameter we use `DPIUtil.getScalingFactor(int)`?

And the scaling down we can do for scale 1, when hopefully we are not in the HiDPI case. Considering that `breeze` supplies a 48x48 SVG icon, that should be scalable however we want... I don't think we want to request more icon sizes for the theme.

Though SWT won't be scaling the SVG in a sensible way... Unless we try to open the SVG by using the path to the icon. Which will make things even more complicated, so I won't be trying that.

bundles/org.eclipse.swt/Eclipse SWT Program/gtk/org/eclipse/swt/program/Program.java
```diff
@@ -168,7 +171,15 @@ public ImageData getImageData(int zoom) {
 			OS.g_object_unref(paintable);
 		} else {
 			long icon_theme = GTK3.gtk_icon_theme_get_default();
-			long gicon_info = GTK3.gtk_icon_theme_lookup_by_gicon(icon_theme, gicon, 16/*size*/, 0);
+			Shell shell = Display.getCurrent().getActiveShell();
+			if (shell != null) {
+				long shellHandle = shell.handle;
+				long window = GTK3.gtk_widget_get_window(shellHandle);
+				if (window != 0) {
+					scale = GDK.gdk_window_get_scale_factor(window);
```
trancexpress: Looks like to get the scaling argument we need a window.

Any better way to get the window @akurtakov ?

trancexpress: Thinking about this, we are requesting icon data that is likely cached by platform UI.

Depending on the screen scaling doesn't seem like a good idea. E.g. if two screens have different scaling, we would be caching the icon for one of them...

trancexpress: @HeikoKlare maybe you have insights here? Where do we handle different DPI for icons in platform UI? Maybe doing the fix in SWT is not the best place...

PR #3006, opened 2026-01-23
[Win32] Update JavaDoc in GC.setXORMode() regarding limitation with GDI+

### LINKED ISSUE #2984: [Win32] XOR mode doesn't work when advanced mode is enabled
**Describe the bug**

When enabling advanced mode, the XOR mode doesn't seem to do anything. Are there any rules I missed that make them incompatible?

**To Reproduce**

```java
package test;

import org.eclipse.swt.SWT;

public class TestXOR {
		Shell shell = new Shell();
		Display d = shell.getDisplay();

			GC gc = event.gc;
			
			// gc.setAdvanced(true);
			gc.setBackground(d.getSystemColor(SWT.COLOR_BLUE));
			gc.fillRectangle(5,5,90,45);
			gc.setXORMode(true);
			gc.setBackground(d.getSystemColor(SWT.COLOR_WHITE));
			gc.fillRectangle(20,20,50,50);
			gc.setBackground(d.getSystemColor(SWT.COLOR_RED));
			gc.fillOval(80,20,50,50);
		});

		shell.open();
		shell.setSize(200, 200);
		
		while(!shell.isDisposed()) {
			while(!d.readAndDispatch()) {
				d.sleep();
			}
		}
	}
}
```

**Expected behavior**

XOR mode should work regardless of advanced mode.

**Screenshots**

Without advanced mode:

Image

With advanced mode:

Image

**Environment:**
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 

Windows 11

3. JRE/JDK version
 
**Version since**

Definitely the 2025-12 release, but earlier versions are probably affected as well.

ptziegler:
Enabling advanced mode causes SWT to switch from GDI to the GDI+. With GDI+, certain features, such as XOR, are not supported. This limitation should be documented.

Closes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2984

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
@@ -5707,14 +5707,22 @@ private void setLineWidthInPixels(int lineWidth) {
 }
 
 /**
- * If the argument is <code>true</code>, puts the receiver
- * in a drawing mode where the resulting color in the destination
- * is the <em>exclusive or</em> of the color values in the source
- * and the destination, and if the argument is <code>false</code>,
- * puts the receiver in a drawing mode where the destination color
- * is replaced with the source color value.
+ * If the argument is <code>true</code>, puts the receiver in a drawing mode
+ * where the resulting color in the destination is the <em>exclusive or</em> of
+ * the color values in the source and the destination, and if the argument is
+ * <code>false</code>, puts the receiver in a drawing mode where the destination
+ * color is replaced with the source color value.
+ * <p>
+ * <b>IMPORTANT:</b> This operation may not work on every platform or with every
+ * configuration. Especially on Windows, GDI+ has no direct support for raster
+ * operations (ROPs). This means that advanced graphics must be <b>disabled</b>
+ * when using XOR mode. More specifically, this means that all operations that
+ * require advanced mode (e.g. {@link Transform}) can't be used in combination
+ * with XOR mode.
+ * </p>
  *
- * @param xor if <code>true</code>, then <em>xor</em> mode is used, otherwise <em>source copy</em> mode is used
+ * @param xor if <code>true</code>, then <em>xor</em> mode is used, otherwise
+ *            <em>source copy</em> mode is used
  *
  * @exception SWTException <ul>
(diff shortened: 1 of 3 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -5695,6 +5695,13 @@ private void setLineWidthInPixels(int lineWidth) {
  * and the destination, and if the argument is <code>false</code>,
  * puts the receiver in a drawing mode where the destination color
  * is replaced with the source color value.
+ * <p>
+ * <b>IMPORTANT:</b> GDI+ has no direct support for raster
```
HeikoKlare: The Javadoc of SWT API needs to be the same for every OS (as it is or at least can be consumed in an OS-agnostic way). So this extension needs to be applied to the GTK and Cocoa implementations of `GC` as well.
To make such a comment reasonable in an OS-agnostic way, it should start with an according statement, such as: "This operation may not work on every platform. Especially on Windows, GDI+ has no direct support [...]".

HeikoKlare (commented):
Thank you for adding this explanation to the Javadoc. It makes sense to have this limited functionality documented.

HeikoKlare:
I lost track of this PR, but I would still appreciate inclusion of that statement in GC Javadoc. @ptziegler do you want to finalize this or should I take over and consistently update the Javadoc for all platforms?

HeikoKlare:
@ptziegler I took the liberty to update your commit with a unification of the Javadoc adaption in the `GC` implementations for every OS. Are you fine with the state?

ptziegler:
@HeikoKlare Apologies, I didn't see the notification up until now. Originally, I wanted to restore the header that was removed with 1af84db9d9097f2c1d4b3b33cc3dc258334486cb, but your description is much more detailed. From my side this looks perfect.

HeikoKlare:
No problem, there is no need to hurry with this 🙂 And thank you for the feedback. I just took over the detailed explanation that you had added and I think it's valuable to provide so much information to the user. E.g., if you want to use the functionality in situations that are know to not be affected by limitations, it will be good to have this precise information instead of being generally concerned that the functionality may not work as expected.

Issue #3011, opened 2026-01-26
ImageData reports wrong size for Image created from SVG InputStream


ptziegler:
**Describe the bug**

When an SVG image is created from an InputStream, the `width` and `height` of the ImageData is `1`. Those images are then drawn with size 1x1.

**To Reproduce**

Execute this snippet based on `Snippet360`.

```java
/*******************************************************************************
 * Copyright (c) 2026 Eclipse Platform Contributors and others.
 *
 * This program and the accompanying materials
 * are made available under the terms of the Eclipse Public License 2.0
 * which accompanies this distribution, and is available at
 * https://www.eclipse.org/legal/epl-2.0/
 *
 * SPDX-License-Identifier: EPL-2.0
 *
 * Contributors:
 *     Eclipse Platform Contributors - initial API and implementation
 *******************************************************************************/
package org.eclipse.swt.snippets;

import java.nio.file.Path;

import org.eclipse.swt.graphics.*;
import org.eclipse.swt.layout.*;
import org.eclipse.swt.widgets.*;

/**
 * Snippet to draw an SVG image using three Image constructors
 * <p>
 * Useful for comparing rendering on different operating systems.
 * <p>
 * For a list of all SWT example snippets see
 * http://www.eclipse.org/swt/snippets/
 *
 * @see org.eclipse.swt.graphics.Image#Image(org.eclipse.swt.graphics.Device,
 *      java.io.InputStream)
 * @see org.eclipse.swt.graphics.Image#Image(org.eclipse.swt.graphics.Device,
 *      String)
 * @see org.eclipse.swt.graphics.Image#Image(org.eclipse.swt.graphics.Device,
 *      org.eclipse.swt.graphics.ImageFileNameProvider)
 */
public class Snippet390 {

	public static void main(String[] args) throws Exception {
		final Display display = new Display();

		final Shell shell = new Shell(display);
		shell.setText("Snippet 390");
		shell.setLayout(new FillLayout());
		shell.setBounds(250, 50, 1000, 400);
		shell.setBackground(new Color(255, 255, 255));

		String imgPath = "eclipse.svg";
		Path fullPath = Path.of(Snippet390.class.getResource(imgPath).toURI());

		// Image loaded in stream
		Image image1 = new Image(null, Snippet390.class.getResourceAsStream(imgPath));

		// Image loaded in file path
		Image image2 = new Image(null, fullPath.toString());

		// Image loaded in ImageFileNameProvider
[...]

PR #3012, opened 2026-01-26
Fix improper ImageData size when loading SVG from InputStream #3011

ptziegler:
On Linux, loading an `Image` from an `InputStream` is internally handled via an `ImageDataAtSizeProvider`.

When loading `ImageData` by zoom, this interface requires a proper implementation of the `getDefaultSize()` method, which in return requires the `ImageData` be loaded at 100% zoom, even when the source data is dynamically sizable. To avoid loading the same resource twice, this data is then used when requesting the `ImageData` with default size.

The tests have been updated to test the loading of an `Image` from an `InputStream` and to also check the size of the `ImageData` at 100% zoom.

Closes https://github.com/eclipse-platform/eclipse.platform.swt/issues/3011

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2025 IBM Corporation and others.
+ * Copyright (c) 2000, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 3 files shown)


bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/Image.java
```diff
@@ -1519,15 +1519,30 @@ private void initUsingImageDataProvider(ImageDataProvider imageDataProvider) {
...
+			}
+
+			@Override
+			public ImageData getImageData(int width, int height) {
+				if (width == defaultSize.x && height == defaultSize.y) {
+					return imageData;
+				}
+				return ImageDataLoader.loadBySize(new ByteArrayInputStream(streamData), width, height);
+			}
+		};
```
HeikoKlare: This should work, still I would propose to just implement the `getImageData(zoom)` method instead of the `getDefaultSize()` method because of two reasons:
- You avoid loading the 100% image data even though you potentially do not need the data (but just the size)
- In general, the advice is to overwrite `getImageData(zoom)` in case the _provider_ (the file) specifies the default 100% size and `getDefaultSize()` in case the _consumer_ does it. In this case, you want to have the provider (the file) specify the 100% size, which is why simply loading the file in the existing by-zoom way is most appropriate. In the current proposa, the provider specifies the size, still `getDefaultSize()` is overwritten, which makes this workaround of loading the data at 100% to extract the provider size which is then passed by the consumer again necessary. 
```suggestion
		ImageDataAtSizeProvider imageDataAtSizeProvider = new ImageDataAtSizeProvider() {
@Override
			public ImageData getImageData(int zoom) {
				return ImageDataLoader.loadByZoom(new ByteArrayInputStream(streamData), FileFormat.DEFAULT_ZOOM, zoom).element();
			}
			
			@Override
			public ImageData getImageData(int width, int height) {
				return ImageDataLoader.loadBySize(new ByteArrayInputStream(streamData), width, height);
			}
		};
```

ptziegler: I wasn't really sure which of the default methods to override, but that makes sense, of course. The call to `getDefaultSize()` no longer returns a meaningful size, but that's probably not releveant for an instance that is only used internally.

PR #3018, opened 2026-01-28
[Win32] Send zoom change event if autoscale disablement changes

HeikoKlare:
When the autoscale disablement state of a control changes such that native zoom of the control is modified, no ZoomChanged event has been sent yet. In case a consumer has created some state on top of the Control that depends on its zoom (such as a GC), this state may become inconsistent.

In order to allow a proper update of that state, with this change a ZoomChanged event is sent like in every other case in which the zoom of the control changes.

Fixes https://github.com/eclipse-gef/gef-classic/issues/984

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
@@ -1313,14 +1313,12 @@ private void updateAutoScalingModeFromData() {
 	Object propagateAutoscaleDisabled = getData(PROPOGATE_AUTOSCALE_DISABLED);
 	boolean propagateAutoscaling = propagateAutoscaleDisabled == null || Boolean.parseBoolean(propagateAutoscaleDisabled.toString());
 	Object autoscaleDisabled = getData(DATA_AUTOSCALE_DISABLED);
-	boolean isAutoscaleDisabled = autoscaleDisabled != null && Boolean.parseBoolean(autoscaleDisabled.toString());
+	boolean isAutoscaleDisabled = autoscaleDisabled != null && Boolean.parseBoolean(autoscaleDisabled.toString()) || parent.autoscalingMode == AutoscalingMode.DISABLED_INHERITED;
+	AutoscalingMode newAutoscalingMode = AutoscalingMode.ENABLED;
 	if (isAutoscaleDisabled) {
-		autoscalingMode = propagateAutoscaling ? AutoscalingMode.DISABLED_INHERITED : AutoscalingMode.DISABLED;
-		this.nativeZoom = 100;
-	} else {
-		autoscalingMode = AutoscalingMode.ENABLED;
-		this.nativeZoom = getShellZoom();
+		newAutoscalingMode = propagateAutoscaling ? AutoscalingMode.DISABLED_INHERITED : AutoscalingMode.DISABLED;
 	}
+	setAutoscalingMode(newAutoscalingMode);
 }
 
 /**
@@ -3403,10 +3401,11 @@ private void fitInParentBounds(Rectangle boundsInPixels, int zoom) {
  */
 public boolean setAutoscalingMode(AutoscalingMode autoscalingMode) {
 	this.autoscalingMode = autoscalingMode;
-	if (isAutoscalingDisabled()) {
-		this.nativeZoom = 100;
-	} else {
-		this.nativeZoom = getShellZoom();
+	int newZoom = isAutoscalingDisabled() ? 100 : getShellZoom();
+	if (nativeZoom != newZoom) {
+		nativeZoom = newZoom;
+		Event zoomChangedEvent = createZoomChangedEvent(newZoom, false);
+		notifyListeners(SWT.ZoomChanged, zoomChangedEvent);
 	}
 	return true;
 }


Phillipus:
@HeikoKlare Thanks for taking a look at this.

I tested the Snippet in https://github.com/eclipse-gef/gef-classic/issues/984 and https://github.com/eclipse-gef/gef-classic/issues/957 and also tested with Archi (RCP app seems to have fixed the problem.

BTW - for me it's hard to know whether the cause of an issue is in GEF/Draw2d or SWT and so my issues are perhaps not so clear as I spend a lot of time on trial and error and understanding this Windows scaling business. ;-)

Phillipus:
Update to my last comment - the exported image is clipped again when running the Snippet in https://github.com/eclipse-gef/gef-classic/issues/977 (and the "Create Image" button is half size).

Phillipus:
Testing this with this snippet on Windows 200%:

```java
import org.eclipse.gef.GraphicalViewer;
import org.eclipse.gef.ui.parts.GraphicalViewerImpl;
import org.eclipse.swt.SWT;
import org.eclipse.swt.widgets.Button;

public class TinyButton {
    
        Shell shell = new Shell();
        shell.setSize(200, 100);
        
        GraphicalViewer viewer = new GraphicalViewerImpl();
        viewer.setControl(shell);
        
        button.setText("Create Image");
        
        shell.open();
        
        while(!shell.isDisposed()) {
                display.sleep();
            }
        }
    }
}
```

The line `viewer.setControl(shell)` results in the control scaled too small.

test2

test

HeikoKlare:
> BTW - for me it's hard to know whether the cause of an issue is in GEF/Draw2d or SWT and so my issues are perhaps not so clear as I spend a lot of time on trial and error and understanding this Windows scaling business. ;-)

We were also not completely sure. You could put the burden on both here: it's not that nice that Draw2d creates a GC on a control before changing the configuration of that control (which requires the GC to adapt). That's something that you could fix in Draw2d (as you have mentioned in the according issue by reordering the statements) or in SWT by sending a ZoomChanged event (like we do here). Both is basically fine, but for the sake of consistency we think it makes sense to send according ZoomChanged events in such a case anyway.

HeikoKlare:
> The line `viewer.setControl(shell)` results in the control scaled too small.

The button has exactly the expected size. The `GraphicalViewerImpl` disables autoscaling for the shell and this setting is inherited to all child controls (such as the button).

Phillipus:
> The button has exactly the expected size. The GraphicalViewerImpl disables autoscaling for the shell and this setting is inherited to all child controls (such as the button).

Before this PR this wasn't the case. The button was full size.

HeikoKlare:
> Before this PR this wasn't the case. The button was full size.

It's the same behavior as it was on 4.38, isn't it?

Phillipus:
> It's the same behavior as it was on 4.38, isn't it?

On 4.38 and the commit before this PR the button is full size.

HeikoKlare:
> On 4.38 and the commit before this PR the button is full size.

I cannot reproduce that. For me, it is not scaled on 4.38 but accidentally scaled on the commit before this PR.

Phillipus:
> I cannot reproduce that. For me, it is not scaled on 4.38 but accidentally scaled on the commit before this PR.

Ugh, I've got so many snippets here I was using the wrong one. Yes, in the above snippet the button is small..._.if_ it's added to the Shell _after_ `viewer.setControl(shell)` but not if added before.

But since the PR the button is small if added before or after.

Use this one:

```java
import org.eclipse.gef.GraphicalViewer;
import org.eclipse.gef.ui.parts.GraphicalViewerImpl;
import org.eclipse.swt.SWT;
import org.eclipse.swt.widgets.Button;

public class TinyButton {
    
        Shell shell = new Shell();
        shell.setSize(200, 100);
        
        GraphicalViewer viewer = new GraphicalViewerImpl();
        
        button.setText("Create Image");
        
        shell.open();
        viewer.setControl(shell);
        
        while(!shell.isDisposed()) {
                display.sleep();
            }
        }
    }
}
```

HeikoKlare:
That last snippet is prone to break anyway. You create a parent control, add a child and then reconfigure autoscaling of the parent control. There is no guarantee that the child will properly inherit that. Is there any need to do it like that?

Phillipus:
> Is there any need to do it like that?

No, I can add it to a child `Canvas` instead. I just thought I'd mention it in case it indicated a problem somewhere else. Going forward, in my snippets I'll do this:

```java
Canvas canvas = new Canvas(shell, 0);
GridDataFactory.fillDefaults().grab(true, true).applyTo(canvas);
viewer.setControl(canvas);
```

Phillipus:
> The issue with the cropped image is reproducible without this change as well, is the Shell is moved forth and back to the original monitor. Reason is the State of e.g. the FlowText is calculated with a disabled autoscaling (=100%), but the GC that is used to print it on the Image is initialized with the native zoom (e.g. of 200%). The font that is used then does not match e.g. the clipping that was calculated before. It worked somehow before because the GC in DrawableFigureUtilities was before initialized with the wrong zoom if the autoscale was disablement.

I linked your comment to the original GEF issue - https://github.com/eclipse-gef/gef-classic/issues/977

Phillipus:
I found another related issue - https://github.com/eclipse-gef/gef-classic/issues/993

PR #3022, opened 2026-01-28
[Win32] Refresh image created by GC when drawn onto another drawable

HeikoKlare:
With the improvements to use/generate the best-fitting handle when drawing an image on some other drawable, the previous refresh of the image when being drawn by a GC was implicitly removed. In consequence, the image-to-be-drawn is not refreshed anymore for a zoom for which no handle exists yet, even though there is a GC that could recreate the image for that zoom.

With this change, in case a GC for the image-to-be-drawn is still active, the image will be requested for the zoom to be drawn at, such that the GC recreates a proper handle for that zoom.


akoch-yatta (approved):
LGTM, ensuring the handle creation for this special case inside GC is the least invasive solution I can think of as well

PR #3023, opened 2026-01-29
[Win32] Update shell icons on DPI change asynchronously

HeikoKlare:
When changing the zoom of the current monitor, the taskbar icon for an SWT application is not updated properly but is replaced with a generic executable icon and will only be replaced with the actual application after quite some time/interaction (if at all). This seems to be caused by some caching effects of the Windows explorer.

This change defers the refresh of the icons to be executed after all DPI change processing has been handled. This results in reliable updates of the application icon once the DPI change processing has finished.


vogella:
Change looks good and works fine locally. 

If this is merged today, I can test this tomorrow also in the client environment, in this environment I cannot build SWT or transfer JAR files to it.

vogella:
Tested in the client enviroment with todays build the images get correctly updated

HeikoKlare:
Great. Thank you for testing and giving the feedback, Lars!

Issue #3020, opened 2026-01-28
Spacing larger in SWT 2026-03 compared to 2025-12  using IBM's Semeru JDK


vogella:
If we use -Dswt.autoScale.updateOnRuntime the empty space in SWT get bigger. As this is enable by default this is also the case if the latest and greatest SWT is used.

Here is the same application side by side using the different SWT versions.

<img width="400" height="852" alt="Image" src="https://github.com/user-attachments/assets/23bb9750-e9b3-43a8-a80e-8e20c7cf32f7" />

Maybe similar to https://github.com/eclipse-platform/eclipse.platform.ui/issues/2852

HeikoKlare:
Are you sure this is about spacing? To me it looks as if the images are larger, which is somehow expected as the previous behavior only used 100% and 200% and layouted around that, leading to rather strange behavior. With monitor-specific scaling, images are always scaled according to the text size. Can you test if you get the same result with 2025-12 when using `-Dswt.autoScale=quarter` than with the current development state and monitor-specific scaling?

vogella:
This is also in areas without images, like the search mask (the navigation bar is a separate view but can not be minized)

<img width="464" height="136" alt="Image" src="https://github.com/user-attachments/assets/2d055058-502d-4962-b6ef-b9676a0f6426" />

HeikoKlare:
From the screenshots, I would still expect the behavior to come from the autoScale mode (also regarding the spacings). Have you tested already?
It would also be interesting to know if this is really something that only happens with Semeru JDK (as indicated by the title).

Is there any chance to see the different behavior with a plain SWT snippet?

vogella:
I think Control should use getShellZoom, maybe this is related. Can you have a look at https://github.com/eclipse-platform/eclipse.platform.swt/pull/3029

HeikoKlare:
I have given a bunch of advices and proposed things to test. Would be great to check them first. And it would be great to have a means to reproduce your issues with some SWT snippet that we can use as a basis for discussion.

vogella:
Happens also with Temurin 21 or Temurin 25 Runtime (which I think is also the basis for JustJ, not sure here).

merks:
Yes the JustJ JREs currently are purely based on Temurin JDKs.

HeikoKlare:
Is there any action we need to take here? Since the issue was reported as a regression in 2026-03 compared to 2025-12, it would be good if we could fix it for the release (which is very soon).

[...]

PR #3029, opened 2026-01-29
Ensuring AdjustWindowRectExForDpi uses consistent zoom

vogella:
Changed Control.adjustWindowRectEx to use getShellZoom() instead of shell.getZoom(). This ensures consistency with nativeZoom which is used by computeTrim and other layout logic, preventing the possibility of incorrect
border sizing when shell.getZoom() and nativeZoom are out of sync

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Control.java
@@ -6161,9 +6161,7 @@ int getSystemMetrics(int nIndex) {
 }
 
 boolean adjustWindowRectEx(RECT lpRect, int dwStyle, boolean bMenu, int dwExStyle) {
-	Shell shell = getShell();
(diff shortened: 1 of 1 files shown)


HeikoKlare:
> shell.getZoom() and nativeZoom are out of sync

This should only be the case if you have disabled autoscaling for a control. And in that case it's necessary to take the proper shell zoom for those metrics as otherwise scaling will be wrong. So from my understanding, the current implementation is correct as is.

Can you give an example where the current implementation does not behave as expected?

vogella:
> > shell.getZoom() and nativeZoom are out of sync
> 
> This should only be the case if you have disabled autoscaling for a control. And in that case it's necessary to take the proper shell zoom for those metrics as otherwise scaling will be wrong. So from my understanding, the current implementation is correct as is.
> 
> Can you give an example where the current implementation does not behave as expected?

https://github.com/eclipse-platform/eclipse.platform.swt/issues/3020 might be related.

vogella:
LTM that the other widgets are using the same approach hence this proposed change.

image


shell.getZoom() is only used in one place

image

Issue #3028, opened 2026-01-29
NPE when setting "AUTOSCALE_DISABLED to false


Phillipus:
Windows 11
Monitor scaling 200%
Latest SWT commit

Playing around with different settings to get Draw2d to play nice with latest SWT I set `AUTOSCALE_DISABLED` to false:

```java
import org.eclipse.swt.widgets.Shell;

public class NPEDisplay {
    
    public static void main(String[] args) {
        Shell shell = new Shell();
        shell.setData("AUTOSCALE_DISABLED", false);
    }
}
```

This leads to:

```
Exception in thread "main" java.lang.NullPointerException: Cannot read field "autoscalingMode" because "this.parent" is null
	at org.eclipse.swt.widgets.Control.updateAutoScalingModeFromData(Control.java:1316)
	at org.eclipse.swt.widgets.Control.setData(Control.java:1302)
	at swt.NPEDisplay.main(NPEDisplay.java:9)
```

Don't know if doing this is breaking the rules but it's still a NPE.

PR #3032, opened 2026-01-29
[Win32] Fix NPE when setting autoscale disablement on top-level control

HeikoKlare:
With a recent change, a check for autoscale disablement inheritance from parent was added without a check if there is a parent at all. This can lead to an NPE. This change fixes that.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/3028


akoch-yatta (approved):
LGTM, fixes the NPE and keeps the logic

PR #3035, opened 2026-01-30
Added documentation about SWT's Autoscale methods

ShahzaibIbrahim:
This PR adds the documentation about SWT's Autoscale methods and their behavior

### DIFF
--- a/docs/hidpi-support.md
+++ b/docs/hidpi-support.md
@@ -1,0 +1,30 @@
+# HiDPI Support in SWT
+
+This document (incompletely) describes certain parts of the HiDPI support in SWT for all platforms (Windows, MacOS, GTK).
+
+## Table of Contents
+
+1. [Autoscale Methods](#autoscale-Methods)
+
+### Autoscale Methods
+
+The `swt.autoScale.method` system property controls the method used to determine how to handle scaling of Images in SWT:
+
+- **`smooth`**
+
+- Uses anti-aliased / bilinear interpolation
+- Smooth edges, may appear slightly blurry
+
+- **`nearest`**
+
+- Uses nearest-neighbor interpolation
+- Edges appear jagged or blocky
+
+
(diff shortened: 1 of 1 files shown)


docs/hidpi-support-for-windows-dev-guide.md
```diff
@@ -330,6 +330,31 @@ The `swt.autoScale.updateOnRuntime` system property controls dynamic DPI changes
...
+- **`smooth`**
+
+- Uses fractional DPI scale factors (e.g. 1.25, 1.5) instead of rounding to integers.
+- Images are scaled by a floating-point factor to match the system DPI exactly.
+- Smooth edges, may look blurry
+
+- **`nearest`**
+
+- Does not use fractional DPI values
+- The system DPI scale is rounded to the nearest integer (e.g. 1.25 → 1, 1.5 → 2)
```
HeikoKlare: Is this correct? In my understanding, `smooth` and `nearest` both use exact scale factors. They only differ in the interpolation mode (nearest neighbor vs. anti-aliasing).

docs/hidpi-support-for-windows-dev-guide.md
```diff
@@ -330,6 +330,31 @@ The `swt.autoScale.updateOnRuntime` system property controls dynamic DPI changes
 **Note**: Runtime rescaling requires compatible autoScale modes (`quarter`, `exact`).
 Modes like `integer` or fixed values are incompatible with monitor-specific scaling and will only be accepted when `force` is used.
 
+### Autoscale Methods
```
HeikoKlare: I think we should create a separate "hidpi-support.md" or the like file for this, as it is not specific to Windows but generic for all platforms.

docs/hidpi-support-for-windows-dev-guide.md
ShahzaibIbrahim: That's correct. I will update the description.

PR #3039, opened 2026-01-30
Replace # with . in hidpi documentation

akoch-yatta:
This PR replaces the remaining hashtag notations for methods with a dot notation.


HeikoKlare (approved):
Only documentation changes, no need to wait for CI.

PR #3040, opened 2026-01-31
[Win32] Add test case to check auto-scaling mode calculation

ptziegler:
This adds test cases to check that the auto-scaling mode is correctly read from the widget data and correctly propagated to child widgets.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/ControlWin32Tests.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/ControlWin32Tests.java
@@ -369,4 +369,45 @@ class ControlWin32Tests {
 		assertEquals(300, childShell.getSizeInPixels().x);
 	}
 
+	@Test
+	public void testAutoScaleDisabledProperty() {
+		Display display = Display.getDefault();
+		Shell shell = new Shell(display);
+		shell.setData("AUTOSCALE_DISABLED", true);
+		assertEquals(AutoscalingMode.DISABLED_INHERITED, shell.autoscalingMode);
+		Composite child = new Composite(shell, SWT.NONE);
+		assertEquals(AutoscalingMode.DISABLED_INHERITED, child.autoscalingMode);
+	}
+
+	@Test
+	public void testAutoScaleDisabled() {
+		Display display = Display.getDefault();
+		Shell shell = new Shell(display);
+		shell.setAutoscalingMode(AutoscalingMode.DISABLED);
+		assertEquals(AutoscalingMode.DISABLED, shell.autoscalingMode);
+		Composite child = new Composite(shell, SWT.NONE);
+		assertEquals(AutoscalingMode.ENABLED, child.autoscalingMode);
+	}
+
+	@Test
+	public void testPropagateAutoScaleDisabledProperty() {
+		Display display = Display.getDefault();
+		Shell shell = new Shell(display);
+		shell.setData("AUTOSCALE_DISABLED", true);
+		shell.setData("PROPOGATE_AUTOSCALE_DISABLED", false);
+		assertEquals(AutoscalingMode.DISABLED, shell.autoscalingMode);
+		Composite child = new Composite(shell, SWT.NONE);
+		assertEquals(AutoscalingMode.ENABLED, child.autoscalingMode);
+	}
+
+	@Test
+	public void testPropagateAutoScaleDisabled() {
+		Display display = Display.getDefault();
+		Shell shell = new Shell(display);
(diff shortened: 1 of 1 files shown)


tests/org.eclipse.swt.tests.win32/JUnit Tests/org/eclipse/swt/tests/win32/Win32AutoScaleTests.java
```diff
@@ -0,0 +1,71 @@
...
+	public void tearDown() {
+		shell.dispose();
+	}
+
+	@Test
+	public void testAutoScaleDisabledProperty() {
+		shell.setData("AUTOSCALE_DISABLED", true);
+		assertEquals(AutoscalingMode.DISABLED_INHERITED, SwtWin32TestUtil.getAutoscalingMode(shell));
+		Composite child = new Composite(shell, SWT.NONE);
+		assertEquals(AutoscalingMode.DISABLED_INHERITED, SwtWin32TestUtil.getAutoscalingMode(child));
```
ptziegler: I think this is what really confused me when debugging a related problem. I've double-checked this with how things used to behave prior to 2166ca67fa658a6cccdbd171a99b06938f67eb57 and that's it. It probably still makes sense to keep those tests, just to avoid any further confusion.

Phillipus:
If there is a `Control.setAutoscalingMode(mode)`then a corresponding `Control.getAutoscalingMode()` method would be useful. Was this a design decision? As you found, this would be a handy thing.

HeikoKlare:
> If there is a `Control.setAutoscalingMode(mode)`then a corresponding `Control.getAutoscalingMode()` method would be useful. Was this a design decision? As you found, this would be a handy thing.

When the autoscale disablement API was introduced, the existing `isAutoScalable()` method has been overwritten. Due to several side effects we found in the last few weeks, we reverted that change:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/3021

We though about providing a replacement method, next to the existing one, but that may obviously lead to confusion as there are two independent methods to retrieve some auto-scale-related state. For that reason, we chose to not provide such a method unless there is an actual use case for it. Do you have such?

Phillipus:
> unless there is an actual use case for it. Do you have such?

At the moment just to see what's going on to implement a workaround for https://github.com/eclipse-gef/gef-classic/issues/977

ptziegler:
To be honest, I don't see any benefit to adding a public getter method. And the only reason I haven't closed this one is because I think unit-tests for the propagation of the auto-scaling mode are useful.

Regarding https://github.com/eclipse-gef/gef-classic/issues/977, I don't think that a getter is of any use. The interesting part is the `nativeZoom` field of the `GC` object. But that, one can already get via `getGCData().nativeZoom`, at least for the sake of debugging.

HeikoKlare:
> To be honest, I don't see any benefit to adding a public getter method. And the only reason I haven't closed this one is because I think unit-tests for the propagation of the auto-scaling mode are useful.

Thank you for the feedback. If you come along a use case that requries a getter, we can still add it. I also consider the tests added here as useful though.

One possibility to avoid that reflective access is to make unit tests out of the added API-level tests: for Windows we have unit tests embedded in the SWT bundle/fragment itself, which can then access state and functionality that is not public API (such as the `autoscalingMode` field). @ptziegler do think the tests should stay API-level tests as they currently are (then we can merge this one as is) or should we transform them to unit tests without reflective access (I can do that and update the PR if you want).

ptziegler:
> One possibility to avoid that reflective access is to make unit tests out of the added API-level tests: for Windows we have unit tests embedded in the SWT bundle/fragment itself, which can then access state and functionality that is not public API (such as the autoscalingMode field).

Moving them to the bundle sounds like the cleaner approach. I've updated the PR and now they are part of the `ControlWin32Tests`.

bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/ControlWin32Tests.java
```diff
@@ -369,4 +369,45 @@ void testChildShellGetSize() {
 		assertEquals(300, childShell.getSizeInPixels().x);
 	}
 
+	@Test
+	public void testAutoScaleDisabledProperty() {
+		Display display = Display.getDefault();
+		Shell shell = new Shell(display);
```
ptziegler, line 375: If I understand this test case correctly then the Shell is automatically disposed after the test run, due to the `ResetMonitorSpecificScalingExtension`?

HeikoKlare: Indeed. The extension disposes the display and with it the shell.

Issue #3044, opened 2026-02-02
Test_org_eclipse_swt_custom_BusyIndicator fails on Linux


iloveeclipse:
Most recent fail: https://download.eclipse.org/eclipse/downloads/drops4/I20260201-1800/testresults/html/org.eclipse.swt.tests_ep439I-unit-linux-x86_64-java25_linux.gtk.x86_64_25.html

```
testShowWhile() timed out after 30 seconds

java.util.concurrent.TimeoutException: testShowWhile() timed out after 30 seconds
at java.base/java.util.ArrayList.forEach(ArrayList.java:1604)
at java.base/java.util.ArrayList.forEach(ArrayList.java:1604)
```

Looking at the code, could it be, the common thread pool that is used by `CompletableFuture.runAsync()` code in the test simply has no free executor threads to run, or that they don't have a free core to be executed? Note, our test instances run on VM's with limited CPU resources.

iloveeclipse:
@laeubi : could you please check ?

laeubi:
The common poll ensures a given concurrency that is the number of cores. If some threads are blocked new ones are spawned so blocking for new tasks should never occur.
It might be that some other error arise so it actually never completes.

iloveeclipse:
> The common poll ensures a given concurrency that is the number of cores. If some threads are blocked new ones are spawned so blocking for new tasks should never occur.

I just remember we had a problem in the past with code blocked by a common pool.
So let merge [that](https://github.com/eclipse-platform/eclipse.platform.swt/pull/3046) and see if this helps?

laeubi:
> I just remember we had a problem in the past with code blocked by a common pool.

Can you give a reference?

iloveeclipse:
> Can you give a reference?

Unfortunately no. I believe this was in platform, platform UI or jdt.

iloveeclipse:
OK, found it. 
See 
* https://github.com/eclipse-platform/eclipse.platform/commit/8271dce78b0afe09aa0b98c08201ea4e72b95b40 
* https://github.com/eclipse-jdt/eclipse.jdt.core/commit/d16aa5db4d3c4d7b0e8d2960fb4f255163b38397
* https://bugs.eclipse.org/bugs/show_bug.cgi?id=576646#c30

laeubi:
Reading through the bugs description it seem to be hindered by this

[...]

PR #3046, opened 2026-02-02
Sort out common pool as the possible root cause for test hangs

iloveeclipse:
The test `Test_org_eclipse_swt_custom_BusyIndicator.testShowWhile` hangs on every second build. Let exclude common thread pool as a possible root cause from equation.

If the test would continue to fail, this patch can be easily reverted.

See https://github.com/eclipse-platform/eclipse.platform.swt/issues/3044


iloveeclipse:
Interesting, previous fix version with `Executors.newSingleThreadExecutor()` reproducible hang forever on Linux (local execution is just fine). Let's try with `Executors.newFixedThreadPool(2)`.

laeubi:
I think a cached pool might even be suitable as well, we do not really care about how many threads used here.

iloveeclipse:
I wanted to be explicit for now, we need at least two threads, **if** the test will become stable, we could switch to whatever else or just forget about it.

iloveeclipse:
OK, the two thread pool seem to be fine also with limited resources. Merging.

Issue #3043, opened 2026-02-02
Widget is disposed on DPI change


laeubi:
I found this in the log of an RCP application and it seems to be related to the handling of a DPI change:

**SWT Version**: `org.eclipse.swt.win32.win32.x86_64_3.133.0.v20260122-1839`

I report it here because apart from that E4 is run the event queue its only SWT in the stack.

```
!ENTRY org.eclipse.e4.ui.workbench 4 0 2026-02-02 08:24:58.758
!MESSAGE Internal Error
!STACK 0
org.eclipse.swt.SWTException: Widget is disposed
        at org.eclipse.swt.SWT.error(SWT.java:4950)
        at org.eclipse.swt.SWT.error(SWT.java:4865)
        at org.eclipse.swt.SWT.error(SWT.java:4836)
        at org.eclipse.swt.widgets.Widget.error(Widget.java:510)
        at org.eclipse.swt.widgets.Widget.checkWidget(Widget.java:429)
        at org.eclipse.swt.widgets.TreeItem.getImage(TreeItem.java:770)
        at org.eclipse.swt.widgets.Item.handleDPIChange(Item.java:231)
        at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:91)
        at org.eclipse.swt.widgets.Display.sendEvent(Display.java:4361)
        at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1217)
        at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1241)
        at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1226)
        at org.eclipse.swt.widgets.Widget.notifyListeners(Widget.java:863)
        at org.eclipse.swt.widgets.TreeItem.handleDPIChange(TreeItem.java:1829)
        at org.eclipse.swt.widgets.Widget.lambda$0(Widget.java:194)
        at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:91)
        at org.eclipse.swt.widgets.Display.sendEvent(Display.java:4361)
        at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1217)
        at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1241)
        at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1226)
        at org.eclipse.swt.widgets.Widget.notifyListeners(Widget.java:863)
[...]

PR #3047, opened 2026-02-03
[Win32] Check widget for disposal before zoom change notification

HeikoKlare:
In most scenarios, zoom change processing happens asynchronously. In consequence, widgets may evolve and even already be disposed until a notification for a zoom change event on them actually is processed. This can lead to disposed widget exceptions in case such a notification is sent on an already disposed widget.

This change adapts all places that send zoom change notification on DPI change to check for disposal of the widget to send the notification to first.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/3043

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CCombo.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CCombo.java
@@ -2039,17 +2039,16 @@ public boolean traverse(int event){
 }
 
 private void handleDPIChange(Event event) {
-	if (text != null) {
-		text.notifyListeners(SWT.ZoomChanged, event);
-	}
-	if (list != null) {
-		list.notifyListeners(SWT.ZoomChanged, event);
-	}
-	if (arrow != null) {
-		arrow.notifyListeners(SWT.ZoomChanged, event);
-	}
-	if (popup != null) {
-		popup.notifyListeners(SWT.ZoomChanged, event);
+	forwardDPIChange(event, text);
+	forwardDPIChange(event, list);
+	forwardDPIChange(event, arrow);
+	forwardDPIChange(event, popup);
+}
+
+private void forwardDPIChange(Event event, Widget widget) {
+	if (widget != null && !widget.isDisposed()) {
+		widget.notifyListeners(SWT.ZoomChanged, event);
 	}
 }
+
 }
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/StyledText.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/StyledText.java
@@ -10932,9 +10932,11 @@ private void handleDPIChange(Event event) {
 			caretSet.add(caret);
 		}
 	}
-	caretSet.stream().filter(Objects::nonNull).forEach(caretToRefresh -> {
-		caretToRefresh.notifyListeners(SWT.ZoomChanged, event);
-	});
+	for (Caret caretToRefresh : caretSet) {
+		if (caretToRefresh != null && !caretToRefresh.isDisposed()) {
+			caretToRefresh.notifyListeners(SWT.ZoomChanged, event);
+		}
+	}
 }
 
 @Override
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/CoolBar.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/CoolBar.java
@@ -1222,7 +1222,7 @@ void handleDPIChange(Event event, float scalingFactor) {
 		CoolItem item = items[index];
 
 		Control control = item.control;
-		if (control != null) {
+		if (control != null && !control.isDisposed()) {
 			control.notifyListeners(SWT.ZoomChanged, event);
 			item.setControl(control);
 		}
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Decorations.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Decorations.java
@@ -1710,13 +1710,13 @@ void handleDPIChange(Event event, float scalingFactor) {
 	super.handleDPIChange(event, scalingFactor);
 
 	Menu menuBar = getMenuBar();
-	if (menuBar != null) {
+	if (menuBar != null && !menuBar.isDisposed()) {
 		menuBar.notifyListeners(SWT.ZoomChanged, event);
 	}
 
 	if (menus != null) {
 		for (Menu menu : menus) {
-			if (menu != null) {
+			if (menu != null && !menu.isDisposed()) {
 				menu.notifyListeners(SWT.ZoomChanged, event);
 			}
 		}
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/ExpandBar.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/ExpandBar.java
@@ -873,7 +873,9 @@ LRESULT wmScroll (ScrollBar bar, boolean update, long hwnd, int msg, long wParam
 void handleDPIChange(Event event, float scalingFactor) {
 	super.handleDPIChange(event, scalingFactor);
 	for (ExpandItem item : getItems()) {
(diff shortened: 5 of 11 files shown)


laeubi:
Thanks for providing a fix for that!

> This can lead to disposed widget exceptions in case such a notification is sent on an already disposed widget.

I'm not sure its relevant here but I mostly see it during remote-desktop sessions when connecting from different computers. So in this particular case it might be that the image needs to be refreshed.

HeikoKlare:
> So in this particular case it might be that the image needs to be refreshed.

As a DPI change event seems to be triggered in that case (according to the stack trace with the disposed error), images should automatically be refreshed via the DPI change handlers. If you notice that anything is wrong with the images, please let me know.

laeubi:
It all seems to work except that ones get an error dialog on DPI change, after dismiss the dialog everything is working.

HeikoKlare:
> It all seems to work except that ones get an error dialog on DPI change, after dismiss the dialog everything is working.

The dialog pops up because of the error reported in https://github.com/eclipse-platform/eclipse.platform.swt/issues/3043, or is there anything else?

laeubi:
> > It all seems to work except that ones get an error dialog on DPI change, after dismiss the dialog everything is working.
> 
> The dialog pops up because of the error reported in #3043, or is there anything else?

Yes it seems to be a standard error dialog of Eclipse (with just text "Widget Disposed") then one can dismiss and application seem to work (but maybe have missed some updates of course). It might be become visible now the automatic scaling is the default?

HeikoKlare:
> It might be become visible now the automatic scaling is the default?

That might be the case, yes. If I understand the description in the issue correctly, it's a pure E4 application, which was (by accident) not affected by the default enablement via preference we had introduced more than half a year ago. So it might be affected by the recent change of SWT to use monitor-specific scaling by default.

The arising question is of course why there is a disposed TreeItem at all. It might be that this disposed item has been there by accident already before, but that it did not lead to any issues as no one has accessed it. That changed now with the DPI change processing being applied and accessing the item, which reveals that disposed state.

laeubi:
> The arising question is of course why there is a disposed `TreeItem` at all. 

SWT has the concept of virtual Table/Tree (using `SWT.VIRTUAL`), in such case there is a model an items are only created when the actual thing becomes visible. It might be things get disposed as well ... e.g. because of the DPI change more/less becomes visible, but this is only a guess!

HeikoKlare:
Sounds like a reasonable guess, but not that easy to verify. And assuming it's correct, just ignoring disposed elements for updates on DPI changes seems fine to me. So should we go with this change? You could then report if the error dialog still pops up for you or if you face any other issues with the trees in your application afterwards.

laeubi:
Yes I think just ignore them would make sense here.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CCombo.java
```diff
@@ -2039,16 +2039,16 @@ public boolean traverse(int event){
...
 	}
-	if (arrow != null) {
+	if (arrow != null && !arrow.isDisposed()) {
 		arrow.notifyListeners(SWT.ZoomChanged, event);
 	}
-	if (popup != null) {
+	if (popup != null && !popup.isDisposed()) {
 		popup.notifyListeners(SWT.ZoomChanged, event);
 	}
 }
```
fedejeanne, line 2052: You could extract it to a method to avoid duplicating the logic and (future) copy-paste errors:

```java
private void handleDPIChange(Event event) {
	handleDPIChange(event, text);
	handleDPIChange(event, list);
	handleDPIChange(event, arrow);
	handleDPIChange(event, popup);
}

private void handleDPIChange(Event event, Widget widget) {
	if (widget != null && !widget.isDisposed()) {
		widget.notifyListeners(SWT.ZoomChanged, event);
	}
}
```

fedejeanne (commented):
The code looks good, I just have some questions:

1. Doesn't this call also need a check?

https://github.com/eclipse-platform/eclipse.platform.swt/blob/c891c43ad54f204d05c9b9f7faf06d99365f9c8d/bundles/org.eclipse.swt/Eclipse%20SWT/win32/org/eclipse/swt/widgets/ToolBar.java#L1775

2. Would it make sense to check for `event.doit` like it's being done in `Control::sendZoomChangeEvent` just in case someone decides to stop the propagation (for whatever reason)?

https://github.com/eclipse-platform/eclipse.platform.swt/blob/c891c43ad54f204d05c9b9f7faf06d99365f9c8d/bundles/org.eclipse.swt/Eclipse%20SWT/win32/org/eclipse/swt/widgets/Control.java#L6121

3. Doesn't the call in `Shell` also need a check? I suppose shell is a special case but I just want to check:

https://github.com/eclipse-platform/eclipse.platform.swt/blob/c891c43ad54f204d05c9b9f7faf06d99365f9c8d/bundles/org.eclipse.swt/Eclipse%20SWT/win32/org/eclipse/swt/widgets/Shell.java#L2534

HeikoKlare:
Thank you for the feedback. The proposed extraction into a separate method definitely makes sense. I have adapted the PR accordingly.

> 1. Doesn't this call also need a check?
> 
> https://github.com/eclipse-platform/eclipse.platform.swt/blob/c891c43ad54f204d05c9b9f7faf06d99365f9c8d/bundles/org.eclipse.swt/Eclipse%20SWT/win32/org/eclipse/swt/widgets/ToolBar.java#L1775

I thought about whether we need a check here as well. The reason why I didn't add it is because the code around that refreshes (or rather recreates) the tool button would actually fail if the item was disposed. When taking a deeper look, I even think that this line of code might be completely useless. The buttons are recreated anyway and the `handleDPIChange` implementation of `Item` has no effect on an `ToolItem` anyway (as setting the same image leads to an early return). Thus I would rather completely remove this line, but for the sake of low risk would wait with that for 2026-06 M1 as keeping the line for now does not hurt and there might still be a risk of unexpected regressions.

> 2. Would it make sense to check for `event.doit` like it's being done in `Control::sendZoomChangeEvent` just in case someone decides to stop the propagation (for whatever reason)?
> 
> https://github.com/eclipse-platform/eclipse.platform.swt/blob/c891c43ad54f204d05c9b9f7faf06d99365f9c8d/bundles/org.eclipse.swt/Eclipse%20SWT/win32/org/eclipse/swt/widgets/Control.java#L6121

`event.doIt` is only (supposed to be) set to `false` by a concurrent DPI change event (see `Shell.handleMonitorSpecificDpiChange()`). Since the `notifyListeners` calls happen synchronously within the UI thread, there is no change that a concurrent DPI change event comes in and sets the `event.doIt` flag to `false` while synchronously processing a change. So the evaluation of `event.doIt` is only necessary when event queue processing has happened (i.e., when processing an asynchronous DPI change event), which happens in `sentZoomChangeEvent` where that check is already present.

> 3. Doesn't the call in `Shell` also need a check? I suppose shell is a special case but I just want to check:
> 
> https://github.com/eclipse-platform/eclipse.platform.swt/blob/c891c43ad54f204d05c9b9f7faf06d99365f9c8d/bundles/org.eclipse.swt/Eclipse%20SWT/win32/org/eclipse/swt/widgets/Shell.java#L2534

This method is only called when a WM_DPICHANGED event is triggered on the shell, which can only happen if the shell is not disposed.

fedejeanne:
Thank you for the clarifications!

> I thought about whether we need a check here as well. The reason why I didn't add it is because the code around that refreshes (or rather recreates) the tool button would actually fail if the item was disposed. When taking a deeper look, I even think that this line of code might be completely useless. The buttons are recreated anyway and the `handleDPIChange` implementation of `Item` has no effect on an `ToolItem` anyway (as setting the same image leads to an early return). Thus I would rather completely remove this line, but for the sake of low risk would wait with that for 2026-06 M1 as keeping the line for now does not hurt and there might still be a risk of unexpected regressions.

Makes sense to postpone the change to M1 to avoid surprises if the current state doesn't hurt 👍 



> `event.doIt` is only (supposed to be) set to `false` by a concurrent DPI change event (see `Shell.handleMonitorSpecificDpiChange()`). Since the `notifyListeners` calls happen synchronously within the UI thread, there is no change that a concurrent DPI change event comes in and sets the `event.doIt` flag to `false` while synchronously processing a change. So the evaluation of `event.doIt` is only necessary when event queue processing has happened (i.e., when processing an asynchronous DPI change event), which happens in `sentZoomChangeEvent` where that check is already present.

Ah, ok. I was mistakenly assuming that the changes were split into several smaller async calls and that there was a (near-zero) chance that some newer DPI change event would happen while processing the older DPI change event.



> This method is only called when a WM_DPICHANGED event is triggered on the shell, which can only happen if the shell is not disposed.

Makes sense 👍

fedejeanne:
All good from my side. 

@laeubi may we merge this one?

laeubi:
You both have commit rights here and as there is no disagreement and its fixing a bug I think it can be merged as soon as you like.

PR #1266, opened 2024-06-07
Add support to display the busy indicator while a Future is running

laeubi:
While the methods to perform an operation with busy indication are already a great help to write responsive UI there is one case missing where a (Completable)Future is already present, this can for example be used to cache some heavy work.

This now adds a new BusyIndicator.showWhile(Future) method that covers the use case of an externally provided future computation so the caller can be sure that while it takes place the UI shows an indicator.

Beside that the Snippet30 was enhanced to demonstrate the new function.

PR #3050, opened 2026-02-03
Assume minimal CPU configuration for the BusyIndicator test on Linux

iloveeclipse:
The test never fail locally with 32 cores, fails every second time on smaller VM configs. Let assume at least 4 cores on Linux.

See https://github.com/eclipse-platform/eclipse.platform.swt/issues/3044

### DIFF
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_custom_BusyIndicator.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_custom_BusyIndicator.java
@@ -16,6 +16,7 @@ package org.eclipse.swt.tests.junit;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotEquals;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assumptions.assumeTrue;
 
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CountDownLatch;
@@ -37,6 +38,11 @@ public class Test_org_eclipse_swt_custom_BusyIndicator {
 	@Test
 	@Timeout(value = 30)
 	public void testShowWhile() {
+		if (SwtTestUtil.isLinux) {
+			int availableProcessors = Runtime.getRuntime().availableProcessors();
+			assumeTrue(availableProcessors >= 4,
+					"Need at least 4 CPUs for this test to be reliable on Linux: " + availableProcessors);
+		}
 		// Executors.newSingleThreadExecutor() hangs on some Linux configurations
 		try (ExecutorService executor = Executors.newFixedThreadPool(2)){
 			Shell shell = new Shell();


tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_custom_BusyIndicator.java
```diff
@@ -37,6 +38,11 @@ public class Test_org_eclipse_swt_custom_BusyIndicator {
 	@Test
 	@Timeout(value = 30)
 	public void testShowWhile() {
+		if (SwtTestUtil.isLinux) {
```
laeubi, line 41: This sound completely off, why should the number of CPU core matter?!? Java can have 100 of parallel threads on a single CPU...

iloveeclipse: Do you have a better idea why the test hangs?
Let try that please. I'm tired to see the test failing every second day.

iloveeclipse:
OK, eclipse.org jenkins runs on smaller config with probably 2 cores, but github job has at least 4 cores (or assumption doesn't work) and is still hanging around on that test, which is strange, because it should be stopped by JUnit Timeout annotation, which seem not to work too on Jenkins (but works on SDK build, which means there is also test setup issue).

@laeubi :  do you prefer:

1) disable this test on Linux (as it seem to be Linux specific issue).
2) assume 8 cores (so it will run on most "real" configurations but not on test machines.
3) analyze and debug the root cause of the test hang and test failure.

My goal is to have a clean build state without hangs or failures, so people can only see test failures related to the actual code changes.

I don't have time for 3) and also feel myself not responsible to do that, as I didn't added this test. 
So 1) or 2) is my current proposal for which I can spend time.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_custom_BusyIndicator.java
laeubi: So I do not have a theory here, but as I'm not aware it fails since its creation in Jun 23, 2024 it might be a bug/regression in GTK part of Linux.

Test is not really complex and the assumption it is in any way related to the number of CPUs does not hold true. Also why should it only fail on Linux then?

As you already noticed it even succeed on Hardware with less course, so any arbitrary number here will simply disable the test and hiding a possible bug.

laeubi:
> also feel myself not responsible to do that, as I didn't added this test

Just for the record the test was [explicitly requested by reviewers](https://github.com/eclipse-platform/eclipse.platform.swt/pull/1266#pullrequestreview-2107252853) so I feel at least to share the responsibility for this with others.

Anyways, I can try to make the timeout and debugging for it a bit better if it helps...

> but github job has at least 4 cores (or assumption doesn't work) and is still hanging around on that test

Github has 4 CPUs, at least it seem to me to hang in GTK4 in the Github actions only so maybe @jonahgraham @akurtakov have a clue what might be a reason here as well.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_custom_BusyIndicator.java
laeubi: See https://github.com/eclipse-platform/eclipse.platform.swt/pull/3051

iloveeclipse:
OK, looks like you have some ideas where the problem is coming from, so closing this PR.

PR #3061, opened 2026-02-05
Let Display#sleep() react to thread interruption

### LINKED ISSUE #3059: sleep() does not react to thread interruption
In the context of
- https://github.com/eclipse-platform/eclipse.platform.swt/issues/3044

I found that `Display#sleep()` does not react to thread interruption. That is also likely the reason why the test timeout is not effective! So if the main thread is hanging (for whatever reason) in a sleep, it can't be interrupted and as there is no other way to reliable end a thread in java the application hangs forever.

laeubi:
Fix https://github.com/eclipse-platform/eclipse.platform.swt/issues/3059

I did not found a good way on Windows (@HeikoKlare ?) and would like to get feedback on the MacOS part, maybe @akurtakov / @jonahgraham can take a peek on the gtk side?

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
@@ -5667,11 +5667,11 @@ public boolean sleep () {
 			OS.g_main_context_check (context, max_priority [0], fds, nfds);
 			OS.g_main_context_release (context);
 		}
-	} while (!result && synchronizer.isMessagesEmpty() && !wake);
+	} while (!result && synchronizer.isMessagesEmpty() && !wake && !thread.isInterrupted());
 	wake = false;
 	if (!GTK.GTK4) GDK.gdk_threads_enter ();
 	sendPostExternalEventDispatchEvent ();
-	return true;
+	return !thread.isInterrupted();
 }
 
 /**
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Display.java
@@ -5211,13 +5211,19 @@ public boolean sleep () {
 	try {
 		addPool();
 		allowTimers = runAsyncMessages = false;
-		NSRunLoop.currentRunLoop().runMode(OS.NSDefaultRunLoopMode, NSDate.distantFuture());
+		/*
+		 * Use a timeout-based approach to allow checking for thread interruption.
+		 * Sleep for a maximum of 50 milliseconds at a time, similar to GTK.
+		 */
+		do {
+			NSRunLoop.currentRunLoop().runMode(OS.NSDefaultRunLoopMode, NSDate.dateWithTimeIntervalSinceNow(0.05));
+		} while (synchronizer.isMessagesEmpty() && !thread.isInterrupted());
 		allowTimers = runAsyncMessages = true;
 	} finally {
 		removePool();
 	}
 	sendPostExternalEventDispatchEvent ();
-	return true;
+	return !thread.isInterrupted();
 }
 
 int sourceProc (int info) {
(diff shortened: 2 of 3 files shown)


HeikoKlare:
I did not find any OS API method that allows to wait for a message until a timeout occurs instead of the inifinitely blocking to `OS.WaitMessage()`.
Maybe something like this replacing the `OS.WaitMessage()` call  in `Display.sleep()`could work (just as a conceptual idea, not a clean solution):
```java
	boolean messageReceived = false;
	while (!messageReceived && !thread.isInterrupted()) {
		messageReceived = OS.PeekMessage (msg, 0, 0, 0, OS.PM_NOREMOVE);
		try {
			Thread.sleep(20);
		} catch (InterruptedException e) {
			thread.interrupt();
			break;
		}
	}
```
At least the test locally passes with it.

laeubi:
> just as a conceptual idea, not a clean solution

Thanks for looking into it, the problem is that it will immediately return but always with 20ms delay (what might be too much), I think we can't do any better then on windows for the moment.

jonahgraham:
On win32 I think you can leave in waitmessage, but always queue a 50ms timer so that you never wait more than 50ms. I don't think it is particularly trivial to do that though as ideally you want the sleep method to handle that particular timer going off locally rather than waiting for the main event loop to handle it.

The gtk side looks good to me on code inspection, but I don't know the implications of changing return value, for example in cases where there are two conditions that caused loop to exit, should it still return true. I don't see any real uses of the sleep return value so far in the code base, so not enough of an issue for me to want to spend more time on it. I guess its too bad we can't throw InterruptedException as that would be a very disruptive API change.

HeikoKlare:
> On win32 I think you can leave in waitmessage, but always queue a 50ms timer so that you never wait more than 50ms. I don't think it is particularly trivial to do that though as ideally you want the sleep method to handle that particular timer going off locally rather than waiting for the main event loop to handle it.

That's a nice idea but you would indeed need to process the timer event inside the `sleep()` method.

I shortly experimented with a simple timer and a timer callback and both of them do not seem to be processed by the event queue afterwards if the timer is killed after it elapsed (and thus made `OS.WaitMessage()` return). But I do not find anything in the Windows documentation that guarantees this behavior.
```java
long wakeupId = OS.SetTimer(0, 0, 50, 0);
boolean result = OS.WaitMessage();
OS.KillTimer(0, wakeupId);
```

```java
long timer = WakeupMessage.sendAfter(50);
boolean result = OS.WaitMessage();
OS.KillTimer(0, timer);

...

private static class WakeupMessage  {
	private final Callback osCallback;

	private WakeupMessage() {
		// Has to have TIMERPROC signature, see https://learn.microsoft.com/en-us/windows/win32/api/winuser/nc-winuser-timerproc
		osCallback = new Callback(this, "run", void.class, new Type[] { int.class, int.class, int.class, int.class} );
	}

	@SuppressWarnings("unused") // Executed as callback method referenced by signature description
	public void run(int hwnd, int msg , int idEvent, int dwTime) {
		OS.KillTimer(hwnd, idEvent);
		System.out.println("lalala");
		osCallback.dispose();
	}

	static long sendAfter(int millis) {
		try {
			WakeupMessage fakeMessage = new WakeupMessage();
			return OS.SetTimer(0, 0, millis, fakeMessage.osCallback.getAddress());
		} catch  (SWTError error) {
		}
		return 0;
	}
}
```

I now found that is the OS method `MsgWaitForMultipleObjectsEx` that is proposed when wanting to have a timeout while waiting for messages to arrive. However, there are several configuration options for that method via the parameters of which I would not be immediately sure how to use them.

laeubi:
I just wanted to mention that I don't want to overcomplicate the matter here and if its not working on windows it is fine for me
The same applies to the return value, I have never seen any use of that and it always returns true unless a mysterious "config change" in gtk.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
trancexpress, line 5663: Isn't the problem setting `wake` to `true` before this line with "bad" timing?

I wonder what this line is for... @akurtakov any idea? Maybe it was from a time before the timeout above was set? It seems very odd that `Display.wake()` would work only if timed to be within the looping 50 ms sleep...

akurtakov: TBH, I have no clue . This code has been there since the famous SWT HEAD reset commit so no clue where/why/how/what it is there for.

HeikoKlare: If I am not mistaken, the `wake = false` line was introduced together with the timeout in this commit: https://github.com/eclipse-platform/eclipse.platform.swt/commit/400a41972553b6a4188a913a649f045d7101753f

PR #3062, opened 2026-02-05
[Gtk] Stop passing widget handle needlessly in Caret

akurtakov:
Widget handle is not used at all thus it only creates confusion.


akurtakov:
Noticed while looking into https://github.com/eclipse-platform/eclipse.platform.swt/issues/2812

PR #3069, opened 2026-02-09
[Gtk4] Fix draw order

akurtakov:
Snapshot children before applying custom draw.


akurtakov:

Issue #3075, opened 2026-02-09
[Windows] [Tree] Column icons and text overlaps in multi resolution monitor setup


glasedom:
Eclipse: 4.38
OS: Windows 11
Monitor Setup: Primary Monitor 100% (1920x1080), Second Monitor 150% (3840x2160)

In our RCP application we have some views and editors where the icons and text of tree cells are overlapped in a monitor setup with multiple resolutions.
I have created a mini snippet that demonstrates the problem.

- [x] I verified I can reproduce this issue against [latest Integration Build of Eclipse SDK](https://download.eclipse.org/eclipse/downloads/) (eclipse-SDK-I20260208-1140-win32-x86_64)

## Steps to reproduce

Open the given sample program.
The tree looks like this:

<img width="586" height="143" alt="Image" src="https://github.com/user-attachments/assets/23da768c-2cba-408c-8826-7e6b5d0eec10" />

Move the window from the primary monitor to the second monitor.
You can see that icon and text overlaps in the second column.

<img width="882" height="216" alt="Image" src="https://github.com/user-attachments/assets/1163f4a3-8612-4f6f-a88f-38c665aeb09e" />

If you move the window back to the primary monitor the icon and text still overlaps.

<img width="586" height="143" alt="Image" src="https://github.com/user-attachments/assets/381d7cdf-c545-47f5-80de-6b1367809a0c" />

 Here is the sample snippet:
```
import java.io.File;

import org.eclipse.jface.action.Action;
import org.eclipse.jface.resource.ImageDescriptor;
import org.eclipse.jface.resource.JFaceResources;
import org.eclipse.jface.viewers.ArrayContentProvider;
import org.eclipse.jface.viewers.DelegatingStyledCellLabelProvider;
import org.eclipse.jface.viewers.DelegatingStyledCellLabelProvider.IStyledLabelProvider;
import org.eclipse.jface.viewers.ITreeContentProvider;
import org.eclipse.jface.viewers.LabelProvider;
import org.eclipse.jface.viewers.StyledString;
import org.eclipse.jface.viewers.TreeViewer;
import org.eclipse.jface.viewers.TreeViewerColumn;
import org.eclipse.swt.SWT;
import org.eclipse.swt.graphics.Image;
import org.eclipse.swt.layout.FillLayout;
import org.eclipse.swt.widgets.Display;
import org.eclipse.swt.widgets.Shell;
import org.eclipse.swt.widgets.TreeColumn;

[...]

PR #3076, opened 2026-02-11
[Win32] Update images of tree columns upon DPI change

HeikoKlare:
When moving a shell with a tree that uses images in any other than the first column to another monitor, those images will be lost in case the other monitor has a different zoom such that the UI is rescaled. The reason is that the DPI change handling for TreeItem does not consider the other images of the tree item at all.

This change explicitly resets all images of a tree item upon DPI change.

Contributes to https://github.com/eclipse-platform/eclipse.platform.swt/issues/3075

### Remaining limitation
One limitation is that if an item does not have an image set for the first column, there will be blank space where the image could be after rescaling. This will be fixed independently via:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/3077

Before rescaling:
image

After rescaling:
image

This is, however, an issue that already happens now (without any rescaling) in case the item has a subitems that uses an image in the first column:
image
And it also happens if you set and then remove an image, like this:
```java
item.setImage(imageColumn1);
item.setImage((Image) null);
```
image

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/TreeItem.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/TreeItem.java
@@ -1815,6 +1815,11 @@ String getNameText () {
 @Override
 void handleDPIChange(Event event, float scalingFactor) {
 	super.handleDPIChange(event, scalingFactor);
+	if (images != null) {
+		for (int i = 1; i < images.length; i++) {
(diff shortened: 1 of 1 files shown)


HeikoKlare:
Merging this as it obviously fixes the missing image update (analogous to an Item's image via `Item#handleDPIChange`), was verified by the issue creator (https://github.com/eclipse-platform/eclipse.platform.swt/issues/3075#issuecomment-3885041517) and should go into M3 for some testing before the release.

Issue #3078, opened 2026-02-11
Windows: SWTException "Invalid image" when rasterizing SVG at size 0 after moving Eclipse window between monitors (DPI change)


oberlehner:
**Describe the bug**
On Windows, after a DPI/scaling change (e.g., moving an Eclipse/SWT application window between monitors with different scaling, or in VM/RDP-like setups), UI painting can throw an unhandled event loop exception. SWT attempts to rasterize an SVG image for a requested size where the computed width or height becomes 0 (or negative). The SVG rasterizer rejects this and throws `IllegalArgumentException`, which bubbles up as `org.eclipse.swt.SWTException: Invalid image`, disrupting painting and sometimes leaving the UI in a broken state.

**To Reproduce**
1) Start an Eclipse-based app that uses SVG icons and renders scaled images during paint (common in editors/views, including those using Draw2D/GEF).

2)Open a view/editor that triggers frequent repaints (e.g., zoomable canvas, thumbnail/overview, outline, etc.).

3)Move the application window between monitors with different DPI scaling (e.g., 100% ↔ 150%/200%), or change scaling while the app is running.

4)Trigger repaint (open/activate editor, zoom, resize window).

5)Observe exception.

**Environment:**

- App: 4diac IDE 3.1.0.qualifier
- Eclipse Release Train: 2026-03 (Platform 4.39)
- SWT: 4.39 (org.eclipse.swt ~3.133.0… in 4.39 stream)
- Java: 21 (required)
- VM args: -Dswt.autoScale=quarter
- OS: Windows (see stack trace shows Win32 DPI utils)

```
org.eclipse.swt.SWTException: Invalid image (java.lang.IllegalArgumentException: Argument not valid [Cannot rasterize SVG for width or height <= 0])
	at org.eclipse.swt.SWT.error(SWT.java:4950)
	at org.eclipse.swt.SWT.error(SWT.java:4865)
	at org.eclipse.swt.internal.image.FileFormat.loadFromStreamBySize(FileFormat.java:140)
	at org.eclipse.swt.internal.image.FileFormat.load(FileFormat.java:167)
	at org.eclipse.swt.internal.NativeImageLoader.load(NativeImageLoader.java:30)
	at org.eclipse.swt.graphics.ImageLoader.loadBySize(ImageLoader.java:169)
	at org.eclipse.swt.graphics.ImageLoader.loadBySize(ImageLoader.java:215)
[...]

PR #3079, opened 2026-02-12
Avoid attempt to draw images with target width/height of 0

HeikoKlare:
Recently, the ability to scale images for the desired target size of a drawImage operation executed on a GC and the incorporation of a Transform potentially applied to that GC have been implemented. With this, the target size for an image to be drawn can become 0 and the request of a handle of that size can fail, in particular when trying to rasterize an SVG at that size.

This change adapts the implementation of the handle retrieval for a desired size in the Image implementation so that the minimum size of the provided handle is 1 in both of the dimensions. The draw operation in the GC will then take care of potentially not drawing anything at all when the actual intended size is really 0.
It also adds a test case for the specific case of requesting a handle at size 0 and also adds a test for the general correctness of image handle retrieval when a transform is applied to the GC.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/3078


HeikoKlare:
Consequences of the addressed issue (failing GC operations) are potentially severe, such that this issue should be fixed for the release. In order to have it tested with the last milestone build to detect potential unintended effects, this should go in M3. Unintended effects are unlikely though as the issue is about the intent to draw an image at width or hight 0, such that the actual image content is irrelevant anyway.

Issue #876, opened 2023-11-10
org.eclipse.swt.tests.junit.Test_org_eclipse_swt_widgets_Text failing in Windows


elsazac:
**Describe the bug**

[org.eclipse.swt.tests.junit](https://download.eclipse.org/eclipse/downloads/drops4/I20231109-0710/testresults/html/org.eclipse.swt.tests_ep430I-unit-win32-java17_win32.win32.x86_64_17.html#org.eclipse.swt.tests.junit) failing in win32.x86_64

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 

```
expected:<[]> but was:<[01567890]>

org.junit.ComparisonFailure: expected:<[]> but was:<[01567890]>
at org.junit.Assert.assertEquals(Assert.java:117)
at org.junit.Assert.assertEquals(Assert.java:146)
at org.eclipse.swt.tests.junit.Test_org_eclipse_swt_widgets_Text.test_cut(Test_org_eclipse_swt_widgets_Text.java:378)
at java.base/java.lang.reflect.Method.invoke(Method.java:568)
```

SyntevoAlex:
Duplicate of https://github.com/eclipse-platform/eclipse.platform.swt/issues/106

deepika-u:
Can this also be closed?
As i see 
<testcase classname="org.eclipse.swt.tests.junit.Test_org_eclipse_swt_widgets_Text" name="test_cut" time="0.015"/>

=> https://download.eclipse.org/eclipse/downloads/drops4/I20250108-1800/testresults/xml/org.eclipse.swt.tests_ep435I-unit-win32-x86_64-java21_win32.win32.x86_64_21.xml

PR #3081, opened 2026-02-14
Improve stability of Text cut/copy/paste tests on Windows

HeikoKlare:
The tests for cut/copy/paste functionality of Text widgets in Test_org_eclipse_swt_widgets_Text sporadically fail on Windows. Locally, they fail on almost every execution (for me). This seems to either be caused by some timing issue at the OS or by the need to spin the event queue for processing the cut/copy/paste operations at the OS.

In some tests, according event processing for spinning the event loop and waiting for events have been implemented already for GTK. This change adapts all calls to text.cut()/copy()/paste() such that the event queue is spinned. For me, it fixes all local test failures of that test class.

This fix attempt was triggered by the according tests spordically but frequently failing in I-Build test execution, such as:
https://download.eclipse.org/eclipse/downloads/drops4/I20260213-1800/testresults/html/org.eclipse.swt.tests_ep439I-unit-win32-x86_64-java21_win32.win32.x86_64_21.html

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/106

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/876

### DIFF
--- a/tests/org.eclipse.swt.tests/META-INF/MANIFEST.MF
+++ b/tests/org.eclipse.swt.tests/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: Eclipse SWT Tests
 Bundle-SymbolicName: org.eclipse.swt.tests
-Bundle-Version: 3.108.0.qualifier
+Bundle-Version: 3.108.100.qualifier
 Bundle-Vendor: Eclipse.org
 Export-Package: org.eclipse.swt.tests.junit,
  org.eclipse.swt.tests.junit.performance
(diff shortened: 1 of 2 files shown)


HeikoKlare:
@jonahgraham you have been working on the Clipboard and according tests quite heavily. What do think about this adaptation of the tests? Do you see any risk? It significantly improves (or even fix) the test behavior on Windows for me, but I would like to avoid that it, e.g., breaks something for GTK.

jonahgraham (approved):
This is ok by me- it aligns well with the change of official API in SWT too - see https://github.com/eclipse-platform/eclipse.platform.swt/issues/2598 and some of its history for more details. This PR also does more of what I had done in https://github.com/eclipse-platform/eclipse.platform.swt/pull/2668

PR #3082, opened 2026-02-14
Avoid compatibility check failure for Uber JARs and Universal Binaries

### LINKED ISSUE #2928: Library.isLoadable() failing unexpectedly
`Library.isLoadable()` fails when `SWT-OS` or `SWT-Arch` manifest entries are missing, with error:
```
Libraries for platform win32 cannot be loaded because of incompatible environment
```

This breaks uber JARs (where build tools lose manifest entries during merge) and universal binaries (where a single manifest can't specify multiple architectures). 

I am including a proposed a simple fix to clarify the error message and work under these conditions. 

**To Reproduce**

**Scenario 1: Uber JAR**
1. Create Maven project with SWT dependency
2. Build with maven-assembly-plugin without preserving manifest entries:
```xml
<plugin>
    <groupId>org.apache.maven.plugins</groupId>
    <artifactId>maven-assembly-plugin</artifactId>
    <configuration>
        <descriptorRefs>
            <descriptorRef>jar-with-dependencies</descriptorRef>
        </descriptorRefs>
    </configuration>
</plugin>
```
3. Run the resulting JAR  (java -jar xyz-jar-with-dependencies.jar) 
4. Application exits with "incompatible environment" error

**Scenario 2: Universal Binary**
1. Create macOS universal binary by combining x86_64 and aarch64 libraries with `lipo`
2. Package with single manifest (can only specify one `SWT-Arch` value)
3. Run on the non-manifest architecture
4. Application exits with "incompatible environment" error

**Expected behavior**

The check should:
- **Allow** when manifest entries are missing (uber JAR or universal binary scenario)
- **Fail** when manifest entries are present but mismatched (actual wrong platform JAR)
- **Provide** clear diagnostic message explaining the mismatch

This distinguishes between acceptable scenarios (missing entries) and actual errors (wrong platform).

**Environment**

1. Platform:
   - [x] All OS
   - [x] Windows
   - [x] Linux
   - [x] macOS

2. Additional OS info: Affects all platforms equally

3. JRE/JDK version: All versions

**Version since**

PR #2054 and #2026 cc @HeikoKlare 
Introduced in commit 360a2702a7 (April 23, 2025) - likely Eclipse 4.33 or later

**Workaround**

**For uber JARs:** Configure build tool to preserve manifest entries:
```xml
<archive>
    <manifestEntries>
        <SWT-OS>win32</SWT-OS>
        <SWT-Arch>x86_64</SWT-Arch>
    </manifestEntries>
</archive>
```

**For universal binaries:** No workaround available (manifest can only contain single architecture)
I've been using universal binaries for Mac for a long time -- as the .classes are identical and it is pretty trivial to use lipo to merge aarch64 and intel.

**Proposed Fix**

Location: `bundles/org.eclipse.swt/Eclipse SWT PI/common/org/eclipse/swt/internal/Library.java` lines 214-222

Current code:
```java
String manifestOS = attributes.getValue("SWT-OS");
String manifestArch = attributes.getValue("SWT-Arch");
if (arch.equals(manifestArch) && os.equals(manifestOS)) {
    return true;
}
return false;  // Fails for both missing AND mismatched
```

Proposed change:
```java
String manifestOS = attributes.getValue("SWT-OS");
String manifestArch = attributes.getValue("SWT-Arch");

// Missing: allow (uber JAR or universal binary)
if (manifestOS == null || manifestArch == null) {
    return true;
}

// Present but mismatched: fail with clear diagnostic
if (!arch.equals(manifestArch) || !os.equals(manifestOS)) {
    System.err.println("SWT platform mismatch:");
    System.err.println("  Manifest: SWT-OS=" + manifestOS + ", SWT-Arch=" + manifestArch);
    System.err.println("  Runtime: os=" + os + ", arch=" + arch);
    return false;
}

return true;
```

This 5-line change:
- Fixes uber JAR and universal binary scenarios
- Still catches wrong platform JARs (e.g., Linux JAR on Windows)
- Provides clear diagnostics when actual mismatch detected
- Maintains security benefit of platform checking

Thank you SWT Team!

HeikoKlare:
When packaging SWT into an Uber Jar or using universal binaries (for multiple architectures), the Jar manifest cannot reference a single OS and/or architecture. As a result, the compatibility check implemented in Library.isLaodable() can never succeed for such a packaging. With those compatibility checks always being executed upon initialization of SWT with a recent change, using such packaging of SWT became difficult or impossible.

This change makes the initialization of SWT not fail in case the JAR manifest is completely lacking the according entries for OS and architecture. This will still make the initialization fail if an OS and/or architecture is specified in the manifest that is incompatible to the current environment, but in case the manifest does not contain such entries, the library will just (try to) boot. The behavior of Platform.isLoadable() is preserved. That method will still return false in case the Manifest does not contain according entries, as this is the long-standing behavior of that method.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2928

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT PI/cocoa/org/eclipse/swt/internal/Platform.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT PI/cocoa/org/eclipse/swt/internal/Platform.java
@@ -21,7 +21,7 @@ public static boolean isLoadable () {
 }
 
 public static void exitIfNotLoadable() {
-	if (!Library.isLoadable ()) {
+	if (!Library.isLoadable (true)) {
 		System.err.println("Libraries for platform " + Platform.PLATFORM + " cannot be loaded because of incompatible environment");
 		System.exit(1);
 	}
--- a/bundles/org.eclipse.swt/Eclipse SWT PI/common/org/eclipse/swt/internal/Library.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT PI/common/org/eclipse/swt/internal/Library.java
@@ -13,15 +13,10 @@
  *******************************************************************************/
 package org.eclipse.swt.internal;
 
-import java.io.File;
-import java.io.IOException;
-import java.io.InputStream;
-import java.net.JarURLConnection;
-import java.net.URL;
-import java.net.URLConnection;
-import java.nio.file.Files;
-import java.nio.file.StandardCopyOption;
-import java.util.jar.Attributes;
+import java.io.*;
+import java.net.*;
+import java.nio.file.*;
+import java.util.jar.*;
 
 public class Library {
 
@@ -192,6 +187,10 @@ static boolean extract (String extractToFilePath, String mappedName) {
 }
 
 static boolean isLoadable () {
+	return isLoadable(false);
+}
+
+static boolean isLoadable (boolean ignoreMissingManifest) {
 	URL url = Platform.class.getClassLoader ().getResource ("org/eclipse/swt/internal/Library.class"); //$NON-NLS-1$
 	if (!url.getProtocol ().equals ("jar")) { //$NON-NLS-1$
 		/* SWT is presumably running in a development environment */
@@ -216,6 +215,9 @@ static boolean isLoadable () {
 	String manifestOS = attributes.getValue ("SWT-OS"); //$NON-NLS-1$
 	String manifestArch = attributes.getValue ("SWT-Arch"); //$NON-NLS-1$
 	if (arch.equals (manifestArch) && os.equals (manifestOS)) {
+		return true;
+	}
+	if (manifestArch == null && manifestOS == null && ignoreMissingManifest) {
 		return true;
 	}
 
--- a/bundles/org.eclipse.swt/Eclipse SWT PI/gtk/org/eclipse/swt/internal/Platform.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT PI/gtk/org/eclipse/swt/internal/Platform.java
@@ -22,7 +22,7 @@ public static boolean isLoadable () {
 }
 
 public static void exitIfNotLoadable() {
-	if (!Library.isLoadable ()) {
+	if (!Library.isLoadable (true)) {
 		System.err.println("Libraries for platform " + Platform.PLATFORM + " cannot be loaded because of incompatible environment");
 		System.exit(1);
 	}
--- a/bundles/org.eclipse.swt/Eclipse SWT PI/win32/org/eclipse/swt/internal/Platform.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT PI/win32/org/eclipse/swt/internal/Platform.java
@@ -23,7 +23,7 @@ public static boolean isLoadable () {
 }
 
 public static void exitIfNotLoadable() {
-	if (!Library.isLoadable ()) {
+	if (!Library.isLoadable (true)) {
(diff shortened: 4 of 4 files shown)


HeikoKlare:
@openaudible would this solution be fine for you?

openaudible (changes requested):
I would say it might do the trick, but you've introduced complexity and the default isLoadable will still return false negatives. 


In my opinion: 

- isLoadable is simply a way for exitIfNotLoadable to catch the case where a developer or user tries to launch SWT code with the wrong jar.
-  Thus, isLoadable should provide a detailed error message right before System.exit is called. 
- isLoadable should not return false unless it is guaranteed to be not loadable. (No false negatives!) 
- loadLibrary failing with one or more error messages has been the real long-standing behavior of SWT. The manifest check was introduced last year.

This function accomplishes the above and doesn't require changes to each Platform. It gives a highly detailed error message which should make it clear exactly which jar is causing the problem and what the architecture mismatch is. 

```
    static boolean isLoadable() {
        URL url = Platform.class.getClassLoader().getResource("org/eclipse/swt/internal/Library.class"); //$NON-NLS-1$
        if (!url.getProtocol().equals("jar")) { //$NON-NLS-1$
            /* SWT is presumably running in a development environment */
            return true;
        }

        String jarPath = null;
        Attributes attributes = null;
        try {
            URLConnection connection = url.openConnection();
            if (!(connection instanceof JarURLConnection jc)) {
                /* should never happen for a "jar:" url */
                return false;
            }
            attributes = jc.getMainAttributes();
            jarPath = jc.getJarFileURL().getPath();
        } catch (IOException e) {
            /* should never happen for a valid SWT jar with the expected manifest values */
            return false;
        }

        String manifestOS = attributes.getValue("SWT-OS"); //$NON-NLS-1$
        String manifestArch = attributes.getValue("SWT-Arch"); //$NON-NLS-1$

        // Missing: allow (uber JAR or universal binary) where manifest entries are stripped
        if (manifestOS == null || manifestArch == null) {
            return true;
        }
        String os = os();
        String arch = arch();

        // Present but mismatched: fail with clear diagnostic
        if (!arch.equals(manifestArch) || !os.equals(manifestOS)) {
            System.err.println("org.eclipse.swt platform mismatch for jar: " + jarPath);
            System.err.println("  Manifest: SWT-OS=" + manifestOS + ", SWT-Arch=" + manifestArch);
            System.err.println("  Runtime: os=" + os + ", arch=" + arch);
            return false;
        }

        return true;
    }

```

Respectfully,

HeikoKlare:
In general, I share your opinion on the expectation of when `Library.isLoadable()` should return true/false. However, this is exposed as public API (via `SWT.isLoadable()`), and we would have to throughly think about changes in behavior of an API method which has been like that since decades. I have no idea how consumers may have used that method and whether they rely on the current behavior, so changing the behavior at least has some risk of breaking consumers. Quite the contrary, no one has complained so far that `SWT.isLoadable()` returns an unexpected result.

In case we agree that changing the behavior of `Library/SWT.isLoadable()` is reasonable, it's nothing for the upcoming release anymore (we are beyond M3).

openaudible:
If you're happy with it, then it's fine with me. 
Thanks for resolving.

akurtakov:

HeikoKlare:
The change should be safe and restores previous behavior for Ubers JARs / universal binaries, but it is obviously not a perfect solution for the semantics of the involved methods. I have no stakes no in this at all and just wanted to help resolve the reported issue. Since there seems to be no interest in moving forward in some other direction, I think we should just merge this for now and proceed just in case there are further complains.

Issue #3085, opened 2026-02-16
[Mac] Buttons with font sizes larger than 23pt are not correctly rendered


xpomul:
**Describe the bug**
When setting the font to 24pt or more for a Button in SWT on MacOS with Cocoa, the button does not grow in height together with the font. This results in the upper part of the button label to be clipped.

**To Reproduce**

```java
package net.winklerweb.test;

import org.eclipse.jface.layout.GridDataFactory;
import org.eclipse.jface.resource.JFaceResources;
import org.eclipse.swt.SWT;
import org.eclipse.swt.events.SelectionListener;
import org.eclipse.swt.graphics.Font;
import org.eclipse.swt.graphics.FontData;
import org.eclipse.swt.graphics.Image;
import org.eclipse.swt.graphics.Rectangle;
import org.eclipse.swt.layout.GridData;
import org.eclipse.swt.layout.GridLayout;
import org.eclipse.swt.widgets.Button;
import org.eclipse.swt.widgets.Combo;
import org.eclipse.swt.widgets.Display;
import org.eclipse.swt.widgets.Label;
import org.eclipse.swt.widgets.Shell;

public class FontScalingTest {

    public static void main(String[] args) {
        Display display = new Display();
        Shell shell = new Shell(display);
        shell.setText("Button Test");
        shell.setLayout(new GridLayout(2, false));

        FontData defaultFontData = display.getSystemFont().getFontData()[0];
        
        new Label(shell, SWT.NONE).setText("Font Size:");
        
        Combo fontSizeCombo = new Combo(shell, SWT.READ_ONLY);
        fontSizeCombo.setItems("8", "10", "12", "14", "16", "18", "20", "21", "22", "23", "24", "25", "26", "27", "28", "29", "30", "36", "48", "72", "96");
        fontSizeCombo.select(4);
        fontSizeCombo.setLayoutData(new GridData(SWT.FILL, SWT.CENTER, true, false));
                
        Button button = new Button(shell, SWT.PUSH);
        button.setText("Button");
        button.setLayoutData(GridDataFactory.swtDefaults().grab(false, false).span(1, 1).create());
                
        updateButtonFont(button, defaultFontData, 16);
        
        fontSizeCombo.addSelectionListener(SelectionListener.widgetSelectedAdapter( e-> {
            int size = Integer.parseInt(fontSizeCombo.getText());
            updateButtonFont(button, defaultFontData, size);
            shell.layout(true);
        }));
        
[...]

PR #3086, opened 2026-02-16
[Mac] Buttons with font sizes>23pt are not correctly rendered #3085

xpomul:
Fix issue #3085 by handling the case when a button uses a custom font with size > 16pt. In this case adjust the button height in cellSizeForBounds(), similar to the other height adjustment fixes already present.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Button.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Button.java
@@ -151,6 +151,17 @@ NSSize cellSizeForBounds (long id, long sel, NSRect cellFrame) {
 		size.height = Math.max(size.height, imageSize.height);
 	}
 
+	// Native buttons in Cocoa always render with a height of 24.0; a font size higher than 16pt is not reflected.
+	// Therefore, we adjust the button height manually, if a custom font with a size > 16 is set.
+	if (font != null && font.handle.pointSize() > 16.0) {
+		NSCell cell = new NSCell (id);
+		NSRect titleRect = cell.titleRectForBounds(cellFrame);
+		// For some reason, the titleRectForBounds() returns too small heights for font sizes > 20
+		// We fix this by adjusting the height by 8 which leads to nicer results for all sizes.
+		final int FONT_SIZE_HEIGHT_ADJUST = 8;
+		size.height = Math.max(size.height, titleRect.height + FONT_SIZE_HEIGHT_ADJUST);
+	}
+
 	if (((style & (SWT.PUSH|SWT.TOGGLE)) !=0) && (style & (SWT.FLAT|SWT.WRAP)) == 0) {
 		if (image != null) {
 			NSCell cell = new NSCell(id);


vogella:
@HeikoKlare can you review?

HeikoKlare:
I have to admit that I am probably not the right person to review this. I am not a common Mac user and not an expert in the SWT implementation for Cocoa.

What I can contribute here is some testing and some considerations. Let's start with the testing:
With the proposed change, the behavior improves in the sense that the button increases its size for larger fonts. However, the behavior I experience feels a bit inconsistent. For example, when I increase the font size from 14 to 20, I get this:
image
When I reduce the font size from 24 to 20, I get this:
image
While already feels strange when manually changing the font size, it will be even be more difficult to achieve an expected result when automatically setting a fixed (but high) font size if the result will depend on which font size was set before.

With respect to the idea of adapting the size of the button at all, it seems as if this is not intended by Apple. From what I find in their documentation, it seems as if they assume the standard font (13 pt?) to be used for buttons. And if buttons with variable height content shall be used, there seem to be specific controls or control configurations available: https://developer.apple.com/design/human-interface-guidelines/buttons
We of course sometime need to overrule the intents of the operating systems and their guidelines to achieve consistent behavior across the operating systems, so it might be okay to do that here, but I still wanted to make that point.

xpomul:
Thanks @HeikoKlare for having a look.
I am not a MacOS/Cocoa developer myself and have just tried to piece together what seemed to work better than what we currently have (which is buttons heights stuck at what is fine for the 13pt font size the system uses). 

I was hoping to get in touch with someone who knows their way around Cocoa and whether there is a way to improve/work around the quirky behavior you have experienced (setting the same height leads to different results depending on whether a smaller or larger height was set before...). So far, I was not able to find one.
What I experienced is that, actually, the 20pt size you have experimented with is the size with the most unpleasant behavior. Other sizes, both smaller and larger that around 20pt seem to work much better... Still, I don't know why... :-(

Maybe it is a more general discussion how we deal with the strict implications from Apple.

* strictly accepting them would mean that Button.setFontSize() should have no effect and we stick to the Apple way of keeping a 13pt font for all buttons
* supporting fonts with bigger sizes (as we are currently doing), but then I'd say we should try to improve on the consequences (scaling of buttons) as well
* the radical approach might even be something like ditching the NSButton completely and doing it completely in SWT (at the cost of visual differences). But that is quite some effort and maybe not worth it...

That said, in the meantime, I have come across more parts where larger font sizes lead to issues with Apple's Cocoa implementation (e.g. Combo boxes). So there might be similar fixes needed (and maybe more strange behavior in Apple API to work around...)

xpomul:
@vogella @HeikoKlare do you know if there is anyone else who could contribute to this discussion?

vogella:
@Phillipus can you help here?

Phillipus:
> @Phillipus can you help here?

I share @HeikoKlare's concerns, but on the other hand this PR doesn't necessarily make things too worse. Apple's Human Interface Guidelines recommend standard button styles with 13pt text. Without this PR the font size is changed and the button doesn't change size.

Hard to say one way or the other.

HeikoKlare:
Following the proposal in [Apple's documentation](https://developer.apple.com/design/human-interface-guidelines/buttons), I created an alternative potential fix:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/3210

xpomul:
Closing in favor of the much better solution in #3210

PR #3087, opened 2026-02-17
Use assertThrows in Menu test

akurtakov:


akurtakov:
Pure tests change. Test results are not updated after rerun of tests thus the results are not the final ones. Merging.

PR #3098, opened 2026-02-25
[Gtk4] Fix Browser.setText

akurtakov:
Disable hardware-acceleration-policy on Gtk 4.x (it is done for Gtk 3.x already) as  first frame is never drawn when using accelerated compositing (GPU path). The DOM loads, signals fire, but the view stays blank.
Different value compared to Gtk 3 is due to the fact that the setting no longer has always/ondemand/never but just always/never on Gtk 4.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT WebKit/gtk/org/eclipse/swt/browser/WebKit.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT WebKit/gtk/org/eclipse/swt/browser/WebKit.java
@@ -786,7 +786,11 @@ public void create (Composite parent, int style) {
 	OS.g_object_set (settings, WebKitGTK.enable_developer_extras, 1, 0);
 	//disable hardware acceleration due to  https://bugs.webkit.org/show_bug.cgi?id=239429#c11
 	//even evolution ended up doing the same https://gitlab.gnome.org/GNOME/evolution/-/commit/eb62ccaa28bbbca7668913ce7d8056a6d75f9b05
-	if (!GTK.GTK4) {
+	if (GTK.GTK4) {
+		// "1" is "never" in webkitgtk6 https://webkitgtk.org/reference/webkitgtk/stable/enum.HardwareAccelerationPolicy.html#never
+		OS.g_object_set (settings, WebKitGTK.hardware_acceleration_policy, 1, 0);
+	} else {
+		// "2" is "never" in webkitgtk4 https://webkitgtk.org/reference/webkit2gtk/stable/enum.HardwareAccelerationPolicy.html#never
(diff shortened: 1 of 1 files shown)


akurtakov:
This is purely Gtk 4 change so should be safe to get into RC2. @HeikoKlare ,  @merks , @iloveeclipse - could someone approve ?

merks:
It looks fine to me.  @HeikoKlare  can approve as a PL.

bundles/org.eclipse.swt/Eclipse SWT WebKit/gtk/org/eclipse/swt/browser/WebKit.java
```diff
@@ -786,7 +786,9 @@ public void create (Composite parent, int style) {
 	OS.g_object_set (settings, WebKitGTK.enable_developer_extras, 1, 0);
 	//disable hardware acceleration due to  https://bugs.webkit.org/show_bug.cgi?id=239429#c11
 	//even evolution ended up doing the same https://gitlab.gnome.org/GNOME/evolution/-/commit/eb62ccaa28bbbca7668913ce7d8056a6d75f9b05
-	if (!GTK.GTK4) {
+	if (GTK.GTK4) {
+		OS.g_object_set (settings, WebKitGTK.hardware_acceleration_policy, 1, 0);
```
iloveeclipse, line 791: Could we please change "1" and "2" below to some named variables or constants, so the code could be also understood by everyone?

iloveeclipse:
I have no objections to merge that, but I would wish a bit more comments in the code explaining the different modes and why the different modes used for which GTK version.

bundles/org.eclipse.swt/Eclipse SWT WebKit/gtk/org/eclipse/swt/browser/WebKit.java
akurtakov: I have put comments instead with links to webkitgtk docs . Having constants most of which end up being 0, 1 just grows the codebase not making things clear esp considering the mess we have in webkitgtk - webkit1/2/3/4 (with other variations in naming) on Gtk 3 and webkit 5/6 (5 being really short lived)  on Gtk 4.

iloveeclipse (approved):
OMG, why should such trivial values like "never" change in different versions?

akurtakov:
> OMG, why should such trivial values like "never" change in different versions?

It's enum on C side (https://github.com/WebKit/WebKit/blob/3af87ab356f096e67488c2e992e06cf729f9f702/Source/WebKit/UIProcess/API/glib/WebKitSettings.h.in#L85 ) thus safe but SWT uses just the "ordinal" as we can't ifdef and having multiple enums for same purpose but different versions would make things even worse.

Issue #3105, opened 2026-02-28
API error in 4.40 for DPITestUtil


iloveeclipse:
Using first 4.40 build I20260227-1600, PDE API tooling reports error in SWT (against 4.39RC2 platform baseline):
```
Description	Resource	Path	Location	Type
The minor version should be incremented in version 3.133.100, since new APIs have been added since version 3.133.0	MANIFEST.MF	/org.eclipse.swt.win32.win32.x86_64/META-INF	line 6	Version Numbering Problem
```

```
List of compatible changes:
- The type org.eclipse.swt.widgets.DPITestUtil has been added to 
org.eclipse.swt.win32.win32.x86_64_3.133.100
```

<img width="1547" height="336" alt="Image" src="https://github.com/user-attachments/assets/37ca6125-1961-4d91-aa1f-59d7198ea275" />

merks:
Yes, I was mentioning this to @HannesWell @akurtakov  and @laeubi  because it happens in my updated SDK Oomph setup.  Also I can reproduce it oomphing just the SWT in a from-scratch setup:

<img width="1044" height="315" alt="Image" src="https://github.com/user-attachments/assets/0fac8aa8-5f79-4947-8e0d-4b2548dcb88a" />

I'm actually not sure it's a bug in SWT versus a bug in the API tools.  😁 

It's been around for a while:

<img width="1008" height="371" alt="Image" src="https://github.com/user-attachments/assets/ae9a40ef-075f-41d8-a3a6-472a9079b539" />

iloveeclipse:
> it happens in my updated SDK Oomph setup

Oomph is unrelated, I see the error without using it.

> I'm actually not sure it's a bug in SWT versus a bug in the API tools.

Not sure if the baseline for SWT contains `DPITestUtil`? If not, it's RC2 issue

merks:
I didn't intend to suggest it is Oomph related; I just mention it because makes the problem very easy to reproduce that way, and, given most folks are using the Oomphed SDK, those people on Windows will eventually/soon see this problem. 

The baseline contains the fragment:

<img width="559" height="505" alt="Image" src="https://github.com/user-attachments/assets/229945b6-df91-4c63-96c9-be8774e33260" />

But now that I look closely, it's "just" in a test  source folder

<img width="503" height="1079" alt="Image" src="https://github.com/user-attachments/assets/f1814055-76a0-45db-ba5a-b6222d267439" />

and is only used in tests:

[...]

PR #3106, opened 2026-02-28
Bump minor version for 4.40 stream

HeikoKlare:
We had to do the same for the 4.39 stream: https://github.com/eclipse-platform/eclipse.platform.swt/pull/2829

Necessary because of API tooling limitation to be addressed via:
- https://github.com/eclipse-pde/eclipse.pde/issues/1496

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/3105

### DIFF
--- a/binaries/org.eclipse.swt.cocoa.macosx.aarch64/.settings/.api_filters
+++ b/binaries/org.eclipse.swt.cocoa.macosx.aarch64/.settings/.api_filters
@@ -234,4 +234,12 @@
             </message_arguments>
         </filter>
     </resource>
+    <resource path="META-INF/MANIFEST.MF">
+        <filter id="926941240">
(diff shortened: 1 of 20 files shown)


HannesWell:
> Necessary because of API tooling limitation to be addressed via:
> 
>     * [[API-Tools] Ignore test src folders by default eclipse-pde/eclipse.pde#1496](https://github.com/eclipse-pde/eclipse.pde/issues/1496)

Since this issue should be fixed now (thanks again Ed!), I suggest to keep this open until that fix is verified with a new I-build or even abandon it until really necessary API-wise.

HeikoKlare:
Closing as obsolete due to:
- https://github.com/eclipse-pde/eclipse.pde/pull/2230

Issue #3108, opened 2026-03-02
[Windows] When hovering over a header on a Tree, tooltips for headers set using TreeColumn.setToolTipText do not appear.


tconnard:
**Describe the bug**
When hovering over a header on a Tree, tooltips for headers set using TreeColumn.setToolTipText do not appear.

**To Reproduce**
Hover mouse cursor over Tree header that has a tooltip such as eclipse Problems view.

**Expected behavior**
I expect the tooltip to display as it does on previous versions.

**Screenshots**
Example from Eclipse IDE
GTK + Older version
<img width="520" height="356" alt="Image" src="https://github.com/user-attachments/assets/babd4632-bab6-47d2-acd1-8954f259232d" />

Windows current version. Cursor is invisible but hovering over "Description" header.
<img width="530" height="375" alt="Image" src="https://github.com/user-attachments/assets/3b0f27d9-f9bc-48a7-9a37-f27f1078c8c2" />

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [x] All OS
- - [ X ] Windows
- - [ ] Linux
- - [ ] macOS 

2. Additional OS info (e.g. OS version, Linux Desktop, etc)
Running in a VDI

3. JRE/JDK version
JDK21
 
**Version since**
 Detected in 3.130 still present in 2d4d435e6b37e9f13f5c89cf4768abe7f1717882

**Workaround (or) Additional context**
As far as I can tell this is caused by  https://github.com/eclipse-platform/eclipse.platform.swt/pull/2142 Which adds a block of code on lines 8187-8195 that hides tooltips in there case there is no target cell. This is incorrect behavior. 

If the bounds of the tooltip needs to be changed it may be necessary to retrieve it using OS.GetWindowPlacement and then modify it, instead of relying on non existent cell data.

PR #3112, opened 2026-03-03
Do not delete OS.SWP_NOSIZE bit for custom tooltips on Trees #3108

fedejeanne:
Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/3108

@HeikoKlare I noticed that commit 65a88c6402562c82ac88532c14412a5602e6bd24 not only moved these 2 lines up but it also added the additional check that I am removing in this PR.

https://github.com/eclipse-platform/eclipse.platform.swt/blob/1e86a0c6c8b35025557cccc4cc6120cfda9ac729/bundles/org.eclipse.swt/Eclipse%20SWT/win32/org/eclipse/swt/widgets/Tree.java#L8166-L8167

 Do you remember _why_ this check was necessary? I don't remember and honestly it seems like I simply missed this added logic back when I merged #2142 .

# How to test
- Hover over the columns headers in the _Problems_ view
- Edge case: position the workbench window in a way that the tooltip is supposed to span between 2 monitors with different scaling.

The edge case is just to be sure that nothing breaks, I remember that some of the stuff that was done around tooltips was to avoid UI hangs and infinite loops when working with monitors at different zooms.

I tested with this configuration but the zoom levels don't play a role, just the fact that the zoom levels are different:
- Primary monitor on the **right** at **125%**
- Secondary monitor on the **left** at **150%** (and also at **100%**)

No issues found. Here are some screenshots, I marked the division line between my 2 monitors with a yellow line (I added it manually to the screenshots).

image

---

image

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
@@ -7404,7 +7404,11 @@ LRESULT wmNotify (NMHDR hdr, long wParam, long lParam) {
 	if (hdr.hwndFrom == itemToolTipHandle && itemToolTipHandle != 0) {
 		LRESULT result = wmNotifyToolTip (hdr, wParam, lParam);
 		if (result != null) return result;
+	} else if (hdr.hwndFrom == headerToolTipHandle && headerToolTipHandle != 0) {
+		// if it's the header, let Windows do its thing.
 	} else if (hdr.code == OS.TTN_SHOW) {
+		// Fallback: it's not about the item or about the header but it's still about a
+		// tooltip.
 		return positionTooltip(hdr, wParam, lParam, false);
 	}
 	if (hdr.hwndFrom == hwndHeader && hwndHeader != 0) {


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Tree.java
```diff
@@ -8164,7 +8164,7 @@ private LRESULT positionTooltip(NMHDR hdr, long wParam, long lParam, boolean man
 	LRESULT result = super.wmNotify (hdr, wParam, lParam);
 	if (result != null) return result;
 	int flags = OS.SWP_NOACTIVATE | OS.SWP_NOZORDER | OS.SWP_NOSIZE;
-	if (isCustomToolTip () || !managedTooltip) flags &= ~OS.SWP_NOSIZE;
+	if (isCustomToolTip ()) flags &= ~OS.SWP_NOSIZE;
```
HeikoKlare: Thank you for the quick fix proposal. This indeed seems to fix https://github.com/eclipse-platform/eclipse.platform.swt/issues/3108.

What this change effectively does is rendering the else block starting in line 8191 obsolete. When having a non-managed tooltip, the flags `OS.SWP_NOSIZE` and `OS.SWP_NOMOVE` (line 8192) will be set. That means that line 8193 will be effectively become a no-op: the intent of setting the tooltip bounds to zero is not fulfilled anymore because `OS.SWP_NOSIZE` is set.
So in my understanding this change is effectively the same as if we would (additionally) remove the else block in line 8191.

What I am not sure about is why/if we need this "fallback" of setting the bounds to zero for an unmanaged tooltip that has no table cell. Precisely, if the call to `positionTooltip` with `managedTooltip == false` (done in `wmNotify`) is relevant at all. If it is necessary, then this fix would be invalid, but I currently don't see where it would be necessary. So in that case one might completely remove this method call.
In particular, the table column header tooltips (which are currently not visible because of the addressed bug) are properly fit into one monitor, such that the original issue of tooltip spanning multiple monitor and freezing the application (https://github.com/eclipse-platform/eclipse.platform.swt/issues/557) will not occur here.

Taking a specific look into the call of `positionTooltip` in `wmNotify`, it seems like this happens when a tooltip is about to be shown and the event is not for the stored item tooltip handle. I wonder if we should maybe just exclude the repositioning for the case that we have a header tooltip handle at hand here and still have the fallback code for case that the tooltip is not managed. This would reduce the risk of further regression as we only address the specific case of table column headers.
Precisely, the idea would be to extend the else-if in line 7403 to something like:
```
} else if ((hdr.hwndFrom != headerToolTipHandle || headerToolTipHandle == 0) && hdr.code == OS.TTN_SHOW) {
```

@fedejeanne  wdyt?

fedejeanne: As discussed in our call, I opted for the safer approach and simply let Windows handle the tooltip for the headers instead of changing anything inside `positionTooltip(...)`. That way the fix in #2142 still applies to all _other_ cases.

It still remains a mystery to me _what exactly are the other cases_. I tried showing header and cell tooltips that span across 2 monitors with different zoom levels but, as you predicted, this falls into the `if`-block inside `positionTooltip(...)`. I don't get the 'else'-block in there.

PR #3113, opened 2026-03-03
Rename Snippet 392.png to Snippet392.png (remove blank space) #3077

fedejeanne:
Contributes to
https://github.com/eclipse-platform/eclipse.platform.swt/pull/3077


merks:
@fedejeanne   You probably shouldn't wait for a build that only causes global warming. 😕

Issue #3092, opened 2026-02-23
SWT Javadoc bash for 4.39


deepika-u:
SWT Javadoc bash for 4.39

Steps to perform
- - [x] Need to pull the latest changes on master.
- - [x] Make sure the git staging view is clean.
- - [x] Then run this standalone Java application: JavadocBasher.java.
- - [x] If there any valid changes it will reflect in git staging.
- - [x] Validate them manually and merge if any.

But this task is to be merged in RC2 only after verifying it.

deepika-u:
Checked on this build `I20260221-1800`, found some updates missing on gtk and cocoa but done on windows. So submitted the same changes in #3093. Thanks.

deepika-u:
Verified on `Version: 2026-03 (4.39), Build id: I20260224-0020`, everything looks good.
Thanks alot everyone.

merks:
And thank you @deepika-u  for taking care of this.

PR #3115, opened 2026-03-03
[RelEng] Add JavaDoc-Basher Pull-Request check

HannesWell:
The added check runs the javadoc-basher on each PR and ensures the JavaDoc of all public API elements is synchronized for all platforms. In case that run leads to changes in the git worktree, the check fails.

This ensures the JavaDoc is always properly in sync und thus avoids the need to verify JavaDoc consistency before the release as separate RelEng task, like it was done for example in
- https://github.com/eclipse-platform/eclipse.platform.swt/issues/3092

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -19,3 +19,23 @@ jobs:
     with:
       botName: Eclipse Platform Bot
       botMail: platform-bot@eclipse.org
+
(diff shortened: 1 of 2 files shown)


HannesWell:
In case of inconsistencies the check fails 
grafik

Clicking on the details of the failed check shows the diff and an expressive message:
grafik

PR #3116, opened 2026-03-03
Remove almost empty o.e.swt.tools.base Plug-in

HannesWell:
Move the remaining view definition into `org.eclipse.swt.tools.spies` (hosted at PDE) and introduce a new category for development tools of SWT itself, named `org.eclipse.swt.dev.tools`.
And use it for the `MacGeneratorView`.

@iloveeclipse and @HeikoKlare as PLs and heavy SWT contributors, is this change fine?

### DIFF
--- a/bundles/org.eclipse.swt.tools/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.swt.tools/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-Name: %pluginName
 Bundle-Vendor: %providerName
 Bundle-SymbolicName: org.eclipse.swt.tools; singleton:=true
-Bundle-Version: 3.111.0.qualifier
+Bundle-Version: 3.112.0.qualifier
 Bundle-ManifestVersion: 2
 Export-Package: org.eclipse.swt.tools.internal; x-internal:=true
 Bundle-ActivationPolicy: lazy
--- a/features/org.eclipse.swt.tools.feature/feature.xml
+++ b/features/org.eclipse.swt.tools.feature/feature.xml
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.swt.tools.feature"
       label="%featureName"
-      version="3.110.0.qualifier"
+      version="3.111.0.qualifier"
       provider-name="%providerName"
       license-feature="org.eclipse.license"
       license-feature-version="0.0.0">


akurtakov:
@HannesWell You have to adjust https://github.com/eclipse-pde/eclipse.pde/blob/master/ui/org.eclipse.swt.tools.spies/plugin.xml#L7 too . This plugin is a leftover from before spies being moved to pde.

HannesWell:
> You have to adjust https://github.com/eclipse-pde/eclipse.pde/blob/master/ui/org.eclipse.swt.tools.spies/plugin.xml#L7 too . This plugin is a leftover from before spies being moved to pde.

I believe the category in PDE doesn't have to change since we move category definition declares the id just like before:
```
<category
      id="org.eclipse.swt.swt.tools"
      name="%viewCategory">
```
The existing `MacGeneratorView` that's already in the `swt.tools` Plugin is also not adjusted.

akurtakov:
"SWT spies" do not depend on "SWT tools"

HannesWell:
> "SWT spies" do not depend on "SWT tools"

Ah, yes. Now I see what you meant.

akurtakov:
Picking up some other existing category for "SWT spies" id probably best but the problem is that they don't depend on anything IIRC. So they may have to define their own category.

HannesWell:
> Picking up some other existing category for "SWT spies" id probably best but the problem is that they don't depend on anything IIRC. So they may have to define their own category.

Sounds good, but I would leave that task for later (and probably somebody else).
I have now also applied a minor-version increment (due to the addition of the new category) to swt.tools and because, for to me unknown reasons, the version-increment workflow can't push to my fork.

I also adapted PDE but required the new major version, which is why the build currently fails:
- https://github.com/eclipse-pde/eclipse.pde/pull/2239

So we either have to make this a two step process or submit the PDE change without successful verification build.
I'm in favor of the latter to keep it simpler.

akurtakov:
Change looks good to me logically.

HannesWell:
> Picking up some other existing category for "SWT spies" id probably best but the problem is that they don't depend on anything IIRC. So they may have to define their own category.

After our private discussion I've no moved the definition of the SWT Tools category to PDE:
- https://github.com/eclipse-pde/eclipse.pde/pull/2239

For the previous reference in swt.tools I introduced a new category swt.dev.tools, for tools to develop SWT itself, the existing one is continue to be used for tools to develop with/target SWT.

PR #3117, opened 2026-03-03
Unify and clean-up code of JavadocBasher

HannesWell:


akurtakov:
Please merge and continue if/when you're fine with it.

HannesWell:
> Haven't tested it but nothing spotted while reading the PR.

Thanks for having a look.
The good thing is, since https://github.com/eclipse-platform/eclipse.platform.swt/pull/3115, the javadoc basher is used on each PR and problems can be seen immediately.

PR #3118, opened 2026-03-04
Migrate JavadocBasher to Path API and fail hard on IO errors

HannesWell:
instead of just printing to the console.
This simplifies code and prevents errors from being hidden.


akurtakov:
Looks good. Thanks for this one.

Issue #3120, opened 2026-03-05
GC.setClipping(path) doesn't paint on Linux when transformed


ptziegler:
**Describe the bug**

This is an issue reported here: https://github.com/eclipse-gef/gef-classic/issues/1044. Draw2D makes extensive use of the SWT Transform class to e.g. handle scrolling. When `setClip(Path)` is called, this transformation is ignored and areas are clipped that should be visible.

**To Reproduce**

```java
import org.eclipse.swt.SWT;
import org.eclipse.swt.graphics.GC;
import org.eclipse.swt.graphics.Path;
import org.eclipse.swt.graphics.Transform;
import org.eclipse.swt.layout.FillLayout;
import org.eclipse.swt.widgets.Display;
import org.eclipse.swt.widgets.Shell;

public class ClipGraphics2 {
    
    public static void main(String[] args) {
        Shell shell = new Shell();
        shell.setLayout(new FillLayout());
        shell.setSize(600, 300);
        
        Display display = shell.getDisplay();
        
        shell.addPaintListener(event -> {
        	GC gc = event.gc;
        	
        	Path p = new Path(display);
        	p.addRectangle(50, 350, 100, 100);
        	
        	Transform t = new Transform(display);
        	t.translate(0, -300);
        	
        	gc.setTransform(t);
        	
        	gc.setBackground(display.getSystemColor(SWT.COLOR_RED));
        	gc.fillPath(p);
        	gc.setClipping(p);
        	gc.setBackground(display.getSystemColor(SWT.COLOR_GREEN));
        	gc.fillPath(p);
        	
        	p.dispose();
        	
        	gc.setTransform(null);
        	t.dispose();
        });
        
        shell.open();
        
        while(!shell.isDisposed()) {
            if(!display.readAndDispatch()) {
                display.sleep();
            }
        }
    }
}
```

**Expected behavior**

The rectangle should be painted green, but is painted red on Linux.

**Screenshots**

From the original GEF issue:

https://github.com/user-attachments/assets/454fb805-15d8-4c6a-894b-fcf2bf47005f

The snippet:

<img width="732" height="459" alt="Image" src="https://github.com/user-attachments/assets/bd93b8bb-02aa-4514-b55b-cd1616543291" />

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [ ] Windows
- - [x] Linux
- - [ ] macOS 
 
**Version since**

[...]

PR #3121, opened 2026-03-05
GC.setClipping(path) doesn't paint on Linux when transformed #3120

ptziegler:
The fix for Bug 531667 (d7ce5978d32a9dd4bf77dfa77e4b1c696992af4a) is incomplete and doesn't properly update the clipping of a `Path` when a `Transform` is set. This is because the clipping is calculated by
intersecting the clipping rectangle when the GC instance was created with the `Path`. The region described by the `Path` is affected by the transformation, while the region of the rectangle is not.

To fix this problem, the transformation needs to be cleared and re-applied when resetting the clipping, so that the region of the clipping rectangle is properly updated.

Closes https://github.com/eclipse-platform/eclipse.platform.swt/issues/3120

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/GC.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/GC.java
@@ -3094,7 +3094,19 @@ public void setClipping(int x, int y, int width, int height) {
 public void setClipping(Path path) {
 	if (handle == 0) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
 	if (path != null && path.isDisposed()) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
+
+	Transform t = null;
+	if (currentTransform != null) {
+		t = new Transform(getDevice());
+		getTransform(t);
+		setTransform(null);
+	}
 	resetClipping();
+	if (t != null) {
+		setTransform(t);
+		t.dispose();
+	}
+
 	if (path != null) {
 		initCairo();
 		long cairo = data.cairo;
--- a/tests/org.eclipse.swt.tests.gtk/ManualTests/org/eclipse/swt/tests/gtk/snippets/Issue3120_GC_setClipping_with_Transform.java
+++ b/tests/org.eclipse.swt.tests.gtk/ManualTests/org/eclipse/swt/tests/gtk/snippets/Issue3120_GC_setClipping_with_Transform.java
@@ -1,0 +1,116 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Patrick Ziegler and others.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Patrick Ziegler - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.swt.tests.gtk.snippets;
+
+import java.util.function.Consumer;
+
+import org.eclipse.swt.SWT;
+import org.eclipse.swt.graphics.Color;
+import org.eclipse.swt.graphics.GC;
+import org.eclipse.swt.graphics.Path;
+import org.eclipse.swt.graphics.Transform;
+import org.eclipse.swt.layout.FillLayout;
+import org.eclipse.swt.widgets.Composite;
+import org.eclipse.swt.widgets.Display;
+import org.eclipse.swt.widgets.Shell;
+
+/**
+ * Description: {@link GC#setClipping(Path)} doesn't respect a previous call to
+ * {@link GC#setTransform(Transform)}.
+ * <ol>
+ * <li>Run the snippet.</li>
+ * </ol>
+ * Expected results: All painted rectangle should be green.<br>
+ * Actual results: The rectangles are at least partially painted red.
+ *
+ * @see <a href="https://bugs.eclipse.org/bugs/show_bug.cgi?id=531667">Bug
+ *      531667 - [GTK3] Cannot draw Canvas with Control.print(GC) </a>
+ */
+public class Issue3120_GC_setClipping_with_Transform {
+	private static Display DISPLAY = Display.getDefault();
+	private static Color COLOR_RED = DISPLAY.getSystemColor(SWT.COLOR_RED);
+	private static Color COLOR_GREEN = DISPLAY.getSystemColor(SWT.COLOR_GREEN);
+
+    public static void main(String[] args) {
+        Shell shell = new Shell();
+        shell.setLayout(new FillLayout());
+		shell.setSize(800, 300);
+
+		Composite composite1 = new Composite(shell, SWT.BORDER);
+		composite1.addPaintListener(
+				event -> paint(event.gc, p -> p.addRectangle(310, 310, 100, 100), t -> t.translate(-300, -300)));
+
+		Composite composite2 = new Composite(shell, SWT.BORDER);
+		composite2.addPaintListener(
+				event -> paint(event.gc, p -> p.addRectangle(20, 5, 200, 50), t -> t.scale(0.5f, 2.0f)));
+
+		Composite composite3 = new Composite(shell, SWT.BORDER);
+		composite3.addPaintListener(
+				event -> paint(event.gc, p -> p.addRectangle(10, 10, 100, 100), t -> t.shear(1.0f, 0.0f)));
+
+		Composite composite4 = new Composite(shell, SWT.BORDER);
+		composite4.addPaintListener(event -> paint(event.gc, p -> p.addRectangle(10, 605, 100, 50), t -> {
+			t.scale(1.0f, 2.0f);
+			t.translate(0, -600);
+		}));
+
+		Composite composite5 = new Composite(shell, SWT.BORDER);
+		composite5.addPaintListener(event -> paint(event.gc, p -> p.addRectangle(10, 10, 100, 100), t -> {}, gc -> {
+			Path p = new Path(DISPLAY);
+			p.addRectangle(55, 55, 10, 10);
+			gc.setClipping(p);
+			p.dispose();
+		}));
+
+		Composite composite6 = new Composite(shell, SWT.BORDER);
+		composite6.addPaintListener(event -> paint(event.gc, p -> p.addRectangle(10, 275, 100, 50), t -> {
+			t.scale(1.0f, 2.0f);
+			t.translate(0, -270);
+		}, gc -> {
+			gc.setClipping(10, 10, 100, 100);
+		}));
+
+		shell.open();
+
+        while(!shell.isDisposed()) {
+			if (!DISPLAY.readAndDispatch()) {
+				DISPLAY.sleep();
+            }
+        }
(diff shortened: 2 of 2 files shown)


ptziegler:
I've tested this with all "Bug531667_" snippets and couldn't see any regression. So I think this change is compatible with the fix for bug 531667.

ptziegler:
@Phillipus I've also tested this change with your `ClipGraphics` example and the problem disappeared there as well.

Phillipus:
> @Phillipus I've also tested this change with your `ClipGraphics` example and the problem disappeared there as well.

And I've tested it with our RCP app, Archi, and the problem disappeared there as well.

Incredible detective work, thank-you so much!

ptziegler:
> Maybe the correct solution would be to still call resetClipping() but in that method temporarily revert the transformation applied to Cairo for setting the original clipping bounds or otherwise adapt the original clipping bounds applied in resetClipping() to be transformed by an inverse of the GC's transform?

Previously the clipping was reset with `setClipping(0);`, which was then changed to `resetClipping()`. I'm fairly certain this also fixed the underlying problem, so perhaps the fix is rather to add this call back to the `else` branch.

Phillipus:
> Previously the clipping was reset with `setClipping(0);`, which was then changed to `resetClipping()`. I'm fairly certain this also fixed the underlying problem, so perhaps the fix is rather to add this call back to the `else` branch.

I tried that and that seems to work for all cases.

Phillipus:
> Previously the clipping was reset with setClipping(0);, which was then changed to resetClipping(). I'm fairly certain this also fixed the underlying problem, so perhaps the fix is rather to add this call back to the else branch.

> I tried that and that seems to work for all cases.

Actually, in real world usage it doesn't. I tested it on our RCP app and as you scroll up and down some of the figures' parts are being drawn outside the GEF editor:

Screenshot from 2026-03-07 15-11-21
.  

And another case:  

Screenshot from 2026-03-07 15-24-06

ptziegler:
> Actually, in real world usage it doesn't. I tested it on our RCP app and as you scroll up and down some of the figures' parts are being drawn outside the GEF editor:

From what I understand, this is because the clipping is shared between all widgets. So callling `setClipping(0)` removes the clipping for the whole application.

https://github.com/eclipse-platform/eclipse.platform.swt/blob/cc2762f9f02888b3b9122e534ddeabacd34b4604/bundles/org.eclipse.swt/Eclipse%20SWT/gtk/org/eclipse/swt/graphics/GC.java#L3132-L3137

In the end, I've applied @HeikoKlare suggestion of still calling `resetClipping()`, but using the clipping rectangle on which the transformation matrix has been applied.

Phillipus:
> In the end, I've applied @HeikoKlare suggestion of still calling resetClipping(), but using the clipping rectangle on which the transformation matrix has been applied.

Many thanks for continuing to look at this. I tested the latest commit and it works in Draw2d if scaling is 100% but not when scaling is any other value.

Here's the snippet from the original report using a `ScalableFreeformLayeredPane` with scaling of 2. The bottom rectangle line is not drawn. Values of 0.5, 1.5, are the same. Only a value of 1 works

```java
import org.eclipse.draw2d.Figure;
import org.eclipse.draw2d.FigureCanvas;
import org.eclipse.draw2d.Graphics;
import org.eclipse.draw2d.ScalableFreeformLayeredPane;
import org.eclipse.draw2d.geometry.Rectangle;
import org.eclipse.swt.graphics.Path;

public class LinuxClipGraphics {
    
        shell.setSize(600, 300);
        shell.setText("Clip Graphics");
        
        ScalableFreeformLayeredPane layeredPane = new ScalableFreeformLayeredPane();
        layeredPane.setOpaque(true);
        layeredPane.setBackgroundColor(new Color(255, 255, 255));
        layeredPane.setPreferredSize(580, 800);
        layeredPane.setScale(2);

        FigureCanvas canvas = new FigureCanvas(shell);
        canvas.setContents(layeredPane);
        
        Figure figure1 = new TestFigure();
        figure1.setBounds(new Rectangle(10, 120, 120, 120));
        layeredPane.add(figure1);
        
        Figure figure2 = new TestFigure();
        figure2.setBounds(new Rectangle(10, 300, 120, 120));
        layeredPane.add(figure2);
        
        shell.open();
        
        while(!shell.isDisposed()) {
                display.sleep();
            }
        }
    }
    
    private static class TestFigure extends Figure {
        Color background = new Color(181, 255, 255);
        
        @Override
        protected void paintFigure(Graphics graphics) {
            Rectangle rect = getBounds().getCopy();
            graphics.setBackgroundColor(background);
            graphics.fillRectangle(rect);
            
            graphics.setLineWidth(6);
            
            Path path = new Path(null);
            path.addRectangle(rect.x, rect.y, rect.width, rect.height);
            graphics.setClip(path);
            graphics.drawPath(path);
            path.dispose();
        }
    }
}
```

ptziegler:
I think there are some infra problems going on? All tests are crashing the moment the Surefire thread is started. Nonetheless, I've added/reactivated some tests to verify some assumptions.

The short version is that clipping on Linux is weird. When using a `Path`, more often than not, the whole client area is used as clip.

akurtakov:
I have noticed the test failure too but so far I have no clue what caused it.

akurtakov:
So the problem is https://github.com/eclipse-platform/eclipse.platform.swt/issues/3129

Phillipus:
I've been doing some trial and error tests and changing the following works in all the above cases:

```java
private void resetClipping() {
	/*
	 * Bug 531667: widgets paint over other widgets
	 *
	 * The Cairo handle is shared by all widgets, and GC.setClipping(0) allows painting outside the current GC area.
	 * So if we reset any custom clipping we still want to restrict GC operations with the initial GC clipping.
	 */
	// setClipping(clipping);
	setClipping(getDevice().getBounds());
}
```

ptziegler:
> I've been doing some trial and error tests and changing the following works in all the above cases:
> 
> ```java
> private void resetClipping() {
> 	/*
> 	 * Bug 531667: widgets paint over other widgets
> 	 *
> 	 * The Cairo handle is shared by all widgets, and GC.setClipping(0) allows painting outside the current GC area.
> 	 * So if we reset any custom clipping we still want to restrict GC operations with the initial GC clipping.
> 	 */
> 	// setClipping(clipping);
> 	setClipping(getDevice().getBounds());
> }
> ```

That pretty much confirms my suspicion that the underlying clipping rectangle is too small. If I understand everything correctly, clipping a `Path` is done by first setting the clip to the initial clipping region and then intersecting it with the given path. If this region is too small, you end up with an empty clip.

Phillipus:
> That pretty much confirms my suspicion that the underlying clipping rectangle is too small. If I understand everything correctly, clipping a Path is done by first setting the clip to the initial clipping region and then intersecting it with the given path. If this region is too small, you end up with an empty clip.

Is this a valid fix then, or an indicator of something else?

akurtakov:
Running tests problems should be fixed. Feel free to rebase to see results if/when you're ready.

HeikoKlare:
> I've been doing some trial and error tests and changing the following works in all the above cases:
> 
> ```java
> private void resetClipping() {
> 	/*
> 	 * Bug 531667: widgets paint over other widgets
> 	 *
> 	 * The Cairo handle is shared by all widgets, and GC.setClipping(0) allows painting outside the current GC area.
> 	 * So if we reset any custom clipping we still want to restrict GC operations with the initial GC clipping.
> 	 */
> 	// setClipping(clipping);
> 	setClipping(getDevice().getBounds());
> }
> ```

Please consider that there has been quite some discussion and work on the current implementation based on using an initially extracted clipping region rather than just taking the device bounds during reset:
- 093c579a4ffd9551acb901bba9617e7aa776989d
- https://bugs.eclipse.org/bugs/show_bug.cgi?id=531667

So I would be a bit afraid that the solution might not be as simple as just using the device bounds. Just applying the transform of the GC to the stored clipping region (or temporarily removing the transform from the GC when resetting the clipping) might be safer.

Phillipus:
> So I would be a bit afraid that the solution might not be as simple as just using the device bounds. Just applying the transform of the GC to the stored clipping region (or temporarily removing the transform from the GC when resetting the clipping) might be safer.

OK. Can you advise on how to do that?

HeikoKlare:
> OK. Can you advise on how to do that?

As a quick attempt, I would try to do something like this in `resetClipping()`:
```java
Transform transform = new Transform();
getTransform(transform);
setTransform(null);
setClipping(clipping);
setTransform(transform);
transform.dispose();
```
Likely there would be a better way, such as:
- Retrieve the transformation matrix and/or reset and reapply it in a "simpler"/less inefficient way
- Transform the clipping rectangle instead of modifying the Cairo instance

Phillipus:
> > OK. Can you advise on how to do that?
> 
> As a quick attempt, I would try to do something like this in `resetClipping()`:
> 
> ```java
> Transform transform = new Transform();
> getTransform(transform);
> setTransform(null);
> setClipping(clipping);
> setTransform(transform);
> transform.dispose();
> ```

Thanks!

I tried that and that works in all the above cases.

ptziegler:
> I tried that and that works in all the above cases.

Same for me. Time is always tight during the week, so I wanted to have a look again on the next weekend. But something like that was my idea as well.

Though given that this issue only appears for `Paths`, I would move this logic to the `setClipping(Path)` method.

Phillipus:
> Time is always tight during the week, so I wanted to have a look again on the next weekend. But something like that was my idea as well.

I appreciate greatly you taking the time to look at this, thank-you. :-)

Phillipus:
> Though given that this issue only appears for Paths, I would move this logic to the setClipping(Path) method.

Good idea. So it would be something like:

```java
public void setClipping(Path path) {
	if (handle == 0) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
	if (path != null && path.isDisposed()) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
	resetClippingForPath();
	if (path != null) {
		initCairo();
		long cairo = data.cairo;
		long copy = Cairo.cairo_copy_path(path.handle);
		if (copy == 0) SWT.error(SWT.ERROR_NO_HANDLES);
		Cairo.cairo_append_path(cairo, copy);
		Cairo.cairo_path_destroy(copy);
		Cairo.cairo_clip(cairo);
	}
}

private void resetClippingForPath() {
    Transform transform = new Transform(getDevice());
    getTransform(transform);
    setTransform(null);
    setClipping(clipping);
    setTransform(transform);
    transform.dispose();
}
```

Not sure about the name `resetClippingForPath`, though.

But in all my testing the general idea of saving and restoring the `Transform` works.

Phillipus:
@ptziegler Thanks for coming back to this. I tested your latest commit and it looks good!

Phillipus:
First of all thank-you to @ptziegler for your engagement on this issue and this patch, and thank-you to @HeikoKlare for the feedback.

Could this PR be merged now? In our RCP app I've re-written all of our `Draw2d` figures so that they are rendered using a `GC#Path` and `GC#setClpping`. All the figures look great on Windows and Mac, but it's this problem on Linux that stops me proceeding.

If there's any more that needs to be done, please let me know. :-)

ptziegler:
I'd wait until tomorrow in case someone has anything to add. Otherwise I'd merge it. The test failure is unrelated (see #3044).

ptziegler:
The test got stuck again, this time with the gtk4 tests. Merging, as the other Linux tests are successful and the BusyIndicator is unrelated to the GC clipping.

Phillipus:
@ptziegler Many thanks!

PR #3133, opened 2026-03-11
Ignore jars.extra.classpath unless we have a better solution

laeubi:


akurtakov:
This one is "simply" changing to previous way of building SWT so unless someone objects I plan to merge it tonight to unblock work on SWT.

laeubi:
Correct, I think this can be merged right away as it fixes the verification builds.

akurtakov:
Let's do it. Other work can continue without the time pressure.

PR #3147, opened 2026-03-26
[Gtk4] Fix Composite background not propagating

akurtakov:
Snapshot background before snapshoting children to actually propagate the background if not overriden.
Fixes one of the most notable problems on Gtk 4 - backgrounds being all white making components look out of place, not distinguish text fields from the rest and so on.


akurtakov:
Before:
image
After:
image

PR #3149, opened 2026-03-26
[GTK] Fix NPE in GC if already disposed

ptziegler:
Calling `stringExtent(...)` or `textExtent(...)` may throw a NullPointerException, rather than an SWTException if the GC object has already been disposed. This is unexpected and inconsistent with how GC behaves on other platforms.


akurtakov:
Thanks for catching this one!

Issue #3091, opened 2026-02-22
[Windows] Text is not displayed if the font has Underline or Strikeout and GC.setAdvanced(true)


Phillipus:
Windows 11
Java 21
Latest SWT from I-builds

This is moved from https://github.com/eclipse-gef/gef-classic/issues/1023

If a `Font` is used to draw text in a `GC` that has either Underline or Strikeout the text is not rendered if `GC.setAdvanced(true)`

The Underline or Strikeout for a Font can be set using a `FontDialog`:

<img width="427" height="471" alt="Image" src="https://github.com/user-attachments/assets/a2523943-4a01-4b16-af87-2ab3a19f9ae1" />

Run the Snippet:

```java
import org.eclipse.swt.graphics.Font;
import org.eclipse.swt.graphics.FontData;
import org.eclipse.swt.graphics.GC;
import org.eclipse.swt.layout.FillLayout;
import org.eclipse.swt.widgets.Display;
import org.eclipse.swt.widgets.Shell;

public class FontTest {
    
    public static void main(String[] args) {
        Display display = new Display();
        Shell shell = new Shell(display);
        shell.setLayout(new FillLayout());
        shell.setSize(200, 150);
        
        // Rather than open a FontDialog we'll create the Font from an equivalent FontData string which has underline set
        String fontString = "1|Arial|10.125|0|WINDOWS|1|-27|0|0|0|400|0|1|0|0|3|2|1|34|Arial";
        FontData fd = new FontData(fontString);
        Font font = new Font(display, fd);

        shell.addPaintListener(e -> {
            GC gc = e.gc;
            gc.setFont(font);

            gc.drawText("Hello World", 10, 10);
            
            gc.setAdvanced(true); // Comment this out to render text
            gc.drawText("Hello World", 10, 50);
        });

        shell.open();
        
        while(!shell.isDisposed()) {
            if(!display.readAndDispatch()) {
                display.sleep();
            }
        }
    }
}
```

It should look like this:

<img width="189" height="144" alt="Image" src="https://github.com/user-attachments/assets/8847681e-d458-46fd-8cbd-1015f7c0955d" />

But looks like this:

<img width="189" height="144" alt="Image" src="https://github.com/user-attachments/assets/a5ab91cc-8e13-4f2c-b2f8-3d25b6727691" />

HeikoKlare:
[...]

PR #3159, opened 2026-03-31
[Win32] Consider SWT.DRAW_DELIMITER flag when drawing text with GDI+

HeikoKlare:
The GC.drawText() implementation falls back to using GDI instead of GDI+ (even if enabled) in many cases for historic reasons. There were issues with rendering tabs back then. One of the few cases where GDI+ is used is when glyphs that are non existing for GDI are used. In that case, however, the DRAW_DELIMITER flag is not properly considered. The behavior is always as if the flag was enabled, i.e., line delimiters is always drawn even if the flag is not set.

This change corrects the behavior of the GDI+ text drawing implementation to consider if the DRAW_DELIMITER flag is not set. In consequence, strings with glyphs that are not supported by GDI are now rendered without line breaks if undesired as per defined flags.

Contributes to https://github.com/eclipse-platform/eclipse.platform.swt/issues/3091

This is the extraction of a fix that was included in https://github.com/eclipse-platform/eclipse.platform.swt/pull/3100

### How to test
The following snippet demonstrates the currently unintended behavior:

```java
	shell.setSize(250, 150);

	shell.addPaintListener(e -> {
		GC gc = e.gc;
		gc.drawText("\u4E2D\u6587:\r\n Should render on single line", 10, 10, SWT.DRAW_TAB);
		gc.setAdvanced(true);
		gc.drawText("\u4E2D\u6587:\r\n Should render on single line", 10, 50, SWT.DRAW_TAB);
	});

	shell.open();

	while (!shell.isDisposed()) {
			display.sleep();
		}
	}
}
```

#### Before
image

#### After
image

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
@@ -3037,6 +3037,9 @@ private RectF drawText(long gdipGraphics, char[] buffer, int start, int length, 
 private void drawTextGDIP(long gdipGraphics, String string, int x, int y, int flags, boolean draw, Point size) {
 	boolean needsBounds = !draw || (flags & SWT.DRAW_TRANSPARENT) == 0;
 	char[] buffer;
+	if ((flags & SWT.DRAW_DELIMITER) == 0) {
+		string = string.replaceAll("[\\r\\n]+", "");
+	}
 	int length = string.length();
 	if (length != 0) {
 		buffer = string.toCharArray();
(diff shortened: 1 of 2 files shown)


fedejeanne (approved):
Tested, works as expected. Thank you!

fedejeanne:
I don't know why the GTK3 test is taking so long but I already checked in GTK4 (logs) and the added test ran green ✅

HeikoKlare:
> I don't know why the GTK3 test is taking so long but I already checked in GTK4 (logs) and the added test ran green ✅

That's caused by:
- https://github.com/eclipse-platform/eclipse.platform.swt/issues/3044

Issue #2449, opened 2025-08-26
Edge browser engine does not properly load BrowserFunction


azoitl:
When using the Nebula RichtTextEditor in our Eclipse RCP (4diac IDE) on Windows we get an SWTException with `getAllOptions is not defined at` reported by `org.eclipse.swt.browser.Edge.evaluateInternal(Edge.java:960)`. See also the full stack trace below.

We tested this on latest Eclipse 2025-09 Milestone.

<details>
<summary>Unhandled event loop exception</summary>

### Stack Trace:
```
org.eclipse.swt.SWTException: getAllOptions is not defined
	at org.eclipse.swt.browser.Edge.evaluateInternal(Edge.java:960)
	at org.eclipse.swt.browser.Edge.evaluate(Edge.java:933)
	at org.eclipse.swt.browser.WebBrowser.evaluate(WebBrowser.java:405)
	at org.eclipse.swt.browser.Browser.evaluate(Browser.java:731)
	at org.eclipse.swt.browser.Browser.evaluate(Browser.java:680)
	at org.eclipse.nebula.widgets.richtext.RichTextEditor$3.completed(RichTextEditor.java:285)
	at org.eclipse.swt.browser.Edge.lambda$43(Edge.java:1176)
	at org.eclipse.swt.widgets.RunnableLock.run(RunnableLock.java:40)
	at org.eclipse.swt.widgets.Synchronizer.runAsyncMessages(Synchronizer.java:132)
	at org.eclipse.swt.widgets.Display.runAsyncMessages(Display.java:4133)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:3749)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine$5.run(PartRenderingEngine.java:1151)
	at org.eclipse.core.databinding.observable.Realm.runWithDefault(Realm.java:339)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine.run(PartRenderingEngine.java:1042)
	at org.eclipse.e4.ui.internal.workbench.E4Workbench.createAndRunUI(E4Workbench.java:153)
	at org.eclipse.ui.internal.Workbench.lambda$3(Workbench.java:678)
	at org.eclipse.core.databinding.observable.Realm.runWithDefault(Realm.java:339)
	at org.eclipse.ui.internal.Workbench.createAndRunWorkbench(Workbench.java:583)
[...]

PR #3166, opened 2026-04-01
[Win32][Edge] Fix BrowserFunction race condition using AddScriptToExecuteOnDocumentCreated

### LINKED ISSUE #20: [Edge][BrowserFunction] BrowserFunctions added to window undefined when URL loaded
I have ran into an issue on the 2021-09 release where BrowserFunctions are sometimes undefined in the window when using the Edge browser. This looks to be caused by some sort of race condition that occurs when a Browser has BrowserFunctions registered and its URL is set quickly after it is constructed. When more time is added between the construction of the Browser instance and registering browser functions/setting the URL, the error occurs far less frequently. This issue is unique to Edge as the code works without issue when using IE.

From briefly looking over the code and documented changes, this looks to be related to [execute(String script) no longer being synchronous](https://www.eclipse.org/swt/faq.php#edgelimitations) when using the Edge browser implementation. When adding a browser function to the window, createFunction(BrowserFunction function) [calls nonBlockingExecute(String script)](https://github.com/eclipse-platform/eclipse.platform.swt/blob/e9f0cda446cb1981839bee8da24483f6cb4a48d6/bundles/org.eclipse.swt/Eclipse%20SWT%20Browser/common/org/eclipse/swt/browser/WebBrowser.java#L368) to execute the JS used to define the browser function in the window. The browser's own implementation of [execute(String script)](https://github.com/eclipse-platform/eclipse.platform.swt/blob/e9f0cda446cb1981839bee8da24483f6cb4a48d6/bundles/org.eclipse.swt/Eclipse%20SWT%20Browser/common/org/eclipse/swt/browser/WebBrowser.java#L399-L403) is then called to run the script. This then asynchronously adds the browser function to the window.

It might be possible to correct this by adding an override implementation of createFunction(BrowserFunction function) to [Edge.java](https://github.com/eclipse-platform/eclipse.platform.swt/blob/e9f0cda446cb1981839bee8da24483f6cb4a48d6/bundles/org.eclipse.swt/Eclipse%20SWT%20Browser/win32/org/eclipse/swt/browser/Edge.java) and using a similar strategy as the Edge evaluate(String script) does to [wait for the callback](https://github.com/eclipse-platform/eclipse.platform.swt/blob/e9f0cda446cb1981839bee8da24483f6cb4a48d6/bundles/org.eclipse.swt/Eclipse%20SWT%20Browser/win32/org/eclipse/swt/browser/Edge.java#L455) to report completion before continuing. 
A good alternative would also be adding support for Window.postMessage() as mentioned in the [Potential API addons](https://git.eclipse.org/r/plugins/gitiles/platform/eclipse.platform.swt/+/refs/heads/master/bundles/org.eclipse.swt/Readme.WebView2.md#potential-api-additions). With asynchronous options being commonly preferred in web development the synchronous behavior of browser functions will likely become harder to maintain and less desired as time goes on.

HeikoKlare:
## Summary

Fixes #20 (see also #2449 for a reproduction snippet).

In Edge/WebView2, `execute()` is asynchronous. When `new BrowserFunction()` is called, the injection script is queued via `ExecuteScript()`, but if a page navigation completes before WebView2 processes that queued script, the function is unavailable in the new document. This race is most easily triggered when two `Browser` instances are created in quick succession, as the second browser's initialization forces event-loop processing that can advance the first browser's navigation past the injection window.

**Fix:** Override `createFunction()` in `Edge` to register every function script via `AddScriptToExecuteOnDocumentCreated`. This WebView2 API guarantees the script runs on **every** future document creation, before any page scripts — eliminating the race condition. The script ID returned by the async registration is stored so it can be cleaned up via `RemoveScriptToExecuteOnDocumentCreated` when the `BrowserFunction` is disposed.

When `createFunction()` is called from within a WebView2 callback (`inCallback > 0`), blocking on the registration would deadlock, so the persistent registration is deferred via `Display.asyncExec()` to complete once the callback returns. This ensures **every** function is always persistently registered, removing the need for any `NavigationStarting` fallback re-injection.

## Changes

- `ICoreWebView2.java`: Add missing `RemoveScriptToExecuteOnDocumentCreated` vtbl binding (index 28, between `AddScriptToExecuteOnDocumentCreated` at 27 and `ExecuteScript` at 29)
- `Edge.java`: Override `createFunction()` to persistently register every function via `AddScriptToExecuteOnDocumentCreated` (deferred via `asyncExec` when inside a callback); override `deregisterFunction()` to call `RemoveScriptToExecuteOnDocumentCreated` on disposal; remove the `NavigationStarting` fallback re-injection since all functions are now always persistently registered
- `Test_org_eclipse_swt_browser_Browser.java`: Two regression tests (`isEdge`-only):
  - `test_BrowserFunction_availableBeforePageScripts_issue20`: registers a function, navigates to a page whose inline `<script>` immediately calls it, asserts the function was invoked (tests the `AddScriptToExecuteOnDocumentCreated` guarantee directly)
  - `test_BrowserFunction_availableOnLoad_concurrentInstances_issue20`: creates two browsers concurrently without waiting for initialization, verifies both functions are callable when their pages complete loading (replicates the original bug report timing)

## Test plan

- [x] Run `test_BrowserFunction_availableBeforePageScripts_issue20` on Windows with Edge
- [x] Run `test_BrowserFunction_availableOnLoad_concurrentInstances_issue20` on Windows with Edge
- [x] Run existing `test_BrowserFunction_*` tests to verify no regressions
- [x] Manually verify with the reproduction snippet from [#2449 comment](https://github.com/eclipse-platform/eclipse.platform.swt/issues/2449#issuecomment-3233601556)

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
@@ -85,6 +85,8 @@ class Edge extends WebBrowser {
 	boolean inNewWindow;
 	private boolean inEvaluate;
 	HashMap<Long, LocationEvent> navigations = new HashMap<>();
+	/** Maps BrowserFunction index to the script ID from AddScriptToExecuteOnDocumentCreated. */
+	private final Map<Integer, String> functionScriptIds = new HashMap<>();
 	private boolean ignoreGotFocus;
 	private boolean ignoreFocusIn;
 	private String lastCustomText;
@@ -1146,14 +1148,6 @@ int handleNavigationStarting(long pView, long pArgs, boolean top) {
 	navigations.put(pNavId[0], event);
 	if (event.doit) {
 		settings.put_IsScriptEnabled(jsEnabledOnNextPage);
-		// Register browser functions in the new document.
-		if (!functions.isEmpty()) {
-			StringBuilder sb = new StringBuilder();
-			for (BrowserFunction function : functions.values()) {
-				sb.append(function.functionString);
-			}
-			execute(sb.toString());
-		}
 	} else {
 		args.put_Cancel(true);
 	}
@@ -1834,6 +1828,51 @@ private boolean setWebpageData(String url, String postData, String[] headers, St
 @Override
 public boolean setUrl(String url, String postData, String[] headers) {
 	return setWebpageData(url, postData, headers, null);
+}
+
+/**
+ * Registers the function script persistently via AddScriptToExecuteOnDocumentCreated so it is
+ * injected on every future document creation before any page scripts run, avoiding the race
+ * condition between async function injection and navigation completion.
+ * If called while inside a WebView2 callback, the persistent registration is deferred via
+ * {@link Display#asyncExec(Runnable)} so it completes once the callback returns.
+ * See <a href="https://github.com/eclipse-platform/eclipse.platform.swt/issues/20">issue #20</a>.
+ */
+@Override
+public void createFunction(BrowserFunction function) {
+	super.createFunction(function);
+	int functionIndex = function.index;
+	String functionString = function.functionString;
+	if (inCallback > 0) {
+		// Cannot wait for a callback result while already inside a WebView2 callback;
+		// defer the persistent registration to after the callback completes.
+		browser.getDisplay().asyncExec(() -> {
+			if (browser.isDisposed() || !functions.containsKey(functionIndex)) return;
+			registerFunctionScript(functionIndex, functionString);
+		});
+		return;
+	}
+	registerFunctionScript(functionIndex, functionString);
+}
+
+private void registerFunctionScript(int functionIndex, String functionString) {
+	String[] scriptId = new String[1];
+	callAndWait(scriptId, completion ->
+		webViewProvider.getWebView(true).AddScriptToExecuteOnDocumentCreated(
+			stringToWstr(functionString), completion.getAddress()));
+	if (scriptId[0] != null) {
+		functionScriptIds.put(functionIndex, scriptId[0]);
+	}
+}
+
+@Override
+void deregisterFunction(BrowserFunction function) {
+	super.deregisterFunction(function);
+	String scriptId = functionScriptIds.remove(function.index);
+	if (scriptId != null) {
+		webViewProvider.getWebView(true).RemoveScriptToExecuteOnDocumentCreated(
+			stringToWstr(scriptId));
+	}
 }
 
 }
--- a/bundles/org.eclipse.swt/Eclipse SWT PI/win32/org/eclipse/swt/internal/ole/win32/ICoreWebView2.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT PI/win32/org/eclipse/swt/internal/ole/win32/ICoreWebView2.java
@@ -67,6 +67,10 @@ public int AddScriptToExecuteOnDocumentCreated(char[] javaScript, long handler) 
 	return COM.VtblCall(27, address, javaScript, handler);
 }
 
+public int RemoveScriptToExecuteOnDocumentCreated(char[] id) {
+	return COM.VtblCall(28, address, id);
+}
+
 public int ExecuteScript(char[] javaScript, IUnknown handler) {
 	return COM.VtblCall(29, address, javaScript, handler.address);
 }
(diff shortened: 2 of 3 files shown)


HeikoKlare:
Test plan is done. The automated tests are run via CI anyway. I also successfully executed the reproduction snippet: https://github.com/eclipse-platform/eclipse.platform.swt/issues/2449#issuecomment-3233601556

@sratz can you please review this Edge enhancement?

HeikoKlare:
@sratz It would be great if this could make it into the upcoming release. Next week is M2 promotion and we should probably not merge something like this after M2. Do you see the chance that you can have a look until then?

sratz:
Aren't the functions now executed twice, as the handling in `org.eclipse.swt.browser.Edge.handleNavigationStarting(long, long, boolean)` still executes them all? Shouldn't that part be removed now?

Also,

`createFunction` internally calls `registerFunction`.

I think it would be nice if we did not need to override `createFunction` here, but only `registerFunction` for symmetry reasons with `deregisterFunction` below. But we require the `function.functionString` to be already set.

sratz:
> Aren't the functions now executed twice, as the handling in `org.eclipse.swt.browser.Edge.handleNavigationStarting(long, long, boolean)` still executes them all? Shouldn't that part be removed now?

Ah, it's not the function itself being executed twice, but only the registration of it is done twice, so `handleNavigationStarting` just replaces/re-sets redundantly what `AddScriptToExecuteOnDocumentCreated` already have added.

HeikoKlare:
> Ah, it's not the function itself being executed twice, but only the registration of it is done twice, so handleNavigationStarting just replaces/re-sets redundantly what AddScriptToExecuteOnDocumentCreated already have added.

Good catch. I have adapted the change so that functions are only registered ones by using the stored `functionScriptIds` to identify if a function was already registered.

> Also,
>
> `createFunction` internally calls `registerFunction`.
>
> I think it would be nice if we did not need to override `createFunction` here, but only `registerFunction` for symmetry reasons with `deregisterFunction` below. But we require the `function.functionString` to be already set.

I agree that for the sake of symmetry, overwriting `registerFunction` would be better, but for the reason you have given that would require a larger refactoring (that would affect all platforms and not only Edge). We do not want to do that, do we?

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -1146,13 +1148,21 @@ int handleNavigationStarting(long pView, long pArgs, boolean top) {
...
+		// AddScriptToExecuteOnDocumentCreated (i.e. those created inside a WebView2
+		// callback where the persistent registration was skipped as a fallback).
+		// Functions that do have a persistent registration are already guaranteed to
+		// run before any page scripts and must not be injected a second time here.
 		if (!functions.isEmpty()) {
 			StringBuilder sb = new StringBuilder();
 			for (BrowserFunction function : functions.values()) {
-				sb.append(function.functionString);
+				if (!functionScriptIds.containsKey(function.index)) {
+					sb.append(function.functionString);
```
sratz: should we attempt to register here again via AddScriptToExecuteOnDocumentCreated? So that for the next navigation iteration we go the new route?

HeikoKlare: It sounds conceptually clean to try to eventually register every function via `AddScriptToExecuteOnDocumentCreated`. Actually, the current solution is quite unclean, as the way how a function is registered now depends on whether the function has initially been reigstered from a callback or not.

I wondered if we could just do everything via `AddScriptToExecuteOnDocumentCreated` without this "fallback" here by just calling that registration asynchronously if we are in a callback. I've adapted the PR accordingly. The only potential issue I see with that is an execution order in which the function is registered in a callback, and then the next browser navigation is started is before the `asyncExec` is processed. I am not sure if that's a relevant scenario we need to consider.

An alternative would be to just temporarily use `AddScriptToExecuteOnDocumentCreated` for the current navigation and to deregister the function right after the next navigation again, so that the default way of registering functions remains the embedding in the document done in `handleNavigationStarting`.

One issue with the solution of using `AddScriptToExecuteOnDocumentCreated` is that it seems to be quite slow. The execution time of `test_TimerRegression_Issue2806` (which creates 10000 browser functions) increases from less than 1 second to more than 2 minutes on my machine.

sratz: Code looks good.

I profiled the test execution and most of the time is actually spent in `callAndWait(...)` -> `org.eclipse.swt.internal.win32.OS.PeekMessage[native]`, probably because of very little OS event activity.

I guess this will not be an actual issue in production.

jonahgraham (approved):

Issue #3060, opened 2026-02-05
Display#wake must likley be volatile


laeubi:
Linux + Mac have both a `boolean wake` field, that is evaluated in `Display#sleep` inside a loop.
Even though the method `wake()` is synchronized on the display and marked with "this method may be called from any thread", the sleep method is not synchronizing when checking this flag.

As wake is not marked as _volatile_ the JVM is allowed to cache the value of wake in a register and only write it back when the method ends and do not see other threads write in the meanwhile.

PR #3167, opened 2026-04-01
Use timer unconditionally in BusyIndicator.showWhile(Future)

akurtakov:
Should prevent hangs like
https://github.com/eclipse-platform/eclipse.platform.swt/issues/3044

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/BusyIndicator.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/BusyIndicator.java
@@ -106,6 +106,19 @@ public class BusyIndicator {
 			} else {
 				Integer busyId = setBusyCursor(display);
 				try {
+					// Install a timer as a safety net to periodically wake the event loop in case
+					// the primary wake signal is lost (e.g. due to a native call may fail to wake the polling function).
+					// That should be short enough to not be noticeable by the user and long
+					// enough to not burn more CPU time than necessary
+					int wakeTime = 10;
+					display.timerExec(wakeTime, new Runnable() {
+						@Override
+						public void run() {
+							if (!future.isDone() && !display.isDisposed()) {
+								display.timerExec(wakeTime, this);
+							}
+						}
+					});
 					if (future instanceof CompletionStage<?> stage) {
 						// let us wake up from sleep once the future is done
 						stage.handle((nil1, nil2) -> {
@@ -118,19 +131,6 @@ public class BusyIndicator {
 								}
 							}
 							return null;
-						});
-					} else {
-						// for plain features we need to use a workaround, we install a timer every
-						// few ms, that should be short enough to not be noticeable by the user and long
-						// enough to not burn more CPU time than necessary
-						int wakeTime = 10;
-						display.timerExec(wakeTime, new Runnable() {
-							@Override
-							public void run() {
-								if (!future.isDone() && !display.isDisposed()) {
-									display.timerExec(wakeTime, this);
-								}
-							}
 						});
 					}
 					while (!future.isDone() && !display.isDisposed()) {


HeikoKlare (approved):
Thank you for proposing a solution for the BusyIndicator getting stuck. Having that fixed so that CI executions terminate regularly again will significantly improve the contributor experience ,

This appears to be a reasonable solution for avoiding that the BusyIndicator gets stuck. It does not add any considerable overhead but makes the implementation more robust against the `wake()` call once the Future has finished not having the intended effect.
I see two possible improvements but consider neither of them mandatory for further processing the PR:
1. After existing the display spinning loop (because the future is done and the display was woken up) the timerExec might be cancelled to avoid an unnecessary wakeup of the display (but it should not be a problem to have that unnecessary wakup though).
2. Using this timer even with a `CompletionStage` future might be limited to platforms on which we know about problems with `wake()` calls (i.e., in particular Linux). But as said above, the overhead is so minimal that it may be considered irrelevant, so making the behavior platform-specific may introduce unnecessary complexity.

This change might also be reverted once https://github.com/eclipse-platform/eclipse.platform.swt/issues/3060 is addressed.

And there is small chance that the underlying issue is addressed by https://github.com/eclipse-platform/eclipse.platform.swt/pull/3161 (as also indicated in https://github.com/eclipse-platform/eclipse.platform.swt/pull/3161#pullrequestreview-4042717868), so we should wait for some builds with that fix and see if the BusyIndicator test still gets stuck.

HeikoKlare:
Turns out that https://github.com/eclipse-platform/eclipse.platform.swt/pull/3161 did not fix the issue of the BusyIndicator test getting stuck.

https://github.com/eclipse-platform/eclipse.platform.swt/pull/3166 was just rebased on master including the the mentioned change, but the current test execution got stuck at the BusyIndicator test again:
https://github.com/eclipse-platform/eclipse.platform.swt/actions/runs/23844044308/job/69506754851?pr=3166

laeubi:
@HeikoKlare if the problem is currently visible then I think the next easy fix would be to try out making it volatile:

- https://github.com/eclipse-platform/eclipse.platform.swt/issues/3060

HeikoKlare:
> It changes production code that never was reported to be stuck instead of fixing the problem (we just do guess working here unless someone was able to reliable reproduce it) and hiding a possibly serve bug in SWT.

It's based on your own assessment that the issue might only be visible on test execution but not in productive usage because there are other events happening because of human input that wake up the display: https://github.com/eclipse-platform/eclipse.platform.swt/issues/3044#issuecomment-3846235139

> I think the next easy fix would be to try out making it volatile:
> 
> * [Display#wake must likley be volatile #3060](https://github.com/eclipse-platform/eclipse.platform.swt/issues/3060)

Yes, we could do that. But then we should proceed with that soon. We could also just revert this change once https://github.com/eclipse-platform/eclipse.platform.swt/issues/3060 is addressed and proves to fix the issue. I just don't think we should further wait for other options to maybe fix the issue if no one is working on them while we have a proposed solution that seems to work and does not harm. The CI builds freqeuently getting stuck are quite annoying and limit productivity, so have a timely solution (even if it's a temporary one) would be great.

akurtakov:
I fully agree that current state can not continue . I would say - if we don't have another fix in and tests still stuck on Monday, let's merge this one as it's already 2 months with this state.

laeubi:
> I fully agree that current state can not continue . I would say - if we don't have another fix in and tests still stuck on Monday, let's merge this one as it's already 2 months with this state.

I disagree here: The change is not backed by any proof it will actually fix the stucked test, this is just guess working, I see not any trace that this is guaranteed to work and it changes code I use every day and never was hanging since it was introduced.

In the worst scenario I would better *disable* the test than changing the implementation for strange CI behavior we see here.

I already suggested a workaround that would only change test code and gave more insights... I now prepare a PR with my proposal and then we can see.

akurtakov:
There have been 2 full rebuilds of this PR without any of the verification jobs stuck. I've just rebased it for third run - if nothing is stuck again , this is a proof enough for me as there haven't been any other PR lately that had even 2 good builds in a row. 
Let's see what the state is on Monday.

HeikoKlare:
CI builds get stuck so often because of the issue that we will easily see if this fix has an impact or not once it has been merged and can, if necessary, simply revert it.
I have to admit that I do not fully understand why the root cause being in production code is challenged so heavily now, as I understood [this comment](https://github.com/eclipse-platform/eclipse.platform.swt/issues/3044#issuecomment-3846235139) as an assumption of exactly that and and explanation why it does not produce any noticable issue during productive use though.

Anyway, I think we have a good proposal/agreement now: let's have a look at the next proposals, such as the upcoming one from Christoph, and see if we can make progress with that. If we do not find a more suitable solution until Monday, let's proceed with this one.

laeubi:
> There have been 2 full rebuilds of this PR without any of the verification jobs stuck. I've just rebased it for third run - if nothing is stuck again , this is a proof enough for me as there haven't been any other PR lately that had even 2 good builds in a row.

I have builds with 20 reruns and without an issue, so that's not really a proof. Anyways I still think it is not the right think to fix a test failure only ever reproducible on CI and only ever through certain times.

akurtakov:
Analysis have shown better ways to do in the original issue.

Issue #3150, opened 2026-03-27
Simplify CTabFolderRenderer to use flat tabs


vogella:
@HeikoKlare suggested that CTabFolderRenderer should use flat tabs instead of rounded ones (which IMHO look really bad).

@HeikoKlare please approve so that can I start working on this. This will be a visible change for SWT UIs but I think most users will welcome it and IMHO we are allowed to improve our L&F.

HeikoKlare:
I am in favor of simplifying the SWT implementation of the CTabFolder rendering via `CTabFolderRenderer` in addition to the simplification of the CSS-based rendering in Platform UI.

I think we have discussed in a community call with @azoitl last year that we should maybe have a deprecation period for such a change (also for the one in Platform UI). But since it's not a breaking change (in the sense that an application may not start/work anymore) we could also skip that for the sake of faster progress. Having such a simplification pending for two years seems inappropriate given the fast moving world 🙂 

Alois also mentioned that removing the rounded tabs in `CTabFolderRenderer` would be beneficial for moving forward here:
https://github.com/eclipse-platform/eclipse.platform.ui/issues/3806#issuecomment-4141204819

azoitl:
Just one word of caution. Event  tough the rounded edges are not API `CTabRendering` in Platform UI is partly using this and even access some values via reflection. So when we move forward here which I would strongly support I think cleaning `CTabRendering` first  and then moving to  `CTabFolderRender` would be the better approach.

vogella:
@azoitl I think the following fields are accessed via reflection and I don't think they are related to rounded corners.

  - selectionGradientVertical                                                                                                                                                                                                                                                                                                                                                                                                       
  - gradientVertical                                                                                                                                                                                                                                                                                                                                                                                                                
  - selectionGradientColors                                                                                                                                                                                                                                                                                                                                                                                                         
  - selectionGradientPercents                                                                                                                                                                                                                                                                                                                                                                                                       
  - gradientColors                                                                                                                                                                                                                                                                                                                                                                                                                  
  - gradientPercents

azoitl:
Interesting I was pretty sure when we did the analysis for starting the work on the vertical tab folders there was something. But also the history is rather clear.

HeikoKlare:
I have to admit that I am always a bit confused about the different types of rounded tabs and how to configure them, so maybe some of the confusion is caused by that and we should clean that up. As far as I know, we have three shapes of tab header:
[...]

PR #3168, opened 2026-04-01
Remove curved tab support from CTabFolder and CTabFolderRenderer

vogella:
Curved tabs are no longer supported. This change simplifies the rendering logic by enforcing square corners and removing all curve-related calculations and legacy highlight gradients.

CTabFolder:
* Deprecate getSimple() and setSimple(boolean).
* Make setSimple(boolean) a no-op to prevent enabling curved tabs.
* Remove layout adjustments based on curve dimensions (curveIndent).
* Remove setSelectionHighlightGradientColor and all usages.

CTabFolderRenderer:
* Remove curve-related fields (curve, curveWidth, curveIndent, etc.).
* Remove selection highlight caching and associated gradient methods.
* Delete obsolete no-op methods including updateCurves(), drawHighlight(), and disposeSelectionHighlightGradientColors().
* Simplify rendering logic:
  - Replace all corner constants with EMPTY_CORNER.
  - Enforce square corners regardless of the 'simple' flag.
  - Streamline drawTabArea and border rendering methods.
  - Remove curve-related width adjustments from computeTrim.
  - Remove anti-aliasing logic (unnecessary for square tabs).
* Force useEllipses() to always return true.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/3150

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
@@ -1251,7 +1251,7 @@ public int getSelectionIndex() {
  * @since 3.0
  * @deprecated Curved tabs are no longer supported.
  */
-@Deprecated(forRemoval = true, since = "2026-04")
+@Deprecated(forRemoval = true, since = "2026-06")
 public boolean getSimple() {
 	checkWidget();
 	return simple;
@@ -2369,7 +2369,6 @@ public void reskin(int flags) {
 @Override
 public void setBackground (Color color) {
 	super.setBackground(color);
-	renderer.createAntialiasColors(); //TODO: need better caching strategy
 	updateBkImages(true);
 	redraw();
 }
@@ -2498,7 +2497,6 @@ public void setBackground(Color[] colors, int[] percents, boolean vertical) {
 @Override
 public void setBackgroundImage(Image image) {
 		super.setBackgroundImage(image);
-		renderer.createAntialiasColors(); //TODO: need better caching strategy
 		redraw();
 }
 /**
@@ -3288,7 +3286,6 @@ public void setSelectionBackground (Color color) {
 	if (selectionBackground == color) return;
 	if (color == null) color = getDisplay().getSystemColor(SELECTION_BACKGROUND);
 	selectionBackground = color;
-	renderer.createAntialiasColors(); //TODO:  need better caching strategy
 	if (selectedIndex > -1) redraw();
 }
 /**
@@ -3449,7 +3446,6 @@ public void setSelectionBackground(Image image) {
 		selectionGradientPercents = null;
 	}
 	selectionBgImage = image;
-	renderer.createAntialiasColors(); //TODO:  need better caching strategy
 	if (selectedIndex > -1) redraw();
 }
 /**
@@ -3504,7 +3500,7 @@ public void setSelectionBarThickness(int thickness) {
  * @since 3.0
  * @deprecated Curved tabs are no longer supported.
  */
-@Deprecated(forRemoval = true, since = "2026-04")
+@Deprecated(forRemoval = true, since = "2026-06")
 public void setSimple(boolean simple) {
 	checkWidget();
 }
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
@@ -33,16 +33,6 @@ public class CTabFolderRenderer {
 	protected CTabFolder parent;
 
 	Color fillColor;
-	/* Selected item appearance */
-	/* Colors for anti-aliasing */
-	Color selectedOuterColor = null;
-	Color selectedInnerColor = null;
-	Color tabAreaColor = null;
-	/*
-	 * Border color that was used in computing the cached anti-alias Colors.
-	 * We have to recompute the colors if the border color changes
-	 */
-	Color lastBorderColor = null;
 
 	private Font chevronFont = null;
 
@@ -163,9 +153,6 @@ public class CTabFolderRenderer {
 		this.parent = parent;
 	}
 
-	void antialias (int[] shape, Color innerColor, Color outerColor, GC gc){
-	}
-
 	/**
 	 * Returns the preferred size of a part.
 	 * <p>
@@ -255,14 +242,9 @@ public class CTabFolderRenderer {
 						int minChars = parent.minChars;
 						text = minChars == 0 ? null : item.getText();
 						if (text != null && text.length() > minChars) {
-							if (useEllipses()) {
-								int end = minChars < ELLIPSIS.length() + 1 ? minChars : minChars - ELLIPSIS.length();
-								text = text.substring(0, end);
-								if (minChars > ELLIPSIS.length() + 1) text += ELLIPSIS;
-							} else {
-								int end = minChars;
-								text = text.substring(0, end);
-							}
+							int end = minChars < ELLIPSIS.length() + 1 ? minChars : minChars - ELLIPSIS.length();
+							text = text.substring(0, end);
+							if (minChars > ELLIPSIS.length() + 1) text += ELLIPSIS;
 						}
 					} else {
 						text = item.getText();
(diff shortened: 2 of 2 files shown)


vogella:
Snippets to test:

```

import org.eclipse.swt.*;
import org.eclipse.swt.custom.*;

public class CTabFolderCurvedTest {

	shell.setText("CTabFolder Curved Test");

	CTabFolder folder = new CTabFolder(shell, SWT.BORDER);
	folder.setSimple(false); // This should now have no effect (remain square)
	
	for (int i = 0; i < 4; i++) {
		CTabItem item = new CTabItem(folder, SWT.CLOSE);
		item.setText("Item " + i);
		Text text = new Text(folder, SWT.MULTI);
		text.setText("Content for item " + i);
		item.setControl(text);
	}
	
	folder.setSelection(0);
	
	shell.setSize(400, 300);
	shell.open();
	while (!shell.isDisposed()) {
			display.sleep();
	}
	display.dispose();
}
}
```

```

import org.eclipse.swt.*;
import org.eclipse.swt.custom.*;

public class CTabFolderGradientTest {

	shell.setText("CTabFolder Gradient Test");

	CTabFolder folder = new CTabFolder(shell, SWT.BORDER);
	folder.setSimple(false); // Should remain square
	
	Color[] colors = new Color[] {
		display.getSystemColor(SWT.COLOR_TITLE_BACKGROUND),
		display.getSystemColor(SWT.COLOR_TITLE_BACKGROUND_GRADIENT),
		display.getSystemColor(SWT.COLOR_TITLE_INACTIVE_BACKGROUND)
	};
	int[] percents = new int[] {50, 100};
	folder.setSelectionBackground(colors, percents, true);
	
	for (int i = 0; i < 4; i++) {
		CTabItem item = new CTabItem(folder, SWT.CLOSE);
		item.setText("Item " + i);
		Text text = new Text(folder, SWT.MULTI);
		text.setText("Content for item " + i);
		item.setControl(text);
	}
	
	folder.setSelection(0);
	
	shell.setSize(400, 300);
	shell.open();
	while (!shell.isDisposed()) {
			display.sleep();
	}
	display.dispose();
}
}
```



Before:

image

image


After (curved support removed):

image

image

vogella:
@HeikoKlare can you have a look?

vogella:
Planning to merge today so that it gets into M1.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
```diff
@@ -1250,7 +1249,9 @@ public int getSelectionIndex() {
  * @return <code>true</code> if the CTabFolder is rendered with a simple shape
  *
  * @since 3.0
+ * @deprecated Curved tabs are no longer supported.
  */
+@Deprecated(forRemoval = true, since = "2026-04")
```
HeikoKlare:
```suggestion
@Deprecated(forRemoval = true, since = "2026-06")
```

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
```diff
@@ -3519,13 +3502,11 @@ public void setSelectionBarThickness(int thickness) {
  * </ul>
  *
  * @since 3.0
+ * @deprecated Curved tabs are no longer supported.
  */
+@Deprecated(forRemoval = true, since = "2026-04")
```
HeikoKlare:
```suggestion
@Deprecated(forRemoval = true, since = "2026-06")
```

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
```diff
@@ -202,45 +164,6 @@ protected CTabFolderRenderer(CTabFolder parent) {
 	}
 
 	void antialias (int[] shape, Color innerColor, Color outerColor, GC gc){
```
HeikoKlare: Shouldn't this method be removed completely?

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
```diff
@@ -1798,52 +1519,11 @@ String shortenText(GC gc, String text, int width, String ellipses) {
...
-					15+d, 10+d,  16+d, 10+d,
-					17+d, 11+d,  18+d, 11+d,  19+d, 11+d,
-					20+d, 12+d,  21+d, 12+d,  22+d,  12+d };
-		}
-	}
-
 	/*
 	 * Return whether to use ellipses or just truncate labels
 	 */
 	boolean useEllipses() {
```
HeikoKlare: Why not completely remove this method?

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
```diff
@@ -3429,22 +3424,12 @@ public void setSelectionBackground(Color[] colors, int[] percents, boolean verti
 		}
 		selectionGradientVertical = vertical;
 		setSelectionBackground(selectionGradientColors[selectionGradientColors.length-1]);
-		setSelectionHighlightGradientColor(highlightBeginColor);
```
HeikoKlare, line 3432: This seems (and all other changes related to `selectionHightlightGradientColor` seem to be unintended/unrelated. They remove the support for gradients in tab header, which is not the intent of this PR. This is only about removing curved tab support. So all those changes should be removed from this PR. In case we want to remove gradients as well, we should discuss that separately.

vogella: Did you see my snippets? Gradient support is not removed just the special case for curved tabs.

setSelectionHighlightGradientColor drew a gradient highlight sheen along the top edge of the selected tab the visual "glow" that curved tabs used to have to make them look 3D/rounded.

It only made visual sense when tabs had curved corners. The renderer's implementation (CTabFolderRenderer.setSelectionHighlightGradientColor) even had an explicit guard:

````
  if (parent.selectionGradientColors.length < 2)
      return; // don't bother if no background gradient
````
And the drawing code in drawSelected() only painted the highlight when parent.simple was false (i.e., curved mode):

````
  if (parent.selectionHighlightBarThickness > 0 && parent.simple) { ... }
````

Since the PR removes curved tab support entirely — all tabs are now always rendered with square corners, simple is effectively always true and the curved  rendering path is gone, the highlight gradient serves no purpose. There's no curved shape to "light up" anymore.

So the removal is safe because:
  1. The visual effect it produced only looked correct on curved tabs
  2. The rendering path that used it (drawHighlight, the gradient cache) is also being deleted
  3. setSimple(false) (the trigger for curved mode) is now a no-op, so the condition under which the highlight was ever drawn can never be reached

File: bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java (hunk 4 above)
Comment (vogella) [reply bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java]: make sense

File: bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java [hunk 6]
```diff
@@ -1250,7 +1249,9 @@ public int getSelectionIndex() {
  * @return <code>true</code> if the CTabFolder is rendered with a simple shape
  *
  * @since 3.0
+ * @deprecated Curved tabs are no longer supported.
  */
+@Deprecated(forRemoval = true, since = "2026-06")
 public boolean getSimple() {
 	checkWidget();
 	return simple;
```
HeikoKlare, line 1257: The `simple` field should be removed, so that it is not accidentally used anymore (or is there reflective access to it)? Then least it should be final. But we can also do that as a follow up.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
HeikoKlare: Alright. Sorry for the confusion because of the false association I had. I had opened the PR page before the snippets were added and it didn't refresh, so I did not see your snippets.
With the explanation, it of course makes sense to get rid of that gradient functionality. I wasn't even aware of that and it's great to see that even more complexity can be removed with the removal of curved tabs then.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
HeikoKlare, line 948: This calculation and probably much code around for the calculations of the shapes could even be further simplified. Probably the whole usage of EMPTY_CORNER could be removed. But that's nothing we need to do in this PR as well, but we can just simplify as a follow up. I can also take that action item.

PR #3169, opened 2026-04-01
[Gtk4] Fix CTabFolder blank content after tab switch

vogella:
When a Composite has both fixedHandle and handle (e.g. CTabFolder with SWT.BORDER), snapshotDrawProc fires for both SwtFixed instances because it is installed on the SwtFixed class. The outer fixedHandle snapshot fires after children are rendered, causing onPaint to overdraw tab content after a visibility change (tab switch and return). Skip snapshotToDraw for the outer fixedHandle, matching the identical guard already in Canvas.


akurtakov:
This change doesn't fix the problem, actually I don't see any change in behavior.

vogella:
Like discussed via chat, unfortunately GTK4 runtime Eclipse does not work for me so I cannot iterate on this patch or test it.

PR #3172, opened 2026-04-01
Simplify shape calculations in CTabFolderRenderer

HeikoKlare:
With the removal of non-simple (curved) header in CTabFolders, the remaining calculations suited for variable corner shapes are overly complex. Most calculations now include coordinates from an EMPTY_CORNER dummy, which just contains a single point of values 0.

This change simplifies all calculations currently including EMPTY_CORNER.

See https://github.com/eclipse-platform/eclipse.platform.swt/pull/3168#discussion_r3022448575


vogella:
https://github.com/eclipse-platform/eclipse.platform.swt/pull/3174 is the other approach to that.

vogella:
LGTM, planning to merge after build verification and afterwards rework https://github.com/eclipse-platform/eclipse.platform.swt/pull/3174 on top of this one considering also @HeikoKlare feedback

PR #3173, opened 2026-04-01
Remove obsolete "simple" field from CTabFolder

HeikoKlare:
With the recent removal of curved tabs from CTabFolder/CTabFolderRenderer, the CTabFolder property "simple" is effectively always true. In order to avoid accidental usages and changes of the according field, this change removes the field and makes the accessors just always yield true.

See https://github.com/eclipse-platform/eclipse.platform.swt/pull/3168#discussion_r3022380040


vogella:
Please do not merge before at least one successful I-Build. platform.ui might still be accessing that fields. I'm currently looking at that

HeikoKlare:
Yes, let's wait for next I-Build. I have also checked Platform UI for accesses and did not find any. I just found `CTabFolderWrapper` in `CTabRendering` doing reflective access to `CTabFolder`. The `simple` property only seems to be accessed via proper accessors.

PR #3174, opened 2026-04-01
Replace polyline/Region drawing with direct line and rect calls in CT…

vogella:
…abFolderRenderer

Now that all tab corners are square, the polyline-based shape approach is unnecessary. Replace with direct gc.drawLine() / gc.drawRectangle() calls:

- Remove EMPTY_CORNER constant (was {0,0}, i.e. no corner offset)
- Remove drawBorder(GC, int[] shape) — callers now draw lines directly
- Simplify drawLeftUnselectedBorder / drawRightUnselectedBorder to a single gc.drawLine() each
- Remove Region-based polygon clipping from drawBackground(); since all shapes are rectangular the clipping was a no-op — remove the shape parameter from the signature entirely
- Remove fillRegion() — no longer called
- Simplify shape arrays in drawSelected and drawTabArea from loop-built polylines (that iterated over EMPTY_CORNER) to inline int[] literals
- Remove the Region.subtract(shape) / fillRegion block in drawTabArea that painted parent background into curved-corner cutouts (corners are now square so this region was always empty)

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
@@ -473,10 +473,10 @@ public class CTabFolderRenderer {
 		int[] percents = selected ? parent.selectionGradientPercents : parent.gradientPercents;
 		boolean vertical = selected ? parent.selectionGradientVertical : parent.gradientVertical;
 
-		drawBackground(gc, null, bounds.x, bounds.y, bounds.width, bounds.height, defaultBackground, image, colors, percents, vertical);
+		drawBackground(gc, bounds.x, bounds.y, bounds.width, bounds.height, defaultBackground, image, colors, percents, vertical);
 	}
 
-	void drawBackground(GC gc, int[] shape, boolean selected) {
+	void drawBackground(GC gc, boolean selected) {
 		Color defaultBackground = selected && parent.shouldHighlight() ? parent.selectionBackground : parent.getBackground();
 		Image image = selected ? parent.selectionBgImage : null;
 		Color[] colors = selected && parent.shouldHighlight() ? parent.selectionGradientColors : parent.gradientColors;
@@ -495,19 +495,10 @@ public class CTabFolderRenderer {
 			x += 1; width -= 2;
 		}
 		int y = parent.onBottom ? size.y - borderBottom - height : borderTop;
-		drawBackground(gc, shape, x, y, width, height, defaultBackground, image, colors, percents, vertical);
+		drawBackground(gc, x, y, width, height, defaultBackground, image, colors, percents, vertical);
 	}
 
-	void drawBackground(GC gc, int[] shape, int x, int y, int width, int height, Color defaultBackground, Image image, Color[] colors, int[] percents, boolean vertical) {
-		Region clipping = null, region = null;
-		if (shape != null) {
-			clipping = new Region();
-			gc.getClipping(clipping);
-			region = new Region();
-			region.add(shape);
-			region.intersect(clipping);
-			gc.setClipping(region);
-		}
+	void drawBackground(GC gc, int x, int y, int width, int height, Color defaultBackground, Image image, Color[] colors, int[] percents, boolean vertical) {
 		if (image != null) {
 			// draw the background image in shape
 			gc.setBackground(defaultBackground);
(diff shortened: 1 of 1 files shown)


vogella:
@HeikoKlare this can be done in additional to your changes, please review

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
```diff
@@ -643,15 +618,8 @@ void drawBody(GC gc, Rectangle bounds, int state) {
...
 				}
-				// If horizontal gradient, show gradient across the whole area
-				if (selectedIndex != -1 && parent.selectionGradientColors != null && parent.selectionGradientColors.length > 1 && !parent.selectionGradientVertical) {
-					drawBackground(gc, shape, true);
-				} else if (selectedIndex == -1 && parent.gradientColors != null && parent.gradientColors.length > 1 && !parent.gradientVertical) {
-					drawBackground(gc, shape, false);
-				} else {
-					gc.setBackground(selectedIndex != -1 && parent.shouldHighlight() ? parent.selectionBackground : parent.getBackground());
-					gc.fillPolygon(shape);
-				}
```
HeikoKlare, line 652: I do not understand how the different cases became obsolete and were thus removed.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
```diff
@@ -717,8 +685,7 @@ void drawClose(GC gc, Rectangle closeRect, int closeImageState) {
 				break;
 			}
 			case SWT.BACKGROUND: {
-				int[] shape = new int[] {x,y, x+10,y, x+10,y+10, x,y+10};
-				drawBackground(gc, shape, false);
+				drawBackground(gc, x, y, 10, 10, parent.getBackground(), null, null, null, false);
```
HeikoKlare, line 688: I do not easily understand why this is equivalent to the previous call either.

HeikoKlare:
I've merged https://github.com/eclipse-platform/eclipse.platform.swt/pull/3172, so this could be rebased on top and then implicitly becomes a bit more simple.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
vogella: The three branches applied gradient fill to the `highlight_margin` L-shaped frame. That frame is **2 pixels wide** (`highlight_margin = 2` for non-FLAT; 0 for FLAT). Rendering a horizontal gradient across a 2-pixel strip is visually indistinguishable from a solid color — the gradient collapses to a single color sample at that scale. `gc.fillPolygon(shape)` with the solid background color produces an equivalent visual result.

bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
vogella: The `SWT.BACKGROUND` case repaints the close button area to erase a previously drawn hot/selected close icon and restore the background. The tab background (including any gradient) was already painted by `drawTabArea`/`drawSelected` before `drawClose` is called. The old `drawBackground(gc, shape, false)` applied `parent.gradientColors` over a **10×10 pixel** area — a gradient fragment at that scale is visually indistinguishable from a solid color. Using `parent.getBackground()` correctly restores the area with no perceptible difference.

vogella:
Thank you for the detailed review (and the AI-assisted analysis)!

Addressing all three `drawBackground` concerns:

**`drawBody` — gradient on highlight_margin frame**
The gradient was applied to an L-shaped frame that is **2px wide** (`highlight_margin = 2` for non-FLAT, 0 for FLAT). A horizontal gradient across 2 pixels is visually indistinguishable from a solid color — the gradient collapses to essentially a single color sample at that scale. Solid fill is equivalent here. (Also replied inline.)

**`drawClose SWT.BACKGROUND` — gradient removed**
This case erases a previously drawn hot/selected close icon by repainting the background. The tab's actual gradient background was already painted by `drawTabArea`/`drawSelected` before `drawClose` is invoked. The old path applied `parent.gradientColors` over a **10×10 pixel** area — a gradient fragment that small is also indistinguishable from solid. Using `parent.getBackground()` is correct. (Also replied inline.)

**`drawSelected` highlight header — draw area widened**
The old shape-clipped call drew into a strip of height `highlight_header - 1` = **2px**. The new `drawBackground(gc, parent.shouldHighlight())` draws over the full tab header height, but tab items are rendered on top of that area immediately afterward — the extra painted area is never visible.

All three cases involve rendering gradients into areas of ≤2px height or ≤10px total size, or areas immediately overdrawn by tab content — which is consistent with your own observation that visual testing shows no regression.

vogella:
@HeikoKlare any additional feedback?

HeikoKlare:
I am currently on vacation and will not be able to have another look before Thursday. If you are confident with the changes, don't wait for me and just move on. IIRC my feedback was just on not understanding certain changes. If you consider them correct, the changes are fine for me.

vogella:
Thursday is not to far away, enjoy your vacation. I will be in forced absense Thursday and Friday so feel free to merge if you reviewed.

vogella:
I split this PR up into multiple smaller changes. I pretty confident that the change here is correct but giving this beautiful spagetti code, its good to have individual commits which can be reverted to reworked if necessary.

vogella:
https://github.com/eclipse-platform/eclipse.platform.swt/pull/3217 and #3218 are the first two split outs. Small and save I would say.

vogella:
Converting this to draft, as we want to split the changes into smaller bits.

HeikoKlare:
> #3217 and #3218 are the first two split outs. Small and save I would say.

Agreed. I've merged them.

vogella:
Closing this for now, this takes at the moment to much time from me, may investigate this later again.

PR #3176, opened 2026-04-01
[Gtk] Motion notify signal is Gtk 3 only

akurtakov:
Rename the handler to gtk3_motion_notify_event to make it obvious and simplify implementations to not try to handle Gtk 4 case as it will never end there.


akurtakov:
Failing test is unstable one - https://github.com/eclipse-platform/eclipse.platform.swt/issues/3044 . Merging.

PR #3218, opened 2026-04-10
Remove fillRegion() and Region.subtract block in CTabFolderRenderer

vogella:
With square tab corners, the Region.subtract block in drawTabArea that painted parent background into curved-corner cutouts was always empty. Remove it along with the now-unused fillRegion() helper.


vogella:
Fast you are Jedi master @HeikoKlare, I was still testing...

But thanks for the merge, up to the next split out...

PR #3220, opened 2026-04-10
Simplify and clean up Win32 Cursor implementation

HeikoKlare:
## Summary

This pull request simplifies and cleans up the Win32-specific `Cursor` class (`bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Cursor.java`) and adds a corresponding set of Win32-specific unit tests.

> **Note:** These changes were developed with the assistance of [GitHub Copilot](https://github.com/features/copilot).

---

## Code Simplifications & Improvements

| Change | Reason |
|--------|--------|
| Replace double map-lookup + `setHandleForZoomLevel()` with `HashMap.computeIfAbsent()` in `win32_getHandle()` | The helper method contained redundant null/containsKey guards; both are collapsed into a single idiomatic call |
| Change `win32_getHandle()` return type from boxed `Long` to primitive `long` | Consistent with `Font.win32_getHandle()`, `Image.win32_getHandle()`, and `Region.win32_getHandle()`; eliminates unnecessary boxing overhead; all call sites already used `long` |
| Fix latent equality bug in `equals()`: `Long` wrapper objects compared with `==` | Cursor handles are OS memory addresses far outside the JVM integer cache range `[-128, 127]`, so `==` on `Long` could silently return `false` for equal handles. With the primitive return type, `==` on `long` is now correct |
| Use Java 16 instanceof pattern matching in `equals()` | Eliminates the redundant explicit cast after the type check |
| Remove unnecessary `clone()` in `ImageDataWithMaskCursorHandleProvider.validateMask()` | The cloned `ImageData` was only ever read for its `width`/`height` fields — the defensive copy was pure overhead |
| Remove redundant `static` modifier on the nested `CursorHandleProvider` interface | Nested interfaces are implicitly `static` in Java |
| Remove redundant `final` modifier on `private static` method `getOSCursorIdFromStyle()` | `private static` methods cannot be overridden; `final` is meaningless here |
| Remove redundant `final` modifier on lambda-captured local variable in `destroyHandlesExcept()` | The variable is already effectively final without the explicit keyword |

---

## New Win32-Specific Tests

A new test class `CursorWin32Tests` was added alongside the existing Win32 test classes, following the same patterns used in `RegionWin32Tests` and `ImagesWin32Tests`.

| Test | What it covers |
|------|----------------|
| `testDisposedCursorReturnsZeroHandle` | Verifies the `isDisposed()` guard at the top of `win32_getHandle()` returns `0L` after disposal |
| `testHandleIsCachedForSameZoomLevel` | Verifies the `computeIfAbsent()` caching — repeated calls with the same zoom return the identical cached OS handle |
| `testImageDataCursorProducesDifferentHandlesForDifferentZoomLevels` | Verifies that `ImageDataCursorHandleProvider` scales the source image and produces a distinct OS cursor handle per zoom level |
| `testDestroyHandlesExceptPreservesRetainedHandle` | Verifies that `destroyHandlesExcept()` leaves the retained zoom entry intact without marking the cursor as disposed |

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/graphics/CursorWin32Tests.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/graphics/CursorWin32Tests.java
@@ -1,0 +1,98 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Vector Informatik GmbH and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.swt.graphics;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+import java.util.*;
+
+import org.eclipse.swt.*;
+import org.eclipse.swt.internal.*;
+import org.eclipse.swt.widgets.*;
+import org.junit.jupiter.api.*;
+import org.junit.jupiter.api.extension.*;
+
+@ExtendWith(PlatformSpecificExecutionExtension.class)
+@ExtendWith(WithMonitorSpecificScalingExtension.class)
+class CursorWin32Tests {
+
+	private Display display;
+
+	@BeforeEach
+	void setUp() {
+		display = Display.getDefault();
+	}
+
+	@Test
+	void testDisposedCursorReturnsZeroHandle() {
+		Cursor cursor = new Cursor(display, SWT.CURSOR_ARROW);
+		cursor.dispose();
+		assertEquals(0L, Cursor.win32_getHandle(cursor, 100), "A disposed cursor must return a zero handle");
+	}
+
+	@Test
+	void testHandleIsCachedForSameZoomLevel() {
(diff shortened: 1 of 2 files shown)


vogella:
The new tests look like they are not so helpful, I would not not test win32_getHandle in isolation.

HeikoKlare:
> The new tests look like they are not so helpful, I would not not test win32_getHandle in isolation.

Why do you think so? Those are actual unit tests (which SWT is mostly lacking). Some of them may not be that useful, but having core functionality of per-zoom handle provision, non-redundancy etc. regression tested sounds reasonable to me to make SWT develop less "let's hope we don't break anything, but we have sufficient time to find implicitly find regressions until the release" 🙂

vogella:
The only useful tests seem to be the last two once to me (starting with         void testImageDataCursorProducesDifferentHandlesForDifferentZoomLevels). 

But if the "a cursor creates a handler" tests are useful to you, that is fine for me. Just keep an open eye for AI slope, no need to pollute our repos with unnecessary tests.

HeikoKlare:
> But if the "a cursor creates a handler" tests are useful to you, that is fine for me. Just keep an open eye for AI slope, no need to pollute our repos with unnecessary tests.

Sure, we must not accept whatever an agent proposed, so I have an eye on slop. As said, I agree that some tests are rather useless, which is why I removed them. I just kept those tests that ensure important contracts of the Win32 cursor implementation.

PR #3224, opened 2026-04-11
[Win32] Simplify and fix Region class

HeikoKlare:
## Summary

This change cleans up the win32 `Region.java` implementation by removing duplication, fixing a formatting bug, and applying idiomatic Java patterns throughout.

### Bug fix

- **`toString()` produced malformed output**: The method opened a `{` but never closed it, producing strings like `Region {handle(zoom:100)` instead of `Region {handle(zoom:100)}`. This made debugging harder and was inconsistent with the disposed-region case which does close its braces.

### Simplifications

- **`getRegionHandle(int)`**: Replaced the manual `containsKey` / `put` / `get` triple with `Map.computeIfAbsent`, which is the idiomatic Java way to lazily populate a map and avoids two redundant lookups.

- **`getBounds()`**: A block lambda containing only a single `return` was converted to an expression lambda.

- **`isEmpty()`**: The `RECT` object was allocated *before* the lambda, even though it is only needed inside it. Moved it inside the lambda to keep allocation and use co-located and to avoid capturing a mutable object across a closure boundary unnecessarily.

### Eliminated code duplication

- **`OperationWithRectangle`**: The three methods `addInPixels`, `subtractInPixels`, and `intersectInPixels` were virtually identical — each created a rect region, called `CombineRgn` with a different mode constant, then deleted the region. These were replaced by a single `combineWithRectInPixels(handle, x, y, w, h, mode)` helper, reducing ~15 lines to ~5.

- **`OperationWithArray`**: The same pattern applied to `addInPixels` and `subtractInPixels`, collapsed into `combineWithPolyInPixels(handle, pointArray, mode)`.

- **`OperationWithRegion`**: Three block lambdas each containing a single `return OS.CombineRgn(...)` call were converted to expression lambdas, removing the syntactic noise.

### Access modifiers

- Removed redundant `public` from the constructors of `OperationWithArray` and `OperationWithPoint`. Both classes are declared `private static`, so `public` on their constructors has no semantic effect and is misleading.

## Test plan

- [ ] Existing `Region`-related JUnit tests pass without modification
- [x] Verify `toString()` output now includes the closing `}` for non-disposed regions
- [x] Smoke-test region operations (add, subtract, intersect, translate) on Windows

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Region.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Region.java
@@ -288,9 +288,8 @@ public boolean equals (Object object) {
  */
 public Rectangle getBounds () {
 	if (isDisposed()) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
-	return applyUsingAnyHandle(regionHandle -> {
-		return Win32DPIUtils.pixelToPoint(getBoundsInPixels(regionHandle.handle()), regionHandle.zoom());
-	});
+	return applyUsingAnyHandle(regionHandle ->
+		Win32DPIUtils.pixelToPoint(getBoundsInPixels(regionHandle.handle()), regionHandle.zoom()));
 }
 
 private Rectangle getBoundsInPixels(long handle) {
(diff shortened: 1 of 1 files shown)


vogella:
I did not check in the code (sorry)but would that cleanup also make sense for Linux ans Mac Implementation or why is this "only" done for windows?

HeikoKlare:
The `Resource` implementations (Region, Transform, Image etc.) are significantly different between Windows and macOS/Linux (due to Windows requiring handles for multiple zooms for proper multi-monitor scaling support, which is non-existent in the limited HiDPI support of macOS and Linux). So, a cleanup may be performed on the Linux and macOS implementations as well, but the result would be significantly different. I chose to apply this to the Windows implementations (first) as those have been recently revised (for monitor-specific scaling).

Issue #3219, opened 2026-04-10
CTabFolder: Improve readability, performance, simplification, and test coverage


vogella:
## Overview

CTabFolder (4,257 lines), CTabFolderRenderer (1,338 lines), CTabItem (540 lines), and CTabFolderLayout (123 lines) have significant opportunities for improvement in readability, performance, code simplification, and test coverage.

---

## 1. Readability Improvements

- [ ] **Extract and simplify `updateItems()` (CTabFolder)**: 83-line method with 4 near-identical loops computing tab widths. Extract into helper methods and add comments explaining the priority-based visibility algorithm.
- [ ] **Split `drawSelected()` (CTabFolderRenderer)**: 187-line method with deep nesting and shape array index arithmetic. Split into shape construction, highlight drawing, and text rendering.
- [ ] **Split `computeControlBounds()` (CTabFolder)**: 182-line method mixing layout, positioning, and wrapping. Split into leading/trailing control computation.
- [ ] **Replace magic numbers with named constants**: e.g. `16` for button width (CTabFolder), unnamed margins/spacing in CTabFolderRenderer.
- [ ] **Fix misleading field initialization in CTabItem**: `int x,y,width,height = 0;` — only `height` is explicitly initialized (others default to 0 by JLS but this is misleading).

## 2. Performance Improvements

- [ ] **Replace array copying in `addTabControl()`/`removeTabControl()` (CTabFolder)**: 4 parallel arrays are fully copied on each add/remove. Use `ArrayList` or grow-by-factor strategy.
- [ ] **Cache `renderer.computeSize()` results in `updateItems()`**: Called 4+ times per item in nested loops. Cache results to avoid redundant computation.
- [x] **Avoid Rectangle allocation on every mouse move** (CTabFolder `onMouseMove()`): Reuse a field instead of `new Rectangle()` in hot path.
- [ ] ~~**Reduce Region allocation in `drawBackground()` (CTabFolderRenderer)**: Creates and disposes Region objects on every call.~~
- [ ] **Cache chevron image in `updateChevronImage()`**: Currently disposes and recreates on every call.

## 3. Code Simplification

- [ ] **Extract duplicate min/max button logic in `updateButtons()` (CTabFolder)**: Near-identical blocks for min/max button creation. Extract to shared helper method.
- [ ] **Unify close button positioning in `setItemLocation()` (CTabFolder)**: Separate code paths for single/multi mode share most logic — consolidate.
[...]

PR #3229, opened 2026-04-13
CTabFolderRenderer: Reuse Region objects in drawBackground()

vogella:
This PR addresses a performance task in #3219 by reusing Region objects in CTabFolderRenderer.drawBackground().

### Changes:
- Added  and  private fields to .
- Lazily initialize these regions in .
- Reuse these regions for clipping and shape intersection to avoid frequent allocation and disposal of native OS resources.
- Updated  to properly clean up the reused regions.

Verified with  (119 tests passed).


vogella:
Copilots analysis is correct, not worth the risk or introducing platform specifc diferences for the little gain: Reusing Region instances here can cause unbounded growth of Region state on Win32.

Issue #2616, opened 2025-10-13
SkiaCanvas: a GC implementation with GPU support


DenisUngemach:
We develop a "Skia Canvas for SWT" that utilizes the Skia 2D graphics library for drawing on the canvas in SWT. As Java binding we use Skija.
The code currently under development can be found here: https://github.com/swt-initiative31/skija-canvas
This feature has several advantages:
 
- Skia is an established 2D graphics library that can be used across multiple operating systems.
- Hardware acceleration is supported for various backends, including OpenGL, Vulkan, Metal, and DirectX12.
- A raster implementation is also available.
- There are two Java bindings: Skija and Skiko. If one binding is deprecated, we can switch to the other.
- big parts of this feature can be developed OS independent
 
To achieve this, a new fragment, org.eclipse.swt.skia, will be created next to the org.eclipse.swt plugin in the SWT repository: https://github.com/eclipse-platform/eclipse.platform.swt. 
All jar dependencies for Skija (these include Skia) will reside in this new plugin.
However, some code modifications to the SWT plugin and its fragments are necessary.
 
We have discussed the API with application developers, who prefer making minimal modifications to their code to use the Skia canvas. 
 
They favor an API resembling the following:
 
-----------------------------------------------
 
SkiaConfiguration.applyRasterCanvas(); // Configures to Skia Raster Canvas.
Canvas c = new Canvas(parent, SWT.NONE); // This canvas will use a Skia delegate, directing specific drawing calls to Skia.
SkiaConfiguration.resetCanvasConfiguration(); // Subsequent canvases will revert to the classical canvas.
 
------------------------------------------------
 
This approach removes the necessity for modifications to class hierarchies. However, if the canvas configuration isn't reset, all future canvases will be Skia canvases. 
Conversely, application developers can opt to switch entirely to Skia canvases. 
This configuration method offers flexibility for multiple scenarios.
 
For the drawing events, the class GC will be refactored from a final class to a sealed class, which is extended by the final class GCExtension. 
This GCExtension delegates all drawing commands with an interface to SkiaGC, which interprets them for Skia.
Thus, application developers might not even notice a difference between drawing events with the classical GC and an embedded "SkiaGC."
 
Please let us know whether this concept is feasible for you.
 
This project requires much resources, so we would like to know whether this development offers advantages for you and whether it can be integrated into the SWT master branch.
 
[...]

PR #3231, opened 2026-04-14
Skia Canvas plugin for hardware-accelerated 2D rendering in SWT

DenisUngemach:
Skia Canvas plugin for hardware-accelerated 2D rendering in SWT.
[Experimental Feature]

- Integrates an SWT’s canvas extension framework, which replaces the
native backend for canvases with the Skia drawing framework which relies in this implementation on
OpenGL.
- adds a new fragment of org.eclipse.swt: bundles/org.eclipse.swt.skia
- Includes core classes for rendering
- resource management
- a factory for SWT integration with the ServiceLoader
- Implements caching for fonts, images, and text, supports DPI scaling
- The plugin is modular an minimizes API changes
- there are some tests in the org.eclipse.swt.skia plugin
- for more information read the skia.md file in bundles/org.eclipse.swt.skia
- uses the service loader, if there are bugs at the loading, the skia plugin will be ignored.

- Since the API was modified the SWT version should be increased (still to do)
- there seems to be an api tools bug because of sealed classes. This happens only in the github PR checks.
- The second commit contains only the ignores for the api tools. It can be reverted and then the versioning could be fixed properly.


DenisUngemach:
Referring to the issue: https://github.com/eclipse-platform/eclipse.platform.swt/issues/2616

vogella:
I think we need another PR to add the Skia libraries to the buiild. @HeikoKlare can you guide what needs to be done for that?

DenisUngemach:
> I think we need another PR to add the Skia libraries to the buiild. @HeikoKlare can you guide what needs to be done for that?

In the current change the source jars will also be loaded to the lib folder. On my builds this was only a warning, it seems here it is an error.
I could just remove this handling.

HeikoKlare:
> I think we need another PR to add the Skia libraries to the buiild. @HeikoKlare can you guide what needs to be done for that?

From my understanding, we need to add the library to the eclipse-sdk-prereqs target, so they become available in our target platform. Just like we did here for the JSVG library used in the SWT SVG fragment:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2776

I can provide a PR once I find the time to do so. But feel free to just do it inspired by the linked PR even earlier.

DenisUngemach:
> > I think we need another PR to add the Skia libraries to the buiild. @HeikoKlare can you guide what needs to be done for that?
> 
> From my understanding, we need to add the library to the eclipse-sdk-prereqs target, so they become available in our target platform. Just like we did here for the JSVG library used in the SWT SVG fragment:
> 
> * [Add jsvg 1.7.0 to eclipse-sdk target eclipse.platform.releng.aggregator#2776](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2776)
> 
> I can provide a PR once I find the time to do so. But feel free to just do it inspired by the linked PR even earlier.

Is this right: https://github.com/swt-initiative31/eclipse.platform.releng.aggregator/pull/1

Of course i will open then the PR to the original releng repo, if this should do it.

bundles/org.eclipse.swt.skia/src/org/eclipse/swt/internal/graphics/SkiaGC.java
```diff
@@ -0,0 +1,1257 @@
...
+		return attributes;
+	}
+
+	LineAttributes getLineAttributesInPixels() {
+		return new LineAttributes(lineWidth, lineCap, lineJoin, lineStyle, lineDashes, dashOffset, miterLimit);
+	}
+
+	@Override
+	public Rectangle getClipping() {
+		return currentClipBounds;
```
ptziegler: The clip returned by this method should never be `null`, even if no clipping is set. From the JavaDoc:

```
If no clipping region is set, the return value will be a rectangle which covers the entire bounds of the object the receiver is drawing on.
```

```suggestion
		if (currentClipRegion != null) {
			return currentClipBounds;
		}
		return canvas.getBounds();
```

Here an example that throws an exception with Skia but not with GTK:

```java
public class Test {
		Shell shell = new Shell();
		shell.setSize(200, 200);
		Canvas canvas = new Canvas(shell, SWT.SKIA);
		canvas.addPaintListener(event -> {
			GC gc = event.gc;
			Objects.requireNonNull(gc.getClipping());
		});
		shell.open();
		while (!shell.isDisposed()) {
				display.sleep();
			}
		}
	}
}
```

bundles/org.eclipse.swt.skia/src/org/eclipse/swt/internal/graphics/SkiaGC.java
```diff
@@ -0,0 +1,1257 @@
...
+public class SkiaGC implements IExternalGC {
+
+	private static boolean logImageNullError = true;
+	private final Surface surface;
+	private int alpha = 255;
+	private int antialias;
+	private boolean xorModeActive;
+	private int lineWidth = 1;
+	private int lineCap = SWT.CAP_FLAT;
+	private int lineStyle;
```
ptziegler: If the line style is never set, 0 is returned, which is an illegal value.

```suggestion
	private int lineStyle = SWT.LINE_SOLID;
```

This will throw an IllegalArgumentException with Skia, but not with GTK.

```
public class Test {
		Shell shell = new Shell();
		shell.setSize(200, 200);
		Canvas canvas = new Canvas(shell, SWT.SKIA);
		canvas.addPaintListener(event -> {
			GC gc = event.gc;
			gc.setLineStyle(gc.getLineStyle());
		});
		shell.open();
		while (!shell.isDisposed()) {
				display.sleep();
			}
		}
	}
}
```

ptziegler:
I'm currently getting a consistent crash with the Draw2D TreeExample when changing the minor/major spacing. I'm not quite sure yet what _exactly_ is going wrong, but it seems like the Skia surface is being disposed.

Immediately before the crash, the following returns `true`: `surface.isClosed()`

Perhaps as a first step, the following check needs to be added to every method, similar to the other GC's?

```
if (surface.isClosed()) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
```

Here the crash log:
[hs_err_pid138498.log](https://github.com/user-attachments/files/27088458/hs_err_pid138498.log)

ptziegler:
I also had a quick look at how it looks with the GEF Logic Editor and most figures seem to be drawn as expected. There seems to be an issue when using `Path`s though, as some of the gates don't show up in the editor.

image

DenisUngemach:
@ptziegler Thanks a lot for the support. I have already some changes that must be added to the PR. I will also check your remarks.

bundles/org.eclipse.swt.skia/src/org/eclipse/swt/internal/graphics/SkiaGC.java
DenisUngemach: Fixed.

bundles/org.eclipse.swt.skia/src/org/eclipse/swt/internal/graphics/SkiaGC.java
DenisUngemach: Fixed.

DenisUngemach:
-> Rebase
-> Fixed the easy issues from @ptziegler 
-> The issue with the crash is still open. I will also check this one.
-> Updated the skija dependency. Now the update site will be used and no longer the maven dependencies directly.
See: See here for setting the target of the sdk releng: https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3800

But this makes it much more difficult to load other targets and still use org.eclipse.swt.skia. 
Maybe @merks can provide a good description how to do that in the best way.

I guess using the update site https://download.eclipse.org/tools/orbit/simrel/orbit-aggregation/milestone/S202604250819 in other targets should work. 
But i have not yet fully found out how to make this run easily.

merks:
@DenisUngemach 

I see no reason not to commit the TP changes from.  **Oops**, I hit some buttons thinking I was updating the branch and actually did merge it now.  😬   But I don't think that should be a problem.  Those IUs will not be pulled into the final product or into the p2 repository until something bundle/fragment/feature actually requires it...

I guess that solves the current problem though?  😀

DenisUngemach:
> @DenisUngemach
> 
> I see no reason not to commit the TP changes from. **Oops**, I hit some buttons thinking I was updating the branch and actually did merge it now. 😬 But I don't think that should be a problem. Those IUs will not be pulled into the final product or into the p2 repository until something bundle/fragment/feature actually requires it...
> 
> I guess that solves the current problem though? 😀

@merks Of course i am completely fine with that 😀.

I will then check the last bug from @ptziegler where the surface is closed and the SkiaGC still seems to be in use. 
I am really wondering where this come from....

ptziegler:
Apologies for the slow response. I wanted to further test the extension but didn't really get the chance. 

> I will then check the last bug from @ptziegler where the surface is closed and the SkiaGC still seems to be in use.
I am really wondering where this come from....

I don't have a simple reproducer yet but as far as I can tell, the problem is that a resize event may be fired within the paint event. Because the resize event invalidates the Skija surface, it also invalidates the GC of the paint event.

Below a screenshot I got during debugging. As you can see the event is fired while painting, specifically when enabling/disabling the scrollbar.

image

Either the local variable in the `SkiaGC` class needs to be updated when the "old" surface is invalidated or the local variable should be removed completely. Otherwise you have no way of knowing when this variable becomes invalid.

ptziegler:
Just as an open question: How difficult would it be to hook the SkiaGC into the GC tests?

ptziegler:
> I don't have a simple reproducer yet but as far as I can tell, the problem is that a resize event may be fired within the paint event. Because the resize event invalidates the Skija surface, it also invalidates the GC of the paint event.

This example should print `false`, then `true` and then run into a segmentation fault when calling `gc.fillRectangle(...)`.

```java
public class Test {
		Shell shell = new Shell();
		shell.setSize(400, 400);
		
		Canvas canvas = new Canvas(shell, SWT.SKIA);

		canvas.addPaintListener(event -> {
			GC gc = event.gc;
			System.out.println(gc.isDisposed());
			canvas.setSize(100, 100);
			System.out.println(gc.isDisposed());
			gc.fillRectangle(0, 0, 100, 100);
		});
		
		shell.open();

		while(!shell.isDisposed()) {
			while(!display.readAndDispatch()) {
				display.sleep();
			}
		}
	}
}
```

If you set a breakpoint inside the `SkiaGC` object, you'll notice that its surface is closed, but the surface of the `SkiaGlCanvasExtension` is not. So the local variable of the `SkiaGC` object is not updated after the resize event.

DenisUngemach:
Hello, the resize bug should be fixed now. Now  the surface will be created and recreate in the paint workflow, if it is necessary.

But i think the plugin setup must be improved. Executing the tests of org.eclipse.swt.skia in the IDE fails now because the build path does not work properly.

merks:
What's wrong with the build path?  What can I do to help?  Are you using an Oomph setup; if not then all bets are off because you'll need to do everything, whatever that may be, manually....

bundles/org.eclipse.swt.skia/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,23 @@
...
+ lib/skija-windows-x64-0.143.11.jar,
+ lib/types-0.1.1.jar,
+ lib/,
+ ./
+Bundle-Name: %fragmentName
+Bundle-Vendor: %providerName
+Bundle-Localization: fragment
+Bundle-RequiredExecutionEnvironment: JavaSE-21
+Eclipse-ExtensibleAPI: true
+Fragment-Host: org.eclipse.swt;bundle-version="[3.134.0.qualifier,4.0.0)"
```
HeikoKlare: We should make a standalone bundle out of this.

bundles/org.eclipse.swt.skia/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,23 @@
...
+Bundle-ManifestVersion: 2
+Bundle-SymbolicName: org.eclipse.swt.skia
+Bundle-Version: 3.134.0.qualifier
+Automatic-Module-Name: org.eclipse.swt.skia
+Bundle-ClassPath: lib/skija-linux-x64-0.143.11.jar,
+ lib/skija-shared-0.143.11.jar,
+ lib/skija-windows-x64-0.143.11.jar,
+ lib/types-0.1.1.jar,
+ lib/,
+ ./
```
HeikoKlare: These can be removed as they are now provided via target platform.

bundles/org.eclipse.swt.skia/.classpath
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<classpath>
+	<classpathentry exported="true" kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
```
HeikoKlare: The plug-in dependencies container must not be exported. This will lead to problems in projects depending on this (such as a test project).

bundles/org.eclipse.swt.skia/.classpath
```diff
@@ -0,0 +1,18 @@
...
+		</attributes>
+	</classpathentry>
+	<classpathentry kind="src" path="src"/>
+	<classpathentry kind="src" path="resources"/>
+	<classpathentry kind="src" output="bin_test" path="SWT Skia Tests">
+		<attributes>
+			<attribute name="test" value="true"/>
+		</attributes>
+	</classpathentry>
+	<classpathentry kind="con" path="org.eclipse.jdt.junit.JUNIT_CONTAINER/5"/>
```
HeikoKlare: I propose to extract the tests into a separate test bundle. Then we don't need this dependency.

HeikoKlare (commented):
I've checked locally for what is still wrong with the project configurations. I have made some comments in this PR pointing to some of the issues, but the whole bunch of changes I tried is covered by this commit: https://github.com/HeikoKlare/eclipse.platform.swt/commit/b42bcb62f3d573a688f6a724dd3d771cad13fbe7

I propose to:
- Make `org.eclipse.swt.skia` a separate bundle instead of a fragment. We need to properly configure access to internals of `org.eclipse.swt` for that (I just silenced the errors in my commit), but that should not be a problem
- Clean up the build path of `org.eclipse.swt.skia`: there is a mixture of PDE dependencies, explicit libraries on the classpath and Maven dependencies; PDE dependencies are even exported; this can now all be reduced to simple PDE dependencies with the Skija libraries being available via target platform
- Move the Skia tests into a separate test bundle. Than the additional test-specific dependencies can also be removed from the `org.eclipse.swt.skia` bundle. Though this is not essentially necessary, it should make everything cleaner (but I would also be fine with leaving the tests in the production bundle as they are now).

What is not working for me is the DLL library loading mechanism of Skija when not explicitly including the win32 Skija fragment in the launch configuration. I would have expected it to be properly picked up via the p2.inf setting, but somehow it does not. However, manually adding it to the launch configuration makes the tests work for me:
Image

HeikoKlare:
> What is not working for me is the DLL library loading mechanism of Skija when not explicitly including the win32 Skija fragment in the launch configuration. 

I just found what's missing for this (and what is different with SWT).

**Why it works for SWT:** the fragments are automatically picked up in a launch because the SWT host declares the `Eclipse-ExtensibleAPI` header and the fragment provide actual Java classes/extensions. The Skija native fragments do not contain any Java code but just native libraries and the skija-shared library does not declare `Eclipse-ExtensibleAPI`, so they are no reasons for the fragment to be picked up in a launch automatically.

**Possible solution for Skija:** We want to express that when launching something with a dependency to skija-shared, a fitting native fragment must be present and loaded as well. The "fitting" aspect is already covered by the Platform-Filter. The "must be present" aspect is missing. This is something we can express with OSGi capabilities. I've locally added a `Require-Capability` to the skija-shared host bundle and a `Provide-Capability` to the fragment, and then the matching fragment is automatically picked for a launch.

If someone sees anything wrong in that approach, please let me know. Otherwise I can try to provide a PR to Orbit for adding that metadata to the wrapped bundles/fragments as soon as I find the time.

HeikoKlare:
See here for the idea of using capabilities to properly include the Skija fragments with native libraries:
- https://github.com/eclipse-orbit/orbit-simrel/pull/62

HeikoKlare:
Just tested with the latest Orbit build including Skija versions 0.143.15.v20260520-1000 with the OSGi capability and they now work fine: when starting a test the native fragment is automatically picked up and the test succeeds. Once we have updated the target definition with that version, it should be available out of the box. And all p2.inf files in this PR should be obsolete.

merks:
FYI, this PR is progressing:

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3831

vogella:
I suggest you squash the commits into one or two (if you want to have the test indepently)

DenisUngemach:
Can someone help me with the rest of these build errors. 

In https://github.com/eclipse-platform/eclipse.platform.swt/actions/runs/27824860967/job/82346595327?pr=3231#logs

```
Error:  Cannot resolve dependencies of project org.eclipse.platform:org.eclipse.swt.skia:eclipse-plugin:3.135.0-SNAPSHOT
Error:   with context {osgi.os=linux, org.eclipse.update.install.features=true, org.eclipse.swt.buildtime=true, osgi.arch=ppc64le, org.eclipse.update.install.sources=true, osgi.ws=gtk, org.eclipse.jdt.buildtime=true}
Error:    Software being installed: org.eclipse.swt.skia 3.135.0.qualifier
Error:    Missing requirement: io.github.humbleui.skija-shared 0.143.16.v20260524-1000 requires 'io.github.humbleui.skija; (library=native)' but it could not be found
Error:    Cannot satisfy dependency: org.eclipse.swt.skia 3.135.0.qualifier depends on: osgi.bundle; io.github.humbleui.skija-shared [0.143.16,1.0.0): See log for details
Error:  -> [Help 1]
Error:  
Error:  To see the full stack trace of the errors, re-run Maven with the -e switch.
Error:  Re-run Maven using the -X switch to enable full debug logging.
Error:  
Error:  For more information about the errors and possible solutions, please read the following articles:
Error:  [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/ResolverException
Error:  
Error:  After correcting the problems, you can resume the build with the command
Error:    mvn <args> -rf :org.eclipse.swt.skia
```

This seems to be a dependency problem. @merks Do you have an idea, what to do?


In https://github.com/eclipse-platform/eclipse.platform.swt/actions/runs/27824860831/job/82346617026?pr=3231#logs

```
Error:  Failed to execute goal org.eclipse.tycho:tycho-apitools-plugin:5.0.3:verify (verify) on project org.eclipse.swt.gtk.linux.x86_64: There are API errors:
Error:  The field org.eclipse.swt.graphics.GC.handle has been added to a class
Error:  META-INF/MANIFEST.MF:1 The major version should be incremented in version 3.135.0, since API breakage occurred since version 3.134.0
Error:  -> [Help 1]
```

I think this api tools error comes up because the GC is a sealed class where GCExtension extends it. It might be possible, that the api tools can't handle this concept.

merks:
I just want to remind you that only these native platforms are supported:

```
      <unit id="io.github.humbleui.skija-linux-arm64" version="0.143.16.v20260527-1000"/>
      <unit id="io.github.humbleui.skija-linux-x64" version="0.143.16.v20260527-1000"/>
      <unit id="io.github.humbleui.skija-macos-arm64" version="0.143.16.v20260527-1000"/>
      <unit id="io.github.humbleui.skija-macos-x64" version="0.143.16.v20260527-1000"/>
      <unit id="io.github.humbleui.skija-windows-arm64" version="0.143.16.v20260527-1000"/>
      <unit id="io.github.humbleui.skija-windows-x64" version="0.143.16.v20260527-1000"/>
```

(Note that an Orbit update is pending to these native libraries, but that's not related to any problems here.)

What is relevant is that when I see a failure like this where I see it resolving for `osgi.arch=ppc64le` I think that cannot work:

> 14:18:41  [INFO] {osgi.os=linux, org.eclipse.update.install.features=true, org.eclipse.swt.buildtime=true, osgi.arch=ppc64le, org.eclipse.update.install.sources=true, osgi.ws=gtk, org.eclipse.jdt.buildtime=true}

You cannot build all the platforms that SWT normally builds, but only the architectures supported by skija...

DenisUngemach:
> I just want to remind you that only these native platforms are supported:
> 
> ```
>       <unit id="io.github.humbleui.skija-linux-arm64" version="0.143.16.v20260527-1000"/>
>       <unit id="io.github.humbleui.skija-linux-x64" version="0.143.16.v20260527-1000"/>
>       <unit id="io.github.humbleui.skija-macos-arm64" version="0.143.16.v20260527-1000"/>
>       <unit id="io.github.humbleui.skija-macos-x64" version="0.143.16.v20260527-1000"/>
>       <unit id="io.github.humbleui.skija-windows-arm64" version="0.143.16.v20260527-1000"/>
>       <unit id="io.github.humbleui.skija-windows-x64" version="0.143.16.v20260527-1000"/>
> ```
> 
> (Note that an Orbit update is pending to these native libraries, but that's not related to any problems here.)
> 
> What is relevant is that when I see a failure like this where I see it resolving for `osgi.arch=ppc64le` I think that cannot work:
> 
> > 14:18:41  [INFO] {osgi.os=linux, org.eclipse.update.install.features=true, org.eclipse.swt.buildtime=true, osgi.arch=ppc64le, org.eclipse.update.install.sources=true, osgi.ws=gtk, org.eclipse.jdt.buildtime=true}
> 
> You cannot build all the platforms that SWT normally builds, but only the architectures supported by skija...

Yes, this is the issue. I try to split the build process of the Skia plugins in those platforms supported and those not supported.

DenisUngemach:
@HeikoKlare 
Only the API Tools errors and version upgrade errors remain. There i need some help.

vogella:
@DenisUngemach can you try this patch?

[skia-loongarch64-api-filter.patch](https://github.com/user-attachments/files/29639797/skia-loongarch64-api-filter.patch)

````
git apply /home/vogella/skia-loongarch64-api-filter.patch
````

DenisUngemach:
> @DenisUngemach can you try this patch?
> 
> [skia-loongarch64-api-filter.patch](https://github.com/user-attachments/files/29639797/skia-loongarch64-api-filter.patch)
> 
> ```
> git apply /home/vogella/skia-loongarch64-api-filter.patch
> ```

Causes new build errors. 

The problem is, that either i get errors, that filters are not used here:
https://github.com/eclipse-platform/eclipse.platform.swt/pull/3231/checks?check_run_id=85804623887

or if i remove these filter entries i get errors in the specific build. 

So something is wrong with the api tools.

vogella:
Hi @DenisUngemach, I tested your code and it looks like you the Skija engine allocates even more native handlers than before. I always assume that the skija work would also reduce the number of allocated native handlers (and therefore remove the restribution of how many widgets a client has). With even more handles I think we are going backwards.

DenisUngemach:
> Hi @DenisUngemach, I tested your code and it looks like you the Skija engine allocates even more native handlers than before. I always assume that the skija work would also reduce the number of allocated native handlers (and therefore remove the restribution of how many widgets a client has). With even more handles I think we are going backwards.

Hi @vogella, could you investigate where these native handle allocations are coming from?

Neither Skija nor Skia should require native handles, as Skia is only a 2D graphics library with its own resources. Therefore, my GC to Skija conversion must be allocating handles somewhere. My initial suspicion is that the necessary default font in SkiaGC (or SkiaResources) is being instantiated too frequently.

Currently, when customer code creates fonts, images, colors, etc., the native handles are created immediately. When these resources are used with a SkiaCanvas, my code extracts the necessary information from the native resource handle to create the corresponding Skija resource. However, this cannot be modified to reduce native handle creations in version 1.

To prevent handle creation, the Font, Image, and Color classes would need to be rewritten, with native resource handling moved to delegates (such as WinFont, WinImage, and WinColor on Windows)—similar to what has already been done for Region in the Windows fragment for the scaling feature. However, this introduces new challenges, particularly for third-party plugins that access native handles directly, but also for other features, where native information will be requested.

This kind of architectural refactoring would be part of a larger overhaul of the SWT source code.

And as it seems nobody has sufficient capacities to do that.

HeikoKlare:
Resource handling/sharing will definitely be an important topic when integrating Skia. In a first step, the goal should be to not introduce an overhead with respect to native handles. Reduction could be a follow-up step, as explained by Denis.

Reworking resource handling could be beneficial beyond the Skia adoption. When I just consider the win32 SWT implementation for image handling, it is quite convoluted. With the recently introduced lazy resource handling and multi-zoom support, it became even more complex. Having SWT as an abstraction layer and then mapping the resources to the format required by the consumer framework (be it win32 or Skia) would be great, because also for the pure win32 usages we could simplify the adoption of different image representation types to a core set to which SWT transforms every image resource it manages.
But as also said, it's a matter of capacities (and also regression risk, necessity to evaluate impacts like performance etc.).

Just a final note on third-party access to resources: removing OS-specific contents of resources (such as handles) should not be an issue, because that is not part of the API. Of course someone may have implemented (OS-specific) code that accesses native internals, but that's invalid anyway and would have tied that third-party extension to a specific OS.

PR #3237, opened 2026-04-14
[Win32] Simplify and clean up Pattern

HeikoKlare:
## Summary

This PR cleans up and simplifies `Pattern.java` (Win32 implementation) and adds a new cross-platform test class to lock in the constructor contracts documented in the API Javadoc.

---

### 1. [Win32] Simplify and clean up Pattern

Pure refactoring of the Win32-specific `Pattern` implementation. No behavioural changes are intended.

**Field naming**
- Renamed `isDestroyed` → `disposed`. Java convention uses an `is`-prefix for *methods*, not for the backing boolean field. Every other `Resource` subclass in SWT already uses `disposed`.

**Constructor validation**
- Moved null/disposed checks for gradient colors from the lazy `createHandle` method into the gradient constructors. The Javadoc for these constructors already documents `ERROR_NULL_ARGUMENT` when a color is `null`, so the error should be raised at construction time, not deferred to the first draw call. This aligns the Win32 implementation with the existing behaviour of the GTK (Cairo) and macOS (Cocoa) implementations, which have always validated eagerly in the constructor. Dispose checks are still performed in `createHandle` as well, since a color can become disposed after the `Pattern` is constructed.

**Map access**
- Replaced the two-lookup `containsKey`/`get` pattern in `getPatternHandle` with `computeIfAbsent`. The redundant `newPatternHandle` helper method is removed.

**Bit manipulation**
- Extracted a private static `colorRefToArgb` helper to eliminate the duplicated Win32 `COLORREF` (`0x00BBGGRR`) → GDI+ ARGB (`0xAARRGGBB`) byte-swap that was copy-pasted for `foreColor` and `backColor`.
- Removed `>> 0` no-op shifts in the `midColor` calculation.

**`PatternHandle.destroy()`**
- Replaced the classic `switch`/`case`/`break` with an arrow-form enhanced switch.

**Inner class constructors**
- Removed the `public` visibility modifier from private inner class constructors. Visibility modifiers on constructors of `private` inner classes are meaningless.

---

### 2. Add cross-platform Pattern constructor precondition tests

Adds `Test_org_eclipse_swt_graphics_Pattern` to the shared `org.eclipse.swt.tests` suite (i.e. not Win32-only) covering the constructor contracts specified in the `Pattern` Javadoc:

| Scenario | Expected |
|---|---|
| Image constructor, `null` image | `IllegalArgumentException` |
| Image constructor, disposed image | `IllegalArgumentException` |
| Gradient constructors (both overloads), `null` color1 | `IllegalArgumentException` |
| Gradient constructors (both overloads), `null` color2 | `IllegalArgumentException` |
| Gradient constructors (both overloads), disposed color1 | `IllegalArgumentException` |
| Gradient constructors (both overloads), disposed color2 | `IllegalArgumentException` |
| `isDisposed()` before and after `dispose()` | correct state |
| `dispose()` called twice | must not throw |
| `toString()` before and after `dispose()` | non-null, non-empty |

The GTK and macOS implementations have always validated null/disposed colors eagerly in the constructor. These tests lock in that shared contract across all three platforms and serve as a regression guard for the Win32 implementation, which was aligned with the other platforms in this PR.

---

## Test Plan

- [x] All 16 tests in `Test_org_eclipse_swt_graphics_Pattern` pass on Windows, Linux (GTK), and macOS
- [x] No regressions in the existing graphics test suite (`AllGraphicsTests`)
- [x] Visual smoke test: gradient patterns and image patterns render correctly (see existing `PatternWin32ManualTest`)

---

*This PR was created with the assistance of [Claude Code](https://claude.ai/code) (claude-sonnet-4-6).*

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Pattern.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Pattern.java
@@ -47,7 +47,7 @@ public class Pattern extends Resource {
 
 	private final Map<Integer, PatternHandle> zoomToHandle = new HashMap<>();
 
-	private boolean isDestroyed;
+	private boolean disposed;
 
 /**
  * Constructs a new Pattern given an image. Drawing with the resulting
@@ -167,6 +167,10 @@ public Pattern(Device device, float x1, float y1, float x2, float y2, Color colo
  */
 public Pattern(Device device, float x1, float y1, float x2, float y2, Color color1, int alpha1, Color color2, int alpha2) {
 	super(device);
+	if (color1 == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
+	if (color1.isDisposed()) SWT.error(SWT.ERROR_INVALID_ARGUMENT);
+	if (color2 == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
+	if (color2.isDisposed()) SWT.error(SWT.ERROR_INVALID_ARGUMENT);
 	this.baseX1 = x1;
 	this.baseX2 = x2;
 	this.baseY1 = y1;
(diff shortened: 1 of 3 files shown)


vogella:
LGTM

PR #3240, opened 2026-04-15
[GTK3] Fix ToolBar.computeSize 60 Hz repaint loop with SWT.WRAP

### LINKED ISSUE #3236: [GTK3] ToolBar.computeSize with SWT.WRAP causes self-sustaining 60 Hz repaint loop in CTabFolder.setTopRight
## Summary

On GTK3, `ToolBar.computeSize` unconditionally toggles `gtk_toolbar_set_show_arrow(handle, false)` before native measurement and, for `SWT.WRAP` toolbars, toggles it back to `true` afterwards. Each state change inside `gtk_toolbar_set_show_arrow` calls `gtk_widget_queue_resize → gtk_widget_queue_draw → gdk_window_invalidate_region` on the toolbar's `GdkWindow`.

When a `SWT.WRAP` `ToolBar` is placed inside a parent that gets measured on every layout pass — for example, the `Composite` passed to `CTabFolder.setTopRight(...)` — the invalidation reschedules the next paint at vsync, which triggers another layout pass, which calls `ToolBar.computeSize` again, which fires another invalidation. A self-sustaining **~60 Hz repaint loop** forms that consumes ~10–25 % of one CPU core **while the application is completely idle**. The loop runs whenever the window is mapped and only stops when the window is unmapped (minimised) or the offending parent is disposed.

This is a latent bug that has been dormant in SWT since the GTK3 implementation of `SWT.WRAP` was added in bug 46025 (SWT 3.8 M7, 2012). It is reproducible in SWT 3.130.0 on stock Ubuntu 25.10 with stock GTK 3.24.18.

## Environment

| | |
|---|---|
| SWT version | `org.eclipse.swt.gtk.linux.x86_64` 3.130.0 |
| Java | OpenJDK 23.0.2 (Temurin / JustJ) |
| OS | Ubuntu 25.10 (kernel 6.17) |
| GTK | 3.24.18 (`libgdk-3.so.0.2418.32`) |
| Desktop | GNOME (also reproducible on Wayland) |

## Minimal reproducer

See `ToolBarWrapRepaintLoop.java` attached. Pure SWT, no dependencies, ~55 lines. Toggle `SWT.WRAP` in the `TOOLBAR_STYLE` constant to reproduce vs fix.

```java
final int TOOLBAR_STYLE = SWT.FLAT | SWT.RIGHT | SWT.WRAP;
// ... creates a Shell with a CTabFolder whose topRight holds a ToolBar(TOOLBAR_STYLE).
```

**With `SWT.WRAP` set** — `top -H -p <pid>` shows the SWT `main` thread at ~10–25 % CPU continuously, with zero user input and no visible change on screen. Remains high as long as the window is mapped. Drops to 0 % if the window is minimised.

**Without `SWT.WRAP`** — `main` sits at 0 % CPU on idle, as expected.

## Root cause

`ToolBar.computeSizeInPixels` in SWT 3.130.0 (`bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/ToolBar.java` around lines 180–196):

```java
if (GTK.GTK4) {
    size = computeNativeSize (handle, wHint, hHint, changed);
} else {
    /*
     * Feature in GTK. Size of toolbar is calculated incorrectly
     * and appears as just the overflow arrow, if the arrow is enabled
     * to display. The fix is to disable it before the computation of
     * size and enable it if WRAP style is set.
     */
    GTK3.gtk_toolbar_set_show_arrow (handle, false);
    size = computeNativeSize (handle, wHint, hHint, changed);
    if ((style & SWT.WRAP) != 0) GTK3.gtk_toolbar_set_show_arrow (handle, true);
}
```

The `set_show_arrow(false)` call is the workaround for the GTK sizing bug the comment describes. For non-WRAP toolbars, `show_arrow` is always `false`, so the first `set_show_arrow(false)` call is a no-op inside GTK and nothing is invalidated.

For **WRAP toolbars**, SWT re-enables `show_arrow` after measurement by calling `set_show_arrow(true)`. On the *next* measurement, the state cycle becomes `true → set(false) → false → measure → set(true) → true`. Each of those two state transitions fires this chain inside GTK:

```
gtk_toolbar_set_show_arrow
  → gtk_widget_queue_resize
    → gtk_widget_queue_draw
      → gdk_window_invalidate_region
```

which marks the toolbar's `GdkWindow` dirty and schedules a paint for the next frame.

The feedback loop forms whenever the WRAP `ToolBar` sits in a parent that `CTabFolder` or another SWT container measures on every layout pass. `CTabFolder.setTopRight(Composite)` is the canonical case — `CTabFolder.onResize` / layout pass → `topRight.computeSize` → `GridLayout.layout` → `ToolBar.computeSize` → toxic toggle → invalidation → next frame → back to `CTabFolder.onResize`. The cycle repeats at vsync rate.

## Evidence — `bpftrace` uprobe trace

Attached uprobes on `gdk_window_invalidate_rect` / `gdk_window_invalidate_region` in a running NetXMS management console (one of many SWT applications that uses `CTabFolder` with a `SWT.WRAP` top-right toolbar). 10-second capture:

```
$ sudo bpftrace -p $PID -e '
uprobe:/usr/lib/x86_64-linux-gnu/libgdk-3.so.0:gdk_window_invalidate_rect   { @rect[ustack(16)] = count(); }
uprobe:/usr/lib/x86_64-linux-gnu/libgdk-3.so.0:gdk_window_invalidate_region { @region[ustack(16)] = count(); }
interval:s:10 { exit(); }'

@region[
    gdk_window_invalidate_region+0
    gtk_widget_queue_draw+150
    gtk_widget_queue_resize+120
    gtk_toolbar_set_show_arrow+110
    Java_org_eclipse_swt_internal_gtk3_GTK3_gtk_1toolbar_1set_1show_1arrow+15
    0x7cfd1112c3dc
    0x70111e838
]: 586
@region[
    gdk_window_invalidate_region+0
    gtk_widget_queue_draw+150
    gtk_widget_queue_resize+120
    gtk_toolbar_set_show_arrow+110
    Java_org_eclipse_swt_internal_gtk3_GTK3_gtk_1toolbar_1set_1show_1arrow+15
    0x7cfd1112c3dc
    0xa00111e838
]: 586
@region[
    gdk_window_invalidate_region+0
    gtk_widget_queue_draw+150
    gtk_widget_queue_resize+120
    gtk_toolbar_set_show_arrow+110
    Java_org_eclipse_swt_internal_gtk3_GTK3_gtk_1toolbar_1set_1show_1arrow+15
    0x7cfd1112c2ec
]: 1172
```

**2344 invalidations in 10 s = ~234 per second**, all from `Java_org_eclipse_swt_internal_gtk3_GTK3_gtk_1toolbar_1set_1show_1arrow`. No other callers appear in the profile.

## Evidence — instrumented `CTabFolderRenderer`

Substituting a `CTabFolderRenderer` subclass that counts `draw(PART_HEADER)` calls and logs a Java stack trace every N calls produced this sample log line (NetXMS console, 9-tab perspective):

```
WARN  CTabFolderRenderer - PART_HEADER draws: 119 in 2000ms (59.5/s)
  folder=2054757222 itemCount=9 size=Point {1260, 1278} parent=ViewFolder
  latest stack:
    at org.eclipse.swt.custom.CTabFolder.onPaint(CTabFolder.java:2091)
    at org.eclipse.swt.custom.CTabFolder.lambda$0(CTabFolder.java:338)
    at org.eclipse.swt.widgets.Display.sendEvent(Display.java:5862)
    at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1656)
    at org.eclipse.swt.widgets.Control.gtk_draw(Control.java:3891)
    at org.eclipse.swt.widgets.Composite.gtk_draw(Composite.java:506)
    at org.eclipse.swt.widgets.Widget.windowProc(Widget.java:2614)
    at org.eclipse.swt.widgets.Control.windowProc(Control.java:6857)
    at org.eclipse.swt.widgets.Display.windowProc(Display.java:6169)
    at org.eclipse.swt.internal.gtk3.GTK3.gtk_main_do_event(Native Method)
    at org.eclipse.swt.widgets.Display.eventProc(Display.java:1605)
    at org.eclipse.swt.internal.gtk3.GTK3.gtk_main_iteration_do(Native Method)
    at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:4519)
```

**59.5 paints per second** — vsync. No Java application code appears above `draw()`; every paint enters through `gtk_main_do_event` → `Display.readAndDispatch`. The application is not calling `redraw()`, `layout()`, or anything else — the entire loop runs below the Java layer and re-enters SWT only for rendering.

## Proposed fixes

In approximate order of invasiveness — any one of these would break the loop:

**1. Cache the computed size and avoid redundant toggles.** The simplest fix. `ToolBar.computeSizeInPixels` can cache the most recent `(wHint, hHint, state)` → `size` result and return it without calling native measurement if neither the items nor the hints have changed. A `changed == true` argument would invalidate the cache. This also benefits Win32 / Cocoa code paths where native measurement has a non-trivial cost.

**2. Guard the `set_show_arrow(true)` re-enable with a state check.** Track SWT's "desired `show_arrow` state" in a Java field; only call `gtk_toolbar_set_show_arrow(handle, true)` if SWT's field says the widget should currently have `show_arrow = true` *and* the previous call flipped it off. Avoid unconditionally re-enabling it after every measurement. This is the smallest diff but doesn't address the first `set_show_arrow(false)` call when the previous state was already `false` (which is the no-op case in GTK and costs nothing anyway).

**3. Replace the toggle workaround with a direct `gtk_widget_get_preferred_width/_height` measurement.** The comment in `ToolBar.computeSizeInPixels` explains that `set_show_arrow(false)` works around a GTK sizing bug. GTK3's preferred-size API has been stable since ~3.10 and the original workaround may no longer be necessary. If `gtk_widget_get_preferred_size` returns the correct dimensions for a WRAP toolbar without the `show_arrow` dance, the entire toggle path can be removed on GTK3.

**4. Document the gotcha.** If none of the above are accepted, at minimum `CTabFolder.setTopRight` and `ToolBar` should document that placing a `SWT.WRAP` toolbar in a frequently-measured parent on GTK3 will cause a repaint loop. Every SWT-on-GTK3 application is currently affected without knowing it.

## Downstream workaround

Applications that cannot wait for an SWT release can work around the bug by **not using `SWT.WRAP`** on any `ToolBar` placed inside `CTabFolder.setTopRight(...)` or any similarly-measured parent. The cost is the loss of the overflow chevron; items that don't fit are clipped instead of hidden behind a dropdown. In practice most SWT-based UIs have ≤ 5 items in their top-right toolbar and fit trivially, so the visible regression is nil.

## Impact estimate

Any SWT-on-GTK3 application that uses `CTabFolder.setTopRight(Composite)` with a `SWT.WRAP`-styled `ToolBar` child — which is the idiomatic pattern in Eclipse IDE, NetXMS nxmc, and many JFace-based applications — is silently burning 10–25 % of one CPU core per open tab folder. The cost is not visible in Java Flight Recorder / YourKit / JProfiler because the loop originates in GTK and re-enters Java only through the SWT event loop; standard Java profiling reports "the main thread is in `Display.sleep()` / `gtk_main_iteration_do`" and does not flag the per-frame measurement cost.

Users experience this as battery drain on laptops, fans running on workstations, and the SWT main thread being less responsive under load. The root cause has been dormant since 2012 and is unlikely to be discovered without tracing at the GDK level.

[ToolBarWrapRepaintLoop.java](https://github.com/user-attachments/files/26721063/ToolBarWrapRepaintLoop.java)

vogella:
## Summary
- Remove the `gtk_toolbar_set_show_arrow` toggle workaround in `ToolBar.computeSizeInPixels` that caused two GTK invalidation cycles per measurement call
- This workaround was added in 2012 (Bug 46025) for GTK 3.4 and is no longer needed on modern GTK 3.24
- When a `SWT.WRAP` ToolBar sits inside a frequently-measured parent (e.g. `CTabFolder.setTopRight`), the toggle created a self-sustaining ~60 Hz repaint loop burning 10-25% CPU while idle
- Add Snippet394 to verify ToolBar sizing works correctly without the workaround

## Test plan
- [ ] Run Snippet394 on GTK3 to verify both WRAP and non-WRAP toolbars report similar widths
- [ ] Run the reproducer from #3236 with `SWT.WRAP` and confirm CPU stays at 0% on idle
- [ ] Verify `SWT.WRAP` toolbar overflow behavior still works correctly (items hidden behind chevron when toolbar is too narrow)
- [ ] Test in Eclipse IDE with CTabFolder top-right toolbars

Fixes #3236

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/ToolBar.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/ToolBar.java
@@ -177,23 +177,7 @@ Point computeSizeInPixels (int wHint, int hHint, boolean changed) {
 	if (wHint != SWT.DEFAULT && wHint < 0) wHint = 0;
 	if (hHint != SWT.DEFAULT && hHint < 0) hHint = 0;
 
-	Point size = null;
-
-	if (GTK.GTK4) {
-		size = computeNativeSize (handle, wHint, hHint, changed);
-	} else {
-		/*
-		 * Feature in GTK. Size of toolbar is calculated incorrectly
-		 * and appears as just the overflow arrow, if the arrow is enabled
-		 * to display. The fix is to disable it before the computation of
-		 * size and enable it if WRAP style is set.
-		 */
-		GTK3.gtk_toolbar_set_show_arrow (handle, false);
-		size = computeNativeSize (handle, wHint, hHint, changed);
-		if ((style & SWT.WRAP) != 0) GTK3.gtk_toolbar_set_show_arrow (handle, true);
-	}
-
-	return size;
+	return computeNativeSize (handle, wHint, hHint, changed);
 }
 
 @Override
@@ -609,18 +593,8 @@ void reskinChildren (int flags) {
 
 @Override
 int setBounds (int x, int y, int width, int height, boolean move, boolean resize) {
-	if (!GTK.GTK4) GTK3.gtk_toolbar_set_show_arrow (handle, false);
 	int result = super.setBounds (x, y, width, height, move, resize);
 	if ((result & RESIZED) != 0) relayout ();
-	if ((style & SWT.WRAP) != 0) {
-		if (GTK.GTK4) {
-			/* TODO: GTK4 will require us to implement our own
-			 * overflow menu. May require the use of the "toolbar" style class
-			 * applied to the widget.  */
-		} else {
-			GTK3.gtk_toolbar_set_show_arrow (handle, true);
-		}
-	}
 
 	return result;
 }
--- a/bundles/org.eclipse.swt/Eclipse SWT PI/gtk/library/gtk3.c
+++ b/bundles/org.eclipse.swt/Eclipse SWT PI/gtk/library/gtk3.c
@@ -1987,16 +1987,6 @@ JNIEXPORT void JNICALL GTK3_NATIVE(gtk_1toolbar_1set_1icon_1size)
 }
 #endif
 
-#ifndef NO_gtk_1toolbar_1set_1show_1arrow
-JNIEXPORT void JNICALL GTK3_NATIVE(gtk_1toolbar_1set_1show_1arrow)
-	(JNIEnv *env, jclass that, jlong arg0, jboolean arg1)
-{
-	GTK3_NATIVE_ENTER(env, that, gtk_1toolbar_1set_1show_1arrow_FUNC);
-	gtk_toolbar_set_show_arrow((GtkToolbar *)arg0, (gboolean)arg1);
-	GTK3_NATIVE_EXIT(env, that, gtk_1toolbar_1set_1show_1arrow_FUNC);
-}
-#endif
-
 #ifndef NO_gtk_1toolbar_1set_1style
 JNIEXPORT void JNICALL GTK3_NATIVE(gtk_1toolbar_1set_1style)
 	(JNIEnv *env, jclass that, jlong arg0, jint arg1)
--- a/bundles/org.eclipse.swt/Eclipse SWT PI/gtk/library/gtk3_stats.h
+++ b/bundles/org.eclipse.swt/Eclipse SWT PI/gtk/library/gtk3_stats.h
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2025 IBM Corporation and others.
+ * Copyright (c) 2000, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -184,7 +184,6 @@ typedef enum {
 	gtk_1toolbar_1insert_FUNC,
 	gtk_1toolbar_1new_FUNC,
 	gtk_1toolbar_1set_1icon_1size_FUNC,
-	gtk_1toolbar_1set_1show_1arrow_FUNC,
 	gtk_1toolbar_1set_1style_FUNC,
 	gtk_1tree_1view_1column_1cell_1get_1size_FUNC,
 	gtk_1tree_1view_1get_1bin_1window_FUNC,
--- a/bundles/org.eclipse.swt/Eclipse SWT PI/gtk/org/eclipse/swt/internal/gtk3/GTK3.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT PI/gtk/org/eclipse/swt/internal/gtk3/GTK3.java
@@ -704,11 +704,6 @@ public class GTK3 {
 	public static final native void gtk_toolbar_insert(long toolbar, long item, int pos);
 	/**
 	 * @param toolbar cast=(GtkToolbar *)
-	 * @param show_arrow cast=(gboolean)
-	 */
-	public static final native void gtk_toolbar_set_show_arrow(long toolbar, boolean show_arrow);
-	/**
-	 * @param toolbar cast=(GtkToolbar *)
 	 * @param style cast=(GtkToolbarStyle)
 	 */
 	public static final native void gtk_toolbar_set_style(long toolbar, int style);


akurtakov:
Very good catch. The snippet is good for testing but it is not smth that any SWT user can use as an example so it shouldn't end up in snippets whenever the PR is merged.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/ToolBar.java
akurtakov: Unconditionally setting to not show arrow whenever setBounds is called and show arrow only for WRAP which Gtk itself doesn't support at all is very wrong. So can be remove altogether.

akurtakov: This can be remove too as WRAP is not supported on Gtk at all and showing or not arrow depending on this flag is totally not what the flag is for.

akurtakov:
With the requested changes in Toolbar GTK3.gtk_toolbar_set_show_arrow will no longer be needed so it can be removed too.

vogella:
````
 * Copyright (c) 2026 Eclipse Foundation and others.
 *
 *
 *
 * Contributors:
 *     Eclipse Foundation - initial API and implementation

import org.eclipse.swt.*;

/**
 * ToolBar computeSize test: verify whether the GTK3 toolbar sizing bug
 * (toolbar reports only the overflow arrow width when show_arrow is enabled)
 * is still present.
 *
 * <p>This snippet creates a ToolBar with several items and compares the
 * computed size with and without the SWT.WRAP style. If the GTK bug is fixed,
 * both sizes should be similar. If the bug is still present, the WRAP toolbar
 * will report a much smaller width (just the arrow).
 *
 * <p>See https://github.com/eclipse-platform/eclipse.platform.swt/issues/3236
 *
 * <p>For a list of all SWT example snippets see
 * https://eclipse.dev/eclipse/swt/snippets/
 */
public class Snippet395 {

	shell.setText("Snippet 395 - ToolBar computeSize with WRAP");

	// Create a toolbar WITHOUT SWT.WRAP
	ToolBar toolBarNoWrap = new ToolBar(shell, SWT.FLAT | SWT.RIGHT);
	addItems(toolBarNoWrap);

	// Create a toolbar WITH SWT.WRAP
	ToolBar toolBarWrap = new ToolBar(shell, SWT.FLAT | SWT.RIGHT | SWT.WRAP);
	addItems(toolBarWrap);

	// Compute sizes
	Point sizeNoWrap = toolBarNoWrap.computeSize(SWT.DEFAULT, SWT.DEFAULT);
	Point sizeWrap = toolBarWrap.computeSize(SWT.DEFAULT, SWT.DEFAULT);

	// Display results
	Label resultLabel = new Label(shell, SWT.WRAP);
	resultLabel.setLayoutData(new GridData(SWT.FILL, SWT.CENTER, true, false));
	StringBuilder sb = new StringBuilder();
	sb.append("ToolBar without WRAP: " + sizeNoWrap.x + " x " + sizeNoWrap.y + "\n");
	sb.append("ToolBar with    WRAP: " + sizeWrap.x + " x " + sizeWrap.y + "\n\n");

	int widthDiff = Math.abs(sizeNoWrap.x - sizeWrap.x);
	if (widthDiff > sizeNoWrap.x / 2) {
		sb.append("RESULT: Large width difference (" + widthDiff + "px). "
			+ "The GTK sizing bug may still be present.\n"
			+ "The WRAP toolbar is likely reporting only the arrow width.");
	} else {
		sb.append("RESULT: Widths are similar (diff=" + widthDiff + "px). "
			+ "The GTK sizing bug appears to be fixed.");
	}
	resultLabel.setText(sb.toString());

	// Also print to console
	System.out.println(sb);

	shell.pack();
	shell.open();
	while (!shell.isDisposed()) {
	}
	display.dispose();
}

private static void addItems(ToolBar toolBar) {
	for (int i = 0; i < 5; i++) {
		ToolItem item = new ToolItem(toolBar, SWT.PUSH);
		item.setText("Action " + (i + 1));
	}
}

}


````

akurtakov:
Changes in GTK3.java should have caused changes in gtk3.c and gtk3_stats.h

vogella:
> Changes in GTK3.java should have caused changes in gtk3.c and gtk3_stats.h

I do not see how. The gtk-dev-guide.md mentions that "SWT Tools also generates JNI bindings (i.e. os.c, os_stat.c, etc.) for you automatically." but my SDK build does not have the SWT Tools installled. And mvn clean install '-Dnative=gtk.linux.x86_64' -DskipTests does also not do this. 

Manually updated gtk3.c. The stats files (gtk3_stats.c/gtk3_stats.h) didn't seemt to have an entry for this function.

akurtakov:
SWT tools work only as eclipse plugin and do their work on save of java file as regular builder. The entry in gtk3_stats is gtk_1toolbar_1set_1show_1arrow_FUNC but I would prefer to see the changes done by the builder manually to reduce extra changes in other PRs where swt tools do the work.

vogella:
> SWT tools work only as eclipse plugin and do their work on save of java file as regular builder. The entry in gtk3_stats is gtk_1toolbar_1set_1show_1arrow_FUNC but I would prefer to see the changes done by the builder manually to reduce extra changes in other PRs where swt tools do the work.

As discussed via chat the tools usage is not document nor part of the SDK build. I will suggest to @HannesWell to document the usage in his pending https://github.com/eclipse-platform/eclipse.platform.swt/pull/3212 for documention improvements

image

vogella:
Coooool. After installing the swt.tools the c files get updated. Very nice feature. Would be nice if that would also be baked into the command line build.

@akurtakov PR updated as requested.

akurtakov:
I'm updating jnigen in https://github.com/eclipse-platform/www.eclipse.org-eclipse/pull/535 - outdated version available at https://eclipse.dev/eclipse/swt/jnigen.html

vogella:
Any further concerns about this PR @akurtakov ?

PR #3245, opened 2026-04-16
[Gtk4] Fix headerbar adjustment on shell resize

akurtakov:
On shell resize headerbar calculations were not accounting for headerbar height thus leading to cutoff dialogs after resize.


akurtakov:
Before:
image
After:
image

PR #3246, opened 2026-04-16
[macOS] Consider transformation scale in GC.drawImage(image, x, y) #2919

HeikoKlare:
## Summary

This PR ports the transformation-scale-aware image drawing logic to Cocoa/macOS, following the same approach previously applied to Windows (e97143c) and currently being introduced for GTK in #3201.

### Background

When a scaled `Transform` is set on a `GC` and `drawImage(image, x, y)` is called, SWT should select the best-fitting image representation (e.g. the high-resolution variant of an SVG-backed image) for the effective on-screen size — accounting for both the destination rectangle *and* the active transformation scale. Without this fix, the code ignores the transformation scale and passes only the nominal destination size to the image-selection logic, which causes a lower-resolution representation to be rendered and then upscaled, producing blurry results especially noticeable with SVG images.

### Changes

**`bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/GC.java`**

- Added `calculateTransformationScale()` — extracts the effective scale factor from the active `NSAffineTransform` using `Math.hypot()` on the matrix columns (`m11`/`m21` and `m12`/`m22`), which yields the correct scale even when the transform combines scaling with rotation.
- Modified `drawImage(Image, int, int)` — when a non-null transform is active, delegates to `drawImage(image, x, y, w, h)` using the image's natural bounds, routing through the scale-aware path instead of the legacy simple path.
- Modified `drawImage(Image, int, int, int, int)` — multiplies the destination dimensions by the transformation scale before passing them to `executeOnImageAtSizeBestFittingSize()`, ensuring the most appropriate image representation is loaded at the effective rendered resolution.

**`tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_GC.java`**

- Removed the `@EnabledOnOs(OS.WINDOWS)` guard (and now-unused `EnabledOnOs`/`OS` imports) from `test_drawImageLorg_eclipse_swt_graphics_ImageIIII_withTransform` — the test now runs on all platforms.
- Added the missing `@Test` annotation to `test_drawImageLorg_eclipse_swt_graphics_ImageIIII_withTransform_zeroTargetSize`, which was previously never executed.

Both tests pass on macOS.

## Relation to other PRs

> **⚠️ This PR is inspired by, built on top of, and must be merged *after* [#3201](https://github.com/eclipse-platform/eclipse.platform.swt/pull/3201).**
>
> PR #3201 introduces the equivalent fix for GTK and changes the test annotation from `@EnabledOnOs(OS.WINDOWS)` to `@DisabledOnOs(OS.MAC)`. This PR removes the remaining OS restriction entirely, completing the cross-platform rollout. Merging this PR before #3201 would result in a trivially conflicting annotation change on the test method.

## Test plan

- [x] `test_drawImageLorg_eclipse_swt_graphics_ImageIIII_withTransform` passes on macOS
- [x] `test_drawImageLorg_eclipse_swt_graphics_ImageIIII_withTransform_zeroTargetSize` passes on macOS
- [x] Verify visually that SVG images drawn via `drawImage(image, x, y)` under a scaled `Transform` render at full sharpness on macOS

---

*This change was created with the assistance of [Claude Code](https://claude.ai/code) (Anthropic).*

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/GC.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/graphics/GC.java
@@ -281,6 +281,19 @@ static int checkStyle (int style) {
 	return style & (SWT.LEFT_TO_RIGHT | SWT.RIGHT_TO_LEFT);
 }
 
+private float calculateTransformationScale() {
+	if (data.transform == null) {
+		return 1.0f;
+	}
+	// this calculates the effective length in x and y
+	// direction without being affected by the rotation
+	// of the transformation
+	NSAffineTransformStruct struct = data.transform.transformStruct();
+	float scaleWidth = (float) Math.hypot(struct.m11, struct.m21);
+	float scaleHeight = (float) Math.hypot(struct.m12, struct.m22);
+	return Math.max(scaleWidth, scaleHeight);
+}
+
 /**
  * Invokes platform specific functionality to allocate a new graphics context.
  * <p>
@@ -1143,7 +1156,12 @@ public void drawImage(Image image, int x, int y) {
 	if (handle == null) SWT.error(SWT.ERROR_GRAPHIC_DISPOSED);
 	if (image == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
 	if (image.isDisposed()) SWT.error(SWT.ERROR_INVALID_ARGUMENT);
-	drawImage(image, 0, 0, -1, -1, x, y, -1, -1, true);
+	if (data.transform != null) {
+		Rectangle imageBounds = image.getBounds();
+		drawImage(image, x, y, imageBounds.width, imageBounds.height);
+	} else {
+		drawImage(image, 0, 0, -1, -1, x, y, -1, -1, true);
+	}
 }
 
 /**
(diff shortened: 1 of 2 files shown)


Phillipus:
From https://github.com/eclipse-gef/gef-classic/issues/922#issuecomment-4315454293:

> If someone of you wants to do a sanity check of the macOS change (even without being able to test), would be nice to have that feedback in the mentioned PR.

Can you tell me how to test this?

HeikoKlare:
This is a slightly adapted snippet from https://github.com/eclipse-platform/eclipse.platform.swt/issues/2919
```java
		shell.setText("Snippet 389");
		ImageFileNameProvider fileNameProvider = zoom -> {
			return "src/org/eclipse/swt/snippets/eclipse.svg";
		};
		Image image = new Image(display, fileNameProvider);
		shell.addPaintListener(e -> {
			e.gc.setTransform(new Transform(display, 4, 0, 0, 4, 0, 0));
			e.gc.drawImage(image, 0, 0, 16, 16);
			e.gc.drawImage(image, 0, 0, 16, 16, 0, 25, 16, 16);
			e.gc.setTransform(null);
		});

		shell.open ();

		}
		display.dispose ();
	}
```

It used to look like this:
image

And with this change it looks like this:
image

I would expect that to also be visible when having images in GEF diagrams that use `SWTGraphics` and are zoomed in, but I have no proper example for that at hand.

Phillipus:
> I would expect that to also be visible when having images in GEF diagrams that use SWTGraphics and are zoomed in

I tested this with our RCP app and added an SVG image to a Draw2d figure and zoomed in to 800% (GEF diagram zoom).

Before this PR:

before

After this PR:

after

Nice.

HeikoKlare:
Great, thank you for testing this, also with GEF!
Then let's merge this to have the proper scaling behavior consistently available across all three SWT implementations with the next release.

Phillipus:
>  Then let's merge this to have the proper scaling behavior consistently available across all three SWT implementations with the next release.

Yes please! Thanks for your work on this. :-)

PR #3253, opened 2026-04-18
Add local Claude settings file to gitignore

HeikoKlare:
See https://code.claude.com/docs/en/settings#available-scopes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ pom.tycho
 /xdg-runtime-*
 /binaries/org.eclipse.swt.*/src/
 tmpdir/
+.claude/settings.local.json


vogella:
Please explain why this is helpful in your commit message.

We currently have @Agent.md in Claude.md and Agent.md describes the repo structure. I think @HannesWell has a pending request to improve it. 

What is missing in these files?

HeikoKlare:
I locally use some repository-specific permissions, which are stored in that file. Do you have any concerns ignoring that file? Maybe that someone puts information in such a local file to deal with missing information for agents instead of contributing them upstream to agents.ds/claude.md? But then such person can still simply ignore that the file is there and just not commit it.

vogella:
Sounds fine for me, just add your reasoning to the commit message.

vogella (approved):

PR #3270, opened 2026-04-24
Proper Maven command to run tests

### LINKED ISSUE #3261: mvn test does nothing for most tests.
https://github.com/eclipse-platform/eclipse.platform.swt/blob/f03d65dfb1ae7daad9596ff77d160af55715db78/AGENTS.md?plain=1#L103

Instruction seems invalid - `test` should probably be replaced with `verify`

Without -pl argument, Maven fails on `eclipse.platform.swt.binaries` project, so the command line should probably include `-pl :org.eclipse.swt.tests` too.

basilevs:
`mvn test` command does not work in Tycho Surefire context.

Added a proper command.

Do not fail Linux and MacOS testing because of absence of platform-specific tests.

Fixes #3261.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -99,8 +99,8 @@ If needed, add packages to `Import-Package` or `Require-Bundle`.
 # Run all tests
 mvn clean verify
 
-# Run specific test class
-mvn test -Dtest=ClassName
+# Run specific test class (append `#testMethodName` for a single test)
+mvn verify -pl :THE_BUNDLE_WITH_THE_ACTUAL_TEST -am -DskipNativeTests=false -Dsurefire.failIfNoSpecifiedTests=false -Dtest=ClassName
 ```
 
 ### Test Location


basilevs:
@HeikoKlare please review

AGENTS.md
```diff
@@ -99,8 +99,11 @@ If needed, add packages to `Import-Package` or `Require-Bundle`.
 # Run all tests
 mvn clean verify
 
-# Run specific test class
-mvn test -Dtest=ClassName
+# Run specific test class (requires prior `mvn install` to build dependencies)
+mvn verify -pl :THE_BUNDLE_WITH_THE_ACTUAL_TEST -Dtest=ClassName
```
HeikoKlare: I think we could just drop this one and always propose to use the `-am` build for agents. Then we can also reduce the explanation for the actual command to "Run specific test class". The shorter the information added to the context, the better.

AGENTS.md
```diff
@@ -99,8 +99,11 @@ If needed, add packages to `Import-Package` or `Require-Bundle`.
...
+
+# Run specific test class without prior `mvn install` (uses -am to build dependencies inline).
+mvn verify -pl :THE_BUNDLE_WITH_THE_ACTUAL_TEST -am -Dtest=ClassName
 ```
```
HeikoKlare, line 104: We might add a sentence about OS-specific tests here. In case the test is OS-specific (fully qualified name contains "win32", "cocoa" or "gtk"), native tests must be enabled via `-DskipNativeTests=false`

binaries/pom.xml
```diff
@@ -29,6 +29,7 @@
 		<buildid>${buildId}</buildid>
 		<maven.compiler.release>21</maven.compiler.release>
 		<swtMainProject>${project.basedir}/../../bundles/org.eclipse.swt</swtMainProject>
+		<surefire.failIfNoSpecifiedTests>false</surefire.failIfNoSpecifiedTests>
```
HeikoKlare: Why is this change required?

HeikoKlare (commented):
Thank you for the proposal. In general, the change looks fine to me. I've tested if an agent (Copilot CLI with Sonnet 4.6) uses a proper command after applying the proposed change. I used the following prompt with and without this change:

> The tests in @tests\org.eclipse.swt.tests.win32\JUnit Tests\org\eclipse\swt\tests\win32\Test_org_eclipse_swt_events_KeyEvent.java are failing when I execute them in the IDE. Can you please confirm that the tests are failing by executing it via Maven?

Without this change, it uses `mvn test -Dtest=Test_org_eclipse_swt_events_KeyEvent -pl :org.eclipse.swt.tests.win32` which does not work because of the wrong Maven goal and the missing `-am` for building dependencies. With this change, it uses `mvn verify -Dtest=Test_org_eclipse_swt_events_KeyEvent -pl :org.eclipse.swt.tests.win32 - am` and thus works fine (except for the missing `-DskipNativeTests=false`, see my additional comment).

binaries/pom.xml
basilevs: It prevents `mvn verify -am` from failing on MacOS and Linux which have no platform specific tests. There are tests for Windows, so we do not want to remove surefire from platform-specific fragments. This will allow to both run tests on Windows and not to fail on other platforms.

I guess this is why you use `-DskipNativeTests`. My approach seems cleaner, as tests are only skipped if they are absent for that platform instead of either skipping them altogether or breaking some platforms.

AGENTS.md
basilevs: Why is this property needed? Why can't native tests always run?

AGENTS.md
basilevs: Removed

AGENTS.md
HeikoKlare: I have to admit that I don't know why it is currently configured like that. Maybe it's related to the Jenkins build, as that one does not execute any of the native tests at all. Without setting this properly, tests inside the OS-specific test bundles will not be executed. E.g., the command I mentioned in the review comment `mvn verify -Dtest=Test_org_eclipse_swt_events_KeyEvent -pl :org.eclipse.swt.tests.win32 - am` will not execute any test without `-DskipNativeTests=false`.

binaries/pom.xml
HeikoKlare: I see. It is because the native fragments use surefire and when specifying a concrete test inside a different test bundle for execution, there will be no matching tests in the fragments to be executed, so the build fails. But this is independent from `-DskipNativeTests` (which is relevant for the OS-specific test bundles and not the unit tests inside the fragments) and not specific to any platform (as there are tests inside the native fragments executed on every platform).
In my opinion, setting `failIfNoSpecifiedTests` to false for the fragments is not correct, as there actually are tests and we want to be notified if we break anything in the configuration that results in non-execution of those tests. I don't have a perfect solution for this. Maybe just adding `-Dsurefire.failIfNoSpecifiedTests=false` to the selective test execution would be the simplest solution?

basilevs: I've added both parameters unconditionally in agent instructions and removed them from POM.
I think none of them do any harm so explaining them to agent is unnecessary.

HeikoKlare: Sounds resonable

HeikoKlare (approved):
Thank you for this enhancement, @basilevs!
It significantly improves the automated feedback cycles for agents. And we have learned something about the configurations and parameterization of SWT tests 🙂

Issue #3271, opened 2026-04-25
Github-hosted agents struggle due to missing LFS


basilevs:
AI agents are confused by LFS test links and have to investigate them in every session:

> Ah, the .so files are ASCII text, not binary ELF files! They're likely Git LFS pointers or some kind of placeholder. That's why they fail with "invalid ELF header".I need to build the native libraries from source. Let me try the native build:

Would it be better to [enable LFS support](https://docs.github.com/en/copilot/how-tos/copilot-on-github/customize-copilot/customize-cloud-agent/customize-the-agent-environment#enabling-git-large-file-storage-lfs), if the repository is using LFS?

https://github.com/eclipse-platform/eclipse.platform.swt/blame/cf3719499a6732cfcb91587bc0aefc04b5a85e35/.github/workflows/copilot-setup-steps.yml#L18-L21

```yml
      - name: Checkout code
        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
        with:
          lfs: false # why?
```

HannesWell:
> Would it be better to [enable LFS support](https://docs.github.com/en/copilot/how-tos/copilot-on-github/customize-copilot/customize-cloud-agent/customize-the-agent-environment#enabling-git-large-file-storage-lfs), if the repository is using LFS?

Unfortunately the git-lfs fetch bandwidth is limited: 
https://docs.github.com/en/billing/concepts/product-billing/git-lfs

Therefore fetching should be avoided if not really necessary. And in github workflows or agents, the natives can can  be build from scratch (for free and little overhead) easier than on developer machines. Developers should be be favored when it comes to the usage of the lfs bandwidth.

> AI agents are confused by LFS test links and have to investigate them in every session:
> 
> > Ah, the .so files are ASCII text, not binary ELF files! They're likely Git LFS pointers or some kind of placeholder. That's why they fail with "invalid ELF header".I need to build the native libraries from source. Let me try the native build:

Instead of just fetching lfs objects, I suggest to explain the situation in the `AGENTS.md` file better and guide AIs what to do.
In general you have to pay great attention what AI suggests you, especially with the setup of SWT as it seems to get confused easily about many aspects (and can give you false suggestions):
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/3198

HannesWell:
@basilevs I considered this for the general documentation of the LFS usage in
[...]

PR #3272, opened 2026-04-25
Document Git Large File Storage usage in SWT

HannesWell:
And move the general Contributing section further up to improve the `README` structure.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -23,6 +23,21 @@ SWT consists of two main parts:
 - `examples/` - Example code and snippets
 - `tests/` - JUnit tests
 
+### Native binaries
+
+This repository uses Git Large File Storage (LFS) to store precompiled native binaries.
+The `.gitattributes` file specifies which files are stored using LFS.
+
+If Git LFS is not installed and configured the LFS pointer files are not resolved on checkout and the binaries have to be compiled locally.
+For that, see the `Build Commands` section below.
+
+**NOTE**:
+For GitHub Copilot agent sessions Git LFS checkout is disabled.
+
+Cross-compiling binaries for operating systems other than the one currently in use is not possible, just as testing code for other platforms is not possible.
+If you make changes to code (Java or native) for other platforms, the binaries cannot be rebuilt and the changes cannot be tested locally.
+Instead, rely on the GitHub PR workflows created for each platform to test the changes.
+
 ## Build System
 
 ### Technology Stack
@@ -49,6 +64,8 @@ Instead: modify Java source (e.g., `OS.java`), clean/rebuild the project, then r
 **CRITICAL**: Never commit any built native binary files to git, i.e. files like `libswt-*.so`, `libswt-*.jnilib` or `swt-*.dll`.
 
 See `docs/*.md` and `bundles/org.eclipse.swt/Readme*.md` files for detailed instructions.
+
+Usually only the native binaries of the targeted platform can be built locally.
 
 ## Coding Standards
 
--- a/README.md
+++ b/README.md
@@ -81,6 +81,32 @@ The `while`-loop processes all GUI related events until the shell is disposed wh
 Before exiting, any claimed GUI resources needs to be freed.
 
 
+# Contributing to SWT
+
+Thanks for your interest in this project.
+
+For information about contributing to Eclipse Platform in general, see the general [CONTRIBUTING](https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md) page.
+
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -81,6 +81,32 @@ The `while`-loop processes all GUI related events until the shell is disposed wh
 Before exiting, any claimed GUI resources needs to be freed.
 
 
+# Contributing to SWT
```
basilevs, line 84: This should go to CONTRIBUTING.md

README.md
```diff
@@ -81,6 +81,32 @@ The `while`-loop processes all GUI related events until the shell is disposed wh
...
+
+The SWT source code is stored in this Git repository.
+
+This repository uses the [Git Large File Storage](https://git-lfs.com/) (LFS) extension to host also the precompiled SWT native binaries,
+which is therefore required unless binaries are compiled locally.
+
+When using Git from the CLI follow the Git LFS installation and setup instructions:
+- https://github.com/git-lfs/git-lfs/wiki/Installation#installing
+
+When using _EGit_ from within Eclipse follow the EGit user guide on _GIT LFS Support_:
```
basilevs, line 104: Why would agent use anything from within Eclipse? They prefer and are much more capable with command line options.

README.md
```diff
@@ -81,6 +81,32 @@ The `while`-loop processes all GUI related events until the shell is disposed wh
...
+"Click to open Eclipse-Installer Auto Launch or drag into your running installer")
+
+## Obtaining the sources
+
+The SWT source code is stored in this Git repository.
+
+This repository uses the [Git Large File Storage](https://git-lfs.com/) (LFS) extension to host also the precompiled SWT native binaries,
+which is therefore required unless binaries are compiled locally.
+
+When using Git from the CLI follow the Git LFS installation and setup instructions:
```
basilevs, line 101: Agent environment should have LFS preinstalled. They would need configuration at most if that. Generic instructions are not useful here.
Instead, specific instructions for each agent type are needed.

basilevs (changes requested):

AGENTS.md
```diff
@@ -23,6 +23,21 @@ SWT consists of two main parts:
...
+
+If Git LFS is not installed and configured the LFS pointer files are not resolved on checkout and the binaries have to be compiled locally.
+For that, see the `Build Commands` section below.
+
+**NOTE**:
+For GitHub Copilot agent sessions Git LFS checkout is disabled.
+
+Cross-compiling binaries for operating systems other than the one currently in use is not possible, just as testing code for other platforms is not possible.
+If you make changes to code (Java or native) for other platforms, the binaries cannot be rebuilt and the changes cannot be tested locally.
+Instead, rely on the GitHub PR workflows created for each platform to test the changes.
```
basilevs, line 39: Github Agent  is already very bad at unattended workflows (often refuses to run commands). I feel that this working might encourage early termination even more.

README.md
basilevs: I've checked. Only a single command is needed for GitHub hosted agent: 

    git lfs pull

basilevs (changes requested):

README.md
HannesWell: This file primarily is for humans, not agents.

README.md
HannesWell: Again, this file is for humans.

But that aside, how should it in general be possible to build and test, e.g. macOS binaries under Windows or Linux?
Cross-compiling might work, but cross-OS execution/testing? I'm not aware that this is possible in any way.

At least Github Copilot Agents on Github.com effectively only work on Linux (theoretically also on Windows):
https://docs.github.com/en/copilot/how-tos/copilot-on-github/customize-copilot/customize-cloud-agent/customize-the-agent-environment

If you run any agent locally or in another cloud environment matching the target environment, this should not affect that agent.

If an agent runs on the targeted platform it can just built the binaries and we can save the limited LFS bandwith for human developers as written in  to https://github.com/eclipse-platform/eclipse.platform.swt/issues/3271#issuecomment-4319636560.

AGENTS.md
HannesWell: As elaborated before, if targeting anything else than Linux the Github Agents (at least at Github.com) cannot use the binaries any-ways (at least I'm not aware of any way).

README.md
HannesWell: The first subsection of this one should be prominently on the landing page.

And then I think it's also fine to keep the git instructions here as they are relatively compact and many other information related to 'modifying the sources' are at this page.
It then doesn't make much sense to keep the `CONTRIBUTING.md` actually, but it's unclear if we might need them for 'legal' reasons:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2448

basilevs (approved):

PR #3274, opened 2026-04-29
[GTK] Use primary monitor for device zoom on multi-monitor setups

### LINKED ISSUE #3273: Eclipse looks broken in Ubuntu 2026-04
Starting Eclipse on my update Ubuntu 2026-04 LTS results in a broken UI.

Image

vogella:
On GTK, `Display._getDeviceZoom(long)` ignored its monitor argument and always queried the monitor at virtual coordinate (0,0); `Device.getDeviceZoom()` did the same on GTK3. On mixed-DPI multi-monitor setups this made every shell pick up the (0,0) monitor's scale factor, so a HiDPI laptop panel positioned at the top-left of the virtual screen would force shells on a standard-density secondary monitor to render at 2x.

This change honors the passed monitor handle in `_getDeviceZoom` and selects the GDK primary monitor for the initial Display zoom on GTK3, falling back to the (0,0) monitor only when no primary is available (e.g. on Wayland). `Monitor.zoom` values returned by `getMonitors()` on X11/Xwayland are now per-monitor instead of all reflecting the (0,0) monitor.

Fixes #3273

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Device.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Device.java
@@ -1107,7 +1107,7 @@ protected int getDeviceZoom() {
 		long surface = GTK4.gtk_native_get_surface(GTK4.gtk_widget_get_native(shellHandle));
 		monitor = GDK.gdk_display_get_monitor_at_surface(display, surface);
 	} else {
-		monitor = GDK.gdk_display_get_monitor_at_point(display, 0, 0);
+		monitor = getPrimaryMonitor(display);
 	}
 
 	// GDK can return null monitor in some cases thus play safe
@@ -1118,6 +1118,20 @@ protected int getDeviceZoom() {
 	}
 
 	return DPIUtil.mapDPIToZoom (dpi);
+}
+
+/**
+ * Returns the GDK primary monitor handle, falling back to the monitor at
+ * virtual coordinate (0,0) when no primary monitor is reported (e.g. on Wayland).
+ *
+ * @noreference This method is not intended to be referenced by clients.
+ */
+protected long getPrimaryMonitor(long display) {
+	long monitor = GDK.gdk_display_get_primary_monitor(display);
+	if (monitor == 0) {
+		monitor = GDK.gdk_display_get_monitor_at_point(display, 0, 0);
+	}
+	return monitor;
 }
 
 }
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
@@ -2712,7 +2712,7 @@ public Monitor[] getMonitors() {
 					int scaleFactor = (int) GDK.gdk_monitor_get_scale_factor(gdkMonitor);
 					monitor.zoom = scaleFactor * 100;
 				} else {
-					monitor.zoom = Display._getDeviceZoom(monitor.handle);
+					monitor.zoom = _getDeviceZoom(monitor.handle);
 				}
 
 				/* workarea was defined in GTK 3.4. If present, it will return the best results
@@ -6256,16 +6256,19 @@ long gdk_device_get_surface_at_position (double[] win_x, double[] win_y) {
 	return GDK.gdk_device_get_surface_at_position (device, win_x, win_y);
 }
 
-static int _getDeviceZoom (long monitor_num) {
+int _getDeviceZoom (long monitor) {
 	/*
 	 * We can hard-code 96 as gdk_screen_get_resolution will always return -1
 	 * if gdk_screen_set_resolution has not been called.
 	 */
 	int dpi = 96;
-	long display = GDK.gdk_display_get_default();
-	long monitor = GDK.gdk_display_get_monitor_at_point(display, 0, 0);
-	int scale = GDK.gdk_monitor_get_scale_factor(monitor);
-	dpi = dpi * scale;
+	if (monitor == 0) {
+		monitor = getPrimaryMonitor(GDK.gdk_display_get_default());
+	}
+	if (monitor != 0) {
+		int scale = GDK.gdk_monitor_get_scale_factor(monitor);
+		dpi = dpi * scale;
+	}
 	return DPIUtil.mapDPIToZoom (dpi);
 }
 


akurtakov (approved):
This looks good. Unfortunately, I don't have a multimonitor setup to test right now. 
FWIW, on X11 monitor at 0,0 was the primary monitor but with Wayland this is no longer mandatory .

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
```diff
@@ -6256,16 +6256,23 @@ long gdk_device_get_surface_at_position (double[] win_x, double[] win_y) {
 	return GDK.gdk_device_get_surface_at_position (device, win_x, win_y);
 }
 
-static int _getDeviceZoom (long monitor_num) {
+static int _getDeviceZoom (long monitor) {
```
HeikoKlare: I find this method and the change to it a bit confusing. It only seems to be called from a code branch is is specific to X11 + GTK3. According to @akurtakov, in that combination the monitor at 0,0 and the primary monitor are the same anyway, so this change should be obsolete (see https://github.com/eclipse-platform/eclipse.platform.swt/pull/3274#pullrequestreview-4194800502). Since the method originally did not use the passed monitor anyway, should there rather be a `getPrimaryMonitor()` method used at both placed changed in this PR?

HeikoKlare (commented):
Conceptually, the change looks sound. I have currently no chance to test it. But from how I understand the control flows, one of the changes seems obsolete and maybe there is further potential for simplification/DRY, see the specific comment.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
vogella: The change isn't obsolete: the primary-monitor fallback only runs when monitor == 0 (and yes, that's a no-op on X11+GTK3).

The fix is honoring the non-zero handle, since getMonitors() (Display.java:2715) passes monitor.handle per monitor and the old code ignored it, making every Monitor.zoom identical.                                                                                                                                                                                                                                                                                                                                        

getPrimaryMonitor() helper sounds ok to me, it is just three lines but I can adjust the change accordingly.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Device.java
```diff
@@ -1107,7 +1108,7 @@ protected int getDeviceZoom() {
 		long surface = GTK4.gtk_native_get_surface(GTK4.gtk_widget_get_native(shellHandle));
 		monitor = GDK.gdk_display_get_monitor_at_surface(display, surface);
 	} else {
-		monitor = GDK.gdk_display_get_monitor_at_point(display, 0, 0);
+		monitor = Display.getPrimaryMonitor(display);
```
akurtakov: Display extends Device and Device calling static method from Display is creating more bad than good IMO.

vogella: I agree, but @HeikoKlare suggested it. Heiko, ok to go back to the orignal change?

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
HeikoKlare: If that's true, it's a bug completely independent of this PR. Are you sure the assessment is correct? To me it sounds as if the zoom set to a monitor on X11 / GTK3 should effectively be the one of the primary monitor to make SWT work correctly (which is why the method argument is ignored). If that is not the case, then your change is probably correct but, as said, independent of the purpose of this PR.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Device.java
HeikoKlare: Depending on the conclusion on the other comment, the method may be obsolete anyway. If not, why not place it in `Device`? Already now, `Device` calls several `gdk_display[...]` methods (including the code extracted to the `getPrimaryMonitor()` method in this), so it should not hurt to place it there. Going back to the original change would not improve the situation as the code is effectively still in `Device` (just duplicated then).

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Display.java
vogella: I fail to understand your response. Without this patch my primary monitor is not used, it always uses the monitor in the fixed coordinates. So Eclipse starts completely broken on my primary device. If I start it on my laptop it starts fine, as this fits to the hard code coordinates

HeikoKlare: Are you using Wayland or X11? I understood this comment https://github.com/eclipse-platform/eclipse.platform.swt/pull/3274#pullrequestreview-4194800502 as if the change should effectively be a no-op on X11 because primary monitor zoom and zoom at (0,0) should be equal. And when using Wayland, this method is not called anyway.

bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/graphics/Device.java
vogella: Ubuntu 2026-04 uses Wayland by default. But OS.isX11() is also true under Xwayland, so on a Wayland session with normal SWT-GTK3 (no SWT_GTK4) _getDeviceZoom does get called.

````
// Display.java:2714                                                                                                                                                                                                                                                                                                                          
  if (!OS.isX11() || GTK.GTK4) {
      int scaleFactor = (int) GDK.gdk_monitor_get_scale_factor(gdkMonitor);                                                                                                                                                                                                                                                                     
      monitor.zoom = scaleFactor * 100;                                                                                                                                                                                                                                                                                                         
  } else {                                                                                                                                                                                                                                                                                                                                      
      monitor.zoom = Display._getDeviceZoom(monitor.handle);   // X11 + GTK3, includes Xwayland                                                                                                                                                                                                                                                 
  }  
````

That's the else in getMonitors() line 2715, which is exactly my setup and where I see the bug. Alex's point that (0,0) isn't necessarily the primary on Wayland applies here too since the compositor decides.                                                                                                                                                                                                                                                                                                                                      
                             
And even if (0,0) and the primary are the same, the real fix is that _getDeviceZoom now actually uses the monitor handle that's passed in. The old code threw the handle away and always queried (0,0). So every Monitor.zoom ended up being the (0,0) monitor's zoom regardless of which monitor it was supposed to represent. That part is independent of where the primary lives.
                                                                                                                                                                                                                                                                                                                                                
On the helper placement: fair, going back to inline duplication isn't really better. IMHO it is not so critical, again it is just few lines but if you prefer I can move getPrimaryMonitor(long display) over to Device instead, that should also address Alex's concern. ok?

vogella: FYI @HeikoKlare - I'm using a custom build of the SDK with this patch otherwise Eclipse is not usable for me  on Ubuntu. Alternatively to this patch: I could align the display settings of my different devices or move Eclipse between monitors, ignoring the "DPI change, so you want to restart"  dialog as a restart would break again the UI.

HeikoKlare: > But OS.isX11() is also true under Xwayland

Is that intended? And if yes, does that fit the intended behavior distinction done by the mentioned if/else block? To me it sounds as if the if and else branch may be somehow unified. In addition, passing a monitor to some `getDeviceZoom()` does not really make sense to me, as the idea of "deviceZoom" in SWT is one zoom for an SWT device, which may consist of multiple monitors. So in my understanding that should at least be named to something like `getMonitorZoom()`.

akurtakov: XWayland is X11 server running on top of Wayland compositor so it's fully intentional, all XWayland apps run in single global space like in every X11 implementation . Probably biggest difference between XWayland and pure Wayland is that DnD go through entirely different path.

HeikoKlare: > Alternatively to this patch: I could align the display settings of my different devices or move Eclipse between monitors, ignoring the "DPI change, so you want to restart" dialog as a restart would break again the UI.

Don't get me wrong, I do not want to challenge this change itself. The current implementation just does not feel that clean to me and seems to introduce more complexity for something that might be solved in a more simple, unified way. Maybe my understanding of the differences between X11 and Wayland is too limited to better assess that (but the concern regarding "deviceZoom" semantics remains). So feel free to move on if you as the ones more involved into the GTK port of SWT agree with the proposal.

akurtakov: Please create a @noreference
protected long getPrimaryMonitor(long display)  method in Device and use it both Display and Device.

vogella: Tomorrow, already off machine

vogella: done

akurtakov (approved):

PR #3279, opened 2026-05-02
Dim dirty tab bullet on inactive tabs in CTabFolderRenderer

### LINKED ISSUE #3278: New Dirty Indicator Should be Lighter for Inactive Tabs in Light Mode
The new dirty indicator replaces the astrisk with a bullet at the place of the close indicator. In darkmode for inactive dirty tabs the bullet is shown a bit darker to better show that the tab is inactive. See the the `TargetInfo.java` tab in the following screenshot:

Image

  
When comparing this to a light mode screenshot:

Image

The bullets have the same color independent of active or non-active tab. The correct solution would be that the dirty bullet of the `TargetInfo.java` tab should be a bit lighter similar to the dark mode. 

FYI @vogella

vogella:
The dirty indicator bullet in CTabFolder previously inherited the tab text foreground color. That color only differs between selected and unselected tabs in dark themes via CSS, so in light themes the bullet looked identical on active and inactive tabs. Fixing this in CSS would also dim the tab title text, which is undesirable.

This change applies a reduced alpha (140, roughly 55%) when drawing the bullet on unselected tabs, so the active/inactive distinction shows up in any theme without configuration. The 3-arg drawClose overload is preserved for binary compatibility.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/3278

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
@@ -692,17 +692,17 @@ public class CTabFolderRenderer {
 	}
 
 	void drawClose(GC gc, Rectangle closeRect, int closeImageState) {
-		drawClose(gc, closeRect, closeImageState, false);
+		drawClose(gc, closeRect, closeImageState, false, false);
 	}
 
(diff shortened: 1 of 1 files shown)


vogella:
image


image

@azoitl WDYT?

azoitl:
@vogella I like it a lot.

vogella:
> @vogella I like it a lot.

@azoitl It was your idea. :-)  Thank you for this great suggestion

azoitl:
Happy that my suggestion was helpful :blush:

Issue #2490, opened 2025-09-09
Integrate LWJGLX-SWT into SWT OpenGL.


FluBBaOfWard:
**Is your feature request related to a problem? Please describe.**
I need support for OpenGL 3 in SWT on MacOS, as far as I can tell that is not possible today with SWT.

**Describe the solution you'd like**
A way to specify which version of OpenGL to use, on MacOS this is necessary to use newer features.

**Describe alternatives you've considered**
On MacOS x86 I can use LWJGLX-SWT, but they have no aarch64 version yet.

**Additional context**
I filed an issue here [Integrate in normal SWT? #17
](https://github.com/LWJGLX/lwjgl3-swt/issues/17)

laeubi:
> I need support for OpenGL 3 in SWT on MacOS, as far as I can tell that is not possible today with SWT.

Can you explain why it is not possible?

> A way to specify which version of OpenGL to use, on MacOS this is necessary to use newer features.

What (native) functions are involved here? e.g. SWT allows customize certain things already so it might be possible to enhance.

FluBBaOfWard:
On MacOS, if you don't specify GL version you get 2.1.
The only example of how to specify which GL version to use, comes from LWJGLX, they use GLData to hold the version values before transferring them to GLCanvas. I haven't seen any other way...
The only additions that are needed seem to be the GLData & GLCanvas classes.

laeubi:
> The only example of how to specify which GL version to use, comes from LWJGLX, they use GLData to hold the version values before transferring them to GLCanvas. I haven't seen any other way...

Why is this way not sufficient? Using OpenGL with SWT usually requires a custom binding so LWJGLX should have ways to do whatever you want.

> The only additions that are needed seem to be the GLData & GLCanvas classes.

If you have concrete requirements, please don't hesitate to open a PR with proposed changes.

FluBBaOfWard:
> Why is this way not sufficient? Using OpenGL with SWT usually requires a custom binding so LWJGLX should have ways to do whatever you want.

They only provide x86 versions of their library, of course this can be fixed on their end but the additions to GLData and GLCanvas seem so small that it would be a shame not to include them in the normal SWT.

merks:
You should assume that we know nothing at all.   So we can't judge on what the small change might be.

Contributing a PR is super easy:

https://github.com/eclipse-platform/eclipse.platform.swt?tab=readme-ov-file#contributing-to-swt

[...]

PR #3280, opened 2026-05-04
#2490 Added version and profile information to GLData,

FluBBaOfWard:
MacOS uses this to request a specific GL version.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT OpenGL/cocoa/org/eclipse/swt/opengl/GLCanvas.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT OpenGL/cocoa/org/eclipse/swt/opengl/GLCanvas.java
@@ -34,6 +34,11 @@ public class GLCanvas extends Canvas {
 	static final int MAX_ATTRIBUTES = 32;
 	static final String GLCONTEXT_KEY = "org.eclipse.swt.internal.cocoa.glcontext"; //$NON-NLS-1$
 
+	static final int NSOpenGLPFAOpenGLProfile = 99;
+	static final int NSOpenGLProfileVersion3_2Core = 0x3200;
+	static final int NSOpenGLProfileVersionLegacy = 0x1000;
+	static final int NSOpenGLProfileVersion4_1Core = 0x4100;
+
 /**
  * Create a GLCanvas widget using the attributes described in the GLData
  * object provided.
@@ -80,6 +85,26 @@ public GLCanvas (Composite parent, int style, GLData data) {
 	if (data.stencilSize > 0) {
 		attrib [pos++] = OS.NSOpenGLPFAStencilSize;
 		attrib [pos++] = data.stencilSize;
+	}
+
+	if (data.profile == Profile.CORE) {
+		attrib[pos++] = NSOpenGLPFAOpenGLProfile;
+		attrib[pos++] = NSOpenGLProfileVersion3_2Core;
+	}
+	if (data.profile == Profile.COMPATIBILITY) {
+		attrib[pos++] = NSOpenGLPFAOpenGLProfile;
+		attrib[pos++] = NSOpenGLProfileVersionLegacy;
+	} else {
+		if (data.majorVersion >= 4) {
+			attrib[pos++] = NSOpenGLPFAOpenGLProfile;
+			attrib[pos++] = NSOpenGLProfileVersion4_1Core;
+		} else if (data.majorVersion >= 3) {
+			attrib[pos++] = NSOpenGLPFAOpenGLProfile;
+			attrib[pos++] = NSOpenGLProfileVersion3_2Core;
+		} else {
+			attrib[pos++] = NSOpenGLPFAOpenGLProfile;
+			attrib[pos++] = NSOpenGLProfileVersionLegacy;
+		}
 	}
 
 	/*
@@ -194,6 +219,19 @@ public GLData getGLData () {
 	data.accumGreenSize = accumColorSize;
 	data.accumBlueSize = accumColorSize;
 	data.accumAlphaSize = accumColorSize;
+
+	pixelFormat.getValues(value, NSOpenGLPFAOpenGLProfile, 0);
+	if (value[0] == NSOpenGLProfileVersion3_2Core) {
+		data.majorVersion = 3;
+		data.minorVersion = 2;
+		data.profile = Profile.CORE;
+	} else if (value[0] == NSOpenGLProfileVersionLegacy) {
+		data.profile = Profile.COMPATIBILITY;
+	} else if (value[0] == NSOpenGLProfileVersion4_1Core) {
+		data.majorVersion = 4;
+		data.minorVersion = 1;
+		data.profile = Profile.CORE;
+	}
 
 	pixelFormat.getValues(value, OS.NSOpenGLPFASampleBuffers, 0);
 	data.sampleBuffers = (int)value [0];
--- a/bundles/org.eclipse.swt/Eclipse SWT OpenGL/cocoa/org/eclipse/swt/opengl/GLCanvas.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT OpenGL/cocoa/org/eclipse/swt/opengl/GLCanvas.java
@@ -87,24 +87,18 @@ public GLCanvas (Composite parent, int style, GLData data) {
 		attrib [pos++] = data.stencilSize;
 	}
 
-	if (data.profile == Profile.CORE) {
-		attrib[pos++] = NSOpenGLPFAOpenGLProfile;
-		attrib[pos++] = NSOpenGLProfileVersion3_2Core;
-	}
 	if (data.profile == Profile.COMPATIBILITY) {
 		attrib[pos++] = NSOpenGLPFAOpenGLProfile;
 		attrib[pos++] = NSOpenGLProfileVersionLegacy;
+	} else if (data.majorVersion >= 4) {
+		attrib[pos++] = NSOpenGLPFAOpenGLProfile;
+		attrib[pos++] = NSOpenGLProfileVersion4_1Core;
+	} else if (data.majorVersion >= 3 || data.profile == Profile.CORE) {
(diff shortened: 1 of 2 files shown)


bundles/org.eclipse.swt/Eclipse SWT OpenGL/cocoa/org/eclipse/swt/opengl/GLCanvas.java
```diff
@@ -82,6 +87,26 @@ public GLCanvas (Composite parent, int style, GLData data) {
...
+			attrib[pos++] = NSOpenGLProfileVersion4_1Core;
+		} else if (data.majorVersion >= 3) {
+			attrib[pos++] = NSOpenGLPFAOpenGLProfile;
+			attrib[pos++] = NSOpenGLProfileVersion3_2Core;
+		} else {
+			attrib[pos++] = NSOpenGLPFAOpenGLProfile;
+			attrib[pos++] = NSOpenGLProfileVersionLegacy;
+		}
+	}
+
```
ferdnyc: If `data.profile == Profile.CORE`, you're going to end up adding two profile attribute pairs to the array.

1. First on lines 91-92, you'll add this pair:
   * `(NSOpenGLPFAOpenGLProfile, NSOpenGLProfileVersion3_2Core)`
2. Then when the test at line 94 fails you'll take the `else` path at line 97, and depending what `data.majorVersion` contains, you'll _**also**_ add one of these three pairs:
   * `(NSOpenGLPFAOpenGLProfile, NSOpenGLProfileVersion4_1Core)`
   * `(NSOpenGLPFAOpenGLProfile, NSOpenGLProfileVersion3_2Core)` (again!)
   * `(NSOpenGLPFAOpenGLProfile, NSOpenGLProfileVersionLegacy)`

That doesn't _seem_ like what you'd want. Should line 94 be an `else if`? Or does the logic here need to be more robust? (The code starting at line 223 makes it seem like `Profile.CORE` should only be used with numbered versions, and `Profile.COMPATIBILITY` should be 1:1 with `VersionLegacy`, but that's not at all what the code _here_ does.)

And then the comments even farther down claim that `Profile` is _only valid_ if the major+minor version is at least `3.0`. Meaning `Profile.COMPATIBILITY` is never actually a valid profile, because it's used when the version is "Legacy" (which seems implicitly to mean "less than `3.0`")?

ferdnyc:
The [Qt6 implementation](https://doc.qt.io/qt-6/qopenglversionprofile.html) of this is `QOpenGLVersionProfile`, which states that:

1. OpenGL 3.2 is the minimum required for profile support
2. A "legacy version" is an OpenGL version 3.1 or less
3. The `OpenGLContextProfile` is always `QSurfaceFormat::NoProfile` with legacy OpenGL versions
4. Code running on OpenGL versions 3.2 or greater can "choose between a restricted core profile, and a compatibility profile which might contain deprecated support functionality." (`QSurfaceFormat::CoreProfile` vs. `QSurfaceFormat::CompatibilityProfile`)

So, in their model "Legacy" doesn't correspond to "Compatibility profile" _at all_, in fact it means that profiles are unsupported so **neither** Core nor Compatibility can be selected, the profile is "No profile".

With versions >= 3.2, either the Core or Compatibility profile can be selected, though Apple's (deprecated) [`NSOpenGLProfile*` constants](https://developer.apple.com/documentation/appkit/opengl-profiles?language=objc) don't appear to have ever had any support for selecting a compatibility profile. (Or if they do, I can't find it in their incredibly-hard-to-parse docs with all of the deprecated feature names struck through for maximum reading difficulty.)

bundles/org.eclipse.swt/Eclipse SWT OpenGL/cocoa/org/eclipse/swt/opengl/GLCanvas.java
FluBBaOfWard: Yes, line 94 should be an "if else".

FluBBaOfWard:
The only reason for this at all is because if you don't ask for a version on MacOS you get the old compatibility version, on Windows (and maybe Linux as well?) you get the latest version _with_ compatibility, so on MacOS you need to specify at least NSOpenGLProfileVersion3_2Core to get the new shader (and lose the old stuff).

bundles/org.eclipse.swt/Eclipse SWT OpenGL/cocoa/org/eclipse/swt/opengl/GLCanvas.java
ferdnyc: OK, but that still doesn't account for the situation where `data.profile == Profile.CORE` and `data.majorVersion >= 4`, which _should_ be valid and should select `NSOpenGLProfileVersion4_1Core` instead of `NSOpenGLProfileVersion3_2Core`.

ferdnyc: (Again going back to the Qt implementation of this) it looks like Qt only respects the equivalent of `data.profile == profile.CORE` on macOS; everything else gets forced to `NSOpenGLProfileVersionLegacy`. (Which seems to confirm my suspicion that compatibility profiles aren't actually available on macOS.)

here's the relevant `qcocoaglcontext.mm` code from Qt6's cocoa platform plugin:

```c++
NSOpenGLPixelFormat*
QCocoaGLContext::pixelFormatForSurfaceFormat(
    const QSurfaceFormat &format)
{
    QVector<NSOpenGLPixelFormatAttribute> attrs;

    attrs << NSOpenGLPFAOpenGLProfile;
    if (format.profile() == QSurfaceFormat::CoreProfile) {
        if (format.version() >= qMakePair(4, 1))
            attrs << NSOpenGLProfileVersion4_1Core;
        else if (format.version() >= qMakePair(3, 2))
            attrs << NSOpenGLProfileVersion3_2Core;
        else
            attrs << NSOpenGLProfileVersionLegacy;
    } else {
        attrs << NSOpenGLProfileVersionLegacy;
    }
```

FluBBaOfWard: Compatibility profile is very much available, that's what you get if you don't specify version/profile.
I think I have a better solution to the code but I will fix that ion Monday, thanks for looking over this PR.

FluBBaOfWard: I guess I understood you wrong, the profiles should probably just be named Legacy/Core since you can't get the core features with backwards compatibility.

bundles/org.eclipse.swt/Eclipse SWT OpenGL/common/org/eclipse/swt/opengl/GLData.java
```diff
@@ -133,6 +138,24 @@ public class GLData {
...
+	 * The minor GL context version to use. If {@link #majorVersion}
+	 * is 0 this field is unused.
+	 */
+	public int minorVersion;
+
+	/**
+	 * The profile to use. This is only valid when
+	 * ({@link #majorVersion}.{@link #minorVersion}) is at least 3.0.
+	 */
+	public Profile profile;
```
ferdnyc, line 157: The comment here is technically no longer correct, since `LEGACY` is one of the enum members and that's used to signal a version less than `3.0`.

<b>Profile.CORE</b> is only valid when `({@link #majorVersion}.{@link #minorVersion}`) is at least 3.0.

bundles/org.eclipse.swt/Eclipse SWT OpenGL/cocoa/org/eclipse/swt/opengl/GLCanvas.java
```diff
@@ -82,6 +87,17 @@ public GLCanvas (Composite parent, int style, GLData data) {
...
+	if (data.profile == Profile.LEGACY) {
+		attrib[pos++] = NSOpenGLPFAOpenGLProfile;
+		attrib[pos++] = NSOpenGLProfileVersionLegacy;
+	} else if (data.majorVersion >= 4) {
+		attrib[pos++] = NSOpenGLPFAOpenGLProfile;
+		attrib[pos++] = NSOpenGLProfileVersion4_1Core;
+	} else if (data.majorVersion >= 3 || data.profile == Profile.CORE) {
+		attrib[pos++] = NSOpenGLPFAOpenGLProfile;
+		attrib[pos++] = NSOpenGLProfileVersion3_2Core;
+	}
```
ferdnyc, line 99: Hm. Would there be a down side to making legacy the default even in circumstances where nothing is specified? And requiring an explicit `Profile.CORE` in addition to the version numbers? IOW,

```suggestion
	if (data.profile == Profile.CORE) {
		if (data.majorVersion >= 4) {
			attrib[pos++] = NSOpenGLPFAOpenGLProfile;
			attrib[pos++] = NSOpenGLProfileVersion4_1Core;
		} else if (data.majorVersion >= 3) {
			attrib[pos++] = NSOpenGLPFAOpenGLProfile;
    		attrib[pos++] = NSOpenGLProfileVersion3_2Core;
    	} else {
			attrib[pos++] = NSOpenGLPFAOpenGLProfile;
			attrib[pos++] = NSOpenGLProfileVersionLegacy;
		}
	} else {
		attrib[pos++] = NSOpenGLPFAOpenGLProfile;
		attrib[pos++] = NSOpenGLProfileVersionLegacy;
	}
```

(Then you wouldn't technically even need `Profile.LEGACY`, except to have a value other than `Profile.CORE` that can be stuck in `data.profile`. But just not setting it would get you the same thing, so it doesn't have to be explicit.)

ferdnyc: But overall it looks workable to me as-is, too.

FluBBaOfWard: I thought it might be good that it works exactly like before if you don't specify anything.

PR #895, opened 2023-11-15
Button's setAlignment() is being considered and right margin value is recalculated depending on alignment set.

deepika-u:
Button's setAlignment() is calculating the right margin value now when text and image are set depending on the alignment of the button.

Fixes
https://github.com/eclipse-platform/eclipse.platform.swt/issues/519

@niraj-modi 
@sravanlakkimsetti 
Can anyone of you review the changes and let me know your comments please.

Issue #519, opened 2022-12-30
Button#setAlignment() has no effect when Text and Image are set


laeubi:
This was already described in [Bug 172850](https://bugs.eclipse.org/bugs/show_bug.cgi?id=172850):

> setAlignment() is not working for Buttons who hold an image and text at the same time. Maybe this is platform-specific behaviour and can not be worked around but it should be documented in the javadoc of setAlignment() that this has no effect in the described situation.

See also:
- https://github.com/eclipse-m2e/m2e-core/pull/1157#issuecomment-1367513300

lathapatil:
is it only the Javadoc change as it is mentioned WONFIX in bugzilla ?

deepika-u:
I tried the below snippet for Button setAlignment() and not seeing any problem as such.
Can anyone give us more details on this issue if this is not the right snippet to showcase the reported problem.

`package org.eclipse.swt.bugs;
import org.eclipse.swt.*;
import org.eclipse.swt.graphics.*;
import org.eclipse.swt.layout.*;
import org.eclipse.swt.widgets.*;

public class ButtonAlignmentExample {

    public static void main(String[] args) {
        Display display = new Display();
        Shell shell = new Shell(display);
        shell.setText("Button Alignment Example");
        shell.setLayout(new GridLayout(1, false));

        // Load an image (you should replace "imagePath" with your image file path)
        Image image = new Image(display, "C:\\Users\\02236C744\\Desktop\\deepika.png");

        // Create a button with text and image
        Button button = new Button(shell, SWT.PUSH);
        button.setText("Button Text");
        button.setImage(image);

        // Set the alignment to CENTER
        button.setAlignment(SWT.CENTER);

        // Set layout data to center the button in the shell
        button.setLayoutData(new GridData(SWT.CENTER, SWT.CENTER, true, true));

        shell.pack();
        shell.open();

        while (!shell.isDisposed()) {
            if (!display.readAndDispatch()) {
                display.sleep();
            }
        }

        // Dispose of the image and resources
        image.dispose();
        display.dispose();
    }
}`

I have tried on below environment
Eclipse SDK
Version: 2023-09 (4.29)
Build id: I20230830-2200
OS: Windows 11, v.10.0, x86_64 / win32
Java vendor: Eclipse Adoptium
Java runtime version: 20+36
Java version: 20

ptziegler:
> I tried the below snippet for Button setAlignment() and not seeing any problem as such.

[...]

PR #3282, opened 2026-05-05
[Win] Enabled Image+Text button ignores alignment #3277

### LINKED ISSUE #3277: [Win] Enabled Image+Text button ignores alignment
**Describe the bug**
Buttons with text and image will ignore the alignment (e.g. `SWT.LEFT`) if a foreground color has been set and they are enabled. Once they be disabled, they regard the alignment.

**To Reproduce**
```java
import org.eclipse.swt.*;

final class ButtonAlignmentProblem {

		display.addListener(SWT.Skin, event -> {
			if (event.widget instanceof Button button
			    && (button.getStyle() & SWT.CHECK) == 0) {
				button.setForeground(new Color(40, 60, 60));
			}
		});

		final Image image = display.getSystemImage(SWT.ICON_WARNING);
		final Button button1 = createButton(shell, "hello", null);
		final Button button2 = createButton(shell, "world", image);
		final Button button3 = createButton(shell, "longer text", image);

		final Button checkBox = new Button(shell, SWT.CHECK);
		checkBox.setText("Buttons enabled");
		checkBox.setSelection(true);
		checkBox.addListener(SWT.Selection, e -> {
			final boolean selection = checkBox.getSelection();
			button1.setEnabled(selection);
			button2.setEnabled(selection);
			button3.setEnabled(selection);
		});

		shell.setSize(400, 300);
		shell.open();

		while (!shell.isDisposed()) {
				display.sleep();
			}
		}

		display.dispose();
	}

	private static Button createButton(Composite parent, String text, Image image) {
		final Button button = new Button(parent, SWT.PUSH | SWT.LEFT);
		button.setLayoutData(new GridData(SWT.FILL, SWT.CENTER, false, false));
		button.setImage(image);
		button.setText(text);
		return button;
	}
}
```

**Expected behavior**
The alignment always should be respected (or ignored). Disabling buttons should not cause such visual noise.

**Screenshots**
Image

**Environment:**
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 

Reproducible for me on older SWT builds as well as with the latest commit 17667a86.

tmssngr:
It looks like the old comment `The default button with an image doesn't respect the text alignment. So we do the same for styled buttons.` is outdated.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Button.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Button.java
@@ -1403,7 +1403,9 @@ LRESULT wmNotifyChild (NMHDR hdr, long wParam, long lParam) {
 							gc.dispose ();
 						}
 
-						left += isRadioOrCheck() ? radioOrCheckTextPadding : 0;
+						left += isRadioOrCheck()
+								? radioOrCheckTextPadding
+								: image != null ? MARGIN : 0;
 						RECT textRect = new RECT ();
 						OS.SetRect (textRect, left, nmcd.top + border, right, nmcd.bottom - border);
 
@@ -1420,13 +1422,7 @@ LRESULT wmNotifyChild (NMHDR hdr, long wParam, long lParam) {
 						}
 						OS.DrawText(nmcd.hdc, buffer, buffer.length, textRect, flags | OS.DT_CALCRECT);
 						OS.OffsetRect(textRect, 0, Math.max(0, (nmcd.bottom  - textRect.bottom - border) / 2));
-						if (image != null) {
-							// The default button with an image doesn't respect the text alignment. So we do the same for styled buttons.
-							flags |= OS.DT_LEFT;
-							if (!isRadioOrCheck()) {
-								OS.OffsetRect(textRect, Math.max(MARGIN, (right - textRect.right) / 2 + 1), 0);
-							}
-						} else if ((style & SWT.LEFT) != 0) {
+						if ((style & SWT.LEFT) != 0) {
 							flags |= OS.DT_LEFT;
 						} else if ((style & SWT.RIGHT) != 0) {
 							flags |= OS.DT_RIGHT;
--- a/tests/org.eclipse.swt.tests/ManualTests/org/eclipse/swt/tests/manual/Issue0519_ButtonAlignmentExample.java
+++ b/tests/org.eclipse.swt.tests/ManualTests/org/eclipse/swt/tests/manual/Issue0519_ButtonAlignmentExample.java
@@ -1,9 +1,11 @@
 package org.eclipse.swt.tests.manual;
 import org.eclipse.swt.SWT;
+import org.eclipse.swt.graphics.Color;
 import org.eclipse.swt.graphics.Image;
 import org.eclipse.swt.layout.GridData;
 import org.eclipse.swt.layout.GridLayout;
 import org.eclipse.swt.widgets.Button;
+import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Display;
 import org.eclipse.swt.widgets.Shell;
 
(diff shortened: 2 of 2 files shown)


tmssngr:
Are there any objections for this PR?

HeikoKlare:
> It looks like the old comment The default button with an image doesn't respect the text alignment. So we do the same for styled buttons. is outdated.

Do we know when the behavior changed? So do we need to guard this behavior with a specific Windows version constraint at which the default button style changed?
The code was introduced with https://github.com/eclipse-platform/eclipse.platform.swt/commit/0822f2030fcfc1ca7d2cdb69aae86e45e1c85556 and https://github.com/eclipse-platform/eclipse.platform.swt/commit/82b8cf40f11f7c0a7dd0bd7e6d9bcc1e41f1c10e, which was after the original Windows 10 release, of which we still support some later versions.

tmssngr:
Unfortunately, I don't have older Windows 10 versions to test.

HeikoKlare:
Did you take a look into the original bug report to maybe find out why that code was introduced, i.e., how the strange default behavior of Windows buttons was detected? There might be some pointer to where you could look up if/when that behavior changed. Or maybe there is at least some information that can be used as input for an AI agent making according research.

ptziegler:
I just want to mention that there is a general problem with the alignment when both text and image is set. See #519. Perhaps something was missed in that fix?

HeikoKlare:
That's a great pointer, thank you. I just reverted the referenced change for the sake of testing and see the disabled buttons not considering the alignment again (using the snippet from https://github.com/eclipse-platform/eclipse.platform.swt/issues/3277).
So it seems like it was not a change on OS side that fixed the default button behavior but the mentioned SWT-side fix. The change proposed here may thus be considered a follow-up change to that PR to make the behavior consistent again.

HeikoKlare (commented):
In general, the change looks sound to me. We do not need to guard this behavior in any way, now that we understood that the inconsistency came from:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/895

I further tested with `Issue0519_ButtonAlignmentExample` added in the mentioned PR. I extended it with setting foreground for the buttons as done in the snippet for the issue to be solved by this PR, and I added a checkbox to enable/disable the buttons. You can see that the alignment of texts in enabled and disabled buttons does not fit for every alignment setting:
Image
Image

@tmssngr can you please extend the mentioned example for this PR to reflect the issue to be solved as well and further enhance the PR so that we see the alignment issues fixed there?

tmssngr:
@HeikoKlare I've updated `Issue0519_ButtonAlignmentExample` (like you've did) and fixed also the tiny text movements between enabled and disabled state.

HeikoKlare (commented):
Thank you for the update. With the snippet, the alignment now looks good.

I just wonder where the added magic numbers are coming from. Having random `-2` and `-1` in the code is difficult to comprehend. The two numbers could already be streamlined as the centering (using `-1`) just requires half the offset, which is why the rest of the offset calculation is divided by 2. Putting the `-1` into that calculation before doing the division would make a `-2` out of it, thus consistent with the right alignment case.
But why is it exactly 2? Few lines after we have a quite similar offset, maybe that is related?
https://github.com/eclipse-platform/eclipse.platform.swt/blob/dc0c6771ed70027d5f88f5fe7762a2bf3b2739dc/bundles/org.eclipse.swt/Eclipse%20SWT/win32/org/eclipse/swt/widgets/Button.java#L1454
Is it `MARGIN / 2` for some reason?

tmssngr:
These magic numbers were determined by try and error. I removed the latest commit from my PR.

tmssngr:
@HeikoKlare Would you mind merging this commit or is still something left to do?

HeikoKlare:
You have removed the latest commit with the fix for the text jumping between enabled/disabled state.

tmssngr:
Yes, I wrote that. I didn't have the energy to explain these numbers found by try and error.

akurtakov:
What is the status here?

Issue #3145, opened 2026-03-25
[GTK] Can draw outside window on Wayland


tmssngr:
**Describe the bug**
It is possible easily by setting clipping to draw outside a window (shell) on Wayland.

**To Reproduce**
Run following snippet on a Wayland system:
```java
import org.eclipse.swt.SWT;
import org.eclipse.swt.widgets.Composite;
import org.eclipse.swt.widgets.Display;
import org.eclipse.swt.widgets.Shell;

public class DrawOutsideShell {

    public static void main(String[] args) {
        Display display = new Display();
        Shell shell = new Shell(display);
        Composite composite = new Composite(shell, SWT.DOUBLE_BUFFERED);
        composite.setBounds(-50, 0, 500, 250);
        composite.addListener(SWT.Paint, event -> {
            event.gc.setClipping(0, 0, 500, 250);
            event.gc.fillRectangle(0, 0, 500, 250);
        });
        shell.setSize(300, 200);
        shell.open();

        while (!shell.isDisposed()) {
            if (!display.readAndDispatch()) {
                display.sleep();
            }
        }

        display.dispose();
    }
}
```

**Expected behavior**
The drawing area is clipped to the controls/shells bounds, no matter what clipping area is set by the user.

**Screenshots**
<img width="398" height="293" alt="Image" src="https://github.com/user-attachments/assets/d739aafa-94d5-4f3b-808c-ad99a4314382" />

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [ ] Windows
- - [x] Linux
- - [ ] macOS 

2. Additional OS info (e.g. OS version, Linux Desktop, etc)
Tried on latest Fedora and on Linux Mint with experimental Wayland support.

Phillipus:
I tested your Snippet on Ubuntu 24.04 Wayland and Linux Mint Wayland and it's working OK. Are you using latest SWT build? I'm using a 200% display FWIW.

tmssngr:
I also can reproduce the problem with the latest SWT build on *master*. I'm using a FullHD monitor only with no special scaling (AKA 100%).

tmssngr:
Even without setting the clipping I can draw outside the shell using the latest commit on stock Ubuntu 26.04:
```java
import org.eclipse.swt.SWT;
import org.eclipse.swt.graphics.*;
import org.eclipse.swt.widgets.Composite;
import org.eclipse.swt.widgets.Display;
[...]

PR #3284, opened 2026-05-06
[GTK] Can draw outside window on Wayland #3145

tmssngr:
`GC.clipping` seems to define the maximum possible clipping rect. However, it did not take into account that controls can be located (partly) outside a client area (e.g. of the shell). Now we reduce the maximum possible clipping rect by the parent client areas so they can't draw over them.


tmssngr:
My patch might cause other problems.

PR #3287, opened 2026-05-06
Add robust log level handling in JVSGLoggerLogManager

HeikoKlare:
The JVSGLoggerLogManager allows to customize the log level with a system property. When an invalid log level is passed, the call currently throws an exception and makes the whole rasterizer initialization fail, leading to the non-availability of an SVG rasterizer throughout the whole application lifecycle.

This change makes the system property processing more robust: when an invalid level is used, an error is printed and the implementation falls back to using "ERROR" as default log level. In addition, capitalization is ignored by always making the passed value upper case. As a result of this change, it not be possible to make the rasterizer implementation fail with an invalid system property anymore.

Resolves https://github.com/eclipse-platform/eclipse.platform.swt/pull/3283#discussion_r3196119236

### DIFF
--- a/bundles/org.eclipse.swt.svg/src/org/eclipse/swt/svg/logging/JVSGLoggerLogManager.java
+++ b/bundles/org.eclipse.swt.svg/src/org/eclipse/swt/svg/logging/JVSGLoggerLogManager.java
@@ -17,7 +17,17 @@ import com.github.weisj.jsvg.logging.Logger;
 
 public class JVSGLoggerLogManager implements LogManager {
 
-	private static final Logger.Level LEVEL = Logger.Level.valueOf(System.getProperty("org.eclipse.swt.svg.logging", "ERROR"));
+	private static final Logger.Level LEVEL = getLogLevel();
+
+	private static Logger.Level getLogLevel() {
+		try {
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.swt.svg/src/org/eclipse/swt/svg/logging/JVSGLoggerLogManager.java
```diff
@@ -14,10 +14,21 @@
 
 import com.github.weisj.jsvg.logging.LogManager;
 import com.github.weisj.jsvg.logging.Logger;
+import com.github.weisj.jsvg.logging.Logger.Level;
```
merks: Maybe use the Logger.Level  like the other parts to avoid this import?  Or make the rest consistent?

merks (commented):
Looks better. Just a tiny name consistency issue.

bundles/org.eclipse.swt.svg/src/org/eclipse/swt/svg/logging/JVSGLoggerLogManager.java
HeikoKlare: Yes, makes sense. `Logger.Level` is better to read than a generic `Level`, so I made the change consistent with how it was used alreday.

PR #3288, opened 2026-05-07
Bump version of org.eclipse.swt.svg for JSVG dependency update

HeikoKlare:
The JSVG dependencies have been updated to a new minor version including an explicit setup of the logging with OSGi capabilities. To properly reflect this change in dependencies, this bumps the version of the SWT SVG fragment to the next minor version.

Follow-up to:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/3283

@merks in my opinion this better reflects the "complexity" of the dependency change. What do you think?

### DIFF
--- a/bundles/org.eclipse.swt.svg/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.swt.svg/META-INF/MANIFEST.MF
@@ -1,7 +1,7 @@
 Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-SymbolicName: org.eclipse.swt.svg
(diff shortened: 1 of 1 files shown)


merks:
I guess we often update dependencies without incrementing the places that use them, but I think in this case it's probably a good idea.  We did actually change the lower bounds, so that's a significantly visible change if not an actual API addition.  In any case, it most definitely won't hurt and it better helps to identify the boundary.

HeikoKlare:
Thank you for the clarification. Yes, a dependency update does usually not mean that we bump the version of its consumers in the same way. This is just a case-specific proposal for a version bump upon a dependency update, motivated by what you have explained.

Issue #3286, opened 2026-05-06
[macOS] Umbrella: Compatibility with macOS 26 SDK


HeikoKlare:
## Background

macOS 26 (Tahoe) introduces behavioral and visual changes when an application binary is linked against the macOS 26 SDK. macOS enforces these changes based on the SDK version embedded in the **process launcher binary** (i.e., the `eclipse` binary from the Equinox launcher or the `java` binary of the JVM used to launch the application). The SDK versions of SWT's own native libraries do not determine this behavior.

This matters for Eclipse because:

- **Eclipse products** (e.g., the Eclipse IDE) are started via the **Equinox launcher** (`eclipse` binary). The macOS SDK version embedded in that binary determines the runtime appearance and behavior.
- **Runtime Eclipse instances** launched from within the IDE (e.g., for plug-in development) are started via the **JVM** (`java` binary). Most JVM distributions (Temurin, official OpenJDK) embed macOS SDK 14 or 15, so runtime instances are unaffected.

The issues described here were first observed when the Jenkins CI agent used to build the Equinox binaries was upgraded from macOS 11 to macOS 26, causing the Equinox launcher to embed the macOS 26 SDK:
- Trigger PR: https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3739
- Temporary fix (revert to macOS 15 build agent): https://github.com/eclipse-equinox/equinox/pull/1273

The temporary fix restored the previous macOS 15 SDK for the Equinox launcher binaries. **The goal of this umbrella issue is to track the work required to make Eclipse fully compatible with the macOS 26 SDK**, so that we can eventually build against it intentionally and with confidence.

> **Note:** The issues were initially only observed on the aarch64 (Apple Silicon) build, because only the aarch64 CI agent was upgraded to macOS 26. The x86_64 build continued to use a macOS 15 SDK and was unaffected.

---

## How to Reproduce (Without Equinox Binaries)

It is not necessary to use Equinox binaries built with the macOS 26 SDK to reproduce these issues. Any JVM binary built against the macOS 26 SDK will trigger the same behavior when used to launch an Eclipse instance from the IDE.

**Confirmed trigger:** The **Homebrew distribution of OpenJDK 25** is built against macOS SDK 26.2 (verified via `otool -l`):

```
% otool -l /opt/homebrew/Cellar/openjdk/25.0.2/.../java | grep -A 8 'Load command 10'
  minos 11.0
    sdk 26.2
[...]

PR #3289, opened 2026-05-07
[macOS] Enable TextField borders on macOS 26

SougandhS:
Restore visible borders for SWT text controls on macOS 26 and later to improve control visibility and readability.

Part of https://github.com/eclipse-platform/eclipse.platform.swt/issues/3286 and https://github.com/eclipse-platform/eclipse.platform.ui/issues/3885

**Before [Light]**
Screenshot 2026-05-07 at 5 24 23 PM

**After[Light]**
Screenshot 2026-05-07 at 5 25 09 PM

**Before [Dark]**
Screenshot 2026-05-07 at 5 26 01 PM

**After [Dark]**
Screenshot 2026-05-07 at 5 26 23 PM

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Text.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/cocoa/org/eclipse/swt/widgets/Text.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2014 IBM Corporation and others.
+ * Copyright (c) 2000, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -518,6 +518,10 @@ void createHandle () {
 		if ((style & SWT.BORDER) == 0) {
 			widget.setFocusRingType (OS.NSFocusRingTypeNone);
 			widget.setBordered (false);
+		} else {
+			if (OS.VERSION >= OS.VERSION(26, 0, 0)) {
+				widget.setBordered (true);
+			}
 		}
 		/*
 		 * Bug in Cocoa: On OSX 10.10, setting the alignment on the search field


Phillipus:
Is this PR intended to be merged after the Equinox launcher is compiled with macOS 26 SDK? This checks only the running macOS version so Text controls on macOS 26 that have `SWT.BORDER` style now look different:

Before this PR (correct border style):

before

After this PR:

after

SougandhS:
>This checks only the running macOS version so Text controls on macOS 26 that have SWT.BORDER style now look different:

Is there any current build agent/ binary check APIs in SWT ?

Phillipus:
> Is there any current build agent/ binary check APIs in SWT ?

Not API. I think the idea as @HeikoKlare says is to build the equinox binary on macOS SDK 26 at some point in the future.

As mentioned in https://github.com/eclipse-platform/eclipse.platform.swt/issues/3286 the idea is to fix problems that occur when running SWT or an app that consumes SWT (Eclipse, RCP app) from a binary that has been compiled with macOS 25 SDK. Right now this seems to be only the Homebrew Java 25. Eclipse is launched with the Equinox launcher compiled with macOS 15 SDK and most Java binaries with 14 or 15.

In this issue setting the Text control's border by checking whether the running macOS is macOS 26 or later will add an additional border in all cases.

tobiasmelcher:
@Phillipus @HeikoKlare   We could use https://github.com/eclipse-platform/eclipse.platform.swt/pull/3300 with method OS.getMachOSDKVersion to query the SDK version and set the borders only when necessary.

HeikoKlare:
Thank you for working on this.

Before discussing about technical details, I would like to raise the question if the proposed solution is actually what we intend to do. From my understanding, this is a workaround, isn't it? Enabling these borders around text widgets makes them visible, but not in a way that seems conforms to native look & feel. At least the dark colored border with non-rounded edges of the text fields shown in the screenshots is nothing I find in native look & feel.

From my understanding, prior to macOS 26 the native look & feel of text widgets comprised light borders and dark "underline", like you can see here in the native TextEdit settings dialog (and in every SWT application):
image

With macOS 26, the native look & feel of text widgets changed to using a light border with rounded edges, as you can see in the native TextEdit settings dialog:
image

It should be somehow possible to adopt that look & feel on both dark and light backgrounds, e.g., see this white background of the search field of the TextEdit application:
image

Maybe macOS provides further options to customize this look & feel. That I have not checked yet.

To adopt macOS 26 look & feel, maybe there are new configuration options to NSTextField that can/need to be set? In particular the configuration around the `bezeled` property might be interesting. Since the rounded border/bezel is present when building an SWT application against macOS 26 SDK, but it only seems to be cut off, maybe it is also about changed control default sizes or the like that need to be adapted in the SWT `Text` implementation.

tobiasmelcher:
@HeikoKlare  I spend now the whole day analyzing the issue and come up with following finding https://github.com/eclipse-platform/eclipse.platform.swt/pull/3302 - I think it is a bug on MacOS side.

SougandhS:
Closing this PR as there's a better handling done via https://github.com/eclipse-platform/eclipse.platform.swt/pull/3302

PR #3260, opened 2026-04-22
Text control should not redraw indefinitely

basilevs:
State: This work is now blocked by https://github.com/eclipse-egit/egit/issues/180

Non-empty `Text` control with search style schedules redraws on every event loop iteration. This happens even if all custom SWT code is removed.
Empty Text control does the same if a backgound color is set.
```
Thread [main] (Suspended (breakpoint at line 2150 in Widget))	
	Text(Widget).setNeedsDisplay(long, long, boolean) line: 2150	
	Display.windowProc(long, long, long) line: 6476	
	OS.objc_msgSend(long, long) line: not available [native method]	
	SWTSearchField(NSControl).stringValue() line: 122	
	Text.drawInteriorWithFrame_inView_searchfield(long, long, NSRect, long) line: 751	
	Text.drawInteriorWithFrame_inView(long, long, NSRect, long) line: 697	
	Display.windowProc(long, long, long, long) line: 6661	
	OS.objc_msgSendSuper(objc_super, long, NSRect) line: not available [native method]	
	Text(Widget).drawRect(long, long, NSRect) line: 813	
	Text.drawRect(long, long, NSRect) line: 759	
	Display.windowProc(long, long, long) line: 6246	
	OS.objc_msgSendSuper(objc_super, long, long, long, long, boolean) line: not available [native method]	
	Display.applicationNextEventMatchingMask(long, long, long, long, long, long) line: 5569	
	Display.applicationProc(long, long, long, long, long, long) line: 5970	
	OS.objc_msgSend(long, long, long, long, long, boolean) line: not available [native method]	
	NSApplication.nextEventMatchingMask(long, NSDate, NSString, boolean) line: 92	
	Display.readAndDispatch() line: 3992	
      .....
```
See https://github.com/eclipse-platform/eclipse.platform.ui/issues/3920

Affected patforms:
- MacOS SDK 14.0 aarch64 - high CPU use
- MacOS SDK 14.4 aarch64 - high CPU use
- MacOS SDK 24.4 aarch64 - Display.asyncExec() never executes
- GTK4

This PR fixes the problem on MacOS by reverting a fix for  https://bugs.eclipse.org/bugs/show_bug.cgi?id=266206, conflicting with overall AppKit paint damage scheduling approach.

AppKit resets needsDisplay flag when control is drawn. If the flag is set asynchronously and draw methods set it, infinite redraw loop happens. This happens even on uncustomized controls, as AppKit's native draw methods do set it (in complex controls with overlaying cells).

PR #3295, opened 2026-05-10
Test CTabItem rendering

basilevs:
Pixel color verification for CTabItem's tabs.

A regression test for #3260 where tabs fail to render if
`NSView.setClipsToBounds()` is omited before scheduling redraw.

This test fails to reproduce the problem, but is useful as rendering is
not tested in this suite yet.

### DIFF
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_custom_CTabFolder.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_custom_CTabFolder.java
@@ -14,13 +14,15 @@
 package org.eclipse.swt.tests.junit;
 
 
+import static org.eclipse.swt.tests.junit.SwtTestUtil.hasPixel;
+import static org.eclipse.swt.tests.junit.SwtTestUtil.openShell;
 import static org.junit.jupiter.api.Assertions.assertArrayEquals;
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertThrows;
-import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.Assertions.fail;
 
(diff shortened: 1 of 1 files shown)


tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_custom_CTabFolder.java
```diff
@@ -110,6 +114,8 @@ public void test_setFontLorg_eclipse_swt_graphics_Font() {
 
 /* custom */
 protected CTabFolder ctabFolder;
+private Color RED;
+private Color BLUE;
```
merks: It seems odd to use an all caps name for something that’s not a constant.

basilevs: Colors in SWT depend on Display, which I access via Shell, which is initialized after test starts. I could try to use `Display.getCurrent()` but this would be riskier (`Display` can be recreated). Therefore these are constants effectively but not in practice. Should I make them lowercase?

basilevs:
Windows CI is passing, Linux and MacOS are not. The test is passing on my MacOS. Configuration issue? The test seems trivial - it should work, as long as graphical context is available.

How do I get screenshots? They are saved to `/tmp` and are not archived.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_custom_CTabFolder.java
merks: Actually there are now Color constructors without a device and colors need not be disposed so they don’t really depend on a display/device. 

When I see an all caps name, I expect it to static final, so yes, I’d prefer the name to match general expectations…

basilevs: I've made constants final using new constructors.

akurtakov:
Would you please rebase on master ?

basilevs:
Tests started to fail on Windows CI. They passed there on a previous rebase.  Screenshots prove tabs are not rendered in 100ms. What am I doing wrong?

org eclipse swt tests junit Test_org_eclipse_swt_custom_CTabFolder 1  false, 0

Now:
Gtk3 - pass
Gtk4 - fail
Macos aarch64 - pass
Macos x86_64 - fail
Windows - fail

basilevs:
@HeikoKlare Windows regression might be related to recent pushes of GC redundancy fixes. Could you please run these tests with and without the changes?

basilevs:
Windows has recovered.

basilevs:
Not sure what to do here. The funcitonality works very slow (?) on some platforms.

Issue #3306, opened 2026-05-15
StyledText treats shortcuts incorrectly with Japanese keyboard layout


tmssngr:
**Describe the bug**
If the US keyboard layout is selected, a configured Cmd+A shortcut works. Not so, if the "Japanese - Kana" keyboard layout is selected.

**To Reproduce**
Run this snippet, one time with US keyboard layout, another time with Japanese keyboard layout.
```java
import org.eclipse.swt.*;
import org.eclipse.swt.custom.*;
import org.eclipse.swt.graphics.*;
import org.eclipse.swt.layout.*;
import org.eclipse.swt.widgets.*;

public class StyledTextCtrlATest {

	public static void main(String[] args) {
		final Display display = new Display();

		final Shell shell = new Shell(display);
		shell.setLayout(new FillLayout());

		final StyledText styledText = new StyledText(shell, SWT.V_SCROLL | SWT.H_SCROLL | SWT.READ_ONLY);
		styledText.setKeyBinding('a' | SWT.MOD1, ST.SELECT_ALL);
		styledText.setText("some text");

		shell.setSize(400, 300);
		shell.open();

		while (!shell.isDisposed()) {
			if (!display.readAndDispatch()) {
				display.sleep();
			}
		}

		display.dispose();
	}
}
```

**Expected behavior**
I would expect in both cases the selection of all text.

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [x] All OS
- - [ ] Windows
- - [ ] Linux
- - [ ] macOS 

2. Additional OS info (e.g. OS version, Linux Desktop, etc)
**Tried on macOS 14**, though I think the problem might apply to all platforms.

tmssngr:
Possible reason: `StyledText.handleKey(Event)` only checks for a key binding with `event.keyCode` (here some unicode), but not for `event.character` (here: `'a'`).

tmssngr:
@ethznn: This bug should be reproducible with Korean keyboard layout, too. Ctrl+A produces a key-event with `keyCode == 12609` and `character == 'a'` for me.

ethznn:
Confirmed with Korean 2-Set on macOS — with one correction to the test itself.

[...]

PR #3307, opened 2026-05-15
StyledText treats shortcuts incorrectly with Japanese keyboard layout #3306

tmssngr:
On MacOS 14 with "Japanese Kana" keyboard pressing Cmd+A results in event.keycode == 12385 while event.character == 'a'.

This PR fixes the problem for me.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/StyledText.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/StyledText.java
@@ -5728,12 +5728,13 @@ void handleHorizontalScroll(Event event) {
  * @param event keyboard event
  */
 void handleKey(Event event) {
-	int action;
 	caretAlignment = PREVIOUS_OFFSET_TRAILING;
+	int action = SWT.NULL;
 	if (event.keyCode != 0) {
 		// special key pressed (e.g., F1)
 		action = getKeyBinding(event.keyCode | event.stateMask);
-	} else {
+	}
+	if (action == SWT.NULL && event.character != 0) {
 		// character key pressed
 		action = getKeyBinding(event.character | event.stateMask);
 		if (action == SWT.NULL) {


Phillipus:
Is this a problem only on on macOS 14? That's out of date now. Is it a problem on 15.7.x and 26.x?

tmssngr:
No, this problem is reproducible with MacOS 26.5, too. I have not tried on other systems though.

Are these failing Windows tests caused by this PR?

tmssngr:
Is this worth merging?

HeikoKlare:
This change seems to be modying core key handling of `StyledText`. Without any explanation of the conceptual idea of the fix, why it may be sound, and how the issue-to-be-solved can be reproduced, I am not sure if anyone will consider this ready to be merged.

tmssngr:
Please find steps to reproduce in #3306. I thought to have explained the reason in the commit message.

HeikoKlare:
Thank you for the reference. I usually do not look at the commit message but only at the PR description, and I did not recognize that there seems to be an implicit reference to the issue in the PR title. It helps to just have a proper reference in the description.

Still I do not see an explanation of the conceptual idea of the fix and why it may be sound in the issue, the commit message or the PR description. It just says that some combination of keycode and character does not work, but that does not say anything about the concept of the fix.

tmssngr:
Before my fix the event had the keycode 12385, so it took the if statement. But for that keycode, no keybinding existed, so nothing was done. Now it first tries with the keycode as before, but if no keybinding was found, it tries to look for a character keybinding, too. And then it finds something.

HeikoKlare:
Yes, so far I understand how exactly the problematic case works because of the change.
My question is regarding how the concept looks like for how the method is supposed to operate in general now. As said, this looks like core interaction functionality of the `StyledText` to me, so we have to understand how that method is supposed to operate now, such that all existing used cases keep working. Since you seem to have already invested the time to understand how the control and data flow of this method works in detail to be able to provide the fix proposal without breaking existing functionality, it would be great to just write that down and share it, so that a potential reviewer that does not have to build this whole knowledge again.

tmssngr:
Sorry, I don't understand what further information you need. If this is not sufficient for you, just close this PR. I had the hope that other users, especially in Japan, also would like to benefit from my fix.

Phillipus:
I tested this on Mac using the Japanese Kana keyboard layout. Without this PR pressing MOD1+A doesn't perform the Select All action, with this PR it does.

I would think that MOD1 + A should perform the Select All action in all keyboard layouts.

@HeikoKlare it seems like this fixes an edge case with the Japanese Kana keyboard layout (and possibly others) in the stand-alone use of the `StyledText` control. As long as it doesn't cause a regression I would have thought this would be useful in that case?

Phillipus:
Further investigations...

A workaround:

```java
styledText.setKeyBinding(12385 | SWT.MOD1, ST.SELECT_ALL);
```

Phillipus:
Seeing as how some folks are using AI to summarise issues/PRs, I'll add this one...

### What happens

When you press Cmd + A (for "Select All") on a Japanese Kana keyboard layout:
- event.keyCode = 12385 (a Unicode value corresponding to a Japanese character, like ち / "chi")
- event.character = 'a' (the expected Latin character)

The original code in StyledText.handleKey(Event event) only checked key bindings using `event.keyCode | event.stateMask` first. Since there's no key binding registered for `12385 | SWT.MOD1`, the shortcut does nothing.

### The fix in this PR
The PR changes `handleKey` to fall back to checking the character if no binding is found for the keyCode.

This makes standard shortcuts (like Cmd+A, Cmd+C, etc.) work even when the underlying macOS event reports a Japanese keycode for Latin keys under certain input methods.

### Why this occurs on Japanese layouts
macOS (and other platforms with complex input methods / IME) often sends the "committed" or native character code in keyCode when using kana/kanji input modes, while still providing the base Latin character for compatibility.

Eclipse's key binding system wasn't handling this mismatch for StyledText.
This is a common class of issue with CJK (Chinese/Japanese/Korean) input methods in cross-platform toolkits.

HeikoKlare:
> it seems like this fixes an edge case with the Japanese Kana keyboard layout (and possibly others) in the stand-alone use of the `StyledText` control. As long as it doesn't cause a regression I would have thought this would be useful in that case?

Sure, it's definitely useful. But you mention the important point as well "as long as doesn't cause a regression". The code does not just add a specific if statement for the edge or the like, it changes how the method processes key event parameters in general. And I am not able to spend the time to get all the knowledge of how this works (including control and data flow of the method), in particular because the creator of such a change needs to have done that anyone and would the person I would expect to post the information together with the change proposal.

So if there is someone who is confident with the change or invests the time into fully understanding the impact, it's of course fine for me to merge this. I just wanted to point out that I am not sure if there will be someone doing that (and that I least I will not be that person).

tmssngr:
Maybe someone can simply merge it and we quickly will notice if it will cause any regressions we are not aware of now. We are using this PR for SmartGit 26.2 internally and did not notice any regression yet. I'm sure, without trying we won't find out.

HeikoKlare:
This change fixes a specific case that no one seems to have noticed/reported for years. It adapts hows combinations of keyCode and character are processed, without any explanation about the existing and new data flows and processings of those combinations. Merging this with just some testing and with involving any conceptual understanding of the change means that we proceed in a try-and-error manner (without any need, as one could just read and understand the method and derive the necessary knowledge). Not unlikely that this change will replace the bug for one long-unnoticed case with another one that will again be noticed after quite a while as the case occurs that seldom.
As an example, we just have a case where we did a change three months ago (also in key event processing) and only now with the release we see a report for a regression that this introduced:
- https://github.com/eclipse-platform/eclipse.platform.ui/issues/4080

With my previous comments, I always assumed that the conceptual knowledge about soundness of this change exists and is just not shared yet, but it more and more sounds as if the fix was just done in a try-and-error manner rather than a conceptual understanding of the affected method's behavior.

So as the last comment from my side on this: I strongly discourage doing such a try-and-error change without having anyone understanding the change conceptually. But I also won't block this, so if anyone invests the time to get that understanding, already has it, or feels confident in any other way, feel free to merge.

tmssngr:
@ethznn Does this PR fix the problem for you?

ethznn:
Yes — measured, on Korean 2-Set.

I could not build `master` here, so I applied this PR's `handleKey` hunk to 3.130.0's `StyledText`, compiled that one class and swapped it into an unsigned copy of the published jar, then ran the same probe I used on #3306. Same machine, same probe, only the hunk differs: macOS 27.0, JDK 21.0.12.1, SWT 3.130.0 (`SWT.getVersion()` = 4969).

`Cmd+C` over a `가나다라` selection, with the clipboard seeded to a sentinel before each attempt:

| build | event | clipboard |
|---|---|---|
| 3.130.0 as published | `keyCode=12618('ㅊ') character=99('c') CMD` | `"SENTINEL-99131"` → `"SENTINEL-99131"` — copy did not run |
| 3.130.0 + this PR | `keyCode=12618('ㅊ') character=99('c') CMD` | `"SENTINEL-71670"` → `"가나다라"` — copy ran |
| 3.130.0 + this PR, U.S. layout | `keyCode=99('c') character=99('c') CMD` | `"SENTINEL-88657"` → `"가나다라"` — copy ran |

So the Hangul keyCode no longer swallows the shortcut, and the U.S. path still resolves the same way.

On the review concern about changing the general flow rather than adding a special case: the new reachability is keys with a non-zero `keyCode` **and** a non-zero `character` whose keyCode lookup misses. Those previously could not reach the character lookup at all, and now can. That is the surface worth probing for a regression — a physical key whose `character` happens to collide with a bound letter while its `keyCode` is bound to nothing. My run exercised only the copy path, so I can speak for that and not for the rest.

One measurement from #3306 that may matter for the fix's shape: the control-character fallback inside that branch still sits behind `event.character <= 31`, and on a Hangul layout `Ctrl+C` arrives as `keyCode=12618('ㅊ') character=3(0x03)`. With this PR the fallback is now reachable for it (`3 + 64 = 'C'`), which is the intended path on platforms where `MOD1` is Ctrl.

PR #3299, opened 2026-05-12
[Gtk4] Paint before child snapshot in Display snapshotDrawProc

akurtakov:
Fixes JFace Forms rendering.

PR #3308, opened 2026-05-15
[Gtk4] Fix PaintListener paint ordering for all widget types

akurtakov:
On Gtk 4, PaintListener on leaf controls (Button, Label, etc.) drew custom content under the native widget appearance instead of over it (issue #2819). Fixing this by simply moving snapshotToDraw after the children loop regressed JFace Forms rendering (PR #3299).

New paint-ordering scheme that mirrors Gtk 3's
signal split (EXPOSE_EVENT_INVERSE for containers, DRAW/EXPOSE_EVENT for leaf controls).

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Composite.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Composite.java
@@ -558,6 +558,18 @@ void snapshotBackground (long handle, long snapshot) {
 	Graphene.graphene_rect_free(rect);
 }
 
+@Override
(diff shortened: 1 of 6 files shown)


akurtakov:

jonahgraham (approved):
This change fixes the top comment case in #2819, but https://github.com/eclipse-platform/eclipse.platform.swt/issues/2819#issuecomment-3583221449 shows that we are calling paint listener twice still on GTK4 even with this change.

image

Snippet for the above is the one from  https://github.com/eclipse-platform/eclipse.platform.swt/issues/2819#issuecomment-3583221449  but I changed the color to red to make it easier to see.

I am approving this as it does move forward, but I am not going to let it close the issue until we discuss further.

Issue #3025, opened 2026-01-29
[Linux] SWT 4.39 - Tree with size "preferred" disappears


pbi-qfs:
**Describe the bug**
A tree control with size "preferred" is not visible or disappears when changing the style.

**To Reproduce**
- Install SWT Controls demo as described in https://eclipse.dev/eclipse/swt/examples.html
- Open Controls demo (standalone or as view in Eclipse)
- Select "Tree" tab -> Tree already is (almost invisible)
- Select a fixed size, then "preferred" again, or resize the window -> Tree is visible
- Toggle any style, w.g. `SWT.CHECK` -> Tree disappears

**Expected behavior**
The tree should be visible all the time.

**Screenshots**
<img width="865" height="698" alt="Image" src="https://github.com/user-attachments/assets/774f678f-6eb6-4819-afe1-5d2575763ff7" />

<img width="1532" height="781" alt="Image" src="https://github.com/user-attachments/assets/919f9990-1e94-4abd-b0d4-a1c8c4d46b74" />

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [ ] Windows
- - [X] Linux
- - [ ] macOS 

2. Additional OS info (e.g. OS version, Linux Desktop, etc)
- Seen on: Ubuntu 18, 20, 24

3. JRE/JDK version
- JDK 21 (Tested: Zulu)
- JDK 25 (Tested: Zulu, Temurin, Oracle)
 
**Version since**
4.39M1, still seen in M2

**Workaround (or) Additional context**
Maybe unrelated: When opening the SWT Controls view in Eclipse, this error is logged six times in the console:
```
(Eclipse:37891): Gtk-WARNING **: 12:36:28.297: gtk_widget_size_allocate(): attempt to allocate widget with width -5 and height 0
```

akurtakov:
It would be nice if you can spend the time to come with a minimal snippet to reproduce the issue. That would allow me to do more work on SWT.

pbi-qfs:
We use the aforementioned default SWT Controls example from the SWT project source as testing base (https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/examples/org.eclipse.swt.examples/src/org/eclipse/swt/examples/controlexample/TreeTab.java), so it should be easy to reproduce and track down using these official sources. I do not have any other source example, and I'm not familiar enough with the SWT sources to strip the controls example down.

[...]

PR #3309, opened 2026-05-15
[Gtk] Fix Tree preferred size becoming invisible after style toggle

akurtakov:
When a Tree with no explicit columns is laid out with SWT.DEFAULT size hints (e.g. GridData(SWT.DEFAULT, SWT.DEFAULT)), the GTK preferred-size call on the not-yet-rendered GtkTreeView returns width 0. The resulting zero-width allocation triggers the ZERO_WIDTH hide path, making the widget invisible.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Tree.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Tree.java
@@ -660,6 +660,23 @@ Point computeSizeInPixels (int wHint, int hHint, boolean changed) {
 		}
 	}
 
+	/*
+	 * Single column Trees may return a very small initial width if the tree has not
+	 * yet been rendered. Walk all root items to compute the required width directly
+	 * from their cell renderers.
+	 */
+	if (wHint == SWT.DEFAULT && size.x == 0 && columnCount == 0) {
+		long col = GTK.gtk_tree_view_get_column(handle, 0);
+		if (col != 0) {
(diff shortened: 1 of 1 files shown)


akurtakov:
Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/3025

akurtakov:
This is alternative fix for a similar problem to https://bugs.eclipse.org/bugs/show_bug.cgi?id=465056 but more dynamic. I'll switch Table to this approach too when we open next stream.

HeikoKlare (commented):
I can reproduce the issue reported in #3025 and can confirm that this change fixes the issue. I've tested on a WSL system.
The code itself also looks functionally sound to me. I just wonder if this iteration of tree items at OS level may have a performance impact. I remember mentions of performance issues of trees on GTK in the past. And we are currently investigating some issues on Windows where retrieving tree items from the OS is very expensive, such that we try to replace the code with the usage of SWT metadata.

In case we are sure there is no risk of performance issue with this, I leave it up to your judgement whether it is still safe to do for RC1.

akurtakov:
In theory this PR can introduce performance costs, but only in a very specific scenario: a single‑column Tree with a very large number of root items (2K+), and only when GTK reports width = 0.  I haven't seen such case in the wild.

akurtakov:
@iloveeclipse As https://bugs.eclipse.org/bugs/show_bug.cgi?id=465056 was reported by you and this one is almost identical wdyt, should we push it for RC1?

iloveeclipse:
I would wait for 4.41, it is not a recent regression.

jonahgraham (approved):

PR #3315, opened 2026-05-21
Save test failure screenshots

### LINKED ISSUE #3303: Test failure screenshots are not archived in GitHub Actions
Screenshots are saved in /tmp folder and do not reach artifacts.

A test harness from platform aggregator relies on tycho-surefire to start Platform and discover a suitable storage location, but SWT is using Maven Surefire, so a tmp fallback is used instead.

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/c82c02b58c46d0082ad127440d8155e2a9d7f632/eclipse.platform.releng/bundles/org.eclipse.test/src/org/eclipse/test/Screenshots.java#L110-L143

```java
    public static File getResultsDirectory() {
        File resultsDir = getJunitReportOutput(); // ends up in testresults/linux.gtk.x86_6.0/<class>.<test>.png


        if (resultsDir == null) {
            File eclipseDir = new File("").getAbsoluteFile();
            if (isRunByGerritHudsonJob()) {
                resultsDir = new File(eclipseDir, "/../").getAbsoluteFile(); // ends up in the workspace root
            } else {
                resultsDir = new File(System.getProperty("java.io.tmpdir"));
            }
        }


        resultsDir.mkdirs();
        return resultsDir;
    }


    private static File getJunitReportOutput() {
		boolean platformAvailable = false;
		try {
			Class.forName("org.eclipse.core.runtime.Platform", false, Screenshots.class.getClassLoader());
			platformAvailable = true;
		} catch (ClassNotFoundException e) {
			platformAvailable = false;
		}
		if (platformAvailable) {
			String[] args = Platform.getCommandLineArgs();
			for (int i = 0; i < args.length - 1; i++) {
				if ("-junitReportOutput".equals(args[i])) { // see library.xml and org.eclipse.test.EclipseTestRunner
					return new File(args[i + 1]).getAbsoluteFile();
				}
			}
		}
		return null;
    }
```

basilevs:
This is a continuation of [Screenshot saving in Platform aggregator][1].

Tests attempt to save screenshots on failure, but so far, the save location has been incorrect (`/tmp`) and screenshots were wasted. Now screenshots are saved in Maven output directory and this PR captures them in artifacts, so that contributors can inspect CI failures in more detail.

[1]: https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3819

Fixes #3303.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -141,6 +141,7 @@ jobs:
         if-no-files-found: warn
         path: |
           ${{ github.workspace }}/**/target/surefire-reports/*.xml
+          ${{ github.workspace }}/**/target/screenshots/*.png
           ${{ github.workspace }}/**/*.log
           ${{ github.workspace }}/**/*.dump
           ${{ github.workspace }}/**/*.dumpstream


HeikoKlare (approved):
Thank you for this follow-up. Looks like the right thing to do.
In my opinion, we can merge this despite freeze period as it's only about GHA build configuration.

Issue #3296, opened 2026-05-11
OutOfMemory when disposing images due to many copies of image in GC.operations


meyertom:
**Describe the bug**
We are using dynamic images that are then displayed on a Canvas for a process-state visualization.
The addition of the operations list in GC for ImageOperations is a problem in our case. the list contains all (Image)Operations since the opening of the Window, leading to a very long list of operations acting on the same image(s).  When such an image is disposed, a copy of the image is made for every ImageOperation (acting on this image) contained in the operations list of the gc.
In our setup this leads to thousands of copies of a single image, that are created on disposal of the image.

Inverting the disposal order is not an option in our case, 

I have no clear understanding on when the operations list is used in GC (it seems to  be used only when GC. refreshFor(Drawable drawable, ImageHandle imageHandle is invoked. Therefore I'm not in a position at the moment to suppose a solution for the problem.
Removing old image operations from operations , when the image that is drawn and the boundaries are already contained in operatons might be an idea but the logic when an operation in operations might be removed might get complicated.

**To Reproduce**
Create an image.
Draw the image to a Canvas
modify the image 
redraw the image in the Canvas.
dispose the image
This leads to two copies of the image beeing created in the ImageOperations of the GC.

**Expected behavior**
Only operations that help determine the state of the Drawable should remain in GC.operations.

**Screenshots**

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [X] Windows
- - [ ] Linux
- - [ ] macOS 

2. Additional OS info (e.g. OS version, Linux Desktop, etc)

3. JRE/JDK version
 
**Version since**
 We encontered this problem whe switching to SWT 4.39, but te probelm exists since the introduction of the operations list in GC.

**Workaround (or) Additional context**
Any known workarounds for the problem?

iloveeclipse:
If I understand it right , you refer to this change: https://github.com/eclipse-platform/eclipse.platform.swt/pull/2331 and it is Windows specific issue.

@HeikoKlare , could you please check?

iloveeclipse:
@meyertom : a self contained example which shows you problem would be for sure helpful.

HeikoKlare:
[...]

PR #3316, opened 2026-05-21
[Win32] Remove unnecessary color disposal in GC

HeikoKlare:
GC operations unnecessarily register stored colors for disposal, though colors do not need to be disposed anymore. This change removes the obsolete disposal tracking.

Contributes to https://github.com/eclipse-platform/eclipse.platform.swt/issues/3296

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
@@ -4859,7 +4859,6 @@ private class SetBackgroundOperation extends ReplaceableOperation  {
 	SetBackgroundOperation(Color color) {
 		RGB rgb = color.getRGB();
 		this.color = new Color(rgb);
-		registerForDisposal(this.color);
 	}
 
 	@Override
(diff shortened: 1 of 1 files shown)


fedejeanne (commented):
LGTM but just for my understanding: the reason we shouldn't dispose the `Color`s is because they are either _system_ colors (which should live through the whole applications LC) or they are created outside of our control (in which case the creator should dispose them). Or did I get it wrong?

iloveeclipse:
> we shouldn't dispose the `Color`s is because they are either _system_ colors (which should live through the whole applications LC) or they are created outside of our control

Colors don't need to be disposed at all anymore (independently of context), they are not system resources anymore after SWT changes.

fedejeanne:
Indeed, thank you for the pointer!

- 359680f217565c4ee358bea50768cbec1b5cf402
- 9f2c5c8725c155f1ee5eecaae552e46aaa30e744

https://github.com/eclipse-platform/eclipse.platform.swt/blob/823aeba299457f36f3c3f63fefe099b8ca9d9e25/bundles/org.eclipse.swt/Eclipse%20SWT/win32/org/eclipse/swt/graphics/Color.java#L24-L26

fedejeanne (approved):

PR #3318, opened 2026-05-21
[Win32] Avoid unnecessary creation of GC operations 

### LINKED ISSUE #3296: OutOfMemory when disposing images due to many copies of image in GC.operations

HeikoKlare:
> [!IMPORTANT]
> This change requires and thus currently contains and needs to be merged after:
> - https://github.com/eclipse-platform/eclipse.platform.swt/pull/3317

The storage of executed operations in a GC has been added to support the situation that a Drawable is created with a GC without knowing the final zoom in which the Drawable is supposed to be used. A common example is the creation of a (buffer) image, which is later to be drawn on a control, but at the time of drawing the image that zoom is not known. The operations allow to recreate the image for that required zoom.

In most use cases of a GC, however, there is no need and no possibility to recreate the underlying drawable from a GC. In particular, when drawing to a control, the zoom of that control is known while drawing such that no storage of GC operations is necessary.

The storage of GC operations can become rather expensive, in particular if the drawing of images is involved and copies of those images need to be created upon their disposal. For that reason, this change reduces the cases in which operations are actually stored to those where they are really required for recreation of the underlying drawable.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/3296

### DIFF
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_GC.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_GC.java
@@ -487,8 +487,8 @@ public void test_drawImageLorg_eclipse_swt_graphics_ImageIIII_ImageDataAtSizePro
 		Exception e = assertThrows(IllegalArgumentException.class, () -> gc.drawImage(image, 0, 0, 1, 1));
 		assertSWTProblem("Incorrect exception thrown for provider == null", SWT.ERROR_INVALID_ARGUMENT, e);
 	} finally {
-		image.dispose();
 		drawToImage.dispose();
+		image.dispose();
 	}
 }
 
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
@@ -6002,8 +6002,11 @@ Point textExtentInPixels(String string, int flags) {
 	return new Point(rect.right, rect.bottom);
 }
 
-void refreshFor(Drawable drawable, ImageHandle imageHandle) {
+void reapplyTo(Drawable drawable, ImageHandle imageHandle) {
 	if (drawable == null) SWT.error(SWT.ERROR_NULL_ARGUMENT);
+	if (!data.reapplicable) {
+		SWT.error(SWT.ERROR_INVALID_ARGUMENT);
+	}
 	destroy();
 	GCData newData = new GCData();
 	originalData.copyTo(newData);
@@ -6109,9 +6112,13 @@ int getZoom() {
 }
 
 private void storeAndApplyOperationForExistingHandle(Operation operation) {
-	removePreviousOperationIfSupercededBy(operation);
-	operations.add(operation);
 	operation.apply();
+	if (data.reapplicable) {
+		removePreviousOperationIfSupercededBy(operation);
+		operations.add(operation);
+	} else {
+		operation.disposeAll();
+	}
 }
 
 private void removePreviousOperationIfSupercededBy(Operation operation) {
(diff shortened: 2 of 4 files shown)


tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_GC.java
```diff
@@ -487,8 +487,8 @@ public void test_drawImageLorg_eclipse_swt_graphics_ImageIIII_ImageDataAtSizePro
 		Exception e = assertThrows(IllegalArgumentException.class, () -> gc.drawImage(image, 0, 0, 1, 1));
 		assertSWTProblem("Incorrect exception thrown for provider == null", SWT.ERROR_INVALID_ARGUMENT, e);
 	} finally {
-		image.dispose();
 		drawToImage.dispose();
+		image.dispose();
```
fedejeanne, line 491: Why is this change necessary? Does this mean that production code also needs to pay attention to the order in which the images have been created when disposing them?

fedejeanne (commented):
I'm not too deep into the technical workings of this but I have some questions:

IIUC these changes mean that:
- There is now a mechanism to **not** allow storage of operations for later re-use (they can be expensive)
- This mechanism is only applied to `Image`s, since the check `if (!data.reapplicable) { ... }` is only done inside `org.eclipse.swt.graphics.GC.reapplyTo(Drawable, ImageHandle)`
- Of all the subclasses of `AbstractImageProviderWrapper`, only `PlainImageProviderWrapper` is allowed to store an reapply operations

Image

Is that correct?

Other than that, I don't see anything wrong with this PR and I consider it ready to be merged as soon as master opens again.

tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_GC.java
HeikoKlare: This is a test for an exception thrown during a `drawImage` call when a null provider is used. The resulting state of the GC after such a failure is undefined and potentially having it unnecessarily try to copy the image that was attempted to be drawn because of the disposal order is just a risk of the test failing because of an unrelated exception. In practice it does not make any difference as the code fails because of the failing draw operation anyway and probably no one wraps a `drawImage` call into a try-finally to clean up the involved resources where this order could make any difference at all.

HeikoKlare:
> IIUC these changes mean that:
> 
> * There is now a mechanism to **not** allow storage of operations for later re-use (they can be expensive)
> * This mechanism is only applied to `Image`s, since the check `if (!data.reapplicable) { ... }` is only done inside `org.eclipse.swt.graphics.GC.reapplyTo(Drawable, ImageHandle)`
> * Of all the subclasses of `AbstractImageProviderWrapper`, only `PlainImageProviderWrapper` is allowed to store an reapply operations
> 
> Image
>
> Is that correct?

Yes, that's correct. The `PlainImageProviderWrapper` is the only one ever having called `refreshFor` on the `GC`, which is why only for that wrapper the storage of operations is relevant.

PR #3321, opened 2026-05-23
[Cocoa] Fix in-place editor freeze by forcing TextKit 1

buchen:
On macOS the AppKit field editor SWT wraps can freeze during mouseDown: due to a TextKit 2 bug (_bellerophonTrack...). Accessing an NSTextView's layoutManager forces a permanent downgrade to TextKit 1 for that view.

Force this downgrade eagerly at field-editor promotion time, before any mouse interaction, rather than relying on the existing lazy layoutManager() calls in getCaretLocation() which run too late:

 - Shell.createHandle(): the per-window field editor (SWTEditorView), covering SWT.SINGLE Text and editable Combo.
 - Text.createWidget(): the separate secure field editor for SWT.PASSWORD (SWTSecureEditorView).

See eclipse-platform/eclipse.platform.ui#1069


buchen:
I am distributing my [RCP application](https://github.com/portfolio-performance/portfolio) with this patch applied. The patch is successful in the wild. Is there anything I can do or provide to help merging it?

PR #3322, opened 2026-05-25
ci: declare workflow-level `contents: read` on maven, pr-checks, version-increments

arpitjain099:
Adds workflow-level `permissions: contents: read` to three workflows that just run build / checks: `maven`, `pr-checks`, `version-increments`. No GitHub API writes.

Same post-CVE-2025-30066 (`tj-actions/changed-files`) hardening pattern. YAML validated locally.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -20,6 +20,9 @@ on:
     branches: [ master ]
     types: [opened, reopened, synchronize, labeled]
 
+permissions:
+  contents: read
+
 jobs:
   event_file:
     name: "Event File"
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -9,6 +9,9 @@ on:
   pull_request:
     branches: [ master ]
 
+permissions:
+  contents: read
+
 jobs:
   check-freeze-period:
     uses: eclipse-platform/eclipse.platform.releng.aggregator/.github/workflows/verifyFreezePeriod.yml@master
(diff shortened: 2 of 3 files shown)


akurtakov:
There is smth wrong with this PR: 
```
Invalid workflow file: .github/workflows/pr-checks.yml#L16
The workflow is not valid. .github/workflows/pr-checks.yml (Line: 16, Col: 3): Error calling workflow 'eclipse-platform/eclipse.platform.releng.aggregator/.github/workflows/verifyFreezePeriod.yml@master'. The workflow is requesting 'issues: read', but is only allowed 'issues: none'.
```
 from https://github.com/eclipse-platform/eclipse.platform.swt/actions/runs/26680432597

arpitjain099:
Thanks for catching that @akurtakov, and sorry for the breakage. You're right: the workflow-level `permissions: contents: read` set every other scope to `none`, which conflicts with the reusable `verifyFreezePeriod.yml` that requests `issues: read`.

I have moved `contents: read` down to the `check-javadoc-consistency` job (the only standalone job here that uses `GITHUB_TOKEN` for checkout) and removed the workflow-level block, so the three reusable callers run with their normal token context again. The other two files in this PR don't have this problem: `maven.yml`'s local `build.yml` declares no permissions, and `version-increments.yml`'s `publishVersionCheckResults.yml` declares `permissions: {}` and does its work via the bot PAT, so both stay within `contents: read`.

Workflow file should validate now.

PR #3323, opened 2026-05-25
[Gtk4] Fix Button image size by using GtkPicture instead of GtkImage

akurtakov:
Gtk 4 works in logical points while Cairo works in pixel thus this is needed for proper size of the pixbuf scale to the original image size.

This makes Button consistent with Label, which already uses GtkPicture for the same reason (see commit
https://github.com/eclipse-platform/eclipse.platform.swt/pull/2766 ).

Also fix a memory leak in Label where the GdkTexture created by gdk_texture_new_for_pixbuf() was not unreffed after being handed to gtk_picture_set_paintable().


akurtakov:
Before:
Screenshot From 2026-05-25 18-19-42
After:
Screenshot From 2026-05-25 18-20-32

jonahgraham (approved):

PR #3324, opened 2026-05-25
[Gtk4] Clip Composite children to parent bounds

akurtakov:
In GTK3, calling gtk_widget_set_has_window(true) on a SwtFixed widget created a backing GdkWindow that automatically clipped all child widgets to the parent's allocated rectangle.  GTK4 removed per-widget GdkWindows,
so children render freely beyond the parent by default (GTK_OVERFLOW_VISIBLE), causing custom-drawn widgets such as CTabFolder to let their content controls draw completely outside the folder bounds.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Composite.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Composite.java
@@ -344,8 +344,8 @@ void createHandle (int index, boolean fixed, boolean scrolled) {
 	if (GTK.GTK4) {
 		GTK4.gtk_widget_set_focusable(handle, true);
 		/*
-		 * Without GTK_OVERFLOW_HIDDEN children can
-		 * draw freely outside the parent, reproducing the GTK3 clipping behaviour.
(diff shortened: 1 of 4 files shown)


bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Composite.java
```diff
@@ -343,6 +343,11 @@ void createHandle (int index, boolean fixed, boolean scrolled) {
 
 	if (GTK.GTK4) {
 		GTK4.gtk_widget_set_focusable(handle, true);
+		/*
+		 * Without GTK_OVERFLOW_HIDDEN children can
+		 * draw freely outside the parent, reproducing the GTK3 clipping behaviour.
```
jonahgraham: nit - this wording confused me.


```suggestion
		 * Without GTK_OVERFLOW_HIDDEN, children can draw freely outside the parent.
		 * With GTK_OVERFLOW_HIDDEN we reproduce the GTK3 clipping behaviour.
```

jonahgraham (approved):
Approved (for when master reopens) - address the nit if you agree.

PR #3328, opened 2026-05-26
[Gtk4] Fix ToolItem image size by using GtkPicture instead of GtkImage

akurtakov:
GTK4 works in logical points while Cairo works in pixels, causing images to appear smaller at HiDPI scaling. Using GtkPicture (instead of GtkImage)
correctly maps the pixbuf to the right display size.

This is the same fix applied to Button
(https://github.com/eclipse-platform/eclipse.platform.swt/pull/3323 ) and Label
(https://github.com/eclipse-platform/eclipse.platform.swt/pull/2766 ) for the same reason.


jonahgraham (approved):

PR #3337, opened 2026-05-28
Control.gtk4_draw: do not send paint events if the control is zero-sized

tmssngr:
A control without width or height can't display anything.


akurtakov:
Can you please share a snippet that reproduces it ? It should be impossible to come up with such a call unless there is "big" problem in SWT Gtk internals and this would most likely hide a problem making it harder to figure it out.

tmssngr:
It was reproducible with your last week's second commit for #3331.

akurtakov:
Yeah I know that but that second commit was pretty incorrect (thus removed) so I believe this PR will only make it harder to debug other issues if landed in. Unless you have another way to reproduce it (which I would rather look into fixing proper) I would like to have this one closed without merging.

akurtakov:
I'm closing this one as I believe it will hide important problems if done that way.

tmssngr:
Maybe you could add an assertion instead there?

PR #3346, opened 2026-05-31
Remove deprecation warnings for SWTEventListener in deprecated methods

HeikoKlare:
The SWTEventListener interface is deprecated for removal. Some methods with parameters having the type of that interface are also marked as deprecated for removal but still show a warning because of the usage of that deprecated type. This change suppresses the unnecessary warning as the method is marked for removal consistently with the used type.

image


akurtakov:
Thanks for this one!

PR #3350, opened 2026-05-31
Promote CTabFolder.setItemOrder(int[]) to public API and add moveItem

vogella:
Promotes the existing package-private `CTabFolder.setItemOrder(int[])` to public API so consumers such as the Terminal view (eclipse-platform/eclipse.platform#2679) can implement drag-to-reorder without disposing and recreating tab items. The Javadoc now mirrors the `Table.setColumnOrder(int[])` contract so the semantics are unambiguous: `indices[i]` is the index of the item that should move to position `i`.

The previous implementation only updated `selectedIndex`, which left the internal `priority[]` array and `firstIndex` field pointing at stale positions and could corrupt the chevron and visible-tab logic. The reorder now builds an old-to-new lookup and remaps both fields so the scroll/chevron state stays correct after a reorder.

It also adds `moveItem(int from, int to)`, an ergonomic primitive for the common single-tab drag gesture: it moves one item and shifts the rest, implemented on top of `setItemOrder` so the index bookkeeping lives in one place. This avoids forcing consumers to construct a full permutation array just to move a single tab.

No symmetric `getItemOrder()` was added: unlike `Table`, a `CTabFolder`'s item array is itself the canonical order, so a getter would always return the identity permutation. Consumers can derive the order via `getItems()` and `indexOf(...)`. Added unit tests covering reordering, single-item moves, selection preservation, chevron behaviour with overflow, and the documented error cases.

### DIFF
--- a/binaries/org.eclipse.swt.cocoa.macosx.aarch64/META-INF/MANIFEST.MF
+++ b/binaries/org.eclipse.swt.cocoa.macosx.aarch64/META-INF/MANIFEST.MF
@@ -3,7 +3,7 @@ Fragment-Host: org.eclipse.swt;bundle-version="[3.128.0,4.0.0)"
 Bundle-Name: %fragmentName
 Bundle-Vendor: %providerName
 Bundle-SymbolicName: org.eclipse.swt.cocoa.macosx.aarch64; singleton:=true
-Bundle-Version: 3.134.100.qualifier
+Bundle-Version: 3.135.0.qualifier
 Bundle-ManifestVersion: 2
 Bundle-Localization: fragment
 Export-Package: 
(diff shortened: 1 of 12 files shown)


HeikoKlare (commented):
I would highly appreciate if we could have more time than like 1 business hour (for contributors seated in Europe) to have a look at a proposed change (in particular one affecting API) before it is merged.

We usually also try to have concrete use cases demonstrating the need for new API. While there is the abstract conceptual need mentioned for the terminal view, there is no concrete use case yet.
I wonder if the `setItemOrder` method is really needed. Isn't the more focussed `moveItem` method sufficient to keep the API lean?

vogella:
https://github.com/eclipse-platform/eclipse.platform/issues/2679 will be the use-case. setItemOrder is not needed as public APi but it felt consistent to release it which was present since a decade, I can change it back to a commented *public* if you prefer.

HeikoKlare:
In my opinion, the demonstrating use case should be an actual consumer at code level. At a high conceptual level it's difficult to assess if that specific API makes sense. In this case, from the conceptual view I would have said the API does not make sense for that case (as indicated by my comment) and you seem to agree with that.

> setItemOrder is not needed as public APi but it felt consistent to release it which was present since a decade

I am not in favor of making any method public as part of an API without any concrete need. Every new API method introduces additional maintenance burden and makes the framework more difficult to evolve.

vogella:
Okey, I will not keep moveItem as API to use it for terminal.

PR #3352, opened 2026-06-01
Make CTabFolder.setItemOrder package-private again, keep moveItem

vogella:
Following review feedback on #3350, `CTabFolder.setItemOrder(int[])` should not be exposed as public API. This demotes it back to a package-private helper and drops its `@since` tag, while keeping the public `moveItem(int, int)` method as the supported primitive for the common single-tab reorder (drag) gesture.

The `priority[]`/`firstIndex` remap fix from #3350 is retained, since `moveItem` delegates to `setItemOrder` and relies on it to keep the chevron and scroll state correct after a reorder. The unit tests that called `setItemOrder` directly are removed (they can no longer compile from the test package); the `moveItem` tests stay and continue to exercise the remap logic.


vogella:
The removed tests are due to the removal of the API. The fix in the incorrect handling of reordering in setItemOrder is of course left in the code and is indirectly tested via the moveItem method.

FYI - @HeikoKlare I plan to merge this today, to avoid having a SWT release with unwanted API.

HeikoKlare:
> The removed tests are due to the removal of the API.

Would it make sense to preserve the tests as unit tests instead of API tests?

vogella:
> > The removed tests are due to the removal of the API.
> 
> Would it make sense to preserve the tests as unit tests instead of API tests?

I would say no, we didn't have them before and the new moveItem API is well-tested.

PR #3353, opened 2026-06-01
[Win32] Correct lifecycle of handles created via GC.getClipping()

HeikoKlare:
The recent enhancement to avoid the unnecessary creation of GC operations

introduced a regression, which particularly manifests in broken rendering of CTabFolders. For an example in the Manifest editor, see the screenshots below.

The GC's getClipping() operation stores region handles representing the clipping region (potentially at different zooms) that is combined with the passed region object. The handles for that passed region object are created lazy, so that the clipping region handle needs to be stored until that point in time. Otherwise it will leads to failures when lazily creating the handle for the passed region later on. Currently, these handles are stored in the GC operations and in case the GC or its operations are disposed too early, the region filled via a GC.getClipping() call will fail to create a proper region handle.

This change moves the responsibility for managing the lifecycle of the stored clipping regions to the Region instance in which they are used. Copies of that region ensure via reference counting that only the last disposed Region instance consuming the clipping region will dispose the according handle. A regression test for the scenario that the GC is disposed before consuming the Region handle to which the clipping region is applied is added.

This is a forward fix as an alternative to the revert proposed in:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/3351

Without this change:
image

With this change:
image

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
@@ -19,6 +19,7 @@ import java.util.function.*;
 
 import org.eclipse.swt.*;
 import org.eclipse.swt.graphics.Image.*;
+import org.eclipse.swt.graphics.Region.*;
 import org.eclipse.swt.internal.*;
 import org.eclipse.swt.internal.gdip.*;
 import org.eclipse.swt.internal.win32.*;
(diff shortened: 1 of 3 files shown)


HeikoKlare:
Since the bug to be fixed by this PR severely degrades usability of Eclipse/SWT applications, I will merge this without keeping it open for two days for potential review, so that we have a usable I-Build state tomorrow. Potential follow-up issues should become easily noticable and could be worked on independently.

PR #3355, opened 2026-06-01
[Win32] Fix font metrics being scaled by GC transform for fallback fonts

### LINKED ISSUE #2978: [Win32] Missing fonts don't play nice with FontMetrics and Transform
**Describe the bug**

Unlike other fonts like e.g. Arial, the FontMetrics returned for the Monospace font are scaled by the Transform.

**To Reproduce**

```java
package test;

import org.eclipse.swt.SWT;
import org.eclipse.swt.graphics.Font;
import org.eclipse.swt.graphics.Transform;

public class Test {
	protected static final Font DISPLAY_FONT = new Font(null, "Monospaced", 24, SWT.NORMAL); //$NON-NLS-1$
		
		Shell shell = new Shell();

			GC gc = event.gc;
			gc.setFont(DISPLAY_FONT);
			
			Transform t = new Transform(gc.getDevice());
			t.scale(3.0f, 3.0f);

			gc.setTransform(t);
			System.out.println(gc.getFontMetrics().getAverageCharacterWidth());
			
			gc.setTransform(null);
			System.out.println(gc.getFontMetrics().getAverageCharacterWidth());

			gc.setFont(DISPLAY_FONT);
			System.out.println(gc.getFontMetrics().getAverageCharacterWidth());

			gc.setTransform(t);
			System.out.println(gc.getFontMetrics().getAverageCharacterWidth());
			
			gc.dispose();
		});

		shell.open();
		
		Display d = shell.getDisplay();
		while(!shell.isDisposed()) {
			while(!d.readAndDispatch()) {
				d.sleep();
			}
		}
	}
}
```

In GEF, I even had a situation where the correct values are returned after calling these no-op operations. Which only works due to calling `getFontMetrics()` while `gc.getTransform()` is `null`:
```java
gc.setTransform(null);
gc.getFontMetrics();
gc.setTransform(transform);
```

**Output**

- Linux (with Monospaced)
15, 15, 15, 15

- Windows (125% zoom, with Monospaced)
53, 53, 18, 18

- Windows (125% zoom, with Arial)
18, 18, 18, 18

**Expected behavior**

The average char width should behave the same across all operating systems.

**Environment:**
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 

**Version since**

Probably since forever. I can see the same inconsistency with the 2023-12.

HeikoKlare:
### Summary

Fixes #2978

When a font that does not exist on the platform is set on a `GC` and a `Transform` is active, GDI+ calls `Font::GetLogFontW` with the transformed `Graphics` to derive a fallback GDI font handle. This causes the `lfHeight` in the resulting `LOGFONT` to be scaled by the transform, so `GetTextMetrics` on the DC returns inflated (transform-scaled) metrics.

The fix temporarily saves and restores the GDI+ transform around the `Font::GetLogFontW` call so the `LOGFONT` always reflects device-pixel metrics, independent of any user-applied transform.

A regression test is added that verifies font metrics are consistent across all four combinations of transform-set/cleared and font-set/kept that the original issue demonstrated.

### How to Test

The snippet from https://github.com/eclipse-platform/eclipse.platform.swt/issues/2978 may be used for testing. That snippet reflects what is also done by the test added with this PR.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
@@ -681,7 +681,22 @@ static long createGdipFont(long hDC, long hFont, long graphics, long fontCollect
 		if (outFont != null && font != 0) {
 			long hHeap = OS.GetProcessHeap();
 			long pLogFont = OS.HeapAlloc(hHeap, OS.HEAP_ZERO_MEMORY, LOGFONT.sizeof);
-			Gdip.Font_GetLogFontW(font, graphics, pLogFont);
+			if (graphics != 0) {
+				// Font_GetLogFontW applies the graphics's world transform when converting the
+				// font size to LOGFONT units. A non-identity user transform would cause lfHeight
+				// to be scaled, producing an incorrectly-sized GDI font for GetTextMetrics.
+				// Temporarily set identity so the LOGFONT reflects device-pixel metrics.
+				long savedMatrix = Gdip.Matrix_new(0, 0, 0, 0, 0, 0);
+				Gdip.Graphics_GetTransform(graphics, savedMatrix);
+				long identityMatrix = Gdip.Matrix_new(1, 0, 0, 1, 0, 0);
+				Gdip.Graphics_SetTransform(graphics, identityMatrix);
+				Gdip.Font_GetLogFontW(font, graphics, pLogFont);
+				Gdip.Graphics_SetTransform(graphics, savedMatrix);
+				Gdip.Matrix_delete(identityMatrix);
+				Gdip.Matrix_delete(savedMatrix);
+			} else {
+				Gdip.Font_GetLogFontW(font, graphics, pLogFont);
+			}
 			outFont[0] = OS.CreateFontIndirect(pLogFont);
 			OS.HeapFree(hHeap, 0, pLogFont);
 		}
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_GC.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_GC.java
@@ -731,6 +731,45 @@ public void test_getFontMetrics() {
 	assertTrue(fm.getHeight() > 0);
 }
 
+/**
+ * Verifies that applying a transform to a GC does not affect the font metrics
+ * when the font does not exist on the current platform and a fallback font is
+ * used instead.
+ *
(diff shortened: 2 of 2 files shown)


HeikoKlare:
@ptziegler I am not sure how severe https://github.com/eclipse-platform/eclipse.platform.swt/issues/2978 still is for you, but this may be a solution to that issue

ptziegler:
@HeikoKlare I think we agreed that this problem primarily happens when using a non-existent font, right? The reason I originally reported this was due to the GEF Logic editor, but otherwise I've never seen it in practice. Especially when the "fix" is to use a valid font instead. :sweat_smile:

HeikoKlare:
Yes, this problem should _only_ happen when using a non-existent font (because the height calculation is wrong in exactly that case).
I remember that you reported the issue because of the numbers in the Logic editor, and I also agree that it should not happen that often in practice, as you should usually use/switch to a font that exists (on every OS you support) for proper UI results. I was just not sure if you may had faced that issue anywhere else as well 🙂

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -680,7 +680,22 @@ static long createGdipFont(long hDC, long hFont, long graphics, long fontCollect
...
 			long hHeap = OS.GetProcessHeap();
 			long pLogFont = OS.HeapAlloc(hHeap, OS.HEAP_ZERO_MEMORY, LOGFONT.sizeof);
-			Gdip.Font_GetLogFontW(font, graphics, pLogFont);
+			if (graphics != 0) {
+				// Font_GetLogFontW applies the graphics's world transform when converting the
+				// font size to LOGFONT units. A non-identity user transform would cause lfHeight
+				// to be scaled, producing an incorrectly-sized GDI font for GetTextMetrics.
+				// Temporarily set identity so the LOGFONT reflects device-pixel metrics.
+				long savedMatrix = Gdip.Matrix_new(1, 0, 0, 1, 0, 0);
+				Gdip.Graphics_GetTransform(graphics, savedMatrix);
```
r-mennig, line 690: Since `savedMatrix` is immediately overwritten by calling `Graphics_GetTransform(graphics, savedMatrix)`, initializing the variable with an empty matrix is easier to understand here (in my opinion).

r-mennig (approved):

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
HeikoKlare: That's a smart proposal. Currently it looks like this needs to be the identity matrix for some reason but actually the whole matrix object will be overwritten by the `GetTransform` call. I have updated the PR to just use zero values for the matrix initialization. I ran the snippet in the linked issue again and it still works as expected (just like the test added with this PR).

PR #3360, opened 2026-06-04
[Edge] Allow SSO with primary account by default

### LINKED ISSUE #2658: SWT upgrade to version 3.126.0: WebView2 no longer passes Device info to Azure
We are using a Java17-based desktop application that federates users via Azure Entra ID using the OAuth2 Authorization Code Flow. Additionally, we enforce Azure Conditional Access policies with Device Assurance enabled.

Recently, we upgraded the **SWT library to version 3.126.0**, which introduced WebView2 as the underlying browser component. Post-upgrade, we observed that Device ID and Join Type information are no longer present in the interactive user sign-in logs in Azure Entra ID.

Observed Behavior:

- Current WebView2 (Edge 140.0.0): Device ID and Join Type are missing in the sign-in logs.

Image

- Previous WebView (IE 7.0): Device ID and Join Type were visible in the sign-in logs.

Image

This change is impacting our Conditional Access policies that rely on device information for compliance and access control.

Please help us for the same.

HeikoKlare:
The WebView feature to allow SSO has been made accessible but disabled by default. With this change, the feature is enabled by default, still allowing to opt-out via system property. This aligns with the previous behavior of the IE engine.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/2658

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
@@ -611,7 +611,7 @@ WebViewEnvironment createEnvironment() {
 	String browserArgs = System.getProperty(BROWSER_ARGS_PROP);
 	String language = System.getProperty(LANGUAGE_PROP);
 
-	boolean allowSSO = Boolean.getBoolean(ALLOW_SINGLE_SIGN_ON_USING_OS_PRIMARY_ACCOUNT_PROP);
+	boolean allowSSO = !"false".equalsIgnoreCase(System.getProperty(ALLOW_SINGLE_SIGN_ON_USING_OS_PRIMARY_ACCOUNT_PROP));
 
 	String dataDir = getDataDir(display);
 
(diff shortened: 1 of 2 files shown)


HeikoKlare:
@sratz you proposed this in https://github.com/eclipse-platform/eclipse.platform.swt/issues/2658#issuecomment-3451060617
I don't see any risk in making this opt-out, despite that someone may want to disable that feature for security reasons and then has to be aware of the change. But we would of course share this via news. What do you think?

r-mennig (approved):

HeikoKlare:
@sratz sorry for pinging you again on this 😳 but in case we want to change this default, it would probably be good to do so before M1, so that we have sufficient time for feedback.

sratz (approved):

sratz:
Sorry, missed that first ping. Yes, makes total sense to enable this by default!

HeikoKlare:
Thank you for the feedback! We are still very good in time for M1, so everything is fine 🙂 
With the latest push, I have just added an update of the readme for the WebView2 runtime, which still explained the SSO feature as opt-in. I've changed it to describe the SSO feature as default with the ability to opt-out. I've not found any other references to the system property in documentation that may require an update yet.

HeikoKlare:
Proposed an N&N to make aware of that change: 
- https://github.com/eclipse-platform/www.eclipse.org-eclipse/pull/570

PR #3363, opened 2026-06-04
[Win32] Remove runtime change of process DPI awareness

HeikoKlare:
SWT initialization modifies the process DPI awareness via the static initializer of the OS class. This has been implemented for DPI awareness on Windows Vista long time ago (https://github.com/eclipse-platform/eclipse.platform.swt/commit/c9bc8bf1f3346d67fe47becef3a67f96a08bd20f). Changing process DPI awareness programmatically is discouraged and should rather be done by the application's manifest: https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setprocessdpiaware

In addition, relevant executables into which SWT is embedded (JVM, Equinox launcher etc.) already have state-of-the-art settings for DPI awareness, which even activate higher levels of DPI awareness than what the call to `setProcessDPIAware` (conforming to "System" DPI awareness) does. In consequence, this call has no benefit but may rather interfere with DPI awareness settings already done via the manifest.

This change thus removes the obsolete programmatic setting of process DPI awareness along with the now unused native method.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT PI/win32/library/os.c
+++ b/bundles/org.eclipse.swt/Eclipse SWT PI/win32/library/os.c
@@ -8772,18 +8772,6 @@ JNIEXPORT jint JNICALL OS_NATIVE(SetPolyFillMode)
 }
 #endif
 
-#ifndef NO_SetProcessDPIAware
-JNIEXPORT jboolean JNICALL OS_NATIVE(SetProcessDPIAware)
-	(JNIEnv *env, jclass that)
-{
-	jboolean rc = 0;
-	OS_NATIVE_ENTER(env, that, SetProcessDPIAware_FUNC);
-	rc = (jboolean)SetProcessDPIAware();
-	OS_NATIVE_EXIT(env, that, SetProcessDPIAware_FUNC);
-	return rc;
-}
-#endif
-
 #ifndef NO_SetProp
 JNIEXPORT jboolean JNICALL OS_NATIVE(SetProp)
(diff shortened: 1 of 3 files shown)


Phillipus:
> In consequence, this call has no benefit but may rather interfere with DPI awareness settings already done via the manifest.

Are there any known side-effects of this in the wild?

HeikoKlare:
> Are there any known side-effects of this in the wild?

No, I am not aware of any. From the API documentation, I would expect the method to "downgrade" the DPI awareness, but even though the method returns "true" (which means "success"), it does actually not change the DPI awareness (otherwise we would have already noticed at many places). The processes still run at PerMonitorV2, so it effectively seems to only upgrade if no DPI awareness is set.

tmssngr:
So it might be necessary to fix the launcher executable for all SWT based applications?

HeikoKlare:
> So it might be necessary to fix the launcher executable for all SWT based applications?

In case you have a custom executable that was configured in a way that setting process DPI awareness from within the process to "System" had any effect, then: in theory, yes.
But I don't see how one should have come into that situation. Conceptually, you would need to have configured a custom manifest for your executable that downgrades default DPI awareness (to something like "unaware") without wanting that, as the process itself then has upgraded that to "system" afterwards (with the statement removed by this PR). This does not sound like a realistic and reasonable scenario.
In addition, practically I seems like even this scenario is not possible to produce: I tried by using a JVM with DPI awareness set to "false"/"unaware" via the manifest and even then this setting was not overridden by the code removed by this PR. 

So in summary, I am not sure how the removed code can have had any effect at all. If you have a concrete scenario where the code did anything reasonable (and maybe was even necessary), please let me know.

PR #3371, opened 2026-06-08
Enable anti-aliasing for dirty indicator rendering

HeikoKlare:
The dirty indicator for editor tabs does currently not use anti-aliasing which makes it look quite jagged. This change enables anti-aliasing for the rendering of the dirty indicator.

Left is current state, right is with this change:
image

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolderRenderer.java
@@ -747,7 +747,10 @@ public class CTabFolderRenderer {
 		if (!selected) {
 			gc.setAlpha(140);
 		}
+		int originalAntialias = gc.getAntialias();
+		gc.setAntialias(SWT.ON);
 		gc.fillOval(x, y, diameter, diameter);
+		gc.setAntialias(originalAntialias);
 		gc.setAlpha(originalAlpha);
 		gc.setBackground(originalBackground);
 	}


merks:
It's great to see such attention to fine detail. 👀  Kudos!

vogella:
> The dirty indicator for editor tabs does currently not use anti-aliasing which makes it look quite jagged. This change enables anti-aliasing for the rendering of the dirty indicator.
> 
> Left is current state, right is with this change: image

Anything specific to configure to see this ugly indicator? Looks fine for me under Linux and Windows.

Phillipus:
> Anything specific to configure to see this ugly indicator? Looks fine for me under Linux and Windows.

I can see the jaggies on Windows at 200% display scale. A screenshot and zooming in reveals all.

vogella:
Change LGTM, thanks @HeikoKlare

Phillipus:
Side question - how does one enable the new indicator for editor parts in an RCP application?

Edit - found it.

```java
IEclipsePreferences prefs = InstanceScope.INSTANCE.getNode("org.eclipse.e4.ui.workbench.renderers.swt");
prefs.putBoolean("SHOW_DIRTY_INDICATOR_ON_TABS", true);
```

HeikoKlare:
Thank you for your feedback. We seem to agree on antialiasing the dirty indicator, thus merging this.

I wonder if we may want to do the same for the close button. In that case it does not seem to be that obvious, still I personally like the antialiased version more (left is non-antialiased, right is antialiased):
image
Maybe I will create a follow-up proposal for discussing that.

Phillipus:
> I wonder if we may want to do the same for the close button. In that case it does not seem to be that obvious, still I personally like the antialiased version more 

I find that whereas antialiasing is better on curves, on straight lines it can make things blurry. Just my personal 2 cents.

HeikoKlare:
> I find that whereas antialiasing is better on curves, on straight lines it can make things blurry. Just my personal 2 cents.

Yes, there will probably be some agreement on this for horizontal/vertical lines, and for others it may be a matter of personal preference.

PR #3381, opened 2026-06-12
[Win32] Replace deprecated URL constructor with URI in examples

HeikoKlare:
The URL(URL context, String spec) constructor used in OlePlugin.getImageFromPlugin() of the OLE examples project is deprecated since Java 20. This replaces it with an URI-based approach.

### DIFF
--- a/examples/org.eclipse.swt.examples.ole.win32/src/org/eclipse/swt/examples/ole/win32/OlePlugin.java
+++ b/examples/org.eclipse.swt.examples.ole.win32/src/org/eclipse/swt/examples/ole/win32/OlePlugin.java
@@ -172,8 +172,8 @@ public class OlePlugin extends AbstractUIPlugin {
 	 * @return the image, or null if not found
 	 */
 	private static Image getImageFromPlugin(Bundle bundle, String iconPath) {
-		URL installUrl = bundle.getEntry("/");
-		try (InputStream is = new URL(installUrl, iconPath).openConnection().getInputStream()) {
+		URL iconUrl = bundle.getEntry(iconPath);
+		try (InputStream is = iconUrl.openConnection().getInputStream()) {
 			ImageData source = new ImageData(is);
 			ImageData mask = source.getTransparencyMask();
 			Image image = new Image(null, source, mask);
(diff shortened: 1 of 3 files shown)


examples/org.eclipse.swt.examples.ole.win32/src/org/eclipse/swt/examples/ole/win32/OlePlugin.java
```diff
@@ -173,11 +173,16 @@ public static void freeResources() {
 	 */
 	private static Image getImageFromPlugin(Bundle bundle, String iconPath) {
 		URL installUrl = bundle.getEntry("/");
```
HannesWell: Would it be possible to pass the `iconPath` directly to `Bundle.getEntry()`?
Then the conversion could be omitted entirely.

Generally one has to take great care because Eclipse was/is quite lax when it comes to escaping special characters for URLs. And URLs accepted that, URIs don't.
But example are under our control, so this isn't an issue here, like in tests.

HannesWell: I tried this out and in fact you can just use the following:
```suggestion
		URL installUrl = bundle.getEntry(iconPath);
		try (InputStream is = installUrl.openConnection().getInputStream()) {
			ImageData source = new ImageData(is);
			ImageData mask = source.getTransparencyMask();
			return new Image(null, source, mask);
```

HannesWell (commented):
The URL/URI conversion can be entirely avoided.

But when trying this out, by launching an Eclipse SDK with the OLE examples bundle included, I generally wonder how useful and useable this example actually is.

HeikoKlare:
> I tried this out and in fact you can just use the following:

Nice proposal. It work great and I have adapted the PR accordingly.

> But when trying this out, by launching an Eclipse SDK with the OLE examples bundle included, I generally wonder how useful and useable this example actually is.

It allows you to easily test SWT's OLE integration in an isolated way. I don't see too much of a benefit in integrating it in an Eclipse SDK, but you can run this example standalone as well.

HeikoKlare:
And with this change we are down to zero compiler warnings in SWT:
image

HannesWell (approved):
> > I tried this out and in fact you can just use the following:
> 
> Nice proposal. It work great and I have adapted the PR accordingly.

Great, for for adapting.

> And with this change we are down to zero compiler warnings in SWT:

Oh that's awesome!

PR #3390, opened 2026-06-17
[GTK4] Implement DragSource content provider for drag-and-drop

akurtakov:
Implement DragSource.dragPrepare on GTK4 to build a real GdkContentProvider from the transfer data collected via DND.DragSetData, replacing the previous stub that returned no data. The DragStart event is
now fired from the GtkDragSource "prepare" signal.

Map drag operations to the GTK4 GdkDragAction values, which differ from the GTK3 values, via new GTK4.GDK_ACTION_* constants.


akurtakov:
This is not testable with SWT to SWT DnD operation as Gtk4 is not proper drop target still, one has to drop to native Gtk4 application e.g. gnome-text-editor. I have tested using Snippet78 as source and dropping into gnome-text-editor.

PR #3392, opened 2026-06-18
[Gtk4] Implement Shell with ON_TOP as GtkPopover

akurtakov:
This is needed as GtkWindows initial positions are entirely managed by the compositor thus positioning them next to each other is not possible. GtkPopover allows to manage this at the price of complicating focus handling.


akurtakov:
This is probably not perfect but IMO lays the foundations so one can have e.g. JDT's javadoc popup not be above the completion popup, CDT's launchbar popups are showing in the correct place and etc.

jonahgraham (approved):

PR #3394, opened 2026-06-19
Removes System.out.println debug code from ToolItem

jonahgraham:
PR #3257 accidentally included a system.out, this commit removes this debugging code.


akurtakov:
Thanks for catching this one!

PR #3395, opened 2026-06-19
Stop using ant-contrib

akurtakov:
Ant 1.9.1 added https://ant.apache.org/manual/ifunless.html which is the sole reason why ant-contrib was used.

### DIFF
--- a/binaries/pom.xml
+++ b/binaries/pom.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!-- 
 ###############################################################################
-# Copyright (c) 2024, 2024 Hannes Wellmann and others.
+# Copyright (c) 2024, 2026 Hannes Wellmann and others.
 # All rights reserved. This program and the accompanying materials
 # are made available under the terms of the Eclipse Public License v1.0
 # which accompanies this distribution, and is available at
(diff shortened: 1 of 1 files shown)


binaries/pom.xml
```diff
@@ -137,37 +137,34 @@
...
+											<and>
+												<isset property="swt.build.natives"/>
+												<equals arg1="${ws}" arg2="win32"/>
+											</and>
+										</condition>
+										<property name="build_dir" value="${project.build.directory}/natives-build-temp"/>
+										<exec if:set="swt.build.natives" executable="${java.home}/bin/java" dir="${swtMainProject}" failonerror="true">
+											<arg line="-Dws=${ws} -Darch=${arch} build-scripts/CollectSources.java -nativeSources '${build_dir}'"/>
+										</exec>
+										<property if:set="swt.build.natives" name="SWT_JAVA_HOME" value="${java.home}"/><!-- Not overwritten if already set, e.g. as CLI argument -->
```
HannesWell: Setting the `SWT_JAVA_HOME` property in case the platform is not build natively doesn't harm, so I'd keep this simpler and avoid the `if`:
```suggestion
										<property name="SWT_JAVA_HOME" value="${java.home}"/><!-- Not overwritten if already set, e.g. as CLI argument -->
```

HannesWell: Instead of `if:set`, wouldn't it be a little bit more readable and save to use `if:true` (even if it's slightly more lengthy as it requires a variable access)?
```suggestion
										<property if:true="${swt.build.natives}" name="SWT_JAVA_HOME" value="${java.home}"/><!-- Not overwritten if already set, e.g. as CLI argument -->
```
When I read the original, I was wondering if the condition might set the attribute to false, so I hope using `if:true` is foolproof?
If this is applied, it should be done for the other tests too.

binaries/pom.xml
```diff
@@ -221,17 +218,6 @@
...
-								<artifactId>ant-contrib</artifactId>
-								<version>1.0b3</version>
-								<exclusions>
-									<exclusion>
-										<groupId>ant</groupId>
-										<artifactId>ant</artifactId>
-									</exclusion>
-								</exclusions>
-							</dependency>
 						</dependencies>
```
HannesWell, line 235: The explicit dependency on ant should not be necessary anymore since `antrun` `3.2.0` already depends on ant `1.10.15`, which should be sufficient for this. So we could just remove the explicit dependency on `ant`.
I don't recall exactly anymore why that was added in the first place, but I assume it was related to override the dependency of `ant-contrib`.

HannesWell (commented):
That's awesome. Thanks for this!

I believe to vaguely remember that I've seen these attributes before but didn't make it work back then. But this makes the build noticeably simpler and avoids a dependency on a dependency last released over 15 years ago.

I just have three minor points below.

binaries/pom.xml
akurtakov: Thanks, applied.

akurtakov: The other proposal to set it unconditionally was applied.

binaries/pom.xml
akurtakov: Removed.

akurtakov:
I truly don't understand the new issues but they are totally not related to this change.

binaries/pom.xml
HannesWell: Both proposals were meant individually.
Just changed it as we spoke about privately.

HannesWell (approved):
Thanks again for this.

> I truly don't understand the new issues but they are totally not related to this change.

Maybe something changed in ECJ?
But I cannot reproduce it in the IDE, so maybe it's just something in the build. I don't think it's related.

PR #3397, opened 2026-06-21
[Gtk4] Fix blurry images on widgets at HiDPI

akurtakov:
On GTK4, GtkPicture measures a GdkTexture by its device-pixel intrinsic size, which inflated widgets at HiDPI.

Introduce a custom scale-aware Gtk4-only GdkPaintable (SwtScaledPaintable) that reports the logical intrinsic size while drawing the full-resolution texture, giving crisp images at the correct widget size.

Use that paintable into the GtkPicture consumers (Button, Label, ExpandItem, ToolItem) and remove the GTK4 downscale in ImageList.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT PI/gtk/library/os_custom.c
+++ b/bundles/org.eclipse.swt/Eclipse SWT PI/gtk/library/os_custom.c
@@ -2263,3 +2263,96 @@ JNIEXPORT jlong JNICALL OS_NATIVE(content_1providers_1create_1gvalue)
 	OS_NATIVE_EXIT(env, that, content_1providers_1create_1gvalue_FUNC)
 	return (jlong)v;
 }
+
+#if defined(GTK4)
+/*
+ * SwtScaledPaintable: a GdkPaintable wrapping a GdkTexture that advertises a
+ * fixed logical (point) intrinsic size. GtkPicture measures a plain GdkTexture
+ * using its device-pixel size, which inflates widgets at HiDPI. By reporting the
+ * logical size here while drawing the full-resolution texture, images keep their
+ * correct logical footprint and render crisply (no down/up-scaling).
+ */
+#define SWT_TYPE_SCALED_PAINTABLE (swt_scaled_paintable_get_type ())
+G_DECLARE_FINAL_TYPE (SwtScaledPaintable, swt_scaled_paintable, SWT, SCALED_PAINTABLE, GObject)
+
+struct _SwtScaledPaintable
+{
+	GObject parent_instance;
+	GdkTexture* texture;
+	int width;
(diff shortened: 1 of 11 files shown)


bundles/org.eclipse.swt/Eclipse SWT PI/gtk/library/os_custom.c
```diff
@@ -2263,3 +2263,96 @@ JNIEXPORT jlong JNICALL OS_NATIVE(content_1providers_1create_1gvalue)
...
+static double swt_scaled_paintable_get_intrinsic_aspect_ratio (GdkPaintable* paintable)
+{
+	SwtScaledPaintable* self = SWT_SCALED_PAINTABLE (paintable);
+	if (self->height <= 0) return 0.0;
+	return (double) self->width / (double) self->height;
+}
+
+static GdkPaintableFlags swt_scaled_paintable_get_flags (GdkPaintable* paintable)
+{
+	return GDK_PAINTABLE_STATIC_SIZE | GDK_PAINTABLE_STATIC_CONTENTS;
```
jonahgraham, line 2319: Can the `gdk_paintable_snapshot (GDK_PAINTABLE (self->texture)...` ever paint differently or need to invalidate? i.e. is the texture always static contents too?

Ref is this demo from GTK4 which checks the child paintable for static

https://gitlab.gnome.org/GNOME/gtk/-/blob/eddff506651609ff02c641066f51f1c32f1b2fd6/demos/gtk-demo/paintable_emblem.c#L115-121

I am not sure this is wrong as we tend to have the paintables (texture) be rather static and recreated when changed.

jonahgraham (approved):
It looks ok to me. I have an open question, which I think isn't relevant, but I thought best to check.

bundles/org.eclipse.swt/Eclipse SWT PI/gtk/library/os_custom.c
akurtakov: It is always static as GdkTexture is wrapped (https://docs.gtk.org/gdk4/class.Texture.html)  and it's clearly stated in its documentation "It is primarily meant for pixel data that will not change over multiple frames, and will be used for a long time." . What's in the demo is due to it handling other types of GdkPaintable e.g https://gnome.pages.gitlab.gnome.org/gdk-pixbuf/gdk-pixbuf/class.PixbufAnimation.html or https://docs.gtk.org/gtk4/class.MediaStream.html but it's not our case.

akurtakov: I'll merge this one now so tomorrow's build has this fix with which I believe I can start using GTK 4 as my daily driver.

PR #3399, opened 2026-06-22
Edge: Ignore multiple gotFocus events if we were the one who caused them

sratz:
Follow-up on #1849 / #1848.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
@@ -87,7 +87,7 @@ class Edge extends WebBrowser {
 	HashMap<Long, LocationEvent> navigations = new HashMap<>();
 	/** Maps BrowserFunction index to the script ID from AddScriptToExecuteOnDocumentCreated. */
 	private final Map<Integer, String> functionScriptIds = new HashMap<>();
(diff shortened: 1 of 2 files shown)


HeikoKlare (approved):
The change looks sound to me. We had already shortly discussed it in https://github.com/eclipse-platform/eclipse.platform.swt/pull/1849 and just didn't do it without knowing about a concrete case that needs it to keep complexity and risk as low as possible.
I also checked again that `ignoreGotFocus` is only accessed from the UI thread, such that no memory consistency mechanisms need to be applied.

Just a nit question: shouldn't the test succeed on every browser (not only Edge), so that removing the assumption may be reasonable?

sratz:
> Just a nit question: shouldn't the test succeed on every browser (not only Edge), so that removing the assumption may be reasonable?

In principle, yes. But timing is slightly different. Also, for some reason on Linux/WebKit I am receiving 2 Events for each Browser.setFocus().

Not sure what exactly is going on there, haven't looked into this further. It's probably fine and the general assumption that "1 SWT.FocusIn event per Browser.setFocus()" just doesn't hold generally.

So only for the Edge case we know that should be the case and can attribute a mismatch to a regression for this particular issue.

I propose to proceed with having the test flagged Edge-only, WDYT?

HeikoKlare:
> I propose to proceed with having the test flagged Edge-only, WDYT?

Sounds good, thank you for the explanations. Since the behavior is so different with the other browsers, it does not make sense to extend the test to them. I am fine with merging this as is.

Issue #3398, opened 2026-06-22
SVG icons with fill="currentColor" render as white due to missing default foreground color in JSVGRasterizer


joeljohansson99:
**Describe the bug**
When rendering SVGs with property fill="currentColor", it will take the foreground color from the graphics. And since JSVGRasterizer creates the graphics with BufferedImage.createGraphics(), and dont set the default foreground color, it will take it from its default which is always white. This causes all SVG icons which has fill="currentColor" to be white. 

**To Reproduce**
Use an SVG with fill="currentColor".

**Expected behavior**
I would expect it to be black for light mode and white for dark mode. You can do this with g.setColor(color) after creating the graphics. (It wont override the fill="color") property when its specified, just the default when fill="currentColor", I have tested it already. 

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [x] All OS
- - [ ] Windows
- - [ ] Linux
- - [ ] macOS 

2. Additional OS info (e.g. OS version, Linux Desktop, etc)

3. JRE/JDK version
 25 

**Version since**
 Eclipse 4.40

**Workaround (or) Additional context**
This issue was introduced with this commit in the underlying Java SVG renderer: https://github.com/weisJ/jsvg/commit/55c481e2e8067a4a31f19ed3ddefc5b5e546248a.

PR #3400, opened 2026-06-22
Set default SVG currentColor to black

DLTKJoelJohansson:
An attempt at fixing #3398.
Let me know what you think.

Thanks!

### DIFF
--- a/bundles/org.eclipse.swt.svg/src/org/eclipse/swt/svg/JSVGRasterizer.java
+++ b/bundles/org.eclipse.swt.svg/src/org/eclipse/swt/svg/JSVGRasterizer.java
@@ -31,6 +31,7 @@ import static java.awt.RenderingHints.VALUE_RENDER_QUALITY;
 import static java.awt.RenderingHints.VALUE_STROKE_PURE;
 import static java.awt.RenderingHints.VALUE_TEXT_ANTIALIAS_ON;
 
+import java.awt.Color;
 import java.awt.Graphics2D;
 import java.awt.RenderingHints.Key;
 import java.awt.image.BufferedImage;
@@ -139,6 +140,7 @@ public class JSVGRasterizer implements SVGRasterizer {
 	private Graphics2D configureRenderingOptions(float widthScalingFactor, float heightScalingFactor,
 			BufferedImage image) {
 		Graphics2D g = image.createGraphics();
+		g.setColor(Color.BLACK);
 		g.setRenderingHints(RENDERING_HINTS);
 		g.scale(widthScalingFactor, heightScalingFactor);
 		return g;
--- a/bundles/org.eclipse.swt.svg/test/org/eclipse/swt/svg/JSVGRasterizerTest.java
+++ b/bundles/org.eclipse.swt.svg/test/org/eclipse/swt/svg/JSVGRasterizerTest.java
@@ -22,6 +22,7 @@ import java.nio.charset.StandardCharsets;
 import org.eclipse.swt.SWT;
 import org.eclipse.swt.SWTException;
 import org.eclipse.swt.graphics.ImageData;
+import org.eclipse.swt.graphics.RGB;
 import org.junit.jupiter.api.Test;
 
 class JSVGRasterizerTest {
@@ -29,7 +30,7 @@ class JSVGRasterizerTest {
 	private final JSVGRasterizer rasterizer = new JSVGRasterizer();
 	private String svgString = """
 			<svg viewBox="0 0 100 100" xmlns="http://www.w3.org/2000/svg">
-			    <rect width="100%" height="100%"/>
+			    <rect width="100%" height="100%" fill="currentColor"/>
 			</svg>
 			""";
 
@@ -89,4 +90,10 @@ class JSVGRasterizerTest {
 		assertEquals(SWT.ERROR_INVALID_IMAGE, exception.code);
 	}
 
+	@Test
+	void testDefaultCurrentColorIsBlack() {
+		ImageData data = rasterizer.rasterizeSVG(svgStream(svgString), 100);
+		RGB rgb = data.palette.getRGB(data.getPixel(0, 0));
+		assertEquals(new RGB(0, 0, 0), rgb);
+	}
 }


DLTKJoelJohansson:
Cant get the build run locally, and jenkins reporting build errors, maybe someone can take a look? Thanks

merks:
Don't do merge commits; we can't merge this into master in the end because we don't accept merge commits.   In fact don't do multiple commits in general.  Use rebase.  Amend you commit and force push it when you need to make subsequent follow up changes.  You'll need to rebase and force push to repair this PR.

merks:
I'll approve the workflow so you can see any errors in the GitHub actions.  Please rebase and force push to get this back on track.

merks:
See we even fail with a check for merge commits:

image

merks:
I've approved the workflow again.  I'll be out of office for a few hours...

DLTKJoelJohansson:
Cant seem to figure it out, its failing for Linux builds

DLTKJoelJohansson:
Okay got it to work now! Moved out the Display.getDefault() call to org.eclipse.swt bundle (In SVGFileFormat, the caller) fixed it. 

Thanks @merks for your help. Let me know what you think

DLTKJoelJohansson:
Maybe not, seems like the Browser tests for Windows are failing. Pretty strange

DLTKJoelJohansson:
@merks any thoughts on this? I have a hard time seeing how these changes can cause the Browser tests to fail on Windows, since the changes only touch the SVG Rasterization

merks:
It could just be a flaky test.  I'll rerun it.

DLTKJoelJohansson:
Looks like the same Browser tests are failing again, but at least the build succeeded now

merks:
@HeikoKlare 

I've not looked closely, but is the failure of the browser test on Windows related to this PR or maybe known to be flaky?

joeljohansson99:
Thanks @HeikoKlare, and @merks! This PR should ready for review now

merks:
@HeikoKlare 

I will leave it to someone qualified to review the changes.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/image/SVGFileFormat.java
```diff
@@ -73,8 +74,15 @@ ImageData loadFromByteStreamBySize(int width, int height) {
...
-		return rasterizedImageData;
+		return RASTERIZER.rasterizeSVG(inputStream, width, height, getWidgetForegroundColor());
+	}
+
+	private static RGB getWidgetForegroundColor() {
+		Display display = Display.getDefault();
+		if (display != null) {
+			return display.getSystemColor(SWT.COLOR_WIDGET_FOREGROUND).getRGB();
+		}
+		return null;
```
HeikoKlare: I am not sure if it is correct to use the system color for widget foreground here. This makes the assumption that this color is always wanted as `currentColor` for every image to be rasterized, but that may depend on the context, doesn't it? As an example, when using the Eclipse CSS engine, the system foreground color of widgets is overwritten (at every single widget, not at the display's system color), so the used color will not be fitting.

So this change (or rather the change to JSVG) seems to require some universal semantics for SVG's `currentColor`. In my understanding, this will depend on the context, so you would need to extend the image loading mechanism (like the `ImageLoader`/`ImageDataLoader`) with this information. Note that the `Image` class already has a `setBackground()` method, which might be a semantic fit for the `currentColor`. But adapting that consistently may be a bit of effort.

Couldn't it be a simpler and maybe sufficient solution to just use a fully transparent color and have everything else done by the context in which the image is used? Or isn't that possible because the this will give you the context's background for the parts to fill in the image whereas you want to have (widget) foreground?

HeikoKlare (commented):
Thank you for proposing a solution to the mentioned issue. I have some concerns regarding the proposed solution (please see the detailed comment). What could also help is an example (e.g., a snippet or an RCP extension) that demonstrates the behavior and the change, so that we can easily test different alternatives.

bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/internal/image/SVGFileFormat.java
DLTKJoelJohansson: Thanks for the review and also the explanation. It makes sense and I agree then that the `SWT.COLOR_WIDGET_FOREGROUND` is the wrong color to use. 

To be honest I am not very knowledgable about the graphic stack in Eclipse. But yes I think the idea is that the `currentColor` should inherit the text/foreground color of the surrounding context. So you are probably right that this information should be provided in some other way.

However I am bit unsure about the transparent color, would it not make the SVG pretty much invisible? What I would propose instead is if we want a sufficient simple solution is just setting the graphics foreground color to black (rather than having the default white). Because before this change: https://github.com/weisJ/jsvg/commit/55c481e2e8067a4a31f19ed3ddefc5b5e546248a it was defaulting to black in the renderer, and it did not seem to cause any problems, I think..?

If you agree then I can do that, and then maybe someone else can take a look at extending the image loading mechanism, and seeing if that would make sense.

HeikoKlare: > However I am bit unsure about the transparent color, would it not make the SVG pretty much invisible? 

You are probably right. I just understood that I probably slightly misunderstood the issue (as I thought the color was taken to fill a background, but it's used for drawing the foreground).

So regarding being specific to the usage context: you would probably need to extend the `Image` class to accept some foreground color that may be used (as a hint) for the image (in case the underlying provider such as SVG supports it). This will still be quite difficult (and maybe even possible to support properly) as it will not be sufficient to set some foreground color to an image but rather to retrieve handles for specific foreground colors (just like for specific zooms) because an image may be used in multiple contexts in which different foreground colors are used. And then the number of handles to manage inside an image increases.

Just setting the used color to black to restore the previous behavior sounds like a good proposal to me. As you said, no one complained about that behavior, so it seems to be sufficient for now. I guess that no so many users have SVGs with `currentColor` but rather use fixed colors, which keeps them unaffected by the change to JSVG.
So yes, it would be great if you can propose that change. And it would also be good to add a regression test for the JSVGRasterizer for this scenario so that we can be sure that we become aware of such changes eariler in future.

DLTKJoelJohansson:
Thanks @HeikoKlare for the feedback. I made the change so now we simply set the default foreground color to black, reverting back to the behaviour it was before the change in the underlying renderer.

I also added a regression test as you proposed. I am just checking one pixel seeing if its black. I guess you could test all pixels but I kinda felt it was overkill. Maybe there is an easier way to test this, but I could not find any.

Jenkins are failing too, not sure why. From the info I can see the compiler is saying `Unnecessary @SuppressWarnings("restriction")` which is very wierd since its not something I added. Any ideas what could cause this?

DLTKJoelJohansson:
@HeikoKlare, any thoughts on this? It might not be the compiler that is failing the build, but I cannot see any details

HeikoKlare (approved):
Thank you for updating the fix with this simple restoration of the previous behavior. We identified just today that our product is also affected by this issue, so I directly tested the proposed fix with it. I can confirm that is properly restores the previous behavior.

The test also looks sound to me. Just testing a single pixel is absolutely sufficient.

The reported warnings in the failing Jenkins check are strange. They are obviously not introduced by this PR. I have restarted the Jenkins build. Let's see if the next run produces proper results.

DLTKJoelJohansson:
Thank you, great! I see that the build is still giving the same results

HeikoKlare:
I have retriggered an SWT  `master` build and we see the same warnings reported there: https://ci.eclipse.org/releng/job/eclipse.platform.swt/job/master/1937/

image

Only difference I see between the previous and the current build is a newer ECJ version, but I did not find any suspicious changes at ECJ.

Anyway, the new warnings are not related to this PR and there is no other degradation reported in the build, so I will merge it as is.

PR #3401, opened 2026-06-24
[GTK] Speed up Combo setItems/removeAll/remove for large item counts

### LINKED ISSUE #506: setting/removing large number of (more than 5000) combo items is too slow
**Description**
org.eclipse.swt.internal.gtk.GTK.gtk_combo_box_text_remove_all\[native]() and  org.eclipse.swt.internal.gtk.GTK.gtk_combo_box_text_insert\[native]() are taking so long, see the screenshot below for profiler result.

**OS:** ubuntu 22.04
**JDK** Java 17

Screenshot from 2022-12-15 17-22-49

Screenshot from 2022-12-15 17-22-23

vogella:
Setting or clearing a large number of `Combo` items (>5000) is very slow on GTK.
Editing the `GtkListStore` a combo is showing costs O(n) per row, because the popup keeps its own handlers on the model and rebuilds an item for every `row-inserted`/`row-changed`, so filling or clearing a large combo is quadratic.

Detaching the model with `gtk_combo_box_set_model(handle, 0)` does not fix this, since the popup keeps the model even when the combo drops it.
Instead `setItems`, `removeAll` and `remove(start, end)` now build a new `GtkListStore`, fill it while nothing observes it, and hand it to the combo in one step, so filling is linear and the popup is rebuilt exactly once.

Measured on GTK3 with 16000 items: `setItems` 41.9s to 0.4s, removing half the items 12.2s to 0.5s, `removeAll` 2.1s to 0.1s.

As a side effect the combo no longer sends the spurious `Modify` events that the per-row insert produced (one per item), because the model swap happens with the `CHANGED` closure blocked.
Two new `Combo` tests cover the selection handling across range removals and the absence of those events.

Verified locally on Linux against self-built natives, on both GTK3 and GTK4: all 143 `Test_org_eclipse_swt_widgets_Combo` tests pass on both.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/506


vogella:
Reworked and force-pushed.

The three review comments about the `CHANGED` closure were right, and following them up showed the original approach did not actually help: detaching the model from the combo leaves it attached to the popup, which is what costs O(n) per row. Benchmarked on GTK3 at 16000 items, the detach version was no faster than master (47.2s vs 41.9s for `setItems`).

The bulk operations now build a fresh `GtkListStore` and swap it in once, with `CHANGED` blocked for the swap. That is linear: `setItems` 41.9s to 0.4s at 16000 items, and it also removes the per-item spurious `Modify` events, which a new test now covers.

vogella:
Both new review points applied.

**Blocking CHANGED on `entryHandle` too** was correct, and it turned out to matter on GTK4 specifically. On GTK3 `gtk_entry_set_text` short-circuits when the text is unchanged, so restoring the selection after the model swap was already silent. GTK4 goes through the entry buffer instead and rewrites the text unconditionally, so `remove(start, end)` on an editable combo fired 2 spurious `Modify` events even though the shown text never changed. `setModelItems` now blocks CHANGED, INSERT_TEXT and DELETE_TEXT on the entry for the swap, matching what `setText` already does.

**Extending the test to editable combos** was the right call too: the read-only-only version did not catch the above. Both event tests now run for `SWT.READ_ONLY` and `SWT.DROP_DOWN`, and there is a second test for the case that actually exposed it, a range removal that keeps the selected item. With the entry blocking reverted it fails on GTK4 with `style 4 ==> expected: <0> but was: <2>`.

Verified locally against self-built natives: all 144 `Test_org_eclipse_swt_widgets_Combo` tests pass on GTK3 and GTK4. Performance is unchanged by the fix (16000 items: `setItems` 0.44s, remove half 0.54s, `removeAll` 0.12s).

akurtakov:
New test test_removeII_keepsSelectedItemWithoutEvents fails on MacOS so as a minimum it should not be run on MacOS.

vogella:
Both review points applied and force-pushed.

@akurtakov right, `test_removeII_keepsSelectedItemWithoutEvents` is now skipped on macOS via `assumeFalse(SwtTestUtil.isCocoa, ...)`. The failure was `style 4 ==> expected: <0> but was: <1>`, so an editable Combo on Cocoa sends one Selection event when items before the selection are removed. The remaining assertions in that test were GTK-only anyway, so nothing is lost by skipping it there. The assume message records the deviation.

Copilot's point about the RTL popup direction was also correct: the model swap rebuilds the popup, so the direction previously applied to its children is lost, and only `setItems` was reapplying it. `removeAll()` and `remove(start, end)` now go through `setModelItems` too, so they had the same gap. The reapplication moved into `setModelItems`, which covers all three bulk paths and removes the duplication.

akurtakov:
Would you please rebase this one on master?

akurtakov:
I have uncovered pretty big flaw in this PR. If you run:
```java
    Combo combo = new Combo(shell, SWT.DROP_DOWN);
    combo.setItems("a", "b", "cat");
    combo.addVerifyListener(e -> e.text = e.text.toUpperCase());
    combo.select(2);
    System.out.println("after select:  " + combo.getText());
    combo.remove(0, 1);
    System.out.println("after remove:  " + combo.getText());
    display.dispose();
  }
```
the output will be 
```
after select:  CAT
after remove:  cat
```
showing the nasty and totally unwanted effect. The cause is the selection restore in setModelItems: gtk_combo_box_set_active makes GTK rewrite the entry from the model row while the entry's CHANGED, INSERT_TEXT and DELETE_TEXT closures are blocked, so the Verify listener never runs and the transformed text is lost.

akurtakov:
@vogella Do you plan to look further into this one?

vogella:
Yes

PR #3402, opened 2026-06-26
[Gtk4] Make resizeCalculationsGTK3 noop on Gtk 4

akurtakov:
On GTK4 the widget is sized via its own
gtk_widget_size_allocate(), which clamps to the minimum size internally, so this adjustment is unnecessary. Worse, when a widget reports a large (possibly stale, pre-realization) minimum size, inflating the allocation to that size leaves an oversized widget that paints as a blank rectangle over the content on first show until an interactive resize re-measures it (e.g. a GtkScrolledWindow whose content's natural height exceeds the requested height). Return the requested size unchanged on GTK4.


akurtakov:
Fixes choose workspace dialog when started via Equinox launcher.
Before:
Screenshot From 2026-06-26 16-01-17
After:
Screenshot From 2026-06-26 15-59-43

PR #3403, opened 2026-06-26
[Gtk4] Stop undersizing toolbar buttons

akurtakov:
With all the recent fixes this is no longer needed and it even makes toolbar look weird (one might say - slightly broken). Remove the custom css instea.


akurtakov:
Before:
Screenshot From 2026-06-26 16-06-11
After:
Screenshot From 2026-06-26 16-08-53

PR #3406, opened 2026-06-29
[GTK] Fix empty ImageData list for single-frame GIFs under glycin

### LINKED ISSUE #3404: Image loading with scaling different than 100% causes crash of CDT (GTK Linux)
There are actions using CDT that cause a null pointer exception when window scaling is other than 100%.
A developer of CDT attributes this to a problem with image loading using SWT.

!ENTRY org.eclipse.jface 4 2 2026-06-26 18:35:49.404
!MESSAGE Problems occurred when invoking code from plug-in: "org.eclipse.jface".
!STACK 0
java.lang.ArrayIndexOutOfBoundsException: Index 0 out of bounds for length 0
	at java.base/java.util.ImmutableCollections$ListN.get(ImmutableCollections.java:687)
	at org.eclipse.jface.resource.URLImageDescriptor.loadImageFromStream(URLImageDescriptor.java:151)
	at org.eclipse.jface.resource.URLImageDescriptor.getImageData(URLImageDescriptor.java:136)
	at org.eclipse.jface.resource.URLImageDescriptor.getImageData(URLImageDescriptor.java:108)

This may be specifically for Linux/GTK which uses gdk-pixbuf. The potentially significant parameters that trigger this issue appear to be:

    Xubuntu 26.04 & Fedora 44
    Wayland
    Scaling not 100%
    GIF image (image file provided by CDT)

> If you are reporting a **crash on Linux**, please check if `webkit` is mentioned in the crash file.
> If yes, please try to set `WEBKIT_DISABLE_COMPOSITING_MODE=1` and/or `WEBKIT_DISABLE_DMABUF_RENDERER=1` environment variables.

This makes no difference.

Please see https://github.com/eclipse-cdt/cdt/issues/1481#issuecomment-4817266046

akurtakov:
NativeImageLoader.load() branches on
gdk_pixbuf_animation_is_static_image():
static images take a fast path that always yields one frame, while non-static animations are decoded by iterating frames and only adding a frame when gdk_pixbuf_animation_iter_advance() reports an update.

As of gdk-pixbuf 2.44 (GNOME 49, e.g. Fedora 44), gdk-pixbuf delegates decoding to glycin (gdk-pixbuf MR !197 drops all loaders except gdi and glycin; NEWS: 2.44.0 "glycin: Implement animation"). The old in-tree GIF loader reported a single-frame GIF as a static image, so SWT took the fast path. glycin instead reports a single-frame GIF that carries a graphic control extension (frame delay/disposal) as a *non-static* animation. For such an image the very first iter_advance() returns false,
so the frame loop breaks immediately having added nothing and load() returns an empty list.

JFace's URLImageDescriptor.loadImageFromStream() calls this and does .get(0) on the result, so the empty list surfaces as an ArrayIndexOutOfBoundsException (not an SWTException, so callers that only
catch SWT.ERROR_INVALID_IMAGE do not handle it). This crashes consumers such as CDT, whose CWizardHandler loads tc_empty.gif from a static initializer, turning the failure into NoClassDefFoundError for MBSWizardHandler.

Make the loader resilient to however the decoding backend classifies frames: if the animation loop produces no frame, fall back to the animation's static image so a valid ImageData is always returned.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/3404


akurtakov:
I'm merging this one so reporters can test it in I-build.

PR #3408, opened 2026-06-29
CTabFolder overlap calculations change

akurtakov:
Update chevron image before deciding visible tabs so size reservations are correct.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
@@ -2678,7 +2678,7 @@ int getChevronCount() {
 private void updateChevronImage(boolean styleChange) {
 	if (styleChange && chevronImage == null) return;
 	int newCount = getChevronCount();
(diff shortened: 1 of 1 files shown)


akurtakov:
Split from https://github.com/eclipse-platform/eclipse.platform.swt/pull/3403 for wider/independent testing.

HeikoKlare (commented):
I've tested this on Windows and I don't find anything that breaks with it. I've used both the `CustomControlExample` for testing `CTabFolder` as well as an SDK with differently layouted part stacks (using `CTabFolders`), some with the chevron, some without, also resizing so that it swaps between the two states.

I don't see anything that becomes "better" on Windows, but if I understand correctly that's the intent, as the fix is supposed to improve behavior on Linux, isnt it?

akurtakov:
What I traced is that setItemLocation() decides how many tabs fit, reserving space for the chevron via getRightItemEdge(). But the chevron's image (hence its real width) is only set later, in setButtonBounds()->updateChevronImage() leading to tab visibility being computed against a image-less chevron, then the chevron is drawn in its real width and placed right after the last tab causing overlap. I believe that the this PR makes CTabFolder items positioning more stable and less fragile.

akurtakov:
I'll merge so it gets wider testing.

PR #3409, opened 2026-06-30
[Gtk4] Fix missing space between TreeItem icon and label

akurtakov:
Measuring the allocation of GtkSeparator before it's realized always gives 0x0 so icon and label are really glued.
Use gtk_cell_renderer_get_padding to query the cell renderer for that info. Gtk 3 could use same code but I decided to not do it as separator has bigger width than cellrenderer padding and I don't want to change Gtk 3 visual behavior (make trees more "condensed").


akurtakov:
Before:
Screenshot From 2026-06-30 14-44-08
After:
Screenshot From 2026-06-30 14-43-39

PR #3412, opened 2026-07-02
[GTK4] Implement DropTarget async drop handling

akurtakov:
Implement GTK4 DropTarget support using GtkDropTargetAsync with accept, enter/motion/leave and drop handlers.

Add GTK4 drop data flow integration by matching offered GTypes, reading drop values asynchronously via gdk_drop_read_value_async/finish, and finishing drops with the selected action.

Tested with DNDExample.


akurtakov:
DnD of CTabs in Eclipse IDE is still not working with this PR , as it's big enough and lays the foundations for drop support on Gtk 4 I would like to see it in before digging further.

PR #3418, opened 2026-07-07
[Gtk4] Fix displaying file permissions

akurtakov:
On GTK4 widgets have no per-widget GdkWindow, so paint order is simply the parent's child-list order (first = bottom, last = top) thus the Gtk3 hack calling moveAbove from moveBelow is not needed.


akurtakov:
Before:
Screenshot From 2026-07-07 17-45-52
After:
Screenshot From 2026-07-07 17-44-18

PR #3431, opened 2026-07-09
[Win32] Improve size calculation for cropped and scaled image drawing

### LINKED ISSUE #1146: IllegalArgumentException using ImagePrintFigureOperation and GC.drawImage
I've opened this new issue for the problem I encountered using `ImagePrintFigureOperation` and the Nebula `Gallery` in https://github.com/eclipse-gef/gef-classic/pull/1001.

@ptziegler said:

> As long as there isn't a reproducer, I consider this to be a Nebula bug.

This Snippet reproduces the exception using SWT only. Run on Windows:

```java
package gef;

import org.eclipse.draw2d.Figure;
import org.eclipse.draw2d.ImagePrintFigureOperation;

public class DrawGC {

        Shell shell = new Shell();

        int width = 310;
        int height = 310;

        Figure figure = new Figure();
        figure.setSize(width, height);

        ImagePrintFigureOperation op = new ImagePrintFigureOperation(shell, figure);
        Image image = op.run();

            event.gc.drawImage(image, 0, 0, width, height, 0, 0, 100, 100);
        });

        shell.open();

        while(!shell.isDisposed()) {
                display.sleep();
            }
        }

        image.dispose();
    }
}
```

The exception is thrown when `zoom` < 100 in `ImagePrintFigureOperation.getImageDataAtZoom(int zoom)`.

```
Exception in thread "main" java.lang.IllegalArgumentException: Argument not valid
	at org.eclipse.swt.SWT.error(SWT.java:4915)
	at org.eclipse.swt.SWT.error(SWT.java:4849)
	at org.eclipse.swt.SWT.error(SWT.java:4820)
	at org.eclipse.swt.graphics.GC$DrawScalingImageToImageOperation.computeSourceRectangle(GC.java:1251)
	at org.eclipse.swt.graphics.GC$DrawScalingImageToImageOperation.lambda$0(GC.java:1223)
	at org.eclipse.swt.graphics.Image.executeOnImageHandleAtBestFittingSize(Image.java:946)
	at org.eclipse.swt.graphics.GC$DrawScalingImageToImageOperation.draw(GC.java:1222)
	at org.eclipse.swt.graphics.GC$DrawScalingImageToImageOperation.apply(GC.java:1204)
	at org.eclipse.swt.graphics.GC.storeAndApplyOperationForExistingHandle(GC.java:6102)
	at org.eclipse.swt.graphics.GC.drawImage(GC.java:1134)
```

I've tried to simulate what I'm getting using `ImagePrintFigureOperation` to create an image from a `Figure` and displaying it in a Nebula `Gallery`. The Gallery widget calls `drawImage (Image image, int srcX, int srcY, int srcWidth, int srcHeight, int destX, int destY, int destWidth, int destHeight)` as in the Snippet.

This may be an SWT issue but I'm wondering if something could be done in `ImagePrintFigureOperation.getImageDataAtZoom` to work around this?

For now I've added:

```java
if(zoom < 100) {
    return null;
}
```

(Edited - happens on all display scales including 100%)

### LINKED ISSUE #3454: [Win32] Inconsistent scaling in GC.drawImage() methods at monitor zoom != 100%
**Describe the bug**
The GC.drawImage() operation for scaled and cropped images (9-arguments method) does not produce consistent results to the other GC.drawImage() methods when drawing an image at its original size on a non-100% monitor.

**To Reproduce**
The issuse can be reproduced with the following snippet and the attached file:
```java
		shell.setText("Graphics.drawImage() Repro");
		shell.setSize(400, 200);

		Image sourceImage = new Image(display, Snippet164.class.getResourceAsStream("source.png"));

		shell.addPaintListener(e -> {
			Rectangle imageBounds = sourceImage.getBounds();

			e.gc.drawImage(sourceImage, //
					10, 0);

			e.gc.drawImage(sourceImage, //
					0, 0, imageBounds.width, imageBounds.height, //
					10, imageBounds.height + 2, imageBounds.width, imageBounds.height);
		});

		shell.open();
		try {
			while (!shell.isDisposed()) {
					display.sleep();
				}
			}
		} finally {
			sourceImage.dispose();
			display.dispose();
		}
	}
```

[source.png](https://github.com/user-attachments/assets/db7b7da3-002b-4bd7-a5ae-041bf3231410)

Both images should be rendered at equal size, but it looks like this:
Image

**Expected behavior**
Images should be rendered consistently at their original size when using the GC.drawImage() operations on every monitor zoom.

**Environment:**
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 

Use the non-cropping GC.drawImage() operation to achieve better/expected results.

HeikoKlare:
## Summary

`GC.drawImage` with explicit source and destination rectangles has several bugs in how it maps the logical source rectangle to the pixel coordinates of the image handle used for rendering. The image handle selection in `Image` also picks the wrong zoom level for asymmetric images. This PR fixes both.

<s>
> [!IMPORTANT]
> This change is based on and should be merged after (or together with):
> - https://github.com/eclipse-platform/eclipse.platform.swt/pull/3455
</s>

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/3454

May also fix https://github.com/eclipse-gef/gef-classic/issues/1146 

## Reproduction / How to Test

See https://github.com/eclipse-platform/eclipse.platform.swt/issues/3454

The snippet should show an image rendered equally two times (in particular at 125% monitor zoom).

### Before this change
image

### After this change
image

### Wrong image handle zoom for asymmetric images (`Image.java`)

When selecting which zoom-level handle to use, the code estimated the required zoom separately for each axis and then took the maximum. For asymmetric images the smaller axis produces a larger integer-division truncation error, so `Math.max` reliably picks the *wrong* zoom.

> **Example:** a 500×19 image at 125% maps the width to zoom 125 (exact) but the height to zoom 126 (truncated from 126.3). `Math.max` picks 126, requesting a handle at the wrong zoom level and feeding incorrect pixel dimensions into all downstream calculations.

**Fix:** use the zoom estimate from the larger hint dimension, which has more pixels and therefore less relative rounding error.

### X and Y axes treated as a single scale factor (`GC.java`)

When converting the source rectangle from logical points to handle pixels, the old code collapsed the independent per-axis scale factors into one value using `Math.min`, then applied it to both axes. For asymmetric images the X and Y factors diverge because each axis is rounded independently when the handle is created.

> **Example:** the 625×24 handle for the 500×19 image above gives scaleFactorX = 1.25 but scaleFactorY ≈ 1.263. Using 1.25 for Y shifts crop boundaries by one pixel for tall sub-regions.

**Fix:** compute and apply `scaleFactorX` and `scaleFactorY` independently for each axis, using the same endpoint-subtraction rounding that `Image` uses when scaling image data — guaranteeing consistent results between the two.

### Error tolerance collapses to zero at sub-100% zoom (`GC.java`)

After computing the pixel source rectangle, the code checks whether it slightly overflows the handle bounds (an expected rounding side-effect) and clamps it within a tolerance. The tolerance was computed as an integer division, which truncates to zero for any zoom below 100%, turning every normal rounding artifact into a spurious `ERROR_INVALID_ARGUMENT`.

> **Example:** at 75% zoom the tolerance was `75 / 100 = 0`, so even a 1-pixel overshoot was rejected as an error.

**Fix:** use `Math.ceil(scaleFactor)` instead, which always allows at least 1 pixel of tolerance regardless of zoom level. Additionally, the bounds correction was previously guarded so it was skipped entirely when only one axis had a non-unity scale factor; the guard is corrected to apply whenever *either* axis is scaled.

## Tests

A new `@ParameterizedTest` in `GCWin32Tests` verifies that the 3-arg, 5-arg, and 9-arg `drawImage` overloads all produce identical pixel output for a 500-wide source image across a 5×9 matrix of zoom levels (100%–200%) and prime image heights (1–19). Prime heights maximise the chance of exposing per-axis rounding divergence.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
@@ -1250,11 +1250,15 @@ private class DrawScalingImageToImageOperation extends ImageOperation {
 		 * computed to pixels depending on the factor of the full image bounds to the
 		 * actual OS handle size that will be used.
 		 */
-		float scaleFactor = Math.min(1f * imageHandle.width() / fullImageBounds.width, 1f * imageHandle.height() / fullImageBounds.height);
-		int closestZoomOfHandle = Math.round(scaleFactor * 100);
-		Rectangle srcPixels = Win32DPIUtils.pointToPixel(drawable, src, closestZoomOfHandle);
+		float scaleFactorX = (1f * imageHandle.width()) / fullImageBounds.width;
+		float scaleFactorY = (1f * imageHandle.height()) / fullImageBounds.height;
+		int srcXPixels = Math.round(scaleFactorX * src.x);
+		int srcWidthPixels = Math.round(scaleFactorX * (src.x + src.width)) - srcXPixels;
+		int srcYPixels = Math.round(scaleFactorY * src.y);
+		int srcHeightPixels = Math.round(scaleFactorY * (src.y + src.height)) - srcYPixels;
+		Rectangle srcPixels = new Rectangle(srcXPixels, srcYPixels, srcWidthPixels, srcHeightPixels);
 
-		if (closestZoomOfHandle != 100) {
+		if (Math.abs(scaleFactorX - 1f) >= 0.01f || Math.abs(scaleFactorY - 1f) >= 0.01f) {
 			/*
 			 * This is a HACK! Due to rounding errors at fractional scale factors,
 			 * the coordinates may be slightly off. The workaround is to restrict
@@ -1263,7 +1267,7 @@ private class DrawScalingImageToImageOperation extends ImageOperation {
 			int errX = srcPixels.x + srcPixels.width - imageHandle.width();
 			int errY = srcPixels.y + srcPixels.height - imageHandle.height();
 			if (errX != 0 || errY != 0) {
-				if (errX <= closestZoomOfHandle / 100 && errY <= closestZoomOfHandle / 100) {
+				if (errX <= Math.max(1, scaleFactorX) && errY <= Math.max(1, scaleFactorY)) {
 					srcPixels.intersect(new Rectangle(0, 0, imageHandle.width(), imageHandle.height()));
 				} else {
 					SWT.error (SWT.ERROR_INVALID_ARGUMENT);
--- a/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/graphics/GCWin32Tests.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/graphics/GCWin32Tests.java
@@ -14,16 +14,19 @@
 package org.eclipse.swt.graphics;
 
 import static org.junit.Assert.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertAll;
-import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.*;
 
+import java.util.*;
 import java.util.concurrent.*;
+import java.util.stream.*;
 
 import org.eclipse.swt.*;
 import org.eclipse.swt.internal.*;
 import org.eclipse.swt.widgets.*;
 import org.junit.jupiter.api.*;
 import org.junit.jupiter.api.extension.*;
+import org.junit.jupiter.params.*;
+import org.junit.jupiter.params.provider.*;
 
 @ExtendWith(PlatformSpecificExecutionExtension.class)
 @ExtendWith(WithMonitorSpecificScalingExtension.class)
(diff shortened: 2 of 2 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -1250,11 +1250,15 @@ private Rectangle computeSourceRectangle(ImageHandle imageHandle, Rectangle full
 		 * computed to pixels depending on the factor of the full image bounds to the
 		 * actual OS handle size that will be used.
 		 */
-		float scaleFactor = Math.min(1f * imageHandle.width() / fullImageBounds.width, 1f * imageHandle.height() / fullImageBounds.height);
-		int closestZoomOfHandle = Math.round(scaleFactor * 100);
-		Rectangle srcPixels = Win32DPIUtils.pointToPixel(drawable, src, closestZoomOfHandle);
-
-		if (closestZoomOfHandle != 100) {
+		float scaleFactorX = 1f * imageHandle.width() / fullImageBounds.width;
+		float scaleFactorY = 1f * imageHandle.height() / fullImageBounds.height;
```
r-mennig: Operator precedence could be made clearer here.

```suggestion
		float scaleFactorX = (1f * imageHandle.width()) / fullImageBounds.width;
		float scaleFactorY = (1f * imageHandle.height()) / fullImageBounds.height;
```

r-mennig (approved):

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
```diff
@@ -1263,7 +1267,7 @@ private Rectangle computeSourceRectangle(ImageHandle imageHandle, Rectangle full
 			int errX = srcPixels.x + srcPixels.width - imageHandle.width();
 			int errY = srcPixels.y + srcPixels.height - imageHandle.height();
 			if (errX != 0 || errY != 0) {
-				if (errX <= closestZoomOfHandle / 100 && errY <= closestZoomOfHandle / 100) {
+				if (errX <= Math.ceil(scaleFactorX) && errY <= Math.ceil(scaleFactorY)) {
 					srcPixels.intersect(new Rectangle(0, 0, imageHandle.width(), imageHandle.height()));
```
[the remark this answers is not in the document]
HeikoKlare: Fixed

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/GC.java
HeikoKlare: Done with the latest commit.

Issue #3214, opened 2026-04-10
[Cocoa] Deprecated OS constants


HeikoKlare:
## Analysis of Deprecated macOS Constants in `OS.java`

The constants defined in [`OS.java`](bundles/org.eclipse.swt/Eclipse%20SWT%20PI/cocoa/org/eclipse/swt/internal/cocoa/OS.java) (lines 2177–2365) are replications of constants from macOS SDK header files. An analysis against the macOS 15.4 SDK revealed **53 deprecated constants** across 14 headers, most deprecated since macOS 10.12 or 10.14.

The tables below list each deprecated constant alongside its replacement. The OpenGL constants are a special case since the entire OpenGL API was deprecated in favor of Metal/MetalKit.

---

- [x] **Window Style Masks (`NSWindow.h`) — deprecated macOS 10.0–10.12**

| Current Constant | Replacement |
|---|---|
| `NSBorderlessWindowMask` | `NSWindowStyleMaskBorderless` |
| `NSTitledWindowMask` | `NSWindowStyleMaskTitled` |
| `NSClosableWindowMask` | `NSWindowStyleMaskClosable` |
| `NSMiniaturizableWindowMask` | `NSWindowStyleMaskMiniaturizable` |
| `NSResizableWindowMask` | `NSWindowStyleMaskResizable` |
| `NSFullScreenWindowMask` | `NSWindowStyleMaskFullScreen` |
| `NSUtilityWindowMask` | `NSWindowStyleMaskUtilityWindow` |
| `NSDocModalWindowMask` | `NSWindowStyleMaskDocModalWindow` |
| `NSNonactivatingPanelMask` | `NSWindowStyleMaskNonactivatingPanel` |

---

- [ ] **Compositing Operations (`NSGraphics.h`) — deprecated macOS 10.0–10.12**

| Current Constant | Replacement |
|---|---|
| `NSCompositeClear` | `NSCompositingOperationClear` |
| `NSCompositeCopy` | `NSCompositingOperationCopy` |
| `NSCompositeSourceOver` | `NSCompositingOperationSourceOver` |
| `NSCompositeSourceAtop` | `NSCompositingOperationSourceAtop` |

---

- [ ] **Alert Styles (`NSAlert.h`) — deprecated macOS 10.3–10.12**

| Current Constant | Replacement |
|---|---|
| `NSWarningAlertStyle` | `NSAlertStyleWarning` |
[...]

PR #3433, opened 2026-07-12
[Cocoa] Replace deprecated NSProgressIndicatorPreferredThickness

HeikoKlare:
`NSProgressIndicatorPreferredThickness` is deprecated since macOS 10.14. The deprecation message in the SDK header (`NSProgressIndicator.h`) prescribes the replacement:

> These constants do not accurately represent the geometry of NSProgressIndicator. Use `controlSize` and `sizeToFit` instead.

This change adds a binding for `NSProgressIndicator#sizeToFit` and uses it in `ProgressBar#computeSize` to determine the default thickness, restoring the original frame afterwards since `computeSize` must not have side effects on the widget's bounds. The header documents exactly the semantics this relies on:

> For the spinning style, it will size the view to its default size. For the bar style, the height will be set to the recommended height.

SWT always uses the bar style and never changes the indicator's `controlSize` (it stays at the default `NSControlSizeRegular`), so `sizeToFit` yields the recommended thickness for that control size.

Note that this changes the default thickness of a `ProgressBar` from the hardcoded 14px to the value AppKit actually reports for the current system (18px on macOS 15): the deprecated constant was a stale Aqua-era value that, per the deprecation message, "does not accurately represent the geometry" of the rendered indicator.

## How to test

Open the **ProgressBar tab** of the SWT ControlExample (`org.eclipse.swt.examples.controlexample.ControlExample`). In the **Size** group, the default **Preferred** setting sizes the bars via `computeSize`, i.e., the changed code path. With this change, the horizontal progress bar becomes 18px instead of 14px tall (the vertical one correspondingly wider), matching the geometry AppKit recommends.

Before:
image

After:
image

Contributes to #3214

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT PI/cocoa/org/eclipse/swt/internal/cocoa/AppKitFull.bridgesupport.extras
+++ b/bundles/org.eclipse.swt/Eclipse SWT PI/cocoa/org/eclipse/swt/internal/cocoa/AppKitFull.bridgesupport.extras
@@ -2376,6 +2376,9 @@
 			<arg swt_gen="true"></arg>
 			<retval swt_gen="true"></retval>
 		</method>
+		<method selector="sizeToFit" swt_gen="true">
+			<retval swt_gen="true"></retval>
+		</method>
 		<method selector="startAnimation:" swt_gen="true">
 			<arg swt_gen="true"></arg>
 			<retval swt_gen="true"></retval>
@@ -4507,7 +4510,6 @@
 	<enum name="NSPortraitOrientation" swt_gen="true"></enum>
 	<enum name="NSPrintPanelShowsPageSetupAccessory" swt_gen="true"></enum>
 	<enum name="NSPrintPanelShowsPrintSelection" swt_gen="true"></enum>
-	<enum name="NSProgressIndicatorPreferredThickness" swt_gen="true"></enum>
 	<enum name="NSRegularSquareBezelStyle" swt_gen="true"></enum>
 	<enum name="NSResizableWindowMask" swt_gen="true"></enum>
 	<enum name="NSRightMouseDown" swt_gen="true"></enum>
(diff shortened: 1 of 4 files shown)


Phillipus:
I thought you couldn't add code to `NS*` classes (`NSProgressIndicator.java`) as they are auto-generated by the MacGenerator tool?

https://github.com/eclipse-platform/eclipse.platform.swt/pull/1081#issuecomment-1984234044

HeikoKlare:
> I thought you couldn't add code to `NS*` classes (`NSProgressIndicator.java`) as they are auto-generated by the MacGenerator tool?

That's a very good hint, thank you. I have to admit that I was not aware of the exact purpose of that tool. I just executed it and turns out that several changes performed in the recent past (by me and others) did not take that into account properly, such that running the MacGenerator now produces some unexpected result. I will try to take a look what needs to be done to make the current state consistent to the MacGenerator again. And we definitely have to update the `Agents.md` with that information to ensure that agents to not produce such changes.

Phillipus:
Actually I missed that you also made changes in `AppKitFull.bridgesupport.extras` in this commit so maybe you already did the right thing?

HeikoKlare:
Might be. But since the agent did definitely not use the MacGenerator tool, we should not further process this before doing so and verifying that the change is sound.
I propose to first clean up the current state (such that executing the MacGenerator does not produce any changes) before proceeding with this one (thus marking it as draft again).

HeikoKlare:
Except for the remaining open PR, the cleanup is done now. `MacGenerator` produces (almost) no changes anymore (except for a long-standing custom `IS_X86_64` macro issue). Turns out that this change is correct, as is only removes the generation instructions in bridgesupport.extras, as already mentioned in https://github.com/eclipse-platform/eclipse.platform.swt/pull/3433#issuecomment-4967412799. So this change is ready.

Issue #3421, opened 2026-07-07
Scrollbars in toolbar pop-up menus


Defective4:
**Describe the bug**
This bug is related to Eclipse `2026-09 (4.41M1)` with `SWT_GTK4` environment variable set to `1`, run on Debian Trixie with the Cinnamon desktop environment.
Some toolbar pop-up menus display useless vertical and horizontal scrollbars in them.

Scrolling the scrollbars does not make them disappear.
This was tested with the following GTK themes:
- Adwaita
- Fluent Purple Dark
- Fluent Purple

All themes have produced the exact same results.

**To Reproduce**
- Run the [4.41M1 Eclipse Platform binary](https://www.eclipse.org/downloads/download.php?file=/eclipse/downloads/drops4/S-4.41M1-202607031319/eclipse-platform-4.41M1-linux-gtk-x86_64.tar.gz) with a new workspace
- Click on `New`, or `Run Last Tool` dropdowns

**Expected behavior**
The menus should appear without any scrollbars

**Screenshots**
<img width="201" height="228" alt="Image" src="https://github.com/user-attachments/assets/605dcce0-bf7f-42d3-991d-c391e0802ac1" />
<img width="289" height="160" alt="Image" src="https://github.com/user-attachments/assets/58dcd3a0-b4aa-44fa-8648-4640e93bc961" />

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [ ] Windows
- - [x] Linux
- - [ ] macOS 

2. Additional OS info (e.g. OS version, Linux Desktop, etc)
`Debian 13 (6.12.86+deb13-amd64), Cinnamon 6.4.10`

3. JRE/JDK version
`OpenJDK 64-Bit (build 21.0.2+13-58)`

**Version since**
4.41M1 (Tested with Eclipse Platform)

akurtakov:
Fixed via the linked PR.

PR #3444, opened 2026-07-16
[Gtk4] Fix spurious scrollbar on popup menus

akurtakov:
GtkPopoverMenu inserts section-separator widgets from a GLib idle callback (gtk_menu_section_box_handle_sync_separators), but SWT popped the popover synchronously in the same call that built the items, so the popover measured its height before the separators existed, undersized itself and fell back to a scrollbar. Defer gtk_popover_popup() via display.asyncExec() so GTK's pending idle work runs first.

Additionally, a bottom-of-screen menu carries GDK_ANCHOR_RESIZE_Y and is shrunk into a scrolled view when it fits neither below nor flipped above the anchor. Wayland has no global window coordinates to compute the overflow, so compare the popover's allocated height to its natural height and iteratively shift the anchor up by the shortfall, keeping the popover hidden (opacity 0) until it fits.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Menu.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/gtk/org/eclipse/swt/widgets/Menu.java
@@ -261,6 +261,88 @@ boolean ableToSetLocation() {
 	return hasLocation;
 }
 
+/*
+ * GtkPopoverMenu inserts section-separator widgets from a GLib idle callback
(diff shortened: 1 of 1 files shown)


akurtakov:
Before toolitem menu:
Screenshot From 2026-07-16 11-51-23

After toolitem menu:
Screenshot From 2026-07-16 11-50-29

Before editor menu:
Screenshot From 2026-07-16 11-51-36

After editor menu:
Screenshot From 2026-07-16 11-50-43

akurtakov:
Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/3421

PR #2882, opened 2025-12-14
Add native PDF output support

laeubi:
Currently if one wants to create a PDF file it requires external
libraries and as SWT does not allows an abstraction like Grahics2D in
AWT one can not export real content of SWT components (e.g. Canvas)
except exporting as an raster image or using some hacks.

This now introduce a new PDFDocument to enable direct
PDF generation from SWT widgets via Control.print(GC). This allows
applications to export widget content to PDF files using the standard
GC drawing API as well as even creating completely customized documents.

## Platform Support

| Platform | Implementation |
|----------|----------------|
| GTK (Linux) | Cairo PDF surface |
| Cocoa (macOS) | Core Graphics CGPDFContext |
| Win32 (Windows) | Microsoft Print to PDF |

I currently have only tested this on Linux GTK, I currently have no way to test it with Cocoa (macOS)  it would be great if someone with a mac can take a look. I'll try to take a look on windows soon but also here some help would be appreciated!

The Snippet388 can be used to produce such PDF after compiling the natives with the "Build-SWT-native-binaries-for-running-platform" run configuration.

### GTK (Linux) 

On GTK (Linux) the snippet looks like this:
grafik

And here is the resulting PDF document:
- [swt_graphics_demo.pdf](https://github.com/user-attachments/files/24148543/swt_graphics_demo.pdf)

### Win32 (Windows)

On Win32 (Windows) the snippet looks like this:

grafik

Due to the usage of the pdf printer that do not supports custom page size, we have the limitation to not getting perfect page sizes and we currently try to find the best matching one, the PDF document then looks like tihs:

- [swt_graphics_demo.pdf](https://github.com/user-attachments/files/24150269/swt_graphics_demo.pdf)

As an interesting observaation I found that Shell#print is not implemented for windows (simply returns), I added an implementation for this as well, but this part is maybe better submited as an own PR.

PR #3446, opened 2026-07-16
[Cocoa] Add generation instructions for PDF export functions

HeikoKlare:
The `CGPDFContext*` functions are all present in the CoreGraphicsFull.bridgesupport files but were never registered in the corresponding .extras files, so MacGenerator silently dropped them on the next regeneration. This adds the missing `swt_gen="true"` entries and reruns the MacGenerator to adopt its ordering of generated entries in the OS class.

Follow-up to:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2882

@laeubi fyi in case this unexpectedly breaks something for PDFDocument on MacOS


laeubi:
Thanks for fixing these!

PR #3455, opened 2026-07-21
[Win32] Fix image handle zoom selection for asymmetric image dimensions

HeikoKlare:
For images where width and height differ greatly (e.g. 500x2), computing the required zoom independently per axis using integer division can give a larger zoom estimate for the smaller axis. The old code selected the maximum of both estimates, which reliably picks the less accurate value and creates a handle at the wrong zoom level, causing noticeably incorrect scaling.

The axis with more pixels has a proportionally smaller integer-division truncation error. The fix therefore derives the zoom from the axis whose pixel hint is larger, giving the most accurate zoom estimate.

Contributes to https://github.com/eclipse-platform/eclipse.platform.swt/issues/3454

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/graphics/ImagesWin32Tests.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/graphics/ImagesWin32Tests.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2024 Yatta Solutions
+ * Copyright (c) 2024, 2026 Yatta Solutions
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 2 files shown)


bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/graphics/ImagesWin32Tests.java
```diff
@@ -30,9 +35,52 @@ public void testImageIconTypeShouldNotChangeAfterCallingGetHandleForDifferentZoo
...
+					"Handle width should match widthHint for a " + logicalWidth + "x" + logicalHeight
+					+ " image at display zoom " + displayZoom + "%: expected " + widthHint
+					+ " but got " + capturedWidth[0]);
+		} finally {
+			image.dispose();
+		}
+	}
+
+	private static Stream<Arguments> provideZoomAndHeightCombinations() {
+		int[] zooms = { 100, 125, 150, 175, 200 };
```
r-mennig, line 81: Shouldn't zooms < 100 also be considered here?

r-mennig (approved):

HeikoKlare:
I am currently not sure how to best enhance the zoom calculation. The goal was to choose the axis for zoom calculation with the lower rounding error, but just doing that based on the given hint for the target dimension does not seem to be correct. Though that value is already rounded because point-to-pixel and pixel-to-point conversions, small bounds of the image in one of the dimensions may also influence the rounding error. In addition, the change optimizes precision of the calculated zoom while ignoring the dominating axis in case the image is scaled asymmetrically (e.g., by factor 5 in y direction and by factor 1 in x direction), where it is most reasonable to better use an image version that is too large rather than using one that is too small.

This was intended to be a preparation step for:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/3431

This is because for heights between 1 and 3 pixels and a much larger widths, the dimensions of the handle being returned here are not as expected, due to large rounding errors with such a small size. However, images of 1 to 3 pixel in height do not seem to be that relevant use cases (as you cannot see anything relevant in them anyway, not matter if a higher or lower resolution version is used) that it justifies the drawback mentioned about and the risk of other unintended changes in behavior / regressions.

For the mentioned reasons, I close this PR for now. Thanks to @r-mennig for challenging the change, leading to this conclusion.

PR #3459, opened 2026-07-22
[Win32] Extend GC.drawImage() handle selection with exact-zoom and monitor-zoom logic

HeikoKlare:
## Summary

This PR extends the `GC.drawImage()` handle-selection logic with two improvements.

**Exact imageZoom lookup before nearestAvailableZoom**
Before consulting `nearestAvailableZoom`, the image's persistent handle manager is now queried at the exact `imageZoom` first. This ensures that a handle previously created for a specific zoom is found and reused even when the provider would not normally supply data at that zoom on its own. For example, a 200% handle pre-created for a 100%-only image is now returned when drawing at the 200%-equivalent pixel size rather than falling back to the 100% handle from `nearestAvailableZoom`.

**Monitor-zoom persistence**
The condition for eagerly persisting a handle is extended from `nearestAvailableZoom == 100%` to also cover any zoom that matches a current monitor's zoom. Monitor zooms are discovered from the zoom reported by each open `Shell`. When `imageZoom` or `nearestAvailableZoom` matches a monitor zoom, the image is likely being drawn repeatedly at that screen's native resolution, so persisting the handle avoids repeated allocations across consecutive draw calls.

As a minor cleanup, `getAvailableMonitorZooms()` is now called once per `getExistingHandle()` invocation rather than twice.

Two regression tests are added to `ImagesWin32Tests` covering both new behaviors.

## Related issue

This contributes to https://github.com/eclipse-platform/eclipse.platform.swt/issues/3454. The issue can be reproduced with the snippet provided there: the second line of text will appear non-antialiased.

Before https://github.com/eclipse-platform/eclipse.platform.swt/pull/3429 was merged, the same artifact also appeared in the snippet when the image was loaded twice (once per draw operation); this PR fixes that case as well.

This PR is a prerequisite for https://github.com/eclipse-platform/eclipse.platform.swt/pull/3431.

### Before
image

### After
image

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
@@ -18,6 +18,7 @@ import static org.eclipse.swt.internal.image.ImageColorTransformer.DEFAULT_DISAB
 
 import java.io.*;
 import java.util.*;
+import java.util.List;
 import java.util.concurrent.*;
 import java.util.concurrent.atomic.*;
 import java.util.function.*;
@@ -29,6 +30,7 @@ import org.eclipse.swt.internal.DPIUtil.*;
 import org.eclipse.swt.internal.gdip.*;
 import org.eclipse.swt.internal.image.*;
 import org.eclipse.swt.internal.win32.*;
+import org.eclipse.swt.widgets.*;
 
 /**
  * Instances of this class are graphics which have been prepared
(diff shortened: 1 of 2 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -277,6 +276,32 @@ private InternalImageHandle getOrCreateImageHandleAtClosestSize(int widthHint, i
...
+			}
+			return null;
+		}
+
+		private Set<Integer> getAvailableMonitorZooms() {
+			if (getDevice() instanceof Display display) {
+				return Arrays.stream(display.getShells()).map(Shell::getZoom).collect(Collectors.toSet());
+			}
+			return Collections.emptySet();
+		}
```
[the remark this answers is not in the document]
HeikoKlare: Done

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -277,6 +276,36 @@ private InternalImageHandle getOrCreateImageHandleAtClosestSize(int widthHint, i
...
+			}
+			Set<Integer> monitorZooms = getAvailableMonitorZooms();
+			if (monitorZooms.contains(imageZoom)) {
+				return getHandleInternal(imageZoom, imageZoom);
+			}
+			bestFittingHandle = imageHandleManager.get(nearestAvailableZoom);
+			if (bestFittingHandle != null) {
+				return bestFittingHandle;
+			}
+			if (monitorZooms.contains(nearestAvailableZoom) || nearestAvailableZoom == 100) {
```
r-mennig: I can see why persisting a handle matching the monitor zoom or 100% zoom would be beneficial here (since those will likely be reused multiple times), but is it really useful to persist a handle for the nearest available zoom if it matches neither of those zooms?

HeikoKlare: You are right, those handles should not be persisted. I have removed this case. Now this PR just adds the exact-zoom retrieval from existing handles and for zooms that fit to one of the shell zooms.

r-mennig (approved):

PR #3463, opened 2026-07-24
[Win32] Fix drawImage() failing for an empty image

### LINKED ISSUE #3442: NPE when painting "empty" image
**Describe the bug**

Trying to paint an "empty" image (an image for which no GC was created) fails with a `NullPointerException`.

**To Reproduce**

```
package gef;

public class DrawGC2 {
		Shell shell = new Shell();

		int width = 310;
		int height = 310;

		Image image = new Image(display, width, height);

			event.gc.drawImage(image, 0, 0, width, height, 0, 0, 100, 100);
		});

		shell.open();

		while (!shell.isDisposed()) {
				display.sleep();
			}
		}

		image.dispose();
	}
}
```

Stacktrace:
```
Exception in thread "main" java.lang.NullPointerException: Cannot invoke "java.lang.Integer.intValue()" because the return value of "java.util.Optional.orElse(Object)" is null
	at org.eclipse.swt.graphics.Image$AbstractImageProviderWrapper.getClosestAvailableImageData(Image.java:2105)
	at org.eclipse.swt.graphics.Image$PlainImageProviderWrapper.loadImageData(Image.java:2388)
	at org.eclipse.swt.graphics.Image$HandleAtSize.getOrCreateImageHandleAtClosestSize(Image.java:254)
	at org.eclipse.swt.graphics.Image$HandleAtSize.lambda$0(Image.java:225)
	at java.base/java.util.Optional.orElseGet(Optional.java:364)
	at org.eclipse.swt.graphics.Image$HandleAtSize.refresh(Image.java:225)
	at org.eclipse.swt.graphics.Image.executeOnImageHandleAtBestFittingSize(Image.java:968)
	at org.eclipse.swt.graphics.GC$DrawScalingImageToImageOperation.draw(GC.java:1240)
	at org.eclipse.swt.graphics.GC$DrawScalingImageToImageOperation.apply(GC.java:1222)
	at org.eclipse.swt.graphics.GC.storeAndApplyOperationForExistingHandle(GC.java:6136)
	at org.eclipse.swt.graphics.GC.drawImage(GC.java:1152)
	at gef.DrawGC2.lambda$0(DrawGC2.java:18)
	at org.eclipse.swt.widgets.TypedListener.handleEvent(TypedListener.java:272)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:4363)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1217)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1241)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1226)
	at org.eclipse.swt.widgets.Composite.WM_PAINT(Composite.java:1627)
	at org.eclipse.swt.widgets.Control.windowProc(Control.java:5021)
	at org.eclipse.swt.widgets.Canvas.windowProc(Canvas.java:336)
	at org.eclipse.swt.widgets.Decorations.windowProc(Decorations.java:1490)
	at org.eclipse.swt.widgets.Shell.windowProc(Shell.java:2446)
	at org.eclipse.swt.widgets.Display.windowProc(Display.java:5131)
	at org.eclipse.swt.internal.win32.OS.UpdateWindow(Native Method)
	at org.eclipse.swt.widgets.Decorations.setVisible(Decorations.java:1322)
	at org.eclipse.swt.widgets.Shell.setVisible(Shell.java:2192)
	at org.eclipse.swt.widgets.Shell.open(Shell.java:1334)
	at gef.DrawGC2.main(DrawGC2.java:21)
```

**Expected behavior**
Either a "blank" image should be drawn or an SWTError be thrown.

**Environment:**
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 

3. JRE/JDK version
 
**Version since**

Only tested with the master.

HeikoKlare:
Fixes drawing an empty (blank) `Image` via `GC.drawImage()` with a scaled source/destination region.

## Problem

Drawing a blank image with a scaled region failed with a `NullPointerException`. The nearest-zoom lookup in `PlainImageProviderWrapper` tried to derive the zoom from an existing image handle, but an empty image has no handle yet, so the lookup resolved to `null`.

Reproducible with the updated `Snippet6` taken from the issue.

## Fix

The nearest-zoom lookup now falls back to using an empty image at 100% zoom when no image handle exists yet, so drawing an empty image succeeds.

## Test

Adds an OS-independent test to `Test_org_eclipse_swt_graphics_GC` that draws an empty image with a scaled region to guard against this regression.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/3442

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
@@ -175,6 +175,15 @@ public final class Image extends Resource implements Drawable {
 			return zoomLevelToImageHandle.keySet();
 		}
 
+		Integer getNearestAvailableZoom(int zoom) {
+			TreeSet<Integer> availableZooms = new TreeSet<>(getAllZooms());
+			if (availableZooms.contains(zoom)) {
+				return zoom;
+			}
+			Integer higher = availableZooms.higher(zoom);
+			return higher != null ? higher : availableZooms.lower(zoom);
+		}
+
 		void destroyHandles(Predicate<Integer> filter) {
 		    zoomLevelToImageHandle.entrySet().removeIf(entry -> {
 		    	if (filter.test(entry.getKey())) {
@@ -2155,8 +2164,7 @@ private abstract class AbstractImageProviderWrapper {
 	abstract AbstractImageProviderWrapper createCopy(Image image);
 
 	ElementAtZoom<ImageData> getClosestAvailableImageData(int zoom) {
-		TreeSet<Integer> availableZooms = new TreeSet<>(imageHandleManager.getAllZooms());
-		int closestZoom = availableZooms.contains(zoom) ? zoom : Optional.ofNullable(availableZooms.higher(zoom)).orElse(availableZooms.lower(zoom));
+		int closestZoom = imageHandleManager.getNearestAvailableZoom(zoom);
 		ImageData imageData = imageHandleManager.get(closestZoom).getImageData();
 		return new ElementAtZoom<>(imageData, closestZoom);
 	}
@@ -2467,7 +2475,10 @@ private class PlainImageProviderWrapper extends AbstractImageProviderWrapper {
 
 	@Override
 	int nearestAvailableZoom(int zoom) {
-		return getClosestAvailableImageData(zoom).zoom();
+		if (imageHandleManager.isEmpty()) {
+			return 100;
+		}
+		return imageHandleManager.getNearestAvailableZoom(zoom);
 	}
 
 	@Override
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_GC.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_graphics_GC.java
@@ -382,6 +382,20 @@ public void test_drawImageLorg_eclipse_swt_graphics_ImageIIIIIIII() {
 	images.dispose();
 }
 
+/**
+ * See https://github.com/eclipse-platform/eclipse.platform.swt/issues/3442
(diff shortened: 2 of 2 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
```diff
@@ -2467,7 +2467,11 @@ protected ElementAtZoom<ImageData> loadImageData(int zoom) {
 
 	@Override
 	int nearestAvailableZoom(int zoom) {
-		return getClosestAvailableImageData(zoom).zoom();
+		if (imageHandleManager.isEmpty()) {
+			return 100;
+		}
+		TreeSet<Integer> availableZooms = new TreeSet<>(imageHandleManager.getAllZooms());
+		return availableZooms.contains(zoom) ? zoom : Optional.ofNullable(availableZooms.higher(zoom)).orElse(availableZooms.lower(zoom));
```
r-mennig: Using Optionals here just for chaining makes the code harder to read (in my opinion):

```suggestion
		if (availableZooms.contains(zoom)) {
      			return zoom; 
		}
		Integer higher = availableZooms.higher(zoom);
		return higher != null ? higher : availableZooms.lower(zoom);
```

r-mennig (approved):

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/graphics/Image.java
HeikoKlare: Agreed. I had taken the code from `getClosestAvailableImageData()`, where the implementation already existed (without the special case for no handle existing yet). It became so complex when we added the case for a handle with the required zoom potentially existing already (before is was just a higher->orElse->lower implementation).

Anyway, I have extracted the functionality into a shared method inside the `ImageHandleManager` now and there implemented it exactly as you have proposed.

PR #3139, opened 2026-03-18
Fix CTabFolder topRight WRAP overflow triggering too aggressively

vogella:
## Summary

- **Include WRAP controls in `updateTabHeight()`** so the tab row grows to accommodate taller controls (e.g., Text widgets on GTK). Previously WRAP controls were excluded from the height calculation, causing a height-based overflow check to force them below the tabs even with ample horizontal space.
- **Use `<=` instead of `<` in `computeControlBounds()`** when checking if a WRAP control fits the available width, preventing unnecessary overflow for controls that exactly fit.

## Root Cause

On GTK, controls like `Text` are slightly taller (~31px) than the default `tabHeight` (~24px) due to theme padding. `updateTabHeight()` explicitly skipped WRAP controls when computing tab row height. This meant the height-based overflow check (`tabControlSize[i].y >= tabHeight`) in `computeControlBounds()` always triggered for WRAP composites containing a Text widget, wrapping them to a second row regardless of available horizontal space.

This affected Eclipse view toolbars (e.g., EGit Staging View) where the toolbar contains a search field via `ControlContribution`.

## Test Plan

- [x] Added `test_topRightWrapOverflow` JUnit test verifying inline rendering with sufficient space and wrapping when tight
- [x] Added GTK manual test snippet `Bug3138_CTabFolderWrapDiagnostic` for visual verification
- [ ] Verify on GTK that Eclipse view toolbars with search fields render inline when space allows
- [ ] Verify wrapping still occurs when the shell is resized narrow
- [ ] Verify no regressions with min/max buttons, chevron, or other topRight configurations

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/3138

PR #3467, opened 2026-07-28
[Win32] Fix wrong CTabFolder header height after zoom change

### LINKED ISSUE #3456: Wrong CTabFolder header size after zoom change
**Describe the bug**
When moving an Eclipse application between monitors of different zooms, the headers of some part stacks / `CTabFolders` are too high. They reduce their size once another tab of the folder is selected.

This is a regression from:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/3139

When reverting that commit, everything is fine again.
@vogella can you please have a look?

**To Reproduce**
Start an Eclipse application on a 100% monitor. It looks like this:
Image

Move the window to a 125% monitor and back. The headers of some part stacks / `CTabFolders` are too high:
Image

Click on another tab and they reduce their size again:
Image

**Expected behavior**
The headers of CTabFolders should aways have correct size after moving between monitors of different zooms.

**Environment:**
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 
 
**Version since**
Eclipse or SWT version since when the behavior is seen: 4.40 / 2026-06

Switch to another tab in the `CTabFolder` to restore proper header height.

vogella:
When an application window is moved between monitors with different zoom levels, the header of a `CTabFolder` with a top-right control can end up too high and stays that way until the folder is updated for another reason, for example by selecting a different tab.

Since PR #3139 the tab height also accounts for the top-left/top-right tab controls, so it depends on `Control#computeSize()` of those controls. A control converts its pixel size to points using its parent's zoom, and on a DPI change the folder is rescaled before its children are. Measuring a child in that window mixes the old zoom's pixel size with the new zoom, which produces a too large tab height when moving from a higher to a lower zoom. Nothing recomputed the tab height afterwards, so the wrong value persisted.

This adds a `SWT.ZoomChanged` listener on each tab control that triggers a folder update. Because the listener is registered after the control's own internal zoom handling, the control is already rescaled by the time the folder measures it again. Non-Windows platforms are unaffected at runtime, as rescaling only happens there.

The new test puts the folder into the same state by letting a tab control report a stale size and then sending a zoom change to it, so CI validates the recomputation on every platform without needing monitors with different scaling. It fails on current master and passes with the fix.

Still a draft because I cannot verify the original scenario locally on a multi-DPI Windows setup.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/3456

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Custom Widgets/common/org/eclipse/swt/custom/CTabFolder.java
@@ -172,6 +172,7 @@ public class CTabFolder extends Composite {
 	int[] priority = new int[0];
 	boolean mru = false;
 	Listener listener;
+	Listener tabControlZoomListener;
 	boolean ignoreTraverse;
 	boolean useDefaultRenderer;
 
@@ -344,6 +345,10 @@ void init(int style) {
 			case SWT.ZoomChanged:	   onZoomChange(event); break;
 		}
 	};
+
+	// A tab control is rescaled after the folder itself, so its size can only be
+	// measured reliably once it has processed the zoom change on its own.
+	tabControlZoomListener = event -> updateFolder(UPDATE_TAB_HEIGHT | REDRAW);
 
 	int[] folderEvents = new int[]{
 		SWT.Dispose,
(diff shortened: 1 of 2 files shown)


vogella:
@HeikoKlare can you test? I currently do not have a dual monitor setup.

HeikoKlare:
I've just tested with a single-monitor setup: just change the monitor scaling from 100% to 125% and back. This change seems to fix the issue in that scenario. I will try to test in multi-monitor scenarios tomorrow.

HeikoKlare (approved):
I've tested with a multi-monitor setup as well and that also works as expected for me. The change sounds reasonable and of low risk.

vogella:
Thanks @HeikoKlare for the finding the issue and the testing.

HeikoKlare:
Thanks to you for fixing it 🙂

Issue #3073, opened 2026-02-10
Tool Bar Item Disappears after DPI Change and Enabled State Change


runaway94:
**Describe the bug**
A toolbar menu item with text and icon (`mode="FORCE_TEXT"`) disappears from the tool bar after a DPI change and the tool bar refreshes (e.g. the handler is disabled/enabled)

**To Reproduce**
You need a tool bar with a few tool bar items (for me there are three items) that show text and icon (`mode="FORCE_TEXT"`). 
When I move the window form a 100% scaled monitor to a higher scaled monitor (e.g. 150% scaling) the tool bar icons are still available but the label text is cut off (from "My Label" to "My La...").
As soon as any tool bar item is disabled, the label texts are written out in full again but the last tool bar item disappears from the tool bar. 
This only happens if text and icon are shown and if the scaling changed, if the menu is initiated with a higher scaling everything looks fine.
The item will also not re-appear if it is enabled again.

**Expected behavior**
The tool bar icon should not disappear.

**Screenshots**
On 100% scaling
<img width="308" height="53" alt="Image" src="https://github.com/user-attachments/assets/5038c09f-dd52-43eb-b78d-ff652db2aafa" />
Moved to 150% scaling
<img width="441" height="80" alt="Image" src="https://github.com/user-attachments/assets/12a4337c-c04e-46ef-b78a-a1276b31c3c0" />
After handlers are enabled 
<img width="416" height="73" alt="Image" src="https://github.com/user-attachments/assets/df6855b8-ae0c-4e35-8e4e-e69624564f70" />

**Environment:**
1. Select the platform(s) on which the behavior is seen:
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 

2. Additional OS info (e.g. OS version, Linux Desktop, etc)

3. JRE/JDK version
 
**Version since**
 Eclipse or SWT version since when the behavior is seen [e.g. 4.23]

**Workaround (or) Additional context**
Any known workarounds for the problem?

HeikoKlare:
Thank you for the report, @runaway94.

Since you already seem to have an example/snippet from which you have taken the screenshots, can you please share that with us so that we can easily reproduce the issue?

runaway94:
Hi @HeikoKlare  thank you for the quick answer. 

Unfortunately, I’m unable to share the exact code snippets here, as they are part of a commercial application in which commands, handlers, and menu contributions are distributed across multiple plugins.

[...]

PR #3468, opened 2026-07-28
[Win32] Reinitialize tool bar item images in place on DPI change

### LINKED ISSUE #3466: [Win32] Tool bar items show wrong or blank icons after a DPI change
**Describe the bug**

On Windows, a `ToolBar` item can render another item's icon — or a blank/black icon — after a monitor zoom (DPI) change. This happens when the tool bar's internal image list contains a "hole", i.e. a freed slot left behind by an earlier `ToolItem.setImage(null)` on some other item. It does not require disposing any image and does not strictly require multiple monitors (any DPI change reproduces it), but it is most easily observed by dragging a window between two monitors with different scaling.

The visible effect is that icons appear swapped between tool items after the DPI change.

This is related to:
- https://github.com/eclipse-platform/eclipse.platform.swt/issues/3385

**Root cause (analysis)**

Clearing an item's image with `setImage(null)` frees that item's slot in the `ImageList` without compacting it, leaving a hole. On a DPI change, `Item.handleDPIChange` refreshes the image (since [#3074](https://github.com/eclipse-platform/eclipse.platform.swt/pull/3074) via `setImage(null); setImage(image);`), which routes through `ToolItem.updateImages`'s add-branch. `ImageList.add()` returns the *first free slot* — the hole — so the item is relocated to a lower index, while `ToolBar.handleDPIChange` has already captured the button's original `TBBUTTON.iBitmap`. The result is a stale image index and the wrong/blank icon.

**To Reproduce**

Requires two monitors with different scaling (e.g. 100% and 150%/200%). Run the standalone SWT snippet below:

1. Start the application on the monitor with the *lower* scaling.
2. Press **"Hide 'Two'"** — this clears the second item's icon, leaving a hole in the image list.
3. Drag the window to the monitor with the *higher* scaling (this triggers a DPI change).

```java
import org.eclipse.swt.SWT;
import org.eclipse.swt.widgets.Button;
import org.eclipse.swt.widgets.ToolBar;
import org.eclipse.swt.widgets.ToolItem;

public class ToolBarDpiIconReproducer {

	private static Image solidIcon(Display display, int size, int r, int g, int b) {
		Image image = new Image(display, size, size);
		GC gc = new GC(image);
		Color color = new Color(display, r, g, b);
		gc.setBackground(color);
		gc.fillRectangle(0, 0, size, size);
		gc.dispose();
		return image;
	}

		shell.setText("ToolBar DPI icon reproducer");

		Image red = solidIcon(display, 16, 220, 40, 40);
		Image green = solidIcon(display, 16, 40, 180, 40);
		Image blue = solidIcon(display, 16, 40, 40, 220);
		Image yellow = solidIcon(display, 16, 230, 200, 30);

		ToolBar bar = new ToolBar(shell, SWT.FLAT | SWT.WRAP);
		bar.setLayoutData(new GridData(SWT.FILL, SWT.TOP, true, false));

		ToolItem one = new ToolItem(bar, SWT.PUSH);   one.setText("One");   one.setImage(red);
		ToolItem two = new ToolItem(bar, SWT.PUSH);   two.setText("Two");   two.setImage(green);
		ToolItem three = new ToolItem(bar, SWT.PUSH); three.setText("Three"); three.setImage(blue);
		ToolItem four = new ToolItem(bar, SWT.PUSH);  four.setText("Four");  four.setImage(yellow);

		Label hint = new Label(shell, SWT.WRAP);
		hint.setText("Item colors must stay: One=red, Two=green, Three=blue, Four=yellow.\n"
				+ "1) Press 'Hide Two' to punch a hole in the image list.\n"
				+ "2) Drag this window to a monitor with a different scaling.\n"
				+ "Bug: after the move, icons are swapped/blank.");
		hint.setLayoutData(new GridData(SWT.FILL, SWT.TOP, true, false));

		Button hide = new Button(shell, SWT.PUSH);
		hide.setText("Hide Two (punch hole)");
		hide.addListener(SWT.Selection, e -> two.setImage(null));

		shell.setSize(420, 200);
		shell.open();
		while (!shell.isDisposed()) {
		}
		red.dispose();
		green.dispose();
		blue.dispose();
		yellow.dispose();
		display.dispose();
	}
}
```

**Expected behavior**

Every tool item keeps its own icon after the DPI change: One=red, Three=blue, Four=yellow (Two intentionally has no icon after step 2).

**Screenshots**

Correct behavior (before #3074) — every item keeps its own icon after the DPI change:
Image

Broken behavior (current master) — icons are swapped/blank after the DPI change:
Image

**Environment**

- [ ] All OS
- [x] Windows
- [ ] Linux
- [ ] macOS

**Version since**

Made observable by [#3074](https://github.com/eclipse-platform/eclipse.platform.swt/pull/3074) (fix for [#3073](https://github.com/eclipse-platform/eclipse.platform.swt/issues/3073)), which forces an image refresh on every DPI change; the underlying image-list hole/relocation fragility predates it.

**Additional context**

The DPI-change image refresh should reinitialize each item's image *in place* at its existing image-list slot rather than clearing and re-adding it (which relocates into holes). A candidate fix is in preparation.

HeikoKlare:
## What & why

Fixes #3466: on Windows, a `ToolBar` item can render another item's icon — or a blank one — after a monitor zoom (DPI) change.

The regression was introduced while fixing #3073 (via #3074), which forced a full image refresh for every `ToolItem` on every DPI change by nulling and re-setting the image in `Item.handleDPIChange` to ensure proper item resizing. That is both costly and fragile:

- **Costly:** it removes and re-adds every item's image on every zoom change.
- **Fragile:** for tool bars it can relocate an item to a lower free slot in the image list when a *hole* is present (a slot freed by an earlier `ToolItem.setImage(null)`). The native button was re-added with a previously captured image index, so it ended up pointing at a stale slot — the item then shows another item's icon or a blank one.

Neither disposing an image nor multiple monitors are required to hit this — only a hole in the image list plus any DPI change.

## The fix

- Revert `Item.handleDPIChange` to the cheap no-op refresh.
- Let `ToolBar.handleDPIChange` reinitialize each item's image *in place* via `ToolItem.updateImages`, after the buttons have been re-added and the image lists refreshed.

`updateImages` writes the rescaled image at the item's **existing** image-list slot (no clear/re-add, so no relocation) and forces the button width to be recomputed. This keeps #3073 fixed without the per-zoom overhead and without the wrong-icon regression.

## Testing

Adds `ToolBarWin32Tests.testIconsRenderedCorrectlyAfterZoomChangeWithImageListHole`, a regression test for #3466 using **public API only**: each item gets a distinctly colored icon, a hole is punched via `setImage(null)`, a DPI change is triggered, and the test asserts (by rendering the tool bar and sampling colors under each item's bounds) that every item still renders its own icon. The test fails on current `master` and passes with this change.

The #3073 scenario (a labelled item disappearing after a DPI change + enable/disable in the RCP workbench) is inherently workbench-level and can't be reproduced headlessly, so it is covered by manual verification with the reproducer from #3073.

## Manual reproduction (#3466)

A standalone SWT reproducer is attached in #3466: four PUSH items with distinct colored icons, a button to clear one item's icon (punch the hole), then drag the window between two monitors of different scaling — without the fix the icons become swapped/blank.

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/ToolBarWin32Tests.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/ToolBarWin32Tests.java
@@ -1,0 +1,191 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Vector Informatik GmbH and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.swt.widgets;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.util.*;
+import java.util.function.*;
+
+import org.eclipse.swt.*;
+import org.eclipse.swt.graphics.*;
+import org.eclipse.swt.internal.*;
+import org.eclipse.swt.layout.*;
+import org.junit.jupiter.api.*;
+import org.junit.jupiter.api.extension.*;
+
+/**
+ * Windows-specific tests for {@link ToolBar}'s image handling across monitor
+ * zoom changes.
+ */
+@ExtendWith(PlatformSpecificExecutionExtension.class)
+class ToolBarWin32Tests {
+
+	private static final int TIMEOUT_MILLIS = 5000;
+
+	private record ToolItemWithExpectedColor(ToolItem toolItem, RGB expectedColor) {
+	}
+
+	/**
+	 * Regression test for a tool bar item rendering the wrong or a blank icon after
+	 * a monitor zoom (DPI) change, as reported in <a href=
+	 * "https://github.com/eclipse-platform/eclipse.platform.swt/issues/3466">issue
+	 * #3466</a>.
+	 * <p>
+	 * The bug is triggered by clearing one item's image (which leaves the image
+	 * list in a state that a later zoom change mishandles); neither disposing an
+	 * image nor multiple monitors are required. The test observes the actual
+	 * rendered result via public API only: each item gets a distinctly colored
+	 * icon, and after the zoom change the dominant color under each item must still
+	 * match that item's own icon.
+	 */
+	@Test
+	void testIconsRenderedCorrectlyAfterZoomChangeWithImageListHole() {
+		Display display = new Display();
+		RGB[] colors = { new RGB(220, 40, 40), new RGB(40, 180, 40), new RGB(40, 40, 220), new RGB(230, 200, 30) };
+		Image[] icons = new Image[colors.length];
+		for (int i = 0; i < colors.length; i++) {
+			icons[i] = solidIcon(display, 16, colors[i]);
+		}
+		try {
+			Shell shell = new Shell(display);
+			shell.setLayout(new FillLayout());
+			ToolBar bar = new ToolBar(shell, SWT.FLAT);
+			ToolItem[] items = new ToolItem[colors.length];
+			for (int i = 0; i < colors.length; i++) {
+				items[i] = new ToolItem(bar, SWT.PUSH);
+				items[i].setImage(icons[i]);
+			}
+			shell.setSize(500, 90);
+			shell.open();
+
+			// Punch a hole below the other items: clear the first item's image.
+			items[0].setImage(null);
+
+			int zoom = bar.getAutoscalingZoom();
+			DPITestUtil.changeDPIZoom(bar.getShell(), zoom * 2);
+
+			// Only the first item is expected to rendered blank
+			Set<ToolItemWithExpectedColor> itemsToCheck = new HashSet<>();
+			for (int i = 1; i < items.length; i++) {
+				itemsToCheck.add(new ToolItemWithExpectedColor(items[i], colors[i]));
+			}
+			// Dispatch events until every item renders its own icon color, or fail on
+			// timeout. With the bug an item permanently renders another item's icon or
+			// a blank one, so the condition is never met and the timeout triggers.
+			assertTrue(waitUntilIconsRenderOwnColor(display, () -> iconsRenderOwnColor(bar, itemsToCheck, colors),
+					TIMEOUT_MILLIS), "every tool item must render its own icon color after a zoom change");
+		} finally {
+			for (Image icon : icons)
+				icon.dispose();
+			display.dispose();
+		}
+	}
+
+	private static boolean waitUntilIconsRenderOwnColor(Display display, BooleanSupplier condition,
+			long timeoutMillis) {
+		long deadline = System.currentTimeMillis() + timeoutMillis;
+		while (System.currentTimeMillis() < deadline) {
+			if (condition.getAsBoolean()) {
+				return true;
+			}
+			if (!display.readAndDispatch()) {
+				try {
+					Thread.sleep(10);
+				} catch (InterruptedException e) {
+					Thread.currentThread().interrupt();
+					break;
+				}
+			}
+		}
+		return condition.getAsBoolean();
+	}
+
+	private static boolean iconsRenderOwnColor(ToolBar bar, Set<ToolItemWithExpectedColor> items, RGB[] candidateColors) {
+		Image snapshot = renderToolBar(bar);
+		try {
+			for (ToolItemWithExpectedColor item : items) {
+				if (!item.expectedColor().equals(dominantIconColor(snapshot, item.toolItem().getBounds(), candidateColors))) {
+					return false;
+				}
+			}
+			return true;
+		} finally {
+			snapshot.dispose();
+		}
+	}
+
+	private static Image solidIcon(Display display, int size, RGB rgb) {
+		Image image = new Image(display, size, size);
+		GC gc = new GC(image);
+		Color color = new Color(display, rgb);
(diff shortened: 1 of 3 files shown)


bundles/org.eclipse.swt/Eclipse SWT/common/org/eclipse/swt/widgets/Item.java
```diff
@@ -230,7 +230,6 @@ private void handleDPIChange(Event event) {
 	// Refresh the image
 	Image image = getImage();
 	if (image != null) {
-		setImage(null);
 		setImage(image);
 	}
 }
```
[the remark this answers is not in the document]
HeikoKlare: Thanks for raising this — I checked it, and removing `setImage(null)` is safe. The affected item types don't rely on the forced `setImage(null); setImage(...)` to rebuild their image list.

**Why `TabItem` is not affected:** unlike the base `Item.setImage` (and `ToolItem.setImage`), `TabItem.setImage` has no early-return on an unchanged image. It always calls `super.setImage(image)` and then unconditionally re-runs `parent.imageIndex(image)` + `TCM_SETITEM`, rebuilding the image list that `TabFolder.handleDPIChange` released just before notifying its items. So a single `setImage(image)` suffices; the extra `setImage(null)` was a no-op for tabs.

**Why `Table`/`Tree` are not affected:** here the item's `setImage` *does* early-return on an unchanged `SWT.ICON`, but these controls are callback-driven and never depend on `setImage` to keep the native image list current. On every paint, Windows sends `LVN_GETDISPINFO`/`TVN_GETDISPINFO` and SWT answers via `imageIndex(image, …)`, which lazily recreates the released image list at the new zoom whenever `imageList == null`. The next repaint after the DPI change rebuilds it regardless of whether `setImage` short-circuited.

The only item type that truly needs a forced reinitialization is `ToolItem`, because it stores its image-list slot eagerly in `TBBUTTON.iBitmap` and has no dispinfo callback — and that case is handled explicitly in `ToolBar.handleDPIChange` (which refreshes the image lists and calls `updateImages` per item). In fact, the four widgets that release their image list on a DPI change are exactly `TabFolder`, `Table`, `Tree`, and `ToolBar`, and all four are covered

bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/ToolBarWin32Tests.java
```diff
@@ -0,0 +1,183 @@
...
+import org.junit.jupiter.api.*;
+import org.junit.jupiter.api.extension.*;
+
+/**
+ * Windows-specific tests for {@link ToolBar}'s image handling across monitor
+ * zoom changes.
+ */
+@ExtendWith(PlatformSpecificExecutionExtension.class)
+public class ToolBarWin32Tests {
+
```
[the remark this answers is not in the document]
HeikoKlare: Done

bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/ToolBarWin32Tests.java
```diff
@@ -0,0 +1,183 @@
...
+			}
+		}
+		return condition.getAsBoolean();
+	}
+
+	private static boolean iconsRenderOwnColor(ToolBar bar, ToolItem[] items, RGB[] colors) {
+		Image snapshot = renderToolBar(bar);
+		try {
+			// items[0] has no image (the hole), so start at 1.
+			for (int i = 1; i < items.length; i++) {
```
r-mennig: This is caller-dependent, which makes it hard to reuse this method if further tests are added at some later point.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/ToolBar.java
```diff
@@ -1738,19 +1738,19 @@ record ToolItemData(ToolItem toolItem, TBBUTTON button) {
...
+	setDisabledImageList(getDisabledImageList());
+	setHotImageList(getHotImageList());
 	for (int i = 0; i < itemCount; i++) {
 		ToolItem item = toolItems[i];
 		// If the separator is used with a control, we must reset the size to the cached value,
 		// cause windows will treat the separator as normal separator and shrinks it accordingly
 		if ((item.style & SWT.SEPARATOR) != 0 && item.getControl() != null) {
 			item.setWidth(seperatorWidth[i]);
 		}
+		toolItems[i].updateImages(toolItems[i].getEnabled() && getEnabled());
```
r-mennig: Does `this.getEnabled()` really needs to be re-checked on each loop?

r-mennig (approved):

bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/widgets/ToolBarWin32Tests.java
HeikoKlare: Indeed. I have moved this context-specific restriction to the caller site, passing a set of records for tool items and their expected colors to this general, reusable method instead.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/ToolBar.java
HeikoKlare: Goo catch. No, that's not necessary. It was just taken from other calling places without considering that when calling this method for every item of a tool bar, we can just retrieve the tool bar enablement state once and then reuse it. I have adapted the code accordingly.

PR #3471, opened 2026-07-29
[Win32] Keep tool bar normal, hot and disabled image lists index-aligned

HeikoKlare:
> [!IMPORTANT]
> This change is based on and should thus be merged after:
> - https://github.com/eclipse-platform/eclipse.platform.swt/pull/3504
> - https://github.com/eclipse-platform/eclipse.platform.swt/pull/3508
> - https://github.com/eclipse-platform/eclipse.platform.swt/pull/3513
> - https://github.com/eclipse-platform/eclipse.platform.swt/pull/3515

### Problem

On Windows a `ToolItem` could render **another item's hot (hover) or disabled icon** — no multiple monitors and no image disposal required. Setting an image, then a hot image, then clearing the image while adding a new item to the tool bar is already enough to trigger it.

A tool bar keeps three native image lists (normal, hot, disabled), but each button stores a **single** image index that addresses all three lists at once. The three lists therefore have to stay index-aligned. Two code paths in `ToolItem.updateImages` broke that invariant:

- **Clearing the normal image while a hot image is still set** freed the normal and disabled slots but kept the hot slot occupied. A later item that reused the freed normal slot then rendered the first item's stale hot icon on hover.
- **Adding a first image** appended it to each list with three independent `add()` calls. Once the lists' free-slot patterns diverged (as caused by the first defect), those calls could return different indices. Only the normal list's index is written to the button, so its slot could end up pointing at another item's hot or disabled image.

Contributes to #3466.

### Fix

Enforce one index per item across all three lists:

- Free the hot slot too when the normal image is cleared, so the hot list's free-slot pattern no longer diverges from the normal and disabled lists.
- Derive the slot index **once** from the normal list and write the hot and disabled lists at that exact index (instead of relying on three independent `add()` calls returning matching indices). `ToolBar.updateOrientation` adopts the same discipline.

This is implemented with a small `ImageList.putAt(index, image)` that stores an image at a caller-chosen index (appending when the index equals the current size, otherwise replacing or, for a `null` image, clearing the slot). `add(image)` is now implemented on top of it by first computing the target index.

### Reproduction

Run the snippet below and **hover the mouse over the second (blue) button**:

- **Before the fix:** the blue button turns **green** — it renders the first item's leftover hot image, even though it has no hot image of its own.
- **After the fix:** the blue button stays **blue**.

**Before (wrong behavior):**
image

**After (corrected behavior):**
image

```java
/*
 * Reproduction for a Win32 tool bar item rendering another item's hot (hover)
 * icon, without any multiple monitors or image disposal involved.
 *
 * Steps performed below:
 *   - item0 gets a RED normal image and a GREEN hot image
 *   - item0's image is cleared (setImage(null)) while the hot image stays set
 *   - item1 gets a BLUE normal image and NO hot image
 *
 * Hover the mouse over the second (blue) button:
 *   Expected (fixed): it stays BLUE (item1 has no hot image).
 *   Bug:              it turns GREEN, i.e. item1 renders item0's leftover hot image.
 */
import org.eclipse.swt.SWT;
import org.eclipse.swt.graphics.RGB;
import org.eclipse.swt.widgets.ToolBar;
import org.eclipse.swt.widgets.ToolItem;

public class ToolItemHotImageReproduction {

		Image red = solidImage(display, new RGB(220, 40, 40));
		Image green = solidImage(display, new RGB(40, 180, 40));
		Image blue = solidImage(display, new RGB(40, 40, 220));

		shell.setText("ToolItem hot image reproduction");

		ToolBar bar = new ToolBar(shell, SWT.FLAT);

		ToolItem item0 = new ToolItem(bar, SWT.PUSH);
		item0.setImage(red);
		item0.setHotImage(green); // item0 shows GREEN on hover
		item0.setImage(null);     // clear item0's image; its hot image stays set

		ToolItem item1 = new ToolItem(bar, SWT.PUSH);
		item1.setImage(blue);     // item1: BLUE, no hot image

		Label hint = new Label(shell, SWT.WRAP);
		hint.setText("Hover the mouse over the second (blue) button.\n"
				+ "Expected: it stays BLUE (it has no hot image).\n"
				+ "Bug: it turns GREEN (item0's leftover hot image).");

		shell.setSize(360, 150);
		shell.open();
		while (!shell.isDisposed()) {
		}

		red.dispose();
		green.dispose();
		blue.dispose();
		display.dispose();
	}

	private static Image solidImage(Display display, RGB rgb) {
		Image image = new Image(display, 16, 16);
		GC gc = new GC(image);
		Color color = new Color(display, rgb);
		gc.setBackground(color);
		gc.fillRectangle(image.getBounds());
		gc.dispose();
		return image;
	}
}
```

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/ImageList.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/ImageList.java
@@ -55,6 +55,27 @@ public int add (Image image) {
 		if (imageAtIndex == null) break;
 		index++;
 	}
+	putAt (index, image);
+	return index;
+}
+
+/**
+ * Stores {@code image} at the given {@code index}.
+ * <ul>
+ * <li>If {@code image} is non-null and {@code index} equals the current size, a
+ * new slot is appended.</li>
+ * <li>If {@code index} is within the current size, the slot is replaced (or
+ * cleared when {@code image} is null).</li>
+ * <li>If {@code index} is out of range (&lt; 0 or &gt; current size), this
+ * method does nothing.</li>
+ * </ul>
+ */
+public void putAt (int index, Image image) {
+	int count = OS.ImageList_GetImageCount (handle);
+	if (index != count || image == null) {
+		put (index, image);
+		return;
+	}
 	if (count == 0) {
 		Rectangle bounds = image.getBounds();
 		width = bounds.width;
@@ -68,7 +89,6 @@ public int add (Image image) {
 		images = newImages;
 	}
 	images [index] = image;
-	return index;
 }
 
 private Image getOrClearIfDisposed(int index) {
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/ImageList.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/ImageList.java
@@ -55,6 +55,11 @@ public int add (Image image) {
 		if (imageAtIndex == null) break;
 		index++;
 	}
+	put (index, image);
+	return index;
+}
+
+private void append (int index, Image image, int count) {
 	if (count == 0) {
 		Rectangle bounds = image.getBounds();
 		width = bounds.width;
@@ -68,7 +73,6 @@ public int add (Image image) {
 		images = newImages;
 	}
 	images [index] = image;
-	return index;
 }
 
 private Image getOrClearIfDisposed(int index) {
@@ -386,9 +390,19 @@ public int indexOf (Image image) {
 	return -1;
 }
 
+/**
+ * Stores the given image at the given index, replacing whatever is stored at that index. Passing
+ * no image clears the index. The index may also address the slot right after the last one, in
+ * which case a new slot is appended for the given image. Nothing happens for any other index
+ * outside the list's current size.
+ */
 public void put (int index, Image image) {
 	if ((0 <= index && index < images.length) && (images [index] == image)) return;
 	int count = OS.ImageList_GetImageCount (handle);
+	if (index == count && image != null) {
+		append (index, image, count);
+		return;
+	}
 	if (!(0 <= index && index < count)) return;
 	if (image != null) setForAllHandles(index, image, count);
 	images [index] = image;
--- a/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/internal/ImageListTests.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Tests/win32/org/eclipse/swt/internal/ImageListTests.java
@@ -132,4 +132,158 @@ public class ImageListTests {
 		}
 	}
 
+	@Test
+	public void testAddReturnsConsecutiveIndicesForConsecutiveImages() {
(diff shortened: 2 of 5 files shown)


bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/ImageList.java
```diff
@@ -55,6 +55,24 @@ public int add (Image image) {
...
 	}
+	putAt (index, image);
+	return index;
+}
+
+/**
+ * Stores {@code image} at the given {@code index}, appending a new slot when
+ * {@code index} equals the current size and replacing (or, for a {@code null}
+ * image, clearing) the existing slot otherwise.
+ */
```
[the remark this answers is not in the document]
HeikoKlare: Done, replaced the code comments with the proposed Javadoc giving the same explanations.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/internal/ImageList.java
```diff
@@ -55,6 +55,27 @@ public int add (Image image) {
...
+ * <ul>
+ * <li>If {@code image} is non-null and {@code index} equals the current size, a
+ * new slot is appended.</li>
+ * <li>If {@code index} is within the current size, the slot is replaced (or
+ * cleared when {@code image} is null).</li>
+ * <li>If {@code index} is out of range (&lt; 0 or &gt; current size), this
+ * method does nothing.</li>
+ * </ul>
+ */
+public void putAt (int index, Image image) {
```
r-mennig: `ImageList` now contains both a `public putAt(int index, Image image)` and a `public put(int index, Image image)` method where `putAt` allows appending and `put` does not. So maybe `putOrAppend` would be a better name for this method?

HeikoKlare: That a valid concern. I had originally named that method `putOrAppend` before I changed it, but I agree that given the already existing `put`, this is very misleading. Rather than renaming, I have now merged the two: `putAt` is gone and `put` now also accepts the index right after the last slot, appending there.

`put` previously ignored such an index silently, which is the surprising behaviour of the two. `Map.put` inserts when the key is absent, so store-or-append is closer to what the name suggests. `add` keeps its shape (find the first free slot, then `put`), which now describes exactly what it does.

The change is safe for existing callers: the sole caller outside `ToolBar`/`ToolItem` is `TabFolder`, whose index comes from `indexOf` and is therefore always inside the list, i.e., it cannot reach the new append path. Every other caller passes a button's `iImage`, likewise always in range.

I have also rebased this change/PR on a sequence of refactorings for properly encapsulating and aligning the three ImageLists of a ToolBar inside the ToolBar class and a specific three-list container. Only the last of the current commits compared to master covers the actual change of the PR.
I have also added a bunch of tests for preserving the enhanced `ImageList` behavior.

r-mennig: Thank you for the update! This API looks a lot cleaner now

r-mennig (approved):
This change looks good to me. Combining the image lists into one abstraction simplifies this code a lot, reduces the overall code size and makes the API less error-prone. Thank you for this refactoring!

PR #3479, opened 2026-07-31
[Win32] Fix BrowserFunction missing on first page

### LINKED ISSUE #3370: Test_org_eclipse_swt_browser_Browser.test_BrowserFunction_availableOnLoad_concurrentInstances_issue20 failed in I-build
See: https://download.eclipse.org/eclipse/downloads/drops4/I20260607-1800/testresults/html/org.eclipse.swt.tests_ep441I-unit-win32-x86_64-java21_win32.win32.x86_64_21.html

```
	options is not defined

org.eclipse.swt.SWTException: options is not defined
at org.eclipse.swt.browser.Edge.evaluateInternal(Edge.java:1003)
at org.eclipse.swt.browser.Edge.evaluate(Edge.java:976)
at org.eclipse.swt.browser.WebBrowser.evaluate(WebBrowser.java:405)
at org.eclipse.swt.browser.Browser.evaluate(Browser.java:735)
at org.eclipse.swt.browser.Browser.evaluate(Browser.java:684)
at org.eclipse.swt.tests.junit.Test_org_eclipse_swt_browser_Browser.lambda$157(Test_org_eclipse_swt_browser_Browser.java:3026)
at org.eclipse.swt.browser.ProgressListener$2.completed(ProgressListener.java:97)
at org.eclipse.swt.browser.Edge.lambda$44(Edge.java:1218)
at org.eclipse.swt.widgets.Synchronizer.runAsyncMessages(Synchronizer.java:131)
at org.eclipse.swt.widgets.Display.runAsyncMessages(Display.java:4136)
at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:3752)
at org.eclipse.swt.tests.junit.Test_org_eclipse_swt_browser_Browser.waitForPassCondition(Test_org_eclipse_swt_browser_Browser.java:3139)
at org.eclipse.swt.tests.junit.Test_org_eclipse_swt_browser_Browser.waitForPassCondition(Test_org_eclipse_swt_browser_Browser.java:3105)
at org.eclipse.swt.tests.junit.Test_org_eclipse_swt_browser_Browser.test_BrowserFunction_availableOnLoad_concurrentInstances_issue20(Test_org_eclipse_swt_browser_Browser.java:3035)
at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
Suppressed: java.lang.RuntimeException: AwtScreenshot VM finished with exit code 1:
Exception in thread "main" java.awt.AWTError: no screen devices
at java.desktop/sun.awt.Win32GraphicsEnvironment.getDefaultScreenDevice(Win32GraphicsEnvironment.java:101)
at java.desktop/java.awt.Robot.<init>(Robot.java:135)
at org.eclipse.test.AwtScreenshot.main(AwtScreenshot.java:43)

at org.eclipse.test.AwtScreenshot.dumpAwtScreenshot(AwtScreenshot.java:113)
at org.eclipse.test.Screenshots.takeScreenshot(Screenshots.java:102)
at org.eclipse.swt.tests.junit.Test_org_eclipse_swt_widgets_Widget.lambda$1(Test_org_eclipse_swt_widgets_Widget.java:79)
at java.base/java.util.Optional.ifPresent(Optional.java:178)
at org.eclipse.swt.tests.junit.Test_org_eclipse_swt_widgets_Widget.lambda$0(Test_org_eclipse_swt_widgets_Widget.java:75)
... 2 more 
```

Nothing relevant in the workspace log: https://download.eclipse.org/eclipse/downloads/drops4/I20260607-1800/testresults/ep441I-unit-win32-x86_64-java21_win32.win32.x86_64_21/org.eclipse.swt.tests.junit.AllBrowserTests.log

```
!SESSION 2026-06-07 22:48:22.360 -----------------------------------------------
eclipse.buildId=4.41.0.I20260607-1800
java.version=21.0.5
java.vendor=Eclipse Adoptium
BootLoader constants: OS=win32, ARCH=x86_64, WS=win32, NL=en_US
Framework arguments:  -application org.eclipse.equinox.p2.director -flavor tooling -installIUs org.eclipse.swt.tests,org.eclipse.test,org.eclipse.ant.optional.junit,org.eclipse.test.performance,org.eclipse.test.performance.win32 -repository file:C:\Users\genie.releng\workspace\AutomatedTests\ep441I-unit-win32-x86_64-java21/workarea/I20260607-1800/eclipse-testing/testRepo
Command-line arguments:  -application org.eclipse.equinox.p2.director -consoleLog -debug -flavor tooling -installIUs org.eclipse.swt.tests,org.eclipse.test,org.eclipse.ant.optional.junit,org.eclipse.test.performance,org.eclipse.test.performance.win32 -repository file:C:\Users\genie.releng\workspace\AutomatedTests\ep441I-unit-win32-x86_64-java21/workarea/I20260607-1800/eclipse-testing/testRepo

!ENTRY org.eclipse.equinox.p2.director 0 0 2026-06-07 22:48:31.350
!MESSAGE Overall install request is satisfiable
!SUBENTRY 1 org.eclipse.equinox.p2.director 0 0 2026-06-07 22:48:31.350
!MESSAGE Add request for Eclipse Automated Testing 3.6.600.v20260520-0751 (org.eclipse.test 3.6.600.v20260520-0751) is satisfiable
!SUBENTRY 1 org.eclipse.equinox.p2.director 0 0 2026-06-07 22:48:31.350
!MESSAGE Add request for Eclipse JUnit Fragment for Ant used in Eclipse Test Framework 3.3.400.v20230714-1309 (org.eclipse.ant.optional.junit 3.3.400.v20230714-1309) is satisfiable
!SUBENTRY 1 org.eclipse.equinox.p2.director 0 0 2026-06-07 22:48:31.350
!MESSAGE Add request for Performance Monitor for Windows 3.1.600.v20230714-1309 (org.eclipse.test.performance.win32 3.1.600.v20230714-1309) is satisfiable
!SUBENTRY 1 org.eclipse.equinox.p2.director 0 0 2026-06-07 22:48:31.350
!MESSAGE Add request for Performance Test Plug-in 3.21.300.v20260602-1216 (org.eclipse.test.performance 3.21.300.v20260602-1216) is satisfiable
!SUBENTRY 1 org.eclipse.equinox.p2.director 0 0 2026-06-07 22:48:31.350
!MESSAGE Add request for Eclipse SWT Tests 3.108.100.v20260602-1543 (org.eclipse.swt.tests 3.108.100.v20260602-1543) is satisfiable
```

HeikoKlare:
Fixes #3370 — the sporadic failure of `test_BrowserFunction_availableOnLoad_concurrentInstances_issue20` on the Edge/WebView2 backend.

Builds on top of #3478 (BrowserFunction dispose/redefine regression tests) and employs the regression tests added there — #3478 should be merged first.

## Problem

A `BrowserFunction` created while an Edge browser is still initializing (with a navigation queued via `setUrl`/`setText`) could be missing on the first loaded page. `AddScriptToExecuteOnDocumentCreated` was issued only *after* the queued navigation and raced with document creation, so the function was sometimes not registered when the page's document was created — surfacing as `SWTException: <function> is not defined`.

## Fix

Register the document-created scripts of all pending `BrowserFunction`s inside `WebViewProvider.initializeWebView()`, before completing the initialization future (which synchronously runs the queued navigation). This is the only point that is reliably "after init, before the first navigation", since the `CompletableFuture` navigation chain cannot be preempted retroactively.

Registration is now issued asynchronously: what makes a function land on a page is *issuing* the registration before the navigation is issued to WebView2, not *waiting* for its completion. This also removes the previous `inCallback > 0` deferral (an async call cannot deadlock inside a WebView2 callback) and the separate synchronous registration method, leaving a single registration method and a single branch in `createFunction`. `deregisterFunction` stays correct for a registration whose script ID has not arrived yet (its async completion detects the removed function and removes the script itself).

## Tests

Adds regression tests for function creation during concurrent initialization: creating a function from inside a callback, availability before the first page's inline scripts run, and multiple functions registered before initialization completes. These complement the dispose/redefine regression tests from #3478.

## Reproducer

The underlying race is a sub-millisecond window and cannot be forced deterministically through the public API with a single browser. The following standalone reproducer makes it deterministic in practice by stressing concurrent initialization: before the fix it reports `options is not defined` within a few iterations; after the fix it completes all iterations cleanly.

<details>
<summary>Standalone reproducer (Windows/Edge)</summary>

```java
 * Copyright (c) 2026 Contributors to the Eclipse Foundation.
 *
 *
package org.eclipse.swt.tests.junit;

import java.util.ArrayList;
import java.util.List;
import java.util.concurrent.atomic.AtomicReference;

import org.eclipse.swt.SWT;
import org.eclipse.swt.SWTException;
import org.eclipse.swt.browser.Browser;
import org.eclipse.swt.browser.BrowserFunction;
import org.eclipse.swt.browser.ProgressAdapter;
import org.eclipse.swt.browser.ProgressEvent;

/**
 * Standalone reproducer for the Edge/WebView2 BrowserFunction registration race fixed by this change.
 *
 * <p>A {@link BrowserFunction} created while a {@link Browser} is still initializing (and while a
 * navigation queued via {@code setUrl}/{@code setText} is pending) must be available on the first
 * loaded page. Before the fix, {@code AddScriptToExecuteOnDocumentCreated} was issued <em>after</em>
 * the navigation, so it raced with document creation and the function was sometimes undefined,
 * producing {@code SWTException: options is not defined}.
 *
 * <p>The underlying race is a sub-millisecond timing window, so it cannot be forced deterministically
 * through the public SWT API with a single browser. This reproducer makes it <em>deterministic in
 * practice</em> by stressing the exact scenario: it creates several browsers that initialize
 * concurrently (their initialization pumps the shared event loop and perturbs each other's timing),
 * repeated over many iterations. Before the fix it reports failures within the first few iterations;
 * after the fix it completes all iterations with zero failures.
 *
 * <p>Run it standalone via {@link #main(String[])} on Windows (Edge backend), or call
 * {@link #reproduce(Display, int, int)} from a test.
 */
public final class BrowserFunctionConcurrentInitReproducer {

	private BrowserFunctionConcurrentInitReproducer() {
	}

	/** Number of browsers initialized concurrently in a single iteration. */
	private static final int DEFAULT_BROWSERS_PER_ITERATION = 8;
	/** Number of iterations to run when stressing the scenario. */
	private static final int DEFAULT_ITERATIONS = 50;
	/** Maximum time to wait for all browsers of one iteration to resolve. */
	private static final int ITERATION_TIMEOUT_MILLIS = 30_000;

	/** Result of a single browser within an iteration. */
	private enum Result {
		PENDING, AVAILABLE, MISSING
	}

		try {
			Browser probe = new Browser(shell, SWT.NONE);
			boolean isEdge = "edge".equals(probe.getBrowserType());
			probe.dispose();
			if (!isEdge) {
				System.out.println("This reproducer targets the Edge/WebView2 backend; current backend is '"
						+ probe.getBrowserType() + "'. Nothing to reproduce.");
				return;
			}

			int failures = reproduce(display, DEFAULT_ITERATIONS, DEFAULT_BROWSERS_PER_ITERATION);
			if (failures > 0) {
				System.out.println("\nREPRODUCED: 'options' was missing in " + failures
						+ " case(s). This indicates the AddScriptToExecuteOnDocumentCreated race.");
				System.exit(1);
			} else {
				System.out.println("\nNo failures observed across " + DEFAULT_ITERATIONS + " iterations of "
						+ DEFAULT_BROWSERS_PER_ITERATION + " concurrent browsers. The race appears fixed.");
			}
		} finally {
			display.dispose();
		}
	}

	/**
	 * Runs the concurrent-initialization scenario and returns the number of browsers for which the
	 * BrowserFunction was <em>not</em> available on the loaded page (i.e. reproductions of the bug).
	 *
	 * @param display            the display to run on (must be the calling thread's display)
	 * @param iterations         number of iterations to run
	 * @param browsersPerIteration number of browsers initialized concurrently per iteration
	 * @return the total number of reproductions observed
	 */
	public static int reproduce(Display display, int iterations, int browsersPerIteration) {
		int totalFailures = 0;
		for (int iteration = 0; iteration < iterations; iteration++) {
			totalFailures += runSingleIteration(display, iteration, browsersPerIteration);
			if (totalFailures > 0) {
				// Fail fast: one reproduction is enough to demonstrate the defect.
				break;
			}
		}
		return totalFailures;
	}

	private static int runSingleIteration(Display display, int iteration, int browsersPerIteration) {
		List<Browser> browsers = new ArrayList<>();
		List<Result[]> results = new ArrayList<>();
		List<AtomicReference<SWTException>> exceptions = new ArrayList<>();

		try {
			for (int i = 0; i < browsersPerIteration; i++) {
				final Result[] result = { Result.PENDING };
				final AtomicReference<SWTException> exception = new AtomicReference<>();
				results.add(result);
				exceptions.add(exception);

				// Deliberately do NOT wait for initialization: keep all browsers initializing
				// concurrently, which is the timing that exposes the race (issue #20).
				final Browser browser = new Browser(shell, SWT.NONE);
				browser.setUrl("about:blank");
				new BrowserFunction(browser, "options") {
					@Override
					public Object function(Object[] arguments) {
						return null;
					}
				};
				browser.addProgressListener(new ProgressAdapter() {
					@Override
					public void completed(ProgressEvent event) {
						try {
							browser.evaluate("options();");
							result[0] = Result.AVAILABLE;
						} catch (SWTException ex) {
							result[0] = Result.MISSING;
							exception.set(ex);
						}
					}
				});
				browsers.add(browser);
			}

			shell.open();
			waitUntilResolved(display, results);

			int failures = 0;
			for (int i = 0; i < results.size(); i++) {
				Result r = results.get(i)[0];
				if (r != Result.AVAILABLE) {
					failures++;
					SWTException ex = exceptions.get(i).get();
					System.out.println("Iteration " + iteration + ", browser " + i + ": " + r
							+ (ex != null ? " (" + ex.getMessage() + ")" : " (timed out)"));
				}
			}
			if (failures == 0) {
				System.out.println("Iteration " + iteration + ": OK (" + browsersPerIteration + " browsers)");
			}
			return failures;
		} finally {
			for (Browser browser : browsers) {
				if (!browser.isDisposed()) {
					browser.dispose();
				}
			}
			shell.dispose();
			// Let Edge process pending disposal work between iterations.
			while (display.readAndDispatch()) {
				// drain
			}
		}
	}

	private static void waitUntilResolved(Display display, List<Result[]> results) {
		long deadline = System.currentTimeMillis() + ITERATION_TIMEOUT_MILLIS;
		// Periodically wake the display so the deadline is honored even if no browser event arrives.
		Runnable[] wake = new Runnable[1];
		wake[0] = () -> {
			if (!display.isDisposed()) {
				display.timerExec(100, wake[0]);
			}
		};
		display.timerExec(100, wake[0]);
		try {
			while (System.currentTimeMillis() < deadline && !allResolved(results)) {
					display.sleep();
				}
			}
		} finally {
			display.timerExec(-1, wake[0]);
		}
	}

	private static boolean allResolved(List<Result[]> results) {
		for (Result[] result : results) {
			if (result[0] == Result.PENDING) {
				return false;
			}
		}
		return true;
	}
}
```
</details>

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
@@ -85,7 +85,13 @@ class Edge extends WebBrowser {
 	boolean inNewWindow;
 	private boolean inEvaluate;
 	HashMap<Long, LocationEvent> navigations = new HashMap<>();
-	/** Maps BrowserFunction index to the script ID from AddScriptToExecuteOnDocumentCreated. */
+	/**
+	 * Maps BrowserFunction index to the script ID from AddScriptToExecuteOnDocumentCreated. An entry
+	 * is added synchronously when the (asynchronous) registration is issued, with a {@code null}
+	 * value until the registration completes and provides the script ID. Therefore a contained key
+	 * denotes a function whose script is already registered, which ensures that a function's script
+	 * is registered exactly once.
+	 */
 	private final Map<Integer, String> functionScriptIds = new HashMap<>();
 	private int ignoreGotFocus;
 	private boolean ignoreFocusIn;
@@ -386,6 +392,11 @@ class WebViewProvider {
 		webViewWrapper.webView_11 = initializeWebView_11(webView);
 		webViewWrapper.webView_12 = initializeWebView_12(webView);
 		webViewWrapper.webView_13 = initializeWebView_13(webView);
+		// Register the scripts of all BrowserFunctions created during initialization
+		// before completing the future below, as that synchronously runs queued
+		// navigation tasks (e.g. from setUrl()/setText()) and thus may already create
+		// the first document
+		registerPendingFunctionScripts(webView);
 		boolean success = webViewWrapperFuture.complete(webViewWrapper);
 		// Release the webViews if the webViewWrapperFuture has already timed out and completed exceptionally
 		if(!success && webViewWrapperFuture.isCompletedExceptionally()) {
@@ -397,6 +408,10 @@ class WebViewProvider {
 
 	private void abortInitialization() {
 		webViewWrapperFuture.cancel(true);
+	}
+
+	boolean isInitialized() {
+		return webViewWrapperFuture.isDone();
 	}
 
 	void releaseWebView() {
@@ -1839,38 +1854,77 @@ public boolean setUrl(String url, String postData, String[] headers) {
 }
 
 /**
- * Registers the function script persistently via AddScriptToExecuteOnDocumentCreated so it is
- * injected on every future document creation before any page scripts run, avoiding the race
- * condition between async function injection and navigation completion.
- * If called while inside a WebView2 callback, the persistent registration is deferred via
- * {@link Display#asyncExec(Runnable)} so it completes once the callback returns.
+ * Registers a BrowserFunction persistently via AddScriptToExecuteOnDocumentCreated so it is
+ * injected on every future document creation before any page scripts run.
+ * <p>
+ * The registration is issued immediately, but without waiting for its (asynchronous) completion:
+ * what makes a function available on a page is <em>issuing</em> the registration before the
+ * navigation that creates the document is issued to WebView2 - not waiting for the registration to
+ * complete. Since nothing is blocked on the completion, this can also safely be issued from within
+ * a WebView2 callback without risking a deadlock.
+ * <p>
+ * If the browser is not yet initialized when the function is created, the registration is issued by
+ * {@link WebViewProvider#initializeWebView(ICoreWebView2Controller)} (via
+ * {@link #registerPendingFunctionScripts(ICoreWebView2)}) before the first navigation, so functions
+ * created concurrently with initialization are available on the first loaded page. Note that
+ * {@code super.createFunction(function)} below may itself complete the initialization, since it
+ * executes the function's script and thus pumps the event loop until the browser is initialized.
+ * Independent of which of the two registers the script, {@link #registerFunctionScript(ICoreWebView2, int, String)}
+ * ensures it is registered exactly once.
  * See <a href="https://github.com/eclipse-platform/eclipse.platform.swt/issues/20">issue #20</a>.
  */
 @Override
 public void createFunction(BrowserFunction function) {
 	super.createFunction(function);
-	int functionIndex = function.index;
-	String functionString = function.functionString;
-	if (inCallback > 0) {
-		// Cannot wait for a callback result while already inside a WebView2 callback;
-		// defer the persistent registration to after the callback completes.
-		browser.getDisplay().asyncExec(() -> {
-			if (browser.isDisposed() || !functions.containsKey(functionIndex)) return;
-			registerFunctionScript(functionIndex, functionString);
-		});
-		return;
+	if (webViewProvider.isInitialized()) {
+		registerFunctionScript(webViewProvider.getWebView(false), function.index, function.functionString);
 	}
-	registerFunctionScript(functionIndex, functionString);
 }
 
-private void registerFunctionScript(int functionIndex, String functionString) {
-	String[] scriptId = new String[1];
-	callAndWait(scriptId, completion ->
-		webViewProvider.getWebView(false).AddScriptToExecuteOnDocumentCreated(
-			stringToWstr(functionString), completion.getAddress()));
-	if (scriptId[0] != null) {
-		functionScriptIds.put(functionIndex, scriptId[0]);
+private void registerPendingFunctionScripts(ICoreWebView2 webView) {
+	for (Map.Entry<Integer, BrowserFunction> entry : functions.entrySet()) {
+		BrowserFunction function = entry.getValue();
+		if (function.functionString != null) {
+			registerFunctionScript(webView, entry.getKey(), function.functionString);
+		}
 	}
+}
+
+/**
+ * Issues the registration of a function's document-created script on the given WebView without
+ * blocking for the asynchronous completion. Functions whose script is already registered are
+ * skipped, so it does not matter whether the registration is issued by
+ * {@link #createFunction(BrowserFunction)} or {@link #registerPendingFunctionScripts(ICoreWebView2)}.
+ * <p>
+ * The resulting script ID is stored in {@link #functionScriptIds} once the completion callback
+ * fires; if the function was deregistered again in the meantime, the script is removed right away
+ * instead, so an immediately following deregistration does not leak the script.
+ */
+private void registerFunctionScript(ICoreWebView2 webView, int functionIndex, String functionString) {
+	if (functionScriptIds.containsKey(functionIndex)) {
+		// The script of this function has already been registered
+		return;
+	}
+	functionScriptIds.put(functionIndex, null);
+	IUnknown completion = newCallback((result, scriptIdPointer) -> {
+		if ((int) result == COM.S_OK) {
+			String scriptId = wstrToString(scriptIdPointer, false);
+			if (functions.containsKey(functionIndex)) {
+				functionScriptIds.put(functionIndex, scriptId);
+			} else if (!browser.isDisposed()) {
+				webView.RemoveScriptToExecuteOnDocumentCreated(stringToWstr(scriptId));
+			}
+		} else {
+			functionScriptIds.remove(functionIndex);
+		}
+		return COM.S_OK;
+	});
+	int hr = webView.AddScriptToExecuteOnDocumentCreated(stringToWstr(functionString), completion.getAddress());
+	if (hr != OS.S_OK) {
+		functionScriptIds.remove(functionIndex);
+		System.err.println("Registering browser function failed with result " + hr + " for function: " + functionString);
+	}
+	completion.Release();
 }
 
 @Override
@@ -1881,6 +1935,8 @@ void deregisterFunction(BrowserFunction function) {
 		webViewProvider.getWebView(true).RemoveScriptToExecuteOnDocumentCreated(
 			stringToWstr(scriptId));
 	}
+	// If scriptId == null, an asynchronous registration has not stored its ID yet; its completion
+	// callback detects the now-removed function (via the functions map) and removes the script itself.
 }
 
 }
--- a/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
+++ b/tests/org.eclipse.swt.tests/JUnit Tests/org/eclipse/swt/tests/junit/Test_org_eclipse_swt_browser_Browser.java
@@ -3046,6 +3046,133 @@ public void test_BrowserFunction_availableOnLoad_concurrentInstances_issue20() {
 }
 
 /**
+ * Regression test: a BrowserFunction created from <em>inside</em> another BrowserFunction's
+ * callback must be registered and available on a page that is navigated to from within that same
+ * callback.
+ * <p>
+ * On the Edge/WebView2 backend this exercises function creation while a WebView2 callback is on the
+ * stack. Registration must be issued (before the navigation queued in the same callback) without
+ * blocking, since blocking inside a callback would deadlock.
+ */
+@Test
+public void test_BrowserFunction_createFunctionInsideCallback() {
+	assumeTrue(isEdge, "BrowserFunction availability before the page's inline scripts is specific to the Edge/WebView2 implementation");
+	AtomicBoolean innerCalled = new AtomicBoolean(false);
+
+	// 'inner' is only created when 'outer' is invoked from JavaScript, i.e. inside a callback.
+	class Inner extends BrowserFunction {
+		Inner() {
+			super(browser, "inner");
+		}
+		@Override
+		public Object function(Object[] arguments) {
+			innerCalled.set(true);
+			return null;
+		}
+	}
+	class Outer extends BrowserFunction {
+		Outer() {
+			super(browser, "outer");
+		}
+		@Override
+		public Object function(Object[] arguments) {
+			new Inner(); // create a new BrowserFunction from inside a callback
+			// Navigate to a page whose inline script calls the just-created function.
+			browser.setText("<html><body><script>inner();</script></body></html>");
+			return null;
+		}
+	}
+	new Outer();
+
+	// Trigger outer() once, after the first page has loaded.
+	AtomicBoolean outerTriggered = new AtomicBoolean(false);
+	browser.addProgressListener(completedAdapter(e -> {
+		if (outerTriggered.compareAndSet(false, true)) {
+			browser.execute("outer();");
+		}
+	}));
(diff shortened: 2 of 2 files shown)


HeikoKlare:
@sratz I may have finally found the last piece for reliable browser function registration with Edge/WebView2. This PR revises the according implementation to address all registration scenarios (during browser initialization, during page navigation or when page is already loaded) in a more simple way than just extending for the not-yet-addressed scenario. Could you please have a look at this fix proposal? I have iterated the fix with an AI agent for quite some time already but would be good to have someone else with experience in this piece of code have a thorough look as well. Thank you in advance!
I do not think this is something we should still merge for 2026-09 but rather early in 2026-12 development, so that we have sufficient time for testing.

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -1839,40 +1848,66 @@ public boolean setUrl(String url, String postData, String[] headers) {
...
+			String scriptId = wstrToString(scriptIdPointer, false);
+			if (functions.containsKey(functionIndex)) {
+				functionScriptIds.put(functionIndex, scriptId);
+			} else if (!browser.isDisposed()) {
+				webView.RemoveScriptToExecuteOnDocumentCreated(stringToWstr(scriptId));
+			}
+		}
+		return COM.S_OK;
+	});
+	webView.AddScriptToExecuteOnDocumentCreated(stringToWstr(functionString), completion.getAddress());
```
r-mennig: Not checking the return value of `AddScriptToExecuteOnDocumentCreated` can lead to browser functions silently failing to register. This is made worse by the fact that those functions can be contributed from other code parts (via `createFunction`), so I think any registration failure should be handled here by throwing an exception.

```suggestion
		int hr = webView.AddScriptToExecuteOnDocumentCreated(stringToWstr(functionString), completion.getAddress());
	if (hr != COM.S_OK) error(SWT.ERROR_NO_HANDLES, hr);
```

(Maybe something other that `SWT.ERROR_NO_HANDLES` would be better here)

r-mennig (commented):
The changes look good to me. I've only found a minor issue regarding a missing return value check on `AddScriptToExecuteOnDocumentCreated`

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
HeikoKlare: That's a very correct concern. Unfortunately, there are problems with either of the callers of this method when throwing an error here:
1. The one caller is the immediate function registration when creating a BrowserFunction (via `new BrowserFunction()`). That constructor does not have any declared errors for the case that function registration fails. Introducing a new type of exception here can lead to consumers fail unexpectedly as they do not deal with that type of exception. Thus, this would be kind of API-breaking (though not in a very strict sense).
2. The other caller is the browser initialization finalization that registers a function that was created before initialization was finished. This initialization process must not fail because of a failing browser function registration as well, such that the error would have to be caught there again and the further error processing is unclear, as the initialization finalization may be triggered by any kind of consumer performing operations on the browser.

While just swallowing a failure is obviously not a good idea (even though we have done that so far as well), throwing an error does not seem feasible here.
I found these types of errors to potentially occur here:
- `E_INVALIDARG` if the script is null or the completion callback is invalid (both cannot happen)
- `E_POINTER` if a pointer argument is null (I do not see how it could happen with this call)
- The HRESULT for `ERROR_INVALID_STATE` if the web view was already closed (in which case failing function registration is not an issue at all).

So I am not sure if any relevant kind of error may occur here and since we have not seen reports about issues with function registration so far, it may be really seldom (if at all) that an error occurs here. For that, we may simply log the error for the sake of traceability and debuggability without throwing any error to avoid completely unexpected or unrecoverable application states. That's what I have implemented with my latest commit.

r-mennig: Thank you for the update and very detailed explanation!

On my initial comment, I've mistakenly assumed that  `org.eclipse.swt.browser.Edge.createFunction(BrowserFunction)` did not guard against registering an invalid / malformed function string which could lead to a silent/hard-to-debug problem in this method. Upon re-reading `createFunction(BrowserFunction)`, I now see that the super-call to `org.eclipse.swt.browser.WebBrowser.createFunction(BrowserFunction)` always ensures a correct function string.

As of now, the only way I can see `hr != COM.S_OK` being true here is if `org.eclipse.swt.browser.WebBrowser.createFunction(BrowserFunction)` would return an invalid function string. But this would be an implementation error (and not a caller passing an invalid argument as I've initially though), so just logging this should be fine.

bundles/org.eclipse.swt/Eclipse SWT Browser/win32/org/eclipse/swt/browser/Edge.java
```diff
@@ -1839,40 +1848,69 @@ public boolean setUrl(String url, String postData, String[] headers) {
...
-	String functionString = function.functionString;
-	if (inCallback > 0) {
-		// Cannot wait for a callback result while already inside a WebView2 callback;
-		// defer the persistent registration to after the callback completes.
-		browser.getDisplay().asyncExec(() -> {
-			if (browser.isDisposed() || !functions.containsKey(functionIndex)) return;
-			registerFunctionScript(functionIndex, functionString);
-		});
-		return;
+	if (alreadyInitialized) {
```
sratz: do we have a race condition here?

when `super.createFunction()` returns, we might be initialized in the meantime as the future is completed from an asynchronous callback.

HeikoKlare: Thank you for asking. The state can indeed change in between. In fact `super.createFunction()` itself executes the function's script, which pumps the event loop via `getWebView()` and thus usually completes the initialization right there. An asynchronous callback, however, cannot complete initialization, as the event queue will not yield any events when being processed from withing an asynchronous callback.

The result was still correct (the function is put into the `functions` map with its `functionString` before that happens, so `registerPendingFunctionScripts()` picks it up exactly once and the stale `false` is precisely the value we want), but I agree that this very difficult to understand.

I made the registration idempotent instead: https://github.com/eclipse-platform/eclipse.platform.swt/compare/3e6cd3530ec7865a520170b69e6c6ab8de21688d..7c05dd1f3af4b576956988d3ba401e8e249c7752
`registerFunctionScript()` now puts the function's index into `functionScriptIds` right when the registration is issued (with a `null` value until the async completion delivers the script ID), and skips functions already contained there. So it no longer matters whether `createFunction()` or `initializeWebView()` issues the registration, or how the two interleave. The script is registered exactly once, and the `isInitialized()` check is only there to avoid blocking in `getWebView()`. I think/hope this makes the logic easier to understand and evolve.

sratz: Ah, right, the event loop handling plays into that.

The new idempotent variant looks more understandable now!

And I suppose this will also safer in the future should anything in that even loop handling change.

sratz (approved):
lgtm for M1

PR #3485, opened 2026-08-05
[Build] Run API check and javadoc only once per operating system

### LINKED ISSUE #3484: Run API tools check and javadoc only once per OS
`.github/workflows/build.yml` passes `-Papi-check -Pjavadoc`, and `maven.yml`
calls that reusable workflow six times per pull request: three Linux jobs
(gtk3/x11, gtk4/x11, gtk4/wayland), one Windows job and two macOS jobs.

All three Linux jobs pass the identical `native: gtk.linux.x86_64` input.
Only `gtk` and `gdk_backend` differ, and those affect test execution, not the
Java API surface or the javadoc, so both checks are computed three times from
identical inputs.

Proposal: gate `-Papi-check` and `-Pjavadoc` behind a new boolean input on
`build.yml` and enable it for one job per OS. The per-OS split has to stay,
since SWT's API surface really does differ between the gtk, win32 and cocoa
fragments. Whether the two macOS architectures can collapse into one run
needs checking.

vogella:
The reusable build workflow always passed `-Papi-check -Pjavadoc`, so the API tools check and the javadoc generation ran in all six PR jobs. Neither depends on the GTK version, the GDK backend or the target architecture, only on the Java sources, so the three Linux jobs and the two macOS jobs were computing identical results. I verified the macOS case rather than assuming it: the two cocoa fragments compile the same source folders and their jars contain an identical set of 768 class files with identical sizes and an identical `.api_description` apart from the component id.

A new boolean input `api_check` (default false) now gates both profiles, and `maven.yml` enables it for exactly one job per operating system: gtk3/x11 on Linux, the single Windows job and aarch64 on macOS. That takes the API check and javadoc from six runs down to three, which cuts a few minutes off the slowest jobs without losing any coverage. Unlike most Eclipse repos, SWT does have a genuinely different API surface per gtk/win32/cocoa fragment, so the check deliberately stays per operating system instead of collapsing to a single global run.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/3484

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,6 +28,15 @@ on:
         type: boolean
         required: false
         default: false
+      api_check:
+        description: |
+          Run the API tools check and generate the javadoc (one of true, false)
+
+          Both only depend on the Java sources, so enable this for a single job
+          per operating system.
+        type: boolean
+        required: false
+        default: false
       gtk:
         description: |
           (Required on Linux only) GTK version to use (one of gtk3, gtk4)
@@ -108,7 +117,7 @@ jobs:
         --threads 1C
         -DforkCount=1
         '-Dnative=${{ inputs.native }}'
-        -Papi-check -Pjavadoc
+        ${{ inputs.api_check && '-Papi-check -Pjavadoc' || '' }}
         '-Dtycho.baseline.replace=none'
         --fail-at-end
         -DskipNativeTests=false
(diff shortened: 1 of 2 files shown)


akurtakov:
The compiler warnings are unrelated and TBH I have no clue why they appear (it's not the first time in my experience). Merging despite them.

PR #3491, opened 2026-08-07
[Win32] Add missing null check in Table tooltip positioning #3487

### LINKED ISSUE #3487: [Win32] Possible NPE handling WM_NOTIFY
**Describe the bug**
A customer reported following exception:
```
java.lang.NullPointerException: Cannot read field "x" because "adjustedTooltipBounds" is null
	at org.eclipse.swt.widgets.Table.positionTooltip(Table.java:7201)
	at org.eclipse.swt.widgets.Table.wmNotifyToolTip(Table.java:7164)
	at org.eclipse.swt.widgets.Table.wmNotify(Table.java:6504)
	at org.eclipse.swt.widgets.Control.WM_NOTIFY(Control.java:5605)
	at org.eclipse.swt.widgets.Control.windowProc(Control.java:5037)
	at org.eclipse.swt.widgets.Table.windowProc(Table.java:5814)
	at org.eclipse.swt.widgets.Display.windowProc(Display.java:5142)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:3755)
```
(using v4972r4)

**To Reproduce**
I don't know.

**Expected behavior**
No crash.

**Environment:**
- - [ ] All OS
- - [x] Windows
- - [ ] Linux
- - [ ] macOS 

Windows 11

Table.positionTooltip():
```java
				Rectangle adjustedTooltipBounds = getDisplay().fitRectangleBoundsIntoMonitorWithCursor(toolRect);
				OS.SetWindowPos(hwndToolTip, 0, adjustedTooltipBounds.x, adjustedTooltipBounds.y,
						adjustedTooltipBounds.width, adjustedTooltipBounds.height, flags);
```
Display.fitRectangleBoundsIntoMonitorWithCursor():
```java
	Rectangle monitorBounds = coordinateSystemMapper.getContainingMonitorBoundsInPixels(getCursorLocation());
	if (monitorBounds == null) {
		return null;
	}
```
can return null, so it needs to be handled by callers.

HeikoKlare:
Table and tree implementations contain a tooltip repositioning functionality to make a tooltip fit into a single monitor. The reason is that tooltips spanning multiple monitors of different zoom can freeze the UI thread on Windows.

While in the Tree implementation the repositioning was properly guarded with a null check for the adjusted position, this check is missing in the Table implementation can lead to a NullPointerException in case the calculated result is actually null. This change fixes that by adding an according null check, which effectively skips the tooltip repositioning in case the correct monitor to fit into could not be found.

Fixes https://github.com/eclipse-platform/eclipse.platform.swt/issues/3487

Originally introduced with: https://github.com/eclipse-platform/eclipse.platform.swt/pull/2241

### DIFF
--- a/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Table.java
+++ b/bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Table.java
@@ -7161,9 +7161,11 @@ private LRESULT positionTooltip(NMHDR hdr, long lParam) {
 				long hwndToolTip = OS.SendMessage(handle, OS.LVM_GETTOOLTIPS, 0, 0);
 				int flags = OS.SWP_NOACTIVATE | OS.SWP_NOZORDER;
 				Rectangle adjustedTooltipBounds = getDisplay().fitRectangleBoundsIntoMonitorWithCursor(toolRect);
-				OS.SetWindowPos(hwndToolTip, 0, adjustedTooltipBounds.x, adjustedTooltipBounds.y,
-						adjustedTooltipBounds.width, adjustedTooltipBounds.height, flags);
-				result = LRESULT.ONE;
+				if (adjustedTooltipBounds != null) {
+					OS.SetWindowPos(hwndToolTip, 0, adjustedTooltipBounds.x, adjustedTooltipBounds.y,
+							adjustedTooltipBounds.width, adjustedTooltipBounds.height, flags);
+					result = LRESULT.ONE;
+				}
 			} else if (isCustomToolTip()) {
 				RECT itemRect = getItemBounds(pinfo, item, hDC);
 				NMTTDISPINFO lpnmtdi = new NMTTDISPINFO();


r-mennig (approved):
The change looks good to me. `org.eclipse.swt.widgets.Table.positionTooltip(NMHDR, long)` now correctly checks if the computed `adjustedTooltipBounds` value is `null`, bubbling up a `null` return value which is handled in `wmNotify(NMHDR, long, long)`.

bundles/org.eclipse.swt/Eclipse SWT/win32/org/eclipse/swt/widgets/Table.java
r-mennig, line 7134: General question: Why does this method return a `LRESULT` which is not the result of a Win-API call?

`Table.positionTooltip(NMHDR, long)` reports one bit of information via its return value: Whether a (re-)positioning occurred (currently `LRESULT.ONE`) or not (currently `null`). 

I think using a `boolean` return value here and converting the result to a `LRESULT` on the caller side (if needed) could help prevent future NPEs.

HeikoKlare: Good catch. The method was extract here:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/2241

We seem to have not further thought about the return type, but it sounds reasonable to have that method with boolean return type and conert it into a proper LRESULT at caller side.
Same applies to the same method in `Tree`, which is why I would keep that as an enhancement independent from this change. Would appreciate if you can create an according PR, @r-mennig

r-mennig: `org.eclipse.swt.widgets.Tree.positionTooltip(NMHDR, long, long, boolean)` returns the non-null result of a `org.eclipse.swt.widgets.Composite.wmNotify(NMHDR, long, long)` which is an `LRESULT` so returing a `LRESULT` for this method seems correct to me.

For `org.eclipse.swt.widgets.Table.positionTooltip(NMHDR, long)` I've created a PR here: https://github.com/eclipse-platform/eclipse.platform.swt/pull/3497

